Issue #3615, opened 2020-05-19
scan: simplifaction


oliver-sanders:
Simplify cylc scan:

* Re-implement removing the old --user junk.
* List stopped suites (based on kwarg).
* Get information from the FS rather than Scheduler when possible (potentially contriversial).

**Pull requests welcome!**

Close #2663

oliver-sanders:
Removed the "small" label as it made sense to go async at the same time making scan faster and moving io to async for better compatibility with the UI Server.

PR #3724, opened 2020-07-30
async scan

oliver-sanders:
Closes: #3615 
Addresses: https://github.com/cylc/cylc-uiserver/issues/96
Sibling to: cylc/cylc-uiserver#150, https://github.com/cylc/cylc-doc/pull/138, https://github.com/conda-forge/cylc-flow-feedstock/pull/10.

Re-write `cylc scan` and it's underlying logic for simplicity, performance and for better incorporation with Cylc UI Server and TUI.

**This is a precursor to:**

* Showing stopped flows in the UI Server - https://github.com/cylc/cylc-uiserver/pull/150.
* Listing flows in TUI.

**Highlights:**

* Removes multi-user & port-scanning logic.
* Separates out filtering, contact file reading and network information requests into separate functions.
* Makes it super easy to roll-you-own scan client by moving logic into reusable components.
* Converts from the legacy REST endpoints to GraphQL.
* Scanning of all workflows in any state under one interface.
* Performs directory listing asynchronously (which can be very slow on network filesystems).
* Interlaces different stages of the scan operation for performance (e.g. can perform directory listings, whilst contacting flows).
* Reduction in worst-case scan times.
* The whole scan logic is now asynchronous.
* Unify `cylc scan` and `cylc print`.

**Async Pipes:**

I've had a pet-project for a little while now of constructing asynchronous pipes in Python so this was a nice chance to use it.

Construct a pipe using `|` syntax like on the command line:

```python
pipe = scan | is_active(True) | contact_info
```

The pipe is an async generator, iterate through it like so:

```python
async for flow in pipe:
    # ...
```

Once items start coming out of `scan` they get pushed into the next stage of the pipe, in theory each stage of the pipe could be running over different items simultaneously.

Sounds complicated but it actually only takes a few lines of Python to make this happen, most of `cylc.flow.async_util` is just documentation.

Here's your minimal scan client under the new system:

```python
import asyncio
 
from cylc.flow.suite_files import ContactFileFields as CFF 
from cylc.flow.network.scan_nt import (
    scan,
    is_active,
    contact_info,
)                                                                                    
 
 
async def scanner():
    pipe = scan | is_active(True) | contact_info
    async for flow in pipe:
        print(f'{flow["name"]} {flow[CFF.HOST]}')
 
 
asyncio.run(scanner())
```

**File System Efficiency:**

In order to perform a scan we need to perform, potentially a very large number of directory listings.

* One for the `cylc-run` directory.
* One for each `cylc-run/*` directory (this is recursive!).
* One for the `.service` dir of each registered flow.

Before all of these directory listings happen one after the other, they are all blocking not only to each other but also to the other operations involved in the scan (e.g. contacting the flows themselves).

Now it is possible for listings to happen simultaneously thanks to a bit of `libuv` magic. This is the library behind node, there are a few lines of code required to marry up the `libuv` event loop to `asyncio`, after that it's plain sailing.

**Scan Client:**

The scan client is now pretty simple, I've knocked together a basic interface though this can be easily changed.

Screenshot 2020-07-30 at 16 27 42

* `cylc scan` with no opts or args still as compatible with Cylc7 as it can be.
* New `name` and `tree` formats replace `cylc print`.
* Can now display stopped workflows.
* Can now filter by workflow state (running, held, stopping, stopped).
* New `rich` format.
* Only contacts workflows when actually necessary (with no opts, no need to do so).
* Uses colour but has `--colour-blind` option which still uses colour (`--color=never` to disable completely) but ensures that colour isn't required to interpret the display.

**TODO:**

* [x] Remove the old scan interface  - **done**.
* [x] Fix any tests broken by syntax changes - **wrote new integration tests**.
* [x] Consider adding a functional test for the new scan client - **done**.
* [x] Fix tests/f/broadcast/00-simple.t

**Questions:**

* [x] What back-compatibility do we require to the Cylc7 `cylc scan` interface? **None**
* [x] Do we want to replace `cylc scan -t raw`? **No**
* [x] Do we need to back-support the old `cylc scan -t json` interface? **No**
* [x] Ok to kill `cylc print`, the only thing print does which can doesn't is to list the reg dir (which scan could do if desired)?

**Requirements check-list**
- [x] Documented in https://github.com/cylc/cylc-doc/pull/138
- [x] Created an issue at https://github.com/conda-forge/cylc-flow-feedstock/pull/10


hjoliver:
> The standard format is a bit dull, for compatibility with Cylc7.

Does it need to be compatible?

oliver-sanders:
> Does it need to be compatible?

That is the question, people may well be scraping `cylc scan`, though given the interface breakages in other commands do we need to preserve this interface?

TomekTrzeciak:
> > Does it need to be compatible?
> 
> That is the question, people may well be scraping `cylc scan`, though given the interface breakages in other commands do we need to preserve this interface?

Is there some machine-readable output format, say JSON? Might be better to provide something like that rather than trying to maintain backward compatibility for human-intended output.

oliver-sanders:
Yes, both the new and old scan clients provide a JSON format, however, I've broken that too:

```console
$ cylc scan -t json
[
    [
        "generic",
        "olivers-macbook-pro.local",
        "43098",
        "43094",
        "5",
        {
            "name": "generic",
            "owner": "oliver",
            "version": "8.0a2"
        }
    ]
]
$ cylc scan-nt -t json
[
    {
        "name": "generic",
        "path": "/Users/oliver/cylc-run/generic",
        "contact": "/Users/oliver/cylc-run/generic/.service/contact",
        "CYLC_API": "5",
        "CYLC_SSH_USE_LOGIN_SHELL": "True",
        "CYLC_SUITE_HOST": "olivers-macbook-pro.local",
        "CYLC_SUITE_NAME": "generic",
        "CYLC_SUITE_OWNER": "oliver",
        "CYLC_SUITE_PORT": "43098",
        "CYLC_SUITE_PROCESS": "54860 /Users/oliver/miniconda3/envs/cylc8/bin/python3.7 /Users/oliver/miniconda3/envs/cylc8/bin/cylc-run generic",
        "CYLC_SUITE_PUBLISH_PORT": "43094",
        "CYLC_SUITE_RUN_DIR_ON_SUITE_HOST": "/Users/oliver/cylc-run/generic",
        "CYLC_SUITE_UUID": "33000c07-565e-4d93-84f4-39d6f315bf31",
        "CYLC_VERSION": "8.0a2"
    }
]
```

It would be possible to re-instate the old JSON format `[name: str, host: str, port: str, api_version: str, data: dict]`.

Note that the addition of the publisher port to this output has already broken JSON scan cylc7/8 compatibility.

TomekTrzeciak:
> Yes, both the new and old scan clients provide a JSON format, however, I've broken that too
> ...
> It would be possible to re-instate the old JSON format `[name: str, host: str, port: str, api_version: str, data: dict]`.

The new format is much nicer though - much easier to `grep`, trivial to turn it into shell variables with a bit of `sed` or `awk`.

My 2 cents - you're allowed to break compatibility at major versions. That kind of opportunity doesn't happen very often, so make the most of it.

(PS. you could use `json.dump(..., sort_keys=True)` for even more stable output).

oliver-sanders:
Sorted the keys, good suggestion. I'd rather not go back to the old JSON format as its less explicit and more fragile.

The other format likely to be used by scrapers is "raw"

```console
$ cylc scan -t raw
generic|oliver|olivers-macbook-pro.local|port|43098
```

I'd rather ditch it but it's trivial to re-implement.

hjoliver:
I don't mind breaking the old `cylc scan` output format (if you could call it that).  Scraping that is a pretty niche application I think, unlikely to be done by the average user.  Maybe by very few ops-type people who want to automate (via scripts) interactions with many suites?  And not too hard to adapt such scripts to the new improved output anyway.

tests/functional/cylc-subscribe/01-subscribe.t
```diff
@@ -38,21 +38,12 @@ run_ok "${TEST_NAME}" cylc subscribe --once --topics="workflow" "${SUITE_NAME}"
 grep_ok "running" "${TEST_NAME}.stdout"
 
 TEST_NAME="${TEST_NAME_BASE}-subscribe-2"
-# Need fields from "cylc scan", cannot quote
-# shellcheck disable=SC2046
-run_ok "${TEST_NAME}" cylc subscribe --once --topics="workflow" \
-    $(cylc scan --color=never --publisher -n "${SUITE_NAME}" \
-        | awk -F' ' -v RS='\n([ \t]*\n)+' '{print $1 " " $3}')
```
oliver-sanders, line 45: Didn't need to provide host:port via `cylc scan`, the internal logic does this for us.

cylc/flow/network/scan.py
```diff
@@ -13,298 +13,361 @@
...
+    return (
+        SERVICE.name in listing
+        or SUITERC.name in listing  # cylc7 flow definition file name
+        or 'flow.cylc' in listing  # cylc8 flow definition file name
+        or 'log' in listing
+    )
+
+
+@pipe
+async def scan(run_dir=None, scan_dir=None, max_depth=4):
```
oliver-sanders, line 105: The `max_depth` is a new feature to avoid going too deep down the rabbit warren in the event that:

* Someone puts something which isn't a workflow into their cylc-run dir (a not entirely uncommon occurrence).
* Incase the way we recognise flows changes (e.g. the suite.rc file gets renamed or converted to a Python file).

What is a sensible default value for this?

Set as 4 for now as that's what's required by cylc functional test battery.

oliver-sanders:
I've managed to do something to anger `tests/f/broadcast/00-simple.t` which is notoriously hard to debug but that aside should be ready for review.

The details of the new `cylc scan` client are all up for debate/input, it's only displaying a small number of fields at the moment but it's trivial to expand that.

cylc/flow/hostuserutil.py
```diff
@@ -113,7 +113,7 @@ def _get_host_info(self, target=None):
         """Return the extended info of the current host."""
         if target not in self._host_exs:
             if target is None:
-                target = socket.getfqdn()
+                target = socket.gethostname()
```
datamel, line 116: whoops again...
Should we think about supporting mac use with cylc...
Could we do something like this
```suggestion
                import platform
                if platform.system() == 'Darwin':
                    target = socket.gethostname()
                else:
                    target = socket.getfqdn()
                   
```

oliver-sanders: > Could we do something like this

No.

oliver-sanders: Fixed, been a while since I last did it!

oliver-sanders:
Deconflicted.

cylc/flow/print_tree.py
```diff
@@ -62,18 +64,28 @@ def print_tree(tree, padding, use_unicode=False, prefix='', labels=None,
...
-                    print(line)
+                    ret.append(line)
             else:
                 if eq:
                     joiner = '= '
                 else:
                     joiner = ''
-                print(line + joiner + str(tree[item]))
+                ret.append(f'{line}{joiner}{tree[item]}')
+    return ret
```
datamel, line 91: https://github.com/cylc/cylc-flow/blob/c56a1b825c63b89196872e342500f43eb7b6b9bb/cylc/flow/config.py#L1398
https://github.com/cylc/cylc-flow/blob/c56a1b825c63b89196872e342500f43eb7b6b9bb/cylc/flow/scripts/cylc_broadcast.py#L271
 ^^^ These calls won't do anything now, I think they were expecting a printed tree rather than a built tree. I see you have fixed the call on scripts/cylc-scan.py `  ret = print_tree(tree, '', sort=False, use_unicode=True)` . Does the same fix have to happen in the above files?

cylc/flow/print_tree.py
```diff
@@ -33,7 +33,8 @@
 
 
 def print_tree(tree, padding, use_unicode=False, prefix='', labels=None,
```
datamel, line 35: Would this be better named now as `build_tree`, to properly reflect what it does now?

cylc/flow/network/client.py
```diff
@@ -62,6 +62,18 @@ class SuiteRuntimeClient(ZMQSocketBase):
             Set the default timeout in seconds. The default is
             ``ZMQClient.DEFAULT_TIMEOUT``.
             Note the default timeout can be overridden for individual requests.
+        host (str):
+            The host where the flow is running if known.
+
+            If both host and port are provided it is not necessary to load
+            the contact file.
+        host (str):
+            The port the flow REQ-REP TCP server is listing on if known.
```
datamel, line 71:
```suggestion
            The port the flow REQ-REP TCP server is listening on if known.
```

cylc/flow/print_tree.py
oliver-sanders: Good spot, I hadn't realised there were other uses, this probably explains the broadcast failure!

cylc/flow/network/scan.py
```diff
@@ -13,302 +13,361 @@
...
+SERVICE = Path(SuiteFiles.Service.DIRNAME)
+CONTACT = Path(SuiteFiles.Service.CONTACT)
+SUITERC = Path(SuiteFiles.SUITE_RC)
 
-def async_map(coroutine, iterator):
-    """Map iterator iterator onto a coroutine.
 
-    * Yields results in order as and when they are ready.
-    * Slow workers can block.
+async def dir_is_flow(listing):
```
datamel, line 80: Would this not be considered CPU-bound not IO-bound? I could be completely wrong here but I thought asyncio was for IO bound and processes were for CPU-bound?
So, if we wait for an IO operation it is some other machine/database doing the processing and we can free up our thread to use elsewhere while we wait. But if we wait for this method we're still burning our CPU anyway but also have the expense of setting up additional async. 
I don't think there is a bottleneck here that can be freed by adding async. 
I really don't feel super-confident on this yet, so might be making mistakes - sorry if that's the case! 
https://leimao.github.io/blog/Python-Concurrency-High-Level/

cylc/flow/scripts/cylc_scan.py
```diff
@@ -60,267 +159,356 @@ def get_option_parser():
...
-    if options.color in ['auto', 'always'] and terminal.supports_color():
-        options.color = True
+def _format_plain(flow, _):
+    """A single line format of the form: <name> [<host>:<port>]"""
+    if flow.get('contact'):
+        return f'<b>{flow["name"]}</b> {flow[Cont.HOST]}:{flow[Cont.PORT]}'
     else:
-        options.color = False
-    color_init(autoreset=True, strip=not options.color)
+        return f'<dim><b>{flow["name"]}</b></dim>'
```
datamel, line 267: Is this a html tag? Should it be div?
```suggestion
        return f'<div><b>{flow["name"]}</b></div>'
```

oliver-sanders: This is an `ansimarkup` tag, `dim` is term speak for "diminished".

cylc/flow/network/scan.py
oliver-sanders: Er, yeah, there is no IO in this function so the `async` is not necessary (there was earlier on in development). I think the `async` here is pretty harmless but I'll remove it anyway.

Note that the async transformer/filter functions which follow use `async` whether they do IO or not, simply because the interface is `async` (if one of them is `async` they all have to be), so the unnecessary `async def`s there should be ignored.

cylc/flow/print_tree.py
hjoliver: The clue was in the name!  `print_tree`

cylc/flow/scripts/cylc_scan.py
datamel: Ah good to know!

cylc/flow/scripts/cylc_scan.py
```diff
@@ -15,39 +15,138 @@
...
+List Cylc workflows, by default this shows only running or held workflows.
 
-Print information about running suites.
+Examples:
+    # list all "active" workflows (i.e. running or held)
+    $ cylc scan
 
-Use the -o/--suite-owner option to get information of running suites for other
-users.
+    # show more information about these workfows
```
datamel, line 26:
```suggestion
    # show more information about these workflows
```

datamel:
Looks fab in the screen grabs above. Unfortunately, I am not able to get this working as there is an un-handled exception happening for me. Looking at the suites these are happening on for me - just old cit-test battery suites that appeared with the nfs problem. 
I suspect this will be a limited problem and for most normal users it will probs be fine but I thought I would flag it as it may be an easy fix and I am limited in my ability to review at the mo.  


image

oliver-sanders:
That's strange looks like you don't have permission to delete a file you created. Have you `chmod`ed anything in your `cylc-run` directory? I think if you clean out your `cylc-run` dir the problem will probably go away.

Shouldn't happen normally but should probably put a try/except around that `Path.exists` call anyway.

cylc/flow/async_util.py
```diff
@@ -0,0 +1,282 @@
+import asyncio
```
dwsutherland, line 1: Add Cylc heading info to this file?

dwsutherland (approved):
I don't fully understand the piping code yet (just the different parts), but functionally this works great 👌 

Few minor comments..

cylc/flow/scripts/cylc_scan.py
```diff
@@ -59,267 +158,356 @@ def get_option_parser():
...
+    # show_active = bool({'running', 'held'} & opts.states)
+    show_inactive = bool({'stopped'} & opts.states)
+
+    # filter by flow name
+    if opts.name:
+        pipe |= filter_name(*opts.name)
+
+    # filter by flow state  -  TODO: reduce this
+    if show_active and show_inactive:
+        pipe |= is_active(True, filter_stop=False)
```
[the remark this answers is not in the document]
oliver-sanders: It's a combo of `show_active` and `show_inactive` and if `not show_active and not show_inactive` the `is_active` step is skipped entirely.

cylc/flow/network/scan.py
oliver-sanders: I think probably too rare to care, or at least the default should be sufficient that we wouldn't expect users to hit it.

oliver-sanders:
Rebased and squashed a `suite.rc` reference.

cylc/flow/scripts/cylc_scan.py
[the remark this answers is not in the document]
dwsutherland:
```
    if show_active and show_inactive:
        pipe |= is_active(True, filter_stop=False)
    elif show_active:
        pipe |= is_active(True)
    elif show_inactive:
        pipe |= is_active(False)
```
Reduce to:
```
    if show_active:
        pipe |= is_active(True, filter_stop=(not show_inactive))
    elif show_inactive:
        pipe |= is_active(False)
```
Since the default for `filter_stop` must be `True`

oliver-sanders:
@datamel the last commit (async_util: skip errored items in async pipes) 092fef8 allows async pipes to keep going in the even of errors. The error gets logged as a warning. This should allow more graceful handling of the error you reported.

Note that "expected" errors should already be handled in the `cylc.flow.network.scan` function (e.g. `SuiteStopped`).

datamel (approved):
Great change. Comments I had raised before have been addressed so I am happy to approve.

cylc/flow/network/scan.py
```diff
@@ -13,302 +13,361 @@
...
-    load_contact_file,
     get_suite_title,
-    get_suite_source_dir
+    load_contact_file_async
 )
 
-DEBUG_DELIM = '\n' + ' ' * 4
-INACTIVITY_TIMEOUT = 10.0
-MSG_QUIT = "QUIT"
-MSG_TIMEOUT = "TIMEOUT"
```
kinow, line 43: This breaks the UIS `master`.

```bash
[I 2020-08-26T11:06:18.893 JupyterHub spawner:1417] Spawning cylc-uiserver --port=57193 -s /home/kinow/Development/python/workspace/cylc-uiserver/../cylc-ui/dist
Failed to set groups [Errno 1] Operation not permitted
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/bin/cylc-uiserver", line 11, in <module>
    load_entry_point('cylc-uiserver', 'console_scripts', 'cylc-uiserver')()
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2852, in load_entry_point
    return ep.load()
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2443, in load
    return self.resolve()
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2449, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/home/kinow/Development/python/workspace/cylc-uiserver/cylc/uiserver/main.py", line 38, in <module>
    from .data_store_mgr import DataStoreMgr
  File "/home/kinow/Development/python/workspace/cylc-uiserver/cylc/uiserver/data_store_mgr.py", line 43, in <module>
    from cylc.flow.network.scan import MSG_TIMEOUT
ImportError: cannot import name 'MSG_TIMEOUT' from 'cylc.flow.network.scan' (/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/network/scan.py)
```

Tested with UIS `master` and sibling PR, it failed to launch the UIS in both cases :cry:

cylc/flow/network/scan.py
```diff
@@ -13,302 +13,361 @@
...
-
-    Walk users' "~/cylc-run/" to get (host, port) from ".service/contact" for
-    active, or all (active plus registered but dormant), suites.
-    Note suites run in cylc 7 or less will only be returned as LOG.info
-    statement.
+    Args:
+        flow (dict):
+            Flow information dictionary, provided by scan through the pipe.
+        fields:
+            Iterable containing the fields to request e.g::
```
kinow, line 299: Missing new parameter `filters`

cylc/flow/network/scan.py
```diff
@@ -13,302 +13,361 @@
...
 
-    Example:
-        >>> async def square(number): return number ** 2
-        >>> generator = async_map(square, ((i,) for i in range(5)))
-        >>> list(generator)
-        [0, 1, 4, 9, 16]
+    Args:
+        flow (dict):
+            Flow information dictionary, provided by scan through the pipe.
+        pattern (str):
```
kinow, line 167: s/`str`/`re.Pattern`? IDE complaining about calling `.match` on a `str` type

cylc/flow/async_util.py
```diff
@@ -0,0 +1,306 @@
...
+def pipe(func=None, preproc=None):
+    """An asynchronous pipe implementation in pure Python.
+
+    Use this to decorate async functions in order to arrange them into
+    asynchronous pipes. These pipes can process multiple items through multiple
+    stages of the pipe simultaneously by doing what processing it can whilst
+    waiting on IO to take place in the background.
+
+    Args:
+        preproc (callable):
```
kinow, line 186: Add `func (callable)` too?

kinow (commented):
Not sure if a blocker, but I couldn't launch UIS + UI with this. The UIS failed to spawn due to a missing import, that I think was removed in this PR? (see comment below)

No other blockers, few comments but nothing important.

Tested in the command line with no issues

```
(venv) kinow@ranma:~/Development/python/workspace/cylc-flow$ cylc scan --states=held,stopped
3698
aws
cat
complex
dog
families2
reflow1
```

@oliver-sanders when would I use the task states? I tried filtering by `states=submitted`. The error indicated me I could use only workflow states. But the `--help` shows also task states?

Running it with Python is a real pleasure! @oliver-sanders I adjusted the example in the issue description, replacing the `scan_nt` by `scan` (looked at `test_scan_nt` to find out which file to import).

image

dwsutherland:
> Not sure if a blocker, but I couldn't launch UIS + UI with this. The UIS failed to spawn due to a missing import, that I think was removed in this PR? (see comment in one of the files modified).

@kinow  - There's a sibling PR (https://github.com/cylc/cylc-uiserver/pull/150), and my review has some fixes to get things working.

cylc/flow/network/scan.py
dwsutherland: https://github.com/cylc/cylc-uiserver/pull/150

hjoliver (commented):
A couple of embarrassingly trivial comments.   The async pipes are cool :sunglasses:  although I'm struggling a bit to get my head around the advanced (by my standards at least!) use of asyncio. No need to let that hold us up though as it obviously works well :+1:  But I'll hold off merging today as Bruno has raised a few minor issues.

hjoliver:
(Still needs a change log entry).

hjoliver:
Weird log output with `-t rich` and `--states=stopped`:
```
cylc scan --states=stopped -t rich
Job State Key: ■ submitted ■ submit-failed ■ running ■ succeeded ■ failed

2020-08-26T17:12:05+12:00 WARNING - Workflow not running: complex
2020-08-26T17:12:05+12:00 WARNING - Workflow not running: foo
2020-08-26T17:12:05+12:00 WARNING - Workflow not running: rog
2020-08-26T17:12:05+12:00 WARNING - Workflow not running: ron
```

hjoliver:
@kinow - 
> when would I use the task states? I tried filtering by states=submitted. The error indicated me I could use only workflow states. But the --help shows also task states?

I guess that's just to document the task state totals visible in the rich output.  

But the command help for `--states` could say "...providing a list of **workflow** states..." to avoid this confusion, @oliver-sanders

kinow:
> @kinow -
> 
> > when would I use the task states? I tried filtering by states=submitted. The error indicated me I could use only workflow states. But the --help shows also task states?
> 
> I guess that's just to document the task state totals visible in the rich output.
> 
> But the command help for `--states` could say "...providing a list of **workflow** states..." to avoid this confusion, @oliver-sanders

Ah, that makes sense! Thanks @hjoliver !

cylc/flow/network/scan.py
oliver-sanders: I need to update that PR, at the moment there will be some issues.

cylc/flow/network/scan.py
oliver-sanders: Good spotting.

oliver-sanders:
> Weird log output with -t rich and --states=stopped:

Fixed with the following diff.

```diff
diff --git a/cylc/flow/scripts/cylc_scan.py b/cylc/flow/scripts/cylc_scan.py
index d5de14e64..14e66b60c 100644
--- a/cylc/flow/scripts/cylc_scan.py
+++ b/cylc/flow/scripts/cylc_scan.py
@@ -417,7 +417,7 @@ def get_pipe(opts, formatter, scan_dir=None):
         graphql_fields.update(RICH_FIELDS)
 
     # add graphql queries / filters to the pipe
-    if graphql_fields:
+    if show_active and graphql_fields:
         pipe |= graphql_query(graphql_fields, filters=graphql_filters)
 
     return pipe
```

oliver-sanders:
> The async pipes are cool 😎 although I'm struggling a bit to get my head around the advanced (by my standards at least!) use of asyncio.

The async pipe wrapper makes it look more advanced than it actually is, the crux of the matter is async generator syntax like this

```python
async for item in async_generator:
    # yield items from async_generator without blocking
    pass
```

All i've done is to nest these async generators.

```
async for item in async_generator:
    async for item in async_generator(item):
        pass
```

It's not very advanced, what happens when the first item is yielded? Does the second generator start processing the first item or does the first generator try loading the second item? The prioritisation is presumably in the implementation of the async library. After a bit of playing I found that it can go either way.

cylc/flow/network/scan.py
oliver-sanders: Oh, actually, I'm not sure about that.

The function accepts `re.Pattern` objects, however, the user passes `str` objects because the `pipe` decorator pre-processes the arguments.

oliver-sanders:
Rebased and feedback addressed.

dwsutherland:
Looks good to me!
image

hjoliver:
I made an example that seems to conform to your documented async pipe usage, and it doesn't seem to work as advertised (or how I'd expect) in terms of concurrency:

A generator yields successive letters of a word, which get transformed by subsequent functions the pipe.

```
bong =  speak("hello") | louder() | longer()
```

The `louder()` function uppercases the letter, and the `longer()` function appends an underscore.

```python
#!/usr/bin/env python3

from cylc.flow.async_util import pipe
import asyncio

@pipe
async def speak(word):
    print(f'SPEAK {word}')
    for c in word:
        print(f'(spork {c})')
        yield c
    print(f'SPOKE {word}')

@pipe
async def louder(c):
    print(f' > upper {c} (sleep...)')
    await asyncio.sleep(2)
    d = c.upper()
    print(f' < upper {c} -> {d}')
    return d

@pipe
async def longer(c):
    print(f' > longer {c} (sleep...)')
    await asyncio.sleep(2)
    d = c + '_'
    print(f' < longer {c} -> {d}')
    return d

bong = speak("hello") | louder() | longer()

async def consumer(pipe):
    res = ""
    async for item in pipe:
        print(f'(consumed {item})')
        res += item
    print(f'HEARD {res}')

asyncio.run(consumer(bong))
```

If this was a classic pipeline (ignoring the letter transformations here for clarity) I'd expect the following sequence of events:
- `speak` yields `h`
- `louder(h)` runs while `speak` is yielding 'e`
- `longer(h)` runs while `louder(e)` runs and `speak` is yielding `l`
- (and so on)

For a 5 letter word and 2 second async sleeps in `louder()` and `longer()` total processing time should be 12 seconds (`2 + 5x2`).

But running this, each letter gets processed through the whole pipeline sequentially (i.e. `h` comes out the end before `e` goes in the front, and so on), total processing time 20 seconds (`5 * (2 + 2)`) 

```
time ./foo.py 
SPEAK hello
(spork h)
 > upper h (sleep...)
 < upper h -> H
 > longer H (sleep...)
 < longer H -> H_
(consumed H_)
(spork e)
 > upper e (sleep...)
 < upper e -> E
 > longer E (sleep...)
 < longer E -> E_
(consumed E_)
(spork l)
 > upper l (sleep...)
 < upper l -> L
 > longer L (sleep...)
 < longer L -> L_
(consumed L_)
(spork l)
 > upper l (sleep...)
 < upper l -> L
 > longer L (sleep...)
 < longer L -> L_
(consumed L_)
(spork o)
 > upper o (sleep...)
 < upper o -> O
 > longer O (sleep...)
 < longer O -> O_
(consumed O_)
SPOKE hello
HEARD H_E_L_L_O_

real    0m20.124s
user    0m0.091s
sys     0m0.012s
```

So, have I done something wrong here?

hjoliver:
Wondering if implementing what I've described would require queues; but a comment in the code seems to gel with my expectation:
> These pipes can process multiple items through multiple stages of the pipe simultaneously by doing what processing it can whilst waiting on IO to take place in the background.

hjoliver:
> All i've done is to nest these async generators.
>
> ```python
> async for item in async_generator:
>     async for item in async_generator(item):
>        pass
>```

Same result here, although in this case it's easier to see what should be the expected result:
```
#!/usr/bin/env python3

import asyncio

async def gen1(sentence):
   """yield words of sentence"""
   for word in sentence.split():
      yield word
      await asyncio.sleep(2)

async def gen2(word):
   """yield letters of word"""
   for letter in word:
      yield letter
      await asyncio.sleep(2)

async def main():
   """nested async generators"""
   async for word in gen1("the quick brown"):
      print(word)
      space = ''
      async for letter in gen2(word):
         print(f'{space}{letter}')
         space += ' '

asyncio.run(main())
```
Result, no concurrency:
```
./foo.py
the
t
 h
  e
quick
q
 u
  i
   c
    k
brown
b
 r
  o
   w
    n
```

I'm sure I haven't fully grokked asyncio yet, but it seems to me we should not expect any concurrency over loop iterations here? The "async generator" syntax allows us to `await` stuff inside the `for` loop, and I've used `await asyncio.sleep(2)`, so that any **other** coroutines could slot in there, but the loop iterations themselves (included nested loops) are still gonna be sequential.

oliver-sanders:
Yeah, you may have a point there, think I may have jiggered it with a botched simplification.

Working on a slightly more advanced queue-based system which would provide scope for prioritisation, fix should be ready soon.

oliver-sanders:
Ok, moved to a queue-based system which allows the pipe to follow paths which aren't depth-first.

Here's an example which inserts random sleeps into each stage of the pipe:

```python
import asyncio
from random import random

from cylc.flow.async_util import pipe


@pipe
async def a_range(x):
    for y in range(x):
        await asyncio.sleep(random())
        print(f'# a_range -> {y}')
        yield y


@pipe
async def a_double(x):
    print(f'# a_double({x})')
    await asyncio.sleep(random())
    return x ** 2


@pipe
async def a_str(x):
    print(f'# a_str({x})')
    await asyncio.sleep(random())
    return str(x)


async def consumer():
    pipe = a_range(5) | a_double | a_str
    async for item in pipe:
        print(
            f'=> {repr(item)}'
        )


asyncio.run(consumer())
```

If it's concurrent you should get a different order of events every time with results not necessarily yielded in order.

**run 1:**

```
# a_range -> 0
# a_double(0)
# a_str(0)
# a_range -> 1
# a_double(1)
# a_str(1)
=> '0'
# a_range -> 2
# a_double(2)
# a_str(4)
=> '1'
# a_range -> 3
# a_double(3)
# a_range -> 4
# a_double(4)
=> '4'
# a_str(9)
=> '9'
# a_str(16)
=> '16'
```

**run 2:**

```
# a_range -> 0
# a_double(0)
# a_str(0)
# a_range -> 1
# a_double(1)
=> '0'
# a_range -> 2
# a_double(2)
# a_str(1)
# a_range -> 3
# a_double(3)
# a_str(4)
=> '1'
=> '4'
# a_range -> 4
# a_double(4)
# a_str(16)
=> '16'
# a_str(9)
=> '9'
```

hjoliver:
Well that certainly sped things up; my "speak" example now runs in 4 seconds instead of 20.

However, your new implementation is what I referred to in Element chat yesterday to as the "extreme concurrency" model.  This is more concurrency than a pipeline is supposed to have, which I think is fine (and actually desirable) for the scan application, but it kind of calls into question the **pipeline** name and special pipeline syntax doesn't it? Because it doesn't preserve order coming out the end of the pipe.

With random sleeps in my example, I get (e.g.):
```
SPEAK hello
...
HEARD  L_E_H_O_L_   # should be H_E_L_L_O_ 
```

To belabor the point, in a shell pipeline I'd expect the file `bar` to be identical to `foo` here:
```bash
cat foo | tr a B | tr B a > bar
```
but your pipe would reorder the lines if some of them took much longer to "translate" than others.

hjoliver:
In Cylc graph notation, here's a pipeline `A | B | C` for processing 3 items:
```
R3/^/P1 = """
   A => B => C
   A[-P1] => A
   B[-P1] => B
   C[-P1] => C"""
```
so C.3, B.2, and A.1 (e.g.) can run concurrently (they're at different points along the one pipe) but A.1, A.2, A.3 can't (that would be 3 concurrent pipes).

But you've done this, which is 3 pipes:
```
R3/^/P1 = "A => B => C"
```

hjoliver:
Maybe we could support both modes, to justify the cool syntax?  Or both separately, with a slightly different name and syntax for the multiple parallel pipes case?

dwsutherland:
> Because it doesn't preserve order coming out the end of the pipe.

I think that's find if your printing just collects whichever is done (reached the end of the pipe) from a queue, and prints one at a time (right?).. Then if we want to order the result, we wait for all concurrency to finish and order the results (which would still be faster) (?)

hjoliver:
But how do you know what the order should be at the other end @dwsutherland ?

dwsutherland:
> But how do you know what the order should be at the other end @dwsutherland ?

That's optional, user determined (or default). (?)

hjoliver:
(Note I'm not saying we don't want this maximum concurrency for `cylc scan`, I'm just arguing about whether or not we can call it a "pipe" and use pipe syntax to do it, because classic pipelines - like shell command pipes for example - don't behave like that).

dwsutherland:
> (Note I'm not saying we don't want this maximum concurrency for `cylc scan`, I'm just arguing about whether or not we can call it a "pipe" and use pipe syntax to do it, because classic pipelines - like shell command pipes for example - don't behave like that).

I guess I need to understand it more, I thought the pipe bit was sequential, and the concurrent bit was the multiple pushes through the pipe

hjoliver:
See my description of the Cylc graph pipeline implementation above.  A pipeline has a special kind of concurrency, like (literally) a single pipe with multiple processors spaced along its length.  That's why it's called a "pipe".  There's only one of each processor, and the data passes through the pipe sequentially, but once the pipe is full all of the processors are running concurrently (on different data elements).  That still more concurrency than passing one data element through the whole pipe at a time, which is what was happening on this branch yesterday.  But it's less concurrency than creating n pipes for n data elements.

hjoliver:
(Sorry @oliver-sanders - we should have this conversation on Element rather than cluttering up your PR!)

dwsutherland:
Latest push looks, appears functionally sound 👍 

And concurrent "pipes" working:
```
(oliver-flow) sutherlander@cortex-vbox:bin$ time ./foo.py 
SPEAK hello
(spork h)
(spork e)
(spork l)
(spork l)
(spork o)
SPOKE hello
 > upper h (sleep...)
 > upper e (sleep...)
 > upper l (sleep...)
 > upper l (sleep...)
 > upper o (sleep...)
 < upper h -> H
 > longer H (sleep...)
 < upper e -> E
 > longer E (sleep...)
 < upper l -> L
 > longer L (sleep...)
 < upper l -> L
 > longer L (sleep...)
 < upper o -> O
 > longer O (sleep...)
 < longer H -> H_
(consumed H_)
 < longer E -> E_
(consumed E_)
 < longer L -> L_
 < longer L -> L_
(consumed L_)
 < longer O -> O_
(consumed L_)
(consumed O_)
HEARD H_E_L_L_O_

real	0m4.079s
user	0m3.981s
sys	0m0.076s
```

cylc/flow/async_util.py
```diff
@@ -0,0 +1,340 @@
...
+            provided to a function when the pipe is created.
+
+            preproc(args: tuple, kwargs: dict) -> (args: tuple, kwargs: dict)
+
+    Example:
+        A generator to begin our pipe with:
+        >>> @pipe
+        ... async def arange():
+        ...    for i in range(10):
+        ...        yield i
```
hjoliver, line 231: > In your example the generator (speak) is not asynchronous as there is no await in it anywhere. 

I was following this example of yours in the code (although I admit to being confused about whether an async generator needs an await or not - I thought maybe `yield` passes control back to the event loop).

hjoliver:
> And concurrent "pipes" working: 

That's only with equal sleeps in my coroutines.  If you change to `asyncio.sleep(random())` the order will not be preserved.

dwsutherland:
> > And concurrent "pipes" working:
> 
> That's only with equal sleeps in my coroutines. If you change to `asyncio.sleep(random())` the order will not be preserved. Which as discussed (at length at this point!) is OK for `cylc scan` but it's not a "pipe".

Agreed. We don't need all the scanned workflows to collectively from something coherent (e.g. `H_E_L_L_O_`) .

cylc/flow/async_util.py
hjoliver: (OK I think I understand how that works now).

oliver-sanders:
>However, your new implementation is what I referred to in Element chat yesterday to as the "extreme concurrency" model.

Yes, and purposefully so.

> Because it doesn't preserve order coming out the end of the pipe.

It never proclaimed to, it's an async pipe.

It's possible to add a "preserve order" mode at some point, however, this isn't desirable for the `cylc scan` case (we have `cylc scan --sort` for those purposes anyway).

The main motives were:

* To move IO time into async-sleep time so that TUI/UIS/whatever can do useful things whilst the scan is happening. In my UIS PR the UIS starts opening connections to new flows whilst the scan it still doing its thing.
* To handle FS/network delays like, for example, NFS taking 5 seconds to read one flow's contact file or to list one directory. Concurrency protects us against worst-case scenarios and delays affect the minimum possible subset of flows.

cylc/flow/async_util.py
oliver-sanders: `yield` yields control to a generator in synchronous code, `await` yields control to a coroutine in asynchronous code. We have an async-generator here, both paradigms still apply but only in their own context.

oliver-sanders:
Ok three more commit which:

* Make async pipes ordered by default.
* Make filesystem scanning more concurrent (was going to do this in followup).

Note that order in async pipes isn't necessarily meaningful as async generators don't necessarily yield results in any kind of dependable order anyway (and in our case the scan generator doesn't).

hjoliver:
> It never proclaimed to, it's an async pipe.

It is async, for sure; I'm just arguing that it's not a pipe, it's a more like a bunch of pipes.

My classic pipe still has concurrency (albeit less of it), so I presume it could still have an async implementation and so still be called an "async pipe".

Of course if "async pipe" is actually a known thing and doesn't have classic pipe behavior then I guess I'm just wrong about the meaning of the term. 

But either there certainly are potential applications where preserved ordering would be expected, like line-by-line file processing.

> Make async pipes ordered by default.
> - This enables "classic" pipes which still benefit from the same concurrency as unordered pipes.

Great, even better :+1:

hjoliver:
I forgot to say up above, it's a great idea (especially with the pipe syntax) and I reckon you should consider making it available as a self-contained Python package on GitHub.  Then, as a generic library package, I'm sure people would want both kinds of behaviour for different applications.

hjoliver (approved):
Happy now :+1:  (Thanks for assuaging my pipe-OCD :grimacing: )

PR #3727, opened 2020-07-31
cyclers: fix ingest_time

oliver-sanders:
Sibling PR: https://github.com/cylc/cylc-doc/pull/141

Fix a new and interesting bug which effects non-gregorian date-time cycling modes, but only on certain days of the month, such as, conveniently, today, the 31st.

Spotted in this test, but may effect others:

```
tests/f/special/07-clock-triggered-360.t
```

The issue has surfaced now since the recent change to isodatetime which means that out-of-bounds date-times (e.g. the 31st of Feb) are rejected on sight. The `ingest_time` logic was attempting to load todays date into a different calendar which causes trouble, if say, it's the 31st of the month and you are using a 360 day calendar.

**Requirements check-list**
- [x] Appropriate tests added.
- [x] Docs PR - https://github.com/cylc/cylc-doc/pull/141
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/cycling/iso8601.py
+++ b/cylc/flow/cycling/iso8601.py
@@ -649,7 +649,6 @@ def ingest_time(value, now=None):
     # remove extraneous whitespace from cycle point
     value = value.replace(" ", "")
     parser = SuiteSpecifics.point_parser
-    timepoint = parser.parse(value)
 
     # integer point or old-style date-time cycle point format
     is_integer = bool(re.match(r"\d+$", value))
@@ -662,15 +661,25 @@ def ingest_time(value, now=None):
     is_prev_next = "next" in value or "previous" in value
     # offset from now (±P...)
     is_offset = value.startswith("P") or value.startswith("-P")
-    # missing date-time components off the front (e.g. 01T00)
-    is_truncated = timepoint.truncated
 
     if (
         is_integer
         or is_expanded
-        or not any((is_prev_next, is_offset, is_truncated))
     ):
         # we don't need to do any fancy processing
+        return value
+
+    # parse the timepoint if needed
+    if is_prev_next or is_offset:
+        # `value` isn't necessarily valid ISO8601
+        timepoint = None
+        is_truncated = None
+    else:
+        timepoint = parser.parse(value)
+        # missing date-time components off the front (e.g. 01T00)
+        is_truncated = timepoint.truncated
+
+    if not any((is_prev_next, is_offset, is_truncated)):
         return value
 
     if now is None:
--- a/cylc/flow/cycling/iso8601.py
+++ b/cylc/flow/cycling/iso8601.py
@@ -19,7 +19,7 @@
 from functools import lru_cache
 import re
 
-from metomi.isodatetime.data import Calendar, Duration
+from metomi.isodatetime.data import Calendar, Duration, CALENDAR
 from metomi.isodatetime.dumpers import TimePointDumper
 from metomi.isodatetime.timezone import (
     get_local_time_zone, get_local_time_zone_format, TimeZoneFormatMode)
@@ -29,7 +29,10 @@ from cylc.flow.cycling import (
     PointBase, IntervalBase, SequenceBase, ExclusionBase, cmp_to_rich, cmp
 )
 from cylc.flow.exceptions import (
-    SequenceDegenerateError, PointParsingError, IntervalParsingError
+    CylcConfigError,
+    IntervalParsingError,
+    PointParsingError,
+    SequenceDegenerateError
 )
 from cylc.flow.wallclock import get_current_time_string
 from cylc.flow.parsec.validate import IllegalValueError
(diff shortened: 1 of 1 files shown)


cylc/flow/cycling/iso8601.py
```diff
@@ -633,125 +636,165 @@ def _get_old_anchor_step_recurrence(anchor, step, start_point):
...
+    is_prev_next = "next" in value or "previous" in value
+    # offset from now (±P...)
+    is_offset = value.startswith("P") or value.startswith("-P")
+    # missing date-time components off the front (e.g. 01T00)
+    is_truncated = timepoint.truncated
+
+    if (
+        is_integer
+        or is_expanded
+        or not any((is_prev_next, is_offset, is_truncated))
```
oliver-sanders, line 671: This is the line which fixes the bug, allowing the code to return earlier skipping the whole `now` logic (which isn't needed).

cylc/flow/cycling/iso8601.py
```diff
@@ -633,125 +636,165 @@ def _get_old_anchor_step_recurrence(anchor, step, start_point):
...
+    Returns
+        tuple - (cycle_point, offset)
+
+    """
+    # are we in gregorian mode (or some other eccentric calendar
+    if CALENDAR.mode != Calendar.MODE_GREGORIAN:
+        raise CylcConfigError(
+            'prev()/next() syntax must be used with integer or gregorian'
+            f' cycling modes ("{value}")'
+        )
```
oliver-sanders, line 732: This is the other logic change.

oliver-sanders:
Ok, not so pretty but should do the job. I've converted the unittests to pytest in a separate commit to make it easier to flush out the real changes. The unittests *should* catch this I think, let me know if you can think of any other cases worth testing.

cylc/flow/cycling/iso8601.py
```diff
@@ -647,111 +666,147 @@ def ingest_time(value, my_now=None):
...
+        # this should only be attempted if the calendar is in
+        # gregorian mode
+        if CALENDAR.mode != Calendar.MODE_GREGORIAN:
+            raise CylcConfigError(
+                f'The value "{value}" can not be calculated '
+                f'in {CALENDAR.mode}'
+            )
+
+        if now is None:
+            now = parser.parse(get_current_time_string())
```
MetRonnie, line 692: Following the reference for `get_current_time_string()` to `wallclock.py` - is there a reason we use the `datetime` library instead of isodatetime to get the current time? Wouldn't the problem be solved by using isodatetime with the `CALENDAR` mode?

oliver-sanders: An interesting idea, however, I'm not entirely sure it's actually possible to get the current date in the 360 day calendar, there being no real mapping between the two.

MetRonnie: So there is no defined "now" for the 360 day calendar (or it is only a relative calendar)? So when `metomi.isodatetime.data.get_timepoint_for_now()` returns `2021-04-28` that's completely arbitrary?

oliver-sanders: Not completely arbitrary but it is essentially counting days from the start of the year when the calendar resets. Since one day in the 360 day calendar lasts longer than one day in the gregorian calendar it doesn't really make sense to say that today is such-and-such a date in the 360 day calendar. The 360 day calendar is essentially the calendar of a fictional planet.

cylc/flow/cycling/iso8601.py
```diff
@@ -647,111 +666,147 @@ def ingest_time(value, my_now=None):
...
+
+    if is_prev_next or is_offset or is_truncated:
+        # we need information about the current time to process this
+        # this should only be attempted if the calendar is in
+        # gregorian mode
+        if CALENDAR.mode != Calendar.MODE_GREGORIAN:
+            raise CylcConfigError(
+                f'The value "{value}" can not be calculated '
+                f'in {CALENDAR.mode}'
+            )
```
MetRonnie, line 689: Is there an equivalent section for https://cylc.github.io/cylc-doc/7.9.1/singlehtml/index.html#scheduling-initial-cycle-point-initial-cycle-point-relative-to-current-time in the cylc8 docs, so that this can be noted there?

cylc/flow/cycling/iso8601.py
```diff
@@ -647,111 +666,147 @@ def ingest_time(value, my_now=None):
...
+     """
+    # break down cycle point into constituent parts.
+    direction, tmp = value.split("(")
+    tmp, offset = tmp.split(")")
 
     # correct for year in 'now' if year only,
     # or year and time, specified in input
     if re.search(r"\(-\d{2}[);T]", value):
-        my_now.year = my_now.year + 1
+        now.year = now.year + 1
```
MetRonnie, line 743: I know these lines are from before, but any idea what is meant by "correct for year"? Why is there a need for this correction?

oliver-sanders: No idea whatsoever, perhaps it's rounding up for the next/prev syntax?

cylc/flow/cycling/iso8601.py
oliver-sanders: I'll add a warning to the docs.

oliver-sanders: Documented in a fairly obscure location, these docs are up for review so the important thing for now it that it's in there https://github.com/cylc/cylc-doc/pull/141/files#diff-47b7a9f6c399cc90f13256a3a895c865R1022-R1027

datamel:
Just a change log conflict and then ready for merge. Interesting bug and I'm happy with fix, thanks @oliver-sanders!

oliver-sanders:
Rebased and fixed the broken tests in the second commit (I was to parsing the timepoint too early in an attempt to nicify the logic).

Issue #3711, opened 2020-07-26
Update security e-mails and point to Discourse


kinow:
**Describe exactly what you would like to see in an upcoming release**

@hjoliver going through old notes, I think @oliver-sanders pointed we still had Matthew's e-mail in our [security reporting document](https://github.com/cylc/cylc-flow/blob/ea6ecca0943dac215263e0257de60947e01d0f27/SECURITY.md).

I don't know what's the new e-mail, so thought it was best to create an issue so we don't forget it (as a message to Element can get lost/forgotten too).

The link to Google Groups also needs to be updated to https://cylc.discourse.group/ :+1: 

Assigning to @hjoliver, but I guess anyone can chime in and fix it, if you know the correct new e-mail.

**Additional context**

**Pull requests welcome!**

PR #3728, opened 2020-08-03
Email swap on security doc.

hjoliver:
(Follow-up from last meeting). One review will do @oliver-sanders


kinow:
@hjoliver should we mark it as "closes #3711" ? There's also the Google Groups e-mail. Not sure if we have a replacement for that.

hjoliver:
> @hjoliver should we mark it as "closes #3711" ? There's also the Google Groups e-mail. Not sure if we have a replacement for that.

@kinow - thanks for spotting that.  I'll close #3711 in a minute with another PR to replace the Google Groups link...

PR #3733, opened 2020-08-04
tests/u: fix test_job_file

oliver-sanders:


tests/unit/test_job_file.py
```diff
@@ -100,7 +100,7 @@ def test_write(mocked_get_remote_suite_run_dir):
 
         assert (os.path.exists(local_job_file_path))
         size_of_file = os.stat(local_job_file_path).st_size
-        assert(size_of_file == 1841)
+        assert(size_of_file == 1845)
```
MetRonnie, line 103: Not sure what it's about but ok

Issue #3740, opened 2020-08-05
more restrictive auto-spawning?


hjoliver:
Under SoD, parent-less and absolute-parented tasks are successively auto-spawned.
```
R1 = "start-up"
P1 = """
   start-up[^] => foo => bar
   woo => baz
        """
```   
Here, `foo.1` will be spawned by `start-up[^]`, then `foo.2` will be auto-spawned when `foo.1` drops out of the runahead pool, and so on.

And, `woo.1` will be auto-spawned at start-up, then `woo.2` will be auto-spawned when `woo.1` drops out of the runahead pool, and so on.

This (or something similar) is unavoidable because we can't feasibly spawn all (potentially infinite) such tasks at once in future cycle points. 

Currently this is done for any absolute-parented task.  But if such a task also has other normal parents we could just wait for the normal parents to spawn it:
```
R1 = "start-up"
P1 = """
   start-up[^] & other => foo  # no need to auto-spawn foo?
        """
```   

But, not sure it matters much.

TomekTrzeciak:
Will this work in SOD?
```
R3//P1 = "foo => bar"
P1 = "bar"
```

hjoliver:
> Will this work in SOD?

It should!  But well done, you've identified a bug :bouquet:   A one-line fix coming soon :relieved:

hjoliver:
Pretty sure this doesn't matter

PR #3756, opened 2020-08-06
Fix sequence bounds check for parent tasks.

hjoliver:
This change addresses a bug (and untested functionality, evidently) found by @TomekTrzeciak in https://github.com/cylc/cylc-flow/issues/3740#issuecomment-669130391

In the following example, `bar` should trigger off of `foo` in the first 3 cycles, then subsequent `bar`'s should trigger automatically (no parents):
```
R3//P1 = "foo => bar"
P1 = "bar"
```
On this branch it works, but on master the scheduler shuts down finished after cycle point 3.

Since spawn-on-demand we auto-spawn parent-less tasks when their previous instance leaves the runahead pool (they have no parents to spawn them "on demand").  In this example, `bar.3` has a parent (`foo.3`) but `bar.4` has no parents and so has to be auto-spawned.  To do this, when `bar.3` leaves the runahead pool we look at the next cycle point for `bar` and see if it has parents there by looking for parents in each of its sequences.  This parent-lookup gave the wrong result because `sequence.is_on_sequence(point)` disregards sequence bounds. The fix is to use `sequence.is_valid(point)` (which correctly says that `bar.4` has no parents).

Marking as Draft until I add a test.

**Requirements check-list**

- [x] No change log entry needed  (fixes a very recent never-released bug)

- [x] No documentation update required

- [x] No dependency changes.

### DIFF
--- a/cylc/flow/taskdef.py
+++ b/cylc/flow/taskdef.py
@@ -139,7 +139,7 @@ class TaskDef:
         """Return the cycle points of my parents, at point."""
         parent_points = set()
         for seq in self.sequences:
-            if not seq.is_on_sequence(point):
+            if not seq.is_valid(point):
                 continue
             if seq in self.dependencies:
                 # task has prereqs in this sequence
@@ -154,7 +154,7 @@ class TaskDef:
         """Return my absolute triggers, if any, at point."""
         abs_triggers = set()
         for seq in self.sequences:
-            if not seq.is_on_sequence(point):
+            if not seq.is_valid(point):
                 continue
             if seq in self.dependencies:
                 # task has prereqs in this sequence
--- a/tests/functional/spawn-on-demand/08-lost-parents.t
+++ b/tests/functional/spawn-on-demand/08-lost-parents.t
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+# THIS FILE IS PART OF THE CYLC SUITE ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+# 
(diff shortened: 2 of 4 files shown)


hjoliver:
Proof the fix works:
```
$ cylc cat-log foo | grep triggered
2020-08-06T15:59:47+12:00 INFO - [foo.1] -triggered off []
2020-08-06T15:59:48+12:00 INFO - [foo.2] -triggered off []
2020-08-06T16:00:01+12:00 INFO - [bar.1] -triggered off ['foo.1']
2020-08-06T16:00:01+12:00 INFO - [bar.2] -triggered off ['foo.2']
2020-08-06T16:00:02+12:00 INFO - [bar.3] -triggered off []
2020-08-06T16:00:02+12:00 INFO - [bar.4] -triggered off []
2020-08-06T16:00:03+12:00 INFO - [bar.5] -triggered off []
2020-08-06T16:00:15+12:00 INFO - [bar.6] -triggered off []
2020-08-06T16:00:16+12:00 INFO - [bar.7] -triggered off []
...
```

hjoliver:
(rebased)

hjoliver:
Minimal functional test added (fails on master, passes here).

oliver-sanders (approved):
Good, tested with the following:

```ini
[scheduling]
    initial cycle point = 1 
    final cycle point = 9 
    cycling mode = integer
    [[graph]]
        R3//P1 = a => b
        P1 = b 
        6/P1 = b => c
```

MetRonnie (approved):

Issue #3746, opened 2020-08-05
Event-driven datastore update


hjoliver:
Event-driven removal of finished tasks is not compatible with the way the datastore is updated by iterating the task pool once per main loop - post SoD merge we miss final status changes.

So we need to be able to update the data store one task state change at a time.

PR #3757, opened 2020-08-06
CLI migration to GraphQL endpoint

dwsutherland:
As a prerequisite to having the CLI interact with the UIS GraphQL endpoint via the HUB, this PR:
- [x] Migrates the CLI to the one GraphQL endpoint.
- [x] Removes the unused endpoints.
- [x] Removes the Cylc7 data generation & structures.

Addressed elsewhere:
- [x] `cylc_scan.py` dealt with in #3724 

Commands scripts to do:
- [x] `cylc_broadcast.py`
- [x] `cylc_checkpoint.py`
- [x] `cylc_dump.py`
- [x] `cylc_ext_trigger.py`
- [x] `cylc_get_suite_version.py`
- [x] `cylc_hold.py`
- [x] `cylc_kill.py`
- [x] `cylc_message.py`/`task_message.py`
- [x] `cylc_nudge.py`
- [x] `cylc_ping.py`
- [x] `cylc_poll.py`
- [x] `cylc_release.py`
- [x] `cylc_reload.py`
- [x] `cylc_remove.py`
- [x] `cylc_set_outputs.py`
- [x] `cylc_set_verbosity.py`
- [x] `cylc_show.py`
- [x] `cylc_stop.py`
- [x] `cylc_trigger.py`

**Requirements check-list**
- [x] No dependency changes.


dwsutherland:
Slowly making my way through.. 

Broadcast deltas are somewhat event driven (create a `workflow.broadcast` delta on set/clear of broadcasts), this could be the template for fixing #3746

dwsutherland:
#3724 Needs to go in first

cylc/flow/scheduler.py
```diff
@@ -364,7 +360,6 @@ async def initialise(self):
             self.suite, context=self.zmq_context, barrier=self.barrier)
 
         self.proc_pool = SubProcPool()
-        self.state_summary_mgr = StateSummaryMgr()
```
oliver-sanders, line 367: Woohoo!

cylc/flow/scheduler.py
```diff
@@ -882,100 +844,22 @@ def info_get_graph_raw(self, cto, ctn, group_nodes=None,
...
-
     def command_stop(
             self,
             mode=None,
             cycle_point=None,
             # NOTE clock_time YYYY/MM/DD-HH:mm back-compat removed
             clock_time=None,
-            task=None
+            task=None,
+            kill_active_tasks=False,
```
oliver-sanders, line 854: `kill_active_tasks` is covered by `StopMode.KILL`.

cylc/flow/scripts/cylc_get_suite_version.py
```diff
@@ -32,8 +36,30 @@ def main():
     (options, args) = parser.parse_args()
     suite = args[0]
 
-    pclient = SuiteRuntimeClient(suite, timeout=options.comms_timeout)
-    print(pclient('get_cylc_version'))
+    try:
+        pclient = SuiteRuntimeClient(suite, timeout=options.comms_timeout)
+    except Exception as exc:
+        print(cparse(f'<red>{exc}</red>'))
+        sys.exit(1)
```
oliver-sanders, line 43: OOh, looks like this script isn't wrapped with `cylc.flow.terminal.cli_function` for some reason, that will take care of exception formatting.

cylc/flow/scripts/cylc_ping.py
```diff
@@ -45,22 +47,59 @@ def get_option_parser():
...
+            query ($wFlows: [ID]) {
+              workflows(ids: $wFlows) {
+                id
+                name
+                port
+                pubPort
+              }
+            }
+        ''',
+        'variables': {'wFlows': [suite]}
```
oliver-sanders, line 64: Is this using a running workflow to list other running workflows?

Or is the intention to use the UIS?

oliver-sanders (commented):
Looks like you had some fun with `cylc broadcast`!

cylc/flow/scheduler.py
dwsutherland: There might be some `TaskProxy.summary` to get rid of too.. But I'll leave that for when the data store moves to event driven updates.

cylc/flow/scheduler.py
dwsutherland: Oh, right! (need to put another mode in the input enum then).

cylc/flow/scripts/cylc_ping.py
dwsutherland: Artifact of UIS use, if the data-store had more than one workflow (every data-store is structured `self.data[w_id]`) or the resolvers used scan to send the query off to other workflows (without causing a network storm! 🌩️ ) we could..

cylc/flow/scheduler.py
dwsutherland: There existed no `StopMode.KILL` so I created the new enum for it (`StopMode.REQUEST_KILL`) and removed `kill_active_tasks`...

cylc/flow/scripts/cylc_get_suite_version.py
dwsutherland: Done.

cylc/flow/scripts/cylc_ping.py
dwsutherland: Also, yes (if that's what you meant, sorry!); once the CLI goes through the UIS this will kick in to target multiple workflows.

dwsutherland:
> No change log entry required (invisible to users).

The only change I can think of that would effect users, is the CLI JSON format is different for a few commands (i.e. `cylc show` )..
The other one; `cylc show --list-prerequisites` only shows prerequisites for live tasks, which may be rectified post n-window (TODO in code)..

cylc/flow/scripts/cylc_get_suite_version.py
```diff
@@ -20,20 +20,42 @@
...
+              }
+            }
+        ''',
+        'variables': {'wFlows': [suite]}
+    }
+
+    result = pclient('graphql', query_kwargs)
+
+    for workflow in result['workflows']:
+        print(workflow['cylcVersion'])
```
kinow, line 58: I think it will always be a single workflow, but harmless to use a loop to unpack the first result I think

kinow (approved):
Tested with the workflow `five`, found no errors using the few commands I am familiar with (run, nudge, poll, broadcast, hold, release, cat-log, dump).

Had a look at the code, left a few comments, but no blockers for me :+1: amazing work @dwsutherland !

Tried to test the UI, but there's some hiccup with that `MSG_TIMEOUT`, causing UIS -> WFS to fail to spawn in jupyterhub.

Thanks!
B

cylc/flow/task_message.py
```diff
@@ -46,6 +46,24 @@
...
+  message (
+    workflows: $wFlows,
+    taskJob: $taskJob,
+    eventTime: $eventTime,
+    messages: $messages
+  ) {
+    result
+  }
+}
+'''
```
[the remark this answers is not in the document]
dwsutherland: > Would it be useful here too? Looks like Graphene provides objects to create mutations too: https://docs.graphene-python.org/en/latest/types/mutations/

Those are schema definition objects.

cylc/flow/scripts/cylc_get_suite_version.py
dwsutherland: UIS API/client can/will peddle to multiple workflows.

cylc/flow/scripts/cylc_show.py
```diff
@@ -57,77 +137,136 @@ def get_option_parser():
...
                             print('  o  %s ... %s' % (key, value))
-            for bad_item in bad_items:
-                ansiprint(
-                    f"<red>No matching tasks found: {bad_item}\n",
-                    file=sys.stderr)
-            if bad_items:
-                sys.exit(1)
+        if not results['taskProxies']:
+            ansiprint(
+                f"<red>No matching tasks found: {task_ids}",
```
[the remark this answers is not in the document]
dwsutherland: That code was already there (see above).. (I suppose it just means "do the whole line")

cylc/flow/task_message.py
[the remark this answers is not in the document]
dwsutherland: Could be an idea to collect the strings into one place, but there's no (or almost no) duplication because each is tailored to the respective command (and I can't think of where else I'd use them).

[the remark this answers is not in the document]
kinow: I wasn't sure if there were queries or fragments being shared among the commands. If there's little or no sharing, I think it's OK to leave as is. 90% of the commands required only the mutation string and that `pclient` call, so quite simple to change/maintain I think.

cylc/flow/scripts/cylc_ping.py
oliver-sanders: > if the data-store had more than one workflow or the resolvers used scan to send the query off to other workflows

???

oliver-sanders: If the UIS is present we can get a workflow listing from there.

If the UIS is not present, how would this work? Surely we would need to do a FS scan to discover workflows?

cylc/flow/scripts/cylc_show.py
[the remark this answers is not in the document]
oliver-sanders: I think ansimarkup has been configured to auto-close tags on each line.

Having said that, this may well be a mistake!

cylc/flow/scripts/cylc_ping.py
dwsutherland: > If the UIS is present we can get a workflow listing from there.

Yes, this is already in place. The UIS mutation resolvers are different from the WFS, they parse the `ids`/`workflows` arguments to figure out which workflow(s) to send it to.

> If the UIS is not present, how would this work? Surely we would need to do a FS scan to discover workflows?

Correct, we would have to use resolvers like those at the UIS to parse the arguments against a list of workflow IDs.. However, we wouldn't want duplicate parsing at both UIS and WFS (so that might have to be an arg too!)

dwsutherland: You can actually test it out in the WUI GraphiQL, just use a glob for the workflow IDs i.e.:
```
mutation {
  stop(workflows: ["*"]) {
    result
  }
}
```
(off the top of my head)

hjoliver:
(This is my next review priority)

hjoliver (commented):
7/55 files viewed ... getting there :+1:

hjoliver:
> > No change log entry required (invisible to users).
> 
> The only change I can think of that would effect users, is the CLI JSON format is different for a few commands (i.e. `cylc show` )..
> The other one; `cylc show --list-prerequisites` only shows prerequisites for live tasks, which may be rectified post n-window (TODO in code)..

:+1: not major problems.

cylc/flow/suite_status.py
```diff
@@ -65,6 +65,9 @@ class StopMode(Enum):
     REQUEST_CLEAN = 'REQUEST(CLEAN)'
     """External shutdown request, will wait for active jobs to complete."""
 
+    REQUEST_KILL = 'REQUEST(KILL)'
+    """External shutdown request, will wait for active jobs to be killed."""
```
hjoliver, line 69: Not the fault of this PR (`git blame` fingers @oliver-sanders) but I don't think we doing this in other modules (are we?) and in fact last I heard Python doesn't have attribute docstrings - the PEP was rejected years ago.  So these strings should probably just be comments before the attributes?  (Or is this for Sphinx doc generation? I'm sure @oliver-sanders will enlighten me!)

dwsutherland: Yeah, it appears only the describe method is used to get different strings (?):
https://github.com/dwsutherland/cylc-flow/blob/e4c6e4cc9571893d6c7b9523d0cb1dbec9c383cc/cylc/flow/network/schema.py#L1297-L1308

tests/functional/graph-equivalence/00-oneline.t
```diff
@@ -25,24 +25,21 @@ set_test_number 5
...
 TEST_NAME="${TEST_NAME_BASE}-validate"
-run_ok "${TEST_NAME}" cylc validate "${SUITE_NAME}"
+run_ok "${TEST_NAME}" cylc validate \
+    --set=TEST_OUTPUT_PATH="${PWD}" "${SUITE_NAME}"
 #-------------------------------------------------------------------------------
 TEST_NAME="${TEST_NAME_BASE}-run"
-suite_run_ok "${TEST_NAME}" \
-    cylc run --reference-test --debug --no-detach "${SUITE_NAME}"
+suite_run_ok "${TEST_NAME}" cylc run --reference-test --debug --no-detach \
+    --set=TEST_OUTPUT_PATH="${PWD}" "${SUITE_NAME}"
```
hjoliver, line 33: Ha-ha, I was wondering why the changes in this test and the associated `test1/flow.cylc`, then thought why in hell is this test passing on master... but I see you explained it:
> The other one; cylc show --list-prerequisites only shows prerequisites for live tasks, which may be rectified post n-window

hjoliver: (and same for the other graph-equivalence tests)

cylc/flow/data_store_mgr.py
```diff
@@ -299,6 +306,7 @@ class DataStoreMgr:
         'min_point',
         'parents',
         'pool_points',
+        'publish_deltas',
```
hjoliver, line 309: I should have picked this up previously, but we probably don't need `__slots__` in this class.  As I understand it, it's only a significant advantage (memory footprint) for classes that instantiate a large number of instances, such as task proxy and cycling-related objects in cylc-flow.

hjoliver: Official docs on explain it as such, anyway:
https://docs.python.org/3/glossary.html#term-slots

> best reserved for rare cases where there are large numbers of instances in a memory-critical application.

cylc/flow/data_store_mgr.py
```diff
@@ -1126,7 +1196,7 @@ def get_publish_deltas(self):
         result.append(
             (ALL_DELTAS.encode('utf-8'), all_deltas, 'SerializeToString')
         )
-        return result
+        return deepcopy(result)
```
hjoliver, line 1199: Why this deepcopy?

cylc/flow/data_store_mgr.py
dwsutherland: Protobuf messages are class instances, does this count?

cylc/flow/data_store_mgr.py
dwsutherland: So the data elements aren't modified between gathering them here and publishing them.

cylc/flow/data_store_mgr.py
hjoliver: It might help if the Protobuf message class itself had slots, but this is the `DataStoreMgr` class, which only gets instantiated once, right?

hjoliver: i.e. wasted/spare memory in a small attributes dict doesn't matter much for a single instance, but it might for a very large number of instances.

dwsutherland: Roger that, will remove slots (although it was helpful for the memory profiler to iterate through)

cylc/flow/data_store_mgr.py
hjoliver: Fair enough.

hjoliver (approved):
Tests as working, passes all tests, code LGTM :+1: Question about use of slots is no biggie.
Great job @dwsutherland :bouquet: 

I still need to get a better understanding of some of the inner workings (and Protobuf unfortunately makes examining sh1t in a debugger more difficult, for me anyway :grimacing: )  but no need to block this (can I hit you up for a bit of a screen-sharing session Thursday or Friday @dwsutherland ?)

hjoliver:
That's 2 approvals, but we should probably wait for at least a sanity check from @oliver-sanders on this one.

dwsutherland:
> can I hit you up for a bit of a screen-sharing session Thursday or Friday @dwsutherland ?

For sure! I'll be out on Thursday afternoon, but Friday should be good if not tomorrow morning.

Thanks for the review!
(I'll get rid of slots in the n-window PR, since I'm modifying it there)

cylc/flow/data_store_mgr.py
dwsutherland: I'll have to check if there's even a chance that they could be.. Might not be needed, probably safer, might depend on whether they need accessed aysnc to other things and how ZMQ publisher works..

cylc/flow/suite_status.py
oliver-sanders: These strings do nothing in Python, however, Sphinx has some special parsing logic which picks them up which makes for a convenient way of documenting them.

cylc/flow/data_store_mgr.py
oliver-sanders: `__slots__` are unnecessary for classes like this that don't get initiated often, however, they are also harmless.

oliver-sanders:
> That's 2 approvals, but we should probably wait for at least a sanity check from @oliver-sanders on this one.

I'm most the way there, haven't found anything significant, looks good.

This functionality is so heavily tested by the functional tests that I'm pretty happy it works.

tests/functional/graph-equivalence/02-splitline.t
```diff
@@ -26,25 +26,22 @@ set_test_number 5
...
 #-------------------------------------------------------------------------------
 TEST_NAME="${TEST_NAME_BASE}-check-a"
-cylc run "${SUITE_NAME}" --hold 1>'out' 2>&1
-poll_grep_suite_log 'Holding all waiting or queued tasks now'
-cylc show "${SUITE_NAME}" 'a.1' | sed -n "/prerequisites/,/outputs/p" > 'a-prereqs'
 cmp_ok "${TEST_SOURCE_DIR}/splitline_refs/a-ref" 'a-prereqs'
 #-------------------------------------------------------------------------------
 TEST_NAME="${TEST_NAME_BASE}-check-b"
-cylc show "${SUITE_NAME}" 'b.1' | sed -n "/prerequisites/,/outputs/p" > 'b-prereqs'
+poll_grep_suite_log 'INFO - DONE'
```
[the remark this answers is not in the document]
dwsutherland: Hmm.. But the suite task runs the `cylc show` now for `b-prereqs`, so at minimum I'd need to poll the task for completion.

cylc/flow/data_store_mgr.py
hjoliver: Yeah no biggie, but still, I don't think we should use __slots__ unnecessarily. There is at least a small maintenance cost, and it makes the class slightly harder to use (in experimental mode, at least).

cylc/flow/data_store_mgr.py
hjoliver: (That's one of the things I'd like to understand better).

cylc/flow/suite_status.py
hjoliver: Should we have an issue up to extend this throughout the code then?

tests/functional/graph-equivalence/02-splitline.t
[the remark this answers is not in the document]
hjoliver: You used `cylc run --no-detach` so the suite will run to completion before any of the `cmp_ok` tests run.

oliver-sanders (approved):
Finished code review, looks good, tested as working.

tests/functional/graph-equivalence/02-splitline.t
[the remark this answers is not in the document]
dwsutherland: removed

PR #3759, opened 2020-08-06
Bugfix: cylc-gui - tasks momentarily out of sequence

### LINKED ISSUE #3698: cylc-gui - tasks momentarily out of sequence
Since upgrading from 7.8.0 to 7.8.6, I'm noticing a new visual oddity with our GUI when running a rather large suite.  For 2-10 seconds the tasks in a family will appear out of sequence:
image
Then the sequence restores itself:
image

It is not a major problem, but it is quite distracting and can lead to confusion when it changes.  It's almost as though an underlying object converts from a Dict to an OrderedDict, and because our suite is so stupidly large, we catch it in the act.

Thanks for your time and hard work!

Glenn

GlennCreighton:
gui: fix tree-view out-of-sequence tasks caused by default sorting on wrong column index

Unclear as to the reason why this bug does not show up with newer versions of Python or GTK, but this fixes the problem on my end and does not appear to hurt anything else.

These changes close #3698 

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/lib/cylc/gui/view_tree.py
+++ b/lib/cylc/gui/view_tree.py
@@ -264,11 +264,11 @@ class ControlTree(object):
         return self._nat_cmp(prop1, prop2)
 
     def default_sort_column(self, model, iter1, iter2):
-        point_string1 = model.get_value(iter1, 1)
-        point_string2 = model.get_value(iter2, 1)
-        if point_string1 is None or point_string2 is None:
-            return cmp(point_string1, point_string2)
-        return self._nat_cmp(point_string1, point_string2)
+        task_name_string1 = model.get_value(iter1, 1)
+        task_name_string2 = model.get_value(iter2, 1)
+        if task_name_string1 is None or task_name_string2 is None:
+            return cmp(task_name_string1, task_name_string2)
+        return self._nat_cmp(task_name_string1, task_name_string2)
 
     def _get_interval_in_seconds(self, val):
         """Convert the IOS 8601 date/time to seconds."""
--- a/lib/cylc/gui/view_tree.py
+++ b/lib/cylc/gui/view_tree.py
@@ -264,8 +264,8 @@ class ControlTree(object):
         return self._nat_cmp(prop1, prop2)
 
     def default_sort_column(self, model, iter1, iter2):
-        point_string1 = model.get_value(iter1, 0)
-        point_string2 = model.get_value(iter2, 0)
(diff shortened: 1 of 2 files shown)


hjoliver:
Looks like an expertly-submitted first PR @GlennCreighton - first of many maybe!! 

Assigning @kinow as main reviewer (when he returns next week) - as author of the original sorting code here.

hjoliver:
>  I have read CONTRIBUTING.md and added my name as a Code Contributor.

You checked this box, but I don't see your name in the file.

hjoliver:
(@GlennCreighton I just saw your comment on editing `CONTRIBUTING.md` via the GH UI edit button - not sure how that works with respect to PR branches, but I don't see any sign of it here ... may be best to do it manually via another local commit to your branch, which you should then push to the PR branch on your fork - and it will magically show up here as a new commit on the PR)

GlennCreighton:
@hjoliver okay, CONTRIBUTING.md updated on my branch.  Thanks for the git lessons!

lib/cylc/gui/view_tree.py
```diff
@@ -264,8 +264,8 @@ def sort_column(self, model, iter1, iter2, col_num):
         return self._nat_cmp(prop1, prop2)
 
     def default_sort_column(self, model, iter1, iter2):
-        point_string1 = model.get_value(iter1, 0)
-        point_string2 = model.get_value(iter2, 0)
+        point_string1 = model.get_value(iter1, 1)
+        point_string2 = model.get_value(iter2, 1)
```
hjoliver, line 268: These variables now have bad names (as @kinow has pointed out, the `model.get_value()` is now retrieving the task name, not the cycle point) - could you just change that @GlennCreighton, just to help avoid confusion in future?

It does seem somewhat surprising that sorting on task name intead of cycle point results in correct sorting by cycle point (in the integer case, with single- and double-digit points at the same time) ... but given this GUI is at end-of-life I don't think we need to spend much time worrying about that.

GlennCreighton:
Variable names changed as requested.  I share your surprise that things worked properly as well.  I don't understand how the default sort is being applied by gtk to the tree, but it definitely seems to be looping through tasks and comparing one to the next.

hjoliver:
Oh, one other thing, sorry - can you also add a new line `CHANGES.md` to reference this Pull Request, and push that up to this branch?

PR #3761, opened 2020-08-07
made job file test work against kgo

wxtim:
### Problem
The comparison of job file sizes in `test_job_file.py` is brittle, and doesn't give any sense of the nature of changes to the job file. Two job files the same size but with no other similarity will pass this test.

### Possible solutions
* Using an MD5 or similar hash to ensure that the content as well as the size of the file remains the same. The test remains fragile and is still totally uninformative.
* explicitly reading files line by line asserting each line is the same. This would be slow, but easy to understand.
* Using `filecmp.cmp(a, b, shallow=False)` - quick, but uninformative.
* Using difflib - still slower, but much more usful if changes occur. ⭐  - I went with this solution.

### Uncertainties
Whether I should have put the kgo inside this file. The trade off here is that it'd be faster if I had done this: I believe that having a standalone file makes it easier for devs to make changes in future - they can use their fave graphical diff program along with the commented code to examine what is happening.

### How a reviewer might test this
* Change items in the file `tests/unit/kgo/job_file.kgo` and see what happens.
* Change items in the `job_conf` dict in `test_write` in `tests/unit/test_job_file.py`.

### DIFF
--- a/tests/unit/kgo/job_file.kgo
+++ b/tests/unit/kgo/job_file.kgo
@@ -1,0 +1,87 @@
+#!/bin/bash -l
+#
+# ++++ THIS IS A CYLC TASK JOB SCRIPT ++++
+# Suite: farm_noises
+# Task: baa
+# Job log directory: 1/baa/01
+# Job submit method: background
+# Job submit command template: woof
+# Execution time limit: moo
+if [[ $1 == 'noreinvoke' ]]; then
+    shift
+else
+    exec bash -l "$0" noreinvoke "$@"
+fi
+
+CYLC_FAIL_SIGNALS='EXIT ERR TERM XCPU'
+export CYLC_VERSION='8.0a3.dev'
+
+cylc__job__inst__cylc_env() {
+    # CYLC SUITE ENVIRONMENT:
+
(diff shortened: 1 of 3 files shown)


tests/unit/kgo/job_file.kgo
```diff
@@ -0,0 +1,87 @@
...
+# Job submit command template: woof
+# Execution time limit: moo
+if [[ $1 == 'noreinvoke' ]]; then
+    shift
+else
+    exec bash -l "$0" noreinvoke "$@"
+fi
+
+CYLC_FAIL_SIGNALS='EXIT ERR TERM XCPU'
+export CYLC_VERSION='8.0a3.dev'
```
oliver-sanders, line 17: Need to replace the version with the current version (this is what caused the last breakage).

wxtim: Dun

oliver-sanders: Every time the cylc version changes this test will break.

wxtim:
This is too thoroughgoing -> This new test simply replicates existing functionality.

PR #3765, opened 2020-08-07
make test for existance of job file more robust

wxtim:


wxtim:
Surplus to requirements

Issue #3595, opened 2020-05-05
portability: dns


oliver-sanders:
DNS issues have long caused portability issues with Cylc.

A lot of our tests make the assumption that `socket.getfqdn` will always match `hostname -f`, sadly this assumption is not realistic. Of the six platforms I use regularly this assumption only holds for two of them.

Here are five different methods for identifying localhost:

```
hostname
hostname -f
python3 -c 'import socket; print(socket.getfqdn())'
python3 -c 'import socket; print(socket.gethostname())'
python3 -c 'from cylc.flow.hostuserutil import get_fqdn_by_host; print(get_fqdn_by_host(None))'
```

And here are the results from a few different platforms:

Ubuntu (Vanilla)
===========

No surprises, all methods return the short hostname:

```
ubuntu
ubuntu
ubuntu
ubuntu
ubuntu
```

Mac OS
======

* `socket.getfqdn` returns an IPV6 loopback.
* Hostnames are case insensitive but are configured with capitals, some socket methods reduce names to lowercase causing diffs.

```
Bobs-MacBook.local
Bobs-MacBook.local
1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.ip6.arpa
Bobs-MacBook.local
1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.ip6.arpa
```

GitHub Actions
==============

* `hostname -f ` disagrees with `socket.getfqdn`.

```
fv-az33
fv-az33.ukytxvjih3denlwyewf2iq3q4b.cx.internal.cloudapp.net
fv-az33.internal.cloudapp.net
fv-az33
fv-az33.internal.cloudapp.net
```

Travis CI
========

* `hostname -f ` disagrees with `socket.getfqdn`.

```
travis-job-e8a47a2f-a1eb-4e70-96e1-d884e48e24bc
travis-job-e8a47a2f-a1eb-4e70-96e1-d884e48e24bc
localhost
travis-job-e8a47a2f-a1eb-4e70-96e1-d884e48e24bc
localhost
```

Networked Linux example 1
=========

No surprises, one fqdn and one short hostname:

```
vld601
vld601.c.m.g.u
vld601.c.m.g.u
vld601
vld601.c.m.g.u
```

Networked Linux example 2
==========

[...]

PR #3766, opened 2020-08-09
Identify localhost by common IP address.

hjoliver:

Tentatively assigning @matthewrmshin as reviewer, as he authored the `is_remote_host` method and this is a small change - Matt, no worries if you can't spare the time. 

On my Centos 7 work laptop one of our unit tests fails because `$HOSTNAME` gets identified as a remote host:
```
>       self.assertFalse(is_remote_host(os.getenv('HOSTNAME')))
E       AssertionError: True is not false
```

Looking at the code, the "extended host info" (from `socket.gethostbyname_ex`) is expected to be identical for `$HOSTNAME` and `localhost`, which isn't the case on my box.  My `/etc/hosts` and `/etc/hostname` files (not sure if they are "correct", but they weren't configured by me):
```
# /etc/hostname
pastry
```

```
# /etc/hosts
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
127.0.0.1   pastry localhost
```
This results in the following extended host info:
```python
>>> socket.gethostbyname_ex("localhost")
('localhost',
 ['localhost.localdomain', 'localhost4', 'localhost4.localdomain4', 'localhost.localdomain', 'localhost6', 'localhost6.localdomain6', 'localhost', 'pastry'],
 ['127.0.0.1', '127.0.0.1', '127.0.0.1'])

>>> socket.gethostbyname_ex("pastry")
('pastry',
 ['localhost'],
 ['127.0.0.1'])
```

So I'm guessing, we should identify "the same host" with different names, by looking for a common IP address (`127.0.0.1` in this case), but not by expecting the full extended host info to be identical.  Does that seem reasonable?

**Requirements check-list**
- [x] Covered by existing tests, sort of (depends on system-level host config)

- [x] No dependency changes.

### DIFF
--- a/cylc/flow/hostuserutil.py
+++ b/cylc/flow/hostuserutil.py
@@ -189,8 +189,10 @@ class HostUtil(object):
                 except IOError:
                     self._remote_hosts[name] = True
                 else:
-                    self._remote_hosts[name] = (
-                        host_info != self._get_host_info())
+                    # look for common IP with localhost
+                    self._remote_hosts[name] = not bool(
+                        set(self._get_host_info()[2]).intersection(
+                            set(host_info[2])))
         return self._remote_hosts[name]
 
     def is_remote_user(self, name):


kinow:
All good, `is_remote_host('127.0.0.1')` returns `True` on my computer for `master` and for this branch. So no regression.

hjoliver:
>  Actually looks like I get True for is_remote_host('127.0.0.1')? 

`is_remote_host` is only called for host *names*, I think.  (And presumably a system should never be configured such that the `hostname` command would return `127.0.0.1` ... out of my depth here though).

hjoliver:
(And similar considerations apply for `0.0.0.0` i would think? -  i.e. its not a host *name*).

cylc/flow/hostuserutil.py
```diff
@@ -189,8 +189,10 @@ def is_remote_host(self, name):
                 except IOError:
                     self._remote_hosts[name] = True
                 else:
-                    self._remote_hosts[name] = (
-                        host_info != self._get_host_info())
+                    # look for common IP with localhost
+                    self._remote_hosts[name] = not bool(
+                        set(self._get_host_info()[2]).intersection(
+                            set(host_info[2])))
```
matthewrmshin, line 195: Looks good. Can also use the `&` operator syntactic sugar:

```suggestion
                    self._remote_hosts[name] = not (
                        set(self._get_host_info()[2]) & set(host_info[2]))
                    )
```

oliver-sanders:
See also the somewhat related https://github.com/cylc/cylc-flow/issues/3595 which also effects localhost self-identification.

datamel:
@hjoliver's setup is probably an edge case with `pastry` pointed at the loopback IP in `/etc/hosts`. But I guess it is something we should cater for?

I think this PR is a move in the right direction although any *remote* hosts pointing to loopback IP(s), aka the local host, should never be getting as far the new `.intersection` check. We could capture all of these system peculiarities with clearer semantics,  calling out to:

`IPAddress(...).is_loopback()`

Or, if you prefer, switch out:

`.startswith("localhost"):`

For the IP address instead, which we can grab with a call to `socket.gethostbyname(name)`.

`.startswith("127."):`

There is one scenario the new set intersection test does not catch still. One similar to where @hjoliver has the hostname `pastry` in his `/etc/hosts`, pointing to his loopback IP, but the FQDN e.g. `pastry.niwa.co.nz` is given as the remote host. The IPs for these would not match and it would classified as remote when the user probably intended it to be a local connection. I've played with some code from stackoverflow that could be adapted to help. Where we can take a peek at all the network interfaces e.g. https://stackoverflow.com/a/27423915. We can then add these (not that a typical system has more than 1 external network interface) to the set of IPs that are the 'local' host to ensure the intersection check is aware of all of the local host's addresses.

This led me to taking a look at our method `get_local_ip_address(target)` and I think it doesn't need a target (and so first line of docstring is wrong). It can be simplied from:

```python
ipaddr = ""
with suppress(IOError):
    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as sock:
        sock.connect((target, 8000))
        ipaddr = sock.getsockname()[0]
return ipaddr
```
to:

`return socket.gethostbyname(socket.gethostname())`

Shall I put a PR in tomorrow?

*Maybe* we could consider adding behaviour in there that fishes out a proper external IP from the network interfaces where 127.0.0.1 is ever returned? Depends how cheeky people are with their `/etc/hosts` and whether `socket.gethostbyname` checks that before trying a DNS server.

hjoliver:
@oliver-sanders - 

> Seems legit and works for my Mac OS environment, however, sadly the assumption does not hold for my networked Linux box.

Which assumption does hold? That there is a common IP address in extended host info for `$HOSTNAME` and `localhost`?  Does that makes sense, or is your box misconfigured?

hjoliver:
@datamel - you sound like you know what you're talking about - which is impressive, this network config stuff is a dark art to me.

> @hjoliver's setup is probably an edge case with pastry pointed at the loopback IP in /etc/hosts. But I guess it is something we should cater for?

Good question.  I'll ask our IT people if that is deliberate. It's possible that I screwed it up myself some time ago and forgot about it.

> Shall I put a PR in tomorrow?

Sure. Shall I cancel this one?

oliver-sanders:
> Does that makes sense, or is your box misconfigured?

My box has the same config as all others on our network, I don't think that there is a rule that says there *must* be a common IP between the two.

oliver-sanders:
Checking the network interface would be a pretty sophisticated check for localhost, however, I don't know, is it possible to have things that look like hosts (e.g. docker containers or virtual systems) on the same network interface?

IPV6 reminder, the loopback IP can now take different forms:

* `127.*` (IPV4)
* `0000:0000:0000:0000:0000:0000:0000:0001` (IPV6)
* `0:0:0:0:0:0:0:1` (IPV6 short)
* `::1` (IPV6 abbrev)
* `::0001` (IPV6 long+abbrev?)
* `:ffff:127.*` (IPV4 mapped IPV6 address?)
* (possibly others)

datamel:
> Checking the network interface would be a pretty sophisticated check for localhost, however, I don't know, is it possible to have things that look like hosts (e.g. docker containers or virtual systems) on the same network interface?
> IPV6 reminder, the loopback IP can now take different forms:
> 
> 127.* (IPV4)
> 0000:0000:0000:0000:0000:0000:0000:0001 (IPV6)
> 0:0:0:0:0:0:0:1 (IPV6 short)
> ::1 (IPV6 abbrev)
> ::0001 (IPV6 long+abbrev?)
> :ffff:127.* (IPV4 mapped IPV6 address?)
> (possibly others)

Yep, good point. Makes more of an argument for using `is_loopback()` to catch all the cases for us.

hjoliver:
For the record, I think the weird line in my `/etc/hosts` was not supposed to be there (by comparison with my work desktop box):
```
127.0.0.1   pastry localhost
```
... and removing it made the test pass.

However, we have decided (via chat platform) to close this PR as superceded by a new umbrella issue that will take into account the upcoming "platforms" changes that should reduce or eliminate our reliance on system configuration for host identification...

PR #3772, opened 2020-08-13
Platforms: Upgrade Cylc 7 Host Syntax

wxtim:
## Summary
As per [the deprecation method section of the platforms proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-platforms.md#deprecation-method) suite tasks where Cylc 7 syntax such as
```
[runtime]
  [[MY_TASK]]
    [[[job]]]
       batch system = some batch system
    [[[remote]]]
      host = myhost
```
should select a sensible batch system, or return a reasonable error if unable to do so.

This PR attempts to address this.

## Work done
1. Removed `cylc.flow.cfgspec.suite.host_to_platform_upgrader` and replaced it with an upgrade warner function to be called on workflow validation. This can't catch every possible problem, but will hopefully fail loudly if something obviously wrong is put in a config.
2. Added the logic to a `get_platform` interface to the platform module: Replaced instances of `platform_from_name` with `get_platform`.
3. Added upgrader logic to `get_platform`.
4. Refactored some functional tests which I missed in the original platforms work, but now fail.
5. Added Functional test ` tests/functional/validate/04-host-platform-upgrade.t ` - this is not meant to comprehensively test the upgrader function, merely to test that it has been plugged into the cylkc
6. Added s unit test `tests/unit/cfgspec/test_suite.py` which is meant to provide comprehensive coverage of the warning function, and a unit test for `get_platform` to `tests/unit/test_platforms.py`.


tests/functional/test_config_upgrader.py
```diff
@@ -221,9 +221,13 @@ def test_upgrader_where_host_is_function(tmp_path, caplog):
     caplog.set_level(logging.DEBUG)
```
MetRonnie, line 221: I don't understand why this file has been moved to `functional/`?

tests/unit/test_suite_config_upgrader.py
```diff
@@ -0,0 +1,212 @@
...
+def test_noplatform_hostname__cantselectplatform(
+    get_task_conf, mock_glbl_cfg
+):
+    """
+    Check that we can select a platform and remove forbidden items
+
+    This test does not need to replicate the tests for the
+    `platform_from_job_info` (a.k.a Reverse Lookup), just confirm
+    that we are using that function correctly here.
+    """
```
MetRonnie, line 164: Docstring copied from test above - needs updating to better describe this test

wxtim: 🐆

tests/functional/test_config_upgrader.py
wxtim: I think that at one point I began to believe that this was more functional than unit, but it isn't.

cylc/flow/config.py
```diff
@@ -305,7 +305,7 @@ def __init__(
         self.mem_log("config.py: before inheritance")
         self.compute_inheritance()
         self.mem_log("config.py: after inheritance")
-
+        self.cfg = host_to_platform_upgrader(self.cfg)
```
hjoliver, line 308: Here your doing the upgrade post-inheritance, so this flow:
```
[runtime]
    [[root]]
          [[[remote]]]
                host = cheese
    [[foo<m>]]  # m = 1.. 1000
```
will result in 1001 upgrades (root and foo_m) instead of 1 (just root) - and 1001 validation warnings!

We have done every other syntax upgrade in the cfgspec module on the sparse config, after parsing and before inheritance processing, and then the upgraded result gets inherited through the family hierarchy as normal.

I guess you can't do that here because the upgrader assumes `localhost` for any namespace with no host or platform specified.  Pre-inheritance, most namespaces are in this category, but they aren't targeting localhost because they inherit the remote host.

This method also pulls every task (via `platform = localhost`) into the sparse config, which is only supposed to hold config items explicitly specified by the user (try `cylc get-config --sparse -i runtime ...`).

So I guess the question is, does platforms upgrade localhost selection prevent us from doing this the usual efficient way?  What about just inserting `platform = localhost` in the root namespace, if no platform or remote is specified there already?

wxtim: Thank goodness - I thought it should not be necessary.

wxtim: > What about just inserting platform = localhost in the root namespace, if no platform or remote is specified there already?

Jargon Brainfreeze - are we talking about setting a default in the config thus: 
```
with Conf('platform', VDR.V_STRING, 'localhost'):
```

hjoliver: Maybe just add the default value at line 493 (no `with` needed) ... I think.  Then adapt your upgrader function to deal with that change.  (Not 100% sure though, since Oliver's fancy new self-documenting `with Conf(...)` refactor ... would need to have a play with it to confirm...)

wxtim: This looks like it's not going to work, because if I have the following suiterc runtime:
```
[[FAMILY]]
  [[[remote]]]
    host = supercomputer_login_node01
[[mytask]]
  inherit = FAMILY
  [[[job]]]
    batch system = slurm
```
and platforms thus:
```
[[supercomputer_slurm]]
hosts = supercomputer_login_node01, supercomputer_login_node02
batch system = slurm

[[smaller_spice_cluster]]
batch system = slurm
hosts = spice01, spice02
```
You'll end up with mytask being assigned `platform = smaller_spice_cluster` if you upgrade before inheritance, but `platform = supercomputer_slurm` if you do it after. 

Worse still, in this scenario FAMILY will fail if you try to upgrade at the normal time because there is no platform with `batch system = background` defined.

@dpmatthews  has suggested that we could put a warning about old syntax on upgrade and do the upgrading quietly where it is...

How much effort should we put into code which ought to be deprecated at Cylc9?

oliver-sanders: Because the upgrader requires multiple pieces of information and because these could come from anywhere in the inheritance hierarchy I don't think we can upgrade families, only tasks.

tests/functional/test_config_upgrader.py
oliver-sanders: So what aspect of this test is functional?

The "line in the sand" that I have in my head is:

* Integration - Involves the interaction of more than one significant piece of logic.
* Functional - Involves interaction with the shell, browser or other systems.

cylc/flow/config.py
wxtim: After a conversation with @dpmatthews and @oliver-sanders  I think we've concluded that because the Cylc 7 to platform upgrade is not a 1:1 mapping, but takes account of _all_ the items in the task `[remote]` and `[job]` sections we don't think that a traditional upgrade (i.e. before config expansion and inheritance is calculated) is possible. We've agreed that we think the best approach is to separate the deprecation warnings from the upgrade:

At suite validation the user will be warned that combinations of `host`, `batch system` and `batch submit command template` will be used to select platforms and will be deprecated at Cylc9. Further down the line this should include links to upgrade documentation.

Upgrade logic will be used on the fly by adding an interface to `platforms.py`:
```python
# Psuedocode really
get_platform(job_config):
  if job_config has platform name and not forbidden cylc 7 settings:
    get_platform_from_name()
  if job_config has platform name and forbidden with platforms cylc 7 settings:
    fail horribly
  if job_config not have platform name and forbidden with platforms cylc 7 settings:
    get_platform_from_cylc_7_settings()
  else:
    return localhost platform config
```

hjoliver: OK, damn it, right.  Can we perhaps give a stronger-than-normal deprecation warning given the unusual complexity of this upgrade process, which could be quite slow in a huge suite? (and check no significant impact on workflows that don't need upgrading).

hjoliver:
(A few conflicts and test failures now).

cylc/flow/config.py
wxtim: > stronger-than-normal deprecation warning

Yes

cylc/flow/config.py
```diff
@@ -1332,7 +1332,7 @@ def configure_sim_modes(self):
             # TODO - to use batch schedulers in dummy mode we need to
             # identify which resource directives to disable or modify.
             # (Only execution time limit is automatic at the moment.)
-            rtc['job']['batch system'] = 'background'
+            rtc['platform'] = 'localhost'
```
wxtim, line 1335: Technically this breaks dummy mode, but we've already agreed in principle that dummy-mode only serves to test network connections, and should go.

cylc/flow/platforms.py
```diff
@@ -24,6 +24,94 @@
...
+                    f'- task_conf[section][key] is '
+                    f' [{section}]{key} = {task_conf[section][key]}\n'
+                )
+            if fail_items:
+                raise PlatformLookupError(
+                    "A mixture of Cylc 7 (host) and Cylc 8 (platform)"
+                    " logic should not be used. In this case the "
+                    "following are not compatible\n" +
+                    fail_items
+                )
```
MetRonnie, line 77:
```suggestion
                raise PlatformLookupError(
                    f"A mixture of Cylc 7 (host) and Cylc 8 (platform) "
                    f"logic should not be used. In this case the "
                    f"following are not compatible\n{fail_items}"
                )
```
Use f-strings instead of string concatenation

cylc/flow/cfgspec/suite.py
```diff
@@ -1360,126 +1357,75 @@ def upg(cfg, descr):
...
+                    # inheritance we cannot catch all cases here.
+                    LOG.error(
+                        "A mixture of Cylc 7 (host) and Cylc 8 (platform)"
+                        " logic should not be used. In this case the "
+                        "following are not compatible\n"
+                        f"[{task_name}][{section}]{key} = "
+                        f"{task_cfg[section][key]} and "
+                        f"[{task_name}][platform] = {task_cfg['platform']}"
+                    )
+                    raise UpgradeError()
```
MetRonnie, line 1412: Is there a reason for logging with level "error" and then raising an empty error, instead of just raising the error with the message above?

wxtim: Yes - you get more error and less traceback.

MetRonnie (approved):

cylc/flow/cfgspec/suite.py
MetRonnie: Won't you get traceback for the empty exception anyway?

cylc/flow/cfgspec/suite.py
```diff
@@ -1360,126 +1357,75 @@ def upg(cfg, descr):
...
+                    limit_count += 1
+        if limit_count >= nwarnings:
+            break
+
+    if warnlines:
+        LOG.warning(
+            f"Cylc 7 Settings:\n{warnlines}"
+            "Cylc will attempt to upgrade these settings before submission "
+            f"but cannot be sure of success. (Only the first {nwarnings} tasks"
+            " are shown)"
```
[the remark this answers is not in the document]
wxtim: If I commit that should I create a follow up issue to add that link once we have the transition document?

cylc/flow/cfgspec/suite.py
```diff
@@ -1360,126 +1357,75 @@ def upg(cfg, descr):
...
-                        if (
-                            task_section in cfg['runtime'][task_name] and
-                            old_spec_item in
-                                cfg['runtime'][task_name][task_section].keys()
-                        ):
-                            poppable = cfg['runtime'][task_name][task_section]
-                            poppable.pop(old_spec_item)
-                    LOG.warning(
-                        f"Cylc 7 {old_spec_item} removed."
+                if 'platform' in task_cfg:
```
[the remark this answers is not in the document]
wxtim: I'm not sure that I agree here: We want to warn for any task with Cylc7 settings whether or not platforms have been set. Unless you feel that only the clash warnings (Which we can guarantee a failure for later) should be raised here.

Are you saying that I should ditch the warning part of the functionality?

cylc/flow/cfgspec/suite.py
wxtim: Oh - I've not got this right.

wxtim: This, it turned out was a proxy for a fairly nasty problem which @oliver-sanders found a fix for, for me. A special handler for this exception had to be added at about line 1800 in scheduler.py

wxtim:
Can I poke @oliver-sanders  and @MetRonnie  to have a final double check that I've answered outstanding issues?

MetRonnie (approved):

MetRonnie (approved):

cylc/flow/platforms.py
```diff
@@ -24,6 +24,93 @@
...
+                fail_items += (
+                    f'- task_conf[section][key] is '
+                    f' [{section}]{key} = {task_conf[section][key]}\n'
+                )
+            if fail_items:
+                raise PlatformLookupError(
+                    f"A mixture of Cylc 7 (host) and Cylc 8 (platform) "
+                    f"logic should not be used. In this case the "
+                    f"following are not compatible:\n{fail_items}"
+                )
```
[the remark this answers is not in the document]
wxtim: Which bit? The actual error message? (yes) 
The need to have it in 3 places - tricky to avoid.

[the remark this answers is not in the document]
wxtim: Or were you thinking of creating a single upgrade_check function which can be called anywhere?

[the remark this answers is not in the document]
wxtim: I'll see if I can.

tests/unit/test_platforms.py
```diff
@@ -73,6 +75,9 @@
 }
 
 
+# ----------------------------------------------------------------------------
+# Tests of platform_from_name
+# ----------------------------------------------------------------------------
```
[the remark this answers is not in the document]
hjoliver: I've always hated them in the bash test scripts too :vomiting_face: (no offense @wxtim !)

hjoliver (approved):
The new upgrade logic seems a pretty good way to do this. :+1:

tests/unit/test_platforms.py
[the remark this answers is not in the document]
wxtim: None Taken.

tests/functional/test_config_upgrader.py
oliver-sanders: I think this has been moved back again?

oliver-sanders:
I think all outstanding comments are resolved, looks ok, merging with two approvals.

PR #3779, opened 2020-08-18
conf: fix work directory reference

oliver-sanders:
Remove whitespace from a configuration reference which causes a parsing error in the docs.


oliver-sanders:
Was planning on fixing in cylc-sphinx-domain but

PR #3787, opened 2020-08-25
family/cycle state totals

dwsutherland:
In order to have state totals per cycle point and more efficient family/workflow state total updates, we should move away from:
- Updating whole cycle points of families if one task state changes.
- Iterating over config to find relationships.
- Iterating over all tasks on every change to update the workflow state totals.

And move towards:
- [x] Event driven family generation.
- [x] Targeted family state updates on event of task state updates.
- [x] Using the relationships set in the node elements (i.e. `first_parent`, `child_families`, `child_tasks`)
- [x] Workflow/Parent-families use child families/tasks to create state totals.

This PR implements this, and draft/diagram used for coding it up is:
family_state_totals
(note: this will recursively update families via `child_families` & `first_parent` fields, and it doesn't matter if you start from `root` or `Fam5` or somewhere in between)

**Requirements check-list**
- [x] No dependency changes.


hjoliver:
> FTU = `(root, Fam1, Fam2, Fam5)`

Shouldn't that include Fam4 as well? (It has a grandchild who's state changed) ... or am I missing something?

dwsutherland:
> > FTU = `(root, Fam1, Fam2, Fam5)`
> 
> Shouldn't that include Fam4 as well? (It has a grandchild whose state changed) ... or am I missing something?

It's a recursive function, and the updated ones add their first parent to the set of those to update, i.e.:
- If `Fam5` is worked on first it will update, then add `Fam4` to FTU.
- If `Fam2` is worked on first it will recurse through to `Fam4` then `Fam5`, update `Fam5` which will add `Fam4` to FTU (then `return` to `Fam4` to update that one and so forth.

cylc/flow/data_store_mgr.py
```diff
@@ -925,63 +894,76 @@ def update_task_proxies(self, updated_tasks=None):
...
-                )
-                self.updated[FAMILY_PROXIES].setdefault(
-                    fp_id, PbFamilyProxy()).MergeFrom(fp_delta)
+        child_fam_nodes = [
+            n_id
+            for n_id in fam_node.child_families
+            if n_id not in self.updated_state_families
+        ]
+        for child_fam_id in child_fam_nodes:
+            self._family_ascent_point_update(child_fam_id)
```
oliver-sanders, line 924: This recursively works its way to the bottom of the tree then works its way up to `fp_id`?

Would it be possible to do this the other way around, starting at the task which has changed state and bubbling that update up the family hierarchy?

dwsutherland: > This recursively works its way to the bottom of the tree then works its way up to fp_id?

Not quite, it starts from whatever first parent family task(s) flag to be updated..

> Would it be possible to do this the other way around, starting at the task which has changed state and bubbling that update up the family hierarchy?

Yes, that's exactly how it works. The updated tasks add their first parent to `self.state_update_families`, however, if it's a middle family then you need to check if any child families are also in `self.state_update_families`  (that's the recursive bit)... This Means all the families are checked (which is a super fast set lookup), but only those that actually need updating are changed (deltas created).

cylc/flow/data_store_mgr.py
```diff
@@ -925,63 +894,76 @@ def update_task_proxies(self, updated_tasks=None):
...
-            cycle_points = self.delta_cycle_states.keys()
-        if not cycle_points:
+    def update_family_proxies(self):
+        self.updated_state_families.clear()
+        while self.state_update_families:
+            for fam_id in self.state_update_families:
+                self._family_ascent_point_update(fam_id)
+                break
+
+    def _family_ascent_point_update(self, fp_id):
```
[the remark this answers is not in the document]
dwsutherland: I agree.

cylc/flow/data_store_mgr.py
```diff
@@ -925,63 +894,76 @@ def update_task_proxies(self, updated_tasks=None):
...
-        family_proxies = self.data[self.workflow_id][FAMILY_PROXIES]
-        if cycle_points is None:
-            cycle_points = self.delta_cycle_states.keys()
-        if not cycle_points:
+    def update_family_proxies(self):
+        self.updated_state_families.clear()
+        while self.state_update_families:
+            for fam_id in self.state_update_families:
+                self._family_ascent_point_update(fam_id)
+                break
```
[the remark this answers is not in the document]
dwsutherland: Probably a better way of doing it .. However:
- `self.state_update_families` is modified within the recursive function, so can't loop on it directly.
- I just want to start on any member of `self.state_update_families`, it doesn't matter which one, and I want to keep it in the set (i.e. not use `pop()`)..

I'm open, if you have any better ideas.. But you can loop on a set you modify, and I don't want to make a copy of it, because that could mean covering the same ground twice..

dwsutherland:
Just re-post for all:

The updated tasks add their first parent to `self.state_update_families`, however, if it's a middle family then you need to check if any child families are also in `self.state_update_families` and update those first, and so on (that's the recursive bit)... This Means all the families are checked (which is a super fast set lookup), but only those that actually need updating are changed (deltas created).

hjoliver (commented):
(I had a quick look at this today, but need to get my head back in the datastore zone to understand it ...)

cylc/flow/data_store_mgr.py
oliver-sanders: Ah, heck, I don't quite understand that? Might need an example.

> Not quite, it starts from whatever first parent family task(s) flag to be updated..

What does "first parent family task" mean?

> if it's a middle family then you need to check if any child families are also in `self.state_update_families`

I think I'm misunderstanding something here. For the following inheritance hierarchy:

* root
  * FOO
    * foot
    * BAR
        * bart

I would expect that:

* If `foot` changed `FOO` should be recomputed.
* If `bar` changed `FOO` and `BAR` should be recomputed.

In this example is `FOO` a middle family?

oliver-sanders: Ah, ok, I think I've been thick, the `child_fam_nodes ` are taken from `self.updated_state_families` so this is just an efficiency improvement to avoid recomputing the family tree twice?

dwsutherland: Sorry "first-parent-family that task(s)"

dwsutherland: > the `child_fam_nodes` are taken from `self.updated_state_families`

The `child_fam_nodes` are taken from `self.state_update_families`  nodes (`PbFamilyProxy.child_families`), and filtered to not include those already updated (`self.updated_state_families`). The filtering avoids re-computation, but I could probably broaden the filtered ones to any family that has been checked (not just the updated ones).

dwsutherland: > but I could probably broaden the filtered ones to any family that has been checked (not just the updated ones).

Actually I'm not sure it's possible to check the same family twice as is.. So no need to widen net, if a child family is checked but not updated it's parent (or ancestor) would have been checked (added to `self.updated_state_families`), so it won't be hit again.

hjoliver:
> [ ]  Add tasks per state to each family and workflow.

Should that check-box be checked, above?

dwsutherland:
> > [ ]  Add tasks per state to each family and workflow.
> 
> Should that check-box be checked, above?

Yeah, I don't know.. This was the Element discussion about the summary panel hover-over, but it was inconclusive what exactly we want to show..

dwsutherland:
> > > [ ]  Add tasks per state to each family and workflow.
> > 
> > 
> > Should that check-box be checked, above?
> 
> Yeah, I don't know.. This was the Element discussion about the summary panel hover-over, but it was inconclusive what exactly we want to show..

Removed... Separate PR for that (after further Element discussion)

kinow (approved):
UI working fine. Had a look at the code, found nothing that called my attention. Thanks!

image

(tested with families2 too, to see if it would cause any issues having multiple families nested, still OK! :tada: )

hjoliver (approved):
LGTM :+1:   @oliver-sanders can merge if happy with the responses to his questions...

hjoliver:
It'd be good to get your diagram in as code documentation somehow (in-comment ascii flow-chart?)

hjoliver:
(meh, maybe a text description will do).

dwsutherland:
> It'd be good to get your diagram in as code documentation somehow (in-comment ascii flow-chart?)

I need to learn more inkscape (>.<)

hjoliver:
Nothing wrong with a photo of a hand sketch!

cylc/flow/data_store_mgr.py
oliver-sanders: Ok, I think I have a rough idea what it's doing now, thanks for clearing it up.

PR #3789, opened 2020-08-25
Tidy cfgspec/parsec slightly

MetRonnie:
Update: I realised the original aim of this PR wasn't practical. Instead I've just tidied the code & comments ever so slightly.

**Requirements check-list**
- [X] Does not need tests (no real changes).
- [X] No change log entry required.
- [X] No dependency changes.

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -134,7 +134,6 @@ with Conf('global.cylc', desc='''
                 Conf('interval', VDR.V_INTERVAL, desc='''
                     The interval with which this plugin is run.
                 ''')
-                Conf('foo', VDR.V_STRING, default='X')
 
             with Conf('health check', meta=MainLoopPlugin, desc='''
                 Checks the integrity of the suite run directory.
--- a/cylc/flow/parsec/OrderedDict.py
+++ b/cylc/flow/parsec/OrderedDict.py
@@ -22,20 +22,18 @@ from collections import OrderedDict
 
 class OrderedDictWithDefaults(OrderedDict):
 
-    """Subclass to provide defaults fetching capability.
+    """Subclass to provide defaults fetching capability."""
 
-    Note that defining a '__missing__' method would work for foo[key],
-    but doesn't for foo.get(key).
-
-    """
+    # Note that defining a '__missing__' method would work for foo[key],
+    # but doesn't for foo.get(key).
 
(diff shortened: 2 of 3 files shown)


MetRonnie:
Something isn't right, I'm looking into it

oliver-sanders:
`None` is basically the default anyway, technically the config node defaults to unset but the end result is the same no?

MetRonnie:
There is still the issue of: if you have `UTC mode = True` in your global config and you want to unset it to the default in the flow config, well, you can't exactly unset it. However, as Oliver has pointed out, just setting `UTC mode = False` in flow config is probably good enough? (UTC mode is the only bool setting in both flow & global cfgspec)

wxtim (approved):
Happy with this.

cylc/flow/parsec/OrderedDict.py
```diff
@@ -95,12 +93,9 @@ def __bool__(self):
...
     def prepend(self, key, value):
-        """Prepend new item in the ordered dict.
-
-        https://stackoverflow.com/questions/16664874/
-           how-can-i-add-an-element-at-the-top-of-an-ordereddict-in-python
-
-        """
+        """Prepend new item in the ordered dict."""
+        # https://stackoverflow.com/questions/16664874/
+        #   how-can-i-add-an-element-at-the-top-of-an-ordereddict-in-python
```
oliver-sanders, line 98: I'm confused why are you converting docstrings to comments?

What's the gain.

MetRonnie: Not much gain, but I figure those don't describe what the methods do, so they're not suitable to be docstrings

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -134,7 +134,6 @@
                 Conf('interval', VDR.V_INTERVAL, desc='''
                     The interval with which this plugin is run.
                 ''')
-                Conf('foo', VDR.V_STRING, default='X')
```
hjoliver, line 137: Huh? :thinking:

hjoliver (approved):
Looks fine to me; Oliver can merge if happy with your docstring/comment response.

cylc/flow/cfgspec/globalcfg.py
oliver-sanders: Yeah, not sure how that one snuck in.

PR #3790, opened 2020-08-26
Platforms.fix tests

wxtim:
I missed a set of tests (mainly those dependent on batch systems that I don't have immediate access to) when I created the platforms work. This PR attempts to remedy this failure.

Additionally, I messed up tests `tests/functional[job kill|directives]/[loadleveler|slurm|lsf|pbs]` because in many cases these have a single test with different `flow.cylc` files for each batch system: I had hardcoded the batch system name into the `global.cylc` file created in the base test which would have broken all the tests which were symlinked to it.

Review -
Unless you happen to have the batch system in question set up locally you probably cannot check that these tests actually work so this review may need to be a proof-read and sanity check only. (Slurm tests work at the Met Office)

### DIFF
--- a/tests/functional/cylc-poll/06-loadleveler.t
+++ b/tests/functional/cylc-poll/06-loadleveler.t
@@ -35,9 +35,9 @@ set_test_number 2
 
 create_test_global_config "" "
 [platforms]
-    [[loadleveler-platform]]
+    [[lsf-platform]]
         hosts = $CYLC_TEST_BATCH_TASK_HOST
-        batch system = loadleveler
+        batch system = lsf
 "
 
 reftest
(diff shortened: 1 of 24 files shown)


tests/functional/cylc-poll/06-loadleveler.t
```diff
@@ -35,9 +35,9 @@ set_test_number 2
 
 create_test_global_config "" "
 [platforms]
-    [[loadleveler-platform]]
+    [[${BATCH_SYS_NAME}-test-platform]]
```
wxtim, line 38: Tests for Slurm and PBS symlinked to this file so it needs to provide platforms.

hjoliver (approved):
Looks sensible. Untested here, but  tests only and can fix with a small follow-up in the event of disaster.

Issue #3672, opened 2020-07-02
platforms: allow sub-shell platform names


wxtim:
## Cylc 7 Behaviour
At Cylc 7 users could set `host = $(myFunction)` for a test, and `myFunction` would be evaluated by `cylc.flow.TaskJobManager.task_remote_mgr.remote_host_select()` (`task_remote_mgr` is a `TaskRemoteMgr` object).

## (Desired) Cylc 8 Behaviour
At Cylc 8 we want the platform value to be evaluated in the same way.

wxtim:
Closed with merger of #3791

PR #3791, opened 2020-08-26
Platforms.eval platform cmd

wxtim:
main response to #3672 

## Synopsis
As per the [deprecation examples in the platforms proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-platforms.md#deprecation-examples) Cylc 8 is to be able to evaluate either:
* `[TASK]platform = $(some-shell-script.sh)`
* `[TASK][remote]host = $(some-shell-script.sh)`
In a similar way at job submission time.

This PR attempts to create this behavior.

## Work undertaken
- [x] Modified the logic to work with an example suite.
- [x] Written functional tests for each of the main scenarios this logic should handle:
    - Not interfering with `platform = name` and `host = name`.
    - Correctly expanding subshell host and platform settings.
    - Throwing a wobbly (raise Exception) if one tries to set both host and platform before attempting to expand subshells to avoid unnecessary calls to `remote_host_select`.

## Follow on work not undertaken
- No attempt to update nomenclature of items in `remote_host_select` - should be straightforward, but would confuse the issue.
- No attempt to stop using backticks for platform.


cylc/flow/task_job_mgr.py
```diff
@@ -798,10 +798,39 @@ def _prep_submit_task_job(self, suite, itask, check_syntax=True):
...
+        #    by the platforms module it's probably worth putting it here too
+        #    to prevent trying to run the remote_host/platform_select logic for
+        #    tasks which will fail anyway later.
+        # - Platform exists, host doesn't = eval platform_n
+        # - host exists - eval host_n
+        if (
+            rtconfig['platform'] is not None and
+            rtconfig['remote']['host'] is not None
+        ):
+            raise PlatformLookupError(
```
wxtim, line 818: This should be a validation error

tests/functional/validate/74-host-to-platform-upgrade-fail-inherit.t
```diff
@@ -15,9 +15,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #-------------------------------------------------------------------------------
-# Check that platform upgrader will fail if, after inheritance but not
-# before a task has both old and new settings - This should be a fail on
-# Job Submit.
+# Parent and child tasks are both valid, before inheritance calculated.
```
wxtim, line 18: Unrelated, but not worth a full, separate PR.

cylc/flow/platforms.py
```diff
@@ -30,6 +31,9 @@
     ('job', 'batch submit command template', [None])
 )
 
+# Regex to check whether a string is a command
+REC_COMMAND = re.compile(r'(`|\$\()\s*(.*)\s*([`)])$')
```
oliver-sanders, line 35: Looks fine, it can be broken but only for facetious examples like:

```
1) $(foo`
2) `foo)
```

Which probs isn't an issue.

If testing for the end of the line, best to permit arbitrary whitespace:

```suggestion
REC_COMMAND = re.compile(r'(`|\$\()\s*(.*)\s*([`)])\s*$')
```

Consider testing for the start of the line too:

```suggestion
REC_COMMAND = re.compile(r'^\s*(`|\$\()\s*(.*)\s*([`)])$')
```

cylc/flow/platforms.py
```diff
@@ -63,10 +70,21 @@ def get_platform(task_conf=None, task_id='unknown task', warn_only=False):
...
     elif 'platform' in task_conf and task_conf['platform']:
+        if REC_COMMAND.match(task_conf['platform']) and warn_only:
+            # In warning mode this function might have been passed an
+            # un-expanded platform string - warn that they won't deal with
+            # with this until job submit.
+            LOG.warning(
+                f'Cannot attempt check validity of [runtime][{task_id}]'
+                f'platform = {task_conf["platform"]}: Platform lookup may fail'
+                ' on job-submit.'
+            )
```
oliver-sanders, line 81:
```suggestion
```

oliver-sanders (commented):
Looks fine, testing to follow.

cylc/flow/platforms.py
wxtim: This was exactly what was used before, but happy to change it.

cylc/flow/platforms.py
wxtim: Why?

wxtim:
> tests/f/job-submission/18-platform_select.t

My fault - Dave suggested that I was raising the wrong error, so I changed the error I raised...

cylc/flow/platforms.py
oliver-sanders: If I'm on the right track - this message would appear for every task which uses the subshell syntax. Saying that platform lookup may fail on job submit is kinda just documenting what this subshell approach does, but once for each task and in a way that few people will likely understand.

tests/functional/job-submission/15-garbage-host-command-2.t
```diff
@@ -17,13 +17,12 @@
...
 # Test recovery of a failed host select command for a group of tasks.
 . "$(dirname "$0")/test_header"
-skip_all "TODO replace this test with one checking that garbage-platform commands are handled ok"
 set_test_number 2
 
 create_test_global_config "
 [platforms]
     [[test platform]]
-        hosts = $(my-host-select)
+        hosts = localhost
```
oliver-sanders, line 25: Debug change?

hjoliver:
One test failing - fixed by #3806?

tests/functional/job-submission/15-garbage-host-command-2.t
wxtim: Nope, just a mistake. I don't think that this will work at all, because we don't evaluate this.

wxtim:
> One test failing - fixed by #3806?

Looks very much like it. 👍🏻 

Can I request re-review from @oliver-sanders  and @hjoliver

oliver-sanders:
Closes #3672?

oliver-sanders:
Looks good, testing I spotted some unexpected behaviour:

A non existent platform will result in a `PlatformLookupError` being raised which causes the suite to shutdown, I would have expected a submit failure:

```ini
platform = $(echo 'elephant')
```

```
2020-09-15T12:54:30+01:00 ERROR - PlatformLookupError: No matching platform "elephant" found
```

There seems to be some unintended pattern matching going on, for instance `localhost-xyz-mess` seems to get evaluated as `localhost` rather than resulting in a submit failure:

```ini
platform = $(echo 'localhost-xyz-mess')
```

```
2020-09-15T12:53:40+01:00 DEBUG - for task a.1: platform = $(echo 'localhost-xyz-mess') evaluated as localhost
```

cylc/flow/task_job_mgr.py
```diff
@@ -693,14 +695,15 @@ def _run_job_cmd(self, cmd_key, suite, itasks, callback):
...
-    def _set_retry_timers(itask, rtconfig=None):
+    def _set_retry_timers(itask, rtconfig=None, retry=True):
         """Set try number and retry delays."""
         if rtconfig is None:
             rtconfig = itask.tdef.rtconfig
-        try:
-            _ = rtconfig[itask.tdef.run_mode + ' mode']['disable retries']
-        except KeyError:
-            retry = True
+        if (
```
wxtim, line 702: This should be faster since in most cases the try was going to fail.

wxtim:
> A non existent platform will result in a PlatformLookupError being raised which causes the suite to shutdown, I would have expected a submit failure:

I think I've fixed this in commit https://github.com/cylc/cylc-flow/pull/3791/commits/3175feeee4bac8e43d30ae7777592eb656ddd6ae

oliver-sanders (commented):
Getting a submit failure now for `platform=$(echo elephant)` 👍. Sadly also getting this traceback:

```
2020-09-15T15:55:08+01:00 ERROR - [Errno 2] No such file or directory:
	'/Users/oliver/cylc-run/foo/log/job/1/foo/01/job-activity.log'
	  File "/Users/oliver/cylc-flow/cylc/flow/task_job_mgr.py", line 865, in
	_prep_submit_task_job
	    platform = get_platform(rtconfig)
	  File "/Users/oliver/cylc-flow/cylc/flow/platforms.py", line 82, in get_platform
	    output = platform_from_name(task_conf['platform'])
	  File "/Users/oliver/cylc-flow/cylc/flow/platforms.py", line 184, in
	platform_from_name
	    f"No matching platform \"{platform_name}\" found")
	cylc.flow.exceptions.PlatformLookupError: No matching platform "elephant" found
	  File "/Users/oliver/cylc-flow/cylc/flow/task_events_mgr.py", line 87, in
	log_task_job_activity
	    with open(os.path.expandvars(job_activity_log), "ab") as handle:
	FileNotFoundError: [Errno 2] No such file or directory: '/Users/oliver/cylc-
	run/foo/log/job/1/foo/01/job-activity.log'
```

I think the old logic may have created the job activity log on failure to avoid this?

wxtim:
> I think the old logic may have created the job activity log on failure to avoid this?

I don't think so, on examination - the logic in `task_events_mgr.log_task_job_activity` says:

```python
    except IOError as exc:
        # This happens when there is no job directory, e.g. if job host
        # selection command causes an submission failure, there will be no job
        # directory. In this case, just send the information to the suite log.
        LOG.exception(exc)
        LOG.info(ctx_str)
```
So it looks to me like it's behaving in the correct way. I could of course modify it to make the file it it's not there already, assuming that there is enough info in this method to do it. 
Alternatively, I could add a bit of logic to fish out platform lookup failures before this point.

wxtim:
Yes, @oliver-sanders  pointed out that at some point I had deleted the line self._create_job_log_path(suite, itask) in task_job_mgr which is required to create the job-activity.log

oliver-sanders (approved):
All good I think.

The log messages contain a bit of duplicated info and could do with being tidied up a bit:

```
DEBUG - ['bash', '-c', 'echo
	localhost']
DEBUG - [remote-host-select cmd] bash -c
	'echo localhost'
	[remote-host-select ret_code] 0
	[remote-host-select out] localhost
DEBUG - for task foo.1: platform =
	$(echo localhost) evaluated as localhost
```

One message should suffice for this case, note the second log message comes from the subprocpool and contains all the required info.

```
DEBUG - Traceback (most recent call
	last):
	  File "/Users/oliver/cylc-flow/cylc/flow/task_job_mgr.py", line 866, in
	_prep_submit_task_job
	    platform = get_platform(rtconfig)
	  File "/Users/oliver/cylc-flow/cylc/flow/platforms.py", line 82, in get_platform
	    output = platform_from_name(task_conf['platform'])
	  File "/Users/oliver/cylc-flow/cylc/flow/platforms.py", line 184, in
	platform_from_name
	    f"No matching platform \"{platform_name}\" found")
	cylc.flow.exceptions.PlatformLookupError: No matching platform "elephant" found

ERROR - No matching platform "elephant" found
ERROR - [jobs-submit cmd] (platform not defined)
	[jobs-submit ret_code] 1
	[jobs-submit err] No matching platform "elephant" found
```

One message should suffice for this case, the traceback isn't helpful here as we already know where the exception was raised from.

hjoliver (approved):
Looks good, thanks @wxtim :+1:

hjoliver:
@wxtim - can  you address the log duplication as a follow-up?

Issue #3686, opened 2020-07-15
platforms: platform aliases


oliver-sanders:
Add support for groups of platforms called platform-aliases e.g:

```ini
[platforms]
    [[hpcl1-bg]]
        hosts = hpcl1
        retrieve job logs = True
        batch system = background
    [[hpcl2-bg]]
        hosts = hpcl2
        retrieve job logs = True
        batch system = background
[platform aliases]
    [[hpc-bg]]
        platforms = hpcl1-bg, hpcl2-bg
```

At its most basic we pick which platform to use from a group randomly.

See the [platforms proposal document](https://cylc.github.io/cylc-admin/proposal-platforms.html#further-enhancements) for more information.

**Extension:**

We might as well integrate the host-select logic currently used for picking suite servers as there is an [existing interface](https://github.com/cylc/cylc-flow/blob/c6c90629546854378a96292226a7a65ee4a9e6a3/cylc/flow/host_select.py#L57-L62
) which was written to fulfil this function e.g:

```ini
[platform aliases]
    [[hpc-bg]]
        platforms = hpcl1-bg, hpcl2-bg
        ranking = """
            cpu_percent()
        """
```

**Pull requests welcome!**

wxtim:
PR #3793 adds the basic functionality described in this issue. The selection method to be addressed in a follow up.

wxtim:
PR #3793 was merged, but leaving this open becuase the advanced selection options are required.

oliver-sanders:
Only the selection component of this issue remains so opening a new issue (#3827) for that and closing this.

PR #3793, opened 2020-08-28
Platforms.aliases

wxtim:
## Synopsis
Part addresses issue #3686 

As per the [platforms proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-platforms.md#further-enhancements) it was agreed that platform aliases should be available. In their most basic form (here) these allow a task with a platform set by the user to be assigned by the site administrator to any of a set of platforms.

## Work Carried Out
- Added the `[platform aliases][ALIAS]platforms` config to the `global.cylc` config spec.
- Used Python's `random.choice` to pick a platform from the list.
- Created a functional test.

## Work not carried out
Additional infrastructure for more sophisticated platform from alias selection.

## Discussion Point
Can anyone suggest a better way to test this. My unit test is horrible, because I cannot see a way to test this functionality nicely.


tests/functional/platform_aliases/00-simple.t
```diff
@@ -0,0 +1,82 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test that a different platform is selected on different tries with the same
+# platform Alias.
+#
+# Because this test can only ever test that all platforms within an alias have
+# been used it is theoretically possible for the random selection from Python's
+# random.choice to continue selecting the same platform forever, which is the
+# same outcome as failure. As a result this test is deliberately slightly flaky:
```
MetRonnie, line 24: Should this not go under `tests/flakyfunctional/` then?

wxtim: Perhaps - I'm hoping that < 1% chance of flaking is probably ok... but happy to move it.

hjoliver (commented):
Looks fine, but I wonder if it might be worth implementing a non-random selection method now to prevent test flakiness - which is likely to be painful even if it occurs infrequently.

hjoliver:
> I wonder if it might be worth implementing a non-random selection method now to prevent test flakiness

One way of doing it would be to create a platform generator the first time an alias is used, then yield the next true platform name each time the alias is used. Then we could implement any selection method quite easily. Just for the hell of it, I tried "random" and "endless cycle" this way (cycling would enable a non-flaky test, and might be a useful real-world option too?) - it seems to work fine, here's the diff in case you want to consider it.  I guess a similar method could be used for host selection within platforms too?
```diff
diff --git a/cylc/flow/platforms.py b/cylc/flow/platforms.py
index 666889fd5..73e991f42 100644
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -19,11 +19,33 @@
 import random
 import re
 from copy import deepcopy
+from itertools import cycle
 
 from cylc.flow.exceptions import PlatformLookupError
 from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
 
 
+platform_generators = {}
+
+
+def random_gen(stuff):
+    while True:
+        yield random.choice(stuff)
+
+
+def choose_platform(alias, platforms, method="random"):
+    method = "cycle"  # TEST ME
+    if alias not in platform_generators:
+        if method == "cycle":
+            gen = cycle(platforms)
+        elif method == "random":
+            gen = random_gen(platforms)
+        platform_generators[alias] = gen
+    else:
+        gen = platform_generators[alias]
+    return next(gen)
+
+
 def platform_from_name(platform_name=None, platforms=None):
     """
     Find out which job platform to use given a list of possible platforms and
@@ -55,9 +77,11 @@ def platform_from_name(platform_name=None, platforms=None):
 
     for platform_name_re in reversed(list(platform_aliases)):
         if re.fullmatch(platform_name_re, platform_name):
-            platform_name = random.choice(
-                platform_aliases[platform_name_re]['platforms']
-            )
+            # platform_name is an alias; replace with a real platform
+            platform_name = choose_platform(
+                platform_name,
+                platform_aliases[platform_name_re]['platforms'])
+            print('>>>>>>>>>>', platform_name)   # DEBUG
 
     # The list is reversed to allow user-set platforms (which are loaded
     # later than site set platforms) to be matched first and override site
```

hjoliver:
(This would require adding `host selection method` to `[platform aliases]` - but I guess we plan to do that anyway)

wxtim:
This seems like a reasonable template @hjoliver - and a good way around the flaky test. Thanks. I'll get on it.

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -469,7 +469,7 @@
             Conf('hosts', VDR.V_STRING_LIST, ['localhost'])
 
     # Platform Groups
-    with Conf('platform groups'):
+    with Conf('platform aliases'):
```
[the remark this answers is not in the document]
oliver-sanders: Yes, you're right, platform groups are listed in the "upgraded config" [here](https://cylc.github.io/cylc-admin/proposal-config-changes.html#updated-global-config).

tests/flakyfunctional/platform_aliases/00-simple.t
```diff
@@ -0,0 +1,80 @@
...
+. "$(dirname "$0")/test_header"
+require_remote_platform
+set_test_number 2
+
+install_suite "${TEST_NAME_BASE}" "${TEST_NAME_BASE}"
+LOCAL_HOSTNAME=$(hostname)
+create_test_global_config "" "
+[platforms]
+  [[platform1]]
+    hosts = ${LOCAL_HOSTNAME}
```
oliver-sanders, line 39:
```suggestion
    hosts = localhost
```
?

tests/flakyfunctional/platform_aliases/00-simple.t
```diff
@@ -0,0 +1,80 @@
...
+  cylc trigger "${SUITE_NAME}" task_with_platform_alias.1
+  if [[ $i -gt 1 ]]; then
+    dbinfo=$(sqlite3 "${DBPATH}" "SELECT name, platform_name from task_jobs")
+    if grep -q 'platform1' <<< "$dbinfo" && grep -q 'platform2' <<< "$dbinfo"; then
+      ok "${TEST_NAME_BASE}-all-platforms-used"
+      break
+    fi
+    if [[ $i > "${TASTE_FOR_FLAKINESS}" ]]; then
+      if [[ -n "${TEST_VERBOSE:-}" ]]; then
+        echo "There is a 1 in 2/2^${TASTE_FOR_FLAKINESS} probability that this is a false-negative" >&2
```
oliver-sanders, line 68: Statistic approach, cool.

Not the nicest test in the world but I can't think of a better approach for functional testing.

cylc/flow/platforms.py
```diff
@@ -46,12 +46,19 @@ def platform_from_name(platform_name=None, platforms=None):
...
     if platforms is None:
         platforms = glbl_cfg().get(['platforms'])
+    platform_aliases = glbl_cfg().get(['platform aliases'])
 
     if platform_name is None:
         platform_data = deepcopy(platforms['localhost'])
         platform_data['name'] = 'localhost'
         return platform_data
 
+    for platform_name_re in reversed(list(platform_aliases)):
```
oliver-sanders, line 56: Funny thought, what if someone defines both a platform and a platform group with the same name?

* Should we validate platform groups to prevent this from occurring.
* Otherwise the approach used here (go through platform groups before platforms) seems sensible to me.

oliver-sanders (commented):
Looks good.

tests/flakyfunctional/platform_aliases/00-simple.t
wxtim: @hjoliver suggested a (sort-of) solution which I'm going to implement.

cylc/flow/platforms.py
wxtim: There are cases where this might be reasonable:
```
[platform aliases]
  # Site user prevents anyone specifying a login node (unless they do it explicitly by creating their own platform)
  [[hpc\d?]]
    platforms = hpc1, hpc2   #, hpc3 is out of service, but everyone's hp3 tasks work anyway!
    ...
```
...but equally allows wonderful brainaches...
```
[platform aliases]
  [[chocolate]]
    platforms = vanilla
  [[vanilla]]
    platforms = raspberry
  [[raspberry]]
    platforms = chocolate
```

cylc/flow/cfgspec/globalcfg.py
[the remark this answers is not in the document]
wxtim: Right then - I'll fix that.

wxtim:
> This would require adding host selection method to [platform aliases]

No, this would be platform selection method - host selection method would be added to the platform...

oliver-sanders:
> One way of doing it would be to create a platform generator the first time an alias is used

Nice simple approach, a `cycle` option would be great for testing.

There's just one barrier to overcome which is that these generators would be long lived? Which means that the list of hosts provided to the generator would be fossilised for the life of the workflow run. This gets in the way of automatically reloading the global config at regular intervals would would enable site admins to, say, remove a login node or drain a system.

So I think we need to be passing the list of platforms to the selection method with each call. For a selection method like `cycle` which needs to maintain local state we will have to get creative, how about something like this:

```python
def cycle_selection(_):
    index = 0
    def _cycle_selection(platforms):
        nonlocal index
        return platforms[index % len(platforms)]
        index += 1
    return _cycle_selection

METHODS = {
    'cycle': cycle_selection
}

def choose_platform(alias, platforms, method="random"):
    meth = METHODS[method](platforms)
```

oliver-sanders:
See the related https://github.com/cylc/cylc-flow/issues/3800 for thoughts about integrating the existing Cylc host-selection logic into platforms.

wxtim:
I find myself uncomfortable doing too much jiggery-pokery for a platform selection method I don't think will ever actually be used. I'll have a quick play though.

hjoliver:
> Nice simple approach, a cycle option would be great for testing.
> 
> There's just one barrier to overcome which is that these generators would be long lived? Which means that the list of hosts provided to the generator would be fossilised for the life of the workflow run

Hmm, I kinda thought the generator approach was more elegant. Presumably we could just "reset" a generator by replacement, at the appropriate time:
```python
>>> from itertools import cycle
>>> lunch = ['cat', 'dog', 'sausage']
>>> gen = cycle(lunch)
>>> next(gen)
1
>>> next(gen)
2
>>> gen = cycle(lunch)  # RESET 
>>> next(gen)
1
>>> next(gen)
2
```

However, I haven't actually thought about the mechanics of how to do that in context, and @wxtim has implemented your approach, which works, - so no biggie :+1:

wxtim:
I'm afraid that it all became a bit wierd and complex - @oliver-sanders original fix didn't remember which platform_group was which, and I had trouble getting modified logic working. Since in the current form the test is very, very unlikely (I can make it slower to fail correctly and less likely to give a false negative if you want?) to give a false negative I'd like to commend the PR in its original form to @oliver-sanders  and @hjoliver. 

No reason that we can't have a follow on request.

hjoliver:
Alright, fine with me :+1:

hjoliver (approved):
Looks good, tested as working. :+1: 

Maybe bang up a follow-up issue for better/additional section methods?

oliver-sanders:
FYI: we realised that we are going to require a more advanced platform/host selection interface in order to be able to automatically handle network/system outages.

Here's a purely illustrative example to explain what I mean by this:

```python
def host_from_platform_group(group):
    for platform in group:
        for host in platform['hosts']
            yield host

def submit_job(job):
    for host in host_from_platform_group(job['platform']):
        try:
            submit(job, host)
            break
        except (PlatformLookupError, HostSelectError, socket.gaierror, Timeout):
            continue
```

wxtim:
> FYI: we realised that we are going to require a more advanced platform/host selection interface in order to be able to automatically handle network/system outages.

Agreed, but can we get this mergéd?

oliver-sanders (approved):

hjoliver:
> FYI: we realised that we are going to require a more advanced platform/host selection interface in order to be able to automatically handle network/system outages.

Is there a follow-up issue up for this?

wxtim:
> Is there a follow-up issue up for this?

https://github.com/cylc/cylc-flow/issues/3827

PR #3795, opened 2020-08-29
Fix cylc ping verbose (and a few minor docs improvements)

kinow:

`cylc ping --verbose five` is broken :cry: 

```bash
(venv) kinow@ranma:~/Development/python/workspace/cylc-flow$ cylc ping five --verbose
2020-08-29T23:25:03+12:00 DEBUG - Loading site/user config
	files
2020-08-29T23:25:03+12:00 DEBUG - zmq:send {'command':
	'ping_suite', 'args': {}, 'meta': {'prog': 'cylc-ping', 'host': 'ranma'}}
2020-08-29T23:25:03+12:00 DEBUG - zmq:recv {'data': True,
	'user': 'kinow'}
  File "/home/kinow/Development/python/workspace/cylc-flow/venv/bin/cylc-ping", line 11, in <module>
    load_entry_point('cylc-flow', 'console_scripts', 'cylc-ping')()
  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/terminal.py", line 183, in wrapper
  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/scripts/cylc_ping.py", line 51, in main
    host, port = SuiteRuntimeClient.get_location(suite)
AttributeError: type object 'SuiteRuntimeClient' has no attribute 'get_location'
```

This PR fixes it, also adding missing docstrings, fixing some typos, and applying a few more suggestions from a linter.

```bash
(venv) kinow@ranma:~/Development/python/workspace/cylc-flow$ cylc ping five --verbose
2020-08-29T23:30:01+12:00 DEBUG - Loading site/user config files
2020-08-29T23:30:01+12:00 DEBUG - zmq:send {'command': 'ping_suite', 'args': {}, 'meta': {'prog': 'cylc-ping', 'host': 'ranma'}}
2020-08-29T23:30:01+12:00 DEBUG - zmq:recv {'data': True, 'user': 'kinow'}
Running on ranma:43082
```

**Requirements check-list**
- [x] Does not need tests (why? let me know if others prefer to have a functional test for this, I thought it would be OK to skip it).
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/network/__init__.py
+++ b/cylc/flow/network/__init__.py
@@ -67,8 +67,6 @@ def get_location(suite: str):
 
     Args:
         suite (str): suite name
-        owner (str): owner of the suite
-        host (str): host name
     Returns:
         Tuple[str, int, int]: tuple with the host name and port numbers.
     Raises:
--- a/cylc/flow/network/client.py
+++ b/cylc/flow/network/client.py
@@ -74,6 +74,10 @@ class SuiteRuntimeClient(ZMQSocketBase):
             the contact file.
 
     Attributes:
+        host (str):
+            Suite host name.
+        port (int):
+            Suite host port.
         timeout_handler (function):
             Optional function which runs before ClientTimeout is raised.
             This provides an interface for raising more specific exceptions in
@@ -121,6 +125,8 @@ class SuiteRuntimeClient(ZMQSocketBase):
             host, port, _ = get_location(suite)
         else:
             port = int(port)
+        self.host = host
+        self.port = port
         if timeout is None:
             timeout = self.DEFAULT_TIMEOUT
         else:
@@ -130,7 +136,7 @@ class SuiteRuntimeClient(ZMQSocketBase):
             self._timeout_handler, suite, host, port)
         self.poller = None
         # Connect the ZMQ socket on instantiation
-        self.start(host, port, srv_public_key_loc)
+        self.start(self.host, self.port, srv_public_key_loc)
         # gather header info post start
         self.header = self.get_header()
 
--- a/cylc/flow/network/resolvers.py
+++ b/cylc/flow/network/resolvers.py
@@ -422,7 +422,7 @@ class BaseResolvers:
             raise
         except Exception:
             import traceback
-            logger.warn(traceback.format_exc())
+            logger.warning(traceback.format_exc())
         finally:
             for w_id in w_ids:
                 if delta_queues.get(w_id, {}).get(sub_id):
--- a/cylc/flow/scripts/cylc_ping.py
+++ b/cylc/flow/scripts/cylc_ping.py
@@ -48,7 +48,7 @@ def main(parser, options, suite, task_id=None):
     # cylc ping SUITE
     pclient('ping_suite')  # (no need to check the result)
     if cylc.flow.flags.verbose:
-        host, port = SuiteRuntimeClient.get_location(suite)
+        host, port = pclient.host, pclient.port
         sys.stdout.write("Running on %s:%s\n" % (host, port))
     if task_id is None:
         sys.exit(0)
--- a/cylc/flow/xtrigger_mgr.py
+++ b/cylc/flow/xtrigger_mgr.py
@@ -122,7 +122,6 @@ class XtriggerManager:
         proc_pool: SubProcPool = None,
         suite_run_dir: str = None,
         suite_share_dir: str = None,
-        suite_work_dir: str = None,
         suite_source_dir: str = None,
     ):
         """Initialize the xtrigger manager.
--- a/.github/workflows/bash.yml
+++ b/.github/workflows/bash.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install Cylc dependencies in the container
         run: |
           docker exec bash python3.7 -m pip install six==1.12
-          docker exec -w /root/cylc-flow bash python3.7 -m pip install -e .[all]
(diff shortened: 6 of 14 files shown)


cylc/flow/network/__init__.py
```diff
@@ -66,8 +66,6 @@ def get_location(suite: str):
 
     Args:
         suite (str): suite name
-        owner (str): owner of the suite
-        host (str): host name
```
kinow, line 70: Not used anymore. Now the `get_location` function handles that.

cylc/flow/network/resolvers.py
```diff
@@ -422,7 +422,7 @@ async def subscribe_delta(self, root, info, args):
             raise
         except Exception:
             import traceback
-            logger.warn(traceback.format_exc())
+            logger.warning(traceback.format_exc())
```
kinow, line 425: `.warn` is deprecated, and actually calls `warning`.

cylc/flow/network/client.py
```diff
@@ -103,17 +109,17 @@ def __init__(
     ):
         super().__init__(zmq.REQ, context=context)
         self.suite = suite
-        host, port, _ = get_location(suite)
+        self.host, self.port, _ = get_location(suite)
```
kinow, line 112: Note, I did this, so that in the `cylc ping` command, we don't have to call `get_location` twice, once when creating the `SuiteRuntimeClient`, and later again to retrieve the `host` and `port` for logging.

cylc/flow/scripts/cylc_ping.py
```diff
@@ -48,7 +48,7 @@ def main(parser, options, suite, task_id=None):
     # cylc ping SUITE
     pclient('ping_suite')  # (no need to check the result)
     if cylc.flow.flags.verbose:
-        host, port = SuiteRuntimeClient.get_location(suite)
+        host, port = pclient.host, pclient.port
```
kinow, line 51: I've stored these `host` and `port` in the `SuiteRuntimeClient`. We create an instance of this class as `pclient` (above in this file's diff :arrow_up:) which will call `get_location`. So I tried to avoid calling that function again as that doesn't appear to be cached.

Happy to revert and use `get_location` if others prefer`. This is to fix the runtime error when running `cylc ping --verbose $SUITE`…

cylc/flow/xtrigger_mgr.py
```diff
@@ -122,7 +122,6 @@ def __init__(
         proc_pool: SubProcPool = None,
         suite_run_dir: str = None,
         suite_share_dir: str = None,
-        suite_work_dir: str = None,
```
kinow, line 125: Not used. These arguments must be created with the key (`, *, ...`), which helped searching for its usage. Appeared to be safe to remove :+1:

hjoliver (approved):
Thanks for the general tidy-up too :+1: 

Re: tests, you could just upgrade the existing `cylc ping` functional test:

```diff
diff --git a/tests/functional/cylc-ping/00-simple/flow.cylc b/tests/functional/cylc-ping/00-simple/flow.cylc
index 754d13a90..972e093b9 100644
--- a/tests/functional/cylc-ping/00-simple/flow.cylc
+++ b/tests/functional/cylc-ping/00-simple/flow.cylc
@@ -3,6 +3,9 @@
         R1 = "foo => bar"
 [runtime]
     [[foo]]
-        script = cylc ping $CYLC_SUITE_NAME
+        script = """
+           cylc ping $CYLC_SUITE_NAME
+           cylc ping --verbose $CYLC_SUITE_NAME
+           """
     [[bar]]
         script = [[ ! $(cylc ping $CYLC_SUITE_NAME-non-existent) ]]
```

kinow:
> Thanks for the general tidy-up too +1
> 
> Re: tests, you could just upgrade the existing `cylc ping` functional test:
> 
> ```diff
> diff --git a/tests/functional/cylc-ping/00-simple/flow.cylc b/tests/functional/cylc-ping/00-simple/flow.cylc
> index 754d13a90..972e093b9 100644
> --- a/tests/functional/cylc-ping/00-simple/flow.cylc
> +++ b/tests/functional/cylc-ping/00-simple/flow.cylc
> @@ -3,6 +3,9 @@
>          R1 = "foo => bar"
>  [runtime]
>      [[foo]]
> -        script = cylc ping $CYLC_SUITE_NAME
> +        script = """
> +           cylc ping $CYLC_SUITE_NAME
> +           cylc ping --verbose $CYLC_SUITE_NAME
> +           """
>      [[bar]]
>          script = [[ ! $(cylc ping $CYLC_SUITE_NAME-non-existent) ]]
> ```

Great idea! Done. Thanks @hjoliver . Will fix the changes conflict now too.

hjoliver:
Not sure why the bash-docker tests failed here @kinow ?

kinow:
@hjoliver reproduced locally. Curiously, the `pkg_resources` module is not able to import `cylc-flow` when I install that with `pip install -e .`. It works when I install with `pip install .`. :thinking: 

Should I just update it to `pip install .` here @hjoliver , or would you like to understand why that happened? I only rebased it, and added the test you suggested (which I don't think is related to this error).

hjoliver:
> Should I just update it to `pip install .` 

I think so, we don't need `-e` in this context anyway.

kinow:
> > Should I just update it to `pip install .`
> 
> I think so, we don't need `-e` in this context anyway.

Done, and build passing now :+1:

dwsutherland (approved):
Looks sensible to me, and functions 👍 

(Will likely cause me conflicts #3757 , but minor)

PR #3755, opened 2020-08-05
conf: change suite.rc to flow.cylc

MetRonnie:
These changes partially address #3689 

Best to look at this one commit by commit. The commits called "Wider rename..." usually consist of the same change done across many files, so you probably only need to take a look at a couple of the files in those.

### Summary
- `suite.rc --> flow.cylc`
- `flow.rc & global.rc --> global.cylc`
- The directory `log/suiterc/ --> log/flow-config/`
- Generally, `suiterc` in variable names `--> flow_config`, and generally `rc` in variable names `--> config`
  - Functional tests: `test_header` function `create_test_globalrc --> create_test_global_config`
  - Integration tests: (in `tests/i/utils/flow_writer.py`): `flow_writer.suiterc() --> flow_writer.flow_config_str()`
  - When referring to a file or filename/path (rather than the config itself), generally `suiterc --> flow_file`
- Every other `*.rc --> *.cylc`

### Backwards compatibility
- The first time a user `cylc run`s a directory that contains only `suite.rc`, it will automatically create a symlink `flow.cylc` to that `suite.rc`.
- If a user `cylc run`s an already-registered suite/workflow and `flow.cylc` doesn't exist (i.e. suite/workflow was run on a previous version of Cylc), it will re-register 

(No backwards compatibility for flow.rc)

---

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).

- [x] Tests for "upgrading" existing `suite.rc` to `flow.cylc` included

- [x] Appropriate change log entry included.

- [x] (master branch) I have opened a documentation PR at https://github.com/cylc/cylc-doc/pull/142

- [x] No dependency changes.

PR #3797, opened 2020-09-01
async util: handle empty dirs better

oliver-sanders:
Just noticed a warning which can occur if `scandir` attempts to scan an empty directory.

I think this is harmless ATM as the only use of `scandir` is workflow scanning, if a workflow dir is empty then it's not a workflow dir so can be safely skipped.

Also add a small update post-#3755.

### DIFF
--- a/cylc/flow/async_util.py
+++ b/cylc/flow/async_util.py
@@ -389,6 +389,7 @@ def _scandir(future, path, request):
     """Callback helper for scandir()."""
     future.set_result([
         Path(path, directory.name)
+        # request.result can be None (maybe because of an empty directory in cylc-run)
         for directory in request.result or []
     ])
 
--- a/cylc/flow/async_util.py
+++ b/cylc/flow/async_util.py
@@ -389,7 +389,7 @@ def _scandir(future, path, request):
     """Callback helper for scandir()."""
     future.set_result([
         Path(path, directory.name)
-        for directory in request.result
+        for directory in request.result or []
     ])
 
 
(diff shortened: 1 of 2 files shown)


cylc/flow/async_util.py
```diff
@@ -389,7 +389,7 @@ def _scandir(future, path, request):
     """Callback helper for scandir()."""
     future.set_result([
         Path(path, directory.name)
-        for directory in request.result
+        for directory in request.result or []
```
hjoliver, line 392: Why does this make any difference?  If I put an empty directory in `~cylc-run` then `request.result` is an empty list already.  (And I can seem to

oliver-sanders: I saw an error where (`request.result` was returning `None`) in the log whilst working on cylc-uiserver so put this in to suppress it.

Perhaps it happens when you scan a non-existent directory or something like that.

hjoliver: OK, fair enough.  Probably needs a comment then, because it looks weird (if you expect `request.result` to always be list).

hjoliver, line 392:
```suggestion
        # request.result can be None (maybe because of an empty directory in cylc-run)
        for directory in request.result or []
```

hjoliver: Huh, GitHub let me create a "suggestion" and commit it to the branch :grin:

hjoliver (approved):

cylc/flow/async_util.py
hjoliver: One review will do for this.

hjoliver: Brilliant, my comment was a style violation (line too long) :rofl:

hjoliver: Hot fix coming.

PR #3799, opened 2020-09-02
Write to stream, then flush

kinow:

Found this one yesterday while reviewing @dwsutherland 's #3757 . I think there was some change in `task_message.py` but on different lines (IOW this was not part of his change, so I added a post-it to review this separately later)

I think if you have 2 messages, one with one severity level, and the other with another severity level, then you could miss flushing one of the handle/streams.

**Requirements check-list**
- [x] Does not need tests (I think it's covered by tests, although we could write a unit test with a mock... if necessary?).
- [x] No dependency changes.


oliver-sanders (approved):
Definitely looks like the logic was flawed before to me.

hjoliver (approved):
:+1:

PR #3802, opened 2020-09-02
Global config hierarchy

### LINKED ISSUE #3689: conf: configuration file changes
As documented in the [config change proposal document](https://cylc.github.io/cylc-admin/proposal-config-changes.html#config-file-naming):

- [x] The `.cylc` extension and the new configuration file names.
  - [x] With support for the deprecated `suite.rc` filename with a warning of some kind for the user.
- [ ] The new global config hierarchy.
- [ ] Any changes to `CYLC_CONF_PATH` required.
- [ ] The new `CYLC_SITE_CONF_PATH` variable.

**Pull requests welcome!**

MetRonnie:
These changes close #3689

> Cylc will attempt to load the global configuration (``global.cylc``) from a hierarchy of locations, including the site directory (defaults to ``/etc/cylc/flow/``) and the user directory at ``~/.cylc/flow/``. E.g. for Cylc version 8.0.1, the hierarchy would be, in order of ascending priority:
>
> * ``${CYLC_SITE_CONF_PATH}/global.cylc``
> * ``${CYLC_SITE_CONF_PATH}/8/global.cylc``
> * ``${CYLC_SITE_CONF_PATH}/8.0/global.cylc``
> * ``${CYLC_SITE_CONF_PATH}/8.0.1/global.cylc``
> * ``~/.cylc/flow/global.cylc``
> * ``~/.cylc/flow/8/global.cylc``
> * ``~/.cylc/flow/8.0/global.cylc``
> * ``~/.cylc/flow/8.0.1/global.cylc``
>
> A setting in a file lower down in the list will override the same setting from those higher up (but if a setting is present in a file higher up and not in any files lower down, it will not be overridden).
>
> To override the default configuration path, set the ``CYLC_CONF_PATH`` environment variable to the directory containing your ``global.cylc`` file. If this is set, the hierarchy above is ignored.
>
> Setting the ``CYLC_SITE_CONF_PATH`` environment variable overrides the default value of ``/etc/cylc/flow/``, without ignoring the hierarchy.

**Requirements check-list**

- [ ] Appropriate tests are included (unit and/or functional) - kind of 😛 

- [X] (master branch) I have opened a documentation PR at https://github.com/cylc/cylc-doc/pull/156

- [X] No dependency changes.

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -790,10 +790,10 @@ class GlobalConfig(ParsecConfig):
                     if conf_type == upgrader.SITE_CONFIG:
                         # Warn on bad site file (users can't fix it).
                         LOG.warning(
-                            'ignoring bad %s %s:\n%s', conf_type, fname, exc)
+                            f'ignoring bad {conf_type} {fname}:\n{exc}')
                     else:
                         # Abort on bad user file (users can fix it).
-                        LOG.error('bad %s %s', conf_type, fname)
+                        LOG.error(f'bad {conf_type} {fname}')
                         raise
 
         self._set_default_editors()
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -42,10 +42,9 @@ with Conf('global.cylc', desc='''
 
        $ cylc get-global-config --sparse
 
-
     Cylc will attempt to load the global configuration (``global.cylc``) from a
     hierarchy of locations, including the site directory (defaults to
-    ``/etc/cylc/flow/``) and the user directory at ``~/.cylc/flow/``. E.g. for
+    ``/etc/cylc/flow/``) and the user directory (``~/.cylc/flow/``). E.g. for
     Cylc version 8.0.1, the hierarchy would be, in order of ascending priority:
 
     * ``${CYLC_SITE_CONF_PATH}/global.cylc``
@@ -61,12 +60,11 @@ with Conf('global.cylc', desc='''
     from those higher up (but if a setting is present in a file higher up and
     not in any files lower down, it will not be overridden).
 
-    To override the default configuration path, set the ``CYLC_CONF_PATH``
-    environment variable to the directory containing your ``global.cylc`` file.
-    If this is set, the hierarchy above is ignored.
-
     Setting the ``CYLC_SITE_CONF_PATH`` environment variable overrides the
-    default value of ``/etc/cylc/flow/``, without ignoring the hierarchy.
+    default value of ``/etc/cylc/flow/``.
+
+    To override the entire hierarchy, set the ``CYLC_CONF_PATH`` environment
+    variable to the directory containing your ``global.cylc`` file.
 
     .. note::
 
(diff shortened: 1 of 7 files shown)


MetRonnie:
I'm not sure how best to test the global config hierarchy

oliver-sanders:
> I'm not sure how best to test the global config hierarchy

* Can test the value of `SITE_CONF_DIR_HIERARCHY`.
* If really going for it could move the logic which loads files into it's own function `load_config(files: list)`.

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -775,6 +775,9 @@ def load(self):
             fname = os.path.join(conf_path_str, self.CONF_BASENAME)
             if os.access(fname, os.F_OK | os.R_OK):
                 self.loadcfg(fname, upgrader.USER_CONFIG)
+            else:
+                LOG.warning(
+                    f'No global.cylc file at CYLC_CONF_PATH: {conf_path_str}')
```
MetRonnie, line 780: Is this useful? I noticed there was a previous commit to raise an error if the path doesn't exist, but that commit was reverted

oliver-sanders: Because this environment variable can get copied to remote platforms (which aren't on the same filesystem) this warning is useful some of the time, and misleading some of the time.

MetRonnie: Ok, removing it on the basis that being misleading some of the time is worse than not being helpful some of the time

MetRonnie:
I've found it messy trying to unit test the exact config dir hierarchy, but at least there is a doctest for the version number hierarchy:
https://github.com/cylc/cylc-flow/blob/564bf83a66e17fe971c4f6518feb69d68c5ab2b9/cylc/flow/cfgspec/globalcfg.py#L698-L707

and a unit test for the override behaviour of `ParsecConfig.loadcfg`
https://github.com/cylc/cylc-flow/blob/564bf83a66e17fe971c4f6518feb69d68c5ab2b9/tests/unit/parsec/test_config.py#L75-L77

MetRonnie:
Just undoing & redoing the last commit and force pushing seems to have done the trick, as the tests run on a merge commit (PR branch merged into master). Ideally the tests should re-run when master changes, then. But I don't think that's possible currently

oliver-sanders:
That's interesting, didn't know it worked like that, Travis CI just checked out your branch locally so you had to rebase to pick up changes which resulted in a couple of false negatives over the years.

MetRonnie:
Travis also creates a merge commit to run the tests, e.g. for https://travis-ci.org/github/metomi/isodatetime/builds/725975670 it says " Commit ba1e4c1 " and if you click on the link it shows
image

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -691,8 +728,19 @@ class GlobalConfig(ParsecConfig):
...
+        self.SITE_CONF_PATH = (os.getenv('CYLC_SITE_CONF_PATH') or
+                               os.path.join(os.sep, 'etc', 'cylc', 'flow'))
+        self.USER_CONF_PATH = os.path.join(self._HOME, '.cylc', 'flow')
+        version_hierarchy = get_version_hierarchy(CYLC_VERSION)
+        self.CONF_DIR_HIERARCHY = [
+            *[(upgrader.SITE_CONFIG, os.path.join(self.SITE_CONF_PATH, ver))
+              for ver in version_hierarchy],
+            *[(upgrader.USER_CONFIG, os.path.join(self.USER_CONF_PATH, ver))
+              for ver in version_hierarchy]
+        ]
```
oliver-sanders, line 742: If converting these class attributes to instance attributes make them lower case.

oliver-sanders (approved):
Looks good, one PEP8 correction.

cylc/flow/cfgspec/globalcfg.py
MetRonnie: It was only `CONF_DIR_HIERARCHY` that needed to be moved inside `__init__` because it's not possible to use another class attribute inside the list comprehension (https://stackoverflow.com/a/13913933/3217306). Seeing as this is the only reason for making it a instance attribute, would it make sense to leave it uppercase if the other two are changed to class attributes?

hjoliver (approved):
LGTM :+1:  thanks @MetRonnie

PR #3806, opened 2020-09-04
Task was destroyed but it is pending

oliver-sanders:
**built on #3805**

On master you get some cryptic feedback when running tests:

```console
$ pytest tests/i/test_scan_api.py
...
RuntimeError: Event loop is closed
Task was destroyed but it is pending!
task: <Task pending coro=<Queue.get() done, defined at /Users/oliver/miniconda3/envs/cylc8/lib/python3.7/asyncio/queues.py:150> wait_for=<Future cancelled>>
Task was destroyed but it is pending!
task: <Task pending coro=<Queue.get() running at /Users/oliver/miniconda3/envs/cylc8/lib/python3.7/asyncio/queues.py:159> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7ffa32096dd0>()]>>
Exception ignored in: <coroutine object Queue.get at 0x7ffa32042dd0>
...
```

These `Queue.get` tasks appear to be orphaned, since we only use `asyncio.Queue` in one place I'm guessing this is the fault of the async pipes (dammit).

I've added some teardown logic to the async generator stuff to ensure all running tasks get canceled when the generator exists:

```python
async for item in pipe:
    print(item)
    # if there are more items in the pipe then the teardown logic will kick in at this stage,
    # due to the maximum concurrency one shouldn't use async pipes like this though
    return
```

However, this didn't clear up the test warnings so I've added logic to the `conftest` to ensure that all tasks in the event loop get canceled before it is closed, which, for async tests, is probably a good idea anyway.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Does not need tests (why?).
- [x] No change log entry required (why? e.g. invisible to users).
- [x] No documentation update required.
- [x] No dependency changes.

Issue #3641, opened 2020-06-02
actions: remote testing functionality


oliver-sanders:
**Use containers in our GitHub Actions testing workflow to act as remote hosts for testing purposes.**

* Apparently we don't have an issue for this!
* Closely related to #3372

**The functional test battery currently supports two types of remote host:**

* remote host
* remote host with shared filesystem

**And there are currently two (soon to be three) task communication methods:**

* tcp (currently called `zmq`)
* polling (`poll`)
* ssh tunnelling (`ssh+tcp`) - https://github.com/cylc/cylc-flow/issues/3327

That means that to fully test Cylc we need six platforms. These can probably be implemented as two containers (📦) where 📦 2 is a subtle variant of 📦 1.

|                                    | tcp | poll | ssh+tcp |
|------------------------------------|-----|------|----------|
| remote host                        | 📦 1  | 📦 1   | 📦 1       |
| remote host with shared filesystem | 📦 2  | 📦 2   | 📦 2       |

**We currently support 9 batch systems (🧮):**

* at
* background
* loadleveler
* lsf
* moab
* pbs
* pbs_multi_cluster
* sge
* slurm

**So that leaves us with 2 * 3 * 9 = 54 containers!**

Ok, so this is getting a bit nuts, for testing we can probably test all of the batch systems (🧮) in the `remote host` + `tcp` image.

|                                    | tcp | poll | ssh+tcp |
|------------------------------------|-----|------|----------|
| remote host                        | 📦1 + 🧮 | 📦 1   | 📦 1       |
| remote host with shared filesystem | 📦 2  | 📦 2   | 📦 2       |

hjoliver:
> How many containers do we need?

Seems to me two is enough as testing batch system support is orthogonal to testing remote host support (isn't it?) with separate or shared FS - i.e. no need for the cross combinations.

> Can we avoid duplication between containers?

Presumably you mean duplication of test runs, not duplication of container content? If the tests are quick enough we could run all batch system tests in both contains, and to hell with the duplication. Or we could randomly select half the batch system tests for one container, half for the other (nice side effect, we would eventually pick up unforeseen cross-combination issues, at the cost of perfect reproducibility).

oliver-sanders:
> Presumably you mean duplication of test runs

Sorry, I actually meant duplication in docker files.

oliver-sanders:
Ping @kinow who knows the most.

kinow:
>Ping @kinow who knows the most.

[...]

Issue #3372, opened 2019-09-20
Use CI to run tests for batch systems


kinow:
**Describe exactly what you would like to see in an upcoming release**

We have some new unit tests for batch schedulers, but no functional test that confirms it is working.

It would be nice to use the same CI infrastructure for running tests against PBS/Slurm/etc.

**Additional context**

See https://twitter.com/jiminy_crist/status/1174681602015813633?s=20 for an example.

**Pull requests welcome!**

hjoliver:
(Re-opening - this issue that closed this was unit tests only).

kinow:
Thanks. I was going to re-open or create a new issue but forgot. My intention with this issue was to actually have Slurm, PBS, etc. Whatever batch system we could use Docker to run on Travis CI or GitHub actions (both platforms support containers). Possibly include some scenarios with SSH too.

PR #3808, opened 2020-09-09
platform orientated tests + cylc swarm

oliver-sanders:
Built on #3806 
Addresses #3641 
Addresses #3372 ("background" and "at" batch systems)

Apologies for the epic write up, I've tried to make it as concise as possible.

## tldr;

You can run the test battery using docker containers now:

```
$ etc/bin/swarm configure -y
$ etc/bin/swarm build
$ etc/bin/swarm run
$ etc/bin/run-functional-tests -p '*' tests/f
$ etc/bin/swarm kill
```

Take a look at:

* `etc/bin/swarm --help`
* `tests/f/README.md`
* The "caveats" section below.
* The "future work" section below.

## The Problem(s)

* We aren't regularly testing batch system functionality.
  * This is due to limited access to compatible platforms.
  * Some batch systems probably haven't been tested in a long time (e.g. loadleveler).
* We aren't running tests in the full assortment of configurations.
  * For example we recently broke `communications method=poll` because we have been running tests on platforms configured with `communications method=tcp`.
  * At present we can't run the test battery with `communications method=poll` as most of the remote tests require `tcp` but we have no way of knowing which ones.
  * Some tests may benefit from being run locally and remotely.
* Test failures are hard to reproduce and debug.
  * Limited access to platforms, hard to factor out configuration, etc.
  * Inconsistencies between platforms.
* Development is difficult.
  * Only people with access to particular platforms can develop for them.
  * Production platforms have complex configurations which can convolute simple matters.
  * Production platforms are restrictive and are generally nasty to develop on.
* Packaging is not tested until release time.
  * The advised way to install Cylc is via Conda, but the test battery doesn't install it in this way.
  * Is anyone currently testing remote hosts with Cylc installed via the Conda Forge recipe?

### Platform Orientated Tests

* Make the tests "platform orientated" so that they describe what they require in order to run.
* Specify test platforms using the `-p` argument to `etc/bin/run-functional-tests`.
* Run each test on the first matching platform.

### The Platform Matrix

A test platform is defined by four main components:

* Local/remote job submission - *(local, remote)*
* Batch system - *(at, background, loadleveler, lsf, moab, pbs, pbs_multi_cluster, sge, slurm)*
* Independent/shared file system - *(indep, shared)*
* `communications method` - *(tcp, poll)*

So thats 72 possible combinations, granted not all of them sensible/possible.

### Locally Available Platforms

* Configure whatever platforms you have at your disposal in your global-tests.
* You can then run the full assortment of tests which are compatible with that platform.
* This runs the maximum possible number of tests on the test platform which is good for release testing.

### Docker Swarm

* Locally available platforms are unlikely to fill all slots in the matrix.
* Build a docker image installing Cylc from a local development copy.
* Use this image to create a "swarm" of docker containers configured in different ways to test different platforms.
* This "swarm" is used for testing in GitHub Actions but can be used locally providing a stable, universal Cylc development environment.

### Docker Details

These containers are designed & configured to behave like regular hosts, ssh to them on the system to run commands. When run the containers sit idle and do not shut down until instructed to do so.

They will work if run on Linux using the host kernel (like normal docker containers), however, they will also run happily inside a virtual machine. This means that they can be run on non-Linux systems (Mac + Windows) via docker desktop and potentially also docker machine etc which may be useful in the future.

### The Container Matrix.

This PR implements:

* local, remote
* at, background
* indep
* poll, tcp

Giving us a total of 8 different testing platforms. A fair way off of the 72 possible combinations, however, the remaining slots should be easy to fill in as and when.

### With Locally Available Platforms

Test platforms must adhere to the naming convention:

```
_<local/remote>_<batch_system>_<indep/shared>_<comms method>
```

Which produces platform names you are certain to forget.

Configure a platform in your `global-tests.cylc` file like so:

```
[platforms]
    # a remote platform with polling task comms
    [[_remote_background_indep_poll]]
        hosts = _remote_background_indep_poll
        communication method = poll
        execution polling intervals = 5*PT1S, PT5S
        submission polling intervals = PT1S
```

You can now run all compatible tests on that platform by running:

```console
$ etc/bin/run-functional-tests -p _remote_background_indep_poll tests/f
```

### With The Swarm

To configure Cylc and SSH, build and run the docker container matrix do the following:

```
$ etc/bin/swarm configure -y
$ etc/bin/swarm build
$ etc/bin/swarm run
```

Now run the test battery using the provided platforms:

```console
$ etc/bin/run-functional-tests -p _remote_background_indep_poll tests/f
```

When you're done shut down the swarm:

```console
$ etc/bin/swarm kill
```

### Running The Test Battery Using Multiple Platforms

* This test matrix is thorough, ensuring that every possible combination is hashed out.
* However, during development we typically just want to run tests once in a sensible configuration.
* To do this you can specify multiple platforms, the first compatible one will be chosen to run the test and each test will be run only once.

```console
$ # run all compatible tests on _remote_background_indep_poll
$ etc/bin/run-functional-tests -p _remote_background_indep_poll tests/f

$ # run all tests on the first compatible host
$ etc/bin/run-functional-tests -p '*' tests/f

$ # run all remote tests on the first compatible host
$ etc/bin/run-functional-tests -p '_remote*' tests/f

$ # run all slurm tests on the first compatible host
$ etc/bin/run-functional-tests -p '*slurm*' tests/f
```

## Caveats

Blockers:

* ~The container is installing Cylc dependencies from conda-forge!~
  * Now using a local conda environment file.
* I've disabled two tests.

Irritations:

* It's a somewhat messy problem with a somewhat messy solution, very much open to ideas to improve it.
* Platform names are lengthy.
* The docker images are quite large, however, my system seems to handle the swarm (in a VM) with little impact.
* The matrix doesn't cater for everything, perhaps we need a better matrix.
  * For example what if we want different containers to test different versions of a batch system, Bash, Python, whatever.
* The framework doesn't filter tests by compatibility, it skips them when run.
  * This is a limitation of the current test framework, perhaps conversion to pytest might solve it.
  * This convolutes the test log.
  * This takes time, a lot of time, a serious lot of time.
     * I think it was pretty slow before.
     * Now the test config is now written whether the test runs or not, that's bad (might be able to change this after the config hierarchy work).
  * At least you can `git grep --name-only 'batch:slurm' -- tests/functional` now.

## Future Work

* Integrate Bruno's Slurm image into the matrix.
* Fill in the remaining gaps in the matrix (or at least the ones we care about).
* In the GH Action build the docker images in the first job then re-use it in the subsequent jobs.
* Try to slim down the docker images.
  * Consider moving to a light weight base image.
  * Consider replacing Conda with Mamba.
  * See if there are any dud files we can remove.
* Migrate from prove/run-functional-tests to `pytest`.

**Requirements check-list**
- [x] No dependency changes.


kinow (commented):
>Which produces platform names you are certain to forget.

:laughing: 

>The container is installing Cylc dependencies from conda-forge! 

I think we could run into issues in CI that are related to Cylc + Conda, and have to troubleshoot them, instead of dealing with failing tests due to changes in Cylc.

>The framework doesn't filter tests by compatibility, it skips them when run. 

I noticed that too. Would be nicer to just say "run just the tests for SLURM" for example. +1 if we can re-use pytest, or maybe just change our current test runner.

>The docker images are quite large, however, my system seems to handle the swarm (in a VM) with little impact.


The bash image has 590 MB. Running the container, it uses 232MB of exclusive memory, and shares 821 MB in layers that may be re-used by other containers (base image and its layers for example).
There's more space used too for checkpoints, container configuration, etc, but not too important when comparing container/image sizes I think.

```bash
kinow@ranma:~/Development/python/workspace/cylc-flow$ docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
bash                test                7564cab841ca        2 minutes ago       590MB
kinow@ranma:~/Development/python/workspace/cylc-flow$ docker ps --size
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES               SIZE
8e76603a2255        bash:test           "/bin/bash"         2 minutes ago       Up 2 minutes                            bash                232MB (virtual 821MB)
```

Building the containers from this PR created three new images (with unnamed).

So about 3GB for the images. When running, 1.18 GB shared by all the cylc-remote images.

```
(venv) kinow@ranma:~/Development/python/workspace/cylc-flow$ docker images
REPOSITORY          TAG                 IMAGE ID            CREATED              SIZE
cylc-remote         latest              99bdf4043079        About a minute ago   1.18GB
cylc-dev            latest              112377f5da1f        2 minutes ago        1.04GB
<none>              <none>              da957b0cba7c        3 minutes ago        1.01GB
...
(venv) kinow@ranma:~/Development/python/workspace/cylc-flow$ docker ps --size
CONTAINER ID        IMAGE                COMMAND                  CREATED             STATUS              PORTS                   NAMES               SIZE
055ca80d95ac        cylc-remote:latest   "/bin/sh -c 'service…"   17 seconds ago      Up 16 seconds       0.0.0.0:42230->22/tcp   focused_knuth       2kB (virtual 1.18GB)
246066281d4e        cylc-remote:latest   "/bin/sh -c 'service…"   17 seconds ago      Up 16 seconds       0.0.0.0:42221->22/tcp   cool_buck           2kB (virtual 1.18GB)
2199fadf3f4d        cylc-remote:latest   "/bin/sh -c 'service…"   18 seconds ago      Up 17 seconds       0.0.0.0:42220->22/tcp   competent_kalam     2kB (virtual 1.18GB)
8e76603a2255        bash:test            "/bin/bash"              18 minutes ago      Up 18 minutes                               bash                232MB (virtual 821MB)
```

Would be nice to reduce the images footprint. But at least the running containers share the image and its layers, saving some space.

---

I am a bit paranoid and try to keep my docker environment as clean as possible. So I prefer to always being able to either use `docker-compose` or manually run the docker commands.

I tried building the container, but it complained about some missing command. Then I used the `./etc/bin/swam ...` and everything worked. It'd still be nice to have instructions to build and start/stop the containers. That's useful for troubleshooting the environment or just understanding how things are build together.

Also, `docker kill` doesn't remove al containers I think.

```
(venv) kinow@ranma:~/Development/python/workspace/cylc-flow$ docker ps -a
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS                     PORTS               NAMES
3de8a27fa523        cylc-remote         "/bin/sh -c 'service…"   5 minutes ago       Exited (1) 5 minutes ago                       sharp_johnson
```

oliver-sanders:
> It'd still be nice to have instructions to build and start/stop the containers

The best instructions I can give are to look at the swarm file and copy the commands it runs. Perhaps I could add an option to swarm to print out the commands it would have run.

> Also, docker kill doesn't remove al containers I think.

Dammit, I'm running containers with `--rm` which I thought would tidy up afterwards? Should that be followed by a `docker rmi`?

> and also alpine that could replace the base image of the bash tests

I like Alpine and tried to use it as the base image for this work, however, minimalism comes at a cost. I couldn't install either Conda or Mamba on Alpine because of missing libraries. Once I saw how much space Conda installations took up I stopped worrying about the base image as 5-200MB is small fry compared to a >1GB image!

Might work well for the bash tests though!

oliver-sanders:
OOh, just discovered `conda install --prune`.

```
REPOSITORY                   TAG                 IMAGE ID            CREATED              SIZE
cylc-remote                  latest              41e678780b77        About a minute ago   972MB
cylc-dev                     latest              b79591fb694d        About a minute ago   832MB
```

That's a 200MB improvement!

oliver-sanders:
If the size of the containers is a barrier to development we could cut down on the number being run by using the same containers for TCP and POLL comms (since the only real difference is the Cylc configuration). I've purposefully not done this to make the tests slightly more rigourous.

oliver-sanders:
I'm now building the environments from a Conda environment file. When dependencies change we can update that file. When we release a new Cylc version we can copy the dependencies across to the Conda Forge Feedstock. Howz that sound?

The image is back up to 1GB as the dependency base has swollen.

dockerfiles/bash/global.cylc
```diff
@@ -0,0 +1,3 @@
+[platforms]
+    [[_local_background_indep_tcp]]
+        hosts = localhost
```
datamel, line 3: Just a quick reminder if this pr breaks (in case my PR (#3796) is merged first) -  these kinds of setups will require `install target = localhost` added.

oliver-sanders: Thanks.

oliver-sanders:
Got the tests passing with and without the swarm in different places. I'm not happy with the system for selecting/skipping tests (or the lack thereof), however, I think this is probably a fundamental limitation of the current test framework to resolve in future work.

hjoliver:
Is the multi-container aspect of this somewhat reinventing the "docker compose" wheel?  https://docs.docker.com/compose/

etc/bin/swarm
```diff
@@ -0,0 +1,401 @@
...
+
+        else
+            exit 1
+        fi
+    fi
+}
+
+# docker run opts for independent filesystem images
+INDEP_FS_OPTS=(
+)
```
hjoliver, line 240: empty array - see next comment...

etc/bin/swarm
```diff
@@ -0,0 +1,401 @@
...
+    if ${detached}; then
+        opts+=('-i')
+    else
+        opts+=('-d')
+    fi
+
+    # set the filesystem args
+    if [[ $fs == indep ]]; then
+        port=$BASE_PORT
+        opts+=("${INDEP_FS_OPTS[@]}")
```
hjoliver, line 277: `"${EMPTY_ARRAY[@]}"` is undefined and fails `set -u` prior to bash-4.4 (and my Centos environment is still bash-4.2, damn it).

hjoliver, line 277:
```suggestion
        opts+=(${INDEP_FS_OPTS[@]+"${INDEP_FS_OPTS[@]}"})
```
This works in bash-4.2 and 5.0.11

hjoliver:
BTW I was wondering if the container IP addresses  (via `docker inspect -f "{{ .NetworkSettings.IPAddress }}" ID`) could be used to avoid the need to mess with ssh config (no need to use the port number then) ... but switching to root user for the container needs ssh config anyway, now that we can no longer set a task "owner" for that purpose).

oliver-sanders:
> Is the multi-container aspect of this somewhat reinventing the "docker compose" wheel?

I did initially take a look at Docker Compose, but I didn't think the ability to specify command line options in a YAML file really offered anything over specifying them on the command line, one less thing to learn seemed like a good idea.

Happy to change if there's a meaningful benefit to doing so, the code that builds/runs containers is only a few lines, quick to change.

hjoliver:
> Happy to change if there's a meaningful benefit to doing so, the code that builds/runs containers is only a few lines, quick to change, happy to replace my scripts.

No biggie, it's a possible future follow-up at best.  (Although @kinow might have a stronger opinion, with more docker experience than me?)

kinow:
> > Happy to change if there's a meaningful benefit to doing so, the code that builds/runs containers is only a few lines, quick to change, happy to replace my scripts.
> 
> No biggie, it's a possible future follow-up at best. (Although @kinow might have a stronger opinion, with more docker experience than me?)

I think it's a good start, and we can refine it later if necessary (not too concerned as this is for our test battery only :+1: )

kinow:
Some conflicts in this PR now @oliver-sanders :+1:

hjoliver:
(Conflicting functional tests :vomiting_face: )

hjoliver (changes requested):
(In case my requested changes to the `swarm` script have scrolled out of sight...)

oliver-sanders:
I'll try and give this one a bit of time since it has such a large diff surface.

I've had a go at using docker-compose:

*dockerfiles/swarm.yaml*
```yaml
version: "3.8"

services:
  cylc_dev:
    # TODO: we don't want this to run
    build:
      context:
        ../
      dockerfile:
        dockerfiles/cylc-dev/Dockerfile
    image:
      cylc-dev:latest

  remote_background_indep_tcp:
    build:
      context:
        ../
      dockerfile:
        dockerfiles/cylc-remote/Dockerfile
    image:
      cylc-remote:latest
    ports:
      - "42220"
    container_name:
      remote_background_indep_tcp
    environment:
      - HOST_HOSTNAME=<paste the output of $(hostname -f) here>
```

However swarm doesn't like the bash entry point and shuts down the container as soon as it has launched:

```console
$ docker-compose -f dockerfiles/swarm.yaml up
Creating network "dockerfiles_default" with the default driver
Creating remote_background_indep_tcp ... done
Attaching to remote_background_indep_tcp
remote_background_indep_tcp    | Starting OpenBSD Secure Shell server: sshd.
remote_background_indep_tcp exited with code 0
```

If I do something daft like set the entry point to `sleep 100` just to keep the container alive the SSH fails with connection refused.

oliver-sanders:
Rebased, deconflicted and fixed the empty array thing (by commenting it out for now).

Had a go at docker compose but hit issues that would take time/knowledge to overcome. Suggest punting that down the line so we can get this diff out the way.

hjoliver (commented):
A couple of trivial comments for now; I'm having trouble getting it to work this time :thinking:

oliver-sanders:
> Does this need doing? (3796 is merged)

Can't reply to this comment for some reason.

The purpose of this is to allow configuration of the localhost platform specifically for test purposes and also to separate the local job submission methods `background` from `at`.

oliver-sanders:
> I'm having trouble getting it to work this time 🤔

Ditto, looks like a missing dependency - `packaging`.

hjoliver:
> > Does this need doing? (3796 is merged)
> 
> Can't reply to this comment for some reason.
> 
> The purpose of this is to allow configuration of the localhost platform specifically for test purposes and also to separate the local job submission methods `background` from `at`.

GitHub UI seems to have screwed up the placement of my comment. In case it wasn't obvious, I was following up on a comment by @datamel

oliver-sanders:
Ah right, that'll be the `install target` thing, I'll add that to the config shortly.

hjoliver:
> > I'm having trouble getting it to work this time thinking
> 
> Ditto, looks like a missing dependency - `packaging`.

Yep, that's my problem too, inside the containers.

hjoliver:
ooh, tests passing now :+1:

dockerfiles/bash/global.cylc
oliver-sanders: Done.

kinow (commented):
Some conflicts now.

The failure on the bash docker image is not from this PR. I've seen it in other PR's, so I might have to investigate why that same test fails intermittently.

```
Test Summary Report
-------------------
tests/functional/broadcast/00-simple.t              (Wstat: 0 Tests: 5 Failed: 4)
  Failed tests:  2-5
Files=11, Tests=40, 84 wallclock secs ( 0.05 usr  0.03 sys + 37.37 cusr  7.08 csys = 44.53 CPU)
Result: FAIL
```

kinow (approved):
All tests passing. Shame we don't have coverage number, but good improvement for out tests :+1:

hjoliver:
This still has problems with arrays bash-4.2 (my system bash, unfortunately), but I suppose that's impossible for you to fix if you only have a more recent version.  So I'm testing within a conda env with bash-5 and will do a follow-up PR for bash back-compat later if necessary...

oliver-sanders:
Rebased
Deconflicted
Upgraded new tests

hjoliver:
Some progress getting this to run on my Centos 7 box:
- upgrade to bash-5 (else bash array problems)
- add `docker run` opt to set `docker.host.internal` for the `configure` script on this branch
```diff
git diff
diff --git a/etc/bin/swarm b/etc/bin/swarm
index 32737ec60..58018b190 100755
--- a/etc/bin/swarm
+++ b/etc/bin/swarm
@@ -247,6 +247,7 @@ SHARED_FS_OPTS=(
 # docker run opts for tcp communication images
 TCP_OPTS=(
     '-e' "HOST_HOSTNAME=$(hostname -f)"
+    '--add-host' "host.docker.internal:172.17.0.1"
 )
 # docker run opts for polling communication images
 POLL_OPTS=(
```
- clear (temporarily!) `iptables` on the host (firewall was evidently blocking tcp comms from the container)

TODO:
- figure out how to determine the host IP (172.17.0.1) on the docker host rather than hardwiring it.  This might be the way:
```
ip addr show docker0
6: docker0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:e3:0f:ed:1b brd ff:ff:ff:ff:ff:ff
    inet 172.17.0.1/16 brd 172.17.255.255 scope global docker0
       valid_lft forever preferred_lft forever
```
- figure out how to open the appropriate channel in iptables (I'm network-challenged)

Any thoughts or advice @oliver-sanders ?

hjoliver:
Will everyone on the team be able run this on their work machines, if it requires configuring iptables??

tests/functional/retries/02-xtriggers.t
```diff
@@ -19,4 +19,24 @@
...
+    'xtrigger satisfied: cylc_retry_b.1' \
+    '\[b.1\] -submit-num=02' \
+    '\[b.1\] status=running: (received)failed/EXIT.*job(02)' \
+    '\[b.1\] -job(02) failed, retrying in PT2S' \
+    'xtrigger satisfied: cylc_retry_b.1' \
+    '\[b.1\] -submit-num=03' \
+    '\[b.1] status=running: (received)succeeded' \
+    '\[c.1] status=running: (received)succeeded'
+
+purge
```
hjoliver, line 41: This test is failing and I'm not quite sure what you're trying to do with it - are the new lines here a cut-and-paste error? (There's no restart DB to copy, and the log scan lines appear to be for a different suite).

hjoliver:
@oliver-sanders - I've posted a PR to your branch to fix a few minor problems here. And there's the weird failing test change I commented on just above.  After that, if you fix the conflicts we can merge this.

tests/functional/retries/02-xtriggers.t
oliver-sanders: Probs a rebase error, will take a look.

oliver-sanders:
Rebased
Deconflicted

oliver-sanders:
Rebased
Deconflicted

hjoliver (approved):
:boom:

hjoliver:
Merging this, thanks @oliver-sanders.  I don't think the failing bash test is related.

PR #3809, opened 2020-09-10
Fix time zone related tests

### LINKED ISSUE #3794: time zone related test failures
Since the recent isodatetime upgrade in Cylc8 two tests are failing on non-UTC time zones.

* tests/functional/cyclers/20-multidaily_local
* tests/functional/cylc-show/03-clock-triggered-non-utc-mode

We should probably consider running tests in non-UTC time zones to catch any potential issues as part of the fix.

MetRonnie:
(The two tests that expected timezone format to be e.g. +01 instead of +0100)

These changes close #3794

**Requirements check-list**
- [X] Does not need tests
- [X] No change log entry required (invisible to users).
- [X] No dependency changes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,7 +64,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Configure Python
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
 
@@ -83,6 +83,9 @@ jobs:
 
       - name: Test
         run: |
+          # Use non-UTC time zone
+          export TZ=XXX-05:30
+
           if [[ "${BASE}" == 'tests/k' ]]; then
               NPROC=4
           else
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -56,6 +56,19 @@ jobs:
...
+            time-zone: 'non-UTC'
+          - python-version: 3.7
+            tests: ['tests/f', '2/4']
+            time-zone: 'non-UTC'
+          - python-version: 3.7
+            tests: ['tests/f', '3/4']
+            time-zone: 'non-UTC'
+          - python-version: 3.7
+            tests: ['tests/f', '4/4']
+            time-zone: 'non-UTC'
```
oliver-sanders, line 71: I don't think there is any need to mess with the matrix, we can just run all functional tests non-UTC.

MetRonnie: Oh ok

MetRonnie: (I've finished messing around with the matrix now)

oliver-sanders:
tests/f/cylc-graph-diff/00-simple.t   is failing due to #3806, tested locally, passes.

oliver-sanders (approved):

MetRonnie:
(Just spotted a small mistake)

hjoliver (approved):

Issue #3658, opened 2020-06-18
Investigate why the web UI is getting jobs under deltas.added without getting its parent task proxy first/at the same time


kinow:
**Describe the bug**

Follow up for the deltas/incremental updates work.

For one workflow, after running it for some minutes, some times 5 minutes, some times 15 or so, the Web UI gets a job under `deltas.added`, without getting the job's parent under `deltas.added`.

The job parent does appear in the same response payload, but under `deltas.updated`.

**Release version(s) and/or repository branch(es) affected?**

`master`, but the Web UI pull request is still under review: https://github.com/cylc/cylc-ui/pull/458

**Steps to reproduce the bug**

1. check out `master` for Cylc UI Server, and install it with `pip` or `conda` in a virtual environment
2. check out the Web UI branch, and run `yarn install`, following by `yarn run build:watch`
3. install the Cylc workflow (code below)
4. start the `jupyterhub` and Cylc workflow
5. open the browser to start the Cylc UI Server
6. visit the /#/workflow/$NAME page, to see the Cylc tree component, not necessary to open any other widgets
7. press F12 (chrome/ffox) and choose the Console tab, keep an eye for warnings
8. optionally, you can choose the Debugger tab, and check the option to break on uncaught exceptions; that way you should be able to hover the variable containing the deltas payload
9. also optional, when performing step 6, choose the Network tab and filter by WS (WebService) and refresh the page, this way you should have a complete log of the messages received by the Web UI, and if you have a breakpoint, you should be able to locate your message and previous one (with some effort)

:vulcan_salute: 

**Expected behavior**

No console errors or warnings.

**Screenshots**

**Additional context**

- Previous discussion on UIS pull request starts around this comment: https://github.com/cylc/cylc-uiserver/pull/118#issuecomment-642389311
- Pull request in the Web UI to reproduce: https://github.com/cylc/cylc-ui/pull/458
- Query used: https://gist.github.com/kinow/a1f99b40d0deb3766e13b0584a84773c
- Workflow used: 

```
[scheduling]
initial cycle point = now

[[dependencies]]
[[[P1Y]]]
graph = "FAM3:succeed-all => FAM6"

[runtime]
    [[root]]
        init-script = echo "I'm first"
        env-script = echo "Hi first, I'm second"
        script = sleep 20; echo "RubyRubyRubyRuby"
        exit-script = echo 'Yay!'
        err-script = echo 'Boo!'
    [[FAM]]
    [[FAM2]]
[...]

PR #3811, opened 2020-09-13
n-distance data-store graph window generation/pruning

### LINKED ISSUE #3747: n=1 scheduler data-store
The scheduler now only needs n=0 (active) tasks to fulfill its primary purpose (scheduling), but we have agreed that it is OK to hold n=1 as well, as this will be good for clients (especially `cylc tui`) that connect directly to the scheduler. (The UI Server will populate higher-n windows for the the web UI as needed). 

See also #3746

dwsutherland:
These changes close #3747 

Closes https://github.com/cylc/cylc-ui/issues/496

**N-Distance node/edge generation:**
The cycle point nodes/edges (i.e. task/family proxies) generation is triggered
individually on transition from staging to active task pool. Each active task
is generated along with any children and parents out recursively out to a
specified maximum graph distance (n_edge_distance), that can be externally
altered (via API). Collectively this forms the N-Distance-Window on the
workflow graph.

- [x] n-distance nodes/edges generation.

**N-Distance nodes/edges pruning:**
Pruning of data-store elements is done using both the collection/set of nodes
generated through the associated graph paths of the the active nodes and the
tracking of the boundary nodes (n_edge_distance+1) of those active nodes.
Once a boundary node becomes active the original node is flagged for prunning.
Set operations are used to do a diff between the nodes of active paths
(paths whose node is in the active task pool) and the nodes of flagged paths
(whose boundary node(s) have become active).

**example one**
```
a* => b => c => d
```
(`*` as active).. With `n=1` window
With `a` active we have
In paths
```
{a: {a, b}}
```
boundary info at n+1
```
{c: {a}}
```
When `c` becomes active we have:
```
{a: {a, b}, b: {a, b, c}, c: {b, c, d}}
```
```
{c: {a}, d: {b, c, d}} 
```
(yes `d` is the prune trigger for itself and `c`, because it is the furthest distance before `n+1` ... **this works for isolates too**)
and since `c` became active we flag the now out of window `a` for pruning..
The pruning mechanism will then take all the active path nodes:
`{b, c, d}`
and do a diff with all the inactive path nodes:
`{a, b} + {a, b, c}`
so we are cleared to prune `{a}`, resulting in:
```
{b: {a, b, c}, c: {b, c, d}}
```
```
{d: {b, c, d}} 
```

**example two** 
```
a => b* => c => d
e => f => g
b => g*
```
(`*` as active).. With `n=1` window
When `b` & `g` turned active, `e` is pruned but not `a`, as `a` is in the path of active `b` (even though `g` flagged it), then when `c` becomes active `a` is pruned.

**example three**
```
a => b => c* => d
a:failed => f
```
Pruning works here because when `c` becomes active, the paths of `a` include `f`, so `f` is pruned (as long as it's not in the active task pool)..

(other convoluted graphs with conditionals and absolute triggers are also handled by this scheme)

So when we generate the window about an active task, we also find/associate the conditions for the pruning of/with it.
This ensures a change of window size won't orphan the task from being pruned, which would trip up alternate pruning schemes.. i.e reducing the window size using a scheme based flagging the ancestors/parents of active tasks would miss nodes formerly in the window.

- [x] n-distance nodes/edges pruning.

**Animated examples:**

Given the following graph with simple chain, and isolate branch:
```
[scheduling]
    initial cycle point = 20190101T00
    [[special tasks]]
        sequential = bin
    [[dependencies]]
        [[[P1M]]]
            graph = """
bin => foo => bar
baa => qux
"""
[runtime]
    [[root]]
        script = sleep 5

    [[baa, bin]]

    [[foo, bar, qux]]
        script = sleep 3
```
At `n=0` we have:
n0_window_tui

And `n=1`:
n1_window_tui

But we can also alter this with a mutation:
n_window_wui

Need to: 
- [ ] work this option into TUI
- [ ] `n=1` is default, but perhaps we can set `n=0` dynamically when the active pool gets beyond some configurable limit. 

(future PRs?)

**Requirements check-list**
- [x] No documentation required (alpha release candidate, visual but (yet) interactive change).
- [x] No dependency changes.


oliver-sanders:
Cylc Tui is looking much better 👍

I noticed that waiting tasks that are ready to run are considered to be active which confused me slightly at first but kinda makes sense as they are quite transitory.

dwsutherland:
> I noticed that waiting tasks that are ready to run are considered to be active

Yeah, I call the window creation on any task released from the runahead pool ... But this could be done elsewhere.

We can actually make n_max configurable with a mutation.. `n=0` is definitely more minimalist, and would be better for gigantic workflows.  We could also make it a subscription argument (choosing the largest n_max to decide the data-store size), and filter independently.. 

It will mean the front end can effect the data-store size on the back end, but perhaps we can set limits according authorization (i.e. read only n=0, ... , full control n=?)

hjoliver:
> > I noticed that waiting tasks that are ready to run are considered to be active
>
> Yeah, I call the window creation on any task added to the runahead pool ... But this could be done elsewhere.

I think n=0 (if that's what you're talking about) should include all tasks in the main scheduler pool, but not the runahead pool.

That will include "active waiting" tasks, as I think I called them in the SoD proposal: tasks that have at least one, but not all prerequisites satisfied (they are spawned by the first upstream, output that they depend on).  Note these aren't necessarily transitory @oliver-sanders .  As also discussed in the proposal, we could decide these waiting tasks are not "active" (i.e. not in n=0) but I don't think it matters too much right now.

hjoliver:
> We can actually make n_max configurable with a mutation.. n=0 is definitely more minimalist, and would be better for gigantic workflows. We could also make it a subscription argument (choosing the largest n_max to decide the data-store size), and filter independently..
>
> It will mean the front end can effect the data-store size on the back end workflow server, ...

Remember the scheduler only **needs** to store n=0, but we decided it can keep up to n=1 for the convenience of `cylc tui` and other direct-connecting clients. 

So the front end should determine the n-size of the UIS datastore, but not the scheduler datastore.

We could optionally allow the scheduler to store arbitrary n windows, up to some max, if you want to see loads of tasks in `cylc tui` without going through the UIS, but I don't see that as a primary use case.  We can come back and do that later if there's a need for it.  (I suppose if the UIS and scheduler code will be the same, for windowing, there's no harm in playing with it, or developing it, in the scheduler, but keep in mind the n=1 scheduler end game).

dwsutherland:
> I think n=0 (if that's what you're talking about) should include all tasks in the main scheduler pool, but not the runahead pool.

Done.

oliver-sanders:
> we could decide these waiting tasks are not "active" (i.e. not in n=0) but I don't think it matters too much right now.

It might be a little bit difficult explaining the concept of "active waiting" tasks to users, it's easier to explain the n=x window as being x edges out from an active task where active is {preparing,submitted,running}, however, it's not that big an issue.

hjoliver:
> > That will include "active waiting" tasks
> 
> > we could decide these waiting tasks are not "active" (i.e. not in n=0) but I don't think it matters too much right now.
> 
> It might be a little bit difficult explaining the concept of "active waiting" tasks to users, it's easier to explain the n=x window as being x edges out from an active task where active is {preparing,submitted,running}, however, it's not that big an issue.

Yeah I agree.  As argued in the SoD proposal, spawning of these as waiting task proxies doesn't matter - users will see a "waiting" task out front whether or not it is backed by a task proxy in the scheduler; **but** they probably shouldn't appear in n=0. 

New Issue:  #3822

dwsutherland:
> Yeah I agree. As argued in the SoD proposal, spawning of these as waiting task proxies doesn't matter - users will see a "waiting" task out front whether or not it is backed by a task proxy in the scheduler; but they probably shouldn't appear in n=0.

We can place the `self.increment_graph_window(itask)` anywhere you like (itask being a `n=0` node).. At the moment it's when a task gets released from the runahead pool.

dwsutherland:
The hard challenge I have at the moment is working out an efficient pruning scheme.. (I have a couple of things in mind, bound to evolve as I attempt)

hjoliver:
> The hard challenge I have at the moment is working out an efficient pruning scheme.. (I have a couple of things in mind, bound to evolve as I attempt)

We can have a brainstorming session about that, so it's not all on you.

dwsutherland:
I've added a pruning mechanism that satisfies the need to remove potential future workflow paths that are not taken
(i.e. 
```
foo => bar
foo:failed => qux
```
).
I'll write up more in the description soon.

There is still a little buggy behaviour, so I'll need to tidy things up somewhat.

dwsutherland:
The other issue is; I've had to revert back to n=0 window (node/edge creation) to be triggered on tasks being added/created to the runahead pool.. The reason is; tasks get removed before their children get released from the runahead pool, so creating nodes of the release from runahead creates a gap in the data-store (i.e. children get removed as a potential workflow path that was never followed, and have to get recreated again)...

So the only other options:
1. Not remove from the active task pool until a child is release from the runahead (which undermines SoD a bit)..
2. Or not remove from data-store until a any child is n=0 ..
3. Or have a node/edge field to record edge distance.. And prune according to that (which would be annoying to increment, and edge lookup is better for queries so probably wouldn't have another use).

I'm leaning towards option 2, as it's just a set lookup, and nodes are soon to know about parents to n-max anyway... Watch this space...

hjoliver:
@dwsutherland - not sure if you got any further with this yet, perhaps we can have a chat tomorrow?

My initial thought was the first order attempt might be something like this:
- if a task becomes active (i.e. if it enters n=0) follow its edges out to n=N and add those tasks to the pool (if not already in it)
- if a task leaves n=0, recompute the window from scratch, by following edges out from remaining n=0 tasks, and "prune" by differencing the old and new windows.

This would be super-easy and it might be enough, at least for low-ish n windows, because following edges is pretty easy.

Then the challenge would be to beat this with an efficient and incremental pruning scheme (i.e. if a task leaves n=0, can we decide what tasks to remove from n=N window without recomputing from scratch again).  But in the mean time, we would have a working system :grin:  - what do you think?  (Maybe there's a flaw in my logic, or you already have a better schemed lined up).

dwsutherland:
> if a task leaves n=0, recompute the window from scratch, by following edges out from remaining n=0 tasks, and "prune" by differencing the old and new windows.

@hjoliver - Yeah I already have something more efficient in place, but I've got another idea in the works which avoids knowing when something has left the pool (i.e. knows when a child at n=N becomes n=0), and does a diff to prune (which should also hit paths not taken)... Anyway I'm just about implement this.

dwsutherland:
Ok, ready for review.
The notable issues I've found (to write up along side this one), are:
- TUI flakes crashes with this every now and then:
```
(flow) sutherlander@cortex-vbox:cylc-flow$ cylc tui jox
'NoneType' object has no attribute 'get_value'
```
- There are `TaskProxy`s that don't leave the task pool when you have paths not taken: i.e.
```
[scheduling]
    initial cycle point = 20190101T00
    [[special tasks]]
        sequential = a, d
    [[dependencies]]
        [[[P1M]]]
            graph = """
a => b
a:fail => c
d => c
"""
[runtime]
    [[root]]
        script = sleep 5

    [[a, b, c]]

    [[d]]
        script = sleep 3
```
alt_path_workflow
(Which again is a task pool issue)

kinow:
Thanks a lot for providing an example (in your GIF) of how to use it via GraphiQL @dwsutherland . That's pretty much what I need for the Cylc UI change. I'll try to quickly hack the code in Cylc UI to add a combobox and play with the n-window over this next week :tada: It looks like fun!!!

kinow:
BTW, @dwsutherland , just to confirm, the mutation sets the n-window for all running workflows of the user? Am I able to set the n-window value for a single workflow (e.g. `five` with n-window =3), while I keep another workflow with a different value? (e.g. `complex` with n-window =1)?

dwsutherland:
> BTW, @dwsutherland , just to confirm, the mutation sets the n-window for all running workflows of the user? Am I able to set the n-window value for a single workflow (e.g. `five` with n-window =3), while I keep another workflow with a different value? (e.g. `complex` with n-window =0)?

Of course, just use the workflow ID instead of a glob.

cylc/flow/data_store_mgr.py
```diff
@@ -539,19 +542,173 @@ def generate_definition_elements(self):
...
+        # Reference set for workflow relations
+        for _, items in graph_children.items():
+            if edge_distance == 1:
+                descendant = True
+            self._expand_graph_window(
+                s_id, s_node, items, active_id, flow_label, edge_distance,
+                descendant, False)
+
+        for _, items in generate_graph_parents(
+                self.schd.config.get_taskdef(name), point).items():
```
kinow, line 618: I think these two `for` loops don't need to iterate `.items()` if they are discarding the key? Could be just `for items in ().values()` ? Probably doesn't make much difference in performance.

cylc/flow/data_store_mgr.py
```diff
@@ -539,19 +542,173 @@ def generate_definition_elements(self):
...
+                s_id, s_node, items, active_id, flow_label, edge_distance,
+                False, True)
+
+        if edge_distance == 1:
+            levels = self.n_window_boundary_nodes[active_id].keys()
+            # Could be self-reference node foo:failed => foo
+            if not levels:
+                self.n_window_boundary_nodes[active_id][0] = {active_id}
+                levels = (0,)
+            # Only trigger pruning for futherest set of boundary nodes
```
kinow, line 629: s/futherest/furthest ?

cylc/flow/data_store_mgr.py
```diff
@@ -539,19 +542,173 @@ def generate_definition_elements(self):
...
+        self.all_task_pool.add(tp_id)
+
+    def generate_ghost_task(self, tp_id, name, point, flow_label):
         """Create task-point element populated with static data.
 
         Args:
-            t_id (str): data-store task ID.
             tp_id (str): data-store task proxy ID.
-            point_string (str): Valid cycle point string.
+            itask (cylc.flow.TaskProxy): Task proxy object.
```
kinow, line 703: `itask` is not in this function? It's missing `name`, `point`, and `flow_label`.

cylc/flow/data_store_mgr.py
```diff
@@ -766,100 +839,122 @@ def update_data_structure(self, updated_nodes=None):
...
-    def increment_graph_elements(self):
-        """Generate and/or prune graph elements if needed.
+    def prune_data_store(self):
+        """Remove old nodes and edges by cycle point.
 
-        Use the task pool and edge source/target cycle points to find
-        new points to generate edges and/or old points to prune data-store.
+        Args:
+            point_strings (iterable):
+                Iterable of valid cycle point strings.
```
kinow, line 855: This arg is not in the function?

cylc/flow/data_store_mgr.py
```diff
@@ -766,100 +839,122 @@ def update_data_structure(self, updated_nodes=None):
...
+            for k, v in self.n_window_nodes.items()
+            if k in self.all_task_pool
+        ]
+        #    if k in self.all_task_pool
+        out_paths = [
+            v
+            for k, v in self.n_window_nodes.items()
+            if k in self.prune_flagged_nodes
+        ]
+        node_ids = set().union(*out_paths).difference(*in_paths)
```
kinow, line 875: Would `set(*out_paths)` produce the same as `set().union(*out_paths)` without the `union` (since the `set()` produces an empty set?)?

cylc/flow/data_store_mgr.py
```diff
@@ -766,100 +839,122 @@ def update_data_structure(self, updated_nodes=None):
...
+            self.deltas[EDGES].pruned.extend(node.edges)
+            parent_ids.add(node.first_parent)
+
+        prune_ids = set()
+        checked_ids = set()
+        while parent_ids:
+            for fp_id in parent_ids:
+                self._family_ascent_point_prune(
+                    fp_id, node_ids, parent_ids, checked_ids, prune_ids)
+                break
```
kinow, line 905: Could we remove the `for` loop with

```py
while parent_ids:
    self._family_ascent_point_prune(next(iter(parent_ids)), node_ids, parent_ids, checked_ids, prune_ids)
```

I **think** that'd be equivalent to iterating once with `for` and `break`'ing right away? (haven't tested it, sorry)

cylc/flow/data_store_mgr.py
```diff
@@ -766,100 +839,122 @@ def update_data_structure(self, updated_nodes=None):
...
+        checked_ids.add(fp_id)
+        if fp_id in parent_ids:
+            parent_ids.remove(fp_id)
+
+    def update_dynamic_elements(self, updated_nodes=None):
+        """Update data elements containing dynamic/live fields."""
+        # If no tasks are given update all
+        if updated_nodes is None:
+            updated_nodes = self.schd.pool.get_all_tasks()
+        elif not updated_nodes:
```
kinow, line 953: Should it be an `if` instead? In case the `updated_nodes = self.schd.pool.get_all_tasks()` makes `updated_nodes` to be `False`?

cylc/flow/data_store_mgr.py
```diff
@@ -1035,6 +1140,45 @@ def _family_ascent_point_update(self, fp_id):
...
+                continue
+            tp_delta = self.updated[TASK_PROXIES].setdefault(
+                tp_node.id, PbTaskProxy(id=tp_node.id))
+            tp_delta.stamp = f'{tp_node.id}@{update_time}'
+            tp_delta.is_held = hold
+            tp_delta.state = tp_node.state
+            self.state_update_families.add(tp_node.first_parent)
+
+    def set_graph_window_extent(self, n_edge_distance):
+        """Set the what the max edge distance will change to.
```
kinow, line 1160: s/Set the what the max edge/Set the max edge ?

kinow (commented):
Really impressive work @dwsutherland ! I don't fully understand everything that is going on here, so I left some comments that are more like questions :grimacing: Feel free to ignore any that is not correct/pertinent.

Will start testing with GraphiQL and then Cylc UI, and finalize the review today or tomorrow :+1: 

Thanks!!
Bruno

cylc/flow/data_store_mgr.py
```diff
@@ -539,19 +542,173 @@ def generate_definition_elements(self):
...
+
+        """
+        # Create this source node
+        s_node = TaskID.get(name, point)
+        s_id = f'{self.workflow_id}{ID_DELIM}{point}{ID_DELIM}{name}'
+        if active_id is None:
+            active_id = s_id
+
+        # Setup and check if active node is another's boundary node
+        # to flag it's paths for pruning.
```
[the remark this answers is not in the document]
dwsutherland: It is possessive, so not short for "it is"..   i.e. a car and it's wheels
(I could be wrong >.>)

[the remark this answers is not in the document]
dwsutherland: Ok looked it up.. and I am wrong... Apparently the apostrophe doesn't play by the same possessive vs plural rules as other words

cylc/flow/data_store_mgr.py
```diff
@@ -539,19 +542,173 @@ def generate_definition_elements(self):
...
+        if (
+                edge_distance > self.n_edge_distance or
+                not any(graph_children.values())
+        ):
+            if descendant or edge_distance == 0:
+                self.n_window_boundary_nodes[
+                    active_id].setdefault(edge_distance, set()).add(s_id)
+
+        if edge_distance > self.n_edge_distance:
+            return
```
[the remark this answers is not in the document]
dwsutherland: Yeah, tell me how, without duplicating the internal check or adding extra checks and I will..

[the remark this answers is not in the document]
dwsutherland: i.e. could do this:
```
        if not any(graph_children.values()):
            if descendant or edge_distance == 0:
                self.n_window_boundary_nodes[
                    active_id].setdefault(edge_distance, set()).add(s_id)

        if edge_distance > self.n_edge_distance:
            if descendant or edge_distance == 0:
                self.n_window_boundary_nodes[
                    active_id].setdefault(edge_distance, set()).add(s_id)
            return
```

maybe that's better?

cylc/flow/data_store_mgr.py
dwsutherland: True, I dashed them out temporarily, thanks for the reminder.

cylc/flow/data_store_mgr.py
dwsutherland: I thought so too.. but I think it would be a set of sets

cylc/flow/data_store_mgr.py
dwsutherland: That would be equivalent to iterating until `parent_ids` are empty .. Will have to look at it again, but I think (like the family update code), I was trying an alternative to `.pop()` where you remove the item at the same time.

cylc/flow/data_store_mgr.py
dwsutherland: Yeah could be `if`.. The assumption is `self.schd.pool.get_all_tasks()` is never empty thereafter, but even if it is, the following loop will do nothing.

cylc/flow/data_store_mgr.py
dwsutherland: Yes, not anymore, oops

cylc/flow/data_store_mgr.py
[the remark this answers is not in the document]
kinow: Ah, I thought that we could have 

```
        if edge_distance > self.n_edge_distance:
            return

        graph_children = generate_graph_children(
            self.schd.config.get_taskdef(name), point)
        if not any(graph_children.values()): # <-- I thought perhaps that condition didn't need to be used here?
            if descendant or edge_distance == 0:
                self.n_window_boundary_nodes[
                    active_id].setdefault(edge_distance, set()).add(s_id)
```

But now I think you have confirmed we need to check that anyway.

>maybe that's better?

It looks better to me, but if you are happy with the other version all good for me too now that I understand we need to check the condition twice :+1:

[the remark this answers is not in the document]
dwsutherland: No can't do this first:
```
        if edge_distance > self.n_edge_distance:
            return
```
As we need to make it as a boundary node if it's a direct descendant.

kinow (commented):
## UI without changing n-distance/n-window

Tested with Cylc UI first. Using `five` as test for reference. It worked OK, found two differences.

The first that before I would normally have 2 cycle points. Now I am seeing—most of the time—3 cycle points.
Two contain the tasks/jobs in running/waiting/succeeded state. And the 3rd cycle point has just `foo` in waiting.
I think that's because the default n-window is 1?

IMG

The other thing is that there are a lot more of JS errors due to missing parent, with the error
`Error applying initial data burst for deltas TypeError: job.node.firstParent is undefined`.
I believe this is a known issue, #3658, but still worth pointing out it is more frequent now I think
(unfortunately I cannot explain why that happens now).

But regardless of the errors, the UI works fine with this PR, so these two issues are no blockers IMO.

## UI changing n-distance/n-window from 0 to 4 with GraphiQL

@dwsurtherland, this test was initializing my environment from scratch, with just `jupyterhub`
and UI from `master`. Left the UI running for some minutes, and then in another window/tab opened
GraphiQL.

There I executed the following query:

```graphql
mutation {
  setGraphWindowExtent (
    workflows: ["kinow|five"],
    nEdgeDistance: 0
  ) {
    result
  }
}
```

Where `nEdgeDistance` started at `0`, and then I incremented slowly, by 1, up to 4. By slowly I mean that I updated,
then waited a couple minutes before running the query again. That was just so the UI had time to receive deltas,
apply them, and display the extended graph.

Here I think we have a real issue with the UI (not sure if TUI works OK?). My guess is that as we change the n-window,
some deltas are not sent due to the new window/distance size (i.e. not in `deltas.pruned`), or they are in
different place (i.e. maybe `inserted` instead of `updated`), or maybe the UI needs some changes?

I thought the UI would need only a combo box to call the mutation (which I guess would either be part of
https://github.com/cylc/cylc-ui/pull/504 or implemented using code from that PR). But maybe we need to
change something in how we apply deltas if/when the n-distance is changed?

Bruno

dwsutherland:
@kinow - I've addressed the code review comments..

I think TUI was crashing from the same no parent reasons with jobs and with task proxies.. I've put a workaround in TUI to address that.. 

I think I'll address the issue after event driven deltas are in (as this could be half the problem)...
I'll try to catch where it's happening before then if I have time, but maybe follow-on PRs (I've put a couple of TODOs in the code).

kinow (approved):
Thanks @dwsutherland !

dwsutherland:
> The first that before I would normally have 2 cycle points. Now I am seeing—most of the time—3 cycle points.
Two contain the tasks/jobs in running/waiting/succeeded state. And the 3rd cycle point has just foo in waiting.
I think that's because the default n-window is 1?

Don't compare it to before.. That was just based on what cycles were in the task pool.
If foo is sequential, then n=1 will give you three cycle points; the current running foo (0), the previous succeeded foo (-1), the next waiting foo (+1).

cylc/flow/data_store_mgr.py
```diff
@@ -539,19 +542,178 @@ def generate_definition_elements(self):
...
+            t_node = TaskID.get(t_name, t_point)
+            t_id = (
+                f'{self.workflow_id}{ID_DELIM}{t_point}{ID_DELIM}{t_name}')
+            # Initiate edge element.
+            if is_parent:
+                e_id = (
+                    f'{self.workflow_id}{ID_DELIM}{t_node}{ID_DELIM}{s_node}')
+            else:
+                e_id = (
+                    f'{self.workflow_id}{ID_DELIM}{s_node}{ID_DELIM}{t_node}')
```
oliver-sanders, line 651: I'm confused by these lines:

* The `{s,t}_node`s are in `<task>.<cycle>` format but being templated into a universal ID?
* The first `{s,t}_node` is going to be interpreted as a cycle point?

oliver-sanders (commented):
Working my way through....

cylc/flow/data_store_mgr.py
```diff
@@ -539,19 +542,178 @@ def generate_definition_elements(self):
...
+        Returns:
+
+            None
+
+        """
+        # Create this source node
+        s_node = TaskID.get(name, point)
+        s_id = f'{self.workflow_id}{ID_DELIM}{point}{ID_DELIM}{name}'
+        if active_id is None:
+            active_id = s_id
```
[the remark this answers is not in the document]
dwsutherland: `active_id` is the original `n=0` .. So I'm trying to carry that information through the recursion

cylc/flow/data_store_mgr.py
dwsutherland: Taskdef ID is `{self.workflow_id}{ID_DELIM}{name}` .. So there's no promise of the the third item being a cycle point.. Whether there should be is a different question.. Maybe?   

> <task>.<cycle> format but being templated into a universal ID

Well the source and target needs to be unique in the edge ID ... Can't be just the name as we have intercycle dependencies.

cylc/flow/config.py
```diff
@@ -1742,7 +1742,7 @@ def generate_triggers(self, lexpression, left_nodes, right, seq,
             # (name is left name)
             self.taskdefs[name].add_graph_child(task_trigger, right, seq)
             # graph_parents not currently used but might be needed soon:
-            # self.taskdefs[right].add_graph_parent(task_trigger, name, seq)
+            self.taskdefs[right].add_graph_parent(task_trigger, name, seq)
```
hjoliver, line 1745: Good thing I left this in the code :grin:

cylc/flow/task_proxy.py
```diff
@@ -32,6 +32,79 @@
 from cylc.flow.wallclock import get_unix_time_from_time_string as str2time
 
 
+def generate_graph_children(tdef, point):
```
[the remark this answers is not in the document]
hjoliver: In taskdef would make sense, given the first arg.

tests/functional/graphql/01-workflow.t
```diff
@@ -90,7 +90,7 @@ cat > expected << __HERE__
                 "title": "foo",
                 "description": "bar"
             },
-            "newestRunaheadCyclePoint": "1",
+            "newestRunaheadCyclePoint": "",
```
hjoliver, line 93: What's the reason for this change?

cylc/flow/task_proxy.py
```diff
@@ -32,6 +32,79 @@
...
+                      trigger.offset_is_from_icp)
+            if is_abs:
+                if parent_point != point:
+                    # If 'foo[^] => bar' only spawn off of '^'.
+                    continue
+            if seq.is_on_sequence(parent_point):
+                # E.g.: foo should trigger only on T06:
+                #   PT6H = "waz"
+                #   T06 = "waz[-PT6H] => foo"
+                graph_parents[seq].append((name, parent_point, is_abs))
```
hjoliver, line 90: The comments here relate to spawning and triggering, which isn't appropriate to graph parents.  But also, have you convinced yourself this code works in the parent case (for the child case, I remember this stuff being a bit brain-twisting...)

tests/functional/graphql/01-workflow.t
dwsutherland: Apparently there's no runahead pool on start up, when now I initiate the pool earlier:
```
        if self.schd.pool.pool:
            pool_points = set(self.schd.pool.pool)
            workflow.oldest_cycle_point = str(min(pool_points))
            workflow.newest_cycle_point = str(max(pool_points))
        if self.schd.pool.runahead_pool:
            workflow.newest_runahead_cycle_point = str(
                max(set(self.schd.pool.runahead_pool))
            )
```
You can see this get's run, as the other's show up:
```
            "meta": {
                "title": "foo",
                "description": "bar"
            },
            "newestRunaheadCyclePoint": "",
            "newestCyclePoint": "1",
            "oldestCyclePoint": "1",
            "reloaded": false
```

hjoliver: OK (I think) but maybe add a comment to that line in the test, because it looks weird.

cylc/flow/task_proxy.py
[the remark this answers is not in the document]
hjoliver: (In fact I should have it there in the first place).

cylc/flow/task_proxy.py
dwsutherland: Yeah, because the appropriate history remains in the graph window when you prune out path nodes.
Another aspect working; it doesn't go back before the initial cycle point.

dwsutherland: Without it, the oldest task is in the active pool.

hjoliver:
>  There are TaskProxys that don't leave the task pool when you have paths not taken: i.e.

(Discussed offline in Element chat) These waiting tasks represent partially-satisfied prerequisite, soon to be hidden from the active pool #3823, no not an issue for this PR)

cylc/flow/data_store_mgr.py
```diff
@@ -539,19 +542,178 @@ def generate_definition_elements(self):
...
+        if (
+                tp_id in self.prune_trigger_nodes and
+                tp_id in self.prune_trigger_nodes[tp_id]
+        ):
+            self.prune_flagged_nodes.update(self.prune_trigger_nodes[tp_id])
+            del self.prune_trigger_nodes[tp_id]
+            self.prune_pending = True
+
+    def add_runahead_node(self, name, point):
+        """Flag removal of this active graph window."""
```
hjoliver, line 693: Wrong docstring?

dwsutherland: Opps!

cylc/flow/data_store_mgr.py
```diff
@@ -32,16 +32,24 @@
 Static data elements are generated on workflow start/restart/reload, which
 includes workflow, task, and family definition objects.
 
+The cycle point nodes/edges (i.e. task/family proxies) generation is triggered
+individually on transition from staging to active task pool. Each active task
+is generated along with any children and parents out recursively out to a
```
hjoliver, line 37:
```suggestion
is generated along with any children and parents recursively out to a
```

cylc/flow/task_pool.py
```diff
@@ -238,6 +245,7 @@ def add_to_runahead_pool(self, itask, is_new=True):
                 "flow_label": itask.flow_label})
             if itask.state.outputs.has_custom_triggers():
                 self.suite_db_mgr.put_insert_task_outputs(itask)
+            # create data-store task graph window
```
hjoliver, line 248: Is this comment supposed to be here?

cylc/flow/task_pool.py
```diff
@@ -229,7 +233,10 @@ def add_to_runahead_pool(self, itask, is_new=True):
         self.runahead_pool[itask.point][itask.identity] = itask
         self.rhpool_changed = True
 
-        # add row to "task_states" table
+        # Register pool node in data-store in external ID format
+        self.data_store_mgr.add_runahead_node(itask.tdef.name, itask.point)
+
+        # add row to "task_states" table & data-store
```
hjoliver, line 239: `& datastore` ?  (Seems to be DB only in the following code block)

cylc/flow/task_pool.py
```diff
@@ -229,7 +233,10 @@ def add_to_runahead_pool(self, itask, is_new=True):
         self.runahead_pool[itask.point][itask.identity] = itask
         self.rhpool_changed = True
 
-        # add row to "task_states" table
+        # Register pool node in data-store in external ID format
+        self.data_store_mgr.add_runahead_node(itask.tdef.name, itask.point)
```
hjoliver, line 237: What's "external ID format"?  (It's not mentioned anywhere else in cylc-flow).

Why are runahead tasks added here, rather than just active tasks?

cylc/flow/data_store_mgr.py
```diff
@@ -32,16 +32,24 @@
...
+specified maximum graph distance (n_edge_distance), that can be externally
+altered (via API). Collectively this forms the N-Distance-Window on the
+workflow graph.
+
+Pruning of data-store elements is done using both the collection/set of nodes
+generated through the associated graph paths of the the active nodes and the
+tracking of the boundary nodes (n_edge_distance+1) of those active nodes.
+Once a boundary node becomes active the active node is flagged for prunning.
```
hjoliver, line 45: (via Element chat - this sentence could be clearer!)

hjoliver (commented):
Part way done reviewing - so far so good @dwsutherland :tada: 

Is a diagram or two (pen-and-paper is fine) feasible to explain the pruning algorithm better?  Still having a hard time getting my head around that from the very brief description and code.

cylc/flow/task_pool.py
dwsutherland: > Why are runahead tasks added here, rather than just active tasks?

I use it to bridge the gap between removed parent and active child.

dwsutherland:
`n=0` now treated as isolates (pruned as soon as finished)..
```
[scheduling]
  [[graph]]
    R1 = foo => bar & baz => qux
[runtime]
  [[root]]
    script = true
  [[baz]]
    script = sleep 10
```
n_0_isolates

hjoliver (approved):
I have played with various workflow graphs at several n values, and this seems to do the right thing - a nice improvement :+1:  

Noticed scheduler reload wipes out jobs from the datastore (probably not just on this branch though; fix with a follow-up?)

Some follow-ups already planned (e.g. waiting-but-held-back tasks in n=1 not n=0) and some other tweaks might be needed in due course, but I think this is a good basis for n-windows in the scheduler.

I did a pretty once-over-lightly code review, but it looked fine at that level ... so let's keep the ball rolling here.

oliver-sanders:
I've not done an especially deep code review, I think I have a vague idea of what's going on, I know @dwsutherland and @hjoliver have debated the implementation to death so I'm happy to pass over that.

I've had a play with a few graphs and it seems to be giving me the expected results.

I've found two things which were unexpected to me, though I may just be demonstrating my ignorance of the discussions over recent weeks.

1) I wasn't expecting the scheduler n-window to be configurable on the fly like that.
    * Not a problem, however, I guess we will need to add some warning notices advising users not to play
    with this as it might have the potential to mess up the UI server?
2) Pre-initial tasks now show in the n=1 window.
    * We should block these from the view as they are potentially quite confusing, but happy to see that in a follow up.
    * Again apologies if I've missed a discussion.
3) I think that I'm getting the `n+1` window (e.g. I ask for `n=1` but get `n=2`).

This capture shows (2) [task `t.0`] and (3) [two waiting tasks in the `n=1` window]:

```
P1 = z[-P1] => a => b => c => d ...
```

Kapture 2020-11-10 at 16 33 29

hjoliver:
> I wasn't expecting the scheduler n-window to be configurable on the fly like that.
    -    Not a problem, however, I guess we will need to add some warning notices advising users not to play with this as it might have the potential to mess up the UI server?

Huh, I was expecting that, and that we'd just have to make it work with the UIS.  If I'm viewing a flow in the UI with the default n=1 window, and want to see more tasks (n=4 say), surely I shouldn't have to restart my UIS to achieve that?

> Pre-initial tasks now show in the n=1 window.
   - We should block these from the view as they are potentially quite confusing, but happy to see that in a follow up.

I also noticed this but forgot to comment on it (definitely not intended from my perspective, but you didn't miss a discussion on it).  Fix as a follow-up is fine.

>   I think that I'm getting the n+1 window (e.g. I ask for n=1 but get n=2).

@dwsutherland can confirm but I think that's because you're holding task `b`, and a ready-to-run-but-held task still looks like n=0 even though it isn't active.  The datastore is currently assuming that the scheduler task pool content is all n=0/active.  Fixing this is another SoD-follow-up (if we take those tasks out of the scheduler task pool) or a datastore follow-up (if we get the datastore manager to deliberately exclude those tasks.

hjoliver:
So, no merge-blockers there I think, but will wait for @dwsutherland  to respond as well.

dwsutherland:
> > I wasn't expecting the scheduler n-window to be configurable on the fly like that.
> > 
> > * Not a problem, however, I guess we will need to add some warning notices advising users not to play with this as it might have the potential to mess up the UI server?
> 
> Huh, I was expecting that, and that we'd just have to make it work with the UIS. If I'm viewing a flow in the UI with the default n=1 window, and want to see more tasks (n=4 say), surely I shouldn't have to restart my UIS to achieve that?

Actually, I don't think it will mess up the UIS, because the boundary pruning isn't effected by it (only new nodes register a new boundary) .. And even if they get out of sync, the UIS will reconcile (via the stamp). (though in the future they may be independent)

> 
> > Pre-initial tasks now show in the n=1 window.
> > 
> > * We should block these from the view as they are potentially quite confusing, but happy to see that in a follow up.
> 
> I also noticed this but forgot to comment on it (definitely not intended from my perspective, but you didn't miss a discussion on it). Fix as a follow-up is fine.

Huh? tasks before the initial cycle point? I haven't seen this (maybe I'm not running the right flows)

> 
> > I think that I'm getting the n+1 window (e.g. I ask for n=1 but get n=2).
> 
> @dwsutherland can confirm but I think that's because you're holding task `b`, and a ready-to-run-but-held task still looks like n=0 even though it isn't active. The datastore is currently assuming that the scheduler task pool content is all n=0/active. Fixing this is another SoD-follow-up (if we take those tasks out of the scheduler task pool) or a datastore follow-up (if we get the datastore manager to deliberately exclude those tasks.

Correct, if it's in the pool then it's `n=0`  .. However, we may also get that boundary pruning gap smoothing with n>0, where tasks aren't pruned until an active child (at n+1) becomes active.

hjoliver:
> Huh? tasks before the initial cycle point? I haven't seen this (maybe I'm not running the right flows)

```ini
[scheduling]
  cycling mode = integer
  initial cycle point = 1
  final cycle point = 6
  [[graph]]
    P1 = "foo[-P1] => foo"
[runtime]
  [[root]]
    script = sleep 20
```
Screenshot from 2020-11-11 09-51-57

dwsutherland:
Oh, it's the integer cycling, as the following doesn't have that issue:
```
[scheduling]
  #cycling mode = integer
  initial cycle point = 20200101
  final cycle point = 20200107
  [[graph]]
    P1D = "foo[-P1D] => foo"
[runtime]
  [[root]]
    script = sleep 20
```

Will have a look.

dwsutherland:
Hmm.. This works fine:
```
[scheduling]
  cycling mode = integer
  initial cycle point = 1
  final cycle point = 6
  [[special tasks]]
    sequential = foo
  [[graph]]
    P1 = """
#foo[-P1] => foo
foo
"""
[runtime]
  [[root]]
    script = sleep 20
```
So might be the trigger logic

hjoliver:
Yes it seems to only happen with intercycle dependence.  But the pre-initial task proxies aren't appearing in the scheduler task pool, right? So is it something to do with cycling sequence methods accessed by the datastore?

dwsutherland:
The cycle point needed to be checked that it is on sequence and valid, for  iso8601 sequences the `is_on_sequence` method already included a bounds check, where integer bounds checking is included in `is_valid` (iso8601 `is_valid` only references the `is_on_sequence` result).

dwsutherland:
@oliver-sanders  (BTW changing the window size doesn't require a reload.. Reloads will wipe and restart the whole data-store)

hjoliver:
(My similar fix here was also for integer cycling: https://github.com/cylc/cylc-flow/pull/3756).

Posted an issue to check up on these cycling sequence methods: #3934

hjoliver:
Checked it works. Thanks @dwsutherland :tada: :bouquet:

PR #3813, opened 2020-09-14
fix resolve via workflow parent lookup

dwsutherland:

Noticed a but when looking up node relationships via the `workflow` parent while resolving GraphQL queries.. This PR addresses this.

One review should do.

**Requirements check-list**
- [x] No change log entry required (pre-release & not visible).
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -412,10 +412,12 @@ async def get_nodes_by_ids(root, info, **args):
             parent_args.update(
                 {'id': parent_id, 'delta_store': False}
             )
-            parent = await resolvers.get_node_by_id(
-                NODE_MAP[get_type_str(info.parent_type)],
-                parent_args
-            )
+            parent_type = get_type_str(info.parent_type)
+            if parent_type in NODE_MAP:
+                parent = await resolvers.get_node_by_id(
+                    NODE_MAP[parent_type], parent_args)
+            else:
+                parent = await resolvers.get_workflow_by_id(parent_args)
             field_ids = getattr(parent, field_name, None)
         if not field_ids:
             return []


cylc/flow/network/schema.py
```diff
@@ -412,10 +412,12 @@ async def get_nodes_by_ids(root, info, **args):
...
-            parent = await resolvers.get_node_by_id(
-                NODE_MAP[get_type_str(info.parent_type)],
-                parent_args
-            )
+            parent_type = get_type_str(info.parent_type)
+            if parent_type in NODE_MAP:
+                parent = await resolvers.get_node_by_id(
+                    NODE_MAP[parent_type], parent_args)
+            else:
+                parent = await resolvers.get_workflow_by_id(parent_args)
```
kinow, line 420: Right, so if we don't have the parent in the map/cache, then we define the `parent` as the workflow (which is retrieved with the `parent_args`)? It won't ever happen that we have a parent missing that is some other type, like a family, or task?

Don't know much about the code involved here, but looks OK to me.

kinow (approved):

cylc/flow/network/schema.py
kinow: (also not sure if it'd be easy to write a test for this, I guess it'd be a bit hard?)

hjoliver:
>  Already covered by existing tests.

Re chat just now on Element: if this was already covered by existing tests, then existing tests should be failing on master (and passing on this branch). :grin:

hjoliver:
@dwsutherland could you just explain briefly what the effect of this bug was, and how to trigger this bit of code (I'm having trouble doing that, via a debugger, so I can observer the content of the variables here...)

cylc/flow/network/schema.py
dwsutherland: > @dwsutherland could you just explain briefly what the effect of this bug was, and how to trigger this bit of code (I'm having trouble doing that, via a debugger, so I can observer the content of the variables here...)

Run this query in GraphiQL on a stopped workflow:
```
{
  workflows (ids: ["nox"]) {
    id
    taskProxies {
      id
    }
  }
}
```
and you'll get:
```
{
  "errors": [
    {
      "message": "'Workflow'",
      "locations": [
        {
          "line": 4,
          "column": 5
        }
      ],
      "path": [
        "workflows",
        0,
        "taskProxies"
      ]
    }
  ],
  "data": {
    "workflows": [
      {
        "id": "sutherlander|nox",
        "taskProxies": null
      }
    ]
  }
}
```

dwsutherland:
> @dwsutherland could you just explain briefly what the effect of this bug was, and how to trigger this bit of code (I'm having trouble doing that, via a debugger, so I can observer the content of the variables here...)

Run this query in GraphiQL on a stopped workflow:
```
{
  workflows (ids: ["nox"]) {
    id
    taskProxies {
      id
    }
  }
}
```
and you'll get:
```
{
  "errors": [
    {
      "message": "'Workflow'",
      "locations": [
        {
          "line": 4,
          "column": 5
        }
      ],
      "path": [
        "workflows",
        0,
        "taskProxies"
      ]
    }
  ],
  "data": {
    "workflows": [
      {
        "id": "sutherlander|nox",
        "taskProxies": null
      }
    ]
  }
}
```

dwsutherland:
It only took this route at the UIS because of the empty-ish data store of the stopped suite..

cylc/flow/network/schema.py
dwsutherland: Probably would be good as a functional test when the CLI can go through the UIS.

dwsutherland:
@hjoliver ping! sorry, just replied to your earlier comment about it not being picked up by tests at the workflow end ☝️

hjoliver (approved):

PR #3816, opened 2020-09-15
Change spawn command name.

### LINKED ISSUE #3735: cylc spawn: command name
Should we rename the `cylc spawn` command?  

This was suggested during SoD implementation discussions.  However I think the existing name is fine.  The command forces a parent task to **spawn** the children of its (specified, or default success) outputs.

hjoliver:
These changes close #3735

After discussion on the Issue and in offline chat, I've gone with `cylc set-outputs` as the new command name. This should be a more intuitively obvious name for users (it makes the scheduler behave just as if specified outputs had been completed naturally).

Notes:
- I have not changed the internal method names that action this command, just the user-facing stuff
- there were already some functional tests that effectively test the command, but only for the default case ("succeeded" outputs); so I've added a functional test to check that setting of custom outputs works.

**Requirements check-list**
- [x] Appropriate tests are included (functional; plus default case coverage by existing tests).
- [x] No documentation update required (will be covered by general Cylc 8 documentation upgrade).
- [x] No dependency changes.


kinow (approved):
>I have not changed the internal method names that action this command, just the user-facing stuff

:+1: 

Without re-creating the virtual environment, `cylc spawn` gives some Python traceback due to missing module. But after re-creating it, it prints the expected output

```bash
(venv) kinow@ranma:~/Development/python/workspace/cylc-flow$ cylc spawn
Error: cylc spawn: unknown utility. Abort.
Type "cylc help all" for a list of utilities.
```

Also did a grep with IDE, found no occurrences of anything similar to `cylc spawn`. Great job!

Q: is `REG` an argument? The usage example doesn't seem to use it. Asking because I'm trying to run a simple test with `five` :grimacing: 

image

hjoliver:
> Q: is REG an argument? The usage example doesn't seem to use it. Asking because I'm trying to run a simple test with five grimacing

Good spotting!

REG stands for "REGistered workflow name"  (i.e. from `cylc register foo /path/to/foo/flow.cylc`).   It should be used, I'll fix the help.

(I'll also put an issue up to swap out "REG" for something less obscure, probably "FLOW", in all command help).

dwsutherland (approved):
Tested & working on my machine, tests passed, code LGTM 👍

PR #3820, opened 2020-09-16
Validate task messages & task outputs separately

### LINKED ISSUE #3428: Cylc Validate check custom task outputs
Cylc message triggers can not have a colon in them unless it is a valid severity level eg. warning.
We should check to ensure that the trigger is valid.

MetRonnie:
Follow on from #3788, these changes close #3428

I didn't quite get it right in #3788. There are two different requirements for _task messages_ and _task outputs/message triggers_.
- Task messages (e.g. `cylc message -- "${CYLC_SUITE_NAME}" "${CYLC_TASK_JOB}" "WARNING: Uh-oh"`):
  - Can have colons only if there's one at the end of the first word of the message, which will be the (possibly custom) severity level
- Task outputs/message triggers (e.g. `[runtime][my_task][outputs]foo = "Uh-oh"`):
  - Cannot have colons at all

Note: Task messages can be used for purposes other than for message triggering, so they can have colons after `SEVERITY:`. However, colons cannot be used in task messaging if indeed they are used for triggering, as suite/flow validation will fail if the task outputs contain colons.

**Requirements check-list**
- [X] Appropriate tests are included (doctests).
- [X] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/157.
- [X] No dependency changes.


cylc/flow/scripts/cylc_message.py
```diff
@@ -132,6 +134,10 @@ def main(parser, options, *args):
         if message_str == '-':
             pass
         elif ':' in message_str:
+            valid, err_msg = TaskMessageValidator.validate(message_str)
+            if not valid:
+                raise UserInputError(
+                    f'Invalid task message "{message_str}" - {err_msg}')
```
MetRonnie, line 140: The trouble with this is, it doesn't fail until the `cylc message ...` command inside `[runtime][<task>]script` is finally run. Not sure if there is a better way; trying to parse the contents of the `script` setting seems overly complicated

hjoliver: Ooh, yeah, hadn't thought of that.  If messages were separate config items, that would be easier, but that train left the station long ago.

hjoliver: Actually, maybe this is OK. It results in job, not scheduler, failure. So it can be fixed on the fly (fix task def, reload and retrigger).

hjoliver: On that basis I'm OK with this; but we'll see what @oliver-sanders thinks.

hjoliver (approved):
(See discussion on validation above)

MetRonnie:
Not sure why tests timed out
```
2020-09-16T20:47:43Z [02:17:43] tests/f/cyclers/01-hourly.t ... ok    67572 ms (0.02 usr  0.00 sys + 96.47 cusr 10.37 csys = 106.86 CPU)
2020-09-16T21:14:41Z ##[error]The operation was canceled.
```
Re-running...

oliver-sanders:
Usually happens when a test is waiting for some form of signal but where no appropriate timeout was set. It's more than likely that this validation has flushed out an error in a test suite? The test should hang locally too, you should be able to see which one it is when run locally:

```console
$ CHUNK=1/4 etc/bin/run-functional-tests tests/f
```

MetRonnie:
It's not hanging for me locally

MetRonnie:
I should have expected that, considering that task message validation failure will stall the suite

oliver-sanders:
Tests passing.

oliver-sanders (approved):

oliver-sanders:
I'll leave the merge to you so you can decide where you want to put the `.. autoclass::` directive.

cylc/flow/cfgspec/suite.py
```diff
@@ -1249,6 +1248,11 @@
 
                        out1 = "sea state products ready"
                        out2 = "NWP restart files completed"
+
+                    Task outputs are validated by
+                    :py:class:`cylc.flow.unicode_rules.TaskOutputValidator`.
+
+                    .. autoclass:: cylc.flow.unicode_rules.TaskOutputValidator
```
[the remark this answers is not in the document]
MetRonnie: In that PR it's `.. autoclass:: cylc.flow.unicode_rules.TaskMessageValidator`

They're quite similarly named; maybe `TaskMessageValidator` should be renamed `CylcMessageValidator`?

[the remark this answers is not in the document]
oliver-sanders: Ah, ok, these classes aren't user facing so don't need to worry *too* much about nomenclature.

PR #3823, opened 2020-09-17
Hide waiting tasks from n=0.

### LINKED ISSUE #3822: Take waiting tasks out of the n=0 window
Post SoD, waiting tasks are spawned "on demand" when the first upstream output that they depend on gets completed. 

These waiting task proxies don't pose a problem in and of themselves: they are just a natural way to keep track of partially-satisfied prerequisites, and users need to see waiting tasks at those points in the graph whether or not there are task proxy objects behind them.  

However, they probably shouldn't appear in the n=0 window, otherwise we have to explain why partially satisfied tasks are "active" when wholly unsatisfied ones aren't.  That isn't entirely insane IMO (these tasks are "in the process of becoming satisfied" which is kind of active, in a sense) ... but still, it probably makes more sense to exclude them from n=0.

We might be able to do this by simply keeping these tasks in the runahead pool until they are fully satisfied...

### LINKED ISSUE #3892: Correct stall diagnosis
[Better workflow completion handling (SoD Proposal)](https://cylc.github.io/cylc-admin/proposal-spawn-on-d.html#better-workflow-completion-handling)

Long story short:

Pre-SoD stall was pragmatically rather than conceptually grounded: the scheduler had literally got stuck and didn't know what to do about it. There were no more tasks to run **and** one or more failed or unsatisfied waiting tasks in the pool.

The *unsatisfied waiting tasks* part could lead to normal workflow completion being incorrectly identified as a stall, because of all the wholly-unsatisfied waiting tasks spawned ahead even though they might not be needed. 

Post-SoD there are no wholly-unsatisfied waiting tasks and there will soon (#3822) be no partially-satisfied ones either 
(just partially-satisfied prerequisites in a hidden pool, and as the example in the doc section linked to above shows they can't be used to reliably identify a stall).

What stall *should* mean: **the scheduler can't do anything more, but it knows that the flow is not finished.**

The only way valid to make that determination now is if there are unhandled failed tasks in the pool.  They are, by definition, task outcomes that were not meant to happen.

So: 
- if the active pool is empty:
     - **completed**
- else if the active pool contains only unhandled failed tasks:
     - **stalled**
- else:
    - still running

At normal shutdown or stall log any partially satisfied prerequisites in case they point to a flow design error, but in general we can't assume they were "meant" to be completed.

(Note special treatment of unhandled failed tasks is still under discussion; if that special treatment is revoked there will be no stall concept at all anymore).

hjoliver:
- close #3822 - keep partially-satisfied waiting tasks hidden in the runahead pool until fully satisfied
- close #3892 - correctly identify scheduler stall, by unhandled failed tasks only 

(The above two issues are closely related through how we handle partially-satisfied waiting tasks here and/or on master)

Also removes several checkpoint tests rather than adapt them to this branch, because they're about to be deleted anyway.

**Requirements check-list**

- [x] No documentation update required. (Will be covered by general Cylc 8 docs on n=windows)
- [x] No dependency changes.


hjoliver:
(I'm about to push up a bunch of commits here, but waiting on local functional tests first)

cylc/flow/scheduler.py
```diff
@@ -1672,6 +1671,7 @@ async def shutdown(self, reason):
             self.proc_pool.process()
 
         if self.pool is not None:
+            self.pool.report_unmet_deps()
```
hjoliver, line 1674: Log any partially satisfied tasks at shutdown as well as stall.

cylc/flow/scheduler.py
```diff
@@ -1525,7 +1524,7 @@ async def update_data_structure(self):
         updated_nodes = set(updated_tasks).union(
             self.pool.get_pool_change_tasks())
         if (
-                has_updated or
+                updated_nodes or
```
hjoliver, line 1527: The datastore wasn't updating on some changes (noticed via tests that use `cylc dump`, which now goes via GraphQL to the datastore).

cylc/flow/task_pool.py
```diff
@@ -176,7 +176,6 @@ def __init__(self, config, suite_db_mgr, task_events_mgr, job_pool):
 
         self.is_held = False
         self.hold_point = None
-        self.stuck_future_tasks = []
```
hjoliver, line 179: Stuck future tasks are another artifact of the old SoS algorithm: a waiting task gets spawned even though it depends on another task that is beyond the stop point.  Under SoD they don't get spawned (or if they have other non-future prereqs as well, they will only be spawned as hidden partially-satisfied tasks).

dwsutherland (approved):
LGTM 👍  

It does make the current data-store jump around a bit, but that will be dealt with soon (#3811), although (after testing with this branch), it will need a little work.

oliver-sanders (commented):
Non-blocking log format improvements:

cylc/flow/task_pool.py
```diff
@@ -914,77 +911,65 @@ def can_stop(self, stop_mode):
...
-                    itask.identity, itask.state.status))
+            if itask.state(*TASK_STATUSES_ACTIVE):
+                if itask.state.kill_failed:
+                    orphans_kill_failed.append(itask)
+                else:
+                    orphans.append(itask)
+        if orphans_kill_failed:
+            LOG.warning("Orphaned task jobs (kill failed):")
+            for itask in orphans_kill_failed:
+                LOG.warning(f"* {itask.identity} ({itask.state.status})")
```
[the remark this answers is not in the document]
hjoliver: Yep, nice, thanks.

hjoliver:
Log message changes applied @oliver-sanders (although they required some tweaking).

oliver-sanders (approved):
Works as expected for my examples.

Leaving un-merged incase you want to add a changelog, however, since SoD hasn't made release yet I don't think it's required.

hjoliver:
(I will update the SoD proposal to reflect this change)

Issue #3825, opened 2020-09-17
Platforms: Remove support for `platform = `some cmd`


wxtim:
In Cylc 7 the host setting for a task could be set as a bash subshell either in the form
```bash
$(my-command)   # or
`my-command`
```
my-command```. @dpmatthews proposes dropping support for the backtick ``my-command`` syntax for platforms at Cylc 8. 

Reasons why:
- This is very, very old syntax.
- Handling of single quotes and backslash escaping inside backticks is horrible.
- Nested backticks are hard to read.
- Backticks are visually difficult to distinguish from single quotes.
- Backticks are increasingly rarely used.

Are there any objections to doing this?
(I already have a branch for this purpose.)

n.b. Trying to write about this in GH Markdown is horrible.

hjoliver:
No objections!  Back-ticks are evil, confusing, and ugly.

PR #3826, opened 2020-09-18
Platforms.eval platform cmd.no backticks

wxtim:
Fix: #3825 , subsidiary response to #3672 

Built on #3791 

In Cylc 7 the host setting for a task could be set as a bash subshell either in the form
```bash
$(my-command)   # or
`my-command`
```
my-command```. @dpmatthews proposes dropping support for the backtick ``my-command`` syntax for platforms at Cylc 8. 

Reasons why:
- This is very, very old syntax.
- Handling of single quotes and backslash escaping inside backticks is horrible.
- Nested backticks are hard to read.
- Backticks are visually difficult to distinguish from single quotes.
- Backticks are increasingly rarely used.

### DIFF
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -31,7 +31,8 @@ FORBIDDEN_WITH_PLATFORM = (
 )
 
 # Regex to check whether a string is a command
-REC_COMMAND = re.compile(r'(`|\$\()\s*(.*)\s*([`)])$')
+HOST_REC_COMMAND = re.compile(r'(`|\$\()\s*(.*)\s*([`)])$')
+PLATFORM_REC_COMMAND = re.compile(r'(\$\()\s*(.*)\s*([)])$')
 
 
 def get_platform(task_conf=None, task_id='unknown task', warn_only=False):
@@ -69,11 +70,17 @@ def get_platform(task_conf=None, task_id='unknown task', warn_only=False):
         output = platform_from_name(task_conf)
 
     elif 'platform' in task_conf and task_conf['platform']:
-        if REC_COMMAND.match(task_conf['platform']) and warn_only:
+        if PLATFORM_REC_COMMAND.match(task_conf['platform']) and warn_only:
             # In warning mode this function might have been passed an
             # un-expanded platform string - warn that they won't deal with
             # with this until job submit.
             return None
+        if HOST_REC_COMMAND.match(task_conf['platform']) and warn_only:
+            raise PlatformLookupError(
+                f"platform = {task_conf['platform']}: "
+                "backticks are not supported; "
+                "please use $()"
+            )
 
         # Check whether task has conflicting Cylc7 items.
         fail_if_platform_and_host_conflict(task_conf, task_id)
--- a/cylc/flow/task_remote_mgr.py
+++ b/cylc/flow/task_remote_mgr.py
@@ -50,7 +50,6 @@ from cylc.flow.platforms import get_platform, get_host_from_platform
 from cylc.flow.remote import construct_platform_ssh_cmd
 
 
-REC_COMMAND = re.compile(r'(`|\$\()\s*(.*)\s*([`)])$')
 REMOTE_INIT_FAILED = 'REMOTE INIT FAILED'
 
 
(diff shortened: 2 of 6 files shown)


cylc/flow/task_remote_mgr.py
```diff
@@ -60,74 +59,86 @@ class TaskRemoteMgr:
...
+            command_re (re.Pattern):
+                A compiled regex pattern designed to match subshell strings.
 
         Return (str):
-            None if evaluate of host_str is still taking place.
-            'localhost' if host_str is not defined or if the evaluated host
-            name is equivalent to 'localhost'.
-            Otherwise, return the evaluated host name on success.
+            - None if evaluation of command is still taking place.
+            - If command is not defined or the evaluated name is equivelent
```
datamel, line 84:
```suggestion
            - If command is not defined or the evaluated name is equivalent
```

cylc/flow/platforms.py
```diff
@@ -63,10 +69,16 @@ def get_platform(task_conf=None, task_id='unknown task', warn_only=False):
         output = platform_from_name(task_conf)
 
     elif 'platform' in task_conf and task_conf['platform']:
+        if REC_COMMAND.match(task_conf['platform']) and warn_only:
```
wxtim, line 72: Fail in warn only mode if `platform == subshell with backticks`

datamel (commented):
So, I have just had a chat about this with you @wxtim. To summarise, I am happy with the change. I think the suite should fail validation if you have a platform = `something in backticks`.  Should be an easy enough change although will also make one of your new tests fail, so that will need fixing too.

cylc/flow/platforms.py
```diff
@@ -30,6 +30,10 @@
     ('job', 'batch submit command template', [None])
 )
 
+# Regex to check whether a string is a command
```
wxtim, line 33: I realized that I probably wanted to keep these in a single, centralized location.

cylc/flow/platforms.py
```diff
@@ -63,10 +70,21 @@ def get_platform(task_conf=None, task_id='unknown task', warn_only=False):
         output = platform_from_name(task_conf)
 
     elif 'platform' in task_conf and task_conf['platform']:
+        if PLATFORM_REC_COMMAND.match(task_conf['platform']) and warn_only:
+            # In warning mode this function might have been passed an
+            # un-expanded platform string - warn that they won't deal with
+            # with this until job submit.
+            return None
+        if HOST_REC_COMMAND.match(task_conf['platform']) and warn_only:
```
wxtim, line 78: This could be annoying if a user has systematically used backticks, but the tasks will fail later anyway, so we  might as well be annoying at validation.

hjoliver:
Just trying to separate the "built on top of 3791" bit, as currently:
```
$ git diff --shortstat wxtim/platforms.eval_platform_cmd.no_backticks master
 104 files changed, 2451 insertions(+), 2167 deletions(-)
```
It looked a lot simpler after local merge from master:
```
$ git diff --shortstat master
 6 files changed, 63 insertions(+), 19 deletions(-)
```

cylc/flow/platforms.py
```diff
@@ -63,10 +70,21 @@ def get_platform(task_conf=None, task_id='unknown task', warn_only=False):
...
+            raise PlatformLookupError(
+                f"platform - {task_conf['platform']} does not support "
+                "use of backticks: use $()"
```
hjoliver, line 81: This message is possibly a bit confusing? ("`command` does not support use of backticks"?):
```
PlatformLookupError: platform - `echo localhost` does not support use of backticks: use $()
```
Maybe:
```suggestion
                f"platform = {task_conf['platform']}: backticks are not supported; "
                "please use $()"
```

wxtim:
On Element @hjoliver  asked why we are doing this: and @dpmatthews said:
> Well, I think it is accepted good practise to use $() rather than backticks nowadays. We can choose to support whatever syntax we like so why are we supporting shell syntax which is no longer recommended? It's not really forcing users to change since this is a new configuration item - backticks will continue to work for host. As you say though, not a biggie - I'm happy either way.

@hjoliver said:
> Fair enough, let's do it.

datamel (approved):
I have tested this change and am happy with it. Initial feedback has been addressed.

oliver-sanders (approved):
Looks good, tests make sense, could do with rebase-squash'ing pre-merge.

wxtim:
> Looks good, tests make sense, could do with rebase-squash'ing pre-merge.

@oliver-sanders  Done.

oliver-sanders (approved):

PR #3829, opened 2020-09-21
initial work on removing dummy local

### LINKED ISSUE #3482: Deprecate dummy mode & make dummy-local work using a special job platform
## Problem
Dummy and dummy-local modes modify suite config `[runtime][TASK]` section that will be moved to a platform in the globalconfig file in Cylc 8. We cannot easily or safely modify the global config object because if we reload it any changes will disappear.

Dummy mode is effectively a residual test battery for the batch sys handlers and should be deprecated because it is very easy to test all things remote with a simple hand-rolled test suite.

[Dummy-local mode will be retained - it is sufficient for testing everything except remote job host connectivity ... I guess we should rename it it to "dummy mode" though, once the current non-local dummy mode is gone.]

## Proposed changes.
1. Deprecate dummy mode.

wxtim:
These changes close #3482, although I do not believe that all the steps outlined in the issue are necessary.

### DIFF
--- a/cylc/flow/cfgspec/suite.py
+++ b/cylc/flow/cfgspec/suite.py
@@ -166,18 +166,6 @@ with Conf(
                (e.g. ``+05:30``), given that the time zone is used as part of
                task output filenames.
         ''')
-        Conf('required run mode', VDR.V_STRING, '',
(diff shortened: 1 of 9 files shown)


wxtim:
Pushing this to one side pending discussions on issue

PR #3830, opened 2020-09-21
Ensure old host/batch system configs don't conflict with dummy-local mode

wxtim:
This is a small change not directly addressing any issue, although it touches on #3482.

@oliver-sanders  observed the following issue:
> Just spotted a platforms issue. When you run with --mode=dummy it adds platform=dummy. If the suite contains stuff like [remote]host or [job]batch system this will result in mixed syntax which causes the upgrader to fail and raise a PlatformLookupError.

This PR attempts to remedy this bug.


tests/functional/validate/76-host-to-platform-upgrade-dummy-mode.t
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
```
oliver-sanders, line 1: This shouldn't be in the validate tests, move in with the run tests or put somewhere else.

oliver-sanders (approved):

tests/functional/validate/76-host-to-platform-upgrade-dummy-mode.t
hjoliver: (maybe a new "platforms" test sub-dir?)

wxtim: how about a `upgrade7to8` folder - if used correctly it'll help us know whether we've deprecated the rather involved upgrade syntax?

hjoliver (approved):
LGTM, tests as working :+1:

Issue #3827, opened 2020-09-18
platforms: platform and host selection methods and intelligent fallback


oliver-sanders:
Follow on from #3686

* Implement an interface for supporting different selection methods (e.g. random, ordered, rank by psutil metrics, etc).
* Implement logic to request another host/platform if the first is not available (e.g. host down or network issues).

### Selection Methods

This should involve the creation of interfaces (preferably implemented as Python entry-points to permit in-house extension) for:

* Selection of a platform from a platform group.
* Selection of a host from a platform.

We will need to allow these to be configured separately e.g:

```ini
[platforms]
    [[one]]
        hosts = a, b, c
        method = random
    [[two]]
        hosts = d, e, f
        method = cycle

[platform-groups]
    [[three]]
        platforms = one, two
        method = random
    [[four]]
        platforms = two, one
        method = cycle
```

The following selection methods would be desirable but only the interface(s) need(s) to be implemented to close this issue:

* random (implemented)
* ordered (i.e. pick the first available host/platform from a list)
* cycle (pick hosts/platforms in order then go back the the start of the list - useful for testing)
* ranked (rank by CPU, memory or whatever and specify cutoffs #3800)

### Intelegent Fallback

In the event that a host is not available (e.g. for job submission) Cylc will need to pick an alternative host from the specified platform or platform group.

oliver-sanders:
@wxtim and I had a talk through the "intelligent fallback" part of this issue which has raised some questions...

The basic premice of this logic is that if a host goes down, rather than just failing Cylc can re-try the operation on another host. 

So how do we know if a host is down, there are two options:

1) Test the connection up-front.
    * This is what the current `rose host-select` logic does (it executes a short bash script on the remote host and sends some numbers back).
    * Testing connection to every host, every time we want to run a remote command is an overhead.
    * Even if SSH has been configured to reuse connections via "control persist" it is still an extra subprocess call.
    * These connection test processes could overwhelm the process pool, especially if there are connection issues.
2) Perform the operation (e.g. job submission) and diagnose any failure.
    * SSH exits 255 in the event of comms issues so could we forgo connection testing and just use this signal?
    * Saves an un-necessary connection/process.

However, not all issues are comms based, for example, what if the platform is not accepting new jobs, say because the queue is full or closed? This seems like the sort of thing Cylc should be able to handle gracefully. In this case there is no point trying another host within the platform, however, it may be worth trying another platform within the group.

Should we:

3) Only handle SSH failure.
[...]

PR #3832, opened 2020-09-22
Platform from platform group basic unit test

wxtim:
Precursor to #3827 
Partial fix of issues outstanding from #3793 

## store platform group name
Probably required for #3827
## refactor platforms tests
The file `tests/unit/test_platforms.py` had become large. I moved tests of the function `get_platform` to a separate file: `test_platforms_get_platform.py`
## add test for random platform selection
The test for random platform selection `tests/flakyfunctional/platform_aliases/00-simple.t` makes a trade off between run time and probability of giving a false negative. For some reason I failed to use the Pytest unit testing framework at `tests/unit/test_platforms.py`. By using the speed of pytest I was able to increase the number of iterations of the function carried out in a reasonable time, and reduce the probability of flakiness to about 1.6x10^-30 - I think this might be overkill, but it's not slow, and effectively not flaky.

### Question
Can/Should I delete `tests/flakyfunctional/platform_aliases`?


hjoliver:
> and reduce the probability of flakiness to about 1.6x10^-30 

:rofl:

hjoliver:
> Can/Should I delete tests/flakyfunctional/platform_aliases?

Do you mean move it out of the "flaky" category? (Not delete it outright).  If so, yes - going by your description of the chance failure, it is now the least flaky functional test in the entire universe.

wxtim:
> > Can/Should I delete tests/flakyfunctional/platform_aliases?
> 
> Do you mean move it out of the "flaky" category? (Not delete it outright). If so, yes - going by your description of the chance failure, it is now the least flaky functional test in the entire universe. [UPDATE: it's not entirely clear to me from your description if that applies to the functional test or the unit tests??]

I think that the not minutely flaky unit test covers exactly the same ground as the slightly flaky functional test. I was hoping that @oliver-sanders might double check my reasoning that the functional test can be discarded.

tests/unit/test_platforms_get_platform.py
```diff
@@ -0,0 +1,199 @@
...
+def test_get_platform_warn_mode(caplog):
+    task_conf = {
+        'remote': {'host': 'cylcdevbox'},
+        'job': {
+            'batch system': 'pbs',
+            'batch submit command template': 'some template'
+        }
+    }
+    output = get_platform(task_conf, warn_only=True)
+    for forbiddenitem in (
```
datamel, line 162: I know this is not from this pr really, since you are just moving this test out, but I find that a forbiddenitem a bit hard to read.
```suggestion
    for forbidden_item in (
```

tests/unit/test_platforms_get_platform.py
```diff
@@ -0,0 +1,199 @@
...
+        'batch submit command template = some template',
+        'host = cylcdevbox',
+        'batch system = pbs'
+    ):
+        assert forbiddenitem in output
```
datamel, line 167:
```suggestion
        assert forbidden_item in output
```

tests/unit/test_platforms_get_platform.py
```diff
@@ -0,0 +1,199 @@
...
+            [[hebrew_letters]]
+                platforms = aleph, bet
+        '''
+    )
+    output = get_platform('hebrew_letters')
+    assert output['group'] == 'hebrew_letters'
+    names = []
+    # this number keeps the test fast but give about 1.6x10^-30 chance of
+    # a false negative.
+    flakiness_n = 100
```
datamel, line 191: Nice to be able to increase the flakiness number from 9 to 100.

datamel (approved):
I have checked out your branch, read the code and run the tests. I am happy with the change and I think it makes sense to move these tests to a new file. 
With regards to the flaky functional test, I think, if I have understood things correctly, that test can be deleted - a flakiness number of 100 is preferential to 9, and I think the unit tests now have this covered.

tests/unit/test_platforms_get_platform.py
wxtim: Given the speed of pytest it could go a lot higher without being a problem, but 10^-30 seems sufficient.

tests/unit/test_platforms_get_platform.py
```diff
@@ -0,0 +1,199 @@
...
+    flakiness = 2 / 2 ** flakiness_n
+    for i in range(flakiness_n):
+        names.append(get_platform('hebrew_letters')['name'])
+        if 'aleph' in names and 'bet' in names:
+            break
```
oliver-sanders, line 196: With this test written in Python there might be a nicer solution, the Python random library is only psudo-random. You can control the series of random numbers it produces by setting the [random seed](https://docs.python.org/3/library/random.html#random.seed).

```python
random.seed(1234)
assert [
    get_platform('hebrew_letters')['name']
    for _ in range(N)
] == [
    # results should come out in the same order each time
]
```

wxtim: This will, I think work rather well. Thanks for the point!

oliver-sanders:
> Can/Should I delete tests/flakyfunctional/platform_aliases?

Yes, I think we've gone to great lengths to prove that the random method does indeed work.

oliver-sanders (approved):

PR #3838, opened 2020-09-25
schema: move default into schema for set-outputs

oliver-sanders:
Related to https://github.com/cylc/cylc-ui/pull/504

The `setOutputs` mutation has no required fields which means that on the surface it looks like something you can do without providing any input (like `kill` or `trigger`).

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -1678,7 +1678,7 @@ class SetOutputs(Mutation, TaskMutation):
     class Arguments(TaskMutation.Arguments):
         outputs = List(
             String,
-            default_value=[],
+            required=True
         )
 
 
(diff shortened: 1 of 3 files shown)


hjoliver:
> The setOutputs mutation has no required fields which means that on the surface it looks like something you can do without providing any input (like kill or trigger).

You can - that was intentional.  90% of the time this will be used to set the "succeeded" output of the target task (IMO), so I made that the default.

hjoliver:
(I think there's a bug though: the explicit `--output=` option is only working for custom outputs? I'll check...)

oliver-sanders:
Aah, just found the default in the task pool. Ok, I've moved the default into the GraphQL schema so that it's more transparent. This works better for api-on-the-fly which is introspective so benefits from having more information in the schema.

dwsutherland (approved):
Great 👍 

(I expect a few of these kind of things might come up, now that the schema is the sole entry to acting on the scheduler)

hjoliver (approved):

PR #3841, opened 2020-09-28
Improve the thoroughness of dummy-local functional test

wxtim:
Follow up a larger PR on #3482 

I do not believe that previously there was any test that dummy-local turns off pre and post scripts.


kinow (commented):
Looks like one functional test job failed @wxtim :+1:

wxtim:
It was a flaky test.

kinow (approved):

datamel (approved):
So, I think once the change goes into master to fix the `tests/functional/platforms/05-host-to-platform-upgrade-fail.t`, this will then pass the test battery for me.

PR #3842, opened 2020-09-29
test-fix: platforms/05

wxtim:

@datamel found a test which hung rather than worked as desired. On examination I also fixed:
- Incorrect tab size for project.
- Added a grep to ensure that the suite had failed for the reason we are testing for.
- Added a description of the workflow to the metadata section of the test `flow.cylc`


datamel (approved):
Test no longer hangs and passes. Thanks @wxtim.

PR #3845, opened 2020-09-30
flow-cfg-update.authentication

wxtim:
Refers to Cylc8 [Config Proposal agreed at CylcCon Wellington](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-config-changes.md#cylcforce-run-mode--required-run-mode---obsolete).

Do we want to attach a follow on to carry out more heavy-duty removal of this functionality?

### DIFF
--- a/cylc/flow/cfgspec/suite.py
+++ b/cylc/flow/cfgspec/suite.py
@@ -21,7 +21,6 @@ from metomi.isodatetime.data import Calendar
 
 from cylc.flow import LOG
 from cylc.flow.parsec.exceptions import UpgradeError
-from cylc.flow.network.authorisation import Priv
 from cylc.flow.parsec.config import ParsecConfig, ConfigNode as Conf
 from cylc.flow.parsec.OrderedDict import OrderedDictWithDefaults
 from cylc.flow.parsec.upgrade import upgrader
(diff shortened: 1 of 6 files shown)


MetRonnie (approved):
Looks good as far as I can tell

cylc/flow/network/server.py
```diff
@@ -15,7 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """Server for suite runtime API."""
 
-import getpass
+import getpass  # noqa: F401
```
oliver-sanders, line 18: What's this F401 for?

cylc/flow/cfgspec/suite.py
```diff
@@ -21,7 +21,6 @@
 
 from cylc.flow import LOG
 from cylc.flow.parsec.exceptions import UpgradeError
-from cylc.flow.network.authorisation import Priv
```
oliver-sanders, line 24: Can remove the `Priv` class (as this PR effectively removes authorisation) and kill the `if usr_priv_level < req_priv_level:` check.

cylc/flow/network/server.py
wxtim: getpass is not used, but causes failures if it's removed.

oliver-sanders:
```
2020-10-01T10:25:53Z ERROR - pb_entire_workflow() got an unexpected keyword
	argument 'user'
	  File "/Users/oliver/cylc-flow/cylc/flow/network/server.py", line 245, in _receiver
	    response = method(**args)
	TypeError: pb_entire_workflow() got an unexpected keyword argument 'user'
```

MetRonnie (approved):

oliver-sanders:
@wxtim BTW this will require a sibling PR in cylc-doc

cylc/flow/network/server.py
hjoliver: What failures occurred?  I just tried removing this line and didn't see any problems on running a flow.

Issue #3667, opened 2020-06-30
runahead limit / max active cycle points


oliver-sanders:
*After discussion on Riot:*

We should consider re-aligning `max active cycle points` to match the SoD implementation (and the configuration name).

**Context:**

The `runahead limit` and `max active cycle points` configurations both limit the span between the oldest and youngest active cycle points.

From the docs for `max active cycle points`:

> It allows up to N (default 3) consecutive cycle points to be active at any time, adjusted up if necessary for any future triggering.

The key word here is *consecutive*.

So `runahead limit` and `max active cycle points` actually both do the same thing, just that one uses integer semantics and the other, date-time.

**Issues:**

1. The consecutive aspect of `max active cycle points` doesn't gel well with SoD. [HO: triggering a reflow in past cycles will completely stop new cycles, depending on the limit; same goes for SoS really, if you insert-and-trigger an old task]
2. The name `max active cycle points` is misleading since it's nothing to do with the number of active cycle points.

**Questions:**

1. Should we combine the current `max active cycle points` configuration in with `runahead limit`?
    * `runahead limit = P1D`
    * `runahead limit = P3` (equivalent to [current consecutive] `max active cycle points = 3`)
    * This would prevent confusing dual-specification of two orthogonal runahead limits. **yes**
2. Should we add a non-consecutive limit which lives up to the name `max active cycle points`? **yes**
     * Is there a good use case for this?

hjoliver:
> Questions

1. yes, good idea
2. yes, if we can't think of a good reason not to (I will see if I can find out why we changed to the consecutive version in the first place)
 
The use case for non-consecutive `max active cycle points` is pretty general, it seems to me: we should be aiming for maximum concurrency within prescribed limits; and why would we stop two tasks from running concurrently when they are (say) 3 cycle points apart but allow them to do it if (say) 2 cycle points apart?

hjoliver:
If we allow non-consecutive `max active cycle points`, should we sort active and waiting(runahead) tasks into cycle point order and prioritize older points?  (Older would tend to allow a reflow to get done first, and if you trigger a reflow in the first place that is probably what you want?)

oliver-sanders:
> should we sort active and waiting(runahead) tasks into cycle point order and prioritize older points

I would have thought so.

There's also the question of defaults, personally I would prefer `runahead limit` and `max active cycle points` to default to `None`. Unfortunately that might not be advisable due to graphs like this:

```
c[-P1] => a
[...]

PR #3848, opened 2020-10-01
Move `max active cycle points` behaviour into `runahead limit`

MetRonnie:
These changes partially address #3667 

The functionality of setting a runahead limit of a certain number of _consecutive_ cycle points is being moved from `[scheduling]max active cycle points` to `[scheduling]runahead limit`.

In future, `max active cycle points` will set a limit on _non-consecutive_ cycle points (see https://github.com/cylc/cylc-flow/issues/3667#issuecomment-701504225 for info on the difference between consecutive and non-consecutive limits), but this is not implemented in this PR. **Edit:** `max active cycle points` is being deprecated and a new setting will set the non-consecutive limit.

Examples:

| before                        | after                 |
| ----------------------------- | --------------------- |
| `max active cycle points = 3` | `runahead limit = P3` |
| `runahead limit = PT12H`      | (same as before)      | 

#### Requirements check-list

- [X] Appropriate change log entry included.

- [X] (master branch) I have opened a documentation PR at https://github.com/cylc/cylc-doc/pull/161

- [x] No dependency changes.


cylc/flow/task_pool.py
```diff
@@ -298,7 +305,7 @@ def release_runahead_tasks(self):
             sequence_points = []
             for sequence in self.config.sequences:
                 point = runahead_base_point
-                for _ in range(limit):
+                for _ in range(number_limit or sequence._MAX_CACHED_POINTS):
```
MetRonnie, line 308: This uses a limit of 100 if `number_limit` is `None`.

Would it also be worth setting a maximum of `runahead limit = P100`?

tests/unit/test_config.py
```diff
@@ -455,3 +455,38 @@ def _test(cp_tz, utc_mode, expected, expected_warnings=0):
...
+        ({'cycling mode': 'gregorian', 'runahead limit': 'P14'}, True),
+        ({'cycling mode': 'gregorian', 'runahead limit': 'PT12H'}, True),
+        ({'cycling mode': 'gregorian', 'runahead limit': 'P7D'}, True),
+        ({'cycling mode': 'gregorian', 'runahead limit': 'P2W'}, True),
+        ({'cycling mode': 'gregorian', 'runahead limit': '4'}, True),
+
+        ({'cycling mode': 'integer', 'runahead limit': 'PT12H'}, False),
+        ({'cycling mode': 'integer', 'runahead limit': 'P7D'}, False),
+        ({'cycling mode': 'integer', 'runahead limit': '4'}, False),
+        ({'cycling mode': 'gregorian', 'runahead limit': ''}, False),
```
MetRonnie, line 473: It is not possible to set `runahead limit` to null

cylc/flow/cfgspec/suite.py
```diff
@@ -351,27 +351,26 @@
...
-            consecutive cycle points to be active at any time, adjusted up if
-            necessary for any future triggering.
-        ''')
+            This limit on the number of consecutive cycle points is specified
+            by an interval between the least and most recent: either an integer
+            interval (e.g. ``P3`` -  works for both :term:`integer cycling` and
+            :term:`datetime cycling`), or a time interval (e.g. ``PT12H`` -
+            only works for datetime cycling). Alternatively, if a raw number is
+            given, it will be taken to mean that number of hours, though this
+            usage is deprecated.
```
MetRonnie, line 365: I don't think this format of `Pn` (e.g. `P3`) for a number limit is good, because if you have
```ini
[[graph]]
    P2 = foo  # where foo fails
```
That means the cycle points will be `foo.1`, `foo.3`, `foo.5` etc. But the runahead limit will stall at `foo.5` not `foo.3` because `foo.5` is actually the 3rd cycle point.

So `runahead limit = 3` instead of `P3` would make more sense. But the trouble is it would break the backwards compatibility where the raw number is interpreted as `PT3H`

cylc/flow/cfgspec/suite.py
```diff
@@ -351,27 +351,26 @@
...
+        ''')  # TODO: mention :cylc:conf:`[..]max active cycle points`
+        Conf('max active cycle points', VDR.V_INTEGER, None, desc='''
+        ''')  # TODO: rewrite desc after non-consecutive implementation
```
oliver-sanders, line 368: What do we want to do with `max active cycle points`? Options:

1) Deprecate it and upgrade value to `runahead limit`. When we get around to the non-consecutive limit we would have to use a different configuration name or move it to a new section.

2) Keep it and upgrade existing non-consecutive limits to consecutive ones.

tests/functional/runahead/default-complex/flow.cylc
```diff
@@ -10,12 +11,12 @@
...
         PT7H = "foo & oops => bar"
-        # T00, T12, T18... 
+        # T00, T12, T18...
         T00, T12, T18 = "foo"
         # T04...
-        T04 = "run_ok & fail"
+        T04 = "run_ok"
         # T05...
-        T05 = "run_ok_2 & fail"
+        T05 = "run_ok_2"
```
MetRonnie, line 19: It doesn't matter whether these succeed or fail, as `foo.T00` has already failed, so that is what limits the runahead

cylc/flow/task_pool.py
oliver-sanders: I don't think that we would want to enforce an upper bound on the runahead limit.

tests/functional/runahead/01-check-default-simple.t
```diff
@@ -30,10 +30,10 @@ run_fail "${TEST_NAME}" cylc run --debug --no-detach "${SUITE_NAME}"
 #-------------------------------------------------------------------------------
 TEST_NAME="${TEST_NAME_BASE}-max-cycle"
 DB="${SUITE_RUN_DIR}/log/db"
-run_ok "${TEST_NAME}" sqlite3 "${DB}" "select max(cycle) from task_states"
-cmp_ok "${TEST_NAME}.stdout" <<'__OUT__'
-20100101T1800Z
-__OUT__
+run_ok "${TEST_NAME}" sqlite3 "${DB}" \
+    "select max(cycle) from task_states where status!='waiting'"
```
MetRonnie, line 34: If I understand correctly @hjoliver , the database contains tasks from the task pool + runahead pool. The latter contains only waiting tasks that have not spawned. So to correctly identify the max cycle point (which has spawned tasks), we can search just the task pool by eliminating the waiting tasks?

cylc/flow/cfgspec/suite.py
oliver-sanders: Well, actually, it wouldn't break backwards compatibility because this issue only occurs in integer cycling mode. It might make it a bit harder to parse the value though.

cylc/flow/task_pool.py
MetRonnie: But if a workflow has no final cycle point, this for loop hang?

oliver-sanders: Would have to re-work the logic to permit that.

cylc/flow/cfgspec/suite.py
oliver-sanders: Thinking about it a bit I think we have multiple suites built to the same pattern which are dependent on consecutive behaviour which would constrain us to (1).

cylc/flow/cfgspec/suite.py
oliver-sanders: The alternatives would be:

* Stick to `P<n>` but document the hell out of it.
* Come up with another prefix character (confusing).

MetRonnie: > Stick with `P<n>`, just document the hell out of it.

That seems the best option to me, if we can't afford to lose the raw hours back compat.

oliver-sanders: Sadly the old feature was never actually deprecated and is widely used.

tests/functional/runahead/01-check-default-simple.t
MetRonnie: Actually I'm not so sure about that explanation as I have just seen the `task_pool` table contains only the waiting tasks at the end. But regardless, the only "active" cycle points (which show up in the TUI) seem to be the ones that have at least 1 task whose state is not `waiting`.

cylc/flow/task_pool.py
hjoliver: > But if a workflow has no final cycle point, there would be an infinite number of points to loop over, wouldn't there?

The loop is over the task pool, and tasks with no prerequisites spawn sequentially (when one is released from the runahead pool, the next one is spawned into the runahead pool). So that provides some constraint, but such tasks still do proliferate pretty quickly.

cylc/flow/cfgspec/suite.py
```diff
@@ -358,27 +358,37 @@
...
+            specified by an interval between the least and most recent: either
+            an integer (e.g. ``P3`` -  works for both :term:`integer cycling`
+            and :term:`datetime cycling`), or a time interval (e.g. ``PT12H`` -
+            only works for datetime cycling). Alternatively, if a raw number is
+            given, e.g. ``7``, it will be taken to mean ``PT7H``, though this
+            usage is deprecated.
 
-            Example: ``PT12H`` - for a 12 hour limit under ISO 8601 cycling.
-        ''')
-        Conf('max active cycle points', VDR.V_INTEGER, 3, desc='''
```
MetRonnie, line 373: Should the conf node item actually be left in for the sake of documentation (with a note that it's deprecated), or is removing it better?

cylc/flow/etc/syntax/cylc.lang
```diff
@@ -136,7 +136,6 @@
         <keyword>number of cycle points</keyword>
         <keyword>members</keyword>
         <keyword>max-polls</keyword>
-        <keyword>max active cycle points</keyword>
```
MetRonnie, line 139: Should these syntax highlighting keywords be removed or not until the setting is obsoleted?

cylc/flow/task_pool.py
MetRonnie: But in this line it's looping over all on-sequence points up to a certain limit, before determining the task pool. For a datetime cycling workflow with a runhead limit specified as a time interval, and no final cycle point, surely there has to be some built-in limit to the number of future points to find, otherwise the loop would hang

cylc/flow/cfgspec/suite.py
oliver-sanders: When we bring this back it will have to be with a different name so best to remove.

cylc/flow/task_pool.py
MetRonnie: Update: I've come up with a way of getting the on-sequence points up to the (integer or time interval) runahead limit without imposing the arbitrary 100 point limit - https://github.com/cylc/cylc-flow/pull/3848/commits/f6b615abf9afb74c41398369f04ab682e240b47e

hjoliver: (Sorry, I didn't look at the actual code before making that comment! :grimacing: )

tests/functional/runahead/01-check-default-simple.t
hjoliver: > If I understand correctly @hjoliver , the database contains tasks from the task pool + runahead pool.

Yes, runahead pool tasks appear as just "waiting" in the task_pool table.  On restart, all task_pool table entries are initially dropped into the runahead pool, so there's no need to distinguish the two pools in the DB.

hjoliver: >  The latter [runahead pool] contains only waiting tasks that have not spawned.

I'd put it as, tasks that have been spawned but whose cycle points are still beyond the current runahead limit.  But in effect you're correct - tasks in the runahead pool can't have spawned successors yet because they haven't submitted yet (also, if a task's own next instance - i.e. same task at next cycle point - has no parents, it will be spawned when when the current instance leaves the runahead pool).

cylc/flow/cfgspec/suite.py
hjoliver: You're probably right. Unfortunately, because "max active cycle points" is the obvious best name for the new item.  I guess we could go for "maximum active cycle points" :grin:

hjoliver: https://github.com/cylc/cylc-flow/issues/3667#issuecomment-705183728

cylc/flow/etc/syntax/cylc.lang
hjoliver: Can't remember if we've done this consistently in the past.  But IMO yes, it might provide additional incentive to upgrade deprecated items.

cylc/flow/task_pool.py
oliver-sanders: Workaround looks good to me.

tests/unit/test_config.py
oliver-sanders: Sensible, an infinite runahead limit is a whole other kettle of fish!

tests/functional/runahead/01-check-default-simple.t
oliver-sanders: (note that Tui [and the UI] gets its data from the "data store" not the DB, the two have a different "windowing")

cylc/flow/etc/syntax/cylc.lang
oliver-sanders: (we will need to re-do these at some point)

MetRonnie: I'll open an issue for that?

tests/functional/runahead/01-check-default-simple.t
hjoliver: BTW the DB query above works in this instance but isn't the best way to diagnose runahead in general, but in investigating that I found that the better way is currently stymied by the "abort on stall" which doesn't check if any runahead tasks can be released before aborting.  But I'll put up a new PR for that, on top of this one.

cylc/flow/etc/syntax/cylc.lang
hjoliver: (I see you've done that #3860 )

cylc/flow/cfgspec/suite.py
```diff
@@ -358,27 +358,37 @@
...
+               task repeats every cycle.
+
+            .. note::
+
+               The runahead limit may be automatically raised if this is
+               necessary to allow a future task to be triggererd, preventing
+               the suite from stalling.
         ''')
+        # TODO: make raw number of hours obsolete and then later allow a raw
+        # number to mean the integer limit ?
```
oliver-sanders, line 391: Future plans should be recored in issues. Comments get mixed up, forgotten about and become hard to interpret later. (That said the above applies to issues too, just, hopefully, to a slightly lesser extent.)

```suggestion
```

oliver-sanders (approved):
Few tiny comments.

tests/functional/runahead/01-check-default-simple.t
hjoliver: (Not worth putting time into my previous comment; it only affects tests. I'll resolve this discussion and simply put a code comment in this test post merge, in case the test gets flaky in the future).

hjoliver (approved):
All good, thanks @MetRonnie :+1:

hjoliver:
(pycodestyle failed; I'll fix that in my add-a-comment follow-up).

Issue #3472, opened 2020-01-02
host selection command failure not handled properly


dpmatthews:
Tested with 7.8.4 using the following suite:
```ini
[scheduling]
    [[dependencies]]
        graph = root
[runtime]
    [[bad-host]]
        [[[remote]]]
            host = bad
#            host = $(rose host-select bad)
```

With `host = bad` the suite log contains
```
2020-01-02T10:00:43Z ERROR - [remote-init cmd] cat '<tempfile.SpooledTemporaryFile instance at 0x7f92c4dd1d40>' | cylc remote-init --host=bad 29cfe816-38b2-4fb3-923e-9fe6f682afe2 '$HOME/cylc-run/test.submision-failure'
	[remote-init ret_code] 1
	[remote-init err]
	ssh: Could not resolve hostname bad: Name or service not known

	ERROR: command returns 255: ssh -oBatchMode=yes -oConnectTimeout=8 -oStrictHostKeyChecking=no bad env CYLC_VERSION=7.8.4 bash --login -c ''"'"'exec "$0" "$@"'"'"'' cylc remote-init 29cfe816-38b2-4fb3-923e-9fe6f682afe2 ''"'"'$HOME/cylc-run/test.submision-failure'"'"''
2020-01-02T10:00:43Z INFO - [bad-host.1] -submit-num=01, owner@host=bad
2020-01-02T10:00:43Z ERROR - [jobs-submit cmd] (init bad)
	[jobs-submit ret_code] 1
	[jobs-submit err] REMOTE INIT FAILED
```
and the job activity log contains
```
[jobs-submit cmd] (init bad)
[jobs-submit ret_code] 1
[jobs-submit err] REMOTE INIT FAILED
```
This looks reasonable although it would be nice if the job activity log contained the ssh output.

With `host = $(rose host-select bad)` the suite log contains
```
2020-01-02T11:07:04Z ERROR - [remote-host-select cmd] timeout 10 bash -c 'rose host-select bad'
	[remote-host-select ret_code] 1
	[remote-host-select err]
	[WARN] bad: (ssh failed)
	[FAIL] No hosts selected.
2020-01-02T11:07:04Z ERROR - rose host-select bad: host selection failed:
	COMMAND FAILED (1): rose host-select bad
	COMMAND STDERR: [WARN] bad: (ssh failed)
	COMMAND STDERR: [FAIL] No hosts selected.

[...]

PR #3850, opened 2020-10-02
fix warning for uneval'd platform subshell

wxtim:
Fix issue discovered in https://github.com/cylc/cylc-flow/issues/3472

### DIFF
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -74,7 +74,10 @@ def get_platform(task_conf=None, task_id='unknown task', warn_only=False):
             # In warning mode this function might have been passed an
             # un-expanded platform string - warn that they won't deal with
             # with this until job submit.
-            return None
+            return (
+                f"{task_conf['platform']} will be evaluated at Job Submission "
+                "and may fail then."
+            )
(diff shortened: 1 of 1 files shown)


cylc/flow/platforms.py
```diff
@@ -74,7 +74,10 @@ def get_platform(task_conf=None, task_id='unknown task', warn_only=False):
             # In warning mode this function might have been passed an
             # un-expanded platform string - warn that they won't deal with
             # with this until job submit.
-            return None
+            return (
+                f"{task_conf['platform']} will be evaluated at Job Submission "
+                "and may fail then."
+            )
```
oliver-sanders, line 80: I think this was removed after a review comment.

This would raise a warning every time someone tries to use a subshell in a platforms item telling the user that the command could fail?

If so this is a bit over the top, a warning in the docs should be more than sufficient.

wxtim:
@oliver-sanders  said:

> I think this was removed after a review comment.

> This would raise a warning every time someone tries to use a subshell in a platforms item telling the user that the command could fail?

> If so this is a bit over the top, a warning in the docs should be more than sufficient.

Issue #3688, opened 2020-07-15
platforms: support symlinking directories to configured locations


oliver-sanders:
Bullet point (2) "Support moving some directories to different locations with symlinks to the original location" from the [platforms proposal document](https://cylc.github.io/cylc-admin/proposal-platforms.html#enhancements-to-support-rose-suite-run-functionality).

Re-implement the [`root-dir`](http://metomi.github.io/rose/doc/html/api/configuration/site-and-user.html#rose:conf:rose.conf[rose-suite-run]root-dir) functionality from `rose suite-run` in Cylc. This allows, for example, the `log` directory to be placed on a more appropriate filesystem.

**Pull requests welcome!**

oliver-sanders:
This issue comes in two parts:

1. The initial creation of the directories on the "localhost" platform when the suite is run.
2. The subsequent creation of these directories on other install targets (filesystems) during the remote-init process when the first job is submitted to a new install target.

hjoliver:
Note that run-dir symlinking will render `cylc register --run-dir=RUNDIR` obsolete. The PR for this Issue should remove that option.

PR #2935, opened 2019-01-29
Support symlinked alternate run directories (cylc 8).

hjoliver:
These changes close #2797

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Appropriate tests are included (functional).
- [x] Includes an appropriate entry in the release change log `CHANGES.md`.
- [x] I have opened a corresponding documentation PR at cylc/cylc-doc#77.

Supersedes #2817 (On reflection my initial take on this was unnecessarily focused - for historical reasons - on the idea of using an absolute suite path (instead of a registered suite name) to distinguish this from a normal run.  In fact, registration is really an entirely orthogonal concept.  The basic requirement (/desired feature) is:

 _Better support for quick-running sub-suites (of jobs that run on a single node?) by allowing sub-suite run directories to be located on fast local disk or RAM disk._

Why? 1)  too many small files kills HPC shared filesystems like Lustre; and 2) top-level suite run directories rapidly proliferate if you use sub-suites in a cycling suite.  

This trivial PR just adds a new registration option `--run-dir=DIR` that symlinks the standard suite run directory to an alternative location, so all suite job logs get written there.

```
$ cylc reg --run-dir=/tmp cat/dog suites/foo
REGISTERED cat/dog -> /home/oliverh/suites/foo

$ ls -l cylc-run/cat/
lrwxrwxrwx 1 oliverh oliverh 12 Jan 29 14:50 dog -> /tmp/cat/dog/

$ ls -a /tmp/cat/dog/
./  ../  .service/
```

Everything works as normal (including `cylc scan` and `cylc review`) but the log files are on (e.g.) `/tmp`. 

We still have (for the moment; see below) proliferation of sub-suite run directories, but at least they're not on the shared FS.

PR #2817, opened 2018-10-24
WIP: Local run dir mode

hjoliver:
Close #2797 

Currently `cylc register NAME PATH` associates suite name NAME and run directory `~/cylc-run/NAME/` with a suite definition in PATH. At run time the server is targetted by name, the run directory located by name (and all task job logs are captured there). 

Following the discussion in https://github.com/cylc/cylc/issues/2749#issuecomment-414978342 this PR  supports a new mode in which suites are not registered and are identified by source path rather than name:
- `cylc run ABS-PATH`
- run directory is `ABS-PATH/cylc-run/`
- suite name is `ABS-PATH`  (thus `cylc stop ABS-PATH` etc.)
- the suite is invisible to plain `cylc scan/gscan` (which looks in the standard run dir) but can be seen with `cylc scan/gscan -a` (port scan) 

These suites still have the full verbose set of cylc job logs, BUT:
- they do not pollute the standard cylc-run directory
  - thus avoiding the sub-suite house-keeping problem
- they could be located on fast local filesystem or RAM disk
  - they can be house-kept by the main-suite tasks that run them.

This might be a good interim solution to some post-processing problems - ping @benfitzpatrick @TomekTrzeciak  - particularly for small quick-running sub-suites that can be treated like tasks (so no need to worry about the fact that mid-run restarts aren't handled automatically by the main suite).

If agreed, I will document (including advice on what to do about restarts etc.) and add tests.

It might be stretch to get this in next-release (next week) ... but it is a small change that is orthogonal to existing functionality.

PR #3852, opened 2020-10-05
Disallow nested run directories

### LINKED ISSUE #3835: Disallow nested run directories
Reported by @kinow 

Currently you can register and run nested flows:
```bash
$ cylc register f1 ~/suites/foo
$ cylc register f1/f2 ~/suites/foo
```

Cylc scan (and hence the UI) does not see the running `f1/f2` flow though, because it stops recursing deeper into `cylc-run/` whenever it finds a valid run directory.

Can't think of a good reason to allow this - particularly given this `cylc scan` problem and the potential for *huge* run directories. 

We should prevent this at the registration stage.

MetRonnie:
These changes close #3835

This PR prevents registering a suite in a subdirectory of another run directory, or registering a suite in a directory whose children contain a run directory. It does this by checking upwards for the existence of a flow config file or a `.service` dir in successive parent dirs up to the cylc run dir, as well as checking downwards in subdirectories to a depth of 4 (the default `cylc scan` depth).

**Requirements check-list**

- [X] Appropriate tests are included (unit).

- [X] Appropriate change log entry included.

- [X] No dependency changes.

### DIFF
--- a/cylc/flow/suite_files.py
+++ b/cylc/flow/suite_files.py
@@ -481,6 +481,7 @@ def register(reg=None, source=None, redirect=False, rundir=None):
             No flow.cylc file found in source location.
             Illegal name (can look like a relative path, but not absolute).
             Another suite already has this name (unless --redirect).
+            Trying to register a suite nested inside of another.
     """
     if reg is None:
         reg = os.path.basename(os.getcwd())
@@ -494,6 +495,8 @@ def register(reg=None, source=None, redirect=False, rundir=None):
     if os.path.isabs(reg):
         raise SuiteServiceFileError(
             "suite name cannot be an absolute path: %s" % reg)
+
+    check_nested_run_dirs(reg)
 
     if source is not None:
         if os.path.basename(source) == SuiteFiles.FLOW_FILE:
@@ -669,3 +672,37 @@ def _load_local_item(item, path):
             return file_.read()
     except IOError:
         return None
+
+
+def check_nested_run_dirs(reg):
+    """Disallow nested run dirs e.g. trying to register foo/bar where foo is
+    already a valid suite directory.
+
+    Args:
+        reg (str): suite name
+    """
+    parent_dir = os.path.dirname(os.path.normpath(reg))
+    while parent_dir != '':
+        if _is_valid_run_dir(parent_dir):
+            raise SuiteServiceFileError(
+                'Nested run directories not allowed - cannot register suite '
+                f'name "{reg}" as {parent_dir} is already a valid '
+                'run directory.')
+        parent_dir = os.path.dirname(parent_dir)
+
+
+def _is_valid_run_dir(path):
+    """Return True if path is a valid suite run directory, else False.
+
+    Args:
+        path (str): if this is a relative path, it is taken to be relative to
+            the Cylc run directory.
+    """
+    if not os.path.isabs(path):
+        cylc_run_dir = os.path.expandvars(get_platform()['run directory'])
+        path = os.path.join(cylc_run_dir, path)
+    if (os.path.isfile(os.path.join(path, SuiteFiles.FLOW_FILE)) or
+            os.path.isdir(os.path.join(path, SuiteFiles.Service.DIRNAME)) or
+            os.path.isfile(os.path.join(path, SuiteFiles.SUITE_RC))):
+        return True
+    return False
--- a/cylc/flow/suite_files.py
+++ b/cylc/flow/suite_files.py
@@ -345,8 +345,7 @@ def get_contact_file(reg):
 def get_flow_file(reg, suite_owner=None):
     """Return the path of a suite's flow.cylc file."""
     return os.path.join(
-        get_suite_source_dir(reg, suite_owner),
-        SuiteFiles.FLOW_FILE)
+        get_suite_source_dir(reg, suite_owner), SuiteFiles.FLOW_FILE)
 
 
 def get_suite_source_dir(reg, suite_owner=None):
@@ -488,13 +487,11 @@ def register(reg=None, source=None, redirect=False, rundir=None):
 
     is_valid, message = SuiteNameValidator.validate(reg)
     if not is_valid:
-        raise SuiteServiceFileError(
-            f'invalid suite name - {message}'
-        )
+        raise SuiteServiceFileError(f'invalid suite name - {message}')
 
     if os.path.isabs(reg):
         raise SuiteServiceFileError(
-            "suite name cannot be an absolute path: %s" % reg)
+            f'suite name cannot be an absolute path: {reg}')
 
     check_nested_run_dirs(reg)
 
@@ -556,14 +553,12 @@ def register(reg=None, source=None, redirect=False, rundir=None):
     if orig_source is not None and source != orig_source:
         if not redirect:
             raise SuiteServiceFileError(
-                "the name '%s' already points to %s.\nUse "
-                "--redirect to re-use an existing name and run "
-                "directory." % (reg, orig_source))
+                f"the name '{reg}' already points to {orig_source}.\nUse "
+                "--redirect to re-use an existing name and run directory.")
         LOG.warning(
-            "the name '%(reg)s' points to %(old)s.\nIt will now"
-            " be redirected to %(new)s.\nFiles in the existing %(reg)s run"
-            " directory will be overwritten.\n",
-            {'reg': reg, 'old': orig_source, 'new': source})
+            f"the name '{reg}' points to {orig_source}.\nIt will now be "
+            f"redirected to {source}.\nFiles in the existing {reg} run "
+            "directory will be overwritten.\n")
         # Remove symlink to the original suite.
         os.unlink(os.path.join(srv_d, SuiteFiles.Service.SOURCE))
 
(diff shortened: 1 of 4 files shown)


cylc/flow/suite_files.py
```diff
@@ -669,3 +667,37 @@ def _load_local_item(item, path):
...
+        path (str): if this is a relative path, it is taken to be relative to
+            the Cylc run directory.
+    """
+    if not os.path.isabs(path):
+        cylc_run_dir = os.path.expandvars(get_platform()['run directory'])
+        path = os.path.join(cylc_run_dir, path)
+    if (os.path.isfile(os.path.join(path, SuiteFiles.FLOW_FILE)) or
+            os.path.isdir(os.path.join(path, SuiteFiles.Service.DIRNAME)) or
+            os.path.isfile(os.path.join(path, SuiteFiles.SUITE_RC))):
+        return True
```
MetRonnie, line 702: Instead of checking for the existence of the `.service` dir, should we check that `.sevice/source` is an existing symlink?

MetRonnie:
Still 2 unresolved points Oliver commented on the issue:
> 1. In-place installations (e.g. run a suite in a temp dir)
>    - Should we permit these, e.g. in the work dir (for subsuites).
> 2. Might be a good time to remove the `cylc register --run-dir=RUNDIR` option.
>    - Obsolete once we have symlink support built-in.

cylc/flow/suite_files.py
hjoliver: Existing or valid?

Are you suggesting we disallow flows inside existing valid run directories, but we allow them inside existing invalid ones?

That seems untidy at best, to me (is there is use case?).

hjoliver:
We can already *run a suite in a temp dir* using `--run-dir` which (as per your second point) creates a symlink under cylc-run so that everything behaves normally even though the run-dir is actually located somewhere else.

> In-place installations (e.g. run a suite in a temp dir)

Presumably by *in-place installations* you mean run a flow in-place without installing it to the normal cylc-run location? (Like the original pre-Rose `cylc run` :rofl: ).  The trouble is, that's not discoverable by FS scan and other tools. But maybe that's OK if we have real built-in support for sub-suites, so that you're actually not supposed to be able to discover them as independent entities as they are now.

>-    Should we permit these, e.g. in the work dir (for subsuites).

I suppose `cylc-run/foo/work/sub1,sub2,...` would be a sensible place for in-place sub-suites.  It wouldn't mess with `cylc scan` too much (only need to look in one defined location below `foo`; or don't make them independently discoverable). 
 
> Might be a good time to remove the cylc register --run-dir=RUNDIR option.
>    Obsolete once we have symlink support built-in.

I'd rather that got removed by the PR that makes it obsolete.

cylc/flow/suite_files.py
MetRonnie: I was wondering whether testing for `.service/source` was a more appropriate gauge of whether the directory was valid, but I guess just `.service` is good enough

oliver-sanders:
@hjoliver I think this is the PR (https://github.com/cylc/cylc-flow/pull/2817) which (re)instated what I called the "run in place" model, you've called it "local run mode".

I think the desire is that functionality should not be effected by any change here.

hjoliver:
> @hjoliver I think this is the PR (#2817) which (re)instated what I called the "run in place" model, you've called it "local run mode".

I called it "local run**-dir** mode", which makes a bit more sense than "local run mode". 

That PR was actually aborted and superseded by #2935 ("support symlinked alternate run-directories") which is where the `cylc reg --run-dir` option appeared.

hjoliver:
> 
> I think the desire is that functionality should not be effected by any change here.

I agree, in which case (in light of my previous comment) I'm a bit confused about your statement *Might be a good time to remove the `cylc register --run-dir=RUNDIR` option.* :thinking:

hjoliver:
By "run in place" model, do you mean "run a suite that has not been registered, by `cylc run /path/to/flow.cylc` ?  We can't do that anymore; but the `cylc register --run-dir` option provides the important part of that old functionality via a symlink in the standard location (e.g. `cylc reg --run-dir=$HOME/flows foo $HOME/flows/foo` (the run-dir is now the src-dir).

cylc/flow/suite_files.py
hjoliver: Actually checking for `flow.cylc` or `suite.rc` or `.service/` as above might be overkill.  Detecting `.service/` is probably sufficient isn't it?  I would say the presence of a service directory is really what defines a Cylc run directory, not the presence of a flow config file, which can appear in other places too.

oliver-sanders:
> cylc register --run-dir option provides the important part of that old functionality via a symlink in the standard location

This should become obsolete with the upcoming changes to symlink run dirs (https://github.com/cylc/cylc-flow/issues/3688) and the `cylc install` command.

Can remove later.

> run a suite that has not been registered, by cylc run /path/to/flow.cylc

This is what I was thinking of.

We used to have the ability to write and run a workflow in an arbitrary directory without symlinking anything into cylc-run. I think it "worked by accident" but proved to have some valid use cases for fire-and-forget suites, `--no-detach` and `make` use cases.

Ok, don't have to worry about that then, can come back to it later if there is demand.

cylc/flow/suite_files.py
MetRonnie: I see... Though checking for `flow.cylc` will catch if you tried to register a suite in a sub dir of an unregistered suite. But come to think of it, I suppose that isn't desirable, actually?

hjoliver:
> This should become obsolete with the upcoming changes to symlink run dirs (#3688) and the cylc install command.

:+1:  I've added a note to that issue. https://github.com/cylc/cylc-flow/issues/3688#issuecomment-705838724

hjoliver:
> >  run a suite that has not been registered, by cylc run /path/to/flow.cylc
> 
> This is what I was thinking of.
> 
> We used to have the ability to write and run a workflow in an arbitrary directory without symlinking anything into cylc-run. I think it "worked by accident" but proved to have some valid use cases for fire-and-forget suites, --no-detach and make-like use cases.

That's right. But when I went to restore that capability (and make it official) we ended up with the `--run-dir` solution for reasons detailed on the PR.  You can still use that for the same make-like use cases, it just requires the one extra registration step first.
 
> Ok, don't have to worry about that then, can come back to it later if there is demand.

Yep.

cylc/flow/suite_files.py
MetRonnie: I've pushed a new commit removes the check for `flow.cylc` / `suite.rc`

oliver-sanders:
Had a go at testing this, for the expected use case `cylc register` blocked the registration for the nested suite due to the presence of the `.service` directory.

```console
$ tree -a
.
└── a
    ├── b
    │   └── suite.rc
    └── suite.rc

2 directories, 2 files
$ cylc reg a a
2020-10-13T17:39:28+01:00 WARNING - The filename "suite.rc" is deprecated in
	favor of "flow.cylc". Symlink created.
REGISTERED a -> /Users/oliver/cylc-run/a
$ cylc reg a/b a/b
SuiteServiceFileError: Nested run directories not allowed - cannot register suite name "a/b" as a is already a valid run directory.
```

However, you can still register nested suites by doing it the other way around:

```console
$ tree -a
.
└── a
    ├── b
    │   └── suite.rc
    └── suite.rc

2 directories, 2 files
$ cylc reg a/b a/b
2020-10-13T17:38:41+01:00 WARNING - The filename "suite.rc" is deprecated in
	favor of "flow.cylc". Symlink created.
REGISTERED a/b -> /Users/oliver/cylc-run/a/b
$ cylc reg a a
2020-10-13T17:38:44+01:00 WARNING - The filename "suite.rc" is deprecated in
	favor of "flow.cylc". Symlink created.
REGISTERED a -> /Users/oliver/cylc-run/a
$ tree -a
.
└── a
    ├── .service
    │   └── source -> ..
    ├── b
    │   ├── .service
    │   │   └── source -> ..
    │   ├── flow.cylc -> /Users/oliver/cylc-run/a/b/suite.rc
    │   └── suite.rc
    ├── flow.cylc -> /Users/oliver/cylc-run/a/suite.rc
    └── suite.rc

6 directories, 4 files
```

This isn't a facetious example as with hierarchically registered suites it's quite likely that someone might try to absentmindedly register a suite somewhere in the hierarchy where it's not meant to be.

Blocking registration in this case would involve searching down the file tree in search of `.service` directories which is a bit of a pain, however, we can depth-limit this search to make it a little less painful.

If only we had a nice convenient database of registrations 😈 (Cylc 6 joke).

hjoliver:
> If only we had a nice convenient database of registrations :smiling_imp: (Cylc 6 joke).

A classic :rofl:

oliver-sanders:
The default depth that `cylc scan` uses is defined in the scan method. Could move this to a constant somewhere and use it for both purposes for consistency.

https://github.com/cylc/cylc-flow/blob/fe83115b561929861ed7b1da1d916fdeace94f6d/cylc/flow/network/scan.py#L106

oliver-sanders:
Nice.

Screenshot 2020-10-15 at 13 43 46

oliver-sanders (approved):

hjoliver (approved):
:tada: nice

PR #3854, opened 2020-10-05
Add Python 3.9 for unit tests in CI (experimental, do not stop other builds if it fails)

kinow:

This is how GH actions handles experimental [Java](https://github.com/apache/commons-imaging/blob/752c546f711a6f27ab71ef7e720de1a9b8826a4d/.github/workflows/maven.yml) versions. I guess it should work with [Python](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-preventing-a-specific-failing-matrix-job-from-failing-a-workflow-run) versions too…

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/.github/workflows/bash.yml
+++ b/.github/workflows/bash.yml
@@ -19,8 +19,8 @@ jobs:
 
       - name: Build Docker container
         run: |
-          docker build -t bash:test dockerfiles/bash/
-          docker run --name bash -v $(pwd -P):/root/cylc-flow --rm -t -d bash:test /bin/bash
+          docker build -t bash:test  -m 1G --memory-swap 1G dockerfiles/bash/
+          docker run --name bash -v $(pwd -P):/root/cylc-flow --rm -t -d -m 1G --memory-swap 1G bash:test /bin/bash
           docker ps -a
 
       - name: Install Cylc dependencies in the container
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,10 +6,15 @@ on: [pull_request]
 jobs:
   fast-tests:
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
     timeout-minutes: 15
     strategy:
       matrix:
         python-version: ['3.7', '3.8']
+        experimental: [false]
+        include:
+          - python-version: 3.9
+            experimental: true
     steps:
       - name: Checkout
         uses: actions/checkout@v2


kinow:
(Should fail for now as I think GH hasn't added support to 3.9 yet)

kinow:
The bash container is failing to build due to… virtual memory I think?

```
2020-10-05T22:17:29.2744656Z    write (2, "virtual memory exhausted\n", 25);
2020-10-05T22:17:29.2744959Z    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2020-10-05T22:17:29.4147064Z [0mgcc -c -g -O2 -Wno-parentheses -Wno-format-security -DHAVE_CONFIG_H  -I. -I../.. -I../.. -I../../lib -I. tparam.c
2020-10-05T22:17:29.4414545Z [91mtparam.c: In function ‘memory_out’:
2020-10-05T22:17:29.4421362Z [0m[91mtparam.c:67:3: warning: implicit declaration of function ‘write’ [-Wimplicit-function-declaration]
2020-10-05T22:17:29.4422068Z    write (2, "virtual memory exhausted\n", 25);
```

kinow:
Ha! https://docs.docker.com/engine/reference/run/#runtime-constraints-on-resources

Interesting, I never had to limit memory utilization of containers. Today-I-Learned :tada:

.github/workflows/bash.yml
```diff
@@ -19,8 +19,8 @@ jobs:
 
       - name: Build Docker container
         run: |
-          docker build -t bash:test dockerfiles/bash/
-          docker run --name bash -v $(pwd -P):/root/cylc-flow --rm -t -d bash:test /bin/bash
+          docker build -t bash:test  -m 1G --memory-swap 1G dockerfiles/bash/
+          docker run --name bash -v $(pwd -P):/root/cylc-flow --rm -t -d -m 1G --memory-swap 1G bash:test /bin/bash
```
kinow, line 23: Was getting some errors due to low virtual memory when building the container. Not sure if others have seen this, but might be good to constraint resources so our builds don't fail sporadically.

kinow:
(low priority, happy to leave this one waiting until GH has 3.9 support if others prefer :+1: )

.github/workflows/test.yml
```diff
@@ -6,10 +6,15 @@ on: [pull_request]
...
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
     timeout-minutes: 15
     strategy:
       matrix:
         python-version: ['3.7', '3.8']
+        experimental: [false]
+        include:
+          - python-version: 3.9
+            experimental: true
```
hjoliver, line 17: Why are `python-version` and `experimental` lists above, but scalar values here?

kinow: Oh, hadn't noticed it. Both appear to work fine. But let me edit that commit to match the other values :+1:

kinow: Done, just need to wait for CI to run and confirm it works too.

hjoliver: (I see the GH Actions docs does it the way your originally did it, so maybe it works either way :crossed_fingers: )

hjoliver: Bugger:
```
Error: .github/workflows/test.yml (Line: 25, Col: 27): A sequence was not expected
Error: The template is not valid. .github/workflows/test.yml (Line: 25, Col: 27): A sequence was not expected
```

kinow: But good to know! I guess if you have multiple versions, then you will have to do something like 

```
- python-version: 1
  experimental: true
- python-version: 2
  experimental: true
...
```

hjoliver: What does the "experimental" mean exactly?

kinow: It's just a feature in GitHub actions that allows to mark a branch of the matrix-tree as not-ready I guess? It's really common in Java and JavaScript, for new versions of the JVM or of Node (which are experimental in relation to the code; meaning that they may not work, and the developers/CI are "experimenting" with it? :smile: )

hjoliver: OK, fair enough.  Just metadata then, no real consequences :+1:

oliver-sanders (approved):

hjoliver (approved):
The 3.9 block fails, as expected.  Any opinions on whether or not we should merge like this or wait till GH-A adds 3.9 support? (It might look a bit odd to have expected test fails all the time?)

kinow:
> The 3.9 block fails, as expected. Any opinions on whether or not we should merge like this or wait till GH-A adds 3.9 support? (It might look a bit odd to have expected test fails all the time?)

We can review/approve and merge later.

They have some pull requests in GH actions virtual environments repo for Ubuntu 18.04 and 20.04, but the next updates don't seem to include Python 3.9. I think first it needs to make its way to Ubuntu package repositories, then they will create a PR to merge?

https://github.com/actions/virtual-environments/pulls

hjoliver:
I'll label this as "blocked" for now, pending 3.9 on GH-A.

kinow:
Ignore last commits. I added one commit to test something I saw in GH actions [docs for setup-python](https://github.com/actions/setup-python):

```
python-version: '3.9.0-alpha - 3.9.0' # SemVer's version range syntax
```

Didn't work :grimacing:

kinow:
I think the 3.9 will pass now, https://github.com/actions/setup-python/issues/148

.github/workflows/test.yml
oliver-sanders: Can remove the experimental thing now 3.9 is out.

kinow:
CI passed, two approvals. Merging :tada:

PR #3856, opened 2020-10-06
Fail the query with a helpful message if the variables defined do not match the expected values

### LINKED ISSUE #3855: [graphql] Error running query: argument of type 'NoneType' is not iterable
**Describe the bug**

Trying a new query to retrieve the offline data, used in the Cylc UI (for the offline UI, https://github.com/cylc/cylc-ui/pull/505).

```
{
  "errors": [
    {
      "message": "argument of type 'NoneType' is not iterable",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "workflows"
      ]
    }
  ],
  "data": {
    "workflows": null
  }
}
```

`master` for UIS (01988bb66307e53a1bda66e9ead70c64e0cda9c9) and Flow (bd293959edcf6db96414d608ef43772299514e36)

Run the query in GraphiQL, with a workflow on hold.

**Expected behavior**

Valid data retrieved, or an error response with the explanation.

**Screenshots**

**Additional context**

Query

```
query ($workflowID: ID) {
  workflows (ids: [$workflowID]) {
    ...WorkflowData
    cyclePoints: familyProxies (ids: ["root"], ghosts: true) {
      ...CyclePointData
    }
    taskProxies (sort: { keys: ["name"], reverse: false }, ghosts: true) {
      ...TaskProxyData
      jobs(sort: { keys: ["submit_num"], reverse:true }) {
        ...JobData
      }
    }
    familyProxies (exids: ["root"], sort: { keys: ["name"] }, ghosts: true) {
      ...FamilyProxyData
    }
  }
}

fragment WorkflowData on Workflow {
  id
  name
  status
  owner
  host
  port
}

fragment CyclePointData on FamilyProxy {
  id
  cyclePoint
}

fragment FamilyProxyData on FamilyProxy {
  id
  name
  state
  cyclePoint
  firstParent {
    id
    name
    cyclePoint
    state
  }
}

fragment TaskProxyData on TaskProxy {
  id
  name
  state
  isHeld
  cyclePoint
  latestMessage
  firstParent {
    id
    name
    cyclePoint
    state
  }
  task {
    meanElapsedTime
    name
  }
}

fragment JobData on Job {
  id
  firstParent: taskProxy {
    id
  }
  batchSysName
  batchSysJobId
  host
  startedTime
  submittedTime
  finishedTime
  state
  submitNum
}
```

**Pull requests welcome!**

kinow:
These changes close #3855 

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,6 +48,10 @@ compatibility, the `cylc run` command will automatically symlink an existing
 
 ### Enhancements
 
+[#3856](https://github.com/cylc/cylc-flow/pull/3856) - fail the GraphQL query
+with a helpful message if the variables defined do not match the expected
+values.
+
 [#3853](https://github.com/cylc/cylc-flow/pull/3853) - Update protobuf and
 pyzmq.
 
--- a/cylc/flow/network/graphql.py
+++ b/cylc/flow/network/graphql.py
@@ -151,27 +151,30 @@ class AstDocArguments:
             if isinstance(defn, ast.OperationDefinition):
                 root_type = get_operation_root_type(schema, defn)
                 definition_variables = defn.variable_definitions or []
-                variables = {}
                 if definition_variables:
-                    variables = get_variable_values(
-                        schema,
-                        definition_variables,
-                        variable_values
-                    )
+                    def_var_names = {
+                        v.variable.name.value
+                        for v in definition_variables
+                    }
+                    var_names_diff = def_var_names.difference({
+                        k
+                        for k in variable_values
+                        if k in def_var_names
+                    })
                     # check if we are missing some of the definition variables
-                    if not variables or \
-                            len(variables) != len(definition_variables):
-                        variable_names = [
-                            v.variable.name.value for v in definition_variables
-                        ]
+                    if var_names_diff:
                         msg = (f'Please check your query variables. The '
-                               f'following variables are expected: '
-                               f'[{", ".join(variable_names)}]')
+                               f'following variables are missing: '
+                               f'[{", ".join(var_names_diff)}]')
                         raise ValueError(msg)
                 self.operation_defs[getattr(defn.name, 'value', root_type)] = {
                     'definition': defn,
                     'parent_type': root_type,
-                    'variables': variables,
+                    'variables': get_variable_values(
+                        schema,
+                        definition_variables,
+                        variable_values
+                    ),
                 }
             elif isinstance(defn, ast.FragmentDefinition):
                 self.fragment_defs[defn.name.value] = defn
--- a/cylc/flow/network/graphql.py
+++ b/cylc/flow/network/graphql.py
@@ -150,14 +150,28 @@ class AstDocArguments:
         for defn in document_ast.definitions:
             if isinstance(defn, ast.OperationDefinition):
                 root_type = get_operation_root_type(schema, defn)
+                definition_variables = defn.variable_definitions or []
+                variables = {}
+                if definition_variables:
+                    variables = get_variable_values(
+                        schema,
+                        definition_variables,
+                        variable_values
+                    )
+                    # check if we are missing some of the definition variables
+                    if not variables or \
+                            len(variables) != len(definition_variables):
+                        variable_names = [
+                            v.variable.name.value for v in definition_variables
+                        ]
+                        msg = (f'Please check your query variables. The '
+                               f'following variables are expected: '
+                               f'[{", ".join(variable_names)}]')
+                        raise ValueError(msg)
                 self.operation_defs[getattr(defn.name, 'value', root_type)] = {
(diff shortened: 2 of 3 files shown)


kinow:
Now if I accidentally misspell a variable in Cylc UI, the message should hint that there's something wrong with the variables.

image

And fixing the variables then the query runs just fine.

image

kinow:
Virtual memory error when running the bash tests in the Docker container. Fix available in #3854 

Will spend some more time on this later trying to add a test. Not sure if a unit test will work, might try our new test classes for the first time :+1:

kinow:
The other PR that contains the fix for the Docker builds may take a while to be merged, while we wait for GH Actions to include Python 3.9.

So I've cherry-picked the commit that should fix Docker, and pushed here :+1:

kinow:
Functional test broadcast/00-simple is failing, but I think it needs to be updated after this fix (just need to understand how it works, and confirm it indeed just needs to be updated)

cylc/flow/scripts/cylc_broadcast.py
```diff
@@ -262,7 +262,7 @@ def get_option_parser():
         "-e", "--expire", metavar="CYCLE_POINT",
         help="Cancel any broadcasts that target cycle "
              "points earlier than, but not inclusive of, CYCLE_POINT.",
-        action="store", default=None, dest="expire")
+        action="store", default='', dest="expire")
```
kinow, line 265: @dwsutherland I had a functional test (`broadcast/00-simple.t`) failing as `bCutoff` (which is created using this `expire` value) was set to `None`.

In `cylc-flow`, in `network/graphql.py`, the `get_variable_values` function removes variables that are set to `None`. Which results in a GraphQL without all the variables defined.

If you run a query in JavaScript, or in GraphiQL and the number of variables in the query definition does not match the number of variables provided in the client query, it results in an error.

Apparently it was working for this functional test, but I think it's safe to set the default here to an empty string? The tests pass fine. WDYT?

kinow:
I fixed the functional test in this PR, and yesterday it passed in the Docker image. Strangely, it failed now, but my only updates were changelog and a new **unit** test. :thinking: ready for review, I think the test is just a bit brittle in Docker (will investigate why later)

hjoliver (approved):
LGTM, but I'll defer to @dwsutherland for the definitive review.

dwsutherland (changes requested):
Good idea @kinow , may I suggest the following instead:
```
        for defn in document_ast.definitions:
            if isinstance(defn, ast.OperationDefinition):
                root_type = get_operation_root_type(schema, defn)
                definition_variables = defn.variable_definitions or []
                if definition_variables:
                    def_var_names = {
                        v.variable.name.value
                        for v in definition_variables
                    }
                    var_names_diff = def_var_names.difference({
                        k
                        for k in variable_values
                        if k in def_var_names
                    })
                    # check if we are missing some of the definition variables
                    if var_names_diff:
                        msg = (f'Please check your query variables. The '
                               f'following variables are missing: '
                               f'[{", ".join(var_names_diff)}]')
                        raise ValueError(msg)
                self.operation_defs[getattr(defn.name, 'value', root_type)] = {
                    'definition': defn,
                    'parent_type': root_type,
                    'variables': get_variable_values(
                        schema,
                        definition_variables,
                        variable_values
                    ),
                }
            elif isinstance(defn, ast.FragmentDefinition):
                self.fragment_defs[defn.name.value] = defn
```

Then you don't have to change the default behaviour of GraphQL (i.e. if `None` is desirable to fall to the schema defined default), and hence not have to change the CLI tests.

kinow:
> Good idea @kinow , may I suggest the following instead:
> 
> Then you don't have to change the default behaviour of GraphQL (i.e. if `None` is desirable to fall to the schema defined default), and hence not have to change the CLI tests.

Hi @dwsutherland ! Thanks for the suggestion! I will just need some time tomorrow to re-contextualize with the change, and understand your change. Then it should be a matter of just updating the PR with a new/edit commit, and check if tests pass :+1: 

Might be able to complete it before the executive roadshow starts tomorrow in Auckland, by the morning.

kinow:
Fixed git conflicts, and applied @dwsutherland 's patch (thanks!). I think we now just need to wait for CI to verify.

cylc/flow/scripts/cylc_broadcast.py
dwsutherland: Probably want to change this back now?

kinow: I've dropped the commit with that change, and push-force'd. Now it's up to GitHub actions to confirm everything's working :+1:

dwsutherland (approved):
Good to go 👍

PR #3423, opened 2019-10-24
re-implement the task retry state using xtriggers

oliver-sanders:
* Remove the following task states:
  * `TASK_STATUS_RETRYING`
  * `TASK_STATUS_SUBMIT_RETRYING`
* Re-implement retry logic using xtriggers.
  * Create an xtirigger called `cylc[_submission]_retry_<task>.<cycle>`
    * Use the `cylc` namespace for special triggers.
  * Create only one xtrigger per task per retry type.
* Reset "retrying" tasks to waiting.

Caveats:
* Can no longer select tasks by retrying states on the CLI (e.g. `<cyclepoint>/*:retrying`)
* Retrying tasks are now waiting so have a different position in `TASK_STATUS_ORDERED`
  * TODO need to check this doesn't break polling logic

These changes partially address https://github.com/cylc/cylc-admin/pull/47

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Appropriate tests are included (unit and/or functional).
- [x] Appropriate change log entry included.
- [x] I have opened a documentation Issue at cylc/cylc-doc#162.

PR #3858, opened 2020-10-07
Fix minor issues found while reviewing task states PR's

kinow:
This is a small change with no associated Issue. Found one strange thing some code that was not in the PR I was reviewing (#3423) so ran a complete code analysis with the IDE, and selected the issues that seemed to be legit (it actually shows heaps more warnings, especially related to typing - for some other day).

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/network/scan.py
+++ b/cylc/flow/network/scan.py
@@ -190,7 +190,7 @@ async def filter_name(flow, pattern):
     Args:
         flow (dict):
             Flow information dictionary, provided by scan through the pipe.
-        pattern (str):
+        pattern (re.Pattern):
             One or more regex patterns as strings.
             This will return True if any of the patterns match.
 
--- a/cylc/flow/parsec/validate.py
+++ b/cylc/flow/parsec/validate.py
@@ -909,7 +909,6 @@ class CylcConfigValidator(ParsecValidator):
         value = cls.strip_and_unquote(keys, value)
         if not value:
             raise IllegalValueError("xtrigger", keys, value)
-        fname = None
         args = []
         kwargs = {}
         match = cls._REC_TRIG_FUNC.match(value)
--- a/cylc/flow/remote.py
+++ b/cylc/flow/remote.py
@@ -181,7 +181,6 @@ def construct_rsync_over_ssh_cmd(
         src_path(string): source path
         dst_path(string): path of target
         platform(dict)): contains info relating to platform
-        logfile(str): the path to the file logging the rsync
         rsync_includes(list): files and directories to be included in the rsync
 
     """
--- a/cylc/flow/remote.py
+++ b/cylc/flow/remote.py
@@ -222,7 +222,7 @@ def construct_rsync_over_ssh_cmd(
 
 
 def construct_ssh_cmd(
-        raw_cmd, user=None, host=None, forward_x11=False, stdin=False,
+        raw_cmd, host=None, forward_x11=False, stdin=False,
         ssh_cmd=None, ssh_login_shell=None, ssh_cylc=None, set_UTC=False,
         allow_flag_opts=False, timeout=None
 ):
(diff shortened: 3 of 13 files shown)


kinow:
There are two warnings that I think could be bugs, but I will need to debug with some test suite. If any of them turns out to be a bug, I'll add tests & changelog. Draft PR for now.

cylc/flow/parsec/validate.py
```diff
@@ -909,7 +909,6 @@ def coerce_xtrigger(cls, value, keys):
         value = cls.strip_and_unquote(keys, value)
         if not value:
             raise IllegalValueError("xtrigger", keys, value)
-        fname = None
```
kinow, line 912: (click the blue down-arrow in GH UI to view `fname` declared further down)

cylc/flow/network/scan.py
```diff
@@ -190,7 +190,7 @@ async def filter_name(flow, pattern):
     Args:
         flow (dict):
             Flow information dictionary, provided by scan through the pipe.
-        pattern (str):
+        pattern (re.Pattern):
```
kinow, line 193: See further down where we call `.match`, method from `re.Pattern` I think.

cylc/flow/remote.py
```diff
@@ -223,15 +222,14 @@ def construct_rsync_over_ssh_cmd(
 
 
 def construct_ssh_cmd(
-        raw_cmd, user=None, host=None, forward_x11=False, stdin=False,
```
kinow, line 226: The `user` appears to be calculated in runtime, with no need to pass a `user` parameter. This `user` parameter is never used within `construct_ssh_cmd`, which can be misleading to a user/caller of the API.

cylc/flow/scheduler.py
```diff
@@ -626,7 +625,7 @@ async def start_scheduler(self):
         finally:
             self.profiler.stop()
 
-    async def run(self, daemonize=False):
```
kinow, line 629: Probably removed some time ago. We still have the `daemonize` function elsewhere. This `daemonize` flag is never used here. Also doesn't seem to be used by the callers of this function.

cylc/flow/suite_files.py
```diff
@@ -389,7 +389,7 @@ def get_suite_srv_dir(reg, suite_owner=None):
     return os.path.join(run_d, SuiteFiles.Service.DIRNAME)
 
 
-def load_contact_file(reg, owner=None, host=None):
```
kinow, line 392: Not used.

cylc/flow/task_state.py
```diff
@@ -485,7 +485,6 @@ def _add_prerequisites(self, point, tdef):
 
         if tdef.sequential:
             # Add a previous-instance succeeded prerequisite.
-            p_prev = None
```
kinow, line 488: See below for more, `p_prev` is re-declared only in the `if adjusted` branch.

cylc/flow/remote.py
hjoliver: I think this was associated with the "task owner" concept we used to have (run a task job as user@host) but recently removed (now use ssh config instead if you need to run as a different user on host).

kinow:
Last issue seems to be a bug but I'm still testing it. Will leave that one for later in a separate PR.

cylc/flow/xtriggers/suite_state.py
```diff
@@ -58,9 +58,6 @@ def suite_state(suite, task, point, offset=None, status='succeeded',
                different location to what is specified in the global
                configuration (usually ``~/cylc-run``).
 
-        debug (bool):
-            Flag to enable debug information.
-
```
hjoliver, line 63: Can you remove `(except ''debug")` from the bottom of the docstring too? (line 71)

hjoliver (changes requested):
(One tiny additional change suggested)

cylc/flow/xtriggers/suite_state.py
kinow: Done!

hjoliver:
One review will do for this @kinow - thanks :bouquet:

PR #3863, opened 2020-10-13
Store task prerequisites in their own DB table

### LINKED ISSUE #3752: Better DB task pool prerequisite storage
Initial SoD implementation uses a nasty hack to store task proxy prerequisites in the DB (because I'm "database-challenged"):

*`cylc.flow.task_pool`*
```python
 415       # Update prerequisite satisfaction status from DB
 416             sat = {}
 417             for k, v in json.loads(satisfied).items():
 418                 sat[tuple(json.loads(k))] = v
 419       # TODO (from Oliver's PR review):
 420       #   Wait, what, the keys to a JSON dictionary are themselves JSON
 421       #     :vomiting_face:!
 422       #   This should be converted to its own DB table pre-8.0.0.
 423             for pre in itask.state.prerequisites:
 424                 for k, v in pre.satisfied.items():
 425                     pre.satisfied[k] = sat[k]
```

This works fine by is fundamentally :vomiting_face:

MetRonnie:
These changes close #3752 

Instead of storing task prerequisites as a JSON string inside the `task_pool` table, remove the `satisfied` column from `task_pool` and introduce a `task_prerequisites` table.

**Note:** This PR removes the ability to restart a Cylc 7 suite.

For the flow (where `foo.1` fails)
```
P1 = foo & ipsum:custom1 => bar
```
the `task_prerequisites` table has the form:
| cycle* TEXT | name* TEXT | prereq_name* TEXT | prereq_cycle* TEXT | prereq_output* TEXT | satisfied TEXT      |
| ----------- | ---------- | ----------------- | ------------------ | ------------------- | ------------------- |
| 1           | bar        | foo               | 1                  | succeeded           | 0 |
| 1           | bar        | ipsum             | 1                  | I'm a custom task message  | satisfied naturally | 

where asterisk in the column header denotes primary key.

A task's prerequisites are inserted into the `task_prerequisites` table when the task is added to the runahead pool (if a task has no prerequisites, no entry is inserted to the `task_prerequisites` table). The `task_prerequisites` table is updated at the same time as before when the prerequisites used to be updated in the `task_pool` table.

**Requirements check-list**

- [X] Appropriate tests are included (functional).

- [x] Change log entry included

- [X] No dependency changes.


cylc/flow/rundb.py
```diff
@@ -301,7 +309,6 @@ class CylcSuiteDAO:
             ["name", {"is_primary_key": True}],
             ["flow_label", {"is_primary_key": True}],
             ["status"],
-            ["satisfied"],
```
MetRonnie, line 304: Not sure what are the implications of removing the `satisfied` state from the task pool checkpoints table. Does there need to be a `task_prerequisites_checkpoints` table?

cylc/flow/unicode_rules.py
```diff
@@ -105,6 +105,24 @@ def disallowed_characters(*chars):
...
+        True
+        >>> bool(regex.match("I'm Batmans"))
+        False
+    """
+    return (
+        re.compile(fr'^{re.escape(string)}$'),
+        f'cannot be the string ``{string}``'
+    )
+
+
```
oliver-sanders, line 125: Did this sneak in from another branch?

oliver-sanders:
Looks good to me, note to others, this change does not have an upgrader for the database. We have abandoned the idea of maintaining restart compatibility with Cylc7 due to installation changes amongst other things so this is fine. Just be aware that `cylc restart` will not work with previously run workflows on this branch.

cylc/flow/unicode_rules.py
MetRonnie: I thought I needed this to further constrain custom task output messages. Then, as soon as I finished writing this function I realised I had got the `satisfied` and `output` columns from the `task_prerequisites` table mixed up in my head, so this wasn't needed. Left it in in case it ends up being useful in future?

oliver-sanders:
Sure, but if removing, remove all upgraders and their associated tests.

cylc/flow/rundb.py
oliver-sanders: > Does there need to be a task_prerequisites_checkpoints table?

In order to preserve checkpointing functionality we would need a checkpoints table subject to the same checkpointing logic as the task pool table.

I have to say I'm not sure I really understand the purpose of checkpointing, especially post Cylc6/SoD so I've put up an issue, partly to bring the number of open issues back up to the golden 256, but mostly to debate the relevance of checkpointing to Cylc8.

My suggestion is, don't add checkpointing in this PR. Await the resolution of that issue, if we need to keep checkpointing we can bring it back in a later PR.

MetRonnie:
Still need to detect when trying to restart a Cylc 7 suite and give a useful error message

cylc/flow/rundb.py
hjoliver: (I've commented on checkpoint in the new issue).

MetRonnie:
Would it be worth renaming the `db` file in order to detect when the user is trying to restart a Cylc < 8.0a3 suite?

oliver-sanders:
> Would it be worth renaming the db file in order to detect when the user is trying to restart a Cylc < 8.0a3 suite?

The Cylc version of the current run (even if the suite is stopped it can still be thought of as the current "run") has been stored in the database since Cylc 7.something:

```console
$ sqlite3 <path-to-suite>/.service/db 'select value from suite_params where key = "cylc_version";'
```

MetRonnie:
Is there any reason not to rename the db file, though? It would make the logic simpler

oliver-sanders:
Yes, people may be using it.

tests/functional/restart/54-incompatible-db.t
```diff
@@ -14,32 +14,36 @@
...
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#-------------------------------------------------------------------------------
-# Compat: Cylc<8
-# Test the upgrade of the old *retrying states to the new xtrigger based
-# retry mechanism.
+#------------------------------------------------------------------------------
+# Test that restarting a Cylc 7 suite does not work due to database
+# incompatibility, and that suitable error message is given
+
```
MetRonnie, line 20: Can this be done with an integration test instead?

oliver-sanders: Kinda, but not so easy, this is fine.

cylc/flow/suite_db_mgr.py
```diff
@@ -578,8 +580,31 @@ def recover_pub_from_pri(self):
...
         pri_dao.vacuum()
         pri_dao.close()
+
+    def check_suite_db_compatibility(self):
+        """Raises SuiteServiceFileError if the existing suite database is
+        incompatible with the current version of Cylc."""
+        pri_dao = self.get_pri_dao()
+        last_run_ver = pri_dao.connect().execute(
+            f'SELECT value FROM {self.TABLE_SUITE_PARAMS} '
+            f'WHERE key == "{self.KEY_CYLC_VERSION}"').fetchone()[0]
```
oliver-sanders, line 599: Need to wrap this in a try/except to protect against the failure of this query (the suite_params table is fairly new).

oliver-sanders (commented):
Looks good, works ok.

One thing left to do, the task_pool table is housekept so as to keep it in sync with the task pool. The same logic will need to apply to the task_prereqs table otherwise it will continue to grow indefinitely.

Had a quick fish about in the code and it looks like the `suite_db_mgr` [wipes and rebuilds](https://github.com/cylc/cylc-flow/blob/master/cylc/flow/suite_db_mgr.py#L406) the task_pool and it's related tables for each batch of changes 🤮. The rundb performs all queued insertions and deletions [in the same transition](https://github.com/cylc/cylc-flow/blob/11d45a8f7c699f7a61522ee5386b7c359b49939c/cylc/flow/rundb.py#L411-L429) so that's probably not as bad as it sounds (depending on the implementation of the database), presumably just means sqlite has to do a bit more legwork to compute what's actually changed?

Probs best to turn a blind eye to this and just fit the `task_prerequsites` table into the existing global-update pattern. Perhaps in a later Cylc8 version as we try to make the main loop more async we will make DB updates more event-driven and remove this per-loop global-update logic in the process.

MetRonnie:
> Probs best to turn a blind eye to this and just fit the `task_prerequsites` table into the existing global-update pattern. 

Like this: https://github.com/cylc/cylc-flow/pull/3863/commits/01e0798004f7a324ddafaa1e9aa1737e58052cfc ?

MetRonnie:
@oliver-sanders Looks like that will need a `task_prerequisites_checkpoints` table, the absence of which seems to be causing `tests/k/restart/19-checkpoint.t` to fail now

oliver-sanders:
Ah, perhaps disable that test for now and add a comment to the checkpoints issue.

cylc/flow/prerequisite.py
```diff
@@ -156,7 +165,7 @@ def is_satisfied(self):
             return self._all_satisfied
         else:
             # No cached value.
-            if not self.satisfied:
+            if self.satisfied == {}:
```
hjoliver, line 168: The former way is more Pythonic, innit?

cylc/flow/rundb.py
hjoliver: (We've agreed to removed named checkpoints - well, as good as, no objections so far - so this PR can forget about them).

hjoliver: I'll add a comment to the checkpoint issue...

cylc/flow/unicode_rules.py
hjoliver: This will get identified as dead code once we get code coverage back.  Better to hide it in a comment if it's worth keeping?

hjoliver:
> @oliver-sanders Looks like that will need a `task_prerequisites_checkpoints` table, the absence of which seems to be causing `tests/k/restart/19-checkpoint.t` to fail now

https://github.com/cylc/cylc-flow/issues/3864#issuecomment-712535988

cylc/flow/prerequisite.py
MetRonnie: I thought it was misleading at a casual glance
```python
def is_satisfied(self):
    if not self.satisfied:
        return True
```

hjoliver: OK, I see your point, I didn't look at the surrounding context. Although the comment on the next line explains what's going on, at least. Probably indicates a badly named dict, if anything (my fault).

cylc/flow/unicode_rules.py
oliver-sanders: We try to avoid [introns](https://en.wikipedia.org/wiki/Intron), generally better to have the bare minimum of code in the repo, easy to revive from a branch if needed in the future.

oliver-sanders:
To checkpoint or not to checkpoint?!

hjoliver:
BLOCKED by #3864

oliver-sanders:
Can we just disable these tests for now and add a note to #3864 for now, then reinstate checkpoint functionality later if required?

hjoliver:
> Can we just disable these tests for now and add a note to #3864 for now, then reinstate checkpoint functionality later if required?

We could, but is there any reason we can't make a quick decision on removing named checkpointing.  All the facts are "on the table" already.

MetRonnie:
If named checkpoints are removed does that mean automatic checkpoints are also removed?

oliver-sanders:
> We could, but is there any reason we can't make a quick decision on removing named checkpointing

Perhaps but it's a bit off topic for this PR to go removing all of the named checkpointing logic.

> If named checkpoints are removed does that mean automatic checkpoints are also removed?

Any logic relating to the `checkpoint` tables or `cylc *checkpoint` commands would be up for removal.

hjoliver:
> > We could, but is there any reason we can't make a quick decision on removing named checkpointing
> 
> Perhaps but it's a bit off topic for this PR to go removing all of the named checkpointing logic.

Arguable, but fair enough - my feeling was, this PR breaks checkpointing as-is, and let's wait on the checkpointing decision to know if this PR should be amended or not.

Moot point now, as @dpmatthews has agreed on removal on Element chat - I'll remove the blocked label...

oliver-sanders (approved):
:+1:

hjoliver (approved):
Great, thanks @MetRonnie :tada:

PR #3868, opened 2020-10-14
Intel host select.get platform group

wxtim:
These changes are a pre-requisite to addressing #3827 
To carry out intelligent host selection the code needs to have access not to platforms but to platform groups.

To this end this PR
- Replaces `platforms.get_platform` with `platforms.get_platform_group` and  `platforms.get_platform_from_name` to `get_platforms_from_name`, so that code calling it has information about the whole group of platforms. If regex only matches a single platform rather than a platform group then a dummy platform group dictionary containing that single platform is returned. Question: Should this have selection method set to "first" to save time?
- Adds function `platforms.get_localhost_platform` - in many places in the code `get_platform()` is called. This function replaces that one, and makes it easier to differentiate cases where we just want info about the localhost platform with cases where something more complex is happening.
- Adds a function `platforms.get_host_from_platform_group`. As with `platforms.get_host_from_platform` this is a placeholder for more sophisticated logic. At the moment it simply picks a random platform from the platform_group and calls `get_host_from_platform`.

### DIFF
--- a/cylc/flow/cfgspec/suite.py
+++ b/cylc/flow/cfgspec/suite.py
@@ -26,7 +26,7 @@ from cylc.flow.parsec.OrderedDict import OrderedDictWithDefaults
 from cylc.flow.parsec.upgrade import upgrader
 from cylc.flow.parsec.validate import (
     DurationFloat, CylcConfigValidator as VDR, cylc_config_validate)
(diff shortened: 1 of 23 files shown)


wxtim:
Closed in favour of cracking on with IHS having reached this milestone point.

PR #3875, opened 2020-10-15
validate: run validate checks on run

oliver-sanders:
Investigation work relating to - https://github.com/cylc/cylc-admin/pull/111

Currently `rose suite-run` calls `cylc validate; cylc run` which is highly inefficient. Since validation does most of the things run does the current plan is to unify the two so that all validation occurs on run.

After this PR the only two validation-only checks are:

* Circular dependency checking.
   * Which can be quite slow, we are still deciding how/when this should be enabled/disabled.
* Throwing errors for naked tasks.
   * Which is controversial and awaiting decision.

This PR also does the following:

* Centralise event handler data provision
* Enumerate event handler data fields
* Rename naked *dummy* tasks => naked tasks

**Requirements check-list**
- [x] No dependency changes.


cylc/flow/config.py
```diff
@@ -538,8 +542,6 @@ def __init__(
 
         self.mem_log("config.py: before load_graph()")
         self.load_graph()
-        if self._is_validate():
-            GraphNodeParser.get_inst().clear()
```
oliver-sanders, line 542: Looking at the code nothing is done to `GraphNodeParser` until this line so wiping the instance will have no effect.

cylc/flow/config.py
```diff
@@ -551,21 +553,21 @@ def __init__(
 
         self.configure_suite_state_polling_tasks()
 
-        # Warn or abort (if --strict) if naked dummy tasks (no runtime
+        self._check_task_event_handlers()
+        self._check_special_tasks()  # adds to self.naked_tasks
```
oliver-sanders, line 557: This is the old `check_tasks` method split into two parts and moved up slightly earlier in the logic to allow unification of naked task checking.

cylc/flow/config.py
```diff
@@ -1521,55 +1523,37 @@ def run_mode(self, *reqmodes):
         else:
             return mode
 
-    def check_tasks(self):
-        """Call after all tasks are defined.
-
-        ONLY IF VALIDATING THE SUITE
-        because checking conditional triggers below may be slow for
-        huge suites (several thousand tasks).
```
oliver-sanders, line 1529: I ran this against a diamond configuration with 100'000 tasks, this method took 0.5% of the wall clock time so I think we're good here.

cylc/flow/taskdef.py
```diff
@@ -126,15 +126,6 @@ def describe(self):
...
 
-    def check_for_explicit_cycling(self):
-        """Check for explicitly somewhere.
-
-        Must be called after all graph sequences added.
-        """
-        if len(self.sequences) == 0 and self.used_in_offset_trigger:
-            raise TaskDefError(
-                "No cycling sequences defined for %s" % self.name)
-
```
oliver-sanders, line 137: I've removed this check as I don't think it's helpful. It is common to define unused tasks, especially in large programatic suites. Another use case is to enable dynamic insertion for diagnostic purposes or whatever.

cylc/flow/config.py
```diff
@@ -1521,55 +1523,37 @@ def run_mode(self, *reqmodes):
...
-            taskdef.check_for_explicit_cycling()
-            # Check custom event handler templates compat with task meta
             if taskdef.rtconfig['events']:
-                subs = dict((key, key) for key in self.TASK_EVENT_TMPL_KEYS)
-                for key, value in self.cfg['meta'].items():
-                    subs['suite_' + key.lower()] = value
-                subs.update(taskdef.rtconfig['meta'])
-                # Back compat.
-                try:
-                    subs['task_url'] = subs['URL']
```
oliver-sanders, line 1551: This is the back-compat for `task_url`, however the back-compat for `suite_url` seems to be missing from here, though present in `task_event_mgr`?

Centralised this logic to remove the problem.

cylc/flow/task_events_mgr.py
```diff
@@ -108,6 +109,44 @@ def log_task_job_activity(ctx, suite, point, name, submit_num=None):
         LOG.debug(ctx_str)
 
 
+class EventData(Enum):
```
oliver-sanders, line 112: Enumeration of the event handler data fields. Not strictly necessary but humour me, I like enumerations. The plan was to auto-document from this enumeration in the config for further safety.

Sadly I can't get it to format nicely in the docs (it renders as `Event='event'` whereas we want `%(event)s`). Would be more work to fix this than we have time to do at the moment so I've dropped it for now.

tests/functional/validate/37-clock-trigger-task-not-defined.t
```diff
@@ -31,6 +31,8 @@ cat >'flow.cylc' <<'__FLOW_CONFIG__'
 __FLOW_CONFIG__
 run_fail "${TEST_NAME_BASE}" cylc validate --strict "${PWD}/flow.cylc"
 cmp_ok "${TEST_NAME_BASE}.stderr" <<'__ERR__'
-SuiteConfigError: clock-trigger task "foo" is not defined.
+WARNING - naked tasks detected (no entry under [runtime]):
+	+       foo
+SuiteConfigError: strict validation fails naked tasks
```
oliver-sanders, line 36: This naked task check for `scheduling[special-tasks]` entries used a different logic for reporting issues to the `[runtime]` check so I unified it.

tests/functional/events/36-task-event-bad-custom-template.t
```diff
@@ -27,10 +27,11 @@ run_fail "${TEST_NAME_BASE}-validate" cylc validate "${SUITE_NAME}"
...
 SuiteConfigError: bad task event handler template t1: echo %(rubbish)s: KeyError('rubbish')
 __ERR__
-suite_run_ok "${TEST_NAME_BASE}-run" \
+suite_run_fail "${TEST_NAME_BASE}-run" \
     cylc run --reference-test --debug --no-detach "${SUITE_NAME}"
-LOG="${SUITE_RUN_DIR}/log/suite/log"
-run_ok "${TEST_NAME_BASE}-log" grep -q -F "ERROR - 1/t1/01 ('event-handler-00', 'succeeded') bad template: 'rubbish'" "${LOG}"
+grep_ok \
+    'SuiteConfigError: bad task event handler template t1: echo %(rubbish)s: KeyError(.rubbish.)' \
+    "${SUITE_RUN_DIR}/log/suite/log"
```
oliver-sanders, line 34: This suite now fails on startup due to a malformed event template.

oliver-sanders:
`tests/k/restart/21-task-elapsed.t` appears to have become super-flaky on GHA, I couldn't get it to pass on multiple runs so I added a "temp" debug action for just that test. Now I can't get it to fail 😠.

hjoliver:
Looks good, but what do you mean by "investigation work" - do you want this PR reviewed now or is it more of an investigative draft?

oliver-sanders:
I mean that I put this branch together whilst going through the code to determine exactly what the difference between the validate use of cylc.flow.config.SuiteConfig object was to other uses.

Yep good to merge once I've striped out that temp commit...

oliver-sanders:
Done.

oliver-sanders:
Now a different flakytest is being flaky (`tests/k/cylc-poll/16-execution-time-limit.t`). This is quite worrying as the tests are automatically retried.

hjoliver:
> Since validation does most of the things run does the current plan is to unify the two so that all validation occurs on run.
>
> After this PR the only two validation-only checks are:..

What about repeatedly validating a flow during the early stages of development, to catch errors early, long before you might first attempt to run the new flow?

In light of that use case, wouldn't it make more sense to have run and validate do *exactly the same validation checks* (loaded from the same code modules, of course) so that you only have to use one or the other?  i.e. standalone validation, or run start-up  validation, but no need for `cylc validate <flow-name>; cylc run <flow-name>`.

oliver-sanders:
> In light of that use case, wouldn't it make more sense to have run and validate do exactly the same validation 

Yes, that's what this PR does!

(with the exception of `--strict` mode which can change once decisions are made).

hjoliver:
> Yes, that's what this PR does!

Ah good.  Skimming your description I read *After this PR the only two validation checks are:..*  (not *validation-ONLY*) :facepalm:

.github/workflows/test.yml
```diff
@@ -108,3 +108,10 @@ jobs:
               -exec echo '==== {} ====' \; -exec cat '{}' \;
           find "${TMPDIR:-/tmp}/${USER}/cylctb-"* -type f \
               -exec echo '==== {} ====' \; -exec cat '{}' \;
+
+      - name: Upload
+        if: failure()
+        uses: actions/upload-artifact@v2
+        with:
+          name: cylc-run
+          path: cylc-run
```
oliver-sanders, line 117: Added this to help with debugging flaky tests. It should upload the cylc-run directory as an [artefact](https://docs.github.com/en/free-pro-team@latest/actions/guides/storing-workflow-data-as-artifacts) that we can download and inspect locally. Artefacts are housekept along with build logs after 90 days.

hjoliver: Nice, good idea.

cylc/flow/config.py
MetRonnie: Are you sure? As far as I can tell, there is this chain of calls:

1. `self.load_graph()` in the line above
2. In `load_graph()`: https://github.com/cylc/cylc-flow/blob/f656c97e5ac81144635d2bcb8bfb69b48d4f75ae/cylc/flow/config.py#L2215-L2216
3. In `_proc_triggers()`: https://github.com/cylc/cylc-flow/blob/f656c97e5ac81144635d2bcb8bfb69b48d4f75ae/cylc/flow/config.py#L2232
4. In `generate_task_defs()`: https://github.com/cylc/cylc-flow/blob/f656c97e5ac81144635d2bcb8bfb69b48d4f75ae/cylc/flow/config.py#L1638

.github/workflows/test.yml
MetRonnie: If you give the `name: Test` step an `id: `, we can check the failure of that step specifically instead of the installation steps. Not really necessary though as most of the time it's the test step that fails.

MetRonnie (approved):

cylc/flow/config.py
hjoliver: Yeah, GraphNodeParser caches already-parsed nodes. On master, this `clear()` results in **more** parsing during a validation, due to reparsing nodes that were already parsed.  So I'm not sure what that line is there for???

cylc/flow/task_events_mgr.py
```diff
@@ -108,6 +109,44 @@ def log_task_job_activity(ctx, suite, point, name, submit_num=None):
...
+    Message = 'message'
+    BatchSysName = 'batch_sys_name'
+    BatchSysJobID = 'batch_sys_job_id'
+    SubmitTime = 'submit_time'
+    StartTime = 'start_time'
+    FinishTime = 'finish_time'
+    PlatformName = 'platform_name'
+    TaskName = 'name'
+    TaskURL = 'task_url'  # deprecated
+    SuiteURL = 'suite_url'  # deprecated
```
hjoliver, line 131: BTW Python docs say:
> Because Enums are used to represent constants we recommend using UPPER_CASE names for enum members, and will be using that style in our examples.

oliver-sanders: Yes, this was a purposful violation that slides through flake8 nicely, can change if it's unpopular.

Python enums are really strange in that what you get out is not what you get in:

```python
>>> from enum import Enum
>>> class Foo(Enum):
...     ANSWER = 42
... 
>>> Foo.ANSWER
<Foo.ANSWER: 42>
# wait that's not an integer

>>> Foo.ANSWER.value
42
# hmmm
```

hjoliver (approved):
Conflict resolution just requires moving the new upload block from the old `tests.yml` to the new `tests_functional.yml`; I'll merge from master, resolve that, and push the result so we can get this in.

PR #3876, opened 2020-10-15
add copyright notices to files where it was missing.

wxtim:
(seriously, people might think there's nothing else going on because of some sort of disease...)

Just say if you think I should remove the change from the trivially small files, otherwise don't wish to waste any more time on this.


oliver-sanders (approved):
Dammit, a few of them are mine.

kinow (approved):

Issue #3592, opened 2020-05-04
Universal suite/task/cycle delimiter (universal identifier)


hjoliver:
specification: cylc/cylc-admin#115

For suite/task IDs, the new hierarchical ID chosen by @dwsutherland for use in the back-end has nice properties but can't be used on the command line because the delimiter is the shell pipe character. 

Supersedes #3542 - please refer to the discussion there.  The general solution to this problem should support the original use case of that issue (allow dots in task names).

dwsutherland:
As mentioned in the superseded [ticket](https://github.com/cylc/cylc-flow/issues/3542#issuecomment-611787467); We may need to use a two character delimiter to meet 
 the constraints of shell/Cylc-Conf/datetime ...

oliver-sanders:
Proposal:

* Use `/` as the separator (as originally intended) but use `//` to separate the flow from the cycle.
* Support `:<arg>` as a state selector (e.g. the current `cylc trigger flow *:failed`).

```bash
# regular form (e.g. element ids in GraphQL schema)
[~user[/]]flow//cycle/task/job

# expanded form with selectors (e.g. CLI querying)
[~user[/]]flow[:state]//cycle[:state]/task[:state]/job[:state]
```

Examples (equivalent GraphQL arguments under current scheme)

```bash
# flow under account user (wid="user|flow")
~user/flow

# the task foo at all cycles in flow (wid="flow|*|foo")
flow//*/foo

# all running tasks in foo (wid="flow|*|*" state="running")
flow//*/*:running

# all running workflows (wid="*" state="running")
*:running
```

cylc-admin proposal PR: https://github.com/cylc/cylc-admin/pull/115

oliver-sanders:
Assigning myself for phase 1 i.e. permitting the new syntax on the CLI.

PR #3880, opened 2020-10-19
schema: separate task and job mutations

oliver-sanders:
Separate task and job mutations in the schema.

* This means you can now selectively kill/poll specific jobs by explicitly providing the submit number.
* However, for regular use there is no interface change.
* This allows the UI to perform mutations on jobs (e.g. right click on running job icon, select kill) (see https://github.com/cylc/cylc-ui/pull/504).

At present there isn't much value to being able to specify a particular job submit number since it would be considered an error for two jobs to run at the same time.

Here's an example specifying the job submit num on the CLI:

```console
$ # cylc kill <flow> '<namespace>.<point>#<job>'
$ cylc kill abc 'foo.1#1'
```

And here via GraphQL:

```graphql
mutation () {
    kill ("abc", "foo.1#1") {
        result
    }
}
```

Obviously `#` isn't ideal for use on the CLI, we could consider something else, see also #3592.

**Requirements check-list**

- [ ] No dependency changes.

### DIFF
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -1263,7 +1263,39 @@ class NamespaceName(String):
 class NamespaceIDGlob(String):
     """A glob search for an active task or family.
 
-    Can use the wildcard character (`*`), e.g `foo*` might match `foot`.
+    Examples:
+
+    * `foo.*` the task named `foo` in any cycle.
+    * `*.1` all tasks in cycle `1`.
+    * `*.*:failed` all failed tasks.
+
+    Can use either of these formats:
+
+    * `namespace[.cycle][:status]`
+    * `point[/namespace][:status]`
+
+    Where both `namespace` and `cycle` can be globs.
+
+    """
+
+
+class NamespaceJobGlob(String):
+    """A glob search for jobs within namespaces and cycles.
+
+    Examples:
+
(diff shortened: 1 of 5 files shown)


cylc/flow/network/schema.py
```diff
@@ -1263,7 +1263,39 @@ class NamespaceName(String):
...
+
+    * `namespace[.cycle][:status]`
+    * `point[/namespace][:status]`
+
+    Where both `namespace` and `cycle` can be globs.
+
+    """
+
+
+class NamespaceJobGlob(String):
```
oliver-sanders, line 1282: This input type enables us to tell that this mutation works on jobs by flicking through the results of an introspection query.

cylc/flow/task_id.py
```diff
@@ -22,6 +22,58 @@
...
+        (?P<namespace>{SEP_PATTERN})
+        (?:\:(?P<state>{SEP_PATTERN}))?
+        (?:\#(?P<job>{SEP_PATTERN}))?
+        $
+    ''',
+    re.X
+)
+
+
+def parse_reference(string):
```
oliver-sanders, line 52: This replaces `TaskPool._parse_task_item`.

oliver-sanders:
Closing this now, will bring back in a new form with the universal identifier.

PR #3882, opened 2020-10-20
Fix polled task:start bug

### LINKED ISSUE #3769: polled start event bug
If using polling to get remote task status, the `started` event will be missed if the job finishes before the first poll is done.

Prior to spawn-on-demand (e.g. 8.0a2):
- the started event is missed
- but `:start` triggers work
   - presumably we artificially update the `started` output as well when the task polls as `succeeded`

After spawn-on-demand (master):
- the started event is missed
- `:start` triggers don't work (triggering is now event-driven)

Test case:
```
[scheduling]
    [[graph]]
        R1 = "foo:start => bar"
[runtime]
    [[foo]]
        platform = maui 
        # script = "sleep 20"  # slow task will poll as started, then succeeded - GOOD
        script = "true"  # fast task will only poll as succeeded - BAD
        [[[events]]]
             started handler = echo "!!!!! STARTED"
```
With the following global config to configure  polling:
```
# ~/.cylc/flow/8.0a3.dev/flow.rc 
[platforms]
    [[maui]]
         hosts = w-nwp01.maui
         communication method = poll
         submission polling intervals = PT5S
         execution polling intervals = PT5S
```
If you run the test case,
- on master: the event  handler will not execute, and the scheduler will shut down as finished without spawning `bar`
- at 8.0a2 (with pre-platforms global and workflow config) the event handler will not execute, but `bar` will trigger

MetRonnie:
These changes close #3769

**Requirements check-list**

- [X] Appropriate tests are included (functional).

- [X] No changelog entry needed

- [X] No dependency changes.

### DIFF
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -365,13 +365,14 @@ class TaskEventsManager():
 
         # Check the `started` event has not been missed e.g. due to
         # polling delay
-        if (message not in [TASK_OUTPUT_SUBMITTED, self.EVENT_SUBMIT_FAILED,
-                            TASK_OUTPUT_STARTED]
+        if (message not in [self.EVENT_SUBMITTED, self.EVENT_SUBMIT_FAILED,
+                            self.EVENT_STARTED]
                 and not itask.state.outputs.is_completed(TASK_OUTPUT_STARTED)):
-            self.setup_event_handlers(itask, 'started', 'job started')
+            self.setup_event_handlers(
+                itask, self.EVENT_STARTED, f'job {self.EVENT_STARTED}')
             self.spawn_func(itask, TASK_OUTPUT_STARTED)
 
-        if message == TASK_OUTPUT_STARTED:
+        if message == self.EVENT_STARTED:
             if (
                     flag == self.FLAG_RECEIVED
                     and itask.state.is_gt(TASK_STATUS_RUNNING)
@@ -379,10 +380,10 @@ class TaskEventsManager():
                 return True
             self._process_message_started(itask, event_time)
             self.spawn_func(itask, TASK_OUTPUT_STARTED)
-        elif message == TASK_OUTPUT_SUCCEEDED:
+        elif message == self.EVENT_SUCCEEDED:
             self._process_message_succeeded(itask, event_time)
             self.spawn_func(itask, TASK_OUTPUT_SUCCEEDED)
-        elif message == TASK_OUTPUT_FAILED:
+        elif message == self.EVENT_FAILED:
             if (
                     flag == self.FLAG_RECEIVED
                     and itask.state.is_gt(TASK_STATUS_FAILED)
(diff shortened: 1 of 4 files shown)


MetRonnie:
Does this actually need a changelog entry, since the bug didn't really exist before Cylc 8.0a3?

hjoliver:
> Does this actually need a changelog entry, since the bug didn't really exist before Cylc 8.0a3?

No.

hjoliver (changes requested):
Looks good, but:
- I think it needs to apply to any message beyond "started", not just succeed and failed - anything that implies the task job is running in spite of its started output not being completed.
- you haven't also triggered the started event handler (as noted in the Issue)

kinow (commented):
The code looks good to me. Great tests too!

But I tried to reproduce the issue first on `master`, with the star trek workflow from this PR. The workflow ran with no errors on `master` and on this PR branch.

I also searched the strings from the reference log in the output, and they are present in both `master` and this branch PR?

I will try to use the example from @hjoliver's issue.

hjoliver:
> But I tried to reproduce the issue first on master, with the star trek workflow from this PR. The workflow ran with no errors on master and on this PR branch.

@kinow it requires a proper remote host configured  by global config for polling task comms.  If you try to run it with a local job, the poll completes too quickly, so you won't see the problem.

I did reproduce it locally using another local user account as a fake remote host (via ssh config) - because that still makes the poll go by ssh, which slows it down enough.

MetRonnie:
> I think it needs to apply to any message beyond "started", not just succeed and failed - anything that implies the task job is running in spite of its started output not being completed.

I thought succeeded and failed were the only two greater than started/running? https://github.com/cylc/cylc-flow/blob/d89afd0131b09b4c3ea9fa83dd62c8d94b6d850f/cylc/flow/task_state.py#L73-L84

The only other place I can think might be this? https://github.com/cylc/cylc-flow/blob/d89afd0131b09b4c3ea9fa83dd62c8d94b6d850f/cylc/flow/task_events_mgr.py#L445-L447

MetRonnie:
> you haven't also triggered the started event handler (as noted in the Issue)

Should be addressed now

hjoliver:
> The only other place I can think might be this? 

That's right, not all incoming messages (via TCP comms or polled) correspond to a state change. Specifically, any custom output messages would imply the task is running even if a previous started message had not been received.

kinow:
I wanted to second-review this one, but looks like the easiest way to reproduce the issue is with a remote platform. While this could be made with Docker (maybe @oliver-sanders 's PR already provides it?) it would take me some time to prepare the environment. Will lurk in here to learn how it was fixed later :+1:

hjoliver (changes requested):
Sorry to be pedantic @MetRonnie ! but strictly speaking this should cover *all* post-start messages, just in case, including "aborted with message", and even the final catch-all "unhandled messages" - because a job can't send any message at all unless it is running.

How about a separate block nearer the top of the method that completes the started output and triggers the started event handler, if the message is anything at all except for submitted or submit-failed? (Then carry on to process the specific message).
   if message is not 'submitted' or 'submit-failed', and task is not started, then

MetRonnie:
> Sorry to be pedantic 

Not at all! Just my lack of exposure to certain areas of Cylc

datamel (approved):
Ronnie, I have read through your code and run the functional tests - the broken tests on this branch are the cylc-cat-log ones, I rebased this branch with master and they pass. 
I have used Hilary's test case and run this on master to observe the issue and again with your branch checked out and can see the bug has been fixed. Thanks Ronnie!

hjoliver (approved):
What Mel said; all good now, thanks :+1:

PR #3883, opened 2020-10-21
Config change.stop after cycle point

### LINKED ISSUE #3694: conf: [scheduling]stop after cycle point
Implement a `[scheduling]stop after cycle point` configuration to match the `cylc run --stop-cycle-point=POINT` cli option.

See the [config change proposal document](https://cylc.github.io/cylc-admin/proposal-config-changes.html#schedulingstop-after-cycle-point---new).

**Pull requests welcome!**

wxtim:
These changes close #3694 : Implement a [scheduling]stop after cycle point configuration to match the cylc run --stop-cycle-point=POINT cli option.

**Requirements check-list**

- [x] Appropriate Auto-doc string added for the command reference.
- [ ] Documentation PR.

- [x] No dependency changes.


cylc/flow/scheduler.py
```diff
@@ -482,8 +482,25 @@ async def configure(self):
...
+                self.options.stopcp is None
+            ):
+                self.pool.set_stop_point(get_point(
+                    self.config.cfg['scheduling']['stop after cycle point']
+                ))
+                self.options.stopcp = self.config.cfg['scheduling'][
+                    'stop after cycle point'
+                ]
+            if self.options.stopcp:
+                self.pool.set_stop_point(get_point(self.options.stopcp))
```
oliver-sanders, line 503: Can this be simplified along the lines of:

```python
if not self.options.stopcp:
    # doesn't matter if stop after cycle point is None
    self.options.stopcp = self.config.cfg['scheduling']['stop after cycle point']
if self.options.stopcp:
    self.pool.set_stop_point(get_point(self.options.stopcp))
```

?

wxtim: Not really: Your suggestion would never allow the command line to over-ride the config: I'm not against tidying or commenting the logic though. It's a bit wierd.

Also, not having `'stop after cycle point' in self.config.cfg['scheduling']` would mean putting this in a `try:/except:` in case the config item was blank(I mean, I could set a default, but that seems weird in this instance).

wxtim: Sources of stopcp by priority
- Command line
- Database
- Config

We can simplify the logic by creating a temporary variable to store our stopcp in, and only overwriting the self.options.stopcp at the end of the process.

oliver-sanders: > Not really: Your suggestion would never allow the command line to over-ride the config

I think it would, the config value is only read if the CLI opt is *not* set.

cylc/flow/scheduler.py
```diff
@@ -1878,3 +1870,32 @@ def _get_events_conf(self, key, default=None):
         """Return a named [cylc][[events]] configuration."""
         return self.suite_event_handler.get_events_conf(
             self.config, key, default)
+
+    def process_cylc_stop_point(self):
```
MetRonnie, line 1874: I would have thought a unit test for this method would be better suited than a functional test

MetRonnie (changes requested):
Looks good apart from question of unit vs functional test (up to you). However still needs a change log entry and maybe a cylc-doc PR?

cylc/flow/scheduler.py
wxtim: I considered it, but because it's a method of the class it'd require some in un-fun mocking. However it'd be nicer, and I like the idea of working out how.

wxtim:
> Looks good apart from question of unit vs functional test (up to you). However still needs a change log entry and maybe a cylc-doc PR?

I'm not sure about a cylc-doc PR. I think that if I were to write anything there it would probably go in `src/tutorial/scheduling/further_scheduling`, but I feel I'd want to do a more in-depth treatment including a discussion of use cases rather than just copying the content of the config reference, which will be auto generated from the changes in this PR>

oliver-sanders:
Don't think this needs adding to the user guide or anything. The best documentation would likely be to note the CLI opt referencing this config (and vice versa).

MetRonnie (approved):

cylc/flow/cfgspec/suite.py
```diff
@@ -383,6 +383,18 @@
             shut down once all tasks have passed this point. If this item
             is provided you can override it on the command line.
         ''')
+        Conf('stop after cycle point', VDR.V_CYCLE_POINT, desc='''
+            Set stop point. Shut down after all tasks have PASSED
+            this cycle point. Will be over-ridden by command line
+            ``--stop-cycle-point=POINT``
```
oliver-sanders, line 389: Could you add a quick note onto the `--stop-cycle-point` option to say that it overrides the `stop after cycle point` configuration in the workflow.

There's no markdown/RST or referencing syntax in the CLI, yet, but once we've sorted out the CLI we can autodoc that too.

tests/functional/restart/08-stop-after-cycle-point/flow.cylc
```diff
@@ -0,0 +1,26 @@
...
+        PT1H = hello[-PT1H] => hello
+
+[runtime]
+    [[hello]]
+        script = """
+        if [[ "$CYLC_TASK_CYCLE_POINT" == 19700101T0000Z ]]; then
+            cylc stop "$CYLC_SUITE_NAME"
+        fi
+
+        """
```
oliver-sanders, line 26: https://stackoverflow.com/a/44704969

oliver-sanders (approved):

tests/functional/restart/08-stop-after-cycle-point/flow.cylc
wxtim: Gah! That again. Now so changed. :(

PR #3884, opened 2020-10-21
Symlink dirs

datamel:
This PR is part of the rose suite run migration.
New section added to the global config - [symlink dirs]. Configuration as documented in https://github.com/cylc/cylc-admin/blob/master/docs/proposal-platforms.md
Configurable directories are: run, share, share/cycle, work, log.

Symlinks are now created locally on `cylc register` and `cylc run`. 
Symlinks are created remotely per install target during remote initialisation. 

These changes close https://github.com/cylc/cylc-flow/issues/3688

- [x] Contains logically grouped changes - I will tidy/squash after any review changes.
- [x] (master branch) I will open a documentation PR at cylc/cylc-doc. Issue opened and self-assigned.
- [x] No dependency changes.


cylc/flow/pathutil.py
```diff
@@ -139,3 +139,60 @@ def make_suite_run_tree(suite):
...
+            dirs_to_symlink[dir_] = os.path.join(link, 'cylc-run', suite, dir_)
+            continue
+
+    return dirs_to_symlink
+
+
+def make_symlink(src, dst):
+    """Makes symlinks for directories.
+    args: src - path to where the files are to be stored
+          dst - path to where the link, pointing to src, will be created.
```
oliver-sanders, line 184:
```suggestion

    Args:
        src (str): path to where the files are to be stored
        dst (str): path to where the link, pointing to src, will be created.
```

cylc/flow/scripts/cylc_remote_init.py
```diff
@@ -14,7 +14,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-"""cylc [task] remote-init [--indirect-comm=ssh] UUID RUND
+"""cylc [task] remote-init [--indirect-comm=ssh] INSTALL TARGET RUND
```
oliver-sanders, line 17:
```suggestion
"""cylc [task] remote-init [--indirect-comm=ssh] INSTALL_TARGET RUND
```

cylc/flow/scripts/cylc_remote_init.py
```diff
@@ -55,8 +56,13 @@ def main():
         dest="indirect_comm",
         default=None,
     )
-    options, (install_target, rund) = parser.parse_args()
-    remote_init(install_target, rund, indirect_comm=options.indirect_comm)
+    options, (install_target, rund, *
+              dirs_to_be_symlinked) = parser.parse_args()
```
oliver-sanders, line 60: I didn't know we could have arbitrary whitespace between the `*` and the variable, I do now!

tests/functional/remote/03-symlink-dirs.t
```diff
@@ -0,0 +1,114 @@
...
+TEST_SYM="${TEST_NAME_BASE}-share-cycle-symlink-exists-ok"
+if [[ $(readlink "$HOME/cylc-run/${SUITE_NAME}/share/cycle") == \
+"$TMPDIR/$USER/cylctb_tmp_share_dir/cylc-run/${SUITE_NAME}/share/cycle" ]]; then
+ok "$TEST_SYM.localhost"
+else
+    fail "$TEST_SYM.localhost"
+fi
+
+for DIR in 'work' 'share' 'log'; do
+   TEST_SYM="${TEST_NAME_BASE}-${DIR}-symlink-exists-ok"
```
oliver-sanders, line 70: Me being fussy but the indentation in this file really bemuses me 3, 4, 5 & 6 space indentation all in the same file, my CDO hurts.

tests/functional/remote/03-symlink-dirs.t
```diff
@@ -0,0 +1,114 @@
...
+
+
+LINK="$(${SSH} "${CYLC_TEST_HOST}" 'readlink "$HOME/cylc-run/'"$SUITE_NAME"/share/cycle'"')"
+if [[ "$LINK" == "$TMPDIR/$USER/ctb_tmp_share_dir/cylc-run/${SUITE_NAME}/share/cycle" ]]; then
+    ok "${TEST_NAME_BASE}-share/cycle-symlink-exists-ok.remotehost"
+else
+    fail "${TEST_NAME_BASE}-share/cycle-symlink-exists-ok.remotehost"
+fi
+
+# shellcheck disable=SC2016
```
oliver-sanders, line 100: Is this one needed? Perhaps it is, but on line 103?

cylc/flow/task_remote_mgr.py
```diff
@@ -177,14 +177,14 @@ def remote_init(self, platform, curve_auth,
                 If waiting for remote init command to complete
 
         """
-        self.install_target = platform['install target']
+        install_target = platform['install target']
```
oliver-sanders, line 180: 🙈

cylc/flow/task_remote_mgr.py
```diff
@@ -336,6 +339,8 @@ def _remote_init_callback(
                         LOG.error(
                             'File installation error on '
                             f'{install_target}:\n {err}')
+                    else:
+                        LOG.info("File installation complete.")
```
oliver-sanders, line 343: Can combine this with the `if out` branch to avoid duplicate log messages.

oliver-sanders (commented):
Looks good, few small comments.

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -498,14 +498,61 @@
...
+            """)
+            Conf('share', VDR.V_STRING, None, desc="""
+                Specifies the directory where share directories are created.
+                If specified the workflow share directory will be created in
+                <share dir>/cylc-run/<workflow-name>/share and a symbolic link
+                will be created from <$HOME/cylc-run/<workflow-name>/share. If
+                not specified the workflow share directory will be created in
+                $HOME/cylc-run/<workflow-name>/share.
+            """)
+            Conf('share/cycle', VDR.V_STRING, None, desc="""
```
datamel, line 538: @oliver-sanders I'm getting namespace problems with this share/cycle when building the docs. Is there a way around this one or does it need changing to something like share-cycle?

oliver-sanders: Ha, yes I suppose you would get a problem with that as `/` will severely jigger up URLs breaking the docs. Thanks for checking the docs!

Two options:

1) Rename as you suggest.
2) Do something fancy to cylc-sphinx-extensions to convert the `/` to something safe, but unique.

I think (2) might be a bit tricky.

How about trying `share>cycle`, `>` is also a nasty character for the web but it might get handled better. IDK.

datamel: So, I have tried `share>cycle`, as you suspected, same namespace error.
Happy to have a go at 2. or go with `share-cycle`. I guess the problem with `share-cycle` is `cycle` isn't obviously a sub-directory of `share`. Hmm, do you think it is worth the time investment in 2?

oliver-sanders: Can try at (2), if you can't fix it quickly-ish but it's probably not worth trying.

Would need to replace `/` with something benign, there are two likely places where this logic could go:

* https://github.com/cylc/cylc-sphinx-extensions/blob/85583e178c31b8d279e6df3553d0a9c4b7d93d60/cylc/sphinx_ext/cylc_lang/domains.py#L470-L493
* https://github.com/cylc/cylc-sphinx-extensions/blob/85583e178c31b8d279e6df3553d0a9c4b7d93d60/cylc/sphinx_ext/cylc_lang/domains.py#L341-L347

If you fancy a go try one or both of these.

I think since the cylc domain was written there is now some minimal Sphinx documentation for "custom domains" which might provide a pointer or two.

tests/functional/remote/03-symlink-dirs.t
datamel: I think you are referring to the shellcheck disable? I have tried removing them and it complains if any of them are removed - I guess SSHing resets, unless there is a nifty way of globally setting it for the whole test.

tests/functional/remote/03-symlink-dirs.t
datamel: Yes, that is not ideal - should be changed now!

oliver-sanders (commented):
Looks good

Tried with this configuration:

```ini
[symlink dirs]
    [[<remote-install-target>]]
        run = $HOME/cylc-ran
        log = $HOME/cylc-log
        share = $HOME/cylc-share
        share/cycle = $HOME/cylc-share-cycle
        work = $HOME/cylc-work
```

And got this result:

```
$ tree cylc-*
cylc-log
└── cylc-run
    └── one
        └── log
            └── job
                └── 1
                    └── foo
                        ├── 01
                        │   ├── job
                        │   ├── job.err
                        │   ├── job.out
                        │   └── job.status
                        └── NN -> 01
cylc-ran
└── cylc-run
    └── one
        ├── log -> /home/oliver/cylc-log/cylc-run/one/log
        ├── share -> /home/oliver/cylc-share/cylc-run/one/share
        └── work -> /home/oliver/cylc-work/cylc-run/one/work
cylc-run
└── one -> /home/oliver/cylc-ran/cylc-run/one
cylc-share
└── cylc-run
    └── one
        └── share
            └── cycle -> /home/oliver/cylc-share-cycle/cylc-run/one/share/cycle
cylc-share-cycle
└── cylc-run
    └── one
        └── share
            └── cycle
cylc-work
└── cylc-run
    └── one
        └── work
            └── 1
```

Nice.

Spotted a couple of minor issues.

1) I misspelt `$HOME` as $HOMe` and got cryptic failures.
2) I think this might be messing with the on-shutdown tidy up logic, see `cylc run --no-detach` traceback below:

```
Exception ignored in: <function BaseEventLoop.__del__ at 0x7fcbb7d89e60>
  File "/Users/oliver/miniconda3/envs/cylc8/lib/python3.7/asyncio/base_events.py", line 628, in __del__
    self.close()
  File "/Users/oliver/miniconda3/envs/cylc8/lib/python3.7/asyncio/unix_events.py", line 55, in close
    super().close()
  File "/Users/oliver/miniconda3/envs/cylc8/lib/python3.7/asyncio/selector_events.py", line 91, in close
    self._close_self_pipe()
  File "/Users/oliver/miniconda3/envs/cylc8/lib/python3.7/asyncio/selector_events.py", line 98, in _close_self_pipe
    self._remove_reader(self._ssock.fileno())
  File "/Users/oliver/miniconda3/envs/cylc8/lib/python3.7/asyncio/selector_events.py", line 271, in _remove_reader
    key = self._selector.get_key(fd)
  File "/Users/oliver/miniconda3/envs/cylc8/lib/python3.7/selectors.py", line 190, in get_key
    return mapping[fileobj]
  File "/Users/oliver/miniconda3/envs/cylc8/lib/python3.7/selectors.py", line 71, in __getitem__
    fd = self._selector._fileobj_lookup(fileobj)
  File "/Users/oliver/miniconda3/envs/cylc8/lib/python3.7/selectors.py", line 225, in _fileobj_lookup
    return _fileobj_to_fd(fileobj)
  File "/Users/oliver/miniconda3/envs/cylc8/lib/python3.7/selectors.py", line 42, in _fileobj_to_fd
    raise ValueError("Invalid file descriptor: {}".format(fd))
ValueError: Invalid file descriptor: -1
```

cylc/flow/pathutil.py
```diff
@@ -139,3 +139,61 @@ def make_suite_run_tree(suite):
...
+    dirs_to_symlink = get_dirs_to_symlink('localhost', suite)
+    rund = get_suite_run_dir(suite)
+    for key, value in dirs_to_symlink.items():
+        if key == 'run':
+            dst = rund
+            src = (os.path.expandvars(value))
+        else:
+            dst = os.path.join(rund, key)
+            src = os.path.expandvars(value)
+        make_symlink(src, dst)
```
MetRonnie, line 155: Niggle: unnecessary parentheses around the first `src = (...`?

cylc/flow/pathutil.py
```diff
@@ -139,3 +139,61 @@ def make_suite_run_tree(suite):
...
+        return dirs_to_symlink
+    base_dir = symlink_conf[install_target]['run']
+    if base_dir is not None:
+        dirs_to_symlink['run'] = os.path.join(base_dir, 'cylc-run', suite)
+    for dir_ in ['log', 'share', 'share/cycle', 'work']:
+        link = symlink_conf[install_target][dir_]
+        if link is None:
+            continue
+        elif link == base_dir:
+            continue
```
MetRonnie, line 173:
```suggestion
        if link is None or link == base_dir:
            continue
```

cylc/flow/pathutil.py
```diff
@@ -139,3 +139,61 @@ def make_suite_run_tree(suite):
...
+        elif (base_dir is None and link is not None) or link != base_dir:
+            dirs_to_symlink[dir_] = os.path.join(link, 'cylc-run', suite, dir_)
+            continue
```
MetRonnie, line 176:
```suggestion
        if link is None or link == base_dir:
            continue
        dirs_to_symlink[dir_] = os.path.join(link, 'cylc-run', suite, dir_)
```

cylc/flow/scripts/cylc_remote_init.py
MetRonnie: Me neither, maybe it would be better to keep them together for readability?

cylc/flow/task_remote_cmd.py
```diff
@@ -84,17 +85,29 @@ def create_client_keys(srvd, install_target):
...
 
-def remote_init(install_target, rund, indirect_comm=None):
+def remote_init(install_target, rund, *dirs_to_symlink, indirect_comm=None):
     """cylc remote-init
 
     Arguments:
         install_target (str): target to be initialised
         rund (str): suite run directory
+        dirs_to_symlink (list): directories to be symlinked in form
+        [directory=symlink_location, ...]
```
MetRonnie, line 95: Either `*dirs_to_symlink` should be `dirs_to_symlink` in the args, or the docstring should not call it `(list)` (but what do you call unpacked positional args?)

cylc/flow/task_remote_cmd.py
```diff
@@ -84,17 +85,29 @@ def create_client_keys(srvd, install_target):
...
     """
     rund = os.path.expandvars(rund)
+    for item in dirs_to_symlink:
+        key, val = item.split("=", 1)
+        if key == 'run':
+            dst = rund
+            src = (os.path.expandvars(val))
+        else:
+            dst = os.path.join(rund, key)
+            src = os.path.expandvars(val)
```
MetRonnie, line 106: As before
```suggestion
        if key == 'run':
            dst = rund
        else:
            dst = os.path.join(rund, key)
        src = os.path.expandvars(val)
```

MetRonnie: Or even
```suggestion
        dst = rund if key == 'run' else os.path.join(rund, key)
        src = os.path.expandvars(val)
```

cylc/flow/task_remote_cmd.py
datamel: I've checked this with @oliver-sanders and unpacked postitional args become a list in docstrings. Does seem a bit non-intuitive though!

cylc/flow/task_remote_cmd.py
datamel: Thanks, much cleaner!

cylc/flow/task_remote_cmd.py
MetRonnie: Yuk, if you pass a list the for loop won't work 😢 Oh well

oliver-sanders:
Conflicts.

cylc/flow/scripts/cylc_remote_init.py
datamel: This was what my auto formatter, I'll change it for readability.

cylc/flow/pathutil.py
```diff
@@ -139,3 +139,66 @@ def make_suite_run_tree(suite):
...
+    for key, value in dirs_to_symlink.items():
+        if key == 'run':
+            dst = rund
+        else:
+            dst = os.path.join(rund, key)
+        src = os.path.expandvars(value)
+        if '$' in src:
+            raise WorkflowFilesError(
+                f'Unable to create symlink to {src}. '
+                'Please check configuration.')
```
oliver-sanders, line 157: Perhaps make it clear that this is due to environment variables, if this is caused by something daft like undefined `$TMPDIR` someone could end up scrabbling around for a while before they realise.

oliver-sanders (approved):
👍

cylc/flow/scripts/remote_init.py
```diff
@@ -14,7 +14,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-"""cylc remote-init [--indirect-comm=ssh] UUID RUND
+"""cylc remote-init [--indirect-comm=ssh] INSTALL_TARGET RUND
```
MetRonnie, line 17: Is `[DIRS_TO_BE_SYMLINKED ...]` meant to be here too?

MetRonnie (approved):
Looks good 👍 I just had one comment, up to you whether or not to address it

cylc/flow/scripts/remote_init.py
datamel: Yes, just redoing the docstrings for this now. I'll get them tidied before merge. Thanks!

oliver-sanders: Ooh, well spotted.

MetRonnie (approved):

oliver-sanders:
Woohoo.

PR #3886, opened 2020-10-21
Config.remove old deprecations

### LINKED ISSUE #3690: conf: remove deprecated configurations
Prior to releasing 8.0.0:

* Remove configurations deprecated in cylc-flow<=7.6.0
* Remove and deprecated global configurations.

See also the [config change proposal document](https://cylc.github.io/cylc-admin/proposal-config-changes.html#deprecation).

**Pull requests welcome!**

wxtim:
These changes probably close #3690 

- I'm happy about the items I have removed from `cylc/flow/cfgspec/suite.py`.
- I'm happy that I have removed items from `tests/functional/deprecations/01...`
- I'm more uncertain about how I've dealt with the global config deprecation items - I don't think that the config object can do without an upgrader, and I was loathe to remove too much infrastructure, because I imagine that after the clean break between Cylc7 and Cylc8 we might want to be able to carry out upgrades to `global.cylc` and `user.cylc` in future.

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -660,34 +660,6 @@ with Conf('global.cylc', desc='''
 def upg(cfg, descr):
     """Upgrader."""
     u = upgrader(cfg, descr)
-
-    u.obsolete('6.4.1', ['test battery', 'directives'])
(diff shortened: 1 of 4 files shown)


oliver-sanders (approved):

hjoliver (approved):
Conforms to the config change proposal :+1:

PR #2276, opened 2017-05-10
Supply URLs to suite and task event handlers.

hjoliver:
Close #2275 

~~Not for review yet - just needs tests.~~  Tests added (or rather, existing tests upgraded).

PR #3905, opened 2020-10-30
conf: remove explicit group meta entry

### LINKED ISSUE #2776: Deprecate the `group` setting
Now that we can register suites in a hierarchy the `group` setting seems somewhat redundant.

The `group` setting is hard-coded in the suite, the suite registration can be decided on by the user allowing them to group suites as they like.

Registration hierarchies can also have multiple levels and are generally more intuitive as they are included in the suite name.

generic

This could be done with #1873, creating as a separate issue in-case there is discussion to be had here.

oliver-sanders:
Closes #2776

(effectively a documentation change)

Remove the "explicit" `group` configuration since it is not actually supported anywhere in Cylc at present. The config is still valid as it's just metadata.

Rather than brining back explicit support for the `group` setting in the new UI we would allow any arbitrary metadata item to be used (e.g. `title`, `description`, etc). See also https://github.com/cylc/cylc-ui/issues/337.

**Requirements check-list**
- [x] No dependency changes.


MetRonnie:
> Closes 2276

I'm not sure that's the right one, #2276 is a PR

oliver-sanders:
Sorry 2776.

MetRonnie:
Just looks like a functional test needs updating

MetRonnie (approved):

PR #3906, opened 2020-10-30
Remove checkpoints

### LINKED ISSUE #3891: Remove named checkpointing
Supersedes #3864 

Long story short:
- no one uses it
- Cylc 8 reflow achieves the same result better
- it complicates the DB and code logic somewhat

So: let's get rid of it.

@MetRonnie - assigning you as  you've run into this somewhat in #3863, but someone else can take it on if need be...

MetRonnie:
These changes close #3891

- [x] Waiting for #3863 to be merged before this is ready

**Requirements check-list**

- [x] (master branch) documentation PR at https://github.com/cylc/cylc-doc/issues/170

- [x] No dependency changes.


cylc/flow/suite_db_mgr.py
```diff
@@ -576,6 +565,8 @@ def on_restart(self):
         self.check_suite_db_compatibility()
         pri_dao = self.get_pri_dao()
         pri_dao.vacuum()
+        self.n_restart = pri_dao.select_suite_params_restart_count() + 1
+        self.put_suite_params_1(self.KEY_RESTART_COUNT, self.n_restart)
         pri_dao.close()
```
MetRonnie: Setting this as a property of the suite database manager class feels a bit weird, but I couldn't rely on reading the updated value from the database from the scheduler, because this "put" operation is merely queued and doesn't get executed until after the scheduler has already logged the number of restarts

MetRonnie:
In terms of documentation, I suspect that this would be better left to the 7-to-8 guide or whoever writes up reflow in the docs?

hjoliver:
New cylc-doc Issue for 7-8 migration, with a checkbox for this topic: https://github.com/cylc/cylc-doc/issues/170

cylc/flow/rundb.py
```diff
@@ -487,78 +455,42 @@ def _execute_stmt(self, stmt, stmt_args_list):
...
-    def select_checkpoint_id(self, callback, id_key=None):
-        """Select from checkpoint_id.
+    def select_suite_params(self, callback):
+        """Select from suite_params.
 
         Invoke callback(row_idx, row) on each row, where each row contains:
-            [id, time, event]
-
-        If id_key is specified, add where id == id_key to select.
+            [key,value]
```
wxtim, line 480: Can you be more explicit about the data the key and value represent?

cylc/flow/scheduler.py
```diff
@@ -546,9 +543,10 @@ async def start_servers(self):
         self.pub_port = self.publisher.port
 
     async def log_start(self):
+        # TODO: remove or reimplement n_restart when DB has a table for the
```
wxtim, line 546: Could/should this have a ticket as well as/instead of a TODO comment? If you wish to leave the comment perhaps it should reference the ticket?

wxtim (approved):
Nothing Major to say - just my usual pickyness about docstrings.

cylc/flow/rundb.py
MetRonnie: The diff shows up weirdly; I didn't actually change this part of the docstring at all:
```
    def select_suite_params(self, callback):
        """Select from suite_params.

        Invoke callback(row_idx, row) on each row, where each row contains:
            [key,value]
```

cylc/flow/scheduler.py
MetRonnie: Oops, that should've been removed

cylc/flow/rundb.py
MetRonnie: Added an example of a row: https://github.com/cylc/cylc-flow/pull/3906/commits/6c07d9fd777fe67940f7c06aab2ac221dae6b456

oliver-sanders (approved):

Issue #3819, opened 2020-09-16
rose suite-run: plugin for rose-suite.conf functionality


oliver-sanders:
Superseedes: #2960
Related to: #1885

Write a Cylc plugin which detects the `rose-suite.conf` file and acts accordingly.

This plugin will call out to Rose2 libraries, this functionality already exists in the Rose it's just a matter of injecting it in the correct places in the Cylc logic.

### The `rose-suite.conf` File

The [`rose-suite.conf`](https://metomi.github.io/rose/doc/html/api/configuration/suite.html#rose:file:rose-suite.conf) file has the following sections:

* `opts` - Optional configuration keys.
* `root-dir` - Affects suite installation, symlinks Cylc directories.
* `[env]` - Environment variables.
* `[jinja2:suiterc]` - Jinja2 variables.
* `[empy:suiterc]` - Empy variavles.
* `[file:*]` - File installation.

#### `opts`

Rose should use this variable to determine which optional configuration files to load. This is already implemented in the existing Rose configuration logic so shouldn't require any extra work on the Cylc side.

#### `root-dir` 

Obsolete, replaced by https://github.com/cylc/cylc-flow/issues/3688, log a warning and do nothing.

**Questions:**

* ~Should we make an attempt to map this onto the new Cylc `[symlink dirs][<install target>]run` configs?~ No.

#### `[env]`

~Export these environment variables into the Cylc environment (i.e. `os.environ[key] = var`).~

> **Note:** We should *not* edit the `flow.cylc` file, if exporting the env vars is insufficient we could consider adding a command line option to `cylc (run|restart)`.

For now the path of least resistance is probably to just go ahead and meddle with the `flow.cylc` file, we can keep this logic in the Cylc part (rather than the plugin part) and improve it later, perhaps with a new Cylc options file or an environment variable table in the database (to match the template variables table).

**Questions:**

* ~How is this supposed to work, Cylc may re-execute on another host?!~
  * **Rose adds items under `[cylc][evironment]`**
* Can we achieve this without meddling with the `flow.cylc` file (really don't want to go down that route again)?

#### `[jinja2:suiterc]`

> *Note:* I think we will probably keep this as `:suite.rc` for now.

With cylc7/rose-2019 Rose prepends these variables to the top of the `suite.rc` file. It should not be necessary to edit any Cylc files to achieve this in Cylc8.

> **Note:** Look at implementing this logic in the parsec layer so that it doesn't have to be manually implemented on a command by command basis.

[...]

PR #3913, opened 2020-11-03
Rose conf.read jinja2

wxtim:
These changes  address #3819 
**Summary of Changes**
- [x] Added ability to use `cylc.configure` entry point to parse variables for Cylc in `cylc/flow/parsec/fileparse.py`.
- [x] Written a template `cylc install` which installs files listed in `rose-suite.conf` using `cylc.install` entry point.
- [x] Written functional tests (this seems pretty odd, since they will only work if the `rose-cylc` plugin is installed, but it's probably ok, for now)

I have not raised a warning for absence of `root-dir` - I think that this needs to be in the plugin.

**Requirements check-list**
  - Unit test all functionality in new python code
  - Functional test basic `rose-suite.conf` reading functionality, but cannot check the optional config reading until it's plumbed into the CLI
- [x] No documentation update required: Full documentation of all sections of the rose-suite.conf work after all done.


cylc/flow/parsec/jinja2support.py
```diff
@@ -35,6 +37,8 @@
     TemplateNotFound,
     TemplateSyntaxError)
 
+from metomi.rose.config_tree import ConfigTreeLoader
```
TomekTrzeciak, line 40: Shouldn't this be an optional dependency? Not everyone uses rose.

oliver-sanders: Yes, we will need to encapsulate this in a plugin framework pre-release.

wxtim: Do I need to deal with this in this PR?

oliver-sanders: Ok for now.

TomekTrzeciak: You could just move that import into `get_rose_vars` after you test for existence of `rose-suite.conf`.

TomekTrzeciak:
Do you plan to get this working for `[empy:suite.rc]` too (currently it doesn't)?

Also, `rose suite-run` adds a few extra variables (`ROSE_ORIG_HOST`, `ROSE_SITE`, `ROSE_VERSION` and `ROSE_SUITE_VARIABLES` metomi/rose#2338).

cylc/flow/parsec/jinja2support.py
```diff
@@ -182,9 +186,78 @@ def jinja2environment(dir_=None):
     env.globals['environ'] = os.environ
     env.globals['raise'] = raise_helper
     env.globals['assert'] = assert_helper
+
+    # Insert key/value pairs from rose-suite.conf into jinja2 env.
+    from ast import literal_eval
+    rose_vars = get_rose_vars(dir_)
+    if rose_vars is not None:
+        for key, value in rose_vars.items():
+            env.globals[key] = literal_eval(value)
```
TomekTrzeciak, line 195: Shouldn't this belong in `template_vars` rather then environment globals?

wxtim:
> Do you plan to get this working for `[empy:suite.rc]` too (currently it doesn't)?

Yes.

cylc/flow/parsec/jinja2support.py
wxtim: This would probably make more sense - I'll give it a go.

TomekTrzeciak: Why? These are suite variables, so logically serve the same purpose as Cylc template vars? Plus, that way you won't need to solve it separately for EmPy.

TomekTrzeciak: I think the easiest place to implement this centrally would be in [`cylc.flow.parsec.fileparse.read_and_proc`](https://github.com/cylc/cylc-flow/blob/218fa6d/cylc/flow/parsec/fileparse.py#L203-L269). All the relevant info is there, you can even test if you need to load from `jinja2:suite.rc` or `empy:suite.rc`.

BTW: Any plans to rename `jinja2:suite.rc` & `empy:suite.rc` into something more in line with the current Cylc naming? Given `suite.rc` -> `cylc.flow` rename it would only make sense.

oliver-sanders: We need to back support both `jinja2:suite.rc`, however, the two can be merged and called something like "template_variables", but not expecting all of these changes in the first PR.

wxtim: Turns out my problem was that we don't _really_ want to mix "CLI" template variables, which must be serializable because they are stored in the suite DB with "File" template variables, which are stored in a file and do not need to be stored in the suite DB: Unless we force a massive change on users the file template variables must be potentially unserializable, so we mustn't add them to the dictionary to be stored in the DB. Fortunately, since we store them in a file this isn't a problem.

cylc/flow/config.py
```diff
@@ -218,10 +218,12 @@ def __init__(
         self.mem_log("config.py: before RawSuiteConfig init")
         if output_fname:
             output_fname = os.path.expandvars(output_fname)
+        # Copy template_vars to prevent addition of non serializable template
+        # vars from file being added by parsec.
         self.pcfg = RawSuiteConfig(
             fpath,
             output_fname,
-            template_vars
+            copy(template_vars)
```
TomekTrzeciak, line 226: Any reason you prefer this copy here instead of in `read_and_proc`, where you modify the input `template_vars`?

wxtim: I'd rather not pass the original across at all. Is there a reason to do it the other way around?

TomekTrzeciak: Well, just a matter of style I suppose. I've always considered mutating input arguments in a function with explicit return value a bit evil.

wxtim:
I've converted this back to a draft PR having decided to steam onwards into dealing with empy, fileinstall and environment variables.

cylc/flow/config.py
wxtim: It turned out to be a mistake anyway - I've moved the copying inside the jinja2/empy processing sections so they cannot accidentally contaminate each other.

.github/workflows/test_fast.yml
```diff
@@ -25,6 +25,7 @@ jobs:
 
       - name: Install
         run: |
+          pip install git+https://github.com/metomi/rose@master
```
wxtim, line 28: TODO: Remove this I think

cylc/flow/parsec/fileparse.py
```diff
@@ -243,23 +263,42 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
+            flines = jinja2process(flines, fdir, tvars)
+
+    # Inject rose-suite.conf environment variables into the workflow env.
+    # Appended to ensure that Rose settings over-ride flow.cylc.
+    if rose_vars['env'] is not None:
+        env_lines = [
+            f'        {key} = {value}' for
+            key, value in rose_vars['env'].items()
+        ]
+        flines += ['[scheduler]']
```
wxtim, line 299: This seems a tad crude - anyone suggest a better way?

cylc/flow/scripts/install.py
```diff
@@ -0,0 +1,55 @@
...
+        __doc__, comms=True,
+        argdoc=[('REG', 'Suite name')])
+
+    return parser
+
+
+@cli_function(get_option_parser)
+def main(parser, options, suite):
+    for entry_point in pkg_resources.iter_entry_points(
+        'cylc.rose_fileinstall'
```
oliver-sanders, line 45: `cylc.install`?

cylc/flow/parsec/fileparse.py
TomekTrzeciak: append/extend the list?
```python
        flines.append('[scheduler]')
        flines.append('    [[environment]]')
        flines.extend( '        %s = %s' % keyval for keyval in rose_vars['env'].items())
```

wxtim: I meant less crude modifying the end of the flow.cylc file. But yes.

wxtim: I figured that given the INI structure that appending was safer...

cylc/flow/parsec/fileparse.py
```diff
@@ -243,23 +263,42 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
+            flines = empyprocess(flines, fdir, tvars)
 
     # process with Jinja2
     if do_jinja2:
         if flines and re.match(r'^#![jJ]inja2\s*', flines[0]):
             LOG.debug('Processing with Jinja2')
+            tvars = copy(template_vars)
+            if rose_vars['jinja2:suite.rc'] is not None:
+                for key, value in rose_vars['jinja2:suite.rc'].items():
+                    tvars[key] = literal_eval(value)
```
TomekTrzeciak, line 284: @wxtim, sorry to be a pest, but is `literal_eval` good enough here? There are two ways in which this can break existing suites:
1. Jinja2 code is no longer valid as a value, so things like `range` can't be used anymore. This is probably a very niche use case, so it might be OK to drop it.
2. Environment variables in this section are no longer processed here. This is a much bigger problem, because it's common to have things like `SOME_VAR="\$HOME/some/path"` in this section that are meant to be expanded on the job rather than suite host. Rose will write this as `{% set SOME_VAR="$HOME/some/path" %}`, unlike the above implementation, which will keep the `\$` escape.

An alternative could be to refactor `jinja2:suite.rc` processor in Rose into two methods:
1. Generate `suite.rc` Jinja2 include lines.
2. Write the above lines into `suite.rc`.

This way you could use the first method here directly to inject the Jinja2 lines the way `rose suite-run` did, so the behaviour should be identical. It would also keep the implementation largely where it belongs, i.e. in Rose rather than Cylc, and EmPy would be supported automatically in this way too.

wxtim: [initial response: Will edit comment to add fuller response after reading more carefully] @TomekTrzeciak  - you're not being a pest - I'd rather have this now, from you than later, from some poor sap whose suite has broken in an unexpected way.

oliver-sanders: 1. I think this is a positive change.
2. Good spot, if this is the only issue we should be able to work around it easy enough with a bit of string manipulation.

We could really do with ditching the `suite.rc` mangling that Rose2019 does so would avoid writing vars directly into the `suite.rc` at all costs (e.g. you cannot override these vars from the Cylc CLI or any future Cylc files).

wxtim: > The #!Jinja2 shebang is no longer automatically added. 

This sounds kinda fail-safe - if you want jinja2 you _really_ have to want it.

> Rose has some special env var expansion logic built-in to do this, we should be able to import it and bring back this logic easily.

Another PR for another day? I feel mission creep.

oliver-sanders: Can do though it should only be a couple of lines as this was in the original Rose code.

TomekTrzeciak: > ```shell
> grep -rn --color  --include="rose-suite*.conf" range\(.*\) \
>     /path/to/*manyusers*/conventional_place_to_store_suites/rose-suite*.conf
> ```
> 
> returns one user, a @TomekTrzeciak.... (Oh, and me, now I have a toy suite I made to understand this.

I did find more instances than this, but it is rather uncommon (and as @oliver-sanders points out, might be a good riddance anyway). Mind you, `range` is just an example here, it could be any valid Jinja2 expression like `some_dict.values() | list`.

wxtim: Yes I got the search term of my original grep wrong.

TomekTrzeciak: > > Rose has some special env var expansion logic built-in to do this, we should be able to import it and bring back this logic easily.
> 
> Another PR for another day? I feel mission creep.

It's just [one import away](https://github.com/metomi/rose/blob/ea60ad9/metomi/rose/config_processors/jinja2.py#L26):

```python
from metomi.rose.env import env_var_process
```

oliver-sanders: Ok, here is my test suite (for MO people /home/h06/osanders/roses/jinja2-test) which I think covers the Rose2019 behaviours that we would want to preserve:

@TomekTrzeciak note 3.3. the escaped env var seems to come through to Jinja2 unescaped though I may have misread your comment above:

*rose-suite.conf*
```ini
[env]
FOO=a
 
[jinja2:suite.rc]
BAR="${FOO}b"
LOCAL_ENV="$XYZ"
ESCAPED_ENV="\$HOME"
INT=42                                                                                                       
BOOL=True
LIST=["a", 1, True]
```

*suite.rc*
```jinja2
# Test with rose `XYZ=xyz suite-run --new -l` (which invokes cylc validate --strict)                         
# Rose env vars are logged to the console
 
# 1. The Jinja2 shebang should be added automatically
{{ assert(True, "failed 1.1") }}
 
##
{% from "subprocess" import Popen %}
{% from "subprocess" import PIPE %}
{# {% from "cylc.flow" import LOG %} #}  # cylc8
{% from "cylc" import LOG %}
 
# 2. ROSE_SUITE_VARIABLES should contain a mapping of all vars
{{ assert(ROSE_SUITE_VARIABLES is defined, "failed 2.1") }}
{{ assert(ROSE_SUITE_VARIABLES["INT"] == 42, "failed 2.2") }}
{% do LOG.info(ROSE_SUITE_VARIABLES) %}
 
# 3. Env vars
# 3.1 - Local env vars should be templated into rose-suite.conf[env]
{{ assert(LOCAL_ENV == "xyz", "failed 3.1") }}
# 3.2 - rose-suite.conf[env] vars should be templated into rose-suite.conf[jinja2:suite.rc]
{{ assert(BAR == "ab", "failed 3.2") }}
# 3.3 - Escaped env vars should not be templated
{{ assert(ESCAPED_ENV == "$HOME", "failed 3.3") }}
 
# 4. Literals should be evaluated
{{ assert(INT == 42, "failed 4.1") }}
{{ assert(BOOL == True, "failed 4.2") }}
{{ assert(LIST == ["a", 1, True], "failed 4.3") }}
 
# 5. ROSE_VERSION should be provided
{{ assert(ROSE_VERSION is defined, "failed 5.1") }}
{% set cli_rose_version = Popen(["rose", "version"], stdout=PIPE).communicate()[0] %}
{% set cli_rose_version = cli_rose_version.replace('Rose ', '') %}  # strip the "Rose " prefix
{% set cli_rose_version = cli_rose_version.split(' ')[0] %}  # strip the (location) suffix
{% do LOG.info("$ rose version: " + cli_rose_version) %}
{{ assert(ROSE_VERSION == cli_rose_version, "failed 5.2") }}
 
# 6. CYLC_VERSION should be provided
{{ assert(CYLC_VERSION is defined, "failed 6.1") }}
{% set cli_cylc_version = Popen(["cylc", "version"], stdout=PIPE).communicate()[0] %}
{% set cli_cylc_version = cli_cylc_version.strip() %}  # remove trailing newline
{% do LOG.info("$ cylc version: " + cli_cylc_version) %}
{{ assert(CYLC_VERSION == cli_cylc_version, "failed 6.2") }}
 
# 7. ROSE_ORIG_HOST should be provided (hard to test it is set correctly here)
{{ assert(ROSE_ORIG_HOST is defined, "failed 7.1") }}
 
# 8. ROSE_SITE should be provided
{{ assert(ROSE_SITE is defined, "failed 8.1") }}
 
 
# chuck in some Cylc stuff to make validate happy
[scheduling]
    initial cycle point = now 
    [[dependencies]]
        [[[R1]]]
            graph = """ 
                x   
            """ 
 
[runtime]
    [[x]]
```

TomekTrzeciak: > @TomekTrzeciak note 3.3. the escaped env var seems to come through to Jinja2 unescaped though I may have misread your comment above:

Do you mean on Cylc7+Rose2019 or Cylc8+rose-plugin from this PR? I haven't actually tested on this PR (setting up Cylc8 for dev work is too complicated these days), just relying on prior experience with these bits of code. E.g. 3.3 comes through escaped with my own `jinja2render` tool that uses `rose config` to retrieve the vars (should be equivalent to what happens in this PR).

oliver-sanders: The above test suite was written against Rose2019, Cylc7, it does not yet pass against this PR.

wxtim: Thankyou, @oliver-sanders  for writing my functional tests for me. 😄

oliver-sanders: Looks good to me.

A nice interface for implementing this might be:

```
install_plugin() -> {
    'template variables': {},
    'template engine': '',
    ...
}
```

To avoid Cylc having to know about Rose configurations.

cylc/flow/parsec/fileparse.py
```diff
@@ -239,27 +259,54 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
         flines = inline(
             flines, fdir, fpath, False, viewcfg=viewcfg, for_edit=asedit)
 
+    # If empy and jinja2 sections are both filled raise an error.
+    if (all([extra_vars['empy:suite.rc'], extra_vars['jinja2:suite.rc']])):
+        raise FileParseError(
+            "Your additional configuration files define both empy and jinja2"
+            "variables. This makes sense."
```
oliver-sanders, line 266:
```suggestion
            "variables. This doesn't makes sense."
```

cylc/flow/parsec/fileparse.py
```diff
@@ -239,27 +259,54 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
         flines = inline(
             flines, fdir, fpath, False, viewcfg=viewcfg, for_edit=asedit)
 
+    # If empy and jinja2 sections are both filled raise an error.
+    if (all([extra_vars['empy:suite.rc'], extra_vars['jinja2:suite.rc']])):
```
oliver-sanders, line 263: Should really handle this in cylc-rose and return something like `template_vars` so that cylc-flow doesn't need to know anything about Rose configurations (especially as they are likely to change).

cylc/flow/parsec/fileparse.py
```diff
@@ -224,6 +229,21 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
+    # Load Rose Vars, if a ``rose-suite.conf`` file is present.
+    extra_vars = {
+        'env': None,
+        'empy:suite.rc': None,
+        'jinja2:suite.rc': None
+    }
+    for entry_point in pkg_resources.iter_entry_points(
+        'cylc.pre_configure'
+    ):
+        extra_vars = entry_point.resolve()(Path(fpath).parent)
```
oliver-sanders, line 245: If more than one plugin is defined then each plugin that is run will override the results of the previous plugin.

Need to do something like:
```suggestion
        plugin_result = entry_point.resolve()(Path(fpath).parent)
        extra_vars['env'].update(plugin_result.get('env', {}))
```

cylc/flow/parsec/fileparse.py
```diff
@@ -239,27 +259,54 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
-            flines = empyprocess(flines, fdir, template_vars)
+            flines = empyprocess(flines, fdir, tvars)
 
     # process with Jinja2
     if do_jinja2:
+        if (
+            extra_vars['jinja2:suite.rc'] is not None and
+            not re.match(r'^#![jJ]inja2\s*', flines[0])
+        ):
+            flines.insert(0, '#!jinja2')
```
oliver-sanders, line 297: This is good, might be an idea to add an extra check here to make sure that some other shebang (e.g. empy) hasn't been defined: along the lines of:

```python
if re.match('^#!'. flines[0]):
    raise Exception()
```

cylc/flow/parsec/fileparse.py
wxtim: I have now moved this check into the plugin. :)

cylc/flow/parsec/fileparse.py
TomekTrzeciak: This makes me wonder if you thought about how to handle name clashes. With multiple plugins you might have a situation, where the same variable is being set by multiple plugins, which one wins then? Is the order predictable/controllable? If not, you may potentially end up with weird edge cases. I wonder if the interface shouldn't look something like this:

```python
result -> {
    'template variables': {},
    'template engine': get_template_engine(flines[0]),
    ...
}
for plugin in iter_plugins():
    result = plugin(result)
```

This way each plugin could at least see what the previous ones did and decide what to do (skip, override, merge, throw an error).

cylc/flow/parsec/jinja2support.py
```diff
@@ -182,6 +182,14 @@ def jinja2environment(dir_=None):
     env.globals['environ'] = os.environ
     env.globals['raise'] = raise_helper
     env.globals['assert'] = assert_helper
+
+    # if template_vars is not None:
+    #     # I expect multiple things in this block
+    #     if 'ROSE_SUITE_VARIABLES' in template_vars:
+    #         env.globals['ROSE_SUITE_VARIABLES'] = template_vars[
```
datamel, line 189: Ahha, @wxtim, this was the code I was questioning. Although uncommenting does not help with the test failures.

wxtim: Yes, I ought to uncomment it.

cylc/flow/parsec/jinja2support.py
```diff
@@ -137,7 +137,7 @@ def _load_jinja2_extensions():
     return jinja2_extensions
 
 
-def jinja2environment(dir_=None):
+def jinja2environment(dir_=None, template_vars=None):
```
TomekTrzeciak, line 140:
```suggestion
def jinja2environment(dir_=None):
```

cylc/flow/parsec/jinja2support.py
```diff
@@ -219,7 +220,7 @@ def jinja2process(flines, dir_, template_vars=None):
     # Convert unicode to plain str, ToDo - still needed for parsec?)
 
     try:
-        env = jinja2environment(dir_)
+        env = jinja2environment(dir_, template_vars)
```
TomekTrzeciak, line 223:
```suggestion
        env = jinja2environment(dir_)
```

cylc/flow/parsec/fileparse.py
```diff
@@ -224,6 +229,27 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
+        'template variables': {},
+        'templating detected': None
+    }
+    for entry_point in pkg_resources.iter_entry_points(
+        'cylc.pre_configure'
+    ):
+        plugin_result = entry_point.resolve()(Path(fpath).parent)
+        for section in ['env', 'template variables']:
+            if plugin_result[section] is not None:
+                extra_vars[section].update(plugin_result.get(section, {}))
```
TomekTrzeciak, line 248: This is potentially problematic in case of multiple plugins trying to write the same variables (e.g., `$PATH`), cf. https://github.com/cylc/cylc-flow/pull/3913#discussion_r530333657.

cylc/flow/parsec/fileparse.py
```diff
@@ -239,27 +265,59 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
+                raise FileParseError(
+                    "Your file has shebang line incompatible with jinja2"
+                    " variables."
+                )
         if flines and re.match(r'^#![jJ]inja2\s*', flines[0]):
             LOG.debug('Processing with Jinja2')
+            tvars = copy(template_vars)
+            if extra_vars['templating detected'] == 'jinja2:suite.rc':
+                for key, value in extra_vars['template variables'].items():
+                    tvars[key] = value
```
TomekTrzeciak, line 314: Are rose vars supposed to override command line suite variables or should this be the other way around?

oliver-sanders: CLI should override.

cylc/flow/parsec/fileparse.py
oliver-sanders: Fundamentally there are two options:

* Fail if two plugins provide the same variable.
* Override based on a stable order.

We could provide the result to plugins, however, there isn't anything intelligent that the Rose plugin could do to resolve to resolve such an issue.

cylc/flow/parsec/fileparse.py
wxtim: I'm not sure I have a quick or clever answer...

@oliver-sanders ?

cylc/flow/parsec/fileparse.py
wxtim: thus?
```suggestion
            if extra_vars['templating detected'] == 'jinja2:suite.rc':
                for key, value in template_vars.items():
                    extra_vars[key] = value
            tvars = extra_vars
```

cylc/flow/parsec/fileparse.py
oliver-sanders: I put a comment on the other thread - https://github.com/cylc/cylc-flow/pull/3913#discussion_r531685998

I think if two plugins try to write the same variables then the flow design is flawed. I don't think we want to build a system to handle inter-dependence between installation plugins. Easier just to say, don't do that.

cylc/flow/parsec/fileparse.py
wxtim: I think that I've fixed this @TomekTrzeciak .
Question for you and @oliver-sanders  : If we use `cylc <app> --set=VAR=99` and in the `rose-suite.conf`: `VAR=144` then var will have a different variable in the `ROSE_SUITE_VARS` because these are set in the plugin. It potentially allowsu workflow designers to prevent users using `--set` by writing `ROSE_SUITE_VARS['VAR']` in the suite. Is this ok?

oliver-sanders: Fine.

wxtim:
I'm please that this is failing - it was quite concerning that it was skipping loads of tests and appearing to be passing without actually running any tests before.

Perhaps as a meta thing we should put in a test that checks that all tests have not been skipped.

cylc/flow/parsec/fileparse.py
TomekTrzeciak: > Fundamentally there are two options:
> 
> * Fail if two plugins provide the same variable.

I would go for that. Can always be revisited later as use cases arise without breaking existing suites.

> * Override based on a stable order.
> 
> We could provide the result to plugins, however, there isn't anything intelligent that the Rose plugin could do to resolve to resolve such an issue.

Fair point, and different plugins might want to implement different and incompatible strategies for conflict resolution.

cylc/flow/parsec/fileparse.py
```diff
@@ -239,27 +268,63 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
+            else:
+                raise FileParseError(
+                    "Your file has shebang line incompatible with jinja2"
+                    " variables."
+                )
         if flines and re.match(r'^#![jJ]inja2\s*', flines[0]):
             LOG.debug('Processing with Jinja2')
+            if extra_vars['templating detected'] == 'jinja2:suite.rc':
+                for key, value in template_vars.items():
+                    extra_vars['template variables'][key] = value
```
TomekTrzeciak, line 318: This brings about another problem: `--set` variables are always strings, while rose vars in here can be of any type allowed by `literal_eval`, so a blanket override like this might change the expected type.

oliver-sanders: Yes and it's exactly the same problem again, we have but two options:

1. Allow the override, the user got exactly what they asked for.
2. Fail, reject the override, it might not be safe.

cylc/flow/parsec/jinja2support.py
TomekTrzeciak: @wxtim, looks like you've blasted this away with a force push.

cylc/flow/parsec/fileparse.py
TomekTrzeciak: > Yes and it's exactly the same problem again, we have but two options:
> 
> 1. Allow the override, the user got exactly what they asked for.
> 2. Fail, reject the override, it might not be safe.
> 
> Unless there are any better ideas I'm inclined to go with (1) for this and all other previously outlined cases, albeit with a warning.

3. Change the semantics of cylc `--set` vars to go via `literal_eval` to match the semantics of rose vars.

This would introduce incompatibility from the current behaviour, but would be better in the long run anyway IMO. Porting from `--set foo=bar` to `--set foo="bar"` is trivial, similarly for the `--set-file` option and there's no fixed-name equivalent of `rose-suite.conf` to worry about.

oliver-sanders: Erm, that's orthogonal, you still have to pick either (1) or (2), changes to the way Cylc interprets CLI inputs is a different kettle of fish.

oliver-sanders: Discussed at the monthly meeting, considering other CLI breakages we are ok with this, if it is a problem then we could potentially use a separate CLI flag for typed as opposed to string options.

I've raised a PR here: https://github.com/cylc/cylc-flow/pull/3974

Note that it upgrades template var set files in the same way which we may or may not want to go ahead with depending on usage.

cylc/flow/parsec/fileparse.py
```diff
@@ -224,6 +228,31 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
     do_empy = True
     do_jinja2 = True
     do_contin = True
+
+    if not template_vars:
+        template_vars = {}
```
oliver-sanders, line 233: The cli `template_vars` should be able to override those coming from the plugin(s).

cylc/flow/parsec/fileparse.py
```diff
@@ -239,27 +268,63 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
         flines = inline(
             flines, fdir, fpath, False, viewcfg=viewcfg, for_edit=asedit)
 
+    template_vars['CYLC_VERSION'] = __version__
+
     # process with EmPy
     if do_empy:
+        if (
+            extra_vars['templating detected'] == 'empy:suite.rc' and
```
oliver-sanders, line 276: Just return the name of the template engine without the `:suite.rc` suffix.

```suggestion
            extra_vars['templating detected'] == 'empy' and
```

cylc/flow/parsec/fileparse.py
```diff
@@ -239,27 +268,63 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
+            not re.match(r'^#![Ee]m[Pp]y\s*', flines[0])
+        ):
+            if not re.match(r'^#!', flines[0]):
+                flines.insert(0, '#!empy')
+            else:
+                raise FileParseError(
+                    "Your file has shebang line incompatible with empy"
+                    " variables."
```
oliver-sanders, line 284: Highly unlikely anyone will hit this but could make it a little more informative, say:

```suggestion
                    f"Your file has shebang line for {extra_vars['templating_detected']}"
                    f" but {flines[0]} configured in flow.cylc.
```

Same for Jinja2 equiv.

(untested)

cylc/flow/parsec/fileparse.py
```diff
@@ -239,27 +268,63 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
+                )
         if flines and re.match(r'^#![Ee]m[Pp]y\s*', flines[0]):
             LOG.debug('Processing with EmPy')
+            if extra_vars['templating detected'] == 'empy:suite.rc':
+                for key, value in template_vars.items():
+                    extra_vars[key] = value
+                template_vars = extra_vars
```
oliver-sanders, line 291: don't have to do this in the empy/jinja2 branches separately.

cylc/flow/parsec/fileparse.py
oliver-sanders: Suggest going with (1) and just raising a warning something along the lines of:

```python
if key in extra_vars['template variables']:
    LOG.warning(f'overriding {key}: {extra_vars["template variables'']} -> {value}')
```

setup.py
```diff
@@ -53,7 +53,9 @@ def find_version(*file_paths):
     'pyuv==1.4.*',
     'pyzmq==19.0.*',
     'psutil>=5.6.0',
+    # 'metomi-rose>=2.*',
     'urwid==2.*',
+    'nest_asyncio',
```
oliver-sanders, line 58: I think we were able to remove this?

setup.py
```diff
@@ -53,7 +53,9 @@ def find_version(*file_paths):
     'pyuv==1.4.*',
     'pyzmq==19.0.*',
     'psutil>=5.6.0',
+    # 'metomi-rose>=2.*',
```
oliver-sanders, line 56: Can kill this.

wxtim: Completely, or just uncomment?

cylc/flow/parsec/fileparse.py
wxtim: This line creates a blank dict if the `template_vars` kwarg isn't given, so I'm not sure why it would prevent kwarg template vars over-riding the plugin. I'll:
- Check whether it's actually necessary
- Check the rest of the code, assuming that this is a general comment.

setup.py
wxtim:
```suggestion
```

setup.py
wxtim:
```suggestion
```

cylc/flow/parsec/fileparse.py
wxtim: Dun.

cylc/flow/parsec/fileparse.py
```diff
@@ -224,6 +228,31 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
+
+    if not template_vars:
+        template_vars = {}
+
+    # Load Rose Vars, if a ``rose-suite.conf`` file is present.
+    extra_vars = {
+        'env': {},
+        'template variables': None,
+        'templating detected': None
+    }
```
TomekTrzeciak: Naming could be tweaked a bit, since this is essentially a public interface (no spaces, so you can use the names as kwargs `extra_vars.update(template_variables={})`:
```suggestion
    extra_vars = {
        'environment_variables': {},
        'template_variables': None,
        'template_engine': None
    }
```

wxtim: Good idea.

oliver-sanders:
Using the `jinja2-test` suite I'm getting the following error when trying to run `cylc validate .` on the suite:

```
Jinja2Error: Jinja2 Assertion Error: failed 2.1
Context lines:

# 2. ROSE_SUITE_VARIABLES should contain a mapping of all vars
{{ assert(ROSE_SUITE_VARIABLES is defined, "failed 2.1") }}	<-- Exception
```

> **Note:** We wouldn't expect "1.1" (auto insertion of the Jinja2 shebang) to work
> for validate (as it is applied on install) so I prepended this to the output first.

After a bit of digging it looks like `viewcfg` is not set for validate cases in `read_and_proc`.

cylc/flow/parsec/fileparse.py
TomekTrzeciak: > I think if two plugins try to write the same variables then the flow design is flawed. I don't think we want to build a system to handle inter-dependence between installation plugins. Easier just to say, don't do that.

I think it's best to detect collisions and throw an error in that case:

```suggestion
                section_update = plugin_result.get(section, {})
                keys_collision = extra_vars[section].keys() & section_update.keys()
                if keys_collision:
                    raise ParsecError(section + " collision: " + ", ".join(sorted(keys_collision)))
                extra_vars[section].update(section_update)
```

cylc/flow/parsec/fileparse.py
```diff
@@ -224,6 +228,31 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
+        for section in ['env', 'template_variables']:
+            if section in plugin_result and plugin_result[section] is not None:
+                extra_vars[section].update(plugin_result.get(section, {}))
+
+        if 'templating_detected' in plugin_result and plugin_result[
+            'templating_detected'
+        ] is not None:
+            extra_vars['templating_detected'] = plugin_result[
+                'templating_detected'
+            ]
```
TomekTrzeciak, line 254: Similarly, this should be tested for collision.

cylc/flow/parsec/fileparse.py
```diff
@@ -239,27 +268,63 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
             LOG.debug('Processing with Jinja2')
             try:
                 from cylc.flow.parsec.jinja2support import jinja2process
             except (ImportError, ModuleNotFoundError):
                 raise ParsecError('Jinja2 Python package must be installed '
                                   'to process file: ' + fpath)
-            flines = jinja2process(flines, fdir, template_vars)
+            flines = jinja2process(
+                flines, fdir, template_vars
+            )
```
TomekTrzeciak, line 327: The logic here looks a bit suspect. I think that if you have template with `#!jinja2` and plugin with `'templating_detected': 'empy'` you end up passing the plugin variables to the template engine regardless.

It might be worth to eliminate this split between jinja2/empy and simplify the logic like so:

```python
# assume viewcfg['empy'] + viewcfg['jinja2'] replaced by viewcfg['template']
do_templating = viewcfg['template']

template_engines = {
    'jinja2': jinja2process,
    'empy': empyprocess,
}

templating_detected = re.match(r'#!(\w+)\s*', flines[0])
templating_detected = templating_detected and templating_detected.group(1).lower()

for entry_point in pkg_resources.iter_entry_points('cylc.pre_configure'):
    ...  # detect collisions and update extra_vars

if extra_vars.get('templating_detected'):
    if not templating_detected:
        templating_detected = extra_vars['templating_detected']
        flines.insert(0, templating_detected)
    elif extra_vars['templating_detected'] != templating_detected:
        raise FileParseError(
            'Mismatch between parsed and configured template processor: %s != %s'
            % (templating_detected, extra_vars['templating_detected'])
        )

template_vars = {**extra_vars['template_variables'], **template_vars}

if do_templating and templating_detected and flines:
    LOG.debug('Processing with ' + templating_detected)
    flines = template_engines[templating_detected](flines, fdir, template_vars)
```

wxtim: This is very likely a good idea, but it's also mission creep. I'm happy to address the suggested re-structure but I'd like to put this PR to bed.

wxtim: @TomekTrzeciak  - If I've understood you correctly the new functional test `08-template-engine-conflict.t` should demonstrate that if you set a shebang in your `flow.cylc` file and the plugin returns a `templating_detected` that is different you will get a validation failure.

wxtim:
@oliver-sanders  wrote:

> Using the jinja2-test suite I'm getting the following error when trying to run cylc validate . on the suite:

Fixed.

cylc/flow/parsec/fileparse.py
```diff
@@ -239,27 +291,71 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
+    # Push template_vars into extra_vars so that duplicates come from
+    # template_vars.
+    if extra_vars['templating_detected'] is not None:
+        will_be_overwritten = set(template_vars.keys()).intersection(
+            set(extra_vars['template_variables'].keys())
+        )
+        for key in will_be_overwritten:
+            LOG.warning(
+                f'Over-riding {key}: {extra_vars["template_variables"]} ->'
+                f' {template_vars[key]}'
```
oliver-sanders, line 305:
```suggestion
                f'Overriding {key}: {extra_vars["template_variables"][key]} ->'
                f' {template_vars[key]}'
```

oliver-sanders:
follow on: https://github.com/cylc/cylc-flow/issues/3988

oliver-sanders:
Trying out different `rose-suite.conf` permutations, I've found one which works in Rose2019 which doesn't here, need to pass environment variables through in `cylc install`. e.g:

```
[env]
x=foo
 
[namelist:foo]
whatever=meh
 
[file:abc]
source=namelist:$x 
```

wxtim:
> Trying out different rose-suite.conf permutations, I've found one which works in Rose2019 which doesn't here, need to pass environment variables through in cylc install. e.g:

Embarrassingly it was because I'd failed to name the entry point I wanted correctly in Cylc Install. Duh!

oliver-sanders:
> These changes partially address #3819

Does this not close the issue (circa any issues open on cylc-rose)

wxtim:
> Does this not close the issue (circa any issues open on cylc-rose)?

@oliver-sanders  alongside the plugin PR...

cylc/flow/parsec/fileparse.py
```diff
@@ -239,27 +291,71 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
     if do_empy:
+        if (
+            extra_vars['templating_detected'] == 'empy' and
+            not re.match(r'^#![Ee]m[Pp]y\s*', flines[0])
+        ):
+            if not re.match(r'^#!', flines[0]):
+                flines.insert(0, '#!empy')
+            else:
+                raise FileParseError(
+                    "You plugins give a templating engine: "
```
oliver-sanders, line 320: Rephrase.

cylc/flow/parsec/fileparse.py
```diff
@@ -239,27 +291,71 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
     if do_jinja2:
+        if (
+            extra_vars['templating_detected'] == 'jinja2' and
+            not re.match(r'^#![jJ]inja2\s*', flines[0])
+        ):
+            if not re.match(r'^#!', flines[0]):
+                flines.insert(0, '#!jinja2')
+            else:
+                raise FileParseError(
+                    "You plugins give a templating engine: "
```
oliver-sanders, line 345: Ditto.

cylc/flow/parsec/jinja2support.py
oliver-sanders: Is this resolved?

oliver-sanders (commented):
Testing went well, nothing unexpected, Jinja2 example worked (but that's in the tests so no surprise there), empy worked too.

Can't come up with any Rose7/Rose2 examples we want to support that don't work :+1:.

Minor comments.

oliver-sanders:
> @oliver-sanders alongside the plugin PR...

Should update the PR description to close the issue?

wxtim:
@oliver-sanders  - Because I/we unified the treatment of template_vars from the CLI and from the plugin(s) CLI strings were `literal_eval`''d twice - i.e:
```python
template_var['Hello'] = literal_eval("world")
...
# Now in the fileparse.py
template_var['Hello'] = literal_eval(template_var['Hello'])
# --> Error!
```

oliver-sanders:
> Because I/we unified the treatment of template_vars from the CLI and from the plugin(s) CLI strings were literal_eval''d twice - i.e:

Er, no.

The Rose template vars were `literal_eval`'ed twice, however, CLI settings were only `literal_eval`ed once. Now they aren't `literal_eval`ed at all.

tests/functional/rose-conf/fileinstall_data/lion.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python3
+
+class TerriblePunException(Exception):
+    pass
+
+
+def lion():
+    raise TerriblePunException("This is a Lion's Main script")
```
oliver-sanders, line 8: 🍌

oliver-sanders:
@TomekTrzeciak some follow on work captured in [these issues](https://github.com/cylc/cylc-rose/issues) otherwise this should be done.

> **Note:** The `cylc install` command (replacement for `rose suite-run`) is not yet implemented hence this PR adds a quick dummy command which doesn't make much sense in of itself. The src/run dir separation and the rest of the install functionality will come in #4000.

oliver-sanders:
Code looks good, need to go through and test again but not expecting any surprises.

tests/functional/rose-conf/fileinstall_data/lion.py
wxtim: wait until you hear about
```python
def count_of():
    ...

count_of(monte_christo)

**Comment by TomekTrzeciak** [comment]
> @TomekTrzeciak some follow on work captured in [these issues](https://github.com/cylc/cylc-rose/issues) otherwise this should be done.

Thanks everyone for perseverance on this one, I think the final result is better for it.

One more thing: I think the suggestion to consolidate jinja2/empy code paths https://github.com/cylc/cylc-flow/pull/3913#discussion_r538660381 is still worth doing at some point, but it would require altering `cylc view` options. It was my mistake to add `--empy` option there alongside `--jinja2` rather than replace it with something more generic. Since you can't have more than one templating engine on the shebang line, it makes no sense to have more than one option for it. Perhaps it's worth to capture this in a separate ticket?

**Comment by oliver-sanders** [comment]
Can do, tag it against `8.x`.

**Comment by TomekTrzeciak** [comment]
What's the plan for replacement of `-O` command line option: https://github.com/cylc/cylc-flow/issues/3819#issuecomment-726075270. Is it this PR or follow on?

**Comment by oliver-sanders** [comment]
Will stay as `-O`, not to be done in this PR (needs a plugin interface that can mutate the CLI)

File: tests/functional/rose-conf/01-empy.t [hunk 31]
```diff
@@ -0,0 +1,33 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test jinja2 from rose-suite.conf file is processed into a suite.
```
MetRonnie, line 18:
```suggestion
# Test empy from rose-suite.conf file is processed into a suite.
```

tests/functional/rose-conf/01-empy/processed.conf.control
```diff
@@ -0,0 +1,30 @@
...
+    [[respond]]
+    [[plunge_control]]
+    [[plunge_yan]]
+    [[plunge_tan]]
+    [[plunge_tethera]]
+    [[allow]]
+    [[1_auf_deutsch_ist_ein]]
+    [[2_auf_deutsch_ist_zwei]]
+    [[3_auf_deutsch_ist_drei]]
+    [[fin]]
```
MetRonnie, line 30: If this is the same as `tests/functional/rose-conf/00-jinja2/processed.conf.control`, why not symlink it instead of duplicating?

tests/functional/rose-conf/03-fileinstall.t
```diff
@@ -0,0 +1,51 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test jinja2 from rose-suite.conf file is processed into a suite.
```
MetRonnie, line 18: I think this needs updating?

tests/functional/rose-conf/04-opts-set-from-env.t
```diff
@@ -0,0 +1,36 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test jinja2 from rose-suite.conf file is processed into a suite.
```
MetRonnie, line 18: Could do with being more specific

tests/functional/rose-conf/05-opts-set-from-rose-suite.conf.t
```diff
@@ -0,0 +1,36 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test jinja2 from rose-suite.conf file is processed into a suite.
```
MetRonnie, line 18: Ditto

tests/functional/rose-conf/06-jinja2.thorough/flow.cylc
```diff
@@ -0,0 +1,63 @@
...
+{{ assert(True, "failed 1.1") }}
+
+##
+{% from "subprocess" import Popen %}
+{% from "subprocess" import PIPE %}
+{% from "cylc.flow" import LOG %} # cylc8
+
+# 2. ROSE_SUITE_VARIABLES should contain a mapping of all vars
+{{ assert(ROSE_SUITE_VARIABLES is defined, "failed 2.1") }}
+{{ assert(ROSE_SUITE_VARIABLES["INT"] == 42, "failed 2.2") }}
```
MetRonnie, line 14: Nice

tests/functional/rose-conf/06-jinja2.thorough/suite.rc
```diff
@@ -0,0 +1,84 @@
...
+[scheduling]
+    initial cycle point = now
+    [[dependencies]]
+        [[[R1]]]
+        graph = """
+            x
+        """
+
+[runtime]
+    [[x]]
```
MetRonnie, line 84: Is this needed for backwards compatibility? If so, why is it slightly different to the flow.cylc one? Couldn't it be a symlink instead?

tests/functional/rose-conf/08-template-engine-conflict.t
```diff
@@ -0,0 +1,35 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test jinja2 from rose-suite.conf file is processed into a suite.
```
MetRonnie, line 18: Could do with updating

MetRonnie (changes requested):

cylc/flow/parsec/fileparse.py
```diff
@@ -239,27 +297,71 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
...
             flines, fdir, fpath, False, viewcfg=viewcfg, for_edit=asedit)
 
+    template_vars['CYLC_VERSION'] = __version__
+
+    # Push template_vars into extra_vars so that duplicates come from
+    # template_vars.
+    if extra_vars['templating_detected'] is not None:
+        will_be_overwritten = set(template_vars.keys()).intersection(
+            set(extra_vars['template_variables'].keys())
+        )
```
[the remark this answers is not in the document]
wxtim: No, there is no reason - the & syntax was suggested by Tomel wheras I wrote these explicitly. Happy to be consistent if you wish it.

tests/functional/rose-conf/01-empy/processed.conf.control
wxtim: Good idea.

tests/functional/rose-conf/06-jinja2.thorough/flow.cylc
wxtim: Oliver's.

tests/functional/rose-conf/06-jinja2.thorough/suite.rc
wxtim: It's inclusion is almost certainly a mistake. I don't think that it's needed at all.

cylc/flow/parsec/fileparse.py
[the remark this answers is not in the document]
MetRonnie: I hadn't heard of using `&` for intersection and `|` for union with sets before, but it seems fairly intuitive once you know

MetRonnie (approved):
LGTM assuming tests pass

oliver-sanders:
Tests now passing, will have a final test today.

oliver-sanders:
Commits will need rationalising pre-merge.

wxtim:
> Commits will need rationalising pre-merge.

@oliver-sanders  down to 9 ok?

oliver-sanders (approved):
🥳

PR #3924, opened 2020-11-06
fix null-stripping of repeated composite fields

### LINKED ISSUE #3922: TaskProxy prerequisites missing from delta-query subscription with stripNull: true
**Describe the bug**

Related to https://github.com/cylc/cylc-ui/issues/402. I'm updating the UI components to display message triggers as in the linked issue. The first step was altering the client GraphQL query to request the message for each job.

That can be done via the job's task proxy, which will contain the prerequisites, wherein it's possible to traverse conditions to find the message in reqState.

The current query in the UI has `stripNull: true` at `deltas(workflows: [$ID], stripNull: true) { }`. And it works well so far.

With the `stripNull: true`, I'm not getting anything in my response. Changing it to `stripNull: false`, the message is returned fine.

When I changed to `stripNull: false`, it brought the message, and I couldn't see anything that was `null`? So I am not sure why `stripNull: true` results in the message being removed/stripped from the results?

`master`

Use the example workflow that @hjoliver shared today on Element

```
# cylc-run/messagetrigger1/flow.cylc

[scheduling]
   cycling mode = integer
   initial cycle point = 1
   [[graph]]
      R1 = """
           foo:out1 => bar
      """
[runtime]
   [[foo]]
       script = """
           sleep 20
           cylc message "data ready for bar"  # <-----
           sleep 20
       """
       [[[outputs]]]
           out1 = "data ready for bar"
```

Now start the Cylc 8 system. Open GraphiQL, and paste the following query.

```graphql
subscription OnWorkflowDeltasData {
  deltas(workflows: ["kinow|messagetrigger1"], stripNull: true) {
    ...WorkflowTreeDeltas
    __typename
  }
}

fragment WorkflowTreeDeltas on Deltas {
  id
  shutdown
  added {
    ...WorkflowTreeAddedData
  }
}

fragment WorkflowTreeAddedData on Added {
  workflow {
    ...WorkflowData
    taskProxies(sort: {keys: ["name"], reverse: false}, ghosts: true) {
      jobs(sort: {keys: ["submit_num"], reverse: true}) {
        ...JobData
      }
    }
  }
  jobs(sort: {keys: ["submit_num"], reverse: true}) {
    ...JobData
    __typename
  }
  __typename
}

fragment WorkflowData on Workflow {
  id
}

fragment JobData on Job {
  id
  taskProxy {
    prerequisites {
      conditions {
        reqState
      }
    }
  }
}
```
Note that `taskProxy` is always empty in the response.

image

Change it to `stripNull: false`, and restart the subscription in GraphiQL.

image

**Expected behavior**

Messages returned correctly, since there are no `null` results in the query when `stripNull: false`?

**Screenshots**

See above.

**Additional context**

**Pull requests welcome!**

dwsutherland:
These changes close #3922 

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/network/graphql.py
+++ b/cylc/flow/network/graphql.py
@@ -95,8 +95,13 @@ def instantiate_middleware(middlewares):
 def null_setter(result):
     """Set type to null if result is empty/null-like."""
     # Only set empty parents to null.
-    if result in EMPTY_VALUES:
-        return NULL_VALUE
+    try:
+        if result in EMPTY_VALUES:
+            return NULL_VALUE
+    except TypeError:
+        # If field is a repeated composite field convert to list.
+        if not result:
+            return NULL_VALUE
     return result
 
 
--- a/tests/unit/network/test_graphql.py
+++ b/tests/unit/network/test_graphql.py
@@ -19,8 +19,13 @@ from typing import Optional
 import pytest
 from graphql import parse
 
-from cylc.flow.network.graphql import AstDocArguments
+from cylc.flow.data_messages_pb2 import PbTaskProxy, PbPrerequisite
+from cylc.flow.network.graphql import AstDocArguments, null_setter, NULL_VALUE
 from cylc.flow.network.schema import schema
+
+
+TASK_PROXY_PREREQS = PbTaskProxy()
+TASK_PROXY_PREREQS.prerequisites.append(PbPrerequisite(expression="foo"))
 
 
 @pytest.mark.parametrize(
@@ -117,3 +122,35 @@ def test_query_variables(
             test()
     else:
         test()
+
+
+@pytest.mark.parametrize(
+    'pre_result,'
+    'expected_result',
+    [
+        (
+            'foo',
+            'foo'
+        ),
+        (
+            [],
+            NULL_VALUE
+        ),
+        (
+            {},
+            NULL_VALUE
+        ),
+        (
+            TASK_PROXY_PREREQS.prerequisites,
+            TASK_PROXY_PREREQS.prerequisites
+        ),
+        (
+            PbTaskProxy().prerequisites,
+            NULL_VALUE
+        )
+    ]
+)
+def test_null_setter(pre_result, expected_result):
+    """Test the null setting of different data types/results."""
+    post_result = null_setter(pre_result)
+    assert post_result == expected_result


hjoliver:
(I'll do 2nd review after @kinow, discoverer of the bug)

cylc/flow/network/graphql.py
```diff
@@ -95,8 +95,13 @@ def instantiate_middleware(middlewares):
 def null_setter(result):
     """Set type to null if result is empty/null-like."""
     # Only set empty parents to null.
-    if result in EMPTY_VALUES:
-        return NULL_VALUE
+    try:
+        if result in EMPTY_VALUES:
+            return NULL_VALUE
```
hjoliver, line 100: I guess you're avoiding booleans here (by not just doing `if not result:...`)?

cylc/flow/network/graphql.py
```diff
@@ -95,8 +95,13 @@ def instantiate_middleware(middlewares):
...
+    except TypeError:
+        # If field is a repeated composite field convert to list.
+        if list(result) in EMPTY_VALUES:
+            return NULL_VALUE
```
hjoliver, line 104: Does a "repeated composite field" not have a "empty value" you could add to `EMPTY_VALUES` to avoid this except-block and list conversion?

cylc/flow/network/graphql.py
dwsutherland: `EMPTY_VALUES`  only contains `[]` and `{}`  as we want to be able to reset booleans and strings to `False`/`""` empties.

cylc/flow/network/graphql.py
dwsutherland: Not sure one repeated composite field is equivalent to another (when empty)... 

Otherwise you'd still need to test that it's a protobuf object/field, and that it's a "repeated composite field" ... And since that type of field might not be too common, I thought it less expensive to just alternate on the exception.

dwsutherland: will check

hjoliver: Sounds reasonable :+1:

kinow (approved):
Tested with the UI, and got the expected values after removing `stripNull: false` from the `taskProxy` in my query :+1: 

Thanks a lot for fixing it so quickly @dwsutherland !

cylc/flow/network/graphql.py
hjoliver: I'll approve and merge once you report back on the result of "will check"...

dwsutherland: Hmm..
```
>>> type(PbPrerequisite().conditions)
<class 'google.protobuf.pyext._message.RepeatedCompositeContainer'>
```
This works:
```
   assert PbTaskProxy().prerequisites == PREREQ.conditions
```
However, I don't think you can add it to the `EMPTY_VALUES` because if it's not empty it will be compared to the other types.. throwing the error..

dwsutherland: Ok I've removed the list conversion.. I think the exception can just test `not result`

hjoliver (approved):
Thanks @dwsutherland :+1:

PR #3925, opened 2020-11-06
Use message not label for API task outputs

dwsutherland:
Small change to use the output message instead of the graph label, as the key to the task outputs:
```
    [[wiz]]
        pre-script = sleep 2
        post-script = sleep 2; cylc message "B I N G O"
        [[[meta]]]
            description = "first task wig"
        [[[environment]]]
            GREETING = "Hello from wig!"
        [[[outputs]]]
            bingo = "B I N G O"
```
image

**Requirements check-list**
- [x] No change log entry required (not released).
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -915,8 +915,8 @@ class DataStoreMgr:
                     prereq_list.append(prereq_obj)
             tp_delta.prerequisites.extend(prereq_list)
             tp_delta.outputs = json.dumps({
-                trigger: is_completed
-                for trigger, _, is_completed in itask.state.outputs.get_all()
+                message: received
+                for _, message, received in itask.state.outputs.get_all()
             })
             extras = {}
             if itask.tdef.clocktrigger_offset is not None:


kinow (commented):
I thought the key was going to be `bingo`, and the value the message?

```
"bingo": "B I N G O"
```

kinow:
Although I think I may have misunderstood it. @hjoliver said something about displaying once the value is `true`, so maybe the key (key for JavaScript, like in an `Object` or `Map` :point_up: ).

If that's the case, happy to approve, just wait for @hjoliver or @dwsutherland to confirm here.

I've got the two messages, now in the originating tasks:

image

dwsutherland:
>I thought the key was going to be bingo, and the value the message?

> Although I think I may have misunderstood it. @hjoliver said something about displaying once the value is true, so maybe the key (key for JavaScript, like in an Object or Map ☝️ ).

Hilary, checked, and reversed that  I thought:
>> I guess the label is used when reporting prerequisite expressions, so we'll need both label and message ...

>Nope, I was wrong there. We now use simple numerical labels for prerequisite expressions.

hjoliver:
> I thought the key was going to be `bingo`, and the value the message?
> 
> ```
> "bingo": "B I N G O"
> ```

@dwsutherland 's got it right.  The general form for both standard and custom outputs is:
```
string(output message): bool(message received)
```

e.g.
```
started: true
the quick brown fox: true
succeeded: false
```
(For a task that started running and sent one custom message, hasn't finished yet)

hjoliver:
The `bingo` (lower case) label is useless to the UI, it just associates the trigger in the graph with the output message in the outputs section of the task definition.

hjoliver (approved):

kinow (approved):

PR #3930, opened 2020-11-10
conf: remove global-tests.cylc and run at tests in actions

oliver-sanders:
Since #3808 the `global-tests.cylc` file no longer has any real purpose.

Test platforms should be available for interactive work as well as automated testing so it makes sense to share the same configuration for both, just configuring test platforms separately from production ones.

This also enables "at" tests in GH actions which were skipped before due to a config issue.

**Requirements check-list**
- [x] No dependency changes.


hjoliver:
@oliver-sanders - is this one ready for review?

oliver-sanders:
Nope!

oliver-sanders:
One for later

PR #3931, opened 2020-11-10
global universal identifier

### LINKED ISSUE #3592: Universal suite/task/cycle delimiter (universal identifier)

oliver-sanders:
closes #3592

> **~Draft~ Done!**
>
> This has been a much harder conversion than anticipated, many apologies :(
> Will try and thrash out the following shortly:
>
> * [x] Need to fix `tests/integration/test_data_store_mgr.py`.
> * [x] Awaiting more tests.
> * [x] Need to fix / re-write `tests/f/graphql/02-root-queries.t`.
> * [x] Need to sort out task and edge IDs in the data store.
> * [x] Needs `UserInputError` messages in `cylc.flow.id_cli`.
> * [x] Needs a little tidying.
> * [x] Need to fix Tui.
> * [x] Should try to make tokens a little safer
>   * Perhaps subclass Dict to screen the tokens.
>   * Or make the dictionary keys use the `Tokens` enum.
> * [x] UIS branch - https://github.com/cylc/cylc-uiserver/pull/297
> * [x] UI branch - https://github.com/cylc/cylc-ui/pull/876

Convert cylc-flow to use the [global universal identifier](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-universal-id.md)

**Overview of UID:**

* See the [cylc-admin proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-universal-id.md)
  * Full format: `~user/workflow//cycle/task/job`.
  * Relative references (CLI only): `//cycle/task/job`.
* Task IDs are now nicely lexicographically sortable (providing all cycle points are of the same length).
  * Consequently there have been some sort order changes in `cylc graph --reference` and `cylc play --reference-log` output.

The universal ID replaces/unifies:

* The data store format:
  * `user|workflow|cycle|task|job`
  * `user|workflow|task-definition`
* The task proxy format:
  * `task.cycle`
  * `task`
* The alternative task proxy format:
  * `cycle/task`
  * `cycle`
* The job dir format:
  * `cycle/task/job`

There should now only be one interface for parsing to and from string IDs as well as a standard
tokens dictionary format for passing around parsed ID information.

Here are some of the parsing interfaces removed in this PR:

* [TaskID](https://github.com/cylc/cylc-flow/blob/5169362f792a7b5bd2fae9f0085325df305e76c5/cylc/flow/task_id.py#L52-L55)
* [parse_reg](https://github.com/cylc/cylc-flow/blob/5169362f792a7b5bd2fae9f0085325df305e76c5/cylc/flow/workflow_files.py#L1179)
* [parse_workflow_id](https://github.com/cylc/cylc-flow/blob/5169362f792a7b5bd2fae9f0085325df305e76c5/cylc/flow/network/schema.py#L133)
* [parse_node_id](https://github.com/cylc/cylc-flow/blob/5169362f792a7b5bd2fae9f0085325df305e76c5/cylc/flow/network/schema.py#L158)
* [collate_node_attrs](https://github.com/cylc/cylc-flow/blob/08bbc0ab1f46880d772a4ae62aae2a6757d31ecf/cylc/flow/network/resolvers.py#L121)
* [parse_task_item](https://github.com/cylc/cylc-flow/blob/5169362f792a7b5bd2fae9f0085325df305e76c5/cylc/flow/task_pool.py#L1646)
* [parse_job_item](https://github.com/cylc/cylc-flow/blob/5169362f792a7b5bd2fae9f0085325df305e76c5/cylc/flow/data_store_mgr.py#L174)

It also unifies the pattern matching in the:

* Data store.
* Task pool.
* Task proxy.
* CLI.

There is plenty of scope for future work better adopting the new tokens internal
interchange format to reduce the amount of needless parsing to / from string IDs.

**CLI Overview:**

```bash
# CLI help
cylc help id

# deprecated (but still supported)
cylc trigger workflow task.cycle

# new
cylc trigger workflow//cycle/task
cylc trigger workflow //cycle/task  # alt

# multi tasks
cylc trigger workflow//cycle1/task1 workflow//cycle2/task2
cylc trigger workflow //cycle1/task1 //cycle2/task2  # alt

# multi workflows
cylc trigger workflow1//*:failed workflow2//:failed workflow2//:submit-failed
cylc trigger workflow1//*:failed workflow2 //:failed //:submit-failed  # alt
```

**Behaviour Changes:**

* CLI:
  * Src workflow paths must now start with "./" or be absolute.
    * We used to allow relative paths and check for ambiguity with run dirs.
    * E.G. `cylc validate foo` could mean `./foo` or `~/cylc-run/foo`.
    * I have not re-integrated this functionality.
    * So now if you want `./foo` you must type `./foo`.
    * Nicer?
  * Using the absolute path to a run directory is no longer supported (use the workflow ID).
  * Task patterns are now explicit rather than implicit.
    * Previously the absence of a task ID meant all tasks (`*`).
    * Now if you want to act on all tasks you must explicitly say so.
  * Cylc clean can now clean multiple workflows and prompts rather than fails in non-force mode.
* Selected CLI commands have been upgraded to support multi-workflow functionality
  via `cylc scan`.
  * E.G. The missing: `cylc stop '*'`!
* Task meta-queries with `cylc show` must now be provided with the `--task-def` argument.
* Trigger now globs against the task pool (and matches future tasks) same as `cylc hold`.
  * It used to match more broadly.
  * Now if you want to trigger non-active tasks you must list them explicitly.
  * Removed tests/functional/cylc-trigger/05-filter-cycles.t
* Namespace objects in the data store (i.e. task and family definitions) now have the ID `//*/<name>`:
  * Definitions don't fit into the UID scheme (hence using `*` for the cycle point).
  * Consider using a special character to cater to them more nicely E.G. `//!namespace`

**Terminology & Variable Names:**

UID gives us a good opportunity to standardise naming conventions, the UID scheme suggests the following:

Variable names:

* User (not owner).
  * Owner was previously used to distinguish from the `--user` (Cylc used to have multi-user functionality).
* Workflow (not flow).
  * Keep "flow" for "reflows".
* Cycle (not point).
* Task (not name).
* Job (not submission_no, submission_num, etc).

CLI:

* WORKFLOW_ID (not REG, FLOW).
* ID (not TASK, CYCLE, TASK_ID, TASK_GLOB, etc).

**Requirements check-list**
- [ ] Contains logically grouped changes - PLEASE REBASE SQUASH
- [ ] Changelog entry.
- [x] Docs - https://github.com/cylc/cylc-doc/pull/377
- [x] Downstream PRs.
- [x] No dependency changes.


oliver-sanders:
@dwsutherlan  I think this will marry well with the pre-existing universal identifier. Tagged you as the original author of the universal identifier to make sure I haven't thrown in any curveballs.

The only thing that I think will trip it up at the moment is the ability to specify selectors on cycle points (as opposed to tasks) though I'm not entirely sure what that would mean in different contexts and we can block that functionality for now.

cylc/flow/id.py
```diff
@@ -0,0 +1,553 @@
...
+    Args:
+        *ids (tuple): Identifier list.
+
+    Raises:
+        ValueError - For invalid identifiers or identifier lists.
+
+    Returns:
+        list - List of tokens dictionaries.
+
+    Examples:
```
oliver-sanders, line 456: These examples show the possible formats and upgrading of legacy ids.

dwsutherland:
@oliver-sanders  - Really amazing work, I wish I was as sophisticated as you!

> The only thing that I think will trip it up at the moment is the ability to specify selectors on cycle points (as opposed to tasks) though I'm not entirely sure what that would mean in different contexts and we can block that functionality for now.

Selectors on cycle points would probably target the root family of that cycle-point (from an API perspective)

Also have you thought about the APIs use of the following ID format?
`'~user/flow:flow_sel//task:task_sel'`

Also, in case you/others need a reminder, `~user` may seem a little redundant at the moment.. However, we have/will-have cross-workflows commands and graph representations (i.e. inter-workflow triggers represented by the nodes in a graph that belong to different workflows)

hjoliver:
> Really amazing work, I wish I was as sophisticated as you!

Join the club - I think it's the "barnyard strength" cider he runs on :beer:

oliver-sanders:
Aw shucks, didn't know there was so much love for regexes.

> Selectors on cycle points would probably target the root family of that cycle-point (from an API perspective)

Could work, seemed like a good idea at the time but it gets a little bit weird:

```
# sets outputs on every task in cycles that have at least one succeeded task in the pool
cylc set-outputs flow/2020:succeeded

# hold all cycles that have at least one failed task in the pool
cylc hold flow//2020:failed
```

> Also have you thought about the APIs use of the following ID format?

Can use a wildcard

```
myflow//*/taskname
```

I think in the cases where you might want to specify a task without a cycle we use the `-n` argument (e.g. `cylc broadcast`) which enables us to sidestep this.

> inter-workflow triggers represented by the nodes in a graph that belong to different workflows

Nice.

The other thing I was thinking about was admin privileges e.g:

```
cylc hold ~alice/*
```

The command would be sent to the UIS for "alice" where it would authenticate as me and check authorisation as it would for a command from the UI.

oliver-sanders:
Untested cycle/task matching for the task_pool (replaces the `filter_task_proxies` method):

```
    def filter_ids(self, ids, out=Tokens.Task):
        match = fnmatchcase
        # TODO: consider using wcmatch which would add support for
        #       extglobs, namely brace syntax e.g. {foo,bar}

        _cycles = []
        _tasks = []
        _jobs = []
        _not_matched = []

        for tokens in ids:
            for lowest_token in reversed(Tokens):
                if tokens.get[lowest_token.value]:
                    break

            cycles = []
            tasks = []
            jobs = []

            # filter by cycle
            if lowest_token == Tokens.Cycle:
                cycle = tokens[Tokens.Cycle.value]
                cycle_sel = tokens.get(Tokens.Cycle.value + '_sel', '*')
                for icycle, itasks in self.pool.items():
                    str_cycle = str(icycle)
                    if not match(str_cycle, cycle):
                        continue
                    if cycle_sel == '*':
                        cycles.append(str_cycle)
                        continue
                    for itask in itasks:
                        if match(itask.state.status, cycle_sel):
                            cycles.append(icycle)
                            break

            # filter by task
            elif lowest_token == Tokens.Task:
                cycle = tokens[Tokens.Cycle.value]
                cycle_sel = tokens.get(Tokens.Cycle.value + '_sel', '*')
                task = tokens[Tokens.Task.value]
                task_sel = tokens.get(Tokens.Task.value + '_sel', '*')
                for icycle, itasks in self.pool.items():
                    str_cycle = str(icycle)
                    if not match(str_cycle, cycle):
                        continue
                    if cycle_sel == '*':
                        cycles.append(str_cycle)
                        continue
                    for itask in itasks:
                        if (
                            match(itask.state.status, cycle_sel)
                            and match(itask.tdef.name, task)
                            and (
                                match(itask.state.status, task_sel)
                                or any(
                                    match(ns, task)
                                    for ns in itask.tdef.namespace_hierarchy
                                )
                            )
                        ):
                            tasks.append(itask)

            else:
                raise NotImplementedError

            if not cycles or tasks or jobs:
                _not_matched.append(tokens)
            else:
                _cycles.extend(cycles)
                _tasks.extend(tasks)
                _jobs.extend(jobs)

        ret = None
        if out == Tokens.Cycle:
            _cycles.extend([
                itask.point
                for itask in _tasks
            ])
            ret = _cycles
        elif out == Tokens.Task:
            _tasks.extend([
                self.pool[icycle]
                for icycle in _cycles
            ])
            ret = _tasks

        return ret, _not_matched
```

hjoliver:
> Especially nice as Cylc Flow would not need to provide any multi-user support,

:+1:

dwsutherland (approved):
Pretty happy with these changes. Only a couple questions:
- I guess the use of `//` & `/` is the best we can do? (slashes seem to be used in many places.. and even the double works in paths.. Can it be overused? or is it the URL safe choice)
- I suppose there's no way to centralise `//` & `/` it can be changed easily? (i.e. f-string substitution in `id.py`)

Apologies if the discussion has already taken place (i.e. [Proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-universal-id.md)) on the above!

PS - If only there was a better way to pattern match selection, but I guess there's no avoiding the looping:
```
                for icycle, itasks in self.pool.items():
                    str_cycle = str(icycle)
                    if not match(str_cycle, cycle):
                        continue
                    if cycle_sel == '*':
                        cycles.append(str_cycle)
                        continue
                    for itask in itasks:
```
at least its only going through each task once I suppose O(n)... (same with non ID lookup resolvers)

oliver-sanders:
Good point about URL safety, will come back to after the beta release.

oliver-sanders:
> I guess the use of // & / is the best we can do?

I think the use of `/` is great (plays into dir structure and CLI auto-completion nicely), the `//` is not so good.

Here's the rationale behind using two characters:

```bash
# chars considered:
/ ! @ % ^ & * ( ) - _ = + [ ] { } / | ; : , . < > ~ #

# conflicts with hierarchical workflow registrations
/
# conflicts with glob chars in the shell and the :sel pattern
[ ] { } * 
# conflicts with the :sel prefix
:
# conficts with shell
; ( ) < > | & #
# conflicts with ~user prefix
~
# conflict with workflow name
/ _ + - . @
# conflict with task name
- + % @
# conflicts with extglob (shell)
!
# partial conflict with shell (substitution?)
^

# chars left:
=,
```

So that just leaves us with the following two char combinations:

```bash
//
/=
=/
/,
,/
```

dwsutherland:
Not `,` by itself?

oliver-sanders:
Possible, but a bit strange:

```
~user/flow=cycle/task/job
~user/flow,cycle/task/job
```

oliver-sanders:
(on draft waiting for https://github.com/cylc/cylc-flow/issues/4230 which will provide the place for this argument parsing to be inserted)

hjoliver:
Note https://github.com/cylc/cylc-flow/issues/4419#issuecomment-999181650

We need to consider non-merging flows and potential task ID implications of that.

oliver-sanders:
Tests all passing (bar the one graphql test I'm re-writing as an integration test).

Final interface change in - https://github.com/cylc/cylc-flow/pull/3931/commits/2702efaf45ca8a544cbc8d11fb67697a7898a75c changed the IDs for namespaces (i.e. task / family names, used for referencing task definitions) and edges. These IDs don't fit into the UID scheme. We were using the UID for these before, however, it's wired and makes tokenising dodge so I've added an explicit prefix of `$namespace` or `$edge`:

```
before:
    workflow|namespace
after:
    workflow//$namespace|namespace

before:
    workflow|task1.cycle1|task2.cycle2
after:
    workflow//$edge|cycle1/task1|cycle2/task2
```

Still working on it but ready for review.

cylc/flow/async_util.py
```diff
@@ -417,3 +417,44 @@ async def asyncqgen(queue):
...
+        pending.append(task)
+
+    # run tasks
+    while pending:
+        done, pending = await asyncio.wait(
+            pending,
+            return_when=asyncio.FIRST_COMPLETED
+        )
+        for task in done:
+            yield task._args, task.result()
```
MetRonnie, line 460: Should this while loop have a `await asyncio.sleep(0)` to avoid blocking the process?

cylc/flow/id.py
```diff
@@ -0,0 +1,664 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Cylc univeral identifier system for referencing Cylc "objets".
```
MetRonnie, line 17:
```suggestion
"""Cylc universal identifier system for referencing Cylc "objects".
```

cylc/flow/async_util.py
oliver-sanders: I think `asyncio.wait` on the results handles this?

oliver-sanders:
Ok, one final change to make tokens a little [type] safer.

Here's the interface for converting between string IDs and Tokens objects:

```python
>>> tokens = Tokens('~u/w//c/t/01')
>>> tokens
<id: ~u/w//c/t/01>
>>> str(tokens)  # or tokens.id
'~u/w//c/t/01'
>>> tokens.workflow_id
~u/w
>>> tokens.relative_id
c/t/01
```

They are still dictionaries under the hood, but all the dict keys get checked against a whitelist to avoid typo mistakes.

cylc/flow/rundb.py
```diff
@@ -706,7 +706,7 @@ def select_task_job_platforms(self):
     def select_submit_nums(self, name, point):
         """Select submit_num and flow_nums from task_states table.
 
-        Fetch submit number and flow_nums for spawning task name.point.
+        Fetch submit number and flow_nums for spawning task point/name.
```
MetRonnie, line 709: This could potentially be confusing
```suggestion
        Fetch submit number and flow_nums for spawning the task
        with the given name and cycle point.
```

oliver-sanders:
UIS branch up, you should now be able to test with GraphiQL - https://github.com/cylc/cylc-uiserver/pull/297

The remainder of the UI will be somewhat broken at the moment, will try to get that branch up shortly.

cylc/flow/id_cli.py
```diff
@@ -0,0 +1,483 @@
...
+                raise UserInputError('IDs cannot be null.')
+        return
+    raise Exception(f'Invalid constraint: {constraint}')
+
+
+def _validate_workflow_ids(*tokens_list, src_path):
+    for ind, tokens in enumerate(tokens_list):
+        if tokens['user']:
+            raise UserInputError(
+                'Operating on others workflows is not supported'
```
MetRonnie, line 360:
```suggestion
                "Operating on other users' workflows is not supported"
```

cylc/flow/id_cli.py
```diff
@@ -0,0 +1,483 @@
...
+def _parse_src_path(id_):
+    src_path = Path(id_)
+    if (
+        id_ == os.curdir
+        or id_.startswith(f'{os.curdir}{os.sep}')
+        or Path(id_).is_absolute()
+    ):
+        src_path = src_path.resolve()
+        if not src_path.exists():
+            raise UserInputError(src_path)
```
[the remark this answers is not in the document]
oliver-sanders:
```suggestion
            raise UserInputError(f'Path does not exist: {src_path}')
```

oliver-sanders:
Dammit, the diff is too big to find suggestions in the file tab.

cylc/flow/network/resolvers.py
```diff
@@ -82,116 +86,184 @@ def collate_workflow_atts(workflow):
...
+    Examples:
+        >>> uniq([1, 1, 2, 3, 5, 8, 1])
+        [1, 2, 3, 5, 8]
+
+    """
+    ret = []
+    for item in iterable:
+        if item not in ret:
+            ret.append(item)
+    return ret
```
MetRonnie, line 111: Although couldn't this be fulfilled using `set(iterable)`?

cylc/flow/scripts/cylc.py
```diff
@@ -69,17 +69,108 @@ def get_version(long=False):
...
 
-  Date-time cycle points are in an ISO 8601 date-time format, typically
-  CCYYMMDDThhmm followed by a time zone - e.g. 20101225T0600Z.
+  You can split an ID at the // so following two IDs are equivalent:
+    workflow//cycle1 workflow//cycle2
+    workflow// //cycle1 //cycle2
 
-  Integer cycle points (including those for non-cycling workflows) are integers
-  - just '1' for non-cycling workflows.
+  IDs to be written as globs:
```
MetRonnie, line 80:
```suggestion
  IDs can be written as globs:
```

cylc/flow/scripts/cylc.py
```diff
@@ -69,17 +69,108 @@ def get_version(long=False):
...
+
+    We will end up with a workflow with the ID "foo/run1".
+
+    This ID can be used to interact with the workflow:
+      $ cylc play foo/run1
+      $ cylc pause foo/run1
+      $ cylc stop foo/run1
+
+    In the case of numbered runs (e.g. "run1", "run2", ...) you can omit
+    the run number:
```
MetRonnie, line 103:
```suggestion
    the run number and Cylc will infer the latest run:
```

cylc/flow/scripts/cylc.py
```diff
@@ -69,17 +69,108 @@ def get_version(long=False):
...
+      # play the workflow with the ID "foo/bar/baz"
+      $ cylc play foo/bar/baz
+
+    The full format of a workflow ID is:
+      ~user/workflow-id
+
+    You can omit the user name when working on your own workflows.
+
+Cycle / Family / Task / Job IDs:
+    Just as workflows have IDs the things within workflows have IDs too.
```
MetRonnie, line 120:
```suggestion
    Just as workflows have IDs, the things within workflows have IDs too.
```

cylc/flow/scripts/pause.py
```diff
@@ -23,19 +23,22 @@
 This prevents submission of any task jobs.
 
 Examples:
+  # pause my_flow
   $ cylc pause my_flow
 
-To resume a paused workflow, use 'cylc play'.
+  # un-pause my_flow
```
MetRonnie, line 29: resume?

cylc/flow/scripts/set_verbosity.py
```diff
@@ -52,30 +54,36 @@ def get_option_parser():
     parser = COP(
         __doc__, comms=True,
         argdoc=[
-            ('WORKFLOW', 'Workflow name or ID'),
-            ('LEVEL', ', '.join(LOG_LEVELS.keys()))
+            ('LEVEL', ', '.join(LOG_LEVELS.keys())),
+            ('ID [ID ...]', 'Workflow ID(s)'),
```
MetRonnie, line 58: Are we sure about reversing the order of these args?

oliver-sanders: Yep, workflow args must always come last (because we don't know how many of them there will be).

tests/unit/test_id_cli.py
```diff
@@ -0,0 +1,466 @@
...
+            raise
+    else:
+        if errors:
+            raise Exception('Should have raised UserInputError')
+
+
+async def test_parse_ids_infer_run_name(tmp_run_dir):
+    """It should infer the run name for auto-numbered installations."""
+    # it doesn't do anything for a named run
+    tmp_run_dir('foo', named=True)
```
MetRonnie: In the `tmp_run_dir` fixture, the `named` arg doesn't do anything unless `installed=True` too. E.g. `tmp_run_dir('foo/bar', installed=True, named=True)`. But in this case I think a simple
```suggestion
    tmp_run_dir('foo/bar')
```
is what you are after

tests/unit/test_id_cli.py
```diff
@@ -0,0 +1,466 @@
...
+    # valid relative path
+    workflow_id, src_path, src_file_path = _parse_src_path('./a')
+    assert workflow_id == 'a'
+    assert src_path == src_dir
+    assert src_file_path == src_dir / 'flow.cylc'
+
+    # broken relative path
+    with pytest.raises(WorkflowFilesError):
+        _parse_src_path('.')
+
```
MetRonnie: Same as below
```suggestion
```

tests/unit/test_id_cli.py
```diff
@@ -0,0 +1,466 @@
...
+    assert src_file_path == src_dir / 'flow.cylc'
+
+    # relative './<flow-file>'
+    workflow_id, src_path, src_file_path = _parse_src_path('./flow.cylc')
+    assert workflow_id == 'a'
+    assert src_path == src_dir
+    assert src_file_path == src_dir / 'flow.cylc'
+
+
+async def test_parse_ids_src_path(src_dir):
```
MetRonnie: Async tests are getting skipped on GH Actions: https://github.com/cylc/cylc-flow/runs/4741547845?check_suite_focus=true#step:11:2594
```
/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/_pytest/python.py:172:
  PytestUnhandledCoroutineWarning: async def functions are not natively supported
  and have been skipped.
  You need to install a suitable plugin for your async framework, for example:
    - anyio
    - pytest-asyncio
    - pytest-tornasync
    - pytest-trio
    - pytest-twisted
    warnings.warn(PytestUnhandledCoroutineWarning(msg.format(nodeid)))
```

cylc/flow/scripts/broadcast.py
```diff
@@ -283,8 +288,10 @@ def get_option_parser():
 
     parser.add_option(
         "-k", "--display-task", metavar="TASK_GLOB",
```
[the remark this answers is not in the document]
oliver-sanders: Good spot, I will have a glob and go through for all these.

cylc/flow/task_events_mgr.py
```diff
@@ -1046,6 +1049,7 @@ def _process_message_submitted(self, itask, event_time, submit_num):
             itask.state.outputs.set_completion(TASK_OUTPUT_STARTED, True)
             self.data_store_mgr.delta_task_output(itask, TASK_OUTPUT_STARTED)
         else:
+            itask.set_summary_time('submitted', event_time)
```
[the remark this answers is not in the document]
oliver-sanders: Not sure, removing and re-testing.

tests/unit/test_id_cli.py
```diff
@@ -0,0 +1,466 @@
...
+    with pytest.raises(WorkflowFilesError):
+        _parse_src_path('.')
+
+    # relative '.' (invalid)
+    with pytest.raises(WorkflowFilesError) as exc_ctx:
+        workflow_id, src_path, src_file_path = _parse_src_path('.')
+    assert 'No flow.cylc or suite.rc in .' in str(exc_ctx.value)
+
+    # move into the src dir
+    os.chdir(src_dir)
```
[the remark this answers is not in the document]
oliver-sanders: Good to know, thanks.

tests/unit/test_id_cli.py
oliver-sanders: Hmmm.

I used to put an explicit `@pytest.mark.asyncio` onto these functions which got them running but I recently found that they ran even without that marker.

I've gotta strange feeling that it's because I've got `pytest-tornasync` installed in my environment (due to UIS development), from poking at the Jupyter Server tests I noticed that `pytest-tornasync` seems to enable itself by default so I guess I'm probably running unmarked async tests through `pytest-tornasync` (which is `asyncio` under the hood)?

Dunno, I guess I will go through and add explicit `pytest.mark.asyncio` lines?

oliver-sanders: There might be a `pytest-asyncio` config...

oliver-sanders: ...Yes but it won't be available until the next release...

cylc/flow/network/resolvers.py
oliver-sanders: That wouldn't preserve order.

tests/unit/test_id_cli.py
```diff
@@ -0,0 +1,528 @@
...
+async def test_parse_ids_tasks(ids_in, ids_out):
+    """It should parse workflow tasks in two formats."""
+    workflows, _ = await parse_ids_async(*ids_in, constraint='tasks')
+    assert {
+        workflow_id: [detokenise(tokens) for tokens in tokens_list]
+        for workflow_id, tokens_list in workflows.items()
+    } == ids_out
+
+
+@pytest.mark.parametrize(
```
MetRonnie: 1 slipped through the net
```suggestion
@pytest.mark.asyncio
@pytest.mark.parametrize(
```

tests/unit/test_id_match.py
```diff
@@ -0,0 +1,297 @@
...
+            1: ['1/a:x'],
+        },
+        {}
+    )
+
+    _matched, _not_matched = filter_ids(
+        [pool],
+        ['1/a'],
+        out=IDTokens.Task,
+        pattern_match=True,
```
MetRonnie: Seems wrong given the name + description of the test?
```suggestion
        pattern_match=False,
```

tests/integration/test_task_pool.py
```diff
@@ -272,37 +268,31 @@ async def test_match_taskdefs(
...
-            ["No active tasks in the family 'FAM' matching: FAM.2"],
+            ['1/FAM', '2/FAM'], ['1/bar'],
+            ["No active tasks in the family 'FAM' matching: 2/FAM"],
             id="Family names hold active tasks only"
         ),
         param(
-            ['foo.*', 'bar', 'pub', 'grogu.*'], ['foo.1', 'bar.1'],
-            ["No active instances of task: pub",
-             "No active tasks matching: grogu.*"],
-            id="Point globs/point omitted hold active tasks only"
```
MetRonnie: Is it no longer possible to hold all instances of a task across cycle points?

MetRonnie (commented):
I've now reviewed 166/731 files, I think that'll do it 😅 (Not going to look at all the functional test files that changed)

tests/integration/test_task_pool.py
oliver-sanders: Can still hold all "active" instances of a task across all cycle points:

```console
$ cylc hold 'workflow//*/task'
```

MetRonnie: Then can the test be reimplemented as
```
        param(
            ['//*/foo', '*/bar', '*/pub', '//*/grogu'], ['1/foo', '1/bar'],
            ["No active tasks matching: */pub",
             "No active tasks matching: //*/grogu"],
            id="Point globs hold active tasks only"
        ),
```

oliver-sanders:
Small regex change in https://github.com/cylc/cylc-flow/pull/3931/commits/efc2cc6f73497bac56d76d01139d58f3fa44c4a7

tldr; I am using the regex from cylc-flow in the cylc-ui but browser support for look behinds isn't up to scratch yet.

cylc/flow/task_events_mgr.py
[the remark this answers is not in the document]
oliver-sanders: Seems fine, I think this was the result of frantic debugging, good catch.

tests/unit/test_id_cli.py
oliver-sanders: Not quite, I wanted to ensure it treats a named run the same way as a hierarchical workflow.

tests/unit/test_id_cli.py
oliver-sanders: Made it do something different.

tests/unit/test_id_match.py
oliver-sanders: Fixed.

cylc/flow/scripts/broadcast.py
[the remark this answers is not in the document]
oliver-sanders: Not resolving this as I've not yet done this everywhere.

tests/unit/test_id_cli.py
oliver-sanders: Changed.

tests/integration/test_task_pool.py
oliver-sanders: You can, tested in id_match.

dwsutherland (changes requested):
Looks good to me, and got it running. (reviewed up to tests, then will let GHActions do the rest)

I guess the only thing is the `~` character that get's expanded in shell:
```
(uid-flow) sutherlander@graphic-vbox:~$ cylc clean ~sutherlander/quz
  File "/home/sutherlander/.envs/uid-flow/bin/cylc", line 33, in <module>
    sys.exit(load_entry_point('cylc-flow', 'console_scripts', 'cylc')())
  File "/home/sutherlander/cylc/cylc-flow/cylc/flow/scripts/cylc.py", line 637, in main
.
.
.
  File "/home/sutherlander/cylc/cylc-flow/cylc/flow/id.py", line 98, in __init__
    kwargs = tokenise(
  File "/home/sutherlander/cylc/cylc-flow/cylc/flow/id.py", line 660, in tokenise
    raise ValueError(f'Invalid Cylc identifier: {identifier}')
ValueError: Invalid Cylc identifier: /home/sutherlander/quz
```
```
(uid-flow) sutherlander@graphic-vbox:~$ cylc clean '~sutherlander/quz'
UserInputError: Operating on others users' workflows is not supported
```
so not sure it'll work (mentioned it in a comment below)

cylc/flow/scripts/broadcast.py
```diff
@@ -331,20 +342,21 @@ def main(_, options: 'Values', workflow: str) -> None:
...
             try:
-                task, point = TaskID.split(options.showtask)
-                query_kwargs['variables']['nIds'] = [
-                    f'{point}{ID_DELIM}{task}']
+                query_kwargs['variables']['nIds'] = [options.showtask]
             except ValueError:
-                raise UserInputError("TASKID must be " + TaskID.SYNTAX)
-        result = pclient('graphql', query_kwargs)
+                # TODO validate showtask?
+                raise UserInputError("TASKID must be cycle/taask")
```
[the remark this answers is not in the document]
MetRonnie: This is fixed in a recent commit

cylc/flow/network/resolvers.py
```diff
@@ -82,116 +86,184 @@ def collate_workflow_atts(workflow):
...
-        (not args.get('states') or n_atts[5] in args['states']) and
-        not (args.get('exstates') and n_atts[5] in args['exstates']) and
-        (args.get('is_held') is None
-         or (node.is_held == args['is_held'])) and
-        (args.get('is_queued') is None
-         or (node.is_queued == args['is_queued'])) and
-        (args.get('mindepth', -1) < 0 or node.depth >= args['mindepth']) and
-        (args.get('maxdepth', -1) < 0 or node.depth <= args['maxdepth']) and
+        (
+            args.get('ghosts')
```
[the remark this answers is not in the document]
dwsutherland: I have a related PR, will do in there.

cylc/flow/broadcast_report.py
```diff
@@ -20,7 +20,7 @@
 BAD_OPTIONS_TITLE = "No broadcast to cancel/clear for these options:"
 BAD_OPTIONS_TITLE_SET = ("Rejected broadcast: settings are not"
                          " compatible with the workflow")
-CHANGE_FMT = "\n%(change)s [%(namespace)s.%(point)s] %(key)s=%(value)s"
+CHANGE_FMT = "\n%(change)s [%(point)s/%(namespace)s] %(key)s=%(value)s"
```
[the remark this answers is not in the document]
oliver-sanders: Ooh dammit, good spot.

MetRonnie:
Doing some testing: unfortunately the trailing slash that bash tab key autocomplete gives you causes this traceback:
```console
$ cylc play temp4/
  ...
  File "~/github/cylc-flow/cylc/flow/scheduler_cli.py", line 285, in scheduler_cli
    (workflow_id,), _ = parse_ids(
  File "~/github/cylc-flow/cylc/flow/id_cli.py", line 178, in parse_ids
    return asyncio.run(parse_ids_async(*args, **kwargs))
  ...
  File "~/github/cylc-flow/cylc/flow/id_cli.py", line 248, in parse_ids_async
    tokens_list.extend(_parse_cli(*ids))
  File "~/github/cylc-flow/cylc/flow/id_cli.py", line 125, in _parse_cli
    tokens = Tokens(id_)
  File "~/github/cylc-flow/cylc/flow/id.py", line 98, in __init__
    kwargs = tokenise(
  File "~/github/cylc-flow/cylc/flow/id.py", line 660, in tokenise
    raise ValueError(f'Invalid Cylc identifier: {identifier}')
ValueError: Invalid Cylc identifier: datetime-cycling/
```

cylc/flow/id.py
```diff
@@ -0,0 +1,854 @@
...
+
+
+class Tokens(dict):
+    """A parsed representation of a Cylc universal identifier (UID).
+
+    Examples:
+        Parse tokens:
+        >>> tokens = Tokens('~u/w//c/t/01')
+        >>> tokens
+        <id: ~u/w//c/t/01>
```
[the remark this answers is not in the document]
oliver-sanders: Yes, also the glob characters like '*' might get expanded too.

Users will need to quote arguments that contain these characters e.g:

```
cylc stop '~other_user/workflow'
```

MetRonnie:
> The bash autocompletion in the Cylc repo is out of date

Sorry I meant the non-Cylc bash autocompletion (filenames and dirs in PWD)

oliver-sanders:
Ah I see, yes a (single) trailing slash after a workflow name is an error.

cylc/flow/id.py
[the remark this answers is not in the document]
dwsutherland: I assumed that didn't work either..
```
(uid-flow) sutherlander@graphic-vbox:~$ cylc clean "~sutherlander/quz"
UserInputError: Operating on others users' workflows is not supported
(uid-flow) sutherlander@graphic-vbox:~$ cylc clean '~sutherlander/quz'
UserInputError: Operating on others users' workflows is not supported
```
as I thought the user might have been recognised as `~sutherlander` not `sutherlander` (which is the same user)

I suppose we've bashed out any shell safe alternatives in the proposal..

oliver-sanders:
Could potentially strip (single) trailing '/' characters from IDs before they are passed through the tokeniser to handle the issue of people using native-shell auto-complete whilst located in the cylc-run directory.

Here's my unfinished Bash auto-completion script which fills in workflows, cycles, tasks and jobs (will raise in a future PR):

```bash
#!/usr/bin/env bash

_cylc_complete() {
    if [[ "${#COMP_WORDS[@]}" == 2 ]]; then
        _cylc_cmd
    elif [[ "${#COMP_WORDS[@]}" -gt 2 ]]; then
        _cylc_uid
    fi
    return
}

_cylc_cmd() {
    COMPREPLY=(play pause stop trigger)
}

_join_by() { local d=${1-} f=${2-}; if shift 2; then printf %s "$f" "${@/#/$d}"; fi; }

_cylc_uid() {
    local id=${COMP_WORDS[-1]}
    if [[ ${COMP_WORDS[-1]: -1} == '/' ]]; then
        local lock=true
    else
        local lock=false
    fi

    readarray -d '/' -t parts <<< "${id}"
    is_relative=false
    workflow_tokens=()
    relative_tokens=()
    for part in "${parts[@]}"; do
        if ! $is_relative; then
            if [[ $part == '' ]]; then
                is_relative=true
            else
                workflow_tokens+=("$part")
            fi
        elif [[ $part != '' ]]; then
            relative_tokens+=("$part")
        fi
    done

    user=''
    workflow=''
    cycle="${relative_tokens[0]:-}"
    task="${relative_tokens[1]:-}"
    job="${relative_tokens[2]:-}"

    if [[ "${id:0:1}" == '~' ]]; then
        # starts: ~user
        # user="${workflow_tokens[0]:1:-1}"
        user="${workflow_tokens[0]}"
        workflow="$(_join_by '/' "${workflow_tokens[@]:1:100}")"
    else
        workflow="$(_join_by '/' "${workflow_tokens[@]}")"
    fi

    user="${user/$'\n'/}"
    workflow="${workflow/$'\n'/}"
    cycle="${cycle/$'\n'/}"
    task="${task/$'\n'/}"
    job="${job/$'\n'/}"

    if ! [[ "$user" == '' || "$user" == "$USER" ]]; then
        # we don't support this yet
        return
    fi

    if $lock; then
        if [[ $job ]]; then
            true
        elif [[ $task ]]; then
            _cylc_job "$user" "$workflow" "$cycle" "$task"
        elif [[ $cycle ]]; then
            _cylc_task "$user" "$workflow" "$cycle"
        elif [[ $workflow ]]; then
            _cylc_cycle "$user" "$workflow"
        else
            _cylc_workflow "$user"
        fi
    else
        if [[ $job ]]; then
            _cylc_job "$user" "$workflow" "$cycle" "$task"
        elif [[ $task ]]; then
            _cylc_task "$user" "$workflow" "$cycle"
        elif [[ $cycle ]]; then
            _cylc_cycle "$user" "$workflow"
        else
            _cylc_workflow "$user"
        fi
    fi

}

detokenise () {
    user="${1:-}"
    workflow="${2:-}"
    cycle="${3:-}"
    task="${4:-}"
    job="${5:-}"
    ret=''
    if [[ $user != '' ]]; then
        ret="~$user"
    fi
    if [[ $workflow != '' ]]; then
        ret+="$workflow//"
    fi
    if [[ $cycle != '' ]]; then
        ret+="$cycle"
        if [[ $task != '' ]]; then
            ret+="/$task"
            if [[ $job != '' ]]; then
                ret+="/$job"
            fi
        fi
    fi
    echo "$ret"
}

_cylc_user() {
    true
}

_cylc_workflow() {
    user="$1"
    workflow="$2"
    words=()
    for _workflow in $(cylc scan --states=all -t name --color=never); do
        words+=("$(detokenise "$user" "$_workflow")")
    done
    COMPREPLY=($(compgen -W "$(echo "${words[@]}")" "${COMP_WORDS[-1]}"))
}

_cylc_cycle() {
    user="$1"
    workfow="$2"
    words=()
    for _cycle in $(ls "$HOME/cylc-run/${workflow}/log/job/"); do
        words+=("$(detokenise "$user" "$workflow" "$_cycle")")
    done
    COMPREPLY=($(compgen -W "$(echo "${words[@]}")" "${COMP_WORDS[-1]}"))
}

_cylc_task() {
    user="$1"
    workflow="$2"
    cycle="$3"
    for _task in $(ls "$HOME/cylc-run/${workflow}/log/job/${cycle}"); do
        words+=("$(detokenise "$user" "$workflow" "$cycle" "$_task")")
    done
    COMPREPLY=($(compgen -W "$(echo "${words[@]}")" "${COMP_WORDS[-1]}"))
}

_cylc_job() {
    user="$1"
    workflow="$2"
    cycle="$3"
    task="$4"
    for _job in $(ls "$HOME/cylc-run/${workflow}/log/job/${cycle}/${task}"); do
        words+=("$(detokenise "$user" "$workflow" "$cycle" "$task" "$_job")")
    done
    COMPREPLY=($(compgen -W "$(echo "${words[@]}")" "${COMP_WORDS[-1]}"))
}

complete -F _cylc_complete cylc
```

cylc/flow/id.py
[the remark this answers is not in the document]
dwsutherland: I guess it's a little redundant; why won't it recognize I'm the same user specified in the above?
Does it just, when someone puts their user in, spit out:
```
UserInputError: Operating on others users' workflows is not supported
```
or is there an issue here?
(yes, I know, why would you use a `~user` form of ID while operating as that user)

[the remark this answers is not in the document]
dwsutherland: Ah, found it.. So the `parse_ids_async` in `id_cli.py` blocks any form of user (even owner) at present:
```
def _validate_workflow_ids(*tokens_list, src_path):
    for ind, tokens in enumerate(tokens_list):
        if tokens['user']:
            raise UserInputError(
                "Operating on others users' workflows is not supported"
            )
```
This is fine, we can modify going forward... Let's just get your work in, as it's clearly the way forward.

dwsutherland (approved):
As long as test are passing, this should go in:
- Code looks good on pass.
- Works along with sibling PRs end-to-end.
- Nothing fundamentally wrong (from what I can see), so further changes can be split off to more PRs.

At present, the more work I do the more the rebase will hurt!

cylc/flow/id_cli.py
```diff
@@ -0,0 +1,482 @@
...
+            workflows - only allow workflows.
+            tasks - require tasks to be defined.
+            mixed - permit tasks not to be defined.
+        max_workflows:
+            Specify the maximum number of workflows permitted to be specified
+            in the ids.
+        max_tasks:
+            Specify the maximum number of tasks permitted to be specified
+            in the ids.
+
```
wxtim, line 224: IMO a "returns" section would be useful.

wxtim:
I have
- [x] Run the Cylc Rose test battery with this branch checked out:
  - One issue discovered: https://github.com/cylc/cylc-rose/pull/108
- [ ] Run the Rose test battery with this branch checked out.
- [ ] Had a play with this branch running workflows, doing things with workflows.
- [ ] Had a quick run through the tutorials using this branch.

MetRonnie (approved):
Bash filename autocompletion (trailing slash) can be tackled in follow-up PR. Otherwise happy.

`tests/f/clock-trigger-inexact/00-big-offset.t` failing because it is new on master but doesn't exist on this branch. I'm guessing can fix by merging master into the branch and updating to UID

wxtim:
When I use Cylc help and press pagedown:

```bash
> cylc help

                Cylc Workflow Engine 8.0rc1.dev
                Copyright (C) 2008-2021 NIWA

Cylc ("silk") orchestrates complex cycling (and non-cycling) workflows.

Version:
  $ cylc version --long           # print cylc-flow version and install path
  8.0rc1.dev (/home/h02/tpilling/miniconda3/envs/cylc8)

Usage:
  $ cylc help all                 # list all commands
  $ cylc validate FLOW            # validate a workflow configuration
  $ cylc play FLOW                # run/resume a workflow
  $ cylc scan                     # list all running workflows (by default)
  $ cylc tui FLOW                 # view a running workflow in the terminal
  $ cylc stop FLOW                # stop a running workflow

Command Abbreviation:
  # Commands can be abbreviated as long as there is no ambiguity in
  # the abbreviated command:
  $ cylc trigger WORKFLOW TASK    # trigger TASK in WORKFLOW
  $ cylc trig WORKFLOW TASK       # ditto
  $ cylc tr WORKFLOW TASK         # ditto
  $ cylc t                        # Error: ambiguous command

Cylc IDs:
  Cylc IDs take the form:
    workflow//cycle/task

  You can split an ID at the // so following two IDs are equivalent:
    workflow//cycle1 workflow//cycle2
    workflow// //cycle1 //cycle2

  IDs can be written as globs:
    *//                 # All workflows
    workflow//*         # All cycles in "workflow"
    workflow//cycle/*   # All tasks in "workflow" in "cycle"

  For more information type "cylc help id".

Selected Sub-Commands:
^[[6~Traceback (most recent call last):
  File "/home/h02/tpilling/miniconda3/envs/cylc8/bin/cylc", line 33, in <module>
    sys.exit(load_entry_point('cylc-flow', 'console_scripts', 'cylc')())
  File "/net/home/h02/tpilling/metomi/cylc.git/cylc/flow/scripts/cylc.py", line 605, in main
    cli_help()
  File "/net/home/h02/tpilling/metomi/cylc.git/cylc/flow/scripts/cylc.py", line 401, in cli_help
    print_command_list(
  File "/net/home/h02/tpilling/metomi/cylc.git/cylc/flow/scripts/cylc.py", line 383, in print_command_list
    contents = [
  File "/net/home/h02/tpilling/metomi/cylc.git/cylc/flow/scripts/cylc.py", line 383, in <listcomp>
    contents = [
  File "/net/home/h02/tpilling/metomi/cylc.git/cylc/flow/scripts/cylc.py", line 361, in iter_commands
    module = __import__(obj.module_name, fromlist=[''])
  File "/net/home/h02/tpilling/metomi/cylc-uiserver/cylc/uiserver/__init__.py", line 21, in <module>
    from cylc.uiserver.app import CylcUIServer
  File "/net/home/h02/tpilling/metomi/cylc-uiserver/cylc/uiserver/app.py", line 46, in <module>
    from cylc.uiserver.authorise import (
  File "/net/home/h02/tpilling/metomi/cylc-uiserver/cylc/uiserver/authorise.py", line 28, in <module>
    from cylc.uiserver.schema import UISMutations
  File "/net/home/h02/tpilling/metomi/cylc-uiserver/cylc/uiserver/schema.py", line 34, in <module>
    from cylc.flow.network.schema import (
ImportError: cannot import name 'parse_workflow_id' from 'cylc.flow.network.schema' (/net/home/h02/tpilling/metomi/cylc.git/cylc/flow/network/schema.py)

```

oliver-sanders:
> If this is replicable it's undesirable

That's a broken import in cylc-uiserver which is fixed in https://github.com/cylc/cylc-uiserver/pull/297

dwsutherland:
> Yikes, I merged @wxtim's log-message tweak branch, and now we have 300 conflicting files here 😱

Another reason this should go in ASAP 😅

hjoliver:
> > Yikes, I merged @wxtim's log-message tweak branch, and now we have 300 conflicting files here scream
> 
> Another reason this should go in ASAP sweat_smile

Yeah, currently attempting to fix the conflicts myself...

hjoliver:
@oliver-sanders - I merged from master and deconflicted, and pushed a couple of commits to your branch.  Looks like all the tests will pass, except for test-tutorial-workflow.

Are you particularly attached to your ~90-commit history on this branch, or happy to have it squashed?

hjoliver (approved):
Did lots of manual testing. No problems found during quick code sanity check (thanks @MetRonnie and @dwsutherland for doing the real work :grin: )

hjoliver:
Feel free to merge this...

dwsutherland:
> Feel free to merge this...

Will see if Oliver is happy with the squash.. I guess

oliver-sanders:
> Did lots of manual testing. No problems found

Thanks, managed to get the diff coverage up to ~90% which is good but glad to have some heavy manual testing!

> Feel free to merge this...

I still have one comment from Ronnie to address about metavars in the CLI.

There appear to be three forms remaining, my best attempt ATM:

* `WORKFLOW_ID` - `workflow`, `~user/workflow`
* `TASK_ID` - `cycle`, `cycle/task`, `cycle/task/job`
* `ID` - `workflow`, `workflow//cycle`, `workflow//cycle/task`, ...

> Are you particularly attached to your ~90-commit history on this branch

No (checkbox unticked in desc) and we should not merge PRs with messy commits like this (boats the repo and makes it very hard to track down bugs).

I'm going to squash at the end, hadn't done it yet to make review easier.

oliver-sanders:
I'll address the remaining comments, rebase squash and push shortly.

oliver-sanders:
Dammit, I don't think I can rebase because of the merge commit, may have to squash merge.

cylc/flow/data_store_mgr.py
```diff
@@ -395,7 +358,11 @@ class DataStoreMgr:
 
     def __init__(self, schd):
         self.schd = schd
-        self.workflow_id = f'{self.schd.owner}{ID_DELIM}{self.schd.workflow}'
+        self.id_ = Tokens(
+            user=self.schd.owner,
+            workflow=self.schd.workflow,
+        )  # TODO: rename and move to scheduler
```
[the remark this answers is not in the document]
oliver-sanders: A bit of possible nicification to reduce the extent of parsing to and from token IDs.

cylc/flow/id_cli.py
oliver-sanders: Done.

cylc/flow/scripts/broadcast.py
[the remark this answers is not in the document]
oliver-sanders: Now done.

oliver-sanders:
Ok, last feedback comments addressed, tests willing that should do it.

@MetRonnie can you once over the docs changes in this commit - https://github.com/cylc/cylc-flow/pull/3931/commits/76770eababa0002aba053e2f5a7cf52d771746a6 - which standardises the CLI metavars and adds more docs.

hjoliver:
> No (checkbox unticked in desc) and we should not merge PRs with messy commits like this (boats the repo and makes it very hard to track down bugs).

Agreed, that's why I did not merge, and asked - I expected you still wanted to do some history rationalization.

hjoliver:
Mysterious functional test failures disappeared when I reran the tests.

hjoliver:
@oliver-sanders - I'll squash and merge, to a single commit.  It doesn't sound as if you wanted to do more complex history surgery to end up with `n > 1` logical commits, but if you did, we can always back up the truck.

oliver-sanders:
Thanks all for reviewing, that was a horrible one.

PR #3938, opened 2020-11-12
Event driven data-store deltas

### LINKED ISSUE #3746: Event-driven datastore update

### LINKED ISSUE #3921: Add xtriggers to GraphQL schema

I couldn't find an existing issue in Cylc Flow. Related to

- https://github.com/cylc/cylc-ui/issues/331
- https://github.com/cylc/cylc-ui/issues/131

The UI needs information about the external triggers, so that they can be displayed in the components/views.

**Additional context**

I tested the UI with the following workflow

```
[scheduling]
initial cycle point = 2000-01-01

[[xtriggers]]
clock_1 = wall_clock(offset=PT1M):PT20S

[[dependencies]]
[[[P1D]]]
graph = "@clock_1 => foo"

[runtime]
[[foo]]
script = true
```

Which works fine in the tree view. It shows `foo` task, but you have no information about the external trigger. I started with the `wall_clock` as that's the only one I could remember using before.

In GraphQL, with the current schema, the closest I found is the `task.proxies.extras` field. The query

```graphql
subscription {
  workflows (ids: ["kinow|xtrigger1"]) {
    id
    tasks {
      id
      proxies {
        id
        extras
      }
    }
  }
}
```

does return some information about the external trigger of the task

```js
{
  "data": {
    "workflows": [
      {
        "id": "kinow|xtrigger1",
        "tasks": [
          {
            "id": "kinow|xtrigger1|foo",
            "proxies": [
              {
                "id": "kinow|xtrigger1|20030725T0000+1300|foo",
                "extras": {
                  "xtrigger \"clock_1 = wall_clock(offset=PT1M)\"": "satisfied"
                }
              },
              {
                "id": "kinow|xtrigger1|20030724T0000+1300|foo",
                "extras": {
                  "xtrigger \"clock_1 = wall_clock(offset=PT1M)\"": "satisfied"
                }
              },
              {
                "id": "kinow|xtrigger1|20030726T0000+1300|foo",
                "extras": {
                  "xtrigger \"clock_1 = wall_clock(offset=PT1M)\"": "satisfied"
                }
              }
            ]
          }
        ]
      }
    ]
  }
}
```

But I think we might need more information/structure than what we have at the moment, which is just `"xtrigger \"clock_1 = wall_clock(offset=PT1M)\"": "satisfied"`.

**Pull requests welcome!**

### LINKED ISSUE #4030: showing prerequites of waiting tasks
Ref: https://github.com/cylc/cylc-flow/pull/3938#issuecomment-757569873 

`cylc show` gives no prerequisites for waiting tasks, since SoD

The reason is the datastore gets this info from task proxies, and:
 1. partially satisfied waiting task proxies are hidden from the task pool (partially satisfied prerequisites might not be held in task proxy objects in future, anyway)
 2. other waiting tasks have no task proxies at all

It would be easy to get the datastore to look at the hidden partially satisfied tasks (1), but that doesn't solve the bigger problem (2).

The scheduler datastore goes to n=1, and it should be simple to generate prerequisite info for all n=1 tasks in the scheduler datastore (easiest kludge: generate a temporary task proxy and extract the prereqs).

The UIS does not currently have access to this information though, for n>1. (Not even from the DB).

PROPOSAL:  allow only n=1 waiting tasks to be queried for prerequisites etc.
- we hope to allow only n=1 into the future anyway, in which case this should be uncontroversial
- (this might warrant a conceptual distinction between "waiting tasks" (n=1: what comes next) and n>1 tasks that will be scheduled in future but are not waiting to run yet)

(Note for n>1 in the past we can as the DB? - do we store prerequisites for past tasks, or just current active ones...)

dwsutherland:
These changes close #3746, closes #3921, closes #4030
Partially satisfies https://github.com/cylc/cylc-ui/pull/530

At present the data-store deltas are only partially event driven, including:
- Node creation (task/familiy proxies and edges).
- Job creation and deltas.

A few points on current state and intentions:
- Task proxy deltas are not created on event.
- The job pool (`job_pool.py`) is separate from the data-store manager because it is event driven, but after converting the data-store to an event driven approach, this is no longer necessary.

Some updates and these deltas should still be batched for application and publishing (for now) to ensure relational consistency, summary updates are less frequent, and UI/UIS is not overwhelmed. i.e. if you have a family of 100 tasks, and each individual state change constitutes a recalculation of the family state and a publishing of these, then you could end up with hundreds if not thousands of group-state/summary and state deltas to calculate and publish in a few seconds (but batched this would only come to a handful if that).

To do:
- [x] Separate the batched task fields delta into separate event driven deltas:
- - [x] state
- - [x] held
- - [x] outputs (separate to capture individual output changes)
- - [x] prerequisites (separate to capture individual output changes)
- - [x] Clock trigger, external trigger, xtrigger changes (separate for each? just a dump for all in one proxy field ATM)
- - [x] Job addition updates
- [x] Combine/merge the event-driven job pool (`job_pool.py`) into the data-store manager.

Most the new delta event triggers will be from within the `task_pool.py` and `task_events_mgr.py` 

Possible future work:
- Publish the deltas individually
- Create a data-store at the UIS off these published events, which will decouple the n-windows at both ends.

**Requirements check-list**
- [x] No dependency changes.


dwsutherland:
Ok, ready for review..

However, there are a few things that I want to change:

**One**
The delta calls have been spread through most the event/task/job managers. I think it want to change this to pass the `data_store_mgr` object to the `TaskProxy` objects and objects created therein (i.e. state, outputs, prerequisites).. The reasons:
- We may not be catching all the delta events (i.e. I may have missed places where the outputs are reset)
- The way it is now makes maintenance/changes harder.. If other locations for state/output/other changes are added, data-store delta calls might not be added...

**Two**
Some fields of the workflow element should be delta driven:
https://github.com/cylc/cylc-ui/pull/543

hjoliver:
I've been playing with this branch using `cylc tui` - it's great to see the ui responding properly now (particularly to succeeded ) :tada: 

However, I'm seeing this error from the UI Server with Cylc UI:
```
[libprotobuf ERROR google/protobuf/wire_format_lite.cc:577] String field 'PbTaskProxy.outputs' contains invalid UTF-8 data when parsing a protocol buffer. Use the 'bytes' t
ype if you intend to send raw bytes. 
2020-12-21 14:48:53,393 tornado.application ERROR    Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncI
OMainLoop object at 0x7f22128d2a58>>, <Task finished coro=<WorkflowsManager.update() done, defined at /home/oliverh/cylc/cylc-uiserver/cylc/uiserver/workflows_mgr.py:227> e
xception=DecodeError('Error parsing message')>)
  File "/home/oliverh/cylc/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/home/oliverh/cylc/cylc-uiserver/venv/lib/python3.7/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/home/oliverh/cylc/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 255, in update
    await self._connect(wid, flow)
  File "/home/oliverh/cylc/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 202, in _connect
    flow
  File "/home/oliverh/cylc/cylc-uiserver/cylc/uiserver/data_store_mgr.py", line 134, in sync_workflow
    await self.entire_workflow_update(ids=[w_id])
  File "/home/oliverh/cylc/cylc-uiserver/cylc/uiserver/data_store_mgr.py", line 320, in entire_workflow_update
    pb_data.ParseFromString(result)
google.protobuf.message.DecodeError: Error parsing message
[libprotobuf ERROR google/protobuf/wire_format_lite.cc:577] String field 'PbTaskProxy.outputs' contains invalid UTF-8 data when parsing a protocol buffer. Use the 'bytes' t
ype if you intend to send raw bytes. 
```

hjoliver:
Seems to be a genuine error. It still happens after recreating my environments.  It occurs after starting the first flow on the back end.

dwsutherland:
I'll have a look. But, on the surface, it looks like the UI Server hasn't picked up the protobuf definition changes

hjoliver:
I did a manual `pip install .` from cylc-flow with your branch checked out, into cylc-uiserver.

hjoliver:
Also checked the correct `data_messages_pb2.py` is installed in the cylc-uiserver venv.

cylc/flow/network/schema.py
```diff
@@ -562,6 +580,26 @@ def resolve_json_dump(root, info, **args):
...
+        key: field[key]
+        for key in args.get('sort_order', [])
+        if key in field}
+    mapping.update(field)
+    satisfied = args.get('satisfied')
+    elements = sort_elements(
+        [
+            val
+            for val in mapping.values()
+            if satisfied is None or val.satisfied is satisfied
```
hjoliver, line 595: So `satisfied` is always `None` or `True` or `False`?

dwsutherland: That's determined by us.. Argument not used, then it's None, otherwise, the user has to set it to True or False (we wan't unset to be distinguishable from True/False, as None will mean include both in the result)

hjoliver: What I was getting at, if it can be anything but those 3 values, you have to use `val.satisfied == satisfied` (not `is`).

dwsutherland:
Working for me:
image

However, separate topic, did the "job runner" cylc-ui end go in?
image

hjoliver:
> However, separate topic, did the "job runner" cylc-ui end go in? 

No but Bruno approved it already (on holiday!) and it's a simple change, so I'll merge it now.

hjoliver:
(cylc-ui change merged)

cylc/flow/network/schema.py
hjoliver: Just to check myself:
```python
>>> foo = "true"
>>> bar = "".join(["t", "r", "u", "e"])
>>> foo == bar
True
>>> foo is bar
False
>>> 
```

dwsutherland: The line about it fetches the argument with `get`
```
   satisfied = args.get('satisfied')
```
If it's not set then `get` will return `None`, and if it is set then GraphQL will fail validation on the Query/Subscription as the field is defined as Boolean.

So it can only be those 3.
And `val.satisfied` is defined as Boolean in the Protobuf definition, so we are covered at both ends.

hjoliver:
Update on this problem for others: `google.protobuf.message.DecodeError: Error parsing message`:

*double check that you've done `pip install .` from this branch of cylc-flow into cylc-uiserver.*

cylc/flow/data_store_mgr.py
```diff
@@ -1117,24 +1216,32 @@ def _family_ascent_point_update(self, fp_id):
             tp_data = self.data[self.workflow_id][TASK_PROXIES]
             tp_updated = self.updated[TASK_PROXIES]
             tp_added = self.added[TASK_PROXIES]
-            # gather child states for count and set is_held
+            # gather child familiy states for count and set is_held
```
hjoliver, line 1219:
```suggestion
            # gather child family states for count and set is_held
```

hjoliver:
(branch conflicted now)

cylc/flow/data_store_mgr.py
dwsutherland: fixed.

dwsutherland:
> (branch conflicted now)

deconflicted

hjoliver:
The batching of deltas is just controlled by the main loop - is that right?

dwsutherland:
> The batching of deltas is just controlled by the main loop - is that right?

Yes, if changed, each main loop it will apply and publish the deltas.

hjoliver:
> The delta calls have been spread through most the event/task/job managers. ... Thoughts?

My feeling is leave it as-is for now. This works, and we could consider moving them into the task proxy object later. However, it may not be worth bothering because we want to move to a proper event-loop driven system before too long: objects raise events (with associated data) and a central event loop processes all events, dispatching them to systems (like the datastore) that have registered interest in them. (Something like that).

hjoliver (approved):
LGTM :+1:, no problems found, so approving this before I bail out for the xmas break. Thanks @dwsutherland :bouquet:

tests/integration/test_data_store_mgr.py
```diff
@@ -78,81 +159,138 @@ def test_increment_graph_window(harness):
...
+    for itask in schd.pool.get_all_tasks():
+        itask.state.set_prerequisites_not_satisfied()
+        schd.data_store_mgr.delta_task_prerequisite(itask)
+    assert not any({
+        p.satisfied
+        for t in schd.data_store_mgr.updated[TASK_PROXIES].values()
+        for p in t.prerequisites})
 
 
 @pytest.mark.skip('TODO: fix this test')
```
oliver-sanders, line 296: Is this test still relevant?

tests/integration/test_data_store_mgr.py
```diff
@@ -13,22 +13,103 @@
...
+
+
+@pytest.fixture
+@pytest.mark.asyncio
+async def myflow(flow, scheduler):
+    schd = scheduler(reg)
+    await schd.install()
+    await schd.initialise()
+    await schd.configure()
+    return schd
```
oliver-sanders, line 84: Unused?

tests/integration/test_data_store_mgr.py
```diff
@@ -78,81 +159,138 @@ def test_increment_graph_window(harness):
...
+    assert schd.data_store_mgr.updated[JOBS].get('j_id') is None
+    schd.data_store_mgr.delta_job_msg(job_d, 'The Atomic Age')
+    assert schd.data_store_mgr.updated[JOBS][j_id].messages
+
+
+def test_delta_job_attr(harness):
+    """Test method modifying job fields to job element."""
+    schd, data = harness
+    schd.data_store_mgr.delta_job_attr(
+        int_id(schd), 'exit_script', 'rm -v *')
```
oliver-sanders, line 235: 😧😨😓😈💀

tests/integration/test_data_store_mgr.py
```diff
@@ -13,22 +13,103 @@
...
+            'graph': {
+                'R1': 'foo => bar'
+            }
+        }
+    }
+    reg = mod_flow(flow_def)
     schd = mod_scheduler(reg)
     async with mod_run(schd):
+        # Think this is needed to save the data state at first start (?)
+        # Fails without it.. and a test needs to overwrite schd data with this.
```
oliver-sanders, line 112: Yeah, because this is a module-scoped fixture it is the same scheduler object provided to each test. When you do a `TaskProxy.reset` in one test that is messing with the object before it is passed to the next test.

You could potentially bolt a clear method onto the scheduler object which does something like `TaskProxy.reset(TASK_STATUS_WAITING, is_held=False)` to wipe the slate called something like `schd._test_clear()`.

Under this framework tests in a module are run in order, however, tests can be run manually one-at-a-time and have the potential to behave differently.

cylc/flow/scripts/show.py
```diff
@@ -258,12 +276,38 @@ def main(_, options, suite, *task_args):
...
+                            t_proxy['clockTrigger']['timeString']
+                            or t_proxy['externalTriggers']
+                            or t_proxy['xtriggers']
+                    ):
+                        if t_proxy['clockTrigger']['timeString']:
+                            state = t_proxy['clockTrigger']['satisfied']
+                            print(
+                                f'  o  Clock trigger time reached ... {state}')
+                            time_str = t_proxy['clockTrigger']['timeString']
+                            print('\nother:')
```
oliver-sanders, line 292:
```suggestion
                            ansiprint('\n<bold>other</bold>')
```

cylc/flow/scripts/show.py
```diff
@@ -258,12 +276,38 @@ def main(_, options, suite, *task_args):
...
+                        print_msg_state(info, output['satisfied'])
+                    if (
+                            t_proxy['clockTrigger']['timeString']
+                            or t_proxy['externalTriggers']
+                            or t_proxy['xtriggers']
+                    ):
+                        if t_proxy['clockTrigger']['timeString']:
+                            state = t_proxy['clockTrigger']['satisfied']
+                            print(
+                                f'  o  Clock trigger time reached ... {state}')
```
oliver-sanders, line 290: There is a `print_msg_state` which standardises message format as well as putting them in colour.

```suggestion
                            print_msg_state(f'  o  Clock trigger time reached', state)
```

cylc/flow/scripts/show.py
```diff
@@ -258,12 +276,38 @@ def main(_, options, suite, *task_args):
...
+                            time_str = t_proxy['clockTrigger']['timeString']
+                            print('\nother:')
+                            print(f'  o  Triggers at ... {time_str}')
+                        else:
+                            print('\nother:')
```
oliver-sanders, line 295:
```suggestion
                            ansiprint('\n<bold>other</bold>')
```

cylc/flow/scripts/show.py
```diff
@@ -258,12 +276,38 @@ def main(_, options, suite, *task_args):
...
+                        for ext_trig in t_proxy['externalTriggers']:
+                            if ext_trig['satisfied']:
+                                msg = 'satisfied'
+                            else:
+                                msg = 'NOT satisfied'
+                            info = f'{ext_trig["label"]}'
+                            print('  o  %s ... %s' % (info, msg))
```
oliver-sanders, line 302:
```suggestion
                            print_msg_state(ext_trig["label"], ext_trig['label'])
```

cylc/flow/scripts/show.py
```diff
@@ -258,12 +276,38 @@ def main(_, options, suite, *task_args):
...
+                        for xtrig in t_proxy['xtriggers']:
+                            if xtrig['satisfied']:
+                                msg = 'satisfied'
+                            else:
+                                msg = 'NOT satisfied'
+                            info = (
+                                f'xtrigger "{xtrig["label"]} = {xtrig["id"]}"')
+                            print('  o  %s ... %s' % (info, msg))
```
oliver-sanders, line 310:
```suggestion
                            print_msg_state(f'xtrigger "{xtrig["label"]}"', xtrig['satisfied'])
```

oliver-sanders (commented):
Looks good, runs fine, happy to have Tui showing succeeded tasks again.

oliver-sanders:
Prerequisites only seem to be listed for active tasks, presumably a SoD consequence, will need to find a way to generate the output for `N>0` tasks as at the moment `cylc show` incorrectly shows a waiting task as having no prerequisites, should bump to another issue for another day.

oliver-sanders:
Does this close #3921?

kinow (approved):
UI working fine with the changes, great work!

@dwsutherland, I noticed while playing with the new code in GraphiQL that I was getting empty deltas, and empty task proxies too. Should them have been stripped/removed from the response instead? Easy to filter in the UI, but looks unnecessary to send `{}`. And you can see that the number of `{}`'s corresponds to the number of task proxies (that's `five`, the same `bar`/`foo` task proxies it's always displaying)

image

No other issues found 👍 I haven't tested it with https://github.com/cylc/cylc-ui/pull/530, as it's addressing that issue partially. @oliver-sanders did a great review. I had a cursory look at the code, and found no other issues +1 pending only @dwsutherland to address @oliver-sanders ' review.

hjoliver:
> Prerequisites only seem to be listed for active tasks, presumably a SoD consequence, will need to find a way to generate the output for `N>0` tasks as at the moment `cylc show` incorrectly shows a waiting task as having no prerequisites, should bump to another issue for another day.

Correct, although I'd point out that the old system only had prerequisites for waiting tasks that happened to have corresponding task proxies spawned ahead already.

cylc/flow/data_store_mgr.py
```diff
@@ -1182,81 +1282,391 @@ def set_graph_window_extent(self, n_edge_distance):
...
+                     if n.name == 'root']
+            ):
+                root_node = self.updated[FAMILY_PROXIES].get(
+                    root_id, data.get(root_id))
+                if root_node is not None and root_node.state:
+                    is_held_total += root_node.is_held_total
+                    state_counter += Counter(dict(root_node.state_totals))
+            w_delta.states[:] = state_counter.keys()
+            for state, state_cnt in state_counter.items():
+                w_delta.state_totals[state] = state_cnt
```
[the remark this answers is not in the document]
dwsutherland: Yes.  Some of the protobuf merging is cumulative and some replace.. In this case the mapping is the latter (for `str: int` anyway), however, if a state is formerly but no longer present it needs to be wiped:
```
>>> from cylc.flow.data_messages_pb2 import PbWorkflow
>>> flow1 = PbWorkflow()
>>> flow1.state_totals['snapper'] = 2
>>> flow1.state_totals['clown'] = 3
>>> flow2 = PbWorkflow()
>>> flow2.state_totals['snapper'] = 1
>>> flow2.state_totals['flounder'] = 1
>>> flow1.state_totals
{'snapper': 2, 'clown': 3}
>>> flow2.state_totals
{'flounder': 1, 'snapper': 1}
>>> flow1.MergeFrom(flow2)
>>> flow1.state_totals
{'snapper': 1, 'flounder': 1, 'clown': 3}
```
So we do that at delta application time (as the delta will contain all totals, which are gathered from the families immediately under root):
```
    # Merge in updated fields
    if getattr(delta, 'updated', False):
        if key == WORKFLOW:
            # Clear fields that require overwrite with delta
            field_set = {f.name for f, _ in delta.updated.ListFields()}
            for field in CLEAR_FIELD_MAP[key]:
                if field in field_set:
                    data[key].ClearField(field)
            data[key].MergeFrom(delta.updated)
```
by flagging state_totals:
```
CLEAR_FIELD_MAP = {
    EDGES: set(),
    FAMILIES: set(),
    FAMILY_PROXIES: {'state_totals', 'states'},
    JOBS: set(),
    TASKS: set(),
    TASK_PROXIES: {'prerequisites'},
    WORKFLOW: {'state_totals', 'states'},
}
```

cylc/flow/data_store_mgr.py
```diff
@@ -1182,81 +1282,391 @@ def set_graph_window_extent(self, n_edge_distance):
...
+        tp_id, tproxy = self.store_node_fetcher(itask.tdef.name, itask.point)
+        if not tproxy:
+            return
+        # update task instance
+        update_time = time()
+        tp_delta = self.updated[TASK_PROXIES].setdefault(
+            tp_id, PbTaskProxy(id=tp_id))
+        tp_delta.stamp = f'{tp_id}@{update_time}'
+        ext_trigger = tp_delta.external_triggers[trig]
+        ext_trigger.message = message
```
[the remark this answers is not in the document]
dwsutherland: This one's a bit different, as I don't set the `message` in on task/node creation.. Also, I think it's user defined on the fly (via command).

cylc/flow/scheduler.py
```diff
@@ -1598,7 +1587,7 @@ def should_process_tasks(self):
                 process = True
             if self.pool.set_expired_task(itask, time()):
                 process = True
-            if itask.is_ready():
+            if all(itask.is_ready()):
```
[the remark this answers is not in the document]
dwsutherland: I wanted to piggy-back on the clock checking done within (to not check twice), so returned the checks within `is_ready`:
```
    def is_ready(self):
        """Am I in a pre-run state but ready to run?

        Queued tasks are not counted as they've already been deemed ready.

        """
        if self.manual_trigger:
            return (True,)
        if self.state.is_held:
            return (False,)
        if self.state.status in self.try_timers:
            return (self.try_timers[self.state.status].is_delay_done(),)
        return (
            self.state(TASK_STATUS_WAITING),
            self.is_waiting_clock_done(),
            self.is_waiting_prereqs_done()
        )
```
Then create the delta if changed, in task_pool (get_ready_tasks):
```
                    if itask.tdef.clocktrigger_offset is not None:
                        self.data_store_mgr.delta_task_clock_trigger(
                            itask, check_items)
```
then in data store:
```
        _, clock, _ = check_items
        # update task instance
        if (
                tproxy.HasField('clock_trigger')
                and tproxy.clock_trigger.satisfied is not clock
        ):
```
Probably a better way of doing it.. 

Off topic, I think that `scheduler.should_process_tasks` needs looking at.. Why does it need to iterate over the task pool twice...

cylc/flow/task_events_mgr.py
```diff
@@ -812,7 +814,8 @@ def _retry_task(self, itask, wallclock_time, submit_retry=False):
                 os.getenv("CYLC_SUITE_RUN_DIR")
             )
             itask.state.add_xtrigger(label)
-        itask.state.reset(TASK_STATUS_WAITING)
+        if itask.state.reset(TASK_STATUS_WAITING):
+            self.data_store_mgr.delta_task_state(itask)
```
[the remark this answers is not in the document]
dwsutherland: I wrestled with this too..

On the one hand it would be simpler to pass the data_store_mgr object into ever `TaskProxy`, and then `Outputs` and `TaskState`, then to Prerequisites object and so on... You can see it being passed around a lot, like public bicycle but worse..

But on the other hand, the data_store_mgr doesn't need to know about all the task proxies (i.e. runahead and other not ready), and/or all their events...

At least this way we can be intentional about it... but I'm open..

tests/integration/test_data_store_mgr.py
dwsutherland: Yes. `update_workflow` is called within `update_data_structure` and `initiate_data_model`, so well covered in integration...

tests/integration/test_data_store_mgr.py
dwsutherland: 😁

cylc/flow/data_store_mgr.py
```diff
@@ -1182,81 +1282,391 @@ def set_graph_window_extent(self, n_edge_distance):
...
+        label, _, satisfied = item
+        # update task instance
+        update_time = time()
+        tp_delta = self.updated[TASK_PROXIES].setdefault(
+            tp_id, PbTaskProxy(id=tp_id))
+        tp_delta.stamp = f'{tp_id}@{update_time}'
+        tp_delta.latest_message = message
+        output = tp_delta.outputs[label]
+        output.label = label
+        output.message = message
```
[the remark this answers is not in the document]
dwsutherland: removed.

cylc/flow/task_events_mgr.py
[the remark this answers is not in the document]
oliver-sanders: Ok, might open a follow-on issue as this isn't entirely a data-store issue. Recently hit a bug where the DB wasn't being updated for a particular state change because the full procedure wasn't being followed, so I'm a bit worried about potential bugs now we have the task proxy, data store and data base to keep in sync.

(Long-term I would really like task events to be pushed to a centralised "event queue" with multiple consumers able to subscribe to the event stream.)

cylc/flow/scheduler.py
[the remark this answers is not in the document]
oliver-sanders: Ouch, if `is_ready` returns a tuple `itask.is_ready()` must always have resolved as `True` in that `if` statement!

I think you've just crushed an efficiency bug here.

cylc/flow/task_events_mgr.py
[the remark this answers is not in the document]
oliver-sanders: For now I guess we could write an callback method in the task_pool and pass that to each task_proxy to avoid having to pass a handful of managers down to every object?

cylc/flow/scheduler.py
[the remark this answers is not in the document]
hjoliver: No, @dwsutherland changed `TaskProxy.is_ready()` to return a tuple, on this branch.

oliver-sanders:
Chucked the prerequisites thing into another issue  - https://github.com/cylc/cylc-flow/issues/4036

cylc/flow/scheduler.py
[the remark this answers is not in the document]
oliver-sanders: Phew.

oliver-sanders (approved):

cylc/flow/task_events_mgr.py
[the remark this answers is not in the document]
oliver-sanders: Bumped - https://github.com/cylc/cylc-flow/issues/4037

Issue #3696, opened 2020-07-15
conf: ensure all proposed changes have been enacted


oliver-sanders:
Ensure all changed document in the [config change proposal document](https://cylc.github.io/cylc-admin/proposal-config-changes.html) have been implemented pre 8.0.0 release.

**Pull requests welcome!**

wxtim:
- [ ] [authentication] -> obsolete
- [ ] [cylc][events]mail * -> [scheduler][mail]
- [ ] [cylc][events] -> [scheduler][events]
- [ ] [cylc]task event mail interval -> [scheduler][mail]task event batch interval
- [ ] [cylc]health check interval -> [scheduler]health check interval
- [ ] process pool * -> [scheduler]process pool *
- [ ] [documentation] -> obsolete
- [ ] [document viewers] -> obsolete
- [ ] [suite logging] -> [scheduler][logging]
- [ ] run directory rolling archive length -> [scheduler]run directory rolling archive length
- [ ] [task events]mail smtp -> obsolete
- [ ] [task events]mail retry delays -> obsolete
- [ ] [suite servers] -> [scheduler]
- [ ]     [suite servers]auto restart delay -> [scheduler]auto restart delay
- [ ]     [suite servers]run hosts -> [scheduler][run hosts]available
- [ ]     [suite servers]condemned hosts -> [scheduler][run hosts]condemned
- [ ]     [suite servers]run ports -> [scheduler][run hosts]ports
- [ ]     [suite servers]ranking -> [scheduler][run hosts]ranking
- [ ] [suite host self-identification] -> [scheduler][host self-identification]
- [ ] [job platforms] -> [platforms]
- [ ] [job platforms][]task communication method -> [platforms][]communication method
- [ ] [cylc]UTC mode -> [scheduler]UTC mode
- [ ] [editors]
- [ ] Discussed at CylcCon. Agreed to keep these settings but discourage their use. Also, change them to use to $EDITOR & - [ ] $GEDITOR if set (otherwise use current defaults).
- [ ] disable interactive command prompts -> obsolete
- [ ] [test battery] -> obsolete
- [ ] [job platforms][]batch system -> [platforms][]job runner
- [ ] [job platforms][]batch submit command template -> [platforms][]job runner command template
- [ ] [job platforms][]task event handler retry delays -> obsolete
- [ ] [job platforms][]scp command -> obsolete
- [ ] No longer used. Also remove the cylc [util] scp-transfer command.

[...]

PR #3942, opened 2020-11-12
Remove "disable automatic shutdown"

MetRonnie:

[From the config proposal](https://cylc.github.io/cylc-admin/proposal-config-changes.html#cylchealth-check-interval---schedulerhealth-check-interval): `[cylc]disable automatic shutdown` -> obsolete
> **hjoliver on 9 Oct 2019**
> I think we can get rid of this item. The original intent was to make suites stay alive for ever, even if finished, so that users could see them in the GUI until explicitly shutting them down. But that is really not necessary, and may lead to forgotten zombie suites.

**Requirements check-list**
- [x] Does not need tests.
- [x] No changelog entry needed (entry will be written after #3696 is closed)
- [x] No dependency changes.


tests/flakyfunctional/events/31-dont-stall-succeeded.t
```diff
@@ -21,18 +21,13 @@ set_test_number 4
...
-done
+poll_suite_stopped
+
 grep_ok '\[t1\.1\] .*succeeded' "${SUITE_RUN_DIR}/log/suite/log"
 
-sleep 5
-cylc stop --max-polls=10 --interval=2 "${SUITE_NAME}" 1>'/dev/null' 2>&1
-run_fail "${TEST_NAME_BASE}-stall" \
+run_fail "${TEST_NAME_BASE}-not-stalled" \
     grep -q -F 'WARNING - suite stalled' "${SUITE_RUN_DIR}/log/suite/log"
```
MetRonnie, line 31: As far as I can tell this will still perform the objective of the test: "Test no stall when task pool has succeeded tasks only."

cylc/flow/scheduler_cli.py
```diff
@@ -108,18 +108,6 @@ def get_option_parser(is_restart, add_std_opts=False):
         help="Do not daemonize the suite (infers --format=plain)",
         action="store_true", dest="no_detach")
 
-    parser.add_option(
-        "-a", "--no-auto-shutdown", help="Do not shut down"
-        " the suite automatically when all tasks have finished."
-        " This flag overrides the corresponding suite config item.",
-        action="store_true", dest="no_auto_shutdown")
```
MetRonnie, line 115: Have I assumed too much in removing the CLI option as well?

hjoliver: No, I don't think so.

hjoliver (changes requested):
Looks good, but `tests/f/events/31...` can be removed; it doesn't really do anything useful now because the scheduler will just shut down as soon as the lone tasks succeeds (and in SoD there are no succeeded tasks kept in the task pool anyway).

hjoliver:
Meh, to save time I've pushed the test deletion to your branch.  One review will do.  All the other tests already passed.

hjoliver:
(Also fixed a conflict in a deprecations test, passed in my environment).

hjoliver (approved):

MetRonnie:
@hjoliver As you seem to be on a roll, you can review https://github.com/cylc/cylc-flow/pull/3943 too if you like

PR #3943, opened 2020-11-12
Global config changes: process pool, logging

MetRonnie:

- `process pool *` -> `[scheduler]process pool *`
- `[suite logging]` -> `[scheduler][logging]`
- `run directory rolling archive length` -> `[scheduler]run directory rolling archive length`

Also the `test_header` function `create_test_global_config` has been simplified to only arg which is appended to the user's global config, as there are no longer any top-level global config items that would need to be prepended.

**Requirements check-list**
- [x] No changelog entry needed (entry will be written after #3696 is closed)
- [x] No dependency changes.


hjoliver (approved):
:+1:

wxtim (approved):

PR #3945, opened 2020-11-13
remote: clarify platform/host commands

oliver-sanders:
* Make the default way of constructing / running ssh stuff platform orientated.
* Prefix the underlying primitive "host" commands with `_` to make the danger (no platform conf) clear.

**Requirements check-list**
- [x] No dependency changes.


oliver-sanders:
> **Note:** We still have a couple of uses of `_remote_cylc_cmd` in the scheduler host select logic. These will need to be upgraded in the future as at the moment we cannot configure the scheduler platforms in the way we can other platforms.

wxtim (approved):
Hippo

PR #3950, opened 2020-11-17
Cylc version: long format

hjoliver:

When working with multiple Cylc 8 installations via the (upcoming) central wrapper, `cylc version` should report the location of the Cylc version that it invokes, as well as the version string.  

Useful for developers too now that we don't use `git describe` for a fine-grained version string (between releases the hardwired and self-reported code `__version__` string stays the same), so playing with multiple git branches at once requires checking out to a different location and changing `$CYLC_HOME` (for the central wrapper) - setting a different `$CYLC_VERSION` for each branch no longer works.

On this branch, `cylc version` prints `VERSION (LOCATION)` with new command options for `VERSION` alone.

I considered requiring `--long` to get the location, but people may not be expecting additional options to a version query, (Also, `rose version` prints the location).

Example for cylc-flow in `/home/oliverh/cylc/cylc-flow` with symlink `cylc-8.0a3.dev -> cylc-flow`:
```
$ cylc version
8.0a3.dev (/home/oliverh/cylc)

$ cylc --version
8.0a3.dev (/home/oliverh/cylc)

$ cylc -V
8.0a3.dev (/home/oliverh/cylc)

$ cylc version --short
8.0a3.dev

$ cylc --version-short
8.0a3.dev
```

**Requirements check-list**

- [x] No dependency changes.

### DIFF
--- a/tests/functional/authentication/01-remote-suite-same-name.t
+++ b/tests/functional/authentication/01-remote-suite-same-name.t
@@ -34,7 +34,7 @@ scp ${SSH_OPTS} -pqr "${TEST_SOURCE_DIR}/${TEST_NAME_BASE}/"* \
 # shellcheck disable=SC2086
 run_ok "${TEST_NAME_BASE}-register" \
     ssh ${SSH_OPTS} "${CYLC_TEST_HOST}" \
-    CYLC_VERSION="$(cylc version)" cylc register "${SUITE_NAME}" \
+    CYLC_VERSION="$(cylc version --short)" cylc register "${SUITE_NAME}" \
     "cylc-run/${SUITE_NAME}"
 
 suite_run_ok "${TEST_NAME_BASE}" \
--- a/cylc/flow/scripts/cylc.py
+++ b/cylc/flow/scripts/cylc.py
@@ -18,6 +18,7 @@
 
 import os
 import sys
+import pathlib
 
 from ansimarkup import parse as cparse
 import click
@@ -339,20 +340,26 @@ def cli_help():
     sys.exit(0)
 
 
-def cli_version():
+def cli_version(short=False):
     """Display the Cylc Flow version."""
-    click.echo(__version__)
+    version = f"{__version__}"
+    if not short:
+        version += f" ({pathlib.Path(__file__).parent.parent.parent})"
+    click.echo(version)
     sys.exit(0)
 
 
 @click.command(context_settings={'ignore_unknown_options': True})
 @click.option("--help", "-h", "help_", is_flag=True, is_eager=True)
 @click.option("--version", "-V", is_flag=True, is_eager=True)
+@click.option("--version-short", is_flag=True, is_eager=True)
 @click.argument("cmd-args", nargs=-1)
-def main(cmd_args, version, help_):
+def main(cmd_args, version, version_short, help_):
     if not cmd_args:
         if version:
-            cli_version()
+            cli_version(short=False)
+        elif version_short:
+            cli_version(short=True)
         else:
             cli_help()
     else:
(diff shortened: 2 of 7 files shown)


kinow (approved):
Code looks good. Working as expected

```
# pip install -e .
(venv) kinow@ranma:~/Development/python/workspace/cylc-flow$ cylc version
8.0a3.dev (/home/kinow/Development/python/workspace/cylc-flow/cylc)

# pip install .
(venv) kinow@ranma:~/Development/python/workspace/cylc-flow$ cylc version
8.0a3.dev (/home/kinow/Development/python/workspace/cylc-flow/venv/lib/python3.8/site-packages/cylc)
```

hjoliver:
> Useful for developers too now that we don't use git describe for a fine-grained version string (between releases the hardwired and self-reported code __version__ string stays the same), so playing with multiple git branches at once requires checking out to a different location and changing $CYLC_HOME (for the central wrapper) - setting a different $CYLC_VERSION for each branch no longer works.

This is slightly brain-twisting.

Task jobs get `$CYLC_VERSION` from `cylc.flow.__version__`, so the wrapper will always invoke that version for job cylc commands. So you can't get away with giving different worktree branches different version directory names like `cylc-8.0a3.dev-master` and `cylc-8.0a3.dev-test` - jobs will not find those versions.

hjoliver:
(The modified tests all passed, phew).

hjoliver:
(Just converted this to draft while I figure out how best to handle both pip and conda installations)

tests/functional/authentication/01-remote-suite-same-name.t
```diff
@@ -34,7 +34,7 @@ scp ${SSH_OPTS} -pqr "${TEST_SOURCE_DIR}/${TEST_NAME_BASE}/"* \
 # shellcheck disable=SC2086
 run_ok "${TEST_NAME_BASE}-register" \
     ssh ${SSH_OPTS} "${CYLC_TEST_HOST}" \
-    CYLC_VERSION="$(cylc version)" cylc register "${SUITE_NAME}" \
+    CYLC_VERSION="$(cylc version --short)" cylc register "${SUITE_NAME}" \
```
oliver-sanders, line 37: In Cylc7 we had `cylc version` and `cylc version --long`, could we swap `--short` for `--long`, It seems safer to only give extra when the user requests extra.

hjoliver: I considered that and came down on the other side (in the description):
> I considered requiring --long to get the location, but people may not be expecting additional options to a version query, (Also, `rose version` prints the location).

To expand on this a bit, on a system that may have multiple cylc-flow versions installed in both conda and pip virtual environments, and where the short version is no longer fine grained (via `git describe`), the installed path is more definitive than the release version string.  

Also: as I noted, `rose version` prints the path - are you going to change that?  And, what are your use cases that *require* the short format?  Presumably you mean just reporting the version string for `conda install cylc-flow=8.0a3` etc.?  IMO `cylc version` long format doesn't hinder that as no one is going it think that the local installation path is part of that version string.

hjoliver: (This isn't a big deal, obviously, but I given the reasons above I thought it would be good to show which cylc is in use by default, so long as that doesn't cause other problems).

kinow: I think showing version with `cylc version` by default, and showing extra info if required makes more sense (least surprise). But either way is fine for me.

kinow: Keeping my +1 here. Now up to @oliver-sanders the final review/merge :+1:

hjoliver:
Superseded by #3952

tests/functional/authentication/01-remote-suite-same-name.t
oliver-sanders: > it's a little surprising to see a member of the Rose team disagree rofl)

Not at all, I'm very glad for the chance to re-engineer Rose, a lot of really nasty behaviour there.

PR #3951, opened 2020-11-17
Define a local "at" platform

hjoliver:

Some local tests were being skipped with `requires batch:at`

One review should do @oliver-sanders 

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/etc/conf/global.cylc
+++ b/etc/conf/global.cylc
@@ -5,6 +5,10 @@
     [[_local_background_indep_tcp]]
         hosts = localhost
         install target = localhost
+    [[_local_at_indep_tcp]]
+        hosts = localhost
+        install target = localhost
+        batch system = at
     [[_remote_background_indep_tcp]]
         hosts = _remote_background_indep_tcp
     [[_remote_at_indep_tcp]]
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Ah, this was very much purposeful.

Atrun is an external batch submission system which is not pre-installed on all OS's and may require extra configuration to function. So, similar to slurm, pbs, etc developers are left to configure their own local test platforms (i.e. by adding those lines to their own `global.cylc` files).

hjoliver:
Well `at` is required for "Linux Standard Base" system compliance: https://refspecs.linuxfoundation.org/LSB_1.0.0/gLSB/command.html ... which I believe includes Debian-based and Red Hat based distros, and it should run "out of the box" on those systems.  But perhaps not on all distros.

BTW `atrun` is apparently not a thing anymore.  This ancient [article](https://www.linuxjournal.com/article/4087) (year 2000) says:
> Linux has separate dæmons for at and cron. Old versions of Linux used a program called “atrun”, which was run in root's crontab once a minute to execute at requests

oliver-sanders:
Sadly MacOS has atrun installed and disabled in such a cunning way that I can't figure out how to turn it on!

hjoliver:
OK I'll close this sucker.

PR #3952, opened 2020-11-19
Cylc version long take 2

hjoliver:
Supersedes #3950 

```
$ cylc --version
8.0a3.dev

$ cylc version
8.0a3.dev

$ cylc version --long  # conda environment
8.0a2 (/home/oliverh/miniconda3/envs/cylc-8.0a2)

$ cylc version --long  # "pip install ." with venv in git clone
8.0a3.dev (/home/oliverh/cylc/cylc-flow.git/venv)
```

Path reported is the top level of the virtual environment, for both conda and Python venv.

@kinow - you might want a quick re-review since I've gone back to the more conventional `--long` option.

### DIFF
--- a/cylc/flow/scripts/__init__.py
+++ b/cylc/flow/scripts/__init__.py
@@ -78,7 +78,7 @@ LOGO_LINES = [
 ]
 
 LICENCE = dedent(f"""
-    The Cylc Suite Engine [{__version__}]
+    The Cylc Workflow Engine [{__version__}]
     Copyright (C) 2008-2020 NIWA
     & British Crown (Met Office) & Contributors.
 """)
--- a/cylc/flow/scripts/cylc.py
+++ b/cylc/flow/scripts/cylc.py
@@ -35,15 +35,16 @@ from cylc.flow.terminal import (
 
 
 def get_version(long=False):
-    """Return the Cylc Flow version and top level installation path.
+    """Return version string, and (if long is True) install location.
 
-    Path reports correctly for normal pip and conda installations
+    The install location returned is the top directory of the virtual
+    environment. This works for normal pip and conda installations
     but not for "pip install -e ." because of the source links.
     """
     version = f"{__version__}"
     if long:
-        parents = Path(__file__).parent.parent.parent
-        version += f" ({parents.parent.parent.parent.parent})"
+        path = Path(__file__).parent.parent.parent.parent.parent.parent.parent
+        version += f" ({path})"
     return version
 
 
--- a/cylc/flow/scripts/cylc.py
+++ b/cylc/flow/scripts/cylc.py
@@ -18,6 +18,7 @@
 
 import os
 import sys
+from pathlib import Path
 
 from ansimarkup import parse as cparse
 import click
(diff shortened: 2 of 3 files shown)


cylc/flow/scripts/__init__.py
```diff
@@ -78,7 +78,7 @@
 ]
 
 LICENCE = dedent(f"""
-    The Cylc Suite Engine [{__version__}]
+    The Cylc Workflow Engine [{__version__}]
```
kinow, line 81: :+1:

cylc/flow/scripts/cylc.py
```diff
@@ -33,15 +34,30 @@
 )
 
 
+def get_version(long=False):
+    """Return the Cylc Flow version and top level installation path.
+
+    Path reports correctly for normal pip and conda installations
+    but not for "pip install -e ." because of the source links.
```
kinow, line 41: Probably doesn't report correctly for Zip imports (but quite sure it won't work anyway due to namespaces in our packages)

cylc/flow/scripts/cylc.py
```diff
@@ -33,15 +34,30 @@
 )
 
 
+def get_version(long=False):
+    """Return the Cylc Flow version and top level installation path.
```
kinow, line 38: Top level installation path only if long is True?

kinow (approved):
Few comments, but nothing major. Tested and reported the short version by default, and `--long` included the path :+1:

cylc/flow/scripts/cylc.py
oliver-sanders: How about displaying the path to the `cylc.flow` module rather than the top-level env?

```python
Path(cylc.flow.__file__).parent
```

hjoliver: I doubt the rather lengthy cylc-flow Python library location is what users really want to see:
```
$ cylc version --long
8.0a3.dev (/home/oliverh/cylc/cylc-flow/worktrees/cylc-version-long/venv/lib/python3.7/site-packages/cylc/flow)
```
If you're using the CLI (`cylc version`) then you really care more about the application(s), which is much closer to the top:
```
$ which cylc
~/cylc/cylc-flow/worktrees/cylc-version-long/venv/bin/cylc
```
and the common parent location is of course the top level of the venv, which is what I currently have it printing:
```
cylc version --long
8.0a3.dev (/home/oliverh/cylc/cylc-flow/worktrees/cylc-version-long/venv)
```
And bear in mind that the conda version (or even a more sensibly named Python venv) is even clearer:
```
8.0a3.dev (/home/oliverh/miniconda3/envs/cylc-8.0a3.dev)
```
What do you think?

cylc/flow/scripts/cylc.py
hjoliver: Yep, I'll clarify that.

cylc/flow/scripts/cylc.py
hjoliver: (Another way to look at it: print where the *package* is installed, which is more than just the Python library).

hjoliver: > Probably doesn't report correctly for Zip imports (but quite sure it won't work anyway due to namespaces in our packages)

I'm sure we can tweak this function in the future, to work for other kinds of installations.

oliver-sanders: > 8.0a3.dev (/home/oliverh/cylc/cylc-flow/worktrees/cylc-version-long/venv/lib/python3.7/site-packages/cylc/flow)

Wow, you've got a funky setup!

> If you're using the CLI (cylc version) then you really care more about the application(s), which is much closer to the top:

In that case how about the path to the Cylc executable:

```
$ which cylc
$ python3 -c 'import cylc.flow.scripts.cylc; print(cylc.flow.scripts.__file__)'
```

oliver-sanders (approved):
Works for "normal" installations via the expected channels so serves the intended use case.

Would be nice to have a more generic solution (especially editable pip installs).

cylc/flow/scripts/cylc.py
hjoliver: > In that case how about the path to the Cylc executable:

No, `cylc.flow.scripts.__file__` is the path to the library modules called by the executable. 

> Would using the Python executable location work?

That's a good idea :tada:  It works even for `pip install -e .`.  I'll push up the change...

(Now I hope we don't have to worry about people going to great lengths to run cylc-flow with the system Python :confounded:)

hjoliver:
Now uses Python interpreter path for reasons described above (and even works with "editable" pip installs) - is that OK with you @kinow ?

kinow (approved):
Tested and still works fine for me. Using either `pip install .` and `pip install -e .`, `cylc version` and `cylc version --long` both work fine :+1: great job @hjoliver !

hjoliver:
Good example of how even the seemingly simplest things can blow out 😬

Issue #3840, opened 2020-09-25
Rsync to subprocpool


datamel:
Follow up to https://github.com/cylc/cylc-flow/pull/3796

As suggested by @oliver-sanders and agreed by @hjoliver, run rsync in the subprocpool.

oliver-sanders:
Closed by #3953

PR #3953, opened 2020-11-19
Rsync to subprocpool

datamel:
Moves rsync to the subprocpool.
  
These changes close Rsync to subprocpool #3840.

- [x] No dependency changes.

### DIFF
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -27,44 +27,85 @@ This module provides logic to:
 import json
 import os
 from copy import deepcopy
-from logging import CRITICAL, DEBUG, INFO, WARNING
+from logging import (
+    CRITICAL,
+    DEBUG,
+    INFO,
+    WARNING
+)
 from shutil import rmtree
 from time import time
 
 from cylc.flow import LOG
 from cylc.flow.batch_sys_manager import JobPollContext
-from cylc.flow.exceptions import (PlatformLookupError, SuiteConfigError,
-                                  TaskRemoteMgmtError)
-from cylc.flow.hostuserutil import get_host, is_remote_host, is_remote_platform
+from cylc.flow.exceptions import (
+    PlatformLookupError,
+    SuiteConfigError,
+    TaskRemoteMgmtError
+)
+from cylc.flow.hostuserutil import (
+    get_host,
+    is_remote_host,
+    is_remote_platform
+)
 from cylc.flow.job_file import JobFileWriter
-from cylc.flow.parsec.util import pdeepcopy, poverride
+from cylc.flow.parsec.util import (
+    pdeepcopy,
+    poverride
+)
 from cylc.flow.pathutil import get_remote_suite_run_job_dir
-from cylc.flow.platforms import (HOST_REC_COMMAND, PLATFORM_REC_COMMAND,
-                                 get_host_from_platform,
-                                 get_install_target_from_platform,
-                                 get_platform)
+from cylc.flow.platforms import (
+    HOST_REC_COMMAND,
+    PLATFORM_REC_COMMAND,
+    get_host_from_platform,
+    get_install_target_from_platform,
+    get_platform
+)
 from cylc.flow.remote import construct_ssh_cmd
 from cylc.flow.subprocctx import SubProcContext
 from cylc.flow.subprocpool import SubProcPool
-from cylc.flow.task_action_timer import TaskActionTimer, TimerFlags
-from cylc.flow.task_events_mgr import TaskEventsManager, log_task_job_activity
-from cylc.flow.task_job_logs import (JOB_LOG_JOB, NN,
-                                     get_task_job_activity_log,
-                                     get_task_job_id, get_task_job_job_log,
-                                     get_task_job_log)
+from cylc.flow.task_action_timer import (
+    TaskActionTimer,
+    TimerFlags
+)
+from cylc.flow.task_events_mgr import (
+    TaskEventsManager,
+    log_task_job_activity
+)
+from cylc.flow.task_job_logs import (
+    JOB_LOG_JOB,
+    NN,
+    get_task_job_activity_log,
+    get_task_job_id,
+    get_task_job_job_log,
+    get_task_job_log
+)
 from cylc.flow.task_message import FAIL_MESSAGE_PREFIX
-from cylc.flow.task_outputs import (TASK_OUTPUT_FAILED, TASK_OUTPUT_STARTED,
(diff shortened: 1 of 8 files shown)


cylc/flow/task_job_mgr.py
```diff
@@ -227,17 +233,30 @@ def submit_task_jobs(self, suite, itasks, curve_auth,
...
             # Re-fetch a copy of platform
             platform = itasks[0].platform
-            is_init = self.task_remote_mgr.remote_init(
-                platform, curve_auth, client_pub_key_dir)
-            if is_init is None:
-                # Remote is waiting to be initialised
+            if install_target not in self.task_remote_mgr.remote_init_map.keys(
+            ) or self.task_remote_mgr.remote_init_map[install_target] is None:
+                self.task_remote_mgr.remote_init(
+                    platform, curve_auth, client_pub_key_dir)
```
oliver-sanders, line 239: I *think* this branch could do with splitting in two to avoid calling `remote_init` when not necessary though might be missing something.

cylc/flow/task_remote_mgr.py
```diff
@@ -165,44 +169,31 @@ def remote_init(self, platform, curve_auth,
...
         install_target = platform['install target']
 
-        # If task is running locally or the install target is localhost
-        # we can skip the rest of this function
-        if (install_target == 'localhost' or
-                self.single_task_mode or
-                not is_remote_host(get_host_from_platform(platform))):
-            LOG.debug(f"REMOTE INIT NOT REQUIRED for {install_target}")
-            return REMOTE_INIT_NOT_REQUIRED
-
```
oliver-sanders, line 189: Will need this logic somewhere? Otherwise could change the default in the remote-init map:

```
def __init__(..._):
    self.remote_init_map = {'localhost': REMOTE_INIT_DONE}
```

oliver-sanders (commented):
Looks like the right approach :+1:.

cylc/flow/task_remote_mgr.py
datamel: I think in order to pickup the hosts defined as something other than localhost we will need to check is_remote_host at some stage. I've moved this out of remote_init so we do the check before going in here.

cylc/flow/task_job_mgr.py
```diff
@@ -25,52 +25,46 @@
 """
 
 import json
-from logging import DEBUG, CRITICAL, INFO, WARNING
 import os
+from copy import deepcopy
```
datamel, line 29: Looks like I have accidentally alphabetized these - me being a bit overzealous with the auto format on vscode. Happy to revert this if it makes reviewing easier, although I'm not sure if it looks tidier that way?

cylc/flow/task_job_mgr.py
datamel: So, now remote_init_map install targets should never be mapped to none. I have removed this condition.

cylc/flow/task_job_mgr.py
oliver-sanders: "alphabetizing" +1
"multiple imports per line" -1

tests/functional/restart/53-task-prerequisites.t
```diff
@@ -44,7 +44,7 @@ cmp_ok "${TEST_NAME}.stdout" << '__EOF__'
 3|bar|apollo|3|The Eagle has landed|0
 __EOF__
 
-suite_run_ok "${TEST_NAME_BASE}-restart" cylc restart "${SUITE_NAME}" --stop-point=3
+suite_run_fail "${TEST_NAME_BASE}-restart" cylc restart "${SUITE_NAME}" --stop-point=3 --no-detach
 poll_suite_stopped
```
oliver-sanders, line 48:
```suggestion
```

cylc/flow/task_job_mgr.py
datamel: > "multiple imports per line" / "right-hand code squash" -1
I'll fix these, thanks!

cylc/flow/task_job_mgr.py
```diff
@@ -268,7 +282,11 @@ def submit_task_jobs(self, suite, itasks, curve_auth,
                     'batch_sys_name': itask.summary['batch_sys_name'],
                 })
                 itask.is_manual_submit = False
-            if is_init == REMOTE_INIT_FAILED:
+
+            if (ri_map[install_target] in [REMOTE_INIT_FAILED,
+                                           REMOTE_FILE_INSTALL_FAILED]):
+                init_error = (ri_map[install_target])
+                del ri_map[install_target]
```
oliver-sanders, line 289: I think there was a code comment with the old `del` line which would be helpful here.

oliver-sanders (approved):

tests/functional/restart/53-task-prerequisites.t
```diff
@@ -28,8 +28,8 @@ install_suite "${TEST_NAME_BASE}" "${TEST_NAME_BASE}"
 
 run_ok "${TEST_NAME_BASE}-validate" cylc validate "${SUITE_NAME}"
 
-suite_run_ok "${TEST_NAME_BASE}-run" cylc run "${SUITE_NAME}" --stop-point=2
-poll_suite_stopped
+suite_run_fail "${TEST_NAME_BASE}-run" cylc run "${SUITE_NAME}" --stop-point=2 --no-detach
+
```
MetRonnie, line 32: Why should this fail now?

oliver-sanders: Because the flow is now being run with `--no-detach` which means that `cylc run` is run in the foreground until the flow finishes. It then exits 0/1 based on the task pool.

Before the flow was being run in the background so `cylc run` always exited 0 as long as it succeeded to launch the flow.

The background approach was flaky as the log file and DB are closed after `poll_suite_stopped` returns.

MetRonnie: Oh I see, and the workflow in question exits with 1 because of the deliberate failed task that causes it to stall

MetRonnie:
Still waiting for a couple of Oliver's suggestions to be implemented I think?

oliver-sanders:
Ah, nothing important.

Please do a rebase squash before merging though.

MetRonnie (approved):

oliver-sanders:
tests/f/remote/04-symlink-dirs.t is unhappy

datamel:
> tests/f/remote/04-symlink-dirs.t is unhappy

Just looking into it now, hopefully I'll get to the bottom of it quickly!

PR #3954, opened 2020-11-19
Update cylc wrapper.

hjoliver:
New `cylc` wrapper for Cylc 8 conda and pip virtual envs.

Based on @dpmatthews original, tweaked a bit.

With this installed in `$PATH`:
- no need to activate venvs manually, the wrapper does it for you
- select versions via `CYLC_VERSION` (releases) or `CYLC_HOME` (between-release venvs in a git clone, e.g.)
- jobs can automatically see the same cylc as their parent scheduler, without hardwiring a single version in your environment
- supports release-version venvs name `cylc-$CYLC_VERSION` ~~and `cylc-flow-$CYLC_VERSION`~~

Need upcoming PR to divorce jobs from scheduler environment (later today...) to test this properly (currently on master local jobs inherit the scheduler's venv, which get in front of the wrapper in `$PATH`).

Documented in comments in the file, for now (explaining it concisely was the hardest part of this PR :grimacing: ).
In due course, some version of this documentation will need to go in the User Guide.
  
Note that we can't use CYLC_VERSION based selection for between-release versions (e.g. in git clones) in Cylc 8, because `cylc.flow.__version__` is now hardwired rather than determined by`git describe`. (Have to use CYLC_HOME  for that).

### DIFF
--- a/usr/bin/cylc
+++ b/usr/bin/cylc
@@ -121,15 +121,15 @@ if [[ -d "${CYLC_HOME}/conda-meta" ]]; then
     # Conda
     if [[ -f "${CYLC_HOME}/bin/activate" ]]; then
         # A conda pack installation
-        . ${CYLC_HOME}/bin/activate || exit 1
+        . "${CYLC_HOME}/bin/activate" || exit 1
     else
         # A normal conda environment
-        . $(dirname ${CYLC_HOME})/../etc/profile.d/conda.sh
+        . "$(dirname ${CYLC_HOME})/../etc/profile.d/conda.sh"
         conda activate "$(basename ${CYLC_HOME})" || exit 1
     fi
 elif [[ -f "${CYLC_HOME}/bin/activate" ]]; then
     # Python venv
-    . ${CYLC_HOME}/bin/activate || exit 1
+    . "${CYLC_HOME}/bin/activate" || exit 1
 else
     echo 1>&2 "ERROR: cylc installation not found in ${CYLC_HOME}"
     exit 1
--- a/usr/bin/cylc
+++ b/usr/bin/cylc
@@ -16,54 +16,123 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-# Wrapper for selection between installed versions of cylc.
-# Author: Matt Shin (Met Office).
-
-#_____________________________
-# INSTALLATION: 
+#------------------------------------------------------------------------------
+# Wrapper script to support multiple Cylc versions installed on the same host.
+# Currently handles Conda and Python virtual environments.
 #
-#  1. Install this script as "cylc" in $PATH for normal users.
+# INSTALL AND CONFIGURE THE WRAPPER
+#----------------------------------
+# Copy this script as "cylc" into the default $PATH on scheduler and job hosts,
+# and modify CYLC_HOME_ROOT(_ALT) below (see "EDIT ME") to point to the parent
+# directory of all installed versions.
 #
-#  2. Install cylc versions as described in the INSTALL file, e.g.:
-#     /opt/cylc-7.6.1/
-#     /opt/cylc-7.7.0/
-#     /opt/cylc -> cylc-7.7.0/  # symlink DEFAULT version
+# HOW IT WORKS
+#-------------
+# Intercept `cylc` commands and re-invoke them with the appropriate cylc-flow
+# version, according to location or version information in the environment:
+# - 1) $CYLC_HOME if defined, or
+# - 2) $CYLC_HOME_ROOT(_ALT)/$CYLC_VERSION
 #
-# Environment variables:
-#  $CYLC_HOME_ROOT: top installation directory, e.g. /opt/
-#  $CYLC_HOME:   a specific version, e.g. /opt/cylc-7.7.0/
-#  $CYLC_VERSION: a specific version string, e.g. 7.7.0
-#
-#  3. Set $CYLC_HOME_ROOT for your site - see "!!! EDIT ME !!! below.
-
-#_____________________________
-# ACTION IN ORDER OF PRIORITY:
+# USERS SHOULD:
+# - Set CYLC_VERSION to select a specific installed version under the ROOT
+# - Set CYLC_HOME in .bashrc (scheduler and job hosts) to select a single
+#   version outside of the ROOT location.
 # 
-# 1) If $CYLC_HOME is defined and exists it will be selected. If it does
-#   not exist the command will fail.
+# The scheduler writes cylc.flow.__version__ to CYLC_VERSION in task job
+# scripts so that jobs see the right version. __version__ only increments with
+# releases so CYLC_VERSION cannot be used for fine-grained selection. Git
+# branches between releases, for example, all have the same version. Use
+# CYLC_HOME to select a venv in your cylc-flow git clone.
 #
-# 2) If $CYLC_VERSION is defined and exists it will be selected. If it
-#   does not exist the default under $CYLC_HOME_ROOT will be selected.
+# CYLC_HOME must be set in .bashrc so that task jobs also see it - otherwise
+# they will try to select their CYLC_VERSION.
+#
+# If USERS set a default CYLC_VERSION in .bashrc it should take an existing
+# value first to avoid overiding task job version selection:
+# - export CYLC_VERSION=${CYLC_VERSION:-8.0.0}
+#
+# INSTALLING CYLC RELEASES WITH CONDA
+#------------------------------------
+# Create environments named cylc-$CYLC_VERSION or cylc-flow-$CYLC_VERSION under
+# a root location such as /opt/cylc:
+#
+# $ conda create -p /opt/cylc/cylc-8.2.0 cylc=8.2.0
 # 
-# 3) If $CYLC_HOME and $CYLC_VERSION are both undefined, the default
-#   under $CYLC_HOME_ROOT will be selected.
+# $ ls /opt/cylc
+# cylc-8.1.0/
+# cylc-8.2.0/
+#
+# Note you can `pip install` cylc-flow from a git clone into a conda
+# environment and then select it by CYLC_VERSION but that will be misleading
+# (the code won't actually be the release version). It is better to use Python
+# venvs and CYLC_HOME for cylc-flow development and testing.
+#
+# INSTALLING CYLC-FLOW RELEASES WITH PIP FOR DEVELOPMENT AND TESTING
+#-------------------------------------------------------------------
+# Create Python virtual environments in-place in your cylc-flow git
+# clones and select them using CYLC_HOME in your .bashrc.
+#
+# Create a new venv and install cylc-flow:
+#   $ cd <my-cylc-flow-clone>
+#   $ python -m venv venv
+#   $ . venv/bin/activate
+#   $ pip install -e . 
+#
+# Set CYLC_HOME in .bashrc to select this venv (see explanation above):
+#   $ echo "export CYLC_HOME=$PWD/venv" >> .bashrc
+#   $ . .bashrc
+#
+# You can also `pip install` cylc-flow releases under a common root location
+# for CYLC_VERSION based selection, like for conda:
+#
+#   $ CYLC_HOME_ROOT=$HOME/cylc-flow/releases  # (set in this wrapper too)
+#   $ python -m venv $CYLC_HOME_ROOT/cylc-flow-8.0a2
+#   $ . $CYLC_HOME_ROOT/cylc-flow-8.0a2/bin/activate
+#   $ python -m pip install cylc-flow==8.0a2
+# 
+# (But there is no need to do this if you have full system releases - i.e. not
+# just cylc-flow - installed via conda anyway.)
  
-#_____________________________
-# NOTES:
-# * Users will typically accept the default or set $CYLC_VERSION, but
-#   they may set $CYLC_HOME to pick up a private cylc installation.
-# * Developers may set $CYLC_HOME to pick up their own repository clone.
-# * Running suites export $CYLC_VERSION to task environments, so that
-#    task messaging commands etc. pick up the compatible cylc version.
-# * Note that $CYLC_ROOT_HOME can be overridden in the environment too.
+##############################!!! EDIT ME !!!##################################
+# Central Cylc conda releases:
+CYLC_HOME_ROOT=${CYLC_HOME_ROOT:-/opt}
+# User Conda releases:
+CYLC_HOME_ROOT_ALT=${CYLC_HOME_ROOT_ALT:-/home/$USER/miniconda3/envs}
+# NOTE: users can set CYLC_HOME in their .bashrc to select a non-release conda
+# or pip virtual environment (located in a cylc-flow git clone, for example).
+###############################################################################
 
-CYLC_HOME_ROOT=${CYLC_HOME_ROOT:-/opt} # !!! EDIT ME !!!
-
+# If CYLC_HOME is not defined, find it using {ROOT} and {VERSION}.
 if [[ -z ${CYLC_HOME:-} ]]; then
-    if [[ -n ${CYLC_VERSION:-} && -d $CYLC_HOME_ROOT/cylc-$CYLC_VERSION ]]; then
-        CYLC_HOME=$CYLC_HOME_ROOT/cylc-$CYLC_VERSION
-    else
-        CYLC_HOME=$CYLC_HOME_ROOT/cylc
+    CYLC_HOME=$CYLC_HOME_ROOT/cylc  # default symlinked version
+    if [[ -n ${CYLC_VERSION:-} ]]; then
+        # Look for matching version in root locations
+        for ROOT in "${CYLC_HOME_ROOT}" "${CYLC_HOME_ROOT_ALT}"; do
+            for NAME in "cylc" "cylc-flow"; do
+                if [[ -d $ROOT/${NAME}-$CYLC_VERSION ]]; then
+                    CYLC_HOME=$ROOT/${NAME}-$CYLC_VERSION
+                    break 2
+                fi
+            done
+         done
     fi
 fi
-exec "$CYLC_HOME/bin/$(basename "$0")" "$@"
+if [[ -d "${CYLC_HOME}/conda-meta" ]]; then
+    # Conda
+    if [[ -f "${CYLC_HOME}/bin/activate" ]]; then
+        # A conda pack installation
+        . ${CYLC_HOME}/bin/activate || exit 1
+    else
+        # A normal conda environment
+        . $(dirname ${CYLC_HOME})/../etc/profile.d/conda.sh
+        conda activate "$(basename ${CYLC_HOME})" || exit 1
+    fi
+elif [[ -f "${CYLC_HOME}/bin/activate" ]]; then
+    # Python venv
+    . ${CYLC_HOME}/bin/activate || exit 1
+else
+    echo 1>&2 "ERROR: cylc installation not found in ${CYLC_HOME}"
+    exit 1
+fi
+# The selected Cylc should now be first in $PATH
+exec "$(basename "$0")" "$@"
--- a/usr/bin/cylc
+++ b/usr/bin/cylc
@@ -130,15 +130,15 @@ if [[ -d "${CYLC_HOME}/conda-meta" ]]; then
         . "${CYLC_HOME}/bin/activate" || exit 1
     else
         # A normal conda environment
-        . "$(dirname ${CYLC_HOME})/../etc/profile.d/conda.sh"
-        conda activate "$(basename ${CYLC_HOME})" || exit 1
+        . "$(dirname "${CYLC_HOME}")"/../etc/profile.d/conda.sh
+        conda activate "$(basename "${CYLC_HOME}")" || exit 1
     fi
 elif [[ -f "${CYLC_HOME}/bin/activate" ]]; then
     # A Cylc 8 Python venv
     . "${CYLC_HOME}/bin/activate" || exit 1
 elif [[ -x "${CYLC_HOME}/bin/cylc" ]]; then
     # A Cylc 7 installation
-    exec $CYLC_HOME/bin/$(basename $0) "$@"
+    exec "${CYLC_HOME}"/bin/"$(basename "$0")" "$@"
 else
     echo 1>&2 "ERROR: cylc not found in ${CYLC_HOME}"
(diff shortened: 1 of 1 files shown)


hjoliver:
(Ignore cancelled tests - they do not test this wrapper at all).

usr/bin/cylc
```diff
@@ -16,54 +16,123 @@
...
-# * Running suites export $CYLC_VERSION to task environments, so that
-#    task messaging commands etc. pick up the compatible cylc version.
-# * Note that $CYLC_ROOT_HOME can be overridden in the environment too.
-
-CYLC_HOME_ROOT=${CYLC_HOME_ROOT:-/opt} # !!! EDIT ME !!!
+##############################!!! EDIT ME !!!##################################
+# Central Cylc conda releases:
+CYLC_HOME_ROOT=${CYLC_HOME_ROOT:-/opt}
+# User Conda releases:
+CYLC_HOME_ROOT_ALT=${CYLC_HOME_ROOT_ALT:-/home/$USER/miniconda3/envs}
```
oliver-sanders, line 100: Nice.

oliver-sanders (commented):
Looks good, I think this is missing Cylc7 support which will prove a barrier to mixed deployment for us (although my preference is to install Cylc7 via Conda too - #3958).

usr/bin/cylc
hjoliver: (Should be `$HOME/miniconda3/envs` actually; I'll fix that).

usr/bin/cylc
```diff
@@ -16,54 +16,123 @@
...
+# Intercept `cylc` commands and re-invoke them with the appropriate cylc-flow
+# version, according to location or version information in the environment:
+# - 1) $CYLC_HOME if defined, or
+# - 2) $CYLC_HOME_ROOT(_ALT)/$CYLC_VERSION
 #
-#  2. Install cylc versions as described in the INSTALL file, e.g.:
-#     /opt/cylc-7.6.1/
-#     /opt/cylc-7.7.0/
-#     /opt/cylc -> cylc-7.7.0/  # symlink DEFAULT version
+# USERS SHOULD:
```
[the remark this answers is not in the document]
hjoliver: Done. (Next commit, when pushed up).

usr/bin/cylc
```diff
@@ -16,54 +16,123 @@
...
+# cylc-8.2.0/
+#
+# Note you can `pip install` cylc-flow from a git clone into a conda
+# environment and then select it by CYLC_VERSION but that will be misleading
+# (the code won't actually be the release version). It is better to use Python
+# venvs and CYLC_HOME for cylc-flow development and testing.
 #
-# 2) If $CYLC_VERSION is defined and exists it will be selected. If it
-#   does not exist the default under $CYLC_HOME_ROOT will be selected.
+# INSTALLING CYLC-FLOW RELEASES WITH PIP FOR DEVELOPMENT AND TESTING
```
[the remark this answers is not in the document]
hjoliver: Done (in next commit, when pushed up).

usr/bin/cylc
```diff
@@ -16,54 +16,123 @@
...
+# Create a new venv and install cylc-flow:
+#   $ cd <my-cylc-flow-clone>
+#   $ python -m venv venv
+#   $ . venv/bin/activate
+#   $ pip install -e . 
+#
+# Set CYLC_HOME in .bashrc to select this venv (see explanation above):
+#   $ echo "export CYLC_HOME=$PWD/venv" >> .bashrc
+#   $ . .bashrc
+#
```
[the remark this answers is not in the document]
hjoliver: (Already done higher up).

hjoliver:
Cylc 7 support added (I had intended to that).

usr/bin/cylc
```diff
@@ -16,54 +16,132 @@
...
+fi
+if [[ -d "${CYLC_HOME}/conda-meta" ]]; then
+    # A Cylc 8 Conda environment
+    if [[ -f "${CYLC_HOME}/bin/activate" ]]; then
+        # A conda pack installation
+        . "${CYLC_HOME}/bin/activate" || exit 1
     else
-        CYLC_HOME=$CYLC_HOME_ROOT/cylc
+        # A normal conda environment
+        . "$(dirname "${CYLC_HOME}")"/../etc/profile.d/conda.sh
```
oliver-sanders, line 133: Perhaps protect this with `[[ -d "${CONDA_EXE}" ]]` to prevent duplicate calls.

dpmatthews: Not safe if a user has activated a different conda environment first?

usr/bin/cylc
```diff
@@ -16,54 +16,132 @@
...
+        conda activate "$(basename "${CYLC_HOME}")" || exit 1
     fi
+elif [[ -f "${CYLC_HOME}/bin/activate" ]]; then
+    # A Cylc 8 Python venv
+    . "${CYLC_HOME}/bin/activate" || exit 1
```
dpmatthews, line 138: venv & conda pack use the same logic so this can be simplified.
Also, use bash parameter expansion to avoid the extra forks?
```bash
if [[ -f "${CYLC_HOME}/bin/activate" ]]; then
    # A Python venv or conda pack installation
    . "${CYLC_HOME}/bin/activate" || exit 1
elif [[ -d "${CYLC_HOME}/conda-meta" && -f "${CYLC_HOME%/*}/../etc/profile.d/conda.sh" ]]; then
    # A normal conda environment
    . "${CYLC_HOME%/*}/../etc/profile.d/conda.sh"
    conda activate "${CYLC_HOME##*/}" || exit 1
```

usr/bin/cylc
oliver-sanders: Not sure, don't want to find out.

usr/bin/cylc
```diff
@@ -16,54 +16,132 @@
...
-# Wrapper for selection between installed versions of cylc.
-# Author: Matt Shin (Met Office).
-
-#_____________________________
-# INSTALLATION: 
-#
-#  1. Install this script as "cylc" in $PATH for normal users.
+#------------------------------------------------------------------------------
+# Wrapper script to support multiple Cylc versions installed on the same host.
+# Handles Cylc 8 Conda and Python virtual envs, and plain Cylc 7 installations.
```
dpmatthews, line 21: Cylc 8 could be done as plain installation couldn't it?
Also, cylc 7 could be installed using Conda or venv.
So, I suggest: "Handles Conda and Python virtual envs, and plain installations"

oliver-sanders: > Cylc 8 could be done as plain installation couldn't it?


I guess, however, Cylc8 doesn't have the same `bin/cylc` pattern as Cylc7 so this would require extra logic. `CYLC_HOME` would have to point to the installed `cylc` script?

usr/bin/cylc
```diff
@@ -16,54 +16,132 @@
...
+##############################!!! EDIT ME !!!##################################
+# Central Cylc Conda releases:
+CYLC_HOME_ROOT="${CYLC_HOME_ROOT:-/opt}"
+# User Conda releases:
+CYLC_HOME_ROOT_ALT="${CYLC_HOME_ROOT_ALT:-$HOME/miniconda3/envs}"
 
-CYLC_HOME_ROOT=${CYLC_HOME_ROOT:-/opt} # !!! EDIT ME !!!
+# Legacy Cylc 7 locations if needed:
+CYLC7_HOME_ROOT="${CYLC7_HOME_ROOT_ALT:-/opt}" 
+CYLC7_HOME_ROOT_ALT="${CYLC7_HOME_ROOT_ALT:-$HOME/cylc/cylc-7}"
```
dpmatthews, line 105: I'd have thought just having CYLC_HOME_ROOT (defaults to /opt) and CYLC_HOME_ROOT_ALT (no default) would be enough to cover most cases. The logic treats them all the same anyway.

usr/bin/cylc
dpmatthews: OK - I'm happy that plain installations only support cylc 7.
"Handles Conda and Python virtual envs, and plain (Cylc 7) installations"

usr/bin/cylc
```diff
@@ -16,54 +16,132 @@
...
-if [[ -z ${CYLC_HOME:-} ]]; then
-    if [[ -n ${CYLC_VERSION:-} && -d $CYLC_HOME_ROOT/cylc-$CYLC_VERSION ]]; then
-        CYLC_HOME=$CYLC_HOME_ROOT/cylc-$CYLC_VERSION
+if [[ -z "${CYLC_HOME:-}" ]]; then
+    # If CYLC_HOME is not defined find it with CYLC_HOME_ROOT and CYLC_VERSION.
+    if [[ -n ${CYLC_VERSION:-} ]]; then
+        # Look for matching version in root locations
+        for ROOT in "${CYLC_HOME_ROOT}" "${CYLC_HOME_ROOT_ALT}" \
+                "${CYLC7_HOME_ROOT}" "${CYLC7_HOME_ROOT_ALT}"; do
+            for NAME in "cylc" "cylc-flow"; do
```
dpmatthews, line 114: I'd have thought we could just stick with using "cylc".

usr/bin/cylc
hjoliver: I was trying to handle user conda envs out of the box, with _ALT.  But I suppose a different variable name would be more appropriate for that.  (Thinking about it...)

usr/bin/cylc
hjoliver: My intention here was to allow separate conda installation of just the back end (`conda install cylc-flow`) with the environment named appropriately to distinguish from a full-system environment. But maybe that's overkill; I'll back it out to just "cylc".

usr/bin/cylc
hjoliver: Yep.

>  Also, use bash parameter expansion to avoid the extra forks?

Good idea, it might be slightly more efficient.  We can also do that with the final `basename` call in the `exec` line (which has been there forever in Cylc 7).

usr/bin/cylc
hjoliver: A non-venv Cylc 8 seems a bit perverse as you'd have to install all the dependencies yourself. If anyone needs to do that, we can tweak the wrapper in a future PR.

hjoliver: Also we can adjust later for Cylc 7 via conda or venv if necessary. That is difficult to do because of the pygtk dependencies - see #3958

usr/bin/cylc
hjoliver: I was also allowing for Cylc 7 and 8 to be installed in different locations, but on reflection I suppose that is not necessary. Simplifying...

usr/bin/cylc
hjoliver: It appears to be safe to source this multiple times. By experiment, it doesn't change my environment after the first time.

hjoliver:
@oliver-sanders and @dpmatthews - I've modified the wrapper as per your suggestions, and in fact gone a bit further as well. Also tidied the documentation (comments) a bit more, and added a short section on Cylc 7.

There's one problem: if using conda, the wrapper will cause infinite recursion for jobs that inherit the scheduler environment. I blew several hours on this :angry: but eventually figured out what's happening. To test successfullym merge in #3955 and set global config `clean job submission environment = True`.

It's possibly a bug in `conda activate`. If not (or in lieu of a fix) we might want to reconsider the making it optional to have a clean job submission environment.

hjoliver:
cylc-doc Issue https://github.com/cylc/cylc-doc/issues/173

hjoliver:
### wrapper fork-bomb

If using a conda environment via `CYLC_HOME` or `CYLC_HOME_ROOT/CYLC_VERSION`, the wrapper will activate the environment before executing `cylc run`.  Then, local jobs will inherit the scheduler's `CONDA` variables. When a job calls `cylc message` (say) the wrapper will locate the right conda environment and do `conda activate` BUT unexpectedly that does not put the environment bin directory at the front of `PATH` ... so the wrapper will end up calling itself again, ad infinitum.

This was :angry:-making to debug because initially I couldn't reproduce it outside of a task job. `conda activate` did the right thing with `PATH` even if I manually exported all the `CONDA` variables in my terminal first. I ended up trying to debug `conda/activate.py`.

It turns out, the presence of the `CONDA` variables causes conda to run an internal `reactivate` method, which works fine *so long as the target environment bin directory is a) already in front of `PATH` or b) not in `PATH` at all.*

If the bin directory is already in `PATH` but not at the front, condas `reactivate` method detects it but reconstructs `PATH` with it in the same position instead of moving it to the front.  That seems stupid to me, but I don't fully understand what reactivate is supposed to do.

In task jobs, the inherited `PATH` has the cylc conda environment at front (from the scheduler activating it) but that gets prepended with other paths (including wrapper location) by login scripts, so the conda path ends up in the middle of PATH, which triggers the above problem.

usr/bin/cylc
```diff
@@ -16,54 +16,130 @@
...
-#  2. Install cylc versions as described in the INSTALL file, e.g.:
-#     /opt/cylc-7.6.1/
-#     /opt/cylc-7.7.0/
-#     /opt/cylc -> cylc-7.7.0/  # symlink DEFAULT version
+# HOW IT WORKS
+#-------------
+# Intercept "cylc" commands and re-invoke them with the appropriate cylc-flow
+# version, according to location or version information in the environment:
+# + 1) $CYLC_HOME if defined, or
+# + 2) $CYLC_HOME_ROOT[_ALT]/$CYLC_VERSION
```
kinow, line 34: Maybe have a third option

```
2) $CYLC_HOME_ROOT/$CYLC_VERSION
3) $CYLC_HOME_ROOT_ALT/$CYLC_VERSION
```
and/or describe that the `_ALT` option is used after `CYLC_HOME_ROOT`?

kinow:
Just reproduced the :fork_and_knife: :bomb: that @hjoliver explained (with a lot of patience) here and on Element.

I'm using `five`, and have:

```
(cylc-8.0a2) kinow@ranma:~/cylc-run/five/log$ cylc dump five
sh: 0: getcwd() failed: No such file or directory
sh: 0: getcwd() failed: No such file or directory
last_updated=1606279607.4708803
namespace definition order=['root', 'prep', 'foo', 'bar']
newest cycle point string=20130808T0000+13
newest runahead cycle point string=None
oldest cycle point string=20130808T0000+13
reloading=False
run_mode=live
state totals={'submitted': 1, 'waiting': 2}
states=['submitted', 'waiting', 'waiting']
status=running
status_string=running to stop at 20130808T12
suite_urls={'root': '', 'prep': '', 'foo': '', 'bar': '', 'suite': ''}
time zone info={'hours': 13, 'minutes': 0, 'string_basic': '+13', 'string_extended': '+13:00'}
bar, 20130808T0000+13, waiting, unspawned, unheld
foo, 20130808T0000+13, waiting, unspawned, unheld
prep, 20130808T0000+13, submitted, spawned, unheld
```

And 

```
(cylc-8.0a2) kinow@ranma:~/cylc-run/five/log$ cylc cat -f e five prep.20130808T0000+13
sh: 0: getcwd() failed: No such file or directory
sh: 0: getcwd() failed: No such file or directory
ERROR: CYLC_HOME not defined or found

```

No idea where these `sh: 0: getcwd() failed: No such file or directory` are coming from. But they can be ignored, not having these on `master`.

So for `five` workflow, looks like it's complaining about `CYLC_HOME` at the moment. I didn't set it, as I preferred to use the minimum required, which I thought was

```
export CYLC_HOME_ROOT=/home/kinow/Development/python/miniconda3/envs
export CYLC_VERSION=${CYLC_VERSION:-8.0a3.dev}
```

And created a conda env `cylc-8.0a2` (`CYLC_VERSION` doesn't match, but that's OK, as I can just prepend `CYLC_VERSION=8.0a2` in the command line and it works).

kinow:
Adding `set -x` in `job.sh`, my `PATH` is indeed wrong. The `~/bin` is coming before the Conda env bin.

```
+ PATH=/home/kinow/cylc-run/five/bin:/home/kinow/cylc-suites/five/bin:/home/kinow/bin:/home/kinow/.rbenv/bin:/home/kinow/.rbenv/versions/2.7.1/bin/:/home/kinow/bin:/home/kinow/.local/bin:/home/kinow/.rbenv/bin:/home/kinow/.rbenv/versions/2.7.1/bin/:/home/kinow/cylc-suites/five:/home/kinow/cylc-suites/five/bin:/home/kinow/Development/python/miniconda3/envs/cylc-8.0a2/bin:/home/kinow/Development/python/miniconda3/condabin:...
```

```
(cylc-8.0a2) kinow@ranma:~/cylc-run/five/log$ echo $PATH | sed 's/:/\n/g'
/home/kinow/Development/python/miniconda3/envs/cylc-8.0a2/bin
/home/kinow/Development/python/miniconda3/condabin
/home/kinow/Development/python/miniconda3/bin
/home/kinow/bin
...
```

kinow:
>It turns out, the presence of the CONDA variables causes conda to run an internal reactivate method, which works fine so long as the target environment bin directory is a) already in front of PATH or b) not in PATH at all.
>
>If the bin directory is already in PATH but not at the front, conda's reactivate method detects it but reconstructs PATH with it in the same position instead of moving it to the front. That seems stupid to me, but I don't fully understand what reactivate is supposed to do.

Probably related to: https://github.com/conda/conda/issues/10296

usr/bin/cylc
hjoliver: OK, I guess my `ROOT[_ALT]` is a bit cryptic.

hjoliver: Text tweaked.

hjoliver:
I've tried to make the comments even clearer (prompted by @kinow ) and moved the user instruction section to the bottom.

hjoliver:
> It might be that reactivate command that's messing with the $PATH perhaps?

The problem is it's NOT messing with PATH.  It should be putting the environment bin directory in front of PATH but it isn't.

kinow:
> > It might be that reactivate command that's messing with the $PATH perhaps?
> 
> The problem is it's NOT messing with PATH. It should be putting the environment bin directory in front of PATH but it isn't.

@hjoliver with this my wrapper was not invoked before my Conda environment file, as the environment variables appear to be correctly set.

```
cylc__job__main() {
    if [[ -n "${CONDA_SHLVL}" ]]; then
    	local SOURCE_FILE=$(realpath "$(conda info --base)/etc/profile.d/conda.sh")
    	local C_ENV="${CONDA_DEFAULT_ENV}"
        source "${SOURCE_FILE}"
    	unset CONDA_SHLVL
        conda activate "${C_ENV}"
    fi
```

I think `conda` doesn't always clean the environment, and if it finds left-over env vars, it gets confused. Unsetting `CONDA_SHLVL` appears to be the short circuit to tell to Conda that the current env can be re-activated (not that shell function, but tell it to run `conda activate blah` and get it actually activated...)

Tested with the wrapper on my `~/bin` and running `five` with Conda env activated, and with Conda env deactivated, but specifying the version in the cmd line, `CYLC_VERSION=8.0a2 cylc run --no-detach --stop-cycle-point=20130808T12 five`.

hjoliver:
Brilliant, thanks @kinow :+1:  I had not experimented with deliberating unsetting CONDA variables.

I think `unset CONDA_SHLVL` should be done in the wrapper though, just before invoking `conda activate`, just in case users load another conda environment in their own job scripting before `cylc` commands get called.  Would you agree? (I tested that it works :+1: )

kinow:
Sounds good to me 🎉

kinow (approved):
Did only a quick functional review yesterday, and read latest changes today, +1. Probably better keep the other two reviewers as they might have a better understanding of this change. Thanks again for the patience explaining the change to me yesterday @hjoliver !

hjoliver:
Thanks for solving the nasty fork bomb problem for me :+1:

hjoliver:
@dpmatthews @oliver-sanders - latest tweaks provide slightly better error messaging, and it works even in a dirty job environment thanks to Bruno's suggestion.  So I think it should be good to go now :tada:

usr/bin/cylc
```diff
@@ -16,54 +16,156 @@
...
+elif [[ -x "${CYLC_HOME}/bin/cylc" ]]; then
+    # A Cylc 7 installation
+    BASENAME_PREFIX="${CYLC_HOME}"/bin/
+else
+    echo 1>&2 "ERROR: cylc not found in ${CYLC_HOME}"
+    exit 1
+fi
+
+# Execute the command via the selected Cylc installation.
+exec "${BASENAME_PREFIX}${0##*/}" "$@"
```
dpmatthews, line 171: Why can't we stick with ```exec "${CYLC_HOME}/bin/${0##*/}" "$@"``` ?
That fixes the fork bomb issue (sorry - didn't notice this before).
As far as I can see this works in all cases.

usr/bin/cylc
```diff
@@ -16,54 +16,156 @@
...
 
-if [[ -z ${CYLC_HOME:-} ]]; then
-    if [[ -n ${CYLC_VERSION:-} && -d $CYLC_HOME_ROOT/cylc-$CYLC_VERSION ]]; then
-        CYLC_HOME=$CYLC_HOME_ROOT/cylc-$CYLC_VERSION
-    else
-        CYLC_HOME=$CYLC_HOME_ROOT/cylc
+# If CYLC_HOME is not defined try CYLC_HOME_ROOT/cylc-CYLC_VERSION
+if [[ -z "${CYLC_HOME}" ]]; then
+    if [[ -z ${CYLC_VERSION} ]]; then
+        echo 1>&2 "ERROR: CYLC_HOME or CYLC_VERSION must be defined"
```
dpmatthews, line 125: This no longer supports the symlink default version?

usr/bin/cylc
hjoliver: Damn it, you're right. I switched to relying on PATH to find `cylc` for the conda case because that's more in the spirit of virtual environments (find the environment and activate it, then rely on the "activated" PATH). It's probably still sensible to force `conda activate` to do the right thing with PATH just in case, but yes, going back to the explicit path/to/cylc would prevent the fork bomb either way. (Jeez, not my best effort, this PR :grimacing: !!)

hjoliver: (Well, in my defense, if `conda activate` had done what it says on the tin (to PATH) my way wouldn't have caused a fork bomb either).

usr/bin/cylc
hjoliver: damn-it_2

hjoliver: fixed (late night, hope I didn't screw it up again!) :wave:

dpmatthews (approved):
Looks good to me

oliver-sanders (approved):
Looks good, tested.

PR #3712, opened 2020-07-26
Divorce local jobs from scheduler environment.

hjoliver:
Close #3710 
Close #3487 

Pass only a minimal `$PATH` and `$HOME` to the subprocess in which the batch system job submission command is executed.

This ensures that local and remote jobs are consistent in not seeing the scheduler environment, and in using the `cylc` wrapper.

~~Draft PR until agreed that we should do this or not, and (if yes)  exactly what minimal env should be passed to jobs.~~

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Appropriate tests are included (unit and/or functional).
- [x] Appropriate change log entry included.
- [x] (master branch) I have opened a documentation Issue at https://github.com/cylc/cylc-doc/issues/152
- [x] No dependency changes.

PR #3955, opened 2020-11-20
Clean job submission environment.

### LINKED ISSUE #3487: PYTHONPATH & PYTHONUNBUFFERED set in cylc jobs
With cylc7, job runs via `background`, `at` or `slurm` have the following environments variables set:
`PYTHONUNBUFFERED=true`
`PYTHONPATH=/path/to/cylc/lib:/path/to/cylc/lib` (same path twice!)
(The setting of `PYTHONPATH` caused a user's python script to fail because it picked up the version of Jinja2 provided with cylc rather than the version they needed.)
I assume it happens because the jobs inherit the environment from the `cylc jobs-submit` command.

This doesn't happen for jobs run via PBS.
It also doesn't happen if you use the `--export=NONE` directive with Slurm.

Ideally, these variables should not be set and the behaviour should not vary with the submission method. As a minimum we should document the behaviour.

With cylc8 it seems that `PYTHONUNBUFFERED` is still set but not `PYTHONPATH`.

See also #3168.

### LINKED ISSUE #3710: Local jobs inherit scheduler environment
Back when we used a fixed multi-processing pool for job submission etc.:
- local background jobs were entirely divorced from the scheduler environment, just like remote and batch system jobs.

Now we just manage sub-processes (up to some max size), and:
- local background jobs inherit the scheduler environment just like normal shell sub-processes, unlike batch system jobs.

This is inconsistent and potentially dangerous, because:
- it could cause jobs to behave differently depending on choice of job launcher (local background vs all other)
- for Cylc 8 it also causes local background jobs, uniquely, to bypass the central `cylc` wrapper, because they inherit the Python or conda virtual environment from the scheduler.

Bypassing the wrapper is not in itself bad, but it is (again) inconsistent, and it makes it harder to get the wrapper right because most testing uses local background jobs, and users might reasonably assume that the wrapper is used by all jobs. 

@cylc/core - should we try to fix this?

hjoliver:
Close #3710 
Close #3487 
Supersede #3712 

Global config platform settings to divorce job submission subprocesses from the parent environment (which for local jobs, means the scheduler environment). 

The "impasse" referred to here https://github.com/cylc/cylc-flow/pull/3712#issuecomment-678092052 was resolved (duh) by having the scheduler read relevant global config and pass the `clean_env` flag and associated environment variable names and executable paths to the job host by new `cylc jobs-submit` command line options.

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -33,6 +33,18 @@ from cylc.flow.parsec.validate import (
 # - value_type: value type (compulsory).
 # - default: the default value (optional).
 # - allowed_2, ...: the only other allowed values of this setting (optional).
+
+# Standard executable search paths to pass to job submission subprocesses.
+SYSPATH = [
+    '/bin',
+    '/usr/bin',
+    '/usr/local/bin',
+    '/sbin',
+    '/usr/sbin',
+    '/usr/local/sbin'
+]
+
+
 with Conf('global.cylc', desc='''
     The global configuration which defines default Cylc Flow settings
     for a user or site.
@@ -548,6 +560,40 @@ with Conf('global.cylc', desc='''
                        install target = localhost
             ''')
 
+            Conf('clean job submission environment', VDR.V_BOOLEAN, False,
+                 desc='''
+                Job submission subprocesses inherit their parent environment by
+                default. So remote job submissions inherit the default
+                non-interactive shell environment, but local ones inherit the
+                scheduler environment. This means local jobs see the scheduler
+                environment unless the local batch system prevents it, which
+                can cause problems - e.g. scheduler ``$PYTHON...`` variables
+                can affect Python programs executed by task job scripts. For
+                consistent handling of local and remote jobs a clean job
+                submission environment is recommended, but it is not the
+                default because it prevents local task jobs from running unless
+                the ``cylc`` version selection wrapper script is installed in
+                ``$PATH`` (a clean environment prevents local jobs from seeing
+                the scheduler's virtual environment).
+
+                Specific environment variables can be singled out to pass
+                through to the clean environment, if necessary.
+
+                A standard set of executable paths is passed through to clean
+                environments, and can be added to if necessary.
+            ''')
+
+            Conf('job submission environment pass-through', VDR.V_STRING_LIST,
+                 desc='''
+                Minimal list of environment variable names to pass through to
+                job submission subprocesses. $HOME is passed automatically.
+                You are unlikely to need this.
+            ''')
+            Conf('job submission executable paths', VDR.V_STRING_LIST, desc='''
+                Additional executable locations to pass to the job
+                submission subprocess beyond the standard locations''' +
+                 ', '.join(SYSPATH) + '''. You are unlikely to need this.
+            ''')
         with Conf('localhost', meta=Platform):
             Conf('hosts', VDR.V_STRING_LIST, ['localhost'])
 
--- a/tests/flakyfunctional/job-submission/19-chatty.t
+++ b/tests/flakyfunctional/job-submission/19-chatty.t
@@ -14,19 +14,25 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#-------------------------------------------------------------------------------
+#------------------------------------------------------------------------------
 # Test job submission with a very chatty command.
 # + Simulate "cylc jobs-submit" getting killed half way through.
+
 export REQUIRE_PLATFORM='runner:at'
+
 . "$(dirname "$0")/test_header"
 set_test_number 15
 
-create_test_global_config "" "
-[scheduler]
-    process pool timeout = PT10S
+# This test relies on jobs inheriting the scheduler environment: the job
+# submission command bin/talkingnonsense reads COPYING from $CYLC_REPO_DIR
+# and writes to $CYLC_SUITE_RUN_DIR.
+
+create_test_global_config "
+process pool timeout = PT10S
 [platforms]
     [[$CYLC_TEST_PLATFORM]]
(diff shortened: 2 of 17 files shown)


cylc/flow/cfgspec/globalcfg.py
```diff
@@ -457,6 +469,40 @@
...
+                default. So remote job submissions inherit the default
+                non-interactive shell environment, but local ones inherit the
+                scheduler environment. This means local jobs see the scheduler
+                environment unless the local batch system prevents it, which
+                can cause problems - e.g. scheduler ``$PYTHON...`` variables
+                can affect Python job applications). For consistent handling of
+                local and remote jobs a clean job submission environment is
+                recommended, but it is not the default because it prevents
+                local task jobs from running unless the ``cylc`` version
+                selection wrapper script is installed in ``$PATH`` (a clean
```
oliver-sanders, line 484: Python job script!

hjoliver: Ah, what's your point?

hjoliver: Are you referring to 
> can affect Python job applications

I should change that to "Python programs executed in task job scripts"

hjoliver: (Done)

hjoliver:
## Question

Should we make a clean job environment compulsory, or at least make it the default?

Pros:
- consistent treatment of local and remote jobs
- safety: avoid undesirable interaction between scheduler and job environment in local task jobs, e.g.
  - scheduler PYTHON variables affecting python scripts run by jobs
  - cylc wrapper fork bomb caused scheduler CONDA variables https://github.com/cylc/cylc-flow/pull/3954#issuecomment-733430758 (worked around for now, but it illustrates the potential for problems)

Cons:
- central cylc wrapper required to run even local jobs

oliver-sanders:
One to add to the cons list is additional barrier to Cylc8 migration, sadly we have found cases that are relying on this functionality :vomiting_face:.

hjoliver:
OK I'll leave it as-is (default to dirty env)

hjoliver:
This one's good to go now -  one bug corrected, and tests fixed (I hope...)

cylc/flow/batch_sys_manager.py
```diff
@@ -637,7 +645,24 @@ def _job_submit_impl(
...
+            # (Note this runs on the job host).
+            env = os.environ
+        else:
+            # $HOME is required by job.sh on the job host.
+            env = {'HOME': os.environ.get('HOME', '')}
+        # Pass selected extra variables to the job submit subprocess.
+        for var in self.env:
+            env[var] = os.environ.get(var, '')
+        if self.path is not None:
+            # append to avoid overriding (e.g.) Python in inherited venvs.
```
kinow, line 659: `(e.g. Python in inherited venvs)` nit-pick :smile:

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -545,6 +557,40 @@
...
+                default because it prevents local task jobs from running unless
+                the ``cylc`` version selection wrapper script is installed in
+                ``$PATH`` (a clean environment prevents local jobs from seeing
+                the scheduler's virtual environment).
+
+                Specific environment variables can be singled out to pass
+                through to the clean environment, if necessary.
+
+                A standard set of executable paths is passed through to clean
+                environments, and can be added to if necessary.
```
kinow, line 580: :clap: this combined with the autodoc generation for cylc-doc provides an excellent documentation for Cylc. I read aloud while reviewing, and it is really simple to understand what's implemented here. Thanks!! :100:

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -545,6 +557,40 @@
...
+
+            Conf('job submission environment pass-through', VDR.V_STRING_LIST,
+                 desc='''
+                Minimal list of environment variable names to pass through to
+                job submission subprocesses. $HOME is passed automatically.
+                You are unlikely to need this.
+            ''')
+            Conf('job submission executable paths', VDR.V_STRING_LIST, desc='''
+                Additional executable locations to pass to the job
+                submission subprocess beyound the standard locations''' +
```
kinow, line 591: s/beyound/beyond

tests/flakyfunctional/job-submission/19-chatty.t
```diff
@@ -14,19 +14,25 @@
...
 export REQUIRE_PLATFORM='batch:at'
+
 . "$(dirname "$0")/test_header"
 set_test_number 15
 
-create_test_global_config "" "
-[scheduler]
-    process pool timeout = PT10S
+# This test relies on jobs inheriting the scheduler environment: the job
+# submission command "bin/talkingnonsense reads COPYING from $CYLC_REPO_DIR
```
kinow, line 27: The `"` here appears to be missing a closing `"`?

kinow (approved):
Looks good to me. One typo and a nit-pick comment :smile: (that can be ignored).

@hjoliver from what I understood, this PR modifies a little the environment variables passed to jobs from the workflow scheduler. Including `PATH`. Q/ is there any chance of this causing issues with that cylc wrapper script you wrote a few days ago?

I think it should still work the same, but just checking if there's any risk and if we need to test that here too?

Thanks!

hjoliver:
> @hjoliver from what I understood, this PR modifies a little the environment variables passed to jobs from the workflow scheduler. Including PATH. Q/ is there any chance of this causing issues with that cylc wrapper script you wrote a few days ago?

No that's all good. In fact you need this branch, with clean job env switched on, to test the wrapper properly.  That was part of the motivation for making this branch.

kinow (approved):

wxtim (approved):
- [x] Read and sanity checked.
- [x] Tests pass locally.
- [x] Played with change locally: Failed to make it behave unexpectedly.

PR #3956, opened 2020-11-20
scan: ping

oliver-sanders:
* Add a `--ping` option to `cylc scan` which forces a connection to the running workflow.
* Ensure that `cylc scan --ping` will remove a contact file from a crashed flow.

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/scripts/scan.py
+++ b/cylc/flow/scripts/scan.py
@@ -215,9 +215,11 @@ def get_option_parser():
     parser.add_option(
         '--ping',
         help=(
-            'Test the connection to the flow.'
-            ' (scanning will use the filesystem where possible, --ping forces'
-            ' a connection to the flow)'
+            'Test the connection to the flow. Scan normally just reads flow'
+            ' contact files, but --ping forces a connection to the scheduler'
+            ' and removes the contact file if it is not found to be running'
+            " (this can happen if the scheduler gets killed and can't clean"
+            ' up after itself).'
         ),
         action='store_true'
     )
--- a/cylc/flow/scripts/scan.py
+++ b/cylc/flow/scripts/scan.py
@@ -212,6 +212,16 @@ def get_option_parser():
         action='store_true'
     )
 
+    parser.add_option(
+        '--ping',
+        help=(
+            'Test the connection to the flow.'
+            ' (scanning will use the filesystem where possible, --ping forces'
+            ' a connection to the flow)'
+        ),
+        action='store_true'
+    )
+
     return parser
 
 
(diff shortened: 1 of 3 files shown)


cylc/flow/scripts/scan.py
```diff
@@ -424,6 +434,10 @@ def get_pipe(opts, formatter, scan_dir=None):
     # add graphql queries / filters to the pipe
     if show_active and graphql_fields:
         pipe |= graphql_query(graphql_fields, filters=graphql_filters)
+    elif opts.ping:
+        # check the flow is running even if not required
+        # by display format or filters
+        pipe |= graphql_query({'status': None})
```
hjoliver, line 440: :tada:

cylc/flow/scripts/scan.py
```diff
@@ -212,6 +212,16 @@ def get_option_parser():
         action='store_true'
     )
 
+    parser.add_option(
+        '--ping',
+        help=(
+            'Test the connection to the flow.'
+            ' (scanning will use the filesystem where possible, --ping forces'
+            ' a connection to the flow)'
```
hjoliver, line 220:
```suggestion
            'Test the connection to the flow. Scan normally just reads flow contact'
            ' files, but --ping forces a connection to the scheduler and removes the'
            ' contact file if it is not found to be running (this can happen if the'
            ' scheduler gets killed and can't clean up after itself).'
```
(Best to be explicit about what's going on?).

hjoliver (approved):
(One minor command help suggestion)

datamel (approved):
Great change. I have read through the code and am happy with it. I have checked out your branch, run tests locally and they pass. 
I have manually tested the change by creating 2 fake workflows - one with a fake empty contact file (cylc-scan-test-ping), and one with a fake contact file with required fields CYLC_SUITE_HOST, CYLC_SUITE_PORT (scan-ping).
I then ran cylc scan with --ping and regular cylc scan. The results are below and behaviour is expected.  Note regular cylc scan errors with traceback on cylc-scan-test-ping - this traceback error doesn't need fixing as the only way for users to get this is if they meddle with the contact file which I believe is unlikely. 
Only one note is that --ping does not remove contact file on hosts it can't connect to because authentication keys are missing - again I think this is unlikely to happen but could remove contact file on this error I guess?
image

oliver-sanders:
> Only one note is that --ping does not remove contact file on hosts it can't connect to because authentication keys are missing - again I think this is unlikely to happen but could remove contact file on this error I guess?

Hmmm, interesting. The underlying functionality here is in `detect_old_contact_file` in `suite_files`. I expect it may need a quick rethink at some point.

datamel:
> Hmmm, interesting. The underlying functionality here is in `detect_old_contact_file` in `suite_files`. I expect it may need a quick rethink at some point.
Yes, possibly somthing I shuold have included in the key work. I guess out of scope with this one. Again, I don't _think_ this is going to be a major issue for users as they would probably have to mess with keys to get this as an error.

PR #3957, opened 2020-11-20
Workflow config changes: expected task failures

MetRonnie:
- `[cylc][reference test]expected task failures` -> `[scheduler][events]expected task failures`

**Requirements check-list**
- [x] No changelog entry needed (entry will be written for completed #3696 changes before next release)
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/cfgspec/suite.py
+++ b/cylc/flow/cfgspec/suite.py
@@ -272,6 +272,10 @@ with Conf(
             Conf('abort on timeout', VDR.V_BOOLEAN)
             Conf('abort on inactivity', VDR.V_BOOLEAN)
             Conf('mail events', VDR.V_STRING_LIST, None)
+            Conf('expected task failures', VDR.V_STRING_LIST, desc='''
+                (For Cylc developers writing a functional reference test
+                only) List of tasks that are expected to fail in the test.
+            ''')
 
         with Conf('mail'):
             Conf('footer', VDR.V_STRING, desc='''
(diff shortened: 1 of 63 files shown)


cylc/flow/cfgspec/suite.py
```diff
@@ -1361,6 +1349,11 @@ def upg(cfg, descr):
     u.obsolete('8.0.0', ['cylc', 'events', 'abort if any task fails'])
     u.obsolete('8.0.0', ['cylc', 'events', 'mail retry delays'])
     u.obsolete('8.0.0', ['cylc', 'disable automatic shutdown'])
+    u.deprecate(
```
wxtim, line 1352: Since this config key is "for internal use only" perhaps we should have no upgraders/obsetors? Should we just ensure that there are no tests which use the old syntax?

cylc/flow/suite_status.py
```diff
@@ -60,7 +60,7 @@ class StopMode(Enum):
     """Suite has reached a state where it can automatically stop"""
 
     AUTO_ON_TASK_FAILURE = 'AUTOMATIC(ON-TASK-FAILURE)'
-    """A task has failed and ``abort if any task fails = True``."""
+    """A task has failed and ``--abort-if-any-task-fails`` was given."""
```
wxtim, line 63:
```suggestion
   """A task has failed and ``--abort-if-any-task-fails`` was used."""
```
Really just a suggestion. Ignore if you don't like..

wxtim (approved):
- [x] Read and commented - I've left some comments, but none are fatal objections to merging. 👍🏻 
- [x] Run tests locally. Will tick when it's done, but not expecting problems.

cylc/flow/cfgspec/suite.py
oliver-sanders: You're quite right, there is no reason for us to provide back-compat here.

oliver-sanders (approved):

Issue #3887, opened 2020-10-22
cylc clean


oliver-sanders:
A new command for housekeeping workflows and their files.

1. Remove stopped workflows on the local scheduler filesystem.
   * *target: 8.0.0*
2. Remove stopped workflows on all filesystems.
   * *target: 8.0.0*
   * Obtain a unique list of platforms from the `task_jobs` table in the database.
   * Use the global config to reduce this to an install target mapping `{install_target: [platform, ...]}`.
   * Shuffle the lists of platforms to randomise them.
   * Attempt to remove workflow files using the first platform for each install target.
   * Use host selection as normal for the platform.
   * If it fails try another host, if that fails move on to the next platform.
3. A targeted version of (1) & (2) e.g. delete just the `log` directory.
   * *target: may be needed for 8.0.0 else 8.x*
   * An extension of (1) and (2) to all more targeted removal of dirs within a workflow.
4. Cycle aware housekeeping of `log`, `work` and `share`.
   * *target: 8.x*
   * An extension of (3) which enables the removal of task files from cycles before `x`.
   * A direct replacement of `rose_prune` functionality.
   * Should be able to target tasks and cycle point ranges.
   * Intended for use in housekeep tasks, should detect and use the `CYLC_TASK_CYCLE_POINT` variable.
   * Closes #1159

oliver-sanders:
Note: @dpmatthews has suggested that we might not want to expose the "cycle aware housekeeping" via the CLI which makes some sense as it would be nicer to configure housekeeping in the workflow rather than having a dedicated housekeep task. So (4) might not be related to the CLI, however, we would expect it to share the same logic as the `cylc clean` command.

MetRonnie:
For part 1,
> this will fail if run on a host which doesn't have access to that filesystem (i.e. no SSH logic required).

Without checking the DB, `cylc clean foo` will remove `~/cylc-run/foo` on localhost, without knowing about any remote installations. I.e., it will appear to succeed. Is this okay?

oliver-sanders:
It's an acceptable half-way-house until part (2) is implemented.

MetRonnie:
From part 2:
>    * Obtain a list of platforms used platforms from the `task_jobs` table in the database.
>    * Use the global config to reduce this to an install target mapping `{install_target: [platform, ...]}`.

[...]

PR #3961, opened 2020-11-24
cylc clean - initial implementation

MetRonnie:
These changes partially address #3887
> 1. Remove stopped workflows on the local scheduler filesystem

Implements a new command
```
$ cylc clean workflow_name
```

Note: Cleaning remote directories will come in a later PR.

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/scripts/clean.py
+++ b/cylc/flow/scripts/clean.py
@@ -24,8 +24,8 @@ NOTE: this command is intended for workflows installed with `cylc install`. If
 this is run for a workflow that was instead written directly in ~/cylc-run and
 not backed up elsewhere, it will be lost.
 
-It will also remove an symlink directory targets. For now, it will fail if
-run on a host which doesn't have access to that filesystem.
+It will also remove an symlink directory targets. For now, it will fail to
+remove workflow files/directories on a remote host.
 
 Suite names can be hierarchical, corresponding to the path under ~/cylc-run.
 
--- a/cylc/flow/scripts/clean.py
+++ b/cylc/flow/scripts/clean.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python3
+
+# THIS FILE IS PART OF THE CYLC SUITE ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""cylc clean [OPTIONS] ARGS
+
+Remove a stopped workflow from the local scheduler filesystem.
+
+NOTE: this command is intended for workflows installed with `cylc install`. If
+this is run for a workflow that was instead written directly in ~/cylc-run and
+not backed up elsewhere, it will be lost.
+
+It will also remove an symlink directory targets. For now, it will fail if
+run on a host which doesn't have access to that filesystem.
+
+Suite names can be hierarchical, corresponding to the path under ~/cylc-run.
+
+Examples:
+  # Remove the workflow at ~/cylc-run/foo
+  $ cylc clean foo
+
+"""
+
+from cylc.flow.option_parsers import CylcOptionParser as COP
+from cylc.flow.terminal import cli_function
+from cylc.flow.suite_files import clean
+
+
+def get_option_parser():
+    parser = COP(
+        __doc__,
+        argdoc=[("REG", "Suite name")]
+    )
+    return parser
+
+
+@cli_function(get_option_parser)
+def main(parser, opts, reg):
+    clean(reg)
+
+
+if __name__ == "__main__":
+    main()
--- a/cylc/flow/suite_files.py
+++ b/cylc/flow/suite_files.py
@@ -28,8 +28,9 @@ import zmq.auth
 import aiofiles
 
 from cylc.flow import LOG
-from cylc.flow.exceptions import SuiteServiceFileError
-from cylc.flow.pathutil import get_suite_run_dir, make_localhost_symlinks
+from cylc.flow.exceptions import SuiteServiceFileError, WorkflowFilesError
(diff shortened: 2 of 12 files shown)


MetRonnie:
Should the docs update be part of the 7-to-8 guide?

cylc/flow/scripts/clean.py
```diff
@@ -0,0 +1,57 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""cylc clean [OPTIONS] ARGS
+
+Remove a stopped workflow from the local scheduler filesystem.
+
+NOTE: this command is intended for workflows installed with `cylc install`. If
+this is run for a workflow that was instead written directly in ~/cylc-run and
+not backed up elsewhere, it will be lost.
```
MetRonnie, line 25: It's just occurred to me, we can use the `.service/source` symlink to find out if the workflow was written directly into `~/cylc-run`. If it is, maybe we could require a `--force` option to do the clean?

oliver-sanders: The cylc-run directory is intended to be volatile so I wouldn't put thought into handling these cases in any special way.

Also note the source symlink will be moving/changing soon.

cylc/flow/suite_files.py
```diff
@@ -559,6 +554,92 @@ def register(reg=None, source=None, redirect=False):
...
+    Deletes the run dir in ~/cylc-run and any symlink dirs. Note: if the
+    run dir has been manually deleted, it will not be possible to clean the
+    symlink dirs.
+
+    Args:
+        reg (str): workflow name.
+    """
+    _validate_reg(reg)
+    reg = os.path.normpath(reg)
+    if reg.startswith('.'):
```
oliver-sanders, line 569: BTW the suite name validator has you covered here:

https://github.com/cylc/cylc-flow/blob/80bb0503692c3c5292e3980b5f07bda4606aeae1/cylc/flow/unicode_rules.py#L190

MetRonnie: The difference is I'm checking after normalising the path, so you could have `cylc clean foo/../..` which corresponds to `rm $HOME`. However, I guess I could change the order to
```python
reg = os.path.normpath(reg)
_validate_reg(reg)
```

oliver-sanders: I don't think we should accept path manipulations, though I'm unaware of a simple method to block them (testing for `..` doesn't work for all cases).

Probs good enough to ensure that the flow exists within the cylc-run dir:

```
reg = Path(reg)
reg.resolve()
try:
    reg.relative_to(run_dir_path)
except ValueError:
    raise Exception('Flow not found')
```

Anyway, not a problem for this PR.

cylc/flow/suite_files.py
```diff
@@ -559,6 +554,92 @@ def register(reg=None, source=None, redirect=False):
...
+    for name, path in possible_symlinks:
+        if path.is_symlink():
+            # Ensure symlink is pointing to expected directory. If not,
+            # something is wrong and we should abort
+            target = path.resolve()
+            if target.exists() and not target.is_dir():
+                raise WorkflowFilesError(
+                    f'Invalid Cylc symlink directory {path} -> {target}\n'
+                    f'Target is not a directory')
+            expected_end = str(Path('cylc-run', reg, name))
```
oliver-sanders, line 597: `str(Path(` 😆

tests/functional/cylc-clean/00-basic.t
```diff
@@ -0,0 +1,132 @@
...
+
+__TREE__
+
+TEST_NAME="test-dir-tree-pre-clean"
+tree --charset=ascii "${TEST_DIR}/sym-"* > "${TEST_NAME}.stdout"
+# Remove last line of output:
+sed -i '$d' "${TEST_NAME}.stdout"
+# Note: backticks need to be escaped in the heredoc
+cmp_ok "${TEST_NAME}.stdout" << __TREE__
+${TEST_DIR}/sym-cycle
```
oliver-sanders, line 62: Nice human readable test.

oliver-sanders (approved):

tests/functional/cylc-clean/00-basic.t
```diff
@@ -0,0 +1,132 @@
...
+    \`-- cylctb-${CYLC_TEST_TIME_INIT}
+        \`-- f
+            \`-- cylc-clean
+                \`-- ${TEST_NAME_BASE}
+                    \`-- share
+                        \`-- cycle
+${TEST_DIR}/sym-log
+\`-- cylc-run
+    \`-- cylctb-${CYLC_TEST_TIME_INIT}
+        |-- f
```
datamel, line 73: These tests will fail when running tests using the `functional`  directory name, rather than the `f` one.

datamel (commented):
I have read code, checked this branch out and run tests. I have one comment that needs addressing, other than that I'm pretty sure it is good to go!

tests/functional/cylc-clean/00-basic.t
```diff
@@ -62,23 +67,23 @@ cmp_ok "${TEST_NAME}.stdout" << __TREE__
 ${TEST_DIR}/sym-cycle
 \`-- cylc-run
     \`-- cylctb-${CYLC_TEST_TIME_INIT}
-        \`-- f
+        \`-- ${FUNCTIONAL_DIR}
```
oliver-sanders, line 70: Just realised that if another test gets added to this directory (i.e. `01-<name>.t`) then this test will break.

MetRonnie: I tried copying another test file into `tests/functional/cylc-clean/01-something.t` and `00-basic.t` still works

oliver-sanders: If 01-something creates a workflow and the two tests are run together then 01-something will change the output of `tree` based on the order in which the tests are run.

MetRonnie: I tried making a copy of `00-basic.t` called `01-foo.t` that has a slightly different expected tree (no symlink `work` dir or `leave-me-alone` dir) and both pass? Doesn't `purge` take care of removing the old stuff?

oliver-sanders: It does, but only once the test has passed but tests run in parallel.

MetRonnie: Hopefully https://github.com/cylc/cylc-flow/pull/3961/commits/b35b10ba000f18dac5303b9bcd5393ceb1a2c6ed prevents that

tests/functional/cylc-clean/00-basic.t
oliver-sanders: This has been fixed.

datamel (approved):
I'm happy with these changes. Thanks @MetRonnie!

PR #3963, opened 2020-11-25
Replace [scheduling][parameters] with [task parameters]

wxtim:
also: Replace [scheduling][parameter templates] with [task parameters][templates]

**Requirements check-list**
- [x] Does not need tests: Renaming Configs PR
- [x] No change log entry required: Log entry change to be made at the end of the issue.
- ~[x] No documentation update required: Log entry change to be made at the end of the issue.~
-  [ ] cylc-doc PR required

### DIFF
--- a/cylc/flow/cfgspec/suite.py
+++ b/cylc/flow/cfgspec/suite.py
@@ -289,10 +289,10 @@ with Conf(
         '''):
             Conf('<parameter>', VDR.V_STRING, desc='''
                 Default for integer parameters:
-                    ``_p%(p)0Nd``
-                    where ``N`` is the number of digits of the maximum integer
-                    value, e.g. ``foo<run>`` becomes ``foo_run3`` for ``run``
-                    value ``3``.
+                   ``_p%(p)0Nd``
+                   where ``N`` is the number of digits of the maximum integer
+                   value, e.g. ``foo<run>`` becomes ``foo_run3`` for ``run``
+                   value ``3``.
                 Default for non-integer parameters:
                     ``_%(p)s`` e.g. ``foo<run>`` becomes ``foo_top`` for
                     ``run`` value ``top``.
@@ -304,10 +304,10 @@ with Conf(
 
                 .. note::
 
-                    The values of a parameter named ``p`` are substituted for
-                    ``%(p)s``.  In ``_run%(run)s`` the first "run" is a string
-                    literal, and the second gets substituted with each value of
-                    the parameter.
+                   The values of a parameter named ``p`` are substituted for
+                   ``%(p)s``.  In ``_run%(run)s`` the first "run" is a string
+                   literal, and the second gets substituted with each value of
+                   the parameter.
             ''')
 
     with Conf('scheduling', desc='''
--- a/cylc/flow/cfgspec/suite.py
+++ b/cylc/flow/cfgspec/suite.py
@@ -203,46 +203,11 @@ with Conf(
             with Conf('<plugin name>'):
                 Conf('interval', VDR.V_INTERVAL)
 
-        with Conf('parameters', desc='''
-            Define parameter values here for use in expanding
-            :ref:`parameterized tasks <User Guide Param>`.
-        '''):
-            Conf('<parameter>', VDR.V_PARAMETER_LIST, desc='''
-                Examples:
-                - ``run = control, test1, test2``
-                - ``mem = 1..5``  (equivalent to ``1, 2, 3, 4, 5``).
-                - ``mem = -11..-7..2``  (equivalent to ``-11, -9, -7``).
-            ''')
+        with Conf('simulation'):
+            Conf('disable suite event handlers', VDR.V_BOOLEAN, True)
 
-        with Conf('parameter templates', desc='''
-            Parameterized task names are expanded, for each parameter value,
-            using string templates.
-
-            You can assign templates to parameter names here to override the
-            default templates.
-        '''):
-            Conf('<parameter>', VDR.V_STRING, desc='''
-                Default for integer parameters:
-                   ``_p%(p)0Nd``
-                   where ``N`` is the number of digits of the maximum integer
-                   value, e.g. ``foo<run>`` becomes ``foo_run3`` for ``run``
-                   value ``3``.
-                Default for non-integer parameters:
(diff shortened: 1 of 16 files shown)


cylc/flow/config.py
```diff
@@ -271,12 +271,17 @@ def __init__(
             self.cfg['runtime']['root'] = OrderedDictWithDefaults()
 
         try:
-            parameter_values = self.cfg['scheduler']['parameters']
+            # Ugly hack to avoid templates getting included in parameters
+            parameter_values = {
+                key: value for key, value in
+                self.cfg['task parameters'].items()
+                if key != 'templates'
+            }
```
wxtim, line 279: Because `[templates]` is a subsection of `[task parameters]` we have to specifically remove them from the parameter values dict. 
There is a potential bug here: If a user were to have a template variable called "templates" (Not inconceivable?) then this would be removed leading to a hard to debug error. 
Question: Do we:
- Hope this problem never arises.
- Document this and forget about it.
- Add a warning if key == 'templates'.
- Prevent (raise error) if parameter_values includes templates that isn't a section. (I don't know if Parsec will do this cleverly (@hjoliver ?) but this is my favoured option).

MetRonnie: > Make the test at line 278 more specific using type of value as well as value of key.

Seems doable to me, something like:
```python
parameter_values = self.cfg['task parameters']
if 'templates' in parameter_values and isinstance(parameter_values['templates'], OrderedDict):
    parameter_values.pop('templates')
```
(Not tested)

MetRonnie:
Is there not a functional test that involves task parameters _templates_? Didn't see one when reviewing the changes to the tests.

cylc/flow/config.py
wxtim: > Seems doable to me

Else I'd not have said it, interested in whether other people thought it a good idea....

wxtim:
> Is there not a functional test that involves task parameters _templates_? Didn't see one when reviewing the changes to the 

tests.https://github.com/cylc/cylc-flow/pull/3963/files#diff-6bb572caa59b07ddc8713d4a5bde55443efeabf8335e53cc89f424c6f02fc8a5

MetRonnie:
Ah, tests/functional/param_expand/01-basic.t  does contain templates, must have missed it

cylc/flow/config.py
```diff
@@ -271,14 +271,28 @@ def __init__(
...
         except KeyError:
             parameter_templates = {}
+
+        # Check that parameter templates are a section
+        if not hasattr(parameter_templates, 'update'):
+            raise SuiteConfigError(
+                '[task parameters][templates] is a section. Don\'t use it '
+                'as a parameter.'
+            )
+
```
MetRonnie, line 295: Won't this raise a NameError if `parameter_templates` is not defined? Also, is it not possible to use `isinstance()` instead of `hasattr()`, as the meaning would be clearer?

cylc/flow/config.py
hjoliver: I'm not sure it's reasonable to have a template variable called "templates". It's reasonable to disallow it, because templates work just the same with different variable names (even "templatesx" or whatever if you insist).  However I don't mind to much if we handle it or exclude it.  Maybe handling it is best actually, just to avoid an exclusion that we have to document.

cylc/flow/config.py
wxtim: If `parameter_templates` isn't defined in the config (line 284) then it is set as an empty dictionary at line 287.
I've used `hasattr` because parameter templates could be a dict or an OrderedDictWithDefaults, and this test covers both classes.

MetRonnie: Ah, I am blind!

MetRonnie (approved):

cylc/flow/cfgspec/suite.py
```diff
@@ -305,6 +270,46 @@
...
+
+            You can assign templates to parameter names here to override the
+            default templates.
+        '''):
+            Conf('<parameter>', VDR.V_STRING, desc='''
+                Default for integer parameters:
+                    ``_p%(p)0Nd``
+                    where ``N`` is the number of digits of the maximum integer
+                    value, e.g. ``foo<run>`` becomes ``foo_run3`` for ``run``
+                    value ``3``.
```
oliver-sanders, line 295: Invalid 4-space indentation.

cylc/flow/cfgspec/suite.py
```diff
@@ -305,6 +270,46 @@
...
+                    ``run`` value ``top``.
+
+                Example:
+
+                ``run = -R%(run)s`` e.g. ``foo<run>`` becomes ``foo-R3`` for
+                ``run`` value ``3``.
+
+                .. note::
+
+                    The values of a parameter named ``p`` are substituted for
```
oliver-sanders, line 307: Ditto.

oliver-sanders (commented):
Couple of RST formatting issues, otherwise looks good. Will require a docs PR to fix broken references, code examples, etc.

cylc/flow/cfgspec/suite.py
wxtim, line 295:
```suggestion
                   ``_p%(p)0Nd``
                   where ``N`` is the number of digits of the maximum integer
                   value, e.g. ``foo<run>`` becomes ``foo_run3`` for ``run``
                   value ``3``.
```

cylc/flow/cfgspec/suite.py
```diff
@@ -305,6 +270,46 @@
...
+                    ``%(p)s``.  In ``_run%(run)s`` the first "run" is a string
+                    literal, and the second gets substituted with each value of
+                    the parameter.
```
wxtim, line 310:
```suggestion
                   The values of a parameter named ``p`` are substituted for
                   ``%(p)s``.  In ``_run%(run)s`` the first "run" is a string
                   literal, and the second gets substituted with each value of
                   the parameter.
```

wxtim:
Sorry @oliver-sanders  - I missed your comments on this. Hopefully fixéd now.

MetRonnie (approved):

MetRonnie (approved):

oliver-sanders (approved):

PR #3971, opened 2020-12-01
Add Slurm heterogeneous job support (master)

### LINKED ISSUE #3964: Support Slurm Heterogenous Jobs
**Context**
Slurm 17.11 instroduced the concept of [hetrogenous jobs](https://slurm.schedmd.com/heterogeneous_jobs.html) (a.k.a "hetjobs") where multiple sections of the same job can be run with different directives such as:
```
#SBATCH --nodes=28
#SBATCH hetjob
#SBATCH --nodes=7
#SBATCH hetjob
#SBATCH --nodes=1
```

This cannot currently be done in Cylc due to:

* Parsec overriding duplicate directive keys.
* Cylc job tracking logic (aka `cylc message`).

See proposed implementation outline - https://github.com/cylc/cylc-flow/issues/3964#issuecomment-734361717

**Question:**

* [ ] Are we happy with the proposed implementation outline below?

**Pull requests welcome!**
This is an Open Source project - please consider contributing code yourself
(please read `CONTRIBUTING.md` before starting any work though).

hjoliver:
These changes close #3964 

Can test without Slurm on your dev platform by running the following flow. The lone task will fail job submission, but you can examine the job script.

Global config:
```ini
# .cylc/flow/8.0a3.dev/global.cylc
[platforms]
   [[local]]
      hosts = localhost
      batch system = slurm
```
Flow config:
```ini
[scheduling]
   [[graph]]
         R1 = foo
[runtime]
   [[foo]]
       platform = local
       [[[directives]]]
          --job-name = fooble  # override main job name
          hetjob_0_--mem = 1G  #  first het job
          hetjob_1_--mem = 2G  #  second het job
```

Resulting job script content:
```bash
# DIRECTIVES:
#SBATCH --job-name=fooble
# ...
#SBATCH --mem=1G
#SBATCH hetjob
#SBATCH --mem=2G
```

For older slurm versions use:
- `batch system = slurm_packjob` instead of `slurm`
- directive prefix `packjob_<N>_` instead of `hetjob_<N>_`

**Requirements check-list**

- [x] documentation: self-documenting code module
- [x] No dependency changes.


hjoliver:
## questions

- I presume this approach is OK as a way to provide working support quickly, given that a user has publicly requested this; there's been opportunity for the team to comment on the approach already, and it should look reasonably intuitive to users (if they read the docs). Plus, it works, and altering parsec to allow repeated items without overriding is a bigger job that might have other implications (and is part of wider future plans) 
- is the directive prefix OK?
- is a functional test requiring slurm the best way to test this, or can I use the new integration test framework @oliver-sanders ? (apologies for asking, but it'll save me spending the time to figure it out!)

hjoliver:
Job poll filtering added https://cylc.discourse.group/t/polling-slurm-hetjobs/292

hjoliver:
(Job kill tested, just in case - it works).

oliver-sanders:
Answers:

1) Seems like a good way to provide support quickly, perhaps we can do something more sophisticated later.
2) Can't think of anything better.
3) Probably best to go with a mix of unit+functional testing for this one, integration tests should stop short of job submission / shell interaction.

I don't have access to a slurm platform with hetjob support to evaluate this one. This would be a good time to add slurm images to the docker swarm, however, that may be a larger job.

hjoliver:
> Probably best to go with a mix of unit+functional testing for this one, integration tests should stop short of job submission / shell interaction.

Sorry, I meant to generate a job script and check the content, as per the example above, not actually submit the job.  Would that be suitable for an integration test?

oliver-sanders:
Ah right. Better to use a unit test for this (as there is no need to launch a scheduler). There is a test pattern for writing/testing job file output: https://github.com/cylc/cylc-flow/tree/master/tests/unit/batch_sys_handlers

hjoliver:
Assigned same reviewers as the 7.8.x version.  This one isn't urgent, but might be best to knock it off while fresh in the mind.

oliver-sanders:
Just noticed that slurm tests were skipping, have fixed in https://github.com/cylc/cylc-flow/pull/3977.

Functional tests all pass for me on my available test batch systems (at, slurm):

```console
$ ctb -p '*' -j 4 $(git grep --name-only 'batch:' -- tests)
```

cylc/flow/batch_sys_handlers/slurm_packjob.py
```diff
@@ -0,0 +1,39 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""SLURM job submission and manipulation.
```
oliver-sanders, line 18: Please add this new batch sys handler to `cylc/flow/batch_sys_handlers/__init__.py`, it will then get auto-documented.

oliver-sanders (approved):
Looks good as far as my hetjob knowledge goes (which isn't very), all tests pass on our slurm system.

cylc/flow/batch_sys_handlers/slurm_packjob.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python2
+
```
hjoliver, line 2:
```suggestion
```

Oops, cut and paste error from 7.8.x

hjoliver:
@wxtim - swapping you in as 2nd reviewer (it's the same as the 7.8.x slurm PR you just reviewed)

cylc/flow/batch_sys_handlers/slurm.py
```diff
@@ -51,6 +51,43 @@
...
+       [[my_task]]
+           [[[job]]]
+               batch system = slurm
+               execution time limit = PT1H
+           [[[directives]]]
+               --account = QXZ5W2
+               hetjob_0_--mem = 1G  # first prefix must be "0"
+               hetjob_0_--nodes = 3
+               hetjob_1_--mem = 2G
+               hetjob_0_--nodes = 6
```
wxtim, line 71: is this deliberate, or should it be 
```suggestion
               hetjob_1_--nodes = 6
```

wxtim (approved):
I've made one comment about the docstring. Otherwise it's fine.

cylc/flow/batch_sys_handlers/slurm.py
wxtim: @hjoliver  - can we get this changed or not then we can close?

hjoliver: Sorry Tim, I had missed your comment. Applied your change.

PR #3974, opened 2020-12-01
templatevars: parse CLI opts with literal_eval

oliver-sanders:
Apologies, oversight on my part, it turned out to be a three line change rather than the two lines predicted.

For this PR I have upgraded the set-file logic along with the CLI option (hence the unexpected third line), can revert though it would be somewhat inconsistent. If it is a problem then we could potentially use a different CLI opt for specifying type variables.

> **Rationale (discussed at meeting)**
> * We would expect the `-s`, `--set` option to be used interactively so the change shouldn't impact user scripts.
> * And if it does the CLI has changed anyway.
> * Set-files may be more controversial, however, I don't think these are very widely used.

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -59,6 +59,11 @@ symlink to `DIR` (see #3884).
 
 ### Enhancements
 
+[#3974](https://github.com/cylc/cylc-flow/pull/3974) - Template variables,
+both in set files and provided via the -s/--set command line options are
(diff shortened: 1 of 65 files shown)


hjoliver:
Hahaha :rofl:  Quickest PR ever.

hjoliver (approved):
LGTM, tested as working :+1:

wxtim (approved):

Issue #3706, opened 2020-07-23
Execution polling intervals not set correctly


dpmatthews:
The execution polling intervals are not set correctly when you specify a list of intervals in combination with an execution time limit.

Example suite:
```
[scheduling]
    [[dependencies]]
        graph = root
[runtime]
    [[poll]]
        [[[job]]]
            execution polling intervals = PT30S, PT1H
    [[limit]]
        [[[job]]]
            execution time limit = PT30M
    [[limit_poll]]
        [[[job]]]
            execution time limit = PT30M
            execution polling intervals = PT30S, PT1H
```
Relevant suite log entries:
```
[poll.1] -health check settings: execution timeout=None, polling intervals=PT30S,PT1H,...
```
This looks fine.
```
[limit.1] -health check settings: execution timeout=PT40M, polling intervals=PT31M,PT2M,PT7M,...
```
This makes sense (although I'm not sure why the execution timeout is set - don't think this is documented anywhere).
```
[limit_poll.1] -health check settings: execution timeout=PT40M, polling intervals=60*PT30S,PT1M,PT2M,PT7M,...
```
This is wrong - it looks like only the first item in the execution polling intervals list is being used.

Tested with 7.8.6 & repeated on master.

hjoliver:
Update: this was giving the right answer for the wrong reasons on current master, but will revert to giving the wrong answer for the right reasons once #3981 is merged.

dpmatthews:
Bug fixed in #4844.
Execution timeout issue moved to #4859.

PR #3981, opened 2020-12-03
Minor post-platforms fix.

hjoliver:

Found some problems while looking at #3706 

- `TaskEventsManager` has a method `get_host_conf` that used to extract config items from the deprecated task `remote` and `job` sections, defaulting to global config.  It's broken because **some** deprecated `job` items get auto-upgraded to top level task settings (so they no longer have a secondary dict key), but `remote` items don't because they've moved to platforms (so in deprecated form they still have a secondary dict key). So I've split these into two separate functions.
- `TaskJobManager` also makes a call to the above event manager method to extract `batch systems` config (`err tailer` etc.) which is now in platforms. And, the `batch_sys_config` dict is not actually used anywhere, so I've removed it.

@wxtim - could you cast your "platforms eye" over this to see I'm on the right track?

**Requirements check-list**
- [x] No dependency changes.


hjoliver:
(Draft until I think about tests)

wxtim:
It looks reasonable - Sorry that I missed doing this.
these look like nicely testable units, but I guess that you probably want to check that over-all functionality hasn't changed in an unexpected way so you want a functional test?

cylc/flow/task_events_mgr.py
```diff
@@ -252,18 +252,35 @@ def check_job_time(self, itask, now):
...
-            ret = itask.tdef.rtconfig[skey][key]
+        SKEY = 'remote'
+        if SKEY in overrides and overrides[SKEY].get(key) is not None:
+            ret = overrides[SKEY][key]
+        elif itask.tdef.rtconfig[SKEY].get(key) not in (None, []):
+            ret = itask.tdef.rtconfig[SKEY][key]
         else:
-            try:
-                ret = itask.platform[key]
-            except (KeyError, ItemNotFoundError):
```
hjoliver, line 265: No way to get `ItemNotFoundError` here

cylc/flow/task_events_mgr.py
```diff
@@ -252,19 +252,27 @@ def check_job_time(self, itask, now):
...
+        """Get deprecated "[remote]" items that default to platforms."""
         overrides = self.broadcast_mgr.get_broadcast(itask.identity)
-        if skey in overrides and overrides[skey].get(key) is not None:
-            ret = overrides[skey][key]
-        elif itask.tdef.rtconfig[skey].get(key) not in (None, []):
-            ret = itask.tdef.rtconfig[skey][key]
-        else:
-            try:
-                ret = itask.platform[key]
-            except (KeyError, ItemNotFoundError):
```
hjoliver, line 265: There's no way to get `ItemNotFoundError` here.  And `KeyError` can only occur if an illegal config item is requested.

tests/unit/test_task_events_mgr_2.py
```diff
@@ -0,0 +1,105 @@
...
+
+from unittest.mock import Mock
+import pytest
+
+
+from cylc.flow.task_events_mgr import TaskEventsManager
+
+
+@pytest.mark.parametrize(
+    "broadcast, remote, platforms, expected",
```
kinow, line 25: Nothing wrong here, but my OCD makes me keeping looking at the parameter names are out of order here and in the function declaration🤣

kinow (approved):
Tests look good to me! LGTM

tests/unit/test_task_events_mgr_2.py
hjoliver: Good spotting. I had reordered the parameters at one point but forgot to do the same for the function args. It doesn't seem to matter, but I fixed it anyway.

kinow:
One functional test job failing. Strange. Could it be similar issue to that from my PR fixing unused imports??

hjoliver:
Damn, it passed before my final trivial commit.

hjoliver:
It is a flaky one. I've restarted the tests.

kinow:
> It is a flaky one. I've restarted the tests.

Two jobs failing now :smile:

wxtim (approved):
Oh dear - tests look flaky this morning

hjoliver:
The test failure here appears to be genuine, but is also happening on master (local testing): 
```
tests/f/cylc-poll/04-poll-multi-hosts.t (Wstat: 0 Tests: 3 Failed: 2)
  Failed tests:  2-3
```
So I'll merge this and try to diagnose the problem on master.

PR #3982, opened 2020-12-07
Fix failure to shutdown on keyboard interrupt

### LINKED ISSUE #3946: Scheduler keeps running after CTRL + C (Py3.8)
**Describe the bug**

Supersedes #3490 , which has similar symptoms. The Scheduler keeps running after a CTRL + C, and a simple cylc scan shows the workflow as running. See details below for more.

Shamelessly copying @hjoliver great explanation of the issue on Element.

> First Ctrl+C says "Aborted!". The scheduler stays up, logs incoming connections, and is responsive to queries (e.g. cylc dump) BUT it is no longer responsive to commands and does not change state or submit new jobs.
>
> (ps command on the RHS confirms that the scheduler process is still running, although that's kind of obvious from the left side).
>
> Second Ctrl+C does kill the scheduler process, but not cleanly -it does not clean up its contact file like it does in Py 3.7.
> 
> The 2nd ps command on the right confirms that it is dead.
> 
> cylc scan still says it is running, because the minimal scan does not try to contact the running scheduler - it just reads the contact file (and assumes a contact file means scheduler running)
> 
> Then cylc scan -t rich does try to contact the scheduler, and finds that it is not running so it deletes the contact file. Any scheduler-connecting command would do the same - this is why your cylc kill five updated the scan result, not because it actually killed something.

 `master` with Python 3.8 (works fine in 3.7)

```
cylc run --no-detach five
# hit CTRL +C
# then in another window
cylc scan
# five is displayed as running
cylc scan -t rich
# now five stops, as scan will check the info from the contact file
```

**Expected behavior**

Nicely exit after CTRL + C is hit the first time.

**Screenshots**

**Additional context**

**Pull requests welcome!**

MetRonnie:
These changes close #3946

<kbd>Ctrl</kbd>+<kbd>C</kbd> was failing to shutdown the scheduler on Python 3.8+ due to the fact that `asyncio.CanclledError` changed from subclassing `Exception` to `BaseException`. See https://stackoverflow.com/a/60167684/3217306 and https://docs.python.org/3.8/library/asyncio-exceptions.html#asyncio.CancelledError

I'm new to asyncio, and there seems to be another way of handling signal interruptions: https://www.roguelynn.com/words/asyncio-graceful-shutdowns/. However, the changes I've made are very simple and seem to work.

**Requirements check-list**
- [x] Does not need tests (I'm not sure we can test this?).
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -612,12 +612,7 @@ class Scheduler:
             raise exc from None
 
         except (KeyboardInterrupt, asyncio.CancelledError, Exception) as exc:
-            try:
-                await self.shutdown(exc)
-            except Exception as exc2:
-                # In case of exceptions in the shutdown method itself
-                LOG.exception(exc2)
-            raise exc from None
+            await self.handle_exception(exc)
 
         else:
             # main loop ends (not used?)
@@ -644,8 +639,7 @@ class Scheduler:
             await self.start_servers()
             await self.log_start()
         except (KeyboardInterrupt, asyncio.CancelledError, Exception) as exc:
-            await self.shutdown(exc)
-            raise
+            await self.handle_exception(exc)
         else:
             # note start_scheduler handles its own shutdown logic
             await self.start_scheduler()
@@ -1619,16 +1613,19 @@ class Scheduler:
 
         """
         if isinstance(reason, SchedulerStop):
-            LOG.info('Suite shutting down - %s', reason.args[0])
+            LOG.info(f'Suite shutting down - {reason.args[0]}')
         elif isinstance(reason, SchedulerError):
-            LOG.error('Suite shutting down - %s', reason)
+            LOG.error(f'Suite shutting down - {reason}')
         elif isinstance(reason, SuiteConfigError):
             LOG.error(f'{SuiteConfigError.__name__}: {reason}')
         elif isinstance(reason, PlatformLookupError):
             LOG.error(f'{PlatformLookupError.__name__}: {reason}')
         else:
             LOG.exception(reason)
-            LOG.critical('Suite shutting down - %s', reason)
+            if str(reason):
+                LOG.critical(f'Suite shutting down - {reason}')
+            else:
+                LOG.critical('Suite shutting down')
 
         if self.proc_pool:
             self.proc_pool.close()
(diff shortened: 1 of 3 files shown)


cylc/flow/scheduler_cli.py
```diff
@@ -406,13 +406,7 @@ async def _run(parser, options, reg, is_restart, scheduler):
     # stop cylc stop
     except SchedulerError:
         ret = 1
-    except KeyboardInterrupt as exc:
-        try:
-            await scheduler.shutdown(exc)
-        except Exception as exc2:
-            # In case of exceptions in the shutdown method itself.
-            LOG.exception(exc2)
-            raise exc2 from None
```
MetRonnie, line 415: No need for this anymore as the shutdown is handled in `scheduler.run()` or `scheduler.start_scheduler()`

cylc/flow/scheduler.py
```diff
@@ -643,7 +643,7 @@ async def run(self):
             await self.configure()
             await self.start_servers()
             await self.log_start()
-        except Exception as exc:
+        except (KeyboardInterrupt, asyncio.CancelledError, Exception) as exc:
             await self.shutdown(exc)
```
oliver-sanders, line 647: Might want to copy over the try/except to ensure that shutdown traceback gets dumped to the flow log.

oliver-sanders (approved):
Works for me, tested 3.7/3.8.

> (I'm not sure we can test this?)

It's do-able via bash though fiddly, we used to do this for `cylc scan` (`tests/functional/cylc-scan/02-sigstop.t`).

Should be able to do via Python though, could try writing an integration test and raising KeyboardInterrupt from within the Scheduler, not sure what would happen but worth a try.

```python
async with run(schd) as flow_log:
    raise KeyboardInterrupt()
assert flow_log.record_tuples[-1] == [...]
```

kinow (approved):
Worked like a charm. Switched to this branch, re-installed my venv, started `five`. Stopped it, and it exited nicely :+1: 

Thanks @MetRonnie !

cylc/flow/scheduler.py
kinow: @MetRonnie two approvals, but will let you reply this one before I merge it (happy if you or @oliver-sanders merge it after you've taken a look on this one too :+1: )

MetRonnie:
Having a go at the integration test

MetRonnie:
Couldn't get the integration test to work. It seemed like the `KeyboardInterrupt` caused the test to quit, but I didn't spend too much time looking into it

cylc/flow/scheduler.py
```diff
@@ -1619,16 +1613,19 @@ async def shutdown(self, reason):
 
         """
         if isinstance(reason, SchedulerStop):
-            LOG.info('Suite shutting down - %s', reason.args[0])
+            LOG.info(f'Suite shutting down - {reason.args[0]}')
         elif isinstance(reason, SchedulerError):
-            LOG.error('Suite shutting down - %s', reason)
+            LOG.error(f'Suite shutting down - {reason}')
```
kinow, line 1618: :+1:

kinow (approved):

hjoliver (commented):
Works for me too; thanks @MetRonnie :+1: 

I'll merge this now, a test for this isn't super-critical. We developers are likely to notice quickly if this breaks in future.

PR #3989, opened 2020-12-11
terminal: hack around stupid emacs stty return value

oliver-sanders:
Reported by @dpmatthews, emacs implemets `stty size` but in a broken way:

```console
$ stty size
0 0
```

Return the defined default if we get 0.

Workaround:

`alias emacs=vim`

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/terminal.py
+++ b/cylc/flow/terminal.py
@@ -53,7 +53,7 @@ def get_width(default=80):
         return default
     try:
         return int(proc.communicate()[0].split()[1]) or default
-    except IndexError:
+    except (IndexError, ValueError):
         return default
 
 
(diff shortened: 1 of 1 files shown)


cylc/flow/terminal.py
```diff
@@ -52,8 +52,8 @@ def get_width(default=80):
     if proc.wait():
         return default
     try:
-        return int(proc.communicate()[0].split()[1])
-    except IndexError:
+        return int(proc.communicate()[0].split()[1]) or default
+    except (IndexError, ValueError):
```
oliver-sanders, line 56: Just to be double certain, this will handle return values like `a` or ``.

dpmatthews (approved):
This appears to fix the issue - thanks!

PR #4000, opened 2020-12-16
Cylc install

### LINKED ISSUE #3889: cylc install
New command for installing and registering workflows from arbitrary locations into the cylc-run directory.

As per the [rose suite-run migration proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-rose-suite-run.md).

#### Overview Of The New Command

* Check for the existence of a `flow.cylc` file else fail.
* Configure the run directory as per symlink configuration in globalrc installing the workflow into
  `~/cylc-run/<flow_name>/<run_name>` where `run_name` defaults to `run_<n + 1>`.
* `rsync` the directory (applying some pre-defined exclusions) into the run directory.
* Register the workflow.

#### Out Of Scope

Inexhaustive list of related topics which aren't expected to be resolved as part of this issue:

* Support for recording any `cylc play` command line arguments for use with future commands.
* Combined actions (e.g. `install-play`).
* `cylc run` -> `cylc play` and related CLI changes.
* Re-installation (#3890).
* Recording of version control information (#3849).
* Inference of run number by `cylc` commands when partial registration is provided.
* Support for running other commands against the workflow source (e.g. `cylc get-config`).

#### Registration

Registration (or at least much of its functionality) isn't actually going anywhere but we will retire it as a user-facing concept.

Registration (defined in `cylc.flow.suite_files.register`) currently does the following:

1. Validates workflow name.
2. Validates run directory structure (#3835).
3. Installs workflow definition from specified source.
4. Symlinks `flow.cylc` -> `suite.rc`.
5. Creates the `.service` directory.
6. Creates the `.service/source` symlink.
7. Allows updating the source with `--redirect`.

As part of this work the following changes should be made:

* Remove (3) - registration from arbitrary locations no longer supported, use `cylc install`.
* Remove (6) - source location now specified by `cylc install`.
* Remove (7) - users can manually update the new `_cylc-install/source` symlink for now.
* Remove the `cylc register` command - use `cylc install` or `cylc play` for in-place workflows.

Note that that `cylc play` should still register if not already done as it *should* still be possible to write a workflow directly in the run directory and run it without using `cylc install`.

datamel:
Cylc Install command added. These changes close #3889

Cylc can now install workflows from from arbitrary locations into the cylc-run directory.

This PR includes:
- Removing cylc register command.
- New install log added in log directory: `log/install/<timestamp>-install.log`

**Requirements check-list**
- [x] Appropriate change log entry included. - TODO before ready for review
- [ ] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/XXXX. _TODO
- [x] No dependency changes.


hjoliver:
Looks impressive @datamel :tada:

oliver-sanders:
Dammit I totally forgot we were approaching PR 4000!

wxtim:
> Dammit I totally forgot we were approaching PR 4000!

IMO this is why Mel waited for a few days to put up the draft PR!

cylc/flow/pathutil.py
```diff
@@ -141,10 +143,13 @@ def make_suite_run_tree(suite):
...
 
-def make_localhost_symlinks(suite):
-    """Creates symlinks for any configured symlink dirs from glbl_cfg."""
-    dirs_to_symlink = get_dirs_to_symlink('localhost', suite)
-    rund = get_suite_run_dir(suite)
+def make_localhost_symlinks(rund, named_sub_dir):
+    """Creates symlinks for any configured symlink dirs from glbl_cfg.
+    Args:
+        rund: the entire run directory path
+        named_sub_dir: e.g flow_name/run1 """
```
wxtim, line 150:
```suggestion
        named_sub_dir: e.g flow_name/run1
    Returns:
         A dictionary of Symlinks with sources as keys and destinations as values: ``{source: destination}``
    """
```

oliver-sanders:
The bash docker test failures are nothing to worry about, those tests pass locally, likely that the test configuration requires a bit of a poke.

Looking at the GH actions `cylc-run` directory it looks like the `flow.cylc` files are not in place suggesting that that installation failed somehow. Perhaps there is some useful error that is being swallowed by `2>/dev/null`?

```diff
--- a/tests/functional/lib/bash/test_header
+++ b/tests/functional/lib/bash/test_header
@@ -430,7 +430,7 @@ install_suite() {
     SUITE_RUN_DIR="${RUN_DIR}/${SUITE_NAME}"
     mkdir -p "${TEST_DIR}/${SUITE_NAME}/" # make source dir
     cp -r "${TEST_SOURCE_DIR}/${TEST_SOURCE_BASE}/"* "${TEST_DIR}/${SUITE_NAME}/"
-    cylc install --no-run-name --flow-name="${SUITE_NAME}" --directory="${TEST_DIR}/${SUITE_NAME}" 2>'/dev/null'
+    cylc install --no-run-name --flow-name="${SUITE_NAME}" --directory="${TEST_DIR}/${SUITE_NAME}"
     cd "${TEST_DIR}/${SUITE_NAME}/"
 }
```

oliver-sanders:
The cylc-rose stuff has been merged now, should be straight forward to merge the two `cylc install` commands, just need to ensure the plugins get called at the right point.

`pre_configure` - Before the flow configuration is read.
`post_install` - Last thing that `cylc install` does.

oliver-sanders:
Took a crack, this seems to do the job:

```diff
diff --git a/cylc/flow/scripts/install.py b/cylc/flow/scripts/install.py
index 5011996d2..f773ffdc4 100755
--- a/cylc/flow/scripts/install.py
+++ b/cylc/flow/scripts/install.py
@@ -58,9 +58,13 @@ multiple workflow run directories that link to the same suite definition.
 
 """
 
+import os
+import pkg_resources
+
 from cylc.flow.option_parsers import CylcOptionParser as COP
-from cylc.flow.suite_files import install_workflow
 from cylc.flow.terminal import cli_function
+from cylc.flow.pathutil import get_workflow_run_dir
+from cylc.flow.suite_files import install_workflow
 
 
 def get_option_parser():
@@ -116,14 +120,30 @@ def get_option_parser():
 def main(parser, opts, flow_name=None, src=None):
     if opts.no_run_name and opts.run_name:
         parser.error(
-            """options --no-run-name and --run-name are mutually exclusive.
+            """opts --no-run-name and --run-name are mutually exclusive.
             Use one or the other""")
-    install_workflow(
+
+    for entry_point in pkg_resources.iter_entry_points(
+        'cylc.pre_configure'
+    ):
+        entry_point.resolve()(opts.source)
+
+    flow_name = install_workflow(
         flow_name=opts.flow_name,
         source=opts.source,
         run_name=opts.run_name,
         no_run_name=opts.no_run_name,
-        no_symlinks=opts.no_symlinks)
+        no_symlinks=opts.no_symlinks
+    )
+
+    for entry_point in pkg_resources.iter_entry_points(
+        'cylc.post_install'
+    ):
+        entry_point.resolve()(
+            dir_=opts.source or os.getcwd(),
+            opts=opts,
+            dest_root=get_workflow_run_dir(flow_name)
+        )
 
 
 if __name__ == "__main__":
```

oliver-sanders:
This might be the answer to the bash test failures:

```diff
diff --git a/dockerfiles/bash/Dockerfile b/dockerfiles/bash/Dockerfile
index 8b9f6a2e1..2396924a1 100644
--- a/dockerfiles/bash/Dockerfile
+++ b/dockerfiles/bash/Dockerfile
@@ -13,6 +13,7 @@ RUN apt-get update && apt-get install -y \
     python3.7-dev \
     sqlite3 \
     wget \
+    rsync \
     && python3.7 -m pip install -U pip setuptools \
     && rm -rf /var/lib/apt/lists/*
```

cylc/flow/suite_files.py
```diff
@@ -787,4 +797,239 @@ def get_cylc_run_abs_path(path):
...
+        flow_name = (Path.cwd().stem)
+    validate_flow_name(flow_name)
+    if run_name == '_cylc-install':
+        raise WorkflowFilesError(
+            'Run name cannot be "_cylc-install".'
+            ' Please choose another run name.')
+    validate_source_dir(source)
+    run_path_base = Path(get_workflow_run_dir(flow_name)).expanduser()
+    relink = False
+    run_num = int()
```
oliver-sanders, line 901:
```python
>>> int()
0
```

```suggestion
    run_num = None  # or 0?
```

?

cylc/flow/suite_files.py
```diff
@@ -787,4 +797,239 @@ def get_cylc_run_abs_path(path):
...
+            rsync_cmd.append(f"--exclude={exclude}")
+    if Path(src).joinpath('.cylcignore').exists():
+        rsync_cmd.append("--exclude-from=.cylcignore")
+    rsync_cmd.append(f"{src}/")
+    rsync_cmd.append(f"{dst}/")
+
+    return rsync_cmd
+
+
+def install_workflow(flow_name=None, source=None, run_name=None,
```
oliver-sanders, line 858: Can break this function down a bit, would make it possible to unit test bits of of (e.g. the runN bit).

cylc/flow/suite_files.py
```diff
@@ -509,49 +535,30 @@ def register(reg=None, source=None, redirect=False):
...
+    """Check to see if the path sent contains installed flow.
+
+       Checks for valid _cylc-install directory in current folder and checks
+       source link exists.
+    """
+    cylc_install_folder = Path(path, SuiteFiles.Install.DIRNAME)
+    source = Path(cylc_install_folder, SuiteFiles.Install.SOURCE)
+    if cylc_install_folder.exists and source.is_symlink():
+        return True
+    return False
```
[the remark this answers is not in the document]
MetRonnie: I think that shouldn't matter, as `clean` uses `shutil.rmtree` rather than `os.rmdir`

oliver-sanders:
Something seems to have angered the tests there. Lots of this diff in the logs:

```
-Parsing cylctb-20210112T182706Z/f/cylc-diff/00-basic-2 (/tmp/tmp.yw9DTu8vUR/cylctb-20210112T182706Z/f/cylc-diff/00-basic-2/flow.cylc)
250
+Parsing cylctb-20210112T182706Z/f/cylc-diff/00-basic-1 (/home/runner/cylc-run/cylctb-20210112T182706Z/f/cylc-diff/00-basic-1/flow.cylc)
```

datamel:
> Something seems to have angered the tests there. Lots of this diff in the logs:
> 
> ```
> -Parsing cylctb-20210112T182706Z/f/cylc-diff/00-basic-2 (/tmp/tmp.yw9DTu8vUR/cylctb-20210112T182706Z/f/cylc-diff/00-basic-2/flow.cylc)
> 250
> +Parsing cylctb-20210112T182706Z/f/cylc-diff/00-basic-1 (/home/runner/cylc-run/cylctb-20210112T182706Z/f/cylc-diff/00-basic-1/flow.cylc)
> ```

Just looking into it now - angering the tests is my speciality!

oliver-sanders:
Looking at the timestamps in the logs none of the four cancelled jobs from the last run produced any output in the minutes before being cancelled.

This suggests that some tests are hanging.

Unfortunately it can be quite hard to work out which of the tests are hanging (I have an idea on this I developed for the Rose tests, will try some time) but you can always try locally.

I think some of the tests are hanging which is causing the timeouts.

oliver-sanders (commented):
Given a test locally, works well.

I found one quirk where you can end up with a combination of custom and automatically named runs:

```
$ cylc install --run-name a
$ cylc install
$ tree ~/cylc-run -L 2
~/cylc-run/
└── testy
    ├── _cylc-install
    ├── b
    ├── run1
    └── runN -> ~/cylc-run/testy/run1
```

I think the correct behaviour should be that the second install would fail, we can tell that `b` is an installed flow not a run because `~/cylc-run/testy/b` contains a ` flow.cylc` file.

datamel:
> I think the correct behaviour should be that the second install would fail, we can tell that `b` is an installed flow not a run because `~/cylc-run/testy/b` contains a ` flow.cylc` file.

Are we happy for someone to do a named run after a named run? Is it only mix of named and run numbered installs we want to avoid? 

e.g.
 ```
$ cylc install --run-name a
$ cylc install --run-name b
$ tree ~/cylc-run/testy -L 1

~/cylc-run/testy
|-- _cylc-install
|-- a
`-- b
  
```

**Comment by oliver-sanders** [comment]
Yes, because the meaning of the `runN` symlink gets fuzzy.

File: .github/workflows/test_functional.yml [hunk 5]
```diff
@@ -5,7 +5,7 @@ on: [pull_request]
 jobs:
   test:
     runs-on: ubuntu-latest
-    timeout-minutes: 35
+    timeout-minutes: 50
```
oliver-sanders, line 8: Suggest `40` mins?

tests/functional/cylc-install/00-simple.t
```diff
@@ -0,0 +1,148 @@
...
+set_test_number 20
+
+export RND_SUITE_NAME
+export RND_SUITE_SOURCE
+export RND_SUITE_RUNDIR
+
+function make_rnd_suite() {
+    # Create a randomly-named suite source directory.
+    # Define its run directory.
+    RND_SUITE_NAME=x$(< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c6)
```
oliver-sanders, line 30: BTW, can use `mktemp` to create a dir in `$PWD`:

```
mktemp -d -p "$PWD"
```

But today I learned about `/dev/urandom` so thanks.

oliver-sanders (commented):
Nice tests!

GH actions is angry again, looks like all tests are failing for the same reason at least.

tests/functional/cylc-install/00-simple.t
datamel: Ah I won't take credit for that - 'twas already in the code base I think :)

cylc/flow/suite_files.py
```diff
@@ -926,4 +935,305 @@ def get_cylc_run_abs_path(path):
...
+        suite_rc_path = source.joinpath(SuiteFiles.SUITE_RC)
+        if suite_rc_path.is_file():
+            flow_file_path.symlink_to(suite_rc_path)
+            INSTALL_LOG.warning(
+                f'The filename "{SuiteFiles.SUITE_RC}" is deprecated in favour'
+                f' of "{SuiteFiles.FLOW_FILE}". Symlink created.')
+        else:
+            raise WorkflowFilesError(
+                f'no {SuiteFiles.FLOW_FILE} or {SuiteFiles.SUITE_RC}'
+                f' in {source}')
```
oliver-sanders, line 1082: I think this logic is a duplicate of the `parse_suite_arg`, factor into another function?

oliver-sanders (commented):
Nearly there!

Screenshot 2021-01-26 at 17 09 25

The argument of `cylc install` is a little funny with its flow/path/PWD opts thing, but we can sort that with the universal id changes later.

cylc/flow/suite_files.py
datamel: Thanks, I've refactored.

cylc/flow/suite_files.py
```diff
@@ -428,71 +444,80 @@ async def load_contact_file_async(reg, run_dir=None):
...
-    make_localhost_symlinks(reg)
-
+    if flow_name is None:
+        flow_name = (Path.cwd().stem)
+    is_valid, message = SuiteNameValidator.validate(flow_name)
+    if not is_valid:
+        raise WorkflowFilesError(f'Invalid workflow name - {message}')
+    if Path.is_absolute(Path(flow_name)):
+        raise WorkflowFilesError(
+            f'Workflow name cannot be an absolute path: {flow_name}')
```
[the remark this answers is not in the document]
datamel: Looking at tests, just the not valid name was untested - I have now added a test for this.

oliver-sanders (approved):
Looks good, a lot of nuances and Cylc history has been blasted aside here, nice.

There are some minor quirks I think we should skip over for this review and come back to later:

* The argument of `cylc install` can be reg/path.
  * Come back to this later (universal id stuff).
* The `flow.cylc` is being created in the "source" dir and `rsync`'ed to the "run" dir.
  * It may be better not to create the symlink in the "source" dir.
  * This will silence future warnings.
  * Users should just `mv suite.rc flow.cylc`.
* The runN symlinks are absolute rather than relative.
  * This blocks registration changes, but perhaps that's not the worst thing?

hjoliver:
> Looks good, a lot of nuances and Cylc history has been blasted aside here, nice.

yeah brilliant, thanks @datamel :tada:   I'll have a close look as soon as possible...

hjoliver (approved):
Looks great, no problems found  :tada: :tada:

PR #3992, opened 2020-12-11
batch system -> job runner

MetRonnie:
These changes close #3910. Yikes.

Global config change:
- `[platforms][X]batch system` -> `[platforms][X]job runner`
- `[platforms][X]batch submit command template` -> `[platforms][X]job runner command template`

**Plus:** Change all permutations of `batch system`, `batch_sys` etc to an equivalent `job runner`, apart from where `batch system` refers to the _workflow_ config item `[runtime][X][job]batch system`.

I've tried to _batch_ the changes into somewhat digestible commits for reviewing.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).

- [x] Appropriate tests are included (functional).

- [ ] Appropriate change log entry included.

- [ ] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/XXXX.

- [x] No dependency changes.

PR #4001, opened 2020-12-17
BACK COMPAT: remove old compat and update comments

oliver-sanders:
Follow on from: https://github.com/cylc/cylc-flow/pull/3992#discussion_r542490192

Use a standard marker for indicating backwards compatibility in code. Its a totally arbitrary marker, all that matters is that we standardise and record the important information to avoid some of the issues weve had in the past with back-compat lasting longer than inteded.

In the process fix a few cases of back-compat which *should* have been removed already but was missed!

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/scripts/message.py
+++ b/cylc/flow/scripts/message.py
@@ -105,7 +105,12 @@ def main(parser, options, *args):
         return parser.error('No message supplied')
     cylc.flow.flags.verbose = os.getenv('CYLC_VERBOSE') == 'true'
     cylc.flow.flags.debug = os.getenv('CYLC_DEBUG') == 'true'
-    if len(args) <= 2:  # compat
+    if len(args) <= 2:
+        # BACK COMPAT: args <= 2
+        # from:
+        #     ????
+        # remove at:
+        #     ????
         suite = os.getenv('CYLC_SUITE_NAME')
         task_job = os.getenv('CYLC_TASK_JOB')
         message_strs = list(args)
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -526,7 +526,7 @@ class TaskEventsManager():
                 itask, ("message %s" % str(severity).lower()), message)
         lseverity = str(severity).lower()
         if lseverity in self.NON_UNIQUE_EVENTS:
-            itask.non_unique_events.update(lseverity)
+            itask.non_unique_events.update({lseverity: 1})
             self.setup_event_handlers(itask, lseverity, message)
 
     def _process_message_check(
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -526,8 +526,7 @@ class TaskEventsManager():
                 itask, ("message %s" % str(severity).lower()), message)
         lseverity = str(severity).lower()
         if lseverity in self.NON_UNIQUE_EVENTS:
-            itask.non_unique_events.setdefault(lseverity, 0)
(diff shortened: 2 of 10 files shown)


cylc/flow/scripts/message.py
```diff
@@ -105,7 +105,12 @@ def main(parser, options, *args):
         return parser.error('No message supplied')
     cylc.flow.flags.verbose = os.getenv('CYLC_VERBOSE') == 'true'
     cylc.flow.flags.debug = os.getenv('CYLC_DEBUG') == 'true'
-    if len(args) <= 2:  # compat
+    if len(args) <= 2:
+        # BACK COMPAT: args <= 2
+        # from:
+        #     ????
+        # remove at:
+        #     ????
```
oliver-sanders, line 113: Since the `cylc message -- xyz` interface is this still considered legacy?

cylc/flow/task_pool.py
```diff
@@ -422,8 +422,12 @@ def load_db_task_pool_for_restart(self, row_idx, row):
...
                         itask.state.outputs.set_completion(message, True)
                 except (AttributeError, TypeError, ValueError):
-                    # Back compat for <=7.6.X
-                    # Each output in separate line as "trigger=message"
+                    # BACK COMPAT: outputs in separate lines
+                    # (as "trigger=message")
+                    # from:
+                    #     <=7.6.X
+                    # remove at:
+                    #     ???
```
oliver-sanders, line 430: Thoughts???

cylc/flow/task_events_mgr.py
```diff
@@ -526,8 +526,7 @@ def process_message(
                 itask, ("message %s" % str(severity).lower()), message)
         lseverity = str(severity).lower()
         if lseverity in self.NON_UNIQUE_EVENTS:
-            itask.non_unique_events.setdefault(lseverity, 0)
-            itask.non_unique_events[lseverity] += 1
+            itask.non_unique_events.update(lseverity)
```
hjoliver, line 529:
```suggestion
            itask.non_unique_events.update({lseverity: 1})
```
Fixes the broken test.

cylc/flow/scripts/message.py
hjoliver: The new interface was introduced 3 years ago (cylc-7.6-ish?). I don't mind if we ditch the back-compat now, but there is a pretty high chance the "class interface" is being used somewhere.  Note we still use it ourselves and some tests.

```
$ git grep "cylc message" | grep -v '\-\-' | wc -l
35
```

cylc/flow/task_pool.py
hjoliver: This is in `load_db_task_pool_for_restart()` and we're not support restart from a Cylc 7 DB, so can remove this, right?

cylc/flow/task_events_mgr.py
oliver-sanders: Er, heck, thats confusing, the two should do the same:

```python
>>> from collections import Counter
>>> c = Counter()

>>> c.update({'a': 1})
>>> c
Counter({'a': 1})

>>> c.update('b')
>>> c
Counter({'a': 1, 'b': 1})
```

hjoliver: Try it with a multi-character string.

cylc/flow/scripts/message.py
oliver-sanders: Think I'll leave it as `???` and kick this down the road.

hjoliver (approved):

PR #4011, opened 2020-12-22
Fix trailing slash in reg (workflow name) bug

### LINKED ISSUE #4010: cylc stop doesn't work when reg ends with slash
**TL;DR**

When you do `cylc run`, if you have a slash at the end of the workflow name, e.g. `cylc run foo/`, then `cylc stop foo` won't work, only `cylc stop foo/` will work. And vice versa.

Cylc 8.0a3.dev master

```console
$ cylc run foo --hold
$ cylc stop foo/  # appears to work (no output as is the case of a success)
$ cylc get-suite-contact foo
CYLC_API=5
CYLC_SUITE_PROCESS=1234 /path/to/python3 /path/to/cylc run foo/ --hold
# etc
```

**Expected behavior**
Workflow stops and contact file removed.

**Additional context**
Often when I'm in the cylc-run dir, I use <kbd>tab</kbd> autocompletion to run workflows, stop them etc. But because the workflows are directories, that adds a `/` to the end of the reg (workflow name).

**Pull requests welcome!**

MetRonnie:
These changes close #4010 

It seems that passing a reg with a trailing slash to the `mutations_kwargs` dict was the reason for this bug. As such I've fixed it everywhere I found a `mutations_kwargs` dict.

Does this need a functional test?

**Requirements check-list**

- [ ] Appropriate tests are included (functional).

- [x] No dependency changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -196,6 +196,10 @@ issue when mixing offset and conditional (e.g. foo<m-1> & baz => foo<m>).
 workflow from shutting down properly on a keyboard interrupt (Ctrl+C) in
 Python 3.8+.
 
+[#4011](https://github.com/cylc/cylc-flow/pull/4011) - Fix bug where including
+a trailing slash in the suite/workflow name would cause `cylc stop`
+(and possibly other commands) to silently fail.
+
 -------------------------------------------------------------------------------
 ## __cylc-8.0a2 (2020-07-03)__
 
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -314,7 +314,7 @@ def scheduler_cli(parser, options, args, is_restart=False):
     functionality.
 
     """
-    reg = args[0]
+    reg = os.path.normpath(args[0])
     # Check suite is not already running before start of host selection.
     try:
         suite_files.detect_old_contact_file(reg)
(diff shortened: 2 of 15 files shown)


oliver-sanders:
Did you get to the bottom of why the exact argument specified to `cylc run` mattered?

oliver-sanders (approved):
I think this is harmless and should work for both REG/path uses, `..` in paths now get expanded.

Not so worried about testing as the universal identifier work will centralise this.

MetRonnie:
> Did you get to the bottom of why the exact argument specified to `cylc run` mattered?

I assumed it's something to do with how mutations are handled, seeing as removing the slash before the `mutations_kwargs` dict is set is what seems to fix it

hjoliver:
I'm not sure this is the right solution.

If the registered suite name is "foo", then "foo/" is not the correct name, so `cylc stop foo/` should say `SuiteStopped: foo/ is not running` rather than magically convert `foo/` to `foo`.  (And vice versa).

If that's confusing because it's easier to accidentally register `foo/` than `foo#` or some other character, then we should disallow `/` at the end of a suite name.

MetRonnie:
The trouble with that is tab autocompletion adds a trailing slash to the directory name, so it would be an inconvenience to have to remove it every time.

hjoliver:
OK fair point. Approving on the basis that suite name often mirrors source dir name

hjoliver (approved):

Issue #3593, opened 2020-05-04
Drop the task "owner" concept.


hjoliver:
We've agreed to get rid of the “task owner” concept in Cylc, used to run jobs on other user accounts.

It is better to use ssh config for this (note we need to document this for users).

oliver-sanders:
Pushing this forward a bit as we have already chipped away at owner functionality in recent PRs.

PR #3990, opened 2020-12-11
Remove `[platforms][PLATFORM][owner]` from conf

wxtim:
These changes partially address #3696 and #3593 

- [x] Already covered by existing tests.
- [x] Doc and config changes to be handled at the end of #3696

PR #4015, opened 2020-12-23
task owner: remove remains of obsolete functionality

oliver-sanders:
follow on from: #3990
closes: #3593

**Requirements check-list**
- [x] No change log entry required (functionality already removed).
- [x] No documentation update required (document as part of larger changes).
- [x] No dependency changes.


cylc/flow/data_messages.proto
```diff
@@ -117,7 +117,6 @@ message PbJob {
     optional string host = 15;
     optional string init_script = 16;
     optional string job_log_dir = 17;
-    optional string owner = 18;
```
oliver-sanders, line 120: @dwsutherland good to remove like this?

wxtim (approved):
- [x] Seems rational.
- [x] Doesn't break any tests.
- [ ] @dwsutherland  okayed changes to `data_messages.proto`.

Approved, conditional on that last.

cylc/flow/data_messages.proto
kinow: I think after the 8.0 final release it shouldn't happen. Instead we should probably use

```
optional string owner = 18 [deprecated=true];  // No longer used, see issue-1234; to be removed in 8.1 or 9.0...
```

But I think it should be OK to remove it now, but @dwsutherland is the best to answer that. Also, David, if this is removed now, do we re-organize the other field numbers (i.e. re-assign the `18` to the `19`, and so on?)?

kinow (approved):
Looks good to me. Pending @dwsutherland to check the review questions :+1:

hjoliver:
Ping @dwsutherland - small question for you, to finish this one off (when you're ready).

cylc/flow/data_messages.proto
dwsutherland: Yeah, that's fine.. Number tag doesn't need to be consecutive

dwsutherland (approved):
Looks good to me! (sorry for the late review)

You'll need to rebase, and the `job_pool.py` is now merged into `data_store_mgr.py` .. (And I'm sure you know; it's always safer to regenerate the `data_messages_pb2.py` then to manually resolve)

👍

oliver-sanders:
Rebased @wxtim could you sanity check and merge.

PR #4016, opened 2020-12-23
cfgspec: fix doc issues

oliver-sanders:
Fix some doc issues / RST formatting problems.

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -223,8 +223,8 @@ with Conf('global.cylc', desc='''
                 DurationFloat(300),
                 desc='''
                     Default for
-                    :cylc:conf:
-                    `flow.cylc[scheduler][mail]task event batch interval`.
+                    :cylc:conf:`flow.cylc
+                    [scheduler][mail]task event batch interval`
                 '''
             )
 
--- a/cylc/flow/cfgspec/suite.py
+++ b/cylc/flow/cfgspec/suite.py
@@ -127,8 +127,8 @@ with Conf(
             format, see :cylc:conf:`flow.cylc[scheduler]cycle point time zone`.
             To just alter the number of expanded year digits (for years
             below 0 or above 9999), see
-            :cylc:conf:
-            `flow.cylc[scheduler]cycle point num expanded year digits`.
+            :cylc:conf:`flow.cylc
+            [scheduler]cycle point num expanded year digits`.
 
             Cylc usually uses a ``CCYYMMDDThhmmZ`` (``Z`` in the special
             case of UTC) or ``CCYYMMDDThhmm±hhmm`` format for writing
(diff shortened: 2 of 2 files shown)


cylc/flow/cfgspec/globalcfg.py
```diff
@@ -178,7 +178,7 @@
                        Manually specified host name or IP address (requires
                        *host*).
             ''')
-            Conf('target', VDR.V_STRING, 'google.com', desc='''
```
oliver-sanders, line 181: remove silly default

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -223,8 +223,8 @@
                 DurationFloat(300),
                 desc='''
                     Default for
-                    :cylc:conf:
-                    `flow.cylc[scheduler][mail]task event batch interval`.
+                    :cylc:conf:`flow.cylc
```
oliver-sanders, line 226: can handle newlines within the reference string thinggy (the bit in the backqoutes) but cant handle a newline at the beginning of the reference string thinggy.

cylc/flow/cfgspec/suite.py
```diff
@@ -463,7 +463,7 @@
 
             Example:
 
-               ``foo(PT1H30M), bar(PT1.5H), baz``
```
oliver-sanders, line 466: Rogue blockquote.

cylc/flow/cfgspec/suite.py
```diff
@@ -327,7 +327,7 @@
             for this, it will be assumed to be local time, or in UTC if
             :cylc:conf:`flow.cylc[scheduler]UTC mode` is set, or in the time
             zone determined by
-            :cylc:conf`flow.cylc[scheduler][cycle point time zone]`.
+            :cylc:conf:`flow.cylc[scheduler]cycle point time zone`.
```
oliver-sanders, line 330: Setting referenced as a section.

Didn't fail the docs build because the reference was corrupted (missing the trailing `:`).

cylc/flow/cfgspec/suite.py
MetRonnie: I think the whole indentation is off for this config node - should be indented by 4 more spaces. This is what it looks like with this PR branch checked out (ignore the orange highlight, that's just due to ctrl+F)
Screenshot 2020-12-23 140521

MetRonnie (changes requested):

cylc/flow/cfgspec/suite.py
oliver-sanders: Ah, looks like I took away one space more than I meant to.

MetRonnie: It renders correctly now, but the indentation of the `desc` content in the code is 4 spaces less than it should be for consistency.

MetRonnie (approved):

wxtim (commented):
👍🏻 Thanks for fixing what may well be my mistakes.

PR #4017, opened 2020-12-24
cylc clean 2: remote clean

### LINKED ISSUE #3887: cylc clean

MetRonnie:
These changes partially address #3887

> 2. Remove stopped workflows on all filesystems

Clean on remote hosts in addition to local filesystem. This is done by getting the platforms from the workflow database and then looking up the install targets for those platforms from `global.cylc`.

**Requirements check-list**
- [x] Documentation will be handled by https://github.com/cylc/cylc-doc/issues/256
- [x] No dependency changes.


MetRonnie:
>```
>tests/f/cylc-clean/01-remote.t .. skipped: 'tree' command not available on remote host _remote_background_indep_tcp
>```
@oliver-sanders `tree` is installed on the GitHub Actions runner but not on the remote host. Is there a way to install `tree` on the host? (I take it this is to do with Docker?)

cylc/flow/rundb.py
```diff
@@ -609,6 +610,11 @@ def select_task_job_run_times(self, callback):
         for row_idx, row in enumerate(self.connect().execute(stmt)):
             callback(row_idx, list(row))
 
+    def select_task_job_platforms(self):
+        """Return the set of platform names from task_jobs table."""
+        stmt = f"SELECT platform_name FROM {self.TABLE_TASK_JOBS}"
+        return set(i[0] for i in self.connect().execute(stmt))
```
oliver-sanders, line 616: SQL has a `DISTINCT` keyword that filters out duplicate items which should be a good bit faster for large flows.

https://www.1keydata.com/sql/sqldistinct.html

cylc/flow/suite_files.py
```diff
@@ -611,6 +663,74 @@ def clean(reg):
...
+
+    Args:
+        reg (str): Workflow name.
+        platform (dict): Config for the platform on which to remove the
+            workflow.
+    """
+    LOG.info(
+        f'Cleaning on install target: {platform["install target"]} '
+        f'(platform: {platform["name"]})')
+    cmd = ['remote-clean', reg, get_remote_suite_run_dir(platform, reg)]
```
oliver-sanders, line 728: Can just call plain `cylc clean` to avoid duplicating it as `remote-clean`. The `remote_*` scripts are things that are called automatically by Cylc during the course of running a flow.

MetRonnie: At the moment, the difference between `clean` and `remote-clean` is the former includes looking up the platforms in the database, and the latter includes an arg for specifying the cylc-run dir on the machine, which shouldn't be an arg for the local command

cylc/flow/suite_files.py
```diff
@@ -21,22 +21,30 @@
 
 import os
 from pathlib import Path
+from random import shuffle
 import re
 import shutil
+from subprocess import Popen, PIPE, DEVNULL
```
MetRonnie, line 27: Just noticed this line at the top:
```
# Note: Some modules are NOT imported in the header. Expensive modules are only
# imported on demand.
```
Should I import `Popen` etc inline instead? (This was done in `detect_old_contact_file()` before this)

oliver-sanders: Eh, I think we can stop doing that and import at the top of the file like normal code.

cylc/flow/suite_files.py
MetRonnie: Ok, I've replaced `remote-clean` with an internal-use option for normal `clean`, so it will be `cylc clean --local-only-run-dir=<run_dir>`. Will push later today

MetRonnie:
`tests/f/cylc-clean/01-remote.t` is passing now on Actions

MetRonnie:
Tests seem a bit flaky at the moment, `tests/k/xtriggers/01-suite_state.t` and `tests/k/restart/21-task-elapsed.t` both passing locally for me

wxtim (commented):
Add myself as a reviewer. GH broken.

cylc/flow/scripts/clean.py
```diff
@@ -18,39 +18,50 @@
...
-        argdoc=[("REG", "Suite name")]
+        argdoc=[('REG', "Suite name")]
     )
+
+    parser.add_option(
+        '--local-only', metavar='RUN_DIR',
+        help="(For internal use) Path of the workflow run dir; will only "
+             "clean on local filesystem.",
+        action='store', dest='local_only__run_dir'
+    )
```
oliver-sanders, line 53: Why do we need to pass the path, can't that be determined from the REG?

When we add more functionality this option will make sense for other purposes (we will want a `--remote-only` option to go with it) so can remove the (internal only) flag.

```suggestion
    parser.add_option(
        '--local-only', '--local',
        help="Only clean on local filesystem.",
        action='store_true',
        default=False
    )
```

oliver-sanders (commented):
Looks good, no problems in testing.

The logging is a little odd, the "Removing directory" notices come through as warnings, looks like I get one per install-target but no information as to where it is coming from.

```
$ cylc clean generic
2021-01-06T15:46:21Z INFO - Cleaning on install target: _remote_background_indep_poll
	(platform: _remote_background_indep_poll)
2021-01-06T15:46:21Z INFO - Cleaning on install target: _remote_background_indep_tcp
	(platform: _remote_background_indep_tcp)
2021-01-06T15:46:21Z WARNING - 2021-01-06T15:46:21Z INFO - Removing directory:
	/root/cylc-run/generic
	
2021-01-06T15:46:21Z WARNING - 2021-01-06T15:46:21Z INFO - Removing directory:
	/root/cylc-run/generic
	
2021-01-06T15:46:21Z INFO - Removing directory: /Users/oliver/cylc-run/generic
```

We will probably want to clean multiple platforms in parallel for 8.0.0 so reporting remote logging is tricky, can just log in the event of error.

cylc/flow/scripts/clean.py
```diff
@@ -18,39 +18,50 @@
...
 
 
 @cli_function(get_option_parser)
 def main(parser, opts, reg):
-    clean(reg)
+    if opts.local_only__run_dir:
+        clean(reg, run_dir=opts.local_only__run_dir)
+    else:
+        init_clean(reg)
+    # print("Welcome, to Mineral Splash!")
```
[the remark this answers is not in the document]
MetRonnie: D'oh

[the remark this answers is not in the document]
MetRonnie: (It was this: https://www.youtube.com/watch?v=pAoK4Pk_Nec for the record)

MetRonnie:
@oliver-sanders Your points should be addressed now, plus I've added another unit test

MetRonnie:
Ah, wait, sorry, forgot about the logging issue. I've noticed too on exvcylcdev that everything that comes back from the Popen subprocess is in stderr instead of stdout for some reason, so it gets logged as a warning. https://github.com/cylc/cylc-flow/blob/06cf5f8ec1d8eafb5fdabe171a98e610cab8430d/cylc/flow/suite_files.py#L686-L692

cylc/flow/platforms.py
```diff
@@ -421,8 +421,28 @@ def get_install_target_from_platform(platform):
...
+
+    Return {install_target_1: [platform_1_dict, platform_2_dict, ...], ...}
+
+    Args:
+        platform_names (list): List of platform names to look up in the
+            global config.
+    """
+    platform_names = set(platform_names)
+    platforms = [get_platform(p_name) for p_name in platform_names]
+    install_targets = set(get_install_target_from_platform(platform)
```
wxtim, line 436: Do you need this second `set`?

cylc/flow/suite_files.py
```diff
@@ -611,6 +656,77 @@ def clean(reg):
...
+
+    queue = []
+    for target, platforms in install_targets_map.items():
+        if target == 'localhost':
+            continue
+        shuffle(platforms)
+        # Issue ssh command:
+        queue.append(
+            (_remote_clean_cmd(reg, platforms[0]), target, platforms)
+        )
```
wxtim, line 684: `random.choice(_list_)` might create cleaner code?
```suggestion
        # Issue ssh command:
        queue.append(
            (_remote_clean_cmd(reg, choice(platforms), target, platforms)
        )
```

wxtim (approved):
- [x] Read code - Nothing obvious stands out, minor comments.
- [x] Read tests.
- [x] Checked-out and poked at. Satisfied it works in the desired way with some of my test workflows.

cylc/flow/platforms.py
oliver-sanders: I think yes because otherwise you would end up with duplicate install targets.

Aside, Python has now given us set comprehensions to match list and dict comprehensions, but we don't get tuple comprehensions, those map to generators.

```
install_targets = {
    get_install_target_from_platform(platform)
    for platform in platforms
}
```

cylc/flow/suite_files.py
oliver-sanders: Would be better for load balancing but would have to be done earlier in the code as in the case of failure the platform has to be popped from the list.

oliver-sanders:
> I've noticed too on exvcylcdev that everything that comes back from the Popen subprocess is in stderr instead of stdout for some reason, so it gets logged as a warning.

Hmmm, what does the ret-code say? Would be good to include that in the logging (note we can use the debug level).

cylc/flow/suite_files.py
MetRonnie: I'm not sure what you mean by it would be better for load balancing?

MetRonnie:
`ret_code` is `0` but I've noticed that the `LOG.info` gets added to the bottom of the blurb in `err`:
> \nWelcome to the Met Office\n\nWARNING: exvcylcdev.......in accordance with the Monitoring policy.\n\n2021-01-12T10:49:59Z INFO - Removing directory: ~/cylc-run/temp5\n

cylc/flow/suite_files.py
```diff
@@ -602,6 +647,77 @@ def clean(reg):
...
+    Args:
+        reg (str): Workflow name.
+        platform (dict): Config for the platform on which to remove the
+            workflow.
+    """
+    LOG.info(
+        f'Cleaning on install target: {platform["install target"]} '
+        f'(platform: {platform["name"]})')
+    cmd = ['clean', '--local-only', reg]
+    cmd = construct_ssh_cmd(cmd, platform, timeout='10s')
```
oliver-sanders, line 716: Sadly 10s is way too eager for our filesystems!

Not sure what a sensible value would be.

For interactive use we don't need a timeout if we capture ctrl+c then kill the SSH process.

oliver-sanders:
If out and error are coming out funny might be worth checking the function that constructs the SSH command. You may need to tell it to capture/return stdout/err.

cylc/flow/suite_files.py
MetRonnie: Note 1: In today's meeting we decided on 2 mins

Note 2: Reminder to self to look into using asyncio futures to paralellise waiting for each subprocess

MetRonnie:
Turns out `LOG.info()` always ends up in stderr! E.g. try
```
cylc run --no-detach my_flow > /dev/null
```
and all the logging will still get printed, or try
```
cylc run --no-detach my_flow 2> /dev/null
```
and none of the logging will get printed

---
> By default, no destination is set for any logging messages. You can specify a destination (such as console or file) by using `basicConfig()` as in the tutorial examples. If you call the functions `debug()`, `info()`, `warning()`, `error()` and `critical()`, they will check to see if no destination is set; and if one is not set, they will set a destination of the console (`sys.stderr`) and a default format for the displayed message before delegating to the root logger to do the actual message output.

from https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial

cylc/flow/suite_files.py
MetRonnie: Should I include a cli option for the timeout in this PR?

MetRonnie:
More on this: https://github.com/cylc/cylc-flow/blob/500a7baaf86c2cfa1a8b32605181e820b37efa4a/cylc/flow/option_parsers.py#L253-L271

hjoliver:
>
>```
> # 1. On choosing STDERR: Log messages are diagnostics, so STDERR is the 
> #    better choice for the logging stream. This allows us to use STDOUT 
> #    for verbosity agnostic outputs.
> ```

That's correct, IMO. (Although I had kind of forgotten about it!)

MetRonnie:
Ok - logging verbosity, remote timeout and concurrency of ssh commands should be addressed now

oliver-sanders:
Looking good.

Should demote the missing database warning as a missing database is not an error (e.g. flow not run or remote flow installation) suggest:

```
INFO: No database - only this platform will be cleaned
```

oliver-sanders:
@hjoliver I think this is good to go now if you want to take a poke.

MetRonnie:
> Should demote the missing database warning as a missing database is not an error (e.g. flow not run or remote flow installation)

Updated the latest commit

MetRonnie:
The tests have passed, however `test (ubuntu-latest, 3.7, tests/f tests/k, 1/1, _remote_background_indep_poll _remote_at_indep_poll)` took a fair bit longer than usual so that run just timed out before the coverage was uploaded. Might have to increase the GH workflow timeout https://github.com/cylc/cylc-flow/blob/c2f6942efc27d5e7f735f7df1d0a2d4bf1d14cdd/.github/workflows/test_functional.yml#L8

oliver-sanders (approved):
> Might have to increase the GH workflow timeout

Dammit, just when I though things were stable, will keep an eye out.

Looks good to me. Going to leave for @hjoliver to merge. Will need to rebase squash pre-merge.

MetRonnie:
> Will need to rebase squash pre-merge.

How come? I tried to keep my commits logically grouped. I can try to squash them down more before merging if needed, but I think it would pay to keep some of the changes separate

oliver-sanders:
Tis how we work, most of the commits here document the process of how this branch came to be, this information doesn't really have any value going forward. We aren't going to go back and revert one commit and knowing that one feature was developed slightly earlier than another isn't useful info to anyone. The extra commits do, however, bloat the repo and make bisects longer.

Some (I think predominantly agile) projects use "squash and merge" for all PRs to strip out unnecessary commits, makes sense as the most important information to preserve the is the link to the PR, the history and purpose of the branch should be documented there. We don't because our PRs often do multiple things (we aren't very agile).

MetRonnie:
> preserve the is the link to the PR, the history and purpose of the branch should be documented there

I don't know why that didn't occur to me before 👍

hjoliver (approved):
Looks great :tada: 

~Please merge after tidying the commit history as suggested by @oliver-sanders~ (duh - he suggested squash and merge :facepalm: )

PR #4018, opened 2020-12-30
pre_configure: capture plugin error

oliver-sanders:
Closes https://github.com/cylc/cylc-rose/issues/9

Catch errors from plugins and wrap them in a Cylc error with plugin context. Should be good enough for now.

*Example `rose-suite.conf`*:

```ini
[jinja2:suite.rc]
Y=X
```

*Output:*

```console
$ cylc install .
PluginError: Error in plugin cylc.post_configure.rose
jinja2:suite.rc=Y=X: Invalid template variable: X
Must be a valid Python or Jinja2 literal (note strings "must be quoted").
```

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/exceptions.py
+++ b/cylc/flow/exceptions.py
@@ -27,6 +27,21 @@ class CylcError(Exception):
     """
 
 
+class PluginError(CylcError):
+    """Represents an error arising from a Cylc plugin."""
+
+    def __init__(self, entry_point, plugin_name, exc):
+        self.entry_point = entry_point
+        self.plugin_name = plugin_name
+        self.exc = exc
+
+    def __str__(self):
+        return (
+            f'Error in plugin {self.entry_point}.{self.plugin_name}'
+            f'\n{self.exc}'
+        )
+
+
 class UserInputError(CylcError):
     """Exception covering erroneous user input to a Cylc interface.
 
--- a/cylc/flow/parsec/fileparse.py
+++ b/cylc/flow/parsec/fileparse.py
@@ -39,7 +39,8 @@ from pathlib import Path
 
 from cylc.flow import __version__
 from cylc.flow import LOG
-from cylc.flow.parsec.exceptions import ParsecError, FileParseError
+from cylc.flow.exceptions import PluginError
+from cylc.flow.parsec.exceptions import FileParseError, ParsecError
 from cylc.flow.parsec.OrderedDict import OrderedDictWithDefaults
 from cylc.flow.parsec.include import inline
 from cylc.flow.parsec.util import itemstr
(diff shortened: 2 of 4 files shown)


cylc/flow/exceptions.py
```diff
@@ -27,6 +27,21 @@ class CylcError(Exception):
     """
 
 
+class PluginError(CylcError):
+    """Represents an error arrising from a Cylc plugin."""
```
kinow, line 31: s/arrising/arising ?

cylc/flow/parsec/fileparse.py
```diff
@@ -223,7 +224,16 @@ def process_plugins(fpath):
     for entry_point in pkg_resources.iter_entry_points(
         'cylc.pre_configure'
     ):
-        plugin_result = entry_point.resolve()(fpath)
+        try:
+            plugin_result = entry_point.resolve()(fpath)
+        except Exception as exc:
+            # NOTE: except Exception (purposefully vague)
+            # this is to separate plugin from core Cylc errors
```
kinow, line 231: :+1:

cylc/flow/scripts/install.py
```diff
@@ -48,16 +49,34 @@ def main(parser, options, reg):
...
         'cylc.pre_configure'
     ):
-        entry_point.resolve()(Path(flow_file).parent)
+        try:
+            entry_point.resolve()(Path(flow_file).parent)
+        except Exception as exc:
+            # NOTE: except Exception (purposefully vague)
+            # this is to separate plugin from core Cylc errors
+            raise PluginError(
+                'cylc.post_configure',
```
kinow, line 58: @oliver-sanders in `fileparse.py`, the entry point iterated and the first argument to `PluginError` both matched (`cylc.pre_configure`).

But here we iterate the plug-ins `cylc.pre_configure` and raise an error `PluginError` with `cylc.post_configure`. Is that correct?

cylc/flow/scripts/install.py
```diff
@@ -48,16 +49,34 @@ def main(parser, options, reg):
...
+            entry_point.resolve()(
+                dir_=os.getcwd(),
+                opts=options,
+                dest_root=get_suite_run_dir(suite)
+            )
+        except Exception as exc:
+            # NOTE: except Exception (purposefully vague)
+            # this is to separate plugin from core Cylc errors
+            raise PluginError(
+                'cylc.post_configure',
```
kinow, line 76: Ditto the above, in this case they are slightly different, though I think they may mean the same thing. The plug-ins iterated are `cylc.post_install`, but the error is for `cylc.post_configure`. Just checking if that's intentional, or if it needs updating.

kinow (commented):
Couple comments, but code is good and ready to be merged after @oliver-sanders has time to reply :+1:

cylc/flow/scripts/install.py
oliver-sanders: Dammit, no, was rushing.

kinow (approved):

wxtim (approved):

Issue #3866, opened 2020-10-13
Decide on the future of naked tasks.


hjoliver:
Should they stay or should they go?

https://github.com/cylc/cylc-admin/pull/111#pullrequestreview-506579884

[edit]

"Naked" tasks are tasks without an explicit runtime definition i.e in this example `bar` is a "naked" task:

```
[scheduling]
  [[graph]]
    R1 = """
      foo
      bar
    """

[runtime]
  [[foo]]
```

Naked tasks should cause configuration errors by default with this behaviour overridden by a new config:

```ini
[scheduling]
  allow implicit = True
```

We have referred to these as dummy or naked-dummy tasks in the past, I think we will go with "implicit" from here on?

[/edit]

dpmatthews:
I think we need to change terminology.
These tasks will run whatever is defined in the `root` runtime section so they are not necessarily dummy tasks.
Maybe "implicit" rather than "naked"?

If we really want to allow naked tasks to only generate a warning by default (not my preference) then I'd suggest we add a `cylc.flow` setting which turns these warning into errors. We could also do the same for deprecation warnings.

hjoliver:
I may be coming around to your point of view.  To be fair it should be more important to avoid accidental "implicit tasks" caused by  a simple typo, than to make quick "experiment with scheduling" flows super-easy to write. 

How about disallow them, but provide a command line one-off opt-in capability? `cylc play --allow-implicit-tasks`

oliver-sanders:
#### The case for naked tasks

The majority of the complexity of a workflow lies in the `[scheduling]` section, this is where the hardest-to-debug errors are made and where support queries regularly lead. This section configures many of the most conceptual parts of a workflow including cycling, recurrences, xtriggers, etc.

As a result I would encourage users to start writing a workflow with the `[scheduling]` section, then move on to the `[runtime]` in due course. This is a more incremental approach which avoids debugging cycling issues in a real workflow with long run times and added complexity. The Cylc tutorial embodies this, introducing the `[scheduling]` section, graph logic and visualisation tools before even mentioning the `[runtime]` section. This is an approach I've seen users take from time to time when writing new suites. It's also an approach I often use myself for mocking up simple workflow examples to send in answer to user queries.

Naked tasks reduce the barrier to experimentation and rapid-prototyping making learning and workflow development easier, especially for new users as they feel their way around Cylc's scheduling concepts.

[...]

PR #4020, opened 2020-12-31
validate: do not check for circular/cyclic graph if size is above limit

### LINKED ISSUE #3869: validate: add separate option for cyclic graph detection
Cyclic graph validation can be quite slow and is currently included as part of  `--strict`.
See https://github.com/cylc/cylc-admin/pull/111#discussion_r503579388
See also https://github.com/cylc/cylc-flow/issues/3866

MetRonnie:
These changes close #3869

> Cyclic graph validation can be quite slow and is currently included as part of `--strict`

This PR (last one of 2020!) introduces an option `--check-circular` to `cylc validate`. If the number of tasks is less than the limit (100), checking for circular graph will always happen. If the number is above the limit, the check will be skipped (with a warning) unless `--check-circular` is used.

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -151,7 +151,7 @@ to control the job submission environment.
 
 [#4020](https://github.com/cylc/cylc-flow/pull/4020) - `cylc validate` will no
 longer check for a cyclic/circular graph if there are more than 100 tasks,
-unless the option  `--check-circular` is used. This is to improved performance.
+unless the option  `--check-circular` is used. This is to improve performance.
 
 
 ### Fixes
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -149,6 +149,10 @@ parse ``rose-suite.conf`` files in repository "cylc-rose".
 [#3955](https://github.com/cylc/cylc-flow/pull/3955) - Global config options
 to control the job submission environment.
 
+[#4020](https://github.com/cylc/cylc-flow/pull/4020) - `cylc validate` will no
+longer check for a cyclic/circular graph if there are more than 100 tasks,
+unless the option  `--check-circular` is used. This is to improved performance.
+
 
 ### Fixes
 
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -132,6 +132,7 @@ class SuiteConfig:
     """Class for suite configuration items and derived quantities."""
 
     Q_DEFAULT = 'default'
+    CHECK_CIRCULAR_LIMIT = 100  # If no. tasks > this, don't check circular
 
     def __init__(
         self,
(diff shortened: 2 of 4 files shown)


cylc/flow/config.py
```diff
@@ -844,12 +845,19 @@ def _check_circular(self):
         """Check for circular dependence in graph."""
         start_point_string = (
             self.cfg['visualization']['initial cycle point'])
+        raw_graph = self.get_graph_raw(start_point_string,
+                                       stop_point_string=None)
+        if (len(raw_graph) > self.CHECK_CIRCULAR_LIMIT and not (
```
oliver-sanders, line 850: Can use `len(self.taskdesf)` to measure the number of tasks rather than graph edges (`get_graph_raw` is slow).

cylc/flow/config.py
```diff
@@ -844,12 +845,19 @@ def _check_circular(self):
...
+                getattr(self.options, 'check_circular', False) or
+                self._is_validate(is_strict=True))):
```
oliver-sanders, line 852: The `--strict` option is going as soon as we decide on [something](https://github.com/cylc/cylc-flow/issues/3866),  can make it so this is just activated by `--check-circular`.

MetRonnie: Ah ok

oliver-sanders (approved):

MetRonnie:
- `tests/k/restart/21-task-elapsed.t` passes locally for me
- `tests/f/rose-conf/04-opts-set-from-env.t` doesn't run locally for me as I haven't installed cylc-rose, but I imagine the problem is not to do with this PR

cylc/flow/scripts/validate.py
```diff
@@ -54,6 +54,13 @@ def get_option_parser():
              "unintentional typographic errors in task names.",
         action="store_true", default=False, dest="strict")
 
+    parser.add_option(
+        "--check-circular",
+        help="Check for circular dependencies in graphs when the number of "
+             "tasks is greater than 100. This can be slow when the number of "
+             "tasks is high.",
+        action="store_true", default=False, dest="check_circular")
```
MetRonnie, line 62: Could also add the abbreviation `--cc` if that's worth doing?

CHANGES.md
```diff
@@ -149,6 +149,10 @@ parse ``rose-suite.conf`` files in repository "cylc-rose".
 [#3955](https://github.com/cylc/cylc-flow/pull/3955) - Global config options
 to control the job submission environment.
 
+[#4020](https://github.com/cylc/cylc-flow/pull/4020) - `cylc validate` will no
+longer check for a cyclic/circular graph if there are more than 100 tasks,
+unless the option  `--check-circular` is used. This is to improved performance.
```
hjoliver, line 154:
```suggestion
unless the option  `--check-circular` is used. This is to improve performance.
```

cylc/flow/scripts/validate.py
```diff
@@ -54,6 +54,13 @@ def get_option_parser():
              "unintentional typographic errors in task names.",
         action="store_true", default=False, dest="strict")
 
+    parser.add_option(
+        "--check-circular",
+        help="Check for circular dependencies in graphs when the number of "
+             "tasks is greater than 100. This can be slow when the number of "
```
hjoliver, line 60:
```suggestion
             "tasks is greater than 100 (smaller graphs are always checked). This can be slow when the number of "
```

hjoliver (approved):

PR #4025, opened 2021-01-06
remove inline #

datamel:

Master currently failing test `tests/functional/rose-conf/04-opts/set-from-env`
This is caused by in-line comment.

This may need to be addressed if it is something we need to support - quick fix for now.

**Requirements check-list**
- [x] Does not need tests (Fix for test).
- [x] No dependency changes.


wxtim (approved):
Thank's Mel

PR #4026, opened 2021-01-06
Fix OrderedDictWithDefault comparison mistake

### LINKED ISSUE #3812: platforms upgrade of root settings
The first [deprecation example](https://cylc.github.io/cylc-admin/proposal-platforms.html#deprecation-examples) in the platforms proposal works as advertised:
```
[scheduling]
    [[graph]]
          R1 = alpha
[runtime]
    [[alpha]]
        [[[remote]]]
            host = localhost
        [[[job]]]
            batch system = background
```

But if I change `[[alpha]]` to `[[root]]` under `[runtime]` the scheduler aborts with:
```
...
2020-09-14T15:09:14+12:00 INFO - Initial point: 1
2020-09-14T15:09:14+12:00 INFO - Final point: 1
2020-09-14T15:09:14+12:00 ERROR - PlatformLookupError: No platform found matching your task
2020-09-14T15:09:15+12:00 INFO - DONE
```

wxtim:
These changes close #3812 

### Cause of bug
From `cylc/flow/parsec/OrderedDictWithDefaults.py`:
```
    def keys(self):
        """Include the default keys, after the list of actually-set ones."""
```

Thus in my code `set(OrderedDictA) ≠ set(OrderedDictA.keys())`. When I originally wrote this code I had not realized that this was the case, and had failed to notice the divergence from standard dict behavior. I had done much of my testing using dicts in plact of OrderedDictWithDefault. In most cases this is fine, but here is an example of a bug which slipped through as a result.

### Checklist

- [x] Unit test added.
- [x] No change log entry required - Fixes a change I hope users will never encounter in anger.
- [x] No documentation update required. Fixes a breakage.

### DIFF
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -307,7 +307,7 @@ def platform_from_job_info(platforms, job, remote):
         # Handle all the items requiring an exact match.
         for task_section in [job, remote]:
             shared_items = set(
-                task_section).intersection(set(platform_spec.keys()))
+                task_section).intersection(set(platform_spec))
             generic_items_match = all((
                 platform_spec[item] == task_section[item]
                 for item in shared_items
--- a/tests/unit/test_platforms.py
+++ b/tests/unit/test_platforms.py
@@ -17,6 +17,7 @@
 # Tests for the platform lookup.
 
 import pytest
+from cylc.flow.parsec.OrderedDict import OrderedDictWithDefaults
 from cylc.flow.platforms import platform_from_name, platform_from_job_info
 from cylc.flow.exceptions import PlatformLookupError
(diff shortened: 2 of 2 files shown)


MetRonnie (commented):
Commenting as a review in case that lets me be added as a reviewer

wxtim:
Thinking about this overnight has led me to wonder if I've not actually fixed the problem, just Hilary's test case. Marked as draft while I have another look.

MetRonnie (approved):
Confirmed that a PlatformLookupError is raised in the test when the change is reverted (and ob that it passes with the change applied)

hjoliver (approved):

PR #4035, opened 2021-01-18
Centralize queue logic, remove queued state

### LINKED ISSUE #2701: Internal queues: task in multiple queues behaviour
From @paulcresswell via email:

> We've noticed that if a task appears in two queues (other than the default queue), it adopts the queue limit of the first one it hits (the suite logs give warnings about ignoring subsequent queues). Is this really desirable? I would have thought that if a user defined two queues, both being picked up through family inheritance, a task should adhere to the smaller of the two values for safety.

See also #1196 #2616.

### LINKED ISSUE #4013: Implement task state rationalization proposal
Complete task state rationalization as described in [this proposal](https://cylc.github.io/cylc-admin/proposal-state-names.html)

- [x] (waiting, queued, runahead) -> waiting (finished by #4035)
- [x] ready -> preparing  #4014
- [x] retrying, submit-retrying: re-implement as xtrigger  #3423
- [x] held: re-implement as attribute  #3230

Note we decided not to do the proposed trivial renaming (succeeded -> run-succeeded, etc.).

hjoliver:
Close #4013 
Close #2701 

- remove the "queued" task state (replacing with a `task.state.is_queued` property)
   - (as first suggested here by @oliver-sanders: https://github.com/cylc/cylc-admin/pull/47#issuecomment-521570792)
- centralize task queue logic
  - easier to understand, to remove the "queued" task state, and potentially to plugin-ize in future
- support (new) overlapping queues(*), but default to Cylc 7 "classic" behavior (independent queues)
  - overlapping queues handle the use cases in #2701 without race conditions or other problems
  - the `default` queue in the overlapping case is a global limit, not an implicit queue for tasks not assigned to other queues
- [UPDATE] fix task hold, so that it actually holds the task as well as adding the `is_held` flag (easier to do here than on master)

 (*) **NOTE** in spite of the config item name (`queues`) this, just like on master, is really *a single FIFO task queue with multiple limit groups*:
- tasks are pushed on to the queue when they become ready to run
- tasks are popped from the queue until it is empty and released if not limited, but those that are limited (by any limit group) are pushed back on in the order they came off. 

You can think of this as multiple queues, sort of, but not really because the global FIFO ordering reflects the fact that it is actually a single queue. (And this is how race conditions - see #2701 - are avoided).

New queue config item "type":
```ini
[scheduling]
    [[queues]]
        type = "overlapping"  # default "classic"
```

**Requirements check-list**

- [x] No dependency changes.


cylc/flow/config.py
```diff
@@ -1216,93 +1216,41 @@ def adopt_orphans(self, orphans):
...
-                            else:
-                                # Ignore: task not used in the graph.
-                                pass
-                        else:
-                            myq[key] = qmember
-                            qmembers.append(qmember)
-                            requeued.append(qmember)
+                    qmembers.add(qmember)
+                else:
+                    print("SHITE", qmember)
```
oliver-sanders, line 1244: Don't change this line for anything.

hjoliver: Oops!! :facepalm:

hjoliver:
```ini
task parameters]
  m = 1..5
[scheduling]
  [[queues]]
     [[[default]]]
        limit = 6 
     [[[small-tasks]]]
        limit = 3 
        members = SMALL
     [[[big-tasks]]]
        limit = 2 
        members = BIG 
  [[graph]]
     R1 = SMALL & BIG & OTHER
[runtime]
  [[root]]
     script = sleep 10
  [[SMALL]]
  [[sml<m>]]
     inherit = SMALL
  [[BIG]]
  [[big<m>]]
     inherit = BIG 
  [[OTHER]]
  [[other<m>]]
     inherit = OTHER
```
^ run this a few times (tasks are pushed onto the FIFO queue as they become ready, and in this example they all become ready at once, so the order will vary for each run).

Using unicode infinity symbol for the queued indicator in Tui - it looks like two tasks queued up in a line :grin: 

Peek 2021-01-19 14-41

oliver-sanders:
> Using unicode infinity symbol for the queued indicator in Tui - it looks like two tasks queued up in a line

Can handle that by using a unicode modifier (like the held icon).

Any of the superscript modifiers in [this table](https://unicode-table.com/en/blocks/combining-diacritical-marks/) should do the trick. [U+030A](https://unicode-table.com/en/030A/) looks pretty good.

Kinda like the idea of a double prime for contrast to the pause icon, but that might be pushing it somewhat when it comes to visual separation.

hjoliver:
3 functional tests hanging locally on this branch

hjoliver:
> Can handle that by using a unicode modifier (like the held icon).

I hadn't noticed that was a "modifier" rather than a separate character. OK, going with your suggestion: 
Screenshot from 2021-01-20 14-07-02

hjoliver:
(rebased onto master)

hjoliver:
(I still need to address the three hanging functional tests)

cylc/flow/tui/util.py
```diff
@@ -52,6 +59,8 @@ def get_task_icon(status, is_held, start_time=None, mean_time=None):
     ret = []
     if is_held:
         ret.append(TASK_MODIFIERS['held'])
+    if is_queued:
```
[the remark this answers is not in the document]
hjoliver: :+1:

hjoliver:
**NOTE** after offline discussion with @dpmatthews I have stepped back from plugin-izing queues and replacing the "classic" queues on this branch.  Please **read the re-written description above**.

hjoliver:
Damn it, looks like I screwed the big rebase...

hjoliver:
Finally got all the tests to pass. The way our job submission and remote select/init/install mechanisms interact is erm, not the clearest.  I'll undraft this after adding some unit tests for the new queue module tomorrow.

hjoliver:
Tests added, got the coverage up :+1:

TomekTrzeciak:
How easy/difficult would be to add custom sort order in queues #1196 on top of the work here?

hjoliver:
@TomekTrzeciak - @dpmatthews has asked the same question, offline. This PR makes future queue enhancements easier in the sense the the queue logic is now localised to one module and (IMO) much easier to understand (which made it easier for me to remove the obsolete queued task state). 

However, our queues (both "classic" and overlapping, here) are really a single queue (see the Issue description above) that allows jumping over tasks that are limited by their membership in one or more limit groups.  In retrospect I regret calling what we have "queues", plural. 

Allowing different sort order in different queues would require actually using multiple queues, I think, with the potential for race conditions if the same task appears in more than one queue. And if there is a global limit we'd also have a cross-queue prioritization problem: if there are more queues than free global slots, which queues do we release tasks from first to reach the global limit?  (And similarly for sub-queues of other queues...). So, it seems to me this this requires more thought, and as that would be new functionality that doesn't seem to be desperately needed it may have to wait ... unless want to take a look?

hjoliver:
@dpmatthews - regarding your concerns about this PR:
- I have punted other potential enhancements to #4068 
- there is one new config item (queue `type = overlapping`) to select the new (non-default) overlapping limit groups. IMO that's worth it because it overlapping limits (#2701) make more sense than the status quo, and we're not sure yet if we're going to change the config again (e.g. for queue plugins) any time soon.

OK?

hjoliver:
(@dpmatthews - assigned you to review too, just to make sure you agree with the previous comment before merge)

hjoliver:
Brilliant, made some trivial changes and rebased on to master, and all the tests broke! :grimacing:

cylc/flow/cfgspec/suite.py
```diff
@@ -427,6 +427,13 @@
             for the whole suite just set the limit on the *default* queue as
             required. See also :ref:`InternalQueues`.
         '''):
+            Conf('type', VDR.V_STRING, "classic",
+                 options=["classic", "overlapping"], desc='''
+                "classic" (FIFO with independent limit groups - "default" queue
+                is for tasks not assigend to other queues) or "overlapping"
```
kinow, line 433: s/assigend/assigned

cylc/flow/tui/util.py
```diff
@@ -31,14 +31,21 @@
 from cylc.flow.wallclock import get_unix_time_from_time_string
 
 
-def get_task_icon(status, is_held, start_time=None, mean_time=None):
+def get_task_icon(
+        status,
```
kinow, line 35: maybe later we could force devs to name the params to avoid accidentally swapping `is_held` by `is_queued`, etc (i.e. `get_task_icon(status, *, is_held=False, ...)`

cylc/flow/task_state.py
```diff
@@ -184,6 +175,8 @@ class TaskState:
             External triggers as {trigger (str): satisfied (boolean), ...}.
         .is_held (bool):
             True if the task is "held" else False.
+        .is_queued (bool):
+            True if the task is queued else False.
```
kinow, line 179: True if the task is "queued"? To follow the example above... (sorry to nit-pick!)

cylc/flow/task_queue.py
```diff
@@ -0,0 +1,201 @@
...
+                released.append(candidate)
+                active.update({candidate.tdef.name: 1})
+
+        # Re-queue rejected tasks, in the original order.
+        for itask in reversed(rejects):
+            self.task_deque.append(itask)
+        if released:
+            LOG.debug("Queue release:")
+        for r in released:
+            LOG.debug(f"  {r.identity}")
```
kinow, line 189: Move the for+log block to under the `if released`?

cylc/flow/task_queue.py
```diff
@@ -0,0 +1,201 @@
...
+        itask.state.reset(is_queued=True)
+        itask.reset_manual_trigger()
+        self.task_deque.appendleft(itask)
+
+    def _is_limited(self, itask: TaskProxy, active: Counter[str]) -> bool:
+        """Return True if the task is limited, else False."""
+        for name, limiter in self.limiters.items():
+            if limiter.is_limited(itask, active):
+                return True
+        return False
```
kinow, line 162: I think it could be too:
```
return any(
  limiter.is_limited(itask, active)
  for name, limiter in self.limiters.items()
)
```
Same but a bit easier to read?

kinow (commented):
Finished reading the code (last task of the day :zzz: ). So easy to read! The types helped me a lot to understand what each object was doing. Found one function only that I had to read to confirm I understood what it was doing. Left a comment there. Also one type :grimacing: 

Will finish reviewing it tomorrow after the Auckland morning tea and after submitting the mid-year performance review :+1:

dpmatthews:
> @dpmatthews - regarding your concerns about this PR:
> 
>    I have punted other potential enhancements to #4068
> 
>    there is one new config item (queue `type = overlapping`) to select the new (non-default) overlapping limit groups. IMO that's worth it because overlapping limits (#2701) make more sense than the status quo, and we're not sure yet if we're going to change the config again (e.g. for queue plugins) any time soon.

@hjoliver - I'm not yet convinced that `type = overlapping` is a good idea - I'd like to explore other options before 8.0.
However, I realise that's not a priority at the moment so I'm happy for this to go in.
I'll add my ideas to #4068 and change it's milestone to 8.0 and then we'll have to see if we manage to find time to take another look before the release.
OK?

cylc/flow/task_state.py
hjoliver: (I'll remove the quotes on "held")

cylc/flow/task_queue.py
hjoliver: Good idea. In fact, single statement then so I'll get rid of the function.

hjoliver:
> I'll add my ideas to #4068 and change it's milestone to 8.0

Good (I actually intended to go through your comments and do that myself, but ran out of time this afternoon).

> and then we'll have to see if we manage to find time to take another look before the release.

All good :+1:

TomekTrzeciak:
> Allowing different sort order in different queues would require actually using multiple queues, I think, with the potential for race conditions if the same task appears in more than one queue.  And if there is a global limit we'd also have a cross-queue prioritization problem: if there are more queues than free global slots, which queues do we release tasks from first to reach the global limit? (And similarly for sub-queues of other queues...). So, it seems to me this this requires more thought,

Sounds like this would require some level of re-engineering. 

> and as that would be new functionality that doesn't seem to be desperately needed it may have to wait ... unless you want to take a look?

While I would not characterise this as desperately needed, there are problems that are hard to solve properly without this functionality.

I had a quick look at implementation and it didn't seem very obvious to me how to add queue sorting if everything ends up in one global queue, which is what prompted my question.

hjoliver:
> Sounds like this would require some level of re-engineering.

Yes, but now it will only require re-engineering the new task_queue module. There is a now a clear API for the scheduler, in terms of queuing and releasing tasks. (I haven't documented it as such yet, but I plan to).

> I had a quick look at implementation and it didn't seem very obvious to me how to add queue sorting if everything ends up in one global queue, which is what prompted my question.

I anticipate you would provide a different queue module, which internally uses multiple queues instead of one.  At that point we would make queues plugins, so users can easily select what kind of queue they want.

kinow (approved):
Found no issues with code. Tested using `five` and another workflow with queues. Found no issues with Cylc Flow. Tested UI `master` and https://github.com/cylc/cylc-ui/pull/575 (removes queue state). Everything working OK for me. +1 :+1:

hjoliver:
Latest commit:  fix task hold, so that it actually holds the task as well as adding the is_held flag (easier to do here than on master)

oliver-sanders:
> I hadn't noticed that was a "modifier" rather than a separate character. OK, going with your suggestion:

Uh, seems to render as two characters in your terminal which is a shame, in the MacOS terminal the modifier is applied to the task state character.



I had noticed some different behaviour with the held modifier in different terminals, I guess differing levels of unicode support. May need to tweak things a little later.

cylc/flow/task_queue.py
```diff
@@ -0,0 +1,197 @@
...
+    def _configure_limiters(self, queues: dict) -> None:
+        """Create a Limiter for each limit group."""
+        for name, config in queues.items():
+            self.limiters[name] = Limiter(config["limit"], config["members"])
+
+    def add(self, itask: TaskProxy) -> None:
+        """Queue a task."""
+        LOG.debug(f"Queue add: {itask.identity}")
+        itask.state.reset(is_queued=True)
+        itask.reset_manual_trigger()
```
oliver-sanders, line 154: Should this be done on entry to the waiting state rather than the queued state?

cylc/flow/task_queue.py
```diff
@@ -0,0 +1,197 @@
...
+        self.task_deque.appendleft(itask)
+
+    def release(self, active: Counter[str]) -> List[TaskProxy]:
+        """Release queued tasks."""
+        released: List[TaskProxy] = []
+        rejects: List[TaskProxy] = []
+        while True:
+            try:
+                candidate = self.task_deque.pop()
```
oliver-sanders, line 163: For large flows with small limits the work of emptying and rebuilding the dequeue could become quite expensive.

Could we just iterate over a list and only remove tasks when dequeued? That would only perform data structure operations when things change:

```python
for candidate in self.task_list:
     if candiate.state.is_held or any(limiter...):
          # no need to track rejects or rebuild the dequeue afterwards
          continue
     self.task_list.remove(candidate)
     released.append(candidate)
     ...
```

cylc/flow/task_queue.py
```diff
@@ -0,0 +1,197 @@
...
+                released.append(candidate)
+                active.update({candidate.tdef.name: 1})
+
+        # Re-queue rejected tasks, in the original order.
+        for itask in reversed(rejects):
+            self.task_deque.append(itask)
+        if released:
+            LOG.debug("Queue release:")
+            for r in released:
+                LOG.debug(f"  {r.identity}")
```
oliver-sanders, line 185: For logging niceness can write this out as a single message with newlines (the logging is configured to auto-indent following lines), I think this'll do it:

```suggestion
            LOG.debug(
                "Queue release:\n"
                + '\n'.join(f"* {r.identity}" for r in released)
            )
```

It might be good to move this to the calling code so that it doesn't have to be re-implemented in other queue implementations if we go that way.

hjoliver:
> Uh, seems to render as two characters in your terminal which is a shame,

No it's a single character, you can tell because it doesn't push the task icon over to the right like in my first gif with the infinity character.

cylc/flow/task_queue.py
hjoliver: Yes it probably should.

hjoliver: Actually, no. The flag gets reset on queued, because it takes two manual triggers to make an un-queued task submit.  However, I've taken this call out of the queue module.

cylc/flow/task_queue.py
hjoliver: I've converted to a list because it's simpler. There are some minuses though: I have to iterate over a copy to allow item removal; and removing list items is more expensive than removing deque items.

cylc/flow/task_queue.py
hjoliver: Done.

hjoliver:
Rebased, addressed feedback, and did some further tidying:
- moved all task state manipulation and logging out of the queue module
- moved some lower-level queue-related code from scheduler to task_pool
- changed implementation from deque to list
- some better public method names

hjoliver:
And ... I broke all the reload tests :grimacing:

hjoliver:
(I forgot to signal yesterday, tests all passing now, this should be ready to go.)

oliver-sanders:
(gh won't let me reply on the thread)

> I've converted to a list because it's simpler.

Thanks, that's clearer to me.

> There are some minuses though: I have to iterate over a copy to allow item removal; and removing list items is more expensive than removing deque items.

Darned, that's not good either, sorry. One way to get around copying the queue is to build the released list first then iterate over that to remove the tasks:

```python
released: List[TaskProxy] = [
    # list comp should yield a small efficiency bonus
    itask
    for itask in self.queue
    if not itask.state.is_held
    and not any(
        limiter.is_limited(itask, active)
        for name, limiter in self.limiters.items()
    )
]
# for large workflows where efficiency matters we would expect the
# released list to be much smaller than the queue so iterating here isn't
# such a big deal
for itask in released:
    self.queue.remove(itask)
    active.update({itask.tdef.name: 1})
return released
```

hjoliver:
> One way to get around copying the queue is to build the released list first then iterate over that to remove the tasks:

Unfortunately the active task counter has to be updated every time a to-be-released task is identified, because the limit for the next task then drops by one.  I can do that inside the list comprehension, but it ain't pretty :vomiting_face: :

```python
released: List[TaskProxy] = [
    # list comp should yield a small efficiency bonus
    itask
    for itask in self.queue
    if not itask.state.is_held
    and not any(
        limiter.is_limited(itask, active)
        for name, limiter in self.limiters.items()
    ) and not active.update({itask.tdef.name: 1})  # <------- BLAAAAARGGHHHHHH: not None
]
# for large workflows where efficiency matters we would expect the
# released list to be much smaller than the queue so iterating here isn't
# such a big deal
for itask in released:
    self.queue.remove(itask)
    # active.update({itask.tdef.name: 1})  #-----^
return released
```

I don't think it will be much of a problem because queues are sort-of self-limiting anyway: only ready-to-run tasks get queued, and queued tasks can't generate any more ready-to-run tasks until released.  

From a quick test with 3000 instantly-ready tasks and an active task limit of 4, my implementation takes ~1.6 sec per call, and this one takes ~1.7.  Maybe the complexity of the list comprehension, with a bunch of function calls inside it, blows out any advantage.
-

hjoliver:
Deconflicted.

oliver-sanders:
> But, a quick test with 3000 instantly-ready tasks and an active task limit of 4: my implementation took ~1.6 sec per call, and this one took ~1.7 sec.

Yikes 1.6s for 3000 tasks, looks like the copy might be a minor factor in the execution time? Maybe this is one to come back to later.

BTW I did a quick scaling test and it looks like the list comprehension advantage has shrunk quite a bit with improvements to for loop efficiency. The two track the same until the number of items passes a threshold one is unlikely to ever encounter (~10^7).

colls

hjoliver:
> Yikes 1.6s for 3000 tasks, looks like the copy might be a minor factor in the execution time? Maybe this is one to come back to later.

Are you suggesting that's bad? Note it was 3000 queued (ready to run) tasks; it would have to be a very nasty workflow to generate a queue like that.

oliver-sanders:
It's not great, however, it should only happen on changes to the task pool(?) so not so bad.

Sorry, this has been a bit of a distraction, perhaps should come back to this later.

> it would have to be a very nasty workflow to generate a queue like that.

A queue of 3000 tasks is not an unlikely use case at all, we see it all the time. With Cylc8 promoting using the default queue as a nicer alternative to runahead limiting I think we will see this more often.

We regularly see Cylc7 suites with multiple thousands of waiting tasks (and sometimes multiple thousands of succeeded tasks thanks to SoS). These are typically data-pipeline type workflows with parallel processing chains hanging off of slow data-retrieval, something like this:

```
get_data => <process> => done
```

There are no inter-cycle dependencies on `get_data` and there might not be any cycling in the flow anyway. If there is cycling then they will want a very large number of `max active cycle points` to allow processing to happen where possible when data becomes available.

tests/unit/test_task_queue.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+READY_TASK_NAMES = ["b3", "s4", "o2", "s3", "b4", "o3", "o4", "o5", "o6", "o7"]
+
+
+@pytest.mark.parametrize(
+    "queue_type,"
+    "active,"
+    "expected_released,"
+    "expected_still_queued,"
```
oliver-sanders, line 110: Are these commas meant to be inside the quotes? Doesn't seem to have broken the test?

oliver-sanders (approved):
Tested both queue types, everything working as expected. Queue logic much easier to grapple with now it's in one place.

hjoliver:
> We regularly see Cylc7 suites with multiple thousands of waiting tasks (and sometimes multiple thousands of succeeded tasks thanks to SoS).

Yes, it's easy to generate thousands of *waiting* (and *succeeded*) tasks in Cylc 7/SOS, and that's a performance issue, but those aren't queued - only ready-to-run (/satisfied) tasks get queued.  And in Cylc 8/SOD, the waiting (and succeeded) tasks are gone. 

It's certainly easy to contrive a case with many thousands in the queue, but I would have thought was unusual (as well as unpleasant):
```ini
graph = "foo => bar<m=1..3000>"
```

tests/unit/test_task_queue.py
hjoliver: Yes (weirdly). It's the `@pytest.mark.parametrize()` arg names, which is a comma-separated string, but split over multiple lines here.

hjoliver:
Closing as superseded by #4088 

@oliver-sanders - I decided to take back out of overlapping queues for now, just in the performance matters on this branch. For "classic" queues there's no need for the more complex queue-wrangling we've been discussing, that only comes in to play when you allow tasks to be in multiple queues.  @dpmatthews ought to be happier too :grin:  In #4088 we have nothing but push and pop operations on simple queues, should be nice and fast.

oliver-sanders:
>  In #4088 we have nothing but push and pop operations on simple queues

:+1:

> It's certainly easy to contrive a case with many thousands in the queue

It's really not as contrived as it seems...

I took a look at our cylc-monitor and pulled off the suite that currently has the highest memory usage. At 1.2GB this is a fairly large flow, though it's not the only member of the GB club at the moment and larger flows routinely crop up:

```
sqlite> select count(status) from task_pool;
34135
sqlite> select count(status) from task_pool where status='waiting';
9131
sqlite> select count(status) from task_pool where status='queued';
3381
sqlite> select count(status) from task_pool where status='succeeded';
21532
```

Firstly, hats off to Cylc7 for happy managing a task pool with 35 thousand odd tasks. It's chewing up 60% of the CPU of one slow virtual core, but that's pretty reasonable really.

In this case the graph is unusually simple and makes the cause of the large queue painfully clear:

```
# a = 1..256
# b =1..2
# c = 1..26

w<b> => x<a><b><c>
X<a>:succeed-all => y => z
```

This is just the worst offender at 9am on a Friday, plenty more where that came from.

The really good news is that SoD should reduce the task pool from 34'135 thousand to 3'481, an order of magnitude difference which aught to sort out the memory footprint.

The bad news is that the queue length in my snapshot is a fraction of what it must have been earlier in the run, expanding those parameters gives us `256 * 2 * 26 = 13'312`. Oh and did I mention this is a cycling suite with `max active cycle points = 2`! A niche example, perhaps, but not an outlier. Queues in excess of 1'000 tasks are pretty common for us in Cylc7. If we start to promote using the default queue for runahead limiting (which is a good idea) these queues are likely to inflate. If nothing else the change in the default `max active cycles` from 3 to 5 will cause an increase by itself.

hjoliver:
(OK, I stand corrected on the existence of insanely large queues in the real world!)

PR #4038, opened 2021-01-19
Fix some functional tests that don't run on GH Actions

MetRonnie:

**Requirements check-list**
- [x] blah blah...

### DIFF
--- a/tests/functional/restart/41-auto-restart-local-jobs.t
+++ b/tests/functional/restart/41-auto-restart-local-jobs.t
@@ -21,7 +21,7 @@ export REQUIRE_PLATFORM='loc:remote fs:shared'
 export CYLC_TEST_HOST2="${CYLC_TEST_HOST}"
 export CYLC_TEST_HOST1="${HOSTNAME}"
 if ${CYLC_TEST_DEBUG:-false}; then ERR=2; else ERR=1; fi
-set_test_number 11
+set_test_number 12
 #-------------------------------------------------------------------------------
 BASE_GLOBAL_CONFIG="
 [scheduler]
@@ -107,7 +107,8 @@ log_scan "${TEST_NAME}-stop" "${FILE}" 40 1 \
     'The Cylc suite host will soon become un-available' \
     'This suite will be shutdown as the suite host is unable to continue' \
     'Suite shutting down - REQUEST(NOW)' \
-    'bar.1: orphaned task'
+    'Orphaned task jobs:' \
+    '* bar.1 (running)'
 
 cylc stop "${SUITE_NAME}" --now --now 2>/dev/null || true
 poll_suite_stopped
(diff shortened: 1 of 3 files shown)


MetRonnie:
One test I'm not sure how to fix: https://github.com/cylc/cylc-flow/blob/af770d4a4c2dc08e187a8d7766ba72f4189ac8e3/tests/functional/platforms/02-host-to-platform-upgrade.t#L52-L55

> grep: ~/cylc-run/cylctb-20210119T155714Z/f/platforms/02-host-to-platform-upgrade/log/job/1/upgradeable_cylc7_settings/NN/job.out: No such file or directory

Seems to be because `job.out` only exists on the remote host, not the local filesystem. But I'm not sure what to do about that

datamel (approved):
All passing now, thanks @MetRonnie

datamel:
> One test I'm not sure how to fix:
> 
> https://github.com/cylc/cylc-flow/blob/af770d4a4c2dc08e187a8d7766ba72f4189ac8e3/tests/functional/platforms/02-host-to-platform-upgrade.t#L52-L55
> 
> > grep: ~/cylc-run/cylctb-20210119T155714Z/f/platforms/02-host-to-platform-upgrade/log/job/1/upgradeable_cylc7_settings/NN/job.out: No such file or directory
> 
> Seems to be because `job.out` only exists on the remote host, not the local filesystem. But I'm not sure what to do about that

Yes, this is failing locally for me but passing on GH actions.

oliver-sanders:
> Seems to be because job.out only exists on the remote host, not the local filesystem.

What platform is this failing for?

* If it is `fs:shared` then the local and remote should be the same filesystem.
  * If the file appears on one host and not another the FS is lagging badly.
  * Try a `sleep 5`, if that fixes it we know what the problem is (an `fsync` is the standard solution).
* If it is `fs:indep` then this means job log retrieval is failing (suite log may enlighten).

datamel:
> > Seems to be because job.out only exists on the remote host, not the local filesystem.
> 
> What platform is this failing for?
> 
> * If it is `fs:shared` then the local and remote should be the same filesystem.
>   
>   * If the file appears on one host and not another the FS is lagging badly.
>   * Try a `sleep 5`, if that fixes it we know what the problem is (an `fsync` is the standard solution).
> * If it is `fs:indep` then this means job log retrieval is failing (suite log may enlighten).

It is failing to return job.out from exvc.. for me. Adding `[remote]retrieve job logs = True` to the flow.cylc returns it (although the `-f` also needs stripping from hostname command), so perhaps it is not mixing old and new correctly.

MetRonnie:
Strange, considering that setting is already in the global config for the test: https://github.com/cylc/cylc-flow/blob/af770d4a4c2dc08e187a8d7766ba72f4189ac8e3/tests/functional/platforms/02-host-to-platform-upgrade.t#L26-L30

wxtim (approved):
- [x] Seem sensible.
- [x] Run locally.

MetRonnie:
It seems that `$CYLC_TEST_PLATFORM` evaluates as `_remote_background_indep_poll`, so the global config gets
`[platforms][_remote_background_indep_poll]retrieve job logs = True` added to it in the test. But this doesn't work.

However, if I instead set `retrieve job logs` to True in **`_remote_at_indep_poll`**, then the `job.out` gets successfully retrieved

PR #4040, opened 2021-01-20
cylc play

### LINKED ISSUE #3897: cli: play|pause|stop
Rename the existing workflow lifecycle commands to match the [rose suite-run migration proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-rose-suite-run.md#cli-changes) and implement the other CLI changes.

I.E. do what it says in the proposal under the following sections:

* `cylc play`
  * With the exception of the `--re-run` option - #3898
* `cylc pause`
* `cylc release`

MetRonnie:
These changes partially address #3897

- Replace `cylc run` and `cylc restart` with `cylc play`.
- Remove the `START_POINT` optional positional argument.
- Tidy up the aliases for cycle points options, so only these are used:

  ```
  --initial-cycle-point, --icp
  --final-cycle-point, --fcp
  --start-cycle-point, --stopcp
  --stop-cycle-point, --startcp
  ```

Note: Running `cylc play` on a workflow that has previously been run now has the effect of `cylc restart`. In order to do a cold start, you will have to either run `cylc clean` or remove the private & public databases before running `cylc play` (a function for the latter has been added to `test_header` for the functional tests)

**Requirements check-list**
- [x] Appropriate tests included?
- [x] (master branch) Will open a documentation PR after `cylc pause` is implemented
- [x] No dependency changes.


MetRonnie:
https://github.com/cylc/cylc-flow/pull/4040/files#annotation_899603301
> cylc/flow/scheduler.py#L1358
> Added line #L1358 was not covered by tests

This is executed as part of `tests/functional/restart/34-auto-restart-basic.t`, but doesn't run on Actions as it requires shared filesystem

hjoliver:
Brilliant to have this ready to go.  Should it be un-Drafted now @MetRonnie ?

MetRonnie:
Still some tests that I should run/add to this

MetRonnie:
@hjoliver Should be ready now

cylc/flow/scheduler_cli.py
```diff
@@ -44,35 +44,39 @@
...
+dependence on tasks prior to the initial cycle point is ignored.
 
-A "warm start" starts from a given cycle point (--startcp=CYCLE_POINT) that is
-later than the initial cycle point specified in flow.cylc. The
-initial cycle point is preserved, but all tasks and dependencies before the
-given warm start cycle point are ignored.
+A "warm start" starts (a freshly-installed workflow) from a given cycle point
+(--startcp=CYCLE_POINT) that is later than the initial cycle point specified in
+flow.cylc. The initial cycle point is preserved, but all tasks and dependencies
+before the given start cycle point are ignored.
```
MetRonnie, line 69: I'm not sure we need to explain the difference between cold and warm starts in the docstring anymore, maybe just explain the `--startcp` option (or maybe not at all, as that's in the help anyway?)

cylc/flow/scheduler.py
```diff
@@ -1778,14 +1776,14 @@ def process_cylc_stop_point(self):
...
         Set stop point.
 
-        In priority order stop cyclepoint (``stopcp``) is set:
-        * From the final point ff ``cylc restart --ignore-stop-cycle-point``.
-        * From the command line (``cylc <run/restart> --stop-point``).
+        In decreasing priority, stop cycle point (``stopcp``) is set:
+        * From the final point for ``cylc play --stopcp=ignore``.
+        * From the command line (``cylc play --stopcp=XYZ``).
         * From the database.
         * From the flow.cylc file (``[scheduler]stop after cycle point``).
```
MetRonnie, line 1809: Actually, why should `cylc play --stopcp=ignore` (for a restart) use the final cycle point as the stop point, considering that `--icp=ignore` and `--fcp=ignore` result in use of the flow.cylc specified value?

cylc/flow/scheduler_cli.py
hjoliver: We could stop using the terminology "warm start", but still need to explain that you can start afresh mid-way through the graph.

cylc/flow/scheduler.py
hjoliver: Sorry, I can't get my head around this! -"result in use of the flow.cylc specified value" of what exactly, the "stop CP"? 

The initial and final CPs define the extent of the graph, and start and stop CP specify where to start and stop within those bounds. So its seems reasonable to me to be able to ignore a previously-specified stop CP at restart, but shutdown at the FCP.  Or am I missing the point?

MetRonnie: Sorry for not being clear! If we're just talking about the pre-existing behaviour, which I don't think I've changed:

The behaviour of `--ignore-initial-cycle-point` is to ignore the icp stored in the database, and use whatever is specified by `[scheduling]initial cycle point`. Same for `--ignore-final-cycle-point`. But for `--ignore-stop-cycle-point`, while it ignores the stopcp in the database, it *doesn't* use `[scheduling]stop after cycle point`, it instead uses the fcp (not sure if from the database or flow.cylc).

So the behaviour is inconsistent between icp/fcp vs stopcp, is what I'm saying.

(For completeness, `--ignore-start-cycle-point` just ignores the startcp in the DB and doesn't do anything else. But as far as I can tell, the option won't make any difference to a restart though? The restart starts from whatever the last cycle point before shutdown was)

MetRonnie: Going to continue this discussion here: https://github.com/cylc/cylc-flow/issues/4062 (haven't written it up yet)

cylc/flow/scheduler.py
```diff
@@ -1771,6 +1775,29 @@ def _get_events_conf(self, key, default=None):
         return self.suite_event_handler.get_events_conf(
             self.config, key, default)
 
+    def process_cycle_point_opts(self) -> None:
+        """Check the values of --icp, --fcp, --startcp, --stopcp.
+
+        Reset the values to None if necessary:
+        * The value 'ignore' is not used in a first start.
+        * The opts --icp and --startcp cannot be used in a restart.
+        """
```
MetRonnie, line 1784: I'm certain that `--startcp` can't be used in a restart. And, previously with `cylc restart`, there was no `--icp` option for changing the icp. However, you could in fact use `--ignore-initial-cycle-point` with `cylc restart`, which allows you to change the icp. 

So, should we
1. Allow the use of `--icp=CYCLE_POINT` for a restart?
2. Or disallow the use of `--icp=ignore` for a restart?

tests/functional/restart/51-ignore-final-point.t
```diff
@@ -47,44 +47,44 @@ init_suite "${TEST_NAME_BASE}" <<'__FLOW_CONFIG__'
...
+                :;;
+            esac
+        """
 __FLOW_CONFIG__
 
 run_ok "${TEST_NAME_BASE}-validate" cylc validate "${SUITE_NAME}"
 
 suite_run_ok "${TEST_NAME_BASE}-run" \
-    cylc run "${SUITE_NAME}" --no-detach --fcp=2018
+    cylc play "${SUITE_NAME}" --no-detach --fcp=2017
```
oliver-sanders, line 67: What's with the FCP changes here?

oliver-sanders (commented):
Looks good to me, should save the remaining initial/final/start/stop questions for #4062 to avoid blocking this PR.

A few small comments...

cylc/flow/scheduler_cli.py
```diff
@@ -42,67 +42,48 @@
...
-A "warm start" (-w/--warm) starts from a given cycle point later than the suite
-initial cycle point (specified in flow.cylc). Any dependence on tasks prior
-to the given warm start cycle point is ignored. The suite initial cycle point
-is preserved."""
+    # Start/restart the workflow with name REG.
+    $ cylc play REG
 
-RESTART_DOC = r"""cylc [control] restart [OPTIONS] ARGS
+    # Install $PWD/flow.cylc as $(basename $PWD) and start it.
+    $ cylc play
```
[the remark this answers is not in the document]
MetRonnie: So I should make the `[REG]` arg non-optional?

cylc/flow/scheduler_cli.py
```diff
@@ -113,63 +94,33 @@ def get_option_parser(is_restart, add_std_opts=False):
...
-            ),
-            metavar="CYCLE_POINT", action="store", dest="startcp")
+    parser.add_option(
+        "--start-cycle-point", "--startcp",
+        help=(
+            "Set the start cycle point, which may be after the initial cycle"
+            "point. If the specified start point is not in the sequeunce, the"
+            "next on-sequence point will be used."
+            "(Not to be confused with the initial cycle point.)"
+        ),
```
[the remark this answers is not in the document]
MetRonnie: Good spot

tests/functional/restart/51-ignore-final-point.t
MetRonnie: Just to make it run quicker

cylc/flow/scheduler_cli.py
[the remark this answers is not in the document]
oliver-sanders: Yes please, it no longer makes sense since the change from registration to installation.

This feature will now only work if you are in the run directory for a suite which has either been written there or `cylc installed` and will fail in unhelpful ways otherwise.

MetRonnie:
Tests re-run timed out. but I ran them locally and they all pass

hjoliver:
The same test batch has suddenly started timing out on #4035, after rebase.  There might be a test broken by some recent change on master (if this branch is up to date too?)

MetRonnie:
I think it's just going to need a longer timeout on GH Actions, that's all. It's running the whole test battery in that job, including the flakyfunctional, and possibly the poll tests take longer than tcp.

Having said that, I've looked through the timed out test runs, and it seems there's more often than not a strange delay at the last couple of tests. Not sure why, doesn't seem to be these 2 tests in particular.
```
[17:33:28] tests/f/events/11-cycle-task-event-job-logs-retrieve.t ..... ok   314944 ms ( 0.02 usr  0.03 sys + 161.12 cusr 18.52 csys = 179.69 CPU)
[17:39:09] tests/f/remote/03-polled-task-started.t ..... ok  1632815 ms ( 0.30 usr  1.48 sys + 876.69 cusr 105.05 csys = 983.52 CPU)
```

oliver-sanders:
Yeah that's not a good sign, if a such a simple test takes so long then something is likely going badly wrong. Sadly there isn't enough evidence to tell what.

I've been stabilising a few tests and added a default workflow timeout in a branch I'm working on at the moment which might potentially help but I'm not sure.

I've seen to outrageously long test runs on polling systems though not as bad as that!

cylc/flow/scheduler_cli.py
```diff
@@ -417,35 +365,9 @@ async def _run(parser, options, reg, is_restart, scheduler):
...
-
-
-@cli_function(partial(get_option_parser, is_restart=False))
-def run(parser, options, *args):
-    """Implement cylc run."""
+@cli_function(get_option_parser)
+def play(parser, options, *args):
+    """Implement cylc play."""
     if not args:
         _auto_install()
```
oliver-sanders, line 372: Can kill this and the `_auto_install` bit now the optional `REG` is no longer optional.

cylc/flow/scheduler.py
hjoliver: (Sorry, missed this response a week ago ... but it's fine to punt it to the new issue)

cylc/flow/scheduler_cli.py
```diff
@@ -42,67 +42,45 @@
...
+initial cycle point (specified in flow.cylc or on the command line). Any
+dependence on tasks prior to the initial cycle point is ignored.
+It is also possible to start from a point that is later than the initial cycle
+point, using the option --startcp=CYCLE_POINT. The initial cycle point is
+preserved, but the workflow does not start there and instead starts part-way
+through the graph (historically known as a "warm start").
 
-RESTART_DOC = r"""cylc [control] restart [OPTIONS] ARGS
+A "restart" continues the workflow from the earliest cycle point that was
+incomplete when shutdown occurred. Tasks recorded as submitted or running are
```
hjoliver, line 68:
```suggestion
A "restart" continues on from the most recent recorded state of the workflow. 
Tasks recorded as submitted or running are
```
I know what you mean, but "continues from the [blah] cycle point ..." could be confusing.  We've traditionally described warm starts that way. A restart often involves tasks at multiple cycle points.

hjoliver (approved):
Looks good, tests as working :tada: 

Still one @oliver-sanders comment to respond to? (on the `SuiteServiceFileError` exception)

MetRonnie:
Both should be sorted now

cylc/flow/scheduler_cli.py
oliver-sanders:
```diff
diff --git a/cylc/flow/scheduler_cli.py b/cylc/flow/scheduler_cli.py
index 2a594e9d1..484940b54 100644
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -207,19 +207,6 @@ RunOptions = Options(
     get_option_parser(add_std_opts=True), DEFAULT_OPTS)
 
 
-def _auto_install():
-    """Register workflow installed in the cylc-run directory"""
-    try:
-        reg = suite_files.register()
-    except SuiteServiceFileError as exc:
-        sys.exit(exc)
-    # Replace this process with "cylc play REG ..." for 'ps -f'.
-    os.execv(
-        sys.argv[0],
-        [sys.argv[0]] + sys.argv[1:] + [reg]
-    )
-
-
 def _open_logs(reg, no_detach):
     """Open Cylc log handlers for a flow run."""
     if not no_detach:
```

cylc/flow/scheduler.py
```diff
@@ -172,7 +172,6 @@ class Scheduler:
     contact_data: Optional[dict] = None
 
     # run options
-    is_restart: bool = False
```
oliver-sanders, line 175: The idea of specifying everything up here is so that we can see all of the Scheduler attributes in one place. Can always default to `None` until set later.

oliver-sanders (approved):
Another Cylc8 brick in the wall, thanks @MetRonnie!

(Small comments above about an unused code block and removed dataclass attr definition).

Ooh, just spotted, a reference to `cylc run` has snuck in probs with rebase in `cylc.flow.scripts.install#24`.

MetRonnie:
Have amended last commit to address those

hjoliver (approved):
Thanks @MetRonnie, good job :tada:

PR #4041, opened 2021-01-20
Platforms.fix generic items match

wxtim:

It was noted by @MetRonnie that the function `get_platform_from_job_info`: Logic checking that items in the job section of the old style config weren't being checked for a match/mismatch against platform spec.

- [x] Appropriate tests are included: I refactored part of the existing code into a more easily tested unit.
- [x] No change log entry required: Small features that should be non-user facing.

### DIFF
--- a/tests/unit/test_platforms.py
+++ b/tests/unit/test_platforms.py
@@ -460,6 +460,12 @@ def test_get_install_target_to_platforms_map(
             {'captain': 'Picard'},
             {'ship': 'Defiant'},
             False
+        ),
+        (
+            {'captain': 'Picard', 'ship': 'Enterprise'},
+            {'captain': 'Picard'},
+            {},
+            True
         )
     ]
 )
--- a/tests/unit/test_platforms.py
+++ b/tests/unit/test_platforms.py
@@ -20,7 +20,9 @@ import pytest
 from cylc.flow.parsec.OrderedDict import OrderedDictWithDefaults
 from cylc.flow.platforms import (
     platform_from_name, platform_from_job_info,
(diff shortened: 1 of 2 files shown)


kinow (approved):
:+1:

tests/unit/test_platforms.py
```diff
@@ -414,3 +428,28 @@ def test_get_install_target_to_platforms_map(
...
+            {'captain': 'Picard'},
+            {},
+            False
+        ),
+        (
+            {'captain': 'Sisko'},
+            {},
+            {'captain': 'Janeway'},
+            False
+        )
```
MetRonnie, line 451: I know it's a simple piece of logic being tested, but for completeness, what about a couple more cases like:
```
(
    {'captain': 'Picard', 'ship': 'Enterprise'},
    {'captain': 'Picard'},
    {'ship': 'Enterprise'},
    True
),
(
    {'captain': 'Picard', 'ship': 'Enterprise'},
    {'captain': 'Picard'},
    {'ship': 'Defiant'},
    False
)
```
(if that's indeed what the expected return values should be, I haven't checked)

MetRonnie (commented):
Looks like it solves the issue, I've just added my two cents

tests/unit/test_platforms.py
wxtim: Yes please. These both look like reasonable cases.

MetRonnie (approved):

MetRonnie:
Functional test job timed out, but that should be fine, it passed for the previous run: https://github.com/cylc/cylc-flow/actions/runs/501247553

PR #4043, opened 2021-01-21
fix actions

oliver-sanders:
Whilst debugging test failures I found a number of errors in the actions setup, mostly my fault, largely rebase errors, need to stop rushing.

Added a couple debugging enhancements:

    * use short job names
    * expand text matrix for clarity
    * make the macos jobs run on macos
    * fix macos installation
    * fix cylc-run dir upload
    * use different artifact name for each job to avoid merging
    * rsync remote cylc-run dirs before the upload
    * rerun failed tests with -j 1 -v for debug niceness
    * set timeout on the step not the job to factor out install time
      fluctuations

Should review carefully and keep an eye on actions runs for a while to make sure nothing else has slipped the net.

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/.github/workflows/test_functional.yml
+++ b/.github/workflows/test_functional.yml
@@ -137,6 +137,7 @@ jobs:
         timeout-minutes: 35
         continue-on-error: true
         run: |
+          echo "::set-output name=finished::false"
           if [[ '${{ matrix.test-base }}' == 'tests/k' ]]; then
               NPROC=4
           else
@@ -147,17 +148,22 @@ jobs:
               -j "${NPROC}" \
               --state=save \
               ${{ matrix.test-base }} \
-              || (touch tests-did-not-time-out && false)
+              || (echo "::set-output name=finished::true" && false)
+
+      - name: Time Out
+        if: steps.test.outcome == 'failure' && steps.test.outputs.finished != 'true'
+        run: |
+          echo '::error:: tests timed-out'
+          # help to identify the tests that were running at the time
+          cylc scan --state=all --format=rich --color-blind
+          # fail the workflow
+          false
 
       - name: Re-run failed tests
         timeout-minutes: 10
-        if: steps.test.outcome == 'failure'
+        if: steps.test.outcome == 'failure' && steps.test.outputs.finished == 'true'
         run: |
-          if [[ ! -f tests-did-not-time-out ]]; then
-            echo "the preious test run timed out -> not re-running tests"
-            # this false causes the workflow to fail
-            false
-          fi
+          # re-run failed tests providing that they didn't time out first time
           # TODO: make the tests deterministic so we don't need to do this
           etc/bin/run-functional-tests \
               -j 1 \
@@ -178,7 +184,6 @@ jobs:
         if: failure() && steps.test.outcome == 'failure'
         timeout-minutes: 1
         run: |
-          cylc scan --state=all --format=rich --color-blind
           find "$HOME/cylc-run" -name '*.err' -type f \
               -exec echo '====== {} ======' \; -exec cat '{}' \;
           find "$HOME/cylc-run" -name 'log.*' -type f \
--- a/.github/workflows/test_functional.yml
+++ b/.github/workflows/test_functional.yml
@@ -1,38 +1,56 @@
 name: functional tests
 
-on: [pull_request]
+on: [pull_request, workflow_dispatch]
 
 jobs:
   test:
-    runs-on: ubuntu-latest
-    timeout-minutes: 35
+    runs-on: ${{ matrix.os }}
+    name: ${{ matrix.name || matrix.chunk }}
+    timeout-minutes: 40
     strategy:
       fail-fast: false
       matrix:
         os: ['ubuntu-latest']
         python-version: ['3.7']
-        tests: [
-          # local tests
-          ['tests/f', '1/4', '_local_background* _local_at*'],
-          ['tests/f', '2/4', '_local_background* _local_at*'],
-          ['tests/f', '3/4', '_local_background* _local_at*'],
-          ['tests/f', '4/4', '_local_background* _local_at*'],
-          ['tests/k', '1/1', '_local_background* _local_at*'],
-          # remote tests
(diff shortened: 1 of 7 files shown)


.github/workflows/test_functional.yml
```diff
@@ -56,9 +74,17 @@ jobs:
...
+            >> "${GITHUB_PATH}"
+          echo \
+            "/usr/local/opt/gnu-sed/libexec/gnubin" \
+            >> "${GITHUB_PATH}"
+
+          # add coreutils to the bashrc too (for jobs)
+          cat >> "${HOME}/.bashrc" <<__HERE__
+          PATH="$(brew --prefix)/opt/coreutils/libexec/gnubin:/usr/local/opt/gnu-sed/libexec/gnubin:$PATH"
+          export PATH
           __HERE__
```
oliver-sanders, line 88: Needed because the way actions sets the PATH doesn't make it as far as job scripts.

Sadly setting the PATH in this way alone doesn't work properly in actions steps so need both.

.github/workflows/test_functional.yml
```diff
@@ -106,54 +133,79 @@ jobs:
...
+          if [[ ! -f tests-did-not-time-out ]]; then
+            echo "the preious test run timed out -> not re-running tests"
+            # this false causes the workflow to fail
+            false
+          fi
+          # TODO: make the tests deterministic so we don't need to do this
+          etc/bin/run-functional-tests \
+              -j 1 \
+              -v \
+              --state=save,failed ${{ matrix.test-base }}
```
oliver-sanders, line 165: This re-runs any failed tests with `-j 1 -v` which will give us much more useful information for debugging.

I had resisted doing this before as it makes it possible for flaky tests that are not parallel-safe to pass as they get run in series the second time around. However, it's not like the way we were doing it before was much better from that perspective. Just need to keep an eye on flaky tests and look to improve/replace them where possible.

.github/workflows/test_functional.yml
```diff
@@ -106,54 +133,79 @@ jobs:
...
+
+      - name: Copy cylc-run out of container
+        if: failure() && steps.test.outcome == 'failure' && contains(matrix.platform, '_remote')
+        run: |
+          # pick the first host in the list
+          host="$(cut -d ' ' -f 1 <<< "${{ matrix.platform }}")"
+          # copy back the remote cylc-run dir
+          rsync -av \
+            "${host}:/root/cylc-run/" \
+            "${HOME}/cylc-run/${host}/"
```
oliver-sanders, line 175: This means that you should expect to find the remote container's `cylc-run` dir in the uploaded artefact.

tests/unit/main_loop/test_auto_restart.py
```diff
@@ -121,12 +121,18 @@ def test_should_auto_restart(
...
 ):
     """Ensure the suite only auto-restarts when appropriate."""
+    # factor out networking and FQDNs for testing purposes
+    monkeypatch.setattr(
+        'cylc.flow.main_loop.auto_restart.get_fqdn_by_host',
+        lambda x: x
+    )
     # mock a scheduler object
     scheduler = Mock(
-        host=get_fqdn_by_host(host),
```
oliver-sanders, line 129: Seemed to be having DNS config issues, simple solution is to factor out DNS from this unit test, there are functional tests to cover that.

MetRonnie (approved):
Couple of comments, otherwise looks good! The re-run of failed tests worked in e.g. the macos run https://github.com/cylc/cylc-flow/runs/1741440821?check_suite_focus=true

kinow (approved):

.github/workflows/test_functional.yml
```diff
@@ -106,54 +133,79 @@ jobs:
...
-              --state=save ${BASE} \
-          || (
-              echo '=== Re-running Failed Tests ==='
-              CHUNK= etc/bin/run-functional-tests \
-                  -j "${NPROC}" \
-                  --state=save,failed
-          )
+              --state=save \
+              ${{ matrix.test-base }} \
+              || (touch tests-did-not-time-out && false)
```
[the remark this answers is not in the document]
oliver-sanders: See what you make of https://github.com/cylc/cylc-flow/pull/4043/commits/7823ab4a7f5559f6705ff70a12a7523fb506381d

.github/workflows/test_functional.yml
```diff
@@ -147,17 +148,13 @@ jobs:
...
-        if: steps.test.outcome == 'failure'
+        if: steps.test.outcome == 'failure' && steps.test.outputs.finished == 'true'
         run: |
-          if [[ ! -f tests-did-not-time-out ]]; then
-            echo "the preious test run timed out -> not re-running tests"
-            # this false causes the workflow to fail
-            false
-          fi
+          # re-run failed tests providing that they didn't time out first time
           # TODO: make the tests deterministic so we don't need to do this
```
MetRonnie: Not sure if the `continue-on-error` for the step `test` will cause the overall workflow to be marked as passed if `test` times out. It might be safer to keep the `if` inside the bash script, i.e.
```
if [[ ! ${{ steps.test.outputs.finished == 'true' } ]]; then
  echo "the preious test run timed out -> not re-running tests"
...
```

oliver-sanders: Derp, that was half finished - 4910e40

MetRonnie (approved):
Will leave to you to merge after tests finish

oliver-sanders:
Tests passed, keep a sharp eye out for any unexpected behaviour, especially the last commit...

oliver-sanders:
The [flaky tests](https://github.com/cylc/cylc-flow/pull/4043/checks?check_run_id=1770474391) went to rerun so that seems to be working fine.

MetRonnie:
@oliver-sanders In this test run the install step on MacOS took 10 minutes, so the workflow was cancelled after 40 mins while the tests were still running: https://github.com/cylc/cylc-flow/pull/4050/checks?check_run_id=1785691809. Might need to increase the timeouts

oliver-sanders:
Wow, what was going on there, 10mins for a `pip install`, that is not normal!

PR #4046, opened 2021-01-25
Shutdown: remove contact file after closing DB connection

MetRonnie:

Detecting whether a workflow is still running or has shut down relies on checking for the existence of a contact file. However, before this PR, when shutting down a workflow, the contact file would get removed before shutting the database, which led to the small possiblity of problems (e.g. if the database is deleted in this short time, it would get regenerated).

This PR reverse the order, so any DB connections are definitely closed when the contact file is removed and the workflow is considered to have shut down.

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -214,6 +214,9 @@ Python 3.8+.
 a trailing slash in the suite/workflow name would cause `cylc stop`
 (and possibly other commands) to silently fail.
 
+[#4046](https://github.com/cylc/cylc-flow/pull/4046) - Fix bug where a workflow
+database could still be active for a short time after the workflow stops.
+
 -------------------------------------------------------------------------------
 ## __cylc-8.0a2 (2020-07-03)__
 
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1642,7 +1642,7 @@ class Scheduler:
             self.server.stop()
         if self.publisher:
             await self.publisher.publish(
-                [(b'shutdown', f'{str(reason)}'.encode('utf-8'))]
+                [(b'shutdown', str(reason).encode('utf-8'))]
             )
             self.publisher.stop()
         self.curve_auth.stop()  # stop the authentication thread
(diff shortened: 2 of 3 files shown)


hjoliver (commented):
@MetRonnie this fix sounds sensible but I can't get the new functional test to fail on master, which makes me wonder if there's any point in having the test. I wonder if there is any way to ensure there is still a connection to the DB at the time...

MetRonnie:
It fails pretty consistently for me on master locally... might be because our filesystem is so slow 😛

oliver-sanders (approved):
Happy with the change.

Would be a good idea to add a comment to the code that removes the contact file highlighting that this should be the last thing the flow should do before shutting down (logging excluded).

If the test is platform dependent, either comment it to explain that or omit it. I think this is just fundamentally a very difficult test to write as the DB closes and reopens so tracking file handles isn't enough and sampling the filesystem could fail to detect a fleeting re-appearance anyway.

MetRonnie:
Perhaps the removal of the contact file should come after this bit too? https://github.com/cylc/cylc-flow/blob/9d92abb93f2cce36ff6cc021bddd14df566c4a88/cylc/flow/scheduler.py#L1677-L1684

That would make it the very last part of `Scheduler.shutdown()`

MetRonnie:
Hopefully if the test passes on master on your machine then the bug too is very unlikely to affect your machine.

oliver-sanders:
I think it makes sense for the shutfown/aborted event handlers to run after the contact file removal as that is the thing that marks the death of the flow as far as the user is concerned.

E.G. if your shutdown handler ran `cylc scan` you wouldn't expect the flow that fired the event to appear in the `cylc scan` results.

MetRonnie:
> Happy with the change.
> 
> Would be a good idea to add a comment to the code that removes the contact file highlighting that this should be the last thing the flow should do before shutting down (logging excluded).
> 
> If the test is platform dependent, either comment it to explain that or omit it. I think this is just fundamentally a very difficult test to write as the DB closes and reopens with normal usage (horrible) so tracking file handles isn't enough and sampling the filesystem could fail to detect a fleeting re-appearance anyway.
> 
> (likely easier to spot on a slow filesystem as unlinking a file is much faster than writing a database).

Addressed and force-pushed

hjoliver (approved):

PR #4048, opened 2021-01-26
tui: update help for MacOS

oliver-sanders:
Keybinding is different in MacOS Terminal.app.

### DIFF
--- a/cylc/flow/tui/overlay.py
+++ b/cylc/flow/tui/overlay.py
@@ -38,6 +38,7 @@ Parameters:
 """
 
 from functools import partial
+import sys
 
 import urwid
 
@@ -144,7 +145,10 @@ def help_info(app):
 
     # mouse interaction
     items.extend([
-        urwid.Text('Shift+Click to select text'),
+        urwid.Text(
+            'fn + ⌥ & click to select text' if sys.platform == 'darwin' else
+            'Shift & click to select text'
+        ),
         urwid.Divider()
     ])
 


MetRonnie (approved):

MetRonnie:
By the way, at what point are these codecov errors for untouched code going to go away?

oliver-sanders:
I was hoping that codecov would settle down after a while, doesn't look like it has, here's the config file - https://github.com/cylc/cylc-flow/blob/master/.codecov.yml.

I explicitly told it not to comment but it appears to have found a way to annotate instead. I don't get why it fails to keep its analysis within the diff.

oliver-sanders:
Looking at the codecov report I can see 12 coverage files corresponding to the 8 functional and 4 fast runs so it's not like any reports got lost in the post:

https://codecov.io/gh/cylc/cylc-flow/commit/e94c5508949c1406b95ada216f5ad3165a300314/build

oliver-sanders:
The functional tests, being somewhat timing sensitive are not totally deterministic, if one fails and re-runs, that could effect the coverage or if the moving parts get into a certain state in one run that they don't usually, that sort of thing.

This could cause a little jitter, the line in this PR could be a result of that.

MetRonnie:
The thing is, every PR says "missing base report" - https://app.codecov.io/gh/cylc/cylc-flow/pulls?page=1&order=-pullid

Could it be this problem? https://github.com/commercetools/nodejs/issues/179#issuecomment-309806395 Could try running coverage locally on master?

oliver-sanders:
Could be, I've no idea really, the docs seem to back that up.

https://docs.codecov.io/docs/error-reference

Try it!

MetRonnie:
Another thing that might be it?
```

MetRonnie:
I've run the unit/integration/doc tests locally and uploaded the coverage, will run the func tests overnight

oliver-sanders:
Presumably once we've got a base report up there it will be continuously updated?

Note the results you'll get with the functional tests may differ from what GH actions mesares unless you run the docker test matrix (the MacOS build shouldn't make a difference).

Issue #3233, opened 2019-07-24
TestHostAppointer.test_remove_bad_hosts unit test fails locally when run within the IDE


kinow:
**Describe the bug**

The `TestHostAppointer.test_remove_bad_hosts` unit test is failing in `_get_host_metrics` when executed in the IDE. I suspect it is due to terminal colours tokens added, which could mean that in the IDE the current approach to identify whether it is running in a terminal or not is missing something.

**Release version(s) and/or repository branch(es) affected?**

`master`

**Steps to reproduce the bug**

Simply run `TestHostAppointer` unit test in PyCharm (not sure if it would pass on Visual Studio/Eclipse/etc). The tests pass fine on the terminal.

**Expected behavior**

Successfully run and debug the tests.

**Screenshots**

![image](https://user-images.githubusercontent.com/304786/61756829-04854d00-ae12-11e9-919b-ccb1e698aa90.png)

**Additional context**

Value copied while debugging the test:

```
'{
    "load:1": 0.86,
    "load:5": 0.43,
    "load:15": 0.56,
    "memory": 1610200
}
[0m'
```

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

kinow:
The test above was removed. But there are two new tests, also related to host selection, that fail now. Same kind of error, spurious character at the end of the string when tests are run under PyCharm.

![image](https://user-images.githubusercontent.com/304786/79401298-d0c00800-7fdc-11ea-88da-b1eadde5c20b.png)

![image](https://user-images.githubusercontent.com/304786/79401304-d3baf880-7fdc-11ea-9d91-d2e27bcebfb9.png)

oliver-sanders:
Closed by #3489

kinow:
And tested the PR #3570 , which fixes the unit tests in the IDE.

![image](https://user-images.githubusercontent.com/304786/82955325-b37e5080-a002-11ea-99b1-3d6d7efa38db.png)

kinow:
Tested today after SoD and other PR's. Thought maybe it had been fixed. But still similar issue, now in two tests.

![image](https://user-images.githubusercontent.com/304786/88865637-86495b00-d25c-11ea-941a-761c165ab268.png)

oliver-sanders:
Ok, so there are two things here.

1) The pycharm terminal behaves differently to normal terminals.
[...]

PR #4050, opened 2021-01-27
Fix Remote Tidy

datamel:
I have kept basic functionality. This will need looking at again once the intelligent host command has been implemented, so we can adapt the tidy in case where platform fails. I have left this as a basic fix as this may be replaced by some version of `cylc clean`.

- [x] Tests added - functional and unit.
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -465,8 +465,9 @@ def get_all_platforms_for_install_target(install_target):
     all_platforms = glbl_cfg(cached=True).get(['platforms'], sparse=False)
     for k, v in all_platforms.iteritems():
         if (v.get('install target', k) == install_target):
-            v['name'] = k
-            platforms.append(v)
+            v_copy = deepcopy(v)
+            v_copy['name'] = k
+            platforms.append(v_copy)
     return platforms
 
 
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -457,3 +457,20 @@ def is_platform_with_target_in_list(install_target, distinct_platforms_list):
     """Determines whether install target is in the list of platforms"""
     for distinct_platform in distinct_platforms_list:
         return install_target == distinct_platform['install target']
+
+
+def get_all_platforms_for_install_target(install_target):
+    """Return list of platform dictionaries for given install target."""
+    platforms = []
+    all_platforms = glbl_cfg(cached=True).get(['platforms'], sparse=False)
+    for k, v in all_platforms.iteritems():
+        if (v.get('install target', k) == install_target):
+            v['name'] = k
+            platforms.append(v)
+    return platforms
+
+
+def get_random_platform_for_install_target(install_target):
+    """Return a randomly selected platform (dict) for given install target."""
+    platforms = get_all_platforms_for_install_target(install_target)
+    return random.choice(platforms)
--- a/cylc/flow/task_remote_mgr.py
+++ b/cylc/flow/task_remote_mgr.py
@@ -226,7 +226,7 @@ class TaskRemoteMgr:
             cmd = ['remote-tidy']
             if cylc.flow.flags.debug:
                 cmd.append('--debug')
-            cmd.append(str(f'{install_target}'))
+            cmd.append(install_target)
             cmd.append(get_remote_suite_run_dir(platform, self.suite))
             cmd = construct_ssh_cmd(cmd, platform, timeout='10s')
             LOG.debug(
--- a/cylc/flow/task_remote_mgr.py
+++ b/cylc/flow/task_remote_mgr.py
@@ -32,7 +32,7 @@ from time import time
 from cylc.flow import LOG, RSYNC_LOG
 from cylc.flow.exceptions import TaskRemoteMgmtError
 import cylc.flow.flags
-from cylc.flow.hostuserutil import (is_remote_host, is_remote_platform)
+from cylc.flow.hostuserutil import is_remote_host
 from cylc.flow.pathutil import (
     get_remote_suite_run_dir,
     get_dirs_to_symlink,
@@ -46,9 +46,7 @@ from cylc.flow.suite_files import (
     KeyType,
     get_suite_srv_dir,
     get_contact_file)
-from cylc.flow.platforms import (
-    get_platform,
-    get_install_target_from_platform)
+from cylc.flow.platforms import get_random_platform_for_install_target
 from cylc.flow.remote import construct_ssh_cmd
 
 # Remote installation literals
@@ -171,7 +169,6 @@ class TaskRemoteMgr:
 
         """
         install_target = platform['install target']
-
         # Set status of install target to in progress while waiting for remote
         # initialisation to finish
         self.remote_init_map[install_target] = REMOTE_INIT_IN_PROGRESS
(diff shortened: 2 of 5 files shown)


MetRonnie (changes requested):
Just making a note: as discussed just now, will need to address the `task_remote_mgr.py` remote init map issue

cylc/flow/task_remote_mgr.py
```diff
@@ -219,21 +216,22 @@ def remote_tidy(self):
...
+                continue
+            if install_target == 'localhost':
                 continue
+            platform = get_random_platform_for_install_target(install_target)
+            platform_n = platform['name']
             cmd = ['remote-tidy']
             if cylc.flow.flags.debug:
                 cmd.append('--debug')
-            cmd.append(str(f'{self.install_target}'))
+            cmd.append(str(f'{install_target}'))
```
MetRonnie, line 229:
```suggestion
            cmd.append(install_target)
```

tests/functional/cylc-ping/04-check-keys-remote.t
```diff
@@ -37,19 +37,23 @@ __FLOW_CONFIG__
...
 SSH='ssh -n -oBatchMode=yes -oConnectTimeout=5'
+ 
 ${SSH} "${CYLC_TEST_HOST}" \
 find "${RSRVD}" -type f -name "*key*"|awk -F/ '{print $NF}'|sort >'find.out'
-cmp_ok 'find.out' <<__OUT__
+sort >'keys' <<__OUT__
+client_${CYLC_TEST_INSTALL_TARGET}.key
 client.key_secret
-client_$CYLC_TEST_PLATFORM.key
 server.key
```
MetRonnie, line 51: I don't think this sort is necessary, as you're sorting `client.*` and `client_*` which should always have the same order anyway

tests/unit/test_platforms.py
```diff
@@ -471,3 +474,54 @@ def test_get_install_target_to_platforms_map(
...
+                 'install target':'forest',
+                 'name':'merida'},
+                {'hosts': ['oak', 'elm', 'fir'],
+                 'name':'forest'}]
+    assert actual == expected
+    platforms = [{'hosts': ['snow', 'ice', 'sparkles'],
+                  'install target':'arendelle',
+                  'name':'olaf'}, {'hosts': ['anna', 'elsa', 'hans'],
+                                   'install target':'arendelle',
+                                   'name':'kristoff'}]
```
MetRonnie, line 526: Nitpick: this could be formatted a bit more clearly

cylc/flow/platforms.py
```diff
@@ -457,3 +457,20 @@ def is_platform_with_target_in_list(install_target, distinct_platforms_list):
...
         return install_target == distinct_platform['install target']
+
+
+def get_all_platforms_for_install_target(install_target):
+    """Return list of platform dictionaries for given install target."""
+    platforms = []
+    all_platforms = glbl_cfg(cached=True).get(['platforms'], sparse=False)
+    for k, v in all_platforms.iteritems():
+        if (v.get('install target', k) == install_target):
+            v['name'] = k
```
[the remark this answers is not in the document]
datamel: Platform name is not already set, since the platform name comes from a different tier of the config.

tests/functional/cylc-ping/04-check-keys-remote.t
```diff
@@ -37,19 +37,22 @@ __FLOW_CONFIG__
...
 RSRVD="${RRUND}/.service"
 poll_grep_suite_log 'Holding all waiting or queued tasks now'
 SSH='ssh -n -oBatchMode=yes -oConnectTimeout=5'
+ 
 ${SSH} "${CYLC_TEST_HOST}" \
 find "${RSRVD}" -type f -name "*key*"|awk -F/ '{print $NF}'|sort >'find.out'
 cmp_ok 'find.out' <<__OUT__
+client_${CYLC_TEST_INSTALL_TARGET}.key
 client.key_secret
-client_$CYLC_TEST_PLATFORM.key
```
MetRonnie, line 49: Need to go back to the original order. I guess `.` comes before `_` alphabetically

tests/functional/cylc-ping/04-check-keys-remote.t
datamel: I have removed the sort. The test passes locally but fails GH Actions: 
image

Putting the sort back fixes this I think.

tests/functional/cylc-ping/04-check-keys-remote.t
datamel: Ah, changing order will break the test locally. Needs the sort.

MetRonnie: This order passes both on GH Actions and locally (can't imagine `sort` works differently on RHEL and Ubuntu):
```
cmp_ok 'find.out' <<__OUT__
client.key_secret
client_${CYLC_TEST_INSTALL_TARGET}.key
server.key
__OUT__
```
Anyway, whatever works.

datamel: Interesting that you are getting that passing locally...
image

MetRonnie (approved):

tests/functional/cylc-ping/04-check-keys-remote.t
MetRonnie: Well maybe `sort` behaves differently on RHEL and RHEL 😢

oliver-sanders: Perhaps consider `sort -d` if behaviour is inconsistent.

oliver-sanders: IDE virtual terminals don't can exhibit some very strange behaviour, I remember fighting some strange buffering behaviour in eclipse and pycharm before I moved to using regular terminals (but hey, learned a bit about text buffering in the process 🤷).

Here are two other recent cases of test failures in cylc-flow caused by IDE terminals:

* https://github.com/cylc/cylc-flow/pull/3989
* https://github.com/cylc/cylc-flow/issues/3233

cylc/flow/platforms.py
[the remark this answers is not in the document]
oliver-sanders: This isn't the nicest thing to do as it modifies the global config itself.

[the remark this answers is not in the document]
MetRonnie: That sort of thing seems to happen a lot in `platforms.py`. E.g. the reason I though `name` was already set is because `platform_from_name()` does that: https://github.com/cylc/cylc-flow/blob/951f265636ee9f828248f832c30ed523ef164b35/cylc/flow/platforms.py#L201-L202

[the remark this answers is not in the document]
oliver-sanders: That example doesn't mutate the global config itself because a few lines earlier it calls `deepcopy` to duplicate the data, it is the copy that it mutates:

https://github.com/cylc/cylc-flow/blob/951f265636ee9f828248f832c30ed523ef164b35/cylc/flow/platforms.py#L189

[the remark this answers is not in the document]
datamel: Trouble is, `platform_from_name()`  makes calls out to `deepcopy`, as such, `platform['name']` is not set by the time we need it here. I'll put up a code change to prevent my change modifying the global config but do we need the deep copies here? I think we could revisit this function, removing the `deepcopy` calls, which would have the following benefits: 

- more efficient given there are a fair few calls to it
- we would be able to set the platform name, and things like install target, early, for use elsewhere  

One of the comments with a `deepcopy` states that it is used because:

>   Deepcopy prevents contaminating platforms with data
>   from other platforms matching platform_name_re  

Unless I'm missing something, I don't think this is an issue in a for loop - we can set values for keys in each element of a for loop independently?

datamel:
I've broken the unit test for this. I'm on it.

cylc/flow/platforms.py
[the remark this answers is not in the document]
MetRonnie: I think Oliver mentioned yesterday in our meeting that we could instead try setting things like `name` on the initial load of the config only

oliver-sanders (approved):

PR #4061, opened 2021-02-04
mypy

oliver-sanders:
Now that we are starting to accumulate some type annotations in the codebase we ought to make sure that they are valid. No documentation is better than broken documentation.

This adds mypy to the CI and fixes outstanding issues to get it passing. We can add type annotations incrementally from here on.

**Requirements check-list**
- [x] No dependency changes.


cylc/flow/c3mro.py
```diff
@@ -150,7 +150,7 @@ def mro(self, C):
 
 
 if __name__ == "__main__":
-    parents = {}
+    parents: dict = {}
```
oliver-sanders, line 153: Have to add type annotations to the odd dict or two to keep mypy happy.

cylc/flow/data_messages_pb2.py
```diff
@@ -1,3 +1,4 @@
+# type: ignore
```
oliver-sanders, line 1: We can't configure per-file ignores at present so we have to stick a comment at the top of it.

cylc/flow/network/schema.py
```diff
@@ -1455,10 +1455,11 @@ class Meta:
 
     class Arguments:
         workflows = List(WorkflowID, required=True)
+
         mode = BroadcastMode(
             # use the enum name as the default value
             # https://github.com/graphql-python/graphql-core-legacy/issues/166
-            default_value=BroadcastMode.Set.name,
+            default_value=BroadcastMode.Set.name,  # type: ignore
```
oliver-sanders, line 1462: mypy doesn't like Graphene enums.

cylc/flow/suite_files.py
```diff
@@ -678,10 +678,10 @@ def remote_clean(reg, platform_names, timeout):
             if ret_code:
                 # Try again using the next platform for this install target:
                 this_platform = platforms.pop(0)
-                exc = TaskRemoteMgmtError(
+                excn = TaskRemoteMgmtError(
```
oliver-sanders, line 681: mypy doesn't like vars called `exc` unless they are in `except:` branches.

hjoliver: Huh, that's an interesting one

cylc/flow/network/graphql.py
```diff
@@ -44,7 +45,7 @@
 
 STRIP_ARG = 'strip_null'
 NULL_VALUE = None
-EMPTY_VALUES = ([], {})
+EMPTY_VALUES: Tuple[list, dict] = ([], {})
```
MetRonnie, line 48: Is it possible to be more specific here, e.g. `Tuple[List[int], Dict[str, str]]` or whatever it's meant to be?

cylc/flow/xtrigger_mgr.py
```diff
@@ -126,16 +126,16 @@ def __init__(
         # For function arg templating.
         if not user:
             user = get_user()
-        self.farg_templ = {
+        self.farg_templ: dict = {
```
MetRonnie, line 129:
```suggestion
        self.farg_templ: Dict[str, str] = {
```

cylc/flow/suite_files.py
MetRonnie: I think it's because there's an earlier line `except Exception as exc:`. https://github.com/python/mypy/issues/5080

MetRonnie (commented):
Instead of just `x: list = []` or `x: dict = {}`, I think it's more useful to specify the type of the entries, e.g. `x: List[str] = []` (needs `from typing import List`)

oliver-sanders:
Agreed, however, the aim of this PR is to get `mypy` passing so that future type annotations are checked (so we don't add broken ones) rather than to upgrade the codebase now.

MetRonnie (approved):
Ok

hjoliver (approved):

PR #4065, opened 2021-02-08
Fix and tidy functional tests

MetRonnie:

**Requirements check-list**
- [x] Does not need tests.
- [x] No dependency changes.

### DIFF
--- a/tests/functional/lib/bash/test_header
+++ b/tests/functional/lib/bash/test_header
@@ -245,8 +245,7 @@ suite_run_ok() {
     SUITE_RUN_FAILS=true
     SUITE_LOG_DIR="${SUITE_RUN_DIR}/log/suite"
     mkdir -p "${SUITE_LOG_DIR}" # directory may not exist if run fails very early
-    cp -p "${TEST_NAME}.stdout" "${SUITE_LOG_DIR}/out"
-    cp -p "${TEST_NAME}.stderr" "${SUITE_LOG_DIR}/err"
+    dump_std "${TEST_NAME}"
 }
 
 suite_run_fail() {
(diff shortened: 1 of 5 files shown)


hjoliver (approved):

tests/functional/lib/bash/test_header
```diff
@@ -245,8 +245,7 @@ suite_run_ok() {
     SUITE_RUN_FAILS=true
     SUITE_LOG_DIR="${SUITE_RUN_DIR}/log/suite"
     mkdir -p "${SUITE_LOG_DIR}" # directory may not exist if run fails very early
-    cp -p "${TEST_NAME}.stdout" "${SUITE_LOG_DIR}/out"
-    cp -p "${TEST_NAME}.stderr" "${SUITE_LOG_DIR}/err"
+    dump_std "${TEST_NAME}"
```
oliver-sanders, line 248: This isn't an error, this is just where the out and error are put for flow runs.

https://github.com/cylc/cylc-flow/blob/c9b9ae99a0bbf785f51f03b049e97538c2490eb8/tests/functional/lib/bash/test_header#L159

MetRonnie: But the logs are already available there anyway?

oliver-sanders: Stderr is typically the same as the suite log though could differ if things go badly wrong.

Why do you want to put this into a temp dir?

MetRonnie: Because that's where all the other `run_ok`, `cmp_ok` etc outputs go
```
stdout and stderr stored in: /var/tmp/rdutta/cylctb-20210209T151507Z/functional/cylc-get-config/00-simple
```

oliver-sanders: Eh, not sure how helpful this is but ok.

I think I'm going to have a crack at moving this dir into the `cylc-run` hierarchy as this tempdir is not included in the "artefact" in the actions and can be hard to access when running the test battery on remote hosts (if the temp file system is not shared) or under a job runner (as the temp files may be removed after the job finishes).

oliver-sanders (approved):

PR #4066, opened 2021-02-08
cfg: document recurrence graph addition in config

oliver-sanders:
Small change, realised that there is nothing in the config docs to explain the irregular behaviour of recurrence settings (now that they are settings not sections) so added a quick note.

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/cfgspec/suite.py
+++ b/cylc/flow/cfgspec/suite.py
@@ -534,12 +534,43 @@ with Conf(
                 Example Recurrences:
 
                 date-time cycling:
-                   ``T00,T06,T12,T18`` or ``PT6H`` - *every six hours*
+                   * ``R1`` - once at the intial cycle point
+                   * ``T00,T06,T12,T18`` - daily at 00:00, 06:00, 12:00
+                     & 18:00
+                   * ``PT6H`` - every six hours starting at the initial
+                     cycle point
                 integer cycling:
-                   ``P2`` - *every other cycle*
+                   * ``R1`` - once at the intial cycle point
+                   * ``P2`` - every other cycle
+                   * ``P3,P5`` - every third or fifth cycle
 
-                See :ref:`GraphTypes` for more on recurrence expressions, and
-                how multiple graphs combine.
+                .. note::
+
+                   Unlike other Cylc configurations duplicate recurrences
+                   are additive and do not override.
+
+                   For example this:
+
+                   .. code-block:: cylc
+
+                      [scheduling]
+                          [[graph]]
+                              R1 = a => b
+                              R1 = c => d
+
+                   Is equivalent to this:
+
+                   .. code-block:: cylc
+
+                      [scheduling]
+                          [[graph]]
+                              R1 = """
+                                  a => b
+                                  c => d
+                              """
+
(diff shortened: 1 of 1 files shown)


cylc/flow/cfgspec/suite.py
```diff
@@ -534,12 +534,43 @@
...
+                   * ``R1`` - once at the intial cycle point
+                   * ``T00,T06,T12,T18`` - daily at 00:00, 06:00, 12:00
+                     & 18:00
+                   * ``PT6H`` - every six hours starting at the initial
+                     cycle point
                 integer cycling:
-                   ``P2`` - *every other cycle*
+                   * ``R1`` - once at the intial cycle point
+                   * ``P2`` - every other cycle
+                   * ``P3,P5`` - every third or fifth cycle
```
wxtim, line 545: If you build this as is the formatting of `integer cycling` and `date-time cycling` is the same as that for config item.
```suggestion
                date-time cycling:

                   * ``R1`` - once at the intial cycle point
                   * ``T00,T06,T12,T18`` - daily at 00:00, 06:00, 12:00
                     & 18:00
                   * ``PT6H`` - every six hours starting at the initial
                     cycle point

                integer cycling:

                   * ``R1`` - once at the intial cycle point
                   * ``P2`` - every other cycle
                   * ``P3,P5`` - every third or fifth cycle
```

wxtim (approved):
The suggestion is tangential to the PR.

cylc/flow/cfgspec/suite.py
oliver-sanders: Yeah, thats a regression bug in the rtd theme I need to poke them about at some point (it effects other things, mainly the API docs).

Your diff above converts the bullet point lists into block quotes. Would rather keep the RST correct and fix the particular theme with the issue:

```rst
definition list item:
    definition list description

plain text:

   blockquote
```

oliver-sanders: Here's the bug report - https://github.com/readthedocs/sphinx_rtd_theme/issues/800, https://github.com/readthedocs/sphinx_rtd_theme/issues/979

It was fixed a while back but seems to be back.

oliver-sanders: Digging a little deeper I think it is because the definition list is being documented within a domain.

Take this example:

```rst
Foo
   foo
Bar
   bar
Baz
   baz


.. py:class:: pub

   Arguments:
      foo: abc
      bar: bcd
      baz: def

   Foo
      foo
   Bar
      bar
   Baz
      baz
```

Which renders like this:

Screenshot 2021-02-12 at 12 37 55

oliver-sanders: Yep that's it - https://sphinx-rtd-theme--981.org.readthedocs.build/en/981/demo/api.html#test_py_module.test.Foo

oliver-sanders: Raised an issue - https://github.com/readthedocs/sphinx_rtd_theme/issues/1052

wxtim:
Having read your comments I'd be happy to merge. But if you'd rather I wait...

Issue #3046, opened 2019-03-27
Improve the speed of our functional tests


kinow:
Besides the issue with flaky tests #2894 , our `test-battery` is quite slow. I just had a look at one of the builds in Travis-CI, it took 32 minutes and 8 seconds. Where 29 minutes and 51 seconds were running the functional tests.

Tasks like git clone, virtualenv creation, installing dependencies, running coverage, do not add up to 2 minutes.

A few possible alternatives:

- Run more tests in parallel
- Move some of the timer based tests (e.g. `sleep` in a loop searching some output) to a more reactive approach when possible
- Fix only the slowest tests
- Pythonize some of the tests
- Remove tests that are covered by unit tests and are not significant enough to be a functional test, or maybe demote/promote them to unit tests

Having a test suite with 4 builds running in parallel, taking up to 30 minutes, and most times needing a kick for the flaky tests is quite infuriating.

But before doing any of the above, I think we should measure. We can already measure the coverage of the functional tests. I think now we could measure how long each test is taking.

This can be easily done with `prove --timer`, which adds a prefix like `[14:59:56] /tmp/tests/cylc-submit/02-remote-with-shared-fs-bg ................. ok`. Doing that we should be able to parse the output and identify what tests are taking the most time, and check whether they can be the first ones to be improved.

For me, under 5 minutes is [perfect](https://en.wikipedia.org/wiki/Continuous_integration#Keep_the_build_fast). Under 10 minutes is OK. Under 15 minutes is still OK if the tests are stable.

hjoliver:
I'm sure we all agree with you @kinow. 

I'm slightly concerned that is is just difficult to *quickly* test a system that manages a bunch of other jobs in a workflow (i.e. they jobs can't all run at once).  However, we must be able to do better than this.

Good idea on recording individual test timings.

hjoliver:
Didn't we get down to about 10 mins when we first started using chunks on Travis CI?

kinow:
>Didn't we get down to about 10 mins when we first started using chunks on Travis CI?

Coverage could have increased it. I will add an option to run coverage only against pull requests and nightly on master too.

kinow:
Side note: GitHub is offering a beta service similar to GitLab and Travis-CI builds: [GitHub Actions](https://github.com/features/actions). Based on containers. Worth if offers the same functionality as Travis, plus performance.

[...]

PR #4077, opened 2021-02-15
test fixes

oliver-sanders:
Miscellaneous test fixes and stability improvements to fix issues encountered when running the fully expanded test matrix against non-containerised platforms.

Test changes:

* Fix test platform configuration errors.
* Fix tests overly reliant on precise timings or extend timeouts where required (e.g. at submission should be expected to take longer than background).
* Fix lsof filesystem polling bug in the `test_header` (mb).
* Fix fork-bomb test - tests/flakyfunctional/job-submission/18-check-chunking.t
* Replace polling with event-driven stuff where possible.

Non-test changes:

* Re-implement the bash `cylc graph-diff` command in Python as `cylc graph --diff`.
  * The way the workflow argument is handled by the old script was different to the Python scripts.
  * The use of `cylc scan` made it fragile on test platforms.
  * This was the last bash command which allows us to remove the back-compat `os.exec` logic.
  * Change `cylc graph` to use the standard workflow argument parsing function.
  * Note: this drops support for graphical diff tools :( can re-implement later if desired.
* Change the test battery base registration.
  * Ensure that each registration is unique so that test runners on different platforms don't fight over the same files.
  * Make the base reg configurable to make automated testing easier.
* Add a new platform configuration for the maximum number of jobs permitted in a single batch submission.
  * Not expecting this configuration to be used, it just replaces a hard-coded default making testing easier.

Some tests now run faster, cc https://github.com/cylc/cylc-flow/issues/3046

**Requirements check-list**
- [ ] Does not contain off-topic changes (off-topic changes listed above)
- [x] No dependency changes.


cylc/flow/task_remote_mgr.py
```diff
@@ -259,7 +259,7 @@ def remote_tidy(self):
                 proc.terminate()
             except OSError:
                 pass
-            out, err = proc.communicate()
+            out, err = (f.decode() for f in proc.communicate())
```
oliver-sanders, line 262: Minor bug that made a mess of reported error but didn't cause traceback in the Scheduler.

etc/bin/run-functional-tests
```diff
@@ -183,6 +183,9 @@ export TMPDIR
 # test environment
 export CYLC_TEST_SKIP=${CYLC_TEST_SKIP:-}
 CYLC_TEST_TIME_INIT="$(date -u +'%Y%m%dT%H%M%SZ')"
+CYLC_TEST_UID="$(< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c4)"
```
oliver-sanders, line 186: Give each test battery run a four character identifier to distinguish it from other runs initiated at the same time.

This is important if running the same tests on multiple platforms simultaneously.

tests/functional/cylc-clean/00-basic.t
```diff
@@ -44,76 +40,60 @@ __FLOW__
...
-                    \`-- share
-                        \`-- cycle -> ${TEST_DIR}/${SYM_NAME}-cycle/cylc-run/${SUITE_NAME}/share/cycle
-${TEST_DIR}/${SYM_NAME}-work
-\`-- cylc-run
-    \`-- cylctb-${CYLC_TEST_TIME_INIT}
-        \`-- ${FUNCTIONAL_DIR}
-            \`-- cylc-clean
-                \`-- ${TEST_NAME_BASE}
-                    \`-- work
+${TEST_DIR}/sym/cycle/cylc-run/${CYLC_TEST_REG_BASE}
```
oliver-sanders, line 59: Move the tree output down a couple of levels to avoid making the tests dependent on the registration base.

tests/functional/cylc-kill/00-multi-hosts-compat/flow.cylc
```diff
@@ -1,23 +1,29 @@
...
+
 [scheduling]
     [[graph]]
-        R1="""
-KILLABLE:start-all => killer
-"""
+        R1 = """
+            # wait for the remote tasks to start before triggering the
+            # local ones in order to factor out remote-init time
+            remote-1:start & remote-2:start => local-1 & local-2 & local-3
```
oliver-sanders, line 13: The preparing stage i.e:

* job file writing
* remote-init
* remote file installation

Is taking >60s on my test platform which results on the local jobs finishing before the remote ones have even started.

tests/functional/cylc-ping/04-check-keys-remote.t
```diff
@@ -16,48 +16,69 @@
...
 client_${CYLC_TEST_INSTALL_TARGET}.key
+server.key
+server.key_secret
+__OUT__
+else
+    cmp_ok "$KEYS_FILE" <<__OUT__
 client.key_secret
+client_${CYLC_TEST_INSTALL_TARGET}.key
 server.key
 __OUT__
```
oliver-sanders, line 68: Different key files are present if the filesystem is shared, it's quite good to have a few dual-modal tests like this to ensure that the shared filesystem usage is working as intended.

tests/functional/job-submission/20-check-chunking.t
```diff
@@ -20,38 +20,29 @@
...
+"
 
-init_suite "${TEST_NAME_BASE}" <<'__FLOW_CONFIG__'
-[scheduler]
-    [[events]]
-        abort on inactivity = True
-        abort on stalled = True
-        inactivity = PT10M
-[task parameters]
-    p = 1..202
```
oliver-sanders, line 35: This test used to submit 200 odd jobs to test that the chunking mechanism for job submission was working correctly. This has a significant impact on the test system which was causing some of the other flaky tests to fail in very strange ways (took ages to realise this was causing it).

Made `max batch submit size` configurable to make this test nicer. Also makes this batching a little more transparent and removes an arbitrary hard-coded default so that's nice too.

tests/functional/job-submission/20-check-chunking.t
```diff
@@ -20,38 +20,29 @@
...
-[scheduling]
-    [[graph]]
-        R1 = t1<p>
-[runtime]
-    [[t1<p>]]
-        # Reduce the load on many jobs sending the "started" message
```
oliver-sanders, line 41: 👿

tests/functional/lib/bash/test_header
```diff
@@ -728,6 +728,15 @@ create_test_global_config() {
...
     echo "$PRE" >'etc/global.cylc'
+    cat >>'etc/global.cylc' <<__HERE__
+    # set a default timeout for all flow runs to avoid hanging tests
+    [scheduler]
+        [[events]]
+            inactivity = PT5M
+            timeout = PT5M
+            abort on inactivity = true
+            abort on timeout = true
+__HERE__
```
oliver-sanders, line 739: Sets a default 5 minute timeout to catch hanging test workflows.

tests/functional/lib/bash/test_header
```diff
@@ -741,7 +750,7 @@ _parse_platform_spec () {
     local requirement="$1"
 
     local locality='local'
-    local job_runner='background'
+    local job_runner='*'
```
oliver-sanders, line 753: This change allows more tests to run under other job submission systems (which was the original intention).

* If a test does *not* specify `REQUIRE_PLATFORM` then it will *only* run on `_local_background`.
* Otherwise it will run on *all* job runners unless `runner:<runner>` is set.

tests/functional/lib/bash/test_header
```diff
@@ -949,5 +960,11 @@ _get_test_host
 # provides: CYLC_TEST_INSTALL_TARGET
 _get_test_install_target
 
+if ${CYLC_SHOW_TEST_EXECUTION:-false}; then
+    # write the name of tests as they are run
+    # (helpful for debugging hanging tests)
+    echo "# $TEST_NAME_BASE" >&2
+fi
```
oliver-sanders, line 967: Minor feature to help with diagnosing hanging tests, this writes the name of each test just before it starts running.

tests/functional/remote/01-file-install.t
```diff
@@ -15,106 +15,137 @@
...
+    for DIR in "bin" "app" "etc" "lib" "dir1" "dir2"
+    do
+        mkdir -p "${SUITE_RUN_DIR}/${DIR}"
+        touch "${SUITE_RUN_DIR}/${DIR}/moo"
+    done
+    for FILE in "file1" "file2"
+    do
+        touch "${SUITE_RUN_DIR}/${FILE}"
+    done
+}
```
oliver-sanders, line 35: Simplified this test to remove some of the moving parts.

tests/functional/remote/01-file-install.t
```diff
@@ -15,106 +15,137 @@
...
+    skip 2
+    exit
+fi
+
 # Test file install completes before dependent tasks are executed
-create_test_global_config "" "
-[platforms]
-    [[cinderella]]
-        hosts = ${CYLC_TEST_HOST}
-        install target = ${CYLC_TEST_INSTALL_TARGET}
```
oliver-sanders, line 78: Need to configure platforms using the `TEST_PLATFORM_NAME` in order to inherit any test configuration (e.g. comms mode) as we don't currently have an `inherit` configuration for platforms.

oliver-sanders:
@hjoliver you might want to take a look into the `cylc graph-diff` -> `cylc graph --diff` changes and the new `max batch submit size` configuration - past that it's just test wrangling.

cylc/flow/task_remote_mgr.py
hjoliver: Huh, why do we need string `decode()` here?

hjoliver: To answer myself:
> `communicate()` returns a tuple (stdout_data, stderr_data). The data will be strings if streams were opened in text mode; otherwise, bytes.

Should we just be specifying text mode (or explicit encoding) instead, for clarity?

tests/functional/cylc-kill/00-multi-hosts-compat/flow.cylc
hjoliver: Ouch

hjoliver (approved):
All looks good to me, very nice tidy up. One review will do as it's just tests (and they're all passing). I'm looking forward to the flagging of hung tests.

hjoliver:
> One review will do as it's just tests (and they're all passing)

Damn it, conflicted now.

cylc/flow/task_remote_mgr.py
oliver-sanders: I've only just learned about "text" mode myself recently, looks good, presumably it just does the decode for you using UTF-8.

tests/functional/remote/01-file-install.t
```diff
@@ -15,108 +15,140 @@
...
+${RUN_DIR_REL}/file2
+${RUN_DIR_REL}/lib/moo
 __OUT__
 
-cylc stop --max-polls=60 --interval=1 "${SUITE_NAME}"
 purge
 # -----------------------------------------------------------------------------
 
+if ! command -v xfs_mkfile; then
+    skip 2
```
MetRonnie, line 112:
```suggestion
    skip 2 "xfs_mkfile not installed"
```

MetRonnie (approved):
Nice, just one small optional suggestion

hjoliver:
I think I rebooted the CI on this today, in which case the functional test fail seems to be repeatable.

hjoliver:
(Pushed a small fix to this branch)

hjoliver:
Tests pass, merging :tada:

PR #4083, opened 2021-02-17
Fix time zone in unit tests.

hjoliver:
Avoid 11 errors like this, caused by rampant timezone bias in the UK team. Go the +13 timezone! :rofl: 
```python
>           assert mocked_config.cfg[
                'scheduling']['final cycle point'] == expected_fcp
E           AssertionError: assert '20210101T0000+13' == '20210101T0000Z'
E             - 20210101T0000Z
E             ?              ^
E             + 20210101T0000+13
E             ?              ^^^

tests/u/test_config.py:635: AssertionError
```

One review will do.

**Requirements check-list**
- [x] Does not need tests [is tests!]
- [ ]  (dependency change: pytest-env addeded, but I don't think we're installing test dependencies via conda).


MetRonnie (changes requested):
We should run the tests in a non-UTC timezone (we already do for the functional tests, must have missed it for the fast tests)
```diff
diff --git a/.github/workflows/test_fast.yml b/.github/workflows/test_fast.yml
index 7fb4699cf..4cf73b01b 100644
--- a/.github/workflows/test_fast.yml
+++ b/.github/workflows/test_fast.yml
@@ -17,6 +17,9 @@ jobs:
         include:
           - os: 'macos-latest'
             python-version: '3.7'
+    env:
+      # Use non-UTC time zone
+      TZ: XXX-05:30
 
     steps:
       - name: Checkout
```

hjoliver:
Timezone Z is as good as any other so long as it's explicitly specified rather than assumed, isn't it?

MetRonnie:
> Timezone Z is as good as any other so long as it's explicitly specified rather than assumed, isn't it?
>
> Or, maybe we should run tests in the local timezone, and change test assertions to expect that. Then our respective local runs will cover all bases.

We can do either of those, at the same time as running tests in an arbitary non-UTC time zone on GH Actions.

> The trouble with setting TZ in the CI environment is local runs will fail

I'm not sure what you mean?

The idea is to catch + fix any cases where tests pass in UTC but fail in other timezones, which we are always likely to forget about here in the UK in winter 😛  - #3794

hjoliver:
>  > The trouble with setting TZ in the CI environment is local runs will fail
> I'm not sure what you mean?

Sorry, I meant if the tests expect TZ but get the setting from the environment, then local test runs will fail unless I also remember to set TZ in my own environment.  So, better to have the timezone, whichever it is, set explicitly in the tests. Or use local time, but have the tests dynamically determine the expected results, given that.

MetRonnie:
> So, better to have the timezone, whichever it is, set explicitly in the tests. Or use local time, but have the tests dynamically determine the expected results, given that.

I agree that these are needed. The point of using an arbitrary TZ in GH Actions is to catch the occasions where we forget to do either of those, because running the tests locally on GMT won't catch them

oliver-sanders:
BTW there is a pytest plugin for setting environment variables - https://github.com/MobileDynasty/pytest-env

hjoliver:
@MetRonnie - simply setting TZ in `pytest.ini` does the trick.

(Looking at the tests, I must say I don't quite understand why it works; the tests mock "now" as a UTC time, regardless of local timezone, which isn't going to happen outside of the tests, and the expected points are all UTC ... but we have higher priorities at the moment)

hjoliver:
(Note there is now a dependency change: `pytest-env`)

MetRonnie:
I don't think the syntax is correct. If I try this
```diff
diff --git a/pytest.ini b/pytest.ini
index 5d321b0b1..fd14a67c8 100644
--- a/pytest.ini
+++ b/pytest.ini
@@ -30,4 +30,4 @@ testpaths =
     tests/unit/
     tests/integration/
 env =
-    TZ=+0530
+    TZ=XXX-05:30
```
then the tests fail again. So I think you still need the original fixes for the tests.

hjoliver:
> I don't think the `$TZ` syntax is correct. 

Oops you're right, I tried the correct syntax first and the tests failed, then I thought maybe isodatetime was actually reading `$TZ` but expecting it our ISO `cycle point timezone` format e.g. `+0530` ... but it turns out a bad `TZ` value defaults to a UTC clock (which is why the tests then worked!).

hjoliver:
Good now @MetRonnie (I hope!)

MetRonnie (approved):

PR #4087, opened 2021-02-18
rsync format

oliver-sanders:
rsync: change file installation rsync options
    
* more friendly output format
* unify local and remote fileinstall options
* use checksums not timestamps for file filtering
* log rsync command with debug level

### DIFF
--- a/cylc/flow/remote.py
+++ b/cylc/flow/remote.py
@@ -152,6 +152,14 @@ def get_includes_to_rsync(rsync_includes=None):
     return configured_includes
 
 
+DEFAULT_RSYNC_OPTS = [
+    '-a',
+    '--checksum',
+    '--out-format=%o %n%L',
+    '--no-t'
+]
+
+
 def construct_rsync_over_ssh_cmd(
         src_path, dst_path, platform, rsync_includes=None):
     """Constructs the rsync command used for remote file installation.
@@ -167,20 +175,14 @@ def construct_rsync_over_ssh_cmd(
 
     """
     dst_host = get_host_from_platform(platform)
-    rsync_cmd = ["rsync"]
     ssh_cmd = platform['ssh command']
-    rsync_options = [
-        "-v",
-        "--perms",
-        "--recursive",
-        "--links",
-        "--checksum",
+    rsync_cmd = [
+        "rsync",
         "--delete",
         "--rsh=" + ssh_cmd,
         "--include=/.service/",
         "--include=/.service/server.key"
-    ]
-    rsync_cmd.extend(rsync_options)
+    ] + DEFAULT_RSYNC_OPTS
     # Note to future devs - be wary of changing the order of the following
     # rsync options, rsync is very particular about order of in/ex-cludes.
 
--- a/cylc/flow/task_remote_mgr.py
+++ b/cylc/flow/task_remote_mgr.py
@@ -372,7 +372,7 @@ class TaskRemoteMgr:
         if ctx.out:
             RSYNC_LOG.info(
                 'File installation information for '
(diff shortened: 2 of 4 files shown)


cylc/flow/rsync.py
```diff
@@ -0,0 +1,94 @@
...
+        [('+', 'added file'), ('-', 'removed file'), ...]
+
+    """
+    ret = []
+    for line in stdout.splitlines():
+        action, path = line.split(' ', 1)
+        if path in ('.', './'):
+            continue
+        ret.append((RSYNC_SYM.get(action, '?'), path))
+    return ret
```
oliver-sanders, line 94: Also omits the src directory itself from the results.

cylc/flow/rsync.py
```diff
@@ -0,0 +1,94 @@
...
+        dry-run (bool):
+            indicate dry-run, rsync will not take place but report output if a
+            real run were to be executed
+
+    Return:
+        list: command to use for rsync.
+
+    """
+
+    rsync_cmd = ["rsync", "-a", "--out-format=%o %n"]
```
oliver-sanders, line 43: `%n` gives us the path relative to the src/run dir.

cylc/flow/scheduler.py
```diff
@@ -276,7 +277,7 @@ async def install(self):
 
         """
         # Install
-        source = suite_files.get_suite_source_dir()
+        source = suite_files.get_workflow_source_dir(Path.cwd())
```
datamel, line 280: `get_workflow_source_dir()` now also returns the source path. This was to enable more informative error messages.

datamel:
Changes look good, I will review properly when converted from draft. Looks like this has been started from a previous version of cylc-reinstall so test failures should be fixed by rebase.

datamel:
Also, would it be worth putting the remote file install rsync through this formatting?

oliver-sanders:
Ping @dpmatthews

dpmatthews:
I think we've agreed the rose files should be excluded so that they don't show up in the rsync output.
Whilst I agree that "+" and "-" are nicer than the "send" and "del." which are returned by rsync, is that a good enough reason to scrape the output (normally best avoided)?

cylc/flow/rsync.py
dpmatthews: I'd suggest "--out-format=%o %n%L"
```
%L the string " -> SYMLINK", " => HARDLINK", or "" (where SYMLINK or HARDLINK is a filename)
```
Should usually be  blank in our case but useful to know when it isn't.

cylc/flow/rsync.py
```diff
@@ -0,0 +1,94 @@
...
+    if dry_run:
+        rsync_cmd.append("--dry-run")
+    if reinstall:
+        rsync_cmd.append('--delete')
```
dpmatthews, line 47: Consider also using "--no-t --checksum"?
This means that files only get reinstalled if their contents have changed.
Otherwise, they get re-installed if their modification time changes.

oliver-sanders: I don't think the `-no-t` effects the `--checksum` functionality does it?

```
t, --times                 preserve modification times
```

oliver-sanders:
Rebased.

datamel (approved):
I am happy with the change. I have run workflows and checked logs after the change. Thanks for the tidy @oliver-sanders.

oliver-sanders:
Applied code suggestions from Dave, @wxtim could you give this a quick pass.

oliver-sanders:
Rebased & fixed indentation of first line of rsync output.

wxtim (approved):
Had a look. Happy.
Played with code. Happy.

PR #4089, opened 2021-02-19
Tidy tests

MetRonnie:

Follow up to #4077, removed `05-check-keys-sharedfs` as it's identical to `04-check-keys-remote`, if you run it with `etc/bin/run-functional-tests -p '*shared*'`. (However, if you use `-p '*'` it will instead choose whichever is the first matching platform in global.cylc)

Also move the new functional tests UID to after the datetime in the base dir name, so that they show up in order in the file explorer etc.

**Requirements check-list**
- [x] Does not need tests (is tests)
- [x] No dependency changes.


tests/functional/cylc-ping/04-check-keys-remote.t
```diff
@@ -53,26 +53,13 @@ suite_run_ok "${TEST_NAME_BASE}-run" cylc play \
...
-server.key
-__OUT__
-fi
 
-if [[ "$CYLC_TEST_PLATFORM" == *shared* ]]; then
-    skip 1
-else
-    grep_ok "Removing authentication keys and contact file from remote: \"${CYLC_TEST_INSTALL_TARGET}\"" "${SUITE_RUN_DIR}/log/suite/log"
-fi
+grep_ok "Removing authentication keys and contact file from remote: \"${CYLC_TEST_INSTALL_TARGET}\"" "${SUITE_RUN_DIR}/log/suite/log"
```
oliver-sanders, line 62: Why has the shared filesystem test been removed?

MetRonnie: Because there's already a test for it, see the other file https://github.com/cylc/cylc-flow/pull/4089/files#diff-18ed6f53d34c84b28d36af7117a10837b94185f6007c54a973675affd1a44052

oliver-sanders: Ok, can we delete

datamel (approved):
Nice to get rid of a surplus test.

oliver-sanders (approved):

PR #4092, opened 2021-02-22
cylc-rose: functional test for re-installation.

wxtim:
Companion of [Cylc-Rose#35](https://github.com/cylc/cylc-rose/pull/35)

Allow cylc-rose to be able to have an awareness of previous installs by preventing `opt/rose-suite-cylc-install.conf` from  being rsynced by cylc reinstall:

**Requirements check-list**

- [x] Does not need tests: Covered by tests in cylc-rose

- [x] No change log entry required. Part of pre-first release work.

### DIFF
--- a/cylc/flow/suite_files.py
+++ b/cylc/flow/suite_files.py
@@ -992,6 +992,7 @@ def get_rsync_rund_cmd(src, dst, reinstall=False, dry_run=False):
         '.svn',
         '.cylcignore',
         'log',
+        'opt/rose-suite-cylc-install.conf',
         SuiteFiles.Install.DIRNAME,
         SuiteFiles.Service.DIRNAME]
     for exclude in ignore_dirs:


cylc/flow/suite_files.py
```diff
@@ -992,6 +992,7 @@ def get_rsync_rund_cmd(src, dst, reinstall=False, dry_run=False):
         '.svn',
         '.cylcignore',
         'log',
+        'opt/rose-suite-cylc-install.conf',
```
datamel, line 995: Just wondering if the intention is for anything else to be kept in the folder `opt`? As it stands, the empty `opt` folder will still be copied.

datamel:
If we want this preserved for future devs, could (artificially) add a folder `opt` here with a fake `rose-suite-cylc-install.conf` file. 

https://github.com/cylc/cylc-flow/blob/078f45c26bb52ef62218488a3297520d075c8af0/tests/functional/cylc-install/03-file-transfer.t#L30

cylc/flow/suite_files.py
wxtim: Yes. Opt can contain other optional configs which we probably want to install, unless we choose to make cylc-rose do it, just as https://github.com/cylc/cylc-rose/issues/36 suggests installing `rose-suite.conf`.
IMO this is reasonable, since it is more consistent.

wxtim: On Element @oliver-sanders  said no:
> I think cylc install should exclude (via the default rsync exclude options):
>    ``rose-suite.conf``
>    ``opts/rose-suite-cylc-install.conf``
> The first file cylc-rose will need to copy into place.
> The second file cylc-rose creates/modifies as part of the post_install hook.

datamel (approved):

cylc/flow/suite_files.py
oliver-sanders: Still need to exclude the `rose-suite.conf`?

oliver-sanders: Ah wait, there's another parallel PR which adds the `rose-suite.conf` that will conflict with this one?

wxtim:
Replaced by #4094

Issue #4060, opened 2021-02-04
install target = localhost


oliver-sanders:
At present we determine if a platform is a "remote" platform based on whether its install target is `localhost`.

This is the default, however, you can mess with it.

```ini
# global.cylc
[platforms]
    [[localhost, linux\d\d\d]]
        install target = filesystem1
    [[[cluster]]]
        install target = filesystem2
```

I think we need the following logic change:

```diff
-if platform['install target'] == 'localhost':
+if platform['install target'] == get_platform('localhost')['install target']:
```

Can tidy this away as a method or whatnot.

MetRonnie:
Would it be a good idea to use a data class to represent a Platform rather than using dicts?

oliver-sanders:
I don't think a dataclass would help, the result would still be Python objets which would still require deepcopy?

MetRonnie:
But we could avoid having to set items on the fly, e.g. instead of https://github.com/cylc/cylc-flow/blob/0a2170a84665849a607e7d6494fd8accde182c89/cylc/flow/platforms.py#L420-L432

we could have
```python
class Platform:
    ...
    @property
    def install_target(self):
        if self._install_target is None:
            return self.name
        return self._install_target
```

oliver-sanders:
That wont help with the platform name which can be a regex value e.g. `linux\d\d\d` which we have to resolve e.g. to `linux012`.

PR #4095, opened 2021-02-24
Replace string literals of localhost with function returning config

datamel:
These changes close https://github.com/cylc/cylc-flow/issues/4060

**Requirements check-list**

- [x] No dependency changes.

### DIFF
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -481,3 +481,9 @@ def get_random_platform_for_install_target(install_target):
     """Return a randomly selected platform (dict) for given install target."""
     platforms = get_all_platforms_for_install_target(install_target)
     return random.choice(platforms)
+
+
+def get_localhost_install_target():
+    """Returns the install target of localhost platform"""
+    localhost = get_platform('localhost')
+    return get_install_target_from_platform(localhost)
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -485,5 +485,5 @@ def get_random_platform_for_install_target(install_target):
 
 def get_localhost_install_target():
     """Returns the install target of localhost platform"""
-    localhost = get_platform('localhost')
+    localhost = get_platform()
     return get_install_target_from_platform(localhost)
--- a/tests/unit/test_platforms_get_platform.py
+++ b/tests/unit/test_platforms_get_platform.py
@@ -18,7 +18,10 @@
 
 import pytest
 import random
-from cylc.flow.platforms import get_platform
+from cylc.flow.platforms import (
+    get_localhost_install_target,
+    get_platform
+)
 from cylc.flow.exceptions import PlatformLookupError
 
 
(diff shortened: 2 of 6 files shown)


cylc/flow/platforms.py
```diff
@@ -481,3 +481,9 @@ def get_random_platform_for_install_target(install_target):
     """Return a randomly selected platform (dict) for given install target."""
     platforms = get_all_platforms_for_install_target(install_target)
     return random.choice(platforms)
+
+
+def get_localhost_install_target():
+    """Returns the install target of localhost platform"""
+    localhost = get_platform('localhost')
```
wxtim, line 488: Is identical to:
```suggestion
    localhost = get_platform()
```

cylc/flow/platforms.py
```diff
@@ -481,3 +481,9 @@ def get_random_platform_for_install_target(install_target):
...
+    return get_install_target_from_platform(localhost)
```
wxtim, line 489: I've tried to kill VSCode's line stripping behavior, but it seems to have become more aggressive. Or perhaps I'm creating more new files....
```suggestion
    return get_install_target_from_platform(localhost)

```

tests/unit/test_platforms_get_platform.py
```diff
@@ -203,3 +206,7 @@ def test_get_platform_warn_mode_fail_if_backticks():
         r'platform = `echo \$\{chamber\}`: '
         r'backticks are not supported; please use \$\(\)'
     )
+
+
+def test_get_localhost_install_target():
+    assert get_localhost_install_target() == 'localhost'
```
wxtim, line 212:
```suggestion
    assert get_localhost_install_target() == 'localhost'

```

tests/unit/test_platforms_get_platform.py
```diff
@@ -203,3 +206,7 @@ def test_get_platform_warn_mode_fail_if_backticks():
         r'platform = `echo \$\{chamber\}`: '
         r'backticks are not supported; please use \$\(\)'
     )
+
+
+def test_get_localhost_install_target():
```
wxtim, line 211: Might you be able to model the case where
```
[platforms]
  [[localhost]]
    install target = raspi123
```

wxtim (changes requested):
- [x] Read code, which seems to make sense. I have jotted some thoughts about the code, but there's nothing critical in the comments, except possibly `get_platform('localhost') which might just possibly later cause future inconsistencies.
- [x] Read Tests - Suggested 1 additional test scenario.
- [x] Downloaded and tested locally - will do.

tests/unit/test_platforms_get_platform.py
datamel: Ah, I think this is to do with the way the diff is displayed on github. Blank line 213 was there, just not displayed as it was no change to the original?

cylc/flow/platforms.py
datamel: I changed it, making the default of  `get_install_target_from_platform()` return localhost but then I changed it back as I think  `if platform['install target'] == get_localhost_install_target()` immediately makes more sense for future devs than `if platform['install target'] == get_install_target_from_platfrom() `, what do you think?

wxtim: It's fine. My suggestion only makes sense in context of platforms. Yours just makes sense.

wxtim (approved):

oliver-sanders (approved):

PR #4099, opened 2021-02-25
Unify get-config

### LINKED ISSUE #4009: get-config: unify commands
Unify the `get-suite-config` and `get-site-config` commands as `config`.

```
cylc config [OPTIONS] [REG]

Examples:
  # print global configuration
  $ cylc config

  # print workflow configuration
  $ cylc config myflow

  # print specific configuration from the global config
  $ cylc config -i '[platforms][myplatform]hosts'

  # print workflow config, first setting the initial cycle point
  $ cylc config --initial-cycle-point=now myflow
```

This handles the renaming of site (global) and suite (flow).

#### Propose removing the following options:

* `--print-run-dir` - not sure what the future for this option is, possibly obsolete
* `--python` - useless (Python can't parse natively), can implement `--format json` if required.
* `--mark-up` - what format was that implemented for?
* `--tasks` - deprecated
* `--run-mode` - obsolete
* `--suite-owner` - obsolete
* `--all-tasks` - untested and appears broken (prints entire sparse config for each task)
* <del>`--user`, `--host` - remove command reinvokation, just use SSH.</del> (now already gone)

#### Proposed change:

Replace `--print-site-dir` with `--print-filenames` which should print the tree of config files that `cylc config` would have read (if they were present).

```console
$ cylc config --print-filenames myflow
/opt/cylc/global.cylc
/opt/cylc/8/global.cylc
/opt/cylc/8.0a3/global.cylc
/opt/cylc/8.0a3.dev/global.cylc
~/.cylc/global.cylc
~/.cylc/8/global.cylc
~/.cylc/8.0a3/global.cylc
~/.cylc/8.0a3.dev/global.cylc
~/cylc-run/myflow/flow.cylc
```

This helps verify `$CYLC_CONF_PATH` has been set correctly and may be helpful for testing.

Not important, can be added later if we want it.

MetRonnie:
These changes close #4009

- Unify the `get-suite-config` and `get-site-config` commands as `config`
- Remove `--python` option
- Remove `--mark-up` option
- Remove `--run-mode` option
- Remove `--suite-owner` option
- Remove `--tasks` and `--all-tasks` options
- Remove `--print-run-dir` option
- Add `--print-hierarchy` option

**Requirements check-list**

- [x] Probably a good idea to wait for #4076 to be merged first
- [x] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/223.
- [x] No dependency changes.


cylc/flow/parsec/util.py
```diff
@@ -120,11 +120,14 @@ def listjoin(lst, none_str=''):
...
 def printcfg(cfg, level=0, indent=0, prefix='', none_str='',
-             handle=sys.stdout):
+             handle=None):
     """Pretty-print a parsec config item or section (nested dict).
 
     As returned by parse.config.get().
     """
+    # Note: default handle=sys.stdout, but explicitly setting this interferes
+    # with pytest capsys:
+    # https://github.com/pytest-dev/pytest/issues/1132#issuecomment-147506107
```
oliver-sanders, line 130: I think it's just the default kwarg value that is messing with pytest, could do this:

```suggestion
    if not handle:
        handle = sys.stdout
```

cylc/flow/scripts/config.py
```diff
@@ -0,0 +1,131 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""cylc config [OPTIONS] ARGS
+
```
oliver-sanders, line 20: Can we have a short description something like:
```suggestion
Parse and print Cylc configuration files.

```

cylc/flow/scripts/config.py
```diff
@@ -0,0 +1,131 @@
...
+
+@cli_function(get_option_parser)
+def main(parser, options, reg=None):
+    if options.print_hierarchy:
+        print("\n".join(get_config_file_hierarchy(reg)))
+        return
+
+    if reg is None:
+        glbl_cfg().idump(options.item, sparse=options.sparse)
```
oliver-sanders, line 114: Should the global config get the oneline and none_str options too?

oliver-sanders (commented):
Great, can't wait to have this on master.

cylc/flow/parsec/util.py
MetRonnie: Unfortunately not, in the issue linked in the comment, it says that redefinition of `sys.stdout` is the problem

MetRonnie:
Note: if this doesn't get merged in time for 8.0b0, will need to edit the changelog

cylc/flow/parsec/util.py
oliver-sanders: This works for me:

```python
import sys


def whatever(handle=None):
    if not handle:
        handle = sys.stdout
    handle.write('hello world')


def test_whatever(capsys):
    whatever()
    assert capsys.readouterr().out == 'hello world'
```

MetRonnie: Oh, the same test fails for a different reason in that case. The capsys replacement for `sys.stdout` is not an instance of `StringIO`, but doesn't have an attribute `.mode`, which causes an exception at this point
https://github.com/cylc/cylc-flow/blob/4b5e8a98b4c9d833ffdec9d67a5c30f5509db988/cylc/flow/parsec/util.py#L142-L143

Could instead do?
```python
if hasattr(handle, 'mode') and 'b' in handle.mode:
    msg = msg.encode()
```

kinow (approved):
Looks good to me :+1:

oliver-sanders (approved):
Great.

Issue #3527, opened 2020-03-19
tui: improve performance


oliver-sanders:
* Use a subscription to GraphQL deltas to update the tree.
* Perform the updates asynchronously so that the app doesn't freeze during the update.

**Pull requests welcome!**

oliver-sanders:
What the change to GraphQL deltas means for communications and how it fits into the [plan](https://github.com/cylc/cylc-admin/blob/master/docs/img/interfaces-and-protocols.png):

![rect4117](https://user-images.githubusercontent.com/16705946/89285829-fd119a00-d648-11ea-83e4-b22187acd3a4.png)

hjoliver:
Blocked by #3731

hjoliver:
Good idea.

PR #4102, opened 2021-02-26
Tui: warn about large flows

oliver-sanders:
Address a minor point on the beta-0 roadmap.

Add a warning about using with large flows until we get it delta-driven.

See https://github.com/cylc/cylc-flow/issues/3527

One off-topic change, got shellchecker to skip the `.git` dir.

**Requirements check-list**
- [x] One off-topic change listed above
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/scripts/tui.py
+++ b/cylc/flow/scripts/tui.py
@@ -21,6 +21,8 @@ View and control running suites in the terminal.
 (Tui = Terminal User Interface)
 
 WARNING: Tui is experimental and may break with large flows.
+An upcoming change to the way Tui receives data from the scheduler will make it
+much more efficient in the future.
 """
 # TODO: remove this warning once Tui is delta-driven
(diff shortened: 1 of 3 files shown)


cylc/flow/scripts/tui.py
```diff
@@ -18,8 +18,13 @@
 
 View and control running suites in the terminal.
 
-(TUI = Terminal User Interface)
+(Tui = Terminal User Interface)
+
+WARNING: Tui is experimental and may break with large flows.
```
hjoliver, line 23: Maybe flag that we expect better performance in the near-ish future? Something like this:
```suggestion
WARNING: Tui is experimental and may break with large flows. An upcoming change to the
way Tui receives data from the scheduler will make it much more efficient in the future.
```

hjoliver (approved):
I think we can forgive the codecov patch fail in this case.

Issue #3874, opened 2020-10-15
Rethink runahead limiting


hjoliver:
*Just getting my thoughts down here after recent discussions; will tidy up on returning from leave next week*

Runahead limiting was originally designed to limit the performance impact of task pool bloat caused by spread over cycles, in the old spawn-on-submit system where the intermediate cycles would fill up with succeeded tasks.

In other words it is really just a crude way of limiting task pool size (plus preferentially running older tasks) - which we couldn't do properly under spawn-on-submit (the flow would stall if the wrong tasks were held back)

But it's a bad way of limiting task pool size because,
- it's no help within a cycle point (many of our worst flows have a large number of tasks per cycle)
- any default value is an unhappy compromise because
   - flows with many tasks per cycle need a very low runahead limit (say, 1)
   - flows with few tasks per cycle can have a very high runahead limit (say, 200)

In the spawn-on-demand scheduler:
- spread over cycles has no impact, except for individual tasks with no prerequisites and no other constraints (xtriggers)
- tasks are spawned "on demand" when ready to run according to the graph. This brings runhead limiting into line with xtriggers, task hold, and queues: it's just another way to temporarily hold back tasks that are otherwise ready to run
- so, runahead limiting under SoD is functionally equivalent to queue that releases older cycle points first
- (also, the global "default" queue is equivalent to a task pool size limit, because queues limit the number of active tasks, and in SoD the task pool *is* the active tasks - see (x) below)

The one thing that different from queues etc. is our use of the hidden runahead pool. Instead of a visible waiting task held back by a queue, we have a hidden waiting task held back by the runahead limit. 

The points above suggest that we no longer **need** runahead limiting because we now have the ability to limit pool size properly even within a cycle.  However, we will keep it for backward compatibility (many existing flows set a value, and some may rely on it as a proxy for intercycle dependence).  But we should get rid of the hidden runahead pool and unify runahead limiting with the other limiting mechansims: queues, xtriggers (now unified with clock triggers and retries), task hold. (And later we may re-implement all of these as xtriggers).

*This was motivated by thinking about represent of limited (ready but held back as waiting) tasks in the Cylc 8 UI n-distance window*

[...]

Issue #3743, opened 2020-08-05
Post-SoD task hold semantics


hjoliver:
“Hold” currently means: prevent waiting task proxies from submitting their jobs. Post-SoD we can still hold a suite, but holding individual tasks is considerably less feasible because the only waiting tasks that exist now as task proxies are those with multiple parents and partially unsatisfied prerequisites. Further, even those waiting tasks might be hidden in future: [proposal](https://cylc.github.io/cylc-admin/proposal-spawn-on-d.html#spawn-on-outputs-or-spawn-when-ready) and #3822. And besides, users should not have to understand the difference between abstract (graph) and concrete (task proxy in the pool) waiting tasks. 

Do we really need the concept of individual task hold?

If we do, options include:
- (maybe) holding an active task causes its spawned children to be held?
- (probably) allow users to hold abstract tasks, which will require the scheduler to check a list of tasks-to-hold at spawn time

oliver-sanders:
I think having the ability to hold an individual task makes sense and is useful.

If it's difficult to hold tasks before they exist in the pool we could, potentially, consider an xtrigger implementation similar to the proposed retry approach.

dwsutherland:
Yeah, or a broadcast style implementation?

hjoliver:
> n=1, would make it easier to select the non-existent also.

Selection is easy, since we plan to allow n=1,2,3,... up to "nmax" which is "all tasks in the cycle point".  

The question is just how to record what's been held when the corresponding task proxies don't exist yet in the scheduler.

Note that the hold only matters once a task proxy exists, so we can still "action" the hold in the old way (essentially, set a held flag in the task proxy).

Regarding xtrigger- or broadcast-style implementations - what's wrong with this simple approach?:
- scheduler maintains a list of task IDs that are held
- when a task is spawned, check if is in that list; if it is, hold it (i.e. spawn it, but set its held flag)
- when a held task is released, remove it from that list, and if an actual held task proxy exists, unset its held flag.

oliver-sanders:
> Regarding xtrigger- or broadcast-style implementations - what's wrong with this simple approach?:

Would be fine, though because the value is not stored on a per-task basis it would introduce a lot of lookups.

That list will need to go the to DB and we will need to re-work the `is_held` functionality as you will have to ask the `task_pool` whether a task is held rather than the `task_proxy` itself.

hjoliver:
[...]

PR #4103, opened 2021-03-01
Post-SoD refactor: remove pflag, runahead pool

hjoliver:
This is a ~~small~~ SOD follow-up change (no specific Issue, but has been discussed variously)

E.g., partially addresses #3874:
> tasks are spawned "on demand" when ready to run according to the graph. This brings runhead limiting into line with xtriggers, task hold, and queues: it's just another way to temporarily hold back tasks that are otherwise ready to run.
...  we should get rid of the hidden runahead pool and unify runahead limiting with the other limiting mechanisms:

- Make runahead limiting consistent with other forms of limiting (see above)
   - Remove the "runahead task pool". Runahead tasks are now held in the main pool, with an `is_runahead` task indicator (c.f. `is_held`, `is_queued`) 
   - Runahead-limited tasks can now be exposed in the UI, like queued and held tasks, to show the user "this task is ready to run but is being held back by the runahead limit (done for `cylc tui` on this branch).
 
 - A new "hidden pool" is used to hide partially-satisfied tasks (in order to get spawn-on-demand from spawn-on-outputs) 
   
- Remove an annoying vestige of the old SOS dependency matching process
  -  We used to set `task_events_mgr.pflag` ("processing required flag") whenever a task event occurred that required a new round of dependency matching, then if the flag was set we would iterate through the task pool and do dependency matching and other things.
  - Generally tidy up and make clearer the aforementioned "other things" which are now done directly without relying on the old flag

- [UPDATE] restores correct force-triggering behaviour in the presence of queues (broken since spawn-on-demand) implementation: triggering an unqueued tasks queues it; triggering a queued task causes it to run.

`cylc tui` showing queue- and runahead-limited tasks:
image

**Requirements check-list**
      - (mention visible runahead tasks) 
      - (mention visible runahead tasks, somewhere...) 
- [x] No dependency changes.


hjoliver:
(I've been working on this on another branch, not pushed to the PR yet; soon...)

hjoliver:
(Tests passing, after some rebase nightmares; I'll update the description and un-Draft this tomorrow/Wednesday)

hjoliver:
(Bumped to beta-1 milestone)

hjoliver:
(rebased)

hjoliver:
Temporarily converted to Draft, pending fix of post-rebase test fails.

hjoliver:
(Tests failing after a moderately nasty conflict resolution :angry: )

hjoliver:
Tests passing again :tada:

cylc/flow/broadcast_mgr.py
```diff
@@ -59,12 +59,23 @@ def __init__(self, workflow_db_mgr, data_store_mgr):
...
         while not ext_trigger_queue.empty():
             ext_trigger = ext_trigger_queue.get_nowait()
             self.ext_triggers.setdefault(ext_trigger, 0)
             self.ext_triggers[ext_trigger] += 1
+        return set(
+            [
+                itask for itask in itasks
+                if self._match_ext_trigger(itask)
+            ]
+        )
```
MetRonnie, line 78:
```suggestion
        return set(
            itask for itask in itasks
            if self._match_ext_trigger(itask)
        )
```

cylc/flow/xtrigger_mgr.py
```diff
@@ -357,16 +356,28 @@ def callback(self, ctx: SubFuncContext):
...
-        self.collate(itasks)
-        for itask in itasks:
-            if itask.state.xtriggers:
-                self.satisfy_xtriggers(itask)
+        satisfied = set(
+            [
+                itask for itask in itasks
+                if itask.state.xtriggers_all_satisfied()
+            ]
+        )
```
MetRonnie, line 379:
```suggestion
        satisfied = set(
            itask for itask in itasks
            if itask.state.xtriggers_all_satisfied()
        )
```

MetRonnie (commented):
Not a full review, just had a read through as it will likely affect my work on #3743

cylc/flow/xtrigger_mgr.py
oliver-sanders: Or better still use a native set comprehension `satisfied = { itask for itask ...`

dpmatthews (approved):
(just approving the description - not tested or reviewed!)

cylc/flow/task_pool.py
```diff
@@ -1268,24 +1311,31 @@ def force_trigger_tasks(self, items, reflow=False):
...
+                itask.is_manual_submit = True
+                itask.reset_try_timers()
                 # (If None, spawner reports cycle bounds errors).
-                itask.manual_trigger = True
                 if itask.state.reset(TASK_STATUS_WAITING):
                     self.data_store_mgr.delta_task_state(itask)
-                LOG.critical('setting %s ready to run', itask)
-                itask.state.set_prerequisites_all_satisfied()
-                self.data_store_mgr.delta_task_prerequisite(itask)
-                self.data_store_mgr.delta_task_outputs(itask)
```
dwsutherland, line 1284: I guess the change and outputs/prereqs on force trigger doesn't happen in SOD?

I notice these are still called elsewhere..

dwsutherland (approved):
Looks good 👍 

Will test after the rebase, so will review my approval after a play.

cylc/flow/xtrigger_mgr.py
hjoliver: Doh, welcome to Python 3 :rofl:

cylc/flow/scheduler.py
```diff
@@ -1431,12 +1405,66 @@ async def main_loop(self):
...
+                    self.is_updated = True
+                    self.reset_inactivity_timer()
+
             self.proc_pool.process()
 
-            if self.should_process_tasks():
-                self.process_task_pool()
+            # Get tasks with newly-satisfied external triggers.
+            check_if_ready = self.broadcast_mgr.check_ext_triggers(
+                [x for x in self.pool.get_tasks()
```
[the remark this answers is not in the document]
hjoliver: Yeah, we have discussed reformulating the task pool to a dict (it's mentioned in an issue somewhere) to avoid some iteration.  Beyond the scope of this PR though!

cylc/flow/task_pool.py
hjoliver: Hmmm, forgotten the details, I'd better check this change.

hjoliver:
(feedback addressed, except for one thing to follow up on, and conflicts).

hjoliver:
(There are no conflicts on this branch right now so long as you don't select "Rebase and merge".)

cylc/flow/task_pool.py
hjoliver: (I remember now: force trigger means "run even though prerequisites are not satisfied" - so why pretend they were satisfied? ... hence, there is no data-store update to do here).

cylc/flow/task_pool.py
```diff
@@ -574,116 +610,121 @@ def remove(self, itask, reason=""):
...
 
-    def _queue_tasks(self):
+    def queue_tasks(self, itasks):
         """Queue tasks that are ready to run."""
-        queue_me = []
+        for itask in itasks:
+            itask.state.reset(is_queued=True)
+            # TODO Reset manual trigger flag. One manual trigger queues and
+            # unqueued task, another one triggers a queued task.
+            self.data_store_mgr.delta_task_state(itask)  # TODO needed?
```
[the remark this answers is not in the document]
MetRonnie: This was marked as resolved, but looks like it might not be?

cylc/flow/task_pool.py
```diff
@@ -585,116 +622,121 @@ def remove(self, itask, reason=""):
...
-        if queue_me:
-            LOG.debug(
-                "Queue pushed:\n"
-                + '\n'.join(f"* {t.identity}" for t in queue_me)
-            )
+            if all(ready_check_items) and not itask.state.is_runahead:
+                to_queue.append(itask)
 
-    def _release_tasks(self):
+        self.queue_tasks(to_queue)
```
MetRonnie, line 737: Why does the "private" `self._queue_tasks()` method call the "public" `self.queue_tasks()` and not the other way round?

cylc/flow/task_pool.py
```diff
@@ -1285,24 +1329,30 @@ def force_trigger_tasks(self, items, reflow=False):
...
-            # Already in pool? Keep merge flow labels.
-            itask = self.get_task(name, point, flow_label)
-            if itask is None:
-                # Spawn with new flow label.
-                itask = self.spawn_task(name, point, flow_label, reflow=reflow)
+            # Already in pool? Keep it and merge flow labels.
+            itask = self.get_task_main(name, point, flow_label)
             if itask is not None:
+                # Trigger existing task proxy
+                LOG.critical('setting %s ready to run', itask)
```
MetRonnie, line 1336: Why is this `critical`? Leftover debugging?

cylc/flow/task_pool.py
```diff
@@ -1332,7 +1382,14 @@ def sim_time_check(self, message_queue):
                 sim_task_state_changed = True
         return sim_task_state_changed
 
-    def set_expired_task(self, itask, now):
+    def set_expired_tasks(self):
+        res = False
+        for itask in self.get_tasks():
+            if self._set_expired_task(itask):
+                res = True
+        return res
```
MetRonnie, line 1390:
```suggestion
        for itask in self.get_tasks():
            if self._set_expired_task(itask):
                return True
        return False
```

MetRonnie (commented):
9/29 files viewed. Seems to be a couple of unresolved point from earlier

oliver-sanders (commented):
There are some quick changes that will reduce the impact of task iteration, have dumped them in a PR:

https://github.com/hjoliver/cylc-flow/pull/16#pullrequestreview-670335428

cylc/flow/task_pool.py
oliver-sanders: Ooh, damn, I didn't spot that one.

cylc/flow/scheduler.py
```diff
@@ -1222,8 +1209,8 @@ def run_event_handlers(self, event, reason):
             event, str(reason), self.workflow, self.uuid_str, self.owner,
             self.host, self.server.port))
 
-    def process_task_pool(self):
-        """Queue and release tasks, and submit task jobs.
+    def release_queued_tasks(self):
+        """Release queued tasks, and submit task jobs.
```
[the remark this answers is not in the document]
hjoliver: Could use "pop" I guess.

cylc/flow/task_pool.py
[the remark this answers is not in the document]
hjoliver: Another conflict regression, I think.  Good spotting.

cylc/flow/task_pool.py
hjoliver: Yikes, I apologize to every coder in the world for this abomination :face_with_head_bandage:

cylc/flow/task_pool.py
```diff
@@ -712,37 +754,17 @@ def _release_tasks(self):
             itask.waiting_on_job_prep = True
             self.data_store_mgr.delta_task_state(itask)
             self.data_store_mgr.delta_task_queued(itask)
-        if released:
-            LOG.debug(
-                "Queue released:\n"
-                + '\n'.join(f"* {r.identity}" for r in released)
-            )
```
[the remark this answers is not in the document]
hjoliver: This involves less iteration, since we're already iterating over the list for the preceding lines. Do you have a strong preference either for debug vs info on this? (It's debatable IMO).

cylc/flow/task_pool.py
```diff
@@ -928,13 +946,17 @@ def warn_stop_orphans(self):
     def is_stalled(self):
         """Return True if the workflow is stalled.
 
-        A workflow is stalled if the active pool contains only unhandled
-        failed tasks.
+        A workflow is stalled if it is not held and the active pool contains
+        only unhandled failed tasks and un-released runahead tasks.
+
         """
+        self.release_runahead_tasks()
```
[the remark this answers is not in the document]
hjoliver: Sorry, I don't follow?

This is here in case there are runahead-limited tasks that could be released (because they are now below the runahead limit) at the time of stall check, in which case, we're not stalled.  Unfortunately I don't recall why I thought that was necessary at this point. ...  I'll go back and see if I can see why.

cylc/flow/task_pool.py
hjoliver: No, this has been (wrongly) `critical` for ages. It's not really a change on this branch, just part of a cut-and-paste move of several lines to another location.

However, I actually changed it to `info` myself in f01c4233c0 from #4218 so I'm a bit confused as to why it's shown up as `critical` again here. Anyhow, I'll (re)fix it.

cylc/flow/scheduler.py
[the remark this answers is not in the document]
hjoliver: (But, out of scope for this PR)

cylc/flow/broadcast_mgr.py
hjoliver: (Fixed by @oliver-sanders ' PR now anyway).

cylc/flow/task_pool.py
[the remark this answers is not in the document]
hjoliver: Fixed.

cylc/flow/task_pool.py
hjoliver: `_queue_tasks()` is only called by `reload_taskdefs()`. It figures out which tasks need to be queued, then calls the public `queue_tasks()` to do it.  I'll inline the private one into `reload_taskdefs()`, there's no good reason to have it as a separate method.

hjoliver: (`_queue_tasks()` used to be called in the main loop, as well as at reload, to iterate the pool and find tasks to be queued. But now have event-driven queuing of tasks so that's not neded).

cylc/flow/scheduler.py
```diff
@@ -672,8 +666,7 @@ def _load_pool_from_point(self):
             parent_points = tdef.get_parent_points(point)
             if not parent_points or all(
                     x < self.config.start_point for x in parent_points):
-                initial_tasks.append(TaskID.get(tdef.name, point))
```
hjoliver, line 675: (Not used anywhere).

cylc/flow/scheduler.py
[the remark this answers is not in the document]
hjoliver: (Reformulating the task pool is definitely on the table, as a possible future enhancement).

cylc/flow/scheduler.py
```diff
@@ -1441,12 +1413,66 @@ async def main_loop(self):
...
+
             self.proc_pool.process()
 
-            if self.should_process_tasks():
-                self.process_task_pool()
+            # Get tasks with newly-satisfied external triggers.
+            check_if_ready = self.broadcast_mgr.check_ext_triggers(
+                [x for x in self.pool.get_tasks()
+                    if x.state(TASK_STATUS_WAITING)
+                    and not x.state.is_queued
```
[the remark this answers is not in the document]
hjoliver: No reason that I can think of. There shouldn't be state changes between each iteration. I've extensively refactored this bit along the lines you've suggested. See what you think.

cylc/flow/task_proxy.py
```diff
@@ -63,9 +63,6 @@ class TaskProxy:
         .late_time:
             Time in seconds since epoch, beyond which the task is considered
             late if it is never active.
-        .manual_trigger:
-            Has this task received a manual trigger command? This flag is reset
-            on trigger.
```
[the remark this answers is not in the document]
hjoliver: Possibly to do with the distinction between force triggering queued and unqueued tasks (broken on current master). I've now fixed that on this branch (without restoring this flag).

hjoliver:
Apologies reviewers!  In addition to my several embarrassing crimes again good iterative form (hopefully I would have found the worst of those myself if I had reviewed my own code properly!) ... there were some `TODO`s indicating unfinished work on this branch, which I had neglected to come back to.  I have now done that, plus addressed all review comments, plus this:

- restored correct force-triggering behaviour in the presence of queues (broken since spawn-on-demand) implementation: triggering an unqueued tasks queues it; triggering a queued task causes it to run.

Hopefully good to go now :crossed_fingers:

cylc/flow/task_pool.py
MetRonnie: Wait, my bad, didn't think about the side effects of `self._set_expired_task()`

MetRonnie (approved):

cylc/flow/task_pool.py
```diff
@@ -303,40 +302,32 @@ def release_runahead_tasks(self):
...
+            for cycle in self.main_pool.values()
+            for itask in cycle.values()
+            if itask.state.is_runahead
+            if itask.state(
+                TASK_STATUS_FAILED,
+                TASK_STATUS_SUCCEEDED,
+                TASK_STATUS_EXPIRED
+            )
+            or itask.is_manual_submit
+        ):
```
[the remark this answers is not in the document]
hjoliver: I used to agree, although I've pretty much got my eye in these constructs now.  In any case, you can blame @oliver-sanders ' side-PR for this one!  Oliver, why exactly do you prefer this kind of thing over nested loops with conditional blocks in them? I doubt that comprehensions are any more efficient unless they work on basic types.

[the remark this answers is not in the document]
hjoliver: (However, the variable `cycle` in this one is badly named, I'll fix that).

hjoliver:
Three approvals should do it!

cylc/flow/task_pool.py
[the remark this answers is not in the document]
oliver-sanders: > I used to agree

Ditto, however, once I got used to comprehensions I changed my mind, I think they are more readable and the filtering is much clearer (especially than the `if <cond>: continue` pattern (negative logic etc).

PR #4109, opened 2021-03-04
Fail by default if implicit/naked tasks present

### LINKED ISSUE #3866: Decide on the future of naked tasks.

MetRonnie:
These changes close #3866

Implicit tasks (task without an (empty) section in `flow.cylc[runtime]`) are now disallowed by default, and will raise an error on validation.

Added a setting `flow.cylc[scheduler]allow implicit tasks` to allow them.

Also removed the `--strict` option from `cylc validate`, as it no longer does anything.

**Requirements check-list**
- [x] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/207
- [x] No dependency changes.


tests/functional/graphing/03-filter/flow.cylc
```diff
@@ -1,27 +0,0 @@
-[scheduling]
```
oliver-sanders, line 1: Another good catch :clap:

oliver-sanders (approved):

hjoliver:
(All functional test batches failing).

MetRonnie:
😫  I think we can call this "tests pass"

image

oliver-sanders:
Looking at the timestamps the last updated as 15 minutes before the timeout so this isn't just a matter of an over-zealous timeout:

```
Thu, 11 Mar 2021 13:27:52 GMT [18:57:52] tests/f/reload/08-cycle.t ...................................... ok    83887 ms ( 0.01 usr  0.01 sys + 97.73 cusr 29.58 csys = 127.33 CPU)
Thu, 11 Mar 2021 13:43:31 GMT Error: The action has timed out.
```

This is the result of `cylc scan --states=all` which gives us an idea of what tests were causing the issue:

```
cylctb-20210311T130819Z-klTO/f/jinja2/11-logging (stopped)
cylctb-20210311T130819Z-klTO/f/restart/28-execution-timeout (running)
```

MetRonnie:
It's passing 🎉 (apart from a seemingly spurious pair of codecov checks)

CHANGES.md
```diff
@@ -70,6 +70,11 @@ queueing logic centralized.
 how workflows are restarted
 ([#4040](https://github.com/cylc/cylc-flow/pull/4040)).
 
+"Implicit"/"naked" tasks (tasks that do not have an explicit definition in
+`flow.cylc[runtime]`) are now disallowed by default
+([#4109](https://github.com/cylc/cylc-flow/pull/4109)). You can allow them by
+setting `flow.cylc[scheduler]allow implicit tasks` to `True`.
+
```
hjoliver, line 77: We'll need to tidy the change log a little before release - some entries (like this) added outside of the useful "enhancements" and "fixes" sections.

hjoliver (approved):
Nice, thanks @MetRonnie :+1:

CHANGES.md
MetRonnie: I wasn't sure if it was a "notable" change for putting above the others or not

hjoliver: OK, we haven't done that in the past on cylc-flow. Probably a good idea, we'll just need to add the appropriate "notable changes" heading.

Issue #3327, opened 2019-08-29
zmq: implement SSH tunnelling


oliver-sanders:
Supersedes #2975 
See also #2978

Implement TCP over SSH communication method.

This could be quite straight forward as ZMQ has [built-in functionality](https://pyzmq.readthedocs.io/en/latest/ssh.html).

(To make searching for this easier, it is about re-implementing the cylc 7 "ssh task communication method")

datamel:
I have looked into this. Although ZMQ does have ssh tunnelling built in, when I tried to implement it became apparent that we would not be able to use this. 
A grep of the sshd_config on one of our vlds returned: AllowTcpForwarding no
Since TCP forwarding is prohibited, this method will not work.
I will have a look at reimplementing the old method, unless there is a preferred method to try?!

hjoliver:
@datamel - what's a "vld"?   By "we" I assume you mean "Met Office"?   I'm not sure that's a good reason not to do this.

Firstly, you don't actually need "cylc ssh comms" at Met Office, do you? (It's only a back-up method for those who can't use our preferred network comms)

Secondly, if you do need it, the first resort should be to persuade your system admins to allow TCP forwarding because a critical bit of their infrastructure needs or, or otherwise give extremely good 
defensible reasons for not doing so (which implies extra work for us and additional complexity in Cylc).

Ping @dpmatthews

oliver-sanders:
Quick note, TCP via SSH isn't (as far as I'm aware) something we would be interested in at the Met Office.

This feature is of most use to people running on infrastructure they have little control over. The worry we have is that the requirement to enable tcp forwarding might make this feature off-limits in such cases.

The old Cylc7 ssh somewhere then open a Cylc client approach only used a few lines of code and won't be hard to resurrect.

datamel:
Apologies @hjoliver I think my post was unclear.
 
…I was not aware that vld was not a widespread use term, the vlds are Met Office virtual linux desktops (I’m still learning what is standard and what is met office jargon!)

By “we” I meant the Cylc team, I was aware that this was for alternative organisations that would not be able to use the standard ZMQ. I should have been more explicit – my thinking was that if (at the Met Office) we have TCP forwarding switched off, other organisations may also have it disabled. It would be unfortunate to push out a feature for users to get around not being allowed to use ZMQ, which is still not viable. 
There is clearly a significant level of risk around allowing SSH tunnelling 
[...]

PR #4111, opened 2021-03-05
tests: prepare for ssh task comms

oliver-sanders:
* use extglobs for platform pattern matching comms:?(tcp|ssh)
* upgrade cylc message tests to run on remote platforms (as well as
  local ones)
* set cylc message tests to run for ssh task comms (as well as tcp)
* set fileinstall tests to run for ssh (as well as tcp)
* improve test battery documentation

This should make testing #3327 easier, deals with some of the less well documented aspects of the functional testing framework.

**Requirements check-list**
- [x] No dependency changes.


tests/functional/README.md
```diff
@@ -32,31 +32,100 @@ command line interfaces / outputs, etc.
...
+
+By default tests require the platform `_local_background_indep_tcp`.
+
+If you want your test to run on any other platform export an environment
+variable called `REQUIRE_PLATFORM` *before* the `test_header` is sourced.
+
+This variable should define a test's requirements of a platform e.g:
+
+```bash
+# require a remote platform with a shared filesystem ans tcp task comms
```
MetRonnie, line 104:
```suggestion
# require a remote platform with a shared filesystem and tcp task comms
```

tests/functional/README.md
```diff
@@ -32,31 +32,100 @@ command line interfaces / outputs, etc.
...
+```
+
+Bash extglob pattern matching can be provided
+(https://www.gnu.org/software/bash/manual/html_node/Pattern-Matching.html)
+e.g:
+
+```bash
+# run for all job runners with tcp comms on remote platforms
+export REQUIRE_PLATFORM='loc:remote runner:* comms:tcp'
+# run for tcp and ssh comms on remote platformexprot REQUIRE_PLATFORM='loc:remote comms:?(tcp|ssh)'
```
MetRonnie, line 115:
```suggestion
# run for tcp and ssh comms on remote platforms
```

etc/conf/ssh_config
```diff
@@ -13,6 +13,9 @@ Host _remote*
 Host _remote_background_indep_tcp
     Port 42220
 
+Host _remote_background_indep_ssh
+    Port 42220
+
```
MetRonnie, line 18: Is this supposed to be different to `_remote_background_indep_tcp`?

datamel: Same port - Oliver told me on Friday :)

oliver-sanders: Yep, SSH defaults to use the TCP protocol anyway so we can use the same container for both.

> **Note:** We use a different container for the polling tests which should not be capable of TCP comms just to make sure. It's a bit over the top but should give us good confidence that polling works correctly.

I've added `_remote_background_indep_ssh` as a separate host to fit in with the same approach (easier to handle in actions, etc) but we could just use the same host here.

We could add another container (duplicate of the TCP one) to the swarm to fully separate them.

MetRonnie (approved):

datamel (approved):
Test failures unrelated, changes look good to me. Thanks Oliver!

hjoliver (approved):
Tests passing now :+1:

Issue #4112, opened 2021-03-05
comms timeout


oliver-sanders:
The `--comms-timeout` option from Cylc7 is still present in Cylc8, however, it doesn't do quite what it says on the tin.

It is now a global timeout, it is implemented in the Python layer, however:

```bash
# this
$ cylc <cmd> --comms-timeout=<time>
# is roughly equivalent to this
$ timeout <time> cylc <cmd>
```

Questions:

1) Reconsider the case for timeouts and determine how they should look in the future

hjoliver:
I guess we need a timeout for cylc commands executed by task jobs, to avoid hanging jobs unnecessarily. For user-run commands it doesn't matter so much.

hjoliver:
The issue is, it's not actually a "comms timeout" - should we rename it, to just "timeout"?

Better check that there is a default timeout esp. for `cylc message`

hjoliver:
But ... it does apply only to commands that contact a scheduler.  

Decision: get rid of the option, replace it with a network timeout configured in global.cylc.  PT30S default?  Current default for messaging - PT30S.

oliver-sanders:
~Current default for messaging - PT30S.~

Current default for messaging - PT5S.

oliver-sanders:
Decision:

* Remove the `--comms-timeout` option (no use case0.
* Make this timeout configurable.
* Set the default to `PT30S`.

dpmatthews:
Note that the Cylc 7 `[task messaging]connection timeout` setting was removed in #3402.

oliver-sanders:
Slight problem with the above decision...

The new `comms timeout` global configuration will be loaded from the global config on the platform from which the request was made. This means that in order to set this configuration properly it must be set on all platforms.

The alternative being to add a field to the contact file to allow the value to be passed through to remote platforms.

Options:

1. Require the `comms timeoout` to be set on all platforms.
2. Copy the `comms timeout` into the contact file.
3. Just do away with it or hardcode a sensible default.

BTW: We are currently doing (3) with a hardcoded default of `PT5S`. Haven't seen any issues yet.

hjoliver:
IMO a hard coded timeout is fine, as well as simpler, so long as it is long enough to handle reasonable latencies.

datamel:
Discussed in VC 18/11/21. Decision remove this from rc1 and track use cases before proceeding with this issue.

hjoliver:
[...]

Issue #4007, opened 2020-12-21
swarm: use docker images for local tests


oliver-sanders:
Related:
* #3372
* #3935 
* #3936 

The Docker “swarm” used by the test battery uses containers to emulate remote job platforms and runs the tests themselves on the local platform.

This largely replicates the old system; it allows us to run the test battery on the system it will be deployed on including the job platforms available which is very useful.

However:

* This does introduce some complexity in that the host system needs to be configured to see these containers as remote hosts. This is highly useful as it allows us to use these containers not just for testing but also for live development, however, it is a pain where this functionality is not desired.
* Also, we don’t have an integrated solution to testing against a build matrix of Python/Bash versions, etc.
* Testing job runners would require installation and configuration of the job submission system on the host machine in order to be able to submit jobs to the container which would be very, very messy.

I suggest that we make it possible to run the whole test battery in containers where desired whilst still supporting running tests locally. We can then install and configure the job runner stuff on these images allowing testing of slurm, pbs, etc. You wouldn’t be able to use these containers from your local system (without doing the heavy lifting yourself) but would still be able to use the background/at containers as well as available platforms for local testing.

Something along these lines:

```console
$ ctb -c cylc_py37_bash_5 -p ‘_remote_background_indep_tcp’ tests/f
$ docker cylc_py37_bash_5 exec pytest tests/u
```

![text4899-9](https://user-images.githubusercontent.com/16705946/102775370-a224d900-4384-11eb-89ce-ec918fea0fe3.png)

oliver-sanders:
See comment on this issue https://github.com/cylc/cylc-flow/pull/4197 about building images with both pip/conda to test dependency changes.

oliver-sanders:
Elaboration on the above comment...

* We use two containers.
* On the first container:
  * We install Conda and Conda pack (can use the miniforge or mambaforge base image)
  * We use this to build a Cylc environment using the `conda-environment.yml`.
  * We `conda-pack` this into a tarball.
* On the second container:
  * We unpack this tarball.
  * We symlink the `<env>/bin/cylc` to the relevant system `bin/` dir.
  * We `pip install -e` the cylc-flow repo.

We then dispose of the first container (job done).

PR #4119, opened 2021-03-11
Task messaging ssh

datamel:
Task communication ssh added.

New contact file fields: `SCHEDULER_SSH_COMMAND`, `SCHEDULER_CYLC_PATH`, `SCHEDULER_USE_LOGIN_SHELL`
Removed unused contact file field: `COMMS_PROTOCOL_2`

These changes close https://github.com/cylc/cylc-flow/issues/3327
In light of discussions about `comms_timeout` at latest Cylc Video Conference, this relate to https://github.com/cylc/cylc-flow/issues/4112 (`comms_timeout` added to ssh comms in this PR)
**Requirements check-list**
- [x] Already covered by existing tests. (ssh platform added by 
- [x] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/214.
- [x] No dependency changes.


datamel:
There is currently an issue with GH actions - looks like the ssh keys are having problems authenticating. Getting a `Host key verification failed` return from the ssh. I'm looking into this now.

datamel:
To update, the ssh comms is currently untestable on our CI build with the current GH actions setup (we are unable to ssh from the remote to the scheduler for security reasons).  Hence the codecov/patch diff is low. 

Tests should run locally, with the addition of https://github.com/cylc/cylc-flow/blob/ad126f2d81c95afc3472a5c54d6ee0abb355f93c/etc/conf/global.cylc#L10 uncommented and added to your `global-tests.cylc` file.

ssh tests can then be run with:
`$ etc/bin/run-functional-tests -j 4 -p _remote_background_indep_ssh $(git grep --name-only 'REQUIRE_PLATFORM' tests/functional/)`

datamel:
> There is currently an issue with GH actions - looks like the ssh keys are having problems authenticating. Getting a `Host key verification failed` return from the ssh. I'm looking into this now.

Incidentally, @oliver-sanders, I got to the bottom of the `tests/functional/platforms/02-host-to-platform-upgrade.t` failure - the upgrader was returning a platform with `task communication =ssh` since it does not take this into consideration.
Failure now solved, I had left some of the ssh platforms uncommented in the `global.cylc`. Although this has drawn my attention to the need for site `global.cylc` to be configured correctly and ensuring the correct ordering of platforms - the upgrader can return a platform with a different communication method depending on order. Wondering if this will need to be communicated to users who are continue with using old remote/host configuration but who can not use ZMQ?

oliver-sanders:
To add to Mel's comments on why we can't use the Docker images to test SSH comms at the moment.

This issue proposes [optionally] running the tests themselves in docker (the test host not just the remote hosts) which should enable Scheduler -> Remote -> Scheduled SSH comms (as well as clearing up much of the configuration junk).

https://github.com/cylc/cylc-flow/issues/4007

hjoliver (approved):
Looks good, tested as working :+1: :tada:

cylc/flow/network/client.py
```diff
@@ -223,15 +224,28 @@ def serial_request(self, command, args=None, timeout=None):
...
-        cmd = sys.argv[0]
+
+        ssh_connection = os.getenv("SSH_CONNECTION")
+        host = socket.gethostname()
+        if ssh_connection is None:
+            comms_method = 'local command'
+        elif self.host and (
+                socket.gethostbyname(host) ==
+                socket.gethostbyname(str(self.host))):
+            comms_method = CommsMeth.SSH
```
oliver-sanders, line 242: Could this be `get_comms_method`?

Many users SSH into their systems to gain access so `SSH_CONNECTION` is non-null even for "local" usage.

cylc/flow/network/ssh_client.py
```diff
@@ -0,0 +1,103 @@
...
+            The host where the flow is running if known.
+
+            If host is provided it is not necessary to load
+            the contact file.
+    """
+    def __init__(
+            self,
+            suite: str,
+            timeout: Union[float, str] = None,
+            host: str = None,
```
oliver-sanders, line 49: Could we swap these around so that `suite` and `host` use the same order as `cylc.flow.network.client.SuiteRuntimeClient`:

```suggestion
            host: str = None,
            timeout: Union[float, str] = None
```

cylc/flow/remote.py
```diff
@@ -371,6 +371,7 @@ def _remote_cylc_cmd(
         remote command.
 
     """
+
```
oliver-sanders, line 374: Not important but BTW there isn't supposed to be a blank line *after* the docstring for functions/methods, only classes (black will strip these out).

oliver-sanders (commented):
Looks good, working on getting the tests passing on local platforms...

Two commands have not been converted to use the new `get_client` interface:

* `cylc scan`.
* `cylc tui`.

Neither are important (as they are unlikely to be used in remote jobs), however, it would be good to convert them to match the other commands. `cylc scan` cannot be converted at the moment as is requies the `async_request` interface which is not yet implemented. No biggie, have punted to another issue #4129, should be simple, can tackle later.

Could do with addressing the first comment bellow.

oliver-sanders:
Ran tests against local platforms, all good, added these two to my regular test matrix:

* _remote_background_shared_ssh
* _remote_background_indep_ssh

cylc/flow/network/client.py
datamel: I've run some tests and `get_comms_method` wont work here unfortunately, since for configured SSH comms at this stage we are on the scheduler which will return zmq for the comms method. I can see my workaround of looking at the latest SSH command and comparing hosts would be problematic in the case where users SSH first. I shall have a little think about this and run some tests to see if I can get a more reliable way of returning the correct communication method.

oliver-sanders: Can we hardcode this on the client:

```python
# cylc.flow.network.client
class SuiteRuntimeClient:
    COMMS_METH = 'zmq'

    def get_headers(cls):
    ...
    return {
        'meta': {
            ...
            'comms method': cls.COMMS_METH
        }
    }

# cylc.flow.network.ssh_client:
    COMMS_METH = 'ssh'
```

datamel: Update, COMMS_METH could not be hard coded (the zmq would overwrite the ssh one) and had to be added as an environment variable (CLIENT_COMMS_METH) and sent via the ssh calling `cylc client`.

oliver-sanders:
Two conflicts with the `cylc pause` changes.

datamel:
> Two conflicts with the `cylc pause` changes.

Now resolved.

oliver-sanders (approved):
Rerun tests, all passing :+1:!

PR #4121, opened 2021-03-11
Fix Mypy

MetRonnie:

I have been driven mad falling down a rabbit hole trying to understand why an imported function from another file in `cylc/flow/`  that was properly type-hinted, did not raise any errors if I passed args of an incorrect type to it and then ran `mypy`!

Turns out we needed to add some mypy options to get it to work fully. There is some useful doc here: https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-paths-to-modules

However, I have also discovered that mypy is skipping lots of methods and functions that don't use type hints in their args, even if type hints are used elsewhere in the file. Sorting that out can come later, I think.

**Requirements check-list**
- [x] Does not need tests
- [x] No dependency changes.

### DIFF
--- a/.github/workflows/test_fast.yml
+++ b/.github/workflows/test_fast.yml
@@ -62,7 +62,9 @@ jobs:
 
       - name: typing
         run: |
-          mypy
+          mypy --explicit-package-bases
+        # Need --explicit-package-bases because we don't have __init__.py in
+        # top-level dir
 
       - name: Doctests
         run: |
--- a/cylc/flow/xtrigger_mgr.py
+++ b/cylc/flow/xtrigger_mgr.py
@@ -243,10 +243,10 @@ class XtriggerManager:
         """Get a real function context from the template.
 
         Args:
-            itask (TaskProxy): task proxy
-            label (str): xtrigger label
+            itask: task proxy
+            label: xtrigger label
         Returns:
-            SubFuncContext: function context
+            function context
         """
         farg_templ = {
             TMPL_TASK_CYCLE_POINT: str(itask.point),
@@ -255,7 +255,6 @@ class XtriggerManager:
         }
         farg_templ.update(self.farg_templ)
         ctx = deepcopy(self.functx_map[label])
-        ctx.point = itask.point
         kwargs = {}
         args = []
         for val in ctx.func_args:
@@ -308,7 +307,7 @@ class XtriggerManager:
                         xtrigger_env = [{'environment': {key: val}} for
                                         key, val in res.items()]
                         self.broadcast_mgr.put_broadcast(
-                            [str(ctx.point)],
+                            [str(itask.point)],
                             [itask.tdef.name],
                             xtrigger_env
                         )
--- a/cylc/flow/xtrigger_mgr.py
+++ b/cylc/flow/xtrigger_mgr.py
@@ -18,7 +18,7 @@ import json
 import re
 from copy import deepcopy
 from time import time
-from typing import List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 from cylc.flow import LOG
 import cylc.flow.flags
@@ -42,7 +42,7 @@ TMPL_TASK_NAME = 'name'
 TMPL_SUITE_RUN_DIR = 'suite_run_dir'
 TMPL_SUITE_SHARE_DIR = 'suite_share_dir'
 TMPL_DEBUG_MODE = 'debug'
-ARG_VAL_TEMPLATES = [
+ARG_VAL_TEMPLATES: List[str] = [
     TMPL_TASK_CYCLE_POINT, TMPL_TASK_IDENT, TMPL_TASK_NAME, TMPL_SUITE_RUN_DIR,
     TMPL_SUITE_SHARE_DIR, TMPL_USER_NAME, TMPL_SUITE_NAME, TMPL_DEBUG_MODE]
 
@@ -100,16 +100,15 @@ class XtriggerManager:
     def __init__(
         self,
         suite: str,
+        broadcast_mgr: BroadcastMgr,
+        data_store_mgr: DataStoreMgr,
+        proc_pool: SubProcPool,
         user: Optional[str] = None,
-        *,  # following must be keyword args
-        broadcast_mgr: Optional[BroadcastMgr] = None,
-        data_store_mgr: Optional[DataStoreMgr] = None,
-        proc_pool: Optional[SubProcPool] = None,
         suite_run_dir: Optional[str] = None,
         suite_share_dir: Optional[str] = None,
     ):
         # Suite function and clock triggers by label.
-        self.functx_map: dict = {}
+        self.functx_map: Dict[str, SubFuncContext] = {}
         # When next to call a function, by signature.
         self.t_next_call: dict = {}
         # Satisfied triggers and their function results, by signature.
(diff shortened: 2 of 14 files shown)


.github/workflows/test_fast.yml
```diff
@@ -62,7 +62,9 @@ jobs:
 
       - name: typing
         run: |
-          mypy
+          mypy --explicit-package-bases
+        # Need --explicit-package-bases because we don't have __init__.py in
+        # top-level dir
```
MetRonnie, line 67: This option will also become available as a config item in `mypy.ini` at some point (it's available on mypy@master but not in the latest 0.812 release)

oliver-sanders:
> With --explicit-package-bases, mypy will locate the nearest parent directory that is a member of the MYPYPATH environment variable, the mypy_path config or is the current working directory. Mypy will then use the relative path to determine the fully qualified module name.

:facepalm: 

I think packaging is perhaps the main weakness of Python.

cylc/flow/xtrigger_mgr.py
```diff
@@ -133,6 +133,13 @@ def __init__(
             TMPL_SUITE_SHARE_DIR: suite_share_dir,
             TMPL_DEBUG_MODE: cylc.flow.flags.debug
         }
+
+        if (proc_pool is None or broadcast_mgr is None or
+                data_store_mgr is None):
+            raise TypeError(
+                "kwargs proc_pool, broadcast_mg and data_store_mgr may not "
+                "be None.")
+
```
oliver-sanders, line 142: I think the tests are passing in `None` here.

MetRonnie: Changed to https://github.com/cylc/cylc-flow/blob/c7af49d4d0de15bb4b499a2fc5fb3b51bea3f7f9/cylc/flow/xtrigger_mgr.py#L137-L142

cylc/flow/xtrigger_mgr.py
```diff
@@ -133,9 +133,13 @@ def __init__(
...
-        self.proc_pool: 'SubProcPool' = proc_pool
-        self.broadcast_mgr: 'BroadcastMgr' = broadcast_mgr
-        self.data_store_mgr: 'DataStoreMgr' = data_store_mgr
+
+        if proc_pool is not None:
+            self.proc_pool: SubProcPool = proc_pool
+        if broadcast_mgr is not None:
+            self.broadcast_mgr: BroadcastMgr = broadcast_mgr
+        if data_store_mgr is not None:
+            self.data_store_mgr: DataStoreMgr = data_store_mgr
```
oliver-sanders, line 142: Is this to suppress mypy issues when these attrs are used later on? Don't undefined attrs cause mypy failures?

oliver-sanders:
Rerunning tests after the cylc-rose merge.

cylc/flow/xtrigger_mgr.py
MetRonnie: 1) Yes; 2) No

The only 2 methods that make use of `self.proc_pool`, `self.broadcast_mgr` and `self.data_store_mgr` are `satisfy_xtriggers()` and `callback()`. Perhaps we should instead raise an error at the start of those methods if the required properties are `None`?

oliver-sanders: 2. :facepalm:

The `None` pattern here is pretty yuck, just done a quick grep through the code and I think the reason it it like this is because `cylc.flow.config` creates an `XTriggerManager` in validate mode.

If that's the only barrier it should be easy to navigate, `add_trig` does some kind of validation so perhaps:

```python
class XTriggerManager:
   ...
   @staticmethod
   def validate(...):
        ...
    def add_trig(self, ...):
        self.validate(...)
        ...
```

```python
if self.xtrigger_mgr:
    self.xtrigger_mgr.add_trig(...)
else:
    # ensure the validation logic is still carried out by cylc validate
    XtriggerManager.validate(...)
```

MetRonnie: > protecting you from the little error whilst ignoring the lethal one.

I guess either way it would still fail loudly in the same way?

(I've edited your comment as the indentation was one space off but rendering badly)

oliver-sanders: My idea was to move that into the `validate` method.

oliver-sanders: > I guess either way it would still fail loudly in the same way?

MyPy will fail if the attr is potentially of the wrong type but won't fail if the attr is potentially not set at all.

MetRonnie:
> ```
> cylc/flow/xtrigger_mgr.py:258: error: "SubFuncContext" has no attribute "point"
> Found 1 error in 1 file (checked 180 source files)
> ```



Here that is on master:

https://github.com/cylc/cylc-flow/blob/ad126f2d81c95afc3472a5c54d6ee0abb355f93c/cylc/flow/xtrigger_mgr.py#L256-L257

oliver-sanders:
Set on line 257, used on line 310?

I think it can be replaced by `itask.point`.

.github/workflows/test_fast.yml
wxtim: [According to the docs](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-explicit-package-bases) I think that you may need to add `--namespace-packages`

> This option is only useful in conjunction with --namespace-packages.

MetRonnie:
Asking Bruno to take a look at the changes to xtrigger manager and its tests

.github/workflows/test_fast.yml
wxtim: Having said which it seems to work locally...

wxtim (approved):
- [x] Read and sanity checked. (1 Question to resolve)
 - [x] Ran locally. 
 - [ ] Happy that it's passing all tests.

pytest.ini
```diff
@@ -26,4 +26,6 @@ addopts = --verbose
     --ignore=cylc/flow/parsec/empysupport.py
     --ignore=cylc/flow/parsec/example
 testpaths =
+    cylc/flow/
     tests/unit/
+    tests/integration/
```
MetRonnie, line 31: This allows VSCode's test explorer to pick up on those two folders

oliver-sanders: This was purposfully omitted so that the default `pytest` just ran the unit tests as the integration test battery will become slower as it grows.

Personally I'm happy to change (can just do `pytest tests/u` if you only want unit tests).

.github/workflows/test_fast.yml
oliver-sanders: Done but in the mypy config - https://github.com/cylc/cylc-flow/pull/4121/files#diff-6f2d4ba9ca9a357d31014946667b7bed1bfdbc6d2530afc77778fa0a36bee457R8

(where this setting will get migrated in due course)

setup.py
```diff
@@ -109,7 +109,7 @@ def find_version(*file_paths):
     package_data={
         'cylc.flow': [
             'etc/*.yaml', 'etc/flow*.eg', 'etc/job.sh',
-            'etc/syntax/*', 'etc/cylc-bash-completion'
+            'etc/syntax/*', 'etc/cylc-bash-completion', 'py.typed'
```
oliver-sanders, line 112: Just to check my understanding this is an empty file that gets bundled with the package (that never gets added to?) that acts as a flagfile to mypy?

oliver-sanders (approved):

setup.py
MetRonnie: Yes, I believe so 🤷

kinow (approved):
Changes look good! Enabled mypy on Pycharm, pointing to the mypy configuration and with `--explicit-package-bases` as arguments. No issues detected :tada: 

image

Thanks @MetRonnie !

kinow:
Three approvals I think. LGTM

PR #4124, opened 2021-03-12
Allow comma lists in platform names.

### LINKED ISSUE #4117: platforms: localhost platform patterns
Platforms can be defined with patterns e.g:

```
[platforms]
    [[one,two]]
    [[one|two]]
    [[(one|two)]]
    [[*]]
```

Unfortunately if `localhost` is defined as a pattern it gets ignored due to a `localhost` default defined in the cfgsepc

```
[platforms]
    [[one,localhost]]  # one will get this config, localhost will not
```

Here's the default which is causing the issue:

https://github.com/cylc/cylc-flow/blob/07c7f5c9c785f097066ffdf8d4d5630b5914d734/cylc/flow/cfgspec/globalcfg.py#L607-L608

**Pull requests welcome!**

wxtim:
- Unit test: Check handling of commas in platforms name.
- Unit tests to check handling of platforms where localhost in list.
- Functional test for comma in platform names bug.

These changes (probably) close #4117 

**Requirements check-list**
- [x] No change log entry required - added a feature that was supposed to exist already and is mainly for site admins anyway.
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -187,7 +187,11 @@ def platform_from_name(platform_name=None, platforms=None):
     for platform_name_re in reversed(list(platforms)):
         # We substitue commas with or without spaces to
         # allow lists of platforms
-        if re.fullmatch(re.sub(r',\s*', '|', platform_name_re), platform_name):
+        if (
+            re.fullmatch(re.sub(r',\s*', '|', platform_name_re), platform_name)
+            or
+            re.fullmatch(platform_name_re, platform_name)
+        ):
             # Deepcopy prevents contaminating platforms with data
             # from other platforms matching platform_name_re
             platform_data = deepcopy(platforms[platform_name_re])
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -189,7 +189,7 @@ def platform_from_name(platform_name=None, platforms=None):
         # allow lists of platforms
         if (
             re.fullmatch(
-                re.sub(r'\s*,[^({\d+,\d+})]\s*', '|', platform_name_re),
+                re.sub(r'\s*(?!{[\s\d]*),(?![\s\d]*})\s*', '|', platform_name_re),
                 platform_name
             )
         ):
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -189,7 +189,7 @@ def platform_from_name(platform_name=None, platforms=None):
         # allow lists of platforms
         if (
             re.fullmatch(
-                re.sub(r',[^({\d+,\d+})]\s*', '|', platform_name_re),
+                re.sub(r'\s*,[^({\d+,\d+})]\s*', '|', platform_name_re),
                 platform_name
             )
         ):
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -189,7 +189,11 @@ def platform_from_name(platform_name=None, platforms=None):
         # allow lists of platforms
         if (
             re.fullmatch(
-                re.sub(r'\s*(?!{[\s\d]*),(?![\s\d]*})\s*', '|', platform_name_re),
+                re.sub(
+                    r'\s*(?!{[\s\d]*),(?![\s\d]*})\s*',
+                    '|',
+                    platform_name_re
+                ),
                 platform_name
             )
         ):
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -188,9 +188,10 @@ def platform_from_name(platform_name=None, platforms=None):
         # We substitue commas with or without spaces to
         # allow lists of platforms
         if (
-            re.fullmatch(re.sub(r',\s*', '|', platform_name_re), platform_name)
-            or
-            re.fullmatch(platform_name_re, platform_name)
+            re.fullmatch(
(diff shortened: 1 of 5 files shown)


wxtim:
I'm not very happy about the change I've made here: In the current state of play `vld123` will match the following platform definitions:
- `vld\d{2,3}`
- `vld123, vld456`
- `vld\d{2}|vld\d{3}`

But not
- `vld\d{2}, vld\d{3}`

Hopefully when I pick this up on Monday I'll have a better idea about how to fix...

cylc/flow/platforms.py
```diff
@@ -187,7 +185,14 @@ def platform_from_name(platform_name=None, platforms=None):
     # later than site set platforms) to be matched first and override site
     # defined platforms.
     for platform_name_re in reversed(list(platforms)):
-        if re.fullmatch(platform_name_re, platform_name):
+        # We substitue commas with or without spaces to
+        # allow lists of platforms
+        if (
+            re.fullmatch(
+                re.sub(r',[^({\d+,\d+})]\s*', '|', platform_name_re),
```
oliver-sanders, line 192: Not quite sure what this regex is doing? Think it's replacing `,` with `|` so `\s*,\s*`?

wxtim: That's exactly what it's doing, with an exception to prevent it splitting regexes:
i.e.:
```
platform_name_re = 'abcd\d{2,4}, other_platform' => 'abcd\d{2,4}|other_platform'
# not
'abcd\d{2|4}|other_platform'
```
It was the only use for a comma I could find in the regex spec.

Your suggestion is a good one.

```suggestion
                re.sub(r'\s*,[^({\d+,\d+})]\s*', '|', platform_name_re),
```

oliver-sanders: Ah right didn't think about the match syntax.

Unfortunately this strips off the first character of the next platform name:

```python
>>> re.split(r'\s*,[^({\d+,\d+})]\s*', 'a,b')
['a', '']
```

oliver-sanders: Might need some look aheads / behinds, this might be the right kind of pattern (untested):

```
\s*(?!{[\s\d]*),(?![\s\d]*})\s*
```

MetRonnie: Using a regex to parse regexes 😬

MetRonnie: What about a platform_re that's like `foo[^_,]`

oliver-sanders: How very meta.

oliver-sanders: I don't think commas make for valid hostnames so don't have to worry about that.

wxtim:
```suggestion
\s*(?!{[\s\d]*),(?![\s\d]*})\s*
```
How _did_ you do _that_ (Write a working regex out of your head)

cylc/flow/platforms.py
```diff
@@ -187,7 +185,14 @@ def platform_from_name(platform_name=None, platforms=None):
     # later than site set platforms) to be matched first and override site
     # defined platforms.
     for platform_name_re in reversed(list(platforms)):
-        if re.fullmatch(platform_name_re, platform_name):
+        # We substitue commas with or without spaces to
+        # allow lists of platforms
+        if (
+            re.fullmatch(
+                re.sub(r'\s*,[^({\d+,\d+})]\s*', '|', platform_name_re),
```
wxtim, line 192:
```suggestion
                re.sub(r'\s*(?!{[\s\d]*),(?![\s\d]*})\s*', '|', platform_name_re),
```

cylc/flow/platforms.py
oliver-sanders: Reviewing Matt's PRs for a few years, you learn so many trinkets that way.

tests/unit/test_platforms_get_platform.py
```diff
@@ -30,13 +30,50 @@ def test_get_platform_no_args():
...
 
-def test_get_platform_from_platform_name_str(mock_glbl_cfg):
+@pytest.mark.parametrize(
+    'platform_re',
+    [
+        'localhost',
+        'localhost, otherplatform',
+        'otherplatform, localhost',
+        'otherplatform|localhost',
+        'localhost|otherplatform'
```
oliver-sanders, line 40: Worth adding a test for the `foo{0,1}` syntax here.

tests/unit/test_platforms_get_platform.py
```diff
@@ -30,13 +30,50 @@ def test_get_platform_no_args():
...
+            [[localhost]]
+                hosts = localhost
+                ssh command = ssh -oConnectTimeout=42
+            [[{platform_re}]]
+                hosts = localhost
+                ssh command = ssh -oConnectTimeout=24
         '''
+    )
+    platform = get_platform('localhost')
+    assert platform['ssh command'] == 'ssh -oConnectTimeout=24'
```
oliver-sanders, line 58: Would be good to test this the other way around for safety, something like:

```python
        'localhost|otherplatform',
        None
)
def ...
...
if platform_re:
    assert platform['ssh command'] == 'ssh -oConnectTimeout=24'
else:
    assert platform['ssh command'] == 'ssh -oConnectTimeout=42'
```

oliver-sanders (approved):
Looks good.

MetRonnie (approved):
LGTM assuming tests pass

PR #4127, opened 2021-03-16
latest state change tasks summary

dwsutherland:
These changes partially address https://github.com/cylc/cylc-ui/pull/543

A summary of the latest 5 tasks per state has been added to the API for use in gscan views.

The way it works is by:
- Using a size limited [queue](https://docs.python.org/3/library/collections.html#collections.deque) for each state, at the Scheduler (`cylc.flow`) end, that adds the preformatted task reference string on state change event.
- If there's been a change this (as with state totals) is set in the workflow delta (published to the UIS) as a mapping.
- This mapping is read by the GraphQL resolver and filtered by argument.

state-tasks-summary

**Requirements check-list**
- [x] Appropriate tests are included (added as field of existing GraphQL functional test).
- [x] No dependency changes.


kinow:
Are you still going to add tests and change anything @dwsutherland ? Otherwise I can start reviewing it today after the team meeting. Thanks!!

dwsutherland:
> Are you still going to add tests and change anything @dwsutherland ? Otherwise I can start reviewing it today after the team meeting. Thanks!!

Added as field to existing test, thanks Bruno.

cylc/flow/data_store_mgr.py
```diff
@@ -1335,6 +1349,11 @@ def update_workflow(self):
             w_delta.is_queued_total = is_queued_total
             delta_set = True
 
+            for state, tp_queue in self.latest_state_tasks.items():
+                if tp_queue:
```
kinow, line 1353: I think this is never `None` as we initialize each `state: deque(...)` in the `__init__` function?

kinow (commented):
Code looks good, one small comment. Was going to do a functional test with the UI @dwsutherland , but will hold back until you've updated it (I think there's something you'll have to update for n-window=0 from today's meeting?).

Thanks!

cylc/flow/data_store_mgr.py
dwsutherland: `tp_queue` is empty it doesn't set it .. otherwise you'd get `failed: []` in the feed.

Which would you prefer on the front end processing? list and item always there? or not?

kinow: Ah, true, it also checks for empty. Sorry @dwsutherland , confused it with JavaScript :smile: easier the way you did IMO :+1:

dwsutherland: All good either way, up to you 😃

dwsutherland:
Updated to leave in empty state lists.

dwsutherland:
Looks like coverage get's triggered too early.. should pass once all tests are done.

kinow (approved):
Code looks good. Already tested with Cylc UI https://github.com/cylc/cylc-ui/pull/617, and found no issues. Well done @dwsutherland ! :clap:

cylc/flow/data_store_mgr.py
```diff
@@ -1408,7 +1425,13 @@ def delta_task_state(self, itask):
         tp_delta.stamp = f'{tp_id}@{update_time}'
         tp_delta.state = itask.state.status
         self.state_update_families.add(tproxy.first_parent)
-        # if state is final work our new task mean.
+        if tp_delta.state in self.latest_state_tasks:
+            tp_ref = f'{tproxy.name}.{tproxy.cycle_point}'
+            tp_queue = self.latest_state_tasks[tp_delta.state]
+            if tp_ref in tp_queue:
+                tp_queue.remove(tp_ref)
+            self.latest_state_tasks[tp_delta.state].appendleft(tp_ref)
```
oliver-sanders, line 1433:
```suggestion
            tp_queue.appendleft(tp_ref)
```

oliver-sanders (approved):

PR #4132, opened 2021-03-17
Configurable source dirs

### LINKED ISSUE #4107: configurable source directories
We would like the UI to be able to display uninstalled source workflows so that they can be installed and run without leaving the UI.

**1) Configuration**

Since users may have established source repositories (e.g. `~/roses`) this should be configurable, proposal:

```ini
# global.cylc
[install]
     source dirs = ~/cylc-src, ~/roses, /some/shared/repo, ...
```

**2) Configuration Default**

Proposing that the default should be `~/cylc-src` as an analogue to `~/cylc-run`.

**3) Cylc Install Default**

The first item in this list could be used by `cylc install` as the "default" source location.

Propose using the first argument of `cylc install` to specify the name of a flow within `source dirs[0]` to install i.e:

```ini
# global.cylc
[install]
     source dirs = ~/cylc-src, ~/roses
```

```console
$ cd "$HOME"
$ cylc install foo
INSTALLED foo from ~/cylc-src/foo -> ~/cylc-run/foo/run1
```

As a more convenient alternative to:

```console
$ cylc install -C ~/cylc-src/foo
# OR
$ cd ~/cylc-src/foo; cylc install
```

> **Note:** `cylc install` currently has an argument, however, it is unused in the code.

**Questions:**

* [x] 1 ok?
* [x] 2 ok?
* [x] 3 ok?

MetRonnie:
These changes close #4107 

- Create a global config item

  ```ini
  # global.cylc
  [install]
       source dirs = ~/cylc-src, ~/roses, /some/shared/repo, ...
  ```
  that acts as a search path for `REG` when doing `cylc install REG`

- Default is `~/cylc-src`

**Requirements check-list**
- [x] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/221.
- [x] No dependency changes.


cylc/flow/cfgspec/globalcfg.py
```diff
@@ -283,6 +283,14 @@
                 file size.
             ''')
 
+    with Conf('install'):
+        Conf('source dirs', VDR.V_STRING_LIST, default=['~/cylc-src'], desc='''
+            A list of paths where workflow source directories may be found.
+            This is used to display and install currently non-installed
+            workflows in the UI, and as a search path for installing workflows
+            using ``cylc install REG``.
```
wxtim, line 291: Consider
"""A list of paths where ``cylc install`` will look for workflow source directories.
If a workflow source is in one of these directories you can install that workflow with ``cylc install REG`` - rather than giving the full path to the workflow source. Cylc UI uses source folders when searching for non-installed workflows.
"""

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -283,6 +283,14 @@
                 file size.
             ''')
 
+    with Conf('install'):
+        Conf('source dirs', VDR.V_STRING_LIST, default=['~/cylc-src'], desc='''
+            A list of paths where workflow source directories may be found.
```
wxtim, line 288: I think I want to say why first, then what second, but you may disagree - my suggestion:

A list of paths where ``cylc install`` will look for source directories.
If a workflow source is in one of these directories you can install that workflow with ``cylc install REG``. Cylc UI uses source folders when searching for non-installed workflows.

cylc/flow/scripts/install.py
```diff
@@ -32,25 +29,35 @@
 A .service directory will also be created and used for server authentication
 files at run time.
 
+If the argument REG is used, Cylc will search for the workflow in the list of
+directories given by `global.cylc[install]source dirs`, and install the first
+match. Otherwise, the workflow in the current working directory, or the one
+specified by --directory, will be installed.
```
wxtim, line 35:
```suggestion
specified by ``--directory``, will be installed.
```

cylc/flow/scripts/install.py
```diff
@@ -32,25 +29,35 @@
...
-  # (if "run1" exists this will increment)
+  # Install workflow "dogs/fido" from the first match in
+  # `global.cylc[install]source dirs`, e.g. ~/cylc-src/dogs/fido/flow.cylc
+  # (with run directory ~/cylc-run/dogs/fido/run1 (if "run1" already exists,
+  # this will increment))
   $ cylc install dogs/fido
 
-  # Install $PWD/flow.cylc with specified flow name: fido
+  # Install $PWD/flow.cylc as "rabbit", if $PWD is ~/bunny/rabbit
+  # (with run directory ~/cylc-run/rabbit/run1)
```
wxtim, line 47: This reminds me of a song... 😉

cylc/flow/suite_files.py
```diff
@@ -1305,3 +1339,18 @@ def link_runN(latest_run):
         run_n.symlink_to(latest_run)
     except OSError:
         pass
+
+
+def search_install_source_dirs(flow_name: str) -> Optional[Path]:
+    """Return the path of a workflow source dir if it is present in the
+    'global.cylc[install]source dirs' search path."""
+    search_path: List[str] = glbl_cfg().get(['install', 'source dirs'])
```
wxtim, line 1347: search_paths might be easier to understand?

cylc/flow/scripts/install.py
oliver-sanders: This doesn't get parsed as markup, I do hope we will do one day... but in the mean time most scripts use `"`.

cylc/flow/cfgspec/globalcfg.py
MetRonnie: How about
```rst
A list of paths where ``cylc install <flow_name>`` will look for
a workflow of that name. All workflow source directories that are 
in these locations will show up in the UI, ready for installation.
```

MetRonnie: I should probably add
```rst
.. note:
   If workflow source directories of the same name exist in more
   than one of these paths, only the first one will be picked up.
```

wxtim:
I set up the following tests:
```bash
for location in "sheep pig cylc-src"; do
  mkdir -p "~/${location}/foo"
  echo "# File from ${location}" > "~/${location}/foo/flow.cylc"
done
```
And added the two extra source directories to my source dirs list, and have some questions:
#. Is it OK to ignore ``cylc-src`` if source dirs are set? I think that at the very least a flow from ``cylc-source`` should be installed if it exists there and no-where else.
#. If a workflow is in multiple source dirs, is it safe to just pick the first one? I think that we should at least warn users that this is going on...

cylc/flow/scripts/install.py
```diff
@@ -32,25 +29,35 @@
...
 
-  # Install PATH/TO/FLOW/flow.cylc
-  $ cylc install --directory=PATH/TO/FLOW
+  # Install $PWD/bunny/rabbit/flow.cylc as "bunny/rabbit"
+  # (with run directory ~/cylc-run/bunny/rabbit/run1)
+  $ cylc install --directory=bunny/rabbit
 
-  # Install cats/flow.cylc
+  # Install $PWD/cats/flow.cylc as "cats", overriding the run1, run2, run3 etc
+  # structure.
```
datamel, line 59: I'm having problems getting these examples to work. I shall wait until this PR is marked ready before reviewing.

MetRonnie:
> Is it OK to ignore cylc-src if source dirs are set? I think that at the very least a flow from cylc-source should be installed if it exists there and no-where else.

If they want to include `~/cylc-src`, they should leave it in the `[install]source dirs` search path. If it's not there then at least they will get an error that states the directories that have been looked in (although I should probably do a special case error message for if they set `[install]source dirs = # (empty)`

> If a workflow is in multiple source dirs, is it safe to just pick the first one? I think that we should at least warn users that this is going on...

I think they get what they ask for in this case. It prints out which source dir the workflow was installed from. The behaviour is a standard search path, like `$PATH` - there isn't a warning when you run an executable with the same name as another in `$PATH`, is there?

cylc/flow/scripts/install.py
MetRonnie: Should be the same use cases that are covered by the cylc-install functional tests, passing for me and on GH Actions

cylc/flow/pathutil.py
```diff
@@ -61,64 +66,59 @@ def get_workflow_run_dir(flow_name, *args):
 
 def get_suite_run_job_dir(suite, *args):
     """Return suite run job (log) directory, join any extra args."""
-    return expandvars(
-        get_workflow_run_dir(suite, 'log', 'job', *args)
-    )
+    return get_workflow_run_dir(suite, 'log', 'job', *args)
```
oliver-sanders, line 69: > **Note To Reviewers:**
>
> I've gone through each of these methods with `*args` and I can't see any (rational) way they could
> be provided env vars.

cylc/flow/suite_files.py
oliver-sanders: We've settled on "installing" a "workflow" from a "source dir" to a "run dir".

So "source dir**s**" makes sense.

cylc/flow/scripts/install.py
```diff
@@ -18,11 +18,8 @@
 
 """cylc install [OPTIONS] ARGS
 
-Install a new workflow.
-
-Install the name REG. The workflow server program can then be started, stopped,
-and targeted by name REG. (Note that "cylc play" can also install workflows on
-the fly).
+Install a new workflow. The workflow can then be started, stopped, and targeted
+by name.
```
oliver-sanders, line 22: The first line is used for the command summary i.e. `cylc help all`:

```suggestion
Install a new workflow.

The workflow can then be started, stopped, and targeted by name.
```

Also just a good system for documenting things:

```
Short concise description.

Longer explanation that can span multiple paragraphs, contain examples and extra information.

The short description should be easily skimable. Even if the full meaning is not immediately apparent to all
users the general scope of its functionality should be obvious enough that someone can quickly decide whether
this is information relevant to their search.
```

oliver-sanders (approved):
Looks good, tested as working, one minor whitespace change:

hjoliver (approved):
Nice, thanks @MetRonnie :+1:   Reviewed and tested, no problems found.

PR #4138, opened 2021-03-19
Make remote init return messages more informative

datamel:
The aim of this PR is to ensure the reason for remote init failures is documented in the workflow log.
This PR also adds an additional check for the `client_public_keys` directory on the remote which will catch misconfigurations of the install target for platforms which share fs with localhost.

- [x] No change log entry required (why? invisible to users).
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -493,7 +493,9 @@ def get_random_platform_for_install_target(install_target):
         return random.choice(platforms)
     except IndexError:
         # No platforms to choose from
-        return None
+        raise PlatformLookupError(
+            f'Could not select platform for install target: {install_target}'
+        ) from None
 
 
 def get_localhost_install_target():
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -489,7 +489,11 @@ def get_all_platforms_for_install_target(install_target):
 def get_random_platform_for_install_target(install_target):
     """Return a randomly selected platform (dict) for given install target."""
     platforms = get_all_platforms_for_install_target(install_target)
-    return random.choice(platforms)
+    try:
+        return random.choice(platforms)
+    except IndexError:
+        # No platforms to choose from
+        return None
 
 
 def get_localhost_install_target():
--- a/cylc/flow/task_remote_mgr.py
+++ b/cylc/flow/task_remote_mgr.py
@@ -230,6 +230,9 @@ class TaskRemoteMgr:
             if install_target == get_localhost_install_target():
                 continue
             platform = get_random_platform_for_install_target(install_target)
+            if platform is None:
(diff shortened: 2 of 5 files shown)


hjoliver (approved):
Much more informative :thumbsup:

cylc/flow/platforms.py
```diff
@@ -489,7 +489,11 @@ def get_all_platforms_for_install_target(install_target):
 def get_random_platform_for_install_target(install_target):
     """Return a randomly selected platform (dict) for given install target."""
     platforms = get_all_platforms_for_install_target(install_target)
-    return random.choice(platforms)
+    try:
+        return random.choice(platforms)
+    except IndexError:
+        # No platforms to choose from
+        return None
```
oliver-sanders, line 496: Before this `IndexError` resulted in a strange error the user wouldn't understand, however, now the error will be pushed into the scripts that call the function.

For example in this code we will get a `KeyError`:

https://github.com/cylc/cylc-flow/blob/4947568bb7e247c71e45d722fae495ab643728af/cylc/flow/task_remote_mgr.py#L232-L233

How about raising a `PlatformLookupError`:

```suggestion
        raise PlatformLookupError(
            f'Could not select platform for install target: {install_target}'
        ) from None
```

datamel: I had changed the code in remote tidy to pick up the None return, to stop the Key Error.
https://github.com/datamel/cylc-flow/blob/96cd9601ad603cb50b4ff3131e6aa225f152e899/cylc/flow/task_remote_mgr.py#L233
Unsure why this updated code is not showing up for you, hmmm.
Happy to raise the error if you would prefer.

oliver-sanders: > Unsure why this updated code is not showing up for you

Because I wasn't looking properly!

cylc/flow/task_remote_mgr.py
```diff
@@ -230,6 +230,9 @@ def remote_tidy(self):
             if install_target == get_localhost_install_target():
                 continue
             platform = get_random_platform_for_install_target(install_target)
+            if platform is None:
+                LOG.debug(f"Remote tidy not required for {install_target}.")
+                continue
```
oliver-sanders, line 235: Not sure we want to silently fail here?

This could just mean that none of the platforms or hosts are accessible.

cylc/flow/platforms.py
oliver-sanders: I think it should be an error because this hides the whole selection stuff so no platform could mean platform, network or configuration issues.

datamel: Yes, and I suppose for future use of the method, safest to raise error, I'll update that now.

oliver-sanders (approved):

PR #4139, opened 2021-03-24
Remove queued/held delta filter

### LINKED ISSUE #4135: Datastore misses some deltas
(After discussions on Element with @kinow and @dwsutherland)

The datastore delta methods include a check like this (e.g. for changes to the task instance `is_queued` flag):
```python
  def delta_task_queued(self, itask):              
        """Create delta for change in task proxy queued state.  
        """    
        tp_id, tproxy = self.store_node_fetcher(itask.tdef.name, itask.point)   
        if not tproxy:    
            return   
        if itask.state.is_queued is not tproxy.is_queued:     # <--------------------------------------                
           # create a new delta...
```                

This sometimes results in a delta being omitted when it should be created (if the `delta_task...` method is called more than once in a main loop iteration?).  Result (e.g.) we see some `is_queued` flags still attached to tasks that are running or finished, in the UIs.
Hacking the above line to use `if True:` seems to fix the problem.

Also: the check above is probably unnecessary, because we only call `delta_task_queued()` (for e.g.) when we know that the queued state has changed, so by definition a delta should be created.  Then, if `delta_task_blah()` gets called multiple times before the deltas get applied only the last one will take effect, but that doesn't really matter - the intermediate changes are essentially just too quick for the UI to see.

Depending on implementation, the fix could result in occasional no-op deltas (i.e. cylc-ui gets a delta that says to "change" a quantity to the state that it already is in) but that doesn't really matter, it should be an infrequent event and of little consequence.

dwsutherland:
These changes close #4135

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -1457,13 +1457,12 @@ class DataStoreMgr:
         tp_id, tproxy = self.store_node_fetcher(itask.tdef.name, itask.point)
         if not tproxy:
             return
-        if itask.state.is_held is not tproxy.is_held:
-            tp_delta = self.updated[TASK_PROXIES].setdefault(
-                tp_id, PbTaskProxy(id=tp_id))
-            tp_delta.stamp = f'{tp_id}@{time()}'
-            tp_delta.is_held = itask.state.is_held
-            self.state_update_families.add(tproxy.first_parent)
-            self.updates_pending = True
+        tp_delta = self.updated[TASK_PROXIES].setdefault(
+            tp_id, PbTaskProxy(id=tp_id))
+        tp_delta.stamp = f'{tp_id}@{time()}'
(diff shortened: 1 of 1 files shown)


kinow (approved):
:+1: I still remember the discussion from last week between you and @hjoliver . Looks good to me, and @hjoliver confirmed it fixed the issue, so LGTM :tada: thanks @dwsutherland !

hjoliver:
@dwsutherland - can we be confident that the same problem won't occur in the other delta fields?

dwsutherland:
> @dwsutherland - can we be confident that the same problem won't occur in the other delta fields?

They aren't filtered (i.e. check that the data store state field is different before creating the delta), as far as I know.

hjoliver (approved):
Good, thanks @dwsutherland :+1:

PR #4141, opened 2021-03-24
CYLC_SITE_CONF_PATH: look in flow sub-dir

oliver-sanders:
Required for: https://github.com/cylc/cylc-uiserver/pull/192

Required for us to use `CYLC_SITE_CONF_PATH` to source the global config for the UIS.

* Look for global config in `${CYLC_SITE_CONF_DIR}/flow/global.cylc`.
* This is to allow this env var to be used by other cylc components - namely the UIS.
* Tried to match the annotations from #4099 to minimise diff.

To explain the `:envvar:` stuff - https://github.com/cylc/cylc-doc/issues/226

**Requirements check-list**
- [x] Docs included
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -90,7 +90,7 @@ with Conf('global.cylc', desc='''
     .. envvar:: CYLC_SITE_CONF_PATH
 
        By default the site configuration is located in ``/etc/cylc/``. For
-       installations where this is not convenient this path can be overridden
+       installations where this is not convenient, this path can be overridden
        by setting ``CYLC_SITE_CONF_PATH`` to point at another location.
 
        Configuration for different Cylc components should be in sub-directories
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -836,6 +836,10 @@ class GlobalConfig(ParsecConfig):
             cls._DEFAULT.load()
         return cls._DEFAULT
 
+    def _load(self, fname, conf_type):
+        if os.access(fname, os.F_OK | os.R_OK):
+            self.loadcfg(fname, conf_type)
+
     def load(self):
         """Load or reload configuration from files."""
         self.sparse.clear()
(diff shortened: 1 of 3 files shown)


hjoliver:
(The new unit test is failing)

oliver-sanders:
That should do it, nicer testing approach which bypasses the FS all-together.

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -59,24 +59,48 @@
...
     from those higher up (but if a setting is present in a file higher up and
     not in any files lower down, it will not be overridden).
 
-    Setting the ``CYLC_SITE_CONF_PATH`` environment variable overrides the
-    default value of ``/etc/cylc/flow/``.
+    The following environment variables can change the files which are loaded:
+
+    .. envvar:: CYLC_CONF_PATH
+
+       If set this bypasses the default site/user configuration hierarchy used
```
MetRonnie, line 84:
```suggestion
       If set, this bypasses the default site/user configuration hierarchy used
```

datamel (approved):
I have read the code. Built/checked the docs and run the tests. I'm happy with this change.

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -59,24 +59,48 @@
...
+
+       This should be set to a directory containing a :cylc:conf:`global.cylc`
+       file.
+
+    .. envvar:: CYLC_SITE_CONF_PATH
 
-    To override the entire hierarchy, set the ``CYLC_CONF_PATH`` environment
-    variable to the directory containing your ``global.cylc`` file.
+       By default the site configuration is located in ``/etc/cylc/``. For
+       installations where this is not convenient this path can be overridden
```
MetRonnie, line 93:
```suggestion
       installations where this is not convenient, this path can be overridden
```

tests/unit/scripts/test_config.py
```diff
@@ -14,31 +14,186 @@
...
+from cylc.flow.cfgspec.globalcfg import GlobalConfig
+
 
 Fixture = Any
+HOME: str = str(Path('~').expanduser())
+
+
+@pytest.fixture
+def conf_env(monkeypatch):
+    """Clear any env vars that effect which conf files get loaded.
```
MetRonnie, line 33:
```suggestion
    """Clear any env vars that affect which conf files get loaded.
```

tests/unit/scripts/test_config.py
```diff
@@ -14,31 +14,186 @@
...
+    )
+    return files
+
+
+def test_get_config_file_hierarchy_global(
+    monkeypatch: Fixture,
+    conf_env: Fixture,
+    capload: Fixture,
+    dummy_version_hierarchy: Fixture
+):
```
MetRonnie, line 96: I didn't realise just by passing fixtures to test functions that they would be executed

MetRonnie (approved):
Built docs and ran tests. Just a few optional suggestions

tests/unit/scripts/test_config.py
oliver-sanders: Up until the `return` or `yield` statement, for generators execution continues after the `yield` statement after the test scope exits.

PR #4142, opened 2021-03-24
Record workflow version control information post-install

### LINKED ISSUE #3849: install: record version control information
closely related to #3819 

Re-implement the Rose ability to record version control information.

Write a plugin (using the on-install entry point developed in #3819) that runs on workflow installation to detect if the workflow is being installed from a version controlled working copy and if so record the revision along with any un-committed diff.

> **Note:** The plugin will probably look like this:
    `install(src_dir: Path, run_dir: Path, run_mode: str) -> None`

This info should get dumped into a file somewhere in the workflow run directory.

This should work for git and svn.

**Questions:**

* Where should we dump this info?
   1) `~/<flow>/log/version-control/<timestamp>-<run_mode>`
   2) Somewhere else?
* Are there any nice abstractions which would save us having to support each VC system individually?

MetRonnie:
These changes close #3849 

On installation of a workflow, will record the version control status to `log/version/vcs.conf` and any uncommitted diff to `log/version/uncommitted.diff`.

Supports git and svn.

**Requirements check-list**
- [x] Appropriate change log entry included
- [x] (master branch) I have opened a documentation PR at cylc/cylc-doc#236
- [x] No dependency changes.


MetRonnie:
Well, the fast test failures have me at a loss.

In this function in the tests I check if git is installed, otherwise instruct pytest to skip the test
https://github.com/cylc/cylc-flow/blob/d3e58d9f60fd83b85cacea1cf85bf789c5166a1e/tests/unit/post_install/test_log_vc_info.py#L44-L48

This works locally, but fails on GitHub actions:
```
FileNotFoundError: [Errno 2] No such file or directory: 'command': 'command'
```
This is despite `command -v` working fine in the functional tests

cylc/flow/post_install/log_vc_info.py
```diff
@@ -0,0 +1,285 @@
...
+) -> None:
+    """Write version control info to the workflow's vcs log dir.
+
+    Args:
+        info: The vcs info.
+        run_dir: The workflow run directory.
+    """
+    if not info:
+        raise ValueError("Nothing to write")
+    info_file = Path(run_dir, 'log', 'version', 'vcs.conf')
```
MetRonnie, line 187: I can't see a way to get the timestamp of the installation, perhaps this needs to be added to the entry point arguments?

MetRonnie:
(Converting to draft to prevent merging until 8.0b0 is released]

datamel (approved):
Minor comments. I have tested the change as working, read the code and run the tests. Thanks Ronnie.

cylc/flow/post_install/log_vc_info.py
```diff
@@ -0,0 +1,285 @@
...
+
+def write_diff(diff: str, run_dir: Union[Path, str]) -> None:
+    """Write a diff to the workflow's vcs log dir.
+
+    Args:
+        diff: The diff.
+        run_dir: The workflow run directory.
+    """
+    if not diff:
+        raise ValueError("Nothing to write")
```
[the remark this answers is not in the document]
MetRonnie: Probably not

MetRonnie:
Just remembered a question I had: do we want to save the diff of _untracked_ files for a git source dir? This is actually not the default behaviour of git, `git diff HEAD` only gets the diff of tracked files (unstaged + staged)

datamel:
> Just remembered a question I had: do we want to save the diff of _untracked_ files for a git source dir? This is actually not the default behaviour of git, `git diff HEAD` only gets the diff of tracked files (unstaged + staged)

Could those _untracked_ files make a difference to the running of the workflow? For example, if they had configured their .cylcignore file to ignore directories for the installation but this was not under source control, they may wish to know about it. Although perhaps this could lead to the log filling with unwanted diffs... difficult choice. I guess by marking files as untracked they should assume it would not be included in the diff, perhaps a one line warning at the top reminding them that there may be untracked changed?

MetRonnie:
> Could those untracked files make a difference to the running of the workflow?

Yes

> I guess by marking files as untracked they should assume it would not be included in the diff, perhaps a one line warning at the top reminding them that there may be untracked changed?

I don't think there is a way of marking files as untracked. It's pretty much only newly-created files that haven't been `git add`ed yet that are untracked.

The untracked files will still show up in the `vcs.conf` file with `??` status, e.g.
```
status = "
?? foo/bar.txt
"
```

cylc/flow/post_install/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+"""Plugins for running Python code after installation of a workflow.
+
+Built In Plugins
+----------------
+
+Cylc Flow provides the following post-install plugins:
+
+.. autosummary::
+   :toctree: built-in
+   :template: main_loop_plugin.rst
```
oliver-sanders, line 25: This sets the template for autosummary to follow when building the pages, you may want to fall-back to the default template as the main loop template automatically documents all members.

cylc/flow/post_install/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Plugins for running Python code after installation of a workflow.
+
+Built In Plugins
+----------------
+
+Cylc Flow provides the following post-install plugins:
```
oliver-sanders, line 21: We can document the pre-install and post-install entry points together as "install plugins" since they have related functionality and plugins may use both (e.g. `cylc-rose`).

tests/unit/post_install/test_log_vc_info.py
```diff
@@ -0,0 +1,220 @@
...
+@pytest.fixture(scope='module')
+def git_source_repo(tmp_path_factory: TempPathFactory) -> Tuple[str, str]:
+    """Init a git repo for a workflow source dir.
+
+    The repo has a flow.cylc file with uncommitted changes. This dir is reused
+    by all tests requesting it in this module.
+
+    Returns (source_dir_path, commit_hash)
+    """
+    skip_if_not_installed('git')
```
oliver-sanders, line 60: Can use `pytest.skipif` as a decorator on the test function to avoid having to run tests that would skip.

oliver-sanders (commented):
Looking good, just need to give it a spin.

oliver-sanders (approved):
Tested with Git, all good.

* Needs a quick changelog entry.
* Can you squash down commits, namely "Tidy" ones.

cylc/flow/install_plugins/log_vc_info.py
```diff
@@ -0,0 +1,284 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Record version control information on workflow install.
+
+The supported version control systems are git and svn.
```
oliver-sanders, line 19: From https://github.com/cylc/cylc-doc/pull/236:

Could do with a little more documentation here, a quick outline of:

* Where the version control information is stored.
* The format of this file.
* The meaning of each field.

Can be quite short.

MetRonnie: Added: https://github.com/cylc/cylc-flow/pull/4142/files#diff-050731cd5a466ce73267f816ff7c03dc72ad30e21d92d29db741db0395ba64fc

cylc/flow/install_plugins/log_vc_info.py
```diff
@@ -0,0 +1,317 @@
...
+
+An example of this information for a git repo:
+
+.. code-block:: cylc
+
+   version control system = "git"
+   repository version = "2.8.0-dirty"
+   commit = "e5dc6573dd70cabd8f973d1535c17c29c026d553"
+   working copy root path = "~/cylc-src/my-workflow-git"
+   status = \"\"\"
```
oliver-sanders, line 32: I don't think these quotes should be escaped here?

MetRonnie: Because the whole docstring is a triple quote string? Rendering fine:

PR #4145, opened 2021-03-25
Remove error logging

datamel:
PR: https://github.com/cylc/cylc-flow/pull/4138 added logging exceptions. This additional logging is making the logs look messier than necessary.
The reason this was added was due to an error added at cylc-tidy stage (scheduler side), was not appearing in logs for some reason. This needs further investigation (I will open an issue) 

- [x] Does not need tests (Removal of code, not affecting tests).
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -351,8 +351,7 @@ async def _run(parser, options, reg, scheduler):
         ret = 1
     except (KeyboardInterrupt, asyncio.CancelledError):
         ret = 2
-    except Exception as exc:
-        LOG.exception(exc)
+    except Exception:
         ret = 3
 
     # kthxbye


wxtim:
I booted @oliver-sanders  because I don't think that this needs 2 reviewers.

cylc/flow/scheduler_cli.py
```diff
@@ -351,8 +351,7 @@ async def _run(parser, options, reg, scheduler):
         ret = 1
     except (KeyboardInterrupt, asyncio.CancelledError):
         ret = 2
-    except Exception as exc:
-        LOG.exception(exc)
+    except Exception:
```
wxtim, line 354: IMO we probably need to have the logging of this particular exception, but perhaps it should be at debug level.

wxtim (approved):
Having looked at the history and context of this code I'm now happy with this change.

oliver-sanders (approved):
Need to come back to this after the beta release.

PR #4156, opened 2021-03-31
setup: remove packaging dependency

oliver-sanders:
We have acquired a dependency on packaging for version parsing. This can be done using setuptools which we are already using for this purpose:

https://github.com/cylc/cylc-flow/blob/8b25ae063cfd3273447ec2339efcb83417d721e2/cylc/flow/network/scan.py#L267

* Switched from `packaging` to `setuptools`.
* Added `setuptools` as a dependency, not sure if this is necessary but it's more explicit.

**Requirements check-list**
- [x] https://github.com/conda-forge/cylc-flow-feedstock/pull/23

### DIFF
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -17,6 +17,7 @@ dependencies:
   - python
   - pyuv >=1.4.0,<1.5.0
   - pyzmq >=18.1.0,<18.2.0
+  - setuptools >=49
   - urwid >=2.0,<2.1
   #- rx >=1.6,<2 # TODO: https://github.com/conda-forge/cylc-feedstock/pull/3#issuecomment-660716268
 # optional dependencies
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -10,13 +10,13 @@ dependencies:
   - cryptography
   - graphene >=2.1,<3
   - jinja2 ==2.11.0,<2.12
-  - metomi-isodatetime ==1!2.0.1
+  - metomi-isodatetime >=1!2.0.2, <1!2.1.0
   - protobuf >=3.15.0,<3.16.0
   - psutil >=5.6.0
   - pyasn1
   - python
   - pyuv >=1.4.0,<1.5.0
-  - pyzmq >=18.1.0,<18.2.0
+  - pyzmq >=19.0.0,<19.1.0
   - setuptools >=49
   - urwid >=2.0,<2.1
   #- rx >=1.6,<2 # TODO: https://github.com/conda-forge/cylc-feedstock/pull/3#issuecomment-660716268
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -19,6 +19,7 @@ import os
 from typing import List, Optional, Tuple
 
 from pkg_resources import parse_version
+from pkg_resources.extern.packaging.version import Version
 
 from cylc.flow import LOG
 from cylc.flow import __version__ as CYLC_VERSION
@@ -802,7 +803,7 @@ def get_version_hierarchy(version: str) -> List[str]:
         ['', '8', '8.0', '8.0.1', '8.0.1a2', '8.0.1a2.dev']
 
     """
-    smart_ver = parse_version(version)
+    smart_ver: Version = parse_version(version)
     base = [str(i) for i in smart_ver.release]
     hierarchy = ['']
     hierarchy += ['.'.join(base[:i]) for i in range(1, len(base) + 1)]
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -17,7 +17,8 @@
 
 import os
 from typing import List, Optional, Tuple
-import packaging.version
+
+from pkg_resources import parse_version
 
 from cylc.flow import LOG
 from cylc.flow import __version__ as CYLC_VERSION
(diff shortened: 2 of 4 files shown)


kinow (approved):
We could also use `packaging` instead of `setuptools` I think. Having `setuptools` as build dependency, but `packaging` as runtime…

I haven't confirmed, but it looks like using Setuptools' `pkg_resources` could also be slower than `packaging`? https://github.com/pypa/packaging/issues/368#issuecomment-738505645 

We also cannot really control the version of `setuptools` (I think it won't update itself it we specify the version in setup.py?) but we can control the version of `packaging`. So in case we have issues we can just pin a newer release, or rollback to an older one if necessary.

Approving because it doesn't break anything and we can think about `packaging` vs `setuptools` later before the next release if necessary :+1:

oliver-sanders:
We're using it to parse version strings which is pretty fast and defined by a PEP so should be a stable interface. Since we need setuptools installed in the first place it's an easy dependency to remove.

hjoliver (approved):

hjoliver (commented):
(Some tests failing) Approve in principle though.

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -768,7 +769,7 @@ def get_version_hierarchy(version: str) -> List[str]:
         ['', '8', '8.0', '8.0.1', '8.0.1a2', '8.0.1a2.dev']
 
     """
-    smart_ver = parse_version(version)
+    smart_ver: Version = parse_version(version)
```
oliver-sanders, line 772: MyPy had issues here, had to set the type to keep it happy.

kinow (approved):

kinow:
Looks good to me, it's still draft, but approved as CI passing.

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -17,7 +17,9 @@
 
 import os
 from typing import List, Optional, Tuple
-import packaging.version
+
+from pkg_resources import parse_version
+from pkg_resources.extern.packaging.version import Version
```
MetRonnie, line 22: VSCode can't resolve this and I don't think mypy can either (I tried accessing a non-existent attribute of `smart_ver` and mypy didn't complain. So might as well do this
```suggestion
from typing import Any
```

cylc/flow/cfgspec/globalcfg.py
MetRonnie:
```suggestion
    smart_ver: Any = parse_version(version)
```

cylc/flow/cfgspec/globalcfg.py
oliver-sanders: > VSCode can't resolve this

Eh?

If the import wasn't resolvable Python would crash with traceback:

```console
$ python
Type "help", "copyright", "credits" or "license" for more information.
>>> from pkg_resources.extern.packaging.version import Version
>>> Version
<class 'pkg_resources.extern.packaging.version.Version'> 
```

MetRonnie:
```diff
diff --git a/cylc/flow/cfgspec/globalcfg.py b/cylc/flow/cfgspec/globalcfg.py
index 960e1affd..c11c29db0 100644
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -770,7 +770,7 @@ def get_version_hierarchy(version: str) -> List[str]:
 
     """
     smart_ver: Version = parse_version(version)
-    base = [str(i) for i in smart_ver.release]
+    base = [str(i) for i in smart_ver.asdf]
     hierarchy = ['']
     hierarchy += ['.'.join(base[:i]) for i in range(1, len(base) + 1)]
     if smart_ver.pre:  # alpha/beta (excluding dev) part of version
```
followed by
```console
$ mypy --explicit-package-bases 
Success: no issues found in 182 source files
```
But if I change it to a `packaging.version.Version` then I get as expected
```console
$ mypy --explicit-package-bases 
cylc/flow/cfgspec/globalcfg.py:773: error: "Version" has no attribute "asdf"
Found 1 error in 1 file (checked 182 source files)
```

oliver-sanders: Doesn't that just mean that setuptools have not added type hints to their code yet?

MetRonnie: Had a look at `~/miniconda3/envs/cylc8/lib/python3.7/site-packages/pkg_resources/extern/__init__.py`, not seen this kind of thing before. I'm guessing mypy can't follow this

https://github.com/pypa/setuptools/blob/ee6a5ffd1b3ec686339e00713542a530dbb65bc8/pkg_resources/extern/__init__.py

oliver-sanders: Swapped to this which seems to be the thing it is "vendoring"?

```diff
-from pkg_resources.extern.packaging.version import Version
+from pkg_resources._vendor.packaging.version import Version
```

MetRonnie (approved):

MetRonnie:
Has the last checklist item been done?
> Created an issue at [cylc-flow conda-forge repository](https://github.com/conda-forge/cylc-flow-feedstock) with version changes (if you changed dependencies in `setup.py`, see `recipe/meta.yaml`).
>     * will do after https://github.com/conda-forge/cylc-flow-feedstock/pull/17

oliver-sanders:
Bumped a couple of conda deps which were pinned to older versions, nice to know Cylc works just fine with the Scheduler using pyzmq 19 and the jobs using pyzmq 18.

oliver-sanders:
Rebased.

conda-environment.yml
```diff
@@ -10,13 +10,13 @@ dependencies:
   - cryptography
   - graphene >=2.1,<3
   - jinja2 ==2.11.0,<2.12
-  - metomi-isodatetime ==1!2.0.1
+  - metomi-isodatetime >=1!2.0.2, <1!2.1.0
```
MetRonnie, line 13: 👍

MetRonnie (approved):

oliver-sanders:
Going in with three approvals.

PR #4168, opened 2021-04-12
Improve error handling if exception occurs during shutdown

### LINKED ISSUE #4147: Catching and Logging of Errors on shutdown
Catching and logging of errors needs some investigation.

Step 1:
Add an error raise in `remote_tidy` in `task_remote_mgr.py` 
(Reminder: this is executed scheduler side and so errors should be raised as elsewhere in Cylc)
e.g. 
```
    LOG.debug("This is where the error is being raised")
    from cylc.flow.exceptions import CylcError
    raise CylcError("!!!!!!!!!!!!!This Error Should Be Raised Correctly!!!!!!!!!!!!!")
``` 
Step 2:
Play a workflow with remote task in `--no-detach` mode.
Step3: 
Note the workflow will run to completion without error.

We would expect this error to be raised correctly.

**Screenshots**
image

**Additional context**

Possibly related to the conversion of the main loop to asyncio. 

**Pull requests welcome!**

MetRonnie:
These changes close #4147

Ensure that if an exception occurs during shutdown, it gets logged before exiting.

<del>This is difficult to test for other than manually; see the steps to reproduce in the issue</del> I've added an integration test, but probably worth manually running the example in #4147 too

**Requirements check-list**
- [x] Tests included
- [x] No dependency changes.

### DIFF
--- a/tests/integration/test_scheduler.py
+++ b/tests/integration/test_scheduler.py
@@ -93,9 +93,9 @@ async def test_resume_does_not_release_tasks(one: Scheduler, run: Callable):
 
 
 @pytest.mark.asyncio
-async def test_exception_logging(one: Scheduler, run: Callable):
-    """Test that if a exception occurs during shutdown, it is
-    logged appropriately."""
+async def test_shutdown_CylcError_log(one: Scheduler, run: Callable):
+    """Test that if a CylcError occurs during shutdown, it is
+    logged in one line."""
     schd = one
 
     async def mock_shutdown(*a, **k):
@@ -104,6 +104,26 @@ async def test_exception_logging(one: Scheduler, run: Callable):
 
     log: pytest.LogCaptureFixture
     with pytest.raises(CylcError) as exc:
+        async with run(schd) as log:
+            pass
+    assert str(exc.value) == "Error on shutdown"
+    last_record = log.records[-1]
+    assert last_record.message == "CylcError: Error on shutdown"
+    assert last_record.levelno == logging.ERROR
+
+
+@pytest.mark.asyncio
+async def test_shutdown_general_exception_log(one: Scheduler, run: Callable):
+    """Test that if a non-CylcError occurs during shutdown, it is
+    logged with traceback (but not excessive)."""
+    schd = one
+
+    async def mock_shutdown(*a, **k):
+        raise ValueError("Error on shutdown")
+    setattr(schd, '_shutdown', mock_shutdown)
+
+    log: pytest.LogCaptureFixture
+    with pytest.raises(ValueError) as exc:
         async with run(schd) as log:
             pass
     assert str(exc.value) == "Error on shutdown"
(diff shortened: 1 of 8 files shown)


tests/integration/test_scheduler.py
```diff
@@ -56,18 +57,17 @@ def ctrl_c():
         raise asyncio.CancelledError("Mock keyboard interrupt")
     # Patch this part of the main loop
     _schd_suite_shutdown = schd.suite_shutdown
-    schd.suite_shutdown = ctrl_c
+    setattr(schd, 'suite_shutdown', ctrl_c)
```
MetRonnie, line 60: This plays better with mypy

MetRonnie:
Ignore test failures (due to DNS issue), recommend running fast tests locally

hjoliver (approved):
Looks good, tested as working :+1:

datamel (commented):
This is an example of a platform lookup error put in remote_init, higher up the program….

 
image


This is exactly the same error put in remote_tidy….

 
image


Is there a reason for leaving traceback in? Surely the same error should produce the same result on screen?

MetRonnie:
My reasoning is that an error during the running of the workflow can still end with a complete shutdown, but an error during shutdown likely means the shutdown did not complete.

Also, there's already a fair bit of error handling in the `Scheduler._shutdown()` method, so if it's not caught by that it's probably something serious enough to fail louder?

datamel:
> My reasoning is that an error during the running of the workflow can still end with a complete shutdown, but an error during shutdown likely means the shutdown did not complete.

Ah ok, my reasoning would be if you are raising an error in, for example `remote_tidy`, that is intentional, and, provided the error message is written correctly, provides sufficient information to the user about the problem. 

The reason why I became aware of this bug and raised the issue, was at one stage (before a change implemented in `remote_init` (https://github.com/cylc/cylc-flow/pull/4138)), the error raise in `get_random_platform_for_install_target()` was not bubbling up correctly in `remote_tidy`. In that case, there is no need for traceback, the error is sufficient (otherwise presumably we would put traceback on all our error raises). 

> Also, there's already a fair bit of error handling in the `Scheduler._shutdown()` method, so if it's not caught by that it's probably something serious enough to fail louder?

From my tests, this error handling will not help here if you look at the call stack, `Scheduler._shutdown()` happens before `remote_tidy` is executed.

MetRonnie:
I've made it so that if a `CylcError` occurs during shutdown, it just gets a one line log entry, but if a general exception occurs, it logs traceback (this is how uncaught exceptions during the running of the workflow are logged).

MetRonnie:
I have locally run:
- ✔️ flake8 & mypy
- ✔️ unit + integration + docstring tests
- ✔️ functional + flaky tests

hjoliver (approved):

datamel (approved):
Thanks @MetRonnie!

PR #4173, opened 2021-04-15
state totals fixes

### LINKED ISSUE #4157: cycle point status with workflow pause
Small bug in the GraphQL deltas which causes cycles to show as running when all tasks have finished.

In this example the cycle "8" shows up as running even though `a.8` has succeeded:

ui-fam-state

It is likely that the pause is the issue here as the status corrects itself once unpaused. Perhaps the "root" family state is being updated one iteration after the task state?

**Pull requests welcome!**

dwsutherland:
These changes close #4157
image

Also fixes incorrect state totals (caused by only the deltas, and not the store totals, being collected).
Although I have noticed additional `succeeded` every now and then, which I will hunt down eventually but is only visible in TUI.

image

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -62,6 +62,10 @@ Replace the job "host" field with "platform" in the GraphQL schema.
 Fix a host ⇒ platform upgrade bug where host names were being popped from task
 configs causing subsequent tasks to run on localhost.
 
+[#4173](https://github.com/cylc/cylc-flow/pull/4173)
+Fix the state totals shown in both the UI and TUI, including incorrect counts
+during workflow run and post pause.
+
 -------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


hjoliver (approved):
:+1:

kinow (approved):
Tested the UI, on `master` got the states with different values than the ones displayed in the tree view, and also some weird state for the toolbar.

With this PR, the toolbar behaves correctly, and the states in GScan and in the tree view are matching. Great job @dwsutherland !

kinow:
Got the issue I had seen on `master` again after trying a little more. Everything else seems to work fine, and this issue appears to be related to how we update the UI state after executing a mutation. Leaving it as approved as the code looks OK and the UI is working OK (cannot tell anything is wrong with stateTotals).

Will leave it to @oliver-sanders to finish the review :+1:

oliver-sanders (approved):
🍾

PR #4180, opened 2021-04-20
Bug fixes

MetRonnie:

- Fix bug where installing a `suite.rc` workflow that had been `cylc validate`d in its source dir would result in a `flow.cylc` symlink being copied over still pointing to the source dir `suite.rc` instead of the one in the run dir
- Fix bug where the deprecated `flow.cylc` config item

  ```ini
  [scheduling]
      [[dependencies]]
          graph = foo
  ```
  (skipped `[[[<recurrence>]]]` section) did not log a warning when upgraded to

  ```ini
  [scheduling]
      [[graph]]
          graph = foo  # (same as R1 = foo)
  ```
- Do not show traceback when any `CylcError` gets raised during running of the workflow, only show traceback for generic uncaught exceptions (unless using `--debug`, in which case show traceback for all exceptions)

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -42,9 +42,7 @@ from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
 from cylc.flow.config import SuiteConfig
 from cylc.flow.cycling.loader import get_point
 from cylc.flow.data_store_mgr import DataStoreMgr, parse_job_item
-from cylc.flow.exceptions import (
-    CyclingError, CylcError, SuiteConfigError, PlatformLookupError
-)
+from cylc.flow.exceptions import CyclingError, CylcError
 import cylc.flow.flags
 from cylc.flow.host_select import select_suite_host
 from cylc.flow.hostuserutil import (
@@ -1636,10 +1634,9 @@ class Scheduler:
                 self.resume_workflow(quiet=True)
         elif isinstance(reason, SchedulerError):
             LOG.error(f'Suite shutting down - {reason}')
-        elif isinstance(reason, SuiteConfigError):
-            LOG.error(f'{SuiteConfigError.__name__}: {reason}')
-        elif isinstance(reason, PlatformLookupError):
-            LOG.error(f'{PlatformLookupError.__name__}: {reason}')
+        elif isinstance(reason, CylcError):
+            LOG.error(
+                f"Suite shutting down - {reason.__class__.__name__}: {reason}")
         else:
             LOG.exception(reason)
             if str(reason):
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -616,7 +616,6 @@ class Scheduler:
 
         """
         try:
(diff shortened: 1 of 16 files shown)


cylc/flow/rundb.py
```diff
@@ -167,7 +167,7 @@ class CylcSuiteDAO:
     CONN_TIMEOUT = 0.2
     DB_FILE_BASE_NAME = "db"
     MAX_TRIES = 100
-    RESTART_INCOMPAT_VERSION = "8.0a2"  # Can't restart suite if <= this vers
+    RESTART_INCOMPAT_VERSION = "8.0b0"  # Can't restart suite if <= this vers
```
MetRonnie, line 170: 8.0a2 was never released

cylc/flow/scheduler.py
```diff
@@ -616,7 +614,6 @@ async def run(self):
 
         """
         try:
-            await self.install()
```
MetRonnie, line 619: `Scheduler.install()` has already just been run in `scheduler_cli._setup()` https://github.com/cylc/cylc-flow/blob/dd8e26ac9cbba3c4b4683380456e59c52e000d8d/cylc/flow/scheduler_cli.py#L293-L310

https://github.com/cylc/cylc-flow/blob/dd8e26ac9cbba3c4b4683380456e59c52e000d8d/cylc/flow/scheduler_cli.py#L334-L337

https://github.com/cylc/cylc-flow/blob/dd8e26ac9cbba3c4b4683380456e59c52e000d8d/cylc/flow/scheduler_cli.py#L342-L347

MetRonnie:
Fast test failures seem to be due to https://github.com/pytest-dev/pytest-asyncio/issues/209

cylc/flow/scheduler.py
oliver-sanders: Hmmm, why the heck was it like that.

Would it work to keep the install in the Scheduler and remove it from `scheduler_cli`?

oliver-sanders:
> passing locally for me on pytest-asyncio 0.14.0

Passing locally for me on pytest-asyncio 0.15.0 but failing on GH actions?!

I've pinned us back to 0.14.0 on #4176 until a fix is out

cylc/flow/scheduler.py
MetRonnie: I tried removing `scheduler_cli._setup()` first, but the
```python
 # setup loggers 
 _open_logs(reg, options.no_detach) 
```
bit fails without the workflow run dir tree set up

oliver-sanders: Dammit, nvm.

MetRonnie:
Final test failure should be fixed by #4185

oliver-sanders (approved):

hjoliver:
(Cancelled Actions after merging a trivial CHANGES.md conflict)

hjoliver (approved):
Good :+1:

PR #4184, opened 2021-04-20
Intelligent Host Select.

wxtim:
These changes partially address #3827 
@datamel  should be credited somewhere since we pair programmed the Remote Init and Filinstall logic. 

### Summary of logic:
- Added a new kwarg, `badhosts: set` to `get_host_from_platform`. This is then subtracted from the platforms list of hosts and the selection is carried out on the leftover hosts.
- Added `TaskRemoteMgr.badhosts = set()`, allowing us to store a list of bad hosts.
- Modified several cases of `get_host_from_platform` in `remote.py` to use `badhosts` if they can.
- Used this infrastructure to allow retrying of remote tasks if they fail with a 255.

### Caveats
- I'm leaving platform from platform host as a follow up.
- Current testing is static - i.e. there is no checking what happens if hostA dies, then comes back online then hostB dies. All testing is done with platforms where lists of hosts contain deliberately fictional hosts, where ssh will fail causing a retry.

### WIP Checklist
- [x] Host within Platform
  - [x] Create a persistent object (`TaskRemoteMgr.badhosts`) in which to store hosts which aren't accessible.
  - [x] Created a mainloop plugin clearing `TaskRemoteMgr.badhosts` from time to time.
  - [x] Remote init working.
  - [x] ... and tested
  - [x] Remote fileinstall working.
  - [x] ... and tested
  - [x] Job submission working.
  - [x] ... and tested
  - [x] Remote tidy working.
  - [x] ... and tested
  - [x] Job log retrieval working.
  - [x] ... and tested
  - [x] Polling Working 
  - [x] ... and tested
  - [x] Kill working
  - [x] ...and tested
- [x] Check existing tests work = Now Passing locally.

### Standard Checklist
  - [x] Test workflow included which may become the basis for future functional tests. 
- [x] No dependency changes.


cylc/flow/platforms.py
```diff
@@ -387,14 +389,28 @@ def get_host_from_platform(platform, method='random'):
             - Random Selection with check for host availability
 
     """
-    if method == 'random':
-        return random.choice(platform['hosts'])
-    elif method == 'first':
-        return platform['hosts'][0]
+    from cylc.flow import LOG
+    if badhosts is not None:
```
wxtim, line 393: This shouldn't be `LOG.critical` when merged, but it makes it easier to see what is happening whilst debugging.

wxtim:
I've noticed that the host selection for job submit doesn't seem to be looking at badhosts. I'll have a poke...

cylc/flow/task_job_mgr.py
```diff
@@ -388,10 +426,13 @@ def submit_task_jobs(self, suite, itasks, curve_auth,
...
 
             if remote_mode:
-                cmd = construct_ssh_cmd(cmd, platform)
+                cmd = construct_ssh_cmd(
+                    cmd, platform, self.task_remote_mgr.badhosts
+                )
             else:
                 cmd = ['cylc'] + cmd
 
+            # Does the enum serve any purpose any-more?
```
wxtim, line 435: Largely irrelevant to the task in hand, but can anyone see any purpose in it?

cylc/flow/task_remote_mgr.py
```diff
@@ -77,8 +79,10 @@ def __init__(self, suite, proc_pool):
         # self.remote_init_map = {(install target): status, ...}
         self.remote_init_map = {}
         self.uuid_str = None
+        # Ready for _what_?
```
wxtim, line 82: Happy to remove, valid comment however IMO.

cylc/flow/task_remote_mgr.py
```diff
@@ -323,6 +327,14 @@ def _remote_init_callback(
                 self.remote_init_map[install_target] = REMOTE_INIT_DONE
                 self.ready = True
                 return
+        elif proc_ctx.ret_code == 255:
+            LOG.critical('Remote-init failed with a 255 error code.')
```
wxtim, line 331: Again, useful for testing, but should be downgraded to debug.

tests/functional/platforms/08-intelligent-host-selection/flow.cylc
```diff
@@ -0,0 +1,52 @@
+
+[meta]
```
wxtim, line 2: This should allow you to have a go at testing.
At the moment the total absence of contactable hosts will cause an exception in platforms. I imagine that if this occurs in Job submission we might want to treat that as submit-fail, and trigger retry logic.

cylc/flow/task_job_mgr.py
```diff
@@ -839,7 +887,12 @@ def _submit_task_job_callback(self, suite, itask, cmd_ctx, line):
             if ctx.ret_code:
                 ctx.cmd = cmd_ctx.cmd  # print original command on failure
         log_task_job_activity(ctx, suite, itask.point, itask.tdef.name)
-
+        if cmd_ctx.ret_code == 255:
+            # Submission failed because host was unreachable. Retry NOW.
+            self.task_events_mgr._retry_task(
```
wxtim, line 892: `_retry_task` seems like a rather low level psuedo-private function, but by using it I bypass using the usual submit fail/retry retry logic which I'm not sure that I want to touch.

cylc/flow/task_remote_mgr.py
hjoliver: In `scheduler.py`:
```python
        if self.task_job_mgr.task_remote_mgr.ready:
             # This flag is turned on when a host init/select command completes
             process = True
             self.task_job_mgr.task_remote_mgr.ready = False  # reset
```
But yes, a comment in the remote manager class would be helpful!

wxtim:
Copied from a comment for @oliver-sanders 

Intelligent Host Selection - Job Submission
 
The idea is Cylc tries to submit the job to one host, if that fails 255 it tries the next one, if it runs out of hosts it puts the job into a submit-failed state.
 
This means the job submission system has to remember the hosts it has previously tried. At the moment on Tim's branch this state is centralised which makes sense. This way Cylc doesn't try to use hosts which is knows are down. This is good for efficiency and also prevents flooding the system with SSH commands that are sitting around before they hit their timeout.
 
Problems:
	
We would have to wipe this global state at some point otherwise hosts which have come back will not be considered for any operation.
	If there is a transient network problem all hosts could be added to the blacklist and the workflow would sit around doing nothing until the state is wiped.

Best solution I can come up with is:
	
Reset this state periodically (use cylc.flow.main_loop.periodical)
	AND also reset the state whenever a submit-failed task is retriggered either manually or by a scheduled automatic retry.

cylc/flow/remote.py
```diff
@@ -162,7 +162,7 @@ def get_includes_to_rsync(rsync_includes=None):
 
 
 def construct_rsync_over_ssh_cmd(
-        src_path, dst_path, platform, rsync_includes=None):
+        src_path, dst_path, platform, rsync_includes=None, badhosts=None):
```
wxtim, line 165: I can't find any tests for this method. I guess it should have some, but I think that this is out of scope for this ticket.

cylc/flow/scripts/cat_log.py
```diff
@@ -439,6 +439,7 @@ def main(parser, options, *args, color=False):
                 cmd.append('--remote-arg=%s' % quote(batchview_cmd))
             cmd.append(workflow_name)
             is_edit_mode = (mode == 'edit')
+            # TODO: Add Intelligent Host selection to this
```
wxtim, line 442: Happy to try and implement now, or declare it follow up. What do you think?

cylc/flow/task_proxy.py
```diff
@@ -247,6 +247,15 @@ def __str__(self):
         """Stringify using "self.identity"."""
         return self.identity
 
+    def __repr__(self):
```
wxtim, line 250: Irrelevent. I added this to help me debug. Tell me if you want it removed.

wxtim:
I've tagged multiple people in this review because I am aware that the half term hols are coming up. Feel free to poke about with this.

cylc/flow/scripts/cat_log.py
hjoliver: Follow-up is fine, since this is a peripheral command.

cylc/flow/task_proxy.py
hjoliver: Best to remove if not used anywhere.  (As an aside, I'd use a debugger to see this sort of thing, without the need to add anything to the code).

cylc/flow/task_remote_mgr.py
```diff
@@ -230,18 +260,19 @@ def remote_tidy(self):
...
-                "Removing authentication keys and contact file "
-                f"from remote: \"{install_target}\"")
-            procs[platform_n] = (
-                cmd,
-                Popen(cmd, stdout=PIPE, stderr=PIPE, stdin=DEVNULL))
+            try:
+                cmd = construct_remote_tidy_ssh_cmd(install_target, platform)
+            except (NoHostsError, PlatformLookupError):
+                LOG.warning(TaskRemoteMgmtError(
+                    TaskRemoteMgmtError.MSG_TIDY, 'A', 'FOO', 'BAR', 'BAZ', 'Q'
```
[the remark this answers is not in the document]
wxtim: Sort of: This Error seems appropriate to me, but it requires information not available at this point in the code. I'll have another look.

cylc/flow/task_remote_mgr.py
```diff
@@ -176,9 +180,10 @@ def remote_init(
         install_target = platform['install target']
         if install_target == get_localhost_install_target():
             self.remote_init_map[install_target] = REMOTE_FILE_INSTALL_DONE
-            return
         # Set status of install target to in progress while waiting for remote
         # initialisation to finish
+            return
+
```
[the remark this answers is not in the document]
wxtim: "?" indeed.

cylc/flow/platforms.py
```diff
@@ -38,6 +39,10 @@
 PLATFORM_REC_COMMAND = re.compile(r'(\$\()\s*(.*)\s*([)])$')
 
 
+class NoHostsError(Exception):
+    ...
```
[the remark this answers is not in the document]
wxtim: Nope. MyBadError.

cylc/flow/platforms.py
```diff
@@ -366,18 +371,15 @@ def generic_items_match(
     return True
 
 
-def get_host_from_platform(platform, method='random'):
+def get_host_from_platform(
+    platform, badhosts: set = None
+) -> str:
```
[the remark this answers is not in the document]
wxtim: Ay!

cylc/flow/task_remote_mgr.py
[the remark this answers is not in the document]
wxtim: Fixed locally, will push soon.

cylc/flow/workflow_files.py
```diff
@@ -603,7 +603,9 @@ def _clean_check(reg, run_dir):
             f"Cannot remove running workflow.\n\n{exc}")
 
 
-def init_clean(reg: str, opts: 'Values') -> None:
+def init_clean(
+    reg: str, opts: 'Values', badhosts: Optional[set] = None
```
MetRonnie, line 607:
```suggestion
    reg: str, opts: 'Values', badhosts: Optional[Set[str]] = None
```

cylc/flow/workflow_files.py
```diff
@@ -627,14 +629,13 @@ def init_clean(reg: str, opts: 'Values') -> None:
...
     if platform_names and platform_names != {'localhost'}:
-        remote_clean(reg, platform_names, opts.remote_timeout)
+        remote_clean(reg, platform_names, opts.remote_timeout, badhosts)
     LOG.info("Cleaning on local filesystem")
     clean(reg)
 
 
 def clean(reg):
-    """Remove a stopped workflow from the local filesystem only.
-
```
MetRonnie, line 637: I don't think this should have been deleted

tests/functional/intelligent-host-selection/00-mixedhost/flow.cylc
```diff
@@ -0,0 +1,23 @@
+
+[meta]
+title = "Try out scenarios for intelligent host selection."
+description = """
+Tasks
+- goodhost: a control to check that everything works
+- mixedhost contains some hosts that will and won't fail
+"""
```
MetRonnie, line 8: For small workflows like this, I think it's easier to put them inside the `.t` file using `init_workflow` or whatever it's called. Up to you though

tests/functional/intelligent-host-selection/01-periodic-clear-badhosts/flow.cylc
```diff
@@ -0,0 +1,18 @@
...
+[scheduling]
+    cycling mode = integer
+    initial cycle point = 1
+    [[graph]]
+        R1 = mixedhosttask
+
+[runtime]
+    [[mixedhosttask]]
+        platform = mixedhostplatform
+        script = true
```
MetRonnie, line 18: No newline at EOF, don't know if that matters

tests/functional/intelligent-host-selection/01-periodic-clear-badhosts.t
```diff
@@ -0,0 +1,89 @@
...
+export REQUIRE_PLATFORM='loc:remote fs:indep'
+
+. "$(dirname "$0")/test_header"
+
+#-------------------------------------------------------------------------------
+set_test_number 6
+
+# We don't use the usual ``create_test_global_config`` because we need to pin
+# the result of ``get_random_platform_for_install_target(install_target)`` to
+# mixedhostplatform.
```
MetRonnie, line 30: Curious as to why this doesn't work for `create_test_global_config`

MetRonnie (commented):
21/32 files viewed

cylc/flow/workflow_files.py
```diff
@@ -627,14 +629,13 @@ def init_clean(reg: str, opts: 'Values') -> None:
...
+    """
```
wxtim, line 638:
```suggestion
"""Remove a stopped workflow from the local filesystem only.

```

tests/functional/intelligent-host-selection/00-mixedhost/flow.cylc
wxtim: I think I agree. It makes debugging a 1-stop-shop. Will do. 👍🏼

tests/functional/intelligent-host-selection/01-periodic-clear-badhosts/flow.cylc
wxtim: It should be there, thanks for pointing it out. It doesn't matter in the grand scheme of things, but I'd like what I write to look more professional and this sort of thing helps. 👍🏼 
Thanks for spotting.

cylc/flow/workflow_files.py
```diff
@@ -627,13 +629,13 @@ def init_clean(reg: str, opts: 'Values') -> None:
...
     if platform_names and platform_names != {'localhost'}:
-        remote_clean(reg, platform_names, opts.remote_timeout)
+        remote_clean(reg, platform_names, opts.remote_timeout, badhosts)
     LOG.info("Cleaning on local filesystem")
     clean(reg)
 
 
 def clean(reg):
-    """Remove a stopped workflow from the local filesystem only.
+"""Remove a stopped workflow from the local filesystem only.
```
MetRonnie, line 638:
```suggestion
    """Remove a stopped workflow from the local filesystem only.
```

cylc/flow/remote.py
```diff
@@ -202,24 +202,27 @@ def construct_rsync_over_ssh_cmd(
...
     return _construct_ssh_cmd(
         raw_cmd,
-        host=get_host_from_platform(platform),
+        host=host,
         ssh_cmd=platform['ssh command'],
         remote_cylc_path=platform['cylc path'],
         ssh_login_shell=platform['use login shell'],
         **kwargs
-    )
+    ), host
```
MetRonnie, line 225: I wonder if it would be better to make `host` an arg of `construct_ssh_cmd()`, replacing `badhosts`. Then `construct_ssh_cmd()` wouldn't have to return the host as well as the cmd

tests/functional/intelligent-host-selection/01-periodic-clear-badhosts.t
wxtim: If I remember rightly, `create_test_global_config` appends stuff to your test config (so it will over-write things if they are there). We need to have only one platform with `install target = ${CYLC_TEST_INSTALL_TARGET}`

cylc/flow/remote.py
wxtim: It might well be. I'll have a look. (I think that this is putting things back to how they were before platforms...)

tests/functional/intelligent-host-selection/00-mixedhost/flow.cylc
wxtim: Actually, on second thoughts I have a good reason for _not_ doing this: I think it makes it easier to run the test workflow outside the test which is helpful for debugging.

cylc/flow/remote.py
wxtim: I have done this

tests/functional/intelligent-host-selection/00-mixedhost/flow.cylc
MetRonnie: Fair enough

tests/functional/platforms/08-intelligent-host-selection/flow.cylc
```diff
@@ -0,0 +1,52 @@
...
+
+[scheduler]
+    [[events]]
+        abort on stalled = True
+
+[scheduling]
+    [[dependencies]]
+        # TODO: Task Badhost should submit-fail when it's tried submitting to
+        # all hosts.
+        R1 = mixedhost & goodhost # & badhost:submit-fail => fin
```
MetRonnie, line 39: Not finished?

tests/unit/test_platforms_get_host.py
```diff
@@ -0,0 +1,45 @@
...
+import pytest
+
+from cylc.flow.platforms import get_host_from_platform, NoHostsError
+
+
+TEST_PLATFORM = {
+    'name': 'Elephant',
+    'hosts': ['nellie', 'dumbo', 'jumbo'],
+    'selection': {'method': 'first'}
+},
```
MetRonnie, line 27:
```suggestion
}
```

tests/unit/test_platforms_get_host.py
```diff
@@ -0,0 +1,45 @@
...
+
+@pytest.mark.parametrize(
+    'badhosts, expect',
+    [
+        pytest.param(None, 'nellie'),
+        pytest.param({'nellie', 'dumbo'}, 'jumbo')
+    ]
+)
+def test_get_host_from_platform(badhosts, expect):
+    platform, = TEST_PLATFORM
```
MetRonnie, line 38:
```suggestion
    platform = TEST_PLATFORM
```

tests/unit/test_platforms_get_host.py
```diff
@@ -0,0 +1,45 @@
...
+    assert get_host_from_platform(platform, badhosts) == expect
+
+
+def test_get_host_from_platform_fails():
+    platform, = TEST_PLATFORM
```
MetRonnie, line 43:
```suggestion
    platform = TEST_PLATFORM
```

cylc/flow/workflow_files.py
```diff
@@ -603,7 +604,9 @@ def _clean_check(reg, run_dir):
             f"Cannot remove running workflow.\n\n{exc}")
 
 
-def init_clean(reg: str, opts: 'Values') -> None:
+def init_clean(
+    reg: str, opts: 'Values', badhosts: Optional[Set[str]] = None
+) -> None:
```
MetRonnie, line 609: `init_clean()` is only called from `cylc.flow.scripts.clean`, so I'm not sure where `badhosts` comes into it?

MetRonnie (commented):
33/33 files viewed

tests/functional/lib/bash/test_header
```diff
@@ -375,6 +375,35 @@ __ERR__
     fail "${TEST_NAME}"
 }
 
+named_grep_ok() {
```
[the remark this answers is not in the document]
wxtim: > Can this not be accomplished by appropriately naming the file you're going to grep in?

Not without repeatedly changing the filename as I'm grepping several things from the same file.

tests/functional/platforms/08-intelligent-host-selection/flow.cylc
wxtim: No. Comment needs rm'd.

cylc/flow/workflow_files.py
wxtim: No, I think that you are right. This is effectively a standalone app? I believe that I have now fixed it.

tests/functional/lib/bash/test_header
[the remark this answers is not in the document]
MetRonnie: I suggest that you make `grep_ok` like this:
```shell
grep_ok() {
    named_grep_ok "$(basename "${FILE}")-grep-ok" "$@"
}
```

cylc/flow/workflow_files.py
MetRonnie: I'm not seeing any changes

wxtim: Change made, but not pushed.

cylc/flow/workflow_files.py
```diff
@@ -688,6 +691,7 @@ def remote_clean(reg, platform_names, timeout):
             config, in order to determine the install targets to clean on.
         timeout (str): Number of seconds to wait before cancelling.
     """
+    badhosts = set()
```
MetRonnie, line 694:
```suggestion
```

cylc/flow/workflow_files.py
```diff
@@ -705,7 +709,13 @@ def remote_clean(reg, platform_names, timeout):
...
         # Issue ssh command:
         pool.append(
-            (_remote_clean_cmd(reg, platforms[0], timeout), target, platforms)
+            (
+                _remote_clean_cmd(
+                    reg, platforms[0], timeout, badhosts=badhosts
+                ),
+                target,
+                platforms
+            )
```
MetRonnie, line 718:
```suggestion
            (_remote_clean_cmd(reg, platforms[0], timeout), target, platforms)
```

cylc/flow/workflow_files.py
```diff
@@ -726,10 +736,12 @@ def remote_clean(reg, platform_names, timeout):
...
                 LOG.debug(excn)
                 if platforms:
-                    pool.append(
-                        (_remote_clean_cmd(reg, platforms[0], timeout),
-                         target, platforms)
-                    )
+                    pool.append((
+                        _remote_clean_cmd(
+                            reg, platforms[0], timeout, badhosts=badhosts
+                        ),
```
MetRonnie, line 742:
```suggestion
                        _remote_clean_cmd(reg, platforms[0], timeout),
```

cylc/flow/workflow_files.py
```diff
@@ -740,7 +752,7 @@ def remote_clean(reg, platform_names, timeout):
             f"Could not clean on install targets: {', '.join(failed_targets)}")
 
 
-def _remote_clean_cmd(reg, platform, timeout):
+def _remote_clean_cmd(reg, platform, timeout, badhosts=None):
```
MetRonnie, line 755:
```suggestion
def _remote_clean_cmd(reg, platform, timeout):
```

cylc/flow/workflow_files.py
```diff
@@ -755,11 +767,15 @@ def _remote_clean_cmd(reg, platform, timeout):
         f'Cleaning on install target: {platform["install target"]} '
         f'(using platform: {platform["name"]})'
     )
+    host = get_host_from_platform(
+        platform, badhosts=badhosts
+    )
```
MetRonnie, line 772:
```suggestion
    host = get_host_from_platform(platform)
```

MetRonnie (commented):
I don't think `badhosts` needs to be brought into `clean` at all?

wxtim:
> I don't think `badhosts` needs to be brought into `clean` at all?

If you don't have badhosts what happens if you try to clean a platform:

```
[platforms]
    [[hpc]]
        hosts = foo, bar, baz
```
if foo happens to be offline?

tests/functional/lib/bash/test_header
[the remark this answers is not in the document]
wxtim: I was trying to work out how to tidy this, but apparently it didn't occur to me that grep_ok should call named_grep_ok rather than the other way around.

MetRonnie:
At the moment the `badhosts` set is empty so it isn't doing anything, right? To deal with this I agree we should open a new issue called IHS for cylc clean

tests/unit/test_workflow_files.py
```diff
@@ -479,7 +479,9 @@ def test_remote_clean(install_targets_map, failed_platforms,
     mocked_shuffle = mock.Mock()
     monkeypatch.setattr('cylc.flow.workflow_files.shuffle', mocked_shuffle)
 
-    def mocked_remote_clean_cmd_side_effect(reg, platform, timeout):
+    def mocked_remote_clean_cmd_side_effect(
+        reg, platform, timeout, badhosts=None
+    ):
```
MetRonnie: I missed this
```suggestion
    def mocked_remote_clean_cmd_side_effect(reg, platform, timeout):
```

tests/unit/test_workflow_files.py
```diff
@@ -503,11 +505,13 @@ def mocked_remote_clean_cmd_side_effect(reg, platform, timeout):
...
     else:
-        workflow_files.remote_clean(reg, platform_names, timeout='irrelevant')
+        workflow_files.remote_clean(
+            reg, platform_names, timeout='irrelevant'
+        )
     if expected_platforms:
         for p_name in expected_platforms:
             mocked_remote_clean_cmd.assert_any_call(
-                reg, PLATFORMS[p_name], 'irrelevant')
+                reg, PLATFORMS[p_name], 'irrelevant', badhosts=set())
```
MetRonnie:
```suggestion
                reg, PLATFORMS[p_name], 'irrelevant')
```

MetRonnie (approved):
I'm happy with the code. I have not tested it, but I will be off work next week, so maybe I could ask one of the other reviewers to do testing only (as there are more than 2 reviewers at the moment)?

wxtim:
@MetRonnie  - I have created issue #4263  to capture the requirement to add IHS to Cylc Clean and Cylc Cat-Log

hjoliver:
(I'll review this tomorrow)

MetRonnie (approved):
(Adding my approval of the changes since my last review)

oliver-sanders:
Could you squash those commits down, 100 might just be a record.

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -704,6 +716,18 @@
                 systems so for safety there is an upper limit on the number
                 of job submissions which can be batched together.
             ''')
+            with Conf('selection'):
+                Conf(
+                    'method', VDR.V_STRING, default='random',
+                    options=['random', 'first'],   # TODO, 'cylc-host-select']
+                    desc='''
+                        Set a method by which a host can be selected from a
+                        ``[platforms][<platform>]``. Currently implemented:
```
oliver-sanders, line 725: I think rst bullets need to be padded with newlines:

```suggestion
                        ``[platforms][<platform>]``. Currently implemented:
                        
```

cylc/flow/platforms.py
```diff
@@ -366,35 +371,46 @@ def generic_items_match(
...
+    if not goodhosts:
+        raise NoHostsError(f'Unable to find valid host for {platform["name"]}')
+    else:
+        if method not in methods:
+            raise CylcError(
+                f'method {method} is not a valid input for '
+                'get_host_from_platform'
+            )
+        else:
+            return methods[method](goodhosts)
```
oliver-sanders, line 413: Perhaps replace the `LOG.warning` above (the one with the TODO comment) with a more generic debug message something like:

```suggestion
            LOG.debug(f'host selection platform={platform["name"]} method={method} badhosts={badhosts}')
            return methods[method](goodhosts)
```

cylc/flow/scripts/cat_log.py
oliver-sanders: Can open a some-day issue with a permalink to this TODO or remove the TODO.

cylc/flow/task_pool.py
```diff
@@ -1512,6 +1512,15 @@ def prune_flow_labels(self):
...
 
+    def empty_badhosts(self):
+        """Periodically clear badhosts list.
+        """
+        LOG.info(
+            'Periodic clearing of badhosts which were: '
+            f'{self.task_events_mgr.badhosts}'
+        )
+        self.task_events_mgr.badhosts.clear()
+
```
oliver-sanders, line 1523: Can we move this into the `task_events_mgr`?

oliver-sanders (commented):
Made a first pass at the code, need to get into detail on each of the failure-fallbacks and do some testing.

One thought, there is some scope for centralising some of this into the subprocpool which might simplify things a little by complementing the `callback` argument with something like `ssh_failure_callback`.

```
SPC(
    cmd,
    callback=mycallback,
    ssh_failure_callback=mysshfailurecallback
)
```

That would allow the checking of the 255 errorcode, recording of the badhost and any logging to be done in the same place (at least for things run through the subprocpool).

hjoliver:
My favourite commit log to date :rofl: :

> creeping mypy
>
> @wxtim
> wxtim committed on 21 Apr

cylc/flow/scheduler.py
```diff
@@ -253,6 +253,7 @@ def __init__(self, reg, options):
         self._profile_amounts = {}
         self._profile_update_times = {}
         self.pre_submit_tasks = []
+        self.badhosts: Set[str] = set()
```
hjoliver, line 256:
```suggestion
        self.bad_hosts: Set[str] = set()
```
(we don't usually normally ram two words together like this)

cylc/flow/scripts/check_versions.py
```diff
@@ -88,7 +88,10 @@ def main(_, options, *args):
     versions = {}
     for platform_name in sorted(platforms):
         platform = get_platform(platform_name)
-        cmd = construct_ssh_cmd(['version'], platform)
+        host = get_host_from_platform(
+            platform, badhosts=None
+        )
+        cmd = construct_ssh_cmd(['version'], platform, host)
```
hjoliver, line 94: Inconsistent styling here, for `host = ...` and `cmd = ...`.

I'd leave it all on one line, or go all the way for readability like this:
```suggestion
        host = get_host_from_platform(
            platform,
            badhosts=None
        )
        cmd = construct_ssh_cmd(
            ['version'],
            platform,
            host
        )
```

hjoliver (commented):
A few comments to get my review started (9/32 files viewed). Some of my comments might be a tad picky, sorry :grimacing:

wxtim:
@hjoliver  Re test for construct_rsync_over_ssh_cmd:

> If your PR touches a method that doesn't have any tests yet, it's a good time to add some tests - if that's feasible.

It does now
I'd not spotted that the logic was a one-track thing wrapping some text manip and calls to platforms stuff, but in itself has not branches. So it was easy.

wxtim:
> A few comments to get my review started (9/32 files viewed). Some of my comments might be a tad picky, sorry grimacing

Picky now → Easier customer service job later.

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -704,6 +716,18 @@
...
+                        - random
+                        - first (not recommended, as likely to cause load
+                          imbalances)
```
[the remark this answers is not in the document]
wxtim: At a meeting of the UK the term "definition order" seemed to come out as a winner. I'm going with this unless you object.

wxtim:
@hjoliver @hjoliver Efforts to produce a decent and useful commit history became rather horrific, so I'm  afraid that this became a 1 commit PR, with all the changes in 1 commit. :(

cylc/flow/task_events_mgr.py
```diff
@@ -1293,3 +1317,12 @@ def unset_waiting_event_timer(self, id_key):
...
         self.event_timers_updated = True
+
+    def reset_bad_hosts(self):
+        """Periodically clear bad_hosts list.
+        """
+        LOG.info(
+            'Periodic clearing of bad hosts which were: '
+            f'{self.bad_hosts}'
+        )
+        self.bad_hosts.clear()
```
oliver-sanders, line 1328: We should avoid logging this if the bad hosts is empty, I think this'll do it:

```suggestion
        if self.bad_hosts:
            LOG.info(
                'Periodic clearing of bad hosts which were: '
                f'{self.bad_hosts}'
            )
            self.bad_hosts.clear()
```

oliver-sanders (approved):
Some small code comments.

cylc/flow/platforms.py
```diff
@@ -366,35 +371,49 @@ def generic_items_match(
...
+        goodhosts = list(set(platform['hosts']) - bad_hosts)
+    else:
+        goodhosts = platform['hosts']
+
+    # Get the selection method
+    method = platform['selection']['method']
+    methods = {
+        'definition order': lambda goodhosts: goodhosts[0],
+        'random': random.choice
+    }
```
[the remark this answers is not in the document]
wxtim: The function isn't in a class. Module attribute?

cylc/flow/subprocpool.py
```diff
@@ -191,16 +197,35 @@ def _proc_exit(self, proc, err_xtra, ctx, callback, callback_args):
...
 
     def process(self):
         """Process done child processes and submit more."""
         # Handle child processes that are done
         runnings = []
-        for proc, ctx, callback, callback_args in self.runnings:
+        for running in self.runnings:
+            if len(running) == 5:
+                proc, ctx, bad_hosts, callback, callback_args = running
+                callback_255, callback_255_args = None, None
```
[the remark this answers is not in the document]
wxtim: 🤦🏼‍♂️

cylc/flow/platforms.py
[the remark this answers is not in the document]
oliver-sanders: Ah yeah sorry, a module attribute.

cylc/flow/exceptions.py
```diff
@@ -253,3 +253,10 @@ def __str__(self):
             for key, value in data.items():
                 ret += f'\n        {key}: {value}'
         return ret
+
+
+class NoHostsError(CylcError):
+    """None of the hsts of a given platform were reachable."""
```
hjoliver, line 259:
```suggestion
    """None of the hosts of a given platform were reachable."""
```

cylc/flow/task_events_mgr.py
```diff
@@ -1293,3 +1317,13 @@ def unset_waiting_event_timer(self, id_key):
         """
         self._event_timers[id_key].unset_waiting()
         self.event_timers_updated = True
+
+    def reset_bad_hosts(self):
+        """Periodically clear bad_hosts list.
```
hjoliver, line 1322: The periodic bit is the responsibility of the caller.
```suggestion
        """Clear bad_hosts list.
```

cylc/flow/task_events_mgr.py
```diff
@@ -1293,3 +1317,13 @@ def unset_waiting_event_timer(self, id_key):
...
+        """
+        if self.bad_hosts:
+            LOG.info(
+                'Periodic clearing of bad hosts which were: '
```
hjoliver, line 1326:
```suggestion
                'Clearing bad hosts: '
```

cylc/flow/task_job_mgr.py
```diff
@@ -252,7 +261,41 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
...
+                    break
+                else:
+                    # If there are no hosts left for this platform.
+                    # Set the task state to submit-failed.
+                    itask.waiting_on_job_prep = False
+                    itask.local_job_file_path = None
+                    self._prep_submit_task_job_error(
+                        workflow, itask, '(remote init)', ''
+                    )
+                    self.bad_hosts.difference_update(itask.platform['hosts'])
```
hjoliver, line 285: So if we run out of hosts on a platform, we put all the (platform's) hosts back in contention right away, rather than wait for the next periodic bad_hosts clear-out?  I guess that makes sense.

cylc/flow/task_job_mgr.py
```diff
@@ -252,7 +261,41 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
...
+                    LOG.critical(TaskRemoteMgmtError(
+                        (
+                            f'{itask.platform["name"]}: initialisation did '
+                            f'not complete because it has no reachable hosts.'
+                            f' Clearing {itask.platform["hosts"]} from '
+                            'bad_hosts'
```
hjoliver, line 291: This might be confusing as a message to users. If the platform has no reachable hosts, why are we removing its hosts from `bad_hosts`?  Also, `bad_hosts` is a variable name rather than the concept exposed to users. Perhaps we should omit the 2nd sentence or rephrase it to something like *"Resetting the list of hosts for {name} to try again next time"*?.

hjoliver (commented):
Looks good @wxtim - just a few trivial comments so far. I still need to have a play with the branch, tomorrow.

cylc/flow/task_job_mgr.py
wxtim: The periodic clear-out is supposed to be a fail-safe backup that I'm really hoping is never needed. Doing it this way may become inefficient if a platform becomes seriously flaky, but I think that is a problem for site admins: Our intended use case is a comms blip or a login node being taken temporarily offline.

cylc/flow/task_job_mgr.py
wxtim: I think that ommiting the second sentence makes more sense. Happy to just let the retry logic be part of Cylc's black box. Not knowing about it is probably less confusing for users.

wxtim:
Thank you @hjoliver .

tests/unit/test_subprocpool.py
```diff
@@ -197,3 +203,98 @@ def test_xfunction_attribute_error(self):
...
+def mock_ctx():
+    def inner_(ret_code=None, host=None, cmd_key=None):
+        """Provide a SimpleNamespace which looks like a ctx object.
+        """
+        inputs = locals()
+        defaults = {
+            'ret_code': 255, 'host': 'mouse', 'cmd_key': 'my-command'
+        }
+        for key in inputs:
+            if inputs[key] == None:
```
datamel, line 218: Take or leave the suggestions in this file. They are auto generated from pycodestyle...
```suggestion
            if inputs[key] is None:
```

tests/unit/test_subprocpool.py
```diff
@@ -197,3 +203,98 @@ def test_xfunction_attribute_error(self):
...
+            cmd_key=inputs['cmd_key']
+        )
+        return ctx
+    yield inner_
+
+
+def _test_callback(ctx, foo=''):
+    """Very Simple test callback function"""
+    LOG.error(f'callback called.{foo}')
+
```
datamel, line 233:
```suggestion


```

tests/unit/test_subprocpool.py
```diff
@@ -197,3 +203,98 @@ def test_xfunction_attribute_error(self):
...
+    ctx = mock_ctx(ret_code=ret_code, cmd_key=cmd_key)
+    SubProcPool._run_command_exit(
+        ctx, callback=_test_callback, callback_255=_test_callback_255
+    )
+    assert expect in caplog.records[0].msg
+    if ret_code == 255:
+        assert f'255 callback called.' in caplog.records[1].msg
+
+
+
```
datamel, line 270:
```suggestion
```

datamel (approved):
Minor comments. I've read the code, checked out the branch, run manual tests and the automated ones. I cannot fins any functional problems with this. Thanks @wxtim, nice to have this.

hjoliver (approved):
OK I reckon I can trust the 3 other reviewers' approvals here, so I don't really need to take it further myself tomorrow. Thanks @wxtim another piece of the puzzle in place :tada:

hjoliver:
@wxtim - do you want to just add a change log entry, then merge this?

wxtim:
> @wxtim - do you want to just add a change log entry, then merge this?

I was going to ask...

tests/functional/intelligent-host-selection/01-periodic-clear-badhosts.t
```diff
@@ -0,0 +1,89 @@
...
+job.status
+job.xtrace
+__HERE__
+
+# remote tidy fails definition order time round"
+named_grep_ok "definition order remote tidy fails" \
+    "Tried to tidy remote platform: 'mixedhostplatform' using host 'unreachable_host' but failed; trying a different host" \
+    "${WORKFLOW_RUN_DIR}/log/workflow/log"
+
+purge
```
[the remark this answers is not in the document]
wxtim: My fault - I've now fixed.

wxtim:
We're waiting on Oliver to do a bit more playing. I think bar @datamel  most of these reviews are mostly read-throughs.

hjoliver:
Roger that :+1:  (I just didn't want to hold things up; I have several non-Cylc meetings tomorrow and I'm away next week...).

wxtim:
> Roger that +1 (I just didn't want to hold things up; I have several non-Cylc meetings tomorrow and I'm away next week...).

No worries. I'm pleased that 4 people have now had a read through, and hopeful that there are no major bugs. 🤞🏼

oliver-sanders:
Having a crack at some heavy testing. Ability to diagnose and respond to connectivity issues tested.

Two issues so far:

1) I'm getting spammed by this message multiple times:

   > Cylc knows that these hosts are unreachable:

2) It turns out that `rsync` is too cool to exit `255` for connection problems:

   I'm getting exit code 12 when trying to rsync to a host with an invalid port. Unfortunately this makes a right mess of things.

   Annoyingly I can't see any obvious way to diagnose a connection failure from `rsync`.

   Options:

   1) Bang head against a wall trying to find a consistent measure of "host being down" from `ssh` and `rsync`.
      * Last ditch if desperate.
   2) After `rsync` failure perform a test `ssh` connection for diagnosis.
      * Fine, this situation is rare so it's not a performance hit.
      * Can produce false positives and negatives with highly transitory issues.
   3) Preserve the host that the `remote_init` was performed on and run the `rsync` on that host.
       * Can produce false negatives with highly transitory issues.

wxtim:
>     1. I'm getting spammed by this message multiple times:
>        > Cylc knows that these hosts are unreachable:

You will be spammed by that message if you are trying to break things. Hopefully, in real use it will be less spammy. Happy to change log level if you like.

> It turns out that rsync is too cool to exit 255 for connection problems:

According to [this website](https://rogerhub.com/~r/sysadmin/2016/07/05/rsync-exit-error-12-or-255/) and following experimentation* it seems like there is a sensitive version dependency with rsync. 🤮  It looks like all rsync 255 errors should be treated as host connectivity failures, but that some connection failures may appear as other errors. It's not totally clear which errors (it's probably version dependent) should be treated this way. It also seems likely that many of these errors may be caused by other problems (lack of disk space, for example).

* Tried

- Employers system → mostly 255 if `--rsh=ssh`, 14 if `--rsh=rsh`
- My Ubuntu → 255
- Oldish Mac 1 → 255
- Oldish Mac 2 → 12

In https://github.com/cylc/cylc-flow/pull/4184/commits/4fd936e2c5b3f6a2edca9d6f90ea6db0aeb5c20d I tried 

> ii. After rsync failure perform a test ssh connection for diagnosis.

oliver-sanders:
> You will be spammed by that message if you are trying to break things. Hopefully, in real use it will be less spammy. Happy to change log level if you like.

My test had one task in one cycle trying to run on a single platform with two hosts (one good, one bad). This log message appeared multiple times in the event that the task tried to run on the bad host. I think three or four times in the run of the workflow.

wxtim:
> > You will be spammed by that message if you are trying to break things. Hopefully, in real use it will be less spammy. Happy to change log level if you like.

I've put it down as a debug setting.

oliver-sanders:
> we might not want to lose the info from the logging of a flow run in a less verbose mode

I don't think this message provides us with any new information?

We should log:

* When a new "bad host" is detected.
* When the "bad hosts" set has been cleared.

No need to log the bad hosts with every single host-selection.

wxtim:
> > we might not want to lose the info from the logging of a flow run in a less verbose mode
> 
> I don't think this message provides us with any new information?
> 
> We should log:
> 
>     * When a new "bad host" is detected.
> 
>     * When the "bad hosts" set has been cleared.
> 
> 
> No need to log the bad hosts with every single host-selection.

Sorry, I've fixed this now - I honestly thought that I had already...

cylc/flow/subprocpool.py
```diff
@@ -384,17 +436,112 @@ def _run_command_init(cls, ctx, callback=None, callback_args=None):
...
+        the failures may be caused by other problems.
+        """
+        rsync_255_fail = False
+        if (
+            ctx.cmd[0] == 'rsync'
+            and ctx.ret_code not in [0, 255]
+            and is_remote_host(ctx.host)
+        ):
+            ssh_test = run(
+                ['ssh', ctx.host, 'echo', 'True'], capture_output=True
```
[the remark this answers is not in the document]
oliver-sanders: Slightly more minimal SSH cmd:

```suggestion
                ['ssh', ctx.host, 'true'], capture_output=True
```

oliver-sanders (commented):
Couple of small comments on the rsync fallback:

cylc/flow/subprocpool.py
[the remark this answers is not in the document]
oliver-sanders: We should really be using the platform configured SSH command here, but I'm not sure how easy it is to get that with the information we have?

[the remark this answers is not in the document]
wxtim: > We should really be using the platform configured SSH command here, but I'm not sure how easy it is to get that with the information we have?

Not easy. Can I punt it into another issue?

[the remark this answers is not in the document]
oliver-sanders: Ok, can you open another issue for that, we will have to tackle it soon is it could cause this SSH to fail incorrectly.

oliver-sanders:
Still testing, no new issues.

cylc/flow/scheduler.py
```diff
@@ -253,6 +253,7 @@ def __init__(self, reg, options):
         self._profile_amounts = {}
         self._profile_update_times = {}
         self.pre_submit_tasks = []
+        self.bad_hosts: Set[str] = set()
```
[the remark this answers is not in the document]
wxtim: Done, will push up a response.

cylc/flow/subprocpool.py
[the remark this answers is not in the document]
wxtim: I have a fix, I think. Just updating the tests, then Will push·

wxtim:
Processes touched
- Task init
- Task file install
- Task Poll
- Remote Tidy
- Retrieve Job Logs

cylc/flow/subprocpool.py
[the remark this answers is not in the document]
wxtim: I didn't have a complete fix, it only worked to installation but not log retrieval. Will continue poking it, but reverted the changes so that this branch can be merged.

Pushed into issue #4184

oliver-sanders (commented):
Still testing, made it most the way through ... :sweat:.

There is a potential bug in the `rsync:fail => ssh` logic if `platform` is not defined. I guess we could fallback to a plain `ssh` command although that makes it easy for the platform logic to be silently ignored. Here's a diff anyway:

```diff
diff --git a/cylc/flow/subprocpool.py b/cylc/flow/subprocpool.py
index 9c533dc3a..31e925f41 100644
--- a/cylc/flow/subprocpool.py
+++ b/cylc/flow/subprocpool.py
@@ -542,15 +542,21 @@ class SubProcPool:
         may cause different rsync failures depending on version, and some of
         the failures may be caused by other problems.
         """
-        if platform is not None:
-            ssh_cmd = platform['ssh command']
-
         rsync_255_fail = False
         if (
             ctx.cmd[0] == 'rsync'
             and ctx.ret_code not in [0, 255]
             and is_remote_host(ctx.host)
         ):
+            ssh_cmd = (
+                platform['ssh command']
+                if platform is not None
+                else 'ssh'
+            )
+            ssh_test_cmd = shlex.split(f'{ssh_cmd} {ctx.host} true')
+            LOG.info(
+                f'testing connectivity for {ctx.host} using {ssh_test_cmd}'
+            )
             ssh_test = run(
                 shlex.split(f'{ssh_cmd} {ctx.host} true'),
                 capture_output=True
```

oliver-sanders:
Save the one small issue above and the one small conflict, all good :tada:!

wxtim:
@oliver-sanders I think that I've responded with https://github.com/cylc/cylc-flow/pull/4184/commits/fff1c7bfd499425f8a02e46bbf2983dc5f2b639f.

Sorry about the close/reopen. I've going to restart my VM. It's very laggy. :(

oliver-sanders:
Going in!

PR #4190, opened 2021-04-22
Remove center deprecated tag in README

kinow:

image

In GitHub I think the image will be still center-aligned. I tested on PyCharm with its Markdown rendered, and the result was the same with/without the `center` tag.

image

`<center>` is an HTML deprecated tag, so it could be the reason why it's rendered as plain-text in PYPI. The `div` tag appears to be accepter, although the `align="center"` is ignored. So on PYPI I think the image should be displayed correctly in the next release, but not center-aligned (which is not a big issue I think?).

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-<center>
 <div
   align="center"
 >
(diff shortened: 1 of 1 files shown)


hjoliver (approved):
Good job, I was being annoyed by this only minutes ago!

oliver-sanders (approved):

PR #4192, opened 2021-04-22
8.0b1 -> 8.0b2.dev

hjoliver:
post release version bump

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -47,6 +47,10 @@ creating a new release entry be sure to copy & paste the span tag with the
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
 -------------------------------------------------------------------------------
+## __cylc-8.0b2 (<span actions:bind='release-date'>Released 2021-??-??</span>)__
+
+
+-------------------------------------------------------------------------------
 ## __cylc-8.0b1 (<span actions:bind='release-date'>Released 2021-04-21</span>)__
 
 ### Enhancements
--- a/cylc/flow/__init__.py
+++ b/cylc/flow/__init__.py
@@ -52,4 +52,4 @@ def environ_init():
 
 environ_init()
 
-__version__ = '8.0b1'
+__version__ = '8.0b2.dev'


kinow (approved):
Should we add the new `8.0ab2` changelog too? For future PR's

hjoliver:
Added next change log header.  Cancelled Actions as tests not needed for this.  @kinow - please merge this if you're happy with it.

CHANGES.md
```diff
@@ -46,6 +46,10 @@ creating a new release entry be sure to copy & paste the span tag with the
 `actions:bind` attribute, which is used by a regex to find the text to be
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
+-------------------------------------------------------------------------------
+## __cylc-8.0b2 (<span actions:bind='release-date'>Released 2021-??-??</span>)__
+
+
 -------------------------------------------------------------------------------
 ## __cylc-8.0b1 (<span actions:bind='release-date'>Released 2021-04-21</span>)__
```
kinow, line 54: I removed this in Cylc UI I think, because I think the GH Action looks for this string to replace by the date (although not sure if having it more than once actually breaks it or not). https://github.com/cylc/release-actions/blob/c4814ec1966191dcd359654a0684e1affeb3777d/stage-1/update-changelog-release-date/action.yml#L22

kinow:
> Added next change log header. Cancelled Actions as tests not needed for this. @kinow - please merge this if you're happy with it.

One last question/comment, sorry. Happy to merge after that.

CHANGES.md
hjoliver: I took this at its word :grin: :+1: 
https://github.com/cylc/cylc-flow/pull/4192/files#diff-d975bf659606195d2165918f93e1cf680ef68ea3c9cab994f033705fea8238b2L44-R48

hjoliver: That link doesn't work...

hjoliver:
```markdown
  44 <!-- The topmost release date is automatically updated by GitHub Actions. When
  45 creating a new release entry be sure to copy & paste the span tag with the
  46 `actions:bind` attribute, which is used by a regex to find the text to be
  47 updated. Only the first match gets replaced, so it's fine to leave the old
  48 ones in. -->
```

kinow (approved):

CHANGES.md
kinow: #TodayILearned :grimacing:

hjoliver: He heh, no it was a good question!

PR #4193, opened 2021-04-23
Cylc install update

datamel:
This PR addresses some feedback on `cylc install` from @dpmatthews.

* Logging of symlink making was inaccurate. (added a condition to dict used for logging)
* Tidied the symlinks created for `suite.rc` and `runN` so they are now relative paths 
* Symlink dirs env variables had been changed to being expanded on the scheduler, (reverted that change) 
* Standard `cylc install` now correctly installs from directories with a `.` in the name
* Updated the `cylc install` help to correctly document location of source symlink.
* Localhost symlinks are now skipped if they exist but point to another location, rather that raising a `Workflow Files Error`
* cylc install now logs the named run rather than flow_name

- [x] Already covered by existing tests (adapted).
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/pathutil.py
+++ b/cylc/flow/pathutil.py
@@ -163,7 +163,7 @@ def make_localhost_symlinks(rund, named_sub_dir):
 
 def get_dirs_to_symlink(install_target, flow_name):
     """Returns dictionary of directories to symlink from glbcfg.
-        Note the paths should remain unexpanded, to be expanded on the remote.
+       Note the paths should remain unexpanded, to be expanded on the remote.
     """
     dirs_to_symlink = {}
     symlink_conf = glbl_cfg().get(['symlink dirs'])
--- a/cylc/flow/pathutil.py
+++ b/cylc/flow/pathutil.py
@@ -161,7 +161,9 @@ def make_localhost_symlinks(rund, named_sub_dir):
 
 
 def get_dirs_to_symlink(install_target, flow_name):
-    """Returns dictionary of directories to symlink from glbcfg."""
+    """Returns dictionary of directories to symlink from glbcfg.
+        Note the paths should remain unexpanded, to be expanded on the remote.
+    """
     dirs_to_symlink = {}
     symlink_conf = glbl_cfg().get(['symlink dirs'])
 
@@ -169,14 +171,12 @@ def get_dirs_to_symlink(install_target, flow_name):
         return dirs_to_symlink
     base_dir = symlink_conf[install_target]['run']
     if base_dir is not None:
-        dirs_to_symlink['run'] = os.path.join(
-            expand_path(base_dir), 'cylc-run', flow_name)
+        dirs_to_symlink['run'] = os.path.join(base_dir, 'cylc-run', flow_name)
     for dir_ in ['log', 'share', 'share/cycle', 'work']:
         link = symlink_conf[install_target][dir_]
         if link is None or link == base_dir:
             continue
-        dirs_to_symlink[dir_] = os.path.join(
-            expand_path(link), 'cylc-run', flow_name, dir_)
+        dirs_to_symlink[dir_] = os.path.join(link, 'cylc-run', flow_name, dir_)
     return dirs_to_symlink
 
 
--- a/cylc/flow/pathutil.py
+++ b/cylc/flow/pathutil.py
@@ -153,10 +153,11 @@ def make_localhost_symlinks(rund, named_sub_dir):
                 f'Unable to create symlink to {src}.'
                 f' \'{value}\' contains an invalid environment variable.'
                 ' Please check configuration.')
-        make_symlink(src, dst)
+        symlink_success = make_symlink(src, dst)
         # symlink info returned for logging purposes, symlinks created
         # before logs as this dir may be a symlink.
-        symlinks_created[src] = dst
+        if symlink_success:
+            symlinks_created[src] = dst
     return symlinks_created
 
 
@@ -189,10 +190,12 @@ def make_symlink(src, dst):
     if os.path.exists(dst):
         if os.path.islink(dst) and os.path.samefile(dst, src):
             # correct symlink already exists
-            return
+            return False
         # symlink name is in use by a physical file or directory
-        raise WorkflowFilesError(
-            f"Error when symlinking. The path {dst} already exists.")
+        # log and return
+        LOG.debug(
+            f"Unable to create {src} symlink. The path {dst} already exists.")
+        return False
     elif os.path.islink(dst):
         # remove a bad symlink.
         try:
@@ -204,6 +207,7 @@ def make_symlink(src, dst):
     os.makedirs(os.path.dirname(dst), exist_ok=True)
     try:
         os.symlink(src, dst, target_is_directory=True)
+        return True
     except Exception as exc:
         raise WorkflowFilesError(f"Error when symlinking\n{exc}")
 
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -549,8 +549,6 @@ def register(
            - Nested workflow run directories.
     """
     validate_flow_name(flow_name)
-    import mdb
-    mdb.debug()
     if source is not None:
         if os.path.basename(source) == WorkflowFiles.FLOW_FILE:
             source = os.path.dirname(source)
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -523,14 +523,15 @@ def parse_workflow_arg(options, arg):
 
 
 def register(
-    flow_name: Optional[str] = None, source: Optional[str] = None
+    flow_name: str, source: Optional[str] = None
 ) -> str:
     """Set up workflow.
     This completes some of the set up completed by cylc install.
-    Called only if running workflow that has not been installed.
+    Called only if running a workflow that has not been installed.
 
     Validates workflow name.
     Validates run directory structure.
+    Creates symlinks for localhost symlink dirs.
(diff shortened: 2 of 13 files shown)


tests/functional/cylc-install/00-simple.t
```diff
@@ -59,7 +59,7 @@ __ERR__
 TEST_NAME="${TEST_NAME_BASE}-REG-install-ok"
 run_ok "${TEST_NAME}" cylc install "${RND_SUITE_NAME}"
 contains_ok "${TEST_NAME}.stdout" <<__OUT__
-INSTALLED $RND_SUITE_NAME from ${RND_SUITE_SOURCE} -> ${RND_SUITE_RUNDIR}/run1
+INSTALLED $RND_SUITE_NAME/run1 from ${RND_SUITE_SOURCE} -> ${RND_SUITE_RUNDIR}/run1
```
oliver-sanders, line 62: If the first and last part of this message are always the same (?) could we change it to:

```suggestion
INSTALLED: ${RND_SUITE_SOURCE} -> ${RND_SUITE_RUNDIR}/run1
```

datamel: They are slightly different (probably not different enough to warrant not accepting your suggestion).... first one is the named workflow i.e what users will use to interact... e.g.  `flow_name/run1`
The last one is the directory in which it is installed to, which does indeed end in the named_workflow.
Happy to accept your change if you confirm you are happy with that.

dpmatthews: I think `$RND_SUITE_NAME/run1` is more useful than `${RND_SUITE_RUNDIR}/run1`.
How about:
```
INSTALLED $RND_SUITE_NAME/run1 from ${RND_SUITE_SOURCE}
```
(also, this is shorter!)

tests/unit/test_pathutil.py
```diff
@@ -212,10 +212,13 @@ def test_make_suite_run_tree(
...
     ], ids=["1", "2", "3"])
 def test_get_dirs_to_symlink(
-        suite, install_target, mocked_glbl_cfg, output, mock_glbl_cfg):
+        suite, install_target, mocked_glbl_cfg,
+        output, mock_glbl_cfg, tmp_path):
+    os.environ['DEE'] = f"{tmp_path}"  # Ensures dirs returned are unexpanded
     mock_glbl_cfg('cylc.flow.pathutil.glbl_cfg', mocked_glbl_cfg)
     dirs = get_dirs_to_symlink(install_target, suite)
     assert dirs == output
+    os.environ.pop('DEE')
```
MetRonnie, line 221: If you use `monkeypatch.setenv('DEE', "whatever")`, then there's no need to remove the env var at the end

cylc/flow/suite_files.py
```diff
@@ -1334,10 +1335,10 @@ def unlink_runN(path: Union[Path, str]) -> bool:
 
 def link_runN(latest_run: Union[Path, str]):
     """Create symlink runN, pointing at the latest run"""
-    latest_run = Path(latest_run).expanduser()
+    latest_run = Path(latest_run)
     run_n = Path(latest_run.parent, SuiteFiles.RUN_N)
     try:
-        run_n.symlink_to(latest_run)
+        run_n.symlink_to(latest_run.relative_to(run_n.parent))
```
MetRonnie, line 1341: Why not
```suggestion
        run_n.symlink_to(latest_run.name)
```

cylc/flow/suite_files.py
```diff
@@ -1278,7 +1277,9 @@ def check_flow_file(
         if flow_file_path.is_symlink():
             # Symlink broken or points elsewhere - replace
             flow_file_path.unlink()
-        flow_file_path.symlink_to(suite_rc_path)
+        flow_file_path.symlink_to(
+            suite_rc_path.relative_to(
+                flow_file_path.parent))
```
MetRonnie, line 1282: Why not
```suggestion
        flow_file_path.symlink_to(SuiteFiles.SUITE_RC)
```

tests/unit/test_pathutil.py
```diff
@@ -233,6 +236,7 @@ def test_make_localhost_symlinks_calls_make_symlink_for_each_key_value_dir(
         'share': '$DEE/suite3/share'}
     mocked_get_workflow_run_dir.return_value = "rund"
     mocked_expandvars.return_value = "expanded"
+    mocked_make_symlink.return_value = True
```
MetRonnie, line 239: Is this needed?

datamel: The `make_symlink` function now returns true/false. Not sure if there is better way of doing it but I think I'd expect to be mocking this return?
The test is checking that `make_symlink` is called for all given directories, the return True has no bearing on the test (False would be grand too), other than to stop it blowing up.

MetRonnie: Strange, I would swear the line wouldn't be needed, because `make_localhost_symlinks()` still returns regardless. But the test does indeed fail if I remove that line. However, if I replace the patch decoration with
```python
mocked_make_symlink = Mock()
monkeypatch.setattr('cylc.flow.pathutil.make_symlink', mocked_make_symlink)
```
then it doesn't need a return value attr and the test passes

oliver-sanders:
Happy conflict day!

cylc/flow/suite_files.py
datamel: Thanks for these suggestions @MetRonnie, I can't commit them as had rebased but did manually add them.

cylc/flow/pathutil.py
```diff
@@ -153,30 +153,31 @@ def make_localhost_symlinks(rund, named_sub_dir):
...
-        dirs_to_symlink['run'] = os.path.join(
-            expand_path(base_dir), 'cylc-run', flow_name)
+        dirs_to_symlink['run'] = os.path.join(base_dir, 'cylc-run', flow_name)
     for dir_ in ['log', 'share', 'share/cycle', 'work']:
         link = symlink_conf[install_target][dir_]
         if link is None or link == base_dir:
             continue
-        dirs_to_symlink[dir_] = os.path.join(
-            expand_path(link), 'cylc-run', flow_name, dir_)
+        dirs_to_symlink[dir_] = os.path.join(link, 'cylc-run', flow_name, dir_)
```
MetRonnie, line 180: Does this work for named/numbered run dirs?

cylc/flow/workflow_files.py
```diff
@@ -1143,11 +1142,11 @@ def install_workflow(
             " name, using the --run-name option.")
     check_nested_run_dirs(rundir, flow_name)
     symlinks_created = {}
+    named_run = flow_name
+    if run_num:
+        named_run = os.path.join(named_run, f'run{run_num}')
```
MetRonnie, line 1147: What about the `run_name` arg?

cylc/flow/pathutil.py
datamel: The `flow_name` should be the full named run.

cylc/flow/workflow_files.py
```diff
@@ -547,8 +548,6 @@ def register(
            - Illegal name (can look like a relative path, but not absolute).
            - Nested workflow run directories.
     """
-    if flow_name is None:
-        flow_name = Path.cwd().stem
```
MetRonnie, line 551: Suggest removing this in the docstring for the function
```diff
diff --git a/cylc/flow/workflow_files.py b/cylc/flow/workflow_files.py
index e4c67fa7f..ed3fc8b19 100644
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -536,7 +536,7 @@ def register(
     Creates the .service directory.
 
     Args:
-        flow_name: workflow name, default basename($PWD).
+        flow_name: workflow name.
         source: directory location of flow.cylc file, default $PWD.
 
     Return:
```

cylc/flow/pathutil.py
MetRonnie: This is what I was concerned about, in the `register()` function: https://github.com/cylc/cylc-flow/blob/fa248cd3d21106a4b9792020718e4991390da93d/cylc/flow/workflow_files.py#L560-L561

datamel: I've traced through this and as far as I can see, the register function has to be called with the full-named run. e.g. `myflow/run1`, I think the variable name `flow_name` perhaps would be better named something like `named_run` for clarity. So, I think the code works as expected in this function, I may be missing something though?

cylc/flow/workflow_files.py
datamel: Thanks, I think I have that covered now.

cylc/flow/pathutil.py
MetRonnie: Yes, I think it should be renamed. However, `named_run` would be `run1` or whatever. Really, it's the run dir, but as a path relative to `~/cylc-run`. So maybe `rel_run_dir`?

cylc/flow/workflow_files.py
MetRonnie: If renaming as in https://github.com/cylc/cylc-flow/pull/4193/files#r631789343, I think `named_run` should be called `rel_run_dir` too

datamel: As discussed in today's team meet, saving renaming of variables for the universal id issue/pr, so whole team can decide on agreed universal terminology.

cylc/flow/pathutil.py
datamel: As, below, leaving the renaming for universal id work.

MetRonnie (approved):

oliver-sanders (approved):
Code looks good, my tests ran fine.

PR #4195, opened 2021-04-23
actions: test manylinux compatibility

oliver-sanders:
Ensure our dependency tree remains `manylinux1` compatible for installation on older OSes.

> **TLDR;**
>
> The `manylinux<v>` identifiers define a standardish "base" environment for Linux binaries to be compiled
> against. The `manylinux1` identifier is based off of centos5 which is beyond EOL, consequently manylinux1
> is also EOL but still fairly well supported. The purpose of this test is to scan dependency changes to ensure
> we don't end up with something which can't be built but which doesn't provide a manylinux1 binary.

[Example run](https://github.com/oliver-sanders/cylc-flow/actions/runs/777370753)

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/.github/workflows/test_manylinux.yml
+++ b/.github/workflows/test_manylinux.yml
@@ -1,0 +1,75 @@
+name: manylinux-compatibility
+
+# Ensure that cylc-flow remains installable with manylinx1 binaries for use
+# with older systems.
+
+# Ideally we would use the Docker images from PyPa
+# (https://github.com/pypa/manylinux).
+# The manylinux1 image is based on CentOS=5, glibc=2.5), however, this is so
+# old that installing contemporary Python on them is quite hard:
+# * native package manager has no package for CentOS5
+# * miniforge can't work with glibc < 2.12
+# * micromamba can't work with glibc < 2.6
+# * need zlib to compile Python / install pip but the package manager can't
+#   download it because the package stream is offline (past EOL)
+
+# So this will have to do. This isn't a perfect test as non manylinux compat
+# packages may attempt to compile from source. This might succeed here but not
+# on an older host.
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'setup.py'
+      - 'setup.cfg'
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        manylinux: ['1']
+        os: ['ubuntu-16.04']  # run on the oldest linux we have access to
+        python-version: ['3.7', '3.8', '3.9']
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


.github/workflows/test_manylinux.yml
```diff
@@ -0,0 +1,74 @@
...
+        manylinux: ['1']
+        os: ['ubuntu-16.04']  # run on the oldest linux we have access to
+        python-version: ['3.7', '3.8', '3.9']
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Configure Python
+        uses: actions/setup-python@v1
```
MetRonnie, line 42:
```suggestion
        uses: actions/setup-python@v2
```

.github/workflows/test_manylinux.yml
```diff
@@ -0,0 +1,74 @@
...
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'setup.py'
+      - 'setup.cfg'
+
+jobs:
+  test:
```
MetRonnie, line 28: Worth having a timeout on the whole job in case something else gets stuck. 
```suggestion
  test:
    timeout-minutes: 35
```

.github/workflows/test_manylinux.yml
```diff
@@ -0,0 +1,74 @@
...
+          manylinux1_compatible = False
+          manylinux2010_compatible = False
+          manylinux2014_compatible = False
+          __HERE__
+          echo \
+            "manylinux${{ matrix.manylinux }}_compatible = True" \
+            >> _manylinux.py
+
+      - name: Install
+        timeout-minutes: 15
```
MetRonnie, line 61: In fact, why have timeouts on these steps? Won't a job-level timeout do?

oliver-sanders: Because these are the only two steps which can hang. It's marginally better this way as it prevents `pip install` hanging for the entire job timeout.

MetRonnie: I see (but I think the checkout and possibly setup-python steps could also hang, though not likely)

oliver-sanders:
Ok, covered the action in timeouts incase a git checkout hangs.

MetRonnie (approved):

.github/workflows/test_manylinux.yml
```diff
@@ -0,0 +1,75 @@
...
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        manylinux: ['1']
+        os: ['ubuntu-16.04']  # run on the oldest linux we have access to
```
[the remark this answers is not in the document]
oliver-sanders: Yep, will have to come up with something else by then.

hjoliver (approved):

PR #4196, opened 2021-04-27
fix type hinting for cylc-rose options

wxtim:

Fixes: Cylc-Rose uses types from `cylc.flow.option_parsers.Options`, although these may not always be present as they only get created if Cylc-Rose is installed. MyPy needs help to see what they should be.

Question: Should the type be `Iterable`, rather than list?

**Requirements check-list**
- [x] Already covered by existing tests (MyPy Checks)
- [x] No documentation update required (internal change).
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/option_parsers.py
+++ b/cylc/flow/option_parsers.py
@@ -315,6 +315,9 @@ class Options:
         False
 
     """
+    opt_conf_keys: list
+    defines: list
+    define_suites: list
 
     def __init__(
         self, parser: OptionParser, overrides: Optional[Dict[str, Any]] = None


MetRonnie:
Actually I added that in error, if you're referring to `cylc/rose/utilities.py`. It should be
```python
diff --git a/cylc/rose/utilities.py b/cylc/rose/utilities.py
index aed4269..437bd8d 100644
--- a/cylc/rose/utilities.py
+++ b/cylc/rose/utilities.py
@@ -34,7 +34,7 @@ from metomi.rose.env import env_var_process, UnboundEnvironmentVariableError
 from metomi.rose.resource import ResourceLocator
 
 if TYPE_CHECKING:
-    from cylc.flow.option_parsers import Options
+    from optparse import Values
 
 
 class MultipleTemplatingEnginesError(Exception):
@@ -150,7 +150,7 @@ def get_rose_vars_from_config_node(config, config_node, environ):
 
 
 def rose_config_exists(
-    srcdir: Union[Path, str, None], opts: 'Options'
+    srcdir: Union[Path, str, None], opts: 'Values'
 ) -> bool:
     """Do opts or srcdir contain a rose config?
```

wxtim:
Understand why this isn't necessary.

MetRonnie:
Sorry for forgetting to fix it, will put up a PR on cylc-rose

PR #4174, opened 2021-04-15
Terminology: suite to workflow

hjoliver:
This is a ~~small~~ *large-but-trivial* change with no associated Issue (but discussed with the team!)

- globally replace suite/Suite/SUITE with workflow/Workflow/WORKFLOW
  - thank @@#$$@# for easy command-line automation
- rename all `*suite*` filenames as `*workflow*`
- restore any back-compat references to `suite.rc`
- fix style violations resulting from slightly longer lines
- restore `CYLC_SUITE_*` job environment variables for back-compat

Note:
- much easier to do this all at once than separate user-facing and internal changes
- some use of the word *suite* should really be translated to *scheduler* rather than *workflow*, but I haven't done that (requires a careful look at each case)

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Already covered by existing tests.
- [x] Appropriate change log entry included.
- [x] documentation PR: cylc/cylc-doc/issues/238
- [x] No dependency changes.

PR #4198, opened 2021-04-27
Terminology: suite to workflow follow-up

MetRonnie:
This is a small change with no associated Issue. Follow-up to #4174 

From Teams discussion:

> define flow is very confusing (since it doesn't even do anything to the flow).
>
> So we gotta change now it anyhoo, how about `-S, --rose-template-variable`, that describes what it does without mentioning either suites or flows.
>
> I think for Rose we can just scale back on the word "suite" and try to talk more about "configurations" or the file itself.

**Requirements check-list**
- [x] Does not need tests
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/etc/job.sh
+++ b/cylc/flow/etc/job.sh
@@ -102,12 +102,16 @@ cylc__job__main() {
         CYLC_WORKFLOW_HOST="$(sed -n 's/^CYLC_WORKFLOW_HOST=//p' "${contact}")"
         CYLC_WORKFLOW_OWNER="$(sed -n 's/^CYLC_WORKFLOW_OWNER=//p' "${contact}")"
         export CYLC_WORKFLOW_HOST CYLC_WORKFLOW_OWNER
-        # Deprecated
+        # BACK COMPAT: DEPRECATED environment variables
+        # remove at:
+        #     Cylc9
         export CYLC_SUITE_HOST="${CYLC_WORKFLOW_HOST}"
         export CYLC_SUITE_OWNER="${CYLC_WORKFLOW_OWNER}"
     fi
 
-    # DEPRECATED environment variables
+    # BACK COMPAT: DEPRECATED environment variables
+    # remove at:
+    #     Cylc9
     export CYLC_SUITE_SHARE_DIR="${CYLC_WORKFLOW_SHARE_DIR}"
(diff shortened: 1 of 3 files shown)


cylc/flow/scripts/install.py
```diff
@@ -154,14 +154,14 @@ def get_option_parser():
             dest="defines"
         )
         parser.add_option(
-            "--define-workflow", "--define-flow", '-S',
+            "--rose-template-variable", '-S',
             help=(
                 "As `--define`, but with an implicit `[SECTION]` for "
                 "workflow variables."
```
MetRonnie, line 160: Feels like we should change the description, and possibly the name of the `--define` option?

cylc/flow/etc/job.sh
```diff
@@ -102,12 +102,16 @@ cylc__job__main() {
         CYLC_WORKFLOW_HOST="$(sed -n 's/^CYLC_WORKFLOW_HOST=//p' "${contact}")"
         CYLC_WORKFLOW_OWNER="$(sed -n 's/^CYLC_WORKFLOW_OWNER=//p' "${contact}")"
         export CYLC_WORKFLOW_HOST CYLC_WORKFLOW_OWNER
-        # Deprecated
+        # BACK COMPAT: DEPRECATED environment variables
+        # remove at:
+        #     Cylc9
```
oliver-sanders, line 107: 👍

oliver-sanders (approved):

cylc/flow/scripts/install.py
wxtim: I suppose `--rose-define` would be acceptable. Not worried.

wxtim (commented):
I'm surprised that this doesn't break the Cylc-Rose tests...or does it?

oliver-sanders:
> I'm surprised that this doesn't break the Cylc-Rose tests...or does it?

Yes it does, however, they are not run in this repo.

MetRonnie:
Corresponding changes in cylc-rose will be included in https://github.com/cylc/cylc-rose/pull/56

wxtim:
> Corresponding changes in cylc-rose will be included in [cylc/cylc-rose#56](https://github.com/cylc/cylc-rose/pull/56)

Will leave you to merge both as near simultaneously as poss once both approved.

PR #2622, opened 2018-04-12
job.sh: replace HERE document with echo commands

matthewrmshin:
Bash may put HERE documents in a temporary file, which is not efficient.
On older versions of Bash, it may even create temporary files under
`/tmp` instead of using `mktemp`, which can be problematic on systems
with only a small amount of space in `/tmp`.

Problem first raised by @steoxley.

PR #4201, opened 2021-04-29
Job script: remove the use of here-string's

dpmatthews:
We've previously removed the use of here-document's (#2622).
However, here-string's suffer from the same problem (using temporary files):
https://askubuntu.com/questions/678915/whats-the-difference-between-and-in-bash
These are a particular problem in bash 3.2 because it ignores $TMPDIR:
https://lists.gnu.org/archive/html/bug-bash/2013-12/msg00056.html

To remove the need for here-string's I've replaced the use of `cut` with bash parameter expansion:
https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

I've also removed the use of separate `export` statements except where needed to avoid failing https://github.com/koalaman/shellcheck/wiki/SC2155.

**Requirements check-list**
- [X] Applied any dependency changes to both `setup.py` and
  `conda-environment.yml`.

- [X] Already covered by existing tests.

- [X] No change log entry required (why? e.g. invisible to users).

### DIFF
--- a/cylc/flow/etc/job.sh
+++ b/cylc/flow/etc/job.sh
@@ -73,17 +73,16 @@ cylc__job__main() {
     export CYLC_WORKFLOW_LOG_DIR="${CYLC_WORKFLOW_RUN_DIR}/log/workflow"
     export CYLC_WORKFLOW_SHARE_DIR="${CYLC_WORKFLOW_RUN_DIR}/share"
     export CYLC_WORKFLOW_WORK_DIR="${CYLC_WORKFLOW_RUN_DIR}/work"
-    CYLC_TASK_CYCLE_POINT="$(cut -d '/' -f 1 <<<"${CYLC_TASK_JOB}")"
-    CYLC_TASK_NAME="$(cut -d '/' -f 2 <<<"${CYLC_TASK_JOB}")"
-    export CYLC_TASK_NAME CYLC_TASK_CYCLE_POINT ISODATETIMEREF
+    export CYLC_TASK_CYCLE_POINT="${CYLC_TASK_JOB%%/*}"
+    export CYLC_TASK_NAME="${CYLC_TASK_JOB#*/}"
+    CYLC_TASK_NAME="${CYLC_TASK_NAME%/*}"
     if [[ "${CYLC_CYCLING_MODE}" != "integer" ]]; then  # i.e. date-time cycling
-        ISODATETIMECALENDAR="${CYLC_CYCLING_MODE}"
-        ISODATETIMEREF="${CYLC_TASK_CYCLE_POINT}"
-        export ISODATETIMECALENDAR ISODATETIMEREF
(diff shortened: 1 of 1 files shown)


cylc/flow/etc/job.sh
```diff
@@ -73,17 +73,16 @@ cylc__job__main() {
     export CYLC_WORKFLOW_LOG_DIR="${CYLC_WORKFLOW_RUN_DIR}/log/workflow"
     export CYLC_WORKFLOW_SHARE_DIR="${CYLC_WORKFLOW_RUN_DIR}/share"
     export CYLC_WORKFLOW_WORK_DIR="${CYLC_WORKFLOW_RUN_DIR}/work"
-    CYLC_TASK_CYCLE_POINT="$(cut -d '/' -f 1 <<<"${CYLC_TASK_JOB}")"
-    CYLC_TASK_NAME="$(cut -d '/' -f 2 <<<"${CYLC_TASK_JOB}")"
-    export CYLC_TASK_NAME CYLC_TASK_CYCLE_POINT ISODATETIMEREF
```
dpmatthews, line 78: This is a minor change. `ISODATETIMEREF` will no longer be exported unless it is set (i.e. it won't be set for integer cycling). I've assumed this isn't a problem.

dpmatthews:
At the moment I think we set the minimum possible in the generated script - we don't set anything that can be derived from other settings. However, I agree it might be simpler to set them.

oliver-sanders (approved):
Tested, happy the vars are coming through right.

Will come back to where these vars are set later.

cylc/flow/etc/job.sh
```diff
@@ -73,17 +73,16 @@ cylc__job__main() {
...
-        ISODATETIMECALENDAR="${CYLC_CYCLING_MODE}"
-        ISODATETIMEREF="${CYLC_TASK_CYCLE_POINT}"
-        export ISODATETIMECALENDAR ISODATETIMEREF
+        export ISODATETIMECALENDAR="${CYLC_CYCLING_MODE}"
+        export ISODATETIMEREF="${CYLC_TASK_CYCLE_POINT}"
     fi
     # The "10#" part ensures that the submit number is interpreted in base 10.
     # Otherwise, a zero padded number will be interpreted as an octal.
-    export CYLC_TASK_SUBMIT_NUMBER=$((10#$(cut -d '/' -f 3 <<<"${CYLC_TASK_JOB}")))
+    export CYLC_TASK_SUBMIT_NUMBER="$((10#${CYLC_TASK_JOB##*/}))"
```
hjoliver, line 85: Huh, I wonder why we weren't just using bash variable substitution in the first place?  (`git blame` pings @oliver-sanders for the first two cases above, in mid 2019, but those were probably adapted from the last case, Matt two years earlier).

hjoliver (approved):

PR #4205, opened 2021-05-06
actions: check shortlog

oliver-sanders:
Add an action which checks that all commits are identified with committers in the contributors log via configured aliases in the `.mailmap` file whenever the `CHANGELOG.md` file is edited.

We could potentially run this for every PR, however, it requires checking out the entire repo[1].

Fixes some minor discrepancies.

> [1]: There is a potential halfway house of checking out the last N commits then testing that those commuters exist in the contributing file rather than testing exact equality but this is starting to sound like more bother than it's worth.

**Requirements check-list**
  `conda-environment.yml`.

### DIFF
--- a/.github/workflows/1_create_release_pr.yml
+++ b/.github/workflows/1_create_release_pr.yml
@@ -27,7 +27,7 @@ jobs:
           fetch-depth: 0  # need to fetch all commits to check contributors
 
       - name: Check CONTRIBUTING.md
-        uses: cylc/release-actions/check-shortlog@c8205dcfb0df1661b99e2c240b75787c4fa02aa3
+        uses: cylc/release-actions/check-shortlog@v1
 
       - name: Setup Python
         uses: actions/setup-python@v2
--- a/.github/workflows/1_create_release_pr.yml
+++ b/.github/workflows/1_create_release_pr.yml
@@ -17,38 +17,42 @@ jobs:
     timeout-minutes: 10
     steps:
 
-    - name: Sanitise workflow inputs # Should be 1st step
-      uses: cylc/release-actions/stage-1/sanitize-inputs@v1
+      - name: Sanitise workflow inputs # Should be 1st step
+        uses: cylc/release-actions/stage-1/sanitize-inputs@v1
 
-    - name: Checkout repo
-      uses: actions/checkout@v2
-      with:
-        ref: ${{ env.BASE_REF }}
+      - name: Checkout repo
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ env.BASE_REF }}
+          fetch-depth: 0  # need to fetch all commits to check contributors
 
-    - name: Setup Python
-      uses: actions/setup-python@v2
-      with:
-        python-version: '3.x'
+      - name: Check CONTRIBUTING.md
+        uses: cylc/release-actions/check-shortlog@c8205dcfb0df1661b99e2c240b75787c4fa02aa3
 
-    - name: Create & checkout PR branch
-      uses: cylc/release-actions/stage-1/checkout-pr-branch@v1
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
 
-    - name: Set the package version
-      uses: cylc/release-actions/stage-1/set-python-package-version@v1
-      with:
-        init-file: 'cylc/flow/__init__.py'
-        pypi-package-name: 'cylc-flow'
+      - name: Create & checkout PR branch
+        uses: cylc/release-actions/stage-1/checkout-pr-branch@v1
 
-    - name: Update "released on" date in changelog
-      continue-on-error: true
-      uses: cylc/release-actions/stage-1/update-changelog-release-date@v1
-      with:
-        changelog-file: 'CHANGES.md'
+      - name: Set the package version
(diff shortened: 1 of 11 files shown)


.github/workflows/shortlog.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: 'ubuntu-latest'
+    timeout-minutes: 10
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0  # need to fetch all commits to check contributors
+
+      - name: Check CONTRIBUTING.md
+        uses: oliver-sanders/release-actions/check-shortlog@ed0d896d4e7b304e3b4f3319eb047164bbc3927e
```
oliver-sanders, line 23: Draft until https://github.com/cylc/release-actions/pull/13 is merged and we can switch to cylc/release-actions.

oliver-sanders:
May have to use `${{ github.sha }}` rather than `${{ github.ref }}`, not sure.

CONTRIBUTING.md
```diff
@@ -72,6 +73,7 @@ requests_).
  - Ronnie Dutta
  - John Haiducek
  - (Andrew Huang)
```
MetRonnie, line 75: CONTRIBUTING.md vs commit authors:
```diff
0a1
> Alex Reinecke
29,30d29
< Patrick Alexander Reinecke
< Patrick Reinecke
41d39
< hilary.oliver
```

Might we need a JSON file or something that maps aliases of people to one name?

oliver-sanders: That's what the `.mailmap` file does.

.github/workflows/shortlog.yml
```diff
@@ -0,0 +1,23 @@
+name: check contributor list
+
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
+    paths-ignore:
+      - '**'
+      - '!CONTRIBUTING.md'  # e.g. when new contributor adds themselves
+      - '!cylc/flow/__init__.py'  # e.g. when __version__ is changed for a new release
```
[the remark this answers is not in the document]
oliver-sanders: Dammit.

CONTRIBUTING.md
oliver-sanders: Already done on this PR ☕

oliver-sanders:
`github.sha` seems to work fine, just tweeking the mailmap.

oliver-sanders:
Switched to running from cylc/release-actions.

.github/workflows/shortlog.yml
oliver-sanders: Done.

oliver-sanders:
Added contributor check to the stage 1 workflow.

.github/workflows/1_create_release_pr.yml
```diff
@@ -17,38 +17,42 @@ jobs:
...
+        uses: cylc/release-actions/stage-1/sanitize-inputs@v1
+
+      - name: Checkout repo
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ env.BASE_REF }}
+          fetch-depth: 0  # need to fetch all commits to check contributors
+
+      - name: Check CONTRIBUTING.md
+        uses: cylc/release-actions/check-shortlog@c8205dcfb0df1661b99e2c240b75787c4fa02aa3
```
MetRonnie, line 30:
```suggestion
        uses: cylc/release-actions/check-shortlog@v1
```
I'll bump the v1 tag on there

.github/workflows/shortlog.yml
```diff
@@ -0,0 +1,23 @@
+name: check contributor list
```
MetRonnie, line 1: Do we still need this? Perhaps it would be better to leave it to the release workflow so we have less stuff to maintain?

oliver-sanders: Better the catch it early than late.

We want contributors to add themselves to the list (and mailmap as necessary) as this infers agreement to the CLA (and the mailmap associates the commit with the contributor list). By release time it's a little too late but better to find out then than never.

.github/workflows/1_create_release_pr.yml
oliver-sanders: Happy using SHA, if we are going to bump the tag might as well create one called "latest".

MetRonnie: The idea is we keep bumping v1 (actually it's a branch not a tag) until we introduce a backwards incompatible change, then we start v2 so we don't break existing workflows across the repos

.github/workflows/shortlog.yml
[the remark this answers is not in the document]
MetRonnie: Should it run for any path, because someone might forget to add themselves to CONTRIBUTING.md?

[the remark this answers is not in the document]
oliver-sanders: Yes, however, it would require us checking out the entire repo (all objects) for each change which is a bit ORR.

MetRonnie, line 10: Following on from https://github.com/cylc/cylc-flow/pull/4205/files#r631898770, I think this could be simplified to
```suggestion
  workflow_dispatch:
  pull_request:
    paths:
      - 'CONTRIBUTING.md'
```

MetRonnie (approved):

PR #4209, opened 2021-05-11
Wrapper script: add global config environment variables & support rose

dpmatthews:
The `cylc` wrapper script only requires minor changes in order to be used as the `rose` wrapper script as well.
For the moment, an extra bit of legacy logic has to used used to support standalone Rose installations but that can be removed once no longer using Cylc 7 / Rose 2.

Note: the alternative is that we maintain a separate `rose` wrapper (identical to the `cylc` wrapper other than the legacy logic. and the setting of `ROSE_SITE_CONF_PATH`) and then advise anyone using Rose to use the `rose` wrapper  in place of the `cylc` one.
Note that `ROSE_SITE_CONF_PATH` has to be set even when calling `cylc` in order for `rose` to work if called by the scheduler (e.g. legacy workflows using `rose host select`).

**Requirements check-list**
- [N/A] Applied any dependency changes to both `setup.py` and
  `conda-environment.yml`.
- [X] Does not need tests (why?).
- [X] No change log entry required (why? e.g. invisible to users).

### DIFF
--- a/usr/bin/cylc
+++ b/usr/bin/cylc
@@ -17,22 +17,24 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 #------------------------------------------------------------------------------
-# Wrapper script to support multiple installed Cylc versions. Handles Conda and
-# Python virtual environments, and legacy plain Cylc 7 installations.
+# Wrapper script to support multiple installed Cylc & Rose versions. Handles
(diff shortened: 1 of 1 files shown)


oliver-sanders (approved):

hjoliver (approved):
:+1:

PR #4211, opened 2021-05-12
Nicer cylc header text

hjoliver:
A quick bit of fluff, because why not.

Several things have long annoyed me about the current "Cylc header" printed at the top of `cylc help` and `cylc play` stdout:
- it is too wide, and too tall
- yellow and green are the wrong way around
- the ASCII art looks:
   -  decidedly crappy
   - and :scream: kind of like a thinly disguised swastika (once you've seen this, you can't un-see it!)

The old:
shot

The new:
shot3
shot2
shot1

Bolder, less verbose, more compact, not offensive! - what's not to like?

### DIFF
--- a/cylc/flow/daemonize.py
+++ b/cylc/flow/daemonize.py
@@ -24,24 +24,9 @@ from time import sleep, time
 from cylc.flow.pathutil import get_workflow_run_log_name
 from cylc.flow.workflow_files import PS_OPTS
 
-
-WORKFLOW_SCAN_INFO_TMPL = r"""
-
-To view scheduler contact information:
- $ cylc get-workflow-contact %(workflow)s
-
-Other ways to see if the workflow is still running:
- $ cylc scan -n '%(workflow)s'
- $ cylc ping -v %(workflow)s
- $ ssh %(host)s ps %(ps_opts)s %(pid)s
- $ cylc tui %(workflow)s    # A terminal graphic UI
-
-"""
-
-_INFO_TMPL = r"""
-*** listening on %(url)s ***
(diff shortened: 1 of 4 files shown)


hjoliver:
(And no, it really does not matter that it's slightly harder to read)

dpmatthews:
I'm happy with the change in header although I wonder if we should support a  `-q` option to suppress it for non-interactive use.

Can we lose most of the remaining output? Users shouldn't typically need to know port numbers and process id's (they are in the log file if they need them) and do they really need instructions on how to contact the scheduler every time they use cylc play? All we really need is something like:
```
Running <workflow name> on <host name>
```

hjoliver:
Good idea @dpmatthews. I've added a `--quiet` option, got rid of the unnecessary header information, and made the logo work better for narrow terminal widths.  Description images updated above...

kinow (commented):
Don't have a strong preference for either, though that dangerous hidden image definitely justifies the change (you can't really unsee it!).

@hjoliver change looks good to me, but some functional tests are failing. Can't tell if related or not to the header (do we have any tests that expect to have the old logo, or perhaps position in the terminal? :confused: )

hjoliver:
Functional test should be fixed now @kinow

kinow:
Now an integration test failed? test_publisher I think? :confused:

hjoliver:
Damn it.

kinow (approved):

kinow:
It's always these small changes that cause the tests to misbehave 😄 I think one review will do, merging it. Thanks @hjoliver !

PR #4213, opened 2021-05-14
Tidy & add type annotations

MetRonnie:

While tackling #3743, I started adding type annotations to help me understand the existing code. Then somehow ended up tidying some of the methods in the cyclers classes

**Requirements check-list**
  `conda-environment.yml`.


cylc/flow/cycling/__init__.py
```diff
@@ -236,15 +261,39 @@ def __add__(self, other):
...
+        return self.__cmp__(other) <= 0
+
+    def __gt__(self, other: 'IntervalBase') -> bool:
+        return self.__cmp__(other) == 1
+
+    def __ge__(self, other: 'IntervalBase') -> bool:
+        return self.__cmp__(other) >= 0
+
+    def __hash__(self) -> int:
+        return hash(self.value)
```
MetRonnie, line 296: I don't think `__hash__` was being implemented in `IntegerInterval` or `ISO8601Interval` before this, so two ostensibly equal intervals would have had different hashes (per-instance hash)

cylc/flow/cycling/loader.py
```diff
@@ -63,8 +63,16 @@ class DefaultCycler:
...
+
+
+@overload
+def get_point(value: None, cycling_type: Optional[str] = None) -> None: ...
+
+
 def get_point(
-    value: str, cycling_type: Optional[str] = None
+    value: Optional[str], cycling_type: Optional[str] = None
 ) -> Optional[PointBase]:
```
MetRonnie, line 76: This is how you type annotate a function whose return value depends on the input

kinow (commented):
`pycodestyle` not happy?

```
./cylc/flow/cycling/loader.py:67:1: E704 multiple statements on one line (def)
./cylc/flow/cycling/loader.py:71:1: E704 multiple statements on one line (def)
./cylc/flow/cycling/loader.py:146:1: E704 multiple statements on one line (def)
./cylc/flow/cycling/loader.py:152:1: E704 multiple statements on one line (def)
```

MetRonnie:
https://www.flake8rules.com/rules/E704.html

I'm proposing we add this rule to the ignored list. What do you think @hjoliver ?

hjoliver:
> Also, we're running pycodestyle and flake8 for some reason - we only need flake8 as it includes pycodestyle

https://github.com/cylc/cylc-flow/pull/3999#issuecomment-745626604

(Not merged yet)

cylc/flow/cycling/loader.py
```diff
@@ -63,8 +63,16 @@ class DefaultCycler:
     TYPE: str
 
 
+@overload
+def get_point(value: str, cycling_type: Optional[str] = None) -> PointBase: ...
```
hjoliver, line 67: Can we do this, to avoid the style violation?:
```suggestion
def get_point(value: str, cycling_type: Optional[str] = None) -> PointBase:
    ...
```

hjoliver:
> https://www.flake8rules.com/rules/E704.html
> 
> I'm proposing we add this rule to the ignored list. What do you think @hjoliver ?

It's generally a sensible rule, so I'd rather not ignore it. The code looks OK as suggested in my previous comment, no?

MetRonnie:
Ok, if you prefer

MetRonnie:
It would be good to remove pycodestyle sooner, as at the moment, flake8 is ignoring more rules than pycodestyle alone

hjoliver:
> It would be good to remove pycodestyle sooner, as at the moment, flake8 is ignoring more rules than pycodestyle alone

OK, feel free to remove it here :+1:

MetRonnie:
Waiting on #4217 which fixes the manylinux tests

cylc/flow/cycling/__init__.py
```diff
@@ -41,34 +41,34 @@ def parse_exclusion(expr):
...
-    _TYPE_SORT_KEY = None
-
     __slots__ = ('value')
 
+    @property
     @abstractmethod
-    def TYPE(self):
-        return self._TYPE
+    def TYPE(self) -> str:
+        """Cycling type of this point."""
```
MetRonnie, line 65: This is how you do an abstract property in Python 3.3+
https://stackoverflow.com/a/48710068/3217306

cylc/flow/cycling/integer.py
```diff
@@ -27,7 +27,7 @@
 )
 
 CYCLER_TYPE_INTEGER = "integer"
-CYCLER_TYPE_SORT_KEY_INTEGER = "a"
+CYCLER_TYPE_SORT_KEY_INTEGER = 0
```
MetRonnie, line 30: Sorting by integer seems more natural than by string

MetRonnie:
Hmm, MacOS functional test failing; looks like it's happened on a previous run too. @oliver-sanders can I ask you to check this branch out and run `tests/f/restart/28-execution-timeout.t`?

oliver-sanders:
I can't replicate it, however, the actions artefact captures the issue, the subprocess pool encountered a `PermissionError` whilst trying to kill a process group. On MacOS this can happen for zombie processes, MacOS fails loudly, I think Linux might fail silently masking the bug? Running a test.

MetRonnie:
MacOS failure seems unrelated, marking as ready for review

cylc/flow/cycling/integer.py
hjoliver: We could conceivably allow other cycling types in the future (e.g. to cycle over a list of named datasets). They could all  be represented as strings, so long as integer cycles are also represented as strings (but compared as integers). The Point and Interval base class docstrings in `cylc/flow/cycling/__init__.py` say that derived classes should be "string based".  

However, (as your comment refers to?) this TYPE_SORT_KEY seems to be used as a proxy for instance comparison when comparing points of different cycling types ... and I don't understand why we'd want to do that either.  Any thoughts @oliver-sanders ? (`git blame` points to Ben on this, but I think it predates your involvement...).

cylc/flow/config.py
```diff
@@ -140,23 +146,23 @@ class WorkflowConfig:
...
+        log_dir: Optional[str] = None,
+        work_dir: Optional[str] = None,
+        share_dir: Optional[str] = None
+    ) -> None:
 
         self.mem_log = mem_log_func
-        if mem_log_func is None:
-            self.mem_log = lambda *a: False
+        if self.mem_log is None:
+            self.mem_log = lambda *a, **k: False
```
hjoliver, line 165: Not sure I understand why Mypy needs this, but it looks weird given that we never call this with keyword args.

It seems to be possible to annotate lambdas properly. This passes for me locally:
```diff
diff --git a/cylc/flow/config.py b/cylc/flow/config.py
index 9691a5329..63c44d29e 100644
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -153,16 +153,17 @@ class WorkflowConfig:
         is_reload: bool = False,
         output_fname: Optional[str] = None,
         xtrigger_mgr: Optional[XtriggerManager] = None,
-        mem_log_func: Optional[Callable] = None,
+        mem_log_func: Optional[Callable[[str], bool]] = None,
         run_dir: Optional[str] = None,
         log_dir: Optional[str] = None,
         work_dir: Optional[str] = None,
         share_dir: Optional[str] = None
     ) -> None:
 
-        self.mem_log = mem_log_func
-        if self.mem_log is None:
-            self.mem_log = lambda *a, **k: False
+        if mem_log_func is None:
+            self.mem_log: Callable[[str], bool] = lambda *a: False
+        else:
+            self.mem_log = mem_log_func
         self.mem_log("config.py:config.py: start init config")
         self.workflow = workflow  # workflow name
         self.fpath = fpath  # workflow definition
```

cylc/flow/config.py
```diff
@@ -206,9 +214,9 @@ def __init__(
             'first-parent descendants': {},
         }
         # tasks
-        self.leaves = []
+        self.leaves = []  # TODO figure out type
```
hjoliver, line 217: Are you planning to address these TODOs on this PR?

MetRonnie: Not really, unless someone can tell me what the type is supposed to be (I struggled to figure it out in a few minutes, and I'm not sure if it's worth spending more time on it until the code needs changing)

cylc/flow/cycling/integer.py
MetRonnie: Yes, I changed both integer's & iso8601's `TYPE_SORT_KEY`s to ints. They are used here
https://github.com/cylc/cylc-flow/blob/87e22ab623ccccb48f990dd9d64357453c1c8a56/cylc/flow/cycling/__init__.py#L110-L118

kinow (approved):
Looks good to me. Used PyCharm integration to run MyPy, it found 23 issues, but none related to the changes here (and could well be false-positives). :+1:

cylc/flow/workflow_db_mgr.py
```diff
@@ -37,6 +38,14 @@
 from cylc.flow.wallclock import get_current_time_string, get_utc_mode
 from cylc.flow.exceptions import ServiceFileError
 
+if TYPE_CHECKING:
+    from cylc.flow.task_pool import TaskPool
+
+# # TODO: narrow down Any (should be str | int) after implementing type
+# # annotations in cylc.flow.task_state.TaskState
```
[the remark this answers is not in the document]
MetRonnie: Maybe the annoyance of seeing double comments will prompt someone into type annotating `cylc.flow.task_state.TaskState` 😄

hjoliver (approved):
:+1:

PR #4221, opened 2021-05-20
subprocpool: handle PermissionError on killpg

oliver-sanders:
Currently workflows can crash during shutdown whilst trying to terminate processes in the subprocpool.

See https://github.com/cylc/cylc-flow/pull/4213#issuecomment-844305766, example traceback (from GH action s):

```
Workflow shutting down - REQUEST(NOW-NOW)
Error during shutdown
   [Errno 1] Operation not permitted
	  File "~/cylc-flow/cylc/flow/scheduler.py", line 1629, in shutdown
	    await self._shutdown(reason)
	  File "~/cylc-flow/cylc/flow/scheduler.py", line 1673, in _shutdown
	    self.proc_pool.terminate()
	  File "~/cylc-flow/cylc/flow/subprocpool.py", line 270, in terminate
	    os.killpg(proc.pid, SIGKILL)
	PermissionError: [Errno 1] Operation not permitted
INFO - DONE
```

Given that we know these subprocess were started by the user this `PermissionError` does not make sense. Here are the only explanations I've been able to come up with:

1) Process exited of its own accord but the pid was taken by a system process.
2) Process group lead or child entered the zombie state (certain systems only).

Given that we've only seen this on Mac OS (which is the certain system noted above) I'm leaning towards (2) as the likely cause.

I'm not sure what Linux does in this case as a zombie is "unkillable" by definition, I guess Mac OS is failing loudly (which is quite sensible as zombies shouldn't exist) whereas Linux fails silently? Dunno.

Suggest catching the error to allow the shutdown to proceed and logging the error.

Kinda a bug as it make a mess of the shutdown, kinda not in that failing loudly in this situation is the right general approach.

**Requirements check-list**
  `conda-environment.yml`.
- [ ] Appropriate tests are included (unit and/or functional) - perhaps later

### DIFF
--- a/cylc/flow/subprocpool.py
+++ b/cylc/flow/subprocpool.py
@@ -34,6 +34,28 @@ from cylc.flow.wallclock import get_current_time_string
 _XTRIG_FUNCS: dict = {}
 
 
+def _killpg(proc, signal):
+    """Kill a process group."""
+    try:
+        os.killpg(proc.pid, signal)
+    except ProcessLookupError:
+        # process group has already exited
+        return False
+    except PermissionError:
+        # process group may contain zombie processes which will result in
+        # PermissionError on some systems, not sure what happens on others
+        #
+        # we could go through the processes in the group and call waitpid on
+        # them but waitpid is blocking and this would be a messy solution for a
+        # problem that shouldn't happen (it's really a bug in the Cylc subproc)
+        LOG.error(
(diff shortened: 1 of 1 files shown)


hjoliver (approved):
Seems reasonable.  I don't think I've seen the error myself, on Linux.

Issue #3666, opened 2020-06-30
public database not closed correctly - consider persisting the connection


oliver-sanders:
**The Issue:**

Cylc is not closing the public database file (despite logic appearing to close it correctly). We also seem to be hanging onto the canonical log dir.

Once Python exits the file descriptor terminates so this hasn't proven symptomatic before, however, with #3654 we can run multiple schedulers in the same event loop, this bug keeps the public database open until Python exits which prevents reliably housekeeping.

**Connect - Close:**

At the moment we connect to and close the database very often, too often. According to the code comments this was an early form of filesystem integrity checking which caused Cylc to exit if the run directory had been emptied.

We now have a separate check for this (which can be improved if necessary) so should be able to dispense with the (highly inefficient) connection/closing of the DB.

Apply this diff then look at the logs. Warning it might scare you, sometimes ignorance is bliss:

```diff
diff --git a/cylc/flow/rundb.py b/cylc/flow/rundb.py
index 37f3789ad..19dda6ed1 100644
--- a/cylc/flow/rundb.py
+++ b/cylc/flow/rundb.py
@@ -367,7 +367,9 @@ class CylcSuiteDAO(object):
 
     def close(self):
         """Explicitly close the connection."""
+        LOG.warning('# close')
         if self.conn is not None:
+            LOG.warning('$ close')
             try:
                 self.conn.close()
             except sqlite3.Error:
@@ -376,7 +378,9 @@ class CylcSuiteDAO(object):
 
     def connect(self):
         """Connect to the database."""
+        LOG.warning('# connect')
         if self.conn is None:
+            LOG.warning('$ re-connect')
             self.conn = sqlite3.connect(self.db_file_name, self.CONN_TIMEOUT)
         return self.conn
```

**The Evidence:**

This diff reveals that the public database (which should be closed during `schd.shutdown()` appears to remain open until Python exits.

```diff
diff --git a/cylc/flow/scheduler.py b/cylc/flow/scheduler.py
index 5eb285f8d..44c0e8369 100644
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -121,6 +121,12 @@ class SchedulerUUID(object):
         return self.value
 
 
+def log_open_files():
+    import psutil
+    proc = psutil.Process()
+    LOG.warning(f'# {proc.open_files()}')
+
+
 class Scheduler(object):
     """Cylc scheduler server."""
 
[...]

PR #1493, opened 2015-06-09
cylc.rundb: add task jobs table and refactor

matthewrmshin:
* Simplify by removing threading logic.
* Group `insert` and similar `update` statements, and write with `executemany` for efficiency.
  * A single queue per table for inserts.
  * A queue for each unique update statement template.
* On data retrieval, use single `select` statement to get every relevant items.
* Die immediately on any problem with private database to guarantee integrity.
  * Database can only be out of sync by a single iteration on the main loop.
* Unlimited number of retries for public database.
  Retry occurs on next iteration of the main loop - no sleep required.
  Suite can only be brought down if queues become so big that memory runs out.
  This is unlikely, because things should recover sooner or later.
* On restart,
  * Vacuum private database
  * Copy private database into public database to ensure they are in sync.
* Add new `task_jobs` table.
* Add new `broadcasts` table to record broadcast settings in searchable form.
  Remove `broadcast_settings` table, which stored unsearchable pickles.

PR #4222, opened 2021-05-20
Close public database correctly

MetRonnie:
These changes partially address #3666

When the scheduler starts, the public database is synced with the private database. This is done by copying the private database into a tempfile, which is then renamed to replace the public database file (the tempfile is not deleted) (see https://github.com/cylc/cylc-flow/pull/1493#issuecomment-111044334 for the rationale behind this). But the file descriptor for the tempfile was not closed at any point, hence the Python process held on to the file beyond the shutdown of the scheduler.

I have not tackled the suggestion of persisting the connection(s) to the databases.

**Requirements check-list**
  `conda-environment.yml`.
- [x] Does not need tests (integration tests will <del>fail</del><ins> error during teardown</ins> if this issue reappears).

### DIFF
--- a/cylc/flow/workflow_db_mgr.py
+++ b/cylc/flow/workflow_db_mgr.py
@@ -25,10 +25,9 @@ This module provides the logic to:
 
 import json
 import os
+from pkg_resources import parse_version
 from shutil import copy, rmtree
 from tempfile import mkstemp
-
-from pkg_resources import parse_version
 
 from cylc.flow import LOG
 from cylc.flow.broadcast_report import get_broadcast_change_iter
@@ -36,6 +35,9 @@ from cylc.flow.rundb import CylcWorkflowDAO
 from cylc.flow import __version__ as CYLC_VERSION
 from cylc.flow.wallclock import get_current_time_string, get_utc_mode
 from cylc.flow.exceptions import ServiceFileError
+
+
+PERM_PRIVATE = 0o600  # -rw-------
 
 
 class WorkflowDatabaseManager:
@@ -112,27 +114,33 @@ class WorkflowDatabaseManager:
             self.TABLE_ABS_OUTPUTS: []}
         self.db_updates_map = {}
 
-    def copy_pri_to_pub(self):
-        """Copy content of primary database file to public database file.
-
-        Use temporary file to ensure that we do not end up with a partial file.
-
-        """
-        temp_pub_db_file_name = None
+    def copy_pri_to_pub(self) -> None:
+        """Copy content of primary database file to public database file."""
         self.pub_dao.close()
+        # Use temporary file to ensure that we do not end up with a
+        # partial file.
+        # If an external connection is locking the old public db, it will
+        # still be connected to its inode, but should no longer affect future
+        # accesses (hopefully that process will soon recover to give up
+        # the lock).
+        temp_pub_db_fd, temp_pub_db_file_name = mkstemp(
+            prefix=self.pub_dao.DB_FILE_BASE_NAME,
+            dir=os.path.dirname(self.pub_dao.db_file_name)
+        )
+        os.close(temp_pub_db_fd)
         try:
-            self.pub_dao.conn = None  # reset connection
-            open(self.pub_dao.db_file_name, "a").close()  # touch
+            # Create the file if it didn't exist; this is done in the hope of
+            # addressing potential NFS file lag, we think
+            open(self.pub_dao.db_file_name, "a").close()
+            # Get default permissions level for public db:
             st_mode = os.stat(self.pub_dao.db_file_name).st_mode
-            temp_pub_db_file_name = mkstemp(
-                prefix=self.pub_dao.DB_FILE_BASE_NAME,
-                dir=os.path.dirname(self.pub_dao.db_file_name))[1]
+
             copy(self.pri_dao.db_file_name, temp_pub_db_file_name)
             os.rename(temp_pub_db_file_name, self.pub_dao.db_file_name)
             os.chmod(self.pub_dao.db_file_name, st_mode)
-        except (IOError, OSError):
-            if temp_pub_db_file_name:
-                os.unlink(temp_pub_db_file_name)
+        except OSError:
+            if os.path.exists(temp_pub_db_file_name):
+                os.remove(temp_pub_db_file_name)
             raise
 
     def delete_workflow_params(self, *keys):
@@ -195,7 +203,7 @@ class WorkflowDatabaseManager:
                 # Just in case the path is a directory!
                 rmtree(self.pri_path, ignore_errors=True)
         self.pri_dao = self.get_pri_dao()
-        os.chmod(self.pri_path, 0o600)
+        os.chmod(self.pri_path, PERM_PRIVATE)
         self.pub_dao = CylcWorkflowDAO(self.pub_path, is_public=True)
         self.copy_pri_to_pub()
 
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -68,10 +68,10 @@ def pytest_runtest_makereport(item, call):
     item.module._module_outcomes = _module_outcomes
 
 
-def _pytest_passed(request):
+def _pytest_passed(request: pytest.FixtureRequest) -> bool:
     """Returns True if the test(s) a fixture was used in passed."""
     if hasattr(request.node, '_function_outcome'):
-        return request.node._function_outcome in {'passed', 'skipped'}
+        return request.node._function_outcome.outcome in {'passed', 'skipped'}
     return all((
         report.outcome in {'passed', 'skipped'}
         for report in request.node.obj._module_outcomes.values()
@@ -105,7 +105,7 @@ def mod_test_dir(request, ses_test_dir):
     yield path
     if _pytest_passed(request):
         # test passed -> remove all files
-        rmtree(path, ignore_errors=True)
+        rmtree(path, ignore_errors=False)
     else:
         # test failed -> remove the test dir if empty
         _rm_if_empty(path)
@@ -119,7 +119,7 @@ def test_dir(request, mod_test_dir):
     yield path
     if _pytest_passed(request):
         # test passed -> remove all files
-        rmtree(path, ignore_errors=True)
+        rmtree(path, ignore_errors=False)
     else:
         # test failed -> remove the test dir if empty
         _rm_if_empty(path)
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -18,11 +18,10 @@
 import asyncio
 from functools import partial
 from pathlib import Path
+import pytest
 import re
 from shutil import rmtree
 from typing import List, TYPE_CHECKING
-
-import pytest
 
 from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
 from cylc.flow.pathutil import get_workflow_run_dir
@@ -139,14 +138,24 @@ def flow(run_dir, test_dir):
 
 @pytest.fixture(scope='module')
 def mod_scheduler():
-    """Return a scheduler object for a flow."""
-    return _make_scheduler
+    """Return a Scheduler object for a flow.
+
+    Usage: see scheduler() below
+    """
+    with _make_scheduler() as _scheduler:
+        yield _scheduler
 
 
 @pytest.fixture
 def scheduler():
-    """Return a scheduler object for a flow."""
-    return _make_scheduler
+    """Return a Scheduler object for a flow.
+
+    Args:
+        reg (str): Workflow name.
+        **opts (Any): Options to be passed to the Scheduler.
+    """
+    with _make_scheduler() as _scheduler:
+        yield _scheduler
 
 
 @pytest.fixture(scope='module')
--- a/tests/integration/utils/flow_tools.py
+++ b/tests/integration/utils/flow_tools.py
@@ -23,11 +23,11 @@ Use the fixtures provided in the conftest instead.
 
 import asyncio
 from async_timeout import timeout
-from async_generator import asynccontextmanager
+from contextlib import asynccontextmanager, contextmanager
 import logging
 from pathlib import Path
 import pytest
-from typing import Optional
+from typing import Any, Optional
 from uuid import uuid1
 
 from cylc.flow import CYLC_LOG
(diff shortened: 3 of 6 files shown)


MetRonnie:
Issue 3666 is addressed by PR 4222 👌

cylc/flow/workflow_db_mgr.py
```diff
@@ -112,27 +115,27 @@ def __init__(self, pri_d=None, pub_d=None):
...
+        os.close(temp_pub_db_fd)
         try:
-            self.pub_dao.conn = None  # reset connection
-            open(self.pub_dao.db_file_name, "a").close()  # touch
-            st_mode = os.stat(self.pub_dao.db_file_name).st_mode
-            temp_pub_db_file_name = mkstemp(
-                prefix=self.pub_dao.DB_FILE_BASE_NAME,
-                dir=os.path.dirname(self.pub_dao.db_file_name))[1]
             copy(self.pri_dao.db_file_name, temp_pub_db_file_name)
             os.rename(temp_pub_db_file_name, self.pub_dao.db_file_name)
```
MetRonnie, line 134: We might want to use this instead?
```suggestion
            os.replace(temp_pub_db_file_name, self.pub_dao.db_file_name)
```
https://docs.python.org/3.7/library/os.html#os.rename
> If you want cross-platform overwriting of the destination, use `replace()`.

cylc/flow/workflow_db_mgr.py
```diff
@@ -112,27 +115,27 @@ def __init__(self, pri_d=None, pub_d=None):
...
-            os.chmod(self.pub_dao.db_file_name, st_mode)
-        except (IOError, OSError):
-            if temp_pub_db_file_name:
-                os.unlink(temp_pub_db_file_name)
+            os.chmod(self.pub_dao.db_file_name, PERM_PUBLIC)
+        except OSError:
```
MetRonnie, line 136: `IOError` was merged into `OSError` in Python 3.3

tests/integration/conftest.py
```diff
@@ -139,14 +138,24 @@ def flow(run_dir, test_dir):
...
 @pytest.fixture(scope='module')
 def mod_scheduler():
-    """Return a scheduler object for a flow."""
-    return _make_scheduler
+    """Return a Scheduler object for a flow.
+
+    Usage: see scheduler() below
+    """
+    with _make_scheduler() as _scheduler:
+        yield _scheduler
```
MetRonnie, line 146: This slightly awkward syntax is how you reuse a "factory as fixture" that yields instead of returns

cylc/flow/workflow_db_mgr.py
hjoliver: It seems that would make no difference for us: https://stackoverflow.com/questions/60369291/os-link-vs-os-rename-vs-os-replace-for-writing-atomic-write-files-what/65553754#65553754

hjoliver (approved):
Looks good :+1:

cylc/flow/workflow_db_mgr.py
```diff
@@ -112,27 +115,27 @@ def __init__(self, pri_d=None, pub_d=None):
...
+        # accesses (hopefully that process will soon recover to give up
+        # the lock).
+        temp_pub_db_fd, temp_pub_db_file_name = mkstemp(
+            prefix=self.pub_dao.DB_FILE_BASE_NAME,
+            dir=os.path.dirname(self.pub_dao.db_file_name)
+        )
+        os.close(temp_pub_db_fd)
         try:
-            self.pub_dao.conn = None  # reset connection
-            open(self.pub_dao.db_file_name, "a").close()  # touch
```
oliver-sanders, line 125: Unhelpfully there is no explaining comment here, however, I think this open/close is more functional than it appears. This helps address potential NFS file lag which can result in a long delay between the DB being written and becoming visible across the network. The commit history might b

cylc/flow/workflow_db_mgr.py
```diff
@@ -112,27 +115,27 @@ def __init__(self, pri_d=None, pub_d=None):
...
+        # Use temporary file to ensure that we do not end up with a
+        # partial file.
+        # If an external connection is locking the old public db, it will
+        # still be connected to its inode, but should no longer affect future
+        # accesses (hopefully that process will soon recover to give up
+        # the lock).
+        temp_pub_db_fd, temp_pub_db_file_name = mkstemp(
+            prefix=self.pub_dao.DB_FILE_BASE_NAME,
+            dir=os.path.dirname(self.pub_dao.db_file_name)
+        )
```
oliver-sanders, line 130: Previously `IOError` was caught for the creation of this file. I think the main reason this would be raised is if the user runs out of disk space (on a network filesystem this may mean they have hit quota). In this event the public DB copy would have been skipped, now it will force scheduler shutdown.

There isn't much Cylc can do in this situation, a failed DB write would result in scheduler shutdown anyway so it's most likely the workflow is going to die anyway. I'm just a little cautious changing this behaviour as I know it has been evolved to handle nasty problems in the past (there should really be comments to reflect this!).

tests/integration/conftest.py
```diff
@@ -105,7 +104,7 @@ def mod_test_dir(request, ses_test_dir):
     yield path
     if _pytest_passed(request):
         # test passed -> remove all files
-        rmtree(path, ignore_errors=True)
+        rmtree(path, ignore_errors=False)
```
oliver-sanders, line 107: 🎉

tests/integration/utils/flow_tools.py
```diff
@@ -54,13 +54,24 @@ def _make_flow(run_dir, test_dir, conf, name=None):
...
+        options = RunOptions(**opts)
+        # create workflow
+        nonlocal schd
+        schd = Scheduler(reg, options)
+        return schd
+
+    yield __make_scheduler
+    # Teardown
+    if hasattr(schd, 'workflow_db_mgr'):
+        schd.workflow_db_mgr.on_workflow_shutdown()
```
oliver-sanders, line 74: Does this mean that `on_workflow_shutdown` isn't being called?

We shouldn't need any teardown logic here as the `Scheduler` is supposed to handle all of this itself (if it isn't that's a bug).

MetRonnie: Integration tests like this don't need it:
https://github.com/cylc/cylc-flow/blob/0230c9da0690ead76e9747a8f2f16bf175f3d51b/tests/integration/test_examples.py#L59-L60

It's tests like this that need the teardown:
https://github.com/cylc/cylc-flow/blob/0230c9da0690ead76e9747a8f2f16bf175f3d51b/tests/integration/test_examples.py#L137-L140

because they don't start the scheduler, and so don't shut it down either

cylc/flow/workflow_db_mgr.py
MetRonnie: But it doesn't need to be inside the `try` block, does it? If it errors out during creation of the file, the file won't actually be created, so no need to `os.remove()` it? Or is that too much of an assumption?

tests/integration/utils/flow_tools.py
oliver-sanders: Ah, dammit, we are creating the DB in order to write the workflow parameters to it at configure time so the `on_workflow_start` method is really `on_workflow_configure`. My bad, I made a rough cut at pulling apart the different stages of workflow startup with the original `async` work but they are still quite tangled and need a bit of TLC.

I've had a quick try at delaying the DB creation until workflow start, this diff passes `tests/i`:

```diff
diff --git a/cylc/flow/scheduler.py b/cylc/flow/scheduler.py
index 06b9de96c..46d662ea5 100644
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -430,8 +430,6 @@ class Scheduler:
         self.load_flow_file()
         self.profiler.log_memory("scheduler.py: after load_flow_file")
 
-        self.workflow_db_mgr.on_workflow_start(self.is_restart)
-
         if not self.is_restart:
             # Set workflow params that would otherwise be loaded from database:
             self.options.utc_mode = get_utc_mode()
@@ -463,6 +461,7 @@ class Scheduler:
 
         self.profiler.log_memory("scheduler.py: before load_tasks")
         if self.is_restart:
+            self.workflow_db_mgr.on_workflow_restart_configure()
             self.load_tasks_for_restart()
             if self.restored_stop_task_id is not None:
                 self.pool.set_stop_task(self.restored_stop_task_id)
@@ -471,10 +470,6 @@ class Scheduler:
         self.process_cylc_stop_point()
         self.profiler.log_memory("scheduler.py: after load_tasks")
 
-        self.workflow_db_mgr.put_workflow_params(self)
-        self.workflow_db_mgr.put_workflow_template_vars(self.template_vars)
-        self.workflow_db_mgr.put_runtime_inheritance(self.config)
-
         self.already_timed_out = False
         self.set_workflow_timer()
 
@@ -567,6 +562,10 @@ class Scheduler:
 
     async def start_scheduler(self):
         """Start the scheduler main loop."""
+        self.workflow_db_mgr.on_workflow_start(self.is_restart)
+        self.workflow_db_mgr.put_workflow_params(self)
+        self.workflow_db_mgr.put_workflow_template_vars(self.template_vars)
+        self.workflow_db_mgr.put_runtime_inheritance(self.config)
         try:
             self._configure_contact()
             if self.is_restart:
diff --git a/cylc/flow/workflow_db_mgr.py b/cylc/flow/workflow_db_mgr.py
index f471d0c72..e04f7038a 100644
--- a/cylc/flow/workflow_db_mgr.py
+++ b/cylc/flow/workflow_db_mgr.py
@@ -181,6 +181,10 @@ class WorkflowDatabaseManager:
         else:
             return json.dumps([type(obj).__name__, obj.__getnewargs__()])
 
+    def on_workflow_restart_configure(self):
+        self.pri_dao = self.get_pri_dao()
+        os.chmod(self.pri_path, 0o600)
+
     def on_workflow_start(self, is_restart):
         """Initialise data access objects.
 
@@ -194,7 +198,8 @@ class WorkflowDatabaseManager:
             except OSError:
                 # Just in case the path is a directory!
                 rmtree(self.pri_path, ignore_errors=True)
-        self.pri_dao = self.get_pri_dao()
+        if not self.pri_dao:
+            self.pri_dao = self.get_pri_dao()
         os.chmod(self.pri_path, 0o600)
         self.pub_dao = CylcWorkflowDAO(self.pub_path, is_public=True)
         self.copy_pri_to_pub()
```

cylc/flow/workflow_db_mgr.py
MetRonnie: Can't find any note of why this open-close/`touch` is done in the original PR #1493. It's there in the first commit that introduced the "copy pri to pub" method: https://github.com/cylc/cylc-flow/pull/1493/commits/aeb784e8520ee18e82e9abb14bc97b35348f4204#diff-6f57f386f371bc9f0ee46a01bd8ab7455793a8b84dbbbd1c98fa9a86d289621aR1380

Is the `os.rename(temp_pub_db_file_name, self.pub_dao.db_file_name)` not sufficient to create it?

cylc/flow/workflow_db_mgr.py
```diff
@@ -112,27 +115,27 @@ def __init__(self, pri_d=None, pub_d=None):
...
-            temp_pub_db_file_name = mkstemp(
-                prefix=self.pub_dao.DB_FILE_BASE_NAME,
-                dir=os.path.dirname(self.pub_dao.db_file_name))[1]
             copy(self.pri_dao.db_file_name, temp_pub_db_file_name)
             os.rename(temp_pub_db_file_name, self.pub_dao.db_file_name)
-            os.chmod(self.pub_dao.db_file_name, st_mode)
-        except (IOError, OSError):
-            if temp_pub_db_file_name:
-                os.unlink(temp_pub_db_file_name)
+            os.chmod(self.pub_dao.db_file_name, PERM_PUBLIC)
```
MetRonnie, line 135: Just occurred to me, could the public db permissions be different to 0644 for a pre-existing db?

cylc/flow/workflow_db_mgr.py
oliver-sanders: Yes, no need to try `os.remove()` for something that's not created. It's more that if if errors during the creation of the file that error will bubble up to the Scheduler and force shutdown.

cylc/flow/workflow_db_mgr.py
oliver-sanders: > Is the os.rename(temp_pub_db_file_name, self.pub_dao.db_file_name) not sufficient to create it?

I think the answer is an awkward combination of yes and no, yes the file exists in the metadata, no it doesn't exist in the metadata everywhere. Network filesystems are very awkward, events simultaneous in one frame aren't in another, relativistic rules apply! Matt did some investigation a while back and found that opening and closing files like this makes a difference even when `fsync` doesn't, unfortunately that's about the limit of my understanding on this.

cylc/flow/workflow_db_mgr.py
oliver-sanders: Looks like before we weren't setting the permission of public DB, so it came out with the default os mask.

I think it should be left that way, if the default mask is not good enough for public access then it doesn't matter what the permissions on the file are as the workflow run directory will not be visible anyway.

The user might have a non-standard mark depending on site-policy e.g. they might want `664`, `640`, whatever.

cylc/flow/workflow_db_mgr.py
MetRonnie: Ok, will revert and add some useful comments

cylc/flow/workflow_db_mgr.py
MetRonnie: Previously it would re-raise the `IOError/OSError` anyway though?
https://github.com/cylc/cylc-flow/blob/0230c9da0690ead76e9747a8f2f16bf175f3d51b/cylc/flow/workflow_db_mgr.py#L123-L136

cylc/flow/workflow_db_mgr.py
MetRonnie: Ah, so `os.rename` is fine for POSIX

cylc/flow/workflow_db_mgr.py
oliver-sanders: Ah, sorry, I missed the `raise` when reading the diff, whatever the behaviour was before is good for now.

cylc/flow/workflow_db_mgr.py
oliver-sanders: Thanks and sorry for these strange bits of unexplained logic, we need to get a little better at documenting the strange quirks we encounter in the fixes. I get a little nervous about introducing regressions around some of these bits of code.

tests/integration/utils/flow_tools.py
MetRonnie: Ok yeah, I've started along this path based on that diff but it's starting to balloon, so I'll put the rest in a later PR

oliver-sanders (approved):

PR #4227, opened 2021-05-25
Fix ICP bug.

### LINKED ISSUE #4225: config.py & iso8601.py try to access the isodatetime parser even if cycling mode is integer
**Describe the bug**

Given the workflow below (NB: `cycling mode = integer`)

```cylc
[scheduling]
  cycling mode = integer
  initial cycle point = 1
  [[queues]]
     [[[default]]]
       limit = 1
  [[graph]]
    R1 = "prep => foo"
    P1 = "foo[-P1] => foo => bar"

[runtime]
  [[root]]
      script="sleep 5"
  [[prep]]
  [[foo]]
  [[bar]]
```

Running a command with the initial CP as now or any invalid text results in the `config.py` and `iso8601.py` modules trying to access the isodatetime parser of the workflow, which is `None` since we specified it's an integer cycling, resulting in a traceback.

```
(venv) kinow@ranma:~/cylc-run/five$ cylc play --no-detach five --initial-cycle-point="2^2"

                Cylc Workflow Engine 8.0b2.dev
                Copyright (C) 2008-2021 NIWA

2021-05-25T11:46:56+12:00 ERROR - 'NoneType' object has no attribute 'parse'
	  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/scheduler.py", line 627, in run
	    await self.configure()
	  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/scheduler.py", line 430, in configure
	    self.load_flow_file()
	  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/scheduler.py", line 1076, in load_flow_file
	    self.config = WorkflowConfig(
	  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/config.py", line 352, in __init__
	    self.process_initial_cycle_point()
	  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/config.py", line 743, in process_initial_cycle_point
	    icp = ingest_time(orig_icp, my_now)
	  File "/home/kinow/Development/python/workspace/cylc-flow/cylc/flow/cycling/iso8601.py", line 683, in ingest_time
	    timepoint = parser.parse(value)
	AttributeError: 'NoneType' object has no attribute 'parse'
2021-05-25T11:46:56+12:00 CRITICAL - Workflow shutting down - 'NoneType' object has no attribute 'parse'
2021-05-25T11:46:56+12:00 INFO - DONE
```

`master`

Install workflow above. Then run with the command in the traceback above.

**Expected behavior**

**Screenshots**

image

**Additional context**

Some discussion about it today in the General room.

**Pull requests welcome!**

hjoliver:
These changes close #4225

```
$ cylc play --no-detach five --initial-cycle-point="2^2"

                Cylc Workflow Engine 8.0b2.dev
                Copyright (C) 2008-2021 NIWA

2021-05-25T13:15:56+12:00 ERROR - Workflow shutting down - PointParsingError: Incompatible value for
        <class 'cylc.flow.cycling.integer.IntegerPoint'>: 2^2: invalid literal for int() with base 10: '2^2'
2021-05-25T13:15:56+12:00 INFO - DONE

```

**Requirements check-list**
- [x] No dependency changes.

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -739,20 +739,22 @@ class WorkflowConfig:
             WorkflowConfigError - if it fails to validate
         """
         orig_icp = self.cfg['scheduling']['initial cycle point']
-        if orig_icp is None:
-            if self.cfg['scheduling']['cycling mode'] == INTEGER_CYCLING_TYPE:
-                orig_icp = '1'
+        if self.cycling_type == INTEGER_CYCLING_TYPE:
+            if orig_icp is None:
+                icp = orig_icp = '1'
             else:
+                icp = orig_icp
+        elif self.cycling_type == ISO8601_CYCLING_TYPE:
+            if orig_icp is None:
                 raise WorkflowConfigError(
                     "This workflow requires an initial cycle point.")
-        if orig_icp == "now":
-            icp = get_current_time_string()
-        else:
-            try:
-                my_now = get_current_time_string()
-                icp = ingest_time(orig_icp, my_now)
-            except IsodatetimeError as exc:
-                raise WorkflowConfigError(str(exc))
+            if orig_icp == "now":
+                icp = get_current_time_string()
+            else:
+                try:
+                    icp = ingest_time(orig_icp, get_current_time_string())
+                except IsodatetimeError as exc:
+                    raise WorkflowConfigError(str(exc))
         if orig_icp != icp:
             # now/next()/prev() was used, need to store evaluated point in DB
             self.options.icp = icp
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -741,9 +741,8 @@ class WorkflowConfig:
         orig_icp = self.cfg['scheduling']['initial cycle point']
         if self.cycling_type == INTEGER_CYCLING_TYPE:
             if orig_icp is None:
-                icp = orig_icp = '1'
-            else:
-                icp = orig_icp
+                orig_icp = '1'
+            icp = orig_icp
         elif self.cycling_type == ISO8601_CYCLING_TYPE:
             if orig_icp is None:
                 raise WorkflowConfigError(
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -56,24 +56,23 @@ def tmp_run_dir(tmp_path: Path, monkeypatch: pytest.MonkeyPatch):
 
 
 @pytest.fixture
-def cycling_mode(monkeypatch: pytest.MonkeyPatch):
-    """Set the Cylc cycling mode.
-
+def cycling_type(monkeypatch: pytest.MonkeyPatch) -> str:
+    """Set the Cylc cycling type.
     Args:
-        mode: The cycling mode.
-        time_zone: If using ISO8601/datetime cycling mode, you can specify a
+        type: The cycling type.
+        time_zone: If using ISO8601/datetime cycling type, you can specify a
             custom time zone to use.
     """
-    def _cycling_mode(
-        mode: str = INTEGER_CYCLING_TYPE, time_zone: Optional[str] = None
+    def _cycling_type(
+        ctype: str = INTEGER_CYCLING_TYPE, time_zone: Optional[str] = None
     ) -> None:
         class _DefaultCycler:
-            TYPE = mode
+            TYPE = ctype
         monkeypatch.setattr(
             'cylc.flow.cycling.loader.DefaultCycler', _DefaultCycler)
-        if mode == ISO8601_CYCLING_TYPE:
+        if ctype == ISO8601_CYCLING_TYPE:
             iso8601_init(time_zone=time_zone)
-    return _cycling_mode
+    return _cycling_type
 
 
 @pytest.fixture
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -56,22 +56,27 @@ def tmp_run_dir(tmp_path: Path, monkeypatch: pytest.MonkeyPatch):
 
 
 @pytest.fixture
-def cycling_type(monkeypatch: pytest.MonkeyPatch) -> str:
-    """Set the Cylc cycling type.
+def cycling_type(monkeypatch: pytest.MonkeyPatch):
+    """Initialize the cycling type according to cycling mode.
     Args:
-        type: The cycling type.
+        mode: The cycling mode.
         time_zone: If using ISO8601/datetime cycling type, you can specify a
             custom time zone to use.
     """
     def _cycling_type(
-        ctype: str = INTEGER_CYCLING_TYPE, time_zone: Optional[str] = None
-    ) -> None:
+        mode: str = 'integer', time_zone: Optional[str] = None
+    ) -> str:
+        if mode == 'integer':
+            ctype = INTEGER_CYCLING_TYPE
+        else:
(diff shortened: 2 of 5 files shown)


cylc/flow/config.py
```diff
@@ -729,20 +729,22 @@ def process_initial_cycle_point(self):
             WorkflowConfigError - if it fails to validate
         """
         orig_icp = self.cfg['scheduling']['initial cycle point']
-        if orig_icp is None:
-            if self.cfg['scheduling']['cycling mode'] == INTEGER_CYCLING_TYPE:
-                orig_icp = '1'
+        if self.cycling_type == INTEGER_CYCLING_TYPE:
```
kinow, line 732: @hjoliver I **think** I saw `cycling_type` in the debugger, and it was `None`? But there was a `cycling mode` in the `cfg['scheduling']` that had the `integer` value. Not sure if that's helpful/important for this issue :+1:

hjoliver: `self.cycling_type` is initialized in `WorkflowConfig.__init__()`.  It shouldn't be `None`, but I'll take another look...

hjoliver: Unfortunately `self.cfg['scheduling']['cycling mode']` was being used incorrectly here.  In the integer case it is equal to `INTEGER_CYCLING_TYPE` but in the date-time case it is equal to the calendar mode, not to `ISO8601_CYCLING_TYPE`.

hjoliver:
Unit tests added.

cylc/flow/config.py
```diff
@@ -729,20 +729,22 @@ def process_initial_cycle_point(self):
...
+            if orig_icp is None:
+                icp = orig_icp = '1'
             else:
+                icp = orig_icp
```
kinow, line 736: Or also

```py
            if orig_icp is None:
                orig_icp = '1'
            icp = orig_icp
```

Either way is OK for me :man_shrugging:

tests/unit/test_config.py
```diff
@@ -272,27 +272,57 @@ def test_family_inheritance_and_quotes(self, mock_glbl_cfg):
...
+        (  # non-ISO8601 ICP for ISO8601 cycling type
+            {
+                'cycling type': loader.ISO8601_CYCLING_TYPE,
+                'initial cycle point': "1",
+                'initial cycle point constraints': []
+            },
+            None,
+            None,
+            (PointParsingError, "Invalid ISO 8601 date representation")
+        ),
```
kinow, line 322: :+1:

kinow (approved):

cylc/flow/config.py
hjoliver: Your way is better :grin:

hjoliver:
(Pushed small tweak with skip-ci; all tests passed on previous commit, except patch coverage 83%)

tests/unit/conftest.py
```diff
@@ -55,17 +55,17 @@ def inner(reg: Optional[str] = None) -> Path:
...
 
 @pytest.fixture
-def cycling_mode(monkeypatch):
-    """Set the Cylc cycling mode and return its value."""
-    def _cycling_mode(integer=True):
-        mode = INTEGER_CYCLING_TYPE if integer else ISO8601_CYCLING_TYPE
+def cycling_type(monkeypatch):
+    """Set the Cylc cycling type and return its value."""
+    def _cycling_type(integer=True):
+        ctype = INTEGER_CYCLING_TYPE if integer else ISO8601_CYCLING_TYPE
```
MetRonnie, line 61: I hadn't realised the difference between cycling type and mode. If I understand correctly:
- Type is either `integer` or `iso8601`
- Mode is either `integer`, `gregorian`, `360day`, `365day` etc

Btw, just a heads up this is going to conflict with my PR #4213 - https://github.com/cylc/cylc-flow/pull/4213/files#diff-0cae8a7ee2d37098b1ad84b543d17cfc1e8535eed5fd6abac88c668bfe354cbb

tests/unit/test_config.py
```diff
@@ -272,27 +272,57 @@ def test_family_inheritance_and_quotes(self, mock_glbl_cfg):
     [
         (  # Lack of icp
             {
-                'cycling mode': loader.ISO8601_CYCLING_TYPE,
+                'cycling type': loader.ISO8601_CYCLING_TYPE,
```
MetRonnie, line 275: Strictly speaking, this ought to be moved out of the `scheduling_cfg` dict into its own parameter. But seeing as I think I wrote these tests, I can do that later.

tests/unit/test_config.py
```diff
@@ -272,27 +272,57 @@ def test_family_inheritance_and_quotes(self, mock_glbl_cfg):
...
+        (  # more non-integer ICP for integer cycling type
+            {
+                'cycling type': loader.INTEGER_CYCLING_TYPE,
+                'initial cycle point': "now",
+                'initial cycle point constraints': []
+            },
+            None,
+            None,
+            (PointParsingError, "invalid literal for int()")
+        ),
```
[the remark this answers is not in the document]
hjoliver: Oops, a cut-and-paste I intended to modify to check a different non-integer ICP.

tests/unit/conftest.py
hjoliver: That's correct, I had forgotten the distinction myself.

hjoliver: > Btw, just a heads up this is going to conflict with my PR

And unfortunately (for me!) I just merged that PR :grimacing:

tests/unit/test_config.py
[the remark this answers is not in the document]
hjoliver: Fixed.

tests/unit/test_config.py
hjoliver: It's now `cycling mode` again, which belongs in the config dict.

tests/unit/conftest.py
hjoliver: Deconflicted.  And I've attempted to straighten out the cycling type vs mode stuff in the process.

tests/unit/conftest.py
```diff
@@ -56,24 +56,28 @@ def inner(reg: Optional[str] = None) -> Path:
...
-    def _cycling_mode(
-        mode: str = INTEGER_CYCLING_TYPE, time_zone: Optional[str] = None
-    ) -> None:
+    def _cycling_type(
+        mode: str = 'integer', time_zone: Optional[str] = None
+    ) -> str:
+        if mode == 'integer':
+            ctype = INTEGER_CYCLING_TYPE
+        else:
+            ctype = ISO8601_CYCLING_TYPE
```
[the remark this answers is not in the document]
hjoliver: True, but I was  trying to maintain the distinction between the `cycling mode` user config item, and `cycling type`, which is internal.  And your parameterized tests mostly do this via a mocked config dict, so they should really specify cycling mode, not type ... and then this fixture sets the internal type according to the mode.  Finally, the `gregorian` cycling mode is the default, so I set those values to empty string rather than hard-wiring another magic value.

So how about we leave this PR as-is, and as a low priority follow-up we could refactor the tests to specify cycling type directly, but separately from the mocked config?

MetRonnie (approved):
I have deconflicted the changelog

PR #4228, opened 2021-05-25
Cylc cli runN fix

datamel:
This fixes a bug whereby if using runN on the cli, the flow name is not correctly mapped to the actual run number.

To reproduce the bug..
Have an installed workflow, myflow/runN which maps to e.g. myflow/run1 
start a workflow off `cylc play myflow/runN`
stop the workflow `cylc stop myflow/run1 --now --now`
a `cylc scan` should confirm the workflow is still running - since the workflow was started with `myflow/runN`, this is used rather than `myflow/run1`  as the registered workflow name.

  `conda-environment.yml`.

### DIFF
--- a/cylc/flow/option_parsers.py
+++ b/cylc/flow/option_parsers.py
@@ -18,13 +18,26 @@
 import logging
 from optparse import OptionParser, OptionConflictError, Values, Option
 import os
+import re
+from ansimarkup import parse as cparse
 import sys
 from typing import Any, Dict, Optional, List
 
 from cylc.flow import LOG, RSYNC_LOG
 import cylc.flow.flags
 from cylc.flow.loggingutil import CylcLogFormatter
-from cylc.flow.terminal import format_shell_examples
+
+
+def format_shell_examples(string):
+    """Put comments in the terminal "dimished" colour."""
+    return cparse(
+        re.sub(
+            r'^(\s*(?:\$[^#]+)?)(#.*)$',
+            r'\1<dim>\2</dim>',
+            string,
+            flags=re.M
+        )
+    )
 
 
 def verbosity_to_opts(verb: int) -> List[str]:
--- a/cylc/flow/terminal.py
+++ b/cylc/flow/terminal.py
@@ -35,6 +35,7 @@ from cylc.flow.exceptions import CylcError
 from cylc.flow.loggingutil import CylcLogFormatter
 from cylc.flow.parsec.exceptions import ParsecError
 from cylc.flow.pathutil import get_workflow_run_dir
+from cylc.flow.workflow_files import WorkflowFiles
 
 
 # CLI exception message format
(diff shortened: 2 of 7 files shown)


hjoliver:
Good spot @datamel, I hadn't noticed this bug.

cylc/flow/terminal.py
```diff
@@ -199,6 +200,26 @@ def parse_dirty_json(stdout):
...
+    arg = arg.rstrip('/')
+    if not arg.startswith(get_workflow_run_dir('')):
+        workflow_dir = get_workflow_run_dir(arg)
+    else:
+        workflow_dir = arg
+    run_number = re.search(
+        r'(?:run)(\d*$)',
+        os.readlink(workflow_dir)).group(1)
+    # importing here to counter circular dependancy errors.
+    from cylc.flow.workflow_files import WorkflowFiles
```
dwsutherland, line 218: Oh really, workflow_files imports from here? or is it more indirect?

dwsutherland (approved):
LGTM .. Reproduced bug, and this branch fixed it:
image

cylc/flow/terminal.py
datamel: Yes, more indirect. Circle of dependencies can be seen pulling the import to the top... something like ```scripts/__init__ => terminal => workflow_files => remote => options_parser => terminal``` .
I'm just pushing a change to fix this, rather than leaving the import at time of use.

cylc/flow/option_parsers.py
```diff
@@ -18,13 +18,26 @@
...
 import sys
 from typing import Any, Dict, Optional, List
 
 from cylc.flow import LOG, RSYNC_LOG
 import cylc.flow.flags
 from cylc.flow.loggingutil import CylcLogFormatter
-from cylc.flow.terminal import format_shell_examples
+
+
+def format_shell_examples(string):
```
datamel, line 31: @oliver-sanders, you may prefer to have this reverted to it's original home in `terminal.py`, in which case I'll put it back and put the imports for `workflow_files` at time of use, rather than at the top of the file.

oliver-sanders: Prefer, yes, but it's really not important so can leave it like this.

cylc/flow/terminal.py
```diff
@@ -199,6 +189,23 @@ def parse_dirty_json(stdout):
...
+
+    Args:
+        arg (str): flow as entered by user on cli e.g. myflow/runN
+    """
+    arg = arg.rstrip('/')
+    if not arg.startswith(get_workflow_run_dir('')):
+        workflow_dir = get_workflow_run_dir(arg)
+    else:
+        workflow_dir = arg
+    run_number = re.search(  #type: ignore
```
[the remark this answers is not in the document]
oliver-sanders: Shouldn't happen but is technically possible, could add an `if` statement to handle this case.

hjoliver (approved):
:+1:

Issue #3872, opened 2020-10-15
database refactor


oliver-sanders:
> **Note:** Tagged against 8.0.0 as this milestone would be a good time to introduce breaking changes, however, this is not required for 8.0.0 and the interface change is less significant now that we have the GraphQL interface.

> **Note:** We have already made breaking changes to the DB in Cylc8.

> **Note:** We will soon need to start building functionality into the UIS to read the DB to provide offline data.

The run database can get quite large >750MB, note we have two copies of the database so that's 1.5GB. It should be pretty easy to bring this down in size.

Suggestions:

0. Tidy the prereqs table.
    * #3863
1. Drop the checkpoint tables.
    * #3864 
    * These are fairly small so there is little space gain here.
    * Fewer tables would make the database a bit clearer though.
2. Record timestamps as unix epoch integers rather than text fields.
    * This should bring a significant reduction in storage requirement.
    * Sqlite has [date/time functions](https://sqlite.org/lang_datefunc.html) to make DB queries with integer timestamps nicer.
3. Enumerate task/job statuses to store them as integers rather than text fields.
    * Same pros/cons as above
4. Move away from compound primary keys e.g. (name, cycle_point).
    * These text fields are duplicated across multiple tables which is inefficient.
    * E.G. create a table of tasks (id INTEGER AUTONUMBER, name TEXT, cycle_point TEXT)
    * Compound text fields could then be replaced with integers.
    * Querying by task name would still work by table relations.
5. Have a quick check to see if any fields can be removed:
   * For example I'm pretty sure we don't need the `time_created` field in task_states.
6. Consider refactoring task_events and task_jobs
   * These tables contain duplicate information in different formats.
   * The task_events table is functionless.
   * These tables account for 60-70% of memory usage.

Since the DB functionality is pretty self-contained and we don't require upgraders at this point these changes shouldn't be too much work.

hjoliver:
Most or all of the above are desirable, but maybe not on the critical path to Cylc 8 - except that, as you note, now is a good time to make breaking changes. I'd still rather get the 100% critical bits in the bag first though.

kinow:
+1 to Cylc 8 or Cylc 9.

For other databases, like Postgres or Oracle, it would be simpler if the Cylc database supported multiple workflows, instead of one database per workflow.

That would require modifying the database and adding more tables, more relationships. But without doing this, it would be quite hard to implement #3360.

oliver-sanders:
[...]

PR #4238, opened 2021-06-02
Allow cylc hold to hold future tasks that haven't spawned yet

### LINKED ISSUE #3743: Post-SoD task hold semantics

MetRonnie:
These changes close #3743

You can now make Cylc hold future tasks when they spawn (without having to use the `--after` option).

Note this only works for "exact" task identifiers, e.g. `cylc hold my_flow mytask.1234`; using globs (e.g. `'mytask.*'`) or status qualifiers (e.g. `mytask.1234:failed`) will only hold active tasks, not unspawned ones.

**Requirements check-list**
  `conda-environment.yml`.
- [x] (master branch) I have opened a documentation PR at cylc/cylc-doc#249.


cylc/flow/task_pool.py
```diff
@@ -705,14 +714,13 @@ def _get_task_by_id(self, id_):
...
-            itask.state.reset(is_queued=True)
-            # TODO Reset manual trigger flag. One manual trigger queues an
-            # unqueued task, another one triggers a queued task.
-            self.data_store_mgr.delta_task_queued(itask)
-        self.task_queue_mgr.push_tasks(itasks)
+    def queue_task(self, itask: TaskProxy) -> None:
+        """Queue a task that is ready to run."""
+        itask.state.reset(is_queued=True)
+        # TODO Reset manual trigger flag. One manual trigger queues an
+        # unqueued task, another one triggers a queued task.
```
hjoliver, line 721: (FYI this TODO was addressed in #4103 but I evidently neglected to remove the comment ... now done, if you update your branch it will disappear).

cylc/flow/task_queues/independent.py
```diff
@@ -107,15 +106,15 @@ def __init__(self,
...
+    def push_task(self, itask: TaskProxy) -> None:
+        """Push a task to the appropriate queue."""
+        for queue in self.queues.values():
+            queue.push_task(itask)
 
     def release_tasks(self, active: Counter[str]) -> List[TaskProxy]:
         """Release tasks up to the queue limits."""
         released: List[TaskProxy] = []
-        for _, queue in self.queues.items():
+        for queue in self.queues.values():
```
hjoliver, line 117: Huh, not sure what I was thinking there! :grimacing:

hjoliver (changes requested):
You missed one call to `TaskPool.queue_tasks()` that actually queues multiple tasks at once - at the end of config reload - hence the broken functional tests. It's probably fine to queue those tasks one at a time rather than revert your change.

hjoliver:
Generally looks great and works well though :tada:  

We probably need a way to show what future tasks are in the hold list. Maybe just add the info to `cylc dump` output.  Or `cylc hold --show` (but that would make `hold` an instant-gratification query command as well as an asynchronous mutation, not sure if that would be a problem...).

MetRonnie:
> You missed one call to `TaskPool.queue_tasks()` that actually queues multiple tasks at once - at the end of config reload - hence the broken functional tests. It's probably fine to queue those tasks one at a time rather than revert your change.

Agh, wasn't picked up by mypy because the method calling it wasn't type annotated. Fixed and rebased

cylc/flow/task_pool.py
```diff
@@ -705,12 +714,11 @@ def _get_task_by_id(self, id_):
...
-        """Queue tasks that are ready to run."""
-        for itask in itasks:
-            itask.state.reset(is_queued=True)
-            self.data_store_mgr.delta_task_queued(itask)
-        self.task_queue_mgr.push_tasks(itasks)
+    def queue_task(self, itask: TaskProxy) -> None:
+        """Queue a task that is ready to run."""
+        itask.state.reset(is_queued=True)
+        self.data_store_mgr.delta_task_queued(itask)
+        self.task_queue_mgr.push_task(itask)
```
MetRonnie, line 721: Might this be a good idea?
```suggestion
        if itask.state.reset(is_queued=True):
            self.data_store_mgr.delta_task_queued(itask)
            self.task_queue_mgr.push_task(itask)
```

hjoliver: It certainly wouldn't hurt.

cylc/flow/task_pool.py
```diff
@@ -1002,53 +1008,106 @@ def report_unmet_deps(self):
...
+                continue
+            taskdefs = self.config.find_taskdefs(name_str)
+            if not taskdefs:
+                LOG.warning(self.ERR_TMPL_NO_TASKID_MATCH.format(name_str))
+                n_warnings += 1
+                continue
+            if point_str is None:
+                LOG.warning(f"No active instances of task: {item}")
+                n_warnings += 1
+                continue
```
oliver-sanders, line 1093: This function is matching future tasks not active ones?

cylc/flow/workflow_db_mgr.py
```diff
@@ -496,9 +504,23 @@ def put_task_pool(self, pool: 'TaskPool') -> None:
...
+    ) -> None:
+        """Replace the tasks in the tasks_to_hold table."""
+        # As we are replacing the maps each time this is called, there
+        # shouldn't be much cost to calling this multiple times between
+        # processing of the db queue
+        self.db_deletes_map[self.TABLE_TASKS_TO_HOLD] = [{}]
+        self.db_inserts_map[self.TABLE_TASKS_TO_HOLD] = [
+            {"name": name, "cycle": str(point)}
+            for name, point in tasks
+        ]
```
MetRonnie, line 522: Hmm, not sure why I did it this way. There may have been a reason why I didn't only insert/delete the differing tasks, but I can't see it

cylc/flow/task_pool.py
```diff
@@ -1002,53 +1008,106 @@ def report_unmet_deps(self):
                 )
             )
 
+    def hold_active_task(self, itask: TaskProxy) -> None:
+        if itask.state.reset(is_held=True):
+            self.data_store_mgr.delta_task_held(itask)
+        self.tasks_to_hold.add((itask.tdef.name, itask.point))
```
oliver-sanders, line 1014: Active tasks don't need to go into the `tasks_to_hold` set, if they were in the set before I think they should come out?

cylc/flow/task_pool.py
```diff
@@ -1266,33 +1331,45 @@ def spawn_task(
...
+            items: Identifiers for matching task definitions, each with the
+                form "name[.point][:state]" or "[point/]name[:state]".
+                Glob-like patterns will give a warning and be skipped.
+        """
         n_warnings = 0
-        task_items = {}
+        task_items: Dict[Tuple[str, 'PointBase'], 'TaskDef'] = {}
         for item in items:
-            point_str, name_str = self._parse_task_item(item)[:2]
+            point_str, name_str, status = self._parse_task_item(item)
```
oliver-sanders, line 1347: Status unused?

oliver-sanders (commented):
Looks great.

cylc/flow/workflow_db_mgr.py
oliver-sanders: I think this might be the approach used for other tables?

For the moment I've got my fingers in my ears going lalala, the DB usage needs some major rationalisation:

* Repeated wiping and rebuilding of tables.
* Repeated creation and destruction of the DB connection.
* Un-normalised tables.

See https://github.com/cylc/cylc-flow/issues/3872, https://github.com/cylc/cylc-flow/issues/3666.

At the moment it's messy but working and we have bigger things to worry about so I think this work will slide back a while to come.

cylc/flow/task_pool.py
MetRonnie: It's a bit confusing I know. This function relies on you having previously filtered out the matching active tasks using `self.filter_task_proxies()`. So if the point is not specified e.g. `mytask`, equivalent to `mytask.*`, and there is no active instance of `mytask`, then I say that is the correct warning to give.

cylc/flow/task_pool.py
MetRonnie: They don't need to, but I think it's easier to deal with if everything that is held or is going to be held is in the set. They should only need to come out when released.

cylc/flow/task_pool.py
```diff
@@ -1004,53 +1012,106 @@ def report_unmet_deps(self):
...
+        """Match future tasks based on the assumption that no active tasks
+        matched the items.
+
+        Returns (n_warnings, matched_tasks)
+        """
+        n_warnings = 0
+        matched_tasks: Set[Tuple[str, 'PointBase']] = set()
+        for item in items:
+            point_str, name_str, status = self._parse_task_item(item)
+            if status or ('*' in item) or ('?' in item) or ('[' in item):
```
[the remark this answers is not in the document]
MetRonnie: I'm not sure what you mean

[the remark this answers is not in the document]
oliver-sanders: Because globs in the cycle point and name should not be allowed for matching future tasks.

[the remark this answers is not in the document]
MetRonnie: This checks if a glob character is in any part of the `name[.point][:status]` string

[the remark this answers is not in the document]
oliver-sanders: Ah, sorry, missread, item includes all three 🤦.

hjoliver (approved):
Nice, this one is quiet important, for users :+1:

oliver-sanders:
Testing is going well, I've hit one issue with the n=1 data store which I had a poke at as I wasn't sure how easy/possible it was to shim, turns out actually quite easy. Here's a PR, feel free to use it or re-work it https://github.com/MetRonnie/cylc-flow/pull/6.

oliver-sanders:
I think some tests have been angered by the recent conftest.py change?

oliver-sanders (approved):
Code looks good, testing went fine, DB housekeeping ok.

Found the following quirks:

* Completed tasks remain in `tasks_to_hold`.

  E.G. If you hold a task then trigger it it will remain in `tasks_to_hold` even after it has passed out of the n=1 window. These stuck tasks can be removed by `cylc release <flow> --all`.

  It's a quirk not a bug, I don't think it's a problem or worth the fuss handling differently.

* Behaviour with reflow.

  I think we will want to add the "flow label" into the `tasks_to_hold` set in order to handle the situation where we have the same task from different flows. E.G. the user might want to hold one flow but the another @hjoliver 

  Easy change to make, happy to do now or later.

* Need to bodge TUI and GUI to handle future families

  > Follow-on not for this PR

  Because globing/matching is not allowed in the n>0 window holding/releasing tasks in the n>0 window from Tui/Gui doesn't work. Easy fix, we should get Tui/Gui to expand the cycle/family to the full list of tasks in the selection.

  Opened:

  * https://github.com/cylc/cylc-flow/issues/4246
  * https://github.com/cylc/cylc-ui/issues/686

oliver-sanders:
(will leave @hjoliver with the reflow comment above)

MetRonnie:
I have also just noticed:
- If you start a simple workflow (`P1 = foo`) paused, then hold all active tasks, then resume; `foo.1` will apparently ignore the held state and will run

```
INFO - Processing 1 queued command(s)
INFO - Command succeeded: hold(['foo.*'])
INFO - Processing 1 queued command(s)
INFO - RESUMING the workflow now
INFO - Command succeeded: resume()
INFO - Queue released: foo.1
INFO - [foo.1] -submit-num=01, host=hostymchostface
INFO - [foo.1] -triggered off []
INFO - [foo.1] status=preparing (held): (internal)submitted at 2021-06-07  for job(01) flow(E)
INFO - [foo.1] -health check settings: submission timeout=None, polling intervals=PT15M,...
INFO - [foo.1] status=submitted (held): (received)started at 2021-06-07  for job(01) flow(E)
INFO - [foo.1] -health check settings: execution timeout=None, polling intervals=PT15M,...
INFO - [foo.1] status=running (held): (received)succeeded at 2021-06-07  for job(01) flow(E)
```
I think this can be fixed in a separate bugfix PR

MetRonnie:
> E.G. If you hold a task then trigger it it will remain in tasks_to_hold even after it has passed out of the n=1 window. These stuck tasks can be removed by `cylc release <flow> --all`.

Hmm, perhaps it should be that triggering a task also releases it?

oliver-sanders:
I think we don't do that so that tasks can be triggered in a way that prevents retries from occurring?

There are other ways we could end up with held tasks drifting outside of the `n=1` window (e.g. suicide triggers). I think the solution would be to remove the task from `tasks_to_hold` when the task is removed from the pool.

MetRonnie:
Another issue, which I realised from Oliver's comment https://github.com/cylc/cylc-flow/issues/4246#issuecomment-855821083:
- If you have a family `A` that has at cycle point 1:
    - a (n=0)
    - b (n=1)
 
   then doing `cylc hold myflow A.1` will only hold `a.1`, and not set `b.1` to be held. (If `a.1` was n=1 instead, then holding `A.1` would set both `a.1` and `b.1` to be held)

MetRonnie:
(Bash test failure is due to Docker container initialisation stalling, don't think it's worth re-running)

hjoliver:
> Behaviour with reflow.
> 
> I think we will want to add the "flow label" into the tasks_to_hold set in order to handle the situation where we have the same
> task from different flows. E.G. the user might want to hold one flow but the another @hjoliver

Yeah, I'll post a follow-up issue .... #4277

hjoliver (approved):
Looks great, no problems found :tada:

Issue #4239, opened 2021-06-03
flake8: simplify


oliver-sanders:
The [flake8-simplify](https://github.com/MartinThoma/flake8-simplify) plugin helps you write better code (I use it in my editor for that purpose).

I think this is a useful tool which would help automate code review, here are some examples of its recommendations - https://github.com/MartinThoma/flake8-simplify#sim102

Raising this as a question as it is somewhat more verbose than plain flake8.

hjoliver:
I'd be happy to use this.  @cylc/core - anyone disagree?

kinow:
Keen to try it :+1:

MetRonnie:
I'll have a go with it

MetRonnie:
Note: to run flake8 without it you can do `flake8 --extend-ignore=SIM`

oliver-sanders:
Yeah there were a fair few violations, I looked through the first so many and the corrections looked sensible, the sort of thing we usually haggle over in code review.

If you install flake8-simplify you should be able to use it in your text editor without any extra config (is how I run it).

hjoliver:
I'm using it in vim now, very useful.

PR #4241, opened 2021-06-03
Flake8

oliver-sanders:
> Recent code review about iteration stuff made me think back to this, I'm sure we used to have checks for this sort of thing...

When we moved away from pylint we lost a lot of useful (and some not so useful) linting.

This PR puts back a good chunk of that updated a little and less pylint'y. You can get as-you-type linting in your text editor:
* For VSCode use the Python plugin thinggy.

  ```ini
  # config
  "python.linting.flake8Enabled": true,  
  "python.linting.flake8Path": "path/to/flake8",  
  ```

* For Vim use [Ale](https://github.com/dense-analysis/ale).

  ```vim
  " config
  Plugin 'dense-analysis/ale'
  let g:ale_linters = {'python': ['flake8']}
  let g:ale_python_flake8_executable = '/path/to/flake8'
  ```

* For Emacs use Vim.

  ```bash
  # config
  alias emacs='vim'
  ```

I've fixed the violations including a few bugs.

There are more plugins (including flake8-simplify #4239 [recommend installing]), however, keeping this simple and, hopefully, uncontroversial.

**Requirements check-list**


cylc/flow/parsec/fileparse.py
```diff
@@ -362,7 +362,7 @@ def read_and_proc(fpath, template_vars=None, viewcfg=None, asedit=False):
             LOG.debug('Processing with EmPy')
             try:
                 from cylc.flow.parsec.empysupport import empyprocess
-            except (ImportError, ModuleNotFoundError):
+            except ImportError:
```
oliver-sanders, line 365: ModuleNotFoundError is a subclass of ImportError

cylc/flow/scheduler.py
```diff
@@ -933,13 +933,13 @@ def command_set_verbosity(lvl):
         except (TypeError, ValueError):
             return
         if lvl <= logging.DEBUG:
-            cylc.flow.flags.verbosity == 2
+            cylc.flow.flags.verbosity = 2
```
oliver-sanders, line 936: 🍌

cylc/flow/terminal.py
```diff
@@ -126,7 +126,7 @@ def ansi_log(name='cylc', stream='stderr'):
     stream_name = f'<{stream}>'
     for handler in logging.getLogger(name).handlers:
         if (
-            getattr(handler, 'formatter')
+            getattr(handler, 'formatter', None)
```
oliver-sanders, line 129: Genuine bug, must specify a default for `getattr` to swallow the `AttributeError`:

```python
>>> getattr(None, 'x')
AttributeError: 'NoneType' object has no attribute 'x'
```

etc/bin/defn-order-test.py
```diff
@@ -1,62 +0,0 @@
-# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
```
oliver-sanders, line 1: This script was broken, it hasn't been run for a while (because it is broken) and nonone has complained so guessing its good to go.

tox.ini
```diff
@@ -6,6 +6,8 @@ ignore=
     W503,
     ; line break after binary operator
     W504
+    ; `.next()` is not a thing on Python 3 (it is for TaskActionTimer)
+    B305
```
oliver-sanders, line 10: This rule checks for `.next()` calls which are a common 2to3 error, however, it checks somewhat naively, our usage is valid, don't think the rule is worth it.

hjoliver:
>  For Emacs use Vim.
> 
> ```
> # config
> alias emacs='vim'
> ```

:rofl: comedy gold

cylc/flow/config.py
```diff
@@ -925,7 +925,7 @@ def _check_circular_helper(x2ys, y2xs):
         """
         # Starter x nodes are those with no incoming, i.e.
         # x nodes that never appear as a y.
-        sxs = set(x01 for x01 in x2ys if x01 not in y2xs)
+        sxs = {x01 for x01 in x2ys if x01 not in y2xs}
```
MetRonnie, line 928: While we're at it, this feels more elegant., and performance is slightly better from a couple of tests I've tried
```suggestion
        sxs = set(x2ys).difference(y2xs)
```

cylc/flow/config.py
```diff
@@ -1268,9 +1268,11 @@ def configure_workflow_state_polling_tasks(self):
             if name not in self.workflow_polling_tasks:
                 continue
             rtc = tdef.rtconfig
-            comstr = "cylc workflow-state" + \
-                     " --task=" + tdef.workflow_polling_cfg['task'] + \
-                     " --point=$CYLC_TASK_CYCLE_POINT"
+            comstr = (
+                "cylc workflow-state"
+                + " --task=" + tdef.workflow_polling_cfg['task']
+                + " --point=$CYLC_TASK_CYCLE_POINT"
```
MetRonnie, line 1274:
```suggestion
                "cylc workflow-state "
                f"--task={tdef.workflow_polling_cfg['task']} "
                f"--point=$CYLC_TASK_CYCLE_POINT"
```

cylc/flow/task_state.py
```diff
@@ -388,8 +391,11 @@ def get_resolved_dependencies(self):
         task with lots of near-simultaneous triggers.
 
         """
-        return list(sorted(dep for prereq in self.prerequisites for dep in
-                           prereq.get_resolved_dependencies()))
+        return sorted(
+            dep
+            for prereq in self.prerequisites
+            for dep in prereq.get_resolved_dependencies()
```
MetRonnie, line 397: likewise
```suggestion
            for dep in prereq.get_resolved_dependencies()
            for prereq in self.prerequisites
```

cylc/flow/terminal.py
MetRonnie: Could use `hasattr`?

MetRonnie (approved):
Nice, I have left some suggestions while we're looking at these lines

cylc/flow/task_state.py
```diff
@@ -349,8 +349,11 @@ def suicide_prerequisites_all_satisfied(self):
 
     def prerequisites_get_target_points(self):
         """Return a list of cycle points targeted by each prerequisite."""
-        return set(point for prerequisite in self.prerequisites for
-                   point in prerequisite.get_target_points())
+        return {
+            point
+            for prerequisite in self.prerequisites
+            for point in prerequisite.get_target_points()
+        }
```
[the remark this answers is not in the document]
oliver-sanders: That would be an error:

```python
>>> prereqs = [[1]]
>>> {point for prereq in prereqs for point in prereq}
{1}
>>> {point for point in prereq for prereq in prereqs}
NameError: name 'prereq' is not defined
```

Some explanation on this SO https://stackoverflow.com/a/38321692

Note the Python way the comprehension order is the same as it would be for a nested for loop:

```python
# list comp
return {
    point
    for point in prerequisite.get_target_points()
    for prerequisite in self.prerequisites
}

# for loop
for point in prerequisite.get_target_points():
    for prerequisite in self.prerequisites:
        points.append(point)
```

cylc/flow/task_state.py
oliver-sanders: as above

cylc/flow/terminal.py
oliver-sanders: Could indeed.

cylc/flow/platforms.py
```diff
@@ -519,7 +524,7 @@ def get_all_platforms_for_install_target(
     """Return list of platform dictionaries for given install target."""
     platforms: List[Dict[str, Any]] = []
     all_platforms = glbl_cfg(cached=True).get(['platforms'], sparse=False)
-    for k, v in all_platforms.iteritems():
+    for k, v in all_platforms.items():
```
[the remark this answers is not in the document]
oliver-sanders: Ooh, good spot, have to add a `noqa` flag for each use (which turns out to only be once!).

cylc/flow/network/schema.py
```diff
@@ -590,7 +592,7 @@ async def resolve_broadcasts(root, info, **args):
                 continue
             t_args['ids'] = [(None, None, None, name, None, None)]
             tasks = await resolvers.get_nodes_all(t_type, t_args)
-            for namespace in set(ns for t in tasks for ns in t.namespace):
+            for namespace in {ns for t in tasks for ns in t.namespace}:
```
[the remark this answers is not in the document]
oliver-sanders: as below

cylc/flow/task_state.py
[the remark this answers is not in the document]
MetRonnie: Ugh, clearly I need to unlearn a bad habit I've picked up from somewhere

hjoliver (approved):
All good :+1:

PR #4249, opened 2021-06-08
Fix mistake in workflow_files

MetRonnie:

Call to `Path.exists` was missing parentheses.

**Requirements check-list**
  `conda-environment.yml`.
- [x] Does not need tests (small fix).
- [x] No change log entry required (small fix).

### DIFF
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -567,7 +567,7 @@ def register(
     return flow_name
 
 
-def is_installed(path):
+def is_installed(path: Union[Path, str]) -> bool:
     """Check to see if the path sent contains installed flow.
 
     Checks for valid _cylc-install directory in current folder and checks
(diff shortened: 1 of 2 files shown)


cylc/flow/workflow_files.py
```diff
@@ -1235,32 +1233,30 @@ def get_run_dir_info(
...
     for entry in Path(run_path_base).iterdir():
-        isNumbered = bool(re.search(r'^run\d+$', entry.name))
+        is_numbered = bool(re.search(r'^run\d+$', entry.name))
         if (
             entry.is_dir()
-            and entry.name not in [
+            and entry.name not in {
                 WorkflowFiles.Install.DIRNAME, WorkflowFiles.RUN_N
-            ]
+            }
```
MetRonnie, line 1254: Something I recently learned is that checking for membership in a set is faster than checking in a list. (Not that it will realistically make any difference in this case)

datamel (approved):
Thanks @MetRonnie

MetRonnie:
1 review should do

Issue #4219, opened 2021-05-18
symlink dirs: move into [install] section


oliver-sanders:
The `[symlink dirs]` section applies only to the `cylc install` command, propose moving it into the `[install]` section.

For reference the config tree currently looks like so:

```python
>>> from cylc.flow.cfgspec.globalcfg import SPEC
>>> print(SPEC.tree())
global.cylc
    ...
    [install]
        source dirs
    ...
    [symlink dirs]
        [<install target>]
            run
            log
            share
            share/cycle
            work
```

**Pull requests welcome!**

MetRonnie:
So it would be `[install][symlink dirs][<install target>]...`? Can't see any reason not to

dpmatthews:
> The [symlink dirs] section applies only to the cylc install command

Also used by `cylc play` (as part of remote init)

datamel:
In my mind, I think it makes sense to move `[symlink dirs]` into the install section, even though it is used by cylc play. I can see see the `install` section as a setup type section, that deals with installation of files for a workflow. Rather than being only strictly used for `cylc install`?

PR #4250, opened 2021-06-08
Symlink dirs cli

datamel:
Adds cli option (`--symlink-dirs`) for over-riding localhost symlink dirs in global config when executing `cylc install`. 
Removes --no-symlink-dirs option, this is replaced by --symlink-dirs=None

Also moves symlink dirs into install section of config...(these changes close https://github.com/cylc/cylc-flow/issues/4219)

  `conda-environment.yml`.

- [x] (master branch) I have opened a documentation PR at https://github.com/cylc/cylc-doc/pull/252

### DIFF
--- a/cylc/flow/pathutil.py
+++ b/cylc/flow/pathutil.py
@@ -142,14 +142,17 @@ def make_localhost_symlinks(
     Returns:
         Dictionary of symlinks with sources as keys and
         destinations as values: ``{source: destination}``
+        symlink_conf: Symlinks dirs configuration passed from cli
 
     """
+    symlinks_created = {}
     dirs_to_symlink = get_dirs_to_symlink(
         get_localhost_install_target(),
         named_sub_dir, symlink_conf=symlink_conf
     )
-    symlinks_created = {}
     for key, value in dirs_to_symlink.items():
+        if value is None:
+            continue
         if key == 'run':
             symlink_path = rund
         else:
--- a/cylc/flow/pathutil.py
+++ b/cylc/flow/pathutil.py
@@ -19,7 +19,7 @@ import os
 from pathlib import Path
 import re
 from shutil import rmtree
-from typing import Dict, Iterable, Set, Union, Optional, OrderedDict, Any
+from typing import Dict, Iterable, Set, Union, Optional, Any
 
 from cylc.flow import LOG
 from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
@@ -131,7 +131,7 @@ def make_workflow_run_tree(workflow):
 
 def make_localhost_symlinks(
     rund: Union[Path, str],
-    named_sub_dir: str, 
+    named_sub_dir: str,
     symlink_conf: Optional[Dict[str, Dict[str, str]]] = None
 ) -> Dict[str, Union[Path, str]]:
     """Creates symlinks for any configured symlink dirs from glbl_cfg.
@@ -190,7 +190,7 @@ def get_dirs_to_symlink(
         dirs_to_symlink: [directory: symlink_path]
     """
     dirs_to_symlink: Dict[str, Any] = {}
-    if not symlink_conf:
+    if symlink_conf is None:
         symlink_conf = glbl_cfg().get(['install', 'symlink dirs'])
     if install_target not in symlink_conf.keys():
         return dirs_to_symlink
@@ -199,7 +199,7 @@ def get_dirs_to_symlink(
         dirs_to_symlink['run'] = os.path.join(base_dir, 'cylc-run', flow_name)
     for dir_ in ['log', 'share', 'share/cycle', 'work']:
         link = symlink_conf[install_target].get(dir_, None)
-        if (not link) is None or link == base_dir:
+        if (not link) or link == base_dir:
             continue
         dirs_to_symlink[dir_] = os.path.join(link, 'cylc-run', flow_name, dir_)
     return dirs_to_symlink
--- a/cylc/flow/pathutil.py
+++ b/cylc/flow/pathutil.py
@@ -138,11 +138,11 @@ def make_localhost_symlinks(
     Args:
         rund: the entire run directory path
         named_sub_dir: e.g flow_name/run1
+        symlink_conf: Symlinks dirs configuration passed from cli
 
     Returns:
         Dictionary of symlinks with sources as keys and
         destinations as values: ``{source: destination}``
-        symlink_conf: Symlinks dirs configuration passed from cli
 
     """
     symlinks_created = {}
@@ -215,7 +215,7 @@ def make_symlink(path: Union[Path, str], target: Union[Path, str]) -> bool:
     path = Path(path)
     target = Path(target)
     if path.exists():
-        if path.is_symlink() and path.samefile(target):
+        if path.is_symlink() and target.exists() and path.samefile(target):
             # correct symlink already exists
             return False
         # symlink name is in use by a physical file or directory
@@ -232,6 +232,11 @@ def make_symlink(path: Union[Path, str], target: Union[Path, str]) -> bool:
             raise WorkflowFilesError(
                 f"Error when symlinking. Failed to unlink bad symlink {path}.")
     target.mkdir(parents=True, exist_ok=True)
+
+    # This is needed in case share and share/cycle have the same symlink dir:
+    if path.exists():
+        return False
+
     path.parent.mkdir(parents=True, exist_ok=True)
     try:
         path.symlink_to(target)
--- a/cylc/flow/pathutil.py
+++ b/cylc/flow/pathutil.py
@@ -215,6 +215,8 @@ def make_symlink(path: Union[Path, str], target: Union[Path, str]) -> bool:
     path = Path(path)
     target = Path(target)
     if path.exists():
+        # note all three checks are needed here due to case where user has set
+        # their own symlink which does not match the global config set one.
         if path.is_symlink() and target.exists() and path.samefile(target):
             # correct symlink already exists
             return False
--- a/cylc/flow/pathutil.py
+++ b/cylc/flow/pathutil.py
@@ -28,7 +28,6 @@ from cylc.flow.exceptions import (
 )
 from cylc.flow.platforms import get_localhost_install_target
 
-
 # Note: do not import this elsewhere, as it might bypass unit test
 # monkeypatching:
 _CYLC_RUN_DIR = os.path.join('$HOME', 'cylc-run')
@@ -131,7 +130,7 @@ def make_workflow_run_tree(workflow):
 
 
 def make_localhost_symlinks(
-    rund: Union[Path, str], named_sub_dir: str
+    rund: Union[Path, str], named_sub_dir: str, symlink_conf=None
 ) -> Dict[str, Union[Path, str]]:
     """Creates symlinks for any configured symlink dirs from glbl_cfg.
     Args:
@@ -144,7 +143,9 @@ def make_localhost_symlinks(
 
     """
     dirs_to_symlink = get_dirs_to_symlink(
-        get_localhost_install_target(), named_sub_dir)
+        get_localhost_install_target(),
+        named_sub_dir, symlink_conf=symlink_conf
+    )
     symlinks_created = {}
     for key, value in dirs_to_symlink.items():
         if key == 'run':
@@ -165,14 +166,14 @@ def make_localhost_symlinks(
     return symlinks_created
 
 
-def get_dirs_to_symlink(install_target: str, flow_name: str) -> Dict[str, str]:
+def get_dirs_to_symlink(install_target: str, flow_name: str, symlink_conf=None) -> Dict[str, str]:
     """Returns dictionary of directories to symlink from glbcfg.
 
     Note the paths should remain unexpanded, to be expanded on the remote.
     """
     dirs_to_symlink: Dict[str, str] = {}
-    symlink_conf = glbl_cfg().get(['symlink dirs'])
-
+    if not symlink_conf:
+        symlink_conf = glbl_cfg().get(['install', 'symlink dirs'])
     if install_target not in symlink_conf.keys():
         return dirs_to_symlink
     base_dir = symlink_conf[install_target]['run']
@@ -213,10 +214,6 @@ def make_symlink(path: Union[Path, str], target: Union[Path, str]) -> bool:
             raise WorkflowFilesError(
                 f"Error when symlinking. Failed to unlink bad symlink {path}.")
     target.mkdir(parents=True, exist_ok=True)
-    if path.exists():
-        # Trying to link to itself; no symlink needed
-        # (e.g. path's parent is symlink to target's parent)
-        return False
     path.parent.mkdir(parents=True, exist_ok=True)
     try:
         path.symlink_to(target)
--- a/cylc/flow/pathutil.py
+++ b/cylc/flow/pathutil.py
@@ -19,7 +19,7 @@ import os
 from pathlib import Path
 import re
 from shutil import rmtree
-from typing import Dict, Iterable, Set, Union
+from typing import Dict, Iterable, Set, Union, Optional, OrderedDict, Any
 
 from cylc.flow import LOG
 from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
@@ -130,7 +130,9 @@ def make_workflow_run_tree(workflow):
 
 
 def make_localhost_symlinks(
-    rund: Union[Path, str], named_sub_dir: str, symlink_conf=None
+    rund: Union[Path, str],
+    named_sub_dir: str, 
+    symlink_conf: Optional[Dict[str, Dict[str, str]]] = None
 ) -> Dict[str, Union[Path, str]]:
     """Creates symlinks for any configured symlink dirs from glbl_cfg.
     Args:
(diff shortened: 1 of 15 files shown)


cylc/flow/workflow_files.py
```diff
@@ -1341,6 +1343,43 @@ def validate_source_dir(source, flow_name):
...
+    if symlink_dirs == 'None':
+        return symdict
+    symlist = symlink_dirs.replace(" ", "").strip(',').split(',')
+    for pair in symlist:
+        try:
+            key, val = pair.split("=")
+        except ValueError:
+            return symdict
+        if key in ['run', 'log', 'share', 'share/cycle', 'work']:
+            symdict['localhost'][key] = val
```
oliver-sanders, line 1375: Just incase users try to "unset" a symlink dir by doing `<dir>=` it might be worth swapping the empty string for `None` (which is think is handled).

```suggestion
            symdict['localhost'][key] = val.strip() or None
```

hjoliver:
>  Removes `--no-symlink-dirs` option, this is replaced by `--symlink-dirs=None`

I'm not sure about this.  The original `--no-symlink-dirs` is pretty unambiguous and intuitive, but `--symlink-dirs=None` kind of exposes Python internals through the CLI and is less intuitive. Users would more likely guess `--symlink-dirs=""` (empty string), which does nothing.

hjoliver:
I'm slight confused about how this works (from a quick test, without examining the code).  If I set (e.g.) `log` and `work` symlinks for `localhost` in `global.cylc`, both symlinks get created by `cylc install`.  If I then override `log` using `cylc install --symlink-dirs`, I find that `cylc install` creates only the overridden `log` symlink, and the global config `work` one gets created later by `cylc play` ??

oliver-sanders (commented):
Tested, looks good:

```console
$ cylc config -i '[install][symlink dirs]'
[[[localhost]]]
    log = $HOME/cylc-log
    run = $HOME/cylc-runn
    share = 
    share/cycle = 
    work = 
$ cylc install one 
INSTALLED one/run1 from .../one
$ cylc install one --symlink-dirs=
INSTALLED one/run2 from .../one
$ cylc install one --symlink-dirs=log=$HOME/cylc-log
INSTALLED one/run3 from .../one
$ tree -l
.
├── _cylc-install
│   └── source -> .../one
│       └── suite.rc
├── run1 -> /cylc-runn/cylc-run/one/run1
│   ├── flow.cylc -> suite.rc
│   ├── log -> /cylc-log/cylc-run/one/run1/log
│   │   ├── install
│   │   │   └── 20210616T102225Z-install.log
│   │   └── version
│   │       ├── uncommitted.diff
│   │       └── vcs.conf
│   └── suite.rc
├── run2
│   ├── flow.cylc -> suite.rc
│   ├── log
│   │   ├── install
│   │   │   └── 20210616T102232Z-install.log
│   │   └── version
│   │       ├── uncommitted.diff
│   │       └── vcs.conf
│   └── suite.rc
├── run3
│   ├── flow.cylc -> suite.rc
│   ├── log -> /cylc-log/cylc-run/one/run3/log
│   │   ├── install
│   │   │   └── 20210616T102253Z-install.log
│   │   └── version
│   │       ├── uncommitted.diff
│   │       └── vcs.conf
│   └── suite.rc
└── runN -> run3  [recursive, not followed]
```


Tested again overriding run and work, the `cylc play` overriding has been addressed.

cylc/flow/pathutil.py
```diff
@@ -161,21 +168,34 @@ def make_localhost_symlinks(rund, named_sub_dir):
...
-    symlink_conf = glbl_cfg().get(['symlink dirs'])
+def get_dirs_to_symlink(
+    install_target: str,
+    flow_name: str,
+    symlink_conf: Optional[Dict[str, Dict[str, Any]]] = None
+) -> Dict[str, Any]:
+    """Parses symlink dirs to the correct format, ready for symlinking.
+
+    Args:
+        install_target (str): Symlinks to be created on this install target
```
oliver-sanders, line 179: Don't need to add the types in the docstrings if using type hints.

cylc/flow/workflow_files.py
```diff
@@ -1090,7 +1105,7 @@ def install_workflow(
     source: Optional[Union[Path, str]] = None,
     run_name: Optional[str] = None,
     no_run_name: bool = False,
-    no_symlinks: bool = False
+    cli_symlink_dirs: Optional[Union[Dict[str, Dict[str, Any]], None]] = None
```
oliver-sanders, line 1108: I think the `Union` isn't doing anything here:

```diff
 Optional[
-    Union[
          Dict[str, Dict[str, Any]],
         None
-    ]
 ]
```

datamel:
> I'm slight confused about how this works (from a quick test, without examining the code). If I set (e.g.) `log` and `work` symlinks for `localhost` in `global.cylc`, both symlinks get created by `cylc install`. If I then override `log` using `cylc install --symlink-dirs`, I find that `cylc install` creates only the overridden `log` symlink, and the global config `work` one gets created later by `cylc play` ??

Sorry about that Hilary. Chatted this through with the team this end and have implemented the following changes. 
- If `cylc play`-ing an installed flow, localhost symlink dirs will now be skipped. 
- Missing symlink dirs entered on the command line will not fill from global config.
- `--symlink-dirs=""` is now the interface for skipping symlink dirs when installing flows.

The dropping of `--no-symlink-dirs` was an effort to keep the cli concise, rather than bloating with too many options.

Hope this is ok, happy to make changes if you wish.

oliver-sanders (approved):

cylc/flow/scripts/install.py
```diff
@@ -105,6 +105,18 @@ def get_option_parser():
         default=None,
         dest="source")
 
+    parser.add_option(
+        "--symlink-dirs",
+        help=(
+            "Enter a list, in the form 'log=path/to/store, share = $...'"
+            ". Use this option to override local symlinks for directories run,"
+            " log, work, share, share/cycle, as configured in global.cylc. "
+            "Enter an empty list \"\" to skip making localhost symlink dirs."
```
hjoliver, line 114:
```suggestion
            "Enter a list, in the form 'log=path/to/log-dir, share=path/to/share-dir, ...'"
            ". Use this option to override local symlinks for directories run,"
            " log, work, share, share/cycle, as configured in global.cylc. "
            "Paths may contain environment variables.  Any use of this "
            "option overrides all global symlink configuration. "
            "Enter an empty list \"\" to skip making localhost symlink dirs."
```

Also, from testing, if I use a space instead of comma to separate `log=..., share=...` (e.g.),  `cylc install` silently fails to make any symlinks. We should probably support comma or space, or else explicitly fail plain spaces to avoid the parsing complication of filenames containing spaces.

hjoliver: And regarding this (my wording):
> Any use of this option overrides all global symlink configuration.

I presume that was a deliberate choice?  What's the argument for not having individual path override?

cylc/flow/workflow_files.py
```diff
@@ -556,26 +559,38 @@ def register(
...
     check_flow_file(source, symlink_suiterc=True, logger=None)
-    symlinks_created = make_localhost_symlinks(
-        get_workflow_run_dir(flow_name), flow_name)
-    if bool(symlinks_created):
-        for src, dst in symlinks_created.items():
-            LOG.info(f"Symlink created from {src} to {dst}")
+    if not is_installed(get_workflow_run_dir(flow_name)):
+        symlinks_created = make_localhost_symlinks(
+            get_workflow_run_dir(flow_name), flow_name)
+        if bool(symlinks_created):
```
hjoliver, line 565:
```suggestion
        if symlinks_created:
```
No reason to use explicit `bool()` here, is there?

cylc/flow/workflow_files.py
```diff
@@ -556,26 +559,38 @@ def register(
...
-    source = Path(cylc_install_folder, WorkflowFiles.Install.SOURCE)
-    return cylc_install_folder.is_dir() and source.is_symlink()
+    rund = Path(rund)
+    cylc_install_dir = Path(rund, WorkflowFiles.Install.DIRNAME)
+    alt_cylc_install_dir = Path(rund.parent, WorkflowFiles.Install.DIRNAME)
+    if cylc_install_dir.is_dir():
+        return True
+    elif alt_cylc_install_dir.is_dir():
+        return True
+    return False
```
hjoliver, line 593:
```suggestion
    return cylc_install_dir.is_dir() or alt_cylc_install_dir.is_dir():
```

cylc/flow/workflow_files.py
```diff
@@ -1145,10 +1159,10 @@ def install_workflow(
         named_run = os.path.join(named_run, run_name)
     elif run_num:
         named_run = os.path.join(named_run, f'run{run_num}')
-    if not no_symlinks:
-        symlinks_created = make_localhost_symlinks(rundir, named_run)
+    symlinks_created = make_localhost_symlinks(
+        rundir, named_run, symlink_conf=cli_symlink_dirs)
     install_log = _get_logger(rundir, 'cylc-install')
-    if not no_symlinks and bool(symlinks_created) is True:
+    if bool(symlinks_created) is True:
```
hjoliver, line 1165:
```suggestion
    if symlinks_created:
```

cylc/flow/workflow_files.py
```diff
@@ -1337,6 +1351,44 @@ def validate_source_dir(source, flow_name):
...
+            e.g. {'localhost': {'log': '$DIR',
+                                'share': '$DIR2'
+                                }
+                }
+    """
+    # Ensures the same nested dict format which is returned by the glb cfg
+    symdict: Dict[str, Dict[str, Any]] = {'localhost': {'run': None}}
+    if symlink_dirs == "":
+        return symdict
+    symlist = symlink_dirs.replace(" ", "").strip(',').split(',')
```
hjoliver, line 1377: (See comment above on accidental (or assumed) use of spaces instead of commas).

cylc/flow/workflow_files.py
```diff
@@ -1337,6 +1351,44 @@ def validate_source_dir(source, flow_name):
...
+    for pair in symlist:
+        try:
+            key, val = pair.split("=")
+        except ValueError:
+            continue
```
hjoliver, line 1382: Better to fail this as a command line error (it's just a command, so can be redone easily with no typos in it)?

hjoliver (commented):
Looks great, tests as working, just a few minor suggestions and questions.

cylc/flow/scripts/install.py
datamel: This was a request from @dpmatthews. We discussed did discuss other options here, with the  suggestion from @oliver-sanders that we could add a further option in time, to fill links from config if users wanted it.

dpmatthews: My reasoning is that this way it's clear to the user what they will get. Otherwise, if I wanted to configure my workflow to use a single filesystem I'd need to either know the global config setup to know what to override or I'd need to set all the possible symlink settings just to be sure. We could support some method of choosing whether to do the individual path override but that seems like unnecessary complexity for an option which we don't expect to be widely used.

cylc/flow/workflow_files.py
datamel: I have implemented explicitly failing plain spaces, due to the complications with file names containing spaces.

MetRonnie (commented):
I think this is the PR you wanted me to check after #4237 was merged

cylc/flow/pathutil.py
```diff
@@ -213,10 +232,6 @@ def make_symlink(path: Union[Path, str], target: Union[Path, str]) -> bool:
             raise WorkflowFilesError(
                 f"Error when symlinking. Failed to unlink bad symlink {path}.")
     target.mkdir(parents=True, exist_ok=True)
-    if path.exists():
-        # Trying to link to itself; no symlink needed
-        # (e.g. path's parent is symlink to target's parent)
-        return False
```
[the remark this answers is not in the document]
datamel: Line 226 has the the return false.

tests/unit/conftest.py
```diff
@@ -65,8 +65,9 @@ def tmp_run_dir(tmp_path: Path, monkeypatch: pytest.MonkeyPatch):
 
     Args:
         reg: Workflow name.
+        installed: named or no-name. Creates
```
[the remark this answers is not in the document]
datamel: oopsie, thanks

cylc/flow/pathutil.py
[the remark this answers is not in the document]
MetRonnie: But this checks if the path exists _after_ the target is created.
1. When the `~/cylc-run/foo/share -> ~/sym-share/cylc-run/foo/share` symlink & target already exist, and we're now on `share/cycle`
2. The line `target.mkdir(...)` above will create `~/sym-share/cylc-run/foo/share/cycle`.
3. At this point `path` exists now - `path` is `~/cylc-run/foo/share/cycle`, but because `~/cylc-run/foo/share` is a symlink, `path` is actually  `~/sym-share/cylc-run/foo/share/cycle`

(Sorry, the comment in the code doesn't give a good explanation I now realise! And I should have created a test case for `share` and `share/cycle` being the same)

cylc/flow/workflow_files.py
```diff
@@ -1491,6 +1502,51 @@ def validate_source_dir(source, flow_name):
     check_flow_file(source, logger=None)
 
 
+def get_sym_dirs(symlink_dirs: str) -> Dict[str, Dict[str, Any]]:
```
MetRonnie, line 1505: Suggest calling this something like `parse_comma_sep_symlink_dirs` to avoid confusion with `get_symlink_dirs`

cylc/flow/workflow_files.py
```diff
@@ -1491,6 +1502,51 @@ def validate_source_dir(source, flow_name):
...
+        try:
+            key, val = pair.split("=")
+            key = key.strip()
+        except ValueError:
+            raise UserInputError(
+                'There is an error in --symlink-dirs option:'
+                f' {pair}. Try entering option in the form '
+                '--symlink-dirs=\'log=$DIR, share=$DIR2, ...\''
+            )
+        if key not in ['run', 'log', 'share', 'share/cycle', 'work']:
```
MetRonnie, line 1539: Rather than rewriting these strings,
```suggestion
        possible_symlink_dirs = WorkflowFiles.SYMLINK_DIRS.symmetric_difference(
            {'', WorkflowFiles.RUN_DIR}
        ).
        if key not in possible_symlink_dirs:
```
But it is a bit awkward due to the use of `''` instead of `'run'` in `WorkflowFiles.SYMLINK_DIRS`.

(If you do use this suggestion, instantiate `possible_symlink_dirs` outside this `for` loop rather than where I've got it)

https://docs.python.org/3.7/library/stdtypes.html#frozenset.symmetric_difference

---

Alternative:
```suggestion
        if key not in {WorkflowFiles.RUN_DIR, WorkflowFiles.LOG_DIR, ...} # etc
```

tests/unit/conftest.py
```diff
@@ -65,8 +65,12 @@ def tmp_run_dir(tmp_path: Path, monkeypatch: pytest.MonkeyPatch):
 
     Args:
         reg: Workflow name.
+        installed: named or no-name. Install type.
```
MetRonnie, line 68:
```suggestion
        installed: If True, make it look like the workflow was installed
            using cylc install (creates _cylc-install dir).
        named: If True and installed is True, the _cylc-install dir will
            be created in the parent to make it look like this is a
            named run.
```

tests/unit/conftest.py
```diff
@@ -75,6 +79,14 @@ def _tmp_run_dir(reg: Optional[str] = None) -> Path:
             run_dir.mkdir(parents=True, exist_ok=True)
             (run_dir / WorkflowFiles.FLOW_FILE).touch(exist_ok=True)
             (run_dir / WorkflowFiles.Service.DIRNAME).mkdir(exist_ok=True)
+            if installed:
+                if named:
```
MetRonnie, line 83:
```suggestion
                if named:
                    if len(Path(reg).parts) < 2:
                        raise ValueError("Named run requires two-level reg")
```

tests/unit/conftest.py
```diff
@@ -65,8 +65,12 @@ def tmp_run_dir(tmp_path: Path, monkeypatch: pytest.MonkeyPatch):
 
     Args:
         reg: Workflow name.
+        installed: named or no-name. Install type.
     """
-    def _tmp_run_dir(reg: Optional[str] = None) -> Path:
+    def _tmp_run_dir(
+        reg: Optional[str] = None,
+        installed: bool = False,
+            named: bool = False) -> Path:
```
MetRonnie, line 73:
```suggestion
        installed: bool = False,
        named: bool = False
    ) -> Path:
```

cylc/flow/pathutil.py
```diff
@@ -196,7 +215,7 @@ def make_symlink(path: Union[Path, str], target: Union[Path, str]) -> bool:
     path = Path(path)
     target = Path(target)
     if path.exists():
-        if path.is_symlink() and path.samefile(target):
+        if path.is_symlink() and target.exists() and path.samefile(target):
```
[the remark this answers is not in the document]
datamel: No, it is not redundant... consider the case where a user sets up a symlink, different to that of what is in the `global.cylc [install] [symlink dirs]` section and then runs `cylc play`. In this scenario...
- `path.exists()` returns True
- `path.is_symlink()` returns True
- `path.samefile(target)` results in error with traceback, since that is the default for `samefile` if either path or target do not exist.
In this case, we want the skipping to be logged and it to pass through without error.

hjoliver (approved):
Approving this now, assuming @MetRonnie's final comments will be addressed (or countered). Very nice, thanks @datamel :+1:

cylc/flow/pathutil.py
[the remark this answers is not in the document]
MetRonnie: Ah ok, might be worth a comment in the code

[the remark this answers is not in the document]
datamel: Thanks, comment added.

tests/unit/test_workflow_files.py
```diff
@@ -1441,3 +1446,61 @@ def test_check_flow_file_symlink(
...
+    ]
+)
+def test_parse_cli_sym_dirs(
+    symlink_dirs: str,
+    err_msg: str,
+    expected: Dict[str, Dict[str, Any]]
+):
+    """Test parse_cli_sym_dirs returns dict or correctly raises errors on cli
+    symlink dir options"""
+    if err_msg != 'None':
```
MetRonnie, line 1485: 🧐
```suggestion
    if err_msg is not None:
```

MetRonnie (approved):
Happy with the response to my review

PR #4251, opened 2021-06-09
Handle mypy 0.900.

hjoliver:

@kinow - this should be uncontroversial, one review will do.

### DIFF
--- a/.github/workflows/test_fast.yml
+++ b/.github/workflows/test_fast.yml
@@ -76,9 +76,11 @@ jobs:
 
       - name: typing
         run: |
-          mypy --explicit-package-bases
-        # Need --explicit-package-bases because we don't have __init__.py in
-        # top-level dir
+          mypy --explicit-package-bases --install-types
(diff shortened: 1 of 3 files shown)


.github/workflows/test_fast.yml
```diff
@@ -76,9 +76,12 @@ jobs:
...
         run: |
-          mypy --explicit-package-bases
-        # Need --explicit-package-bases because we don't have __init__.py in
-        # top-level dir
+          yes | mypy --explicit-package-bases --install-types
+          # Need --explicit-package-bases because we don't have __init__.py in
+          # top-level dir
+          # Need --install-types because mypy>=0.900 only ships with stubs for
+          # stdlib modules (else we need to install e.g. types-aiofiles etc.)
+          # Need "yes" until mypy fix: https://github.com/python/mypy/issues/10598
```
kinow, line 84: :+1:

kinow (approved):
Approving pending only CI to pass. Thanks for fixing it so quickly!!!

PR #4254, opened 2021-06-09
Mypy 0.9 follow-up

MetRonnie:

**Requirements check-list**
  `conda-environment.yml`. (No, because they're only dependencies for tests)
- [x] Does not need tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ install_requires = [
     'pyuv==1.4.*',
     'pyzmq==19.0.*',
     'setuptools>=49',
-    'urwid==2.*'
+    'urwid==2.*',
 ]
 tests_require = [
     'async-timeout>=3.0.0',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -73,27 +73,34 @@ def find_version(*file_paths):
...
     'pytest>=6',
-    'testfixtures>=6.11.0'
+    'testfixtures>=6.11.0',
+    # Type annotation stubs
+    # (http://mypy-lang.blogspot.com/2021/05/the-upcoming-switch-to-modular-typeshed.html)  # noqa: E501
+    'types-Jinja2>=0.1.3',
+    'types-aiofiles>=0.1.3',
+    'types-pkg_resources>=0.1.2',
+    'types-protobuf>=0.1.10',
+    'types-six>=0.1.6',
```
hjoliver, line 83: I initially tried this, then figured using `--install-types` instead was a lower-maintenance solution. What's your reason for going this way?

MetRonnie: It means you don't have to run `mypy --install-types` on setting up a development copy, only `pip install`.

Also, it turns out the `yes | mypy --install-types` isn't sufficient for GH Actions, we'd have to also run `mypy` again afterwards

oliver-sanders: I think this is better as it cuts out an extra (non-obvious) installation step.

Note: This only applies to modules which previously had stubs bundled with MyPy and can only become a problem when adding a dependency in which case you are already editing the `setup.py` anyway.

hjoliver: Hmmm, that's arguable.

> It means you don't have to run `mypy --install-types` on setting up a development copy, only `pip install`

Not really, you can use `mypy.ini` just as you've done here for `--explicit-package-bases`:
```ini
# mypi.ini
install_types = True
```
> we'd have to also run `mypy` again afterwards

Yeah, it is annoying that two passes are required. Yesterday I thought only it was only one, but I must have tested with a type error that didn't involved the missing (/yet to be installed) type defs :grimacing:  However there's little cost to doing two passes (it is quick; on GH Actions it would be automatic; and locally it only matters for the first run in a new venv). 

> I think this is better as it cuts out an extra (non-obvious) installation step.

But at the cost of having to manage extra dependencies that aren't needed for anything but static type checking. 

Anyhow, I guess its a minor annoyance each way ... so I'll go with the consensus!

hjoliver (approved):

Issue #4256, opened 2021-06-11
Handling of partially satisfied prerequisites


hjoliver:
In Cylc 7 partially satisfied prerequisites would eventually (by runahead limiting) stall a suite and prevent the scheduler from shutting down. (Although, so did fully unsatisfied prerequisites! - due to pre-spawning waiting tasks before they are needed)

In Cylc 8 we now have proper "spawn on demand" in that task proxies only appear in the n=0 task pool once *all* of their task-prerequisites are satisfied. Until then, for tasks with more than one prerequisite, prerequisite satisfaction is tracked invisibly (the fact that they are stored in task proxy objects is really just an implementation detail).

If there is nothing else to run, according to the graph, the scheduler shuts down. However, it logs a warning if any partially satisfied prerequisites are present.

## question

(Raised in the last meeting by @dpmatthews)

Should we consider it an error if the scheduler has nothing else to run but there are partially satisfied prerequisites present, and stay up as stalled rather than shut down?

hjoliver:
## answer?

> Should we consider it an error if the scheduler has nothing else to run but there are partially satisfied prerequisites present, and stay up as stalled rather than shut down?

In my opinion, no.

While it is quite possible for a user to make a mistake by triggering a task that only leads on to a sub-graph when they intended the whole graph to run, it is also perfectly legit to do that intentionally, so it's not possible for Cylc to to determine which is which and we shouldn't automatically assume it was an error.

hjoliver:
[Cut-and-paste from my defense of the status quo on Element chat!]

> Example:
```
[scheduling]                                      
  [[graph]]                                       
    R1 = "foo => bar & baz => qux"                
[runtime]                                         
   [[foo, bar, baz, qux]]                         
       script = sleep 10  
```
> If I `cylc play` this as normal, or do `cylc play --start-task=foo.1`, all four tasks will run then the scheduler will shut down.
But if I do `cylc play --start-task=bar.1`, only `bar.1` will run then the scheduler will shut down.
> 
> But this will be logged at shutdown:
```
2021-06-11T12:04:31+12:00 WARNING - Partially satisfied prerequisites left over:
        qux.1 is waiting on:
        * baz.1 succeeded
2021-06-11T12:04:32+12:00 INFO - DONE
```
> IMO that's the correct thing to do. qux.1 should appear in n=0 (active task) only when/if it becomes ready to run, which it never does because baz.1 has no reason to be spawned in this scenario.
[...]

PR #4258, opened 2021-06-11
Don't count hidden pool for runahead limiting.

hjoliver:

I noticed that partially-satisfied prerequisites (aka hidden pool task proxies) are being counted for runahead limiting, which is not ~~necessary~~ the right thing to do.

On this branch the runahead limit simply restricts the spread of active tasks.

(This will break a few runahead tests...)

**Requirements check-list**
  `conda-environment.yml`. (None)

### DIFF
--- a/tests/functional/reload/runahead/flow.cylc
+++ b/tests/functional/reload/runahead/flow.cylc
@@ -8,7 +8,6 @@
     final cycle point = 2010-01-05
     runahead limit = PT12H # marker
     [[graph]]
-        # oops is stuck waiting task to hold back runahead
         R1 = "foo:start => reloader"
         PT6H ="foo" 
 [runtime]
--- a/tests/functional/reload/runahead/flow.cylc
+++ b/tests/functional/reload/runahead/flow.cylc
@@ -1,24 +1,30 @@
 [scheduler]
     UTC mode = True
     [[events]]
-        timeout = PT0.2M
-        abort on timeout = True
+        abort on inactivity = True
+        inactivity = PT10S
 [scheduling]
     initial cycle point = 2010-01-01
     final cycle point = 2010-01-05
-    runahead limit = P2 # marker
+    runahead limit = PT12H # marker
     [[graph]]
         # oops is stuck waiting task to hold back runahead
-        R1/T00 = "foo & reloader => oops"
-        T00/PT6H = "foo => bar"
+        R1 = "foo:start => reloader"
+        PT6H ="foo" 
 [runtime]
     [[foo]]
-        script = false
-    [[bar, oops]]
-        script = true
+        script = """
+            if [[ "$CYLC_TASK_JOB" == '20100101T0000Z/foo/01' ]]; then
+                # keep running until scheduler is gone
+                while cylc ping $CYLC_WORKFLOW_NAME; do
+                    sleep 10
+                done
+            else
+                true
+            fi
+        """
     [[reloader]]
         script = """
-cylc__job__poll_grep_workflow_log '\[foo.* (received)failed'
-perl -pi -e 's/(runahead limit = )P2( # marker)/\1 P4\2/' $CYLC_WORKFLOW_RUN_DIR/flow.cylc
+perl -pi -e 's/(runahead limit = )PT12H( # marker)/\1 P1D\2/' $CYLC_WORKFLOW_RUN_DIR/flow.cylc
 cylc reload $CYLC_WORKFLOW_NAME
 """
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -56,8 +56,14 @@ Third beta release of Cylc 8.
 
 ### Enhancements
 
+[#4258](https://github.com/cylc/cylc-flow/pull/4258) -
+Make runahead limiting depend only on active tasks.
+
 [#4284](https://github.com/cylc/cylc-flow/pull/4284)
  - Make `--color=never` work with `cylc <command> --help`.
+
+[#4259](https://github.com/cylc/cylc-flow/pull/4259) -
+Ignore pre-initial dependencies with `cylc play --start-task`
 
 [#4103](https://github.com/cylc/cylc-flow/pull/4103) -
 Expose runahead limiting to UIs; restore correct force-triggering of queued
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -303,7 +303,6 @@ class TaskPool:
         # Any finished tasks can be released immediately (this can happen at
         # restart when all tasks are initially loaded into the runahead pool).
         # And any manually-triggered task.
-
         for itask in (
             itask
             for point_id_map in self.main_pool.values()
@@ -320,7 +319,7 @@ class TaskPool:
             released = True
 
         points = []
-        for point, itasks in sorted(self.get_tasks_by_point().items()):
+        for point, itasks in sorted(self._get_main_tasks_by_point().items()):
             if points or any(
                 not itask.state(
                     TASK_STATUS_FAILED,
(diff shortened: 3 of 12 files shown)


tests/functional/reload/runahead/flow.cylc
```diff
@@ -1,24 +1,30 @@
...
+        script = """
+            if [[ "$CYLC_TASK_JOB" == '20100101T0000Z/foo/01' ]]; then
+                # keep running until scheduler is gone
+                while cylc ping $CYLC_WORKFLOW_NAME; do
+                    sleep 10
+                done
+            else
+                true
+            fi
+        """
```
hjoliver, line 25: Several functional tests updated to test the runahead limit like this, by keeping the earliest task running until the scheduler shuts down on an inactivity timeout of `PT10S`.  This is a potential cause of test flakiness under heavy load, but not sure if there's an alternative when runahead limiting is only caused by active tasks.

hjoliver:
Ready for review.

oliver-sanders:
Potentially blocking comment from Dave posted to the partially satisfied prereqs issue: https://github.com/cylc/cylc-flow/issues/4256#issuecomment-860817762

> if #4258 gets merged I think it would run forever with no indication of a problem

hjoliver:
> Potentially blocking comment from Dave posted to the partially satisfied prereqs issue: [#4256 (comment)](https://github.com/cylc/cylc-flow/issues/4256#issuecomment-860817762)
> 
> > if #4258 gets merged I think it would run forever with no indication of a problem

OK, but what's the purpose of runahead limiting?

Well, it was invented as a crude way to prevent the scheduler from being overwhelmed by too many task proxies. We can do that better now with proper task pool limiting, which works within as well as across cycle points, but we're retaining runahead limiting to support existing suites.

Users understand runahead limiting as *preventing the fastest tasks from getting too far ahead of the slowest*, i.e. as restricting the number of active cycle points.

So, on what basis should runahead limiting be determined by partially satisfied prerequisites rather than active tasks?

Maybe we do want to flag it as a problem if partially satisfied prerequisites accumulate over many cycle points, but that is a different problem, orthogonal to runahead limiting.

dpmatthews:
> So, on what basis should runahead limiting be determined by partially satisfied prerequisites rather than active tasks?

It's the same reason unhandled failed tasks need to count for runahead limiting. It indicates a problem which potentially means there are still tasks that need to be run at that cycle.

hjoliver:
> > So, on what basis should runahead limiting be determined by partially satisfied prerequisites rather than active tasks?
> 
> It's the same reason unhandled failed tasks need to count for runahead limiting.

Failed tasks did not even count for runahead limiting in Cylc 7, if I recall correctly.

> It indicates a problem which potentially means there are still tasks that need to be run at that cycle.

That's true, but why should that require a complete stall, unnecessarily reducing time to completion (or to catch up from a delay or  whatever) even though unaffected parts of the workflow can perfectly well keep going?

A complete stall is not needed to conserve resources. I really can't think of a good reason why the scheduler should arbitrarily stall n cycle points ahead of a failed task. Pre-Cylc 8 we had no choice because there were loads of waiting tasks in the pool, which multiplied with spread over cycles, and that in itself was a performance problem for the scheduler. But now we don't have that problem, and runahead limiting could just do just what I intended: conserve scheduler and external resources by limiting *task activity* to a range of cycle points. 

If unhandled failed tasks and partially-met prerequisites potentially indicate a problem (and they do!) users should be able to tell the scheduler what to do in response to specifically those problems. (Which could include stalling as an option, but I'd rather it wasn't weirdly linked to the concept of runahead limiting).

hjoliver:
A sub-text here, for me, is that with spawn-on-demand we don't technically need runahead limiting any more for constraining scheduler activity. We need to keep it for historical reasons, but users might eventually stop using it if we don't conflate it unnecessarily with other real problems that do need to be managed.

dpmatthews:
> Failed tasks did not even count for runahead limiting in Cylc 7, if I recall correctly.

They do for me based on a simple test (it's getting very hard to remember how Cylc 7 works!).

> > It indicates a problem which potentially means there are still tasks that need to be run at that cycle.
> 
> That's true, but why should that require a complete stall, unnecessarily ~reducing~ [OOPS increasing!] time to completion (or to catch up from a delay or whatever) even though unaffected parts of the workflow can perfectly well keep going?

The trouble is that runahead limit could be used to prevent how far ahead tasks could get. Users may be relying on this (e.g. to constrain disk usage). I agree that in future we should aim to stop using runahead limiting in its current form and get constraints like this properly represented in the graph.

tests/functional/reload/runahead/flow.cylc
```diff
@@ -1,24 +1,30 @@
...
-        abort on timeout = True
+        abort on inactivity = True
+        inactivity = PT10S
 [scheduling]
     initial cycle point = 2010-01-01
     final cycle point = 2010-01-05
-    runahead limit = P2 # marker
+    runahead limit = PT12H # marker
     [[graph]]
         # oops is stuck waiting task to hold back runahead
```
MetRonnie, line 11: Still a reference to the deleted `oops` task

MetRonnie (approved):
Regardless of the outcome of the discussion, leaving my review of the code.

hjoliver:
> >  Failed tasks did not even count for runahead limiting in Cylc 7, if I recall correctly.
> 
> They do for me based on a simple test

You were probably stalling on pre-spawned waiting tasks downstream of the failed ones.  Here's my test with 7.9.3:

```ini
$ cat suite.rc
[scheduling]
  cycling mode = integer
  initial cycle point = 1
  [[dependencies]]
     [[[P1]]]
          graph = "foo => bar"
[runtime]
   [[root]]
   [[foo]]
   [[bar]]
       script = false
```

Result in the old `cylc monitor`:
shot


>  (it's getting very hard to remember how Cylc 7 works!).

Ain't that the truth.

hjoliver:
> The trouble is that runahead limit could be used to prevent how far ahead tasks could get. Users may be relying on this (e.g. to constrain disk usage). 

I don't understand this. On this branch, it still constrains how far ahead tasks can get relative to the slowest tasks, and the result will be the ~same in the absence of task failures or stuck partially satisfied multi-prerequisite tasks.  Hopefully users aren't relying on task failures or graphing errors to constrain disk usage.

Users could be *expecting* failed or stuck waiting tasks (Cylc 7) to stall the workflow, but again:
- my Cylc 7.9.3 demo above shows that the former is not actually true at Cylc 7
- at Cylc 8 the latter is only possible for partially satisfied multi-prerequisite tasks, and not wholly unsatisfied single or multi-prerequisite tasks ... so it would be weird to single out partially satisfied multi-prereq tasks as a particular problem here. UNLESS perhaps we implement the `?=>` notation to make optional dependence explicit.

hjoliver:
>.. we implement the ?=> notation to make optional dependence explicit.

If we do this, that would enable all unsatisfied "expected" dependencies to be treated equally.  Then I guess we could make it optional to have unsatisfied expected dependencies to cause a runahead stall.  And it could be the default if you really want to insist.

Or (also with `?=>`) we could have a main loop plugin periodically check for unsatisfied expected dependencies, and have various options for what to do about it if any are found.

dpmatthews:
> You were probably stalling on pre-spawned waiting tasks downstream of the failed ones

Ah - that explains it. Slightly surprised at this behaviour. However, I suppose if there are no downstream tasks that will be triggered by re-running the failed task then you can argue there is no reason to stall.

If there are downstream tasks then I think we still need the stall. I think that means unhandled failed tasks need to cause a stall at Cylc 8 (because we can't rely on waiting tasks to do this).

hjoliver:
Closing as superseded by the new "optional outputs" proposal, which allows us to treat single- and multi-prerequisite tasks more equitably in terms of stall and runahead behaviour (in the single case, we will be able to identify "incomplete tasks")

PR #4259, opened 2021-06-11
Ignore pre-initial deps for --start-tasks.

### LINKED ISSUE #4257: --start-task: ignore pre-initial dependencies?
Raised by @dpmatthews in the last meeting.

On a clean start at the initial (or other) cycle point we ignore pre-start-cycle dependencies as a convenience for users. Should we do the same if starting a new run from one or more individual tasks in the graph? e.g. `cylc play --start-task=blah.2`

### LINKED ISSUE #4266: cylc play: strange behaviour when using --start-task with --startcp
Consider this simple workflow:
```ini
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    final cycle point = 5
    [[dependencies]]
        P1 = "foo[-P1] => foo"
[runtime]
    [[foo]]
        script = sleep 5
```
If I run this with `cylc play --starttask=foo.4 --startcp=2`  it appears to run OK.
It starts from `foo.4` and the `--startcp=2` doesn't appear to have any affect expect that the log reports:
```
INFO - Start point: 2
```

However, if I run this using `cylc play --starttask=foo.2 --startcp=4` I get very strange results:
```
2021-06-14T13:58:20+01:00 INFO - Start task: ['foo.2']
2021-06-14T13:58:20+01:00 INFO - Start point: 4
2021-06-14T13:58:20+01:00 INFO - [foo.2] -released from runahead
2021-06-14T13:58:20+01:00 INFO - (foo.2) spawned foo.3 flow(x)
2021-06-14T13:58:29+01:00 INFO - [foo.2] status=running: (received)succeeded at 2021-06-14T13:58:28+01:00  for job(01) flow(x)
2021-06-14T13:58:29+01:00 INFO - [foo.3] -released from runahead
2021-06-14T13:58:29+01:00 INFO - (foo.3) spawned foo.4 flow(x)
2021-06-14T13:58:29+01:00 INFO - [foo.4] -released from runahead
2021-06-14T13:58:30+01:00 INFO - Queue released: foo.3
2021-06-14T13:58:30+01:00 INFO - Queue released: foo.4
2021-06-14T13:58:30+01:00 INFO - [foo.3] -submit-num=01, host=vld182.cmpd1.metoffice.gov.uk
2021-06-14T13:58:30+01:00 INFO - [foo.4] -submit-num=01, host=vld182.cmpd1.metoffice.gov.uk
2021-06-14T13:58:39+01:00 INFO - [foo.3] status=running: (received)succeeded at 2021-06-14T13:58:37+01:00  for job(01) flow(x)
2021-06-14T13:58:39+01:00 INFO - [foo.4] status=running: (received)succeeded at 2021-06-14T13:58:37+01:00  for job(01) flow(x)
2021-06-14T13:58:39+01:00 INFO - (foo.4) spawned foo.5 flow(x)
2021-06-14T13:58:48+01:00 INFO - [foo.5] status=running: (received)succeeded at 2021-06-14T13:58:47+01:00  for job(01) flow(x)
2021-06-14T13:58:48+01:00 INFO - Workflow shutting down - AUTOMATIC
```
(log entries thinned to just show the important bits)
Note that foo.3 gets spawned as soon as foo.2 gets released from runahead rather than when it completes.
Also foo.3 and foo.4 get run in parallel (I think this makes it reasonable to mark this as a bug?).

I suspect we ought to make `--starttask` and `--startcp` mutually exclusive options?

hjoliver:
These changes:
- close #4257 
- close #4266 

**Requirements check-list**
  `conda-environment.yml`. (None)
- [x] Appropriate tests are included (extended unit tests).
- [x] No documentation update required. (Command help updated).

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -645,12 +645,21 @@ class WorkflowConfig:
             self.start_point
         """
         if getattr(self.options, 'startcp', None) is not None:
-            # Warm start from a point later than initial point.
+            # Start from a point later than initial point.
             if self.options.startcp == 'now':
                 self.options.startcp = get_current_time_string()
             self.start_point = get_point(self.options.startcp).standardise()
+        elif getattr(self.options, 'starttask', None) is not None:
+            # Start from designated task(s).
+            # Select the earliest start point for use in pre-initial ignore.
+            self.start_point = min(
+                get_point(
+                    TaskID.split(taskid)[1]
+                ).standardise()
+                for taskid in self.options.starttask
+            )
         else:
-            # Cold start.
+            # Start from the initial point.
             self.start_point = self.initial_point
 
     def process_final_cycle_point(self):
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -48,8 +48,13 @@ from cylc.flow import LOG
 from cylc.flow.c3mro import C3
 from cylc.flow.listify import listify
 from cylc.flow.exceptions import (
-    CylcError, WorkflowConfigError, IntervalParsingError, TaskDefError,
-    ParamExpandError)
+    CylcError,
+    WorkflowConfigError,
+    IntervalParsingError,
+    TaskDefError,
+    ParamExpandError,
+    UserInputError
+)
 from cylc.flow.graph_parser import GraphParser
 from cylc.flow.param_expand import NameExpander
 from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
@@ -648,7 +653,7 @@ class WorkflowConfig:
         starttask = getattr(self.options, 'starttask', None)
 
         if startcp is not None and starttask is not None:
-            raise WorkflowConfigError(
+            raise UserInputError(
                 "--start-cycle-point and --start-task are mutually exclusive"
             )
         if startcp:
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -644,12 +644,19 @@ class WorkflowConfig:
             self.options.startcp
             self.start_point
         """
-        if getattr(self.options, 'startcp', None) is not None:
+        startcp = getattr(self.options, 'startcp', None)
+        starttask = getattr(self.options, 'starttask', None)
+
+        if startcp is not None and starttask is not None:
+            raise WorkflowConfigError(
+                "--start-cycle-point and --start-task are mutually exclusive"
+            )
+        if startcp:
             # Start from a point later than initial point.
             if self.options.startcp == 'now':
                 self.options.startcp = get_current_time_string()
             self.start_point = get_point(self.options.startcp).standardise()
-        elif getattr(self.options, 'starttask', None) is not None:
+        elif starttask:
             # Start from designated task(s).
             # Select the earliest start point for use in pre-initial ignore.
             self.start_point = min(
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -300,14 +300,14 @@ class TaskPool:
 
         released = False
 
-        # Any finished tasks can be released immediately (this can happen at
-        # restart when all tasks are initially loaded into the runahead pool).
-        # And any manually-triggered task.
+        # At restart all tasks are loaded as runahead-limited, but finished and
+        # manually-triggered ones (including --start-task) can be released
+        # immediately. Note runahead release can cause the task pool to change
+        # size because we spawn parentless tasks on previous-instance release.
 
         for itask in (
             itask
-            for point_id_map in self.main_pool.values()
-            for itask in point_id_map.values()
+            for itask in self.get_tasks()
             if itask.state.is_runahead
             if itask.state(
                 TASK_STATUS_FAILED,
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -366,13 +366,20 @@ def test_process_icp(
 
 
 @pytest.mark.parametrize(
-    'startcp, expected',
-    [('20210120T1700+0530', '20210120T1700+0530'),
-     ('now', '20050102T0615+0530'),
-     (None, '18990501T0000+0530')]
+    'startcp, starttask, expected',
+    [
+        ('20210120T1700+0530', None, '20210120T1700+0530'),
+        ('now', None, '20050102T0615+0530'),
+        (None, None, '18990501T0000+0530'),
+        (
+            None,
+            ['foo.20090802T0615+0530', 'bar.20090802T0515+0530'],
+            '20090802T0515+0530'
+        )
+    ]
 )
 def test_process_startcp(
-    startcp: Optional[str], expected: str,
+    startcp: Optional[str], starttask: Optional[str], expected: str,
     monkeypatch: pytest.MonkeyPatch, set_cycling_type: Fixture
 ) -> None:
     """Test WorkflowConfig.process_start_cycle_point().
@@ -387,9 +394,9 @@ def test_process_startcp(
     set_cycling_type(ISO8601_CYCLING_TYPE, time_zone="+0530")
     mocked_config = Mock(initial_point='18990501T0000+0530')
     mocked_config.options.startcp = startcp
+    mocked_config.options.starttask = starttask
     monkeypatch.setattr('cylc.flow.config.get_current_time_string',
                         lambda: '20050102T0615+0530')
-
     WorkflowConfig.process_start_cycle_point(mocked_config)
     assert str(mocked_config.start_point) == expected
 
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -366,20 +366,48 @@ def test_process_icp(
 
 
 @pytest.mark.parametrize(
-    'startcp, starttask, expected',
+    'startcp, starttask, expected, expected_err',
     [
-        ('20210120T1700+0530', None, '20210120T1700+0530'),
-        ('now', None, '20050102T0615+0530'),
-        (None, None, '18990501T0000+0530'),
+        (
+            '20210120T1700+0530',
+            None,
+            '20210120T1700+0530',
+            None
+        ),
+        (
+            'now',
+            None,
+            '20050102T0615+0530',
+            None
+        ),
+        (
+            None,
+            None,
+            '18990501T0000+0530',
+            None
+        ),
         (
             None,
             ['foo.20090802T0615+0530', 'bar.20090802T0515+0530'],
-            '20090802T0515+0530'
+            '20090802T0515+0530',
+            None
+        ),
+        (
+            '20210120T1700+0530',
+            ['foo.20090802T0615+0530'],
+            None,
+            (
+                WorkflowConfigError,
+                "--start-cycle-point and --start-task are mutually exclusive"
+            ),
         )
     ]
 )
 def test_process_startcp(
-    startcp: Optional[str], starttask: Optional[str], expected: str,
+    startcp: Optional[str],
+    starttask: Optional[str],
+    expected: str,
+    expected_err: Optional[Tuple[Type[Exception], str]],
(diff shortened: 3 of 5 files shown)


hjoliver:
A good test case for this:
```
[scheduling]
   cycling mode = integer
   initial cycle point = 1
   runahead limit = P2
   [[graph]]
      P1 = """
         a => b => c => d & e
         c[-P1] => c
      """
[runtime]
   [[a, b, c, d, e]]
      script = "sleep 5"
```

cylc/flow/task_pool.py
```diff
@@ -299,14 +299,14 @@ def release_runahead_tasks(self):
...
+        # At restart all tasks are loaded as runahead-limited, but finished and
+        # manually-triggered ones (including --start-task) can be released
+        # immediately. Note runahead release can cause the task pool to change
+        # size because we spawn parentless tasks on previous-instance release.
 
         for itask in (
             itask
-            for point_id_map in self.main_pool.values()
-            for itask in point_id_map.values()
+            for itask in self.get_tasks()
```
hjoliver, line 309: This was actually a bug, because the task pool can change size during iteration (as described in the comment above).

hjoliver:
One fast test failed. It passes locally and is unrelated to this change. 
`FAILED tests/integration/test_publisher.py::test_publisher - asyncio.exceptio...`

cylc/flow/config.py
```diff
@@ -789,12 +789,21 @@ def process_start_cycle_point(self):
...
             self.start_point = get_point(self.options.startcp).standardise()
+        elif getattr(self.options, 'starttask', None) is not None:
+            # Start from designated task(s).
+            # Select the earliest start point for use in pre-initial ignore.
+            self.start_point = min(
+                get_point(
+                    TaskID.split(taskid)[1]
+                ).standardise()
+                for taskid in self.options.starttask
+            )
```
hjoliver, line 804: The only thing I had to do for this PR was define the start cycle point, for the start task case. Then pre-initial ignore gets done automatically.

cylc/flow/task_pool.py
hjoliver: (In fact the above example triggers this bug, on master)

MetRonnie: I don't understand the difference with how it was previously, considering the definition of
https://github.com/cylc/cylc-flow/blob/ae6f0652e7ec656e20efc8940214ae30d3234eaf/cylc/flow/task_pool.py#L664-L672

(apart from the fact that `self.get_tasks()` can mutate `self.main_pool_list` and `self.main_pool_changed`)

tests/unit/test_config.py
```diff
@@ -366,13 +366,20 @@ def test_process_icp(
...
+    [
+        ('20210120T1700+0530', None, '20210120T1700+0530'),
+        ('now', None, '20050102T0615+0530'),
+        (None, None, '18990501T0000+0530'),
+        (
+            None,
+            ['foo.20090802T0615+0530', 'bar.20090802T0515+0530'],
+            '20090802T0515+0530'
+        )
+    ]
```
MetRonnie, line 379: I think there should be a test case where both `startcp` and `starttask` are specified (with the former winning, according to how you've implemented it)

MetRonnie (commented):
Several minor comments and a request for an additional test case

cylc/flow/task_pool.py
hjoliver: If I run the example above, with `--start-task=a.2`:
```
2021-06-23T10:17:47+12:00 INFO - [a.2] -released from runahead
2021-06-23T10:17:47+12:00 INFO - (a.2) spawned a.3 flow(i)
2021-06-23T10:17:47+12:00 ERROR - dictionary changed size during iteration
          File "/home/oliverh/cylc/cylc-flow/cylc/flow/scheduler.py", line 572, in start_scheduler
            await self.main_loop()
          File "/home/oliverh/cylc/cylc-flow/cylc/flow/scheduler.py", line 1409, in main_loop
            if not self.is_paused and self.pool.release_runahead_tasks():
          File "/home/oliverh/cylc/cylc-flow/cylc/flow/task_pool.py", line 306, in release_runahead_tasks
            for itask in (
          File "/home/oliverh/cylc/cylc-flow/cylc/flow/task_pool.py", line 306, in <genexpr>
            for itask in (
        RuntimeError: dictionary changed size during iteration
2021-06-23T10:17:47+12:00 CRITICAL - Workflow shutting down - dictionary changed size during iteration
```

hjoliver: On this branch, we iterate over the list of tasks (returned by `self.get_tasks()`) that reflects the state of the pool before iteration begins.

On master, we iterate over `main_pool.values()`. In Python 3, `dict.values()` does not return a list of values, it's a "view" of the actual dictionary, so modifying it in-flight is bad.  https://docs.python.org/3/library/stdtypes.html#dict-views


```
for point_id_map in self.main_pool.values():                                                                                                                                                                                                   
    for itask in point_id_map.values():   
        #...
```

tests/unit/test_config.py
hjoliver: I've made the two options mutually exclusive (which they should be, as noted by @dpmatthews in Element chat recently) and updated the unit test for that.

cylc/flow/config.py
```diff
@@ -788,12 +788,19 @@ def process_start_cycle_point(self):
             self.options.startcp
             self.start_point
         """
-        if getattr(self.options, 'startcp', None) is not None:
+        startcp = getattr(self.options, 'startcp', None)
+        starttask = getattr(self.options, 'starttask', None)
+
+        if startcp is not None and starttask is not None:
+            raise WorkflowConfigError(
+                "--start-cycle-point and --start-task are mutually exclusive"
```
MetRonnie, line 796: Would `UserInputError` be better for this?

MetRonnie (approved):
👍

cylc/flow/task_pool.py
dwsutherland: I found this bug too:
```
2021-06-30T12:27:08+12:00 INFO - [a.2] -released from runahead
2021-06-30T12:27:08+12:00 INFO - (a.2) spawned a.3 flow(r)
2021-06-30T12:27:08+12:00 ERROR - dictionary changed size during iteration
          File "/home/sutherlander/cylc/cylc-flow/cylc/flow/scheduler.py", line 572, in start_scheduler
            await self.main_loop()
          File "/home/sutherlander/cylc/cylc-flow/cylc/flow/scheduler.py", line 1405, in main_loop
            if not self.is_paused and self.pool.release_runahead_tasks():
          File "/home/sutherlander/cylc/cylc-flow/cylc/flow/task_pool.py", line 309, in release_runahead_tasks
            for point_id_map in self.main_pool.values()
          File "/home/sutherlander/cylc/cylc-flow/cylc/flow/task_pool.py", line 308, in <genexpr>
            itask
        RuntimeError: dictionary changed size during iteration
2021-06-30T12:27:08+12:00 CRITICAL - Workflow shutting down - dictionary changed size during iteration
```
Changed it to:
```
         for itask in (
             itask
-            for point_id_map in self.main_pool.values()
+            for point_id_map in list(self.main_pool.values())
             for itask in point_id_map.values()
             if itask.state.is_runahead
             if itask.state(

```
For the purposes of reviewing this:
```
2021-06-30T12:33:26+12:00 INFO - (a.2) spawned b.2 flow(w)
2021-06-30T12:33:26+12:00 INFO - [a.3] status=running: (received)succeeded at 2021-06-30T12:33:25+12:00  for job(01) flow(w)
2021-06-30T12:33:26+12:00 INFO - (a.3) spawned b.3 flow(w)
2021-06-30T12:33:26+12:00 INFO - Workflow shutting down - AUTOMATIC
2021-06-30T12:33:26+12:00 WARNING - Partially satisfied prerequisites left over:
        b.2 is waiting on:
        * b.1 succeeded
ESC[0m2021-06-30T12:33:26+12:00 INFO - DONE
```

cylc/flow/config.py
dwsutherland: Hmm.. Is this the proper way of doing it?

What if the workflow has some R1 / start-up tasks for a few cycles, and then runs it's normal tasks (containing the start task) at some offset from ICP?

dwsutherland (commented):
Just a comment about forcing the ICP to match the start task (what if the the start task is defined on some offset from the ICP?).

cylc/flow/task_pool.py
hjoliver: Ah, I'm slightly confused @dwsutherland - isn't that exactly the same bug that I fixed *on this branch*?? (although you've fixed it in a slightly different way, same result).

cylc/flow/config.py
hjoliver: Sorry, I don't follow here either :thinking:  If we start up the scheduler with `--start-task` all that matters (for the purpose of this PR) is that other tasks don't get stuck waiting on tasks at cycle points earlier than the start-task(s).  How do R1 tasks and offsets from ICP affect that?

dwsutherland: The start task (and perhaps dependents) may not exist at the ICP.  They could be defined as starting from some offset of the ICP...

So if we force the ICP to a non-ICP start task.. What is the result?
(might test this tomorrow)

cylc/flow/task_pool.py
dwsutherland: My bad should have phrased it:

"While trying to reproduce the problem on master I found that same bug too:"

Yes, it's fixed on this branch.. But it wasn't the purpose of this branch (though clearly related), so when trying to reproduce the master branch I ran into the same bug that is fixed in this branch.....

It was important to fix master to test the actual purpose of this PR, to know how master handles pre-initial dependencies (because we can't even get that far in the presence of the bug master)... So I made that above temporary fix.

hjoliver: OK, no worries - thanks for clarifying!

cylc/flow/config.py
hjoliver: You have to give the name.point of start-tasks, and they are checked for validity at the specified point.
```ini
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    [[graph]]
        R1 = prep => foo
        P1 = foo[-P1] => foo
[runtime]
    [[prep, foo]]
```
 
shot

hjoliver: Hmm, maybe my statement above wasn't clear:
> The only thing I had to do for this PR was define the start cycle point, for the start task case. Then pre-initial ignore gets done automatically.

Rather than *define the start cycle point for start tasks* I should have said, *extract the cycle point from start tasks and set the workflow start point to that*.

dwsutherland: What I'm asking is; how do you square the making the workflow starting point the with the task starting point when the task is defined in the workflow graph to be some offset from the starting point?
```
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    [[graph]]
        R1 = prep => foo
        +P5/P1 = foo[-P1] => foo => bar
[runtime]
    [[prep, foo, bar]]
```
And then run with:
`$ cylc play --start-task=bar.10 demo2/run1`

dwsutherland: by start point I assume you mean `initial cycle point` at run time (?).

hjoliver: Right, understood now. No, ICP defines the extent of the graph (and where things happen relative to ICP) but *start point* (or start task now too) is where you start up within that graph. So we don’t mess with ICP here.

dwsutherland: Thanks, well, apart from the minor conflict.. wer're ready to go with this one.

dwsutherland (approved):

dwsutherland:
Bah, rebase has conflicts even after manually fixing the Changes.md

hjoliver:
Rebased. The conflict was only in the change log so no need to wait for the tests.

PR #4265, opened 2021-06-12
Remove on-the-fly installation text from play help.

hjoliver:
```diff
+ 
- If the workflow is not already installed (by "cylc install" or a previous run)
- it will be installed on the fly before start up.
```
This seems to be a direct translation from Cylc 7, with "register" replaced by "install", and not actually implemented.


datamel (approved):
Thanks @hjoliver.

PR #4267, opened 2021-06-15
Tidy tidy

oliver-sanders:
Fiddling about whilst tuning into the UMUW.

Remove some old things that aren't required any more.

**Requirements check-list**
  `conda-environment.yml`.


cylc/flow/etc/global.cylc.eg
```diff
@@ -1,23 +0,0 @@
-#------------------------------------------------------------------------------
-# How to create a site or user global.cylc global config file.
-#------------------------------------------------------------------------------
-# The "cylc config" command prints global config defaults,
-# overridden by site global settings (if any), overridden by user global
-# settings (if any).
-#
-# To generate a new global config file:
-#   % cylc config > global.cylc
```
oliver-sanders, line 9: File not referenced from the docs so no one is likely to find it.

The advice to generate a global config from the hardcoded defaults is something we probably shouldn't push ahead with.

etc/bin/live-graph-movie.sh
```diff
@@ -1,127 +0,0 @@
...
-#  * http://wiki.webmproject.org/howtos/convert-png-frames-to-webm-video
-#  * http://www.webmproject.org/docs/encoder-parameters/
-
-if [[ $1 == "-h" || $1 == "--help" || $1 == "help" ]]; then
-    usage
-    exit 0
-fi
-
-WORKFLOW=$1
-if ! cylc pr --fail $WORKFLOW >/dev/null; then
```
oliver-sanders, line 49: Broken script relying on the removed `cylc pr` command and GCylc frames output. No one has noticed it's broken so I'm guessing it's safe to remove.

etc/bin/make-tarball
```diff
@@ -1,62 +0,0 @@
...
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
-set -e; trap "echo ERROR" ERR
-
-function usage {
-cat <<eof
-USAGE: make-tarball
```
oliver-sanders, line 23: No longer part of the release process, GitHub does this for us.

etc/bin/start-group-demo.sh
```diff
@@ -1,98 +0,0 @@
...
-for GROUP in nwp tst opr; do
-  for N in 1 2 3 4; do
-    if (( SKIP == N )); then
-      SKIP=$(( SKIP - 1))
-      break
-    fi
-    WORKFLOW=${GROUP}-$N
-    mkdir -p $DEST/$WORKFLOW
-    cp -r flow.cylc $DEST/$WORKFLOW
-    cylc reg $DEST/$WORKFLOW $DEST/$WORKFLOW
```
oliver-sanders, line 93: Broken script not referenced from the docs.

hjoliver (approved):
Nice tidying, thanks

PR #4269, opened 2021-06-15
Tweak terminal logo.

hjoliver:
A bit of fluff. Pushing up the tweak I made for the demo in my presentation last night.

Master:
shot-old

This branch:
shot
 
IMO the new one is more readable: the "Y" looks less like a "4"; the word looks less like "chic" because the "L" doesn't look like an "I"; and it is also more balanced with the tail on the "L" ... while still looking colourful and graphic.

### DIFF
--- a/cylc/flow/scripts/__init__.py
+++ b/cylc/flow/scripts/__init__.py
@@ -27,19 +27,19 @@ _copyright_year = 2021  # This is set by GH Actions update_copyright workflow
 # fmt: off
 LOGO_LETTERS = (
     (
-        "oooo",
-        "oo  ",
+        "ooo",
+        "oo ",
         "oooo",
     ),
     (
         "oo oo",
         "ooooo",
-        "   oo",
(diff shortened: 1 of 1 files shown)


hjoliver:
(Cancelling CI)

hjoliver:
Pinging @MetRonnie and @wxtim for review as I think they both commented on the deficiencies of the original.

MetRonnie (approved):
Is the Y meant to overlap the first C?

Personally I might be more in favour of an L like this

image

Nevertheless, a welcome change

wxtim (approved):
What Ronnie said. Not worth holding it up

hjoliver:
> Is the Y meant to overlap the first C?

Yes

> Personally I might be more in favour of an L like this

Yeah I tried all the above, and more, and decided I liked the look of the one I settled on best - viewing it as an abstract block of colours as well as the word.

PR #4271, opened 2021-06-16
flake8: simplify

### LINKED ISSUE #4239: flake8: simplify

oliver-sanders:
More meeting code fiddling, closes #4239 

List of simplify codes here: https://github.com/MartinThoma/flake8-simplify#SIM106

Used a few `noqa` flags where appropriate, there are a few things in the code that are hard to refactor, usually patterns which should be avoided but which aren't worth the pain of totally overhauling at this stage.

Suggest adding `?=1` to the diff URL to hide the whitespace changes.

**Requirements check-list**
  `conda-environment.yml`.

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -708,11 +708,12 @@ with Conf('global.cylc', desc='''
             Conf('hosts', VDR.V_STRING_LIST, ['localhost'])
 
     # Platform Groups
-    with Conf('platform groups'):
+    with Conf('platform groups'):  # noqa: SIM117 (keep same format)
         with Conf('<group>'):
             Conf('platforms', VDR.V_STRING_LIST)
     # Symlink Dirs
-    with Conf('symlink dirs', desc="""
+    with Conf('symlink dirs',   # noqa: SIM117 (keep same format)
+              desc="""
         Define directories to be moved, symlinks from the the original
         ``$HOME/cylc-run`` directories will be created.
     """):
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -15,9 +15,9 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """Define all legal items and values for cylc workflow definition files."""
 
-import re
-
+import contextlib
 from itertools import product
+import re
 from typing import Any, Dict, Set
 
 from metomi.isodatetime.data import Calendar
@@ -228,7 +228,7 @@ with Conf(
                task output filenames.
         ''')
 
-        with Conf('main loop'):
+        with Conf('main loop'):  # noqa: SIM117 (keep same format)
             with Conf('<plugin name>'):
                 Conf('interval', VDR.V_INTERVAL)
 
@@ -639,7 +639,8 @@ with Conf(
                      foo:submit-fail => bar
             ''')
 
-    with Conf('runtime', desc='''
+    with Conf('runtime',  # noqa: SIM117 (keep same format)
+              desc='''
         This section is used to specify how, where, and what to execute when
         tasks are ready to run. Common configuration can be factored out in a
         multiple-inheritance hierarchy of runtime namespaces that culminates
@@ -1538,7 +1539,7 @@ def upgrade_graph_section(cfg: Dict[str, Any], descr: str) -> None:
     """Upgrade Cylc 7 `[scheduling][dependencies][X]graph` format to
     `[scheduling][graph]X`."""
     # Parsec upgrader cannot do this type of move
-    try:
+    with contextlib.suppress(KeyError):
         if 'dependencies' in cfg['scheduling']:
             msg_old = '[scheduling][dependencies][X]graph'
             msg_new = '[scheduling][graph]X'
@@ -1568,8 +1569,6 @@ def upgrade_graph_section(cfg: Dict[str, Any], descr: str) -> None:
                         f' * (8.0.0) {msg_old} -> {msg_new} - for X in:\n'
                         f"       {', '.join(sorted(keys))}"
                     )
-    except KeyError:
-        pass
 
 
 def upgrade_param_env_templates(cfg, descr):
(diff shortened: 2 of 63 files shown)


cylc/flow/cfgspec/globalcfg.py
```diff
@@ -708,11 +708,12 @@
             Conf('hosts', VDR.V_STRING_LIST, ['localhost'])
 
     # Platform Groups
-    with Conf('platform groups'):
+    with Conf('platform groups'):  # noqa: SIM117 (keep same format)
         with Conf('<group>'):
```
oliver-sanders, line 712: We could combine these two `with` statements into a single one, however, since all of the others in this file are written individually this would break the pattern somewhat.

cylc/flow/cfgspec/workflow.py
```diff
@@ -1538,7 +1539,7 @@ def upgrade_graph_section(cfg: Dict[str, Any], descr: str) -> None:
     """Upgrade Cylc 7 `[scheduling][dependencies][X]graph` format to
     `[scheduling][graph]X`."""
     # Parsec upgrader cannot do this type of move
-    try:
+    with contextlib.suppress(KeyError):
```
oliver-sanders, line 1542:
```python
with suppress(Something):
    something()
```

Is generally cleaner and more readable than:

```python
try:
    something()
except Something:
    pass
```

I'm not so fussed, if this causes us undue pain we can always disable the rule, however, it does have a potential advantage from the perspective of code coverage so thats nice.

cylc/flow/cycling/__init__.py
```diff
@@ -30,11 +30,17 @@ def parse_exclusion(expr):
                         "permitted" % expr)
     else:
         remainder, exclusions = expr.split('!')
-        if ',' in exclusions:
-            if (not exclusions.strip().startswith('(') or not
```
oliver-sanders, line 34: Needlessly nested if's

cylc/flow/cycling/__init__.py
```diff
@@ -417,7 +423,7 @@ def __hash__(self) -> int:
         pass
 
 
-class ExclusionBase(object, metaclass=ABCMeta):
+class ExclusionBase(metaclass=ABCMeta):
```
oliver-sanders, line 426: Python class.

cylc/flow/daemonize.py
```diff
@@ -70,22 +70,24 @@ def daemonize(schd):
...
-                        if workflow_url and pub_url:
-                            break
-                        elif ' ERROR -' in line or ' CRITICAL -' in line:
-                            # ERROR and CRITICAL before workflow starts
-                            try:
-                                sys.stderr.write(open(logfname).read())
-                                sys.exit(1)
-                            except IOError:
-                                sys.exit("Workflow schd program exited")
+                    with open(logfname, 'r') as logfile:
```
oliver-sanders, line 73: `with open()` is the better pattern.

Note that this:

```
with open() as h:
    h.write()
```

Is *not* equivalent to this:

```python
h = open()
h.write()
h.close()
```

As the `with open` brings extra error handling to the table.

cylc/flow/dbstatecheck.py
```diff
@@ -136,16 +136,16 @@ def task_state_met(self, task, cycle, status=None, message=None):
...
-            for outputs_str, in res:
-                for value in json.loads(outputs_str).values():
-                    if message == value:
-                        return True
-            return False
+            return any(
+                message == value
+                for outputs_str, in res
+                for value in json.loads(outputs_str).values()
+            )
```
oliver-sanders, line 143: `any` is faster and more readable.

cylc/flow/network/resolvers.py
```diff
@@ -187,7 +192,7 @@ def get_data_elements(flow, nat_ids, element_type):
     ]
 
 
-class BaseResolvers:
+class BaseResolvers:  # noqa: SIM119 (no real gain + mutable default)
```
oliver-sanders, line 195: flake8-simplify is keen on dataclasses.

I.E. this:

```python
from dataclasses import dataclass

@dataclass
class Foo:
    a: str
    b: str
```

Rather than this:

```python
class Foo:

    def __init__(self, a: str, b: str):
        self.a = a
        self.b = b
```

Dataclasses have additional benefits because they add other methods like `__str__` and `__eq__` which saves a lot of boilerplate.

Dataclasses are generally a good idea, however, a lot of our classes (like this one) don't fit the pattern so well.

Nice to be reminded that we could use a dataclass, just stick a `noqa` in if not appropriate.

cylc/flow/cfgspec/workflow.py
hjoliver: I quite like the `with suppress...` version.

hjoliver (approved):
:+1: Quite a big change but just straightforward switch to other valid coding constructs, so one review will do on this.

cylc/flow/cfgspec/workflow.py
MetRonnie: One thing to bear in mind is `with suppress` is slower, so we might want to avoid it sometimes
```python

def f():
    try:
        int('a')
    except ValueError:
        pass
    
def g():
    with suppress(ValueError):
        int('a')
```
```python
In [5]: %timeit f()
976 ns ± 46 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [6]: %timeit g()
1.63 µs ± 32.8 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

MetRonnie:
Is this not going to be added to GH Actions?

oliver-sanders:
(we can use the flake8 `--select` option to make it more explicit, otherwise flake8 runs all installed flakes)

PR #4272, opened 2021-06-17
Constrain window by final cycle point

### LINKED ISSUE #4255: Datastore window: stop at final cycle point
(what it says)

dwsutherland:
These changes close #4255

Constrain the data store n-window by the final cycle point (FCP)... (ICP doesn't appear to be an issue)

There would be two ways of doing this:
- change the `taskdef.py` function that generates children
- filter at the window expansion for children whose cycle point extends beyond the FCP.

I chose the latter, because we may wish to (or already do) record the workflow state of tasks beyond the FCP for the purpose of restarting the workflow with an extended FCP.

~~The issue only applies to those tasks in the `[scheduling][[special tasks]]sequential` section, not those defined in the graph as sequential.~~

The test `tests/functional/graphql/01-workflow.t` will fail with the filter commented out.

**Requirements check-list**
  `conda-environment.yml`.

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -698,7 +698,10 @@ class DataStoreMgr:
             self, s_id, s_node, items, active_id, flow_label, reflow,
             edge_distance, descendant=False, is_parent=False):
         """Construct nodes/edges for children/parents of source node."""
+        final_point = self.schd.config.final_point
         for t_name, t_point, _ in items:
+            if t_point > final_point:
+                continue
             t_node = f'{t_name}.{t_point}'
             t_id = (
                 f'{self.workflow_id}{ID_DELIM}{t_point}{ID_DELIM}{t_name}')
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -82,6 +82,9 @@ workflow's public database file was not closed properly.
 
 ### Fixes
 
+[#4272](https://github.com/cylc/cylc-flow/pull/4272) - Workflow visualisation
+data (data-store) now constrained by final cycle point.
+
 [#4248](https://github.com/cylc/cylc-flow/pull/4248)
  - Fix parameter expansion in inherited task environments.
 
--- a/tests/functional/graphql/01-workflow.t
+++ b/tests/functional/graphql/01-workflow.t
@@ -92,8 +92,8 @@ cmp_json "${TEST_NAME}-out" "${TEST_NAME_BASE}-workflows.stdout" << __HERE__
                 "title": "foo",
                 "description": "bar"
             },
-            "newestActiveCyclePoint": "1",
-            "oldestActiveCyclePoint": "1",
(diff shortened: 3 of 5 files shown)


hjoliver:
> The issue only applies to those tasks in the [scheduling][[special tasks]]sequential section, not those defined in the graph as sequential.

That doesn't seem to be true, e.g. @dpmatthews example from Element chat shows an `n=1` task in `2000-01-04`:
```
[scheduling]
    initial cycle point = 2000-01-01
    final cycle point = 2000-01-03
    [[graph]]
        T00 = "task1[-PT24H] => task1"
[runtime]
    [[task1]]
        script = sleep 10
```
Same again for a variation to avoid sequentials: `task1[-PT24H] => task2 => task1`

But ... this PR fixes it :tada:

cylc/flow/data_store_mgr.py
```diff
@@ -698,7 +698,10 @@ def _expand_graph_window(
             self, s_id, s_node, items, active_id, flow_label, reflow,
             edge_distance, descendant=False, is_parent=False):
         """Construct nodes/edges for children/parents of source node."""
+        final_point = self.schd.config.final_point
         for t_name, t_point, _ in items:
+            if t_point > final_point:
+                continue
```
hjoliver, line 704: I was going to say the `final_point` local variable is unnecessary here, but I guess there would be a small performance hit from looking up `self.schd.config` attributes inside the loop :thinking:

hjoliver (approved):
looks good :+1:

cylc/flow/data_store_mgr.py
dwsutherland: Yeah, I'm not certain what's fastest.. I suppose I could do a test to find out.

hjoliver: I googled it, it is faster to look up a local variable than nested attributes. It just requires more low-level ops. Inside a loop that can matter.  In this case I doubt it'll be noticeable either way.

dwsutherland:
> > The issue only applies to those tasks in the [scheduling][[special tasks]]sequential section, not those defined in the graph as sequential.
> 
> That doesn't seem to be true, e.g. @dpmatthews example from Element chat shows an `n=1` task in `2000-01-04`:

I was just looking on startup (`cylc play --pause fox/run1`) with ICP = FCP, but didn't test as the workflow progresses to an FCP further out... The fix, as you've seen, is agnostic to either case.

cylc/flow/data_store_mgr.py
dwsutherland: Yeah, how many children can a node have 100? 1000?

hjoliver: Let's leave it as-is, just in case.

oliver-sanders: This way is better for performance :+1:.

oliver-sanders (approved):

Issue #3379, opened 2019-09-25
Replace the old `[visualisation]` section


oliver-sanders:
See the Cylc-UI counterpart - https://github.com/cylc/cylc-ui/issues/263

Open up the ability for minimal customisation of the appearance of tasks / families.

The main use case for this will be visually distinguishing between different groups of tasks (e.g. ones belonging to a particular model or system).

Now that task icons are not colour dependent we are free to use colour for this, see the design issue (#87) for more info.

This is roughly analogous to the old `[visualisation]` section in Cylc 7 which allowed users to specify custom Graphviz attributes.

hjoliver:
(See my comment on the UI issue: https://github.com/cylc/cylc-ui/issues/263#issuecomment-535243975)

PR #4273, opened 2021-06-18
Remove obsolete visualization section.

hjoliver:
This is a ~~small~~ change ~~with no associated Issue~~. Partially addresses #3379.  May c-l-o-s-e #3696 (final check needed, of that umbrella issue).

- remove obsolete `[visualization]` config section. 
  - future Cylc 8 viz config will be different. The old config was very much tailored for graphviz. 
- simplify `get_graph_raw()` which was keeping track of family collapse state for the old UI. It now has a single argument to specify which families to collapse, and does not remember the collapsed state
- modify `cylc graph` and `cylc list`, and functional tests that use them, accordingly (slightly enhanced too: they can take an interval for the stop point).

The PR would have been a lot smaller if I'd retained `[visualization]initial/final cycle point` which was being used by most of the functional tests that use `cylc graph --reference` ... but, better to remove the whole `[visualization]` section now.  `flow.cylc` should define the whole graph, and `cylc graph` etc. should just specify what part of it they want to see.

**Requirements check-list**
- [x] Applied any dependency changes to both `setup.py` and `conda-environment.yml`. [None]
- [x] Already covered by existing tests. (Removed some `[vis]` tests)
- [x] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/253


hjoliver:
Draft while I fix tests that use visualization settings to define the graph extent.

hjoliver:
@oliver-sanders maybe one review will do as this is mostly functional test tweaks. Assign another reviewer if you like though.

cylc/flow/config.py
```diff
@@ -180,14 +181,11 @@ def __init__(
...
         self.start_point: Optional['PointBase'] = None
         self.final_point: Optional['PointBase'] = None
-        self.first_graph = True
         self.clock_offsets = {}
         self.expiration_offsets = {}
         self.ext_triggers = {}  # Old external triggers (client/server)
         self.xtrigger_mgr = xtrigger_mgr
         self.workflow_polling_tasks = {}  # type: ignore # TODO figure out type
-        self._last_graph_raw_id: Optional[tuple] = None
-        self._last_graph_raw_edges = []  # type: ignore # TODO figure out type
```
oliver-sanders, line 190: This takes out the old graph cashing system, I guess we don't need it anymore?

(could quickly reimplement via `lru_cache` if ever needed again [via a partial or something like that to ignore the non-serialisable `self` argument]).

hjoliver: I meant to comment on this, sorry. I was thinking we'd need something different for the Cylc 8 graph view, with delta updates via the UIS.  And even if we have to regenerate all graph edges from scratch on each change, it'll be push (to a single UIS) so we still won't need caching in case of another request for the full graph.  Haven't properly thought it through though (damn, that's a lot of *ough* in a row). Easy enough to restore those lines for now, just in case. What do you think?

wxtim:
Very likely that this closes #3696 - after this has gone in that will need a final check, and can then be closed... 🎉

hjoliver:
> Very likely that this closes #3696 - after this has gone in that will need a final check, and can then be closed... tada

Good spotting @wxtim - I'll update the description above.

cylc/flow/config.py
oliver-sanders: Yeah, not too sure how we are going to unroll the graph in the n>1 window, ideally via some incremental approach at the UIS (would require the UIS to load the config)? Nothing much is using `get_graph_raw` at the moment so I think we can afford to loose cashing.

Easy to reimplement using standard tooling:

```python
@staticmethod
@lru_cache(cache_size=1)
def _get_graph_raw(_, *args):
    ...

def get_graph_raw(self, *args):
    return self._get_graph_raw(id(self), *args)
```

cylc/flow/etc/syntax/cylc.lang
```diff
@@ -90,7 +90,6 @@
         <keyword>xtriggers</keyword>
         <keyword>work sub-directory</keyword>
         <keyword>warning handler</keyword>
-        <keyword>visualization</keyword>
```
wxtim, line 93: I think that there are a number of other keywords that should be removed. Try running:
`./etc/bin/syntax_generator.py`.

wxtim (approved):
- [x] read it
- [x] tried it on random local workflows.

cylc/flow/etc/syntax/cylc.lang
hjoliver: You're right; last commit cleans this up.

oliver-sanders (approved):

oliver-sanders:
Moved the milestone forward on #3696

Issue #3731, opened 2020-08-04
graphql: subscriptions at the scheduler


oliver-sanders:
Prerequisite to #3527 

Implement the following component of the [interfaces and protocols diagram](https://github.com/cylc/cylc-admin/blob/master/docs/img/interfaces-and-protocols.png):

![rect4117](https://user-images.githubusercontent.com/16705946/89286489-3696d500-d64a-11ea-936c-675d7f413f2d.png)

Running GraphQL subscriptions over ZMQ/TCP is non-standard so this will require a bit of work at our end.

This may just involve patching in our pub-sub / req-rep endpoints to the [existing websockets functionality](https://github.com/graphql-python/graphql-ws/blob/master/graphql_ws/websockets_lib.py) as conceptually only the connections and transport differ.

Perhaps we should take this opportunity to consider whether there is a nice simplification of the existing req-rep / pub-sub duo, perhaps we could merge them with xpub-xsub or something else? Ping @dwsutherland.

**Pull requests welcome!**

dwsutherland:
From:
https://zguide.zeromq.org/docs/chapter5/

>Killing back-chatter is essential to real scalability. With pub-sub, it’s how the pattern can map cleanly to the PGM multicast protocol, which is handled by the network switch. In other words, subscribers don’t connect to the publisher at all, they connect to a multicast group on the switch, to which the publisher sends its messages.
>
>When we remove back-chatter, our overall message flow becomes much simpler, which lets us make simpler APIs, simpler protocols, and in general reach many more people. But we also remove any possibility to coordinate senders and receivers. What this means is:
>
>- Publishers can’t tell when subscribers are successfully connected, both on initial connections, and on reconnections after network failures.
>
>- Subscribers can’t tell publishers anything that would allow publishers to control the rate of messages they send. Publishers only have one setting, which is full-speed, and subscribers must either keep up or lose messages.
>
>- Publishers can’t tell when subscribers have disappeared due to processes crashing, networks breaking, and so on.
>
>The downside is that we actually need all of these if we want to do reliable multicast. The ZeroMQ pub-sub pattern will lose messages arbitrarily when a subscriber is connecting, when a network failure occurs, or just if the subscriber or network can’t keep up with the publisher.
>
[...]

Issue #3329, opened 2019-08-29
zmq: implement REQ-REP + PUB-SUB comms


oliver-sanders:
Issue migrated from #2978

At present the suite runtime API is REQ-REP.

1. Client sends a request.
2. Server sends a response.

This pattern works fine for information requests, but for mutations the client usually just receives an unhelpful "command queued" message. There is no mechanism to communicate command output to the user.

To solve this we could add a PUB-SUB pattern on-top of the REQ-REP whereby:

1. Client sends a request.
2. Server sends a subscription ID.
3. Client subscribes to the subscription.
4. Server publishes updates until the command has completed.

From the client side this might look like:

```
SEND stop_suite
RECV <subid>
RECV <subid> "command queued"
RECV <subid> "waiting for X active tasks to complete before"
RECV <subid> "waiting for suite shutdown handlers to complete"
RECV <subid> "shutting down"
RECV <subid> "command succeeded"
```

This pattern should be used for mutations (and not for information requests).

TomekTrzeciak:
Relevant for #2798.

dwsutherland:
PUB/SUB has been merged as part of the WSchd <=> UIS data store sync PR .. So available as an option for mutation feedback

oliver-sanders:
See the older https://github.com/cylc/cylc-flow/issues/423 which proposes much the same thing.

dwsutherland:
See #4274

hjoliver:
Does that close this?

dwsutherland:
> Does that close this?

Yes, it does.

hjoliver:
(My bad - I thought I looked and the two issues hadn't been connected)

oliver-sanders:
It may be possible to achieve this on a single socket using router-dealer or some fancy combination of req-rep and router-dealer (see https://zguide.zeromq.org/docs/chapter3/), but this will require investigation.

Note that req-rep is synchronous (e.g. "kill" followed by "trigger" will be queued in that order), whereas router-dealer is asynchronous ("kill" followed by "trigger" could be queued in either order) so we may need to leverage both (because queue order is somewhat important to us, although we have never claimed to guarantee it).

Adding a new socket is not desirable (port efficiency matters), ideally we would look to consolidate the pub-sub socket too (see also xpub-xsub).

PR #4274, opened 2021-06-18
Node (Network) Coordinated Subscriptions (part 1/3) - Centralise REP and PUB into single runtime server

### LINKED ISSUE #3564: Calling publisher coroutine from another thread
While reviewing #3492 I noticed the publisher was being called differently.
I had a closer look after re-basing a branch and realised the publisher is running in a different thread, and awaiting the `publish` coroutine from the main thread:
https://github.com/cylc/cylc-flow/blob/e904325da6189dc87fc7aad92fd81f0163c7aa00/cylc/flow/scheduler.py#L1608-L1609

This isn't desirable (although may not be an issue):
https://docs.python.org/3/library/asyncio-dev.html#concurrency-and-multithreading

So, to be safe, we have two options here:

1. Use a threadsafe call to this coroutine from the main thread (i.e. [`loop.call_soon_threadsafe()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.call_soon_threadsafe)).
2. As the `publisher.py` has no blocking calls or loops (unlike the REQ/RES `server.py`), we could run the publisher in the main thread and keep the awaited publish the same.

In my opinion, and assuming publishing is not that intensive/blocking; we should go with option 2, as it simplifies the workflow scheduler somewhat (if extra thread isn't needed, why use it).

Thoughts? (ping @oliver-sanders)

**Pull requests welcome!**

### LINKED ISSUE #4706: Pytest Unhandled Thread Exception Warning
We should fix this warning in `tests/unit/network/test_publisher.py::test_start_stop`.

```
tests/unit/network/test_publisher.py::test_start_stop
  /opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/_pytest/threadexception.py:73: PytestUnhandledThreadExceptionWarning: Exception in thread Thread-1
  
    File "/home/runner/work/cylc-flow/cylc-flow/cylc/flow/network/__init__.py", line 197, in _socket_bind
      server_public_key, server_private_key = zmq.auth.load_certificate(
    File "/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/zmq/auth/certs.py", line 98, in load_certificate
      raise IOError("Invalid certificate file: {0}".format(filename))
  OSError: Invalid certificate file: /home/runner/cylc-run/beef/.service/server.key_secret
  
    File "/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/threading.py", line 973, in _bootstrap_inner
      self.run()
    File "/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/threading.py", line 910, in run
      self._target(*self._args, **self._kwargs)
    File "/home/runner/work/cylc-flow/cylc-flow/cylc/flow/network/__init__.py", line 165, in _start_sequence
      self._socket_bind(*args, **kwargs)
    File "/home/runner/work/cylc-flow/cylc-flow/cylc/flow/network/__init__.py", line 206, in _socket_bind
      raise ServiceFileError(
  cylc.flow.exceptions.ServiceFileError: IO error opening server's private key from /home/runner/cylc-run/beef/.service/server.key_secret.
  
    warnings.warn(pytest.PytestUnhandledThreadExceptionWarning(msg))
```

**Question though - why is this exception not bubbling up and causing the test to fail?**

>These features only work on Python>=3.8.
>
>Unhandled exceptions are exceptions that are raised in a situation in which they cannot propagate to a caller. The most common case is an exception raised in a [`__del__`](https://docs.python.org/3/reference/datamodel.html#object.__del__) implementation.
>
>Unhandled thread exceptions are exceptions raised in a [Thread](https://docs.python.org/3/library/threading.html#threading.Thread) but not handled, causing the thread to terminate uncleanly.
>
>Both types of exceptions are normally considered bugs, but may go unnoticed because they don’t cause the program itself to crash. Pytest detects these conditions and issues a warning that is visible in the test run summary.

https://docs.pytest.org/en/latest/how-to/failures.html#warning-about-unraisable-exceptions-and-unhandled-thread-exceptions

**Pull requests welcome!**

dwsutherland:
This change:
- partially addresses #3731
- partially addresses #3329 
- closes #3564
- closes #4706

This PR completes step one of the following changes needed for GraphQL subscriptions at the workflow:
1. Centralise REP and PUB into single runtime server.
2. Setup coordination scheme for REQ of PUB topics to SUB to.
3. Implement a [GraphQL-WS](https://github.com/graphql-python/graphql-ws) style handler for GraphQL subscriptions.

In order to coordinate between PUB and REP sockets it makes sense to revive the single runtime server concept to initialize both and manage requests (REQ) for publishing (PUB) topics to subscribe (SUB) to.

- [x] Step 1 - Can now publish in response to a REQ (in addition to REP).

This also results in less network code in the scheduler.

At present we have each network object (publisher, replier/server) instantiated by the scheduler in the main thread, then their startup/socket-bind happening in a separate thread:
image

This PR reduces the number of threads, with the publisher, subscriber coordinated in the same thread by the parent server object (instantiated by the scheduler still):
image

Published articles/data/deltas are now put into a (threadsafe) queue, instead of called directly by the main thread...

Since the threading is setup by the scheduler now, the client no longer needs to import the threading library.

Running the complex workflow takes about the same time as the current master:
```
        Command being timed: "cylc play complex --color=never -n"
        User time (seconds): 515.45
        System time (seconds): 36.81
        Percent of CPU this job got: 36%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 25:18.28
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 117568
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 20396247
        Voluntary context switches: 182301
        Involuntary context switches: 172133
        Swaps: 0
        File system inputs: 8
        File system outputs: 3725008
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

**Requirements check-list**
  `conda-environment.yml`.


MetRonnie:
Ok if this is pushed back to rc3?

cylc/flow/network/publisher.py
```diff
@@ -70,7 +69,6 @@ def _socket_options(self):
     def _bespoke_stop(self):
         """Bespoke stop items."""
         LOG.debug('stopping zmq publisher...')
-        self.stopping = True
```
MetRonnie, line 73: Method is now a dud?

cylc/flow/network/server.py
```diff
@@ -148,89 +149,122 @@ def __init__(self, schd, context=None, barrier=None,
...
+        self.stopped = False
+
+        self.operate()
+
+    async def stop(self, reason):
+        """Stop the TCP servers, and clean up authentication."""
+        self.queue.put('STOP')
+        if self.thread and self.thread.is_alive():
+            while not self.stopping:
+                sleep(self.STOP_SLEEP_INTERVAL)
```
MetRonnie, line 217: I think this should be an async sleep to avoid blocking, unless there is a specific reason to block?
```suggestion
                await asyncio.sleep(self.STOP_SLEEP_INTERVAL)
```

cylc/flow/network/server.py
```diff
@@ -148,89 +149,122 @@ def __init__(self, schd, context=None, barrier=None,
...
-    def _bespoke_stop(self):
-        """Stop the listener and Authenticator.
+            # Publish all requested/queued.
+            while self.publish_queue.qsize():
+                articles = self.publish_queue.get()
+                self.loop.run_until_complete(self.publisher.publish(articles))
 
-        Overwrites Base method.
+            # Yield control to other threads
+            sleep(self.OPERATE_SLEEP_INTERVAL)
```
MetRonnie, line 255: As above
```suggestion
            await asyncio.sleep(self.OPERATE_SLEEP_INTERVAL)
```

cylc/flow/network/server.py
```diff
@@ -69,7 +73,7 @@ def filter_none(dictionary):
     }
 
 
-class WorkflowRuntimeServer(ZMQSocketBase):
+class WorkflowRuntimeServer:
     """Workflow runtime service API facade exposed via zmq.
 
     This class contains the Cylc endpoints.
```
MetRonnie, line 79: Docstring args now out of date

cylc/flow/network/server.py
```diff
@@ -148,89 +149,122 @@ def __init__(self, schd, context=None, barrier=None,
...
 
-    def _socket_options(self):
-        """Set socket options.
+        self.queue = Queue()
+        self.publish_queue = Queue()
+        self.stopping = False
+        self.stopped = True
 
-        Overwrites Base method.
+    def configure(self):
```
MetRonnie, line 157: Is there any reason to keep this as a separate method and not put all of this straight into `__init__()`?

cylc/flow/network/replier.py
```diff
@@ -0,0 +1,93 @@
...
+            else:
+                # success case - serve the request
+                res = self.server.receiver(message)
+                # send back the string to bytes response
+                if isinstance(res.get('data'), bytes):
+                    response = res['data']
+                else:
+                    response = encode_(res).encode()
+                self.socket.send(response)
+
```
MetRonnie, line 89:
```suggestion

            await asyncio.sleep(0)  # avoid blocking event loop
            
```

cylc/flow/network/replier.py
```diff
@@ -0,0 +1,93 @@
...
+from cylc.flow.network import encode_, decode_, ZMQSocketBase
+
+
+class WorkflowReplier(ZMQSocketBase):
+    """Initiate the REP part of a ZMQ REQ-REP pattern.
+
+    This class contains the logic for the ZMQ message replier.
+
+    Usage:
+        * Define ...
```
MetRonnie, line 33: Incomplete?

cylc/flow/scheduler.py
```diff
@@ -621,7 +570,10 @@ async def run_scheduler(self):
                     self
                 )
             )
-            await self.publisher.publish(self.data_store_mgr.publish_deltas)
+            self.server.publish_queue.put(
+                self.data_store_mgr.publish_deltas)
+            # Yield control to other threads
+            sleep(0)
```
MetRonnie, line 576: As above
```suggestion
            await asyncio.sleep(0)
```

cylc/flow/scheduler.py
```diff
@@ -1634,8 +1595,10 @@ async def update_data_structure(self):
             # Publish updates:
             if self.data_store_mgr.publish_pending:
                 self.data_store_mgr.publish_pending = False
-                await self.publisher.publish(
+                self.server.publish_queue.put(
                     self.data_store_mgr.publish_deltas)
+                # Yield control to other threads
+                sleep(0)
```
MetRonnie, line 1601: As above
```suggestion
                await asyncio.sleep(0)
```

MetRonnie (commented):
Partial review (still haven't looked at tests or tested out)

cylc/flow/network/server.py
dwsutherland: `time.sleep` is used to yield control to other threads (halting the current thread).. Whereas, I believe `asyncio.sleep` just allows other ayncio execution for the duration.

cylc/flow/network/server.py
dwsutherland: `time.sleep` is used to yield control to other threads (halting the current thread).. Whereas, I believe `asyncio.sleep` just allows other ayncio execution for the duration.

cylc/flow/network/replier.py
dwsutherland: Nah, it's not an asyncio function, and on a separate thread to the scheduler.. This loop can run until no more messages are available, also the REQ/REP means you can't send or receive twice in a row.

cylc/flow/scheduler.py
dwsutherland: `time.sleep` is used to yield control to other threads (halting the current thread).. Whereas, I believe `asyncio.sleep` just allows other ayncio execution for the duration.

cylc/flow/scheduler.py
dwsutherland: Just as the comment above it suggests:

`time.sleep` is used to yield control to other threads (halting the current thread).. Whereas, I believe `asyncio.sleep` just allows other ayncio execution for the duration.

tests/integration/test_replier.py
```diff
@@ -0,0 +1,36 @@
...
+        async with timeout(2):
+            # wait for the server to consume the STOP item from the queue
+            while True:
+                if one.server.replier.queue.empty():
+                    break
+                await asyncio.sleep(0.01)
+        # ensure the server is "closed"
+        with pytest.raises(ValueError):
+            one.server.replier.queue.put('foobar')
+            one.server.replier.listener()
```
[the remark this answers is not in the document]
dwsutherland: The `queue.put()` won't raise the `ValueError`, the listener pulling the value from the queue will though.

cylc/flow/network/publisher.py
dwsutherland: Still called, and adds that log entry for publisher stop  (the Base class (`ZMQSocketBase`) has the `stop()` method for all the derivatives) .

MetRonnie: In that case is there any reason to override the `ZMQSocketBase._bespoke_stop()`? I can't see the harm in it setting `self.stopping = True` even if it's not used

cylc/flow/network/server.py
MetRonnie: OK, would be good to have a comment mentioning this somewhere

cylc/flow/network/replier.py
MetRonnie: OK, would be good to have a comment explaining this imho

tests/integration/test_replier.py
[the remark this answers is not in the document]
MetRonnie: Ah, then the `queue.put()` should be above the context manager?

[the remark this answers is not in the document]
dwsutherland: Yeah, perhaps it ought to be.

cylc/flow/network/publisher.py
dwsutherland: Fair enough.

cylc/flow/network/replier.py
dwsutherland: Incomplete.

dwsutherland: Filled out (same with the publisher).

cylc/flow/network/server.py
dwsutherland: Yes no reason, configure was called right after server instantiation... Combined.

cylc/flow/network/server.py
dwsutherland: updated.

cylc/flow/network/server.py
dwsutherland: I put more into the docstring of the method.

cylc/flow/network/publisher.py
dwsutherland: reinstated.

MetRonnie: Ah I thought there was no point in overriding this method in `WorkflowPublisher` at all because I didn't spot that the log message was slightly different for `ZMQSocketBase`

https://github.com/cylc/cylc-flow/blob/f9efea2e649ffd0ffc1fe2cfc63a122034813412/cylc/flow/network/__init__.py#L293-L296

https://github.com/cylc/cylc-flow/blob/f9efea2e649ffd0ffc1fe2cfc63a122034813412/cylc/flow/network/publisher.py#L69-L72

Could just do
```python
class ZMQSocketBase:
    ...
    def _bespoke_stop(self):
        """Bespoke stop items."""
        LOG.debug(f'stopping {type(self).__name__}...')
        self.stopping = True
```
and not have to override. Not a big deal

cylc/flow/network/server.py
MetRonnie: 👍  But now there are several properties that are instantiated as `None` only to be redefined a few lines down, like `self.zmq_context`. Would be nice to tidy those up.

cylc/flow/network/publisher.py
dwsutherland: Yeah, the purpose for the override is not the logging (the replier does something different).. So happy as is..

cylc/flow/scheduler.py
MetRonnie: The distinction of yielding to event loop vs other threads is a bit subtle imho, something I missed the first time round. I think it pays to be explanatory about it, something like
```python
# Non-async sleep - yield to other threads rather than event loop

Comment (dwsutherland) [reply cylc/flow/scheduler.py:1602]: Well, I commented what we're doing with the `time.sleep`... 
What's `asyncio.sleep` got to do with it?
Why would devs assume they're the same and that it needs to be changed?

Comment (MetRonnie) [reply cylc/flow/scheduler.py:1602]: > What's `asyncio.sleep` got to do with it?

Because most patterns involving a `while True` inside an async function should have `asyncio.sleep(short_duration)` in them to yield to the event loop.

As you've explained, in this case we want to yield to other threads instead. But the fact that it took some explanation for me to notice the difference is why I'm requesting a comment along those lines.

Comment (dwsutherland) [reply cylc/flow/scheduler.py:1606]: Amended.. I appreciated your code hygiene 👍

**Review by MetRonnie — APPROVED** [review]
Thanks

**Review by oliver-sanders — COMMENTED** [review]
Looks good, need to go through this one very thoroughly and do some hands-on-testing, so far so good.

File: cylc/flow/network/server.py [hunk 11]
```diff
@@ -148,89 +147,123 @@ def __init__(self, schd, context=None, barrier=None,
...
+            workflow_files.KeyType.PUBLIC,
+            workflow_files.KeyOwner.CLIENT,
+            workflow_srv_dir=workflow_srv_dir)
+        self.client_pub_key_dir = client_pub_keyinfo.key_path
+
+        # Initial load for the localhost key.
+        self.curve_auth.configure_curve(
+            domain='*',
+            location=(self.client_pub_key_dir)
+        )
```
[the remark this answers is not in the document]
MetRonnie: At first in this PR I think this stuff was in a separate `configure()` method, though that was getting called immediately after `__init__()`. In response to my review it was moved into `__init__()` - https://github.com/cylc/cylc-flow/pull/4274#discussion_r847993672. But perhaps `start()` would be a better place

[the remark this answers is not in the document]
oliver-sanders: The keyinfo stuff is fine but I don't think the curve auth thread should be started until `.start()`?

cylc/flow/scheduler.py
```diff
@@ -1750,15 +1713,7 @@ async def _shutdown(self, reason: Exception) -> None:
...
-            self.server.stop()
-        if self.publisher:
-            await self.publisher.publish(
-                [(b'shutdown', str(reason).encode('utf-8'))]
-            )
-            self.publisher.stop()
-            self.publisher = None
-        if self.curve_auth:
-            self.curve_auth.stop()  # stop the authentication thread
+            await self.server.stop(reason)
```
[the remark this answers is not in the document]
MetRonnie: Introduced in a917e65948b80590a313b741bb7079437f1d0dcc, but not used apparently. Must have been accidentally leftover from an earlier attempt at doing something

cylc/flow/network/server.py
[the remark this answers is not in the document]
dwsutherland: Ya, we could do that... Will give it a try.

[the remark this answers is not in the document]
dwsutherland: Done.

cylc/flow/network/__init__.py
```diff
@@ -295,7 +275,6 @@ def _socket_options(self):
     def _bespoke_start(self):
         """Initiate bespoke items on thread at start."""
```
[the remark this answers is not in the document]
dwsutherland: Fixed, other places through out the file too.

MetRonnie (approved):

hjoliver:
> This also results in less network code in the scheduler.

Well that's nice :grin:

cylc/flow/network/__init__.py
```diff
@@ -293,22 +273,19 @@ def _socket_options(self):
...
         """Stop the server.
 
         Args:
-            stop_loop (Boolean): Stop running IOLoop of current thread.
+            stop_loop (Boolean): Stop running IOLoop.
 
         """
         self._bespoke_stop()
         if stop_loop and self.loop and self.loop.is_running():
             self.loop.stop()
```
MetRonnie, line 288: In my debugging, I have found that `self.loop` is still running after the call to `self.loop.stop()`. Any idea why this is the case? (Not specific to this branch, also happens on master).

dwsutherland: Probably waiting for loop to be clear of tasks.. What's still running is probably the question..
Can wait for them to complete after the call to stop:
```
    # Let's also finish all running tasks:
    pending = asyncio.Task.all_tasks()
    loop.run_until_complete(asyncio.gather(*pending))
```
or cancel them
```
    # Let's also cancel all running tasks:
    pending = asyncio.Task.all_tasks()
    for task in pending:
        task.cancel()
        # Now we should await task to execute it's cancellation.
        # Cancelled task raises asyncio.CancelledError that we can suppress:
        with suppress(asyncio.CancelledError):
            loop.run_until_complete(task)
```
(from these:
https://stackoverflow.com/a/37430948
https://stackoverflow.com/a/37345564
)

dwsutherland: Will give it a try

dwsutherland: how did you track the running processes on shutdown by the way? pycharm?

MetRonnie: I tried `asyncio.all_tasks(self.loop)` and yes there was a task `coro=<WorkflowSubscriber.subscribe() ...` still running. The scenario was: one workflow played then stopped in the UI. I didn't track anything all the way to shutdown, just inspected the loop while stopping the workflow (in VSCode). But by keeping a reference to the loop I noticed that after stopping it, playing and stopping a different workflow, the loop for the first workflow subscription was still running.

MetRonnie (approved):

MetRonnie (approved):

oliver-sanders (approved):
Works great, cheers.

I've poked and prodded this as best I'm able, works great.

PR #4275, opened 2021-06-18
config: make --sparse the defalt

### LINKED ISSUE #4143: config: consider making sparse the default
By default `cylc config` shows us all of the default values (mostly blank) which isn't especially helpful for most uses.

Since the command is changing anyway (`get-config`, `get-site-config` -> `config`) now would be a good opportunity to rethink this.

**Question:** Should we make `--sparse` the default and add a `--defaults` option (or something like that).

oliver-sanders:
closes #4143

Something we could do with getting into the next beta to avoid having two rounds of disruption.

**Requirements check-list**
  `conda-environment.yml`.

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -54,7 +54,7 @@ with Conf('global.cylc', desc='''
 
     To view your global config run::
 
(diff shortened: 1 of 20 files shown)


hjoliver (approved):
Thanks @oliver-sanders :+1:

Issue #4261, opened 2021-06-11
Review: Small Bugs Meta Issue


wxtim:
Meta issue, describing current known problems with Cylc Review in the 7.8.x branch, esp, but not exclusively, WRT Cylc 8 Workflows.

Some of these issues have their own issues open in GH, and in this case I have listed rather than describing the issue.

## Bugs
- [x] Log files with timezone (#4260 #4276 ). 
- [ ] Failed jobs from Cylc 8 workflows not showing (#4207 ).
- [x] Running Cylc 8 Workflows showing as stopped under some circumstances. (#4280 )
- [x] Syntax of `cylc.flow` files is not highlighted (#4264 ). 

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

wxtim:
All the otherwise undocumented issues referenced here are closed.
The unclosed one can't be reproduced, and has its own ticket anyway.

Issue #4260, opened 2021-06-11
Review: Not showing files ending +ZZ (a timezone)


wxtim:
**Describe the bug**
@MetRonnie  found the following bug whilst reviewing something else.

> However, when I try to access a specific log file (one with a datetime in the filename), I get a 404 page with this traceback, both for Cylc 7 & 8 workflows:
>```
> Traceback (most recent call last):
>   File "~/github/cylc-7/lib/cherrypy/_cprequest.py", line 679, in respond
>     response.body = self.handler()
>   File "~/github/cylc-7/lib/cherrypy/lib/encoding.py", line 230, in __call__
>     self.body = self.oldhandler(*args, **kwargs)
>   File "~/github/cylc-7/lib/cherrypy/_cpdispatch.py", line 60, in __call__
>     return self.callable(*self.args, **self.kwargs)
>   File "~/github/cylc-7/lib/cylc/review.py", line 666, in view
>     user, suite, path, path_in_tar=path_in_tar, mode=mode)
>   File "~/github/cylc-7/lib/cylc/review.py", line 505, in get_file
>     f_name = self._get_user_suite_dir(user, suite, path)
>   File "~/github/cylc-7/lib/cylc/review.py", line 930, in _get_user_suite_dir
>     return self._check_dir_access(path)
>   File "~/github/cylc-7/lib/cylc/review.py", line 789, in _check_dir_access
>     raise cherrypy.HTTPError(404)
> HTTPError: (404, None)
> ```
> Seeing as it happens for both 7/8, I guess this is a separate issue

I have confirmed that this is still an issue. Further investigation adding 
```diff
# lib/cylc/review.py@line772
- raise cherrypy.HTTPError(404)
+ raise cherrypy.HTTPError(404, 'File Not Found: {}'.format(path))
```
Shows that the `+` between the time and timezone has been replaced by a space. Depending on time constraints this problem.

**Release version(s) and/or repository branch(es) affected?**
7.8.x

**Steps to reproduce the bug**
- With Cylc 7.8.x or 8 run a workflow.
- With Cylc@7.8.x run `cylc review start`.
- Open `localhost:8080` in a browser and navigate to a workflow
- Enter either the cycles or task job listing pages then goto the cylc files dropdown (top rightish)
- Select a file in the form `??/??/log.YYYYMMDDTHHMMSS+ZZ`

**Expected behavior**
You should be shown a Cylc Log file.

**Pull requests welcome!**
[...]

PR #4276, opened 2021-06-21
Hack-y fix of failure to show log files with + in title

wxtim:
These changes address #4260 

- [x] Reviewer testing only.
- [x] Small changes to be summarized on closure of #4261 

I think that the problem is actually being caused by a problem in the HTML/JS - the + looks like it may be missing from the get command.

(3 Reviewers Tagged. 2nd can merge)

### DIFF
--- a/lib/cylc/review.py
+++ b/lib/cylc/review.py
@@ -661,8 +661,10 @@ class CylcReviewService(object):
     def view(self, user, suite, path, path_in_tar=None, mode=None,
              no_fuzzy_time="0"):
         """View a text log file."""
-        if re.search(r"log\..*\d\d\s\d\d$", path):
-            path = path.replace(' ', '+')
+        # Log files with +TZ in name end up with space instead of plus sign, so
+        # put plus sign back in (https://github.com/cylc/cylc-flow/issues/4260)
+        path = re.sub(r"(log\.\S+\d{2})\s(\d{2,4})$", r"\1+\2", path)
+        
         # get file or serve raw data
         file_output = self.get_file(
             user, suite, path, path_in_tar=path_in_tar, mode=mode)
--- a/lib/cylc/review.py
+++ b/lib/cylc/review.py
@@ -661,7 +661,7 @@ class CylcReviewService(object):
     def view(self, user, suite, path, path_in_tar=None, mode=None,
              no_fuzzy_time="0"):
         """View a text log file."""
-        if re.findall(r"log\..*\d\d\s\d\d$", path):
+        if re.searc(r"log\..*\d\d\s\d\d$", path):
             path = path.replace(' ', '+')
         # get file or serve raw data
         file_output = self.get_file(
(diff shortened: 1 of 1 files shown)


lib/cylc/review.py
```diff
@@ -661,6 +661,8 @@ def viewsearch(self, user, suite, path=None, path_in_tar=None, mode=None,
     def view(self, user, suite, path, path_in_tar=None, mode=None,
              no_fuzzy_time="0"):
         """View a text log file."""
+        if re.findall(r"log\..*\d\d\s\d\d$", path):
```
hjoliver, line 664: Don't need "find all" here?  (`re.match` or `re.search` would do).

lib/cylc/review.py
```diff
@@ -661,6 +661,8 @@ def viewsearch(self, user, suite, path=None, path_in_tar=None, mode=None,
...
+            path = path.replace(' ', '+')
```
hjoliver, line 665: Replacing space with plus in path?  (This doesn't seem to gel with the description of the problem: failing to display logs with `+` in the name ... however, I haven't gone any deeper than that!)

wxtim: I may be misunderstanding how the CherryPi app is working, but the `CylcReviewService.view` method is being called (by the GET?) with `+` removed from filenames in the form `log.YYYYMMDDTHHMM+ZZ`. This is a crude hack because I don't want to spend too long understanding what CherryPi is up to.

hjoliver: OK fair enough (I certainly haven't looked at what CherryPy is doing!)

lib/cylc/review.py
wxtim: Done

lib/cylc/review.py
```diff
@@ -661,6 +661,8 @@ def viewsearch(self, user, suite, path=None, path_in_tar=None, mode=None,
     def view(self, user, suite, path, path_in_tar=None, mode=None,
              no_fuzzy_time="0"):
         """View a text log file."""
+        if re.search(r"log\..*\d\d\s\d\d$", path):
+            path = path.replace(' ', '+')
         # get file or serve raw data
```
MetRonnie, line 666: This doesn't work for timezones with minutes (e.g. `+0430`)

Also I think `re.sub` is better suited to this
```suggestion
        # Log files with +TZ in name end up with space instead of plus sign, so
        # put plus sign back in (https://github.com/cylc/cylc-flow/issues/4260)
        path = re.sub(r"(log\.\S+\d{2})\s(\d{2,4})$", r"\1+\2", path)
        
        # get file or serve raw data
```

wxtim: Sound.

MetRonnie (approved):
Tested with timezones `+01`, `+0430` and `-03`, all work

hjoliver (approved):
:+1:

Issue #4133, opened 2021-03-18
clean: traceback removing local files


oliver-sanders:
Traceback from running `cylc clean` on a local flow.

The flow probably hadn't completely finished shutting down yet or NFS was keeping a file open or whatever. This sort of thing is quite common, especially on network filesystems.

At the least we should catch this error and present it more elegantly i.e:

```python
try:
     ...
except OSError as exc:
    # this just changes the presentation of the error
    raise CylcError(exc)
```

```console
$ cylc clean foo                               
INFO - Cleaning on local filesystem    
Traceback (most recent call last):    
  File "~/cylc/bin/cylc", line 33, in <module>    
    sys.exit(load_entry_point('cylc-flow', 'console_scripts', 'cylc')())    
  File "~/cylc/lib/python3.7/site-packages/click/core.py", line 829, in __call__    
    return self.main(*args, **kwargs)    
  File "~/cylc/lib/python3.7/site-packages/click/core.py", line 782, in main    
    rv = self.invoke(ctx)    
  File "~/cylc/lib/python3.7/site-packages/click/core.py", line 1066, in invoke    
    return ctx.invoke(self.callback, **ctx.params)    
  File "~/cylc/lib/python3.7/site-packages/click/core.py", line 610, in invoke    
    return callback(*args, **kwargs)    
  File "/net~/cylc-flow/cylc/flow/scripts/cylc.py", line 497, in main    
    execute_cmd(command, *cmd_args)    
  File "/net~/cylc-flow/cylc/flow/scripts/cylc.py", line 174, in execute_cmd    
    COMMANDS[cmd].resolve()(*args)    
  File "/net~/cylc-flow/cylc/flow/terminal.py", line 247, in wrapper    
    wrapped_function(*wrapped_args, **wrapped_kwargs)    
  File "/net~/cylc-flow/cylc/flow/scripts/clean.py", line 78, in main    
    init_clean(reg, opts)    
  File "/net~/cylc-flow/cylc/flow/suite_files.py", line 594, in init_clean    
    clean(reg)      
  File "/net~/cylc-flow/cylc/flow/suite_files.py", line 639, in clean    
    remove_dir(run_dir)    
  File "/net~/cylc-flow/cylc/flow/pathutil.py", line 252, in remove_dir    
    rmtree(path)    
  File "~/cylc/lib/python3.7/shutil.py", line 494, in rmtree    
    _rmtree_safe_fd(fd, path, onerror)    
[...]

Issue #3659, opened 2020-06-19
review use of rmtree


oliver-sanders:
Nasty bug with `shutil.rmtree` spotted by @datamel whilst reviewing https://github.com/cylc/cylc-flow/pull/3654.

Python's `shutil.rmtree` (like most things) does not work reliably on NFS filesystems. Other solutions may be affected by the same underlying issue.

The safest way I can think of handling it would is to be to give NFS a brief window to think about what it's done before taking it seriously and allowing traceback to propagate.

I'm not aware that this is causing issues in Cylc at the moment, but it is a potential bug, just yearning for the chance to surface when you really don't want it to so tagging this against `bug?`.

See [this comment](https://github.com/cylc/cylc-flow/pull/3654/files#r442920929) which includes examples of what other projects have done to deal with the NFS menace.

> - https://github.com/hashdist/hashdist/pull/116/files#diff-f8e43180100ba3086d46440b72e3e8f3
> - https://github.com/davidemms/OrthoFinder/commit/3d56a6e4870058f78721fa733ae27f60ab4d0c21#diff-4b6271e9a0c35f1d832125bcadf25034

MetRonnie:
Superseded by #4133? Although it doesn't suggest retrying the deletion after a delay

MetRonnie:
Closed by #5359

PR #4285, opened 2021-06-30
Infer run number & centralised reg parse

### LINKED ISSUE #3895: cli: infer run number where possible
Attempt after ~#3889~, #4230

The default behaviour of `cylc install` will be to install workflows into an automatically named run (i.e. `run<N>`).

This means that the CLI will become a little cumbersome e.g.

```bash
# before
$ cylc stop my-flow
# after
$ cylc stop my-flow/run1
```

To combat this we should make the relevant Cylc subcommands (those with a REG argument) run-dir aware. That is to say that if the REG provided is not a workflow (denoted by the presence of a `.service` directory) but does contain automatically named runs (denoted by the presence of a `runN` symlink), then default to whatever run `runN` is pointing at.

This change returns the CLI REG argument to the way it was before for most cases whilst providing the ability to control named runs individually where required.

### LINKED ISSUE #4230: workflow reg argument parsing
At the moment the parsing of the workflow argument is not fully centralised.

For the "offline" commands (e.g. `cylc diff`, `cylc config`, `cylc list`) we have the centralised `parse_workflow_arg` function.

For the "online" commands (e.g. `pause`, `stop`, `trigger`) we don't have any centralised function, we *usually* (but don't always) do `os.path.normpath(workflow)`.

I think we should pull this functionality into a single function something along the lines of:

```python
# cylc.flow.workflow_files
...
def parse_reg(reg, src=False, run=False) -> str:
    """
        Args:
            src: Reg can be a source workflow located by an absolute path.
            run: Reg can be a run directory located by relative path from ~/cylc-run/
    """
```

```python
# cylc.flow.scripts.pause
...
def main(reg, ...):
     reg = parse_reg(reg, run=True)
```

This is better for maintenance and would provide us with a good place to handle future changes to registration parsing:

* handling the runN symlink: #4228
* inferred run number: #3895
* universal identifier: #3592
* multi-workflow functionality (same issue as above)

MetRonnie:
These changes close #4230 and close #3895

- The `cylc.flow.workflow_files.parse_reg()` function provides a centralised way to process the `REG` or `WORKFLOW` arg in cylc commands. 
- Using this function provides inference of the latest run number, so instead of having to type `cylc pause foo/run3` you can do `cylc pause foo` assuming `run3` is the latest (also if you do `foo/runN` it processes this as `foo/run3`)
- "Offline" commands that work using a workflow file (or any file actually) need to use `parse_reg(reg, src=True)`. With `src=True` the `reg` can be:
  - a path to an existing file or a dir containing a `flow.cylc` file (either absolute, or relative to PWD, or relative to `~/cylc-run` (PWD takes precedence\*))
- "Online" commands that work by using graphQL/detecting workflow contact, need to use  `src=False` (the default). With `src=False` the `reg` can be:
  -  a path relative to `~/cylc-run` only (the dir need not exist but it can't be a path to a file)

**Requirements check-list**
  `conda-environment.yml`.

---

\*This is a change in behaviour. Previously it was the case that the path relative to `~/cylc-run` took precedence over PWD, apart from `.` for PWD itself, strangely. See https://github.com/cylc/cylc-flow/pull/4285#issuecomment-881568370 and https://github.com/cylc/cylc-flow/pull/4285#issuecomment-882661098


cylc/flow/workflow_files.py
```diff
@@ -1036,18 +1004,149 @@ def get_platforms_from_db(run_dir):
...
+            - relative path to the run dir from ~/cylc-run;
+            - absolute path to a run dir, source dir or workflow file (only
+                if src is True);
+            - '.' for the current directory (only if src is True).
+        src: Whether the workflow arg can be a workflow source (i.e. an
+            absolute path (which might not be in ~/cylc-run) and/or a
+            flow.cylc file (or any file really), or '.' for cwd).
+        require_flow_file: (Ignored if src is False, or if reg is already a
+            path to a file.) Whether a flow.cylc (or suite.rc) file is required
+            to be present in the path looked up from the workflow arg.
```
MetRonnie, line 1054: For some reason I thought this extra argument would be needed, but having gone through all the scripts I didn't find a single one where you need `src=True, require_flow_file=False`. @oliver-sanders might this be useful in the future or should I just get rid of it?

cylc/flow/workflow_files.py
```diff
@@ -1036,18 +1004,149 @@ def get_platforms_from_db(run_dir):
...
+        if not os.path.lexists(runN_path):
+            return path
+    if not runN_path.is_symlink() or not runN_path.is_dir():
+        raise WorkflowFilesError(
+            f"runN directory at {runN_path} is a broken or invalid symlink"
+        )
+    numbered_run = os.readlink(str(runN_path))
+    if not re.match(r'run\d+$', numbered_run):
+        # Note: the link should be relative
+        # TODO: what about cylc 8.0b1 workflows where it was absolute?
```
MetRonnie, line 1144: Should I tackle this or should we just leave it and not support the 8.0b1 absolute path targets of `runN` symlinks?

oliver-sanders: I don't think we need to worry about 8.0b1 compat.

MetRonnie:
One consequence of this change is that the "offline" commands no longer accept paths to a file relative to PWD. Any relative path is interpreted as relative to `~/cylc-run` (apart from `.`, which is kept as PWD). Does this need to be changed back to the original behaviour?

oliver-sanders (commented):
> One consequence of this change is that the "offline" commands no longer accept paths to a file relative to PWD.
> ...
> Does this need to be changed back to the original behaviour?

Hmm, it is kinda nasty accepting a path that is *either* relative to `$PWD` or `~/cylc-run`, however, this has worked for us before. This interface accepts `.` which is interpreted relative to `$PWD` so not accepting `./foo/` or `foo/` would be counter intuitive.

Options:

1) Don't allow paths relative to `$PWD` but make an exception for `.` (implemented.
2) Allow paths relative to `$PWD`, check `$PWD` before `~/cylc-run`.
3) Allow paths relative to `$PWD`, check `~/cylc-run` before `$PWD`.
4) Allow paths relative to `$PWD` but enforce them to begin with `.` (e.g. `./foo/`).

Cons:

1: Counter intuitive.

```console
$ cd cylc-src/foo
$ cylc validate .
ok
$ cd cylc-src
$ cylc validate ./foo
error
```

2&3: Has the potential for unexpected results:

```console
$ cd ~/cylc-src/
$ cylc install foo
$ cylc validate foo  # is this validating the src or the run copy?
```

4: Counter intuitive.

```console
$ cd ~/cylc-src
$ ls foo
flow.cylc
$ cylc validate foo
error
```

cylc/flow/platforms.py
```diff
@@ -548,5 +564,5 @@ def get_random_platform_for_install_target(
 
 def get_localhost_install_target() -> str:
     """Returns the install target of localhost platform"""
-    localhost = platform_from_name()
+    localhost = get_platform()
```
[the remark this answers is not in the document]
MetRonnie: This is just changing it back to how it was before; both identical as can be seen from `def get_platform(task_conf = None, task_id = UNKNOWN_TASK):`
https://github.com/cylc/cylc-flow/blob/bc72520ce03b09d492d3870fa8e2bfc76aa79da7/cylc/flow/platforms.py#L85-L87

The reason I had previously changed it to `platform_from_name()` was because mypy was complaining that `localhost` might be `None` due to the `Optional` return of `get_platform()`. However this now is fixed by the use of `@overload` to confirm that if `task_conf` is `None`, the return value will definitely be a dictionary
https://github.com/cylc/cylc-flow/blob/bc72520ce03b09d492d3870fa8e2bfc76aa79da7/cylc/flow/platforms.py#L43-L47

cylc/flow/scheduler_cli.py
```diff
@@ -15,13 +15,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """Common logic for "cylc play" CLI."""
 
+from ansimarkup import parse as cparse
```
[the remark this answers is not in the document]
MetRonnie: Didn't know that

cylc/flow/scripts/edit.py
```diff
@@ -102,8 +106,8 @@ def get_option_parser():
 
 
 @cli_function(get_option_parser)
-def main(parser, options, *args):
-    flow_file = parse_workflow_arg(options, args[0])[1]
+def main(parser: COP, options: 'Values', reg: str) -> None:
+    flow_file = str(parse_reg(reg, src=True)[1])
```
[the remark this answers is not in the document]
MetRonnie: Not 100% sure but the variable was previously a string, so I think I did this to be safe in case something further down doesn't like using pathlib Paths instead of strings

cylc/flow/platforms.py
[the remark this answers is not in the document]
MetRonnie: I've just found out about a nasty omission in mypy's checking of `@overload`ed function implementations: https://github.com/python/mypy/issues/9503

MetRonnie:
@hjoliver I've tagged you in case you want to respond to any of the discussion on this, you don't need to review the code necessarily

hjoliver:
> 
>  >  One consequence of this change is that the "offline" commands no longer accept paths to a file relative to PWD.
>    ...
> Hmm, it is kinda nasty accepting a path that is either relative to $PWD or ~/cylc-run, however, this has worked for us before. This interface accepts . which is interpreted relative to $PWD so not accepting ./foo/ or foo/ would be counter intuitive.
>
> Options:
>
>  1.  Don't allow paths relative to $PWD but make an exception for . (implemented.
>  2.  Allow paths relative to $PWD, check $PWD before ~/cylc-run.
>  3. Allow paths relative to $PWD, check ~/cylc-run before $PWD.
>  4. Allow paths relative to $PWD but enforce them to begin with . (e.g. ./foo/).

IMO it has to be 2. because non-absolute paths are expected be relative to `$PWD` as general principle. To break that in this case would definitely be counterintuitive and surprising for many users.

oliver-sanders:
> IMO it has to be 2

Agreed, however, we could do with handling the ambiguity.

It's easy enough to detect the cases where the user could be referring either to a flow relative to cylc-run or `$PWD` (excluding the case where `$PWD` == `~/cylc-run`)  and log a warning.

Slightly nasty as I don't think we have a way of forcing Cylc to pick the run copy in this eventuality.

With universal-id they could achieve this by prefixing the user name on:

```console
$ cylc validate ~user/foo
```

MetRonnie:
> > Options:
> > 
> > 1. Don't allow paths relative to $PWD but make an exception for . (implemented.
> > 2. Allow paths relative to $PWD, check $PWD before ~/cylc-run.
> > 3. Allow paths relative to $PWD, check ~/cylc-run before $PWD.
> > 4. Allow paths relative to $PWD but enforce them to begin with . (e.g. ./foo/).
> 
> IMO it has to be 2. because non-absolute paths are expected be relative to `$PWD` as a general principle. To break that in this case would definitely be counterintuitive and surprising for many users.

Just discovered this test which suggests the opposite, however

https://github.com/cylc/cylc-flow/blob/ff94ee25697d24bf4ea746cd04e18ff4473b2693/tests/functional/validate/48-reg-then-pwd.t#L18

tests/unit/test_templatevars.py
```diff
@@ -107,20 +108,6 @@ def test_load_template_vars_from_string_and_file_2(self):
...
-
-    n.b. Does not attempt to test ``load_template_vars``
-    """
-    monkeypatch.setattr(
-        'cylc.flow.templatevars.load_template_vars',
-        lambda templatevars, templatevars_file: {'foo': 'bar'}
-    )
-    opts = SimpleNamespace(templatevars='', templatevars_file='')
-    result = get_template_vars(opts, '', names=('eg/runN', 'eg/runN'))
-    assert result == {'foo': 'bar'}
```
MetRonnie, line 121: As this seemed to be testing the logic of handling the `names` tuple at heart, I removed it. Not sure if it needs a replacement. @wxtim ?

MetRonnie:
@wxtim Can I ask you to review the last 2 commits? https://github.com/cylc/cylc-flow/pull/4285/files/e3d3391aa1de7311a0c49f56593de39c978cc049..c1205dc6e91aa3cd6e3dc5385742ba43f0fe9d0a

tests/unit/test_templatevars.py
wxtim: Looks unnecessary to me, given your changes: Good work. It's easy to write tests, much harder to get rid. 👍🏼

wxtim:
> @wxtim Can I ask you to review the last 2 commits? https://github.com/cylc/cylc-flow/pull/4285/files/653f52bd3043152a31946bad2db5cb543a3a8a60..57be42068f71079c2493472b3e76a0eccdce3f46

Approved. Nice. A much better approach than mine. Still not sure that the extra options shouldn't be added from Cylc Rose somehow. But let's put that in a change of its own?

MetRonnie:
> Just discovered this test which suggests the opposite, however
>```
># Test validation order, installed workflows before current working directory.
>```

I have removed that test; the new behaviour (cwd takes precedence over installed workflows) is covered by unit tests. I will document the change in cylc-doc, possibly in the migration guide. Should the behaviour be documented in the `cylc validate` docstring/usage text? (Bearing in mind there are other "offline" commands that also follow the same behaviour)

MetRonnie:
If it helps, here is the final iteration of the flowchart I made in order to make the `parse_reg()` function

parse_reg flowchart v2

hjoliver:
> If it helps, here is the final iteration of the flowchart I made in order to make the parse_reg() function

Wow, well done. This was clearly a tougher job than we might have hoped!

cylc/flow/workflow_files.py
```diff
@@ -1048,18 +1029,181 @@ def get_platforms_from_db(run_dir):
...
+
+
+@overload
+def parse_reg(reg: str, src: Literal[True]) -> Tuple[str, Path]:
+    ...
+
+
+@overload
+def parse_reg(reg: str, src: bool) -> Union[str, Tuple[str, Path]]:
+    ...  # Need this 3rd overload https://github.com/python/mypy/issues/6113
```
datamel, line 1044: Bizarre!

datamel (approved):
I have read the code, checked out the branch, ran the tests and checked functionality with some manual tests, running commands successfully.

cylc/flow/config.py
```diff
@@ -178,15 +180,15 @@ def __init__(
...
-        self.start_point: Optional['PointBase'] = None
-        self.final_point: Optional['PointBase'] = None
         self.clock_offsets = {}
         self.expiration_offsets = {}
         self.ext_triggers = {}  # Old external triggers (client/server)
         self.xtrigger_mgr = xtrigger_mgr
         self.workflow_polling_tasks = {}  # type: ignore # TODO figure out type
 
+        self.initial_point: 'PointBase'
+        self.start_point: 'PointBase'
```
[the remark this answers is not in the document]
MetRonnie: I didn't explain; it's because they should always be set by https://github.com/cylc/cylc-flow/blob/135ab1aceac04cfa9dfa40b631dbd9c3de3bb459/cylc/flow/config.py#L599 and https://github.com/cylc/cylc-flow/blob/135ab1aceac04cfa9dfa40b631dbd9c3de3bb459/cylc/flow/config.py#L647

(`start_point` takes on value of `initial_point` if not specified)

This stops mypy errors complaining that `start_point` or `initial_point` may be `None` when trying to use `PointBase` methods later.

Only problem is, if they somehow don't end up getting set, we'd end up with `AttributeError`s, which won't get picked up by mypy. But in that case, we'd still get `TypeError`s if they were set to `None` in `__init__` like before, I guess. Not sure what the best practice is, but at least this silences mypy.

oliver-sanders (changes requested):
Code looks good, thanks for the docstrings.

Codacy has highlighted a few error cases which aren't covered by the tests, would be good to cover a few of them if possible.

Functional testing has mostly gone well, a few small issues:

1) **I think this is tricking `cylc clean` into thinking that the workflow is stopped.**

To replicate try:

```
cylc install one; cylc play one; tree -a ~/cylc-run; cylc clean one; tree -a ~/cylc-run
```

This should confirm that the contact file is present when the clean command is run.

On a network filesystem I get the expected error when you remove a dir where a running process has an open file handle:

```
FileRemovalError: [Errno 39] Directory not empty: 'workflow'. This is probably a temporary issue with the filesystem, not a problem with Cylc.
```

2) **Traceback for `cylc clean`.**

This might be unrelated, further attempts to clean the workflow in (1) resulted in traceback:

```
$ cylc cleal one
INFO - No workflow database - will only clean locally
INFO - Cleaning on local filesystem
...
OSError: [Errno 16] Device or resource busy: '.nfsc0000207034ffdad00000025'
```

I couldn't replicate this every time.

3) **Duplicate warnings for suite.rc deprecation**

```
$ cylc validate ./whatever
WARNING - The filename 'suite.rc' is deprecated in favour of 'flow.cylc'
WARNING - The filename 'suite.rc' is deprecated in favour of 'flow.cylc'
```

(where whatever also exists in the cylc-run directory)

4) **Ambiguity where it isn't needed**

(small issue, not a biggie, can punt to future work if it isn't a quick fix)

If I have the following setup:

```
~/
    cylc-src/
        whatever
    cylc-run/
        whatever
```

And try to validate the source workflow like so:

```
$ cd ~/cylc-src
$ cylc validate whatever
```

I get the expected warning due to ambiguity:

```
WARNING - Workflow files found in both ./whatever/suite.rc and /home/me/cylc-run/whatever/run1/flow.cylc. This command will use ./whatever/suite.rc.
```

However, if I prepend `./` to the flow name there shouldn't be any ambiguity any more, however, I still get the warning.

```
$ cd ~/cylc-src
$ cylc validate ./whatever
```

cylc/flow/workflow_files.py
```diff
@@ -315,6 +317,20 @@ class RemoteCleanQueueTuple(NamedTuple):
...
+)
+
+NO_FLOW_FILE_MSG = (
+    f"no {WorkflowFiles.FLOW_FILE} or {WorkflowFiles.SUITE_RC} "
+    "in {}"
+)
+
+REG_CLASH_MSG = (
+    "Workflow files found in both {0} and {1}. This command will use {0}."
+)
```
oliver-sanders, line 332: This message could be a little clearer.

MetRonnie:
> I think this is tricking `cylc clean` into thinking that the workflow is stopped.

This should be fixed by #4289

cylc/flow/workflow_files.py
```diff
@@ -1055,18 +1036,181 @@ def get_platforms_from_db(run_dir):
...
+            flow.cylc file (or any file really), or '.' for cwd).
+
+    Returns:
+        reg: The normalised workflow arg.
+        path: (Only if src is True) The absolute path to the workflow file
+            (flow.cylc or suite.rc).
+    """
+    if not src:
+        validate_workflow_name(reg)
+    reg: Path = Path(expand_path(reg))
```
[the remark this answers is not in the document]
MetRonnie: I think `validate_workflow_name()` is designed to validate what the user has typed verbatim, so we don't want to normalise/expand/etc it until after

cylc/flow/workflow_files.py
MetRonnie: What do you have in mind?

I suppose I can change `{1}` from being `/home/blah/blah/user/cylc-run/<run_dir>` to just `~/cylc-run/<run_dir>`, but I'm guessing that's not what you mean

MetRonnie:
> 4. Ambiguity where it isn't needed

Will open a new issue

cylc/flow/workflow_files.py
oliver-sanders: Not sure I had anything particular in mind.

I think "Workflow files" surprised me a little, might not be so obvious to a user that this implies a reference ambiguity between a src and run dir.

```
{input} could mean {src_dir} or ~/cylc-run/{run_dir.relative_to(cylc_run_dir)}, using {src_dir}
```

> I suppose I can change {1} from being /home/blah/blah/user/cylc-run/<run_dir> to just ~/cylc-run/<run_dir>

Would be good.

oliver-sanders:
I've found one more small issue, this time with the `reinstall` command which gives a really confusing traceback talking about `run1` rather than `run2`:

```console
$ cylc install one
...
$ cylc install one
...
$ cylc reinstall one
WorkflowFilesError: Nested run directories not allowed - cannot install workflow name "one" as "/home/me/cylc-run/one/run1" is already a valid run directory.
```

Although I'm not convinced that is necessarily to do with the reg parsing?

MetRonnie:
>> This seems to happen (on NFS) the second time shutil.rmtree ... trying to find a solution as a follow-up to #4133, but not managed to yet. See also #3659
>
> Although is this fix not related - MetRonnie@b02d74f?

Yes - I've been having a bit of a go at improving that `rmtree` error handler

> I've found one more small issue, this time with the `reinstall` command

I can reproduce on master. (`reinstall` isn't using the centralised reg parse)

oliver-sanders:
> I can reproduce on master

Ok, I'll open a separate issue.

> (reinstall isn't using the centralised reg parse)

Ah, it should be because it's working with workflow identifiers (run dir only).

MetRonnie:
In that case, could you add "make it use the centralised reg parse" to the issue please?

oliver-sanders:
> In that case, could you add "make it use the centralised reg parse" to that issue please?

But then we wouldn't be able to close the issues and would have mixed behaviour?

Seems straightforward to convert, this did the trick for me:

```diff
diff --git a/cylc/flow/scripts/reinstall.py b/cylc/flow/scripts/reinstall.py
index b25e2af23..1653b5861 100644
--- a/cylc/flow/scripts/reinstall.py
+++ b/cylc/flow/scripts/reinstall.py
@@ -45,9 +45,11 @@ from cylc.flow.exceptions import PluginError, WorkflowFilesError
 from cylc.flow.option_parsers import CylcOptionParser as COP
 from cylc.flow.pathutil import get_workflow_run_dir
 from cylc.flow.workflow_files import (
+    WorkflowFiles,
     get_workflow_source_dir,
+    parse_reg,
     reinstall_workflow,
-    WorkflowFiles)
+)
 from cylc.flow.terminal import cli_function
 
 if TYPE_CHECKING:
@@ -112,26 +114,20 @@ def get_option_parser():
 
 @cli_function(get_option_parser)
 def main(
-    parser: COP, opts: 'Values', named_run: Optional[str] = None
+        parser: COP, opts: 'Values', workflow: str
 ) -> None:
-    if not named_run:
-        source, _ = get_workflow_source_dir(Path.cwd())
-        if source is None:
-            raise WorkflowFilesError(
-                f'"{Path.cwd()}" is not a workflow run directory.')
-        base_run_dir = Path(get_workflow_run_dir(''))
-        named_run = str(Path.cwd().relative_to(base_run_dir.resolve()))
-    run_dir = Path(get_workflow_run_dir(named_run))
+    workflow = parse_reg(workflow)
+    run_dir = Path(get_workflow_run_dir(workflow))
     if not run_dir.exists():
         raise WorkflowFilesError(
-            f'"{named_run}" is not an installed workflow.')
+            f'"{workflow}" is not an installed workflow.')
     if run_dir.name in [WorkflowFiles.FLOW_FILE, WorkflowFiles.SUITE_RC]:
         run_dir = run_dir.parent
-        named_run = named_run.rsplit('/', 1)[0]
+        workflow = workflow.rsplit('/', 1)[0]
     source, source_path = get_workflow_source_dir(run_dir)
     if not source:
         raise WorkflowFilesError(
-            f'"{named_run}" was not installed with cylc install.')
+            f'"{workflow}" was not installed with cylc install.')
     if not Path(source).exists():
         raise WorkflowFilesError(
             f'Workflow source dir is not accessible: "{source}".\n'
@@ -153,7 +149,7 @@ def main(
             ) from None
 
     reinstall_workflow(
-        named_run=named_run,
+        named_run=workflow,
         rundir=run_dir,
         source=source,
         dry_run=False  # TODO: ready for dry run implementation
```

MetRonnie:
@datamel Can I ask you to review the latest commit touching `cylc reinstall`?

oliver-sanders (approved):
Have tested the reinstall changes, works great.

No more issues from testing, great improvement, good to go :tada:.

datamel:
> No more issues from testing, great improvement, good to go 🎉.

Excellent, I'm back reviewing this and will merge once tested.

datamel (approved):
I've read through the code, run the tests locally and had a manual test with the change. No problems found. Thanks @MetRonnie.

PR #4286, opened 2021-07-01
scan: show uninstalled workflows

oliver-sanders:
Add the `cylc scan --uninstalled` option.

> **Note:** Source isn't a workflow state because source workflows are orthogonal to run workflows. The lifecycle of the source is not the same as the lifecycle of the run. The UI will have to track source and run workflows separately.

First step towards displaying source workflows in the UI (as the UIS can use the `scan_multi` interface). Next we need to jam source workflows into the schema at the UIS and make them available to the UI.

**Requirements check-list**
  `conda-environment.yml`.
- [x] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/XXXX.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -56,6 +56,9 @@ Third beta release of Cylc 8.
 
 ### Enhancements
 
+[#4286](https://github.com/cylc/cylc-flow/pull/4286) -
+Add an option for displaying source workflows in `cylc scan`.
+
 [#4291](https://github.com/cylc/cylc-flow/pull/4291)
  - Remove obsolete `cylc edit` and `cylc search` commands.
 
--- a/cylc/flow/scripts/scan.py
+++ b/cylc/flow/scripts/scan.py
@@ -41,6 +41,10 @@ Examples:
   # filter workflows by name
   $ cylc scan --name '^f.*'  # show only flows starting with "f"
 
+  # list source workflows in a tree
+  # (looks in the dirs configured by "global.cylc[install]source dirs")
+  $ cylc scan --source -t tree
+
   # get results in JSON format
   $ cylc scan -t json
 """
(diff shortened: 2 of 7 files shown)


cylc/flow/scripts/scan.py
```diff
@@ -175,14 +181,24 @@ def get_option_parser():
...
         default='running,paused,stopping',
         action='store'
     )
 
+    parser.add_option(
+        '--uninstalled', '-u',
+        help=(
+            'List uninstalled workflows from configured source directories.'
+            ' (overrides the --states option)'
+        ),
```
MetRonnie, line 197: I wonder if "uninstalled" is the right word? Technically they're non-installed workflows if they haven't been installed at all yet. Perhaps `--source`, `--src` would be more appropriate?

tests/conftest.py
```diff
@@ -0,0 +1,71 @@
...
+        glbl_cfg.loadcfg(global_config_path)
+
+        def _inner(cached=False):
+            nonlocal glbl_cfg
+            return glbl_cfg
+
+        monkeypatch.setattr(pypath, _inner)
+
+    yield _mock
+    rmtree(tmp_path)
```
MetRonnie, line 71: I know this was here before, but is the `rmtree(tmp_path)` actually needed? Doesn't `tmp_path` get removed anyway afterwards?

tests/integration/test_scan.py
```diff
@@ -144,6 +145,33 @@ def nested_run_dir():
     rmtree(tmp_path)
 
 
+@pytest.fixture
+def source_dirs(mock_glbl_cfg):
+    src = Path(TemporaryDirectory().name)
+    src.mkdir()
```
MetRonnie, line 151: Why not use `tmp_path / 'cylc-src'` instead?

MetRonnie (commented):
:+1:

tests/integration/test_scan.py
oliver-sanders: That is an option for this fixture, but not for the others above due to scoping. It made sense to maintain consistency within the file.

tests/conftest.py
oliver-sanders: I don't think it is needed.

cylc/flow/scripts/scan.py
oliver-sanders: Yeah, `--source, -s` makes sense.

tests/integration/test_scan.py
MetRonnie: I see. The `tmp_path_factory` fixture would probably be cleaner but happy to leave as is

datamel (approved):
I have read the code, checked out the branch, manually tested the change and run the tests. Thanks @oliver-sanders.

CHANGES.md
```diff
@@ -56,6 +56,9 @@ Third beta release of Cylc 8.
 
 ### Enhancements
 
+[#4286](https://github.com/cylc/cylc-flow/pull/4286) -
+Add an option for displaying un-installed workflows in `cylc scan`.
```
MetRonnie, line 60: Nitpick
```suggestion
Add an option for displaying non-installed source workflows in `cylc scan`.
```

MetRonnie (approved):

oliver-sanders:
Merging with two approvals.

PR #4287, opened 2021-07-02
Implement named flows, and improve task logging.

hjoliver:
This has been discussed, but we don't seem to have an issue up.

Motivation: while investigating "pre-initial dependency ignore" for individual flows (viewed as essential by @dpmatthews) I realized several problems with the initial automatic flow-label management which make named flows more than just a nice-to-have. The idea behind the initial flow labeling implementation, which uses a set finite set of simple labels (`ascii_letters`) was: users don't have to do anything, just identify and use the ascii character name for the flow of interest, *and* merged flows can be named by simply concatenating those characters rather than dealing with sets of multiple separate names.  BUT it turns out: 
- you can't automatically prune flows that (seem to have) taken over the whole graph, because they might not have taken over, in which case pruning them can result in "conditional reflow" problems
- and even if you could prune them, you can't returned pruned labels for re-use because flow labels have to be unique for the life of the workflow, which means a finite pool of automatic names can run out.  
 
**Requirements check-list**
- [x ] Does not contain off-topic changes (use other PRs for other changes).
- [x] Already covered by (modified) existing tests.


wxtim:
@hjoliver  What is the status of this PR? Would you like me to:
- Resolve the conflict with changes.md?
- Review?

hjoliver:
> @hjoliver What is the status of this PR? Would you like me to:
> 
>     * Resolve the conflict with changes.md?
> 
>     * Review?

Thanks for asking @wxtim .  It's blocked for a short while pending an off-side meeting with @dpmatthews and @oliver-sanders on several spawn-on-demand topics (scheduled for Wednesday). I should have converted to draft ..will do that now.

oliver-sanders:
This makes a lot of sense, nice.

> All task state (and state attribute) changes are now logged automatically (via the state reset method)

:+1:

To stop a flow ... This removes the warm1 flow label from all task proxies

:tada:

hjoliver:
Outstanding questions:
- is a user-supplied name sufficient as flow metadata? (I think it probably is)
  - or do we need additional metadata to be associated with the flow?
- is it problem that users need to supply a name that must be unique in the lifetime of the workflow (unless there is no chance of overlap)?
  - maybe not, they could just increment an integer suffix each time (`--flow=warm9`) or use the current datetime (`--flow=$(date +%Y/%m/%d:%H:%m)`)
  - we could automatically maintain an integer flow counter, to use with a user-supplied name prefix
  - or automatically add the datetime, but resulting flow labels become pretty big and harder to read and type out accurately

hjoliver:
Discussed in meeting:
- general approach agreed
- automatically incrementing integer flow labels will
  - much easier to read and type than massive datetime strings
  - with user-supplied name as an option

oliver-sanders:
I think we may want to accumulate a bit of metadata around [re]flows so happy with a short auto-generated integer rather than a timestamp (nicer for logging and the CLI).

Some metadata we could add in due course:

* timestamp (time of the first occasion the flow label was created)
* user (if actioned from the WUI, the user who actioned it)
* start tasks? (the tasks [or cycles?] at the head of the flow)
* description (unlikely to be provided in most cases, however, might be of interest to operations centres)

Looking ahead we will likely want to merge the `--flow` CLI opt into the universal identifier for a few reasons:

* This allows the WUI to filter by flow
  * e.g. user might want want a graph with the two flows shown separately
  * related issue https://github.com/cylc/cylc-ui/issues/468
* Possible future [re]flow development.
* Multi workflow functionality (`cylc stop a// b// 'c*//'`)
* General completeness.

``` bash
# current proposal / reference implementation
~user/workflow//cycle/task/job
~user/workflow:status//cycle:status/task:status/job:status  # with all selectors

# possible extension for [re]flow
~user/workflow^flow//

# example
cylc stop myflow^orig
```

PR #4289, opened 2021-07-02
Prevent cylc clean from removing workflow base dirs containing run dirs/named runs

### LINKED ISSUE #4163: cylc clean and numbered/named run dirs
Say you have a workflow `foo` with numbered run dirs, i.e. `~/cylc-run/foo/runN/`.

Currently if you do `cylc clean foo` it will just remove `~/cylc-run/foo/`, without removing [symlink dir](https://cylc.github.io/cylc-doc/8.0b0/html/reference/config/global.html#global.cylc[symlink%20dirs]) targets, or cleaning on remote install targets. (I think the rationale behind this is to be able to clean any partially cleaned dirs.)

The same problem exists for hiearchical workflow names, e.g. workflow `alpha/bravo` and you do `cylc clean alpha`. (Named/numbered run dirs is basically a special case of hiearchical workflow names)

How to address this?

1. Raise an error if `cylc clean REG` is run and `~/cylc-run/REG` is not a *run* dir (i.e. it does not contain a `.service` dir?). Suggested by @dpmatthews. Perhaps include a `--force` option to simply remove `~/cylc-run/REG`?
2. Clean each run dir in turn (`~/cylc-run/REG/run1`, `~/cylc-run/REG/run2` etc) and finally remove `~/cylc-run/REG` (which would be in line with #3895<sup>1</sup>). However this still runs the risk of a user accidentally deleting more than they meant to.
3. Do (1) by default, but include an `--all` option to do (2).

---

<sup>1</sup> Except whereas `cylc play foo` would be the same as `cylc play foo/runN`, `cylc clean foo` would be the same as `cylc clean foo/run1 && cylc clean foo/run2 ...`

MetRonnie:
These changes close #4163 (assuming we're happy not having an `--all` option and will instead wait until [universal ID](https://github.com/cylc/cylc-admin/pull/115) is implemented, allowing `cylc clean foo/*`)

**Explanation**

If you have
```
~/cylc-run
`-- foo
    `-- run1
        |-- flow.cylc
            etc
```
Up til now, doing `cylc clean foo` would just delete `~/cylc-run/foo`, and if you had any symlink dirs or remote installations, they wouldn't get cleaned.

With this PR, it now raises an error if you try to clean a dir containing multiple (>1) runs - you have to explicitly type out the run dir. However:
- If the dir contains 1 run dir, it will clean it (cleans `foo/run1` followed by deleting `foo`)
- There is now the option `--force` to allow possibly incomplete cleaning of the specified dir (e.g. it will just delete the `foo` dir)

**Requirements check-list**
  `conda-environment.yml`.
- [x] Documentation will be handled by https://github.com/cylc/cylc-doc/issues/256


MetRonnie:
As I'm off work in the coming week, happy for reviewers to push changes to the branch if it helps get this merged for 8.0b2

cylc/flow/workflow_files.py
```diff
@@ -600,7 +601,13 @@ def is_installed(path: Union[Path, str]) -> bool:
     return cylc_install_folder.is_dir() and source.is_symlink()
 
 
-def _clean_check(reg: str, run_dir: Path) -> None:
+async def get_contained_workflows(path: Path) -> List[str]:
+    """Return the sorted names of any workflows in the path provided"""
+    from cylc.flow.network.scan import scan
+    return sorted([i['name'] async for i in scan(scan_dir=path)])
```
MetRonnie, line 607: I suspect we'd get a circular import error if this import was at the top of the file.

We kind of need to separate out the different parts of `workflow_files` into different files

tests/integration/test_scan.py
```diff
@@ -80,19 +80,20 @@ def sample_run_dir():
     rmtree(tmp_path)
 
 
-@pytest.fixture(scope='session')
-def badly_messed_up_run_dir():
-    tmp_path = Path(TemporaryDirectory().name)
-    tmp_path.mkdir()
+@pytest.fixture
```
MetRonnie, line 83: This fixture only used once so I've made it function-scoped

cylc/flow/workflow_files.py
MetRonnie: By the way, should we increase the scan depth in this case (default is 4 levels deep)? Seeing as this is a safety feature?

wxtim (approved):
- [x] Read, all makes sense.
- [x] Tried it out → Working.

cylc/flow/network/scan.py
```diff
@@ -132,20 +133,26 @@ async def scan(run_dir=None, scan_dir=None, max_depth=MAX_SCAN_DEPTH):
...
     async def _scandir(path, depth):
         contents = await scandir(path)
         return path, depth, contents
 
     # perform the first directory listing
-    for subdir in await scandir(scan_dir):
+    scan_dir_listing = await scandir(scan_dir)
+    if scan_dir != cylc_run_dir and dir_is_flow(scan_dir_listing):
+        # If the scan_dir itself is a workflow run dir, yield nothing
+        return
```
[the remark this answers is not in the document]
MetRonnie: There's this test which specifies `scan(cylc_run_dir)` should not err (even if it somehow ends up containing a `.service` dir)
https://github.com/cylc/cylc-flow/blob/acf5fdb5e9832c5b4352b87eec66355661aacfe2/tests/integration/test_scan.py#L218-L230

cylc/flow/workflow_files.py
```diff
@@ -620,8 +627,18 @@ def _clean_check(reg: str, run_dir: Path) -> None:
     try:
         detect_old_contact_file(reg)
     except ServiceFileError as exc:
-        raise ServiceFileError(
-            f"Cannot remove running workflow.\n\n{exc}")
+        raise ServiceFileError(f"Cannot remove running workflow.\n\n{exc}")
+    # Check dir does not contain other workflows:
+    contained_workflows = asyncio.run(get_contained_workflows(run_dir))
```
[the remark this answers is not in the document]
MetRonnie: I was a bit worried about that

cylc/flow/network/scan.py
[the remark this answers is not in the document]
oliver-sanders: Ah, fair enough, I forgot about that issue (caused issues with `cylc scan`, `cylc review`, etc).

oliver-sanders:
Should `cylc clean` ignore the `runN` directory or will that come with https://github.com/cylc/cylc-flow/pull/4285 (or future work)?

```
$ cylc install two
$ cylc clean two
...
contains the following workflow(s):
    - two/run1
    - two/runN
```

MetRonnie:
(As discussed on Teams) `cylc clean` will not use the inferred run number functionality. Marking as draft while investigating the `runN` symlink being detected as a separate workflow

cylc/flow/pathutil.py
```diff
@@ -295,6 +295,40 @@ def remove_dir_or_file(path: Union[Path, str]) -> None:
...
+        tail: The tail of the path to work our way up, e.g. a/b/c
+
+    Example:
+        remove_empty_parents('/foo/bar/a/b/c', 'a/b/c') would remove
+        /foo/bar/a/b (assuming it's empty), then /foo/bar/a (assuming it's
+        empty).
+    """
+    path = Path(path)
+    if not path.is_absolute():
+        raise ValueError('path must be absolute')
```
oliver-sanders, line 314: Would be good to cover some of these exception lines.

oliver-sanders (approved):

cylc/flow/workflow_files.py
```diff
@@ -649,31 +665,50 @@ def init_clean(
...
+        init_clean(contained_workflows[0], opts)
+        _suppress_no_db_msg = True
+    elif len(contained_workflows) > 1:
+        bullet = "\n    - "
+        msg = (
+            f"{local_run_dir} contains the following workflows:"
+            f"{bullet}{bullet.join(contained_workflows)}"
+        )
+        if not opts.force:
+            raise WorkflowFilesError(f"Cannot clean - {msg}")
```
wxtim, line 688: Might be worth adding a note to the end of the message to the effect that the user can use `--force` if that is what they genuinely want to do...

cylc/flow/workflow_files.py
```diff
@@ -649,31 +665,50 @@ def init_clean(
...
         return
 
+    # Check dir does not contain other workflows:
+    contained_workflows = asyncio.get_event_loop().run_until_complete(
+        get_contained_workflows(local_run_dir, MAX_SCAN_DEPTH + 1)
+    )  # Note: increased scan depth for safety
+    _suppress_no_db_msg = False
+    if len(contained_workflows) == 1:
+        init_clean(contained_workflows[0], opts)
+        _suppress_no_db_msg = True
```
wxtim, line 680: Codecov warning seems to me to be untrue. Do we know why it's flagged these lines as untested?

wxtim (commented):
- [x] Scan read changes since last time.
- [x] Played with result branch, run some examples. 
- [x] Had an in-depth play with `test_init_clean__multiple_runs`. Assured myself that Codecov is wrong in stating some logic not tested.
- [x] Had an in-depth play with `test_remove_empty_parents`

cylc/flow/pathutil.py
MetRonnie: Codecov is lying 😢 https://github.com/cylc/cylc-flow/pull/4289/files#diff-a9a7ea82148a0b1c8f631db722616f6cca1f468f3bf77bf83cd783a93220c7edR439-R442

oliver-sanders: Yep, confirmed locally, went into the codecov report and found that the problem doesn't exist there. Will try fiddling the codecov config as I think these comments may have been left *before* all jobs had finished.

cylc/flow/workflow_files.py
oliver-sanders: No, having a play.

cylc/flow/pathutil.py
MetRonnie: @oliver-sanders See #4348

MetRonnie:
Merging as 2 approvals

Issue #4175, opened 2021-04-15
tests/integration: data store mgr tests unstable


oliver-sanders:
The data store mgr tests or a subset thereof sometimes fail and sometimes pass.

Some example traceback:

```python
    def test_update_data_structure(harness):
        """Test update_data_structure. This method will generate and
        apply adeltas/updates given."""
        schd, data = harness
        w_id = schd.data_store_mgr.workflow_id
        schd.data_store_mgr.data[w_id] = data
        assert TASK_STATUS_FAILED not in set(collect_states(data, TASK_PROXIES))
        assert TASK_STATUS_FAILED not in set(collect_states(data, FAMILY_PROXIES))
        assert TASK_STATUS_FAILED not in data[WORKFLOW].state_totals
>       assert len({t.is_held for t in data[TASK_PROXIES].values()}) == 2
E       assert 0 == 2
E         +0
E         -2
```

The test file:

```
tests/integration/test_data_store_mgr.py
```

dwsutherland:
Have run it twenty times locally and no failure... I will see if there's a more robust way to test the update.

dwsutherland:
I'd have to understand how the test harness works at a deeper level to know why it's inconsistent..

i.e. why does this:
```
@pytest.mark.asyncio
@pytest.fixture(scope='module')
async def harness(mod_flow, mod_scheduler, mod_run):
    flow_def = {
        'scheduler': {
            'allow implicit tasks': True
        },
        'scheduling': {
            'graph': {
                'R1': 'foo => bar'
            }
        }
    }
    reg: str = mod_flow(flow_def)
    schd: 'Scheduler' = mod_scheduler(reg)
    async with mod_run(schd):
        schd.pool.hold_tasks('*')
        schd.resume_workflow()
        # Think this is needed to save the data state at first start (?)
        # Fails without it.. and a test needs to overwrite schd data with this.
        data = schd.data_store_mgr.data[schd.data_store_mgr.workflow_id]
        yield schd, data
```
Sometimes yield the scheduler with no task proxies?  Did the tasks not exist when the hold command was issued?

Hard to tell unless I can reproduce the issue somehow..

dwsutherland:
I think the same harness yield gets shared amongst the tests..
Whenever I changes the order of the tests, it appears to effect what data the next tests are working with.

dwsutherland:
On a side note this also works:
```
(flow) sutherlander@cortex-vbox:cylc-flow$ git diff
[...]

PR #4291, opened 2021-07-05
Remove old "edit" and "search" commands.

hjoliver:
This is a small change with no associated Issue - discussed on Element chat.

**Requirements check-list**
- [x] Does not need tests (Removed code and tests).
- [x] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/261


hjoliver:
Should we remove `cylc diff` too?

```
$ cylc diff --help
Usage: cylc diff [OPTIONS] WORKFLOW1 WORKFLOW2

Compare two workflow configurations and display any differences.

Differencing is done after parsing the flow.cylc files so it takes
account of default values that are not explicitly defined, it disregards
the order of configuration items, and it sees any include-file content
after inlining has occurred.

Files in the workflow bin directory and other sub-directories of the
run directory are not currently differenced.
```
I suppose this could be useful as described. The diffs are presented after parsing so order of sections doesn't matter.  Or shall we just tell users to go to standard `diff` (perhaps with `flow.cylc.processed` to see diffs after inlining and Jinja2 processing)?

hjoliver:
And similarly, `cylc search`?

```$ cylc search --help

Usage: cylc search [OPTIONS] WORKFLOW PATTERN [PATTERN2...]

Search for patterns in workflow configurations.

Search for pattern matches in workflow definitions and any files in the
workflow bin directory. Matches are reported by line number and workflow
section. An unquoted list of PATTERNs will be converted to an OR'd
pattern. Note that the order of command line arguments conforms to
normal cylc command usage (workflow name first) not that of the grep
command.

Note that this command performs a text search on the workflow definition,
it does not search the data structure that results from parsing the
workflow definition - so it will not report implicit default settings.

For case insensitive matching use '(?i)PATTERN'.
```

hjoliver:
I'm not sure the benefits are sufficient to bother (with `cylc diff` and `cylc search`) c.f. standard `diff` and `grep`.

wxtim:
> I'm not sure the benefits are sufficient to bother (with `cylc diff` and `cylc search`) c.f. standard `diff` and `grep`.

Cylc diff - Looks like it does some useful extra stuff. I can see uses for it in terms of convincing customers that "yes, they have changed their workflow"

Cylc search - Don't see the point

oliver-sanders:
Yeah, same opinion here:

* `edit` use a text editor.
* `search` use `grep` or a text editor.

However, `diff` and `graph --diff` are somewhat different as they compare the parsed config. They wouldn't be worth the effort of implementing now, however, since they still work and require little maintenance there is little impetus to remove them.

hjoliver:
Roger that. Updated to remove `cylc search` as well as `cylc edit`. Retaining `cylc diff`.

hjoliver:
One review will do.

kinow (approved):

kinow:
CI failure appears to be unrelated

```
=================================== FAILURES ===================================
_____________________________ test_delta_task_held _____________________________
[gw4] linux -- Python 3.8.10 /opt/hostedtoolcache/Python/3.8.10/x64/bin/python

harness = (Scheduler(profiler=<cylc.flow.profiler.Profiler object at 0x7f3e0dd0b4c0>, pool=<cylc.flow.task_pool.TaskPool object ...e: 2
}
is_held_total: 0
is_queued: false
is_queued_total: 0
is_runahead: true
is_runahead_total: 1
}, 'jobs': {}, ...})

    def test_delta_task_held(harness):
        """Test update_data_structure. This method will generate and
        apply adeltas/updates given."""
        schd, data = harness
>       assert True in {t.is_held for t in data[TASK_PROXIES].values()}
E       assert True in {False}

tests/integration/test_data_store_mgr.py:191: AssertionError
__________________________ test_update_data_structure __________________________
[gw4] linux -- Python 3.8.10 /opt/hostedtoolcache/Python/3.8.10/x64/bin/python

harness = (Scheduler(profiler=<cylc.flow.profiler.Profiler object at 0x7f3e0dd0b4c0>, pool=<cylc.flow.task_pool.TaskPool object ...e: 2
}
is_held_total: 0
is_queued: false
is_queued_total: 0
is_runahead: true
is_runahead_total: 1
}, 'jobs': {}, ...})

    def test_update_data_structure(harness):
        """Test update_data_structure. This method will generate and
        apply adeltas/updates given."""
        schd, data = harness
        w_id = schd.data_store_mgr.workflow_id
        schd.data_store_mgr.data[w_id] = data
        assert TASK_STATUS_FAILED not in set(collect_states(data, TASK_PROXIES))
        assert TASK_STATUS_FAILED not in set(collect_states(data, FAMILY_PROXIES))
        assert TASK_STATUS_FAILED not in data[WORKFLOW].state_totals
>       assert len({t.is_held for t in data[TASK_PROXIES].values()}) == 2
E       assert 1 == 2
E         +1
E         -2
```

Opened the IDE and searched for other parts using edit and found nothing, except `cat-log` that appears to support an edit mode, but I think that's not related to this edit command that was removed.

LGTM :tada:

oliver-sanders:
I think those failures are caused by this issue - https://github.com/cylc/cylc-flow/issues/4175

These tests are about 50/50 on my machine.

PR #4301, opened 2021-07-12
coverage: change ignore rules

oliver-sanders:
This aught to stop mypy stuff being flagged as uncovered code.

**Requirements check-list**
  `conda-environment.yml`.


cylc/flow/scripts/broadcast.py
```diff
@@ -409,7 +409,3 @@ def main(_, options: 'Values', workflow: str) -> None:
             print(get_broadcast_change_report(
                 modified_settings, is_cancel=report_cancel))
     sys.exit(report_bad_options(bad_options, is_set=report_set))
-
-
-if __name__ == "__main__":
-    main()
```
MetRonnie, line 415: Why is this no longer needed? Is it because all commands are mapped to in `setup.cfg`?

oliver-sanders: Yes, this allows the commands to be run by executing the files directly which was never intended.

PR #4302, opened 2021-07-12
have scheduler warn if stopcp > finalcp

### LINKED ISSUE #4226: Warn the user if the stop CP is never going to be used

If you have a start CP `1`, and final CP `2`, and define stop CP as `100`, your stop CP is just going to be ignored and never used. It would be helpful to users to log a warning.

```
rm -rf .service && cylc play --no-detach five --initial-cycle-point=1 --final-cycle-point=2 --stop-cycle-point=4
```

**Additional context**

From General element chat room.

**Pull requests welcome!**

wxtim:
These changes close #4226
(CP = Cycle Point)

> If you have a start CP 1, and final CP 2, and define stop CP as 100, your stop CP is just going to be ignored and never used. It would be helpful to users to log a warning.

  `conda-environment.yml`.
- [x] No change log entry required (Extra logging).

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1910,4 +1910,4 @@ class Scheduler:
             LOG.warning(
                 f'Stop Cycle point \'{self.options.stopcp}\' will have no '
                 'effect as it is after the final Cycle '
-                f'point \'{self.options.fcp}\'.')
+                f'point \'{self.config.final_point}\'.')
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1906,10 +1906,7 @@ class Scheduler:
         """
         if self.config.final_point is None:
             return
-        if (
-            get_point(self.options.stopcp, self.config.cycling_type)
-            > get_point(self.config.final_point, self.config.cycling_type)
-        ):
+        if get_point(self.options.stopcp) > self.config.final_point:
             LOG.warning(
                 f'Stop Cycle point \'{self.options.stopcp}\' will have no '
                 'effect as it is after the final Cycle '
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1888,6 +1888,7 @@ class Scheduler:
         if stoppoint is not None:
             self.options.stopcp = str(stoppoint)
             self.pool.set_stop_point(get_point(self.options.stopcp))
+            self.validate_finalcp()
 
     async def handle_exception(self, exc: Exception) -> NoReturn:
         """Gracefully shut down the scheduler given a caught exception.
@@ -1899,3 +1900,17 @@ class Scheduler:
         """
         await self.shutdown(exc)
         raise exc from None
+
+    def validate_finalcp(self):
+        """Warn if Stop Cycle point is on or after the final cycle point
+        """
+        if self.config.final_point is None:
+            return
+        if (
+            get_point(self.options.stopcp, self.config.cycling_type)
+            > get_point(self.config.final_point, self.config.cycling_type)
+        ):
+            LOG.warning(
+                f'Stop Cycle point \'{self.options.stopcp}\' will have no '
+                'effect as it is after the final Cycle '
+                f'point \'{self.options.fcp}\'.')
--- a/tests/functional/cylc-play/06-warnif-scp-after-fcp.t
+++ b/tests/functional/cylc-play/06-warnif-scp-after-fcp.t
@@ -1,0 +1,85 @@
+#!/usr/bin/env bash
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#------------------------------------------------------------------------
+
+# test that cylc play warns if FCP before Stop Point.
+
+. "$(dirname "$0")/test_header"
+
+set_test_number 14
+
+# integer cycling
+
+init_workflow "${TEST_NAME_BASE}" <<'__FLOW_CONFIG__'
+[scheduler]
+    allow implicit tasks = True
+[scheduling]
+    initial cycle point = 1
+    final cycle point = 2
+    cycling mode = integer
+    [[dependencies]]
+        P1 = foo
+__FLOW_CONFIG__
+
+TEST_NAME="${TEST_NAME_BASE}-validate"
+run_ok "${TEST_NAME}" cylc validate "${WORKFLOW_NAME}"
+
+for SCP in 1 2 3; do
+    TEST_NAME="${TEST_NAME_BASE}-play-integer-scp=${SCP}"
+        workflow_run_ok "${TEST_NAME}" cylc play "${WORKFLOW_NAME}" \
+        --no-detach --stopcp="${SCP}"
+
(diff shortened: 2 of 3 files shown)


cylc/flow/scheduler.py
```diff
@@ -1885,6 +1888,7 @@ def process_cylc_stop_point(self):
         if stoppoint is not None:
             self.options.stopcp = str(stoppoint)
             self.pool.set_stop_point(get_point(self.options.stopcp))
+            self.warnif_stopcp_gte_finalcp()
```
kinow, line 1891: Maybe call the function `validate_finalcp`, or `validate_cyclepoints`… in case we decide to have more checks in there?

cylc/flow/scheduler.py
```diff
@@ -1896,3 +1900,25 @@ async def handle_exception(self, exc: Exception) -> NoReturn:
...
+                and (
+                    ISO8601Point(self.options.stopcp) >=
+                    ISO8601Point(self.options.fcp)
+                )
+            )
+        ):
+            LOG.warning(
+                f'Stop Cycle point \'{self.options.stopcp}\' will not do '
+                'anything: It is the same or later than the final Cycle '
+                f'point \'{self.options.fcp}\'.')
```
kinow, line 1924: :+1:

tests/unit/test_scheduler.py
```diff
@@ -152,3 +152,32 @@ def test_check_startup_opts(
...
+    """Scheduler warns is stopcp ≥ finalcp"""
+    mocked_scheduler = Mock()
+    mocked_scheduler.options = SimpleNamespace(stopcp=scp, fcp=fcp)
+    mocked_scheduler.config = SimpleNamespace(cycling_type='integer')
+    
+    Scheduler.warnif_stopcp_gte_finalcp(mocked_scheduler)
+    if err == True:
+        assert caplog.records[0].message == (
+            f'Stop Cycle point \'{scp}\' will not do anything: It is the same '
+            f'or later than the final Cycle point \'{fcp}\'.'
```
kinow, line 180: Maybe here we don't need to check the exact string is logged? Either just assert the log function has been called and the message is not empty, or that it contains the stop cycle point and the final cycle point?

kinow (commented):
I think some functional tests had this condition of stop after final cp?

```
Failed 5/9 subtests 

    stdout and stderr stored in: /tmp/runner/cylctb-20210712T164034Z-sq7V/f/hold-release/01-beyond-stop
    workflow logs can be found under: /home/runner/cylc-run/cylctb-20210712T164034Z-sq7V/f/hold-release/01-beyond-stop/log/workflow/
[22:10:56] tests/f/hold-release/01-beyond-stop.t ........ 
```

From what I could see reading the [flow.cylc](https://github.com/cylc/cylc-flow/blob/4a180b273f773245cd2ff1fdd5b8322369a72252/tests/functional/hold-release/01-beyond-stop/flow.cylc#L17-L18) for the failed test above, it defines an Initial CP and a Stop CP with the same value.

I **think** internally Cylc must compute the Final CP too, so the three CP's all have same value. In this case, then I guess it's a valid condition of Final CP == Stop CP?? If so, perhaps we must validate only that the Stop CP > Final CP (i.e. `gt` only)?

cylc/flow/scheduler.py
wxtim: Good idea.

tests/unit/test_scheduler.py
wxtim: Agree. Lazy writing on my part ⇒ fragile test.

wxtim:
Turns out that many of the test failures were caused by final cycle point being optional.

kinow (approved):

cylc/flow/scheduler.py
MetRonnie: How about:
> The stop cycle point will have no effect as it is after the final cycle point.

cylc/flow/scheduler.py
```diff
@@ -1896,3 +1900,27 @@ async def handle_exception(self, exc: Exception) -> NoReturn:
         """
         await self.shutdown(exc)
         raise exc from None
+
+    def validate_finalcp(self):
+        """Warn if Stop Cycle point is on or after the final cycle point
+        """
+        if not self.options.fcp:
```
MetRonnie, line 1907: `self.config.final_point` is used elsewhere in `process_cylc_stop_point()`
```suggestion
        if self.config.final_point is None:
```

cylc/flow/scheduler.py
```diff
@@ -1896,3 +1900,27 @@ async def handle_exception(self, exc: Exception) -> NoReturn:
...
+            )
+            or
+            (
+                self.config.cycling_type == ISO8601_CYCLING_TYPE
+                and (
+                    ISO8601Point(self.options.stopcp) >
+                    ISO8601Point(self.options.fcp)
+                )
+            )
+        ):
```
MetRonnie, line 1922: No need to handle the cycling modes differently, `get_point()` takes care of that.
```suggestion
        if get_point(self.options.stopcp) > get_point(self.config.final_point):
```

wxtim: This is lovely. Thank you @MetRonnie

MetRonnie:
Something seems to have gone wrong with merging/rebasing. I suggest rebasing onto master again, just the commits that should be in this branch. Happy to help out if it's messy

wxtim:
> Something seems to have gone wrong with merging/rebasing. I suggest rebasing onto master again, just the commits that should be in this branch. Happy to help out if it's messy

That was horrid. Thanks for the offer, but I think that I have fixed it now.

wxtim:
@kinow  @MetRonnie  Probably needs a total re-review. Changed the unit test for functional testing because the level of mocking involved was becoming silly.

tests/functional/cylc-play/06-warnif-scp-after-fcp.t
```diff
@@ -0,0 +1,85 @@
...
+    if [[ "${SCP}" -lt 3 ]]; then
+        V="-v"
+    fi
+    grep_ok "Stop Cycle point '.*'.*after.*final Cycle point '.*'" \
+        "${RUN_DIR}/${WORKFLOW_NAME}/log/workflow/log" "$V"
+done
+
+purge
+
+# Gregorian Cycling
```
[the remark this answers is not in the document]
wxtim: IMO I think that it's worth testing both integer and ISOdatetime.

kinow (commented):
Code looks good @wxtim but I think functional tests are failing?

```
INSTALLED cylctb-20210811T132330Z-3GBZ/f/cylc-play/06-warnif-scp-after-fcp from /tmp/tmp.v80U1UzwkG/cylctb-20210811T132330Z-3GBZ/f/cylc-play/06-warnif-scp-after-fcp
ok 1 - 06-warnif-scp-after-fcp-validate
ok 2 - 06-warnif-scp-after-fcp-play-integer-scp=1
ok 3 - grep-ok: log-grep-ok
ok 4 - 06-warnif-scp-after-fcp-play-integer-scp=2
ok 5 - grep-ok: log-grep-ok
ok 6 - 06-warnif-scp-after-fcp-play-integer-scp=3
not ok 7 - grep-ok: log-grep-ok
ok 8 - 06-warnif-scp-after-fcp-validate
ok 9 - 06-warnif-scp-after-fcp-play-gregorian-scp=1348
ok 10 - grep-ok: log-grep-ok
ok 11 - 06-warnif-scp-after-fcp-play-gregorian-scp=1349
ok 12 - grep-ok: log-grep-ok
ok 13 - 06-warnif-scp-after-fcp-play-gregorian-scp=1350
not ok 14 - grep-ok: log-grep-ok
Failed 2/14 subtests 
[18:53:50]

Test Summary Report
-------------------
tests/f/cylc-play/06-warnif-scp-after-fcp.t (Wstat: 0 Tests: 14 Failed: 2)
  Failed tests:  7, 14
Files=1, Tests=14, 20 wallclock secs ( 0.02 usr  0.00 sys +  9.80 cusr  1.10 csys = 10.92 CPU)
Result: FAIL
```

wxtim:
> Code looks good @wxtim but I think functional tests are failing?

@kinow  - Passes locally, attempting to work out why not on GHA.

MetRonnie:
It's probably the case sensitivity of grep

wxtim:
> It's probably the case sensitivity of grep

@wxtim  kicks himself hard.

MetRonnie (approved):

Issue #4062, opened 2021-02-04
Re-evaluate the --initial/final/start/stop-cycle-point options for restarts


MetRonnie:
The `ignore` value for the cycle point cli options (`--icp=ignore`, `--startcp=ignore` etc) for `cylc play` are potentially confusing, and the behaviour is a little inconsistent.

### Final cycle point

When the scheduler shuts down, the final cycle point is saved to the database. On restart, this DB value is loaded.

`--fcp=ignore` causes Cylc to ignore the value stored in the DB, and instead use the value in `flow.cylc[scheduling]final cycle point`. The two might be different:
- If the cli option `--fcp=<point>` was used in a previous run to override flow.cylc (then the cli-supplied value would have be saved in the DB).

#### Issues

- [x] If the workflow has reached the final cycle point in a previous run, it has basically packed up and gone home, so `--fcp` won't do anything. Possibly this just needs to be documented clearer. (Doc PR: https://github.com/cylc/cylc-doc/pull/362)

These apply to all 4 types of cycle point:
- [x] The name `ignore` is potentially misleading, as one might get the idea it will basically set the final cycle point to `None`. Maybe `--*-cycle-point=reload/refresh/reset` is clearer? However, the cycle point is reloaded from `flow.cylc[scheduling]` anyway if not previously set in the cli*. (Resolved in #4313)

### Initial cycle point

Initial cycle point is handled in much the same way as final cycle point. However, this time there is a second way the value stored in the DB might be different from `flow.cylc[scheduling]`:
- If `flow.cylc[scheduling]initial cycle point = now` or `next()` or `previous()`, in which case the timepoint is evaluated at the time of first start and stored in the DB

#### Issues

- [x] In a restart, you can't change the initial cycle point using the cli opt. Any value for `--icp=<point>` apart from `ignore` will raise an error. However, you could in fact change the initial cycle point in `flow.cylc[scheduling]`. So why not using the cli opt? (Resolved: `--icp` option not valid for restart since #4313)

### Start cycle point

Similar to above, but there is no way of specifying the start cycle point in flow.cylc.

#### Issues

- [x] It looks like `--startcp=ignore` doesn't actually do anything? The start point for a restart is whatever the last incomplete cycle point at shutdown was. (Resolved: `--startcp=ignore/reload` not valid since #4313)

#### Issues

[...]

PR #4313, opened 2021-07-21
Change 'ignore' to 'reload' for cycle point cli opts

MetRonnie:
These changes partially address #4062 

> The name `ignore` [in the `--*-cycle-point=ignore` cli options] is potentially misleading, as one might get the idea it will basically set the final cycle point to `None`. 

- Change `ignore` to `reload`, as we are actually reloading from the `flow.cylc` file

Also

> It looks like `--startcp=ignore` doesn't actually do anything? The start point for a restart is whatever the last incomplete cycle point at shutdown was.

- Disallow `--icp=reload` and `--startcp=reload` because they weren't doing anything

**Requirements check-list**
  `conda-environment.yml`.


tests/functional/restart/50-ignore-stop-point.t
```diff
@@ -1,77 +0,0 @@
...
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#-------------------------------------------------------------------------------
-# Test restart with ignore stop point
```
MetRonnie, line 18: Merged this test with `08-stop-after-cycle-point.t `

oliver-sanders:
> Change ignore to refresh, as we are actually refreshing from the flow.cylc file

Perhaps "reload" would make more sense (matches `cylc reload`).

MetRonnie:
> Perhaps "reload" would make more sense (matches `cylc reload`).

Happy to, @hjoliver do you agree?

hjoliver:
> > Perhaps "reload" would make more sense (matches `cylc reload`).
> 
> Happy to, @hjoliver do you agree?

Yes, agreed

cylc/flow/scheduler.py
```diff
@@ -1844,35 +1844,39 @@ def _check_startup_opts(self) -> None:
...
-        else:
-            for opt in ('icp', 'fcp', 'startcp', 'stopcp', 'starttask'):
-                if getattr(self.options, opt, None) == 'ignore':
-                    raise SchedulerError(
-                        f"option --{opt}=ignore is only valid for restart.")
+        for opt in ('icp', 'startcp', 'starttask'):
+            value = getattr(self.options, opt, None)
+            if self.is_restart:
+                if value is not None:
+                    raise UserInputError(
```
oliver-sanders, line 1853: Checked that this change of exception doesn't mess with the error catching.

Works ok, changes the exit code from `1` to `3` but that doesn't matter.

Error message changes from:

```
2021-08-05T10:39:28+01:00 ERROR - Workflow shutting down - option --startcp=reload is not valid (only --fcp and --stopcp can be 'reload')
```

To:

```
2021-08-05T10:38:53+01:00 ERROR - Workflow shutting down - UserInputError: option --startcp=reload is not valid 
```

oliver-sanders (approved):

hjoliver (approved):
:+1:

PR #4324, opened 2021-07-23
graph: restore basic cylc 7 static graphing functionality

oliver-sanders:
Restore basic Cylc 7 static graphing functionality.

Basically a Pythonic version of my bash script with a couple of extra goodies.

Should be sufficient to reimplement the tutorial:

* Namespace graphing.
* Group by cycle point.
* Transpose graph.

Requires graphviz to be installed. Has an optional dependency on pillow for opening images in the default application.

**Requirements check-list**
  `conda-environment.yml`.
- [x] No tests, temporary functionality that will be removed once the GUI is finished.
- [ ] Documentation updates required, partially covered by existing issues however the cheat sheet needs updating.


hjoliver:
I had a quick play with this, looks good.

It revealed a strange feature of `cylc graph --reference` though (also on master). Using this recent troublemaker:
```
[scheduling]
  [[graph]]
      R1 = """
a1 => b1
a1:fail | b1 => c => d1
a1:fail & c => !a1 & !b1 & !d1
a2 => b2
a2:fail | b2 => c => d2
a2:fail & c => !a2 & !b2 & !d2
"""
[runtime]
  [[root]]
     script = sleep 5
  [[a1, b1, d1]]
  [[a2, b2, d2]]
  [[c]]
  [[a2]]
     script = false
```
... each run of `cylc graph` seems to choose a random combination of `d1` and `d2` to omit from the graph: `d1` or `d2` or both :thinking:

hjoliver:
Yikes, I just found the above issue affects running the workflow too. Investigating...

hjoliver:
Forget it, that graph is wrong and will fail validation after #4335 :grim

wxtim (changes requested):
- [x] Read through code changes.
- [x] Had a good play. 

The only change I might suggest is a quick review of existing options as descibed by `cylc graph --help`: I think that the following are effectively disabled by these changes:

- Headline is out of date: "Implements the old ``cylc graph --reference command`` for producing a textual graph of a workflow."
- `--show-suicide` - doesn't seem to be doing anything for me.
- `--color` - doesn't seem to do anything for me.

oliver-sanders:
> Headline is out of date: "Implements the old cylc graph --reference command for producing a textual graph of a workflow."

Done

> --show-suicide - doesn't seem to be doing anything for me.

Try this example, you should see an arrow between `a` and `b`, it doesn't get a special symbol like in Cylc 7:

```
[scheduler]
    allow implicit tasks = True

[scheduling]
    [[graph]]
        R1 = """
            a => !b
            a => c
        """
```

> --color - doesn't seem to do anything for me.

`--color` is a standard Cylc opt added to all commands, it is not used in `cylc graph`.

cylc/flow/scripts/graph.py
```diff
@@ -215,19 +243,123 @@ def get_option_parser():
...
+
+    """
+    if not which('dot'):
+        sys.exit('Graphviz must be installed to render graphs.')
+
+    # set filename and output format
+    if opts.output:
+        filename = opts.output
+        try:
+            fmt = filename.split('.')[1]
```
MetRonnie, line 261: In case someone puts more than 1 dot in the filename, should it use the first or last?
```suggestion
            fmt = filename.rsplit('.', 1)[1]
```

cylc/flow/scripts/graph.py
```diff
@@ -215,19 +243,123 @@ def get_option_parser():
...
+                    f'  subgraph "cluster_{cycle}" {{ ',
+                    f'    label="{cycle}"',
+                    '    style="dashed"',
+                ]
+            )
+        dot.extend(
+            [
+                rf'    "{task}.{cycle}" [label="{task}\n{cycle}"]'
+                for task in tasks
+            ]
```
MetRonnie, line 312: I thought we had a flake8 extension to catch these, but regardless `list.extend()` accepts iterables
```suggestion
            rf'    "{task}.{cycle}" [label="{task}\n{cycle}"]'
            for task in tasks
```

cylc/flow/scripts/graph.py
```diff
@@ -215,19 +243,123 @@ def get_option_parser():
...
+        )
+        if opts.cycles:
+            dot.append('  }' if opts.cycles else '')
```
MetRonnie, line 315: Should this be
```suggestion
        dot.append('  }' if opts.cycles else '')
```
or
```suggestion
        if opts.cycles:
            dot.append('  }')
```

MetRonnie (commented):
Tested with the `tutorial-family-triggers` workflow :+1:

meow

MetRonnie:
> > Headline is out of date: "Implements the old cylc graph --reference command for producing a textual graph of a workflow."
> 
> Done

I'm not seeing that change, has it not been pushed?

oliver-sanders:
> I'm not seeing that change, has it not been pushed?

Ach, no, because GH was down.

cylc/flow/scripts/graph.py
oliver-sanders: I don't think `dot` accepts any formats like that but why not.

cylc/flow/scripts/graph.py
oliver-sanders: Nice, I didn't know that.

Here's where the comprehension suggestions come from: https://github.com/adamchainz/flake8-comprehensions

Doesn't look like they have a rule for that, could create an issue.

MetRonnie (approved):
(Not clicking merge as there's a test failure, may be a one-off though)

wxtim (approved):
Happy with response to previous.

wxtim:
> it doesn't get a special symbol like in Cylc 7

I had been hoping for a special symbol.

oliver-sanders:
Could potentially bring back but would be more work (would have to work with the graph data rather than just reformatting the reference output).

wxtim:
> Could potentially bring back but would be more work (would have to work with the graph data rather than just reformatting the reference output). Hopefully we can get the GUI graph view done before too long...

Well exactly. Just looking at the tutorial and decided to avoid changing some pictures to use this.

PR #4334, opened 2021-07-30
Suicide takes precedence in otherwise identical triggers.

### LINKED ISSUE #4333: Bug in suicide trigger handling
See https://github.com/cylc/cylc-flow/pull/4324#issuecomment-889033099

Consider this Cylc 7 graph:
```
[scheduling]
  [[graph]]
      R1 = """
a1 => b1
a1:fail | b1 => c => d1
a1:fail & c => !a1 & !b1 & !d1
a2 => b2
a2:fail | b2 => c => d2
a2:fail & c => !a2 & !b2 & !d2
"""
[runtime]
  [[root]]
     script = sleep 5
  [[a1, b1, d1]]
  [[a2, b2, d2]]
  [[c]]
  [[a2]]
     script = false
```

(Example first noted here: https://github.com/cylc/cylc-admin/pull/129#issuecomment-882672675)

With `a2` failing, sometimes `d1` runs, sometimes it doesn't.  In the latter case we get spawn-time warnings of invalid cycle point (out of sequence bounds) for `d1` (and sometimes `d2`).

hjoliver:
These changes close #4333 

**Requirements check-list**

I'm not sure how to test this, as the bug is not deterministic. 

- [x] No documentation update required. (now conforms to advertised behavior)

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -1516,9 +1516,6 @@ class WorkflowConfig:
             # Only add sequence to taskdef if explicit (not an offset).
             if offset:
                 taskdef.used_in_offset_trigger = True
-            elif suicide and name == right:
-                # "foo => !bar" should not create taskdef bar
-                pass
             else:
                 taskdef.add_sequence(seq)
 
--- a/cylc/flow/taskdef.py
+++ b/cylc/flow/taskdef.py
@@ -216,8 +216,6 @@ class TaskDef:
             if seq in self.dependencies:
                 # task has prereqs in this sequence
                 for dep in self.dependencies[seq]:
-                    if dep.suicide:
-                        continue
                     for trig in dep.task_triggers:
                         parent_points.add(trig.get_parent_point(point))
         return parent_points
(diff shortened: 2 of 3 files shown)


cylc/flow/taskdef.py
```diff
@@ -216,8 +216,6 @@ def get_parent_points(self, point):
             if seq in self.dependencies:
                 # task has prereqs in this sequence
                 for dep in self.dependencies[seq]:
-                    if dep.suicide:
-                        continue
```
hjoliver, line 220: This was preventing the upstream parent of a suicide trigger being recognized as a parent (matters for spawning if the suicided task has no other parents).

cylc/flow/config.py
```diff
@@ -1516,9 +1516,6 @@ def generate_taskdefs(self, orig_expr, left_nodes, right, seq, suicide):
             # Only add sequence to taskdef if explicit (not an offset).
             if offset:
                 taskdef.used_in_offset_trigger = True
-            elif suicide and name == right:
-                # "foo => !bar" should not create taskdef bar
-                pass
```
hjoliver, line 1521: Comment incorrect. We still need to know the valid sequences of a suicided task, in case it isn't spawned by any non-suicide triggers (as is the case in the example that triggers the bug).

hjoliver:
Invalid!  :stuck_out_tongue_winking_eye:

PR #4335, opened 2021-07-30
Fail "a => b & !b" in validation.

hjoliver:
This is a small change with no associated Issue. Follow up #4334 

**Requirements check-list**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,6 +48,19 @@ creating a new release entry be sure to copy & paste the span tag with the
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
 -------------------------------------------------------------------------------
+## __cylc-8.0b3 (<span actions:bind='release-date'>???</span>)__
+
+Fourth beta release of Cylc 8.
+
+(See note on cylc-8 backward-incompatible changes, above)
+
+### Enhancements
+
+[#4335](https://github.com/cylc/cylc-flow/pull/4335) - have validation catch
+erroneous use of both `expr => bar` and `expr => !bar` in the same graph.
+
+
+-------------------------------------------------------------------------------
 ## __cylc-8.0b2 (<span actions:bind='release-date'>Released 2021-07-28</span>)__
 
 Third beta release of Cylc 8.
--- a/cylc/flow/graph_parser.py
+++ b/cylc/flow/graph_parser.py
@@ -16,6 +16,7 @@
 """Module for parsing cylc graph strings."""
 
 import re
+import contextlib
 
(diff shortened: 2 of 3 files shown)


kinow (commented):
Looks good to me!

Tested and confirmed this is now invalid and raises an error:

```
a & b => c
a => !c
b => !c
```

@hjoliver the example below is not invalid, because `c` is executed only if `a` and `b` succeed. But is not when only one of `a` or `b` succeeds, right?

```
a & b => c
a | b => !c
b | a => !c
```

Also, this runs OK, with no errors:

```
a & b => c
!a => c
```

While this one raises a `GraphParseError`:

```
a & b => c
a:fail => c
```

Isn't `!a` the same as `a:fail`?

My Cylc-Fu is a bit rusty, sorry!

hjoliver:
Good testing @kinow :+1: 

> Isn't `!a` the same as `a:fail?`

No, `a:fail` is when the task job fails (or is killed) while executing.

`!a` is a suicide trigger.  `a => !b` means "if `a` succeeds, remove the task proxy `b` from the scheduler task pool".  You can be forgiven for not remembering that as spawn-on-demand largely consigned suicide triggers to history.  This all just came up again in the context of backward compatibility, to support Cylc 7 graphs full of the bastards.

This should fail, but the parser doesn't recognize `a | b` and `b | a` as the same expression. 
```
a & b => c
a | b => !c
b | a => !c
```
It might be easy to fix that for simple expressions, but in general equivalence of complex logical expressions is ah, beyond the scope of this PR!

hjoliver:
> Also, this runs OK, with no errors:

and

> While this one raises a GraphParseError:

Did you get those two the wrong way around?

The first one should fail, because a suicide `!` doesn't make sense on the LHS of a trigger.

The second one should validate and run, but shutdown with `c` still waiting on `a:failed`.

hjoliver:
> It might be easy to fix that for simple expressions, but in general equivalence of complex logical expressions is ah, beyond the scope of this PR!

Probably not worth the effort given the limited-to-no use for suicide triggers in future.

kinow:
>Did you get those two the wrong way around?

Yes! Sorry :rofl: I used your unit test for the examples. And I got confused when trying to say whether the test passed (i.e. an expected exception was raised) or not.

PR looks good! And I'll be one of the first to review our new documentation & tutorial to brush up on Cylc graph syntax

Thanks @hjoliver !

kinow (approved):

kinow:
And the coverage report is wrong. I debugged @hjoliver and it **is** testing the lines of code added in this PR (which codecov added a comment above saying it is not :angry: ).

PR #4336, opened 2021-07-30
Fix docs build failure.

wxtim:
Cause: In cylc.flow.parsec.config.ConfigNode.__doc__
> Leaf nodes inherited from the generic config will have ``meta=True``.

But this doesn't lead to the localhost platform building properly into the docs.

Fix: Explicitly add copy of section to "localhost" platform.

**Requirements check-list**
  `conda-environment.yml`.
- [x] Does not need tests (Doc change).
- [x] No change log entry required (Only for getting doc build to work).
- [x] _Is_ a documentation update.

Review assigned to @oliver-sanders  as most likely person to be able to suggest a better fix.

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -781,6 +781,17 @@ with Conf('global.cylc', desc='''
                 )
         with Conf('localhost', meta=Platform):
             Conf('hosts', VDR.V_STRING_LIST, ['localhost'])
+            with Conf('selection'):
+                Conf(
+                    'method', VDR.V_STRING, default='random',
+                    options=['random', 'definition order'],
+                    desc='''
+                        .. seealso::
+
+                           :cylc:conf:`global.cylc[platforms]
+                           [<platform name>][selection]`
+                    '''
+                )
 
     # Platform Groups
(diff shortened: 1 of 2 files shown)


oliver-sanders:
Docs build failing on master with, I think that is probably somewhere in one of the cfgspec files (unknown means outside of cylc-doc, few files have 967 lines of RST).

>  <unknown>:967:cylc:conf reference target not found: True

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -781,6 +781,17 @@
                 )
         with Conf('localhost', meta=Platform):
             Conf('hosts', VDR.V_STRING_LIST, ['localhost'])
+            with Conf('selection'):
```
oliver-sanders, line 784: I don't think we should be duplicating this section.

I think the root of the error is that something is trying to reference the concrete platform name `localhost` rather than the abstract `<platform name>`.

We could make a change to cylc-sphinx-extensions to allow concrete references to default to the abstract sections.

Otherwise we can change the reference to point at the abstract config.

MetRonnie:
Closing as superseded by https://github.com/cylc/cylc-sphinx-extensions/pull/50 ?

PR #4339, opened 2021-08-02
Change `[selection]method` to `host selection method` in  `global.cylc[platforms][<name>]`

MetRonnie:
Something about  `global.cylc[platforms][<name>][selection]` was causing the [docs build to fail](https://github.com/cylc/cylc-doc/pull/276#issuecomment-889730698).
> Could not reference "global.cylc[platforms][localhost][selection]True" from the context "global.cylc[platforms][localhost][selection]" in file "reference/config/global".

Simple solution: change the section into a single setting.

Unless we are going to be putting more stuff in a `[selection]` method later?

**Requirements check-list**
  `conda-environment.yml`.
- [x] Does not need tests.
- [x] No documentation update required (can't find any reference to  `global.cylc[platforms][<name>][selection]` in the docs).

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -24,11 +24,17 @@ from cylc.flow import LOG
 from cylc.flow import __version__ as CYLC_VERSION
 from cylc.flow.hostuserutil import get_user_home
 from cylc.flow.network.client_factory import CommsMeth
-from cylc.flow.parsec.config import ParsecConfig, ConfigNode as Conf
+from cylc.flow.parsec.config import (
+    ConfigNode as Conf,
+    ParsecConfig,
+)
 from cylc.flow.parsec.exceptions import ParsecError
 from cylc.flow.parsec.upgrade import upgrader
 from cylc.flow.parsec.validate import (
-    DurationFloat, CylcConfigValidator as VDR, cylc_config_validate)
+    CylcConfigValidator as VDR,
+    DurationFloat,
+    cylc_config_validate,
+)
 
 # Nested dict of spec items.
 # Spec value is [value_type, default, allowed_2, allowed_3, ...]
@@ -709,8 +715,7 @@ with Conf('global.cylc', desc='''
             Conf('install target', VDR.V_STRING, desc='''
                 This defaults to the platform name. This will be used as the
                 target for remote file installation.
-                For example, to indicate to Cylc that Platform_A shares a file
-                system with localhost, we would configure as follows:
+                For example, if Platform_A shares a file system with localhost:
 
                 .. code-block:: cylc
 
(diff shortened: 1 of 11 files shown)


cylc/flow/cfgspec/globalcfg.py
```diff
@@ -745,13 +744,15 @@
...
             ''')
-            Conf('job submission executable paths', VDR.V_STRING_LIST, desc='''
+            Conf('job submission executable paths', VDR.V_STRING_LIST,
+                 desc=f'''
                 Additional executable locations to pass to the job
-                submission subprocess beyond the standard locations''' +
-                 ', '.join(SYSPATH) + '''. You are unlikely to need this.
+                submission subprocess beyond the standard locations
+                {", ".join(f"``{i}``" for i in SYSPATH)}.
+                You are unlikely to need this.
```
MetRonnie, line 755: Wouldn't it be better to explain why you _would_ need this instead of just saying you probably won't need this?

oliver-sanders:
> Unless we are going to be putting more stuff in a [selection] method later?

That's the plan!

oliver-sanders:
The error here is:
```
global.cylc[platforms][localhost][selection]
```
Which should be:
```
global.cylc[platforms][<platform name>][selection]
```

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -763,22 +764,20 @@
...
-                          unless that host has been unreachable. In many cases
-                          this is likely to cause load imbalances, but might
-                          be appropriate if your hosts were
-                          ``main, backup, failsafe``.
-                    '''
-                )
+            Conf('host selection method', VDR.V_STRING, default='random',
+                 options=['random', 'definition order'],
+                 desc='''
+                Method for choosing the job host from the platform.
```
oliver-sanders, line 770:
```suggestion
                Method for choosing the job host from the platform.

```

In RST you need a blank line to separate two paragraphs.

MetRonnie:
>> Unless we are going to be putting more stuff in a [selection] method later?
>
>That's the plan!

Ah right, is there a proposal that details this? I had a look at the [platforms proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-platforms.md) but didn't find anything about it.

MetRonnie:
> The error here is:
> 
> ```
> global.cylc[platforms][localhost][selection]
> ```
> 
> Which should be:
> 
> ```
> global.cylc[platforms][<platform name>][selection]
> ```
> 
> Because the docs thinggy doesn't support referencing absolute config values unless explicitly defined.

But there isn't anything that references that, it's something to do with the fact that `global.cylc[platforms][localhost]` (which has `meta=Platform`) picks up on the `[selection]` section.

If you build without `-W` sphinx option:

MetRonnie:
As discussed today, this is probably an issue in cylc-sphinx-extensions. Leaving this PR open but drafted for now to act as a source of reference, will close later.

MetRonnie:
https://github.com/cylc/cylc-sphinx-extensions/pull/50

PR #4348, opened 2021-08-10
Fix coverage

MetRonnie:

- Apparently coverage doesn't work properly with pytest if you don't `pip install -e` (editable mode)
- Switched from the deprecated Codecov bash uploader to Codecov's dedicated GH Action

**Requirements check-list**
  `conda-environment.yml`.
- [x] Does not need tests.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -44,9 +44,19 @@ timid = False
 exclude_lines =
     pragma: no cover
     def __repr__
+
+    # Don't complain if tests don't hit defensive assertion code:
     raise NotImplementedError
+
+    # Don't complain if non-runnable code isn't run:
     if __name__ == .__main__.:
+    if (typing\.)?TYPE_CHECKING:
+
+    # Don't complain about ellipsis (exception classes, typing overloads etc):
+        \.\.\.
+
     def parse_args
+
 fail_under=0
 ignore_errors = False
 include=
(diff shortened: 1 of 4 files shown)


kinow (approved):
Looks good to me! Codecov bot hasn't commented yet, but maybe it's slow today, or maybe the GH action needs a parameter for that? Not a blocker IMO.

Checked the Codecov page, and it's showing +4% increase for this PR. Good job @MetRonnie ! https://app.codecov.io/gh/cylc/cylc-flow/compare/4348

MetRonnie:
> Codecov bot hasn't commented yet, but maybe it's slow today, or maybe the GH action needs a parameter for that?

We have bot commenting switched off in `codecov.yml`. Is it something we want to turn on?

oliver-sanders:
> We have bot commenting switched off in codecov.yml. Is it something we want to turn on?

NO!

oliver-sanders (approved):
I have absolutely no idea why this would help, nothing in the codecov docs to suggest it would, however, it does appear to bump the coverage score.

My only guess is that uploading a mix of `pip install` and `pip install -e` reports could cause issues, however, that doesn't make much sense as the file paths are all correctly identified in the reports :confused:.

No reason not to do this, should remember to test non-editable installs manually whenever making changes that impact packaging.

PR #4349, opened 2021-08-11
Deltas batched post reload

### LINKED ISSUE #4047: datastore ignores orphaned tasks
If an active task gets orphaned by a reload (i.e. removed from the graph) it disappears from the datastore even though the scheduler allows it to continue running until finished.  It should stay visible as n=0 until done.

(Not a super high priority)

(current master branch)

dwsutherland:
These changes partially address https://github.com/cylc/cylc-ui/issues/667
Closes #4047

This PR batches all the data-store deltas post reload, and sets the `reloaded` flag on all..

Previously only the definition elements (`workflow`, `tasks`, `families`)would be batched and sent, and then the cycle point nodes/elements (`task_proxies`, `family_proxies`, `jobs`, `edges`) would be sent on the next main loop iteration (after being released from the runahead/spawn pool)..

The reason we want one batch, is to replicate an initial burst for the UI.

Along with the sibling PR: https://github.com/cylc/cylc-uiserver/pull/234
this fixes the freeze issue at the UI..

~~There is still a little delay with the reload, but is likely related to the processing of the newly loaded pool (releasing etc... which is when the data-store receives nodes.)~~

- The n-window elements are created as the task pool is reloaded.
- The job elements are created from the DB thereafter.
- Orphaned tasks are now included post reload.

**Requirements check-list**
  `conda-environment.yml`.

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -268,35 +268,41 @@ def apply_delta(key, delta, data):
         for del_id in delta.pruned:
             if del_id not in data[key]:
                 continue
-            # remove relationships
+            # Remove relationships.
+            # The suppression of key/value errors is to avoid
+            # elements and their relationships missing on reload.
             if key == TASK_PROXIES:
                 # remove relationship from task
                 data[TASKS][data[key][del_id].task].proxies.remove(del_id)
                 # remove relationship from parent/family
-                with suppress(KeyError):
+                with suppress(KeyError, ValueError):
                     data[FAMILY_PROXIES][
                         data[key][del_id].first_parent
                     ].child_tasks.remove(del_id)
                 # remove relationship from workflow
-                getattr(data[WORKFLOW], key).remove(del_id)
+                with suppress(KeyError, ValueError):
+                    getattr(data[WORKFLOW], key).remove(del_id)
             elif key == FAMILY_PROXIES:
                 data[FAMILIES][data[key][del_id].family].proxies.remove(del_id)
-                with suppress(KeyError):
+                with suppress(KeyError, ValueError):
                     data[FAMILY_PROXIES][
                         data[key][del_id].first_parent
                     ].child_families.remove(del_id)
-                getattr(data[WORKFLOW], key).remove(del_id)
+                with suppress(KeyError, ValueError):
+                    getattr(data[WORKFLOW], key).remove(del_id)
             elif key == EDGES:
                 edge = data[key][del_id]
-                if edge.source in data[TASK_PROXIES]:
+                with suppress(KeyError, ValueError):
                     data[TASK_PROXIES][edge.source].edges.remove(del_id)
-                if edge.target in data[TASK_PROXIES]:
+                with suppress(KeyError, ValueError):
                     data[TASK_PROXIES][edge.target].edges.remove(del_id)
-                getattr(data[WORKFLOW], key).edges.remove(del_id)
+                with suppress(KeyError, ValueError):
+                    getattr(data[WORKFLOW], key).edges.remove(del_id)
             elif key == JOBS:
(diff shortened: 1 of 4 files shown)


cylc/flow/data_store_mgr.py
```diff
@@ -1770,38 +1786,54 @@ def store_node_fetcher(
...
+                            self.deltas[WORKFLOW].added.CopyFrom(elements)
+                        continue
+                    self.deltas[key].added.extend(elements.values())
+            for key, elements in self.updated.items():
+                if elements:
+                    if key == WORKFLOW:
+                        if elements.ListFields():
+                            self.deltas[WORKFLOW].updated.CopyFrom(elements)
+                        continue
+                    self.deltas[key].updated.extend(elements.values())
```
kinow, line 1820: @dwsutherland I think these three blocks of code could be simplified. This part is always the same I think?

```py
            for key, elements in ??????:
                if elements:
                    if key == WORKFLOW:
                        if elements.ListFields():
                            self.deltas[WORKFLOW].?{updated|added}?.CopyFrom(elements)
                        continue
                    self.deltas[key].?{updated|added}?.extend(elements.values())
```

Maybe a function like `gather_delta_elements(elements=elements, key='added')`? Though not too complicated to understand as-is, if you prefer it this way :+1:

kinow (approved):
One comment but not too important. @dwsutherland tested with the other UIS PR and my modified UI, everything worked with no problems. I couldn't tell if the reload was faster, but probably because `five` is such a simple workflow. So leaving the details of the review for Hilary & Oliver now. Thanks!!!

cylc/flow/data_store_mgr.py
dwsutherland: By `?{updated|added}?` you mean `.getattr(delta_type)`?

BTW - key isn't `added` or `updated` .. it's `jobs`, `families` ...etc.

Yeah, I could, in two minds....

kinow: >By ?{updated|added}? you mean .getattr(delta_type)?

I think so, was still thinking in JavaScript, where I could just access a property of an object as in an array element :smile: 

But no biggie on this one, code is working is well contained, so not a necessary change unless you want to :+1:

dwsutherland: Fixed

oliver-sanders (commented):
Makes sense, looks ok.

Had a go at testing with `cylc subscribe`.

```console
$ cylc install <flow>
$ cylc play --pause <flow>
$ cylc subscribe <flow> &
$ cylc reload <flow>
```

When I reload my workflow I seem to get the last delta repeated. I'm not sure if the delta is being sent twice of whether it's a quirk of `cylc subscribe`?

```console
$ cylc subscribe <flow>
Received: workflow
{
    "time": 1630423507.7344673,
    ...
}
# did reload
Received: workflow
{
    "time": 1630423507.7344673,  # hmm, I've already seen this
    ...
}
Received: workflow
{
    "time": 1630423523.7595105,  # this one looks like what I was expecting
    ...
}
```

hjoliver (approved):
:+1: Looks fine to me, tested as working with the sibling PRs.

I'll hold off merging pending response to @oliver-sanders question.

dwsutherland:
> I'm not sure if the delta is being sent twice of whether it's a quirk of cylc subscribe?

@oliver-sanders - It actually needs to be fixed... I think I saw the same when monitoring the network traffic with the UI.

Will investigate.

dwsutherland:
Ok, addressed the duplicate issue.
Happened post reload and pre stop, due to changes being flagged (`has_updated`) in the scheduler but no resultant data_store deltas (which is fine, but the publish call would send the previous delta).

hjoliver (approved):

PR #4354, opened 2021-08-12
codecov: fiddle configuration

oliver-sanders:
Remove some configs which were set to the default values. Some of these configs appear to have since moved, I wonder if that was causing us problems?


hjoliver (approved):
:+1:

PR #4364, opened 2021-08-18
Attempt to fix unit test flakiness on Python 3.7

MetRonnie:

It seems `asyncio.get_event_loop().run_until_complete()` might be more buggy than `asyncio.run()`? But this is the opposite of expected - https://github.com/cylc/cylc-flow/pull/4289#discussion_r664025464

Traceback from unit tests on GH Actions:

```python
>       workflow_files.init_clean(reg, opts=opts)

tests/unit/test_workflow_files.py:377: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
cylc/flow/workflow_files.py:674: in init_clean
    contained_workflows = asyncio.get_event_loop().run_until_complete(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <asyncio.unix_events._UnixDefaultEventLoopPolicy object at 0x7ffa73b578d0>

>                              % threading.current_thread().name)
E           RuntimeError: There is no current event loop in thread 'Dummy-1'.

/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/asyncio/events.py:644: RuntimeError
```

**Requirements check-list**
  `conda-environment.yml`.
- [x] Does not need tests

### DIFF
--- a/.github/workflows/test_functional.yml
+++ b/.github/workflows/test_functional.yml
@@ -1,8 +1,17 @@
 name: functional tests
 
 on:
-  pull_request:
   workflow_dispatch:
+  pull_request:
+    paths-ignore:
+      - '.github/workflows/*.ya?ml'
+      - '!.github/workflows/test_functional.yml'
+      - 'cylc/flow/etc/syntax/**'
+      - 'etc/syntax/**'
+      - 'tests/unit/**'
+      - 'tests/integration/**'
+      - '**.md'
+      - '**/README*/**'
   push:
     branches: [master]
     paths-ignore:
--- a/.github/workflows/bash.yml
+++ b/.github/workflows/bash.yml
@@ -2,10 +2,28 @@
 name: bash
 
 on:
-  pull_request:
   workflow_dispatch:
+  pull_request:
+    paths-ignore:
+      - '.github/workflows/*.ya?ml'
+      - '!.github/workflows/bash.yml'
+      - 'cylc/flow/etc/syntax/**'
+      - 'etc/syntax/**'
+      - 'tests/unit/**'
+      - 'tests/integration/**'
+      - '**.md'
+      - '**/README*/**'
   push:
     branches: [master]
+    paths-ignore:
+      - '.github/workflows/*.ya?ml'
+      - '!.github/workflows/bash.yml'
+      - 'cylc/flow/etc/syntax/**'
+      - 'etc/syntax/**'
+      - 'tests/unit/**'
+      - 'tests/integration/**'
+      - '**.md'
+      - '**/README*/**'
 
 jobs:
   bash-docker:
--- a/tests/unit/test_workflow_files.py
+++ b/tests/unit/test_workflow_files.py
@@ -191,6 +191,7 @@ def test_clean_check__fail(
     assert err_msg in str(exc.value)
 
 
+@pytest.mark.asyncio
 @pytest.mark.parametrize(
     'db_platforms, opts, clean_called, remote_clean_called',
     [
@@ -240,6 +241,7 @@ def test_init_clean(
     assert mock_remote_clean.called is remote_clean_called
 
 
+@pytest.mark.asyncio
 def test_init_clean__no_dir(
     monkeymock: MonkeyMock, tmp_run_dir: Callable,
     caplog: pytest.LogCaptureFixture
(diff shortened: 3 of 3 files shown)


kinow (commented):
@MetRonnie I think this change should be fine. But I suspect we might be experiencing this `pytext-xdist` issue: https://github.com/pytest-dev/pytest-xdist/issues/620

I tried to reproduce it on PyCharm, and noticed my test was always running on `MainThread`. I think the ioloop for tests (created by `pytest-asyncio`?) should be available for that thread. But from your traceback, it looks like the test failed when running on `Dummy-1` thread.

Looking at the IDE, I can see other `Dummy-2`, `Dummy-6` threads, which were probably created by libs/tools for running the tests. But the tests themselves appear to be running on the MainThread.

Isn't it risk to use `asyncio.run` if we may have other ioloop already running? I'm not sure if we have as I haven't looked much at `workflow_files.py` and how it's used in the code. @oliver-sanders can probably answer that.

From [their docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.run):

>This function cannot be called when another asyncio event loop is running in the same thread.

Maybe decorating the brittle test with `@pytest.mark.asyncio` would ensure the ioloop to be present no matter what thread it's running on?

MetRonnie:
> @MetRonnie I think this change should be fine. But I suspect we might be experiencing this `pytext-xdist` issue: [pytest-dev/pytest-xdist#620](https://github.com/pytest-dev/pytest-xdist/issues/620)

Good spot!

> Maybe decorating the brittle test with `@pytest.mark.asyncio` would ensure the ioloop to be present no matter what thread it's running on?

Hadn't thought of it, it's worth a shot. However, perhaps we should instead just strip out the async-ness? There is already some replication of the scan functionality here
https://github.com/cylc/cylc-flow/blob/26cc7091f91f66344fd7df8ec44fcb3794117649/cylc/flow/workflow_files.py#L1100-L1108

MetRonnie:
(Cancelled the functional + bash tests)

kinow:
>Hadn't thought of it, it's worth a shot. However, perhaps we should instead just strip out the async-ness? There is already some replication of the scan functionality here

Sounds like a possible solution too :+1: , but will leave that one for @oliver-sanders to answer since I think he wrote it and he's way more familiar with asyncio than I.

.github/workflows/test_functional.yml
```diff
@@ -1,8 +1,17 @@
...
+  pull_request:
+    paths-ignore:
+      - '.github/workflows/*.ya?ml'
+      - '!.github/workflows/test_functional.yml'
+      - 'cylc/flow/etc/syntax/**'
+      - 'etc/syntax/**'
+      - 'tests/unit/**'
+      - 'tests/integration/**'
+      - '**.md'
+      - '**/README*/**'
```
kinow, line 14: :+1: good idea! We need to start adding these to the other repos later too. Thanks!

datamel (approved):
Thanks Ronnie, looks good.

oliver-sanders:
> It seems asyncio.get_event_loop().run_until_complete() might be more buggy than asyncio.run()?

I don't think these changes move us to `asyncio.run`?

> Traceback from unit tests on GH Actions:

I think this is an error in the Cylc code in that it is assuming that an event loop will have been initiated which isn't necessarily the case.

Here's how we have handled this situation in other code:

https://github.com/cylc/cylc-flow/blob/7f635a285bee80210477ac3afb4c8f635083f581/cylc/flow/network/__init__.py#L158-L162

My guess it that if these tests get run *before* other `async` tests we get an error so the flakeyness is a race condition in our test battery.

Adding `pytest.mark.asyncio` to non-async fixtures that the test depends on will cause the event loop policy we configured in `pytest-asyncio` to be activated which creates one event loop per test module:

https://github.com/cylc/cylc-flow/blob/7f635a285bee80210477ac3afb4c8f635083f581/tests/integration/conftest.py#L235-L257

MetRonnie:
> > It seems asyncio.get_event_loop().run_until_complete() might be more buggy than asyncio.run()?
> 
> I don't think these changes move us to `asyncio.run`?

Sorry, originally that was what I did but then force-pushed with the `@pytest.mark.asyncio` markers instead, following Bruno's feedback

Issue #4165, opened 2021-04-08
cylc wrapper: arbitrary environment names


hjoliver:
Currently the `cylc` wrapper requires virtual envs to be named as the cylc-flow version, so that users can continue to use `$CYLC_VERSION` Cylc 7 style.

Setting `CYLC_VERSION=my-env-name` doesn't work, because `CYLC_VERSION` is set to `cylc.flow.__version__` in job scripts, so that's what the wrapper sees for job-invoked commands.

This means we can't have multiple installs at the same release version, which isn't ideal because we might want to mix-and-match component versions within the full package, or have different plugins installed in different envs.

We should make the wrapper work with arbitrary environment names.  Probably requires a new variable `CYLC_VIRTUAL_ENV` or similar, that gets written to jobs scripts in addition to `CYLC_VERSION`.

hjoliver:
Meeting: agreed we need this. Probably need to use a new variable in addition to `CYLC_VERSION` .. .e.g.  `CYLC_ENVIRONMENT_NAME`, to allow fully arbitrary names.  Also consider `CYLC_ENVIRONMENT_TYPE` (e.g. "conda", "conda-pack", ...) so we don't have to figure out what the environment is and hence how to activate it.

oliver-sanders:
Clarification.

After this change `CYLC_VERSION` becomes a user-facing feature allowing users to conveniently select an environment without having to learn the env naming pattern.

### input(CYLC_VERSION)

We derive `CYLC_ENV_NAME` from `CYLC_VERSION` and export it.

### input(CYLC_ENV_NAME)

We activate `$CYLC_ENV_NAME`

### input(CYLC_ENV_NAME, CYLC_VERSION)

We activate `$CYLC_ENV_NAME`

oliver-sanders:
(closed by #4367)

Issue #3704, opened 2020-07-22
Investigate wrappers for environment activation


hjoliver:
Conda environment activation is reputedly quite slow.

`conda run` might address this in future, but doesn't yet? (and we [can't use it in shared environments](https://github.com/conda/conda/issues/10098))

This small project makes some impressive claims, we should take a look: [exec wrappers](https://github.com/gqmelo/exec-wrappers). (from an offline comment: *(it might just be sourcing the python interpreter which won't work for all cases)*?)

oliver-sanders:
The slow activate times for Conda really confuse me, what the heck is it doing?

Well according to [the code](https://github.com/conda/conda/blob/901afcffb77b9a1132bfdb8381b2ce8afdc28c69/conda/activate.py#L33-L35):

1. Set and unset environment variables
2. Execute/source activate.d/deactivate.d scripts
3. Update the command prompt

(1) should be quick, for (2) I cant see any `activate.d`/`deactivate.d` dirs in my 8.0a2 environment and we don't need to update the command prompt for non-interactive commands.

So what's it actually doing?

```diff
$ env > before
$ conda activate cylc8
$ env > after
$ diff before after
--- before	2020-07-22 23:27:46.000000000 +0100
+++ after	2020-07-22 23:27:53.000000000 +0100
@@ -8,19 +8,22 @@
 GPG_TTY=/dev/ttys002
 PWD=/Users/user/miniconda3
 LOGNAME=user
+CONDA_PREFIX=/Users/user/miniconda3/envs/cylc8
 HOME=/Users/user
 LANG=en_GB.UTF-8
+CONDA_PROMPT_MODIFIER=(cylc8) 
 TMPDIR=/var/folders/rg/691l5s513_97d5s7ktr1rtt00000gn/T/
 TERM=xterm-256color
 _CE_CONDA=
 USER=user
-CONDA_SHLVL=0
+CONDA_SHLVL=1
 SHLVL=1
 HOME_HOST=myhost
 XPC_SERVICE_NAME=0
 CONDA_PYTHON_EXE=/Users/user/miniconda3/bin/python
-PS1=\[\][\W] $\[\] 
-PATH=/Users/user/miniconda3/condabin:/usr/local/opt/gnu-sed/libexec/gnubin:/usr/local/opt/coreutils/libexec/gnubin:/Users/user/bin:/Users/user/.user/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
[...]

PR #4367, opened 2021-08-19
Cylc Wrapper uses arbitrary env set at CYLC_ENV_NAME

wxtim:
These changes address the bulk of #4165 

**Ticket Specific Checklist**
- [x] Use a new variable, `CYLC_ENV_NAME`, to allow fully arbitrary names.
- [ ] Also consider CYLC_ENVIRONMENT_TYPE. Not considered. Mind if I leave it to a future PR?
- [ ] Testing?
  - [x] Manual testing
    - [x] `CYLC_ENV_NAME=foo cylc version --long`
    - [x] `CYLC_ENV_NAME=foo CYLC_VERSION=8.0b3 cylc version --long`
    - [x] `CYLC_VERSION=8.0b3 cylc version --long`
    - [x] Ran a workflow with a remote task and checked the logfiles
  - [ ] Automated testing @hjoliver  - any thoughts? Are there any existing automated tests for this functionality? 

**Requirements check-list**

### DIFF
--- a/usr/bin/cylc
+++ b/usr/bin/cylc
@@ -145,7 +145,7 @@ fi
 # Note: the code above is only needed if still using standalone Rose 2019
 # versions
 
-# CYLC_NAME construction:
+# CYLC_ENV_NAME construction:
 #                           ╔════════════════════════════╗
 #                           ║        CYLC_VERSION        ║
 #                           ╠══════════════════╦═════════╣
--- a/usr/bin/cylc
+++ b/usr/bin/cylc
@@ -147,9 +147,16 @@ fi
 
 if [[ -z "${CYLC_HOME}" ]]; then
     # Look for specified or default version under the root locations
+    # Build CYLC_NAME from CYLC_VERSION if you can:
     if [[ -n "${CYLC_VERSION}" ]]; then
         CYLC_NAME="cylc-$CYLC_VERSION"
-    else
+        echo "cylc name is $CYLC_NAME"
+    fi
+    # Build CYLC_NAME from CYLC_ENV_NAME if you can:
+    if [[ -n "${CYLC_ENV_NAME}" ]]; then
+        CYLC_NAME="${CYLC_ENV_NAME}"
+    fi
+    if [[ -z "${CYLC_ENV_NAME}" && -z "${CYLC_VERSION}" ]]; then
         # Default version symlink
         CYLC_NAME="cylc"
     fi
--- a/usr/bin/cylc
+++ b/usr/bin/cylc
@@ -64,7 +64,8 @@
 # And create a default version symlink "cylc" to the latest version:
 #   $ ln -s cylc-8.2.0 cylc
 #
-# Alternatively use CYLC_ENV_NAME to use Cylc from an arbitrary Conda environment:
+# Alternatively use CYLC_ENV_NAME to use Cylc from an arbitrary Conda or Python
+# virtual environment:
 #
 # Create environments with arbitrary names and install cylc:
 #   $ conda create cylc-violet-capybara
--- a/usr/bin/cylc
+++ b/usr/bin/cylc
@@ -64,15 +64,7 @@
 # And create a default version symlink "cylc" to the latest version:
 #   $ ln -s cylc-8.2.0 cylc
 #
-# Note you can `pip install` from a cylc-flow git clone into a Conda
-# environment and then select it by CYLC_VERSION but that may be misleading
-# if the code is not the true release version. Better to use Python venvs and
-# CYLC_HOME for development and testing.
-#
-# INSTALLING cylc 8+ WITH PIP IN CONDA ENVIRONMENTS WITH ARBITRARY NAMES
-#-----------------------------------------------------------------------
-#
-# You can use CYLC_ENV_NAME to use Cylc from an arbitrary Conda environment:
+# Alternatively use CYLC_ENV_NAME to use Cylc from an arbitrary Conda environment:
 #
 # Create environments with arbitrary names and install cylc:
 #   $ conda create cylc-violet-capybara
@@ -81,7 +73,7 @@
 #   $ conda deactivate
 #
 #   $ CYLC_ENV_NAME=cylc-violet-capybara cylc version --long
-#   > 8.0.whateva (/home/user/miniconda3/envs/cylc-violet-capybara)
+#   > 8.0.0 (/home/user/miniconda3/envs/cylc-violet-capybara)
 #
 # n.b. if CYLC_VERSION and CYLC_ENV_NAME are both set, CYLC_ENV_NAME will be
 # used and CYLC_VERSION ignored.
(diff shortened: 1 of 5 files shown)


usr/bin/cylc
```diff
@@ -147,9 +147,16 @@ fi
...
+        echo "cylc name is $CYLC_NAME"
+    fi
+    # Build CYLC_NAME from CYLC_ENV_NAME if you can:
+    if [[ -n "${CYLC_ENV_NAME}" ]]; then
+        CYLC_NAME="${CYLC_ENV_NAME}"
+    fi
+    if [[ -z "${CYLC_ENV_NAME}" && -z "${CYLC_VERSION}" ]]; then
         # Default version symlink
         CYLC_NAME="cylc"
     fi
```
oliver-sanders, line 162: We would expect *both* the env name and the version to be specified much of the time so these two need to be exclusive.

The env name should be used in preference to the version.

oliver-sanders:
> CYLC_ENVIRONMENT_TYPE

Actually hold off on that, I think it might become obsolete with https://github.com/cylc/cylc-flow/issues/3704

> Testing? 

Manual for now.

> Automated testing

Manual for now.

usr/bin/cylc
wxtim: > Note we need to export the env name in this script.

Why?

wxtim:
> Manual for now.

😢 😞 😿

usr/bin/cylc
wxtim: > so these two need to be exclusive.

I thought that I had made them exclusive:
```
if CYLC_VERSION
  use it
if CYLC_ENV_NAME
  use it (regardless of whether you just used it for CYLC_VERSION)
else:
  whateva
```
But I think I've made the logic clearer.

usr/bin/cylc
```diff
@@ -145,23 +145,33 @@ fi
 # Note: the code above is only needed if still using standalone Rose 2019
 # versions
 
+# CYLC_NAME construction:
```
hjoliver, line 148:
```suggestion
# CYLC_ENV_NAME construction:
```

usr/bin/cylc
```diff
@@ -145,23 +145,33 @@ fi
...
+#                           ╔════════════════════════════╗
+#                           ║        CYLC_VERSION        ║
+#                           ╠══════════════════╦═════════╣
+#                           ║        set       ║ not set ║
+# ╔═══════════════╦═════════╬══════════════════╩═════════╣
+# ║ CYLC_ENV_NAME ║   set   ║      Use CYLC_ENV_NAME     ║
+# ║               ╠═════════╬══════════════════╦═════════╣
+# ║               ║ not set ║ Use CYLC_VERSION ║  "cylc" ║
+# ╚═══════════════╩═════════╩══════════════════╩═════════╝
```
hjoliver, line 157: Nice :grin:

hjoliver: Note the comments from lines 23-114 will need updating.  (Possibly to be moved to the user guide in due course).

usr/bin/cylc
```diff
@@ -145,23 +145,33 @@ fi
...
         # Default version symlink
-        CYLC_NAME="cylc"
+        CYLC_ENV_NAME="cylc"
     fi
     for ROOT in "${CYLC_HOME_ROOT}" "${CYLC_HOME_ROOT_ALT}"; do
-        if [[ -d "${ROOT}/${CYLC_NAME}" ]]; then
-            CYLC_HOME="${ROOT}/${CYLC_NAME}"
+        if [[ -d "${ROOT}/${CYLC_ENV_NAME}" ]]; then
+            CYLC_HOME="${ROOT}/${CYLC_ENV_NAME}"
+            export CYLC_ENV_NAME
```
hjoliver, line 168: I don't think we need this `export`?  I guess it would allow a tiny shortcut in the scripting above if `cylc` is invoked again in subshells, but it would be clearer to have subshells work out the path in exactly the same way as the parent.

Ping @oliver-sanders as it was evidently his suggestion?

hjoliver:
> Automated testing @hjoliver - any thoughts? Are there any existing automated tests for this functionality?

There's been no automated wrapper testing to date.  I'm sure we could do it by setting up the right environment in a container, but might be best to leave that to 8.x.

usr/bin/cylc
oliver-sanders: We may have multiple environments containing the same Cylc version which could lead to different results when activating via the `CYLC_VERSION` or `CYLC_ENV_NAME`.

usr/bin/cylc
wxtim: I've had a look.

usr/bin/cylc
hjoliver: I don't think we even need to set it here. By this point, it is used to set CYLC_HOME, and that's what's used to activate the environment.

hjoliver:
(Seems to be working well, in my testing)

usr/bin/cylc
dpmatthews: I think the only case where we need to `export CYLC_ENV_NAME` is when we're using the default symlink version. We have to export it because cylc will always set `CYLC_VERSION` and we can no longer assume that the symlink version points to `cylc-$CYLC_VERSION`.

usr/bin/cylc
```diff
@@ -174,16 +168,18 @@ fi
...
-        CYLC_ENV_NAME="cylc-$CYLC_VERSION"
-    elif [[ -z "${CYLC_ENV_NAME}" && -z "${CYLC_VERSION}" ]]; then
-        # Default version symlink
-        CYLC_ENV_NAME="cylc"
+    if [[ -z "${CYLC_ENV_NAME}" ]]; then
+        if [[ -n "${CYLC_VERSION}" ]]; then
+            CYLC_ENV_NAME="cylc-$CYLC_VERSION"
+        else
+            # Default version symlink - export CYLC_ENV_NAME to ensure it
+            # does nott get overridden by CYLC_VERSION
```
wxtim, line 176:
```suggestion
            # does not get overridden by CYLC_VERSION
```

dpmatthews (approved):
Not tested by me but logic looks good

wxtim:
@hjoliver  & @oliver-sanders , Dave has now approved the logic. Might you test the functionality?

hjoliver:
> @hjoliver & @oliver-sanders , Dave has now approved the logic. Might you test the functionality?

:+1: I'll take a final look and test first thing tomorrow.

usr/bin/cylc
```diff
@@ -57,16 +57,26 @@
...
 #   $ conda create -p /opt/cylc-8.2.0 cylc=8.2.0
 #
 # And create a default version symlink "cylc" to the latest version:
 #   $ ln -s cylc-8.2.0 cylc
 #
-# Note you can `pip install` from a cylc-flow git clone into a Conda
-# environment and then select it by CYLC_VERSION but that may be misleading
-# if the code is not the true release version. Better to use Python venvs and
-# CYLC_HOME for development and testing.
+# Alternatively use CYLC_ENV_NAME to use Cylc from an arbitrary Conda environment:
```
hjoliver, line 67:
```suggestion
# Alternatively use CYLC_ENV_NAME to use Cylc from an arbitrary Conda or Python
# virtual environment:
```

usr/bin/cylc
```diff
@@ -57,16 +57,26 @@
...
+#
+# Create environments with arbitrary names and install cylc:
+#   $ conda create cylc-violet-capybara
+#   $ conda activate cylc-violet-capybara
+#   $ pip install -e <path to cylc>
+#   $ conda deactivate
+#
```
hjoliver, line 74:
```suggestion
# (or similarly for a native Python venv)
```

hjoliver (approved):
Looks good, thanks @wxtim

hjoliver:
Worth noting in the comments (and later, user guide) that `CYLC_ENV_NAME` can be an absolute path to an environment, or relative to `CYLC_HOME_ROOT` - particularly as the latter can be needed when setting up a matching venv on a remote job host. (Tested).  This is only a comment however, so I'll merge this now and add it along with a change log entry.

usr/bin/cylc
hjoliver: (Oops redundant with subsequent comments, I'll remove in the change log commit)

Issue #4369, opened 2021-08-19
Should `cylc clean` re-link the runN symlink to the next-highest run number?


MetRonnie:
Follow-up to #4323

Say you have
```
~/cylc-run/myflow
|-- run1
|-- run2
`-- runN -> run2
```

Currently, `cylc clean myflow/run2` will just remove `runN`. Should it instead change the target of `runN` to the next-latest run, in this case `run1`?

Discussion from #4323:

> @MetRonnie: What reason is there not to choose the next-highest run number?

> @dpmatthews: The `runN` link is designed to make it easy to access the latest run (e.g. paths to data files). If you delete the latest run does that mean that you're happy for `runN` to point to an older run? Maybe in some cases but not others? I think removing the link is the safer option.

---

Note: currently, this behaviour leaves `cylc install` broken afterwards, see #4368

MetRonnie:
If the answer is no, perhaps we could include a cli option like `--relink` to recreate the `runN` symlink? Could be an option of `cylc clean` or `cylc install`?

MetRonnie:
Maybe if the user is not happy for `runN` to point to an older run after deletion of the latest, they should be using named runs instead of numbered runs?

MetRonnie:
Pinging @cylc/core

oliver-sanders:
I agree with Dave, if you remove the latest run the `runN` link should go with it.

A subsequent installation would add a new run and restore the link.

hjoliver:
@wxtim to double check for dangling symlink and close this.

wxtim:
> @wxtim to double check for dangling symlink and close this.

I can confirm that dangling symlinks are removed by Cylc Clean.

PR #4373, opened 2021-08-23
Attempt to infer run number if there is no runN symlink.

### LINKED ISSUE #4368: `cylc install` broken after `cylc clean`
**Describe the bug**

`cylc install` breaks after cleaning the latest run of a workflow, leaving behind older runs. This is likely due to the lack of `runN` symlink afterwards.

master

```console
$ cylc install myflow
INSTALLED myflow/run1 from ~/cylc-src/myflow
$ cylc install myflow
INSTALLED myflow/run2 from ~/cylc-src/myflow
$ tree -L 1 ~/cylc-run/myflow
~/cylc-run/myflow/
|-- _cylc-install
|-- run1
|-- run2
`-- runN -> run2

$ cylc clean myflow/run2
INFO - No workflow database - will only clean locally
INFO - Cleaning on local filesystem: ~/cylc-run/myflow/run2
$ tree -L 1 ~/cylc-run/myflow
~/cylc-run/myflow/
|-- _cylc-install
`-- run1

$ cylc install myflow
WorkflowFilesError: "~/cylc-run/myflow/run1" exists. Try using cylc reinstall.
    Alternatively, install with another name, using the --run-name option.
```

**Expected behavior**

`cylc install` should create `myflow/run2` and symlink `runN -> run2`.

**Pull requests welcome!**

wxtim:
These changes close #4368 

As Cylc clean now removed the `runN` symlink we may have to guess what the next `runX` folder should be from files rather than reading the symlink.

**Requirements check-list**
- [x] No change log entry required: Immediate fix to change made very recently in unreleased beta branch.

### DIFF
--- a/cylc/flow/pathutil.py
+++ b/cylc/flow/pathutil.py
@@ -335,15 +335,33 @@ def remove_empty_parents(
             break
 
 
-def get_next_rundir_number(run_path):
-    """Return the new run number"""
-    run_n_path = os.path.expanduser(os.path.join(run_path, "runN"))
-    try:
+def get_next_rundir_number(run_path: Union[str, Path]) -> int:
+    """Return the new run number
+
+    Args:
+        run_path: Top level installed workflow dir
+        (often ``~/cylc-run/workflow``).
+
+    """
+    run_n_path = Path(os.path.expanduser(os.path.join(run_path, "runN")))
+    if run_n_path.exists() and run_n_path.is_symlink():
         old_run_path = os.readlink(run_n_path)
-        last_run_num = re.search(r'(?:run)(\d*$)', old_run_path).group(1)
-        return int(last_run_num) + 1
-    except OSError:
-        return 1
+        # Line below could in theory not return a match group, so mypy objects.
+        # This function unlikely to be called in circumstances where this will
+        # be a problem.
+        last_run_num = re.search(
+            r'(?:run)(\d*$)', old_run_path).group(1)  # type: ignore
+        last_run_num = int(last_run_num)
+    else:
+        # If the ``runN`` symlink has been removed, make a sensible guess at
+        # what the next numbered run should be.
+        last_run_num = 0
+        for rundir in Path(run_path).glob('run[0-9]*'):
+            highest_yet = int(str(rundir).split('run')[-1])
+            if highest_yet > last_run_num:
+                last_run_num = highest_yet
+
+    return last_run_num + 1
 
 
 def parse_rm_dirs(rm_dirs: Iterable[str]) -> Set[str]:
--- a/tests/unit/test_pathutil.py
+++ b/tests/unit/test_pathutil.py
@@ -19,6 +19,7 @@ import logging
 import os
 from pathlib import Path
 import pytest
+from pytest import param
 from typing import Callable, Dict, Iterable, List, Set
 from unittest.mock import Mock, patch, call
 
@@ -26,6 +27,7 @@ from cylc.flow.exceptions import UserInputError, WorkflowFilesError
 from cylc.flow.pathutil import (
     expand_path,
     get_dirs_to_symlink,
+    get_next_rundir_number,
     get_remote_workflow_run_dir,
     get_remote_workflow_run_job_dir,
     get_workflow_run_dir,
@@ -491,3 +493,23 @@ def test_parse_rm_dirs__bad(dirs: List[str], err_msg: str):
     with pytest.raises(UserInputError) as exc:
         parse_rm_dirs(dirs)
     assert err_msg in str(exc.value)
+
+
+@pytest.mark.parametrize(
+    'expect, files, runN',
+    [
+        param(1, [], False, id='1st run (from filenames)'),
+        param(2, ['run1'], False, id='2nd run (from filenames)'),
+        param(1000, ['run999'], False, id='1000th run (from filenames)'),
+        param(6, ['run1', 'run5'], False,
+            id='Non-sequential (from filenames)'),
+        param(2, ['run1'], True, id='2nd run (from symlink)'),
+        param(100, ['run1', 'run99'], True, id='100th run (from symlink)')
+    ]
+)
+def test_get_next_rundir_number(tmp_path, expect, files, runN):
+    for file_ in files:
+        (tmp_path / file_).mkdir()
+    if runN:
+        (tmp_path / 'runN').symlink_to(tmp_path / files[-1])
+    assert get_next_rundir_number(tmp_path) == expect


MetRonnie:
> As Cylc clean now removed the `runN` symlink we may have to guess what the next `runX` folder should be from files rather than reading the symlink.

I'm not sure we want to do that. Perhaps it is best to wait until #4369 is answered

wxtim:
If we don't get the number from the symlink, where else is there to go?

cylc/flow/pathutil.py
```diff
@@ -335,15 +335,33 @@ def remove_empty_parents(
...
         old_run_path = os.readlink(run_n_path)
-        last_run_num = re.search(r'(?:run)(\d*$)', old_run_path).group(1)
-        return int(last_run_num) + 1
-    except OSError:
-        return 1
+        # Line below could in theory not return a match group, so mypy objects.
+        # This function unlikely to be called in circumstances where this will
+        # be a problem.
+        last_run_num = re.search(                        # type: ignore
+            r'(?:run)(\d*$)', old_run_path).group(1)     # type: ignore
```
MetRonnie, line 353: Maybe error handling for if the `runN` symlink points to an invalid location would be nice, à la
https://github.com/cylc/cylc-flow/blob/7f635a285bee80210477ac3afb4c8f635083f581/cylc/flow/workflow_files.py#L1213-L1216

But it should ok to ignore this. Even better, ignore the specific rule only. Also, only needs to be on the first line.
```suggestion
        last_run_num = re.search(  # type: ignore[union-attr]
            r'(?:run)(\d*$)', old_run_path
        ).group(1)
```

cylc/flow/pathutil.py
```diff
@@ -335,15 +335,33 @@ def remove_empty_parents(
...
+        last_run_num = int(last_run_num)
+    else:
+        # If the ``runN`` symlink has been removed, make a sensible guess at
+        # what the next numbered run should be.
+        last_run_num = 0
+        for rundir in Path(run_path).glob('run[0-9]*'):
+            highest_yet = int(str(rundir).split('run')[-1])
```
MetRonnie, line 360: I think it would make sense to re-use the logic above. Regex ought to be `run(\d+)$` (we want to match at least 1 digit)

MetRonnie:
Ah I see, you're not really guessing, you're taking the max number and adding 1 to it

tests/unit/test_pathutil.py
```diff
@@ -491,3 +493,23 @@ def test_parse_rm_dirs__bad(dirs: List[str], err_msg: str):
...
         parse_rm_dirs(dirs)
     assert err_msg in str(exc.value)
+
+
+@pytest.mark.parametrize(
+    'expect, files, runN',
+    [
+        param(1, [], False, id='1st run (from filenames)'),
+        param(2, ['run1'], False, id='2nd run (from filenames)'),
+        param(1000, ['run999'], False, id='1000th run (from filenames)'),
```
MetRonnie, line 503: These two are the same test

cylc/flow/pathutil.py
```diff
@@ -335,15 +335,33 @@ def remove_empty_parents(
             break
 
 
-def get_next_rundir_number(run_path):
-    """Return the new run number"""
-    run_n_path = os.path.expanduser(os.path.join(run_path, "runN"))
-    try:
+def get_next_rundir_number(run_path: Union[str, Path]) -> int:
+    """Return the new run number
```
MetRonnie, line 339:
```suggestion
    """Return the next run number for a new install.
```

tests/unit/test_pathutil.py
wxtim: Yes, and that's a mistake. Will fix.

cylc/flow/pathutil.py
wxtim: Agree.

tests/unit/test_pathutil.py
wxtim: Also realized that I should include test case: There is a file not in the form runX in the directory.

cylc/flow/pathutil.py
```diff
@@ -335,15 +335,37 @@ def remove_empty_parents(
...
         old_run_path = os.readlink(run_n_path)
-        last_run_num = re.search(r'(?:run)(\d*$)', old_run_path).group(1)
-        return int(last_run_num) + 1
-    except OSError:
-        return 1
+        # Line below could in theory not return a match group, so mypy objects.
+        # This function unlikely to be called in circumstances where this will
+        # be a problem.
+        last_run_num = re.search(  # type: ignore[union-attr]
+            r'(?:run)(\d*$)', old_run_path
```
MetRonnie, line 353:
```suggestion
            r'run(\d+)$', old_run_path
```

cylc/flow/pathutil.py
```diff
@@ -335,15 +335,37 @@ def remove_empty_parents(
...
+        # If the ``runN`` symlink has been removed, get next numbered run from
+        # file names:
+        last_run_num = 0
+        for rundir in Path(run_path).glob('run[0-9]*'):
+            highest_yet = re.search(  # type: ignore[union-attr]
+                r'(?:run)(\d*$)', str(rundir.name)
+            ).group(1)
+            highest_yet = int(highest_yet)
+            if highest_yet > last_run_num:
+                last_run_num = highest_yet
```
MetRonnie, line 366: Could do a comprehension like
```python
last_run_num = max(
    (
        m.group(1) for m in (
            re.match(r'run(\d+)$', i.name) for i in os.scandir()
        ) if m
    ),
    default=0
)
```

wxtim: This would probably be quicker, but IMO to read: Stet?

cylc/flow/pathutil.py
```diff
@@ -335,15 +335,37 @@ def remove_empty_parents(
...
+            r'(?:run)(\d*$)', old_run_path
+        ).group(1)
+        last_run_num = int(last_run_num)
+    else:
+        # If the ``runN`` symlink has been removed, get next numbered run from
+        # file names:
+        last_run_num = 0
+        for rundir in Path(run_path).glob('run[0-9]*'):
+            highest_yet = re.search(  # type: ignore[union-attr]
+                r'(?:run)(\d*$)', str(rundir.name)
```
wxtim, line 362:
```suggestion
                r'^run(\d*)$', str(rundir.name)
```

cylc/flow/pathutil.py
MetRonnie: Efficiency concerns aside, the main benefit I see is avoiding the type error in the case of the regex not matching (because the glob `run[0-9]*` can match `run2B`).

cylc/flow/pathutil.py
```diff
@@ -335,15 +335,36 @@ def remove_empty_parents(
...
+        # This function unlikely to be called in circumstances where this will
+        # be a problem.
+        last_run_num = re.search(  # type: ignore[union-attr]
+            r'run(\d+)$', old_run_path
+        ).group(1)
+        last_run_num = int(last_run_num)
+    else:
+        # If the ``runN`` symlink has been removed, get next numbered run from
+        # file names:
+        paths = Path(run_path).glob('run[0-9]*')
```
MetRonnie, line 359: Btw, `os.scandir()` returns a generator too, so 
```suggestion
        paths = os.scandir()
```
would do the job

cylc/flow/pathutil.py
```diff
@@ -335,15 +335,36 @@ def remove_empty_parents(
...
+        run_numbers = (
+            int(m.group(1)) for m in (
+                re.match(r'^run(\d+)$', i.name) for i in paths
```
MetRonnie, line 362: Might as well define the regex string separately (or compile it) as it's used for both

cylc/flow/pathutil.py
wxtim: The glob matches run2B, but surely this just acts as a pre-filter for the regex?

MetRonnie (approved):
Thanks Tim (sorry for the barrage of review comments!)

wxtim:
> Thanks Tim (sorry for the barrage of review comments!)

Hopefully produces better code at the end.

PR #4378, opened 2021-08-24
GH Actions: Update action

MetRonnie:


MetRonnie:
Coverage failure due to only running fast tests. Hopefully #4379 should stop that in future

PR #4380, opened 2021-08-25
Add selection to global.cylc[platforms][localhost]

wxtim:
Prevents documentation build failure.

**Requirements check-list**
- [ ] Applied any dependency changes to both `setup.py` and
  `conda-environment.yml`.
- [ ] Does not need tests (Tested by documentation build in cylc-doc).

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -770,7 +770,7 @@ with Conf('global.cylc', desc='''
                 systems so for safety there is an upper limit on the number
                 of job submissions which can be batched together.
             ''')
-            with Conf('selection'):
+            with Conf('selection') as Selection:
                 Conf('method', VDR.V_STRING, default='random',
                      options=['random', 'definition order'],
                      desc='''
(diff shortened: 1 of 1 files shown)


cylc/flow/cfgspec/globalcfg.py
```diff
@@ -785,8 +785,12 @@
                       be appropriate if following the pattern
                       ``hosts = main, backup, failsafe``.
                 ''')
-        with Conf('localhost', meta=Platform):
+        with Conf('localhost', meta=Platform, desc='''
+            A default platform defining settings for the local computer.
```
datamel, line 789: Purely a suggestion and not a blocker to merge (equally happy for you to ignore or correct me!)...
Wondering if this description could be an opportunity to explain to users that, if they use cylc to submit to a cylc server, the server is the localhost, since certainly here at the MO, it is potentially a bit confusing grasping the idea of the server being the localhost, rather than the machine they are on.
Perhaps an extra sentence after this one to explain?

datamel (approved):
I have checked out the branch and rebuilt the docs. This fixes the error I was seeing.
Thanks @wxtim!

PR #4387, opened 2021-09-02
GH Actions: Do not fail if problem in remote coverage file

MetRonnie:

A lot of the time when `_remote_background_indep_poll` fails it's due to an error in the "Fetch remote coverage" step

> Error: no such table: file

This PR aims to skip such failures and just not upload that particular bit of coverage (there are like 100 remote coverage files, so skipping 1 or 2 shouldn't matter too much)

**Requirements check-list**
- [x] Does not need tests 

### DIFF
--- a/.github/workflows/test_functional.yml
+++ b/.github/workflows/test_functional.yml
@@ -258,12 +258,17 @@ jobs:
             --exclude='*' \
             >rsyncout
           cat rsyncout
+
+      - name: Fiddle remote coverage paths
+        if: env.REMOTE_PLATFORM == 'true'
+        run: |
           # fiddle the python source location to match the local system
           for db in $(grep --color=never '.coverage\.' rsyncout); do
             sqlite3 "$db" "
               UPDATE file
               SET path = REPLACE(path, '/cylc/cylc/', '$PWD/cylc/')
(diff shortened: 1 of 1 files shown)


.github/workflows/test_functional.yml
```diff
@@ -258,12 +258,17 @@ jobs:
...
+        if: env.REMOTE_PLATFORM == 'true'
+        run: |
           # fiddle the python source location to match the local system
           for db in $(grep --color=never '.coverage\.' rsyncout); do
             sqlite3 "$db" "
               UPDATE file
               SET path = REPLACE(path, '/cylc/cylc/', '$PWD/cylc/')
-            "
+            " || \
+            (echo "::warning::${db}" && sqlite3 "$db" ".schema" && rm "$db")
```
oliver-sanders, line 271: Just a thought, if sqlite3 commands are occasionally failing in the for loop, it might be the for loop that's the problem.

Perhaps if we `.join(':')` the SQL statements and issue them as one command we wouldn't have to ignore failures?

MetRonnie:
I'm closing this for now, as this doesn't seem to be the main source of flakiness anymore

PR #4395, opened 2021-09-08
ensure workflow tries to kill orphan tasks before shutting down

wxtim:
These changes close #4393 

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.py` and
  `conda-environment.yml`.
- [x] Appropriate tests are included (unit and/or functional).
- [x] Appropriate change log entry included.
- [x] No documentation update required.

PR #4401, opened 2021-09-09
tests/f: fix shutdown/22

oliver-sanders:
Remove non-POSIX compliant grep.

See https://github.com/cylc/cylc-flow/pull/4395#issuecomment-916012828

### DIFF
--- a/tests/functional/shutdown/22-stop-now.t
+++ b/tests/functional/shutdown/22-stop-now.t
@@ -19,7 +19,7 @@
 
 . "$(dirname "$0")/test_header"
 
-set_test_number 3
+set_test_number 4
 
 init_workflow "${TEST_NAME_BASE}" <<'__FLOW_CONFIG__'
 [scheduling]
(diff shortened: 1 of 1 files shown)


tests/functional/shutdown/22-stop-now.t
```diff
@@ -42,6 +42,8 @@ run_ok "${TEST_NAME_BASE}" cylc play "${WORKFLOW_NAME}" --no-detach --debug
 
 WORKFLOW_LOG="${WORKFLOW_RUN_DIR}/log/workflow/log"
 
-grep_ok 'Orphaned task jobs.*\n.*foo.1' "${WORKFLOW_LOG}" -Pz
+log_scan "${TEST_NAME_BASE}-orphaned" "${WORKFLOW_LOG}" 1 1 \
+    'Orphaned task jobs.*' \
+    '.*foo.1'
```
oliver-sanders, line 47: This ensures that the "Orphaned" message appears on a line *before* the "foo.1" message.

As there is only one task in the workflow presence of the "Orphaned" message is more or less good enough but good to check the `foo.1` anyhoo.

wxtim (approved):

PR #4404, opened 2021-09-10
graph line continuation on boolean

### LINKED ISSUE #4390: Allow graph line splitting on AND and OR?
Ref: https://cylc.discourse.group/t/cylc-8b2-graphparseerror/374

Currently we only join on `=>`.

I can't think of any reason not to allow this. It's simple line joining so we don't need to worry about operators embedded in complex conditional expressions.

wxtim:
These changes close #4390 

**Requirements check-list**
- [x] Additionally refactored unittest test to pytest.
- [x] (master branch) I have opened a documentation PR at https://github.com/cylc/cylc-doc/pull/296.

I have left commits separate to make changes to the test easier to understand. If merging please squash.


cylc/flow/graph_parser.py
```diff
@@ -89,6 +89,8 @@ class GraphParser:
     FAM_TRIG_EXT_ANY = '-any'
     LEN_FAM_TRIG_EXT_ALL = len(FAM_TRIG_EXT_ALL)
     LEN_FAM_TRIG_EXT_ANY = len(FAM_TRIG_EXT_ANY)
+    CONTINUATION_STRS = [ARROW, OP_AND, OP_OR]
+    BAD_STRS = [OP_AND_ERR, OP_OR_ERR]
```
kinow, line 93: CI is happy, docs look good. Great job (and really quick) @wxtim . These bad strings are the only bit I didn't understand. I thought parsec would raise an error when finding these bad strings in the next line (or other invalid values like => after a boolean). Thanks!

hjoliver: These represent bad content inside the graph strings. Parsec is (mostly) a generic `item = value` config parser, so to it the entire graph string is just a string.

cylc/flow/graph_parser.py
```diff
@@ -229,7 +231,13 @@ def parse_graph(self, graph_string):
...
                         "trailing arrow: %s" % this_line)
             part_lines.append(this_line)
-            if (this_line.endswith(ARROW) or next_line.startswith(ARROW)):
+            if (any(
+                this_line.endswith(seq) or next_line.startswith(seq) for
+                seq in self.CONTINUATION_STRS
+            ) and not (any(
+                this_line.endswith(seq) or next_line.startswith(seq) for
+                seq in self.BAD_STRS
+            ))):
```
hjoliver, line 240: Not your fault @wxtim as you've just extended the existing conditional ... but it would be good to catch the error here if one line ends with `seq` *and* the next line starts with `seq`.  Current this causes different (and somewhat hard to understand) downstream validation errors.

Also, this (correctly) fails validation with `error: trailing arrow`:
```
foo =>
# comment
```
 But this (incorrectly) passes, and results in a runtime error:
```
foo &
# comment
```

hjoliver (changes requested):
(The one test that's failing should be an easy fix: tests/f/validate/62-null-task-name.t)

hjoliver:
(Oops, didn't notice this was still a draft PR, so maybe you handn't finished :grimacing: )

wxtim:
Only because that test broken and I'm holiday

cylc/flow/graph_parser.py
wxtim: I think that I have fixed both of these issues.

tests/unit/test_graph_parser.py
```diff
@@ -561,6 +597,38 @@ def test_param_expand_graph_parser():
...
+    """
+    parser = GraphParser()
+    with pytest.raises(GraphParseError) as raised:
+        parser.parse_graph(f'foo & bar {before}\n{after}baz')
+    assert isinstance(raised.value, GraphParseError)
+    assert 'Consecutive lines end and start' in raised.value.args[0]
+
+
+if __name__ == "__main__":
+    unittest.main()
```
hjoliver, line 629:
```suggestion
```

cylc/flow/graph_parser.py
```diff
@@ -328,16 +331,39 @@ def parse_graph(self, graph_string: str) -> None:
...
                 next_line = ''
-                if this_line.endswith(self.__class__.ARROW):
-                    # Last line can't end with an arrow.
-                    raise GraphParseError(
-                        f"Trailing arrow: {this_line}")
+                for seq in self.CONTINUATION_STRS:
+                    if this_line.endswith(seq):
+                        # Last line can't end with an arrow, & or |.
+                        raise GraphParseError(
+                            f"Trailing continuation sequence ({seq}):"
```
hjoliver, line 338:
```suggestion
                            f"Dangling ({seq}):"
```
"Trailing" is OK, so long as there's a subsequent line that fits.

Also I'm don't think `sequence` is a great name for these things in the user-facing validation message. Probably sufficient to just print what's dangling without naming it.

hjoliver (changes requested):
Still not getting consistent treatment of the three continuation items.

```
# OK:
 a =>  # GraphParseError: Trailing continuation sequence (=>):a=>

# OK, but use consistent terminology)
 => a  # GraphParseError: Leading arrow: =>a

# OK:
 a &  # GraphParseError: Trailing continuation sequence (&):a&
 a |  # GraphParseError: Trailing continuation sequence (|):a|

# Not OK:
 & a  # valid
 | a  # valid

# OK:
 => a => b  # GraphParseError: Leading arrow: =>a=>b

# Not OK:
 & a => b  # GraphParseError: Null task name in graph: &a => b
 | a => b  # WorkflowConfigError: ERROR: bad trigger
```
  | a  # ...
  
```

**Comment by wxtim** [comment]
> Still not getting consistent treatment of the three continuation items.

I there was 1 remaining case (leading continuation symbol) where the code was checking for `=>` rather than the whole list. I think that I've fixed this now.

File: tests/unit/test_graph_parser.py [hunk 5]
```diff
@@ -49,17 +50,42 @@ def test_parse_graph_fails_null_task_name(graph):
         assert "Null task name in graph:" in str(cm.value)
 
 
+@pytest.mark.parametrize('seq', ('&', '|', '=>'))
```
wxtim, line 53: Factored out of the original test to allow this double parametrization.

hjoliver (changes requested):
All working well now :+1: Just one code simplification suggestion.

cylc/flow/graph_parser.py
```diff
@@ -321,23 +324,47 @@ def parse_graph(self, graph_string: str) -> None:
...
+                if (
+                    this_line.endswith(sequences[0])
+                    and next_line.startswith(sequences[1])
+                ):
+                    raise GraphParseError(
+                        'Consecutive lines end and start with continuation '
+                        'sequence:\n'
+                        f'{this_line}\n'
+                        f'{next_line}'
+                    )
```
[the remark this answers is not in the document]
wxtim: Yep, follow up changes very nice.

cylc/flow/graph_parser.py
```diff
@@ -321,23 +323,46 @@ def parse_graph(self, graph_string: str) -> None:
...
+            # begin the next:
             if (
-                this_line.endswith(self.__class__.ARROW) or
-                next_line.startswith(self.__class__.ARROW)
+                this_line.endswith(self.CONTINUATION_STRS) and
+                next_line.startswith(self.CONTINUATION_STRS)
             ):
+                raise GraphParseError(
+                    'Consecutive lines end and start with continuation '
+                    'characters:\n'  # (not "sequence")
```
hjoliver, line 351:
```suggestion
                    'characters:\n'
```
(sorry, my comment was only aimed at you :-)

cylc/flow/graph_parser.py
```diff
@@ -321,23 +323,46 @@ def parse_graph(self, graph_string: str) -> None:
...
+                    f'{this_line}\n'
+                    f'{next_line}'
+                )
+
+            # Check that line ends with a valid continuation sequence:
+            if (any(
+                this_line.endswith(seq) or next_line.startswith(seq) for
```
hjoliver, line 358: Use `endswith(tuple)` here too (and elsewhere in the file?) instead of iterating?

wxtim: COFFEE ERROR: Tim has not had coffee yet.

wxtim: On inspection I've decided to leave 
```
            for seq in self.CONTINUATION_STRS:
                if i == 0 and this_line.startswith(seq):
                    # First line can't start with an arrow.
                    raise GraphParseError(f"Leading {seq}: {this_line}")
```
Alone, because I think that providing the offending leading/dangling sequence to users is a worthwhile good. If you can see a nice way of doing it let me know. I tried seq = ternery expression, but this ended up being 5 lines long to fit our line length.

hjoliver: Yep, no worries 👍🏻

hjoliver (approved):
Thanks @wxtim all good :tada:

One review will do for this because: "time is of the essense" right now, plust it is somewhat peripheral functionality, and I've tested the hell out of it.

PR #4386, opened 2021-09-01
Fix mutation error when stopping workflow in UI

MetRonnie:
This is a small change with no associated Issue. Fixes bug probably introduced by #4331.

#### The bug

Stopping a workflow using the default stop mode in the UI (e.g. pressing the stop button) was resulting in an error

> command failed: stop - Error: Response not successful: Received status code 400 

image

Inspecting the sent mutation in the devtools console showed that the mode was 
```
mode: "REQUEST(CLEAN)"
```
However, if you clicked on the hamburger and clicked the pencil icon to edit the Stop command, and actively chose "Clean" from the mode dropdown, then it would work fine. The mutation in that case contained
```
mode: "Clean"
```
instead.

#### The fix

In `cylc.flow.network.schema.WorkflowStopMode`, remove the `.value` from `StopMode.REQUEST_*.value`.

I'm not really sure why this works, `graphene.Enum` is confusing and not well documented.

#### Testing it out

Run a workflow in no-detach mode and try stopping it using
- `cylc stop`
- `cylc stop --kill` (or `--now`)
- The web UI using the stop button
- The web UI using the hamburger button > stop (edit) > mode = Clean
- The TUI

#### Requirements check-list
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.py` and
  `conda-environment.yml`.
- [ ] Appropriate tests are included (unit and/or functional).
- [x] No change log entry needed (fixes recently introduced change).
- [x] No documentation update required.

PR #4409, opened 2021-09-14
Access an enum's enum's value

### LINKED ISSUE #4408: graphql.error.located_error.GraphQLLocatedError
I am getting `GraphQLLocatedError`'s in the UIS log even when no workflows are present and no mutations are called:

```
graphql.error.located_error.GraphQLLocatedError
```

See also cylc/cylc-uiserver#242 which may be related.

This is likely a cylc-flow issue:

```
cylc/flow/network/graphql.py", line 426, in async_resolve
```

kinow:
These changes close #4408

Found the bug I think. I was (am) having some issues UI & UIS while writing some new code, and the tracebacks were getting in the way and making it harder to debug :angry: 

So I i) isolated the query that quickly reproduces the issue:

```graphql
query {
  __schema {
    __typename
    mutationType {
      fields {
        args {
          defaultValue # <-- this causes the error, use name instead of another field and there should be no error
        }
      }
    }
  }
}
```

Then ii) looked at the traceback, and attached Pycharm's remote debugger to the `cylc hubapp` process/PID. iii) fired the query, and iv) inspected the context of the breakpoint.

Screenshot_2021-09-14_15-20-40

That value highlighted is an **enum**. The issue appears to be that the GraphQL Core has a bunch of if/else statements filtering the possible types of the `value` passed. When it finds the `Enum` it raises an error as the last option is to have `value is a dict`.

Looking at the schema, looks like `WorkflowStopMode` is an `enum`. And the values of this `enum` are of type `Enum`. i.e. it's an Enum of Enums. Which means that `WorkflowStopMode.Clean` is the `WorkflowStopMode` enum, and `WorkflowStopMode.Clean.value` is the `enum`'s enum` value :dizzy: :dizzy_face: 

You can test it this way:

```py
>>> from cylc.flow.network.schema import WorkflowStopMode
>>> WorkflowStopMode.Clean
<EnumMeta.Clean: <StopMode.REQUEST_CLEAN: 'REQUEST(CLEAN)'>>
>>> WorkflowStopMode.Clean.value
<StopMode.REQUEST_CLEAN: 'REQUEST(CLEAN)'>
>>> WorkflowStopMode.Clean.value.value
'REQUEST(CLEAN)'
>>> 
```

As in this case the `defaultValue` that we want to specify in the GraphQL schema is a string `REQUEST(CLEAN)`, we need to use `WorkflowStopMode.Clean.value.value` :+1: 

@dwsutherland is there an easy way to unit/functional test it?

Thanks!

**Requirements check-list**
- [ ] Applied any dependency changes to both `setup.py` and
  `conda-environment.yml`.


dwsutherland:
> @dwsutherland is there an easy way to unit/functional test it?

Yeah, just add as graphql functional tests..
Or instantiate the schema with a dummy resolver (or in integration test).. and execute the schema directly.

oliver-sanders:
Dammit, this enum has caused us a long string of issues.

Assigning @MetRonnie to review as he recently fixed some other issues with this enum.

BTW, so long as you don't need to test graph/script execution (shouldn't need to) can test GraphQL using the integration test battery:

https://github.com/cylc/cylc-flow/blob/a873629faa1f6a359deb58945d8a43dd7b215298/tests/integration/test_client.py#L34-L45

MetRonnie:
This has been confusing me. 

What happens if you instead go back to using `StopMode.*.value` here?

https://github.com/cylc/cylc-flow/blob/a873629faa1f6a359deb58945d8a43dd7b215298/cylc/flow/network/schema.py#L1444-L1454

Originally I had removed the `.value` from them in #4386, but I can no longer reproduce the bug when I put the `.value`s back in! Can you reproduce the bug from #4386 with the `.value`s back in (and only having `.value` once on the `default_value`)?

N.B. One point that might help avoid confusion:
- `cylc.flow.workflow_status.StopMode` is an `enum.Enum`
- `cylc.flow.network.schema.WorkflowStopMode` is a `graphene.Enum`

kinow:
@MetRonnie thanks a lot for the suggestions, and the links. I had a look at the previous change that removed the `.value`.

I've added it back now, so that the Graphene enum (hadn't noticed it wasn't a Python enum :dizzy_face: ) has values that are strings.

After doing that, I went back to the original issue you tried to fix. The Stop mutation stated throwing errors. After thinking for a while, I decided to do the laziest thing: search other enums and see why they were working (or at least nobody complained about those yet?)

I had a look at the Broadcast mutation. Its args has an enum, with a default value. **But**, the enum's default value is using **an enum name**, not the value. And there's also a link above in a comment to this issue: https://github.com/graphql-python/graphql-core-legacy/issues/166

Changed to `.name`, and now the stop mutation works like a charm, no tracebacks or errors anywhere.

Does this sound like a possible solution to both this current GraphQLError exception, and also the default mode of the Stop mutation?

Thanks!

kinow:
@oliver-sanders I started writing an integration test. I thought this mutation could be used for reproducing the error:

```
mutation {
  stop (workflows: ["user|cylc"]) {
    result
  }
}
```

However, looks like the issue is only triggered if `user|cylc` workflow is running? At the moment the query appears to generate a few errors when used against nonexistent workflows, but not directly related to this issue (https://github.com/cylc/cylc-uiserver/issues/250).

So in this case I think we will need a running workflow, and use the client to submit a migration?

MetRonnie:
> So I simply removed the `.value` or `.name` from the default value. Now everythings works ™

I think that's it! I have tested this and it does indeed work™

(And actually, I can still reproduce the bug from #4386 if I go back to using `.value`s on the `WorkflowStopMode` Enum alone, it's the removal of the `.value` from `default_value` that sorts it out)

While we are at it, I think that means we ought to update these type annotations
```diff
diff --git a/cylc/flow/network/resolvers.py b/cylc/flow/network/resolvers.py
index 31990654e..4b96cf81e 100644
--- a/cylc/flow/network/resolvers.py
+++ b/cylc/flow/network/resolvers.py
@@ -21,7 +21,7 @@ from fnmatch import fnmatchcase
 import logging
 import queue
 from time import time
-from typing import Iterable, Optional, Tuple, TYPE_CHECKING
+from typing import Iterable, Optional, Tuple, TYPE_CHECKING, Union
 from uuid import uuid4
 
 from graphene.utils.str_converters import to_snake_case
@@ -735,7 +735,7 @@ class Resolvers(BaseResolvers):
 
     def stop(
         self,
-        mode: 'StopMode',
+        mode: Union[str, 'StopMode'],
         cycle_point: Optional[str] = None,
         clock_time: Optional[str] = None,
         task: Optional[str] = None,
diff --git a/cylc/flow/scheduler.py b/cylc/flow/scheduler.py
index 9ecd1f13a..08bd5aa74 100644
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -30,7 +30,7 @@ import sys
 from threading import Barrier
 from time import sleep, time
 import traceback
-from typing import Iterable, NoReturn, Optional, List, Set, Dict
+from typing import Iterable, NoReturn, Optional, List, Set, Dict, Union
 from uuid import uuid4
 import zmq
 from zmq.auth.thread import ThreadAuthenticator
@@ -831,7 +831,7 @@ class Scheduler:
 
     def command_stop(
         self,
-        mode: 'StopMode',
+        mode: Union[str, 'StopMode'],
         cycle_point: Optional[str] = None,
         # NOTE clock_time YYYY/MM/DD-HH:mm back-compat removed
         clock_time: Optional[str] = None,
```
(I know we no longer expect these methods to be called with `StopMode` instances anymore, but it's still possible in tests for example, and it helps indicate what the acceptable values of the string should be)

kinow:
>While we are at it, I think that means we ought to update these type annotations

+1 ! Would you like to write one of those `suggest` blocks or push a commit here? Otherwise I can update it in some minutes.

Thanks @MetRonnie !

MetRonnie:
Hmm, just been digging into this a bit more and it's getting confusing again 😢

If using `default_value=WorkflowStopMode.Clean`, then when running the mutation it shows in the log as 
```
INFO - Command succeeded: stop(mode=EnumMeta.Clean)
```

Also if I remove the `.name`s from the lines Mypy is complaining about (e.g. `cylc/flow/scripts/stop.py:186: error: "str" has no attribute "name"`), then `cylc stop --kill` breaks with traceback.

Are you sure `default_value=WorkflowStopMode.Clean.name` doesn't work? It does seem to be working for me. And in the log it shows as
```
INFO - Command succeeded: stop(mode=REQUEST(CLEAN))
```
For clarity:
```diff
diff --git a/cylc/flow/network/schema.py b/cylc/flow/network/schema.py
index 7d3662cfe..c4df42b25 100644
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -1448,10 +1448,10 @@ class WorkflowStopMode(Enum):
     # * Graphene requires special enums.
     # * We only want to offer a subset of stop modes (REQUEST_* only).
 
-    Clean = StopMode.REQUEST_CLEAN
-    Kill = StopMode.REQUEST_KILL
-    Now = StopMode.REQUEST_NOW
-    NowNow = StopMode.REQUEST_NOW_NOW
+    Clean = StopMode.REQUEST_CLEAN.value
+    Kill = StopMode.REQUEST_KILL.value
+    Now = StopMode.REQUEST_NOW.value
+    NowNow = StopMode.REQUEST_NOW_NOW.value
 
     @property
     def description(self):
@@ -1725,7 +1725,7 @@ class Stop(Mutation):
     class Arguments:
         workflows = List(WorkflowID, required=True)
         mode = WorkflowStopMode(
-            default_value=WorkflowStopMode.Clean.value
+            default_value=WorkflowStopMode.Clean.name
         )
         cycle_point = CyclePoint(
             description='Stop after the workflow reaches this cycle.'
```
N.B. I don't think type annotations are implemented correctly for `graphene.Enum`, so we should just stick `# type: ignore[attr-defined]` after each use of `WorkflowStopMode.*.name`

kinow:
Logged off from my notebook already, but I think I tried the value/name. Though not sure if at the same time... will try it again tomorrow morning.

The way I tested was by changing, restarting, then play, pause, and stop a workflow from the UI.👍

MetRonnie:
If I try again tomorrow I'll probably find `.name` doesn't work 😅

kinow:
> If I try again tomorrow I'll probably find `.name` doesn't work sweat_smile

It worked for me! :sweat_smile: let me fix the type errors now.

kinow:
@MetRonnie 

>N.B. I don't think type annotations are implemented correctly for graphene.Enum, so we should just stick # type: ignore[attr-defined] after each use of WorkflowStopMode.*.name

I had a look at the debugger, and even though we specify a string as the value, I think internally (with some metaclass-fu?) the type of `WorkflowStopMode.Clean` is defined as an `Enum` (an enum of an enum :dizzy: ). So I've defined it as `# type: Enum`.

It appeared to satisfy mypy. WDYT?

cylc/flow/network/schema.py
```diff
@@ -1725,7 +1725,7 @@ class Meta:
     class Arguments:
         workflows = List(WorkflowID, required=True)
         mode = WorkflowStopMode(
-            default_value=WorkflowStopMode.Clean.value
+            default_value=WorkflowStopMode.Clean.name
```
MetRonnie, line 1728: Argh, with this the UI stop button works (also when you use the edit button, "Clean" is pre-populated in the mode dropdown), but `cylc stop foo` results in
```
WARNING - Invalid stop mode: 'Clean'
WARNING - Command failed: stop(mode=Clean)
```
But with `default_value=WorkflowStopMode.Clean.value`, `cylc stop foo` works whereas the stop button in the UI has the 400 error!

MetRonnie:
So I've been doing some debugging (with this branch checked out, i.e. using `.name` for the `default_value`) and found this: if you put a breakpoint in https://github.com/cylc/cylc-flow/blob/9b6930e8ef3529c582b4e231db5ba9163af95dab/cylc/flow/network/schema.py#L1262-L1265

then you find that:
- for the stop button in the UI, the value of `args` is `{'mode': 'REQUEST(CLEAN)'}` (works)
- for `cylc stop`, the value of `args` is `{'mode': 'Clean'}` (doesn't work)

I don't know where the value of `args` is set, I couldn't find anything out by stepping out - the frames above seem to be generic graphql and asyncio event loop stuff.

___

Also if you use the GraphiQL with the stop mutation with a `stopMode` of `null` in the query variables, i.e. 
```
{
  "wFlows": ["foo/run1"],
  "stopMode": null,
  "cyclePoint": null,
  "clockTime": null,
  "task": null,
  "flowLabel": null
}
```
then the result is identical to `cylc stop`.

kinow:
>for the stop button in the UI, the value of args is {'mode': 'REQUEST(CLEAN)'} (works)

Yup, I tested Kill and Clean, both using the Stop button and also expanding and Editing the Stop mutation. In every case the workflow was stopped. The parameter passed from the UI, in the HTTP Post to trigger the mutation in UIS appears to be the value of the enum, so "Clean", not "REQUEST(CLEAN)".

image

The command `cylc stop five` is indeed failing, and I set the breakpoint in the `mutator` function in the `schema` module, but got similar scenario. The mode is set to `Clean`. But debugging the stop command, it looks like the value is `mode: Null`. Not sure where the `Clean` is coming from :monocle_face: 

image

kinow:
Actually, I think if you go back a few frames, we have the schema with the default value of `'Clean'`.

image

So perhaps somewhere while the `stop` command's GraphQL message is being serialized, the value `'Clean'` is being attached to the mode. Which I think it what the schema defines. However, the `scheduler.py` is expecting a stop mode, like `REQUEST(CLEAN)`...

kinow:
I tested again removing the `.name`, and using `default_value=WorkflowStopMode.Clean`. The UI and CLI appear to work, but if you go to the mutation form to edit the mutation, the default value (Clean) is not pre-selected. And if you click the stop button, the mode is now null. So I think everything is working simply because we are not attaching any value to the `mode`. :dizzy: :dizzy_face:

MetRonnie:
Looks like we can solve this with a way to map from `'Clean'` back to `'REQUEST(CLEAN)'`!

I have opened a PR against this branch

kinow (commented):
@MetRonnie synced the code (using master for UI and UIS) then with `five` running:

## cli

- `cylc stop five` :heavy_check_mark: 
- `cylc stop five --now` :heavy_check_mark: 

## graphiql

- Stopped with default mode :heavy_check_mark: 

```graphql
mutation {
  stop (workflows: ["kinow|five"]) {
    result
  }
}
```
- Stopped specifying the mode :heavy_check_mark: 

```graphql
mutation {
  stop (workflows: ["kinow|five"], mode: NowNow) {
    result
  }
}
```

## UI

At `http://localhost:8000/user/kinow/cylc/#/workflows/five`:

- Stopped using the big Stop button from the toolbar :heavy_check_mark: 
- Stopped using the mutation menu, without modifying anything in the form (confirmed the `Clean` mode is selected by default :+1: nice! :heavy_check_mark: 

image

- Stopped using the mutation menu, choosing `Kill` mode :heavy_check_mark: 

I think you've nailed it @MetRonnie . Found no errors testing it. A few warnings about orphan tasks when killing with nownow or kill, as it's expected. Thank **you** for solving the enums mystery!

MetRonnie:
I thought about doing that in `workflow_status` but it seemed to me that there is a distinction between automatic stops and user-requested stops, so I didn't want to mess with that

https://github.com/cylc/cylc-flow/blob/5e6ebe3cf1280e29fdc6b744d49271bb8f1cfec6/cylc/flow/workflow_status.py#L59-L78

PR #4411, opened 2021-09-14
Add colour to GH Actions pytest log

MetRonnie:
Also use [`PYTEST_ADDOPTS`](https://docs.pytest.org/en/stable/reference.html#envvar-PYTEST_ADDOPTS) to avoid repetition


hjoliver (approved):
Nice. One review will do.

PR #4425, opened 2021-09-22
Remove Conda Activate from Wrapper Scipt

### LINKED ISSUE #3704: Investigate wrappers for environment activation

wxtim:
These changes close #3704 

**Requirements check-list**

  `conda-environment.yml`.

Tagged @dpmatthews for wrapper script change review.

Tagged @oliver-sanders for review of the test.


.github/workflows/conda-build-test.yml
```diff
@@ -0,0 +1,29 @@
...
+        run: |
+            conda env create \
+              -f conda-environment.yml
+
+      - name: check cylc-version
+        run: |
+          find /usr/share/miniconda/envs/cylc-dev/. -name "activate.d" | tee > activates.txt
+          if [[ $(cat activates.txt | wc -l) -ne 0 ]]; then
+              exit 1
+          fi
```
hjoliver, line 29: No newline at end of file.

usr/bin/cylc
```diff
@@ -191,7 +191,7 @@ elif [[ -d "${CYLC_HOME}/conda-meta" && \
         -f "${CYLC_HOME%/*/*}/etc/profile.d/conda.sh" ]]; then
     # A normal Conda environment
     . "${CYLC_HOME%/*/*}/etc/profile.d/conda.sh"
-    conda activate "${CYLC_HOME##*/}" || exit 1
+    # conda activate "${CYLC_HOME##*/}" || exit 1
```
hjoliver, line 194: Huh, simple as that, is it?!

wxtim: As far as I can tell - I was confused by it, so I asked @oliver-sanders  to look over my shoulder...

wxtim:
```suggestion
```

wxtim:
```suggestion
```

.github/workflows/conda-build-test.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions/checkout@v2
+
+      - name: build conda env
+        run: |
+            conda env create \
+              -f conda-environment.yml
+
+      - name: check cylc-version
+        run: |
+          find /usr/share/miniconda/envs/cylc-dev/. -name "activate.d" | tee > activates.txt
```
wxtim, line 26: @oliver-sanders Do you know of any packages which use activate scripts?
I have struggled to web-search such a thing, and therefore to check this test against a control.

oliver-sanders: Here's one:

https://github.com/conda-forge/cupy-feedstock/blob/5f118fca132164c0d0699d508e45bdf18fba4f5a/recipe/activate.sh#L4-L6

Which is copied into the `activate.d` directory in the build:

https://github.com/conda-forge/cupy-feedstock/blob/5f118fca132164c0d0699d508e45bdf18fba4f5a/recipe/meta.yaml#L49-L52

usr/bin/cylc
oliver-sanders: Yep, truely that simple, just call the executable using it's absolute path, Python automatically inserts the `lib/python<v>/site-packages/` (relative the the Python executable location) into `sys.path`.

usr/bin/cylc
```diff
@@ -191,7 +191,6 @@ elif [[ -d "${CYLC_HOME}/conda-meta" && \
         -f "${CYLC_HOME%/*/*}/etc/profile.d/conda.sh" ]]; then
     # A normal Conda environment
     . "${CYLC_HOME%/*/*}/etc/profile.d/conda.sh"
-    conda activate "${CYLC_HOME##*/}" || exit 1
```
hjoliver, line 194: Can we have a comment in the file for future reference, on why we're not using `conda activate` and why that is OK?

hjoliver: Note there's an existing `conda activate` comment and variable above lines 180-4, which is presumably now redundant.

usr/bin/cylc
```diff
@@ -191,7 +191,6 @@ elif [[ -d "${CYLC_HOME}/conda-meta" && \
         -f "${CYLC_HOME%/*/*}/etc/profile.d/conda.sh" ]]; then
     # A normal Conda environment
     . "${CYLC_HOME%/*/*}/etc/profile.d/conda.sh"
```
hjoliver, line 193: Don't need this line either, I think.

hjoliver (changes requested):
One small mystery to solve.

If I comment the `conda activate` lines out of my wrapper, make sure my environment and `.bashrc` is clean of `CYLC_HOME` etc., and my Cylc conda environment is not activated ... then I do `export CYLC_ENV_NAME=<my-conda-env>` and try to run Cylc via the wrapper ... everything seems to work fine except for this:
```
$ cylc cat-log demo >/dev/null
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'cylc.flow'
```
It works fine if I leave `conda activate` in the wrapper.

And, without `conda activate` in the wrapper, this does work!:
```
$ cylc cat-log --color=never demo >/dev/null  # OK
```
The culprit is in `scripts/cat-log.py`:
```python
  139     if color:
  140         color_proc = Popen(
  141             [
  142                 'python3', '-c',
  143                 '; '.join([
  144                     'import sys',
  145                     'from cylc.flow.loggingutil import re_formatter',
  146                     'print(re_formatter(sys.stdin.read()), end="")'
  147                 ]),
  148             ],
  149             stdin=PIPE
  150         )
  151         color_proc.communicate(cat_proc.communicate()[0])
  152     else:
  153         cat_proc.wait()
```

hjoliver:
Finally, I think we can ditch the Python venv (as opposed to conda) activation too (aside from the same `cat-log` coloring problem above. ... in which case we can delete the whole environment type identification and activation part of the script.  Not sure about "conda pack" though - can you comment on that @dpmatthews ?

usr/bin/cylc
wxtim: Do we need to do the unset below it?

dpmatthews:
> Finally, I think we can ditch the Python venv (as opposed to conda) activation too (aside from the same `cat-log` coloring problem above. ... in which case we can delete the whole environment type identification and activation part of the script. Not sure about "conda pack" though - can you comment on that @dpmatthews ?

It should be no different - I agree it can all go.

usr/bin/cylc
dpmatthews: No - it was only needed due to the "conda activate"

wxtim:
@hjoliver  - RE: Cylc Cat Log 
I can't for the life of me reproduce this, but I think that [this change](https://github.com/cylc/cylc-flow/pull/4425/commits/62423634ffd149f6835ba4af6e70c28af1e398c3) should fix it.

.github/workflows/conda-build-test.yml
```diff
@@ -0,0 +1,40 @@
...
+  test_conda_install:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: checkout cylc-flow
+        uses: actions/checkout@v2
+
+      - name: build conda env
+        run: |
+            # The appended lines make sure that we actually build with Cylc
```
wxtim, line 21: @oliver-sanders 
This feels wierd.

Is it really necessary to check the `cylc version`, which makes all the crud here necessary?

dpmatthews:
2 problems I'm afraid.

1) We've forgotten about Rose. Rose is still using a bash wrapper which just calls `exec python3 ...`. So that still relies on the activation to set `$PATH` until we tackle metomi/rose#2444.

2) I was wrong about conda pack. It sets the shebang lines to `#!/usr/bin/env python3.8` so we'll need to continue calling `activate` (or perhaps just set `$PATH`?).

.github/workflows/conda-build-test.yml
oliver-sanders: > This feels wierd.

It's fine (I do the same thing in the script I use to build my dev environments)

> Is it really necessary to check the cylc version.

Er, didn't you add this bit?

The point is more about checking the conda packages than checking the conda environment so it's not really necessary.

dpmatthews:
For the moment we could just add the following:
```
PATH=${CYLC_HOME}/bin:${PATH}
```
It's not perfect but we'd still be better of than we were.

hjoliver:
> @hjoliver - RE: Cylc Cat Log
> I can't for the life of me reproduce this, but I think that [this change](https://github.com/cylc/cylc-flow/pull/4425/commits/62423634ffd149f6835ba4af6e70c28af1e398c3) should fix it.

Two things to check, to make this reproducible:
1. check that you don't have your cylc conda environment activated in the terminal where you run `cylc cat-log` (i.e., rely only on the wrapper to find `cylc`)
2. check that you don't have a default system python3 in your default `PATH` with `cylc.flow` installed in `~/.local/lib`

cylc/flow/scripts/cat_log.py
```diff
@@ -139,7 +139,7 @@ def colorise_cat_log(cat_proc, color=False):
     if color:
         color_proc = Popen(
             [
-                'python3', '-c',
+                sys.executable, '-c',
```
hjoliver, line 142: Yep, that works :+1:

usr/bin/cylc
```diff
@@ -177,25 +177,16 @@ if [[ -z "${CYLC_HOME}" ]]; then
...
     exit 1
 fi
-# Execute the command in the selected installation.
+# Execute Cylc from the selected installation.
+# n.b. If CYLC_HOME points to a cylc binary inside a
+# ``conda/envs/bin directory``this is sufficent without using
+# ``conda activate``. We avoid ``conda activate`` because:
+# - It's slow.
+# - Subsequent environment changes (conda activate or deactivate)
+#   are not straightforward.
```
wxtim, line 191:
```suggestion
# Add CYLC_HOME to the start of the path to ensure the right version of python is used.
PATH=${CYLC_HOME}/bin:${PATH}
```

usr/bin/cylc
```diff
@@ -177,25 +177,16 @@ if [[ -z "${CYLC_HOME}" ]]; then
...
-    . "${CYLC_HOME%/*/*}/etc/profile.d/conda.sh"
-    conda activate "${CYLC_HOME##*/}" || exit 1
-fi
+# Warn if executable not at $CYLC_HOME
 if [[ ! -x "${CYLC_HOME}/bin/${0##*/}" ]]; then
     echo 1>&2 "ERROR: ${0##*/} not found in ${CYLC_HOME}"
     exit 1
 fi
-# Execute the command in the selected installation.
+# Execute Cylc from the selected installation.
```
wxtim, line 185: @dpmatthews  How's this?

```suggestion
# Add CYLC_HOME to the start of the path to ensure the right version of python is used.
PATH=${CYLC_HOME}/bin:${PATH}

# Execute Cylc from the selected installation.
```

cylc/flow/scripts/cat_log.py
oliver-sanders: Nice, note to self to ensure we use `sys.executable` when performing Python subprocess calls.

usr/bin/cylc
wxtim: Yes vote?

oliver-sanders: It's a "further thought needed".

dpmatthews:
conda pack supports `--dest-prefix` which fixes the shebang lines. The downside is that the install path is then fixed so you might need to generate several different pack files if you install to different locations on different platforms (annoying but acceptable).

wxtim:
Now ready for another examination after the merger of https://github.com/metomi/rose/pull/2494

usr/bin/cylc
```diff
@@ -177,25 +177,19 @@ if [[ -z "${CYLC_HOME}" ]]; then
...
-    conda activate "${CYLC_HOME##*/}" || exit 1
-fi
+# Warn if executable not at $CYLC_HOME
 if [[ ! -x "${CYLC_HOME}/bin/${0##*/}" ]]; then
     echo 1>&2 "ERROR: ${0##*/} not found in ${CYLC_HOME}"
     exit 1
 fi
-# Execute the command in the selected installation.
+# Add CYLC_HOME to the start of the path to ensure the right version of python is used.
+PATH=${CYLC_HOME}/bin:${PATH}
```
oliver-sanders, line 186: After this Rose change the PATH manipulation here is subtracted from the Rose app-run environment.

https://github.com/metomi/rose/pull/2506

hjoliver (approved):
Looks good, works for me :+1:

hjoliver:
(conflicted)

wxtim:
Also closes https://github.com/metomi/rose/issues/2464

cylc/flow/etc/cylc
```diff
@@ -177,25 +177,19 @@ if [[ -z "${CYLC_HOME}" ]]; then
...
-    conda activate "${CYLC_HOME##*/}" || exit 1
-fi
+# Warn if executable not at $CYLC_HOME
 if [[ ! -x "${CYLC_HOME}/bin/${0##*/}" ]]; then
     echo 1>&2 "ERROR: ${0##*/} not found in ${CYLC_HOME}"
     exit 1
 fi
-# Execute the command in the selected installation.
+# Add CYLC_HOME to the start of the path to ensure the right version of python is used.
+PATH=${CYLC_HOME}/bin:${PATH}
```
dpmatthews, line 186: In most cases I don't think is needed (as long as you use the `--dest` option when using conda pack).
The only exception is `cylc hub` because `configurable-http-proxy` uses `#!/usr/bin/env node`.
Therefore  I suggest:
```
# Set PATH when running cylc hub so that configurable-http-proxy can find node
if [[ ${0##*/} =~ ^cylc && ${1:-} == "hub" && ! ${CYLC_VERSION:-} =~ ^7 ]]; then
  PATH=${CYLC_HOME}/bin:${PATH}
fi
```

dpmatthews (changes requested):

cylc/flow/etc/cylc
oliver-sanders: Not sure this is a good idea. It would work for running the Cylc executable due to the shebang in the Cylc Python script, however, if we don't modify the `$PATH` then calls we make out of Cylc will not pick up the Cylc environment.

wxtim:
```suggestion
# Set PATH when running cylc hub so that configurable-http-proxy can find node
if [[ ${0##*/} =~ ^cylc && ${1:-} == "hub" && ! ${CYLC_VERSION:-} =~ ^7 ]]; then
  PATH=${CYLC_HOME}/bin:${PATH}
fi
```

dpmatthews: We've decided to go ahead with this change assuming we don't find find any issues during testing

cylc/flow/scripts/cat_log.py
wxtim: @oliver-sanders  and @hjoliver and @dpmatthews I've realised that I might be able to make tests run without Activating conda by creating a `cylc get-python-exec` CLI, allowing the tests to invoke the correct python.

Is this a good idea? Perhaps it's for a follow on?

dpmatthews (approved):
Seems fine based on my limited testing

hjoliver (approved):
:tada: looks good

PR #4426, opened 2021-09-23
Fix bug with task name as substring of task name.

### LINKED ISSUE #3664: conditional prerequisite bug
**Describe the bug**

Diagnosed after [user bug report on Discourse forum](https://cylc.discourse.group/t/suite-shutting-down-with-name-archive-bool-is-not-defined-error/254/3)

Construction of conditional prerequisite expression fails for `NAME | preNAME => blah` due to overly simplistic string replacement in the `Prerequisite.set_condition()` method of `prerequisite.py`:
```python
# Make a Python expression so we can eval() the logic.
for message in self.satisfied:
    expr = expr.replace(self.MESSAGE_TEMPLATE % message,
                        self.SATISFIED_TEMPLATE % message)
```

- 7.8.x
- master (Cylc 8)

This simple case even fails validation (some only fail at runtime because we don't evaluate all prerequisites during validation):
```ini
[scheduling]
   [[graph]]
      R1 = "foo | xfoo => bar"
```
Validation:
```
$ cylc val foo
name 'xbool' is not defined
SuiteConfigError: bar: failed to evaluate triggers.
```

This probably warrants another Cylc-7 release too :grimacing: 

**Pull requests welcome!**

hjoliver:
These changes close #3664

**Requirements check-list**

### DIFF
--- a/cylc/flow/prerequisite.py
+++ b/cylc/flow/prerequisite.py
@@ -140,10 +140,11 @@ class Prerequisite:
             for message in self.satisfied:
                 # Use '\b' in case one task name is a substring of another
                 # and escape special chars ('.', timezone '+') in task IDs.
-                pattern = (
-                    r"\b" + re.escape(self.MESSAGE_TEMPLATE % message) + r"\b")
-                replace = self.SATISFIED_TEMPLATE % message
-                expr = re.sub(pattern, replace, expr)
+                expr = re.sub(
+                    r"\b" + re.escape(self.MESSAGE_TEMPLATE % message) + r"\b",
+                    self.SATISFIED_TEMPLATE % message,
+                    expr
+                )
             self.conditional_expression = expr
 
     def is_satisfied(self):
--- a/cylc/flow/prerequisite.py
+++ b/cylc/flow/prerequisite.py
@@ -17,6 +17,7 @@
 """Functionality for expressing and evaluating logical triggers."""
 
 import math
+import re
 
 from cylc.flow import ID_DELIM
 from cylc.flow.cycling.loader import get_point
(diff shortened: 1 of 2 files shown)


cylc/flow/prerequisite.py
```diff
@@ -115,16 +116,35 @@ def get_raw_conditional_expression(self):
...
         self._all_satisfied = None
         if '|' in expr:
             # Make a Python expression so we can eval() the logic.
             for message in self.satisfied:
-                expr = expr.replace(self.MESSAGE_TEMPLATE % message,
-                                    self.SATISFIED_TEMPLATE % message)
+                # Use '\b' in case one task name is a substring of another
+                # and escape special chars ('.', timezone '+') in task IDs.
+                expr = re.sub(
+                    r"\b" + re.escape(self.MESSAGE_TEMPLATE % message) + r"\b",
```
kinow, line 144: f-string for cylc-8.x branch only (7.8.x probably needs to stay as-is)?

```suggestion
                    fr"\b{re.escape(self.MESSAGE_TEMPLATE % message)}\b",
```

kinow (approved):
Tested on `master` and got a strange `xfoo`.

```py
['bool(self.satisfied[("foo", "1", "succeeded")])', 'bool(self.satisfied[("xfoo", "1", "succeeded")])']
```

Switched to this branch, and now got the correct result as per doctests.

```py
['bool(self.satisfied[("foo", "1", "succeeded")])', 'bool(self.satisfied[("xfoo", "1", "succeeded")])']
```

Minor comment, but looks good to me! Thanks @hjoliver

cylc/flow/prerequisite.py
hjoliver: Huh, I wasn't sure if I could nested templating like that!.

hjoliver:
> Tested on master and got a strange xfoo.

I think you pasted the wrong line there? The error case should have `xbool` in it, due to a bad string replacement.

kinow:
> > Tested on master and got a strange xfoo.
> 
> I think you pasted the wrong line there? The error case should have `xbool` in it, due to a bad string replacement.

Ops, copied from the wrong section from the terminal output, fixed! Thanks.

cylc/flow/prerequisite.py
hjoliver: But it does work, so I'll take it.

dwsutherland (approved):
Reproduced bug on master, couldn't on this branch 👍

Issue #4407, opened 2021-09-13
Review contents of log/flow-config directory


dpmatthews:
I've been looking at the contents of `log/flow-config` after the merge of #3763.
For a workflow which I started, reloaded and then stopped and restarted I get the following files:
```
20210913T124336Z-run.cylc
20210913T125325Z-reload.cylc
20210913T143828Z-restart.cylc
flow.cylc.processed
```

- [x] `*-run.cylc` - this would be better named `*-start.cylc`

- [x] `flow.cylc.processed` would be better named something like  `processed-flow.cylc` or `latest-processed-flow.cylc` so that it has the correct extension.

- [ ] As discussed in #3763, the contents of the parsed config files would be better replaced with the output of `cylc config` (they currently have lots of empty sections and doesn't match the output of `cylc config` or `cylc config --defaults`).

hjoliver:
> `*-run.cylc` - this would be better named `*-start.cylc`

:+1: 

> `flow.cylc.processed` would be better named something like `processed-flow.cylc` or `latest-processed-flow.cylc` so that it has the correct extension.

:+1: 

> As discussed in future of the `suite.rc.processed` file #3763, the contents of the parsed config files would be better replaced with the output of `cylc config` (they currently have lots of empty sections and doesn't match the output of `cylc config` or `cylc config --defaults`).

We also discussed this at length in a recent meeting. I'm pretty sure you were heavily in favour of keeping the processed file itself, because it can be useful when debugging others' workflows.  Hence the final comment on the issue: https://github.com/cylc/cylc-flow/issues/3763#issuecomment-900139968 

... but we still need to add the `cylc config`

dpmatthews:
> I'm pretty sure you were heavily in favour of keeping the processed file itself, because it can be useful when debugging others' workflows. Hence the final comment on the issue: #3763 (comment) ... but we still need to add the cylc config

Yes, we want both. I just want the parsed config file to contain the output of `cylc config` rather than the current contents.

wxtim:
@dpmatthews - `run` ⇒ `start` or `run` ⇒ `play`?

dpmatthews:
> @dpmatthews - `run` ⇒ `start` or `run` ⇒ `play`?

I think it has to be `start` if we're going to stick with `restart`.

PR #4430, opened 2021-09-23
Update filenames in the `log` directory [small, but can be pushed to rc1 if it doens't make b3]

wxtim:
These changes partially address #4407 

**Requirements check-list**


cylc/flow/scheduler.py
```diff
@@ -1052,7 +1052,8 @@ def load_flow_file(self, is_reload=False):
             mem_log_func=self.profiler.log_memory,
             output_fname=os.path.join(
                 self.workflow_run_dir, 'log', 'flow-config',
-                workflow_files.WorkflowFiles.FLOW_FILE + '.processed'),
+                'flow-processed.cylc'
```
hjoliver, line 1055: I suppose we should have a `WorkflowFiles.PROCESSED_FLOW_FILE`, but no biggie.

hjoliver (approved):

cylc/flow/scheduler.py
wxtim: Fair-enough: I didn't think it was going to  be re-used.

datamel (approved):
I have read the code, checked out the branch and run the tests. I am aware that the tests are not passing CI but I think this is unrelated.

Issue #4131, opened 2021-03-17
data store: triggering tasks beyond the scheduler window


oliver-sanders:
If you trigger a task way out beyond the n=0 window two strange things happen (or not).

1. The triggered task runs, however, it is not included in the n=0 window (or even the n=1 window).
2. When the scheduler catches up and the task enters the n=2 window it's dependencies (likely in the n=1 window) will be marked as succeeded as if they had run.

Example workflow:

```ini
# a simple workflow which runs a sequential task every cycle

[scheduler]
    allow implicit tasks = True

[scheduling]
    initial cycle point = 1
    cycling mode = integer
    [[graph]]
        P1 = """
            a[-P1] => a
        """
```

Example Steps:

```bash
# run the workflow
cylc play <flow> --hold
# trigger a task way out in the future
cylc trigger <flow> a.10
cylc tui <flow>
```

You should see `a.1` & `a.2` as expected, however, `a.10` and `a.11` should appear too:

![Screenshot from 2021-03-17 14-34-23](https://user-images.githubusercontent.com/16705946/111484861-0958ec00-872e-11eb-8a2d-6d1ed4e2258f.png)

```bash
# open tui
cylc tui <flow>
# wait for the scheduler to catch up
cylc unhold <flow>
```

`a.10` will appear now, however `a.9` shows up as if it had previously run and succeeded which it didn't:

![Screenshot from 2021-03-17 14-35-17](https://user-images.githubusercontent.com/16705946/111485008-28577e00-872e-11eb-976d-af145fa0a05a.png)

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

hjoliver:
This might be partly a runahead thing.  I'll take a close look after finishing #4103 (which severely messes with runahead code).

hjoliver:
Back for a look at this one, finally.  Running you example on current master:

- At start up, with `cylc play --pause`, the first task `a.1`now shows as runahead-limited rather than paused.
   -  That's because we don't let tasks do anything (other than change state via receive job messages) when the workflow is paused.
   - I guess it would be clearer to release from runahead and hold the waiting task, instead.

- When triggered, `a.10` appears, as runahead-limited for the same reasons as `a.1`
- `a.11` also appears, correctly as it is now at `n=1` relative to `a.10` (which is now in the scheduler's active pool)
- `a.9` appears wrongly as succeeded
[...]

PR #4436, opened 2021-09-28
Release tasks from runahead even if workflow paused.

hjoliver:
Partially addresses #4131 

**Requirements check-list**
- [ ] Applied any dependency changes to both `setup.py` and `conda-environment.yml`.


oliver-sanders:
Tested, task now shows up as queued rather than runahead which makes much more sense :+1:.

hjoliver:
The example from #4131, with the workflow paused and `a.10` triggered manually:

- `a.1` and `a.10` are waiting in the scheduler task pool, for the workflow to be released
- `a.2` and `a.11` are in the `n=1` window relative to the scheduler task pool
- `a.9` is a historical task at `n=1`, in the sense that it comes before an active task in the graph; the datastore marks these as `expired` for now (in future, it will consult the database)

shot

hjoliver:
Note I chose a circle with a strike through it for the expired task state icon. Think of a waiting task that has been struck out because it has expired and won't run now. Is it too similar to the submit-failed icon?  Maybe not. The meaning isn't entirely different (both mean the task is no longer waiting but never actually started executing) and it's the most rarely seen state.  However, it's easy to change if someone has a better idea.

shot

MetRonnie:
> Note I chose a circle with a strike through it for the expired task state icon. Think of a waiting task that has been "cancelled" because it has expired. Is it too similar to the submit-failed icon? Maybe not. The meaning isn't entirely different (both mean the task is no longer waiting but never started running) and it's the most rarely seen state. However, it's easy to change if someone has a better idea.

Some ideas:
- U+229D "Circled Dash": &#8861;
- U+2296 "Circled Minus": &#8854;
- U+24D4 "Circled Latin Small Letter E": &#9428;
- U+25CC "Dotted Circle": &#9676;
- U+29B8 "Circled Reverse Solidus": &#10680;

I'm not sure about font support for these

hjoliver:
New feature added to the Issue description above:
- spawn parentless tasks out to the runahead limit immediately instead of relying partially on the main loop to do it

This branch was giving me trouble with some functional and integration tests because it was hard to understand and predict the task pool content before the limit was reached. Now, even with `cylc play --pause` parentless tasks will exist out to the limit.

```
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    runahead limit = P5
    [[graph]]
        P1 = foo => bar => baz
[runtime]
    [[foo, bar, baz]]
```
Start up paused:
shot1

Delayed case, not paused:
```
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    runahead limit = P5
    [[graph]]
        R1 = foo
        P1 = foo[^] => bar => baz
[runtime]
    [[foo, bar, baz]]
        script = sleep 2
```
shot2

hjoliver:
On current master, the start-paused case will show only `foo.1` (as runahead limited, which is confusing) and the delayed absolute trigger case will spawn out the runahead limit somewhat incrementally.

hjoliver:
@oliver-sanders - tagging you for review, can you assign another from your end according to availability?

oliver-sanders:
> U+25CC "Dotted Circle": ◌

I thought that was how I had set it up in the first instance. Perhaps I thought that expired might become a badge rather than a top-level task state. Dunno.

Dotted circle comes from the same font table as the other circles so should have the same relative dimensions and font support.

hjoliver:
> > U+25CC "Dotted Circle": ◌
> 
> I thought that was how I had set it up in the first instance. Perhaps I thought that expired might become a badge rather than a top-level task state. Dunno.
> 
> Dotted circle comes from the same font table as the other circles so should have the same relative dimensions and font support.

OK, changing to dotted circle. It looks OK in TUI.

Issue #4431, opened 2021-09-23
Polling can result in incorrect submit-failed


dpmatthews:
See problem reported on Discourse: https://cylc.discourse.group/t/how-does-cylc-read-pbs-job-exit-code/375/4

I've looked at the polling logic: https://github.com/cylc/cylc-flow/blob/7.8.7/lib/cylc/batch_sys_manager.py#L299
If I understand it correctly, we read all the job status files and then check the queue status for any jobs which haven't previously exited. This means there is a chance of inconsistency if the polling checks the job status file and finds it hasn't started yet but then, by the time it checks for the job in the queue it has already run and exited the queue. The polling would then report the job as exited without having started which results in submit-failed.
In normal circumstances this is very unlikely to happen but if we are polling a large number of jobs at the same time and/or the login node is heavily loaded it is a possibility.

I have been able to reproduce the problem (using Cylc 7.8.7) by inserted a sleep in the code in between the read of the status files and the queue status check (to mimic the read of the status files taking a long time).

To fix the problem I think we should re-read the status file for any job found to have exited the batch system to ensure we report up to date information.

hjoliver:
- [x] fix for 7.8.x.  #4437 merged
- [  ] fix for 8.x  (presumably it will have the same problem)

hjoliver:
Moving to an 8.0 milestone now that the 7.8.x fix is done.

PR #4437, opened 2021-09-28
prevent slow polling producing incorrect submit-fail

wxtim:
These changes partially address #4431  - The remaining thing to do is to check that same bug doesn't exist at Cylc8

**Requirements check-list**
  `conda-environment.yml`.
- [x] Does not need tests (Manual testing only?).
- [x] No documentation update required - should fix a largely unencountered bug.

### DIFF
--- a/lib/cylc/batch_sys_manager.py
+++ b/lib/cylc/batch_sys_manager.py
@@ -605,7 +605,6 @@ class BatchSysManager(object):
                     ctx.batch_sys_exit_polled = 0
                 else:
                     debug_flag = True
-
             # Add information to "job.status"
             if ctx.batch_sys_exit_polled:
                 try:
--- a/lib/cylc/batch_sys_manager.py
+++ b/lib/cylc/batch_sys_manager.py
@@ -168,6 +168,10 @@ class JobPollContext(object):
                 else:
                     raise ValueError('Invalid kwarg "%s"' % key)
 
+    def update(self, other):
+        for i in self.__slots__:
+            setattr(self, i, getattr(other, i))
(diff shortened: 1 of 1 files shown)


lib/cylc/batch_sys_manager.py
```diff
@@ -601,6 +605,7 @@ def _jobs_poll_batch_sys(self, job_log_root, batch_sys_name, my_ctx_list):
                     ctx.batch_sys_exit_polled = 0
                 else:
                     debug_flag = True
+
```
wxtim, line 608:
```suggestion
```

dpmatthews (approved):
I've tested that this change fixes the problem

lib/cylc/batch_sys_manager.py
```diff
@@ -613,6 +617,10 @@ def _jobs_poll_batch_sys(self, job_log_root, batch_sys_name, my_ctx_list):
                 except IOError as exc:
                     sys.stderr.write(str(exc) + "\n")
 
+                file_ctx = self._jobs_poll_status_files(
```
hjoliver, line 620:
```suggestion
                # Re-read the status file in case the job started and exited
                # between the file and batch system checks, which would be
                # interpreted as submit-failed (job exited without starting).
                # Possible if polling many jobs and/or system heavily loaded.
                file_ctx = self._jobs_poll_status_files(
```

hjoliver (approved):
Nice, thanks @wxtim - just added doc string and comment suggestions, which I'll merge so we can get the final_73 cylc-7 release out.

PR #4440, opened 2021-09-29
cylc clean improvements

MetRonnie:
This is a series of changes with no associated Issue.

- Fix traceback during remote clean when doing `cylc clean myflow --remote`  where `~/cylc-run/myflow` contains exactly 1 run dir 
- Improve logging to terminal
  - Explicitly log what's being deleted when using `-v` option
  - `--rm` now implies `-v`
  - Make everything below logging level `WARNING` go to stdout instead of stderr (for `cylc clean` only); this allows stdout to be logged during remote clean so the above 2 bullet points can work for remote clean
- Allow `cylc clean myflow/runN` to infer the run number
- Prevent `cylc clean myflow --rm something` (where `~/cylc-run/myflow` contains exactly 1 run dir) from trying to delete `~/cylc-run/myflow/something` (i.e. it will only delete `~/cylc-run/myflow/run1/something`)

**Requirements check-list**


wxtim (approved):
- [x] Read
- [x] Played with - working as promised.

cylc/flow/pathutil.py
```diff
@@ -260,24 +261,24 @@ def remove_dir_and_target(path: Union[Path, str]) -> None:
     Args:
         path: the absolute path of the directory to delete.
     """
+    log = LOG.info if cylc.flow.flags.verbosity == 1 else LOG.debug
```
oliver-sanders, line 264: I think these should stay at `LOG.debug`, the visibility of the messages being decided by the level of the log handlers.

Is there a reason that approach won't work here?

cylc/flow/pathutil.py
```diff
@@ -288,16 +289,17 @@ def remove_dir_or_file(path: Union[Path, str]) -> None:
     Args:
         path: the absolute path of the directory/file/symlink to delete.
     """
+    log = LOG.info if cylc.flow.flags.verbosity == 1 else LOG.debug
```
oliver-sanders, line 292: Dynamic log level?

cylc/flow/pathutil.py
```diff
@@ -315,6 +317,7 @@ def remove_empty_parents(
         /foo/bar/a/b (assuming it's empty), then /foo/bar/a (assuming it's
         empty).
     """
+    log = LOG.info if cylc.flow.flags.verbosity == 1 else LOG.debug
```
oliver-sanders, line 320: Dynamic log level?

oliver-sanders (commented):
Looks good, yet to test.

I'm not convinced by the dynamic log level stuff, is there a reason we have to change the level of the message rather than the level of the handler?

cylc/flow/pathutil.py
MetRonnie: Ideally there would be a log level between `INFO` and `DEBUG` for the `-v` level, I suppose. Or are you saying we can manually set the `CYLC_LOG` handler level to `DEBUG` for `cylc clean -v`, inside the `init_clean` function?

oliver-sanders: Ok, got it, DEBUG messages don't show in VERBOSE mode so you made them INFO level to make them visible.

VERBOSE mode seems a bit pointless ATM, I'm trying to remember what the original idea was...

Options:

1. We could create a custom log level for VERBOSE (`logging` supports this). 
   * However, do we really need it? Would it really be helpful to separate log messages like this?
   * (related) We have talked about creating a TASK log level (at ~ the same level as info) to help separate task and scheduler events in the log.
2. We could set the log level to DEBUG in verbose mode
   * I've forgotten what the original idea was, but this would make sense.

With (2) the levels would become:

> QUIET

Shows WARNING or higher.

> INFO

Shows INFO or higher.

> VERBOSE

Shows DEBUG or higher.

> DEBUG

Shows DEBUG or higher with full traceback for debug purposes.

MetRonnie: I've decided to make the `cylc.flow.pathutil` functions for deleting files/dirs/symlinks log at `INFO` level so that it's clear what's being deleted. This undoes 6066648b9d1390dd1da386ec7630410ce852e3a6. I hope this is fair - it only logs the bare minimum of deleted things (if it's removing a directory, it only logs that, not any constituent files; if a glob matches children of a matching parent, it only logs the parent etc). The only places where the logging may get a bit much are when   the workflow has run on many different install targets, or when doing `--rm 'log/job/*'` instead of `--rm log/job` for a workflow with many cycle points.

cylc/flow/workflow_files.py
```diff
@@ -996,30 +1006,36 @@ def remote_clean(
...
                 )
             else:  # Exhausted list of platforms
-                failed_targets.append(item.install_target)
+                failed_targets[item.install_target] = excp
         elif err:
-            LOG.debug(err)
+            # Only show stderr from remote host in debug mode if ret code 0
+            # because stderr often contains useless stuff like ssh login
+            # messages
+            LOG.debug(f"[{item.install_target}] {err}")
```
oliver-sanders, line 1031:
```suggestion
            LOG.debug(f"[{item.install_target}]\n{err}")
```

oliver-sanders (commented):
Small change that should make the remote logs a little easier to read.

oliver-sanders (approved):
I think the logging looks good, the output is pretty minimal and it's very clear what `cylc clean` is doing :+1:.

wxtim (approved):

PR #4443, opened 2021-10-01
Fix job polling bug.

### LINKED ISSUE #4431: Polling can result in incorrect submit-failed

hjoliver:
Manual cherry-pick 6ea49470dc8f5823e57433f47252b3b35239aacc from #4437 
Credit @wxtim 

Close #4431 

But reproduced on master with this workflow *and `time.sleep(10)` between status file and job runner check in `job_runner_mgr.py`*:
```
[scheduling]
    [[graph]]
        R1 = "poller & fast & waiter"
[runtime]
    [[poller]]
        script = cylc poll $CYLC_WORKFLOW_NAME fast.1
    [[fast]]
        init-script = sleep 5  # fake delayed start
    [[waiter]]
        script = "cylc__job__poll_grep_workflow_log '\(polled\)'"
```

**Requirements check-list**
- [ ] Applied any dependency changes to both `setup.py` and `conda-environment.yml` (none)
- [x] Does not need tests (note testable without bodging the code).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -100,6 +100,9 @@ now accepts ``&`` and ``|`` as valid line breaks in the same way as ``=>``.
 
 ### Fixes
 
+[#4443](https://github.com/cylc/cylc-flow/pull/4443) - fix for slow polling
+generating an incorrect submit-failed result.
+
 [#4421](https://github.com/cylc/cylc-flow/pull/4421) -
 Remove use of the `ps` system call (fixes a bug reported with Alpine Linux).
 
--- a/cylc/flow/job_runner_mgr.py
+++ b/cylc/flow/job_runner_mgr.py
@@ -176,6 +176,11 @@ class JobPollContext():
                     raise ValueError('Invalid kwarg "%s"' % key)
                 setattr(self, key, value)
 
+    def update(self, other):
+        """Update my data from given file context."""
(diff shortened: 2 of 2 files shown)


wxtim (approved):

dpmatthews (approved):
Reproduced problem with Cylc 8 and confirmed this fixes it.

PR #4343, opened 2021-08-06
Optional outputs

hjoliver:
These changes address https://github.com/cylc/cylc-admin/pull/129

Close #4372 
Close #4256
Close #2914 
Close #3749
Close #4397
Close #4402

- [x]  retrigger incomplete tasks without needing `--reflow` and without merging a new flow number
- [x] Cylc 7 back-compat:
      - suicide triggers still work if needed and have no impact if not needed
      - detect "suite.rc" and print a deprecation warning
      - infer `a:succeed ` and `a:fail` are optional if both appear in the graph
      - make all custom outputs optional (because we can't tell if they should always be completed or not)
- [x] Count incomplete tasks and partially satisfied prereqs toward runahead limit
- [x] If nothing else to run, stall if incomplete tasks or partial prereqs are present 

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.py` and  `conda-environment.yml`. (None)
- [x] Appropriate tests are included (unit and/or functional).
- [x] Appropriate change log entry included.
- [ ] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/XXXX.

PR #4446, opened 2021-10-01
Allow implicit tasks in Cylc 7 backwards compatibility mode

### LINKED ISSUE #4385: [scheduler]allow implicit tasks = True in Cylc7 compat mode?
The `[scheduler]allow implicit tasks` config has no Cylc 7 equivalent which makes writing Cylc 7 / 8 compatible workflows impossible when this is relied on.

Most documented Cylc 7 examples will probably fail validation at Cylc 8 because of this.

Should we make `allow implicit tasks = True` the default in Cylc 7 compat mode:

* Big help for Pure Cylc users.
* Could be confusing for Rose 2019 users until they upgrade (although we could make it mandatory if a `rose-suite.conf` file is present).

**Pull requests welcome!**

MetRonnie:
These changes close #4385 

Allow implicit tasks in Cylc 7 backwards compatibility mode, even if `[scheduler]allow implicit tasks` is not set (actually, even if it is False - is that undesirable?).

**Requirements check-list**
- [x] No change log entry required (Follow-up to #4343, not yet released).


wxtim:
> actually, even if it is False - is that undesirable?

Yes - I don't imagine it'll happen often, but I think so.

cylc/flow/config.py
```diff
@@ -751,22 +751,26 @@ def _check_implicit_tasks(self) -> None:
         queue config, unless allowed by config."""
         if self.implicit_tasks:
             print_limit = 10
-            implicit_tasks_str = '\n    * '.join(
+            impl_tasks_str = '\n    * '.join(
```
wxtim, line 754: I can't tell whether `impl` expands to `implicit` or `implementation` or something else. I'd prefer the original term.

wxtim (changes requested):
- [x] Read
- [x] Tested as working
- [ ] Requested changes: I believe that the current functionality is confusing.

MetRonnie:
> > actually, even if it is False - is that undesirable?
> 
> Yes - I don't imagine it'll happen often, but I think so. Imagine how confused the user would be if they explicitly set False and that was ignored!

Actually it isn't possible to address that because `allow implicit tasks = False` is the default. There is no way to distinguish between the default and explicitly setting it to `False`.

Issue #4394, opened 2021-09-07
Disallow runX as a workflow name


MetRonnie:
As described by Bruno on Element, `cylc install --flow-name=foo/run4` results in a run directory `~/cylc-run/foo/run4/run1` which is confusing, so should probably be disallowed. This should mean disallowing names matching the regex `run\d+` <del>(I think `runN` is already disallowed, but need to check that)</del> and `runN`.

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

kinow:
>(I think runN is already disallowed, but need to check that).

You are allowed to install `a/b/c/runN`. It creates the folders, but somewhere down the line Cylc Flow understands that is invalid and doesn't show it on `cylc scan`.

![image](https://user-images.githubusercontent.com/304786/133022137-fb60c247-fe19-4f9f-b71d-42bd42329783.png)

```
(venv) kinow@ranma:~/cylc-src/three$ cylc install --flow-name=a/b/c/runN
INSTALLED a/b/c/runN/run1 from /home/kinow/cylc-src/three
(venv) kinow@ranma:~/cylc-src/three$ cylc scan --state=all
five
747
three/run8
a/run1
a/b/run1
one/two/three/run2
one/two/three/run4 ranma:43010
one/two/three/run3 ranma:43023
one/two/three/run1 ranma:43085
```

PR #4448, opened 2021-10-04
Ensure `check_nested_run_dirs()` checks for symlink run dirs

MetRonnie:

**Requirements check-list**


cylc/flow/workflow_files.py
```diff
@@ -1345,30 +1344,29 @@ def check_nested_run_dirs(run_dir: Union[Path, str], flow_name: str) -> None:
...
-        '"{0}" as "{1}" is already a valid run directory.'
+        "Nested run directories not allowed - cannot install workflow in "
+        "'{0}' as '{1}' is already a valid run directory."
     )
+    reg_path = Path(os.path.normpath(run_dir))
 
     def _check_child_dirs(path: Union[Path, str], depth_count: int = 1):
         for result in os.scandir(path):
-            if result.is_dir() and not result.is_symlink():
+            if result.is_dir():
```
MetRonnie, line 1354: Not sure why I excluded symlinks in the first place

tests/unit/test_workflow_files.py
```diff
@@ -1553,44 +1571,6 @@ def test_remote_clean_cmd(
...
-                is_symlink=lambda: True
-            )
-        ]
-    )
-
-    try:
-        check_nested_run_dirs(run_dir, 'placeholder_flow')
-    except ServiceFileError:
-        pytest.fail(
-            "Unexpected ServiceFileError, Check symlink logic.")
```
MetRonnie, line 1591: @datamel I have removed this test and replaced it with the one above. This test seemed to do the opposite of what it said it should? Surely it should have been using `with pytest.raises()`?

MetRonnie:
In fact, it looks like we don't even need to check the children of a run dir when installing because if the specified dir already exists, `cylc install` fails with
```
WorkflowFilesError: "~/tempflow" exists. Try using cylc reinstall. Alternatively, install with
another name, using the --run-name option.
```
The checking of the children was necessary for registering a workflow in earlier pre-release versions of Cylc 8 (#3852), but not anymore with `cylc install`.

@wxtim If I'm right, it means we don't have to check the children for `_cylc-install` dirs either

wxtim:
> @wxtim If I'm right, it means we don't have to check the children for `_cylc-install` dirs either

I believe that you are right.

wxtim:
IMO we probably need to have a chat about exactly what we want to allow or not. 

I think that this shouldn't be allowed:

```
> cylc install recursive/simplest/run1
WorkflowFilesError: Nested run directories not allowed - cannot install workflow in '/home/h02/tpilling/cylc-run/recursive/simplest/run1/run2' as '/home/h02/tpilling/cylc-run/recursive/simplest/run1' is already a valid run directory.
> cylc install recursive/simplest/run42 
INSTALLED recursive/simplest/run42/run1 from /home/h02/tpilling/cylc-src/recursive/simplest/run42
```

I think that the rules I want to formulate are:

- Run dirs must not be nested
- Run dirs paths must not include any folders of the form `run\d+` or `runN` (I think that the latter is already covered) between themselves and the install dir.
- Install dirs must not be nested.

I think that the child checks are unnecessary in both cases.

wxtim (commented):
I'm part-way through reviewing. I think that it mostly works as intended.
I still need to test it a bit more.

MetRonnie:
> Run dir paths must not include any folders of the form `run\d+` or `runN` (I think that the latter is already covered)

And there is an issue for the former: #4394

wxtim (approved):

tests/unit/test_workflow_files.py
```diff
@@ -100,44 +100,25 @@ def test_is_valid_run_dir(is_abs_path: bool, tmp_run_dir: Callable):
     assert workflow_files.is_valid_run_dir(Path(prefix, 'foo/bar')) is True
 
 
-def test_check_nested_run_dirs_parents(tmp_run_dir: Callable):
+def test_check_nested_run_dirs__parents(tmp_run_dir: Callable):
```
datamel, line 103:
```suggestion
def test_check_nested_run_dirs_parents(tmp_run_dir: Callable):
```

datamel (approved):
No functional problems found from me. Tests might need a kick to make sure they pass. Timeout unlikely to be related to change.
Small double `_` to single `_` suggestion.
Thanks @MetRonnie

datamel:
You have skipped CI but the tests had timed out and never passed.

MetRonnie:
Unfortunately that is quite common for `remote_background_indep_poll`

PR #4460, opened 2021-10-11
Disallow implicit tasks in Cylc 7 back-compat when `rose-suite.conf` present

MetRonnie:
These changes finish closing #4385. This is a followup to #4446.

Now:
- `[scheduler]allow implicit tasks = True` will always allow implicit tasks
- `[scheduler]allow implicit tasks = False` (explicitly set) will always disallow implicit tasks, even in Cylc 7 backward compatibility mode
- If `[scheduler]allow implicit tasks` is not set, Cylc 7 back compat mode is on, but `rose-suite.conf` is present in the run dir, implicit tasks will be disallowed

> **Context:** Rose 2019 used to call `cylc validate --strict` which did not allow implicit tasks. Consequently Rose users are used to the idea that Cylc forbids implicit tasks. One of the complex relations between the two we are glad to be rid of!

**Requirements check-list**
- [x] No changelog entry needed (Followup to #4343, not yet released.)
- [x] (master branch) Docs PR ongoing: cylc/cylc-doc/pull/304

### DIFF
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -160,8 +160,9 @@ with Conf(
             ``False`` after finishing the :cylc:conf:`flow.cylc[runtime]`
             section.
 
-           Note it is not possible to disallow implicit tasks in
-            :ref:`Cylc 7 backwards compatiblity mode <BackCompat>`.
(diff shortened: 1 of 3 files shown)


hjoliver:
> Ah, the information we need is preserved, just in the totally obvious pcfg object (with sparse=True):

Apologies @MetRonnie I should have remembered that, for your first PR!

PR #4470, opened 2021-10-18
graphql: use a common interface for added / updated deltas

oliver-sanders:
The added and updated delta types are essentially identical. Usually we subscribe to the same fields for both added and updated.

This PR uses a GraphQL interface to make this slightly nicer, example using `GscanSubscriptionQuery`:

```diff
  fragment GScanTreeDeltas on Deltas {
    id
    added {
-    ...GscanAddedData
+    ...GscanDelta
    }
    updated {
-    ...GscanUpdatedData
+    ...GscanDelta
    }
    pruned {
      workflow
    }
  }

- fragment GscanAddedData on Added {
+ fragment GscanData on Added {
    workflow {
      ...WorkflowData
    }
  }

- fragment GscanUpdatedData on Updated {
-   workflow {
-     ...WorkflowData
-   }
- }

```

Boiler plate "hello world" deltas subscription after this change:

```graphql
fragment ExampleTreeData on Delta {
  workflow {
    id
  }
  cyclePoints: familyProxies(ids: ["root"], ghosts: true) {
    id
  }
  familyProxies(exids: ["root"], sort: {keys: ["name"]}, ghosts: true) {
    id
  }
  taskProxies(sort: {keys: ["cyclePoint"], reverse: false}, ghosts: true) {
    id
  }
  jobs(sort: {keys: ["submit_num"], reverse: true}) {
    id
  }
}

subscription ExampleTreeQuery {
  deltas(stripNull: true) {
    id
    added {
      ...ExampleTreeData
    }
    updated {
      ...ExampleTreeData
    }
    pruned {
      workflow
    }
  }
}
```

This doesn't change the existing types so no changes to the UI subscriptions should be required.

**Requirements check-list**
  `conda-environment.yml`.

### DIFF
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -27,7 +27,7 @@ from typing import AsyncGenerator, Any
 from graphene import (
     Boolean, Field, Float, ID, InputObjectType, Int,
     List, Mutation, ObjectType, Schema, String, Union, Enum,
-    Argument
+    Argument, Interface
 )
 from graphene.types.generic import GenericScalar
 from graphene.utils.str_converters import to_snake_case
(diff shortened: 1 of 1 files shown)


MetRonnie (approved):
LGTM
- `Added` seems to be the same as before from what I can tell from https://docs.graphene-python.org/en/stable/types/interfaces/
- Tested out running workflows in the UI and everything looks to be working

dwsutherland (approved):
Looks good to me 👍

Issue #4453, opened 2021-10-06
Cylc Install, nested source directory


datamel:
Currently, you can perform cylc install workflow1 on a source directory with the following structure...
```
   $ tree ~/cylc-src
   └── workflow1
       ├── workflow2
       │   └── flow.cylc 
       └── flow.cylc
```

Cylc install does check for nested _run_ directories (by looking for `.service` directory) but this current scenario will install without error. 

Cylc scan, will not currently pick up `workflow2` (`workflow1` does appear on scan).

Some thought needed into whether we wish to check and forbid a scenario like this (although may be undesirable for performance to nested scan), or perhaps clear documentation on best practice for sub-workflows and/or a tutorial would be an acceptable solution.

hjoliver:
(My comments above might be slightly off-piste, if you're only worried about `cylc scan` seeing these source dirs, or having `cylc install` balk at them?)

hjoliver:
Closed by #4861 ... 

...  The nested source structure in the description above is the natural way to store sub-workflow source directories, and #4861 makes this work in Cylc 8 by allowing source directories under `cylc-run` (for on-the-fly installation of sub-workflow instances).

PR #4477, opened 2021-10-26
Allow nested sub-workflows

hjoliver:
Allow sub-workflow definitions under the `~/cylc-run` directory.  

This simple change address part of the discussion under #4453. It just changes (and rewords) an exception into a warning, so that `cylc install` does not balk at installing the sub-workflow at runtime, when its `flow.cylc` is found under `~/cylc-run`.

Context: a sub-workflow definition is just another source flie for the main workflow. As such, it should be installed with other main workflow source files into the main workflow run directory.  Then at run-time, the sub-workflow gets "instantiated" anew for each main workflow cycle point (via another `cylc install`) from the *installed* sub-workflow source.

**Requirements check-list**
  `conda-environment.yml`.

### DIFF
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -1791,9 +1791,9 @@ def validate_source_dir(source, workflow_name):
     cylc_run_dir = Path(get_cylc_run_dir())
     if (os.path.abspath(os.path.realpath(cylc_run_dir))
             in os.path.abspath(os.path.realpath(source))):
-        raise WorkflowFilesError(
-            f"{workflow_name} installation failed. Source directory "
-            f"should not be in {cylc_run_dir}")
+        LOG.warning(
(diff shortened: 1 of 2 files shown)


hjoliver:
> Drop cylc install for the installation of sub-workflows from installed source.

Agreed.  Install and reinstall work correctly in my example (given that a sub-workflow instance needs to be based on the installed main workflow source) but there's not much point in explicitly exposing that because we don't want to encourage users to edit sub-workflow sources (or anything else) in the main run directory.

Note from last meeting we agreed not to let this issue hold up 8.0rc1. If it's much work to implement the above we can start by recommending users take the straightforward manual (but ugly) approach of managing subworkflows as entirely separate workflows, which is how it works at Cylc 7 anyway.

hjoliver:
Superseded by #4811

PR #4482, opened 2021-10-29
Task "preparing" state bug fixes

hjoliver:

Fixes:
- the datastore was creating job proxies for tasks in the `preparing` state, but labeling them as `submitted`
- the task state change to `preparing` was not being logged
- TUI had not been updated for the change from `ready` to `preparing`

Also:
- new icon to distinguish `preparing` from `waiting` in TUI

**Requirements check-list**
  `conda-environment.yml`.

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -76,6 +76,7 @@ from cylc.flow.task_job_logs import JOB_LOG_OPTS, get_task_job_log
 from cylc.flow.task_proxy import TaskProxy
 from cylc.flow.task_state import (
     TASK_STATUS_WAITING,
+    TASK_STATUS_PREPARING,
     TASK_STATUS_SUBMITTED,
     TASK_STATUS_SUBMIT_FAILED,
     TASK_STATUS_RUNNING,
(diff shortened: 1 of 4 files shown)


hjoliver:
> I don't think preparing should be job state, I thought job proxies are supposed to reflect interactions with the job runner.
>
> So long as the task has the preparing state I don't see any benefit of having an associated pseudo job.

Pasting from the chat forum:

> Yes, not intended to be a job state (it's not one in my old "task state rationalization" proposal on cylc-admin either.
>
> However, the GUI, TUI, and the datastore are all broken with respect to that intention right now (on master branches).
>
> I was trying to compare task state transitions carefully in the scheduler and the UIs at the same time, in an attempt to see what was going wrong in the GUI (the non-waiting tasks wrongly marked as runahead - and other errors actually), so I put artificial delays in the scheduler and jobs to extend the preparing and submitted states. That revealed:
> 
> - the datastore is creating job proxies initially for the preparing state, but labeling them as submitted, and those appear in the deltas
> - TUI is displaying a submitted job icon when the task is preparing (with no job ID)
the GUI is getting the task state icon sequencing wrong between preparing and submitted
The outright errors in the GUI and TUI were most easily fixed (with only a few lines of code each) to correctly display "preparing" job icons, to match the data coming from the datastore.
> 
> This makes the task/job state sequencing work correctly in the GUI and TUI, which gets one set of problems out of the way if you're trying to diagnose apparent task status errors.
>
> It will be super easy to get rid of the preparing job status again once it's dealt with in the datastore (which I wasn't confident to attempt fixing in a hurry, especially with david working on the delta issues right now).

hjoliver:
> Perhaps \u229A:

I considered that, but it makes preparing look more important than submitted. Doesn't matter too much though I guess, as it's normally a transient state.

hjoliver:
Two approvals, the job proxy bit explained (and #4484), and we're in a desperate hurry ... merging.

PR #4485, opened 2021-11-01
Preparing job status

### LINKED ISSUE #4484: datastore: create job proxies at the submitted state
"Preparing" is not supposed to be a job state - see https://cylc.github.io/cylc-admin/proposal-state-names.html#taskjob-states  Strictly speaking it's not needed as a job state because the preparing task state shows what the scheduler is doing.

However, job proxies currently get created at the transition to task state "preparing", and they contain some transient data available in the scheduler at that stage.

#4482 and https://github.com/cylc/cylc-ui/pull/824 make the UIs correctly represent the data that they're receiving (attempting to interpret the data as if the datastore implemented the proposal in this respect was causing some buggy behaviour).

We should:
- probably: delay job proxy creation in the store, or hide job proxies until they reach the submitted state.
- maybe: consider keeping the preparing job icons, as it can be argued both ways - need a quick decision as this is user facing

oliver-sanders:
Removes the legacy of the Cylc 7 "ready" job state and fixes some issues discovered along the way.

Fixes #4484 
Addresses https://github.com/cylc/cylc-admin/pull/47

Cylc job items are now created in the data store on submission rather than on preparation.

**Misc Fixes:**

* Jobs which failed during preparation (before job-submit) are now added to the data store.
  * Before the task failed with no jobs, technically correct, but unhelpful.
  * Now we get a submit-failed job with the platform information provided so users can debug.
* Tasks now have jobs when running in simulation mode.
* Fixes a bug where the job platform in the data store may be incorrect.
  * This can happen if the platform it has changed during the preparation process as a result of intelligent host selection.

**Note:**

The "submitting" job is still written to the database so that Cylc can reconnect with it if the workflow is restarted resulting in a small discontinuity between the DB and data store for this edge case. The discontinuity for "prep-failed" tasks has been removed.

**Testing Suggestions:**

* Configure a workflow with a task running on a non-existent *platform*, run it, view it in Tui/Gui, you should *now* see a submit-failed job.
* Configure a workflow with a task running on a platform with a non-existant *host*, run it, view it in Tui/Gui, you should *still* see a submit-failed job.
* Run a workflow in simulation mode, view in Tui/Gui, you should now see simulated jobs.
* Run a regular workflow, check the job information in Tui/Gui matches expectations.

**Question:**

This PR currently relies on the receipt of the "submitted" message. In extreme cases I think it might be possible for the "started" (or even "succeeded") message to be received *before* the "submitted" message? Can't remember if we built some logic in to protect against this? Note that the messages are coming from different ZMQ clients so re cannot rely on any ZMQ message order guarantees.

Should we use some sort of flag or check the data store manually for all messages to cover for this eventuality or is there another layer of protection **(y/n)**?

**Requirements check-list**
  `conda-environment.yml`.

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -76,7 +76,6 @@ from cylc.flow.task_job_logs import JOB_LOG_OPTS, get_task_job_log
 from cylc.flow.task_proxy import TaskProxy
 from cylc.flow.task_state import (
     TASK_STATUS_WAITING,
-    TASK_STATUS_PREPARING,
     TASK_STATUS_SUBMITTED,
     TASK_STATUS_SUBMIT_FAILED,
     TASK_STATUS_RUNNING,
@@ -146,7 +145,6 @@ DELTAS_MAP = {
 DELTA_FIELDS = {DELTA_ADDED, DELTA_UPDATED, DELTA_PRUNED}
 
 JOB_STATUSES_ALL = [
-    TASK_STATUS_PREPARING,
     TASK_STATUS_SUBMITTED,
     TASK_STATUS_SUBMIT_FAILED,
     TASK_STATUS_RUNNING,
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -997,7 +997,7 @@ class DataStoreMgr:
         elif child_fam not in fp_parent.child_families:
             fp_parent.child_families.append(child_fam)
 
-    def insert_job(self, name, point_string, job_conf):
+    def insert_job(self, name, point_string, status, job_conf):
         """Insert job into data-store.
 
         Args:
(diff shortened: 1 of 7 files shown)


cylc/flow/data_store_mgr.py
```diff
@@ -1025,30 +1023,30 @@ def insert_job(self, name, point_string, job_conf):
...
-            post_script=job_conf['post-script'],
-            pre_script=job_conf['pre-script'],
-            script=job_conf['script'],
-            work_sub_dir=job_conf['work_d'],
             name=tproxy.name,
             cycle_point=tproxy.cycle_point,
-            directives=json.dumps(job_conf['directives']),
-            environment=json.dumps(job_conf['environment']),
-            param_var=json.dumps(job_conf['param_var'])
+            job_runner_name=job_conf.get('job_runner_name'),
```
oliver-sanders, line 1030: Allow these fields to be `None` for "prep-failed"-like jobs which did not get far enough through the process for these fields to be populated.

tests/functional/remote/05-remote-init.t
```diff
@@ -48,10 +45,10 @@ sqlite3 "${DB_FILE}" \
      FROM task_jobs ORDER BY name' \
     >"${NAME}"
 cmp_ok "${NAME}" <<__SELECT__
-a|1||
-b|1||
```
oliver-sanders, line 52: The database now contains the platform for "pre-failed" jobs.

tests/functional/remote/05-remote-init.t
```diff
@@ -24,9 +24,6 @@ export REQUIRE_PLATFORM='loc:remote fs:indep comms:tcp'
 set_test_number 6
 create_test_global_config "" "
 [platforms]
-    [[ariel]]
-        hosts = ${CYLC_TEST_HOST}
-        install target = ${CYLC_TEST_INSTALL_TARGET}
```
oliver-sanders, line 29: > **Note:**
>
> This form of platform definition is not safe for tests as it overrides
> any platform config set in `global-tests.cylc` which may be responsible
> for configuring the task communications method, for example.

dwsutherland (approved):
Looks good to me 👍 

Tested in and out simulation mode via TUI and WUI
Screenshot from 2021-11-03 16-34-00

Screenshot from 2021-11-03 16-36-35

hjoliver:
> Question:
> 
> This PR currently relies on the receipt of the "submitted" message. In extreme cases I think it might be possible for the "started" (or even "succeeded") message to be received before the "submitted" message? Can't remember if we built some logic in to protect against this? Note that the messages are coming from different ZMQ clients so re cannot rely on any ZMQ message order guarantees.
> 
> Should we use some sort of flag or check the data store manually for all messages to cover for this eventuality or is there another layer of protection (y/n)?

We definitely have some logic to handle misordering of task messages, but I don't recall if that particular case was covered. Will have to delve into it. Might help to contrive a test case that does it, and see what happens.

cylc/flow/task_events_mgr.py
```diff
@@ -1000,27 +1005,70 @@ def _process_message_submitted(self, itask, event_time):
...
-                self.setup_event_handlers(
-                    itask, self.EVENT_SUBMITTED, f'job {self.EVENT_SUBMITTED}')
-                self.data_store_mgr.delta_task_state(itask)
-                self.data_store_mgr.delta_task_queued(itask)
-            self._reset_job_timers(itask)
+            self.reset_inactivity_timer_func()
+            if itask.state.status == TASK_STATUS_PREPARING:
+                # The job started message can (rarely) come in before the
+                # submit command returns - in which case do not go back to
+                # 'submitted'.
```
hjoliver, line 1026: OK, if this happens, then this branch does the wrong thing. It doesn't create job proxies until the submit message is received, at which point the job stays in the submitted state instead of running. But, time being of the essence right now!, I will approve and merge this and post a follow-up fix.

hjoliver (approved):

cylc/flow/task_events_mgr.py
```diff
@@ -1000,27 +1005,70 @@ def _process_message_submitted(self, itask, event_time):
...
+                    )
+                    self.data_store_mgr.delta_task_state(itask)
+                    self.data_store_mgr.delta_task_queued(itask)
+                self._reset_job_timers(itask)
+
+        # register the newly submitted job with the data base and data store
+        self._insert_task_job(itask, event_time, submit_num, 0)
+
+    def _insert_task_job(self, itask, event_time, submit_num, submit_status):
+        job_conf = itask.jobs[submit_num - 1]
```
hjoliver, line 1042: This a bug, see comment on my follow-up PR.

PR #4487, opened 2021-11-01
Trigger mutation fix

hjoliver:
This is a small change with no associated Issue.

Fix bug in manual triggering, recently introduced by #4300.

Default triggering (i.e. with no reflow) is failing from the GUI and TUI, because missing reflow args (which are not needed for this use case) were not optional.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.py` and  `conda-environment.yml`. (None)
- [x] Already covered by existing tests.
- [x] Does not need tests (would need end-to-end testing?).
- [x] No change log entry required (fix for unreleased bug).
- [x] No documentation update required.

PR #4488, opened 2021-11-02
set-outputs mutation fix

hjoliver:

Similar to #4487, but for `cylc set-outputs` and equivalent via the GUI.

**Requirements check-list**
- [x] No change log entry required (fix for unreleased bug).

### DIFF
--- a/cylc/flow/network/resolvers.py
+++ b/cylc/flow/network/resolvers.py
@@ -662,7 +662,10 @@ class Resolvers(BaseResolvers):
             return (False, 'Edge distance cannot be negative')
 
     def force_spawn_children(
-        self, tasks: Iterable[str], outputs: Iterable[str], flow_num: int
+        self,
+        tasks: Iterable[str],
+        outputs: Optional[Iterable[str]] = None,
+        flow_num: Optional[int] = None
     ) -> Tuple[bool, str]:
         """Spawn children of given task outputs.
 
--- a/cylc/flow/network/resolvers.py
+++ b/cylc/flow/network/resolvers.py
@@ -674,7 +674,7 @@ class Resolvers(BaseResolvers):
         Args:
             tasks: List of identifiers, see `task globs`
             outputs: List of outputs to spawn on
-            flow_num: Flow number to attribute the outputs.
+            flow_num: Flow number to attribute the outputs
         """
         self.schd.command_queue.put(
             (
--- a/cylc/flow/network/resolvers.py
+++ b/cylc/flow/network/resolvers.py
@@ -672,9 +672,9 @@ class Resolvers(BaseResolvers):
         User-facing method name: set_outputs.
 
         Args:
-            tasks: List of identifiers, see `task globs`
-            outputs: List of outputs to spawn on
-            flow_num: Flow number to attribute the outputs
+            tasks: List of identifiers or task globs.
+            outputs: List of outputs to spawn on.
+            flow_num: Flow number to attribute the outputs.
         """
         self.schd.command_queue.put(
             (
@@ -729,7 +729,7 @@ class Resolvers(BaseResolvers):
 
         Args:
             tasks (list):
-                List of identifiers, see `task globs`_
+                List of identifiers or task globs.
             reflow (bool):
                 Start new flow from triggered tasks.
             flow_descr (str):
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1119,7 +1119,7 @@ class TaskPool:
         """
         return self.abort_task_failed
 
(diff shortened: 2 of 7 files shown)


cylc/flow/task_pool.py
```diff
@@ -1119,7 +1119,7 @@ def check_abort_on_task_fails(self):
         """
         return self.abort_task_failed
 
-    def spawn_on_output(self, itask, output):
+    def spawn_on_output(self, itask, output, forced=False):
```
MetRonnie, line 1122: Would be good to have docstring explaining the args

cylc/flow/task_pool.py
```diff
@@ -1387,21 +1389,33 @@ def match_taskdefs(
                     continue
         return n_warnings, task_items
 
-    def force_spawn_children(self, items, outputs, flow_num):
-        """Spawn downstream children of given task outputs on user command.
+    def force_spawn_children(
+        self,
+        items: Iterable[str],
+        outputs: Optional[List[str]] = None,
+        flow_num: Optional[int] = None
```
MetRonnie, line 1396: Again, would be nice to explain args in the docstring

MetRonnie (approved):

hjoliver:
Failed test fixed by #4495

hjoliver:
> Similar to #4487, but for cylc set-outputs a

Note this PR also adds a missing feature that I'd forgotten wasn't working yet (and a test for it): `cylc set-outputs` needs to  be able to spawn a downstream children without reflow, as well as with reflow.  But it works like this:

`cylc set-outputs <workflow> foo.1`  

The above command creates a transient task proxy `foo.1` purely in order to spawn its children. But in normal (non-manual) spawning, a task does not spawn if it has no flow numbers. For the manual case with no reflow, the transient task needs to be able able to spawn its own children (even though it has no flow number) but the children should have no flow number so that they don't spawn further.  That's what was missing.

cylc/flow/network/resolvers.py
```diff
@@ -671,7 +674,7 @@ def force_spawn_children(
         Args:
             tasks: List of identifiers, see `task globs`
             outputs: List of outputs to spawn on
-            flow_num: Flow number to attribute the outputs.
+            flow_num: Flow number to attribute the outputs
```
wxtim, line 677: [Pernickity] IMO they should _all_ have full-stops. I've double checked and they do in the G**gl* style guide.
```suggestion
            tasks: List of identifiers, see `task globs`.
            outputs: List of outputs to spawn on.
            flow_num: Flow number to attribute the outputs.
```

cylc/flow/task_pool.py
```diff
@@ -1119,17 +1119,25 @@ def check_abort_on_task_fails(self):
...
+            - update existing children but don't spawn new ones
+            - unless forced (manual command): spawn but with no flow number
 
         If an absolute output is completed update the store of completed abs
         outputs, and update the prerequisites of every instance of the child
         in the pool. (And in self.spawn() use the store of completed abs
         outputs to satisfy any tasks with abs prerequisites).
 
+        Args:
+            tasks: List of identifiers, see `task globs`
```
wxtim, line 1137: Where am I looking to see about task globs?
Is this markdown or rst?
- If RST use `` or a full `:ref:`

cylc/flow/task_pool.py
```diff
@@ -1119,17 +1119,25 @@ def check_abort_on_task_fails(self):
...
+            outputs: List of outputs to spawn on
+            flow_num: Flow number to attribute the outputs
+            forced: whether this is a manual spawn command or not
```
wxtim, line 1140:
```suggestion
            forced: If true this is a manual spawn command.
```

wxtim (approved):
- [x] Read
- [x] Played with
- [ ] Tests passing

Some Grammatical questions - take or leave, I don't think I need to re-review based on your decision before you merge.

You probably want to check why the tests are failing on MacOS. I gave them a kick, just in case that sorts them, but Shellchecks grievance looks legit. I'd be interested to know if you find out why they fail, but only on MacOS.

cylc/flow/network/resolvers.py
hjoliver: (Huh, your "suggestion" covers deleted lines too, so I can't commit it directly from GitHub UI)

cylc/flow/task_pool.py
hjoliver: Haha, good question - I pasted that line from another method with the same arg list.  I'll take a look.

hjoliver:
Addressed feedback and rebased (both of which only involved minor changes to docstrings, so I'll merge this).

PR #4497, opened 2021-11-08
Forbid case where both `suite.rc` and `flow.cylc` files are found in src/run dir

datamel:
Currently if you have both `flow.cylc` and `suite.rc` files in the run directory and cylc play. You get a log message warning that `suite.rc` has been detected and back compatibility mode has been activated. The workflow proceeds to run ignoring the `suite.rc` file and using the `flow.cylc` file.

This could cause some confusion for users and so we should raise an error if both files are detected (discussed UK end), forcing the user to decide which one should be used, rather than assuming either way.
 
Ideally this would be included in upcoming beta release, apologies, I understand this is frozen.

This is a small change with no associated Issue, bit of a quick fix. Looking at the code, I think there is potential for some refactoring but that would be a bigger job for another day.

To test on master:
```
$ tree `~/cylc-src/workflow`
├── flow.cylc
└── suite.rc
$ cylc install
Installed....
```
Then run `cylc play` on newly installed workflow.
Deprecation warnings occur and back compat mode is enabled, `flow.cylc` is used.

On this branch:

Cylc install/reinstall will raise an error.
Cylc play on a workflow containing `flow.cylc` and `suite.rc` will also raise error.

  `conda-environment.yml`.
- [x] (master branch) I have opened a documentation PR at https://github.com/cylc/cylc-doc/pull/313

### DIFF
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -1705,7 +1705,8 @@ def detect_both_flow_and_suite(path):
     """
     msg = (f"Both {WorkflowFiles.FLOW_FILE} and {WorkflowFiles.SUITE_RC} "
            "files are present in the run directory. Please remove one and"
-           " try again.")
+           " try again. For more information visit: https://cylc.github.io/"
+           "cylc-doc/latest/html/7-to-8/summary.html#backward-compatibility")
     if path.resolve().name == WorkflowFiles.SUITE_RC:
         flow_cylc = path.parent / WorkflowFiles.FLOW_FILE
         if flow_cylc.is_file() and not flow_cylc.is_symlink():
@@ -1767,7 +1768,9 @@ def check_flow_file(
             raise WorkflowFilesError(
                 f"Both {WorkflowFiles.FLOW_FILE} and "
                 f"{WorkflowFiles.SUITE_RC} files are present in the source "
-                "directory. Please remove one and try again."
+                "directory. Please remove one and try again. For more "
+                "information visit: https://cylc.github.io/cylc-doc/latest/"
+                "html/7-to-8/summary.html#backward-compatibility"
             )
         if flow_file_path.resolve() == suite_rc_path.resolve():
             # A symlink that points to existing suite.rc
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -1220,7 +1220,7 @@ def parse_reg(reg: str, src: bool = False, warn_depr=True) -> Tuple[str, Path]:
                 f"Workflow name must refer to a directory, not a file: {reg}"
             )
         abs_path, reg = infer_latest_run(abs_path)
-
+    detect_both_flow_and_suite(abs_path)
     check_deprecation(abs_path, warn=warn_depr)
     return (str(reg), abs_path)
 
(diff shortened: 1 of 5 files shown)


cylc/flow/workflow_files.py
```diff
@@ -1698,6 +1698,24 @@ def get_run_dir_info(
     return relink, run_num, rundir
 
 
+def detect_both_flow_and_suite(path):
+    """Detects if both suite.rc and flow.cylc are in directory.
+    Raises:
+        WorkflowFilesError: If both flow.cylc and suite.rc are in directory
+    """
+    msg = (f"Both {WorkflowFiles.FLOW_FILE} and {WorkflowFiles.SUITE_RC} "
```
wxtim, line 1706: IMO a link to https://cylc.github.io/cylc-doc/latest/html/7-to-8/summary.html#backward-compatibility would be helpful.
I've asked Element whether we can use a link shortener.

datamel: Happy to add a link. I shall be making a doc PR to make this change clear to users too.
I personally would prefer not to use link shortener. I think getting the cylc url in there is good, sometimes I think link shorteners look a bit spammy, as always happy to be overruled though!

datamel:
@hjoliver, I have tagged you as a reviewer, more to make you aware of the issue (and confirmation of suitability for freezed 8.0b3). If you are happy with the combo of `flow.cylc` and `suite.rc` being forbidden in run dirs and source dirs, we can get the code reviewed UK end if it would ease your time pressures.

cylc/flow/workflow_files.py
```diff
@@ -1698,6 +1698,24 @@ def get_run_dir_info(
...
+           "files are present in the run directory. Please remove one and"
+           " try again.")
+    if path.resolve().name == WorkflowFiles.SUITE_RC:
+        flow_cylc = path.parent / WorkflowFiles.FLOW_FILE
+        if flow_cylc.is_file() and not flow_cylc.is_symlink():
+            raise WorkflowFilesError(msg)
+    elif (path / WorkflowFiles.SUITE_RC).is_file():
+        flow_cylc = path / WorkflowFiles.FLOW_FILE
+        if flow_cylc.is_file() and not flow_cylc.is_symlink():
+            raise WorkflowFilesError(msg)
```
wxtim, line 1716:
```suggestion
    if path.resolve().name == WorkflowFiles.SUITE_RC:
        flow_cylc = path.parent / WorkflowFiles.FLOW_FILE
    elif (path / WorkflowFiles.SUITE_RC).is_file():
        flow_cylc = path / WorkflowFiles.FLOW_FILE

    if flow_cylc.is_file() and not flow_cylc.is_symlink():
        raise WorkflowFilesError(msg)
```

MetRonnie:
Is there any need for us to continue creating a flow.cylc symlink to suite.rc? I think it would simplify things if we stopped doing that.

cylc/flow/workflow_files.py
```diff
@@ -1744,7 +1762,13 @@ def check_flow_file(
     suite_rc_path = Path(expand_path(path), WorkflowFiles.SUITE_RC)
     if flow_file_path.is_file():
         if not flow_file_path.is_symlink():
-            return flow_file_path
+            if not suite_rc_path.is_file():
+                return flow_file_path
+            raise WorkflowFilesError(
+                f"Both {WorkflowFiles.FLOW_FILE} and "
```
wxtim, line 1768: If you agree about the link, thenh here too.

cylc/flow/workflow_files.py
wxtim: Discussed on phone: This doesn't work because flow_cylc is only defined inside the first if..else loop.

wxtim (approved):
- [x] Code looks good.
- [x] Tested as working.

datamel:
> Is there any need for us to continue creating a flow.cylc symlink to suite.rc? I think it would simplify things if we stopped doing that.

I will remove this in follow up issue.

hjoliver:
> Ideally this would be included in upcoming beta release, apologies, I understand this is frozen.

No problem, newly discovered bugs that can be quickly fixed are fair game.

hjoliver (commented):
Looks good, tested as working, but could you make a quick test for the `detect_both_flow_files()` function?

(Or could the two separate checks be made to use the same logic?)

MetRonnie:
I have created a question issue for stopping symlinking: #4500

datamel:
> I have created a question issue for stopping symlinking: #4500

Thanks @MetRonnie, I've got a branch with it removed, just waiting for tests to pass and I'll open the PR.

hjoliver (approved):
All good, thanks @datamel

PR #4501, opened 2021-11-09
job: don't use $HOME to source the job script

oliver-sanders:
The archer2 machine does not have access to `$HOME` on the compute nodes.

On the whole we have handled this use case ok, however, one sticking point remains.

Because Cylc 8 sources the job script from the `cylc-run` directory jobs cannot run in this environment at the moment:

```bash
. "$HOME/cylc-run/two/.service/etc/job.sh"
```

This PR uses `$0` to locate the job script rather than `$HOME` to avoid this issue.

**Requirements check-list**
  `conda-environment.yml`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,6 +58,9 @@ Fourth beta release of Cylc 8.
 
 ### Enhancements
 
+[#4501](https://github.com/cylc/cylc-flow/pull/4501) -
+Allow job scripts to run in environments without access to `$HOME`.
+
 [#4367](https://github.com/cylc/cylc-flow/pull/4367) -
 Make the central wrapper work with arbitrary virtual environment names.
 
--- a/cylc/flow/job_file.py
+++ b/cylc/flow/job_file.py
@@ -338,6 +338,11 @@ class JobFileWriter:
     @staticmethod
     def _write_epilogue(handle, job_conf, run_d):
         """Write epilogue."""
-        handle.write(f'\n\n. "{run_d}/.service/etc/job.sh"\ncylc__job__main')
+        # NOTE: don't use $HOME to locate the job.sh file as some systems might
+        # not have acess to $HOME on compute nodes
+        handle.write(
+            '\n\n. "$(dirname "$0")/../../../../../.service/etc/job.sh"'
+            '\ncylc__job__main'
+        )
         handle.write("\n\n%s%s\n" % (
             JobRunnerManager.LINE_PREFIX_EOF, job_conf['job_d']))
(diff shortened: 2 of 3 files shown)


oliver-sanders:
Reverting to draft as `CYLC_WORKFLOW_RUN_DIR` is currently defined using `$HOME` so would require the same treatment.

Whilst this diff would do the job:

```diff
diff --git a/cylc/flow/job_file.py b/cylc/flow/job_file.py
index d59c73864..c983f809f 100644
--- a/cylc/flow/job_file.py
+++ b/cylc/flow/job_file.py
@@ -207,7 +207,12 @@ class JobFileWriter:
 
         handle.write('\n')
         # override and write task-host-specific workflow variables
-        handle.write('\n    export CYLC_WORKFLOW_RUN_DIR="%s"' % run_d)
+        handle.write(
+            # NOTE: don't use $HOME to define the CYLC_WORKFLOW_RUN_DIR as some
+            # systems might not have acess to $HOME on compute nodes
+            '\n    export CYLC_WORKFLOW_RUN_DIR="%s"' %
+            '"$(dirname "$0")/../../../../../../"'
+        )
         handle.write(
             '\n    export CYLC_WORKFLOW_UUID="%s"' % job_conf['uuid_str'])
```

It would leave `../../../../../../` unexpanded in that var which is undesirable. Will rethink tomorrow unless anyone beats me to it.

hjoliver:
> It would leave ../../../../../../ unexpanded in that var which is undesirable.

I haven't thought of a way around that. It doesn't look great but isn't really a problem.

oliver-sanders:
Dave has found some issues with this approach to do with relative paths in symlinks and `$0` *not* being the job script path for slurm.

Note: at present this doesn't handle hierarchical workflows either.

oliver-sanders:
Update: We are currently considering making the `global init-script` run *before* the sourcing of `job.sh` to allow `$HOME` to be set.

This comes with the following caveats:

1. The `global init-script` will no longer be covered by error trapping.
2. The job environment will no longer be available to the `global init-script`.
3. In debug mode the `global init-script` will not be included in xtrace output.

These are fairly small issues that we can document. The `global init-script` is unlikely to be widely used and there are plenty of other `*-script` items to which the above caveats do not apply which can be used for regular purposes.

hjoliver:
That's a good solution IMO :+1:

oliver-sanders:
Good, closing this PR, leaving this issue in its place - https://github.com/cylc/cylc-flow/issues/4520

PR #4504, opened 2021-11-10
actions: manylinux - bump the timeout

oliver-sanders:


MetRonnie (approved):

MetRonnie (commented):
A useful change, while we're here, would be to add the manylinux yml filepath to the `on:` field so it gets triggered when you edit it

MetRonnie (approved):

Issue #4500, opened 2021-11-09
Stop symlinking flow.cylc to suite.rc?


MetRonnie:
With the introduction of Cylc 7 backward compatibility mode, there is little point in creating a flow.cylc symlink to suite.rc. In fact, it has disadvantages:
- Users will have to delete the symlink before renaming suite.rc to flow.cylc
- Unnecessary complexity in the codebase to create/deal with the symlink

Should we get rid of the symlinking? Possible disadvantages to doing this:
- Cylc more or less works as-is
- We might have to account for the existence of the symlink in workflows run before on a previous Cylc 8 beta

However, if we don't remove symlinking, it's likely we'll be stuck with it for Cylc 8. It may be safer to remove it now in case any unforeseen consequences emerge later?

**Pull requests welcome!**

hjoliver:
I vote to remove it.

PR #4506, opened 2021-11-10
Remove the suite.rc flow.cylc symlink

datamel:
Removes the `flow.cylc` symlink created when using a `suite.rc` file. This also enabled a refactor of duplicate logic from https://github.com/cylc/cylc-flow/pull/4497.
These changes close https://github.com/cylc/cylc-flow/issues/4500

Milestone is 8.0rc1 to give time for any unforeseen problems to surface.

**Requirements check-list**
  `conda-environment.yml`.
- [x] Already covered by existing tests (removed now redundant tests).
- [x] No documentation update required (I can not find reference in the docs to this symlink being created).

### DIFF
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -84,7 +84,6 @@ from cylc.flow.wallclock import get_current_time_string
 
 if TYPE_CHECKING:
     from optparse import Values
-    from logging import Logger
 
 
 class KeyType(Enum):
@@ -551,8 +550,8 @@ def get_flow_file(reg: str) -> str:
     Creates a flow.cylc symlink to suite.rc if only suite.rc exists.
     """
     run_dir = get_workflow_run_dir(reg)
-    check_flow_file(run_dir, symlink_suiterc=True)
-    return os.path.join(run_dir, WorkflowFiles.FLOW_FILE)
+    path = check_flow_file(run_dir)
+    return path
 
 
 def get_workflow_source_dir(
@@ -678,7 +677,7 @@ def register(
         source = os.getcwd()
     # flow.cylc must exist so we can detect accidentally reversed args.
     source = os.path.abspath(source)
-    check_flow_file(source, symlink_suiterc=True, logger=None)
+    check_flow_file(source)
     if not is_installed(get_workflow_run_dir(workflow_name)):
         symlinks_created = make_localhost_symlinks(
             get_workflow_run_dir(workflow_name), workflow_name)
@@ -1297,7 +1296,7 @@ def _parse_src_reg(reg: Path, cur_dir_only: bool = False) -> Tuple[Path, Path]:
                         ))
                 else:
                     try:
-                        abs_path = check_flow_file(abs_path, logger=None)
+                        abs_path = check_flow_file(abs_path)
                     except WorkflowFilesError:
                         return (run_dir_reg, run_dir_path)
                     LOG.warning(REG_CLASH_MSG.format(
@@ -1587,7 +1586,7 @@ def reinstall_workflow(named_run, rundir, source, dry_run=False):
         reinstall_log.warning(
             f"An error occurred when copying files from {source} to {rundir}")
         reinstall_log.warning(f" Error: {stderr}")
-    check_flow_file(rundir, symlink_suiterc=True, logger=reinstall_log)
+    check_flow_file(rundir)
     reinstall_log.info(f'REINSTALLED {named_run} from {source}')
     print(f'REINSTALLED {named_run} from {source}')
     close_log(reinstall_log)
@@ -1691,7 +1690,7 @@ def install_workflow(
         install_log.warning(f" Warning: {stderr}")
     cylc_install = Path(rundir.parent, WorkflowFiles.Install.DIRNAME)
     check_deprecation(
-        check_flow_file(rundir, symlink_suiterc=True, logger=install_log)
+        check_flow_file(rundir)
     )
     if no_run_name:
         cylc_install = Path(rundir, WorkflowFiles.Install.DIRNAME)
@@ -1761,11 +1760,13 @@ def get_run_dir_info(
 
 def detect_both_flow_and_suite(path):
     """Detects if both suite.rc and flow.cylc are in directory.
+
+    Permits flow.cylc to be a symlink.
     Raises:
         WorkflowFilesError: If both flow.cylc and suite.rc are in directory
     """
     msg = (f"Both {WorkflowFiles.FLOW_FILE} and {WorkflowFiles.SUITE_RC} "
-           "files are present in the run directory. Please remove one and"
+           f"files are present in {path}. Please remove one and"
            " try again. For more information visit: https://cylc.github.io/"
            "cylc-doc/latest/html/7-to-8/summary.html#backward-compatibility")
     if path.resolve().name == WorkflowFiles.SUITE_RC:
@@ -1806,53 +1807,30 @@ def detect_flow_exists(
 
 def check_flow_file(
     path: Union[Path, str],
-    symlink_suiterc: bool = False,
-    logger: Optional['Logger'] = LOG
 ) -> Path:
-    """Raises WorkflowFilesError if no flow file in path sent.
+    """Checks the path for a suite.rc or flow.cylc file.
+
+    Raises:
+        WorkflowFilesError
+            - if no flow file in path sent
+            - both suite.rc and flow.cylc in path sent.
 
     Args:
         path: Absolute path to check for a flow.cylc and/or suite.rc file.
-        symlink_suiterc: If True and suite.rc exists, create flow.cylc as a
-            symlink to suite.rc. If a flow.cylc symlink already exists but
-            points elsewhere, it will be replaced.
-        logger: A custom logger to use to log warnings.
 
     Returns the path of the flow file if present.
     """
     flow_file_path = Path(expand_path(path), WorkflowFiles.FLOW_FILE)
     suite_rc_path = Path(expand_path(path), WorkflowFiles.SUITE_RC)
     if flow_file_path.is_file():
-        if not flow_file_path.is_symlink():
-            if not suite_rc_path.is_file():
-                return flow_file_path
-            raise WorkflowFilesError(
-                f"Both {WorkflowFiles.FLOW_FILE} and "
-                f"{WorkflowFiles.SUITE_RC} files are present in the source "
-                "directory. Please remove one and try again. For more "
-                "information visit: https://cylc.github.io/cylc-doc/latest/"
-                "html/7-to-8/summary.html#backward-compatibility"
-            )
-        if flow_file_path.resolve() == suite_rc_path.resolve():
-            # A symlink that points to existing suite.rc
-            return flow_file_path
-    if suite_rc_path.is_file():
-        if not symlink_suiterc:
-            return suite_rc_path
-        if flow_file_path.is_symlink():
-            # Symlink broken or points elsewhere - replace
-            flow_file_path.unlink()
-        flow_file_path.symlink_to(WorkflowFiles.SUITE_RC)
-        if logger:
-            logger.warning(
-                f"Symlink created: "
-                f"{WorkflowFiles.FLOW_FILE} -> {WorkflowFiles.SUITE_RC}"
-            )
+        detect_both_flow_and_suite(Path(path))
         return flow_file_path
+    if suite_rc_path.is_file():
+        return suite_rc_path
     raise WorkflowFilesError(NO_FLOW_FILE_MSG.format(path))
 
 
-def create_workflow_srv_dir(rundir=None, source=None):
+def create_workflow_srv_dir(rundir=None):
     """Create workflow service directory"""
 
     workflow_srv_d = rundir.joinpath(WorkflowFiles.Service.DIRNAME)
@@ -1882,7 +1860,7 @@ def validate_source_dir(source, workflow_name):
         raise WorkflowFilesError(
             f"{workflow_name} installation failed. Source directory "
             f"should not be in {cylc_run_dir}.")
-    check_flow_file(source, logger=None)
+    check_flow_file(source)
 
 
 def parse_cli_sym_dirs(symlink_dirs: str) -> Dict[str, Dict[str, Any]]:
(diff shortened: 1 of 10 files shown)


cylc/flow/workflow_files.py
```diff
@@ -1700,11 +1699,13 @@ def get_run_dir_info(
 
 def detect_both_flow_and_suite(path):
     """Detects if both suite.rc and flow.cylc are in directory.
+
+    Permits flow.cylc to be a symlink.
```
wxtim, line 1703: Is this desirable?

wxtim (approved):
- [x] Read -looks sensible.
- [x] Played with.

I had one question about the functionality - do we want to allow the symlink to pass?

cylc/flow/workflow_files.py
datamel: @oliver-sanders, I believe you said we should not stop users from manually making a symlink.

MetRonnie: Shouldn't we check that the symlink points to the right suite.rc, and raise an error if not?

cylc/flow/workflow_files.py
```diff
@@ -1745,53 +1746,30 @@ def detect_flow_exists(
...
-            )
+        detect_both_flow_and_suite(Path(path))
         return flow_file_path
+    if suite_rc_path.is_file():
+        return suite_rc_path
     raise WorkflowFilesError(NO_FLOW_FILE_MSG.format(path))
 
 
-def create_workflow_srv_dir(rundir=None, source=None):
+def create_workflow_srv_dir(rundir=None):
```
MetRonnie, line 1772: This would not pass Mypy
```suggestion
def create_workflow_srv_dir(rundir: Path) -> None:
```

tests/functional/cylc-install/00-simple.t
```diff
@@ -76,20 +75,6 @@ run_ok "${TEST_NAME}" cylc install --flow-name="${RND_WORKFLOW_NAME}" -C "${RND_
 contains_ok "${TEST_NAME}.stdout" <<__OUT__
 INSTALLED $RND_WORKFLOW_NAME/run1 from ${RND_WORKFLOW_SOURCE}
 __OUT__
```
MetRonnie, line 77: I'm not sure there is a test that the suite.rc deprecation message is printed on `cylc install`? If not, perhaps it should go here

cylc/flow/workflow_files.py
```diff
@@ -546,8 +545,8 @@ def get_flow_file(reg: str) -> str:
     Creates a flow.cylc symlink to suite.rc if only suite.rc exists.
     """
     run_dir = get_workflow_run_dir(reg)
-    check_flow_file(run_dir, symlink_suiterc=True)
-    return os.path.join(run_dir, WorkflowFiles.FLOW_FILE)
+    path = check_flow_file(run_dir)
+    return path
```
MetRonnie, line 549: Mypy flagged type error

cylc/flow/workflow_files.py
datamel: If there is a flow.cylc there it will use that, regardless of if there is a suite.rc file there. Do we want to forbid users from symlinking other places? I can't think of a reason to forbid, I am probably missing something! I think so long as it is a valid workflow (which will be picked up with other checks) then do we really care how they are sourcing the workflow, it is a bit messy but that is the user's choice? I can't think of a reason to forbid although I am possibly missing something!

MetRonnie: It sounds reasonable to me that users could have a flow.cylc that is a symlink to either its neighbour suite.rc or somewhere else. But if they have a flow.cylc that symlinks to somewhere else *and* a suite.rc in there, that sounds to me like it should fail in the same way as a normal flow.cylc file and a suite.rc in there.

datamel: Resolving this as there is a use case (at the MO) for users to source workflow from outside the run directory.

MetRonnie: Sorry Mel, not sure if you saw my latest comment above https://github.com/cylc/cylc-flow/pull/4506#discussion_r750305478

datamel: Ah I didn't see that, thanks @MetRonnie, mustn't have refreshed the page before posting.

cylc/flow/workflow_files.py
```diff
@@ -545,14 +544,14 @@ def get_contact_file(reg):
         get_workflow_srv_dir(reg), WorkflowFiles.Service.CONTACT)
 
 
-def get_flow_file(reg: str) -> str:
+def get_flow_file(reg: str) -> Path:
     """Return the path of a workflow's flow.cylc file.
 
     Creates a flow.cylc symlink to suite.rc if only suite.rc exists.
     """
```
MetRonnie, line 551:
```suggestion
    """Return the path of a workflow's flow.cylc file."""
```

tests/unit/test_workflow_files.py
```diff
@@ -1824,87 +1829,50 @@ def test_detect_both_flow_and_suite(tmp_path):
...
-            expected_file = WorkflowFiles.FLOW_FILE
-            if flow_file_target != WorkflowFiles.SUITE_RC:
-                log_msg = "Symlink created: flow.cylc -> suite.rc"
-        assert result == tmp_path.joinpath(expected_file)
-
-    if log_msg:
-        assert log_msg in caplog.messages
+def test_is_forbidden_symlink_returns_false_for_non_symlink(tmp_path):
+    """Test sending a non symlink path is not marked as forbidden"""
+    Path(tmp_path)
```
MetRonnie, line 1871:
```suggestion
```

tests/unit/test_workflow_files.py
```diff
@@ -1824,87 +1829,50 @@ def test_detect_both_flow_and_suite(tmp_path):
...
-        err: Type of exception if expected to get raised.
-        expected_file: Which file's path should get returned, when
-            symlink_suiterc_arg is FALSE (otherwise it will always be
-            flow.cylc, assuming no exception occurred).
-        symlink_suiterc_arg: Value of the symlink_suiterc arg passed to
-            check_flow_file().
-    """
+def test_detect_both_flow_and_suite_symlinked(tmp_path):
+    """Test flow.cylc symlinked to suite.rc together in dir is permitted."""
+    Path(tmp_path)
```
MetRonnie, line 1835: Unnecessary line
```suggestion
```

tests/unit/test_workflow_files.py
```diff
@@ -1824,87 +1829,50 @@ def test_detect_both_flow_and_suite(tmp_path):
...
-    tmp_path.joinpath('other-path').touch()
-    if suiterc_exists:
-        suiterc.touch()
-    if flow_file_target:
-        flow_file.symlink_to(flow_file_target)
+    flow_file.symlink_to(WorkflowFiles.SUITE_RC)
+    try:
+        detect_both_flow_and_suite(tmp_path)
+    except WorkflowFilesError:
+        pytest.fail("Unexpected WorkflowFilesError")
```
MetRonnie, line 1842: This will do?
```suggestion
    detect_both_flow_and_suite(tmp_path)
```

tests/unit/test_workflow_files.py
```diff
@@ -1824,87 +1829,50 @@ def test_detect_both_flow_and_suite(tmp_path):
...
+
+
+def test_flow_symlinked_elsewhere_and_suite_present(tmp_path):
+    """flow.cylc symlinked to suite.rc elsewhere, and suite.rc in dir raises"""
+    tmp_path = Path(tmp_path)
```
MetRonnie, line 1847:
```suggestion
def test_flow_symlinked_elsewhere_and_suite_present(tmp_path: Path):
    """flow.cylc symlinked to suite.rc elsewhere, and suite.rc in dir raises"""
```

tests/unit/test_workflow_files.py
```diff
@@ -1824,87 +1829,50 @@ def test_detect_both_flow_and_suite(tmp_path):
...
-            assert flow_file.samefile(suiterc)
-            expected_file = WorkflowFiles.FLOW_FILE
-            if flow_file_target != WorkflowFiles.SUITE_RC:
-                log_msg = "Symlink created: flow.cylc -> suite.rc"
-        assert result == tmp_path.joinpath(expected_file)
-
-    if log_msg:
-        assert log_msg in caplog.messages
+def test_is_forbidden_symlink_returns_false_for_non_symlink(tmp_path):
+    """Test sending a non symlink path is not marked as forbidden"""
```
MetRonnie, line 1870: Shouldn't a symlink also be not marked as forbidden?

datamel: Yes, thanks @MetRonnie. Small adjustment and test added. I hope all bases are covered this time!

cylc/flow/workflow_files.py
```diff
@@ -1759,23 +1753,50 @@ def get_run_dir_info(
...
+        flow.cylc and suite.rc in same directory but no symlink
+    Args:
+        flow_file : Absolute Path to the flow.cylc file
+    """
+    if not flow_file.is_symlink():
+        if flow_file.parent.joinpath(WorkflowFiles.SUITE_RC).exists():
+            return True
+        return False
+    link = flow_file.resolve()
+    if link.parent == flow_file.parent:
```
MetRonnie, line 1793: This would allow
```
myflow
|-- something.cylc
|-- flow.cylc -> something.cylc
`-- suite.rc
```
which is equivalent to having distinct `flow.cylc` and `suite.rc` files in the same run dir.

Perhaps it should be something like
```python
if flow_file.resolve() == flow_file.parent / WorkflowFiles.SUITE_RC:
```

tests/unit/test_workflow_files.py
```diff
@@ -1817,10 +1818,12 @@ def test_check_flow_file(
...
-    """Test flow.cylc and suite.rc together in dir raises error."""
+    """Test flow.cylc and suite.rc (as files) together in dir raises error."""
     tmp_path.joinpath(WorkflowFiles.FLOW_FILE).touch()
     tmp_path.joinpath(WorkflowFiles.SUITE_RC).touch()
 
+    forbidden = is_forbidden(tmp_path / WorkflowFiles.FLOW_FILE)
+    assert forbidden is True
     with pytest.raises(WorkflowFilesError) as exc:
         detect_both_flow_and_suite(tmp_path)
         assert str(exc.value) == (
```
MetRonnie, line 1829: Just spotted, this `assert` will never be run, because `detect_both_flow_and_suite()` raises and so it will exit out of the `with` block at the point. The `assert` needs to be un-indented

tests/unit/test_workflow_files.py
```diff
@@ -1832,87 +1835,45 @@ def test_detect_both_flow_and_suite(tmp_path):
...
+    assert forbidden is True
+    with pytest.raises(WorkflowFilesError) as exc:
+        detect_both_flow_and_suite(run_dir)
+        assert str(exc.value) == (
+            "Both flow.cylc and suite.rc files are present in the "
+            "source directory. Please remove one and try again. "
+            "For more information visit: "
+            "https://cylc.github.io/cylc-doc/latest/html/7-to-8/summary.html"
+            "#backward-compatibility"
+        )
```
MetRonnie, line 1868: As above (re: indentation) + simplify
```suggestion
    assert str(exc.value).startswith(
        "Both flow.cylc and suite.rc files are present in the "
        "source directory. Please remove one and try again. 
    )
```

cylc/flow/workflow_files.py
datamel: This case (I think) is covered by the following condition: `if flow_file.parent.joinpath(WorkflowFiles.SUITE_RC).exists():`.

datamel:
@wxtim I've made some logic changes since your approval, sorry! So requesting re-review.

cylc/flow/workflow_files.py
MetRonnie: It would be, but it will `return False` before getting to that

cylc/flow/workflow_files.py
```diff
@@ -1759,23 +1753,50 @@ def get_run_dir_info(
...
+        # link points within dir (permitted)
+        return False
+    # link points elsewhere, check that suite.rc does not also exist in dir
+    if flow_file.parent.joinpath(WorkflowFiles.SUITE_RC).exists():
+        return True
+    return False
```
MetRonnie, line 1799:
```suggestion
    return flow_file.resolve() != flow_file.parent / WorkflowFiles.SUITE_RC
```

wxtim: Sorry @datamel  _ I got confused and thought this was my PR.

tests/unit/test_workflow_files.py
```diff
@@ -1817,102 +1818,58 @@ def test_check_flow_file(
...
-        if symlink_suiterc_arg is True:
-            assert flow_file.samefile(suiterc)
-            expected_file = WorkflowFiles.FLOW_FILE
-            if flow_file_target != WorkflowFiles.SUITE_RC:
-                log_msg = "Symlink created: flow.cylc -> suite.rc"
-        assert result == tmp_path.joinpath(expected_file)
-
-    if log_msg:
-        assert log_msg in caplog.messages
+def test_is_forbidden_symlink_returns_false_for_non_symlink(tmp_path):
```
wxtim, line 1867: Can we test some scenarios where `is_forbidden` → `True`?

cylc/flow/workflow_files.py
datamel: Sorry for my confusion! All testing for me was raising the error....
image
The return False does not return before that (where it should!) because `flow_file.parent` needed a `resolve()` on the end to hit the false.
Will go with your suggestion (adding a resolve()), thanks @MetRonnie

cylc/flow/workflow_files.py
datamel: Hehe no probs @wxtim, thought I had totally lost it there, twas blaming sleep deprivation.

tests/unit/test_workflow_files.py
datamel: There are currently 2 tests covering this scenario, `is forbidden → True`. I think that the two that are there are sufficient but let me know if you'd like more added.

cylc/flow/workflow_files.py
datamel: Ok, coming back to this (confusion after your suggestion was added by Tim so it wasn't visible to me). This suggestion will forbid symlinking elsewhere. 
Going to push a change which I think now covers all scenarios, based on your `if flow_file.resolve() == flow_file.parent / WorkflowFiles.SUITE_RC:` suggestion.

cylc/flow/workflow_files.py
MetRonnie: Oops, doesn't help that I meant `return False` on the last line

MetRonnie (approved):
The code could probably be refactored to simplify it, but that can wait.

PR #4511, opened 2021-11-15
Fix clock-trigger offset computation.

### LINKED ISSUE #4510: wall_clock xtrigger offset bug
The `wall_clock` xtrigger offset computation is too simplistic, for offsets involving months and years. Turns out there are valid use cases for that: https://cylc.discourse.group/t/my-wall-clock-trigger-always-satisfied-for-large-offset/397

The offset is converted to seconds without context, then added to the cycle point (unix epoch seconds). We need to do the datetime computation (cycle point + offset) before converting to seconds.

hjoliver:
These changes close #4510

Fixes clock-trigger time computation for clock xtriggers and old-style (task property) clock triggers. Also restore main-loop call to check old-style clock-triggers (must have lost that somewhere during Cylc 8 development).

The new way is very efficient, even for inexact offsets (months, years): the computation is done once at first call (add offset to cycle point, then convert to absolute trigger time in seconds since unix epoch which is used for all subsequent checks.

~~TODO:~~
- [x] ~~is there a less convoluted way to do the offset computation?~~
- [x] ~~this is probably quite inefficient compared to the old broken way? (complex ISO datetime computation in every clock check, instead of a simple float subtraction). Can this be mitigated?~~

**Requirements check-list**

### DIFF
--- a/cylc/flow/task_proxy.py
+++ b/cylc/flow/task_proxy.py
@@ -24,7 +24,6 @@ from typing import Any, Dict, List, Set, Tuple, Optional, TYPE_CHECKING
 
 from metomi.isodatetime.timezone import get_local_time_zone
 
-import cylc.flow.cycling.iso8601
 from cylc.flow import LOG
 from cylc.flow.cycling.loader import standardise_point_string
 from cylc.flow.exceptions import PointParsingError
@@ -34,6 +33,11 @@ from cylc.flow.task_action_timer import TimerFlags
 from cylc.flow.task_state import TaskState, TASK_STATUS_WAITING
 from cylc.flow.taskdef import generate_graph_children
 from cylc.flow.wallclock import get_unix_time_from_time_string as str2time
+from cylc.flow.cycling.iso8601 import (
+    point_parse,
+    interval_parse,
+    ISO8601Interval
+)
 
 if TYPE_CHECKING:
     from cylc.flow.cycling import PointBase
@@ -47,6 +51,7 @@ class TaskProxy:
     Attributes:
         .clock_trigger_time:
             Clock trigger time in seconds since epoch.
+            (Used for both old-style clock triggers and wall_clock xtrigger).
         .expire_time:
             Time in seconds since epoch when this task is considered expired.
         .identity:
@@ -267,7 +272,7 @@ class TaskProxy:
     @staticmethod
     def get_offset_as_seconds(offset):
         """Return an ISO interval as seconds."""
-        iso_offset = cylc.flow.cycling.iso8601.interval_parse(str(offset))
+        iso_offset = interval_parse(str(offset))
         return int(iso_offset.get_seconds())
 
     def get_late_time(self):
@@ -285,8 +290,7 @@ class TaskProxy:
     def get_point_as_seconds(self):
         """Compute and store my cycle point as seconds since epoch."""
         if self.point_as_seconds is None:
-            iso_timepoint = cylc.flow.cycling.iso8601.point_parse(
-                str(self.point))
+            iso_timepoint = point_parse(str(self.point))
             self.point_as_seconds = int(iso_timepoint.get(
                 'seconds_since_unix_epoch'))
             if iso_timepoint.time_zone.unknown:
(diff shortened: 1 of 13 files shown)


cylc/flow/xtriggers/wall_clock.py
```diff
@@ -35,16 +36,23 @@ def wall_clock(offset=None, absolute_as_seconds=None, point_as_seconds=None):
...
+        if not offset:
+            offset = 'PT0S'
+        trigger_time = int(
+            point_parse(
+                str(
+                    get_point_relative(
+                        offset, point
+                    )
+                )
+            ).get('seconds_since_unix_epoch')
```
MetRonnie, line 55: > is there a less convoluted way to do the offset computation?

I am not sure about the need to convert from metomi-isodatetime objects to Cylc date-time objects and back
https://github.com/cylc/cylc-flow/blob/d8d797b2ba42e5a5e1307e794d69721623db30ec/cylc/flow/cycling/iso8601.py#L866-L876

Can this not be handled by metomi-isodatetime alone, by adding the `Duration` to the `TimePoint`?

> this is probably quite inefficient compared to the old broken way? (complex ISO datetime computation in every clock check, instead of a simple float substraction). Can this be mitigated?

Unfortunately I think we must maintain the correct inexact units (years & months) if used. But I suppose an optimisation can be made if years or months are not in the offset: because all the other units are exact[^1], you could convert to seconds since unix epoch before doing the addition/subtraction in that case, might be faster.

[^1]: Not strictly speaking true for ISO 8601, but correct for our isodatetime implementation. See https://github.com/metomi/isodatetime/blob/99d509813bcee8c9fd64523221ef47d45a0c4a72/README.md?plain=1#L282-L294

hjoliver: Fixed now, see updated description.

hjoliver:
Ready for review @wxtim. See updated description above.

tests/flakyfunctional/xtriggers/00-wall_clock/flow.cylc
```diff
@@ -1,10 +1,10 @@
 #!Jinja2
 
 # Test wall_clock xtrigger: workflow will run to completion and exit if the
-# clock trigger is not satisfied, else stall and abort.
+# clock trigger is not satisfied, else abort on inactivity.
 
 [scheduler]
-    cycle point time zone = '+0530'
+    # Default to time zone Z
```
hjoliver, line 7: (This test also fails on master, if system time is local in my timezone).

wxtim: I'm working my a home pc, so I can jimmy the system clock to test this.

cylc/flow/task_proxy.py
```diff
@@ -297,6 +301,26 @@ def get_point_as_seconds(self):
                 self.point_as_seconds += utc_offset_in_seconds
         return self.point_as_seconds
 
+    def get_clock_trigger_time(self, offset_str):
```
wxtim, line 304: This feels like it might be nice to unit test it: It would make fixing any future issues with this easier.

wxtim (approved):
- [x] Read
- [x] Had a good play with the branch checked out.
- [x] Tests passing

I did make a suggestion about adding a unit test, but I don't think it should block merging it if you want to do so.

tests/functional/clock-trigger-inexact/00-big-offset.t
```diff
@@ -0,0 +1,38 @@
...
+workflow_run_ok "${TEST_NAME_BASE}-run" cylc play --no-detach "${WORKFLOW_NAME}"
+
+cylc cat-log "${WORKFLOW_NAME}" > log 
+START_HOUR=$(grep 'Workflow:' log | cut -c 1-13)
+START_MINU=$(grep 'Workflow:' log | cut -c 15-16)
+TRIGG_MINU=$(( 10#${START_MINU} + 1))
+[[ $START_MINU == 0* ]] && TRIGG_MINU=0${TRIGG_MINU}
+
+for NAME in foo bar baz; do
+   grep_ok "${START_HOUR}:${TRIGG_MINU}.* INFO - \[${NAME}\..*] => waiting$" log
```
MetRonnie, line 35:
```suggestion
   grep_ok "${START_HOUR}:${TRIGG_MINU}.* INFO - \[${NAME}\..*\] => waiting$" log
```

MetRonnie (changes requested):
I agree with Tim about the unit test, I think it should be included before merging

hjoliver:
Unit test added.

tests/unit/test_task_proxy.py
```diff
@@ -46,6 +46,47 @@ def test_point_match(
...
+    offset_str: str,
+    expected: int,
+    set_cycling_type: Callable
+) -> None:
+    """Test get_clock_trigger_time() for exact and inexact offsets."""
+    mock_itask = Mock(
+        point=itask_point.standardise(),
+        clock_trigger_time=None
+    )
+    set_cycling_type(itask_point.TYPE)
```
MetRonnie, line 86: Nice test, passes when running the whole test file (`pytest tests/unit/test_task_proxy.py`) but it fails when I run it individually (`pytest tests/unit/test_task_proxy.py::test_get_clock_trigger_time`). Need to call the `set_cycling_type` fixture before `itask_point.standardise()`
```suggestion
    set_cycling_type(itask_point.TYPE)
    mock_itask = Mock(
        point=itask_point.standardise(),
        clock_trigger_time=None
    )
```

MetRonnie (approved):
Consider this approved once my suggestion for the test has been resolved

Issue #4294, opened 2021-07-07
mutations: log the authenticated user


oliver-sanders:
sibling: https://github.com/cylc/cylc-uiserver/issues/224

Since v8.0 cylc-flow is a single-user application, the cylc-uis provides an authorisation layer enabling multi-user functionality.

When mutations are actioned from the UIS we should log the mutation and the user who actions it at the UIS but also pass the user through as metadata and record it in the cylc-flow workflow log.

PR #3469, opened 2019-12-18
Api on the fly - GraphQL schema

oliver-sanders:
Addresses the data-provision component of #2123 

**Not ready for review, putting this up early as it effects downstream activities.**

Improve the GraphQL schema so that it contains all of the information required to auto-generate interfaces (e.g. HTML forms or CLI option parsers).

* Split the `TaskActions` mutation into separate mutations:
  * So we can auto-generate an interface for each mutation individually.
  * So we can document each mutation individually.
  * So we can assign options/arguments to each mutation individually.
* Combine the many `hold_*`, `release_*`, `broadcast_*` and `stop_*` methods
  * So they form a more logical single interface
  * (and also just for niceness)
  * Still need to re-implement:
    * [ ] expire broadcast
    * [ ] display broadcast (do we need a mutation for this, a query should do, but what about the CLI)?
* Create new input types for mutation arguments
  * So we know the meaning of fields and can handle them accordingly (e.g. validate and auto-complete cycle points in an HTML form).
  * These new types mostly just extend the built-in ones with no modifications, it's just for documentation and name-spacing.
* Use enumerations where applicable.
  * We can use these to auto-generate options lists client side.
  * Also permits centralisation of documentation.

This PR pushes mutations through the same `cylc.flow.network.server.SuiteRuntimeServer` interface as the CLI currently uses. This is temporary, hence the docstrings have been left unchanged, this interface will be removed partially/entirely when the CLI is migrated to use the mutations interface.

**Example Mutation:**

```graphql
mutation {
  stop(workflows: ["sanderso|generic"], mode: Now) {
    result
  }
}
```

**Caveats:**
* A little messy *but* much of the mess will get deleted when the CLI is refactored.
* Not quite complete *but* we can fill in the gaps later.
* Not quite perfectly documented *but* the documentation needs a review anyway.

**TODO:**
* [x] Standardise argument names.
* [x] Improve documentation.
* [x] ~Testing.~ Mutation testing will come for free once the CLI is converted, replicating these tests will be a lot of work.
* [x] Document enumerations.
* [x] Review required/optional arguments.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [ ] Appropriate tests are included (unit and/or functional).
- [ ] Appropriate change log entry included.
[...]

PR #4522, opened 2021-11-17
Log Authenticated User from UI-Server in Workflow Log

datamel:
Log the authenticated user executing the mutation command from the UI.

A test user pausing the workflow of another user, will result in a log message below in the workflow log:
image

The owner then resuming the workflow, will result in a log message..
image
 

These changes close https://github.com/cylc/cylc-flow/issues/4294 and https://github.com/cylc/cylc-uiserver/issues/224 by its sibling:

**Requirements check-list**
  `conda-environment.yml`.
- [x] Appropriate tests are _not_ included (unit and/or functional). - Only manually tested. This will require multi user testing setup to get this into the automated test battery.
- [x] No change log entry required (why? Logging change).

### DIFF
--- a/cylc/flow/network/authorisation.py
+++ b/cylc/flow/network/authorisation.py
@@ -17,8 +17,6 @@
 
 from functools import wraps
 
-from cylc.flow import LOG
-
 
 def authorise():
     """Add authorisation to an endpoint.
@@ -40,16 +38,6 @@ def authorise():
         def _call(self, *args, user='?', meta=None, **kwargs):
             if not meta:
                 meta = {}
-            host = meta.get('host', '?')
-            prog = meta.get('prog', '?')
-            comms_method = meta.get('comms_method', '?')
-
-            # Hardcoded, for new - but much of this functionality can be
-            # removed more swingingly.
-            LOG.debug(
-                '[client-command] %s %s://%s@%s:%s',
-                fcn.__name__, comms_method, user, host, prog
-            )
             return fcn(self, *args, meta=meta, **kwargs)
 
         return _call
--- a/cylc/flow/network/authorisation.py
+++ b/cylc/flow/network/authorisation.py
@@ -50,7 +50,7 @@ def authorise():
                 '[client-command] %s %s://%s@%s:%s',
                 fcn.__name__, comms_method, user, host, prog
             )
-            return fcn(self, *args, **kwargs)
+            return fcn(self, *args, meta=meta, **kwargs)
 
         return _call
     return wrapper
--- a/cylc/flow/network/client.py
+++ b/cylc/flow/network/client.py
@@ -37,6 +37,8 @@ from cylc.flow.network import (
     encode_,
     decode_,
     get_location,
+    get_authenticated_user,
+    set_authenticated_user,
     ZMQSocketBase
 )
 from cylc.flow.network.client_factory import CommsMeth
@@ -78,6 +80,8 @@ class WorkflowRuntimeClient(ZMQSocketBase):
 
             If both host and port are provided it is not necessary to load
             the contact file.
+        auth_user(str):
+            The authenticated user who has executed the command.
 
     Attributes:
         host (str):
@@ -123,7 +127,8 @@ class WorkflowRuntimeClient(ZMQSocketBase):
             port: int = None,
             context: object = None,
             timeout: Union[float, str] = None,
-            srv_public_key_loc: str = None
+            srv_public_key_loc: str = None,
+            auth_user: str = None
     ):
         super().__init__(zmq.REQ, context=context)
         self.workflow = workflow
@@ -143,8 +148,7 @@ class WorkflowRuntimeClient(ZMQSocketBase):
         self.poller = None
         # Connect the ZMQ socket on instantiation
         self.start(self.host, self.port, srv_public_key_loc)
-        # gather header info post start
-        self.header = self.get_header()
+        self.header = self.get_header(auth_user)
 
     def _socket_options(self):
         """Set socket options after socket instantiation before connect.
@@ -159,7 +163,13 @@ class WorkflowRuntimeClient(ZMQSocketBase):
         self.poller = zmq.Poller()
         self.poller.register(self.socket, zmq.POLLIN)
 
-    async def async_request(self, command, args=None, timeout=None):
+    async def async_request(
+        self,
+        command,
+        args=None,
+        timeout=None,
+        auth_user=None
+    ):
         """Send an asynchronous request using asyncio.
 
         Has the same arguments and return values as ``serial_request``.
@@ -175,6 +185,8 @@ class WorkflowRuntimeClient(ZMQSocketBase):
         # there is no need to encrypt messages ourselves before sending.
 
         # send message
+        if auth_user:
+            set_authenticated_user(self.header, auth_user)
         msg = {'command': command, 'args': args}
         msg.update(self.header)
         LOG.debug('zmq:send %s', msg)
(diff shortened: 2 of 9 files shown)


cylc/flow/network/client.py
```diff
@@ -159,7 +162,13 @@ def _socket_options(self):
         self.poller = zmq.Poller()
         self.poller.register(self.socket, zmq.POLLIN)
 
-    async def async_request(self, command, args=None, timeout=None):
+    async def async_request(
+        self,
+        command,
+        args=None,
+        timeout=None,
+        auth_user=None
```
oliver-sanders, line 170: If it's quick, it might be a good idea to generalise this to handle arbitrary metadata in a dictionary.

We could do:

```python
# add the shared header
msg.update(self.header)
# add the request metadata
msg['meta'].update(request_meta)
```

This would make it easier to pass through additional information e.g. the app that issued the command or the GraphQL operation name (could be useful if the request contains multiple mutations).

hjoliver:
@datamel - I'm quite keen to get this in! Are you planning to address @oliver-sanders 's comments above? (Pinging you just in case you missed them).

datamel:
> @datamel - I'm quite keen to get this in! Are you planning to address @oliver-sanders 's comments above? (Pinging you just in case you missed them).

Yes, on it today. Sorry for the delay.

cylc/flow/scheduler.py
```diff
@@ -836,6 +854,11 @@ def process_command_queue(self) -> None:
             )
             sep = ', ' if kwargs_string and args_string else ''
             cmdstr = f"{name}({args_string}{sep}{kwargs_string})"
+            # print("moooooo")
+            # if 'tasks' in kwargs:
+            #     import mdb
+            #     print("atttttttatchchhc")
+            #     mdb.debug(ui_server=True)
```
hjoliver, line 861: ? :rofl:

hjoliver:
Tested both branches, looks good :+1: 

What gets logged for direct commands (not via the UIS) looks dodgy however:
```
INFO - Authenticated user ? has submitted request kill_tasks
```
vs
```
INFO - Authenticated user oliverh has submitted request force_trigger_tasks
```

Perhaps we could say `owner has submitted request...` for direct commands?

cylc/flow/scheduler.py
datamel: hehe

cylc/flow/network/resolvers.py
```diff
@@ -639,23 +639,29 @@ async def nodes_mutator(self, *m_args):
...
         if result is None:
             result = (True, 'Command queued')
         return [{'id': w_id, 'response': result}]
 
     async def _mutation_mapper(
-        self, command: str, kwargs: Dict[str, Any]
+        self, command: str, kwargs: Dict[str, Any], meta: Dict[str, Any]
     ) -> Optional[Tuple[bool, str]]:
         """Map between GraphQL resolvers and internal command interface."""
         method = getattr(self, command, None)
```
oliver-sanders, line 651: I think this is the first point at which the GraphQL routing has been finalised so in theory we can move the `client-command` logging down here, raise the authorised user logging up here and merge the two together to avoid having two separate log messages for each mutation.

If we do that I don't think we need to pass meta any further down the chain?

WDYT?

datamel: 👍 , I'll give that a go.

cylc/flow/network/resolvers.py
```diff
@@ -592,60 +593,21 @@ def __init__(self, data: 'DataStoreMgr', schd: 'Scheduler') -> None:
...
             workflows = list(self.data_store_mgr.data.keys())
             return [{
                 'response': (False, f'No matching workflow in {workflows}')}]
         w_id = w_ids[0]
-        result = await self._mutation_mapper(command, args)
-        if result is None:
-            result = (True, 'Command queued')
-        return [{'id': w_id, 'response': result}]
-
-    async def nodes_mutator(self, *m_args):
```
MetRonnie, line 608: Why is `nodes_mutator` being removed?

datamel: This was picked up when adding the unpacking of the meta and was discussed with @oliver-sanders, who looked at codecov for verification. It is seemingly dead code and was asked to remove it. Pinged @dwsutherland on the sibling to verify it is indeed unused code.

cylc/flow/network/authorisation.py
```diff
@@ -40,17 +38,7 @@ def wrapper(fcn):
         def _call(self, *args, user='?', meta=None, **kwargs):
             if not meta:
                 meta = {}
-            host = meta.get('host', '?')
-            prog = meta.get('prog', '?')
-            comms_method = meta.get('comms_method', '?')
-
-            # Hardcoded, for new - but much of this functionality can be
-            # removed more swingingly.
-            LOG.debug(
```
datamel, line 49: @oliver-sanders, just thinking about removing this line. It does remove the `comms_method` which may be useful for debugging in future? Want me to put any logging information about the comms method to replace this, or are you happy with the info loss from the removal?

cylc/flow/network/resolvers.py
```diff
@@ -655,7 +617,16 @@ async def _mutation_mapper(
             self.schd.get_command_method(command)
         except AttributeError:
             raise ValueError(f"Command '{command}' not found")
-        self.schd.command_queue.put((command, tuple(kwargs.values()), {}))
+        if command != "put_messages":
+            log_msg = f"[command] {command} "
+            user = meta.get('auth_user', self.schd.owner)
+            if user != self.schd.owner:
+                log_msg += (f"(issued by {user})")
```
oliver-sanders, line 624: Minor point to avoid logging trailing whitespace:

```suggestion
            log_msg = f"[command] {command}"
            user = meta.get('auth_user', self.schd.owner)
            if user != self.schd.owner:
                log_msg += (f" (issued by {user})")
```

tests/functional/cylc-message/00-ssh.t
```diff
@@ -21,16 +21,13 @@
...
 install_workflow "${TEST_NAME_BASE}" "${TEST_NAME_BASE}"
 
 run_ok "${TEST_NAME_BASE}-validate" \
     cylc validate "${WORKFLOW_NAME}"
 workflow_run_ok "${TEST_NAME_BASE}-run" \
     cylc play --debug --no-detach --reference-test "${WORKFLOW_NAME}"
 
-grep_ok "\[client-command\] graphql ssh" \
-    "$RUN_DIR/${WORKFLOW_NAME}/log/workflow/log" 
-
```
oliver-sanders, line 34: This was a safety test to ensure that the comms method is logged.

Perhaps we could add a Cylc client command into the workflow e.g. `cylc hold $CYLC_WORKFLOW_ID//1/elephant` which should give us a `[command]` entry that we can check.

oliver-sanders:
Looking good, will have a quick test soon.

tests/functional/cylc-message/00-ssh.t
datamel: Yes, are you happy if I do this in the ssh async request PR as this test is broken anyway?

cylc/flow/network/authorisation.py
oliver-sanders: We can pass the comms_method down with the metadata.

oliver-sanders (approved):

MetRonnie (approved):

cylc/flow/network/resolvers.py
dwsutherland: It was in use, until:
https://github.com/cylc/cylc-flow/pull/3469
specifically:
https://github.com/cylc/cylc-flow/pull/3469/files#diff-97f4d62be41218e2d70a203edfca6d44d7c0046b6908cefa8e3af7c02cd31920

And the idea behind it was that mutations being multi-workflow could be given node ids from different workflows, and then these ids would be parsed and mutations sent to the appropriate workflow..

The fact that we moved away from this, means we are happy with not using this functionality... i.e. a node mutation will always be associated with only one workflow and/or the workflow parsing happens somewhere else.

PR #4526, opened 2021-11-22
Disallow run n and run x as flow names

### LINKED ISSUE #4394: Disallow runX as a workflow name

wxtim:
closes #4394  - IMO a single reviewer PR.

**Requirements check-list**
  `conda-environment.yml`.
- [x] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/327.


cylc/flow/workflow_files.py
```diff
@@ -1326,6 +1326,11 @@ def validate_workflow_name(name: str) -> None:
             "Workflow name cannot be a path that points to the cylc-run "
             "directory or above"
         )
+    if re.findall('run(N|[0-9]+)', Path(name).name):
```
MetRonnie, line 1329: Can't this be simplified to?
```suggestion
    if re.findall('run(N|\d+)', Path(name).name):
```

wxtim: Flake8 objects, unless you do 
```suggestion
     if re.findall(r'run(N|[\d+)', Path(name).name): 
```

cylc/flow/workflow_files.py
```diff
@@ -1331,6 +1331,11 @@ def validate_workflow_name(name: str) -> None:
             "Workflow name cannot be a path that points to the cylc-run "
             "directory or above"
         )
+    if re.findall(r'^run(N|\d+)$', Path(name).name):
```
wxtim, line 1334: @MetRonnie  - If this doesn't suite then I think this is easier resolved talking on the phone. I've changed the condition not the message because I'm not sure I want to exclude potentially legit flow names like:
- underrun2
- roadrunNer-beep-beep 
Should be ok. Having said which if you think we should ban any name containing the expression I don't think I can think of any non contrived examples actually likely to cause problems. (https://www.thefreedictionary.com/words-containing-run)

cylc/flow/workflow_files.py
```diff
@@ -1311,7 +1311,7 @@ def _parse_src_reg(reg: Path, cur_dir_only: bool = False) -> Tuple[Path, Path]:
     return (reg, abs_path)
 
 
-def validate_workflow_name(name: str) -> None:
+def validate_workflow_name(name: str, runNcheck=False) -> None:
```
MetRonnie, line 1314: Why do we need this `runNcheck` argument?

cylc/flow/workflow_files.py
```diff
@@ -1331,9 +1331,9 @@ def validate_workflow_name(name: str) -> None:
             "Workflow name cannot be a path that points to the cylc-run "
             "directory or above"
         )
-    if re.findall('run(N|[0-9]+)', Path(name).name):
+    if runNcheck and re.findall(r'^run(N|\d+)$', Path(name).name):
```
MetRonnie, line 1334: We should ban it anywhere in the name (e.g. `bart/runN/lisa` should be banned)
```suggestion
    if runNcheck and any(
        re.match(r'^run(N|\d+)$', dir_name)
        for dir_name in Path(name).parts
    ):
```

cylc/flow/workflow_files.py
```diff
@@ -1331,9 +1331,9 @@ def validate_workflow_name(name: str) -> None:
...
         raise WorkflowFilesError(
-            "Workflow name cannot end with a folder called 'runN' or "
+            "Workflow name cannot be a folder called 'runN' or "
```
MetRonnie, line 1336: See above
```suggestion
            "Workflow name cannot contain a folder called 'runN' or "
```

cylc/flow/workflow_files.py
wxtim: Other commands that use `validate_workflow_name` can (sometimes should) be passed a name which ends in `runN` or `runX`: Consider `cylc clean `~/cylc-run/my-workflow/runN`. Cylc clean uses validate workflow, and in that case it certainly doesn't want to claim that `~/cylc-run/my-workflow/runN` is not valid.

MetRonnie (approved):

datamel (approved):
Thanks @wxtim. I have read and checked out the code and run the tests.

PR #4529, opened 2021-11-23
Refactor client/server & GraphQL workflow mutations

MetRonnie:
These changes partially address https://github.com/cylc/cylc-uiserver/issues/271

- Refactored the message format sent from server to client. Instead of just the raw result of the GraphQL mutation (or whatever request was sent[^1]), the server now responds with a tuple of `(content, error, user)`
  - This allows errors to be returned over the ZMQ socket without mixing with/polluting the data content
  - `user` is included like this instead of adding it to the GraphQL `{data: ..., errors: ...}` response (the spec does not permit adding extra keys)
  - A `NamedTuple` called `ResponseTuple` is used instead of a regular tuple for better clarity and type safety
- RefactoredGraphQL workflow mutation output type. Instead of the sometimes-inconsistent `GenericResponse` output which could be any object in practice, the output is now a `graphene.List` of `GenericReponse`s, and `GenericResponse` has 3 fields: `workflowId = graphene.String()`, `success = graphene.Boolean()` and `message = graphene.String()`
  - This pins down what the expected response should be, which should avoid traceback issues like https://github.com/cylc/cylc-uiserver/issues/271
  - In practice this looks like:
    ```graphql
    mutation {
      hold (
        workflows: ["temp4"]
        tasks: ["foo.2"]
      ) {
        results {
          workflowId
          success
          message
        }
      }
    }
    ```
    ```json
    {
      "data": {
        "hold": {
          "results": [
            {
              "workflowId": "rdutta|temp4",
              "success": true,
              "message": "Command queued"
            }
          ]
        }
      }
    }
    ```

[^1]: The exception to this is that Protobuf results are sent back as-is in `bytes` format instead of the tuple, because `bytes` is not JSON-serializable

**Requirements check-list**


cylc/flow/network/__init__.py
```diff
@@ -46,19 +45,17 @@
 MSG_TIMEOUT = "TIMEOUT"
 
 
-def encode_(message):
+def JSON_stringify(message: object) -> str:
```
dwsutherland, line 48: It's usually convention to have functions and methods all lowercase in python, no?

cylc/flow/network/__init__.py
```diff
@@ -46,19 +45,17 @@
...
     except TypeError as exc:
         return json.dumps({'errors': [{'message': str(exc)}]})
 
 
-def decode_(message):
-    """Convert an encoded message string to JSON with an added 'user' field."""
-    msg = json.loads(message)
-    msg['user'] = getpass.getuser()  # assume this is the user
-    return msg
+def JSON_parse(message: str) -> object:
```
dwsutherland, line 56: Same here.. You'd be hard pressed to find any other mixed case functions in our code.

cylc/flow/network/schema.py
```diff
@@ -206,7 +219,7 @@ def parse_node_id(item, node_type=None):
 
 # Field args (i.e. for queries etc):
 class SortArgs(InputObjectType):
-    keys = List(String, default_value=['id'])
+    keys = graphene.List(String, default_value=['id'])
```
dwsutherland, line 222: I originally had it `graphene.List(` but it was deemed unnecessary by reviewers.. I guess it can't be helped with `typing`..

cylc/flow/network/schema.py
```diff
@@ -1246,9 +1259,21 @@ class Meta:
...
-        description = """Container for command queued response"""
+        description = """Container for workflow command queued response"""
+
+    workflowId = String()
+    success = Boolean(required=True)
+    message = String(required=True)
+
 
-    result = GenericScalar()
+class GenericResponseTuple(NamedTuple):
```
dwsutherland, line 1269: You can just use above Graphene `ObjectType`, `GenericResponse`, as a data container.. Why do we need this one?

dwsutherland (commented):
So looks good.. Few minor comments below..

The main issue I'm having (when trying to run this and the UIS sibling), is getting it working
image
(even after reinstall of venvs)

Perhaps after a rebase, it would help, as there are breaking changes in this and master (?).

cylc/flow/network/schema.py
MetRonnie: Yes it's an unfortunate clash. But seeing as `typing` is widely-used now I think it should take precedence

cylc/flow/network/schema.py
MetRonnie: > So looks good.. Few minor comments below..
> 
> The main issue I'm having (when trying to run this and the UIS sibling), is getting it working

Sorry, should have mentioned: It's not working from the UI yet as some changes need to be made there. But GraphiQL is working.

MetRonnie:
> So looks good.. Few minor comments below..
> 
> The main issue I'm having (when trying to run this and the UIS sibling), is getting it working

Sorry, should have mentioned: It's not working from the UI yet as some changes need to be made there. But GraphiQL is working.

MetRonnie:
Not going to be finished by Christmas so bumping to 8.0.0

Issue #3606, opened 2020-05-11
actions: cache environment


oliver-sanders:
Investigate caching the environment used by Cylc GH Actions in order to reduce installation time.

This is made difficult by the dependence on system modules (shellcheck, sqlite3, mail, postfix, etc).

**Pull requests welcome!**

kinow:
Will watch this issue. In Cylc UI we are caching only the `node_modules`, but we might be able to cache more and speed up more the build time.

Here's the cache configuration for Cylc UI: https://github.com/cylc/cylc-ui/blob/dcc4b4529ce6dcd40e67d1dc2c0aecb49e3afc58/.github/workflows/main.yml#L13-L17

PR #4532, opened 2021-11-25
GH Actions: Cache pip dependencies

MetRonnie:
This is a small change with no associated Issue. Hopefully will shave some time off GH Actions checks [Update: not really 🤷‍♂️]

https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching/

**Requirements check-list**
- [x] Does not need tests 

### DIFF
--- a/.github/workflows/test_manylinux.yml
+++ b/.github/workflows/test_manylinux.yml
@@ -47,6 +47,8 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+          cache: pip
+          cache-dependency-path: setup.cfg
 
       - name: Configure Manylinux Compatibility
         # Make this platform look older than it is. For info see:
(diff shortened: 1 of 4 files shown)


.github/workflows/test_manylinux.yml
```diff
@@ -47,6 +47,8 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+          cache: pip
+          cache-dependency-path: setup.cfg
```
MetRonnie, line 51: Don't know if there is a reason to avoid caching for the manylinux test?

oliver-sanders:
Partially addresses https://github.com/cylc/cylc-flow/issues/3606

oliver-sanders:
Do we have the ability to reset this cache manually?

MetRonnie:
https://github.com/actions/setup-python#caching-packages-dependencies says:
> Please Note: Restored cache will not be used if the requirements.txt file is not updated for a long time and a newer version of the dependency is available that can lead to an increase in total build time.

(It only talks about requirements.txt but https://github.com/actions/setup-python/issues/282#issuecomment-978921695 says you can use setup files too)

---

To be honest, I'm minded to close this as it isn't really making a dent on install times

oliver-sanders:
> To be honest, I'm minded to close this as it isn't really making a dent on install times

Fine with me, the big cost ATM is the Conda envs.

I think we should containerise our build env and install deps from Conda so we are always testing the correct stack and aren't don't get messed around with CI env changes. We could build and cache the env based on the hash of the `conda-environment.yml` and `setup.cfg` files which would massively cut down on build time.

PR #4536, opened 2021-11-26
made cylc wrapper script available using cylc extract-resources

### LINKED ISSUE #4480: wrapper: package wrapper script
Move the wrapper script into `cylc/flow/etc` so that it can be extracted from the installed source with `cylc extract-resources`.

**Pull requests welcome!**

wxtim:
Closes #4480 : Cylc wrapper script made available through `cylc extract-resource`.

Also makes the following small changes, as per [discussion on Element](https://matrix.to/#/!hwZqSYihGPuhDdIzIP:matrix.org/$163793887755181YeRkN:matrix.org?via=matrix.org):
- Source argument now before target.
- Metadata as well as names from ``--list`` option.
- Files extracted to to ``target/source_name`` rather than ``target/full/source/path``.

**Requirements check-list**
  `conda-environment.yml`.
- [x] Does not need tests (small change).
- [x] (master branch) I have opened a documentation PR at https://github.com/cylc/cylc-doc/pull/346.


cylc/flow/scheduler.py
```diff
@@ -306,7 +306,7 @@ async def install(self):
 
         # Extract job.sh from library, for use in job scripts.
         extract_resources(
-            workflow_files.get_workflow_srv_dir(self.workflow),
+            workflow_files.get_workflow_srv_dir(self.workflow) + '/etc/',
```
MetRonnie, line 309: Using `os.path.join` or pathlib would be safer

cylc/flow/task_remote_cmd.py
```diff
@@ -147,7 +147,7 @@ def remote_init(install_target: str, rund: str, *dirs_to_symlink: str) -> None:
     oldcwd = os.getcwd()
     os.chdir(rund)
     # Extract job.sh from library, for use in job scripts.
-    extract_resources(WorkflowFiles.Service.DIRNAME, ['etc/job.sh'])
+    extract_resources(WorkflowFiles.Service.DIRNAME + '/etc/', ['etc/job.sh'])
```
MetRonnie, line 150: Same comment about path joining

setup.cfg
```diff
@@ -145,7 +145,7 @@ cylc.command =
     diff = cylc.flow.scripts.diff:main
     dump = cylc.flow.scripts.dump:main
     ext-trigger = cylc.flow.scripts.ext_trigger:main
-    extract-resources = cylc.flow.scripts.extract_resources:main
+    resources = cylc.flow.scripts.extract_resources:main
```
MetRonnie, line 148: Has this rename been discussed? Personally `extract-resources` is clearer to me

wxtim: between @oliver-sanders  and I. Perhaps this should be thrown open as a discussion. 
Would `get-resouces` or `get-files` do for you?

MetRonnie: `get-resources` seems fine to me. I was just curious about the reason for the change

wxtim: make it shorter since users will be typing it to get the tutorials.

oliver-sanders: Open to offers, IMO `get-resources` is too long:

Context:
* `extract-resources` is a bit of a mouthful and not especially obvious anyway.
* This command will probably be used to extract the Cylc tutorial so we want something convenient to type.
* The rose equivalent is called `rose resource`.

wxtim: I was going for a plural, because by default the script retrieves everything.
Is that desirable behavior if we are listing all the tutorials?

Given the somewhat different users (cylc extract-resources in its current form is an internal/site admin command, whilst tutorial is explicitly for newbies) might it  be sensible to have a  `cylc tutorial` command. This would in effect be an alias, with slightly differing behavior (i.e. a different list of resources) depending on `sys.argv[2]`.

oliver-sanders: Alternative name suggestions?

MetRonnie: If it's just a command that most users will only use when first learning about Cylc and doing the tutorials, I feel it's ok to have a bit of a long command name

wxtim: I'm going to make a decision and go for `cylc get-resources`. 
Reasoning:
- 95% of users will be copy and pasting it out of tutorials anyway.
- Get < Extract
- reading_age('get') < reading_age('extract') for ESL users.
- `resources` alone not massively obvious.
- `resources` because it's designed to extract 1+ resources.

The only negative I can see is that it's inconsistent with `rose resource`, but since I think most people using it will either be copying and pasting from the tutorial or expert users I'm not too worried about this. If it becomes a problem we can always put an alias in.

CHANGES.md
```diff
@@ -59,6 +59,13 @@ First Release Candidate for Cylc 8.
 [#4442](https://github.com/cylc/cylc-flow/pull/4442) - Prevent installation
 of workflows inside other installed workflows.
 
+[#4536](https://github.com/cylc/cylc-flow/pull/4536) - ``cylc extract-resource``
+renamed ``cylc resource`` and small changes made:
```
MetRonnie, line 63:
```suggestion
[#4536](https://github.com/cylc/cylc-flow/pull/4536) - `cylc extract-resources`
renamed `cylc get-resources` and small changes made:
```

cylc/flow/scripts/extract_resources.py
```diff
@@ -29,8 +29,8 @@ def get_option_parser():
     parser = COP(
```
MetRonnie, line 29: Might want to rename this file to `get_resources.py`?

MetRonnie (approved):
Apart from tests needing updating, approved

datamel (approved):
I have checked out the branch, read the code, run the tests and can confirm it is working as expected. Thanks @wxtim.

PR #4537, opened 2021-11-29
Turn off broken flake8-bugbear rule B018 (for now)

MetRonnie:
This is a small change with no associated Issue. Fixes failing style test on GH Actions.

We should re-enable this rule when https://github.com/PyCQA/flake8-bugbear/issues/208 is fixed


datamel (approved):
Thanks @MetRonnie, this solves the problem until we can do the code fixes.

PR #4541, opened 2021-12-02
Pin to Python 3.9.

hjoliver:


MetRonnie (approved):
Oliver's on leave today, might as well bump this through

PR #4543, opened 2021-12-03
Make `--stopcp=reload` use `[scheduling]stop after cycle point` instead of the final cycle point

MetRonnie:
These changes partially address #4062 

`cylc play --stopcp=reload` now does what it says - reload the value from `[scheduling]stop after cycle point`. (Previously it would use the value of the final cycle point)

**Requirements check-list**
- [x] Appropriate tests are included/updated (unit and functional).
- [x] (master branch) I have opened a documentation PR at https://github.com/cylc/cylc-doc/pull/362


tests/functional/cylc-play/06-warnif-scp-after-fcp.t
```diff
@@ -53,35 +51,3 @@ for SCP in 1 2 3; do
 done
 
 purge
-
-# Gregorian Cycling
```
MetRonnie, line 57: No need to check this for different cycling types; it should be the responsibility of the `tests/unit/cycling` and `tests/functional/cyclers` tests to ensure cycle point comparison works

tests/functional/restart/44-stop-point.t
```diff
@@ -1,117 +0,0 @@
...
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#-------------------------------------------------------------------------------
-# Test restart with stop point
```
MetRonnie, line 18: This covered the same stuff as `tests/functional/restart/08-stop-after-cycle-point.t`

cylc/flow/scheduler.py
```diff
@@ -678,17 +678,15 @@ def _load_pool_from_point(self):
...
+            start_type = "Warm"
+        LOG.info(f"{start_type} start from {self.config.start_point}")
 
         flow_num = self.flow_mgr.get_new_flow(
             f"original flow from {self.config.start_point}"
         )
         for name in self.config.get_task_name_list():
-            if self.config.start_point is None:
-                # No start cycle point at which to load cycling tasks.
-                continue
```
MetRonnie, line 691: Pretty sure this condition is no longer possible. And L691 showing as never hit in code coverage: https://app.codecov.io/gh/cylc/cylc-flow/blob/master/cylc/flow/scheduler.py

cylc/flow/scheduler.py
```diff
@@ -708,9 +706,6 @@ def _load_pool_from_point(self):
 
     def _load_pool_from_db(self):
         """Load task pool from DB, for a restart."""
-        if self.options.startcp:
-            self.config.start_point = TaskID.get_standardised_point(
-                self.options.startcp)
```
MetRonnie, line 713: This seems to be unnecessary as duplication of https://github.com/cylc/cylc-flow/blob/b366e13d93aaed34e11c764fbf1483557fb5da2a/cylc/flow/config.py#L669-L687

cylc/flow/scheduler.py
```diff
@@ -1125,10 +1120,6 @@ def load_flow_file(self, is_reload=False):
             handle.write(b"# cylc-version: %s\n" % CYLC_VERSION.encode())
             printcfg(self.config.cfg, none_str=None, handle=handle)
 
-        if not self.config.initial_point and not self.is_restart:
-            LOG.warning('No initial cycle point provided - no cycling tasks '
-                        'will be loaded.')
-
```
MetRonnie, line 1131: Again pretty sure this condition is no longer possible (`config.initial_point` should be set by this point)

cylc/flow/scheduler.py
```diff
@@ -1159,23 +1150,17 @@ def _load_workflow_params(self, row_idx, row):
...
         key, value = row
         if key in self.workflow_db_mgr.KEY_INITIAL_CYCLE_POINT_COMPATS:
-            if self.is_restart and self.options.icp == 'reload':
-                LOG.debug(f"- initial point = {value} (ignored)")
-            elif self.options.icp is None:
-                self.options.icp = value
-                LOG.info(f"+ initial point = {value}")
+            self.options.icp = value
+            LOG.info(f"+ initial point = {value}")
         elif key in self.workflow_db_mgr.KEY_START_CYCLE_POINT_COMPATS:
```
MetRonnie, line 1155: Since #4313, initial and start cycle points cannot change on restart. But still need to set `self.options.icp`/`startcp` as they are used later to set `self.config.initial_point`/`start_point`.

cylc/flow/scheduler.py
wxtim: Checked in practice. Confirmed I can't get it to be true.

cylc/flow/scheduler.py
wxtim: Confirm looks reasonable.

cylc/flow/scheduler.py
wxtim: Legit.

cylc/flow/scheduler.py
```diff
@@ -678,17 +678,15 @@ def _load_pool_from_point(self):
         released from runhead.)
 
         """
-        if self.config.start_point is not None:
-            start_type = "Warm" if self.options.startcp else "Cold"
-            LOG.info(f"{start_type} start from {self.config.start_point}")
+        start_type = "Cold"
+        if self.config.start_point > self.config.initial_point:
+            start_type = "Warm"
+        LOG.info(f"{start_type} start from {self.config.start_point}")
```
wxtim, line 684: This is definitley more readable.

One more alternative?
```suggestion
LOG.info(
    f"{'warm' if self.options.startcp else 'cold'} start from {self.config.start_point}.")
```

cylc/flow/scheduler.py
```diff
@@ -1159,23 +1150,17 @@ def _load_workflow_params(self, row_idx, row):
...
-                LOG.debug(f"- start point = {value} (ignored)")
-            elif self.options.startcp is None:
-                self.options.startcp = value
-                LOG.info(f"+ start point = {value}")
+            self.options.startcp = value
+            LOG.info(f"+ start point = {value}")
         elif key in self.workflow_db_mgr.KEY_FINAL_CYCLE_POINT_COMPATS:
             if self.is_restart and self.options.fcp == 'reload':
-                LOG.debug(f"- override final point = {value} (ignored)")
+                LOG.debug(f"- final point = {value} (ignored)")
```
wxtim, line 1160: Should this not be info, or even warning?

wxtim (approved):
- [x] Read, looks reasonable.
- [x] Checked to see if I can trigger conditions removed as surplus to reqs.

cylc/flow/scheduler.py
hjoliver: I'll take this and merge :grin:

hjoliver (approved):
Thanks @MetRonnie - good to have this confusing mess cleared up!

PR #4544, opened 2021-12-06
Tidy + reinstate deleted workflow_files unit test

MetRonnie:

Modified and reinstated a unit test that was deleted in #4506. I think this test neatly parametrises several scenarios for workflow install


datamel (approved):
One review should do.

PR #4547, opened 2021-12-07
Install scan depth

### LINKED ISSUE #4363: `cylc install` should respect max scan depth
No point in installing to a deep hierarchy if we can't find the resulting workflow!

@MetRonnie I think you worked most recently on `cylc install`? otherwise we can reassign...

MetRonnie:
These changes close #4363 

- `cylc install` now fails if the workflow ID (including the `/run1` bit) would exceed the max scan depth. For example, `one/two/three/four/run1` would exceed a max scan depth of 4.
- Make the max scan depth configurable in `global.cylc[install]max depth`

**Requirements check-list**


tests/functional/cylc-install/07-no-recursive-installations.t
```diff
@@ -31,49 +31,33 @@ __HEREDOC__
 run_ok "$TEST_NAME_BASE" cylc validate "$PWD"/flow.cylc
```
MetRonnie, line 32: In this file I have cut down on duplication - no need to test that install into `cylctb/child` fails if you're also testing that install into `cylctb/child/grandchild` fails.

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -627,15 +627,37 @@
...
+            .. versionadded:: 8.0.0
+        ''')
         Conf('source dirs', VDR.V_STRING_LIST, default=['~/cylc-src'], desc='''
             List of paths that Cylc searches for workflows to install.
 
             All workflow source directories in these locations will
             also show up in the GUI, ready for installation.
 
-            .. caution::
+            .. note::
```
wxtim, line 656: I think this needs a higher level than note: If a site has multiple source directories it is very likely that sooner or later a user will install the wrong workflow by this mechanism, so I think that the risk needs to be highlighted to site admins.

wxtim (commented):
- [x] Read changes, look sane
- [ ] Double checked whether tests are rational. 
- [ ] All tests passing. 
      
 ```
> ctb tests/f/cylc-play/07-provided-vars.t -vv
WorkflowFilesError: Cannot install: workflow ID 'cylctb-20211207T130736Z-KilN/f/cylc-play/07-provided-vars/run1' would exceed global.cylc[install]max depth = 4
```
- [ ] Double Checked hole commented on by codecov -

File: cylc/flow/cfgspec/globalcfg.py (hunk 2 above)
Comment (MetRonnie) [reply cylc/flow/cfgspec/globalcfg.py:656]: Imho the `note` admonition provides the right level of highlight, seeing as this is standard search-path behaviour. Can see what Oliver says

File: tests/unit/conftest.py [hunk 3]
```diff
@@ -120,6 +120,40 @@ def mod_tmp_run_dir(tmp_path_factory: pytest.TempPathFactory):
...
+    return tmp_src_dir(tmp_path)
+
+
+@pytest.fixture(scope='module')
+def mod_tmp_src_dir(tmp_path_factory: pytest.TempPathFactory):
+    """Module-scoped version of tmp_src_dir()"""
+    tmp_path = tmp_path_factory.getbasetemp()
+    return tmp_src_dir(tmp_path)
+
+
```
oliver-sanders, line 156: I think there's a better way of doing this following the pattern used for `run_dir`. Can plug in the `flow()` stuff too:

```suggestion
# copying the patter used for run_dir

@pytest.fixture(scope='session')
def ses_src_dir():
    ...
    
@pytest.fixture(scope='module')
def mod_src_dir(ses_src_dir, request):
    return srs_src_dir / request.module.__name__
    
@pytest.fixture
def src_dir(mod_src_dir, request):
    return mod_src_dir / request.function.__name__
    
# integrating the flow stuff

@pytest.fixture(scope='module')
def mod_src_flow(mod_src_dir):
    yield partial(_make_flow, mod_src_dir, mod_src_dir)
    
@pytest.fixture
def src_flow(mod_src_dir):
    yield partial(_make_flow, src_dir, src_dir)
```

tests/unit/test_workflow_files.py
```diff
@@ -2050,3 +2071,31 @@ def test_delete_runN_skipif_cleanedrun_not_runN(tmp_path):
...
+    [
+        ('foo/' * (MAX_SCAN_DEPTH - 1), False),
+        ('foo/' * MAX_SCAN_DEPTH, True)  # /run1 takes it beyond max depth
+    ]
+)
+def test_install_workflow__max_depth(
+    workflow_name: str,
+    err_expected: bool,
+    tmp_path: Path,
+    tmp_run_dir: Callable,
```
oliver-sanders, line 2087: We already have a `test_dir` / `run_dir` fixture.

MetRonnie: Hmm but those are in `tests/integration/conftest` whereas this test is in `tests/unit/`

oliver-sanders: Ah, this is more an integration test than a unit test since it's testing a full CLI function involving FS.

Can leave it here, we have bigger fish to fry right now, but probably best to move those fixtures into the test module to prevent them being re-used.

oliver-sanders (approved):

cylc/flow/cfgspec/globalcfg.py
wxtim: Accepted.

wxtim (approved):

PR #4550, opened 2021-12-08
remove the color/no-colour option from cylc cyclepoint

wxtim:
This is a small change with no associated Issue:

While looking at tutorial workflows I noticed the following use case:
```python
    Popen(
        ['cylc', 'cyclepoint', '--offset-hours', str(int(offset_hours))],
        stdout=PIPE
    ).communicate()
```

In this case the subsequent commands fail because colour information chars are included in the output.
Because this command is likely to used by other scripts this should not have the colour options.

**Requirements check-list**
- [x] No change log entry required (v. small pre release change).

### DIFF
--- a/cylc/flow/scripts/cycle_point.py
+++ b/cylc/flow/scripts/cycle_point.py
@@ -67,6 +67,7 @@ from metomi.isodatetime.exceptions import IsodatetimeError
 def get_option_parser():
     parser = COP(
         __doc__,
+        color=False,
         argdoc=[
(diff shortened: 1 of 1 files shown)


datamel (approved):
I agree this option does not need a color option. I've checked out the branch and can confirm this has been removed. Thanks @wxtim.

hjoliver:
>  Because this command is likely to used by other scripts this should not have the colour options.

Maybe we should just turn off colour output automatically, for all commands, if a non-interactive shell is detected.

Issue #4110, opened 2021-03-04
Update isodatetime


MetRonnie:
**Describe exactly what you would like to see in an upcoming release**

Update the version of isodatetime used in cylc-flow once https://github.com/metomi/isodatetime/issues/176 is done and 2.1 is released. This should give performance improvements for datetime cycling workflows (see https://github.com/metomi/isodatetime/issues/162). Also, if https://github.com/metomi/isodatetime/issues/80 gets fixed that vanquishes a long-standing bug that is in cylc-flow.

*Optimistically tagged against 8.0rc1 as it would be good to have done before then, but not a priority. See discussion below.*

**Pull requests welcome!**

MetRonnie:
@oliver-sanders @hjoliver I have this tagged against 8.x, but seeing as there is a breaking change (https://github.com/metomi/isodatetime/pull/183), would this be be better suited to 8.0.0?

oliver-sanders:
The breaking change shouldn't be a problem (Cylc is used as a system rather than a library), good to keep 8.0.0 to the more mission-critical aspects right now.

MetRonnie:
Forgot to mention, we could just release isodatetime 2.1.0 now and finish metomi/isodatetime#176 later in 2.1.1. Upgrading cylc-flow to isodatetime 2.1.0 shouldn't take too long.

MetRonnie:
Optimistically tagged against 8.0.0 as it would be good to have done before then, but not a priority

PR #2815, opened 2018-10-23
Relative initial cycle point

lhuggett:
Closes #984. 

NOTE: includes latest version of isodatetime, and changes to Cylc from @kinow for compatibility with latest version of isodatetime

NOTE 2: relative ICP may break when order of tick over in isodatetime is swapped as in https://github.com/metomi/isodatetime/pull/101.

NOTE 3: fully updated documentation with information on use of days, weeks, months and years in relative ICP to follow.

PR #4554, opened 2021-12-14
Update isodatetime

MetRonnie:
These changes address https://github.com/cylc/cylc-flow/issues/4110#issuecomment-888995310. When we release isodatetime 3.0.0, this PR allows Cylc to work with the changed API.

- [x] Relies on isodatetime 3.0.0 being released

isodatetime 3.0.0 fixes a longstanding mistake in the implementation of recurrence format number 1 (https://github.com/metomi/isodatetime/issues/45).

See also [isodatetime changelog](https://github.com/metomi/isodatetime/blob/master/CHANGES.md#isodatetime-210-upcoming-2020)

Edit: as recurrence format 1 is undocumented (in Cylc docs) and rarely used, will postpone this to 8.0rc3

**Requirements check-list**
- [x] (master branch) I have opened a documentation PR https://github.com/cylc/cylc-doc/pull/454


conda-environment.yml
```diff
@@ -9,7 +9,7 @@ dependencies:
   - graphene >=2.1,<3
   - graphviz  # for static graphing
   - jinja2 ==2.11.0,<2.12
-  - metomi-isodatetime >=1!2.0.2, <1!2.1.0
+  - metomi-isodatetime >=1!3.0.0, <1!3.1.0
```
oliver-sanders, line 12: I think the upper pin is a good idea for isodatetime!

cylc/flow/cycling/iso8601.py
```diff
@@ -765,21 +763,28 @@ def prev_next(
...
     # case 1 - year only
     if re.search(r"\(-\d{2}[);T]", value):
-        cycle_point.month_of_year = 1
-        cycle_point.day_of_month = 1
+        cycle_point._month_of_year = 1
+        cycle_point._day_of_month = 1
     # case 2 - month only or year and month
     elif re.search(r"\(-(-\d{2}|\d{4})[;T)]", value):
-        cycle_point.day_of_month = 1
+        cycle_point._day_of_month = 1
```
oliver-sanders, line 787: Should we create a new cycle point instead?

MetRonnie: I think this is the cleanest option for now. In future we could make a method on `isodatetime.TimePoint` to copy and replace kwargs, .e.g
```python
a = TimePoint(year=2022)
b = a.copy(month_of_year=4)
```

oliver-sanders: We should be able to implement a clone method at this end for now. I think this will do?:

```python
from metomi.isodatetime.data import TimePoint


def clone_time_point(time_point):
    return TimePoint(
        **{
            attr: getattr(time_point, attr)
            for attr in TimePoint.DATA_ATTRIBUTES
            if attr != 'time_zone'
        },
        **{
            'time_zone_hour': time_point.time_zone.hours,
            'time_zone_minute': time_point.time_zone.minutes,
        },
    )
```

MetRonnie: There is already a `._copy()` method on `TimePoint` that does that. It just doesn't support changing an property. Adding the ability to change a property on copy would make it worthwhile to make it a public method IMHO. Otherwise, at the moment there is no gain to be had from making a copy here, as the `TimePoint` is a local variable in the function and not a dictionary key, so it is safe to mutate it in this circumstance

tests/functional/cyclers/51-recurrence_fmt_1.t
```diff
@@ -15,26 +15,8 @@
...
-    [[graph]]
-        P0M = foo  # OOPS! zero-width interval
-[runtime]
-    [[foo]]
-        script = true
-__FLOW_CONFIG__
-
-run_fail "${TEST_NAME_BASE}" cylc validate .
-cmp_ok "${TEST_NAME_BASE}.stderr" <<'__ERR__'
-SequenceDegenerateError: R/20100101T0000Z/P0Y, point format CCYYMMDDThhmmZ: equal adjacent points: 20100101T0000Z => 20100101T0000Z.
```
MetRonnie, line 37: This error was only raised on validation, was never raised on run. I have replaced it with a warning, see `tests/unit/test_config.py`

cylc/flow/cycling/iso8601.py
oliver-sanders: The caveats you've outlined on hashing and cashing seem a bit scary to me and we *are* performing calculations on the timepoint which *could* potentially hit cached interfaces or force the hash to be calculated somehow?:

```
my_diff = [abs(my_time - now) for my_time in timepoints]
```

Seems like it should be easy enough to side-step this problem?

```python
def clone_time_point(time_point, **kwargs):
    return TimePoint(
        **{
            attr: getattr(time_point, attr)
            for attr in TimePoint.DATA_ATTRIBUTES
            if attr != 'time_zone'
        },
        **{
            'time_zone_hour': time_point.time_zone.hours,
            'time_zone_minute': time_point.time_zone.minutes,
        },
        **kwargs,
    )

clone_time_point(time_point, month_of_year=1, day_of_month=1)
```

Yes, we should add that to isodatetime one day but that can come later.

MetRonnie:
(Test failures due to usual integration tests hanging on GH Actions)

cylc/flow/cycling/iso8601.py
MetRonnie: We don't actually cache any methods yet - https://github.com/metomi/isodatetime/issues/176. So in isodatetime 3.1 we can add both caching and the `.copy()` method that supports changing attributes.

MetRonnie: I'm a little afraid of passing the attributes through `__init__()` again, I'm not 100% sure they will be unaffected. The `._copy()` method currently creates a new `TimePoint` bypassing `__init__()` and setting the attributes directly https://github.com/metomi/isodatetime/blob/731008b51346e75f3ca553c5b924586df6f69a34/metomi/isodatetime/data.py#L1621

MetRonnie: Not yet, and I guess we should create an issue for updating to the copy method when it is implemented

cylc/flow/cycling/iso8601.py
```diff
@@ -668,13 +665,14 @@ def ingest_time(value: str, now: Optional['TimePoint'] = None) -> str:
 
     # correct for year in 'now' if year only,
     # or year and time, specified in input
+        # TODO: Figure out why this correction is needed
```
wxtim, line 668: Should this still be here? If yes, is there a matching issue?

tests/unit/test_time_parser.py
```diff
@@ -14,187 +14,197 @@
...
-            if ctrl_results is None:
-                continue
-            test_results = []
-            for i, test_result in enumerate(recurrence):
-                self.assertEqual(str(test_result), ctrl_results[i])
-                test_results.append(str(test_result))
-            self.assertEqual(test_results, ctrl_results)
+def test_fourth_recurrence_format(parsers):
+    """Test the fourth ISO 8601 recurrence format."""
+    tests = [("PT6H/20000101T0500Z", "R25/19991226T0500Z/PT6H"),
```
wxtim, line 149: Can we not turn this into pytest parameters? 
If one fails it will make it easier to see _which_ one.

tests/functional/lib/bash/test_header
```diff
@@ -779,7 +779,7 @@ mock_smtpd_kill() {  # Logic borrowed from Rose
 install_and_validate() {
     # First part of the reftest function, to allow separate use.
     # Expect 1 OK test.
-    local TEST_NAME="${1:-${TEST_NAME_BASE}}-validate"
+    local TEST_NAME="${1:-${TEST_NAME_BASE}}"
```
wxtim, line 782: What was wrong with this?

wxtim (commented):
Read through, need to do some manual testing tomorrow.

tests/unit/test_time_parser.py
MetRonnie: Effort

tests/functional/lib/bash/test_header
MetRonnie: It was resulting in `NN-testname-validate-validate`

cylc/flow/cycling/iso8601.py
oliver-sanders: Yep, please crack open an issue for that. I see you've fixed up the comment about caching, I think I'm happy this is ok.

tests/unit/test_time_parser.py
oliver-sanders: (fair justification IMO)

cylc/flow/cycling/iso8601.py
MetRonnie: Not sure if it's worth an issue, whatever it's doing it seems to be working. But if someone can figure this out down the line it would be good to update the comment above to improve the explanation

wxtim:
> (Test failures due to usual integration tests hanging on GH Actions)

Pity grumpy face isn't available...

cylc/flow/cycling/iso8601.py
wxtim: I've spent far too long trying to figure this out - I think I'd quite like an issue raised to poke this - I'm not thrilled that I've not been able to get the bottom of it.

I'm also not clear that I can find definitive docs of what `--icp` can accept.

tests/unit/test_time_parser.py
wxtim: No worries.

wxtim (approved):
Yay

cylc/flow/cycling/iso8601.py
MetRonnie: >[oliver-sanders](https://github.com/oliver-sanders) [on 16 Nov 2018](https://github.com/cylc/cylc-flow/pull/2815#discussion_r234184335)
>Comments like this have a habit of becoming a bit cryptic to future developers. Perhaps copy in the URL of the isodatetime issue.

https://github.com/cylc/cylc-flow/pull/2815#discussion_r234184335

Sadly not done for this particular comment 😬

MetRonnie: I have opened 2 issues:
- https://github.com/cylc/cylc-flow/issues/4805
- https://github.com/cylc/cylc-doc/issues/455

wxtim: > I'm also not clear that I can find definitive docs of what --icp can accept.

This. Yes. #4807 (Put in `cylc-flow` not `cylc-doc` because although it should touch both I think the CLI help is upstream)

PR #4556, opened 2021-12-17
Fix remote psutil invocation.

hjoliver:
Remote check for a running workflow (e.g. for `cylc scan - t rich` ~~or any scheduler-targetting `cylc` command~~) is failing, because it tries to execute `cylc cylc psutil` (too many `cylc`s!) on the workflow host.

This is the quick fix as deployed at NIWA.

TODO: check:
- [x] the fix doesn't break anything else 
- [x] a similar problem does not occur for other remote commands

(off-topic: the `construct_ssh_cmd` methods could use some refactoring - e.g. there's one specifically for `cylc` commands, but it isn't used in this case).
 

**Requirements check-list**
- [x] ~~Does not need tests (why?)~~ can't think of how to test this; it requires a shared FS but the shared swarm container won't work because `cylc scan` needs the source dir as well as the run dir to be shared)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -83,6 +83,9 @@ are now sparse, i.e. they will no longer be fleshed-out with defaults.
 
 ### Fixes
 
+[#4566](https://github.com/cylc/cylc-flow/pull/4566) - Fix `cylc scan`
+invocation for remote scheduler host on a shared filesystem.
+
 [#4526](https://github.com/cylc/cylc-flow/pull/4526),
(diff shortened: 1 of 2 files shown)


oliver-sanders (approved):

wxtim (approved):

hjoliver:
Merging: small bug fix; two approvals; checked that the same problem does not occur for other remote commands; difficult to test at the moment (see comment in description).

Issue #4557, opened 2021-12-17
`cylc config --list-platforms`


wxtim:
Users at sites with pre-set platforms will want to query the list of platforms and platform groups.

Using `cylc config -i [platforms]; cylc config -i [platform groups]` is a bit clunky, and may overwhelm some users with the quantity of more esoteric settings (ssh commands &c) returned.

This ticket proposes the addition of the option `--list-platforms` to `cylc config`. This option should return a list of platforms and platform groups, with the following items from the config:
- Platform regex (it's "name" in the global config)
- Hosts list / Platform method
- Job Runner
- Any fields defined under meta

**questions**
- Are there other fields we consider essential, such as "selection method"?
- Do we want to allow `cylc config --platform <name>` to give a preview of what will happen if you use <name> in a workflow? 

E.g 
```
<name> matches platform <platform> and will be run on a host in {hosts} using {batch system}

<name> matches platform group <group name> and will be run on a platform from the following list:
  - foo and will be run on a host in {hosts} using {batch **system}`**
```

**Additional context**
Discussed in an [element conversation](https://matrix.to/#/!hwZqSYihGPuhDdIzIP:matrix.org/$1639746821549588VCnCm:matrix.org?via=matrix.org)

**Pull requests welcome!**

hjoliver:
Agreed this is a good idea.

However, I suggest `cylc config --describe-platforms` or similar, for giving platform (and group) names (or regexes) plus metadata. `--list-platforms` sounds like it would list all platform config (which of course can be done with `config -i`).

wxtim:
> 

    cylc config --platform-names # just the names
    cylc config --platform-meta # names + metadata


I like this.

Don't forget that I want this CLI to access both platforms and platform groups, because when a user sets `[runtime][TASK]platform=foo` we don't know whether foo is a platform or a group.

Do we want  to reverse the config order or document that the selection process goes from the bottom up -i.e if 

```ini
[platforms]
    [[foo]]
        hosts = foo
    [[fo.]]
        hosts = bar
[platform groups]
    [[f.*]]
        hosts = baz
```

then the example of `platform=foo` will run on host `baz`?

wxtim:
### Proposal

Add the following options to `cylc config`:
- `--platform-names`, which prints out a list of platform and platform group regexes.
- `--platform-meta` which prints out:
  - regexes
  - metadata
  - hosts
  - job runner
  - install target
- In both cases the list should be prepended by a note explaining that
[...]

PR #4558, opened 2021-12-17
added metadata section to platforms and platform groups config

### LINKED ISSUE #4528: Add Metadata to Platforms
Add a metadata section to platforms so that users can do `cylc config -i [platforms]` to get information on platforms and usage.

For example I anticipate local usage to include details of maximum memory and "CPU"'s allowed on a platform.

[Copied from conversation on Element](https://matrix.to/#/!hwZqSYihGPuhDdIzIP:matrix.org/$1637628839316531Hdgzu:matrix.org?via=matrix.org]

[I've put this under 8.x because it's not a breaking change to the config]

**Pull requests welcome!**

wxtim:
These changes close #4528 

**Requirements check-list**
- [x] Does not need tests (Config change).
- [x] No documentation update required, yet. See #4557

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -78,6 +78,9 @@ workflow version control info log file format from modified-INI to JSON.
 logs (that get written in `log/flow-config/` on start/restart/reload)
 are now sparse, i.e. they will no longer be fleshed-out with defaults.
 
+[#4558](https://github.com/cylc/cylc-flow/pull/4558) -
+Added a metadata section to the platform and platform group configurations.
+
 ### Fixes
 
 [#4526](https://github.com/cylc/cylc-flow/pull/4526) - Prevent `runN` and
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -63,6 +63,25 @@ MOVEDFROMJOB = '''
 '''
 
 
+PLATFORM_META_DESCR = '''
+    Metadata for this platform or platform group.
+
+    Allows writers of platform configurations to add information
+    about platform usage. There are no-preset items because
+    Cylc does not use any platform (or group) metadata internally.
+
+    Users can then see information about defined platforms using::
+
+        cylc config -i [platforms]
+        cylc config -i [platform groups]
+
+    .. seealso::
+
+        :ref:`AdminGuide.PlatformConfigs`
+
+'''
+
+
 # Event config descriptions shared between global and workflow config.
 EVENTS_DESCR = {
     'startup handlers': (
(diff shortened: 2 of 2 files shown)


hjoliver (commented):
Looks good. Just wondering about a few things though:
- worth putting the common descriptive text in a string constant?
- would a short-cut option would be useful (e.g. `--describe-platforms`) that just prints platform names and metadata?
- there's no internal use for this, so I wonder if a single `meta` (or `description`?) items would suffice, rather than a section for multiple arbitrary items?  (I suppose it's possible that individual items that can be extracted separately could be useful at a site ... e.g. `cost per cpu hour` or something?).

wxtim:
>     * worth putting the common descriptive text in a string constant?

Yes

>     * would a short-cut option would be useful (e.g. `--describe-platforms`) that just prints platform names and metadata?

I think so; see https://github.com/cylc/cylc-flow/issues/4557 - I'm think that I'm going to take your comment as a yes to the question in that ticket.

>     * there's no internal use for this, so I wonder if a single `meta` (or `description`?) item would suffice, rather than a section for multiple arbitrary items?  (I suppose it's possible that individual items that can be extracted separately could be useful at a site ... e.g. `cost per cpu hour` or something?).

It would, but I think it's reasonable to leave it open ended - examples might include
```ini
[meta]
description = """
Our Cluster.
"""
url=http://somedocs.internalwebsite.com
max-memory-request=200G
max-cpu-request=12
```

Though to be honest, I'm not completely convinced that this is required.

wxtim:
@hjoliver : I think I've now 
- added the description field.
- justified my decision to leave other fields open ended: Just because Cylc has no internal use for formally defined fields doesn't mean that someone else can't use them. For example, my personal workflows repo has [a script](https://github.com/wxtim/workflows/blob/master/.github/workflows/readmemaker) which GH actions uses to extract metadata and dump it in the repo `README.md`.
- Created a new ticket to discuss the platform listing.

hjoliver:
> added the description field.

I think you might have misinterpreted my comment on this, which was:
> worth putting the common descriptive text in a string constant?

I meant, this descriptive text is identical for platform and group:
```
                Allows writers of platform configurations to add information
                about platform usage. There are no-preset items because
                Cylc does not use any platform metadata internally.

                Users can then see information about defined platforms using::

                   cylc config -i [platforms]

                .. seealso::
                   :ref:`AdminGuide.PlatformConfigs`
```

... so, it should really be stored in a string constant to avoid the duplication.

Also, *There are no-preset items* is now not true since you've added a preset `description` item (which I don't think we need, if going with the arbitrary metadata solution).

wxtim:
@hjoliver  - completely got the wrong end of the stick here - have pushed another commit which hopefully hits nearer the mark.

hjoliver (approved):
That's got it, thanks @wxtim

oliver-sanders (approved):

PR #4562, opened 2021-12-22
CLI: don't color if stdout redirected.

hjoliver:
These changes close #4552  and supersede #4559 

Current colour output disabling is a bit broken.

On master, this correctly generates colour output in the terminal:

```console
$ cylc cat-log test 
```

But these cases (for example) get unnecessary color-code pollution:

```console
$ cylc cat-log test | vim -   # IN YOUR FACE emacs!!! (can't read from stdin)
$ cylc cat-log test > log
```

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.

- [ ] Appropriate tests are included (unit and/or functional).
- [ ] Already covered by existing tests.
- [ ] Does not need tests (why?).

- [ ] Appropriate change log entry included.
- [ ] No change log entry required (why? e.g. invisible to users).

- [x] No documentation update required.

PR #4568, opened 2022-01-05
Tweak terminal colour handling.

hjoliver:
Follow-up to #4562 

- [x] strip colorama markup from command help if not needed
- [x] strip hardwired ansimarkup tags from command help if not needed (`cylc scan --help`)
  

**Requirements check-list**

### DIFF
--- a/cylc/flow/option_parsers.py
+++ b/cylc/flow/option_parsers.py
@@ -133,9 +133,8 @@ class CylcHelpFormatter(IndentedHelpFormatter):
     def _format(self, text):
         """Format help (usage) text on the fly to handle coloring.
 
-        This has to be done here because help is printed to the terminal before
-        command line parsing is completed, and color initialization for general
-        command output is done after that to make use of parsed options.
+        Help is printed to the terminal before color initialization for general
+        command output.
 
         If coloring is wanted:
           - Add color tags to shell examples
@@ -143,22 +142,11 @@ class CylcHelpFormatter(IndentedHelpFormatter):
           - Strip any hardwired color tags
 
         """
-        # Crudely parse sys.argv for --color options here (see docstring above;
-        # only default option values are available here in self.parser.values).
-        # Join the command line with '=' to unify "--opt=val" and "--opt val":
-        joined_args = '='.join(sys.argv[2:])
         use_color = (
-            '--color=always' in joined_args
-            or
-            '--colour=always' in joined_args
+            self.parser.values.color == "always"
             or (
-                supports_color()
-                and not
-                (
-                    '--color=never' in joined_args
-                    or
-                    '--colour=never' in joined_args
-                )
+                self.parser.values.color == "auto"
+                and supports_color()
             )
         )
         if use_color:
--- a/cylc/flow/option_parsers.py
+++ b/cylc/flow/option_parsers.py
@@ -21,16 +21,22 @@ from optparse import OptionParser, OptionConflictError, Values, Option
 import os
 import re
 from ansimarkup import parse as cparse
+from colorama import init as color_init
+
 import sys
 from textwrap import dedent
 from typing import Any, Dict, Optional, List, Tuple
 
 from cylc.flow import LOG, RSYNC_LOG
+from cylc.flow.terminal import supports_color
 import cylc.flow.flags
 from cylc.flow.loggingutil import (
     CylcLogFormatter,
     setup_segregated_log_streams,
 )
+
+
+TAG_REC = re.compile(r'<(.*?)>(.*?)</\1>')
 
 
 def format_shell_examples(string):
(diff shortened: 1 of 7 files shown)


hjoliver:
Note, it's kind of annoying that we have to treat command help separately for color switching (because it is printed by the option parser during option parsing) but I don't think it matters much. (There are more important things to think about right now).

hjoliver:
(Assigned reviewers from the precursor PR)

cylc/flow/option_parsers.py
```diff
@@ -181,10 +188,17 @@ def __init__(
...
-        if '--color=never' not in '='.join(sys.argv[2:]):
-            # Before option parsing, for `--help`, make comments grey in usage.
-            # (This catches both '--color=never' and '--color never'.)
+        joined_args = '='.join(sys.argv[2:])  # for "--opt=arg" and "--opt arg"
+        if (
+            '--color=never' not in joined_args
+            and
+            '--colour=never' not in joined_args
+        ):
+            # Before option parsing, make comments grey in --help output.
```
oliver-sanders, line 197: This logic can be easily broken because the `--color` arg can appear at any position:

```bash
cylc --color=never
cylc config --color=never
cylc config foo --color=never
```

It can also take another format:

```bash
cylc config --color never
```

And won't get disabled automatically in non-interactive modes.

I think we can allow opt_parse to parse the color argument for us, but intercept the usage/description before they are printed using a help-formatter. We recently did something with help formatters in Rose:

https://github.com/metomi/rose/blob/3afc4d9c289433f6fc90f44e5b646ed880dc74d4/metomi/rose/opt_parse.py#L25-L37

I think this will do the trick:

```python
class CylcHelpFormatter(optparse.IndentedHelpFormatter):
    def _format(self, text):
        # note you can access cli opts here via self.parser.values
        if not self.parser.getattr('use_color'):
            text = strip(text)  # strip = AnsiMarkup().strip()
        return text

    def format_usage(self, usage):
        return super().format_usage(self._format(usage))

    def format_descrtiption(self, description):
        return super().format_description(self._format(description))


class CylcOptionParser(OptionParser):
    ...
    def __init__(self):
        ...
        kwargs['formatter'] = CylcHelpFormatter()
        OptionParser.__init__(self, *args, **kwargs)

    ...
    def parse_args(self):
        ...
        (options, args) = OptionParser.parse_args(self, api_args)                          
                                                                   
        self.use_color = (                                               
           opts.color == 'always'                                       
          or (opts.color == 'auto' and cylc.flow.terminal.supports_color())            
        )  
        color_init(autoreset=True, strip=not self.use_color)
```

(with the duplicated lines removed from `cylc.flow.terminal`).

cylc/flow/option_parsers.py
```diff
@@ -181,10 +188,17 @@ def __init__(
...
             usage = format_shell_examples(usage)
+        else:
+            # Strip hardwired colour tags in usage (e.g. "cylc scan --help").
+            usage = TAG_REC.sub(r'\2', usage)
```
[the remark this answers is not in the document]
hjoliver: A-ha, nice. Thanks.

cylc/flow/option_parsers.py
hjoliver: > This logic can be easily broken ... It can also take another format

Nope, the logic handles all of that. Proof (plus, see the tests I added):
shot

It also works with the usage text:  `cylc scan --help | cat -` etc.

hjoliver:
@oliver-sanders - I've gone with a hybrid approach: a help formatter that uses my crude `sys.argv` parsing to get at the color options, because as I noted above help formatters do not have access to parsed option values. (I suppose that makes sense:  help is often printed in response to errors on the command line). 

This is not really any different than what I had before, except that the help formatter is formally nicer than overriding `parser.print_help()` :grin:  However, the internal logic is robust enough (not fragile) it seems to me.

oliver-sanders:
> help formatters do not have access to parsed option values

That didn't make a lick of sense to me since the CLI args have to be parsed in order for the parser to activate the help formatter.

I think I've found the problem, if the `--help` argument was used any other args were stripped from the invoked command causing the strange dual state where the args passed to the parser did not match the args in `sys.argv`.

I think this change should fix the problem making args available to the help formatter.

```diff
diff --git a/cylc/flow/scripts/cylc.py b/cylc/flow/scripts/cylc.py
index 036cf4c05..6fa753315 100644
--- a/cylc/flow/scripts/cylc.py
+++ b/cylc/flow/scripts/cylc.py
@@ -532,7 +532,7 @@ def main():
                 # check if this is a command abbreviation or exit
                 command = match_command(command)
             if opts.help_:
-                execute_cmd(command, "--help")
+                execute_cmd(command, *cmd_args, "--help")
             else:
                 if opts.version:
                     cmd_args.append("--version")
```

MB!

hjoliver:
> I think I've found the problem, if the --help argument was used any other args were stripped from the invoked command causing the strange dual state where the args passed to the parser did not match the args in sys.argv.

OK, brilliant, well done :+1:

hjoliver:
Over-confident declaration of rightness retracted on Element chat!

hjoliver:
I came back to this late in the day (out of time) - I'm not sure why the new unit tests are failing.  Monkeypatching of `sys.argv` appears to have worked (e.g. put `print(sys.argv)` in there) but is not affecting command parsing in the tests. I must have something wrong...

oliver-sanders:
Tests timed out, poking them to see if it is reproducible.

oliver-sanders:
... verily verily, not sure why.

oliver-sanders:
Lets see if a merge sorts things out...

hjoliver:
Tests all passed, except codecov/patch 87.5%

hjoliver:
Minor tweak, should be good to go now.

oliver-sanders (approved):
🌈

oliver-sanders:
(maybe squash)

wxtim (approved):

PR #4570, opened 2022-01-06
Fix host selection bug for deprecated `[runtime][<task>][remote]host` syntax

### LINKED ISSUE #4569: platform: submits to localhost if remotehost not in known_hosts
If a host is not in your `~/.ssh/known_hosts` then the SSH may fail (interactively) like so:

```console
$ ssh somehost
The authenticity of the host 'somehost' can't be established
...
Are you sure you want to continue connecting (yes/no) ?
```

Non-interactively the SSH will fail. This seems to mess with the platforms logic causing it to submit to localhost rather than the remote host.

### To replicate:

First, remove the host from `~/.ssh/known_hosts` and confirm that you get the SSH (yes/no) prompt (ctrl+c to exit).

Then define a platform using that host:

```ini
[platforms]
     [[remote-platform]]
          hosts = <remote-host>
```

Then run a task on that platform:

```ini
[scheduling]
    [[graph]]
        R1 = remote
[runtime]
    [[remote]]
        platform = remote-platform
```

The job will be run on localhost despite being configured otherwise. Inspect the suite or job log to see where it actually ran:

```console
$ cylc cat-log <workflow> remote-platform.1
...
User@Host: <user>@<localhost>
...
```

Next, SSH to the remote host and answer `yes` to the prompt and repeat the experiment:

```console
$ cylc clean <workflow> --rm '.service/db --rm 'log/job'; cylc play <workflow>
```

The job should now be correctly run on the remote host.

MetRonnie:
These changes close #4569

Prevent Cylc from incorrectly falling back to localhost if a host specified by `[runtime][<task>][remote]host` is unreachable.

**Requirements check-list**

### DIFF
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -288,17 +288,18 @@ class TaskJobManager:
 
                     # Get another platform, if task config platform is a group
                     use_next_platform_in_group = False
-                    try:
-                        platform = get_platform(
-                            itask.tdef.rtconfig['platform'],
-                            bad_hosts=self.bad_hosts
-                        )
-                    except NoPlatformsError:
-                        pass
-                    else:
-                        # If were able to select a new platform;
-                        if platform and platform != itask.platform:
-                            use_next_platform_in_group = True
+                    if itask.tdef.rtconfig['platform']:
+                        try:
+                            platform = get_platform(
+                                itask.tdef.rtconfig['platform'],
+                                bad_hosts=self.bad_hosts
+                            )
+                        except NoPlatformsError:
+                            pass
+                        else:
+                            # If were able to select a new platform;
+                            if platform and platform != itask.platform:
+                                use_next_platform_in_group = True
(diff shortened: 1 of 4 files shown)


cylc/flow/task_job_mgr.py
```diff
@@ -288,16 +288,18 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
 
                     # Get another platform, if task config platform is a group
                     use_next_platform_in_group = False
-                    try:
-                        platform = get_platform(
-                            itask.tdef.rtconfig['platform'],
```
MetRonnie, line 293: Problem was `itask.tdef.rtconfig['platform']` is `None` so `get_platform()` fell back to the localhost platform

wxtim (approved):
- [x] Read code.
- [x] New test makes sense. 
- [x] New test fails on master and passes on branch.

oliver-sanders (approved):

PR #4571, opened 2022-01-06
Fix start type log formatting

MetRonnie:
This is a small change with no associated Issue. Fixes lack of space in
```
INFO - cold start from1.
```

**Requirements check-list**
- [x] Does not need tests

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -678,9 +678,11 @@ class Scheduler:
         released from runhead.)
 
         """
-        LOG.info(
-            f"{'warm' if self.options.startcp else 'cold'} start from"
-            f"{self.config.start_point}.")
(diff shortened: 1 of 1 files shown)


cylc/flow/scheduler.py
```diff
@@ -678,9 +678,11 @@ def _load_pool_from_point(self):
         released from runhead.)
 
         """
-        LOG.info(
-            f"{'warm' if self.options.startcp else 'cold'} start from"
-            f"{self.config.start_point}.")
+        start_type = (
+            "Warm" if self.config.start_point > self.config.initial_point
```
MetRonnie, line 682: I feel like this is more robust than checking if `self.options.startcp` is set.

wxtim (approved):

PR #4577, opened 2022-01-13
graph: handle potential failure of PIL.Image.show

oliver-sanders:
Hopefully closes https://github.com/cylc/cylc-doc/issues/375

* We use Pillow to display images (when installed).
* Pillow uses other systems to handle the display itself and includes
  functionality to detect which of these systems are present.
  https://github.com/python-pillow/Pillow/blob/82541b6dec8452cb612067fcebba1c5a1a2bfdc8/src/PIL/ImageShow.py#L117-L263
* We have had reports that this detection has failed for one system
  causing the image not to show with no error displayed.
  https://github.com/cylc/cylc-doc/issues/375
* This change always prints the image file name incase the display
  command fails so that the user is never left high and dry.

**Requirements check-list**

### DIFF
--- a/cylc/flow/scripts/graph.py
+++ b/cylc/flow/scripts/graph.py
@@ -344,10 +344,12 @@ def dot(opts, lines):
 
 def gui(filename):
     """Open the rendered image file."""
+    print(f'Graph rendered to {filename}')
     try:
         from PIL import Image
     except ModuleNotFoundError:
-        print(f'Graph rendered to {filename}')
+        # dependencies required to display images not present
+        pass
     else:
         img = Image.open(filename)
         img.show()


hjoliver (approved):
Dunno what's up with the tests, but it can't be this change.

MetRonnie:
Can we not do something like this: https://stackoverflow.com/a/35305473/3217306

oliver-sanders:
That's what Pillow is doing under the hood, for Linux Pillow supports multiple viewers because you don't know which, if any will be installed:

https://github.com/python-pillow/Pillow/blob/82541b6dec8452cb612067fcebba1c5a1a2bfdc8/src/PIL/ImageShow.py#L117-L263

MetRonnie:
Ah ok. Our use of it seems so simple that we could recreate the viewer detection strategy in few lines of code if we wanted to.

PR #4580, opened 2022-01-14
added cylc get-resources --tutorials

wxtim:
These changes close https://github.com/cylc/cylc-doc/issues/40

The intention with this is that the documentation can set out the following workflow to users:

```
cylc get-resources --tutorials
cd ~/cylc-src/cylc-tutorials/<tutorial-name>

# ... make whatever changes the tutorial requires...

cylc install cylc-tutorials/<tutorial-name>
cylc validate cylc-tutorials/<tutorial-name>
cylc play cylc-tutorials/<tutorial-name>
```

**Requirements check-list**
- [x]  Comprehensive documentation update required: This ticket should be merged first.


cylc/flow/resources.py
```diff
@@ -61,7 +70,10 @@ def get_resources(target_dir, resources=None):
     if resources is None:
         resources = resource_names
     for resource in resources:
-        if resource not in resource_names:
+        if (
+            resource not in resource_names
+            and resource != 'etc/tutorial'
+        ):
```
MetRonnie, line 76: Why not just add `'etc/tutorial'` to `resource_names`?

cylc/flow/resources.py
```diff
@@ -72,3 +84,20 @@ def get_resources(target_dir, resources=None):
         res = pr.resource_string('cylc.flow', resource)
         with open(path, 'wb') as h:
             h.write(res)
+
+
+def extract_tutorials():
+    """Extract tutorial workflows to Cylc Source Directory."""
+    src = Path(cylc.flow.__file__).parent / 'etc/tutorial'
+    dest = glbl_cfg().get(['install', 'source dirs'])[0]
```
MetRonnie, line 92: What if no source dirs are configured?

tests/unit/test_resources.py
```diff
@@ -60,3 +62,20 @@ def test_cli(tmpdir):
...
         )
+
+
+def test_extract_tutorials(monkeypatch, tmp_path, caplog):
+    test_dest = tmp_path/ 'destination'
+    class MockGlobalConfig:
+        def get(*args):
+            return [str(test_dest)]
+
+    monkeypatch.setattr(GlobalConfig, '_DEFAULT', MockGlobalConfig)
```
MetRonnie, line 73: You could also use the `mock_glbl_cfg` fixture
```
mock_glbl_cfg(
    'cylc.flow.resources.glbl_cfg',
    f'''
    [install]
        source dirs = {test_dest}
     '''
)
```

cylc/flow/resources.py
wxtim: Doesn't it have a default value in the config of `$HOME/cylc-src`?

tests/unit/test_resources.py
wxtim: Thanks - I thought that such a thing existed, but I was clearly not thinking straight at the time I wrote this.

cylc/flow/resources.py
wxtim: Because this function handles files, but not directories.

oliver-sanders: We could make the main function support directories and make `~/cylc-src` the default dest directory for resources in the tutorial function. That way users wouldn't have to type the `--tutorial` option and the `[DIR]` positional argument would still work.

MetRonnie (approved):

cylc/flow/resources.py
wxtim: OK, I'll do that then.

hjoliver (changes requested):
I think it's fine to treat the tutorials differently from other resources, but we should update the documentation accordingly.

- This doesn't list the tutorials:
```
  --list                List available package resources.`
```
- This doesn't extract the tutorials:
```Arguments:
   [RESOURCES...]               Resources to extract (default all).
```
- This should say *to ~/cylc-src/cylc-tutorials*?
```
  --tutorials           Extract tutorials workflows to ~/cylc-src
```
- And maybe `cylc get-resources --tutorials` should also print "extracted to ~cylc-src/cylc-tutorials`?

wxtim:
@oliver-sanders I think that I have responded to Hilary's review comments. If you are particularly keen to merge you could check that.

oliver-sanders (changes requested):
Unfortunately I may have forgotten to mention an undocumented feature of `rose tutorial`, that replaced API keys in the tutorials automatically. My bad, I should have spotted this sooner.

oliver-sanders (approved):

cylc/flow/resources.py
```diff
@@ -14,99 +14,180 @@
...
+                            b'DATAPOINT_API_KEY',
+                            api_key.encode(),
+                        )
+                    )
+
+            # then move the tmpfile over the original
+            # NOTE: shutil interfaces don't fully support Path objects at all
+            # python versions
+            path.unlink()
+            shutil.move(str(tmp_path), str(path))
```
MetRonnie, line 193: Why not simplify this
```suggestion
            with open(path, 'rb') as f:
                # NOTE: open the file in bytes mode for safety
                # (prevents decode errors surfacing here)
                contents = f.read()
            contents = contents.replace(
                b'DATAPOINT_API_KEY',
                api_key.encode()
            )
            with open(path, 'wb') as f:
                f.write(contents)
```

cylc/flow/scripts/get_resources.py
```diff
@@ -17,14 +17,34 @@
 """cylc get-resources [OPTIONS] ARGS
 
 Extract resources from the cylc.flow package.
-"""
+
+If the directory is omitted the resource will be coppied either to your
```
MetRonnie, line 21:
```suggestion
If the directory is omitted the resource will be copied either to your
```

cylc/flow/scripts/get_resources.py
```diff
@@ -17,14 +17,34 @@
...
+current working directory for regular resources, or to the first configured
+"source" directory (~/cylc-src by default).
```
MetRonnie, line 23: I think?
```suggestion
"source" directory (~/cylc-src by default) for tutorials.
```

cylc/flow/resources.py
oliver-sanders: More efficient reading/writing one line at a time.

Issue #4036, opened 2021-01-19
compute prerequisites for tasks outside n=0


oliver-sanders:
See https://github.com/cylc/cylc-flow/pull/3938#issuecomment-752652607

Tasks that are not active are listed as having no prerequisites. This is especially unhelpful as we are not especially interested in the prerequisites of an active task.

Suggest computing prerequisites for the entire scheduler window (n=1 by default).

hjoliver:
Need to test current situation again.

Can we interrogate the hidden pool task proxies (used for tracking prerequisites).  Further ahead, we need compute the prerequisites (none of which will be satisfied).

hjoliver:
The datastore should have these anyway, if the node is in the window.  DS to check.

hjoliver:
Need to check partially satisfied prerequisites.

hjoliver:
Currently, the scheduler and DB record partially satisfied prerequisites. (And fully satisfied n=0 tasks). 

Historical tasks can be assumed to have satisfied prerequisites, and future tasks unsatisfied.

However, if the scheduler is down the UIS can't ask it to instantiate a transient task to get its prerequisites.  And, future tasks are not in the DB.

Maybe the scheduler could dump relevant task information to file so that the UIS can use that to compute prerequisites on the fly for any task in the graph, if the scheduler is down?

hjoliver:
Summary as of #4581:

Historical task data in the n-window is now loaded from the DB, but that does not include prerequisite satisfaction status - which we want to infer from task status rather than add to permanent DB bloat.

So currently:
- window expansion (i.e. increase `n` at run time) will cause some succeeded tasks to re-appear that don't show prerequisites satisfied  
- if you trigger a reflow back in the graph, window tasks behind the reflow will not show prereqs satisfied, and those ahead of won't either - but in this case that's correct *for the reflow*

hjoliver:
#4581 is merged

Bumping this Issue to rc2.  

> Historical task data in the n-window is now loaded from the DB, but that does not include prerequisite satisfaction status (except for partially satisfied prerequisites, which are stored in the DB for restarts).

To close this Issue, we need to **infer other prerequisites (fully satisfied or fully unsatisfied) from task status** rather than add to permanent DB bloat.

hjoliver:
@dwsutherland - if you haven't started on this, I'll have a crack at it myself (just taking a look at the code now...)

hjoliver:
[...]

PR #4581, opened 2022-01-17
data-store n > 0 window DB loads

dwsutherland:
These changes partially addresses #4036
(prerequisites were always available, but satisfaction not set)

At present history is only populated via the evolution/forward-march of the active pool.
History is not recovered on restart nor on the window appearing/reappearing about a trigger/reflow.

This pull request aims to fill in the history of those n>0 window nodes:
- [x] Load DB task status and other information of the latest submit.
- [x] Load DB prerequisite satisfaction for the flow (`flow_nums`) of the respective tasks.
- [x] Load DB jobs of the respective tasks.
- [x] Set reflow future window task state to waiting (if not in active-pool/data-store).

This PR also fixes a bug where triggered tasks would hang around after completion, and never get pruned.

**Note: Prerequisite DB history has been taken out (only restart prereqs available and loaded)**
So the prereq satisfaction, shown with GraphiQL below, will no longer be correct for historical tasks.
(However, this is just a one line change.. So, I've left it in)

As an example:
```
[scheduling]
    initial cycle point = 20210101T00
    [[special tasks]]
        clock-trigger = foo(PT0H)
    [[graph]]
        P1M = """
foo => bar => qux => qaz
foo[-P1M] => foo
"""
```
With prerequisites, we can't visualize them at the moment.. But they can be confirmed with GraphiQL:
```
query {
  workflows (ids: ["linear/run1"]) {
    id
    taskProxies {
      id
      flowNums
      state
      prerequisites {
        satisfied
        conditions {
          taskId
          satisfied
        }
      }
    }
  }
}
```
**Trigger reflow:**
```$ cylc trigger --reflow linear //20210201T00/foo```
n-gt-zero-loads-reflow
Data at start:
```
{
  "data": {
    "workflows": [
      {
        "id": "~sutherlander/linear/run1",
        "taskProxies": [
          {
            "id": "~sutherlander/linear/run1//20220101T00/foo",
            "flowNums": "[1]",
            "state": "succeeded",
            "prerequisites": [
              {
                "satisfied": true,
                "conditions": [
                  {
                    "taskId": "20211201T00/foo",
                    "satisfied": true
                  }
                ]
              }
            ]
          },
          {
            "id": "~sutherlander/linear/run1//20210201T00/bar",
            "flowNums": "[2]",
            "state": "waiting",
            "prerequisites": [
              {
                "satisfied": false,
                "conditions": [
                  {
                    "taskId": "20210201T00/foo",
                    "satisfied": false
                  }
                ]
              }
            ]
          },
          {
            "id": "~sutherlander/linear/run1//20210301T00/foo",
            "flowNums": "[2]",
            "state": "waiting",
            "prerequisites": [
              {
                "satisfied": false,
                "conditions": [
                  {
                    "taskId": "20210201T00/foo",
                    "satisfied": false
                  }
                ]
              }
            ]
          },
          {
            "id": "~sutherlander/linear/run1//20220301T00/foo",
            "flowNums": "[1]",
            "state": "waiting",
            "prerequisites": [
              {
                "satisfied": false,
                "conditions": [
                  {
                    "taskId": "20220201T00/foo",
                    "satisfied": false
                  }
                ]
              }
            ]
          },
          {
            "id": "~sutherlander/linear/run1//20210201T00/foo",
            "flowNums": "[2]",
            "state": "preparing",
            "prerequisites": [
              {
                "satisfied": false,
                "conditions": [
                  {
                    "taskId": "20210101T00/foo",
                    "satisfied": false
                  }
                ]
              }
            ]
          },
          {
            "id": "~sutherlander/linear/run1//20210101T00/foo",
            "flowNums": "[1]",
            "state": "succeeded",
            "prerequisites": [
              {
                "satisfied": true,
                "conditions": [
                  {
                    "taskId": "20201201T00/foo",
                    "satisfied": true
                  }
                ]
              }
            ]
          },
          .
          .
          .
        ]
      }
    ]
  }
}
```

(not the past task is loaded in it's previous state, while the future tasks are set to waiting)

**Trigger (noflow):**
```$ cylc trigger linear //20210201T00/bar```
n-gt-zero-loads-noflow
Data at start:
```
{
  "data": {
    "workflows": [
      {
        "id": "~sutherlander/linear/run1",
        "taskProxies": [
          {
            "id": "~sutherlander/linear/run1//20220101T00/foo",
            "flowNums": "[1]",
            "state": "succeeded",
            "prerequisites": [
              {
                "satisfied": true,
                "conditions": [
                  {
                    "taskId": "20211201T00/foo",
                    "satisfied": true
                  }
                ]
              }
            ]
          },
          {
            "id": "~sutherlander/linear/run1//20210201T00/bar",
            "flowNums": "[]",
            "state": "submitted",
            "prerequisites": [
              {
                "satisfied": false,
                "conditions": [
                  {
                    "taskId": "20210201T00/foo",
                    "satisfied": false
                  }
                ]
              }
            ]
          },
          {
            "id": "~sutherlander/linear/run1//20220301T00/foo",
            "flowNums": "[1]",
            "state": "waiting",
            "prerequisites": [
              {
                "satisfied": false,
                "conditions": [
                  {
                    "taskId": "20220201T00/foo",
                    "satisfied": false
                  }
                ]
              }
            ]
          },
          {
            "id": "~sutherlander/linear/run1//20210201T00/qux",
            "flowNums": "[1]",
            "state": "succeeded",
            "prerequisites": [
              {
                "satisfied": true,
                "conditions": [
                  {
                    "taskId": "20210201T00/bar",
                    "satisfied": true
                  }
                ]
              }
            ]
          },
          {
            "id": "~sutherlander/linear/run1//20210201T00/foo",
            "flowNums": "[1]",
            "state": "succeeded",
            "prerequisites": [
              {
                "satisfied": true,
                "conditions": [
                  {
                    "taskId": "20210101T00/foo",
                    "satisfied": true
                  }
                ]
              }
            ]
          },
          .
          .
          .
        ]
      }
    ]
  }
}
```

**Requirements check-list**
- [x] No documentation update required (non-interactive changes).

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -76,13 +76,11 @@ from cylc.flow.workflow_status import get_workflow_status
 from cylc.flow.task_job_logs import JOB_LOG_OPTS, get_task_job_log
 from cylc.flow.task_proxy import TaskProxy
 from cylc.flow.task_state import (
-    TASK_STATUS_WAITING,
     TASK_STATUS_SUBMITTED,
     TASK_STATUS_SUBMIT_FAILED,
     TASK_STATUS_RUNNING,
     TASK_STATUS_SUCCEEDED,
     TASK_STATUS_FAILED,
-    TASK_STATUS_EXPIRED,
     TASK_STATUSES_ORDERED
 )
 from cylc.flow.task_state_prop import extract_group_state
@@ -593,10 +591,8 @@ class DataStoreMgr:
         """Generate graph window about given origin to n-edge-distance.
 
         Args:
-            name (str):
-                Task name.
-            point (cylc.flow.cycling.PointBase):
-                PointBase derived object.
+            itask (cylc.flow.task_proxy.TaskProxy):
+                Update task-node from corresponding task proxy object.
             edge_distance (int):
                 Graph distance from active/origin node.
             active_id (str):
@@ -613,6 +609,12 @@ class DataStoreMgr:
         s_tokens = self.id_.duplicate(itask.tokens)
         if active_id is None:
             active_id = s_tokens.id
+
+        # flag manual triggers for pruning on deletion.
+        if itask.is_manual_submit:
+            self.prune_trigger_nodes.setdefault(active_id, set()).add(
+                s_tokens.id
+            )
 
         # Setup and check if active node is another's boundary node
         # to flag its paths for pruning.
@@ -649,7 +651,7 @@ class DataStoreMgr:
         edge_distance += 1
 
         # Don't expand window about orphan task.
-        if is_orphan is not True:
+        if not is_orphan:
             # TODO: xtrigger is workflow_state edges too
             # Reference set for workflow relations
             for items in itask.graph_children.values():
@@ -756,6 +758,7 @@ class DataStoreMgr:
         ).id
         if tp_id in self.all_task_pool:
             self.all_task_pool.remove(tp_id)
+            self.updates_pending = True
         # flagged isolates/end-of-branch nodes for pruning on removal
         if (
                 tp_id in self.prune_trigger_nodes and
@@ -763,6 +766,7 @@ class DataStoreMgr:
         ):
             self.prune_flagged_nodes.update(self.prune_trigger_nodes[tp_id])
             del self.prune_trigger_nodes[tp_id]
+            self.updates_pending = True
 
     def add_pool_node(self, name, point):
         """Add external ID reference for internal task pool node."""
@@ -785,7 +789,7 @@ class DataStoreMgr:
 
         Returns:
 
-            None
+            True/False
 
         """
         name = itask.tdef.name
@@ -794,11 +798,13 @@ class DataStoreMgr:
         task_proxies = self.data[self.workflow_id][TASK_PROXIES]
 
         is_orphan = False
+        if name not in self.schd.config.taskdefs:
+            is_orphan = True
+
         if tp_id in task_proxies or tp_id in self.added[TASK_PROXIES]:
             return is_orphan
 
-        if name not in self.schd.config.taskdefs:
-            is_orphan = True
+        if is_orphan:
             self.generate_orphan_task(itask)
 
         # Most the time the definition node will be in the store,
@@ -858,7 +864,7 @@ class DataStoreMgr:
             self.process_internal_task_proxy(itask, tproxy)
         else:
             # Batch node for DB load
-            self.db_load_task_proxies[tp_id] = itask
+            self.db_load_task_proxies[(point_string, name)] = itask
 
         self.updates_pending = True
 
@@ -974,11 +980,14 @@ class DataStoreMgr:
         tproxy.state = itask.state.status
         tproxy.flow_nums = json.dumps(list(itask.flow_nums))
 
+        prereq_list = []
         for prereq in itask.state.prerequisites:
             # Protobuf messages populated within
             prereq_obj = prereq.api_dump()
             if prereq_obj:
-                tproxy.prerequisites.append(prereq_obj)
+                prereq_list.append(prereq_obj)
+        del tproxy.prerequisites[:]
+        tproxy.prerequisites.extend(prereq_list)
 
         for label, message, satisfied in itask.state.outputs.get_all():
             output = tproxy.outputs[label]
@@ -1005,7 +1014,7 @@ class DataStoreMgr:
             xtrig.id = sig
             xtrig.label = label
             xtrig.satisfied = satisfied
-            self.xtrigger_tasks.setdefault(sig, set()).add(tp_id)
+            self.xtrigger_tasks.setdefault(sig, set()).add(tproxy.id)
 
         if tproxy.state in self.latest_state_tasks:
             tp_ref = Tokens(tproxy.id).relative_id
@@ -1023,21 +1032,14 @@ class DataStoreMgr:
         flow_db = self.schd.workflow_db_mgr.pri_dao
 
         # Batch load rows with matching cycle & name column pairs.
-        cycle_name_pairs = {
-            (f'{itask.point}', f'{itask.tdef.name}')
-            for itask in self.db_load_task_proxies.values()
-        }
+        cycle_name_pairs = set(self.db_load_task_proxies.keys())
 
-        state_rows = flow_db.select_tasks_for_datastore(cycle_name_pairs)
+        prereq_tasks_args = set()
         for (
                 cycle, name, flow_nums, status, submit_num, outputs_str
-        ) in state_rows:
-            itask = self.db_load_task_proxies[
-                self.id_.duplicate(
-                    cycle=cycle,
-                    task=name,
-                ).id
-            ]
+        ) in flow_db.select_tasks_for_datastore(cycle_name_pairs):
+            itask = self.db_load_task_proxies[(cycle, name)]
+            itask.submit_num = submit_num
             itask.state_reset(status)
             itask.flow_nums = set(json.loads(flow_nums))
             if (
@@ -1051,13 +1053,35 @@ class DataStoreMgr:
                 for message in json.loads(outputs_str).values():
                     itask.state.outputs.set_completion(message, True)
 
-        #prereq_rows = flow_db.select_prereqs_for_datastore(cycle_name_pairs)
+            prereq_tasks_args.add((cycle, name, flow_nums))
 
+        prereqs_rows = flow_db.select_prereqs_for_datastore(prereq_tasks_args)
+        prereqs_map = {}
+        for (
+                cycle, name, prereq_name,
+                prereq_cycle, prereq_output, satisfied
+        ) in prereqs_rows:
+            prereqs_map.setdefault((cycle, name), {})[
+                (prereq_name, prereq_cycle, prereq_output)
+            ] = satisfied if satisfied != '0' else False
 
-        for tp_id, itask in self.db_load_task_proxies.items():
+        for ikey, prereqs in prereqs_map.items():
+            for itask_prereq in (
+                    self.db_load_task_proxies[ikey].state.prerequisites
+            ):
+                for key in itask_prereq.satisfied.keys():
+                    itask_prereq.satisfied[key] = prereqs[key]
+
+        for ikey, itask in self.db_load_task_proxies.items():
+            cycle, name = ikey
             self.process_internal_task_proxy(
                 itask,
-                self.added[TASK_PROXIES][tp_id]
+                self.added[TASK_PROXIES][
+                    self.id_.duplicate(
+                        cycle=cycle,
+                        task=name,
+                    ).id
+                ]
             )
 
         self.db_load_task_proxies.clear()
@@ -1229,11 +1253,13 @@ class DataStoreMgr:
         if not self.prune_flagged_nodes:
             return
 
+        # Keep all nodes in the path of active tasks.
         in_paths_nodes = set().union(*[
             v
             for k, v in self.n_window_nodes.items()
             if k in self.all_task_pool
         ])
+        # Gather all nodes in the paths of tasks flagged for pruning.
         out_paths_nodes = self.prune_flagged_nodes.union(*[
             v
             for k, v in self.n_window_nodes.items()
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -869,12 +869,12 @@ class DataStoreMgr:
             if itask.submit_num > 0:
                 flow_db = self.schd.workflow_db_mgr.pri_dao
                 for row in flow_db.select_jobs_for_datastore(
-                        {(tproxy.cycle_point, tproxy.name)}
+                        {itask.tokens.relative_id}
                 ):
                     self.insert_db_job(1, row)
         else:
             # Batch non-active node for load of DB history.
-            self.db_load_task_proxies[(point_string, name)] = (
+            self.db_load_task_proxies[itask.tokens.relative_id] = (
                 itask,
                 is_parent,
             )
@@ -1044,14 +1044,18 @@ class DataStoreMgr:
 
         flow_db = self.schd.workflow_db_mgr.pri_dao
 
-        cycle_name_pairs = set(self.db_load_task_proxies.keys())
+        task_ids = set(self.db_load_task_proxies.keys())
 
         # Batch load rows with matching cycle & name column pairs.
-        prereq_tasks_args = set()
+        prereq_ids = set()
         for (
                 cycle, name, flow_nums_str, status, submit_num, outputs_str
-        ) in flow_db.select_tasks_for_datastore(cycle_name_pairs):
-            itask, is_parent = self.db_load_task_proxies[(cycle, name)]
+        ) in flow_db.select_tasks_for_datastore(task_ids):
+            tokens = self.id_.duplicate(
+                cycle=cycle,
+                task=name,
+            )
+            itask, is_parent = self.db_load_task_proxies[tokens.relative_id]
             itask.submit_num = submit_num
             flow_nums = set(json.loads(flow_nums_str))
             # Do not set states and outputs for future tasks in flow.
@@ -1076,15 +1080,19 @@ class DataStoreMgr:
                 for message in json.loads(outputs_str).values():
                     itask.state.outputs.set_completion(message, True)
             # Gather tasks with flow id.
-            prereq_tasks_args.add((cycle, name, flow_nums_str))
+            prereq_ids.add(f'{tokens.relative_id}/{flow_nums_str}')
 
         # Batch load prerequisites of tasks according to flow.
         prereqs_map = {}
         for (
                 cycle, name, prereq_name,
                 prereq_cycle, prereq_output, satisfied
-        ) in flow_db.select_prereqs_for_datastore(prereq_tasks_args):
-            prereqs_map.setdefault((cycle, name), {})[
+        ) in flow_db.select_prereqs_for_datastore(prereq_ids):
+            tokens = self.id_.duplicate(
+                cycle=cycle,
+                task=name,
+            )
+            prereqs_map.setdefault(tokens.relative_id, {})[
                 (prereq_cycle, prereq_name, prereq_output)
             ] = satisfied if satisfied != '0' else False
 
@@ -1096,22 +1104,16 @@ class DataStoreMgr:
                     itask_prereq.satisfied[key] = prereqs[key]
 
         # Extract info from itasks to data-store.
-        for ikey, task_info in self.db_load_task_proxies.items():
-            cycle, name = ikey
+        for task_info in self.db_load_task_proxies.values():
             self.process_internal_task_proxy(
                 task_info[0],
                 self.added[TASK_PROXIES][
-                    self.id_.duplicate(
-                        cycle=cycle,
-                        task=name,
-                    ).id
+                    self.id_.duplicate(task_info[0].tokens).id
                 ]
             )
 
-        flow_db.select_prereqs_for_datastore(prereq_tasks_args)
-
         # Batch load jobs from DB.
-        for row in flow_db.select_jobs_for_datastore(cycle_name_pairs):
+        for row in flow_db.select_jobs_for_datastore(task_ids):
             self.insert_db_job(1, row)
 
         self.db_load_task_proxies.clear()
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -1084,6 +1084,10 @@ class DataStoreMgr:
                 ]
             )
 
+        # Load jobs from DB.
+        for row in flow_db.select_jobs_for_datastore(cycle_name_pairs):
+            self.insert_db_job(1, row)
+
         self.db_load_task_proxies.clear()
 
     def insert_job(self, name, point_string, status, job_conf):
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -398,6 +398,7 @@ class DataStoreMgr:
         self.n_window_nodes = {}
         self.n_window_edges = {}
         self.n_window_boundary_nodes = {}
+        self.db_load_task_proxies = {}
         self.family_pruned_ids = set()
         self.prune_trigger_nodes = {}
         self.prune_flagged_nodes = set()
@@ -677,6 +678,8 @@ class DataStoreMgr:
                     True,
                 )
 
+        # If this is the active task (edge_distance has been incremented),
+        # then add the most distant child as a trigger to prune it.
         if edge_distance == 1:
             levels = self.n_window_boundary_nodes[active_id].keys()
             # Could be self-reference node foo:failed => foo
@@ -818,14 +821,7 @@ class DataStoreMgr:
             ),
             depth=task_def.depth,
             name=name,
-            state=TASK_STATUS_WAITING,
-            flow_nums=json.dumps(list(itask.flow_nums))
         )
-        if is_parent and tp_id not in self.n_window_nodes:
-            # TODO: Load task info from DB, including itask prerequisites
-            tproxy.state = TASK_STATUS_EXPIRED
-        else:
-            tproxy.state = TASK_STATUS_WAITING
 
         tproxy.namespace[:] = task_def.namespace
         if is_orphan:
@@ -846,39 +842,6 @@ class DataStoreMgr:
             ]
         tproxy.first_parent = tproxy.ancestors[0]
 
-        for prereq in itask.state.prerequisites:
-            # Protobuf messages populated within
-            prereq_obj = prereq.api_dump()
-            if prereq_obj:
-                tproxy.prerequisites.append(prereq_obj)
-
-        for label, message, satisfied in itask.state.outputs.get_all():
-            output = tproxy.outputs[label]
-            output.label = label
-            output.message = message
-            output.satisfied = satisfied
-            output.time = update_time
-
-        if itask.tdef.clocktrigger_offset is not None:
-            tproxy.clock_trigger.satisfied = itask.is_waiting_clock_done()
-            tproxy.clock_trigger.time = itask.clock_trigger_time
-            tproxy.clock_trigger.time_string = time2str(
-                itask.clock_trigger_time)
-
-        for trig, satisfied in itask.state.external_triggers.items():
-            ext_trig = tproxy.external_triggers[trig]
-            ext_trig.id = trig
-            ext_trig.satisfied = satisfied
-
-        for label, satisfied in itask.state.xtriggers.items():
-            sig = self.schd.xtrigger_mgr.get_xtrig_ctx(
-                itask, label).get_signature()
-            xtrig = tproxy.xtriggers[sig]
-            xtrig.id = sig
-            xtrig.label = label
-            xtrig.satisfied = satisfied
-            self.xtrigger_tasks.setdefault(sig, set()).add(tp_id)
-
         self.added[TASK_PROXIES][tp_id] = tproxy
         getattr(self.updated[WORKFLOW], TASK_PROXIES).append(tp_id)
         self.updated[TASKS].setdefault(
@@ -890,12 +853,13 @@ class DataStoreMgr:
         ).proxies.append(tp_id)
         self.generate_ghost_family(tproxy.first_parent, child_task=tp_id)
         self.state_update_families.add(tproxy.first_parent)
-        if tproxy.state in self.latest_state_tasks:
-            tp_ref = Tokens(tproxy.id).relative_id
-            tp_queue = self.latest_state_tasks[tproxy.state]
-            if tp_ref in tp_queue:
-                tp_queue.remove(tp_ref)
-            self.latest_state_tasks[tproxy.state].appendleft(tp_ref)
+
+        if tp_id in self.n_window_nodes:
+            self.process_internal_task_proxy(itask, tproxy)
+        else:
+            # Batch node for DB load
+            self.db_load_task_proxies[tp_id] = itask
+
         self.updates_pending = True
 
         return is_orphan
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: cylc/flow/workflow_db_mgr.py [hunk 1]
```diff
@@ -440,7 +440,8 @@ def put_task_pool(self, pool: 'TaskPool') -> None:
         relevant insert statements for the current tasks in the pool.
         """
         self.db_deletes_map[self.TABLE_TASK_POOL].append({})
-        self.db_deletes_map[self.TABLE_TASK_PREREQUISITES].append({})
+        # We now keep prereq history:
```
MetRonnie, line 443: Might be a bit clearer
```suggestion
        # We now keep prereq history, no need to do:
```

MetRonnie (commented):
Code looks good. However I don't think I'm knowledgeable enough about reflows and the data store to give a proper sign off, might need to wait for @hjoliver

MetRonnie (changes requested):
Ah, actually, 1 change that is needed:
```diff
diff --git a/cylc/flow/rundb.py b/cylc/flow/rundb.py
index 71a932577..abf4ec0e7 100644
--- a/cylc/flow/rundb.py
+++ b/cylc/flow/rundb.py
@@ -168,7 +168,7 @@ class CylcWorkflowDAO:
     CONN_TIMEOUT = 0.2
     DB_FILE_BASE_NAME = "db"
     MAX_TRIES = 100
-    RESTART_INCOMPAT_VERSION = "8.0b1"  # Can't restart if <= this version
+    RESTART_INCOMPAT_VERSION = "8.0rc1"  # Can't restart if <= this version
     TABLE_BROADCAST_EVENTS = "broadcast_events"
     TABLE_BROADCAST_STATES = "broadcast_states"
     TABLE_INHERITANCE = "inheritance"
```

cylc/flow/workflow_db_mgr.py
dwsutherland: Done.

dwsutherland:
> Ah, actually, 1 change that is needed:
> 

Done, thanks.

dwsutherland:
> > Ah, actually, 1 change that is needed:
> 
> Done, thanks.

Actually, not sure it worked.. cause we're at 8.0rc1.dev0

```
ERROR    cylc:scheduler.py:1679 Workflow shutting down - ServiceFileError: Cannot restart - Workflow database is incompatible with Cylc 8.0rc1.dev (workflow last run with Cylc 8.0rc1.dev0)
```

MetRonnie:
D'oh, I should have said `RESTART_INCOMPAT_VERSION = '8.0b3'` instead of `8.0rc1`

dwsutherland:
> D'oh, I should have said `RESTART_INCOMPAT_VERSION = '8.0b3'` instead of `8.0rc1`

Done.

hjoliver:
One thing, from an earlier conversation with @oliver-sanders, probably can be a follow-up optimization though:

We need to avoid unbounded DB growth as far as possible (although #4608 could solve that for endless operational workflows) because that will eventually impact on scheduler performance. 

We should be able to avoid the unbounded prerequisites table by only storing partially-satisfied prerequisites (and perhaps those of any force-triggered tasks too).

If a task started running, its prerequisites can be inferred as satisfied. 

Otherwise, if a task doesn't have partially satisfied prerequisites, they can be inferred as entirely unsatisfied. 

For conditional prerequisites we can't infer the individual upstream task status like that, but we can infer that the conditional prerequisites expression was satisfied, which is good enough (as we can query the upstream tasks as well if needed).

hjoliver:
Hmmm. I must be doing something wrong, as all the tests passed...

```
[scheduling]
    cycling mode = integer
    initial cycle point = 1
   [[graph]]
     P1 = "foo => bar"
[runtime]
   [[foo, bar]]
```

In my environment this runs on master, but on this branch I get:

```console
...
2022-01-25T17:45:52+13:00 INFO - [1/foo waiting job:00 flows:1] => preparing
2022-01-25T17:45:52+13:00 INFO - [2/foo waiting job:00 flows:1] => preparing
2022-01-25T17:45:52+13:00 INFO - [3/foo waiting job:00 flows:1] => preparing
2022-01-25T17:45:52+13:00 INFO - [4/foo waiting job:00 flows:1] => preparing
2022-01-25T17:45:52+13:00 INFO - [5/foo waiting job:00 flows:1] => preparing
2022-01-25T17:45:52+13:00 INFO - [1/foo preparing job:01 flows:1] host=niwa-1007885.niwa.local
2022-01-25T17:45:52+13:00 INFO - [2/foo preparing job:01 flows:1] host=niwa-1007885.niwa.local
2022-01-25T17:45:52+13:00 INFO - [3/foo preparing job:01 flows:1] host=niwa-1007885.niwa.local
2022-01-25T17:45:52+13:00 INFO - [4/foo preparing job:01 flows:1] host=niwa-1007885.niwa.local
2022-01-25T17:45:52+13:00 INFO - [5/foo preparing job:01 flows:1] host=niwa-1007885.niwa.local
2022-01-25T17:45:52+13:00 ERROR - near ",": syntax error
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/scheduler.py", line 615, in start_scheduler
        await self.main_loop()
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/scheduler.py", line 1538, in main_loop
        has_updated = await self.update_data_structure()
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/scheduler.py", line 1599, in update_data_structure
        self.data_store_mgr.update_data_structure(
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/data_store_mgr.py", line 1238, in update_data_structure
        self.apply_task_proxy_db_history()
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/data_store_mgr.py", line 1048, in apply_task_proxy_db_history
        ) in flow_db.select_tasks_for_datastore(cycle_name_pairs):
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/rundb.py", line 914, in select_tasks_for_datastore
        return list(self.connect().execute(stmt))
    sqlite3.OperationalError: near ",": syntax error
2022-01-25T17:45:52+13:00 CRITICAL - Workflow shutting down - near ",": syntax error
2022-01-25T17:45:53+13:00 INFO - DONE
```

Any ideas?

dwsutherland:
> Any ideas?

Something to do with an empty request I think., Will put a request in soon..

Will respond to the DB stuff soon.. I think the Job and task states table also grow unbounded (?), so not unique to prerequisites..

hjoliver:
Yes, but job and task states are essential history, whereas prerequisites aren’t (the latter can be derived from the former)

dwsutherland:
> Yes, but job and task states are essential history, whereas prerequisites aren’t (the latter can be derived from the former)

Ok, well, if you're happy to hold a finite history of them...
I can jettison them based on some criterion.. If there was a timestamp we could cull based on that, because we need some history not in the pool or window (otherwise what's the point at all?).

hjoliver:
I’m only talking about the database. We *do* have to store partially satisfied prereqs (actually already did that for SOD!), but otherwise we can infer prereq satisfaction from task state.  So if you have to go to the database for task data then (a) check for partially satisfied prereqs, else (b) infer from the state recorded in the DB.

dwsutherland:
> I’m only talking about the database. We _do_ have to store partially satisfied prereqs (actually already did that for SOD!), but otherwise we can infer prereq satisfaction from task state. So if you have to go to the database for task data then (a) check for partially satisfied prereqs, else (b) infer from the state recorded in the DB.

Even before this PR the prerequisites were available and updated by the active tasks.

Conditionals could be hard to infer from the task state.

dwsutherland:
@hjoliver - Prereq history is no longer kept..

And I've fixed that bug you encountered, give it a test.

oliver-sanders:
We have the `task_events` table which is one of the perma store DB tables. I think, with the exception of `flow_nums`, we should have everything in that table required to play back the graph evolution.

I think there should be enough info there to reconstruct conditionals (at least reconstruct at a specific point in time e.g. for `a | b => c` the task `a` might trigger `c`, however, `b` might succeed at a later time).

Three cases for prereqs:

1) Outputs satisfied naturally.
2) Some or all output satisfied artificially by `cylc set-outputs`.
3) Outputs ignored by `cylc trigger`.

For case (1) I think all the required information to reconstruct the prereqs at a later date is written to the `task_events` table (except possibly `flow_nums`).

I'm not sure if we get `task_events` entries for cases (2) & (3), if not it would be good to add them there as a permanent record of user intervention (#4609).

dwsutherland:
We have the task_events table which is one of the perma store DB tables. I think, with the exception of flow_nums, we should have everything in that table required to play back the graph evolution.

Yes, we should be able to construct it from that table.. Follow-on PR for sure.
We have `submit_num` in that table, so we can tie it to any particular `flow_num`.

hjoliver:
The query syntax here uses (e.g.):
```
...
WHERE
        (task_states.cycle, task_states.name) IN (
    VALUES ('3', 'bar'), ('5', 'bar'), ('2', 'bar'), ('1', 'bar'), ('4', 'bar'), ('6', 'bar')
...
```
which is only supported since sqlite v 3.16.0 (2016) https://www.sqlite.org/rowvalue.html (see **4. Backwards Compatibility**)

My work Centos 7 box has 3.7.17 (2013)!

**QUESTION** do we need to support older sqlite versions?  If so, need use different query syntax.

dwsutherland:
Have changed the query for back compat to:
```
.
.
.
            WHERE
                %(task_states)s.cycle || '/' || %(task_states)s.name IN (
                    %(task_ids)s
                )
            ORDER BY
                %(task_states)s.submit_num DESC
        """
        form_data = {
            "task_states": self.TABLE_TASK_STATES,
            "task_jobs": self.TABLE_TASK_JOBS,
            "task_ids": ', '.join(f"'{val}'" for val in task_ids),
        }
```

Should work now.

oliver-sanders:
> QUESTION do we need to support older sqlite versions? If so, we need to use different query syntax.

Belated (and now obsolete answer), I think we should support whatever Python supports. We have no way of specifying the sqlite3 in our pypi deps.

hjoliver:
Ran into a bug with a `n=2` window.

```diff
         self.updated_state_families = set()
-        self.n_edge_distance = 1
+        self.n_edge_distance = 2
         self.next_n_edge_distance = None
         self.latest_state_tasks = {
```

with
```ini
[scheduling]
   [[graph]]
     R1 = "foo => bar => baz"
[runtime]
   [[foo, bar, baz]]
      script = sleep 10
```

result:
```console
$ cylc install && cylc play -n junk   
INSTALLED junk/run3 from /home/oliverh/cylc-src/junk

                Cylc Workflow Engine 8.0rc1.dev
                Copyright (C) 2008-2022 NIWA

2022-01-27T15:24:55+13:00 INFO - Extracting etc/job.sh to /home/oliverh/cylc-run/junk/run3/.service/etc/job.sh
2022-01-27T15:24:55+13:00 INFO - Workflow: junk/run3
2022-01-27T15:24:56+13:00 INFO - Cold start from 1
2022-01-27T15:24:56+13:00 INFO - New flow: 1 (original flow from 1) 2022-01-27 15:24:56
2022-01-27T15:24:56+13:00 INFO - Scheduler: url=tcp://niwa-1007885.niwa.local:43017/ pid=28662
2022-01-27T15:24:56+13:00 INFO - Workflow publisher: url=tcp://niwa-1007885.niwa.local:43037
2022-01-27T15:24:56+13:00 INFO - Run: (re)start=0 log=1
2022-01-27T15:24:56+13:00 INFO - Cylc version: 8.0rc1.dev
2022-01-27T15:24:56+13:00 INFO - Run mode: live
2022-01-27T15:24:56+13:00 INFO - Initial point: 1
2022-01-27T15:24:56+13:00 INFO - Final point: 1
2022-01-27T15:24:56+13:00 INFO - [1/foo waiting(runahead) job:00 flows:1] => waiting
2022-01-27T15:24:56+13:00 INFO - [1/foo waiting job:00 flows:1] => waiting(queued)
2022-01-27T15:24:56+13:00 INFO - [1/foo waiting(queued) job:00 flows:1] => waiting
2022-01-27T15:24:56+13:00 INFO - Queue released: 1/foo
2022-01-27T15:24:56+13:00 INFO - [1/foo waiting job:00 flows:1] => preparing
2022-01-27T15:24:56+13:00 INFO - [1/foo preparing job:01 flows:1] host=niwa-1007885.niwa.local
2022-01-27T15:24:57+13:00 INFO - [1/foo preparing job:01 flows:1] (internal)submitted at 2022-01-27T15:24:56+13:00
2022-01-27T15:24:57+13:00 INFO - [1/foo preparing job:01 flows:1] submitted to localhost:background[28674]
2022-01-27T15:24:57+13:00 INFO - [1/foo preparing job:01 flows:1] => submitted
2022-01-27T15:24:57+13:00 INFO - [1/foo submitted job:01 flows:1] health: submission timeout=None, polling intervals=PT15M,...
2022-01-27T15:24:58+13:00 INFO - [1/foo submitted job:01 flows:1] (received)started at 2022-01-27T15:24:57+13:00
2022-01-27T15:24:58+13:00 INFO - [1/foo submitted job:01 flows:1] => running
2022-01-27T15:24:58+13:00 INFO - [1/foo running job:01 flows:1] health: execution timeout=None, polling intervals=PT15M,...
2022-01-27T15:25:08+13:00 INFO - [1/foo running job:01 flows:1] (received)succeeded at 2022-01-27T15:25:07+13:00
2022-01-27T15:25:08+13:00 INFO - [1/foo running job:01 flows:1] => succeeded
2022-01-27T15:25:08+13:00 INFO - [1/bar waiting(runahead) job:00 flows:1] spawned
2022-01-27T15:25:08+13:00 ERROR - '~oliverh/junk/run3//1/bar'
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/scheduler.py", line 615, in start_scheduler
        await self.main_loop()
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/scheduler.py", line 1532, in main_loop
        self.process_queued_task_messages()
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/scheduler.py", line 813, in process_queued_task_messages
        if self.task_events_mgr.process_message(
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/task_events_mgr.py", line 473, in process_message
        self.spawn_func(itask, TASK_OUTPUT_SUCCEEDED)
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/task_pool.py", line 1181, in spawn_on_output
        self.add_to_pool(t)
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/task_pool.py", line 193, in add_to_pool
        self.create_data_store_elements(itask)
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/task_pool.py", line 216, in create_data_store_elements
        self.data_store_mgr.increment_graph_window(itask)
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/data_store_mgr.py", line 674, in increment_graph_window
        self._expand_graph_window(
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/data_store_mgr.py", line 745, in _expand_graph_window
        self.increment_graph_window(
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/data_store_mgr.py", line 685, in increment_graph_window
        levels = self.n_window_boundary_nodes[active_id].keys()
    KeyError: '~oliverh/junk/run3//1/bar'
2022-01-27T15:25:08+13:00 CRITICAL - Workflow shutting down - '~oliverh/junk/run3//1/bar'
2022-01-27T15:25:08+13:00 INFO - DONE
```

hjoliver:
@dwsutherland - this now populates `n > 0` task data from the DB :+1: but it doesn't close #4036 - if inferring prereqs is to be a follow-up can you remove the *closes* directive from the issue description above plz?

dwsutherland:
> Ran into a bug with a n=2 window.

Fixed

dwsutherland:
> @dwsutherland - this now populates `n > 0` task data from the DB 👍 but it doesn't close #4036 - if inferring prereqs is to be a follow-up can you remove the _closes_ directive from the issue description above plz?

Prerequisites were always computed, this was just making sure they were also available in the n>0 window (on restart)..

The thing I took out was the history for anything other than a restart.

hjoliver:
Summary of my understanding of this PR *with regard to prerequisites*:

https://github.com/cylc/cylc-flow/issues/4036#issuecomment-1022860303

hjoliver (approved):
Approved!

Approval may be slightly premature given all the action on this in the last couple of days, but a) I'm about to go on leave for 5 days; b) now that I understand where we are w.r.t. prerequisite availability I agree this is a good improvement as-is.

dwsutherland:
Just a point about prereqs:
- They are available on all tasks
- They are loaded from the DB if available (on restart or otherwise), which is usually just the n=1 future tasks with partially satisfied prerequisites.
- All tasks that were active but have never left the window, and all future tasks will have the correct state of prerequisites.

This is what has been done here.

Only one fringe case remains:
- Historical tasks (succeeded/not-waiting/not-active) that spawn about a triggered task, reflow, restart. The prerequisites are available, but satisfaction not properly set.

oliver-sanders:
Thanks @dwsutherland 

Sorry, I've been distracted bug chasing, code looks good, got some testing to do.

oliver-sanders:
More bugs more distractions :angry:

oliver-sanders:
Yes more bugs 😢 but trying to review this at the same time. Testing this alongside `master` trying to spot the differences.

@dwsutherland could you elaborate on the four points in the PR description to help me understand the changes better:

> Load DB task status and other information of the latest submit.

So this fixes task/job info that wasn't being loaded properly on restart?

> Load DB prerequisite satisfaction for the flow (flow_nums) of the respective tasks.

I don't quite understand what this means. Is it that we weren't checking `flow_nums` before so could get the wrong result if multiple flows were present?

> Load DB jobs of the respective tasks.
> Set reflow future window task state to waiting (if not in active-pool/data-store).

I think these two points are about triggering a task outside the active window and the scheduler reloading the tasks around it from the DB?

dwsutherland:
> > Load DB task status and other information of the latest submit.
> 
> So this fixes task/job info that wasn't being loaded properly on restart?

For tasks in the n>0 respective tasks include:
- Restart
- Reappearance (parent becomes inactive, but another parent becomes active later)
- The window surrounding forced triggered tasks, usually non-active tasks.
- Window paternal of reflow

The following fields of the latest submit are populated from the DB:
`flow_nums_str`, `status`, `submit_num`, `outputs_str`
(states and other information)
 
> > Load DB prerequisite satisfaction for the flow (flow_nums) of the respective tasks.
> 
> I don't quite understand what this means. Is it that we weren't checking `flow_nums` before so could get the wrong result if multiple flows were present?

We use the cycle, name, and flow number (found in the above first point) to extract the matching prerequisites from the DB.

The DB info covers restart and reappearing tasks (as the DB only holds enough prerequisite info for existing and/or waiting/hidden-pool).

 
> > Load DB jobs of the respective tasks.
> > Set reflow future window task state to waiting (if not in active-pool/data-store).
> 
> I think these two points are about triggering a task outside the active window and the scheduler reloading the tasks around it from the DB?

The loading jobs from the DB applies to all respective tasks; restart, reappearance, trigger, trigger reflow.
(if the submit number is > 0 and a new task to the window, then even active tasks will have their job history loaded)

The future tasks of a reflow only have jobs loaded, their status is set to waiting..

All this is new, not currently done in master... 
In master things might be fudged; prerequisites there but not satisfied, task states set to `expired` state if reappeared or restart or historical n>0, and no jobs loaded from the DB.

hjoliver:
@oliver-sanders if you haven't seen this comment: https://github.com/cylc/cylc-flow/issues/4036#issuecomment-1022860303 (was based on my understanding of this PR)

MetRonnie:
Might be useful to have some simple examples to compare before and after in the GUI or TUI

oliver-sanders:
That's what I'm trying to come up with.

I think try restarts, reflows and triggers outside of the n=1 window?

datamel (commented):
I have had a look at this, trying to review rc1 tagged prs where I can. 
I suspect there is a problem with the database on this branch. 
Running `tests/functional/job-submission/01-job-nn-localhost.t` locally for me produces an error: 
`Error: unable to open database "../cylctb-20220208T111512Z-ulqT/functional/job-submission/01-job-nn-localhost/.service/db": unable to open database file`
This test passes locally on master pretty quickly. 
Error confirmed by @oliver-sanders.

dwsutherland:
> I have had a look at this, trying to review rc1 tagged prs where I can. I suspect there is a problem with the database on this branch. Running `tests/functional/job-submission/01-job-nn-localhost.t` locally for me produces an error: `Error: unable to open database "../cylctb-20220208T111512Z-ulqT/functional/job-submission/01-job-nn-localhost/.service/db": unable to open database file` This test passes locally on master pretty quickly. Error confirmed by @oliver-sanders.

Not an error of this branch, a rebase fixed it (the base master had the same issue).

dwsutherland:
> Might be useful to have some simple examples to compare before and after in the GUI or TUI

I'll run the examples in the description using `master`, and post up the gifs.  (it's pretty obvious, especially the jobs)

>  I think try restarts, reflows and triggers outside of the n=1 window?

Yes, do a trigger of a historical task..

Also a reflow of something like this:
` a | b => c`
from `a`, should show `b` history when `c` becomes active (`n=0`)...

oliver-sanders:
> I'll run the examples in the description using master, and post up the gifs

Happy to run the testing ourselves, if you could recommend some experiments for us to try out.

dwsutherland:
> if you could recommend some experiments for us to try out.

The example in the description is good to play around with. Let it run to the end then .. then do restarts, historical triggers and reflows.

cylc/flow/data_store_mgr.py
```diff
@@ -1002,6 +985,139 @@ def generate_ghost_family(self, fp_id, child_fam=None, child_task=None):
...
+            )
+            prereqs_map.setdefault(tokens.relative_id, {})[
+                (prereq_cycle, prereq_name, prereq_output)
+            ] = satisfied if satisfied != '0' else False
+
+        for ikey, prereqs in prereqs_map.items():
+            for itask_prereq in (
+                    self.db_load_task_proxies[ikey][0].state.prerequisites
+            ):
+                for key, _ in itask_prereq.satisfied.items():
```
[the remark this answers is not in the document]
dwsutherland: Fixed.. was transplanted from another piece of code  >.<

oliver-sanders:
Testing going well, I've tried various nasty triggers and reflows, information has all been dragged in from the DB correctly.

We could do with a nicer error message for incompatible databases as there are now a few folks out there using cylc-8.0b3. Nothing fancy needed, just something to make it clear this is a purposeful error not a random traceback.

MetRonnie:
Actually, not sure what could be improved here?
```
INFO - Extracting job.sh to /net/home/h04/rdutta/cylc-run/linear/.service/etc/job.sh
INFO - Workflow: linear
ERROR - Workflow shutting down - ServiceFileError: Cannot restart - Workflow database is
    incompatible with Cylc 8.0rc1.dev (workflow last run with Cylc 8.0b3)
INFO - DONE
```

oliver-sanders:
> Actually, not sure what could be improved here?

That's not what I get:

```console
$ git checkout upstream/master
$ cylc play <id>
$ cylc stop <id>   # let it stop with tasks in the pool
$ git checkout dwsutherland/n-gt-zero-db-loads
$ cylc play <id>
...
2022-02-10T10:16:54Z ERROR - no such column: flow_nums
      File "~/cylc-flow/cylc/flow/scheduler.py", line 657, in run
        await self.configure()
      File "~/cylc-flow/cylc/flow/scheduler.py", line 481, in configure
        self._load_pool_from_db()
      File "~/cylc-flow/cylc/flow/scheduler.py", line 721, in _load_pool_from_db
        self.pool.load_db_task_pool_for_restart)
      File "~/cylc-flow/cylc/flow/rundb.py", line 824, in select_task_pool_for_restart
        callback(row_idx, list(row))
      File "~/cylc-flow/cylc/flow/task_pool.py", line 439, in load_db_task_pool_for_restart
        flow_nums,
      File "~/cylc-flow/cylc/flow/rundb.py", line 844, in select_task_prerequisites
        return list(self.connect().execute(stmt, stmt_args))
    sqlite3.OperationalError: no such column: flow_nums
2022-02-10T10:16:54Z CRITICAL - Workflow shutting down - no such column: flow_nums
```

MetRonnie:
That'll only happen for workflows run with 8.0rc1.dev prior to this PR being merged

dwsutherland:
Ya, the old DB is incompatible for restart.. So I guess that should be handled better?

Probably a follow-on PR though?

oliver-sanders:
Ah, ok if it won't happen for the released code.

MetRonnie:
> Ya, the old DB is incompatible for restart.. So I guess that should be handled better?
> 
> Probably a follow-on PR though? Do you want it in for rc1?

I don't think we need to do anything more. It was decided a while ago that DB incompatibility between pre-release versions is ok.

cylc/flow/data_store_mgr.py
```diff
@@ -1002,6 +985,139 @@ def generate_ghost_family(self, fp_id, child_fam=None, child_task=None):
...
+                    and itask.state(
+                        TASK_STATUS_RUNNING,
+                        TASK_STATUS_FAILED,
+                        TASK_STATUS_SUCCEEDED
+                    )
+            ):
+                for message in json.loads(outputs_str).values():
+                    itask.state.outputs.set_completion(message, True)
+            # Gather tasks with flow id.
+            prereq_ids.add(f'{tokens.relative_id}/{flow_nums_str}')
```
oliver-sanders, line 1083: Because `flow_nums` are defined as a set on the TaskProxy there is no guarantee of order when they are written to the DB which could mess with this comparison logic.

We should push serialisation/deserialisation through a sort to make it safe. Seems to work now so we can do this later - https://github.com/cylc/cylc-flow/issues/4672

cylc/flow/data_store_mgr.py
```diff
@@ -1002,6 +985,139 @@ def generate_ghost_family(self, fp_id, child_fam=None, child_task=None):
...
+
+        # Batch load prerequisites of tasks according to flow.
+        prereqs_map = {}
+        for (
+                cycle, name, prereq_name,
+                prereq_cycle, prereq_output, satisfied
+        ) in flow_db.select_prereqs_for_datastore(prereq_ids):
```
oliver-sanders, line 1090: FYI: we have ditched `pylint` which is the tool which used to tell us to double indent these statements.

oliver-sanders (approved):
:+1:

From experimentation when I trigger a historical task I get four DB queries:

1. Fetches the triggered task & its jobs.
2. Fetches the n=1 tasks.
3. Fetches the n=1 prerequisites (specifying the flow numbers).
4. Fetches the n=1 jobs.

> **Note:** Queries (1) & (2) don't specify the flow numbers because they are loading historical tasks

Here are the DB calls for triggering task `2/b` in the graph:

```
a => b => c
b[-P1] => b
```

```sql
# Fetches the triggered task & its jobs.
                SELECT
                    task_states.cycle,
                    task_states.name,
                    task_states.status,
                    task_states.submit_num,
                    task_jobs.time_submit,
                    task_jobs.time_run,
                    task_jobs.time_run_exit,
                    task_jobs.job_runner_name,
                    task_jobs.job_id,
                    task_jobs.platform_name
                FROM
                    task_jobs
                JOIN
                    task_states
                ON  task_jobs.cycle == task_states.cycle AND
                    task_jobs.name == task_states.name AND
                    task_jobs.submit_num == task_states.submit_num
                WHERE
                    task_states.cycle || '/' || task_states.name IN (
                        '2/b'
                    )
                ORDER BY
                    task_states.submit_num DESC

# 2. Fetches the n=1 tasks.
                SELECT
                    task_states.cycle,
                    task_states.name,
                    task_states.flow_nums,
                    task_states.status,
                    MAX(task_states.submit_num),
                    task_outputs.outputs
                FROM
                    task_states
                LEFT OUTER JOIN
                    task_outputs
                ON  task_states.cycle == task_outputs.cycle AND
                    task_states.name == task_outputs.name
                WHERE
                    task_states.cycle || '/' || task_states.name IN (
                        '1/b', '2/a', '2/c', '3/b'
                    )
                GROUP BY
                    task_states.cycle, task_states.name

# 3. Fetches the n=1 prerequisites (specifying the flow numbers).
                SELECT
                    cycle,
                    name,
                    prereq_name,
                    prereq_cycle,
                    prereq_output,
                    satisfied
                FROM
                    task_prerequisites
                WHERE
                    cycle || '/' || name || '/' || flow_nums IN (
                        '2/c/[1]', '2/a/[1]', '3/b/[1]', '1/b/[1]'
                    )

# 4. Fetches the n=1 jobs.
                SELECT
                    task_states.cycle,
                    task_states.name,
                    task_states.status,
                    task_states.submit_num,
                    task_jobs.time_submit,
                    task_jobs.time_run,
                    task_jobs.time_run_exit,
                    task_jobs.job_runner_name,
                    task_jobs.job_id,
                    task_jobs.platform_name
                FROM
                    task_jobs
                JOIN
                    task_states
                ON  task_jobs.cycle == task_states.cycle AND
                    task_jobs.name == task_states.name AND
                    task_jobs.submit_num == task_states.submit_num
                WHERE
                    task_states.cycle || '/' || task_states.name IN (
                        '1/b', '2/a', '2/c', '3/b'
                    )
                ORDER BY
                    task_states.submit_num DESC
```

Checked some GraphiQL queries which show the flow_numbs are being loaded correctly. All good thanks @dwsutherland.

oliver-sanders:
(leaving the PR up for a short while incase any other reviews are still working on this)

hjoliver:
Two approvals, and no one has raise any unresolved problems ... merging. Thanks @dwsutherland

cylc/flow/data_store_mgr.py
dwsutherland: Oh right, what's your preference? this:
```
        for (
            cycle, name,
            prereq_name,
            prereq_cycle,
            prereq_output,
            satisfied
        ) in flow_db.select_prereqs_for_datastore(prereq_ids):
```

hjoliver: Yep. (Indent-wise; not necessarily each variable on a separate line).

Issue #4586, opened 2022-01-18
port range: alter configuration


oliver-sanders:
The port range configuration configures a list of ports to be chosen from, however, the scheduler only passes the minimum and maximum values to the ZMQ server meaning the bit in the middle is not used.

Double check to ensure that only the min/max port numbers are used and consider changing the configuration  to a min/max pair so that it does what it says on the tin.

**Pull requests welcome!**

PR #4588, opened 2022-01-18
config: convert ports from list to min..max pair

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/4586
* Sibling https://github.com/cylc/cylc-sphinx-extensions/pull/66
* At Cylc 7 we could configure a list of ports to choose from.
* At Cylc 8 we use ZMQ to pick an available port for us which it does
  using min/max values.
* As a result we were only considering the first and last items in the
  configured list which could cause the port range to be smaller than
  configured.
* The config has now been reduced to a min..max pair in a format that is
  back compatible with the old format (providing that there is only one
  range).

**Requirements check-list**
- [x] No change log entry required (why? back compatible)

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -213,12 +213,6 @@ def mod_one(mod_one_conf, mod_flow, mod_scheduler):
     return schd
 
 
-@pytest.fixture(scope='session')
-def port_range():
-    ports = glbl_cfg().get(['scheduler', 'run hosts', 'ports'])
-    return min(ports), max(ports)
-
-
 @pytest.fixture(scope='module')
 def event_loop():
     """This fixture defines the event loop used for each test.
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -34,6 +34,7 @@ from cylc.flow.parsec.upgrade import upgrader
 from cylc.flow.parsec.validate import (
     CylcConfigValidator as VDR,
     DurationFloat,
+    Range,
     cylc_config_validate,
 )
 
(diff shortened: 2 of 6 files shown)


tests/integration/conftest.py
```diff
@@ -215,8 +215,7 @@ def mod_one(mod_one_conf, mod_flow, mod_scheduler):
 
 @pytest.fixture(scope='session')
 def port_range():
-    ports = glbl_cfg().get(['scheduler', 'run hosts', 'ports'])
-    return min(ports), max(ports)
+    return glbl_cfg().get(['scheduler', 'run hosts', 'ports'])
```
MetRonnie, line 218: Why not put this in `tests/conftest` to avoid duplication?

tests/unit/conftest.py
```diff
@@ -196,3 +197,8 @@ def xtrigger_mgr() -> XtriggerManager:
             create_autospec(Scheduler, workflow=workflow_name, owner=user)
         )
     )
+
+
+@pytest.fixture(scope='session')
+def port_range():
+    return glbl_cfg().get(['scheduler', 'run hosts', 'ports'])
```
MetRonnie, line 204: See above

MetRonnie (approved):

wxtim (approved):

MetRonnie (approved):

oliver-sanders:
That should fix the broken test, found a better way to change the __str__ representation so no longer need https://github.com/cylc/cylc-sphinx-extensions/pull/66

oliver-sanders:
One test failure:

```
 tests/f/intelligent-host-selection/05-from-platform-group.t 
```

Which is the result of the platform set issue?

PR #4590, opened 2022-01-19
Fix ordering of good hosts

wxtim:

### Issue description

The order of sets is not immutable. The selection method "definition order" relies on the objects storing lists of hosts and platforms storing order as well as content.

It's conceivably possible that removing the uniqueness property of set may give us a performance hit. I discuss alternatives in [this Element post](https://matrix.to/#/!hwZqSYihGPuhDdIzIP:matrix.org/$1642580156240880RFttk:matrix.org?via=matrix.org).

**Requirements check-list**
- [x] No change log entry included - pre-release change to an internal bug spotted by the dev team.

### DIFF
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -250,10 +250,10 @@ def get_platform_from_group(
     if bad_hosts:
         good_platforms = []
         for platform in group['platforms']:
-            if [
+            if any(
                 host for host in platform_from_name(platform)['hosts']
                 if host not in bad_hosts
-            ]:
+            ):
                 good_platforms.append(platform)
 
         platform_names = list(good_platforms)
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -20,7 +20,7 @@ import random
 import re
 from copy import deepcopy
 from typing import (
-    Any, Dict, Iterable, List, Optional, Tuple, Union, overload
+    Any, Dict, Iterable, List, Optional, Tuple, Set, Union, overload
 )
 
 from cylc.flow import LOG
@@ -67,7 +67,7 @@ def log_platform_event(
 def get_platform(
     task_conf: Union[str, None] = None,
     task_id: str = UNKNOWN_TASK,
-    bad_hosts: Optional[List[str]] = None
+    bad_hosts: Optional[Set[str]] = None
 ) -> Dict[str, Any]:
     ...
 
@@ -76,7 +76,7 @@ def get_platform(
 def get_platform(
     task_conf: Dict[str, Any],
     task_id: str = UNKNOWN_TASK,
-    bad_hosts: Optional[List[str]] = None
+    bad_hosts: Optional[Set[str]] = None
 ) -> Optional[Dict[str, Any]]:
     ...
 
@@ -92,7 +92,7 @@ def get_platform(
 def get_platform(
     task_conf: Union[str, Dict[str, Any], None] = None,
     task_id: str = UNKNOWN_TASK,
-    bad_hosts: Optional[List[str]] = None
+    bad_hosts: Optional[Set[str]] = None
 ) -> Optional[Dict[str, Any]]:
     """Get a platform.
 
@@ -151,7 +151,7 @@ def get_platform(
 def platform_from_name(
     platform_name: Optional[str] = None,
     platforms: Optional[Dict[str, Dict[str, Any]]] = None,
-    bad_hosts: Optional[List[str]] = None
+    bad_hosts: Optional[Set[str]] = None
 ) -> Dict[str, Any]:
     """
     Find out which job platform to use given a list of possible platforms and
@@ -228,7 +228,7 @@ def platform_from_name(
 def get_platform_from_group(
     group: Dict[str, Any],
     group_name: str,
-    bad_hosts: Optional[List[str]] = None
+    bad_hosts: Optional[Set[str]] = None
 ) -> str:
     """Get platform name from group, according to the selection method.
 
@@ -463,7 +463,7 @@ def generic_items_match(
 
 
 def get_host_from_platform(
-    platform: Dict[str, Any], bad_hosts: Optional[List[str]] = None
+    platform: Dict[str, Any], bad_hosts: Optional[Set[str]] = None
 ) -> str:
     """Placeholder for a more sophisticated function which returns a host
     given a platform dictionary.
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -251,8 +251,8 @@ def get_platform_from_group(
         good_platforms = []
         for platform in group['platforms']:
             if any(
-                host for host in platform_from_name(platform)['hosts']
-                if host not in bad_hosts
+                host not in bad_hosts
+                for host in platform_from_name(platform)['hosts']
             ):
                 good_platforms.append(platform)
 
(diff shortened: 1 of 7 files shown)


MetRonnie:
Assuming the list of hosts is not very long, I shouldn't imagine it will be much of a performance hit

cylc/flow/task_job_mgr.py
```diff
@@ -273,10 +273,10 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
             for itask in itasks:
                 # If there are any hosts left for this platform which we
                 # have not previously failed to contact with a 255 error.
-                if (
-                    set(itask.platform['hosts']) -
```
oliver-sanders, line 277: This `set()` was removing duplicates, not sure if that was important.

There is an existing interface in cylc.flow.network.resolvers for producing unique lists if it helps.

https://github.com/cylc/cylc-flow/blob/7b4e95c7dfeac69100530d06b686c9345067fd58/cylc/flow/network/resolvers.py#L96-L111

cylc/flow/platforms.py
```diff
@@ -67,7 +67,7 @@ def log_platform_event(
 def get_platform(
     task_conf: Union[str, None] = None,
     task_id: str = UNKNOWN_TASK,
-    bad_hosts: Optional[Set[str]] = None
+    bad_hosts: Optional[List[str]] = None
```
oliver-sanders, line 70: Do `bad_hosts` need to be ordered too or just `good_hosts`?

wxtim: No, but it makes more sense to me if hosts and badhosts are the same type.

cylc/flow/task_job_mgr.py
wxtim: Probably not hugely, but here./..

```suggestion
                     list(set(itask.platform['hosts'])) - 
```

cylc/flow/task_job_mgr.py
```diff
@@ -273,10 +273,10 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
...
-                    self.task_remote_mgr.bad_hosts
-                ):
+                if [
+                    host for host in itask.platform['hosts']
```
wxtim, line 277:
```suggestion
                    host for host in list(set(itask.platform['hosts']))
```

cylc/flow/task_job_mgr.py
wxtim: I don't think so, because this is in essence user error, but I've put in a `list(set(...))` just to be safe.

cylc/flow/platforms.py
```diff
@@ -248,10 +248,13 @@ def get_platform_from_group(
...
     if bad_hosts:
-        good_platforms = set()
+        good_platforms = []
         for platform in group['platforms']:
-            if set(platform_from_name(platform)['hosts']) - bad_hosts:
-                good_platforms.add(platform)
+            if [
+                host for host in platform_from_name(platform)['hosts']
+                if host not in bad_hosts
+            ]:
```
MetRonnie, line 256: Can do
```suggestion
            if set(platform_from_name(platform)['hosts']).difference(bad_hosts):
```
> Note, the non-operator versions of union(), intersection(), difference(), and symmetric_difference(), issubset(), and issuperset() methods will accept any iterable as an argument

https://docs.python.org/3.7/library/stdtypes.html#set

Or
```suggestion
            if any(
                host for host in platform_from_name(platform)['hosts']
                if host not in bad_hosts
            ):
```

cylc/flow/task_job_mgr.py
```diff
@@ -273,10 +273,10 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
...
                 # If there are any hosts left for this platform which we
                 # have not previously failed to contact with a 255 error.
-                if (
-                    set(itask.platform['hosts']) -
-                    self.task_remote_mgr.bad_hosts
-                ):
+                if [
+                    host for host in list(set(itask.platform['hosts']))
+                    if host not in self.task_remote_mgr.bad_hosts
+                ]:
```
MetRonnie, line 279: As above

MetRonnie (commented):
Having looked through the existing code, I can't find anywhere where the order of `bad_hosts` actually matters?

cylc/flow/platforms.py
wxtim: Can't use the first solution because conversion to set will remove any ordering.

cylc/flow/task_job_mgr.py
wxtim: Ay, one more instance and it becomes a helper function.

wxtim:
@MetRonnie & @oliver-sanders ; I've 
- Put `bad_hosts` and `bad_hosts_to_clear` back to being sets.
- Used suggestion for logic of the if statements.

cylc/flow/platforms.py
MetRonnie: But surely the order doesn't matter there - all that's doing is checking if there are any hosts that are are not bad?

MetRonnie: Basically, as far as I can see, the only diff that's needed is
```diff
-        good_platforms = set()
+        good_platforms = []
         for platform in group['platforms']:
             if set(platform_from_name(platform)['hosts']) - bad_hosts:
-                good_platforms.add(platform)
+                good_platforms.append(platform)
```

cylc/flow/scheduler.py
```diff
@@ -269,7 +269,7 @@ def __init__(self, reg: str, options: Values) -> None:
         self._profile_amounts = {}
         self._profile_update_times = {}
         self.pre_submit_tasks = []
-        self.bad_hosts: Set[str] = set()
+        self.bad_hosts: Optional[List[str]] = set()
```
oliver-sanders, line 272: Revert?

oliver-sanders (commented):
One comment, otherwise fine.

cylc/flow/scheduler.py
wxtim:
```suggestion
         self.bad_hosts: Set[str] = set() 
```

cylc/flow/platforms.py
```diff
@@ -248,10 +248,13 @@ def get_platform_from_group(
...
     """
     if bad_hosts:
-        good_platforms = set()
+        good_platforms = []
         for platform in group['platforms']:
-            if set(platform_from_name(platform)['hosts']) - bad_hosts:
-                good_platforms.add(platform)
+            if any(
+                host for host in platform_from_name(platform)['hosts']
+                if host not in bad_hosts
```
MetRonnie, line 255: When I think about it, the use of `any()` may sometimes be advantageous in terms of performance due to short-circuiting.

One suggestion to simplify
```suggestion
                host not in bad_hosts
                for host in platform_from_name(platform)['hosts']
```

cylc/flow/task_job_mgr.py
```diff
@@ -273,9 +273,9 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
             for itask in itasks:
                 # If there are any hosts left for this platform which we
                 # have not previously failed to contact with a 255 error.
-                if (
-                    set(itask.platform['hosts']) -
-                    self.task_remote_mgr.bad_hosts
+                if any(
+                    host for host in itask.platform['hosts']
+                    if host not in self.task_remote_mgr.bad_hosts
```
MetRonnie, line 278:
```suggestion
                    host not in self.task_remote_mgr.bad_hosts
                    for host in itask.platform['hosts']
```

oliver-sanders (approved):
Looks good.

MetRonnie (approved):

PR #4592, opened 2022-01-19
validate: move seqence bounds validation into the config

oliver-sanders:
* `cylc play` should do more or less everything `cylc validate` does.
* The sequence bounds check was only being run for `cylc validate`.
* This moves the logic into `cylc.flow.config` so it is run for both
  `cylc validate` and `cylc play`.
* Already tested by tests/functional/validate/71-task-proxy-sequence-bounds-err.t

(not essential for rc1, but small enough to be low risk)

**Requirements check-list**

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -32,6 +32,7 @@ from fnmatch import fnmatchcase
 import os
 from pathlib import Path
 import re
+from textwrap import wrap
 import traceback
 from typing import (
     Any, Callable, Dict, List, Mapping, Optional, Set, TYPE_CHECKING, Tuple,
@@ -476,6 +477,7 @@ class WorkflowConfig:
         self._check_explicit_cycling()
 
         self._check_implicit_tasks()
+        self._check_sequence_bounds()
         self.validate_namespace_names()
 
         # Check that external trigger messages are only used once (they have to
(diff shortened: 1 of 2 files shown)


wxtim (changes requested):
- [x] Read the code.
- [x] Played with code, added breakpoints. 
- [x] Manually tested the described behavior

The test you refer to (`f/validate/71-...`) may not be sufficient to test this change: It only tests whether this function is called & logs a warning for validate - If it's being used as a proxy unit test it's OK, but can we create a ticket to make it a Python unit test. If it's being used as a functional test shouldn't it check that Cylc Play actually uses the function?

wxtim:
> This PR moves a bit of validate logic into the config so that it is performed automatically every time the config is loaded rather than only for the `validate` command.

I'm confident that this is what is happening, but I had to manually test it to assure myself that it _was_ called by ``cylc play``, which I thought was the main point of the change.

> I'm not sure how a Python unit test would help here, it would check the expected warning is logged the same way the functional test would?

It'd do the same job faster and more read-ably.

oliver-sanders:
> I'd like more test

What aspect of this feature do you think is not covered by the pre-existing tests?

wxtim:
> > I'd like more test
> 
> What aspect of this feature do you think is not covered by the pre-existing tests?

I don't think that you have tested whether this warning will be logged if you run and command other than ``cylc validate``. From what you are saying I think that's probably OK.

hjoliver (approved):
Failing flaky test unrelated; passes locally.

I think this can go in as-is.

The function is unit testable, but we do have loads of functional validation tests, and they are fast for functional tests, so no biggie. (Plus, I guess we should test that these methods are used via the cli commands).

wxtim (approved):
I agree with @hjoliver  that this is ready to merge.

Issue #4129, opened 2021-03-17
ssh client: add async_request method


oliver-sanders:
Follow on from #4119

> **Note:** Low priority.

Add an `async_request` method to the SSH client to match the ZMQ client.

```python
async def async_request(self, ...)
    proc = remote_cylc_cmd(...)
    while True:
        if proc.poll():
            break
        await asyncio.sleep(self.SLEEP_INTERVAL)
    ...
```

This can be used in `cylc scan` via the `get_client` abstraction to support the SSH interface in this command.

**Pull requests welcome!**

datamel:
Unfortunately UID has caused this to be a breaking change which needs fixing before 8.0rc1. Updating the milestone.

oliver-sanders:
This has now become a bug, my bad.

PR #4595, opened 2022-01-19
add async request to ssh client

datamel:
Closes https://github.com/cylc/cylc-flow/issues/4129

**Requirements check-list**

### DIFF
--- a/cylc/flow/network/ssh_client.py
+++ b/cylc/flow/network/ssh_client.py
@@ -14,9 +14,10 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import asyncio
 import json
 import os
-from typing import Union
+from typing import Union, Dict
 
 from cylc.flow.exceptions import ClientError
 from cylc.flow.network.client_factory import CommsMeth
@@ -45,11 +46,34 @@ class WorkflowRuntimeClient():
             timeout: Union[float, str] = None
     ):
         self.workflow = workflow
-
+        self.SLEEP_INTERVAL = 0.1
         if not host:
             self.host, _, _ = get_location(workflow)
 
-    def send_request(self, command, args=None, timeout=None):
+    async def async_request(self, command, args=None, timeout=None):
+        """Send asynchronous request via SSH.
+        """
+        cmd, ssh_cmd, login_shell, cylc_path, message = self.prepare_command(
+            command, args, timeout)
+        proc = _remote_cylc_cmd(
+            cmd,
+            host=self.host,
+            stdin_str=message,
+            ssh_cmd=ssh_cmd,
+            remote_cylc_path=cylc_path,
+            ssh_login_shell=login_shell,
+            capture_process=True)
+        while True:
+            if proc.poll() is not None:
+                break
+            await asyncio.sleep(self.SLEEP_INTERVAL)
+            out, err = (f.decode() for f in proc.communicate())
+            return_code = proc.wait()
+            if return_code:
+                raise ClientError(err, f"return-code={return_code}")
+        return json.loads(out)
+
+    def serial_request(self, command, args=None, timeout=None):
         """Send a request, using ssh.
 
         Determines ssh_cmd, cylc_path and login_shell settings from the contact
@@ -69,6 +93,18 @@ class WorkflowRuntimeClient():
         Returns:
             object: Deserialized output from function called.
         """
+        loop = asyncio.new_event_loop()
+        task = loop.create_task(
+            self.async_request(command, args, timeout))
+        loop.run_until_complete(task)
+        loop.close()
+        return task.result()
+
+    def prepare_command(
+        self, command: str, args: Dict, timeout: Union[float, str]
+    ):
+        """Prepare command for submission.
+        """
         # Set environment variable to determine the communication for use on
         # the scheduler
         os.environ["CLIENT_COMMS_METH"] = CommsMeth.SSH.value
@@ -84,19 +120,6 @@ class WorkflowRuntimeClient():
         if not args:
             args = {}
         message = json.dumps(args)
-        proc = _remote_cylc_cmd(
-            cmd,
-            host=self.host,
-            stdin_str=message,
-            ssh_cmd=ssh_cmd,
-            remote_cylc_path=cylc_path,
-            ssh_login_shell=login_shell,
-            capture_process=True)
+        return cmd, ssh_cmd, login_shell, cylc_path, message
 
-        out, err = (f.decode() for f in proc.communicate())
-        return_code = proc.wait()
-        if return_code:
-            raise ClientError(err, f"return-code={return_code}")
-        return json.loads(out)
-
-    __call__ = send_request
+    __call__ = serial_request
--- a/cylc/flow/network/ssh_client.py
+++ b/cylc/flow/network/ssh_client.py
@@ -71,10 +71,10 @@ class WorkflowRuntimeClient():
                     if proc.poll() is not None:
                         break
                     await asyncio.sleep(self.SLEEP_INTERVAL)
-                    out, err = (f.decode() for f in proc.communicate())
-                    return_code = proc.wait()
-                    if return_code:
-                        raise ClientError(err, f"return-code={return_code}")
+                out, err = (f.decode() for f in proc.communicate())
+                return_code = proc.wait()
+                if return_code:
+                    raise ClientError(err, f"return-code={return_code}")
                 return json.loads(out)
         except asyncio.TimeoutError:
             raise ClientTimeout(
--- a/cylc/flow/network/ssh_client.py
+++ b/cylc/flow/network/ssh_client.py
@@ -50,13 +50,15 @@ class WorkflowRuntimeClient():
         self.SLEEP_INTERVAL = 0.1
         if not host:
             self.host, _, _ = get_location(workflow)
+        # 5 min default timeout
+        self.timeout = timeout if timeout is not None else 300
 
     async def async_request(self, command, args=None, timeout=None):
         """Send asynchronous request via SSH.
         """
-        self.timeout = timeout if timeout else 300  # 5 min default timeout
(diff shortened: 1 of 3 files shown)


cylc/flow/network/ssh_client.py
```diff
@@ -84,19 +130,6 @@ def send_request(self, command, args=None, timeout=None):
...
-            capture_process=True)
-
-        out, err = (f.decode() for f in proc.communicate())
-        return_code = proc.wait()
-        if return_code:
-            raise ClientError(err, f"return-code={return_code}")
-        return json.loads(out)
+        return cmd, ssh_cmd, login_shell, cylc_path, message
 
     __call__ = send_request
```
oliver-sanders, line 135:
```suggestion
    __call__ = serial_request
```

cylc/flow/network/ssh_client.py
```diff
@@ -69,6 +93,28 @@ def send_request(self, command, args=None, timeout=None):
...
+        cmd, ssh_cmd, login_shell, cylc_path, message = self.prepare_command(
+            command, args, timeout)
+        proc = _remote_cylc_cmd(
+            cmd,
+            host=self.host,
+            stdin_str=message,
+            ssh_cmd=ssh_cmd,
+            remote_cylc_path=cylc_path,
+            ssh_login_shell=login_shell,
+            capture_process=True)
```
oliver-sanders, line 105: In the ZMQ client we avoid duplication in a really lazy way by always calling the async interface:

```python
task = self.loop.create_task(                                                      
    self.async_request(command, args, timeout))          
self.loop.run_until_complete(task)          
return task.result()   
```

Might be a good idea to prevent these interfaces from diverging.

cylc/flow/network/ssh_client.py
```diff
@@ -45,10 +46,33 @@ def __init__(
...
+        while True:
+            if proc.poll() is not None:
+                break
+            await asyncio.sleep(self.SLEEP_INTERVAL)
```
oliver-sanders, line 69: Ooh, just spotted we hadn't implemented `timeout` for this client, would need to wrap with `async_timeout` or compare `time() - time_started`, can bump that though.

datamel:
The test, `tests/functional/cylc-message/00-ssh.t`, which is broken on master, has now been fixed with this change.

oliver-sanders (commented):
Looks good. Tests pass.

cylc/flow/network/ssh_client.py
datamel: So we have added the timeout in this client to the command (in `prepare_command`), via the comms_timeout cli option. Do you think that will need changing?

oliver-sanders: That passes the timeout through to the ZMQ client, but I don't think it protects us from network issues between the hosts on either end of the SSH command.

cylc/flow/network/ssh_client.py
datamel: Ok, I've had a go at implementing this. I've had to start new `asyncio.loop` as the standard client.py I think is using the loop from ZMQ which we don't have here. See what you think, shout if there is a better way that I haven't spotted!

cylc/flow/network/ssh_client.py
```diff
@@ -69,6 +93,18 @@ def send_request(self, command, args=None, timeout=None):
         Returns:
             object: Deserialized output from function called.
         """
+        loop = asyncio.new_event_loop()
+        task = loop.create_task(
+            self.async_request(command, args, timeout))
+        loop.run_until_complete(task)
+        loop.close()
```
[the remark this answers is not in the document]
oliver-sanders: Dammit asyncio is annoying, where's the "just do it" interface?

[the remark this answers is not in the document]
oliver-sanders: I think you can skip the creat_task stage and `loop.run_until_complete(the_thing_you_want_to_run())` but I think this is fine.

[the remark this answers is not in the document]
datamel: > Dammit asyncio is annoying, where's the "just do it" interface?

I think this is it...
```
task = asyncio.run(
            self.async_request(command, args, timeout))
```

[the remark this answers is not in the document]
oliver-sanders: Except that it can raise `RuntimeError`.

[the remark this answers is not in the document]
datamel: Perhaps a solution in the style of: https://stackoverflow.com/a/61331974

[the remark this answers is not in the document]
datamel: Or stick with the original that opens a new loop and closes it manually.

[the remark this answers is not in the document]
oliver-sanders: > Or stick with the original 

I think that's fine.

We have an existing issue with the RuntimeError problem in the functional tests, may need a solution like the one you linked to.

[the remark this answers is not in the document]
oliver-sanders: FYI: The way I've done async/sync interfaces elsewhere is:

```
def async_whatever(...):
    ...

def whatever(*args, **kwargs):
    # This will work, but only if your code is not called by async code.
    # This assumption can be broken in the unit/integration tests.
    asyncio.run(async_whatever(*args, **kwargs))
```

cylc/flow/network/ssh_client.py
```diff
@@ -31,7 +32,7 @@ class WorkflowRuntimeClient():
 
     Determines host from the contact file unless provided.
 
-    Args:
+    Arg:l
```
oliver-sanders, line 35:
```suggestion
    Args:
```

cylc/flow/network/ssh_client.py
```diff
@@ -14,12 +14,13 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from async_timeout import timeout as ascyncto
```
oliver-sanders, line 17: This library is currently in our test dependencies, we would need to bump it into the regular dependencies in `setup.cfg`

oliver-sanders:
(running remote tests ...)

cylc/flow/network/ssh_client.py
```diff
@@ -45,11 +47,43 @@ def __init__(
...
-
+        self.SLEEP_INTERVAL = 0.1
         if not host:
             self.host, _, _ = get_location(workflow)
 
-    def send_request(self, command, args=None, timeout=None):
+    async def async_request(self, command, args=None, timeout=None):
+        """Send asynchronous request via SSH.
+        """
+        self.timeout = timeout if timeout else 300  # 5 min default timeout
```
MetRonnie, line 57: I think it would make more sense to move this to `self.__init__()` and instead do the same thing as the ZMQ client's `async_request()`
https://github.com/cylc/cylc-flow/blob/7377b3cffe57ac01102cbc77e0ae669266bd95cf/cylc/flow/network/client.py#L168

cylc/flow/network/ssh_client.py
```diff
@@ -45,11 +47,43 @@ def __init__(
...
+                    ssh_login_shell=login_sh,
+                    capture_process=True)
+                while True:
+                    if proc.poll() is not None:
+                        break
+                    await asyncio.sleep(self.SLEEP_INTERVAL)
+                    out, err = (f.decode() for f in proc.communicate())
+                    return_code = proc.wait()
+                    if return_code:
+                        raise ClientError(err, f"return-code={return_code}")
```
MetRonnie, line 77: Both `Popen.communicate()` and `Popen.wait()` wait for the subprocess to finish, so that makes the while loop unnecessary. I'm not convinced this is a genuine async method (but does that matter?)

oliver-sanders: I think this was an indentation error: d82212ab3832c9330ec44ead007f80e15cc30007

cylc/flow/network/ssh_client.py
oliver-sanders: (note `async_timeout` and the `--comms-timeout` interfaces take time in seconds not miliseconds)

oliver-sanders:
I can't get the fast tests to pass at the moment on GHA at the moment, looks like the DNS issues prevail?

Suggest `pytest`ing locally, tests pass for me.

cylc/flow/network/ssh_client.py
```diff
@@ -45,11 +47,45 @@ def __init__(
...
+                    cmd,
+                    host=self.host,
+                    stdin_str=msg,
+                    ssh_cmd=ssh_cmd,
+                    remote_cylc_path=cylc_path,
+                    ssh_login_shell=login_sh,
+                    capture_process=True)
+                while True:
+                    if proc.poll() is not None:
+                        break
```
MetRonnie, line 74:
```suggestion
                while proc.poll() is None:
```

oliver-sanders:
Punting further work to https://github.com/cylc/cylc-flow/issues/4611

MetRonnie (approved):
> tests pass for me.

But are there any tests for this?

From the issue #4129:

> @mhall: Unfortunately UID has caused this to be a breaking change which needs fixing before 8.0rc1.
> 
> @oliver-sanders: This has now become a bug, my bad.

I'm not clear what the bug is/why this is a bug

cylc/flow/network/ssh_client.py
oliver-sanders: True but not important.

oliver-sanders:
> I'm not clear what the bug is/why this is a bug

This became a bug because of the multi-workflow functionality added by the UID PR.

Before all client commands used `serial_request`, after selected client commands used `async_request` for convenience with the `async` interface (the commands are still run in series).

oliver-sanders:
(e.g. `cylc stop`, `cylc hold`, `cylc broadcast`)

oliver-sanders:
re-running functional tests on a local platform...

MetRonnie:
Ok, tested `tests/functional/cylc-message/00-ssh.t` on master and this branch - failed on master, passed on this branch

PR #4597, opened 2022-01-20
uid: make task match warning explicit

oliver-sanders:
* Make implicit globs explicit so that warning messages are more
  informative.
* E.G. `cylc hold foo` should produce the warning
  `no tasks matching foo/*`.

### DIFF
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1620,6 +1620,7 @@ class TaskPool:
                 tokens = Tokens(id_, relative=True)
             except ValueError:
                 LOG.warning(f'Invalid task ID: {id_}')
+                continue
             if (
                 tokens['cycle_sel']
                 or tokens['task_sel']
@@ -1680,6 +1681,7 @@ class TaskPool:
                 tokens = Tokens(id_, relative=True)
             except ValueError:
                 LOG.warning(f'Invalid task ID: {id_}')
+                continue
             point_str = tokens['cycle']
             if not tokens['task']:
                 # make task globs explicit to make warnings clearer
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1629,21 +1629,19 @@ class TaskPool:
                 LOG.warning(f"No active tasks matching: {id_}")
                 unmatched_tasks.append(id_)
                 continue
+            if not tokens['task']:
+                # make task globs explicit to make warnings clearer
+                tokens['task'] = '*'
             point_str = tokens['cycle']
-            name_str = tokens['task'] or '*'
+            name_str = tokens['task']
             if name_str not in self.config.taskdefs:
                 if self.config.find_taskdefs(name_str):
                     # It's a family name; was not matched by active tasks
                     LOG.warning(
-                        f"No active tasks in the family '{name_str}' "
-                        f"matching: {id_}"
+                        f"No active tasks matching: {tokens.relative_id}"
                     )
                 else:
                     LOG.warning(self.ERR_TMPL_NO_TASKID_MATCH.format(name_str))
-                unmatched_tasks.append(id_)
-                continue
-            if point_str is None:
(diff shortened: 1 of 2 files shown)


MetRonnie:
Just noticed something (not currently touched by this PR, but probably worth bunging in a fix) - shouldn't there be a `continue` after this warning?
https://github.com/cylc/cylc-flow/blob/1f6364d1a02187850e1da5a9c9c222f5992e066f/cylc/flow/task_pool.py#L1619-L1622

cylc/flow/task_pool.py
```diff
@@ -1677,25 +1675,33 @@ def match_taskdefs(
         """
         n_warnings = 0
         task_items: Dict[Tuple[str, 'PointBase'], 'TaskDef'] = {}
-        for item in items:
-            tokens = Tokens(item, relative=True)
+        for id_ in ids:
+            try:
+                tokens = Tokens(id_, relative=True)
+            except ValueError:
+                LOG.warning(f'Invalid task ID: {id_}')
```
MetRonnie, line 1682: As with https://github.com/cylc/cylc-flow/pull/4597#issuecomment-1017699862
```suggestion
                LOG.warning(f'Invalid task ID: {id_}')
                continue
```

cylc/flow/task_pool.py
```diff
@@ -1629,23 +1629,21 @@ def match_future_tasks(
                 LOG.warning(f"No active tasks matching: {id_}")
                 unmatched_tasks.append(id_)
                 continue
+            if not tokens['task']:
+                # make task globs explicit to make warnings clearer
+                tokens['task'] = '*'
```
MetRonnie, line 1634: I think if this is moved to above the previous `if` condition, then the `No active tasks in the family` message below would be fine

oliver-sanders: The "in family" bit is misnomer, for the reported example the task is unspecified i.e. "*", so reporting it this way is confusing. Clearer to say "no tasks match ...".

cylc/flow/task_pool.py
oliver-sanders: Will take a look

oliver-sanders: Yep, that's a glaring bug.

cylc/flow/task_pool.py
```diff
@@ -1629,23 +1630,21 @@ def match_future_tasks(
...
                     LOG.warning(
-                        f"No active tasks in the family '{name_str}' "
-                        f"matching: {id_}"
+                        f"No active tasks matching: {tokens.relative_id}"
                     )
                 else:
                     LOG.warning(self.ERR_TMPL_NO_TASKID_MATCH.format(name_str))
                 unmatched_tasks.append(id_)
                 continue
-            if point_str is None:
```
oliver-sanders, line 1645: This should have been removed with the UID branch as point_str cannot be None.

cylc/flow/task_pool.py
MetRonnie: But if this is moved above this bit:
https://github.com/cylc/cylc-flow/blob/1f6364d1a02187850e1da5a9c9c222f5992e066f/cylc/flow/task_pool.py#L1623-L1628

then the only way of reaching here:
https://github.com/cylc/cylc-flow/blob/1f6364d1a02187850e1da5a9c9c222f5992e066f/cylc/flow/task_pool.py#L1637-L1639

is as the comment says, if the name is a family name. So it wouldn't be a misnomer

oliver-sanders:
Ok, how about this?

MetRonnie:
Nice, it keeps the ability to distinguish family names that have no active tasks.

Just a couple of test failures due to error message wording
```
FAILED tests/integration/test_task_pool.py::test_hold_tasks[Family names hold active tasks only]
FAILED tests/integration/test_task_pool.py::test_hold_tasks[Specifying task state works for active tasks, not future tasks]
```

PR #4598, opened 2022-01-20
Put `@pytest.mark.asyncio` decorators back in for now

MetRonnie:
Unfortunately, now getting a conflict with pytest-tornasync after removing `@pytest.mark.asyncio` decorators - https://github.com/eukaryote/pytest-tornasync/issues/13. pytest-tornasync is a dependency of cylc-uiserver, not cylc-flow, so it passed on GH Actions but now flakily fails locally.

This might be fixed by https://github.com/pytest-dev/pytest-asyncio/pull/261 when it makes its way into a release

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.
- [x] Does not need tests
- [x] No change log entry required
- [x] No documentation update required.

PR #4599, opened 2022-01-20
tests: disable pytest_tornasync

oliver-sanders:
I don't like the way this plugin invades our tests, nicer to kick it out.


oliver-sanders:
(the usual test failure)

MetRonnie (approved):
Fixes the problem (#4598), tested locally

PR #4602, opened 2022-01-21
cli: update dead end commands

oliver-sanders:


wxtim (approved):
Seem fine, should we put links to the 7-to-8 docs in here?

If not, it's ready to merge IMO.

oliver-sanders:
I think these one liners should be enough to provide users with their next step.

We could link into the docs, however, the 728 summary page just says the exact same thing.

PR #4607, opened 2022-01-24
graph: make workflow_id a required argument

oliver-sanders:
Fixes a minor bug where `cylc graph` would result in traceback due to WORKFLOW_ID erroneously being marked as optional.


hjoliver (approved):
(Ignoring spurious test failures)

PR #4617, opened 2022-01-27
Don't release from runahead if paused.

### LINKED ISSUE #4278: Resuming after `play --pause` can make held tasks run
> If you start a simple workflow (`P1 = foo`) paused, then hold all active tasks, then resume; `foo.1` will apparently ignore the held state and will run
>
>```
>INFO - Processing 1 queued command(s)
>INFO - Command succeeded: hold(['foo.*'])
>INFO - Processing 1 queued command(s)
>INFO - RESUMING the workflow now
>INFO - Command succeeded: resume()
>INFO - Queue released: foo.1
>INFO - [foo.1] -submit-num=01, host=hostymchostface
>INFO - [foo.1] -triggered off []
>INFO - [foo.1] status=preparing (held): (internal)submitted at 2021-06-07  for job(01) flow(E)
>INFO - [foo.1] -health check settings: submission timeout=None, polling intervals=PT15M,...
>INFO - [foo.1] status=submitted (held): (received)started at 2021-06-07  for job(01) flow(E)
>INFO - [foo.1] -health check settings: execution timeout=None, polling intervals=PT15M,...
>INFO - [foo.1] status=running (held): (received)succeeded at 2021-06-07  for job(01) flow(E)
>```

Follow-up to #4238

hjoliver:
These changes close #4278

**Requirements check-list**

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1455,7 +1455,7 @@ class Scheduler:
             tinit = time()
 
             # Useful for debugging core scheduler issues:
-            # self.pool.log_task_pool(logging.CRITICAL)
+            self.pool.log_task_pool(logging.CRITICAL)
 
             if self.pool.do_reload:
                 # Re-initialise data model on reload
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1470,7 +1470,7 @@ class Scheduler:
 
             self.process_command_queue()
 
-            if self.pool.release_runahead_tasks():
+            if not self.is_paused and self.pool.release_runahead_tasks():
                 self.is_updated = True
                 self.reset_inactivity_timer()
 
--- a/tests/integration/test_task_pool.py
+++ b/tests/integration/test_task_pool.py
@@ -79,7 +79,7 @@ def assert_expected_log(
 
 
 @pytest.fixture(scope='module')
-async def mod_example_flow(
+async def mod_example_flow_paused_start(
     mod_flow: Callable, mod_scheduler: Callable, mod_run: Callable
 ) -> Scheduler:
     """Return a scheduler for interrogating its task pool.
@@ -92,6 +92,23 @@ async def mod_example_flow(
     async with mod_run(schd):
         pass
     return schd
+
+
+@pytest.fixture(scope='module')
+async def mod_example_flow(
+    mod_flow: Callable, mod_scheduler: Callable, mod_run: Callable
+) -> Scheduler:
+    """Return a scheduler for interrogating its task pool.
+
+    This is module-scoped so faster than example_flow, but should only be used
+    where the test does not mutate the state of the scheduler or task pool.
+    """
+    reg = mod_flow(EXAMPLE_FLOW_CFG)
+    schd: Scheduler = mod_scheduler(reg, paused_start=False)
+    async with mod_run(schd):
+        pass
+    return schd
+
 
 
 @pytest.fixture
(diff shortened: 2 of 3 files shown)


hjoliver:
If starting the #4278 example paused, task pool logging now shows only one task `//1/foo`:
- `    * 1/foo waiting(held)(runahead) job:00 flows:1 status=waiting runahead=True queued=False`

i.e. we before the preparing stage, so all good.

The initial task pool looks different because subsequent `foo`s have not even been spawned yet, but that's fine.

hjoliver:
Integration tests fixed and extended. 

This could break func tests, if any, that check the task pool at paused start-up.  Hopefully there are none, gotta go ... :wave:

hjoliver:
Damn it, one integration test is failing here, but I can't get it to fail locally.  That's me done for now though.

cylc/flow/scheduler.py
```diff
@@ -1455,7 +1455,7 @@ async def main_loop(self):
             tinit = time()
 
             # Useful for debugging core scheduler issues:
-            # self.pool.log_task_pool(logging.CRITICAL)
+            self.pool.log_task_pool(logging.CRITICAL)
```
MetRonnie, line 1458: I assume you'll want to re-comment-out this?
```suggestion
            # self.pool.log_task_pool(logging.CRITICAL)
```

tests/integration/test_task_pool.py
```diff
@@ -114,6 +131,82 @@ async def example_flow(
...
+    items: List[str],
+    expected_task_ids: List[str],
+    expected_bad_items: List[str],
+    expected_warnings: List[str],
+    mod_example_flow_paused_start: Scheduler,
+    caplog: pytest.LogCaptureFixture
+) -> None:
+    """Test TaskPool.filter_task_proxies().
+
+    The NOTE before EXAMPLE_FLOW_CFG above explains which tasks should be
```
MetRonnie, line 191: Does the note need updating? https://github.com/cylc/cylc-flow/blob/24147eaa4cda64013c1afb6753e735fd8c7f4b48/tests/integration/test_task_pool.py#L28-L29

```diff
+# NOTE: foo & bar have no parents so at start-up they are
+spawned out to the runahead limit (only when not paused at start-up).
```

MetRonnie (commented):
I also get `test_filter_task_proxies[Task state]` passing locally. I think the reason it might fail is because you've changed it so it is no longer paused at start, and the GHA runner must be quicker than our local machines, so the tasks get released and are no longer waiting when the `mod_run` fixture shuts down?
https://github.com/cylc/cylc-flow/blob/072acd3ec8c6254052933dd54a48da2bbd789464/tests/integration/utils/flow_tools.py#L78-L105

cylc/flow/scheduler.py
```diff
@@ -1470,7 +1470,7 @@ async def main_loop(self):
 
             self.process_command_queue()
 
-            if self.pool.release_runahead_tasks():
+            if not self.is_paused and self.pool.release_runahead_tasks():
```
oliver-sanders, line 1473: Hmm, this would appear to be the reverse of a previous commit:

https://github.com/cylc/cylc-flow/commit/22e7a202ff5ad3c8e6becdc66beeea2df2e3f3e3#diff-90711b005fe27b9f894e6cd09f08816d847dcf853518894c6b7e2139784cbcb8R1444

oliver-sanders: I think runahead tasks should still be released in "paused" mode, but that the "paused" state should stop them from running.

https://github.com/cylc/cylc-flow/pull/4436

oliver-sanders (changes requested):
This is a regression of https://github.com/cylc/cylc-flow/pull/4436, suggesting #4620 as an alternative.

cylc/flow/scheduler.py
hjoliver: Good spotting! Admittedly this atempted fix was done in a desperate rush.

MetRonnie:
Closing this, then (if needs be can reopen)

PR #4620, opened 2022-01-27
queues: fix interactions with the scheduler paused and task held states

oliver-sanders:
> [edit] pushed up a new commit with a new approach after discovering the problem went deeper than initially thought.
>
> There are now three issues each has its own steps to reproduce and integration test.
> The new integration tests run on master but you will need to overwite `tests/integration/conftest.py` and `sed -i 's/pre_submit_tasks/prep_submit_tasks/' cylc/flow/scheduler.py`.

* Closes:        
  * https://github.com/cylc/cylc-flow/issues/4278        
  * https://github.com/cylc/cylc-flow/issues/4627        
  * https://github.com/cylc/cylc-flow/issues/4628        
* Makes the following changes:        
  * Held tasks are no longer be released from queues.        
  * `pre_prep_tasks` (previously `pre_submit_tasks`) are now included        
    with active tasks for the computation of queue limits.        
  * Queues are no longer processed whilst the workflow is paused.        
* User's should now be able to safely hold/release tasks:        
  * When they are not in the pool (future tasks).        
  * When they are in the pool but not yet queued.        
  * When they are queued.        
  * When they are in `pre_prep_tasks` (previously `pre_submit_tasks`) which        
    is an intermediary state tasks pass through *after* they have been        
    released form the queue but *before* they are passed into the job        
    preparation pipeline (and acquired the preparing job status).        
* Tasks can also be held whilst they are preparing, submitted & running,        
  however, this will continue to have no effect (except on automatic        
  retries, note `cylc kill`).     

**Requirements check-list**

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -237,8 +237,8 @@ class Scheduler:
     curve_auth: Optional[ThreadAuthenticator] = None
     client_pub_key_dir: Optional[str] = None
 
-    # queue-released tasks still in prep
-    pre_submit_tasks: Optional[List[TaskProxy]] = None
+    # queue-released tasks awaiting job preparation
+    pre_prep_tasks: Optional[List[TaskProxy]] = None
 
     # profiling
     _profile_amounts: Optional[dict] = None
@@ -268,7 +268,7 @@ class Scheduler:
         # mutable defaults
         self._profile_amounts = {}
         self._profile_update_times = {}
-        self.pre_submit_tasks = []
+        self.pre_prep_tasks = []
         self.bad_hosts: Set[str] = set()
 
         self.restored_stop_task_id = None
@@ -1250,20 +1250,27 @@ class Scheduler:
 
         """
         # Forget tasks that are no longer preparing for job submission.
-        self.pre_submit_tasks = [
-            itask for itask in self.pre_submit_tasks if
+        self.pre_prep_tasks = [
+            itask for itask in self.pre_prep_tasks if
             itask.waiting_on_job_prep
         ]
 
-        # Add newly released tasks to those still preparing.
-        self.pre_submit_tasks += self.pool.release_queued_tasks()
-
         if (
-            self.pre_submit_tasks and
-            not self.is_paused and
-            self.stop_mode is None and
-            self.auto_restart_time is None
+            not self.is_paused
+            and self.stop_mode is None
+            and self.auto_restart_time is None
         ):
+            # Add newly released tasks to those still preparing.
+            self.pre_prep_tasks += self.pool.release_queued_tasks(
+                # the number of tasks waiting to go through the task
+                # submission pipeline
+                self.pre_prep_tasks
+            )
+
+            if not self.pre_prep_tasks:
+                # No tasks to submit.
+                return
+
             # Start the job submission process.
             self.is_updated = True
 
@@ -1272,7 +1279,7 @@ class Scheduler:
 
             for itask in self.task_job_mgr.submit_task_jobs(
                 self.workflow,
-                self.pre_submit_tasks,
+                self.pre_prep_tasks,
                 self.curve_auth,
                 self.client_pub_key_dir,
                 self.config.run_mode('simulation')
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -598,10 +598,9 @@ class Scheduler:
             LOG.info("Quiet mode on")
             LOG.setLevel(log_level)
 
-    async def start_scheduler(self):
+    async def run_scheduler(self):
         """Start the scheduler main loop."""
         try:
-            self._configure_contact()
             if self.is_restart:
                 self.restart_remote_init()
             self.run_event_handlers(self.EVENT_STARTUP, 'workflow starting')
@@ -641,15 +640,10 @@ class Scheduler:
         finally:
             self.profiler.stop()
 
-    async def run(self):
-        """Run the startup sequence.
+    async def start(self):
+        """Run the startup sequence but don't set the main loop running.
 
-        * initialise
-        * configure
-        * start_servers
-        * start_scheduler
-
-        Lightweight wrapper for convenience.
+        Lightweight wrapper for testing convenience.
 
         """
         try:
@@ -657,11 +651,19 @@ class Scheduler:
             await self.configure()
             await self.start_servers()
             await self.log_start()
+            self._configure_contact()
         except (KeyboardInterrupt, asyncio.CancelledError, Exception) as exc:
             await self.handle_exception(exc)
-        else:
-            # note start_scheduler handles its own shutdown logic
-            await self.start_scheduler()
+
+    async def run(self):
+        """Run the startup sequence and set the main loop running.
+
+        Lightweight wrapper for testing convenience.
+
+        """
+        await self.start()
+        # note run_scheduler handles its own shutdown logic
+        await self.run_scheduler()
 
     def _load_pool_from_tasks(self):
         """Load task pool with specified tasks, for a new run."""
--- a/cylc/flow/task_queues/independent.py
+++ b/cylc/flow/task_queues/independent.py
@@ -42,6 +42,7 @@ class LimitedTaskQueue:
         """Release tasks if below the active limit."""
         # The "active" argument counts active tasks by name.
         released: List[TaskProxy] = []
+        held: List[TaskProxy] = []
         n_active: int = 0
         for mem in self.members:
             n_active += active[mem]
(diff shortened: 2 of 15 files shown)


oliver-sanders:
Tested manually @ 124b2b2 on Mac OS, passed the following:

* Unit
* Integration
* _local_background_indep_tcp
* _remote_background_indep_tcp
* _remote_background_indep_poll

MetRonnie:
Haven't had a proper look at the code but I cannot reproduce the bug on this branch 👍

cylc/flow/scheduler.py
```diff
@@ -1239,6 +1241,25 @@ def run_event_handlers(self, event, reason=""):
             event, str(reason), self.workflow, self.uuid_str, self.owner,
             self.host, self.server.port))
 
+    def release_unheld_tasks(self):
```
MetRonnie, line 1244: I find this name quite confusing

tests/integration/test_scheduler.py
```diff
@@ -129,3 +129,65 @@ async def mock_shutdown(*a, **k):
     assert last_record.exc_text.startswith("Traceback (most recent call last)")
     assert ("During handling of the above exception, "
             "another exception occurred") not in last_record.exc_text
+
+
```
MetRonnie, line 133: I noticed this test is passing on master, perhaps it shouldn't be?
https://github.com/cylc/cylc-flow/blob/124b2b29e96f041a4f01b2f86d4e8c198ab2b437/tests/integration/test_scheduler.py#L76

Using `pytest -r A` to debug, I think the reason it is passing might be because the task is in runahead at scheduler start-up, but the bug was found by holding a non-runahead task

MetRonnie: Yes, it seems runahead tasks are not affected by the bug

cylc/flow/scheduler.py
oliver-sanders: Picked to follow `release_queued_tasks`, feel free to propose an alternative.

tests/integration/test_scheduler.py
MetRonnie: I think I've found a fix

oliver-sanders: Not sure what you mean. All tasks a runahead at startup. The tasks end up in `pre_submit_tasks` as a result of being released from runahead by `one.pool.release_runahead_tasks`?

oliver-sanders: Test fails for me on master:

```
>           assert one.pre_submit_tasks == []
E           AssertionError: assert [<TaskProxy '//1/one'>] == []
E             Left contains one more item: <TaskProxy '//1/one'>
E             Full diff:
E             - []
E             + [<TaskProxy '//1/one'>]
```

MetRonnie: Sorry if unclear, I'm talking about `test_resume_does_not_release_tasks`

oliver-sanders: Deleting this broken test, logic covered by the new one.

oliver-sanders: This test fails because it is testing for the wrong thing. The task still has the held status, the bug was because the task was being run despite the held status.

oliver-sanders:
FYI: I'm taking a look at preventing tasks being released from queues whilst the Scheduler is paused as I think with this proposed solution the held tasks are occupying queue slots (because they were held after being de-queued).

MetRonnie:
For what it's worth / comparison purposes, here is what I came up with: https://github.com/cylc/cylc-flow/compare/master...MetRonnie:pause-hold-resume-bug

oliver-sanders:
> For what it's worth / comparison purposes, here is what I came up with

Unfortunately that works because of faulty queueing logic (#4628), because `pre_submit_tasks` have been dequeued they should be occupying queue slots preventing other tasks from running. So if you were to hold all tasks, then release a few from the bottom of the queue, you would expect the workflow to hang with this logic because the first few held tasks would be dequeued and placed in `pre_submit_tasks` where they would get stuck (because they would never get into `tasks_to_submit`), but no new tasks can be dequeued until they have run, which they won't because they are held.

oliver-sanders:
Unfortunately this issue has revealed further problems:

* https://github.com/cylc/cylc-flow/issues/4627
* https://github.com/cylc/cylc-flow/issues/4628

I think they can be fixed here fairly simply by:

* Preventing `held` tasks from being dequeued.
  * This has to be implemented at the queue level so will have to be done for every queue implementation sadly.
* Preventing tasks from being dequeued whilst the scheduler is paused.
  * This should address #4627 and avoid held tasks getting into the queue whilst the scheduler is paused (for niceness).

tests/integration/test_scheduler.py
oliver-sanders: I have left this test in.

Because it is only checking that the `is_held` state is preserved it did not identify this bug, it needed a more functional-like test to pick up this problem.

oliver-sanders:
Ok, implemented the above with one additional modification to make `pre_prep_tasks` count towards the queue limits.

Re-wrote the description to match the new approach.

tests/integration/test_scheduler.py
MetRonnie: This diff makes the test fail on master but passes on this branch
```diff
diff --git a/tests/integration/test_scheduler.py b/tests/integration/test_scheduler.py
index 8b7393bf8..ae9c90631 100644
--- a/tests/integration/test_scheduler.py
+++ b/tests/integration/test_scheduler.py
@@ -21,6 +21,7 @@ from typing import Any, Callable
 
 from cylc.flow.exceptions import CylcError
 from cylc.flow.scheduler import Scheduler
+from cylc.flow.task_state import TASK_STATUS_WAITING
 
 Fixture = Any
 
@@ -78,6 +79,10 @@ async def test_resume_does_not_release_tasks(one: Scheduler, run: Callable):
     schd: Scheduler = one
     async with run(schd):
         assert schd.is_paused
+        # Partial main loop iteration:
+        schd.pool.release_runahead_tasks()
+        schd.release_queued_tasks()
+
         itasks = schd.pool.get_all_tasks()
         assert len(itasks) == 1
         itask = itasks[0]
@@ -85,8 +90,13 @@ async def test_resume_does_not_release_tasks(one: Scheduler, run: Callable):
 
         schd.command_hold('*')
         schd.resume_workflow()
+        # Partial main loop iteration:
+        schd.pool.release_runahead_tasks()
+        schd.release_queued_tasks()
+
         assert not schd.is_paused
         assert itask.state.is_held
+        assert itask.state(TASK_STATUS_WAITING)
 
 
 async def test_shutdown_CylcError_log(one: Scheduler, run: Callable):
```

MetRonnie (commented):
Looks good apart from the new integration test seems to be flaky on GH Actions

tests/integration/test_scheduler.py
oliver-sanders: This is covered by the new test. Happy to remove this one or leave it as is.

oliver-sanders: Removed.

cylc/flow/task_queues/independent.py
```diff
@@ -51,9 +52,14 @@ def release(self, active: Counter[str]) -> List[TaskProxy]:
...
-            n_active += 1
-            active.update({itask.tdef.name: 1})
+            if itask.state.is_held:
+                held.append(itask)
+            else:
+                released.append(itask)
+                n_active += 1
+                active.update({itask.tdef.name: 1})
+        for itask in held:
+            self.deque.appendleft(itask)
```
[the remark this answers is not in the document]
oliver-sanders: Applied this but it broke things.

oliver-sanders:
@MetRonnie Haven't managed to replicate the flakyness locally, have pushed up a commit which I think will help.

It adds a new integration fixture that allows us to start workflows without running the main loop which should remove an unnecessary moving part. Can't say if that was causing the issue though.

oliver-sanders:
Looks like it worked, re-running to be safe...

oliver-sanders:
Test failure in 4/4 totally unrelated - https://github.com/cylc/cylc-flow/issues/4633

MetRonnie (approved):

tests/integration/test_data_store_mgr.py
```diff
@@ -178,11 +176,19 @@ def test_delta_task_state(harness):
     assert TASK_STATUS_FAILED in set(collect_states(
         schd.data_store_mgr.updated, TASK_PROXIES))
 
+    # put things back the way we found them
+    for itask in schd.pool.get_all_tasks():
+        itask.state.reset(TASK_STATUS_WAITING)
+        schd.data_store_mgr.delta_task_state(itask)
+    await schd.update_data_structure()
```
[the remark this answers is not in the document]
oliver-sanders: This test needs a larger re-think - https://github.com/cylc/cylc-flow/issues/4175

wxtim (approved):
Partial review posted to show that I'm looking
- [x] Checking against three tickets marked as closed by change.
- [x] Check the code changes.
- [ ] Check the test changes.
- [ ] Check the source against the bullet points in the PR description
- [ ] Had a really good go at breaking this logic

edit 
meant to post this as a comment not an approval - have re-requested my review.

Issue #4619, opened 2022-01-27
Triggering runahead tasks causes workflow to finish early


MetRonnie:
To reproduce:
```
[scheduler]
    allow implicit tasks = True
[scheduling]
    cycling mode = integer
    runahead limit = P3
    [[graph]]
        P1 = foo
```
When starting this workflow the first 3 cycles spawn and `4/foo` is in the runahead state.

If you quickly trigger `4/foo` while it is in runhead (before the other tasks have finished running), it will start running and when all tasks finish, the workflow shuts down (permanently finishes) even though there is no final cycle point.

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

oliver-sanders:
I think I can see what's going on here.

hjoliver:
This is related to how we auto-spawn next-instances of parentless tasks on release from runahead.

PR #4621, opened 2022-01-27
task_pool: fix spawning when runahead tasks are triggered

oliver-sanders:
> **Draft:** Candidate fix for #4619, not sure what the consequences of this change are as yet

* Closes https://github.com/cylc/cylc-flow/issues/4619

**Requirements check-list**

### DIFF
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -240,6 +240,7 @@ class TaskPool:
 
         # At restart all tasks are runahead-limited but finished and manually
         # triggered tasks (incl. --start-task's) can be released immediately.
+        runahead_limit_point = self.compute_runahead()
         for itask in (
             itask
             for itask in self.get_tasks()
@@ -251,10 +252,9 @@ class TaskPool:
             )
             or itask.is_manual_submit
         ):
-            self.release_runahead_task(itask)
+            self.release_runahead_task(itask, runahead_limit_point)
             released = True
 
-        runahead_limit_point = self.compute_runahead()
         if not runahead_limit_point:
             return released
 
@@ -595,10 +595,6 @@ class TaskPool:
 
         if itask.tdef.sequential:
             # implicit prev-instance parent
-            return
-
-        if not itask.flow_nums:
-            # No reflow
             return
 
         if not runahead_limit_point:
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -556,9 +556,17 @@ class TaskPool:
             or all(x < self.config.start_point for x in parent_points)
             or itask.tdef.has_only_abs_triggers(next_point)
         ):
+            task = itask.tdef.name
+            cycle = str(next_point)
+
+            # Get submit number by flow_nums {flow_nums: submit_num, ...}
+            if self.workflow_db_mgr.pri_dao.select_submit_nums(task, cycle):
+                # Task already spawned in this flow.
+                return None
+
             taskid = Tokens(
-                cycle=str(next_point),
-                task=itask.tdef.name,
+                cycle=cycle,
+                task=task,
             ).relative_id
             next_task = (
                 self._get_hidden_task_by_id(taskid)
--- a/tests/functional/spawn-on-demand/15-runahead-trigger.t
+++ b/tests/functional/spawn-on-demand/15-runahead-trigger.t
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+# 
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Tests three different trigger cases which should all have the same result.
+# 
+# https://github.com/cylc/cylc-flow/issues/4644
+# https://github.com/cylc/cylc-flow/issues/4619
+
+. "$(dirname "$0")/test_header"
+set_test_number 3
+for cycle in 2 3 4; do
+    TEST_NAME="${TEST_NAME_BASE}-${cycle}"
+    install_workflow "${TEST_NAME}" "${TEST_NAME_BASE}"
+    run_ok "${TEST_NAME}" \
+        cylc play "${WORKFLOW_NAME}" \
+            -n \
+            -s "cycle=$cycle" \
+            --reference-test \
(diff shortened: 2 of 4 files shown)


cylc/flow/task_pool.py
```diff
@@ -597,10 +597,6 @@ def release_runahead_task(
             # implicit prev-instance parent
             return
 
-        if not itask.flow_nums:
-            # No reflow
-            return
```
oliver-sanders, line 602: @hjoliver is this ok?

hjoliver: You're on the right track, but this doesn't fix the problem.

hjoliver:
Superseded by #4640

oliver-sanders:
https://github.com/cylc/cylc-flow/pull/4640#pullrequestreview-870739423

MetRonnie:
I'm afraid this doesn't work
Can confirm it doesn't work for Oliver's modified case:
```ini
[scheduling]
    cycling mode = integer
    runahead limit = P1
    [[graph]]
        P1 = foo
[runtime]
    [[foo]]
        script = sleep 20
```
```shell
$ cylc play workflow
# 1/foo spawns and starts running
# 2/foo spawns in runahead
$ cylc trigger workflow//2/foo
# When 1/foo and 2/foo finish, workflow shuts down, even though no FCP
```

oliver-sanders:
Hmm, strange things going on here.

Here's the simplest working test:

```ini
[scheduling]                                                
    initial cycle point = 1                                 
    cycling mode = integer                                  
    runahead limit = P1                                     
    [[graph]]                                               
        P1 = x      
               
[runtime]      
    [[x]]      
        script = """    
            sleep 1    
            if ((CYLC_TASK_CYCLE_POINT == 1)); then    
                cylc trigger "${CYLC_WORKFLOW_ID}//3/x"    
            fi      
            sleep 1    
        """ 
```

On this branch example works providing the trigger point is >3 but breaks if it is <3, dammit.

oliver-sanders:
@MetRonnie have a try now, I think that should do it but waiting on tests to see if there are ill consequences.

MetRonnie:
Haven't had a go with your most recent change yet, but it seems to me a combination of this and #4640 works for both cases?
```diff
diff --git a/cylc/flow/task_pool.py b/cylc/flow/task_pool.py
index 6bed8fff5..b8622c28e 100644
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -538,7 +538,9 @@ class TaskPool:
             self.workflow_db_mgr.pri_dao.select_tasks_to_hold()
         )
 
-    def spawn_successor(self, itask: TaskProxy) -> Optional[TaskProxy]:
+    def spawn_successor_if_parentless(
+        self, itask: TaskProxy
+    ) -> Optional[TaskProxy]:
         """Spawn next-cycle instance of itask if parentless.
 
         This includes:
@@ -597,16 +599,13 @@ class TaskPool:
             # implicit prev-instance parent
             return
 
-        if not itask.flow_nums:
-            # No reflow
-            return
-
-        if not runahead_limit_point:
-            return
-
-        # Autospawn successor of itask if parentless.
-        n_task = self.spawn_successor(itask)
-        if n_task and n_task.point <= runahead_limit_point:
+        # Autospawn successor of itask if parentless and reflowing.
+        n_task = self.spawn_successor_if_parentless(itask)
+        if (
+            n_task and
+            runahead_limit_point and
+            n_task.point <= runahead_limit_point
+        ):
             self.release_runahead_task(n_task, runahead_limit_point)
 
     def remove(self, itask, reason=""):
```
However it looks like it means there is no such thing as a non-reflow trigger?

MetRonnie:
Had a go with 24b7d2c and it seems a better solution than the diff I posted above. Say you do
```shell
$ cylc play workflow
# 1/foo spawns and starts running; 2/foo spawns in runahead
$ cylc trigger workflow//5/foo
# 5/foo spawns and starts running; 6/foo spawns in runahead
# 1/foo finishes; 2/foo starts running and 3/foo spawns in runahead
# 5/foo finishes; 6/foo is still waiting
...
# 3/foo finishes; 4/foo starts running; 5/foo spawns (again) in runahead
...
# 5/foo finishes; 6/foo starts running...
```
wheras with my diff, `6/foo` would start running the first time `5/foo` finished, leading to 2 ongoing flows.

hjoliver:
> On this branch example works providing the trigger point is >3 but breaks if it is <3, dammit.

They're different problems: a) triggering a runahead-limited task that already exists - #4640; b) triggering a new task that causes a flow-merge problem - #4644.

hjoliver:
I had a go with https://github.com/cylc/cylc-flow/commit/24b7d2cde6faf63952bcfeea7cda5110da81a61c and it doesn't work. Maybe you tested with a) triggering the runahead task; and b) triggering a future task that finishes before the main flow catches up? I'd expect it to work in those cases, but not for triggering the very next task right after the runahead task.

hjoliver:
Re-closing!

MetRonnie:
> for triggering the very next task right after the runahead task.

Works for me?


https://user-images.githubusercontent.com/61982285/152327180-44fb58a0-d672-4d3c-8e6e-3120150ae463.mp4

oliver-sanders:
Works for me too.

MetRonnie:
Ok so combining #4640 and #4645 results in this:

demo3

The difference is that instead of `4/foo` spawning in runahead when `3/foo` is triggered, `4/foo` spawns in runahead when `2/foo` starts running (i.e. when the original flow catches up, I presume)

oliver-sanders:
Re Re Opening...

oliver-sanders:
(rebased and removed the old test)

MetRonnie:
Tested and it avoids running the triggered task (`4/x`) twice like in https://github.com/cylc/cylc-flow/pull/4651#issuecomment-1029210464. But if using `cylc trigger --reflow` should it run `4/x` twice? Because at the moment, it doesn't; `--reflow` doesn't seem to make any difference

oliver-sanders:
That DB check will need to be a bit more advanced, it will have to fetch all results for the cycle/task, then go through the flow numbers json'loading them, then test that against the previous instance flow numb.

MetRonnie:
Hmm, can't imagine that will be good for performance

hjoliver:
> Ok so combining https://github.com/cylc/cylc-flow/pull/4640 and https://github.com/cylc/cylc-flow/pull/4645 results in this:
> ...
> The difference is that instead of 4/foo spawning in runahead when 3/foo is triggered, 4/foo spawns in runahead when 2/foo starts running (i.e. when the original flow catches up, I presume)

That's correct. When `3/foo` is triggered it is not supposed to reflow so it does not spawn `4/foo`.  But when the main flow catches up, it merges so that `3/foo` IS now supposed to reflow, so we retroactively spawn subsequent tasks to make that happen.

So, can someone explain why this issue has been re-re-opened?

hjoliver:
> >  for triggering the very next task right after the runahead task.
> 
> Works for me?

OK, by *Had a go with https://github.com/cylc/cylc-flow/commit/24b7d2cde6faf63952bcfeea7cda5110da81a61c and it works for both cases.* I thought you meant literally just the changes in that commit.  If I do that, the scheduler shuts down prematurely.

To get the result in your screen recording above I need the line ~600 changes as well, as in @oliver-sanders current first commit on this branch: https://github.com/cylc/cylc-flow/pull/4621/commits/8bf1d4906fc8c2dadbd49634aa7b45e9e4ee3a33

However, that is still the wrong result.  When you triggered `3/foo` you did not ask for a reflow, so it should not spawn `4/foo` at that point.

PR #4622, opened 2022-01-27
trigger: don't warn if task not already in the pool

oliver-sanders:
* Small fix probably resulting from the UID change.
* A warning currently shows when you trigger a task not present in the pool.

**Requirements check-list**

### DIFF
--- a/cylc/flow/id_match.py
+++ b/cylc/flow/id_match.py
@@ -134,8 +134,7 @@ def filter_ids(
             id_tokens_map[id_] = Tokens(id_, relative=True)
         except ValueError:
             _not_matched.append(id_)
-            if warn:
-                LOG.warning(f'Invalid ID: {id_}')
+            LOG.warning(f'Invalid ID: {id_}')
 
     for id_, tokens in id_tokens_map.items():
         for lowest_token in reversed(IDTokens):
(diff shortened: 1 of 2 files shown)


cylc/flow/id_match.py
```diff
@@ -134,8 +134,7 @@ def filter_ids(
             id_tokens_map[id_] = Tokens(id_, relative=True)
         except ValueError:
             _not_matched.append(id_)
-            if warn:
-                LOG.warning(f'Invalid ID: {id_}')
+            LOG.warning(f'Invalid ID: {id_}')
```
oliver-sanders, line 137: I think we should always warn for invalid IDs.

MetRonnie:
> A warning currently shows when you trigger a task not present in the pool.

I'm not sure exactly how to reproduce this

oliver-sanders:
> I'm not sure exactly how to reproduce this

Trigger any task in a workflow that is not in the pool.

```
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    [[graph]]
        P1 = a[-P1] => a

[runtime]
    [[a]]
```

```console
$ cylc play <id>
$ cylc trigger <id>//10/a
$ cylc cat-log <id>
```

On master the log should contain a warning saying that no tasks matched `10/a`, on this branch the warning should be gone.

MetRonnie:
Got it

MetRonnie (approved):

PR #4625, opened 2022-01-27
tests/f: add a skip for non GNU/Linux platforms for a non-posix test

oliver-sanders:
Test fails with BSD `script`, I don't think the test's usage is POSIX.

I don't think we need to test this everywhere so adding a skip.

> **Note:** The Mac OS tests only run `CHUNK=1/4` so won't necessarily hit this, for now...

**Requirements check-list**


wxtim (approved):
Nothing the matter, one question.

Test fails look unrelated - have kicked the CI.

Issue #4633, opened 2022-01-31
tests/f/ fix flaky clock-trigger-inexact/00


oliver-sanders:
`tests/functional/clock-trigger-inexact/00-big-offset.t` will not pass on the 31st of the month so has been disabled.

Between February and September it may fail from the 29th to 31st of the month too.

Fix this test to resolve the datetime dependence!

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

MetRonnie:
Seems to be caused by https://github.com/metomi/isodatetime/issues/250

PR #4634, opened 2022-01-31
tests/f: skip flaky clock-trigger-inexact/00

oliver-sanders:
* See https://github.com/cylc/cylc-flow/issues/4633

Skip the test for now, fix it later.


MetRonnie (approved):
I was of the opinion, "if it ain't broke for 90% of the time, don't skip it". However, looks like the code coverage is unaffected so this must be covered by passing unit tests.

PR #4640, opened 2022-02-02
Fix manual triggering of runahead-limited parentless tasks.

### LINKED ISSUE #4619: Triggering runahead tasks causes workflow to finish early

hjoliver:
These changes close #4619 

~~Note: this branch is built on top of #4641~~ (now merged; rebased onto master)

The problem was, we were not auto-spawning the successor of a manually-triggered parentless task unless runahead limit had been recomputed, and we don't recompute the limit after manual triggering (the limit is irrelevant for that). We need to auto-spawn the successor task regardless, but only release it from runhead if the limit has been recomputed (and if the task is below below the limit, of course).
 

**Requirements check-list**

### DIFF
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -538,7 +538,9 @@ class TaskPool:
             self.workflow_db_mgr.pri_dao.select_tasks_to_hold()
         )
 
-    def spawn_successor(self, itask: TaskProxy) -> Optional[TaskProxy]:
+    def spawn_successor_if_parentless(
+            self, itask: TaskProxy
+    ) -> Optional[TaskProxy]:
         """Spawn next-cycle instance of itask if parentless.
 
         This includes:
@@ -597,17 +599,15 @@ class TaskPool:
             # implicit prev-instance parent
             return
 
-        if not itask.flow_nums:
-            # No reflow
-            return
-
-        if not runahead_limit_point:
-            return
-
-        # Autospawn successor of itask if parentless.
-        n_task = self.spawn_successor(itask)
-        if n_task and n_task.point <= runahead_limit_point:
-            self.release_runahead_task(n_task, runahead_limit_point)
+        # Autospawn successor of itask if parentless and reflowing.
+        if itask.flow_nums:
+            n_task = self.spawn_successor_if_parentless(itask)
+            if (
+                n_task and
+                runahead_limit_point and
+                n_task.point <= runahead_limit_point
+            ):
+                self.release_runahead_task(n_task, runahead_limit_point)
 
     def remove(self, itask, reason=""):
         """Remove a task from the pool (e.g. after a reload)."""
--- a/tests/functional/spawn-on-demand/13-trigger-runahead/flow.cylc
+++ b/tests/functional/spawn-on-demand/13-trigger-runahead/flow.cylc
@@ -1,0 +1,21 @@
+# This workflow should behave the same as if foo.1 did not trigger foo.2.
+
+[scheduling]
+   cycling mode = integer
+   initial cycle point = 1
+   final cycle point = 3
+   runahead limit = P1
+   [[graph]]
+      P1 = foo
+[runtime]
+   [[foo]]
+      script = """
+         cylc__job__wait_cylc_message_started
+         if ((CYLC_TASK_CYCLE_POINT == 1)); then
+            expected="foo, 1, running, not-held, not-queued, not-runahead
+foo, 2, waiting, not-held, not-queued, runahead"
+            diff <(cylc dump -t "${CYLC_WORKFLOW_NAME}") <(echo "$expected")
+            # Force trigger next instance while it is runahead limited.
+            cylc trigger $CYLC_WORKFLOW_NAME//2/foo
+         fi
+      """
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -97,6 +97,9 @@ access to information on configured platforms.
 
 ### Fixes
 
+[#4640](https://github.com/cylc/cylc-flow/pull/4640) - Fix manual triggering of
+runahead-limited parentless tasks.
+
 [#4566](https://github.com/cylc/cylc-flow/pull/4566) - Fix `cylc scan`
 invocation for remote scheduler host on a shared filesystem.
 
--- a/cylc/flow/scripts/dump.py
+++ b/cylc/flow/scripts/dump.py
@@ -267,7 +267,7 @@ def main(_, options: 'Values', workflow_id: str) -> None:
                                 item['name'],
                                 item['cyclePoint'],
                                 item['state']]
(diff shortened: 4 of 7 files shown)


cylc/flow/task_pool.py
```diff
@@ -597,17 +599,15 @@ def release_runahead_task(
...
-            self.release_runahead_task(n_task, runahead_limit_point)
+        # Autospawn successor of itask if parentless and reflowing.
+        if itask.flow_nums:
+            n_task = self.spawn_successor_if_parentless(itask)
+            if (
+                n_task and
+                runahead_limit_point and
+                n_task.point <= runahead_limit_point
+            ):
+                self.release_runahead_task(n_task, runahead_limit_point)
```
oliver-sanders, line 610: I think this is identical to #4621?

hjoliver: No, according to my tests 4621 did not actually fix the problem… but this does, so it must be different!

MetRonnie: The difference in the logic is that here it will still `spawn_successor_if_parentless()` regardless of whether `runahead_limit_point` is `None` or not

oliver-sanders: Ah gotcha, good, helps me understand.

MetRonnie (commented):
Tested out manually, fixes the bug.

Test failure is due to changes in https://github.com/cylc/cylc-flow/pull/4641#issuecomment-1027906245

tests/functional/spawn-on-demand/13-trigger-runahead/flow.cylc
```diff
@@ -0,0 +1,23 @@
...
+   initial cycle point = 1
+   final cycle point = 3
+   runahead limit = P1
+   [[graph]]
+      P1 = foo
+[runtime]
+   [[foo]]
+      script = """
+         cylc__job__wait_cylc_message_started
+         if $((CYLC_TASK_CYCLE_POINT == 1)); then
```
oliver-sanders, line 14: This syntax is wrong causing the body of this `if` statement to be ignored skipping the test logic:

```
bash: 1: command not found
```

```suggestion
         if ((CYLC_TASK_CYCLE_POINT == 1)); then
```

oliver-sanders (changes requested):
This does fix the example reported in #4619, however, fails for a slightly more nuanced example.

The test on this branch has three cycles with a runahead limit of 1:

```
1  # active
2  # trigger this one
3  # workflow should run onto this task
```

This example works fine on this branch (and on #4621).

If we take this example and trigger the task in cycle 3 instead we create a bigger headache for the Scheduler as it needs to step over cycles 2 & 3 to spawn cycle 4 as runahead *before* cycle 2 has even run.

```
1  # active
2  # runahead
3  # trigger this one
4  # workflow should run onto this task
```

This example does not work on this branch (Scheduler shuts down), however, does work on #4621.

Reopened the original with a modification of the test from this branch to cover the 4 cycle variant.

tests/functional/spawn-on-demand/13-trigger-runahead/flow.cylc
MetRonnie: I think this fixes it (plus shellcheck would say put vars in quotes)
```diff
diff --git a/tests/functional/spawn-on-demand/13-trigger-runahead/flow.cylc b/tests/functional/spawn-on-demand/13-trigger-runahead/flow.cylc
index 60e5708e8..8bfd625dd 100644
--- a/tests/functional/spawn-on-demand/13-trigger-runahead/flow.cylc
+++ b/tests/functional/spawn-on-demand/13-trigger-runahead/flow.cylc
@@ -11,13 +11,13 @@
    [[foo]]
       script = """
          cylc__job__wait_cylc_message_started
-         if $((CYLC_TASK_CYCLE_POINT == 1)); then
+         if [[ "$CYLC_TASK_CYCLE_POINT" == 1 ]]; then
 
-            expected = "foo, 1, running, not-held, not-queued, not-runahead
+            expected="foo, 1, running, not-held, not-queued, not-runahead
 foo, 2, waiting, not-held, not-queued, runahead"
-            diff <(cylc dump -t $CYLC_WORKFLOW_NAME) <(echo "$expected")
+            diff <(cylc dump -t "$CYLC_WORKFLOW_NAME") <(echo "$expected")
 
             # Force trigger next instance while it is runahead limited.
-            cylc trigger $CYLC_WORKFLOW_NAME//foo.2
+            cylc trigger "${CYLC_WORKFLOW_NAME}//2/foo"
          fi
       """
```

MetRonnie:
Can confirm it doesn't work for Oliver's modified case:
```ini
[scheduling]
    cycling mode = integer
    runahead limit = P1
    [[graph]]
        P1 = foo
[runtime]
    [[foo]]
        script = sleep 20
```
```shell
$ cylc play workflow
# 1/foo spawns and starts running
# 2/foo spawns in runahead
$ cylc trigger workflow//3/foo
# 3/foo spawns and runs
# When 2/foo and 3/foo have finished, workflow shuts down, even though no FCP
```

hjoliver:
> This test is broken because:

Oops, sorry I did actually have a un-pushed commit fixing the stupid bash syntax error.

hjoliver:
Fixed the new test, checked it passes here and fails on master (and for the right reasons).

hjoliver:
> Can confirm it doesn't work for Oliver's modified case:

Interestingly, it works fine if I trigger any foo with cycle point `>= 4` (or `2/foo`, on this branch), just not `3/foo`.

tests/functional/spawn-on-demand/13-trigger-runahead/flow.cylc
hjoliver: (Fixed, by mining un-pushed changes on the work computer :grimacing: )

hjoliver:
The new example is technically a different problem, so I'll put up a new PR for that ... 

This one is good to go.

MetRonnie (approved):

oliver-sanders:
I think this is good, doing some testing with the other PR merged on top...

oliver-sanders (approved):

hjoliver:
By the way, this branch is equivalent to the approach of #4621 (for triggering runahead tasks only) but this one might have a small efficiency advantage via not recomputing the runahead limit when it's not needed.

PR #4578, opened 2022-01-13
moved triggering from INFO to DEBUG

wxtim:
These changes partially address #3647 

Scope kept small because of the volume of test changes required as consequence of logging change.

Only asked for 1 review because I believe that the change is small in meaning, despite huge diff.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.
- [x] Already covered by existing tests.
- [x] No change log entry required (Small change to the logging).
- [x] No documentation update required.

PR #4641, opened 2022-02-02
Fix reference tests

hjoliver:
~~UID~~ #4578 follow-up - "reference tests" cannot fail!!  :man_facepalming: 

This fixes that, and several functional test failures that were hidden by this bug.

**Requirements check-list**

### DIFF
--- a/tests/functional/broadcast/00-simple/flow.cylc
+++ b/tests/functional/broadcast/00-simple/flow.cylc
@@ -92,9 +92,7 @@
            sed -i '/BASH_XTRACEFD/d' ${PREPLOG}.err
 
            diff -u "${CYLC_WORKFLOW_RUN_DIR}/expected-prep.out" ${PREPLOG}.out
-           # Let's not compare stderr here; it can be contaminated by debug
-           # output from parsec (repeated config items).
-           # diff -u "${CYLC_WORKFLOW_RUN_DIR}/expected-prep.err" ${PREPLOG}.err
+           diff -u "${CYLC_WORKFLOW_RUN_DIR}/expected-prep.err" ${PREPLOG}.err
         """
     [[ENS]]
     [[ENS1]]
--- a/tests/functional/broadcast/00-simple/flow.cylc
+++ b/tests/functional/broadcast/00-simple/flow.cylc
@@ -92,7 +92,9 @@
            sed -i '/BASH_XTRACEFD/d' ${PREPLOG}.err
 
            diff -u "${CYLC_WORKFLOW_RUN_DIR}/expected-prep.out" ${PREPLOG}.out
-           diff -u "${CYLC_WORKFLOW_RUN_DIR}/expected-prep.err" ${PREPLOG}.err
+           # Let's not compare stderr here; it can be contaminated by debug
+           # output from parsec (repeated config items).
+           # diff -u "${CYLC_WORKFLOW_RUN_DIR}/expected-prep.err" ${PREPLOG}.err
         """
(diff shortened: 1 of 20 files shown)


tests/functional/cylc-cat-log/03-bad-workflow.t
```diff
@@ -20,7 +20,7 @@
 set_test_number 4
 
 CYLC_RUN_DIR="$RUN_DIR"
-BAD_NAME="$(basename "$(mktemp -u "${CYLC_RUN_DIR}/XXXXXXXX")")"
+BAD_NAME="NONEXISTENTWORKFLOWNAME"
```
hjoliver, line 23: (Unrelated: this test was flaky - mktemp can generate strings starting with digits, resulting in "invalid workflow name" errors instead of the intended ones).

MetRonnie:
`tests/k/cylc-poll/16-execution-time-limit.t` also failing for me locally on this branch

tests/functional/broadcast/00-simple/flow.cylc
```diff
@@ -92,7 +92,9 @@
            sed -i '/BASH_XTRACEFD/d' ${PREPLOG}.err
 
            diff -u "${CYLC_WORKFLOW_RUN_DIR}/expected-prep.out" ${PREPLOG}.out
-           diff -u "${CYLC_WORKFLOW_RUN_DIR}/expected-prep.err" ${PREPLOG}.err
+           # Let's not compare stderr here; it can be contaminated by debug
+           # output from parsec (repeated config items).
+           # diff -u "${CYLC_WORKFLOW_RUN_DIR}/expected-prep.err" ${PREPLOG}.err
```
oliver-sanders, line 97: Remove the `--debug` and the problem goes away.

oliver-sanders:
> UID follow-up

Is this a UID issue?

> tests/k/cylc-poll/16-execution-time-limit.t also failing for me locally on this branch

The test workflow isn't being run in `--debug` mode so the triggering isn't being logged.

> Let's not compare stderr here; it can be contaminated by debug output from parsec (repeated config items)

There's another option, we could log the trigger lines at info level when in `--reftest` mode.

oliver-sanders (approved):
Pre-emptive approval with https://github.com/hjoliver/cylc-flow/pull/20 or similar.

hjoliver:
>  >  UID follow-up
> 
> Is this a UID issue?

Yes, the `_load_reflog` function was looking for triggering info matching the pre-UID task ID format. So the result was `ref-log == test-log` is `True` (via `[] == []`) for ALL reference tests.

tests/functional/broadcast/00-simple/flow.cylc
hjoliver: > Remove the --debug and the problem goes away.

And so does the triggering info, so the reference test fails.

I figured there should not be any problem with the stderr file if the stdout is 100% correct. (However, I didn't check the details).

hjoliver:
> > There's another option, we could log the trigger lines at info level when in --reftest mode.
> 
> hjoliver#20

Agreed that's a good option. Not sure why we didn't do that in the first place.

tests/functional/broadcast/00-simple/flow.cylc
hjoliver: (merged your log-level change, so this is moot point now anyway!)

hjoliver:
Merged the side PR. One flaky test failed, passes locally. No need for a second review on this.

PR #4645, opened 2022-02-03
Fix flow merge into no-flow task.

### LINKED ISSUE #4644: A triggered task can block a flow.
A task triggered manually without `--reflow` has no flow numbers and does not spawn children.

If a task from an ongoing flow catches up to such a task (while it is still running), the two will merge. The merged task will have the ongoing flow number(s) and will spawn children on any subsequent outputs, but:
- any already-completed outputs won't have spawned children
- this effectively blocks the flow from already-completed outputs, potentially causing premature shutdown if those are the only spawning outputs

This is pretty niche; I can't imagine why a user would want to force-trigger a task that was just about to be spawned by the main flow anyway. Still, it's a bug because the main flow would be expected to keep going regardless of a single triggered task out in front of it.

Example:
```ini
[scheduling]
    [[graph]]
        R1 = """  # OK to force-trigger bar while foo is still running
           foo => bar => baz
        """
[runtime]
    [[foo, bar, baz]]
        script = sleep 10
```
If I trigger `bar` while `foo` is still running, when `foo` succeeds it will try to spawn `bar` and find that it already exists, so the scheduler will merge `foo`'s flow number(s) into `bar`. So by the time `bar` finishes it belongs to the flow and will spawn `baz` as normal.

However, if `baz` triggers off of an earlier output of `bar` then under the same scenario the scheduler will shut down without running `baz` (because `baz` did not belong to a flow at the time it completed that output):

```ini
        R1 = """  # BUG: NOT OK to force-trigger bar while foo is still running
           foo => bar
           bar:start => baz
        """
```

hjoliver:
These changes close #4644 

- [x] parented tasks
- [x] parentless tassk

**Requirements check-list**

### DIFF
--- a/cylc/flow/scripts/trigger.py
+++ b/cylc/flow/scripts/trigger.py
@@ -17,7 +17,7 @@
 
 """cylc trigger [OPTIONS] ARGS
 
-Manually trigger tasks.
+Manually trigger workflow tasks.
 
 Examples:
   # trigger task foo in cycle 1234 in my_flow
@@ -29,10 +29,28 @@ Examples:
   # start a new "flow" by triggering 1234/foo
   $ cylc trigger --reflow my_flow//1234/foo
 
-Note: waiting tasks that are queue-limited will be queued if triggered, to
-submit as normal when released by the queue; queued tasks will submit
-immediately if triggered, even if that violates the queue limit (so you may
-need to trigger a queue-limited task twice to get it to submit immediately).
+Manually triggering a task queues it to run regardless of satisfaction of its
+prerequisites. (This refers to Cylc internal queues).
+
+Manually triggering a queued task causes it to submit immediately, regardless
+of the queue limit.
+
+Manually triggering an active task has no effect, because multiple concurrent
+instances of the same task job aren't allowed. Active tasks are in the
+submitted or running states.
+
+Manually triggering an incomplete task queues it to run again and continue its
+assigned flow. Incomplete tasks finished without completing expected outputs.
+
+Manually triggering an "active-waiting" task queues it to run immediately in
+its flow. Active-waiting tasks have satisfied task-prerequisites but are held
+back by queue limit, runahead limit, clock trigger, xtrigger, or task hold.
+They are already assigned to a flow - that of the parent tasks that satisfied
+their prerequisites.
+
+Other tasks (those outside of the n=0 graph window) do not yet belong to a
+flow. Manual triggering of these with --reflow will start a new flow; otherwise
+only the triggered task will run.
 
 """
 
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1259,11 +1259,11 @@ class TaskPool:
     def spawn_on_all_outputs(self, itask, completed_only=False):
         """Spawn on all (or all completed) task outputs.
 
-        completed_only=False:
+        If completed_only is False:
            Used in Cylc 7 Back Compat mode for pre-spawning waiting tasks. Do
            not set the associated prerequisites of spawned children satisfied.
 
-        completed_only=True:
+        If completed_only is True:
            Used to retroactively spawn on already-completed outputs when a flow
            merges into a force-triggered no-flow task. In this case, do set the
            associated prerequisites of spawned children to satisifed.
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -566,9 +566,19 @@ class TaskPool:
             next_task = (
                 self._get_hidden_task_by_id(taskid)
                 or self._get_main_task_by_id(taskid)
-                or self.spawn_task(
-                    itask.tdef.name, next_point, itask.flow_nums)
             )
+            if next_task is None:
+                # Does not exist yet, spawn it.
+                next_task = self.spawn_task(
+                    itask.tdef.name, next_point, itask.flow_nums)
+            else:
+                # Already exists
+                if not next_task.flow_nums and not next_task.state.is_runahead:
+                    # Does not belong to a flow (force-triggered). Now (merge)
+                    # it does, so spawn successor, and children if needed.
+                    self.spawn_successor(next_task)
+                    self.spawn_on_all_outputs(next_task, completed_only=True)
+                next_task.merge_flows(itask.flow_nums)
             if next_task:
                 self.add_to_pool(next_task)
                 return next_task
@@ -1152,7 +1162,11 @@ class TaskPool:
                 or self._get_main_task_by_id(c_taskid)
             )
             if c_task is not None:
-                # Child already spawned, update it.
+                # Child already exists, update it.
+                if not c_task.flow_nums:
+                    # Child does not belong to a flow (force-triggered). Now
+                    # (merging) it does, so spawn outputs completed so far.
+                    self.spawn_on_all_outputs(c_task, completed_only=True)
                 c_task.merge_flows(itask.flow_nums)
                 LOG.info(
                     f"[{c_task}] Merged in flow(s) "
@@ -1229,14 +1243,25 @@ class TaskPool:
             if itask.identity == self.stop_task_id:
                 self.stop_task_finished = True
 
-    def spawn_on_all_outputs(self, itask):
-        """Spawn on all of itask's outputs regardless of completion.
+    def spawn_on_all_outputs(self, itask, completed_only=False):
+        """Spawn on all (or all completed) task outputs.
 
-        Do not set the associated prerequisites of spawned children satisfied.
-        Used in Cylc 7 Back Compat mode for pre-spawning waiting tasks.
+        completed_only=False:
+           Used in Cylc 7 Back Compat mode for pre-spawning waiting tasks. Do
+           not set the associated prerequisites of spawned children satisfied.
+
+        completed_only=True:
+           Used to retroactively spawn on already-completed outputs when a flow
+           merges into a force-triggered no-flow task. In this case, do set the
+           associated prerequisites of spawned children to satisifed.
 
         """
-        for output in itask.state.outputs._by_message:
+        if completed_only:
+            outputs = itask.state.outputs.get_completed()
+        else:
+            outputs = itask.state.outputs._by_message
+
+        for output in outputs:
             try:
                 children = itask.graph_children[output]
             except KeyError:
@@ -1256,8 +1281,13 @@ class TaskPool:
                     continue
                 # Spawn child only if itask.flow_nums is not empty.
                 c_task = self.spawn_task(c_name, c_point, itask.flow_nums)
-                if c_task is not None:
-                    self.add_to_pool(c_task)
+                if completed_only:
+                    c_task.state.satisfy_me({
+                        (str(itask.point), itask.tdef.name, output)
+                    })
+                    self.data_store_mgr.delta_task_prerequisite(c_task)
+
+                self.add_to_pool(c_task)
 
     def can_spawn(self, name: str, point: 'PointBase') -> bool:
         """Return True if name.point is within various workflow limits."""
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -585,8 +585,9 @@ class TaskPool:
                 if retroactive_spawn:
                     # Did not belong to a flow (force-triggered) before merge.
                     # Now it does, so spawn successor, and children if needed.
-                    LOG.info(
-                        f"[{next_task}] retroactive spawning post flow merge."
+                    LOG.debug(
+                        f"[{next_task}] spawning children retroactively"
+                        " post flow merge."
                     )
                     self.spawn_successor_if_parentless(next_task)
                     self.spawn_on_all_outputs(next_task, completed_only=True)
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -573,12 +573,24 @@ class TaskPool:
                     itask.tdef.name, next_point, itask.flow_nums)
             else:
                 # Already exists
-                if not next_task.flow_nums and not next_task.state.is_runahead:
-                    # Does not belong to a flow (force-triggered). Now (merge)
-                    # it does, so spawn successor, and children if needed.
-                    self.spawn_successor(next_task)
-                    self.spawn_on_all_outputs(next_task, completed_only=True)
+                retroactive_spawn = (
+                    not next_task.flow_nums
+                    and not next_task.state.is_runahead
+                )
                 next_task.merge_flows(itask.flow_nums)
+                LOG.info(
+                    f"[{next_task}] merged in flow(s) "
+                    f"{','.join(str(f) for f in itask.flow_nums)}"
+                )
+                if retroactive_spawn:
+                    # Did not belong to a flow (force-triggered) before merge.
+                    # Now it does, so spawn successor, and children if needed.
+                    LOG.info(
+                        f"[{next_task}] retroactive spawning post flow merge."
+                    )
+                    self.spawn_successor_if_parentless(next_task)
+                    self.spawn_on_all_outputs(next_task, completed_only=True)
+
             if next_task:
                 self.add_to_pool(next_task)
                 return next_task
@@ -1169,7 +1181,7 @@ class TaskPool:
                     self.spawn_on_all_outputs(c_task, completed_only=True)
                 c_task.merge_flows(itask.flow_nums)
                 LOG.info(
-                    f"[{c_task}] Merged in flow(s) "
+                    f"[{c_task}] merged in flow(s) "
                     f"{','.join(str(f) for f in itask.flow_nums)}"
                 )
                 self.workflow_db_mgr.put_insert_task_states(
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -119,6 +119,9 @@ tasks are not already preparing for job submission.
 [#4640](https://github.com/cylc/cylc-flow/pull/4640) - Fix manual triggering of
 runahead-limited parentless tasks.
 
+[#4645](https://github.com/cylc/cylc-flow/pull/4645) - Fix behaviour when a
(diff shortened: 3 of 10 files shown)


hjoliver:
This wasn't difficult to fix once I understood the problem, so bumping up to 8.0rc1.

MetRonnie (approved):
(Note this is not currently compatible with #4640):
- unheld -> not-held in `cylc dump`
- `spawn_successor` -> `spawn_successor_if_parentless`

cylc/flow/task_pool.py
```diff
@@ -1244,8 +1269,13 @@ def spawn_on_all_outputs(self, itask):
                     continue
                 # Spawn child only if itask.flow_nums is not empty.
                 c_task = self.spawn_task(c_name, c_point, itask.flow_nums)
-                if c_task is not None:
-                    self.add_to_pool(c_task)
+                if completed_only:
+                    c_task.state.satisfy_me({
+                        (str(itask.point), itask.tdef.name, output)
+                    })
+                    self.data_store_mgr.delta_task_prerequisite(c_task)
```
oliver-sanders, line 1276: This branch isn't covered, need to manually test...

oliver-sanders:
I think this functionality is broken when merged with #4640, see https://github.com/cylc/cylc-flow/pull/4651

oliver-sanders:
I'm not getting the same results for flow.cylc/suite.rc, need to polish off the test and make it repeatable before confirming.

cylc/flow/task_pool.py
oliver-sanders: I've run the test from #4651 against a flow.cylc and a suite.rc variants and god the same result, happy.

oliver-sanders (approved):

oliver-sanders:
Fixes the reported bug, however, is missing a DB check which causes case (4) to fail in the #4651 tests.

cylc/flow/task_pool.py
hjoliver: Something wrong with code coverage here.  The functional test on this branch should go into this branch...

... brute force test: I put a `print` statement in there and ran the test, it does get executed. :thinking:

hjoliver:
Rebased and deconflicted, post merge of #4640

hjoliver:
> Fixes the reported bug, however, is missing a DB check which causes case (4) to fail in the #4651 tests.

Not needed - see https://github.com/cylc/cylc-flow/pull/4651#issuecomment-1029544132

hjoliver:
OK, with my linked explanation of the controversial `4/x` case, hopefully this is good to go as-is.

cylc/flow/task_pool.py
hjoliver: Any idea what could cause coverage to miss this?

cylc/flow/scripts/trigger.py
```diff
@@ -29,10 +29,25 @@
...
-need to trigger a queue-limited task twice to get it to submit immediately).
+Triggering a task queues it to run (Cylc internal queues) regardless of its
+prerequisites.
+
+Triggering a queued task causes it to submit immediately, regardless of queue
+limit.
+
+An "active task" is already submitted or running. Triggering an active task is
+dangerous: multiple instances of the same job could fight over the same Cylc
+log files at run time (and possibly over task-specific data files too).
```
MetRonnie, line 40: Doesn't triggering active tasks get ignored?
```
INFO - [1/foo submitted job:01 flows:1] => running
INFO - [1/foo running job:01 flows:1] health: execution timeout=None, polling intervals=PT15M,...
INFO - Processing 1 queued command(s)
WARNING - [1/foo running job:01 flows:1] ignoring trigger - already active
```

cylc/flow/task_pool.py
MetRonnie: I think the coverage is accurate in this case. I put a `LOG.critical("Here")` at L1273 and it didn't show up in the workflow log

oliver-sanders:
>> Fixes the reported bug, however, is missing a DB check which causes case (4) to fail in the https://github.com/cylc/cylc-flow/pull/4651 tests.
>
> Not needed - see https://github.com/cylc/cylc-flow/pull/4651#issuecomment-1029544132

Needed, see https://github.com/cylc/cylc-flow/pull/4651#issuecomment-1029891848

cylc/flow/task_pool.py
oliver-sanders: Doesn't this branch only get hit in Cylc 7 back-compat mode?

I tested back-compat to make sure it worked, but didn't run coverage or insert a marker to test whether it hit this branch.

oliver-sanders (changes requested):
This does not work across a restart:

```ini
[scheduling]                                                  
    initial cycle point = 1                                   
    final cycle point = 5                                     
    cycling mode = integer                                    
    runahead limit = P1                                       
    [[graph]]                                                 
        P1 = x                                                
                                                              
[runtime]                                                     
    [[x]]                                                     
        script = """                                          
            if (( CYLC_TASK_CYCLE_POINT == 1 )); then         
                cylc trigger "${CYLC_WORKFLOW_ID}//4/x"       
            elif (( CYLC_TASK_CYCLE_POINT == 3 )); then       
                cylc stop "${CYLC_WORKFLOW_ID}" --now --now    
                contact="${HOME}/cylc-run/${CYLC_WORKFLOW_ID}/.service/contact"    
                while [[ -f "${contact}" ]]; do    
                    sleep 0.1                                                        
                done                                          
                cylc play "${CYLC_WORKFLOW_ID}"               
            fi                                                
        """  
```

Log from the second run:

```
2022-02-04T12:20:38Z INFO - Scheduler: url=tcp://...:43024/ pid=50836
2022-02-04T12:20:38Z INFO - Workflow publisher: url=tcp://vld601.cmpd1.metoffice.gov.uk:43020
2022-02-04T12:20:38Z INFO - Run: (re)start=1 log=1
2022-02-04T12:20:38Z INFO - Cylc version: 8.0rc1.dev
2022-02-04T12:20:38Z INFO - Run mode: live
2022-02-04T12:20:38Z INFO - Initial point: 1
2022-02-04T12:20:38Z INFO - Final point: 5
2022-02-04T12:20:38Z INFO - [3/x running(runahead) job:01 flows:1] => running
2022-02-04T12:20:40Z INFO - [3/x running job:01 flows:1] (polled)started at 2022-02-04T12:20:35Z
2022-02-04T12:20:40Z INFO - [3/x running job:01 flows:1] health: execution timeout=None, polling intervals=PT15M,...
2022-02-04T12:20:40Z INFO - [4/x waiting(runahead) job:01 flows:1] (polled)succeeded at 2022-02-04T12:20:28Z
2022-02-04T12:20:40Z INFO - [4/x waiting(runahead) job:01 flows:1] => succeeded(runahead)
2022-02-04T12:20:41Z INFO - [3/x running job:01 flows:1] (received)succeeded at 2022-02-04T12:20:40Z
2022-02-04T12:20:41Z INFO - [3/x running job:01 flows:1] => succeeded
2022-02-04T12:20:41Z INFO - Workflow shutting down - AUTOMATIC
2022-02-04T12:20:41Z INFO - DONE
```

cylc/flow/task_pool.py
hjoliver: No, `completed_only` is True for retroactive spawning on completed outputs (when an ongoing flow merges with a one-off manually-triggered task (thus making it part of the ongoing flow) ... i.e. what this PR is about). In back-compat mode, `completed_only` is False and we spawn on all outputs ahead of time (before they are completed) to approximate Cylc 7 spawning of tasks before they are needed.

hjoliver: > I think the coverage is accurate in this case. I put a `LOG.critical("Here")` at L1273 and it didn't show up in the workflow log for `tests/functional/spawn-on-demand/14-trigger-flow-blocker.t`

Strange. I just tried the same, to double check, and it does show up in my workflow log. (I did have to disable the purge in the `reftest` shell function, to keep the log for inspection).

hjoliver: (Maybe the test is not actually working correctly in some environments, namely yours and GitHub actions ... I'll investigate later...)

cylc/flow/scripts/trigger.py
hjoliver: Oops, you're right. Had it in my head that we still allowed that, in case the user knows better than the scheduler.

hjoliver: Fixed, and mentioned triggering incomplete tasks too.

cylc/flow/scripts/trigger.py
```diff
@@ -29,10 +29,29 @@
...
+difference here.
+
+An "incomplete task" finished without completing expected outputs. These are
+retained by scheduler in expectation of retriggering to continue the same
+flow, so the --reflow option makes no difference here either.
+
+Other tasks, behind or ahead of current active (and active waiting) tasks in
+the graph, do not yet belong to a flow. If you trigger one of these, the
+scheduler can run just the triggered task (the default) independent of any
+flow, or it can start a new flow to contineu on from that task (--reflow).
```
MetRonnie, line 54:
```suggestion
flow, or it can start a new flow to continue on from that task (--reflow).
```

cylc/flow/scripts/trigger.py
```diff
@@ -29,10 +29,29 @@
...
-immediately if triggered, even if that violates the queue limit (so you may
-need to trigger a queue-limited task twice to get it to submit immediately).
+Triggering a task queues it to run (Cylc internal queues) regardless of its
+prerequisites.
+
+Triggering a queued task causes it to submit immediately, regardless of queue
+limit.
+
+An "active task" is already submitted or running. You re-trigger an active task
+because multiple concurrent instances of the same task job would conflict.
```
MetRonnie, line 39: I'm not sure what the second sentence means

hjoliver: Stupid typos, it's late here!  ... fixed.

cylc/flow/scripts/trigger.py
hjoliver: fixed.

hjoliver:
> This does not work across a restart:

That's a bug, but it's not in scope for this PR (and not really specific to reflow either).

  1.  submit no. does not increment until the `preparing` stage, so a) polling of a `waiting(runahead)` task at restart sees the previous job; and b) we should not be polling `waiting` tasks anyway 
  2. in the context of your example, this PR is about triggering `3/x` so that the main flow catches up with it while it is still running; not `4/x` which is a future task that the scheduler doesn't catch up with (in the sense of requiring a merge)

oliver-sanders:
> future task

If I subtract one from the trigger point I get the same result:

```ini
[scheduling]
    initial cycle point = 1
    final cycle point = 5
    cycling mode = integer
    runahead limit = P1
    [[graph]]
        P1 = x

[runtime]
    [[x]]
        script = """
            if (( CYLC_TASK_CYCLE_POINT == 1 )); then
                cylc trigger "${CYLC_WORKFLOW_ID}//3/x"
            elif (( CYLC_TASK_CYCLE_POINT == 2 )); then
                cylc stop "${CYLC_WORKFLOW_ID}" --now --now
                contact="${HOME}/cylc-run/${CYLC_WORKFLOW_ID}/.service/contact"
                while [[ -f "${contact}" ]]; do
                    sleep 0.1
                done
                cylc play "${CYLC_WORKFLOW_ID}"
            fi
        """
```

```
2022-02-04T15:30:57Z INFO - Initial point: 1
2022-02-04T15:30:57Z INFO - Final point: 5
2022-02-04T15:30:57Z INFO - [2/x running(runahead) job:01 flows:1] => running
2022-02-04T15:30:57Z WARNING - Task 3/x already spawned in {1}
2022-02-04T15:30:58Z INFO - [2/x running job:01 flows:1] (polled)started at 2022-02-04T15:30:51Z
2022-02-04T15:30:58Z INFO - [2/x running job:01 flows:1] health: execution timeout=None, polling intervals=PT15M,...
2022-02-04T15:30:59Z INFO - [2/x running job:01 flows:1] (received)succeeded at 2022-02-04T15:30:59Z
2022-02-04T15:30:59Z INFO - [2/x running job:01 flows:1] => succeeded
2022-02-04T15:31:00Z INFO - Workflow shutting down - AUTOMATIC
2022-02-04T15:31:00Z INFO - DONE
```

Whereas with this diff, the workflow runs on to cycle 5 correctly:

```diff
             elif (( CYLC_TASK_CYCLE_POINT == 2 )); then
+                return
                 cylc stop "${CYLC_WORKFLOW_ID}" --now --now
```

hjoliver:
> > That's a bug, but it's not in scope for this PR (and not really specific to reflow either).
> > ...
> >  future task
> 
> If I subtract one from the trigger point I get the same result:

So:
 -2. now your test does do something that this PR addresses (flow merge) :heavy_check_mark: 
 -1. but the reason it fails is still off-topic :x:  ... 

... see #4658  

Both variants run to completion with that change in.

hjoliver:
(Rebase again)

hjoliver:
(Rebased again to pick up unit test fixes).

cylc/flow/task_pool.py
```diff
@@ -565,9 +565,31 @@ def spawn_successor_if_parentless(
...
+                LOG.info(
+                    f"[{next_task}] merged in flow(s) "
+                    f"{','.join(str(f) for f in itask.flow_nums)}"
+                )
+                if retroactive_spawn:
+                    # Did not belong to a flow (force-triggered) before merge.
+                    # Now it does, so spawn successor, and children if needed.
+                    LOG.info(
+                        f"[{next_task}] retroactive spawning post flow merge."
+                    )
```
MetRonnie, line 589: Not sure how useful this log line is, it seems to suggest `next_task` itself is being spawned, but it was already spawned before this

oliver-sanders: I don't think users are likely to understand what this message means.

oliver-sanders:
> Full credit, you're hammering this issue so hard that we're flushing other bugs out of the system

So much so I think I got a new one https://github.com/cylc/cylc-flow/issues/4665

oliver-sanders:
I'm ok with this approach, due to the number of trigger-related bugs uncovered recently I'd like to test more whist we are getting other PRs in. Would be good to generalise the trigger tests to make sure we are covering all cases correctly.

cylc/flow/task_pool.py
hjoliver: `next_task` is spawning its children.

I think that is consistent with other log messages (need to check).  

I guess it is pretty cryptic!  Should be reworded and/or shifted to debug level. But the point is, there should be some explanation of why the scheduler suddenly spawns on task outputs that were completed some time ago, which is unusual behaviour.

hjoliver:
> > Full credit, you're hammering this issue so hard that we're flushing other bugs out of the system
> 
> So much so I think I got a new one #4665

Fixed!  (And not really spawn-on-demand related).

cylc/flow/task_pool.py
oliver-sanders: I think the "merged in flows" line covers that base.

Suggest demoting this to debug.

oliver-sanders:
One remaining issue with the coverage thing https://github.com/cylc/cylc-flow/pull/4645#discussion_r799365450

cylc/flow/task_pool.py
hjoliver: demoted

hjoliver:
Fast test fail needs #4684

hjoliver:
> One remaining issue with the coverage thing [#4645 (comment)](https://github.com/cylc/cylc-flow/pull/4645#discussion_r799365450)

The test was flaky, so sometimes the merge didn't occur. Should be good now.

cylc/flow/task_pool.py
hjoliver: Test fixed. It was flaky.

hjoliver:
Coverage 100% :tada:

cylc/flow/scripts/trigger.py
```diff
@@ -29,10 +29,28 @@
...
-need to trigger a queue-limited task twice to get it to submit immediately).
+Manually triggering a task queues it to run regardless of satisfaction of its
+prerequisites. (This refers to Cylc internal queues).
+
+Manually triggering a queued task causes it to submit immediately, regardless
+of the queue limit.
+
+Manually triggering an active task has no effect, because multiple concurrent
+instances of the same task job aren't allowed. Active tasks are in the
+submitted or running states.
```
MetRonnie, line 40: If the second sentence is an explanation of what you are referring to in the first, I think this would be clearer:
```suggestion
Manually triggering an active task (a task that is in the submitted or
running states) has no effect, because multiple concurrent instances of
the same task job aren't allowed.
```

cylc/flow/scripts/trigger.py
```diff
@@ -29,10 +29,28 @@
...
+
+Manually triggering an incomplete task queues it to run again and continue its
+assigned flow. Incomplete tasks finished without completing expected outputs.
```
MetRonnie, line 43:
```suggestion
Manually triggering an incomplete task (a task that finished without completing
expected outputs) queues it to run again and continue its assigned flow.
```

MetRonnie (approved):

cylc/flow/scripts/trigger.py
hjoliver: I deliberately changed it to this way around to get the important information first in one concise sentence (*Triggering an xxx task does this.*), followed by some additional information for those who need a reminder of what an "incomplete task" actually is, etc.

MetRonnie: Then perhaps
```suggestion
Manually triggering an incomplete task queues it to run again and continue its
assigned flow (incomplete tasks are those which finished without completing
expected outputs).
```

oliver-sanders: Fine either way.

oliver-sanders (approved):

Issue #3737, opened 2020-08-05
reflow and parentless or absolute-parented tasks


hjoliver:
Should "reflow" mean ongoing spawning of child-tasks only, or does it also apply to parentless or absolute-parented tasks whose next instances get auto-spawned as they drop out of the runahead pool?   IMO its the latter, otherwise we can't trigger a reflow that is identical to the original flow in all of its tasks.  However, maybe ongoing auto-spawning needs to be optional. E.g.:

```
R1 = start
P1 = start[^] => foo => bar & baz => qux
```

If I retrigger `foo.6` with `--reflow` it should cause `bar.6` and `baz.6` then `qux.6` to run, but it will also result in `foo.7` being spawned and so on - is that to be expected or not? (Note we can get restricted child-only reflow by triggering `foo.6` without reflow, and then `cylc spawn foo.6` to spawn the children of `foo.6`).

oliver-sanders:
If we had a workflow like this:

```ini
R1 = """
    # n = 1..3
    a<n> => b<n> => c<n> & d<n>
"""
```

And we did `cylc trigger --reflow <flow> b_2.1` then I think the reflow should only effect the downstream tasks like so:

![*-1](https://user-images.githubusercontent.com/16705946/109684721-75592300-7b78-11eb-898f-8b8084e01d15.png)

There is no relation between `b_2` and `b_3` so we wouldn't expect these parallel graphs to interact.

I don't think swapping parameters for cycles should make any difference in this case difference:

```ini
P1 = """
    a => b => c & d
"""
```

![*-0](https://user-images.githubusercontent.com/16705946/109685276-ff08f080-7b78-11eb-9b71-b0f0117abdbb.png)

hjoliver:
Agreed, but my question really only relates to triggering (in your example) `a.2` (say), which has no parents. Then, with `--reflow` should `a.3` (and so on) get spawned too?

We might need to both. E.g. default to strictly graph-children; but include auto-spawned tasks as well with some new option flag.

hjoliver:
Clear documentation required!!

oliver-sanders:
@hjoliver can this become a cylc-doc issue now?

hjoliver:
It's should be easy to add an option to prevent flow-on to subsequent cycles.  I'll assume we want that (I believe this issue was motivated by you and/or Dave thinking it shouldn't flow on) and remove the question. Lemme know quickly if you've changed your mind on that.

[...]

PR #4651, opened 2022-02-03
tests/f: add test to cover triggering bugs

oliver-sanders:
**Skip to conclusions - https://github.com/cylc/cylc-flow/pull/4651#issuecomment-1029278095**

Adds three tests designed to cover the cases reported in #4644 & #4619 and related discussions:

> **Note:** Including #4644 since it changes the results.

To avoid disputed test results I suggest we have a standard test to cover all cases and apply it to all branches. Here's my proposed test, it seems to give deterministic results, all three cases fail on master.

Results for these tests on the three candidate fix branches:

* hjoliver/fix-4644 (PR #4645)
  2) :x: 
  3) :heavy_check_mark: 
  4) :x:
* hjoliver/fix-4619 (PR #4640)
  2) :heavy_check_mark:
  3) :x:
  4) :x:
* hjoliver/fix-4644 merged with hjoliver/fix-4619
  2) :heavy_check_mark:
  3) :heavy_check_mark:
  4) :x:
  [evidence](https://github.com/cylc/cylc-flow/runs/5055804191?check_suite_focus=true)
* oliver-sanders/4619 (PR #4621)
   2) :heavy_check_mark:
   3) :heavy_check_mark:
   4) :heavy_check_mark:
   [evidence](https://github.com/cylc/cylc-flow/runs/5055938570?check_suite_focus=true)

**Requirements check-list**

### DIFF
--- a/tests/functional/spawn-on-demand/15-runahead-trigger.t
+++ b/tests/functional/spawn-on-demand/15-runahead-trigger.t
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+# 
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Tests three different trigger cases which should all have the same result.
+# 
+# https://github.com/cylc/cylc-flow/issues/4644
+# https://github.com/cylc/cylc-flow/issues/4619
+
+. "$(dirname "$0")/test_header"
+set_test_number 3
+for cycle in 2 3 4; do
+    TEST_NAME="${TEST_NAME_BASE}-${cycle}"
+    install_workflow "${TEST_NAME}" "${TEST_NAME_BASE}"
+    run_ok "${TEST_NAME}" \
+        cylc play "${WORKFLOW_NAME}" \
+            -n \
+            -s "cycle=$cycle" \
+            --reference-test \
+            --debug
+    purge
+done
+exit
--- a/tests/functional/spawn-on-demand/15-runahead-trigger/flow.cylc
+++ b/tests/functional/spawn-on-demand/15-runahead-trigger/flow.cylc
@@ -1,0 +1,19 @@
+#!Jinja2
+
+[scheduling]
+    initial cycle point = 1
+    final cycle point = 5
+    cycling mode = integer
+    runahead limit = P1
+    [[graph]]
+        P1 = x
+
+[runtime]
+    [[x]]
+        script = """
+            cylc__job__wait_cylc_message_started
+            if ((CYLC_TASK_CYCLE_POINT == 1)); then
+                cylc trigger "${CYLC_WORKFLOW_ID}//{{cycle}}/x"
+            fi
+            sleep 1
+        """
--- a/tests/functional/spawn-on-demand/15-runahead-trigger/reference.log
+++ b/tests/functional/spawn-on-demand/15-runahead-trigger/reference.log
@@ -1,0 +1,7 @@
+Initial point: 1
+Final point: 5
+1/x -triggered off []
+2/x -triggered off []
+3/x -triggered off []
+4/x -triggered off []
+5/x -triggered off []


oliver-sanders:
(obviously tests will fail on master)

MetRonnie:
> hjoliver/fix-4644 merged with hjoliver/fix-4619
> 2) ✔️
> 3) ❌
> 4) ❌

But I get

2) ✔️ 
3) ✔️ 
4) ❌ 

And the only reason no. 4 fails is because `4/x` spawns/runs twice instead of once, which is not a problem as far as I can tell?
```diff
--- reference
+++ this run
@@ -2,4 +2,5 @@
 2/x -triggered off []
 3/x -triggered off []
 4/x -triggered off []
+4/x -triggered off []
 5/x -triggered off []
```

demo4

My branch combining Hilary's 2 PRs: https://github.com/MetRonnie/cylc-flow/tree/combined

MetRonnie:
Although when I watch no. 4 running on your branch #4621 I see `4/foo` spawn/run twice, yet the ref test passes 🤔

oliver-sanders:
Yeah just figured it out, that was the reftest bug which was missing from my branch, updated the description and rebased the branch.

oliver-sanders:
> And I think the only reason no. 4 fails is because 4/x spawns/runs twice instead of once, which is not a problem as far as I can tell?

Yes that is a problem!

It's a reflow that the user did not ask for.

oliver-sanders:
> But I get

Looks like I messed up the merge, confirmed.

oliver-sanders:
I've added a DB check to my branch which has got test (4) passing - a8b3fbb85b24398853d546e46dc3c96d53019658.

This approach can be applied to any of the approaches.

oliver-sanders:
* Right, I think I have a good idea what's going on here now.
* I've approved both of Hillary's PRs but not merged.
* I would rather the missing DB check is added into #4645 before merge but leaving that to @hjoliver since that is probably more dangerous in a real-world context than the original issue.
* Will leave @hjoliver to presumably re-re-close #4621 in preference to his branch but this time with the confirmation that it **definitely does work**.

My brain's fried, signing off.

hjoliver:
> Will leave @hjoliver to presumably re-re-close https://github.com/cylc/cylc-flow/pull/4621 in preference to his branch but this time with the confirmation that it definitely does work.

Done, but note I commented there that it works for the wrong reasons! (in the trigger-future-task-without-reflow case).

hjoliver:
> >   And I think the only reason no. 4 fails is because 4/x spawns/runs twice instead of once, which is not a problem as far as I can tell?
> 
> Yes that is a problem, and the same problem as on my branch.
> 
> It's a reflow that the user did not ask for (but it only impacts the one task because the flow numbers have not changed).

I disagree with this. It's the test that is wrong in no. 4, and the DB check isn't necessary.

What the user asked for, and got, was:
- the main flow, which should encompass all tasks including `4/x`
- and additionally, a one-off forced run of `4/x`. This should result in 2 runs of `4/x` in total regardless of whether or not the manual triggering is done ahead of or behind the main flow.

In contrast the other two tests should not result in the triggered task running twice because:
- no. 2: manually triggering an existing main-flow task that is runahead-limited, means "run it now rather than waiting for the runahead limit to move on" (c.f. manually triggering an existing but queued task).
- no. 3: a manually triggered future task that is still running when the main flow catches up to it becomes part of the main flow when the main flow merges into it.

hjoliver:
> My brain's fried, signing off.

I know that feeling :grin:

oliver-sanders:
**tldr;**

> I disagree with this. It's the test that is wrong in no. 4, and the DB check isn't necessary.

* Not performing the lookup for orphan tasks is contradictory.
* If the user didn't want the task to belong to the main flow then they would have used a reflow.
* Runahead limits are like queues, we pre-spwan the first "runahead queued" task for internal implementation reasons, however, all future tasks are "runahead queued". The first "runahead queued" should not behave differently to the second "runahead queued" task or the third or so on.
* The test cases 2-4 are all exactly the same from the user perspective, they all ask the Scheduler to trigger a task that is ahead of the workflow execution. The user would not expect different behaviour and it would be hard to explain to them why the task runs twice because it requires understanding undocumented internal implementation details. This required understanding of the task pool is something SOD was meant to put an end to.
* I'm pretty sure that not performing the DB check creates a bug where the Scheduler will behave differently on restart (where it does perform a DB check).
* From a user/data perspective I cannot think of any reason we would ever want this behaviour, however, I think we do have use cases for the DB lookup approach.

**Rants/responses:**

> without --reflow it should just run that task, no reflow, but still not part of the main flow.

So omitting the `--reflow` argument should result in an implicit reflow which only impacts on task, this is highly confusing (do the DB & data store reflect this edge case?).

Moreover the user cannot override this behaviour by specifying a flow on the command line, the previous success of the task will be ignored either way.

> consistency with manual triggering behind the main flow (rather than in front of it)

The past and future clearly distinct things, you would expect a discontinuity across this barrier, however, you wouldn't expect a discontinuity between the near future and far future.

As you say the runahead limit should be thought of as a queue, so why should the first "runahead queued" task behave differently to all subsequent ones?

> if there was more than one flow in process, which should the triggered task belong to?

This is an icky consequence of reflow, however, how is it different if the task is first runahead limited task or the second? The way the runahead limit is implemented is an implementation detail, the fundamental ambiguity is the same in both cases.

One option is to always use the original flow, it will only get merged anyway, another option is to demand clarification from the user. But inheriting the flow from a task which isn't supposed to logically exist yet is pretty strange, and if we are doing this why do we only do it for the first runahead limited task, not the second, or third, ...

> consider triggering a far- (rather than near-) future task in this way - should the main flow just jump over that bit of the graph when it finally gets there??

Yes, the previous state should be loaded from the DB, from there it will make its way into the data store and be displayed in the `n=1` window correctly.

> However, if you think there is a use case

**Here's a parallel chain example common in Climate data pipelines:**

```ini
runahead limit = P10Y
[graph]
    P1Y = """
        get_data => do_things
    """
```

Use case 1:

I run the workflow, but then I decide I want to get some results for a specific year so trigger the appropriate `get_data` task. I do not want the processing to be repeated later on and would not expect that to happen since the cycle has already completed successfully.

Use case 2:

The `get_data` bit is slow so I want to speed things up by triggering some of these ahead of time.

**Here's a push-type workflow used for satellite data processing (as documented), CI and other listener patterns:**

```ini
external-trigger = get_data("new dataset ready for processing")
[graph]
    P1 = """
        get_data => do_things
    """
```

Cycles are kicked off like so:

```console
$ cylc external-trigger <id> <msg> <trig>
```

We cannot guarantee the order these external events come into the workflow. I'm not sure how `ext-trigger` logic plays out here but I would definitely not expect previously run cycles to show up again.

hjoliver:
Still disagree, but too much to respond to now, I'll have to wait till daylight hours!

oliver-sanders:
Amendments to the above:

* My concern is a broader problem with the no-flow trigger approach as the same is also true for non-orphan scenarios were I also think that the tasks should not be re-run, or not, based on runtime conditions.
* I think the DB concern I noted above is not a problem because the no-flow status should be consistent between the task pool and DB.

oliver-sanders:
I've opened an issue with a less ranty brain-fried summation to help us separate the "future task" stuff from the related issues: https://github.com/cylc/cylc-flow/issues/4657

hjoliver:
I'll try to distill the main topics here before going over to the new issue

1. **DB check**

> Not performing the lookup for orphan tasks is contradictory, we always do it for [non-orphan] tasks [etc.]

Parentless tasks (I assume that's what you're calling orphans) **do** already get DB checked when they're spawned.  All tasks, including these ones, get spawned via the `spawn_task()` method in the `task_pool` module, which does that.

The difference is, the existing DB check looks at flow numbers, whereas your suggested additional check (as per the commit linked to above) bails out if any instance of the task was spawned before, regardless of flow.

2. **Exactly what "reflow" means**

> So omitting the --reflow argument should result in an implicit reflow which only impacts on task, this is highly confusing (do the DB & data store reflect this edge case?).
 
As I've been think of and documenting it, if you manually trigger a task **anywhere in the graph** (outside of n=0) you can:
- Run that task alone
   - this is not a "reflow", because a single task does not make an ongoing "flow"
   - it needs no flow number to pass on, because it spawns no children
- Or start a new flow from that task
   - this is a "reflow", because it's an ongoing flow
   - the task gets a new flow number to pass on to children that continue the flow

In other words, there is the original flow, there are standalone manually triggered tasks, and there are manually triggered reflows - which are "reflows" in the sense that we are "reusing the workflow graph for another flow" - whether the new flow is triggered behind or in front of the original

You seem to be suggesting that "reflow" should mean any standalone task or re-covers parts of the graph already covered by the original flow?  Using the terminology in that way could be really confusing too. E.g. to my mind, a standalone triggered task is not a "flow" even if it's behind the original flow. And what if a new flow only partially overlaps a previous flow (or vice versa)?

Either way, we have to explain exactly what the terminology means, and users will have to understand what they're doing if they want to make use of this new capability.  
 
3. **Should parentless tasks spawn ahead if triggered with `--reflow`?**

There's an issue somewhere, or at least we discussed it somewhere. We really need both capabilities, but as yet only the spawning version is implemented.

If you manually trigger a parentless task with `--reflow`, you might want only its children to run. But equally, you might want it to spawn ahead to future cycles too.  Otherwise, if the original flow was stopped, how do you run to the whole graph to completion?

4. **Should `n=0` waiting tasks behave the same as `n>0`?**

I appreciate what you're saying here, but we need the "active task pool" to contain some waiting tasks (which I've been calling "active-waiting" tasks, and there are several reasons why triggering those is different from triggering entirely virtual future (or past) tasks. I'll go into that in the new issue...

Users will have to understand this, but it's vastly less confusing and problematic than the Cylc 7 task pool.

oliver-sanders:
Closing pending possible reincarnation with #4657

hjoliver:
>     Should parentless tasks spawn ahead if triggered with --reflow?
> 
> There's an issue somewhere, or at least we discussed it somewhere. We really need both capabilities, but as yet only the spawning version is implemented.
> 
> If you manually trigger a parentless task with --reflow, you might want only its children to run. But equally, you might want it to spawn ahead to future cycles too. Otherwise, if the original flow was stopped, how do you run to the whole graph to completion?

#3737

oliver-sanders:
Note these tests now all pass on https://github.com/cylc/cylc-flow/pull/4739

Issue #4555, opened 2021-12-15
Further cylc wrapper improvements?


hjoliver:
Should we get the wrapper to:
- automatically invoke cylc-7.9.5 if `cylc review` is called in a Cylc 8 environment?
- list available Cylc versions?

hjoliver:
Quick hack at available versions:
```bash
if [[ $1 == "list-versions" ]]; then
   ls -l "${CYLC_HOME_ROOT}" | awk '{print $9,$10,$11}'
   exit 0
fi
```
result:
```console
$ cylc list-versions
                                                        
7.9.1 
7.9.2 
7.9.5 
8.0b3 
cylc -> 7.9.1  
```

hjoliver:
Quick hack at `cylc review` invocation:
```bash
if [[ "${CYLC_VERSION:-x}" =~ ^8 ]]; then
   if [[ "${1:-}" == "review" ]]; then
      echo "WARNING: starting Cylc 7.9.5 'cylc review' for Cylc 8 logs."
      CYLC_VERSION=7.9.5  # latest Cylc 7 "cylc review" required for Cylc 8.
   fi
fi
```

hjoliver:
### site config locations

The wrapper could be a convenient place to set `CYLC_SITE_CONF_PATH` to a non-standard location (for instance, `/etc` is not normally (ever?) on the shared filesystem?).  List set up for the cylc admin, because users **only** need the wrapper in their executable search path.

However, I came across a problem with this at NIWA, getting the wrapper to work with both Cylc 8 and Cylc 7 site configs in the same directory hierarchy, because Cylc 7's `CYLC_CONF_PATH` is more explicit than Cylc 8's `CYLC_SITE_CONF_PATH`. Had to do this in the wrapper:
```bash
if [[ "${CYLC_ENV_NAME:-x}" =~ ^8 || "${CYLC_VERSION:-x}" =~ ^8 ]]; then  # NIWA
   # Cylc 8 site config
   export CYLC_SITE_CONF_PATH="${CYLC_SITE_CONF_PATH:-/opt/nesi/share/cylc/etc}"  # /flow/VERSION assumed
   unset CYLC_CONF_PATH  # and this interferes with Cylc 8 too
else
   # Cylc 7 site config
   export CYLC_CONF_PATH="${CYLC_CONF_PATH:-/opt/nesi/share/cylc/etc/flow/7}"  # /flow/VERSION required
fi
```

[...]

PR #4655, opened 2022-02-04
Wrapper script enhancements

dpmatthews:
Support the use of environments with names that do not match the cylc flow version (e.g. cylc-8.0.0-1)
Ensure workflows always run under fixed environments rather than symlinked versions which could change
Provide legacy support for rose edit & cylc review (partially address #4555)

### DIFF
--- a/cylc/flow/etc/cylc
+++ b/cylc/flow/etc/cylc
@@ -17,95 +17,71 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 #------------------------------------------------------------------------------
-# Wrapper script to support multiple installed Cylc & Rose versions. Handles
-# Conda and Python virtual environments, and legacy plain installations.
+# Wrapper script to support multiple installed Cylc, Rose & Isodatetime
+# versions. Handles Conda and Python virtual environments, and legacy plain
+# installations.
 #
 # WRAPPER INSTALLATION AND CONFIGURATION
-#---------------------------------------
+# --------------------------------------
 # - Copy this script as "cylc" into default $PATH, on scheduler and job hosts
-# - If using Rose, create a "rose" symlink (ln -s cylc rose)
+# - Create an "isodatetime" symlink (ln -s cylc isodatetime)
+# - If using Rose, create "rose" & "rosie" symlinks
 # - Set CYLC_HOME_ROOT ("EDIT ME" below) to default to the parent directory of
 #   installed versions and set the global config locations if necessary
 #
 # HOW IT WORKS
-#-------------
-# Intercept "cylc" and "rose" commands and re-invoke them with the version
-# selected. If $CYLC_HOME is defined it is used as the installation location.
-# Otherwise, the script looks for a directory named "cylc-$CYLC_VERSION"
-# (or just "cylc" if $CYLC_VERSION is not defined) in $CYLC_HOME_ROOT or
-# $CYLC_HOME_ROOT_ALT.
+# ------------
+# Intercept "cylc", "rose", "rosie" and "isodatetime" commands and re-invoke
+# them with the version selected via the environment variables described below.
 #
-# Additional legacy logic is used when calling rose with Cylc 7.
+# Additional legacy logic is used when calling Rose with Cylc 7.
 #
 # ENVIRONMENT VARIABLES
 # ---------------------
 #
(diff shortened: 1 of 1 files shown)


cylc/flow/etc/cylc
```diff
@@ -119,77 +95,89 @@ CYLC_HOME_ROOT="${CYLC_HOME_ROOT:-/opt}"
...
+                if [[ -n "${ROSE_VERSION:-}" ]]; then
+                    CYLC_HOME="${ROSE_HOME_ROOT}/rose-${ROSE_VERSION}"
+                else
+                    # Use default version (symlink)
+                    CYLC_HOME="${ROSE_HOME_ROOT}/rose"
+                fi
+            else
+                CYLC_HOME="${ROSE_HOME}"
+            fi
+        elif [[ ${1:-} == "edit" || ${1:-} == "config-edit" ]]; then
```
wxtim, line 151: I think it would be reasonable to add a warning - consider the following silly, but plausible scenario:

```bash
> export CYLC_VERSION=8.0b1; export ROSE_VERSION=2019.01.1;   rose version
  File "<string>", line 1, in <module>
AttributeError: module 'metomi.rose' has no attribute '__version__'
```

wxtim (commented):
I've found one scenario which produces an error which should probably be replaced by a warning.

Otherwise seems reasonable to read, and test manually.

cylc/flow/etc/cylc
```diff
@@ -17,95 +17,71 @@
...
+# Wrapper script to support multiple installed Cylc, Rose & Isodatetime
+# versions. Handles Conda and Python virtual environments, and legacy plain
+# installations.
 #
 # WRAPPER INSTALLATION AND CONFIGURATION
-#---------------------------------------
+# --------------------------------------
 # - Copy this script as "cylc" into default $PATH, on scheduler and job hosts
-# - If using Rose, create a "rose" symlink (ln -s cylc rose)
+# - Create a "isodatetime" symlink (ln -s cylc isodatetime)
```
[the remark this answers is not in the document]
dpmatthews: Fixed

cylc/flow/etc/cylc
dpmatthews: This isn't related to this change - you can't use the wrapper with Cylc 8 for anything prior to rc1 because we no longer set PATH so the old rose bash cli doesn't work.

wxtim (approved):

cylc/flow/etc/cylc
```diff
@@ -17,95 +17,71 @@
...
+# CYLC_HOME - full path to the Cylc environment.
+# Can be set by users in order to use an environment outside of the ROOT
+# locations. However, it is not passed by the scheduler to remote platforms so
+# use of CYLC_VERSION & CYLC_HOME_ROOT_ALT is preferred.
+#
+# INSTALLING ENVIRONMENTS
+# -----------------------
+#
+# Releases should be installed into environments in the ROOT location, named
+# cylc-$CYLC_VERSION. We recommend using the cylc-flow version plus an additional
```
hjoliver, line 71: A small change would be handy at NIWA, to avoid having to edit some code down in the guts of the wrapper:
```
CYLC_ENV_NAME_PREFIX="cylc-"   # editable prefix
# ...
${CYLC_ENV_NAME_PREFIX}${CYLC_VERSION}
```
Our HPC admins use this convention: `.../share/cylc/8.0rc1` etc.

hjoliver: (However, not a biggie)

dpmatthews: I like this idea. However, I think it also means making the name of the default version configurable (also like the idea of this). Do you mind if we do this in a separate PR?

oliver-sanders (approved):

wxtim:
Is there any reason why this has stopped?

hjoliver:
Probably my fault - I mean to approve, aside from the change suggestion above that I agreed (via :+1: ) to punt to another issue.

hjoliver (approved):

cylc/flow/etc/cylc
dpmatthews: See https://github.com/cylc/cylc-flow/issues/4555#issuecomment-1066618751

PR #4667, opened 2022-02-09
Check job not in prep already before triggering.

hjoliver:
These changes close #4665

Don't retrigger a task that is already preparing for job submission.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.
- [x] Added test.
- [x] Appropriate change log entry included.
- [x] No documentation update required.

PR #4668, opened 2022-02-09
Preparing vs active

oliver-sanders:
> Built on #4667

* RC2 with no associated issues.
* Follow on from https://github.com/cylc/cylc-flow/pull/4667.

Found three instances where `TASK_STATUS_PREPARING` might be missing from a `TASK_STATUSES_ACTIVE` state check.

If we agree the check should be extended I'll add a test to enforce it.

**Requirements check-list**
- [x] Logging change only - no changelog.

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -737,6 +737,8 @@ class Scheduler:
             itask.platform['install target'] = (
                 get_install_target_from_platform(itask.platform))
             if (
+                # we don't need to remote-init for preparing tasks because
+                # they will be reset to waiting on restart
                 itask.state(*TASK_STATUSES_ACTIVE)
                 and not (
                     is_platform_with_target_in_list(
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -822,7 +822,7 @@ class TaskPool:
                 itask.copy_to_reload_successor(new_task)
                 self._swap_out(new_task)
                 LOG.info(f"[{itask}] reloaded task definition")
-                if itask.state(*TASK_STATUSES_ACTIVE):
+                if itask.state(*TASK_STATUSES_ACTIVE, TASK_STATUS_PREPARING):
                     LOG.warning(
                         f"[{itask}] active with pre-reload settings"
                     )
@@ -898,6 +898,8 @@ class TaskPool:
                 and itask.state(*TASK_STATUSES_ACTIVE)
                 and not itask.state.kill_failed
             )
+            # we don't need to check for preparing tasks because they will be
+            # reset to waiting on restart
             for itask in self.get_tasks()
         )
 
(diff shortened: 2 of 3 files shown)


cylc/flow/scheduler.py
```diff
@@ -737,7 +737,7 @@ def restart_remote_init(self):
             itask.platform['install target'] = (
                 get_install_target_from_platform(itask.platform))
             if (
-                itask.state(*TASK_STATUSES_ACTIVE)
+                itask.state(*TASK_STATUSES_ACTIVE, TASK_STATUS_PREPARING)
```
oliver-sanders, line 740: Without this remote-init on restart could be skipped for preparing tasks?

cylc/flow/task_pool.py
```diff
@@ -822,7 +822,7 @@ def reload_taskdefs(self) -> None:
                 itask.copy_to_reload_successor(new_task)
                 self._swap_out(new_task)
                 LOG.info(f"[{itask}] reloaded task definition")
-                if itask.state(*TASK_STATUSES_ACTIVE):
+                if itask.state(*TASK_STATUSES_ACTIVE, TASK_STATUS_PREPARING):
```
oliver-sanders, line 825: We should warn for preparing tasks as pre-reload settings may have been baked into the filesystem?

cylc/flow/task_pool.py
```diff
@@ -895,7 +895,7 @@ def can_stop(self, stop_mode):
         return not any(
             (
                 stop_mode in [StopMode.REQUEST_CLEAN, StopMode.REQUEST_KILL]
-                and itask.state(*TASK_STATUSES_ACTIVE)
+                and itask.state(*TASK_STATUSES_ACTIVE, TASK_STATUS_PREPARING)
```
oliver-sanders, line 898: Not sure about this one.

cylc/flow/scheduler.py
hjoliver: Tasks can't be `preparing` here. On restart, we immediately reset `preparing` tasks to `waiting` so that they re-prep from scratch (with a new submit number).

hjoliver: A comment on this in the code would be good though, for future reference.  (I just had to do some quick testing to confirm the behaviour).

cylc/flow/task_pool.py
hjoliver: OK as-is I think. We can safely shut down with (unsubmitted) `preparing` tasks, because they will be re-prepped on restart as noted above.

hjoliver (changes requested):
IMO, 1/3 good, 2/3 not needed.

cylc/flow/scheduler.py
oliver-sanders: Added a comment and an integration test.

cylc/flow/task_pool.py
oliver-sanders: Makes sense, reverted the change, added a comment and an integration test.

Issue #4618, opened 2022-01-27
`cylc clean --rm` - improve interactive message


MetRonnie:
When using the `--rm` option to do a targeted clean, the message that comes up for interactive prompt suggests the entire workflow will be removed, which is a bit alarming.
```
$ cylc clean temp4 --rm .service
Would remove the following workflows:
  temp4/run1
Remove these workflows (y/n):
```

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

MetRonnie:
Reopening as 8.x issue

Issue #4673, opened 2022-02-10
cylc clean: improvements


MetRonnie:
- [x] Replace current `--force` option with `--yes / -y` - more apt (#4675)
  - Should `--force` come back in a different form? E.g. #4450
- [x] Doing `cylc clean toplevel` and typing `y` at the interactive step for a workflow dir `toplevel` containing multiple runs, does not actually remove `~/cylc-run/toplevel` at the moment (moved to #4963)
- [x] Improve message for incompatible database - give instructions for how to manually delete db (#4876)
  - alternative to #4450

**Pull requests welcome!**

Issue #4450, opened 2021-10-05
Allow `cylc clean` to work with `--force` when DB is corrupted/incompatible


MetRonnie:
**Describe exactly what you would like to see in an upcoming release**

Instead of
```
$ cylc clean temp5 --force
ServiceFileError: Cannot clean - Workflow database is incompatible with Cylc 8.0b3.dev, or is corrupted
```
we should be able to use `--force` to force it to clean, possibly with a message similar to when there is no DB:
```
INFO - No workflow database - will only clean locally
```

**Pull requests welcome!**

MetRonnie:
Marking as a question to bring up on Friday in case there is a reason not to do this

oliver-sanders:
Low priority.

MetRonnie:
Every now and then I run into this, which is a bit frustrating. I'd be tempted to promote this to 8.0.0 or rc2

oliver-sanders:
How are you getting corrupted DBs?

MetRonnie:
Not so much corrupted DBs as trying to clean Cylc 7 runs

oliver-sanders:
Ah, do we not prohibit cleaning Cylc 7 runs?!

oliver-sanders:
For a Cylc 7 workflow you get:

```
Workflow database is incompatible with Cylc 8.0rc1.dev, or is corrupted
```

Which is fine. I don't think `cylc clean` should attempt to remove Cylc 7 workflows, users should continue to use `rose suite-clean` (which has remote support) or employ whatever local practices they usually use.

We could detect a Cylc 7 workflow and print this advice for clarity?

PR #4675, opened 2022-02-10
`cylc clean`: rename `--force` to `--yes`

MetRonnie:
These changes partially address #4673 and #4618 

The `--force` behaviour is more of a `--yes` than `--force`, so rename it.

`--force` can come back later if we want to allow users to override the current "cannot clean" scenarios - see #4450

**Requirements check-list**
- [ ] Appropriate tests included
- [x] Docs PR: https://github.com/cylc/cylc-doc/pull/403

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -54,14 +54,23 @@ First Release Candidate for Cylc 8.
 
 (See note on cylc-8 backward-incompatible changes, above)
 
+Also note: you will not be able to restart workflows run with previous
+Cylc 8 pre-releases due to changes in the workflow database structure
+([#4581](https://github.com/cylc/cylc-flow/pull/4581))
+
 ### Enhancements
 
-[#4581](https://github.com/cylc/cylc-flow/pull/4581) - Job and task history
-is now loaded into the window about active tasks. Reflow future tasks now set
-to waiting.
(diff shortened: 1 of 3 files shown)


oliver-sanders:
> partially address https://github.com/cylc/cylc-flow/issues/4673 and https://github.com/cylc/cylc-flow/issues/4618

What's left?

MetRonnie:
The second task on #4673

As for #4618, I've changed the multi workflow clean message
```diff
-Would remove the following workflows:
+Would clean the following workflows:
```
which is better because it doesn't explicitly say the whole run dir will be removed when you're only doing `--rm subdir`, but it still kinda implies it, so have left the issue open

oliver-sanders:
So closes https://github.com/cylc/cylc-flow/issues/4618

oliver-sanders (approved):

MetRonnie:
I still think the prompt message is a bit alarming for targeted clean, perhaps the prompt could do with a slight rethink

oliver-sanders:
I think it's ok, to explain better we would need something like a dry run listing exactly what we would expect to be removed where (perfectly possible).

oliver-sanders:
Very small scope, can single review this.

PR #4688, opened 2022-02-15
log removal of contact file

oliver-sanders:
If a client fails to connect and the workflow can be proven not to be running, then the contact file will be removed.

This PR logs when contact files are removed which seems like sensible logging to have.

**Requirements check-list**

### DIFF
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -513,6 +513,10 @@ def detect_old_contact_file(reg: str, contact_data=None) -> None:
         # ... the process isn't running so the contact file is out of date
         # remove it
         try:
+            LOG.info(
+                f'Removing contact file for {reg}'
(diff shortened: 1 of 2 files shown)


datamel (approved):
Thanks @oliver-sanders. I've checked out and run the new test locally. Unsure why the macos fast tests are unhappy, can't think of a reason it would be related.

oliver-sanders:
> Unsure why the macos fast tests are unhappy

There's an issue with the integration test battery where it sometimes hangs on completion. I've yet to work out why, I mostly see it on Mac, probs that, rerunning for safety (especially given how close we are to release!).

I think it may be related to the slow shutdown time for the integration tests, a bit tricky to debug.

MetRonnie (approved):

oliver-sanders:
Previously stuck integration tests passed on round two.

PR #4689, opened 2022-02-15
Fix contact file exists message

datamel:
While investigating healthcheck bug, I noticed the `CONTACT_FILE_EXISITS_MSG` is incorrect. It was being sent the original command rather than the process id.  

To test, 
`cylc play workflow` 
`cylc play workflow`
On master the pid is not passed to the kill command.

- [x] Does not need tests (why? Documentation).
- [x] No change log entry required (why? trivial change).


wxtim (approved):

oliver-sanders (approved):
MB :(

PR #4703, opened 2022-02-21
Upper pin markupsafe

MetRonnie:
```
ImportError: cannot import name 'soft_unicode' from 'markupsafe'
```

Jinja2 @ 2.11.3 has [a dependency `MarkupSafe>=0.23`](https://github.com/pallets/jinja/blob/cf215390d4a4d6f0a4de27e2687eed176878f13d/setup.py#L53 ). Unfortunately Pallets have made a breaking change in the API in MarkupSafe 2.1.0, but have no plans to patch version 2.11 of Jinja2 with an upper pin on MarkupSafe, so we have to do it ourselves.

https://github.com/pallets/jinja/issues/1585

Note: we have discussed moving to version 3.0 of Jinja2, which is the only supported version going forward (see comments from the Pallets team on the issue above). The difficulty is convincing ourselves this will not introduce breaking changes to Cylc workflows that are dependent on Jinja2. It is is unfeasible (and not our job) to write a test spec for Jinja2; however, it might be enough to have a suitably complex test workflow that we can diff with a reference file to see if Jinja2 updates cause changes in output.

**Requirements check-list**
- [x] Does not need tests 
- [x] (master branch) I have opened a documentation PR: cylc/cylc-doc/pull/408

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,18 @@ creating a new release entry be sure to copy & paste the span tag with the
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
 -------------------------------------------------------------------------------
+## __cylc-8.0rc2 (<span actions:bind='release-date'>Upcoming</span>)__
+
+Second Release Candidate for Cylc 8 suitable for acceptance testing.
+
+### Fixes
+
+[#4703](https://github.com/cylc/cylc-flow/pull/4703) - Fix `ImportError` when
+validating/running a Jinja2 workflow (for users who have installed Cylc
+using `pip`.)
+
+
+-------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


oliver-sanders (approved):

MetRonnie:
- `conda builds / test_conda_install` failing on master also, fix in #4674
- `manylinux-compatibility / test` failing for Python 3.9 - not sure why

oliver-sanders:
The manylinux failure is due to a combination of kiwisolver dropping manylinux1 support and pip22 failing on error (so no longer falling back to the previous version as it used to).

The (transient) kiwisolver dep is `kiwisolver>=1.0.1`, manylinux support was only dropped at the most recent `1.3.2` release so we are good here, for now...

wxtim (approved):
Annoying, but valid·

PR #4704, opened 2022-02-21
actions: pin manylinux to pip<22

oliver-sanders:
> The manylinux failure is due to a combination of kiwisolver dropping manylinux1 support and pip22 failing on error (so no longer falling back to the previous version as it used to).
> 
> The (transient) kiwisolver dep is `kiwisolver>=1.0.1`, manylinux support was only dropped at the most recent `1.3.2` release so we are good here, for now...
 
This might be enough to patch things over for now 

**Requirements check-list**

### DIFF
--- a/.github/workflows/test_manylinux.yml
+++ b/.github/workflows/test_manylinux.yml
@@ -48,6 +48,16 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Downgrade PIP
+        run: |
+          # pip v22+ will fail on error rather than falling back to the
+          # previous release
+          # the old behaviour is better for this check
+          if (( $(pip --version | sed 's/pip \([0-9]*\).*/\1/') >= 22 ))
+          then
+            pip install 'pip<22'
+          fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/test_manylinux.yml
```diff
@@ -48,6 +48,16 @@ jobs:
...
+      - name: Downgrade PIP
+        run: |
+          # pip v22+ will fail on error rather than falling back to the
+          # previous release
+          # the old behaviour is better for this check
+          if (( $(pip --version | sed 's/pip \([0-9]*\).*/\1/') >= 22 ))
+          then
+            pip install 'pip<22'
+          fi
+
```
MetRonnie, line 60: Can we not do `pip install --prefer-binaries` instead?

oliver-sanders: Not sure how that would help?

MetRonnie: It might lead to pip trying to install the compatible wheel instead of the tar.gz that it's been failing with. I haven't tried it though, whatever

PR #4707, opened 2022-02-21
Fix warnings and tidy

MetRonnie:
This is a small change with no associated Issue. In light of the markupsafe incident, I thought it would be a good idea to address the warnings that currently occur during pytest runs (although actually there were no warnings coming through for the markupsafe breaking change). Only exception is the `pytest.PytestUnhandledThreadExceptionWarning` which is tracked in #4706 

**Requirements check-list**
- [x] Does not need tests

### DIFF
--- a/cylc/flow/main_loop/__init__.py
+++ b/cylc/flow/main_loop/__init__.py
@@ -48,7 +48,7 @@ Main loop plugins can be activated either by:
 
      $ # run a workflow using the "health check" and "auto restart" plugins:
      $ cylc play my-workflow --main-loop 'health check' \
-       --main-loop 'auto restart'
+--main-loop 'auto restart'
 
 * Adding them to the default list of plugins in
   :cylc:conf:`global.cylc[scheduler][main loop]plugins` e.g:
--- a/cylc/flow/network/server.py
+++ b/cylc/flow/network/server.py
@@ -324,7 +324,7 @@ class WorkflowRuntimeServer(ZMQSocketBase):
             executed: ExecutionResult = schema.execute(
                 request_string,
                 variable_values=variables,
-                context={
+                context_value={
                     'resolvers': self.resolvers,
                     'meta': meta or {},
                 },
--- a/tests/integration/test_examples.py
+++ b/tests/integration/test_examples.py
@@ -211,10 +211,10 @@ def test_module_two(myflow):
     assert myflow.uuid_str
 
 
-async def test_db_select(one, run, db_select):
+async def test_db_select(one, start, db_select):
     """Demonstrate and test querying the workflow database."""
     schd = one
-    async with run(schd):
+    async with start(schd):
         # Note: can't query database here unfortunately
         pass
     # Now we can query the DB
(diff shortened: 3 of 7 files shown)


hjoliver (approved):
Looks sensible :+1:

cylc/flow/main_loop/__init__.py
```diff
@@ -46,7 +46,7 @@
 
   .. code-block:: console
 
-     $ # run a workflow using the "health check" and "auto restart" plugins:
+     # run a workflow using the "health check" and "auto restart" plugins:
```
oliver-sanders, line 49: This was correct before. Console lines should start with a PS1/2 otherwise they are considered command output.

Take a look at this example with the GH lexer to see the issue:

**console:**

```console
$ echo foo
foo
$ echo '# bar'
# bar
```

**bash:**

```bash
$ echo foo
foo
$ echo '# bar'
# bar
```

With pygments the result is worse.

cylc/flow/network/server.py
```diff
@@ -324,7 +324,7 @@ def graphql(
             executed: ExecutionResult = schema.execute(
                 request_string,
                 variable_values=variables,
-                context={
```
oliver-sanders, line 327: Internal deprecation within Graphene?

cylc/flow/main_loop/__init__.py
oliver-sanders:
```suggestion
     $ # run a workflow using the "health check" and "auto restart" plugins:
```

oliver-sanders (approved):

cylc/flow/network/server.py
oliver-sanders: At which version?

cylc/flow/main_loop/__init__.py
```diff
@@ -48,7 +48,7 @@
 
      $ # run a workflow using the "health check" and "auto restart" plugins:
      $ cylc play my-workflow --main-loop 'health check' \
-       --main-loop 'auto restart'
+--main-loop 'auto restart'
```
oliver-sanders, line 51: Does this really work!

oliver-sanders: Surprised that works, cool.

cylc/flow/network/server.py
oliver-sanders: @MetRonnie just checking that this doesn't remove compatibility with our specified version range before merging.

MetRonnie: I have done some digging; we are pinned at `graphene>=2.1,<3`, and graphene 2.1.0 is pinned at `graphql-core>=2.0,<3`. I have looked through graphql-core 2.0.0 and only the `context_value` arg is used. (This was deprecated in favour of `context` before they changed their minds and made `context` deprecated in favour of `context_value` in between 2.0.0 and the latest 2.x release)

oliver-sanders: Ok, that's strange!

PR #4710, opened 2022-02-22
tests/i: attempt to stabilise test_server tests

### LINKED ISSUE #4708: `tests/integration/test_server.py::test_receiver` hangs flakily
It always seems to be `tests/integration/test_server.py::test_receiver` that is the logged running test when the GH Actions run times out on MacOS. Note it also times out on Ubuntu sometimes, though less frequently

**Pull requests welcome!**

oliver-sanders:
* Hopefully closes #4708.
* Switch from run to start (no need to have main loop running for tests).
* Add a timeout just in case.
* Fix tests and add a positive test case to prevent them breaking again.

**Requirements check-list**

### DIFF
--- a/tests/integration/test_server.py
+++ b/tests/integration/test_server.py
@@ -78,32 +78,48 @@ def test_pb_entire_workflow(myflow):
     assert data.workflow.id == myflow.id
 
 
-@pytest.fixture
-async def accident(flow, scheduler, run, one_conf):
-    reg = flow(one_conf)
-    schd = scheduler(reg)
-    async with run(schd):
-        yield schd
-
-
-async def test_listener(accident):
+async def test_listener(one, start):
     """Test listener."""
-    accident.server.queue.put('STOP')
-    async with timeout(2):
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Hard to verify whether or not this actually fixes the issue as it's hard to impossible to replicate locally.

With these changes if the test hangs it should at least fail and allow pytest to wrap up.

However, I'm not sure whether this test ever did time out (can't see a way for it to have) so it's likely that the timeout was in tear down.

MetRonnie (approved):
Looks good

> However, I'm not sure whether this test ever did time out (can't see a way for it to have) so it's likely that the timeout was in tear down.

The last test to run after `tests/integration/test_server.py::test_receiver` is `tests/integration/test_task_pool.py::test_preparing_tasks_on_restart`, but the latter doesn't show up as started in the log, so it does seem like the former was the problem

MetRonnie:
Dang, you were right, this wasn't it

PR #4713, opened 2022-02-24
scheduler: process DB after configure complete

oliver-sanders:
> Result of test instability spotted by @MetRonnie, posted on Element
>
>> While trying to debug the flakiness of the integration tests on GH Actions MacOS, I noticed that running the integration tests with -n 0 --dist=no to turn off pytest-xdist results in tests/integration/test_[scan.py](http://scan.py/)::test_workflow_params failing 100% of the time (on linux)

I think this is a bug because if the workflow bails during the first main loop iteration the workflow will be left in a state we cannot restart from.

If it's an issue RC2, otherwise 8.1.0.

**Requirements check-list**

### DIFF
--- a/tests/integration/test_scan.py
+++ b/tests/integration/test_scan.py
@@ -17,6 +17,7 @@
 
 from contextlib import suppress
 from pathlib import Path
+import re
 from shutil import rmtree
 from tempfile import TemporaryDirectory
 from typing import List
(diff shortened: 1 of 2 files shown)


tests/integration/test_scan.py
```diff
@@ -339,6 +339,8 @@ async def test_workflow_params(
             assert uuid_key in flow['workflow_params']
             # check the workflow uuid key matches the scheduler value
             assert flow['workflow_params'][uuid_key] == schd.uuid_str
+        else:
```
MetRonnie, line 342: Need a `break` in the loop otherwise the `else` block will always run
```suggestion
            break
        else:
```

oliver-sanders: (still looking at this as the situation with xdist is now reversed)

oliver-sanders:
Ok, I think that's done it. The reason the test results differed when run under xdist is because the workflow registration differed which is in turn because the [configured environment](https://github.com/cylc/cylc-flow/blob/75a6235d19aa261fbe5be66b95895a81f01e8cd6/pytest.ini#L34-L36) is not getting picked up consistently.

MetRonnie (approved):
Do we want a 2nd reviewer?

hjoliver (approved):

PR #4718, opened 2022-02-24
tests/i: fix hanging tests

oliver-sanders:
* Some tests were not correctly shutting down their resources so would cause pytest to hang during the cleanup phase.
* Due to some interaction, possibly between asyncio, pytest-asyncio and pytest-xdist, this does not impact the tests when run in distributed mode.

**Requirements check-list**

### DIFF
--- a/tests/integration/test_scheduler.py
+++ b/tests/integration/test_scheduler.py
@@ -20,7 +20,7 @@ import pytest
 from typing import Any, Callable
 
 from cylc.flow.exceptions import CylcError
-from cylc.flow.scheduler import Scheduler
+from cylc.flow.scheduler import Scheduler, SchedulerStop
 from cylc.flow.task_state import (
     TASK_STATUS_WAITING,
     TASK_STATUS_SUBMIT_FAILED,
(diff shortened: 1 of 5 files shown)


tests/integration/test_scheduler.py
```diff
@@ -86,33 +86,44 @@ async def test_shutdown_CylcError_log(one: Scheduler, run: Callable):
...
             pass
+
+    # check the log records after attempted shutdown
     assert str(exc.value) == "Error on shutdown"
     last_record = log.records[-1]
     assert last_record.message == "CylcError: Error on shutdown"
     assert last_record.levelno == logging.ERROR
 
+    # shut down the scheduler properly
+    await Scheduler._shutdown(schd, Exception('because I said so'))
```
MetRonnie, line 107: This is cleaner in the log, less likely to distract if debugging failure due to some other reason (may have to import `SchedulerStop`)
```suggestion
    await Scheduler._shutdown(schd, SchedulerStop('because I said so'))
```

MetRonnie (commented):
Nice

MetRonnie:
Confirmed `pytest tests/integration/ -n 0 --dist=no` passes when combining this with #4713

MetRonnie:
Just had a thought, might be good to write down in `tests/integration/readme.md` about the need for the Scheduler to shutdown

MetRonnie:
Just had a thought, might be good to write down in `tests/integration/readme.md` about the need for the Scheduler to shutdown

oliver-sanders:
Added some more documentation.

MetRonnie (approved):
Might be good to have a 2nd reviewer, they might learn a thing or two

datamel (approved):
Code read and checked out locally, tests running smoothly. Thanks for adding the extra info in README.

PR #4720, opened 2022-02-28
Stop traceback when `ParsecError` occurs during config load on workflow run

### LINKED ISSUE #4216: config: handle error on config (re)load
* A `ParsecError` raised in `cylc.flow.config` should be an "expected error" that gets formatted nicely for the user.
* A `ParsecError` raised during workflow runtime is an "unexpected error" (i.e. bug) that should be reported and fixed.

Additionally, a `ParsecError` (or `CylcError`) raised during config reload will cause the scheduler to crash. This is a change in behaviour for those moving from Rose 2019 as `rose suite-run --reload` invoked `cylc validate --strict` before attempting reload.

Suggest:

* The scheduler should not die on config errors at reload.
* `ParsecError` should be handled differently at config-time as at run-time.

One approach for separating config-time errors is to add a flag on the exception:

```python
try:
    self.cfg = WorkflowConfig(...)
except ParsecError as exc:
    exc.config_time_error = True
    raise
```

**Pull requests welcome!**

MetRonnie:
These changes close #4216

`ParsecError`s that occur during config load (either on workflow start or reload) no longer cause traceback, instead just giving a one line message, because we expect these to occur.

'ParsecError`s at any other time during workflow run are unexpected and still give traceback.

Additionally, the Scheduler won't crash if a `ParsecError` occurs during reload (not sure if this was already fixed, but anyway it is fixed now for sure).

**Requirements check-list**
- [x] Appropriate tests are included (integration)


MetRonnie:
Codecov upload and MacOS tests being particularly flaky today. Tests passing otherwise

wxtim (commented):
Overall happy
- [x] Read PR
  - Made a couple of comments about grammar &c, mostly reflecting dissatisfaction with my own previous work!
- [x] Manually tested as many scenarios as I could think of
- [x] Run tests locally
- [x] Looked at testing of this
  - I'm a little surprised that this change didn't affect `tests/unit/parsec/test_validate.py`, but I think it's probably a sign that the tests have been formatted to be robust against small changes to exception messages which I think is a _good thing_™.

CHANGES.md
```diff
@@ -441,6 +445,10 @@ not the whole workflow. ([#4076](https://github.com/cylc/cylc-flow/pull/4076))
 ([#4109](https://github.com/cylc/cylc-flow/pull/4109)). You can allow them by
 setting `flow.cylc[scheduler]allow implicit tasks` to `True`.
 
+Named checkpoints have been removed ([#3906](https://github.com/cylc/cylc-flow/pull/3906))
```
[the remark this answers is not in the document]
MetRonnie: Went missing in https://github.com/cylc/cylc-flow/commit/831ca58190c0ae577ac2d63166678d51b5059c0b

MetRonnie:
Changes in response to Tim: https://github.com/cylc/cylc-flow/compare/eddf71b0b41a3d71dd1b8ef96fb8729512c2fcfe..5d7e1d9da1c90f7b50bb0c7ad08adf466014f1cd

also viewable by clicking on "force pushed" above where it says
> [MetRonnie](https://github.com/MetRonnie) [force-pushed](https://github.com/cylc/cylc-flow/compare/eddf71b0b41a3d71dd1b8ef96fb8729512c2fcfe..5d7e1d9da1c90f7b50bb0c7ad08adf466014f1cd) the parsec-error-tb branch from [eddf71b ](https://github.com/cylc/cylc-flow/commit/eddf71b0b41a3d71dd1b8ef96fb8729512c2fcfe)to [5d7e1d9](https://github.com/cylc/cylc-flow/commit/5d7e1d9da1c90f7b50bb0c7ad08adf466014f1cd)

tests/integration/utils/flow_tools.py
```diff
@@ -78,22 +84,20 @@ async def _start_flow(
...
     await schd.install()
 
-    # start
     try:
         await schd.start()
-        yield caplog
 
-    # stop
     finally:
+        yield caplog
```
oliver-sanders, line 97: What's the logic of moving the yield, feels wrong.

tests/integration/utils/flow_tools.py
```diff
@@ -102,31 +106,26 @@ async def _run_flow(
...
+        # Cleanup: need to shut down, but time out in case something goes wrong
         async with timeout(5):
-            # ask the scheduler to shut down nicely
+            # ask the scheduler to shut down nicely, let main loop handle it:
             schd._set_stop(StopMode.REQUEST_NOW_NOW)
-            await task
-
+            if task:
+                await task
```
oliver-sanders, line 128: If the workflow failed to start then the main loop will not be running. This is why it separates the errors and calls `schd.shutdown`.

tests/integration/utils/flow_tools.py
MetRonnie: Yes. Otherwise you won't get the caplog yielded

tests/integration/utils/flow_tools.py
MetRonnie: The trouble is that
```python
try:
    schd.start()
except Exception as exc:
```
was making it impossible to test the Scheduler's error handling

tests/integration/utils/flow_tools.py
MetRonnie: But you have a good point... this now breaks the teardown if an assertion fails in the test

MetRonnie: Should be addressed now - https://github.com/cylc/cylc-flow/compare/5d7e1d9da1c90f7b50bb0c7ad08adf466014f1cd..04164177e63578bf69578ecf4f36f76e740ffe97

MetRonnie:
Test failures due to GH Actions flakiness 😒

cylc/flow/scheduler.py
```diff
@@ -1025,7 +1031,10 @@ def command_reload_workflow(self):
         pri_dao = self.workflow_db_mgr.get_pri_dao()
         pri_dao.select_workflow_params(self._load_workflow_params)
 
-        self.load_flow_file(is_reload=True)
+        try:
+            self.load_flow_file(is_reload=True)
+        except (ParsecError, CylcConfigError) as exc:
+            raise CommandFailedError(f"{type(exc).__name__}: {exc}")
```
oliver-sanders, line 1037: Why the re-raise here?

cylc/flow/scheduler.py
```diff
@@ -441,7 +443,11 @@ async def configure(self):
                 pri_dao.close()
 
         self.profiler.log_memory("scheduler.py: before load_flow_file")
-        self.load_flow_file()
+        try:
+            self.load_flow_file()
+        except ParsecError as exc:
+            exc.schd_expected = True
```
oliver-sanders, line 449: Could you add a comment explaining what the `exc.schd_expected` is about. Not so keen on the name `schd_expected` but can't think of anything better.

cylc/flow/scheduler.py
```diff
@@ -1771,6 +1784,8 @@ async def _shutdown(self, reason: Exception) -> None:
             except OSError as exc:
                 LOG.warning(f"failed to remove workflow contact file: {fname}")
                 LOG.exception(exc)
+            else:
+                self.contact_data = None
```
oliver-sanders, line 1788: What's this do? Could you add a comment.

tests/integration/test_examples.py
```diff
@@ -164,8 +164,7 @@ def killer():
     # make sure that this error causes the flow to shutdown
     with pytest.raises(MyException):
         async with run(schd) as log:
-            # evil sleep - gotta let the except mechanism do its work
-            await asyncio.sleep(0.1)
+            pass
```
oliver-sanders, line 167:
```suggestion
            # the main loop will run when the workflow is being shut down
            pass
```

tests/integration/test_scheduler.py
```diff
@@ -237,3 +245,108 @@ async def test_no_poll_waiting_tasks(
...
+
+    if reload:
+        one_conf.update(test_conf)
+        run_dir = Path(get_workflow_run_dir(reg))
+        async with run(schd) as log:
+            # Shouldn't be any errors at this stage:
+            assert not log_filter(log, level=logging.ERROR)
+            # Modify flow.cylc:
+            _make_flow(get_cylc_run_dir(), run_dir, one_conf, '')
+            schd.queue_command('reload_workflow', {})
```
oliver-sanders, line 307: Should be able to reload the workflow from the scheduler interface (the main loop isn't running whilst this sync code is run so we can act as it would), this way the test doesn't rely on the fine-details of shutdown logic.

```suggestion
            schd.command_reload_workflow()
```

tests/integration/test_scheduler.py
```diff
@@ -237,3 +245,108 @@ async def test_no_poll_waiting_tasks(
...
+    if not reload:
+        one_conf.update(test_conf)
+    reg: str = flow(one_conf)
+    schd: Scheduler = scheduler(reg)
+    log: pytest.LogCaptureFixture
+
+    if reload:
+        one_conf.update(test_conf)
+        run_dir = Path(get_workflow_run_dir(reg))
+        async with run(schd) as log:
```
oliver-sanders, line 302: If we can avoid queueing tasks we don't need to run the main loop:

```suggestion
        async with start(schd) as log:
```

tests/integration/test_scheduler.py
```diff
@@ -237,3 +245,108 @@ async def test_no_poll_waiting_tasks(
...
+    start: Callable,
+    scheduler: Callable,
+    log_filter: Callable,
+    monkeypatch: pytest.MonkeyPatch
+):
+    """Test that ParsecErrors - that occur at any time other than config load
+    when running a workflow - are displayed with traceback, because they are
+    not expected."""
+    reg: str = flow(one_conf)
+    schd: Scheduler = scheduler(reg)
```
oliver-sanders, line 336: FYI there's a fixture called "one" which returns a scheduler created with the "one_conf" saving these steps.

cylc/flow/scheduler.py
MetRonnie: That wouldn't work for `CylcConfigError`s

cylc/flow/scheduler.py
MetRonnie: Isn't this straightforward? If the contact file is removed successfully this property is updated to reflect that.

Seeing as we don't normally do anything with the Scheduler after shutdown, the main use for this is testing convenience. It allows checking for `self.contact_data` during tear down of the `run` fixture
https://github.com/cylc/cylc-flow/blob/a917e65948b80590a313b741bb7079437f1d0dcc/tests/integration/utils/flow_tools.py#L146-L149

tests/integration/test_examples.py
MetRonnie: Counter suggestion?
```suggestion
            # The `run` fixture's shutdown logic waits for the main loop to run
            pass
```

tests/integration/test_scheduler.py
MetRonnie: That doesn't work as it bypasses the Scheduler's nominal way of running commands and handling their errors. Perhaps we should add a leading underscore to the `Scheduler.command_*`s

MetRonnie:
Will push up a commit addressing your review comments after your response to my response

tests/integration/test_scheduler.py
oliver-sanders: Ah, yeah that makes sense, ok.

Yes we should mark most of the Scheduler methods as private, however, I think the scheduler needs some serious tidying first.

> [This issue](https://github.com/cylc/cylc-flow/issues/3329) (originally tagged at Cylc 8) will overhaul the way commands are implemented and the way they send their responses back replacing `(True, "command queued")` with a subscription ID to an event stream. I'd like to take that opportunity to move the commands out of the Scheduler.

tests/integration/test_examples.py
oliver-sanders: Fine, so long as it's explained.

cylc/flow/scheduler.py
oliver-sanders: Wouldn't be asking if it was obvious, high chance I'll come along in a few months having forgotten about this and remove it. Worth adding a quick comment so we know this odd line is for the integration tests.

cylc/flow/scheduler.py
oliver-sanders: Do we have a problem if `CylcConfigError`s are raised during reload?

MetRonnie: It would produce traceback when we surely only want a one-liner?

oliver-sanders: Would have thought so.

oliver-sanders:
@MetRonnie waiting on `CylcConfigError` which should probably be given the same treatment?

https://github.com/cylc/cylc-flow/pull/4720#discussion_r822676878

MetRonnie:
`CylcConfigError`s are `CylcError`s so don't need any special treatment.#

E.g. try out
```diff
diff --git a/cylc/flow/config.py b/cylc/flow/config.py
index 44e7b2a6e..70046be76 100644
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -58,6 +58,7 @@ from cylc.flow.id import Tokens
 from cylc.flow.cycling.integer import IntegerInterval
 from cylc.flow.cycling.iso8601 import ingest_time, ISO8601Interval
 from cylc.flow.exceptions import (
+    CylcConfigError,
     CylcError,
     WorkflowConfigError,
     IntervalParsingError,
@@ -1963,6 +1964,7 @@ class WorkflowConfig:
         # Generate a map of *task* members of each family.
         # Note we could exclude 'root' from this and disallow use of 'root' in
         # the graph (which would probably be quite reasonable).
+        raise CylcConfigError("Moo")
         family_map = {
             family: [
                 task for task in tasks
```
followed by `cylc play`

oliver-sanders:
> CylcConfigErrors are CylcErrors so don't need any special treatment.

Ok, I'll resolve that comment then.

cylc/flow/scheduler.py
oliver-sanders: Resolved - https://github.com/cylc/cylc-flow/pull/4720#issuecomment-1084698176

oliver-sanders (approved):
Tested, works nicely, couldn't break it :+1:.

PR #4721, opened 2022-02-28
Actions: skip certain tests on MacOS

MetRonnie:
~Turn off xdist in the hope it makes the integration tests stop hanging flakily on GH Actions~

Reduce homebrew installation and skip style, mypy, in order to speed up MacOS runs

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - 'dependencies'
+      - 'infrastructure'
--- a/.github/workflows/test_fast.yml
+++ b/.github/workflows/test_fast.yml
@@ -9,8 +9,9 @@ on:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 15
+    timeout-minutes: 20
     strategy:
+      fail-fast: false # Don't let a failed MacOS run stop the Ubuntu runs
       matrix:
         os: ['ubuntu-latest']
         python-version: ['3.7', '3.8', '3.9']
@@ -32,13 +33,15 @@ jobs:
       - name: Brew Install
         if: startsWith(matrix.os, 'macos')
         run: |
+          # speed up install (https://docs.brew.sh/Manpage#environment)
+          export HOMEBREW_NO_AUTO_UPDATE=1 HOMEBREW_NO_INSTALL_CLEANUP=1
+          echo "[command]brew update"
           brew update
-          brew install shellcheck sqlite3 bash coreutils
-
+          echo "[command]brew install ..."
+          brew install bash coreutils
           # add GNU coreutils and sed to the user PATH
           # (see instructions in brew install output)
-          echo \
-            "$(brew --prefix)/opt/coreutils/libexec/gnubin" \
+          echo "$(brew --prefix)/opt/coreutils/libexec/gnubin" \
             >> "${GITHUB_PATH}"
 
       - name: Apt-Get Install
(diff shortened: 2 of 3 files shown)


MetRonnie:
So as soon as I do this it decides to hang on the "brew install" step instead 🤬

image

MetRonnie:
Yes, had one today: https://github.com/cylc/cylc-flow/runs/5391678146?check_suite_focus=true

And it's about a 50% failure rate over the last week: https://github.com/cylc/cylc-flow/actions/workflows/test_fast.yml?query=branch%3Amaster

image

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - 'dependencies'
+      - 'infrastructure'
```
MetRonnie, line 13: I don't think dependabot can do PEP 517 projects, but at least this will give us updates for GH Actions

MetRonnie:
Dammit, `brew install` taking an inordinate amount of time suddenly seems to be very common

MetRonnie:
Updated to remove the homebrew step completely, think we don't need it

oliver-sanders:
> Updated to remove the homebrew step completely, think we don't need it

* shellcheck
  * May be pre-installed, though most likely at an older version.
* sqlite3
  * Don't think this is needed, if something needs it homebrew will install it as a transient dep anyway.
* bash
  * Needed for any Bash interaction, MacOS default shell is zsh so the Bash install is very old.
* coreutils
  * Needed for `ls`, `rm`, that sort of stuff. MacOS has these by default but they are the BSD rather than GNU variants.

So might be able to get away without it but you would have to do some digging to make sure these things aren't needed, the tests passing wouldn't be quite enough.

MetRonnie:
- shellcheck: made it skip on MacOS
- sqlite: removed
- bash & coreutils:
  > might be able to get away without it but you would have to do some digging to make sure these things aren't needed, the tests passing wouldn't be quite enough.

  I would argue the tests passing is enough for now, but I guess it might not be in future. I've added them back in, good news is they seem to install without problem

.github/dependabot.yml
oliver-sanders: Cylc-flow isn't PEP517 yet but its deps are stored in the .cfg file, let's see how it gets on...

.github/workflows/test_fast.yml
```diff
@@ -77,8 +83,10 @@ jobs:
           bandit -r --ini .bandit cylc/flow
 
       - name: Integration Tests
+        timeout-minutes: 6
         run: |
-          pytest tests/integration
+          pytest -n 0 --dist=no tests/integration
+        # Turned off x-dist due to flakiness on GH Actions for some reason
```
oliver-sanders, line 89: Would rather not change something just to cover up issues at our end, at least until we understand what the issues are?

Can we at least try with just the other fixes first and see whether this is necessary after?

```suggestion
          pytest tests/integration
```

.github/workflows/test_fast.yml
```diff
@@ -77,8 +83,10 @@ jobs:
           bandit -r --ini .bandit cylc/flow
 
       - name: Integration Tests
+        timeout-minutes: 6
```
oliver-sanders, line 86: This might be a little on the tight side?

oliver-sanders:
Most recent run timed out: https://github.com/cylc/cylc-flow/runs/5451461210?check_suite_focus=true#step:13:71

.github/workflows/test_fast.yml
MetRonnie: Removed this change

.github/workflows/test_fast.yml
oliver-sanders: Fair enough.

oliver-sanders (approved):
Ok, let's see what dependabot does for Python.

MetRonnie:
Checks haven't yet run since I pushed, probably due to https://www.githubstatus.com/incidents/tyc8wpsgr2r8

MetRonnie:
Looks like it's done the trick, in the most recent run, ubuntu/py3.7 hung instead of macos

MetRonnie:
I've also added uploading of the cylc-run directory as an artifact if `test_fast.yml` fails, hopefully that might help us debug next time it hangs

PR #4732, opened 2022-03-04
actions: speed up homebrew installs

oliver-sanders:
Hopefully speed up homebrew installs.

Homebrew automatically upgrades all formulae on upgrade/install, great for home machines, less good for CI.

GHA installs a fair few packages with homebrew, as the images drift with time the number of packages upgraded will increase.

Found some option which disables this behaviour, lets see how it works out.

**Requirements check-list**

### DIFF
--- a/.github/workflows/test_fast.yml
+++ b/.github/workflows/test_fast.yml
@@ -32,8 +32,10 @@ jobs:
       - name: Brew Install
         if: startsWith(matrix.os, 'macos')
         run: |
+          # speed up install (https://docs.brew.sh/Manpage#environment)
+          export HOMEBREW_NO_AUTO_UPDATE=1
           brew update
-          brew install shellcheck sqlite3 bash coreutils
+          brew install shellcheck bash coreutils
 
           # add GNU coreutils and sed to the user PATH
           # (see instructions in brew install output)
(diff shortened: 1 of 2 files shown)


oliver-sanders:
Example showing the final step of azure-cli installation taking 4 mins.

https://github.com/cylc/cylc-flow/runs/5422432557?check_suite_focus=true#step:4:1167

oliver-sanders:
Failed to speed it up, maybe I didn't use it right?

oliver-sanders:
I don't think we need to install sqlite for MacOS, I think this was just copied from the apt-get deps.

Looking at the homebrew logs the install is "keg only" which basically means it's not in the "$PATH" but is there incase anything needs to link into it:

```
sqlite is keg-only, which means it was not symlinked into /usr/local,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.
```

Better to let homebrew decide whether it needs to install its own sqlite? We don't install for the functional tests.

oliver-sanders:
Brew install took 1:34 this time, can't say whether that's reliable since the install times seem to be jumping around.

Definitely not upgrading `azure-cli` now anyhoo.

MetRonnie:
Now the `pip install` took 8 minutes! This is turning into whack-a-mole 😣

oliver-sanders:
Superseeded.

PR #4735, opened 2022-03-07
client: suppress IOError when removing defunct contact files

oliver-sanders:
Fixes possible traceback reported [here](https://github.com/cylc/cylc-uiserver/pull/324#issuecomment-1059078983).

* If we can prove that a workflow isn't running but still has a contact file we remove it.
* It is possible that the contact file will already have been removed, e.g. if there are lots of jobs running any one of their clients could remove the contact file, potentially whilst other clients are active.
* Consequently IOError can be rasied, can't remove the contact file if it has already been removed.
* Silence this error.

**Requirements check-list**

### DIFF
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -466,10 +466,8 @@ def detect_old_contact_file(reg: str, contact_data=None) -> None:
 
     Raises:
         CylcError:
-            * If it is not possible to tell for sure if the workflow is running
-              or not.
-            * If the workflow is not, however, the contact file cannot be
-              removed.
+            If it is not possible to tell for sure if the workflow is running
+            or not.
         ServiceFileError(CylcError):
             If old contact file exists and the workflow process still alive.
 
@@ -513,15 +511,21 @@ def detect_old_contact_file(reg: str, contact_data=None) -> None:
         # ... the process isn't running so the contact file is out of date
         # remove it
         try:
-            LOG.info(
-                f'Removing contact file for {reg}'
-                ' (workflow no longer running).'
-            )
             os.unlink(fname)
-            return
+        except FileNotFoundError:
+            # contact file has been removed by another process
+            # (likely by another cylc client, no problem, safe to ignore)
+            pass
         except OSError as exc:
-            raise CylcError(
-                f'Failed to remove old contact file: "{fname}"\n{exc}'
+            # unexpected error removing the contact file
+            # (note the FileNotFoundError incorporated errno.ENOENT)
+            LOG.error(
+                f'Failed to remove contact file for {reg}:\n{exc}'
+            )
+        else:
+            LOG.info(
+                f'Removed contact file for {reg}'
+                ' (workflow no longer running).'
             )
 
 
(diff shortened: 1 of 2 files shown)


datamel (approved):
Thansk @oliver-sanders, this will fix it! Code read, tests run.
The failing macos is unrelated so approving.

cylc/flow/workflow_files.py
```diff
@@ -512,17 +512,12 @@ def detect_old_contact_file(reg: str, contact_data=None) -> None:
...
             LOG.info(
                 f'Removing contact file for {reg}'
                 ' (workflow no longer running).'
             )
             os.unlink(fname)
-            return
-        except OSError as exc:
-            raise CylcError(
-                f'Failed to remove old contact file: "{fname}"\n{exc}'
-            )
```
MetRonnie, line 525: Do we not want to keep a warning at least about this? What if there is an `OSError` due to a reason that isn't the file no longer existing?

oliver-sanders: This will most likely be a non-interactive command so there isn't much benefit to raising a warning. We could check if the contact file still exists just before removal and allow the OS error to be raised otherwise?

MetRonnie: I meant a `LOG.warning()`

Btw, just noticed the docstring will be out of date with this change

oliver-sanders: The contact file is loaded at the start of this function, if it is removed before we try to remove it at the end of the function that is of no consequence so I don't think there's a good reason to log this. Also, this will most likely be a non-interactive command.

Have updated the docstring.

tests/integration/test_workflow_files.py
```diff
@@ -160,3 +160,39 @@ def test_detect_old_contact_file_none(workflow):
...
+    monkeypatch,
+    caplog,
+    log_filter,
+):
+    """It should not error if the contact file is removed by something else.
+
+    E.G:
+    * detect_old_contact_file in another client.
+    * cylc clean.
+    * Aliens.
```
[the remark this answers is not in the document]
oliver-sanders: https://xkcd.com/2259/

cylc/flow/workflow_files.py
MetRonnie: > if it is removed before we try to remove it at the end of the function that is of no consequence so I don't think there's a good reason to log this

But I'm talking about an `OSError` that is not an instance of `FileNotFoundError`, e.g. what if some process still has the contact file open?

Just want to check we're happy that's not something to worry about, because if it were to happen, the logging will confusingly say the contact file was removed when it wasn't.

oliver-sanders: Ok, I've change it so that the `removing contact file` message is now `removed contact file` and only shown on success AND logged any non `FileNotFoundError` instance of `OSError` (this log will probably go to a random `job.err` file, but heh, why not).

MetRonnie (approved):

datamel (approved):
Tested, working as expected, thanks @oliver-sanders.

Issue #4687, opened 2022-02-14
reflow: tasks with incomplete output should rerun


oliver-sanders:
### Problem

* Reflow allows us to rerun an entire subgraph of a workflow.
* When the new [re-]flow catches up with an active task from another flow it merges with it.
* Because we keep tasks with incomplete outputs in the pool for practical purposes, tasks with incomplete outputs are *not* re-run (as a result of current merging logic) whereas tasks with complete output are.
* This prevents a reflow from continuing in a situation where it should have been able to.
* This creates a confusing disparity between reflow behaviour with optional vs expected outputs.

### Proposal

When a reflow catches a task with incomplete outputs, the task should be reset to waiting in order to allow the task to re-run and the new flow to continue.

### Example

```ini
[scheduling]
    initial cycle point = 1
    cycling mode = integer
    [[graph]]
        P1 = """
            a => b => c
            b[-P1] => b
        """

[runtime]
    [[a, c]]
        script = sleep 1
    [[b]]
        script = """
            if [[ $CYLC_TASK_SUBMIT_NUMBER == 1 && $CYLC_TASK_CYCLE_POINT == 2 ]]; then
                exit 1
            fi
        """
```

This workflow will stall after the task `2/b` fails.

```console
$ cylc play <id>
...
WARNING - Incomplete tasks:
      * 2/b did not complete required outputs: ['succeeded']
WARNING - Unsatisfied prerequisites:
      * 3/b is waiting on ['2/b:succeeded']
      * 4/b is waiting on ['3/b:succeeded']
CRITICAL - Workflow stalled
```

If I reflow-trigger `1/b`:

```console
$ cylc trigger --reflow <id>//1/b
```

Then I would expect its downstream tasks `1/c` and `2/b` to be re-run in the new flow, however:

* `1/b` *is* re-run.
* `2/b` *is not* re-run because it has incomplete outputs.

If the graph were changed like so:

```
            a => b? => c
            b[-P1]? => b
```

The behaviour would be as expected.

PR #4737, opened 2022-03-09
task pool: fix flow behaviour with incomplete outputs

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/4687.
* Tasks that have incomplete outputs are held in the n=0 pool.
* This means as new flows approach them they merge but are not re-run.
* This change resets such tasks to waiting and re-queues them to allow them to re-run in the same way a task with complete outputs would.

**Requirements check-list**

### DIFF
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -68,7 +68,7 @@ if TYPE_CHECKING:
     from cylc.flow.taskdef import TaskDef
     from cylc.flow.task_events_mgr import TaskEventsManager
     from cylc.flow.workflow_db_mgr import WorkflowDatabaseManager
-    from cylc.flow.flow_mgr import FlowMgr
+    from cylc.flow.flow_mgr import FlowMgr, FlowNums
 
 Pool = Dict['PointBase', Dict[str, TaskProxy]]
 
@@ -581,11 +581,7 @@ class TaskPool:
                     not next_task.flow_nums
                     and not next_task.state.is_runahead
                 )
-                next_task.merge_flows(itask.flow_nums)
-                LOG.info(
-                    f"[{next_task}] merged in flow(s) "
-                    f"{','.join(str(f) for f in itask.flow_nums)}"
-                )
+                self.merge_flows(next_task, itask.flow_nums)
                 if retroactive_spawn:
                     # Did not belong to a flow (force-triggered) before merge.
                     # Now it does, so spawn successor, and children if needed.
@@ -1173,15 +1169,7 @@ class TaskPool:
             )
             if c_task is not None:
                 # Child already exists, update it.
-                if not c_task.flow_nums:
-                    # Child does not belong to a flow (force-triggered). Now
-                    # (merging) it does, so spawn outputs completed so far.
-                    self.spawn_on_all_outputs(c_task, completed_only=True)
-                c_task.merge_flows(itask.flow_nums)
-                LOG.info(
-                    f"[{c_task}] merged in flow(s) "
-                    f"{','.join(str(f) for f in itask.flow_nums)}"
-                )
+                self.merge_flows(c_task, itask.flow_nums)
                 self.workflow_db_mgr.put_insert_task_states(
                     c_task,
                     {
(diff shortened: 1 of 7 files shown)


hjoliver:
On reflection I'm not 100% sure about this.

Incomplete tasks are retained in `n=0` in the expectation of manual retriggering at some point, to complete their expected outputs and continue their own flow.

They have by definition not been fixed and retriggered yet (otherwise they would not still be incomplete) so we should not really expect them to run successfully if triggered automatically in another flow.

So, on that basis we should expect an incomplete task to block other flows as well as its own, and wait for manual retriggering to allow the merged flow to carry on

Caveats:
  - If different flows are processing different data through the graph, it's not necessarily true that an incomplete task will be broken in all flows, but we won't be recommending that sort of use case until/if flows can be entirely independent (i.e. no merging).
  - It won't break anything to re-run a task that is just destined to fail again, but it's still unnecessary use of compute resource

oliver-sanders:
I don't think the outputs the task has or has not generated should affect subsequent flows. We don't distinguish between waiting and finished tasks so why distinguish between complete and incomplete ones?

Seems a bit odd for the flow front to just stop because it met a finished task (which could be a succeeded task that didn't produce a required custom output).

The way we hold incomplete tasks in the n=0 window is an implementation detail we added to increase the visibility of these tasks in the UIs and to assist with stall detection/reporting not originally present in the original "pure" SoD implementation.

> I suppose there is a chance that an upstream change, for the new flow, will "fix" the incomplete task by changing its input data, but ... rather unlikely??

I think this is one of, if not the main use case for reflow?

E.G:

"My data got poisoned at some point upstream resulting in a failure downstream, now I need to fix the problem in an upstream task and re-run the sub-graph".

(I've seen this use case arise due to a faulty node in an archive system which caused data corruption resulting in a few workflows ending up in this position).

Or:

"I didn't configure my model to output the right fields resulting in a failure downstream ...".

Or:

"I'm writing a workflow incrementally in a write-reload-run loop trying to get downstream tasks succeeding".

hjoliver:
Fair enough, good argument. 

The majority of incomplete tasks will be failed tasks that did not succeed for internal reasons, and so will fail again in any flow. But upstream reasons are possible, so my 2nd caveat above justifies this even if it's not super likely.

Not sure I've ever seen your use cases in the wild (it's usually about getting the next task running successfully by understanding and tweaking its own config) but I like them in principle.

Approving...

cylc/flow/task_pool.py
```diff
@@ -1776,3 +1764,34 @@ def match_taskdefs(
                     n_warnings += 1
                     continue
         return n_warnings, task_items
+
+    def merge_flows(self, itask: TaskProxy, flow_nums: 'FlowNums') -> None:
+        """Merge itask.flow_nums with flow_nums.
+
+        This also performs requied spawning / state changing for edge cases.
```
hjoliver, line 1771:
```suggestion
        This also performs required spawning / state changing for edge cases.
```

tests/functional/spawn-on-demand/15-reflow-incomplete-outputs/flow.cylc
```diff
@@ -0,0 +1,20 @@
...
+        R1 = """
+            a => b => c
+        """
+
+[runtime]
+    [[b]]
+        script = """
+            # test $CYLC_TASK_SUBMIT_NUMBER -gt 1
+            if [[ $CYLC_TASK_SUBMIT_NUMBER -eq 1 ]]; then
+                cylc trigger --reflow "$CYLC_WORKFLOW_NAME//1/a"
```
hjoliver, line 16: This will need tweaking after #4738

hjoliver (approved):

datamel (approved):
Manually tested (on master to recreate and on this branch) with the example in the issue. Fixes the issue. Read the code (I appreciate the code explanations) and have run the relevant tests locally with no problems.

oliver-sanders:
Manually pytest'ed locally to get around the bugbear failures, all good.

PR #4744, opened 2022-03-18
cycle-point: add note about isodatetime

oliver-sanders:
I think we should probably advertise the `isodatetime` command more, particularly for `isodatetime R/P1Y/2020` like uses:

https://github.com/cylc/cylc-doc/issues/428

The caveat being that we need to add something about creating a wrapper script for `isodatetime` into the installation docs:

https://github.com/cylc/cylc-doc/issues/429

### DIFF
--- a/cylc/flow/scripts/cycle_point.py
+++ b/cylc/flow/scripts/cycle_point.py
@@ -19,7 +19,9 @@
 
 """cylc cycle-point [OPTIONS] ARGS
 
-Utility for date-time cycle point arithmetic.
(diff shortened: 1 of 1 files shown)


hjoliver (approved):
Fine.

Does `cylc cyclepoint` do anything `isodatetime` doesn't? (apart from, presumably, know about the `CYLC_TASK_CYCLE_POINT` variable)?

oliver-sanders:
> Does cylc cyclepoint do anything isodatetime doesn't

Don't think so.

PR #4745, opened 2022-03-18
Handle missing optional dependencies better

### LINKED ISSUE #4719: setup: disable uninstalled components
Sadly placing optional dependency groups in square brackets doesn't actually disable them:

> This syntax indicates that the entry point (in this case a console script) is only valid when the PDF extra is installed. It is up to the installer to determine how to handle the situation where PDF was not indicated 
>
> -- [setuptools](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html?highlight=optional%20dependencies#optional-dependencies)

So this:

```ini
[options.entry_points]
# top level shell commands
console_scripts =
     rose = metomi.rose.rose:rose
     rosie = metomi.rose.rose:rosie
     rosa = metomi.rose.rose:rosa [rosa]
```

Isn't enough to disable `rosa` which is annoying since it seems to suggest it would.

Sadly I don't think there is a way to tell which optional dependencies were installed, I don't think pip works like that. So we need protect places where we call `entry_point.load` against {module,distribution}-not-found errors otherwise they can leak into the system (e.g. try `cylc help all` without jupyterhub installed). If Cylc is installed and working it's reasonable to assume that a {module,distribution}-not-found error in an optional part of the system is a result of the optional dependency not being installed.

See also:

* [cylc-uiserver](https://github.com/cylc/cylc-uiserver/issues/325)
* [rose](https://github.com/metomi/rose/issues/2527)

**Pull requests welcome!**

MetRonnie:
These changes close #4719 - when optional dependencies are not installed, there is traceback when trying to run `cylc help all` or a command that uses the optional dependency.

Steps to reproduce bug (before checking out this branch):
1. Create a clean conda env
2. `pip install <cylc-flow-dev-location> <cylc-uiserver-dev-location>`
3. `cylc help all`
4. `cylc hubapp`

The last two should give traceback.

Now check out this branch and test the fix:
1. `cylc help all`
2. `cylc hubapp`

**Requirements check-list**
- [ ] Does not need tests (can't).

### DIFF
--- a/cylc/flow/scripts/cylc.py
+++ b/cylc/flow/scripts/cylc.py
@@ -21,6 +21,7 @@ from contextlib import contextmanager
 import os
 from pathlib import Path
 import sys
+from typing import Iterator, NoReturn, Optional, Tuple
 
 from ansimarkup import parse as cparse
 from colorama import init as color_init
@@ -182,8 +183,7 @@ USAGE = cparse(USAGE)
 # {name: entry_point}
 COMMANDS: dict = {
     entry_point.name: entry_point
-    for entry_point
-    in iter_entry_points('cylc.command')
+    for entry_point in iter_entry_points('cylc.command')
 }
 
 
@@ -272,17 +272,21 @@ DEAD_ENDS = {
 # fmt: on
 
 
-def execute_cmd(cmd, *args):
+def execute_cmd(cmd: str, *args: str) -> NoReturn:
     """Execute a sub-command.
 
     Args:
-        cmd (str):
-            The name of the command.
-        args (list):
-            List of command line arguments to pass to that command.
+        cmd: The name of the command.
+        args: Command line arguments to pass to that command.
 
     """
-    COMMANDS[cmd].resolve()(*args)
+    entry_point: pkg_resources.EntryPoint = COMMANDS[cmd]
+    try:
+        entry_point.resolve()(*args)
+    except ModuleNotFoundError as exc:
+        msg = handle_missing_dependency(entry_point, exc)
+        print(msg, file=sys.stderr)
+        sys.exit(1)
     sys.exit()
 
 
@@ -366,15 +370,19 @@ def parse_docstring(docstring):
     return (usage, desc)
 
 
-def iter_commands():
+def iter_commands() -> Iterator[Tuple[str, Optional[str], Optional[str]]]:
     """Yield all Cylc sub-commands.
 
     Yields:
-        tuple - (command, description, usage)
+        (command, description, usage)
 
     """
-    for cmd, obj in sorted(COMMANDS.items()):
-        module = __import__(obj.module_name, fromlist=[''])
+    for cmd, entry_point in sorted(COMMANDS.items()):
+        try:
+            module = __import__(entry_point.module_name, fromlist=[''])
+        except ModuleNotFoundError as exc:
+            handle_missing_dependency(entry_point, exc)
+            continue
         if getattr(module, 'INTERNAL', False):
             # do not list internal commands
             continue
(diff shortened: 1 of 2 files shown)


cylc/flow/scripts/cylc.py
```diff
@@ -663,3 +671,30 @@ def main():
...
+    # should have an attr 'extras'
+    if not entry_point.extras:
+        raise exc
+
+    msg = f"cylc {entry_point.name}: missing optional dependency '{exc.name}'"
+    if entry_point.dist:
+        msg += "\nInstall using \"pip install '{0}[{1}]'\"".format(
+            entry_point.dist.project_name,
+            ','.join(entry_point.extras)
+        )
```
MetRonnie, line 699: Wondering if this is OTT

cylc/flow/scripts/cylc.py
```diff
@@ -663,3 +671,30 @@ def main():
...
+    Args:
+        entry_point: The entry point that was attempted to load but caused
+            a ModuleNotFoundError.
+        exc: The ModuleNotFoundError that was caught.
+
+    Re-raises the given ModuleNotFoundError if it is unexpected.
+    """
+    # If the missing module is an optional dependency, the EntryPoint
+    # should have an attr 'extras'
+    if not entry_point.extras:
```
oliver-sanders, line 691: Oh nice!

cylc/flow/scripts/cylc.py
oliver-sanders: I think it might be, somewhat confusing when installed via Conda, could give them a link to the installation section of the docs (although at present it doesn't really cover this).

cylc/flow/scripts/cylc.py
```diff
@@ -663,3 +671,30 @@ def main():
                 if opts.version:
                     cmd_args.append("--version")
                 execute_cmd(command, *cmd_args)
+
+
+def handle_missing_dependency(
```
wxtim, line 676: You could probably unit test this bit fairly trivially, which would at least leave the untested surface smaller.

wxtim (changes requested):
I'm not sure that you can feasibly functional test this with current infrastructure, but a unit test would be nice.

cylc/flow/scripts/cylc.py
oliver-sanders: (This can happen for both Pip & Conda installs)

cylc/flow/scripts/cylc.py
oliver-sanders: Can patch `iter_entry_points` to return something with a `resolve` method that raises `ModuleNotFoundError`.

cylc/flow/scripts/cylc.py
MetRonnie: - Created issue for documenting pip extras: https://github.com/cylc/cylc-doc/issues/430
- For Conda installs, wouldn't it only happen if they installed the `-base` recipe? E.g. `cylc-uiserver-base` instead of `cylc-uiserver`?

oliver-sanders: > For Conda installs, wouldn't it only happen if they installed the -base recipe?

Yes.

MetRonnie:
Writing a test for this is proving to be non-trivial (the mocking is getting nasty). Unless someone has a test they would like to contribute, I propose leaving this one for manual testing

wxtim:
> Writing a test for `execute_cmd()` or `iter_commands()` is proving to be non-trivial (the mocking is getting nasty). Unless someone has a test they would like to contribute, I propose leaving this one for manual testing. (Not sure it's worth writing a unit test for `handle_missing_dependency()` as you'd basically be testing Python's `try...except...else` logic?)

Understood. Can you raise an issue for the testing (8.x or someday milestone?)? Even if we never get around to testing it then at least it's recorded somewhere.

wxtim (approved):
Happy with changes made since last review and to record absence of tests because testing is awkward in the extreme.

oliver-sanders:
https://github.com/MetRonnie/cylc-flow/pull/12

oliver-sanders (approved):

PR #4746, opened 2022-03-21
Flake8 bugbear fixes

hjoliver:

With flake8-bugbear 22.3.20:
```console
$ flake8 cylc/flow
cylc/flow/task_pool.py:247:13: 
  B020 Found for loop that reassigns the iterable it is iterating with each iterable value.
cylc/flow/task_job_mgr.py:945:17:
  B020 Found for loop that reassigns the iterable it is iterating with each iterable value.
cylc/flow/cycling/iso8601.py:394:6:
  B019 Use of `functools.lru_cache` or `functools.cache` on class methods can lead to memory leaks. 
  The cache may retain instance references, preventing garbage collection.
cylc/flow/main_loop/__init__.py:341:13:
  B020 Found for loop that reassigns the iterable it is iterating with each iterable value.
cylc/flow/main_loop/__init__.py:341:24:
  B020 Found for loop that reassigns the iterable it is iterating with each iterable value.
```

**Requirements check-list**

### DIFF
--- a/cylc/flow/cycling/iso8601.py
+++ b/cylc/flow/cycling/iso8601.py
@@ -391,7 +391,7 @@ class ISO8601Sequence(SequenceBase):
         if self.exclusions:
             self.value += '!' + str(self.exclusions)
 
-    @lru_cache(100)
+    @lru_cache(100)  # noqa B019 TODO?
     def is_on_sequence(self, point):
         """Return True if point is on-sequence."""
         # Iterate starting at recent valid points, for speed.
--- a/cylc/flow/main_loop/__init__.py
+++ b/cylc/flow/main_loop/__init__.py
@@ -339,10 +339,10 @@ def load(config, additional_plugins=None):
         # load coroutines
         log = []
         for coro_name, coro in (
-                (coro_name, coro)
-                for coro_name, coro in getmembers(module)
-                if isfunction(coro)
-                if hasattr(coro, 'main_loop')
+                (corou_name, corou)
+                for corou_name, corou in getmembers(module)
+                if isfunction(corou)
+                if hasattr(corou, 'main_loop')
         ):
             log.append(coro_name)
             plugins.setdefault(
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -942,7 +942,7 @@ class TaskJobManager:
                 cmd = construct_ssh_cmd(
                     cmd, platform, host
                 )
-            for itask in sorted(itasks, key=lambda itask: itask.identity):
+            for itask in sorted(itasks, key=lambda jtask: jtask.identity):
                 job_log_dirs.append(
                     itask.tokens.duplicate(
                         job=str(itask.submit_num)
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -245,15 +245,15 @@ class TaskPool:
         # At restart all tasks are runahead-limited but finished and manually
         # triggered tasks (incl. --start-task's) can be released immediately.
         for itask in (
-            itask
-            for itask in self.get_tasks()
-            if itask.state.is_runahead
-            if itask.state(
+            jtask
+            for jtask in self.get_tasks()
+            if jtask.state.is_runahead
+            if jtask.state(
                 TASK_STATUS_FAILED,
                 TASK_STATUS_SUCCEEDED,
                 TASK_STATUS_EXPIRED
             )
-            or itask.is_manual_submit
+            or jtask.is_manual_submit
         ):
             self.release_runahead_task(itask)
             released = True


cylc/flow/cycling/iso8601.py
```diff
@@ -391,7 +391,7 @@ def set_offset(self, i_offset):
         if self.exclusions:
             self.value += '!' + str(self.exclusions)
 
-    @lru_cache(100)
+    @lru_cache(100)  # noqa B019 TODO?
```
hjoliver, line 394: Need to think about this.

Note recent official Python docs advice:

[How do I cache method calls](https://docs.python.org/3/faq/programming.html#how-do-i-cache-method-calls)

wxtim: I'm happy to approve this for now, but I think that we should create an issue to investigate the context and decide which caching approach is appropriate.

AFAI can tell this advice has been added to the FAQ at 3.10, and the `functools.cached_property` was [added at 3.8](https://docs.python.org/3.8/whatsnew/3.8.html#functools) so it's not really available to us unless we manually back-port the [class](https://github.com/python/cpython/blob/7ad52d174a12951d260aecfcdfe0fcfd2437a66b/Lib/functools.py#L966) in our code base. 🤮

cylc/flow/main_loop/__init__.py
```diff
@@ -339,10 +339,10 @@ def load(config, additional_plugins=None):
         # load coroutines
         log = []
         for coro_name, coro in (
-                (coro_name, coro)
-                for coro_name, coro in getmembers(module)
-                if isfunction(coro)
-                if hasattr(coro, 'main_loop')
+                (corou_name, corou)
```
wxtim, line 342: I think the similarity of the two names could be tricky to read. Why not `coroutine_name` and `coroutine`. It's a mouthful, but only exists on four lines.

Also at the start of a loop like this it's a little like the written English language convention of "Three Letter Acronym [TLA]" Use TLA in rest of document".

cylc/flow/task_job_mgr.py
```diff
@@ -942,7 +942,7 @@ def _run_job_cmd(
                 cmd = construct_ssh_cmd(
                     cmd, platform, host
                 )
-            for itask in sorted(itasks, key=lambda itask: itask.identity):
+            for itask in sorted(itasks, key=lambda jtask: jtask.identity):
```
wxtim, line 945: Same as previous point: Although I think most of us learn that `itask` is our convention for `task_proxy` I think it would be more helpful to call it `task_proxy` than `jtask`. Unlike the previous version I think that it might require a newline, so I will be slightly less keen on the change.

wxtim (approved):
I've made some comments/suggestions about language, but it's not something which should block merger of PR, so it's an approval from me.

cylc/flow/main_loop/__init__.py
```diff
@@ -339,10 +339,10 @@ def load(config, additional_plugins=None):
...
+                for corou_name, corou in getmembers(module)
+                if isfunction(corou)
+                if hasattr(corou, 'main_loop')
```
MetRonnie, line 345: I think a cleaner alternative (also avoiding iterating twice) would be
```diff
diff --git a/cylc/flow/main_loop/__init__.py b/cylc/flow/main_loop/__init__.py 
index 8c2035506..4bb2a2417 100644
--- a/cylc/flow/main_loop/__init__.py
+++ b/cylc/flow/main_loop/__init__.py
@@ -338,17 +338,13 @@ def load(config, additional_plugins=None):
             raise CylcError(f'Could not load plugin: "{plugin_name}"')        
         # load coroutines
         log = []
-        for coro_name, coro in (
-                (coro_name, coro)
-                for coro_name, coro in getmembers(module)
-                if isfunction(coro)
-                if hasattr(coro, 'main_loop')
-        ):
-            log.append(coro_name)
-            plugins.setdefault(
-                coro.main_loop, {}
-            )[(plugin_name, coro_name)] = coro
-            plugins['timings'][(plugin_name, coro_name)] = deque(maxlen=1)    
+        for coro_name, coro in getmembers(module):
+            if isfunction(coro) and hasattr(coro, 'main_loop'):
+                log.append(coro_name)
+                plugins.setdefault(
+                    coro.main_loop, {}
+                )[(plugin_name, coro_name)] = coro
+                plugins['timings'][(plugin_name, coro_name)] = deque(maxlen=1)
         LOG.debug(
             'Loaded main loop plugin "%s":\n%s',
             plugin_name,
```

cylc/flow/task_job_mgr.py
MetRonnie: IMHO for a lambda function it would be ok to use a single letter variable, it should be relatively obvious what it's referring to in this case? (And I find `jtask` a bit weird)
```suggestion
            for itask in sorted(itasks, key=lambda i: i.identity):
```

cylc/flow/cycling/iso8601.py
oliver-sanders: I don't think this usage is an issue for us as on the whole Cylc doesn't make a habit of releasing `ISO8601Sequence` instances (they are used by taskdefs which live for the life of the config).

The `cached_property` thing can't handle arguments so doesn't help, the issue is that the cache lives for the life of the class/global scope rather than the life of the instance, it looks like there is an easy way to solve that.

cylc/flow/task_pool.py
```diff
@@ -245,15 +245,15 @@ def release_runahead_tasks(self):
...
+            if jtask.state(
                 TASK_STATUS_FAILED,
                 TASK_STATUS_SUCCEEDED,
                 TASK_STATUS_EXPIRED
             )
-            or itask.is_manual_submit
+            or jtask.is_manual_submit
         ):
             self.release_runahead_task(itask)
             released = True
```
MetRonnie, line 259: As above
```diff
diff --git a/cylc/flow/task_pool.py b/cylc/flow/task_pool.py
index 20d2cd98b..f180e0283 100644
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -244,19 +244,17 @@ class TaskPool:
 
         # At restart all tasks are runahead-limited but finished and manually
         # triggered tasks (incl. --start-task's) can be released immediately.
-        for itask in (
-            itask
-            for itask in self.get_tasks()
-            if itask.state.is_runahead
-            if itask.state(
-                TASK_STATUS_FAILED,
-                TASK_STATUS_SUCCEEDED,
-                TASK_STATUS_EXPIRED
-            )
-            or itask.is_manual_submit
-        ):
-            self.release_runahead_task(itask)
-            released = True
+        for itask in self.get_tasks():
+            if itask.state.is_runahead and (
+                itask.state(
+                    TASK_STATUS_FAILED,
+                    TASK_STATUS_SUCCEEDED,
+                    TASK_STATUS_EXPIRED
+                )
+                or itask.is_manual_submit
+            ):
+                self.release_runahead_task(itask)
+                released = True
 
         runahead_limit_point = self.compute_runahead()
         if not runahead_limit_point:
```

MetRonnie (changes requested):

cylc/flow/cycling/iso8601.py
oliver-sanders: > The lru_cache approach can be made to work, but the class needs to define the __eq__ and __hash__ methods so the cache can detect relevant attribute updates:

https://github.com/cylc/cylc-flow/blob/85694062b09d10e64541f76835c5507f1c8a0895/cylc/flow/cycling/iso8601.py#L589-L604

cylc/flow/task_pool.py
MetRonnie: Oh, unless the same thing as written below applies?  https://github.com/cylc/cylc-flow/blob/4ca97df6b05e3670dae42ca4d5e680912d0cfdbc/cylc/flow/task_pool.py#L265-L269

in which case using an intermediate list would be fine (also clearer than the current code)

cylc/flow/task_job_mgr.py
wxtim: >  (And I find jtask a bit weird)

Who let the electrical engineers in?

cylc/flow/cycling/iso8601.py
MetRonnie: (Off topic but it looks to me like the `__hash__` method should `return hash((self.TYPE, self.value))` based on the `__eq__`

oliver-sanders: Got a fix on https://github.com/cylc/cylc-flow/pull/4747

oliver-sanders: > Off topic but it looks to me

I don't think it need to because `self.TYPE` is a class variable and the cache "belongs" to the class so `self.TYPE` should not differ.

The reason that the `__eq__` method does this is to allow comparisons with other sequence types.

PR #4747, opened 2022-03-21
flake8: fix B019 issue

oliver-sanders:
Fix valid B019 bug flagged by flake8-bugbear.

This bug will cause a memory leak with `ISO8601Sequence` instances. As it happens Cylc does not release these instances anyway (since they are required by the taskdefs which must live for the life of the Scheduler) so this memory leak is of no consequence at present.

**Requirements check-list**
- [x] No change log entry required (inconsequential at present).

### DIFF
--- a/cylc/flow/cycling/__init__.py
+++ b/cylc/flow/cycling/__init__.py
@@ -365,6 +365,8 @@ class SequenceBase(metaclass=ABCMeta):
         """Deprecated: alter state to offset the entire sequence."""
         pass
 
+    # NOTE: not using @abstractmethod because we need to
+    # patch this method for caching purposes
     def is_on_sequence(self, point):
         """Is point on-sequence, disregarding bounds?"""
         raise NotImplementedError
--- a/cylc/flow/cycling/__init__.py
+++ b/cylc/flow/cycling/__init__.py
@@ -365,10 +365,9 @@ class SequenceBase(metaclass=ABCMeta):
         """Deprecated: alter state to offset the entire sequence."""
         pass
 
-    @abstractmethod
     def is_on_sequence(self, point):
         """Is point on-sequence, disregarding bounds?"""
-        pass
+        raise NotImplementedError
 
     @abstractmethod
     def is_valid(self, point):
--- a/cylc/flow/cycling/iso8601.py
+++ b/cylc/flow/cycling/iso8601.py
@@ -292,7 +292,7 @@ class ISO8601Sequence(SequenceBase):
                  'offset', '_cached_first_point_values',
                  '_cached_next_point_values', '_cached_valid_point_booleans',
                  '_cached_recent_valid_points', 'spec', 'abbrev_util',
-                 'recurrence', 'exclusions', 'step', 'value')
+                 'recurrence', 'exclusions', 'step', 'value', 'is_on_sequence')
 
     @classmethod
     def get_async_expr(cls, start_point=None):
@@ -306,6 +306,10 @@ class ISO8601Sequence(SequenceBase):
         SequenceBase.__init__(
             self, dep_section, context_start_point, context_end_point)
         self.dep_section = dep_section
+
+        # cache is_on_sequence
+        # see B019 - https://github.com/PyCQA/flake8-bugbear#list-of-warnings
+        self.is_on_sequence = lru_cache(maxsize=100)(self._is_on_sequence)
 
         if (
             context_start_point is None
@@ -391,8 +395,8 @@ class ISO8601Sequence(SequenceBase):
         if self.exclusions:
             self.value += '!' + str(self.exclusions)
 
-    @lru_cache(100)
-    def is_on_sequence(self, point):
+    # lru_cache'd see __init__()
+    def _is_on_sequence(self, point):
         """Return True if point is on-sequence."""
         # Iterate starting at recent valid points, for speed.
         if self.exclusions and point in self.exclusions:


oliver-sanders:
To understand the issue, here are three examples of classes using `lru_cache`. In the `test()` method we create an instance of each class and get it to cache one result. Then we unset the sole pointer to the instance which should trigger garbage collection of the instance since there is no reason for it to hang around any more. As a failsafe I've explicitly called `gc.collect()` too. You would expect the instance to be collected and its `__del__` method to be called, however, this does not happen for the first two cases, only the third:

```python
from gc import collect                                                               
from functools import lru_cache
 
 
class A:
 
    def __init__(self, y):
        self.y = 1
 
    @lru_cache(maxsize=None)
    def z(self, x):
        return self.y * x
 
    def __del__(self):
        print('del(A)')
 
 
class B:
 
    def __init__(self, y):
        self.y = 1
 
    @lru_cache(maxsize=None)
    def z(self, x):
        return self.y * x
 
    def __hash__(self):
        return hash(self.y)
 
    def __eq__(self, other):
        return self.y == other.y
 
    def __del__(self):
        print('del(B)')
 
 
class C:
 
    def __init__(self, y):
        self.y = 1
        self.z = lru_cache(maxsize=None)(self._z)
 
    def _z(self, x):
        return self.y * x
 
    def __hash__(self):
        return hash(self.y)
 
    def __eq__(self, other):
        return self.y == other.y
 
    def __del__(self):
        print('del(C)')


def test():
    print('test A')
    a = A(2)
    a.z(3)
    a = None
    print(collect())
 
    print('test B')
    b = B(2)
    b.z(3)
    b = None
    print(collect())
 
    print('test C')
    c = C(2)
    c.z(3)
    c = None
    print(collect())
 
    print('DONE')
 
 
if __name__ == '__main__':
    test() 
```

Why? Because for cases A & B the cache belongs to the class whereas for C the cache belongs to the instance.

cylc/flow/cycling/__init__.py
```diff
@@ -365,10 +365,9 @@ def set_offset(self, i_offset):
         """Deprecated: alter state to offset the entire sequence."""
         pass
 
-    @abstractmethod
     def is_on_sequence(self, point):
         """Is point on-sequence, disregarding bounds?"""
-        pass
+        raise NotImplementedError
```
MetRonnie, line 370: Why? I thought `@asbtractmethod` is the correct thing to do?

oliver-sanders: "Correct" is subjective, both are valid approaches, asbtractmethod is more OO, NotImplementedError is more functional. Python supports both paradigms but they can clash.

OO approaches are often less flexible, in this case the `asbtractmethod` mandated that this method must exist in the child class. The solution on this PR uses the flexibility of Python's functional side to add the method at `__init__` time, OO just doesn't like this.

MetRonnie (commented):
Alternative?
```diff
diff --git a/cylc/flow/cycling/__init__.py b/cylc/flow/cycling/__init__.py
index 6a07b42ba..5c1ec26d2 100644
--- a/cylc/flow/cycling/__init__.py
+++ b/cylc/flow/cycling/__init__.py
@@ -365,9 +365,10 @@ class SequenceBase(metaclass=ABCMeta):
         """Deprecated: alter state to offset the entire sequence."""
         pass
 
+    @abstractmethod
     def is_on_sequence(self, point):
         """Is point on-sequence, disregarding bounds?"""
-        raise NotImplementedError
+        pass
 
     @abstractmethod
     def is_valid(self, point):
diff --git a/cylc/flow/cycling/iso8601.py b/cylc/flow/cycling/iso8601.py
index 2dff5a95b..9dc87b72a 100644
--- a/cylc/flow/cycling/iso8601.py
+++ b/cylc/flow/cycling/iso8601.py
@@ -292,7 +292,7 @@ class ISO8601Sequence(SequenceBase):
                  'offset', '_cached_first_point_values',
                  '_cached_next_point_values', '_cached_valid_point_booleans',
                  '_cached_recent_valid_points', 'spec', 'abbrev_util',
-                 'recurrence', 'exclusions', 'step', 'value', 'is_on_sequence')
+                 'recurrence', 'exclusions', 'step', 'value')
 
     @classmethod
     def get_async_expr(cls, start_point=None):
@@ -307,10 +307,6 @@ class ISO8601Sequence(SequenceBase):
             self, dep_section, context_start_point, context_end_point)
         self.dep_section = dep_section
 
-        # cache is_on_sequence
-        # see B019 - https://github.com/PyCQA/flake8-bugbear#list-of-warnings
-        self.is_on_sequence = lru_cache(maxsize=100)(self._is_on_sequence)
-
         if (
             context_start_point is None
             or isinstance(context_start_point, ISO8601Point)
@@ -395,26 +391,30 @@ class ISO8601Sequence(SequenceBase):
         if self.exclusions:
             self.value += '!' + str(self.exclusions)
 
-    # lru_cache'd see __init__()
-    def _is_on_sequence(self, point):
+    def is_on_sequence(self, point):
         """Return True if point is on-sequence."""
+        return self._is_on_sequence(self, point)
+
+    @staticmethod
+    @lru_cache(100)
+    def _is_on_sequence(instance: 'ISO8601Sequence', point):
+        # see B019 - https://github.com/PyCQA/flake8-bugbear#list-of-warnings
         # Iterate starting at recent valid points, for speed.
-        if self.exclusions and point in self.exclusions:
+        if instance.exclusions and point in instance.exclusions:
             return False
 
-        for valid_point in reversed(self._cached_recent_valid_points):
+        for valid_point in reversed(instance._cached_recent_valid_points):
             if valid_point == point:
                 return True
             if valid_point > point:
                 continue
             next_point = valid_point
             while next_point is not None and next_point < point:
-                next_point = self.get_next_point_on_sequence(next_point)
+                next_point = instance.get_next_point_on_sequence(next_point)
             if next_point is None:
                 continue
             if next_point == point:
                 return True
-        return self.recurrence.get_is_valid(point_parse(point.value))
+        return instance.recurrence.get_is_valid(point_parse(point.value))
 
     def is_valid(self, point):
         """Return True if point is on-sequence and in-bounds."""
```

oliver-sanders:
> Alternative?

Nope, the cache still belongs to the class not the instance.

MetRonnie:
But we have an awful lot of static methods that are `lru_cache`d - does that mean they are not safe either?

oliver-sanders:
It does look like staticmethod is ok so I guess so. If you think it's worth the effort, please close this PR, open a new one and include an example like the one above to make sure the GC works.

MetRonnie:
Think I'm beginning to understand. The other static methods that are `lru_cache`d are ok because they don't take an instance of the class as an arg, only stuff like strings. But in this case it would suffer the same problem as originally, because it takes an instance of `ISO8601Sequence` as an arg, it will keep the reference to that instance. So only your fix would work, my `staticmethod` alternative is pointless

oliver-sanders:
Ah damn, I was coming around to thinking your approach might work.

oliver-sanders:
`lru_cache` should be using `weakref` under the hood rather than regular references. That way if an instance gets garbage collected the weak reference in the cache won't keep it alive. In which case passing an instance as an argument should be ok?

MetRonnie:
I had a go at testing it and the `del(D)` only came after `DONE`....

oliver-sanders:
Derp, ok, need option C then.

cylc/flow/cycling/__init__.py
MetRonnie: I think it's probably adding a comment about why we can't use `@abstractmethod`

MetRonnie (approved):
1 suggestion: https://github.com/cylc/cylc-flow/pull/4747/files#r833614530

cylc/flow/cycling/__init__.py
```diff
@@ -365,10 +365,9 @@ def set_offset(self, i_offset):
         """Deprecated: alter state to offset the entire sequence."""
         pass
 
-    @abstractmethod
     def is_on_sequence(self, point):
```
oliver-sanders, line 368:
```suggestion
    # NOTE: not using @abstractmethod because we need to
    # patch this method for caching purposes
    def is_on_sequence(self, point):
```

cylc/flow/cycling/__init__.py
oliver-sanders: Done - https://github.com/cylc/cylc-flow/pull/4747/files#r834249914

PR #4748, opened 2022-03-21
Make TUI refuse to show non-installed workflows

### LINKED ISSUE #4715: TUI: non-existent workflows are reported as stopped
```
$ cylc tui non-exist

non-exist - stopped
TUI is experimental and may break with large flows

  non-exist
```

**Expected behavior**
```
$ cylc tui non-exist

non-exist - workflow not found
TUI is experimental and may break with large flows

  non-exist
```

**Pull requests welcome!**

wxtim:
These changes close #4715 

### Summary

When Cylc TUI is started it should check whether the workflow has actually been installed before doing anything to display it.

### Requirements check-list
- [ ] Change log rebased after 8.0rc2 release. 

### DIFF
--- a/cylc/flow/tui/app.py
+++ b/cylc/flow/tui/app.py
@@ -21,6 +21,7 @@ import sys
 import urwid
 from urwid import html_fragment
 from urwid.wimp import SelectableIcon
+from pathlib import Path
 
 from cylc.flow.network.client_factory import get_client
 from cylc.flow.exceptions import (
@@ -28,6 +29,7 @@ from cylc.flow.exceptions import (
     ClientTimeout,
     WorkflowStopped
 )
+from cylc.flow.pathutil import get_workflow_run_dir
 from cylc.flow.task_state import (
     TASK_STATUSES_ORDERED,
     TASK_STATUS_SUBMITTED,
@@ -56,6 +58,7 @@ from cylc.flow.tui.util import (
     get_workflow_status_str,
     render_node
 )
+from cylc.flow.workflow_files import WorkflowFiles
 
 
 urwid.set_encoding('utf8')  # required for unicode task icons
@@ -310,10 +313,20 @@ class TuiApp:
                 }
             )
         except WorkflowStopped:
+            # Distinguish stopped flow from non-existent flow.
+            full_path = Path(get_workflow_run_dir(self.reg))
+            if(
+                (full_path / WorkflowFiles.SUITE_RC).exists()
+                or (full_path / WorkflowFiles.FLOW_FILE).exists()
+            ):
+                message = "stopped"
+            else:
+                message = "workflow not found"
+
             return dummy_flow({
                 'name': self.reg,
                 'id': self.reg,
-                'status': 'Workflow not found',
+                'status': message,
                 'stateTotals': {}
             })
         except (ClientError, ClientTimeout) as exc:
--- a/cylc/flow/tui/app.py
+++ b/cylc/flow/tui/app.py
@@ -321,7 +321,10 @@ class TuiApp:
             ):
                 message = "stopped"
             else:
-                message = "workflow not found"
+                message = (
+                    f"No {WorkflowFiles.SUITE_RC} or {WorkflowFiles.FLOW_FILE}"
+                    f"found in {self.reg}."
+                )
 
             return dummy_flow({
                 'name': self.reg,
--- a/cylc/flow/tui/app.py
+++ b/cylc/flow/tui/app.py
@@ -310,11 +310,10 @@ class TuiApp:
                 }
             )
         except WorkflowStopped:
-            self.client = None
             return dummy_flow({
                 'name': self.reg,
                 'id': self.reg,
-                'status': 'stopped',
+                'status': 'Workflow not found',
                 'stateTotals': {}
             })
         except (ClientError, ClientTimeout) as exc:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -33,6 +33,10 @@ ones in. -->
 
 ### Fixes:
 
+[#4748](https://github.com/cylc/cylc-flow/pull/4748) -
+`cylc tui` gives a more helpful error message when a workflow is not running;
+distinguishing between workflow not running and not in run-directory.
+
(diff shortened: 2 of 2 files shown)


cylc/flow/scripts/tui.py
```diff
@@ -89,6 +93,12 @@ def main(_, options: 'Values', workflow_id: str) -> None:
         workflow_id,
         constraint='workflows',
     )
+
+    # Throw out workflow_id's referring to not-installed workflows.
+    if not is_installed(get_workflow_run_dir(workflow_id)):
+        LOG.error(f'{workflow_id} is not an installed workflow')
+        exit(1)
```
oliver-sanders, line 100: This disables Tui for manually installed workflows, this needs to be solved in `cylc.flow.tui.app` (look for `WorkflowStopped`)

cylc/flow/tui/app.py
```diff
@@ -310,11 +310,10 @@ def get_snapshot(self):
                 }
             )
         except WorkflowStopped:
-            self.client = None
```
oliver-sanders, line 313: I think this is needed. It is what allows Tui to attempt to reconnect at a later time.

Try:

* install
* tui <flow>
* play <flow>
* stop <flow>
* play <flow>
* stop <flow>

Tui should connect/disconnect automatically.

Similarly Tui should connect automatically when the workflow is played in this example:

* tui <flow>
* install
* clean <flow>
* install
* play <flow>

cylc/flow/tui/app.py
```diff
@@ -310,11 +310,10 @@ def get_snapshot(self):
...
             return dummy_flow({
                 'name': self.reg,
                 'id': self.reg,
-                'status': 'stopped',
+                'status': 'Workflow not found',
```
oliver-sanders, line 316: But what if the workflow is genuinely stopped?

Need to check manually before setting "workflow not found" as the status.

cylc/flow/tui/app.py
wxtim: Removed by accident along with debugging stuff.

```suggestion
             self.client = None 
```

cylc/flow/tui/app.py
wxtim: That was what my question on the issue was getting at

cylc/flow/tui/app.py
```diff
@@ -310,11 +310,10 @@ def get_snapshot(self):
                 }
             )
         except WorkflowStopped:
```
oliver-sanders, line 312: FYI: I'm going to take a look at getting the client to raise a `WorkflowDoesNotExist` error in place of the WorkflowStopped when appropriate, just need to make sure there aren't any unintended side-effects - see https://github.com/cylc/cylc-flow/pull/4776 (which has a similar problem)

wxtim: This PR is on the back burner. I can close and un-assign the issue since it's right at the bottom of my TODO list if you think that's appropriate.

MetRonnie (commented):
My observations from testing out:
- If a workflow dir does not exist, TUI refuses to load and instead you get a message
  ```
  UserInputError: Path does not exist: ~/cylc-run/dir
  ```
- If the dir does exist in `~/cylc-run`, but does not contain a flow.cylc or suite.rc, you get the "workflow not found" message inside TUI as expected
- However, you still have the option to play it in the menu when you press Enter, only for it come up with an error
  ```
  WorkflowFilesError: No flow.cylc or suite.rc in ~/cylc-run/dir
  ```

wxtim:
@MetRonnie - Can you clarify exactly what you mean in #4715 by `cylc tui non-exist`? Is `non-exist` a non-existant workflow folder? Or a real folder without a clearly identified workflow within it?

MetRonnie:
I can't remember. Either it was a non-existent folder, which means the `UserInputError` is new since then, or it was an empty folder or something like that

wxtim:
> Also, if the dir is a parent of a run dir, it says "workflow not found" inside TUI, which is perhaps misleading

Can it be the parent of a run-dir?

MetRonnie:
> > Also, if the dir is a parent of a run dir, it says "workflow not found" inside TUI, which is perhaps misleading
> 
> Can it be the parent of a run-dir?

Not sure what you mean. Here is an example of what I was talking about:
```
$ tree -L 1 ~/cylc-run/blah
~/cylc-run/blah
|-- _cylc-install
|-- run1
`-- runN -> run1

$ cylc tui blah
```

wxtim:
@MetRonnie - using the equivalent of your example I get Cylc TUI saying

`temp1/run1 - workflow not found`

By the point that the logic I've inserted takes place any input has been expanded to a full workflow name.

MetRonnie:
Ah, I oversimplified my example. Try this instead (another level in the directory structure):
```
$ cd ~/cylc-run
$ tree -L 2 foo
foo
`-- bar
    |-- _cylc-install
    |-- run1
    `-- runN -> run1

$ cylc tui foo
```

oliver-sanders:
My observations from testing out:

> If a workflow dir does not exist, TUI refuses to load and instead you get a message
> UserInputError: Path does not exist: ~/cylc-run/dir

Good

> If the dir does exist in ~/cylc-run, but does not contain a flow.cylc or suite.rc, you get the "workflow not > found" message inside TUI as expected

Good

> However, you still have the option to play it in the menu when you press Enter, only for it come up with an error
> WorkflowFilesError: No flow.cylc or suite.rc in ~/cylc-run/dir

Fine IMO. I don't think we need to obsess about these niche cases.

> Also, if the dir is a parent of a run dir, it says "workflow not found" inside TUI, which is perhaps misleading

Fine IMO, same result as you would get with any other Cylc command.

oliver-sanders (approved):

MetRonnie:
Perhaps "No flow.cylc or suite.rc in ~/cylc-run/dir" would be better than "Workflow not found"?

datamel (approved):
I have manually tested this, working well for me.

cylc/flow/tui/app.py
oliver-sanders: This suggestion was not applied - https://github.com/cylc/cylc-flow/issues/4818

PR #4751, opened 2022-03-22
reduce transient imports

oliver-sanders:
Small things spotted reviewing the `cylc.flow.scripts.cylc` script.

Imports cost time usually this is unimportant, however, for scripts that are likely to be called many times it adds up. We should try to avoid unnecessary imports where possible for these cases. Init files are imported by default so it can help to move things out of them.

Should really be an 8.1.0 change, however, RC3 will contain some big diffs.

**Requirements check-list**


hjoliver (approved):
Looks fine :+1:

wxtim (approved):

PR #4724, opened 2022-03-01
config: assorted fixes for python template variables supported in various configurations

oliver-sanders:
Closes https://github.com/cylc/cylc-flow/issues/4764
Closes https://github.com/cylc/cylc-flow/issues/4763

> **Note:** Issue has widened in scope since initial post.

> Sibling: https://github.com/cylc/cylc-doc/pull/422

Some assorted fixes for Python template variables supported in various configs.

* `[mail]footer`
  * The "suite" template var is provided for scheduler events but not for task events.
  * (bug) KeyError was being caught for scheduler events but not for task events.
  * (bug) ValueError was not being caught for either (would tear the workflow down for a broken template).
* `[scheudler][events]`
  * Formalised event data into a self-documenting enum.
  * Changed `uuid_str` to `uuid` (was previously `suite_uuid`).
  * (bug) Event handlers configured in the global config were ignored (4764).
* `[runtime][<namespace>][events]`
  * Moved documentation from global config into enum.
  * (bug) `user@host` was replaced by `platform_name` but no deprecated config was left in its place, added it back.
* `[meta]URL`
  * (bug) Restored `suite_name` (had been renamed `workflow_name`), added deprecation notice.
  * Deprecated `task_name` in favour of `task` for consistency with event handlers.
  * (bug) Catch KeyError and ValueError in config load and handle better.

Added some integration tests to ensure template var issues in mail footer and custom event handlers are handled properly.

PR #4765, opened 2022-03-24
Fix spurious "Default For" in workflow config reference

MetRonnie:
But leave it in the global config reference.

Before:

After:

**Requirements check-list**
- [x] Does not need tests 
- [x] No change log entry required (docs).
- [x] Is documentation update.

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -17,6 +17,7 @@
 
 import os
 from sys import stderr
+from textwrap import dedent
 from typing import List, Optional, Tuple, Any
 
 from contextlib import suppress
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Beet you to it in #4724

MetRonnie:
Ah, didn't realise. However, couple of issues with #4724
- it only does it for `* handlers`, missing out the other types of item
- there is a typo `:Default For;` ending with a semicolon on one of them

oliver-sanders:
Should be sorted on master now.

hjoliver:
I've confirmed the problem no longer exists, with cylc-doc + cylc-flow master.
(Re-open if you disagree @MetRonnie)

PR #4771, opened 2022-03-25
scan: exclude Cylc 7 workflows from scan results

oliver-sanders:
An unintended side-effect of #4506 is that `cylc scan` is no longer able to tell between A Cylc 7 workflow that is running under Cylc 8 in compatibility mode and one that is running under Cylc 7.

This PR uses the `log/workflow` directory as a marker to exclude Cylc 7 workflows from scan results.

Note this fix also impacts the UIS which currently displays Cylc 7 workflows.

**Requirements check-list**


CHANGES.md
```diff
@@ -28,6 +28,14 @@ creating a new release entry be sure to copy & paste the span tag with the
 `actions:bind` attribute, which is used by a regex to find the text to be
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
+-------------------------------------------------------------------------------
+## __cylc-8.0rc2 (<span actions:bind='release-date'>Pending</span>)__
```
MetRonnie, line 32:
```suggestion
## __cylc-8.0rc3 (<span actions:bind='release-date'>Pending</span>)__
```

MetRonnie (changes requested):
Unfortunately, in my testing this merely makes a Cylc7 workflow that used to show up as `myflow` show up as `myflow/log/suite` now, when running `cylc scan --state=all`

oliver-sanders:
Needed to change `dir_is_flow` to cover three cases:

* It's a compatible workflow (show it in scan results).
* It's an incompatible workflow (don't show it in scan results but also don't descend further looking for workflows).
* It's not a workflow (descend further looking for workflows).

cylc/flow/network/scan.py
```diff
@@ -206,13 +233,14 @@ def _scan_subdirs(listing: List[Path], depth: int) -> None:
         for task in done:
             path, depth, contents = task.result()
             running.remove(task)
-            if dir_is_flow(contents):
+            is_flow = dir_is_flow(contents)
```
MetRonnie, line 236: Might be worth repeating the type annotation here so that nobody removes the `is_flow is False` check below
```suggestion
            is_flow: Optional[bool] = dir_is_flow(contents)
```

tests/integration/test_scan.py
```diff
@@ -406,3 +406,43 @@ async def test_scan_sigstop(
...
+    Path(cylc7, WorkflowFiles.LOG_DIR, 'suite').mkdir(parents=True)
+    Path(cylc7, WorkflowFiles.LOG_DIR, 'suite', 'log').touch()
+
+    # a Cylc 7 workflow running under Cylc 8 in compatibility mode
+    # (should appear in scan results)
+    cylc8 = tmp_path / 'cylc8'
+    cylc8.mkdir()
+    (cylc8 / WorkflowFiles.SUITE_RC).touch()
+    Path(cylc8, WorkflowFiles.LOG_DIR, 'workflow').mkdir(parents=True)
+    Path(cylc8, WorkflowFiles.LOG_DIR, 'workflow', 'log').touch()
```
MetRonnie, line 434: I think we should test that a `suite.rc` workflow previously run under Cylc 7 but now running under Cylc 8 in back-compat mode should appear in the scan results. I.e.
```
~/cylc-run/my_workflow
|-- log
|   |-- suite
|   |   `-- log
|   `-- workflow
|       `-- log
`-- suite.rc
```
I have tested this manually and it does work.

MetRonnie: If someone has deleted the Cylc 7 DB but left the `log/suite` dir behind

oliver-sanders: Somewhat over the top but ok, why not.

MetRonnie (approved):

datamel (approved):
Checked out and tested. 
I initially had a problem with this (a consequence of not tidying my run directories 🤦 ), I had a log directory from a previous run in the directory above my run dir which prevented it from showing. I suspect scan is going to need to be thought of when doing https://github.com/cylc/cylc-flow/pull/4477, as this will (I think) legitimize this scenario.

PR #2938, opened 2019-01-31
review: fix waiting tasks

oliver-sanders:
#### At present waiting tasks do not appear in Cylc Review due to a SQL query error.

We were [joining](https://www.w3schools.com/sql/sql_join.asp) the tasks and jobs tables. In SQL terms `JOIN` means intersection. As waiting tasks do not have jobs they were excluded in the intersection. I've changed the `JOIN` to `LEFT JOIN` to include tasks with out jobs.

PR #4781, opened 2022-03-29
waiting turned on by default for Cylc 8 flows.

wxtim:

#### Issue summary

User report that when in the "task jobs list" view, and selecting failed from the "Select by job status menu" they don't see failed jobs for the task which is waiting for a retry because the "waiting" checkbox in the "Select by task statuses" is not selected by default.

At Cylc 7 this was done in https://github.com/cylc/cylc-flow/pull/2938. It probably isn't necessary at Cylc 8. It's also necessary to enable "waiting" by default at Cylc 8 because tasks which would previously have been described as "retrying" &c are now classified as "waiting".

#### Requirements check-list
- [x] Manually tested

### DIFF
--- a/lib/cylc/review.py
+++ b/lib/cylc/review.py
@@ -306,8 +306,12 @@ class CylcReviewService(object):
         if not task_status:
             # default task statuses - if updating please also change the
             # $("#reset_task_statuses").click function in cylc-review.js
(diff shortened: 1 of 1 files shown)


wxtim:
Closed for now, pending understanding of #2938

Issue #4780, opened 2022-03-29
Flake 8 SIM90*


datamel:
As mentioned on Element by @hjoliver. 
Flake8 has had an update which requires action:
```
./cylc/flow/broadcast_mgr.py:224:17: SIM902 Use keyword-argument instead of magic boolean for 'self.data_store_mgr.delta_task_ext_trigger(itask, qid, qmsg, True)'
./cylc/flow/config.py:536:12: SIM902 Use keyword-argument instead of magic boolean for 'getattr(self.options, 'is_validate', False)'
./cylc/flow/config.py:820:21: SIM902 Use keyword-argument instead of magic boolean for 'getattr(self.options, 'check_circular', False)'
./cylc/flow/config.py:898:21: SIM903 Use keyword-argument instead of magic number for 'wrap(', '.join(out_of_bounds), 70)'
./cylc/flow/config.py:1506:17: SIM908 Use '['clock-trigger', 'clock-expire', 'external-trigger'].get(task_type)' instead of 'if task_type in ['clock-trigger', 'clock-expire', 'external-trigger']: ['clock-trigger', 'clock-expire', 'external-trigger'][task_type]'
./cylc/flow/config.py:1782:23: SIM902 Use keyword-argument instead of magic boolean for 'getattr(self.options, 'is_validate', False)'
./cylc/flow/data_store_mgr.py:194:8: SIM902 Use keyword-argument instead of magic boolean for 'getattr(delta, 'added', False)'
./cylc/flow/data_store_mgr.py:200:8: SIM902 Use keyword-argument instead of magic boolean for 'getattr(delta, 'updated', False)'
./cylc/flow/exceptions.py:217:29: SIM902 Use keyword-argument instead of magic boolean for 'str(item).splitlines(True)'
./cylc/flow/job_runner_mgr.py:278:16: SIM902 Use keyword-argument instead of magic boolean for 'getattr(self._get_sys(job_runner_name), 'SHOULD_KILL_PROC_GROUP', False)'
./cylc/flow/job_runner_mgr.py:304:29: SIM902 Use keyword-argument instead of magic boolean for 'err.splitlines(True)'
./cylc/flow/job_runner_mgr.py:398:29: SIM902 Use keyword-argument instead of magic boolean for 'value.splitlines(True)'
./cylc/flow/job_runner_mgr.py:411:41: SIM903 Use keyword-argument instead of magic number for 'st_file_path.rsplit(os.sep, 6)'
[...]

PR #4783, opened 2022-03-29
flake8

datamel:
These changes close https://github.com/cylc/cylc-flow/issues/4780

- [x] Does not need tests (why? flake8).

### DIFF
--- a/cylc/flow/scripts/cylc.py
+++ b/cylc/flow/scripts/cylc.py
@@ -650,9 +650,8 @@ def main():
                 else:
                     command = cmd_args.pop(0)
 
-            if command in ALIASES:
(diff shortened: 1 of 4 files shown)


cylc/flow/scripts/cylc.py
```diff
@@ -650,9 +650,9 @@ def main():
                 else:
                     command = cmd_args.pop(0)
 
+            # this is an alias to a command
             if command in ALIASES:
-                # this is an alias to a command
```
MetRonnie, line 654: Nitpick: I disagree with the move of the comment (but actually I don't think the comment adds anything anyway)

PR #4785, opened 2022-03-29
Ignore flake8-simplify SIM9xx rules.

hjoliver:

Discussed in Element chat: ignore "experimental" flake8-simplify rules. 

(These do eventually get moved into the mainstream, if stabilized and approved, so we should remember to check locally sometimes, to avoid surprises).


datamel (approved):
👍

Issue #4766, opened 2022-03-24
Misleading error message when run directory doesn't exist


jhaiduce:
If I try to run `cylc play` on a workflow or run that doesn't exist, cylc issues an error message of the form:

```
WorkflowFilesError: runN directory at /home/user/cylc-run/nonexistent-workflow/runN is a broken or invalid symlink
```

The error message implies that the workflow exists, but the link to the run directory is broken. But this same error message is issued if the requested workflow/run doesn't exist at all.

hjoliver:
@jhaiduce I can't reproduce this unless I explicitly target runN (which we're not supposed to do) in a non-existent run directory.

Below, if there is no workflow called "none" installed at all:

![Screenshot from 2022-03-25 10-01-44](https://user-images.githubusercontent.com/2362137/160009716-bbfe95fe-cf5a-48de-b9f1-3243404f0c48.png)

hjoliver:
Could you describe the exact circumstance that causes this for you?

jhaiduce:
I was explicitly targeting runN, and I had a typo in the workflow name.

jhaiduce:
> @jhaiduce I can't reproduce this unless I explicitly target runN (which we're not supposed to do) in a non-existent run directory.

I didn't know we weren't supposed to explicitly target runN. Someone gave me instructions on the matrix.org chat a while back (maybe a year ago) and I've been doing it that way ever since.

hjoliver:
Ah, OK.  There was a short period when the `runN` symlink was implemented but not yet used by commands to infer that lastest run automatically.

Now you can do this:
```
~/cylc-src/foo$ cylc install  # install foo/run1
~/cylc-src/foo$ cylc play foo  # play foo/run1

~/cylc-src/foo$ cylc install  # install foo/run2
~/cylc-src/foo$ cylc play foo  # play foo/run2

~/cylc-src/foo$ cylc stop foo  # stop foo/run2
~/cylc-src/foo$ cylc stop foo/run1  # (explicit run1 required here)
```

hjoliver:
However, I'll leave this issue open. We should:
- make the error message better if you do explicitly target `runN`
- perhaps also warn that explicit use of `runN` is not necessary

jhaiduce:
Thanks for explaining that @hjoliver. It's certainly easier not to put /runN there every time!

PR #4789, opened 2022-03-30
Fix Error Message for RunN Workflow IDs

datamel:
These changes close https://github.com/cylc/cylc-flow/issues/4766

**Requirements check-list**
- [x] No change log entry required (Error change).

### DIFF
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -1236,6 +1236,7 @@ def infer_latest_run(
 ) -> Tuple[Path, Path]:
     """Infer the numbered run dir if the workflow has a runN symlink.
 
+    Warns users that exiplicit use of runN is unnessary.
     Args:
         path: Absolute path to the workflow dir, run dir or runN dir.
         implicit_runN: If True, add runN on the end of the path if the path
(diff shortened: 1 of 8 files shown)


cylc/flow/workflow_files.py
```diff
@@ -1246,14 +1247,22 @@ def infer_latest_run(
...
     try:
         reg = path.relative_to(cylc_run_dir)
     except ValueError:
         raise ValueError(f"{path} is not in the cylc-run directory")
+    if not path.exists():
+        raise UserInputError(f'Path does not exist: {path}')
     if path.name == WorkflowFiles.RUN_N:
+        LOG.warning(
+            f"The use of {WorkflowFiles.RUN_N} in the Workflow ID is not"
+            f" necessary."
```
hjoliver, line 1264: Not a biggie, but this ought to be an infrequently-seen warning, so I think we could afford to be more explicit, and take the opportunity to explain what the `runN` symlink is actually for:
```suggestion
            f"Explicit use of {WorkflowFiles.RUN_N} in the Workflow ID is not"
            f" necessary. It is used automatically to select the latest run number."
```

tests/unit/test_id_cli.py
```diff
@@ -56,7 +57,9 @@ def abc_src_dir(tmp_path_factory):
 )
 async def test_parse_ids_workflows(ids_in, ids_out):
     """It should parse workflows & tasks."""
-    workflows, _ = await parse_ids_async(*ids_in, constraint='workflows')
+    with patch.object(Path, 'exists') as mock_exists:
```
wxtim, line 60: There are a lot of very similar context managers here - I wonder if you could turn all these into a fixture?

tests/unit/test_workflow_files.py
```diff
@@ -301,13 +301,32 @@ def test_infer_latest_run(
         assert infer_latest_run(path) == expected
 
 
+def test_infer_latest_run_raise_warning_for_runN(caplog, tmp_run_dir):
+    caplog.set_level(logging.WARNING, 'log')
+    cylc_run_dir: Path = tmp_run_dir()
```
wxtim, line 306: I don't think that this assignment is necessary either - `temp_run_dir` has the default fixture scope of function so you can dispense with this line if you wish, just using `tmp_run_dir()` each time you need it.

tests/unit/test_workflow_files.py
```diff
@@ -301,13 +301,32 @@ def test_infer_latest_run(
...
+    run_dir = cylc_run_dir / 'run1'
+    run_dir.mkdir()
+    runN_path = cylc_run_dir / 'runN'
+    runN_path.symlink_to('run1')
+    infer_latest_run(runN_path)
+    assert ("The use of runN in the Workflow ID is not necessary."
```
wxtim, line 312: Consider using a regex or similar to check that the essentials of this message are there, without making it very fragile in terms of exact words used?

wxtim (approved):
Sorry, I went a bit pedantic on your test. It works, so feel free to implement or not my suggestions.

I also agree with Hilary's comment on verbosity of warning.

Approved.

hjoliver (approved):

PR #4791, opened 2022-03-31
fix graphql node state and remove ghosts filters

dwsutherland:
These changes partially address https://github.com/cylc/cylc-uiserver/issues/341

Sibling PR https://github.com/cylc/cylc-ui/pull/980

image

Delta's without a state field set will be filtered by state arguments... This is done by supplementing the node state using the data-store when not set..

**Requirements check-list**


hjoliver:
> Already covered by existing tests.

How can we test that non-state updates get through though?  (Damn, "through though" is indictment of the English language, innit).

dwsutherland:
> > Already covered by existing tests.
> 
> How can we test that non-state updates get through though? (Damn, "through though" is indictment of the English language, innit).

Opps, I meant graphql functional tests (and the entire CLI) all use the node filter I changed.
WRT graphql subscriptions, yes we can't test at the scheduler yet.

hjoliver:
> WRT graphql subscriptions, yes we can't test at the scheduler yet.

Yeah, but a UIS test maybe?

dwsutherland:
> > WRT graphql subscriptions, yes we can't test at the scheduler yet.
> 
> Yeah, but a UIS test maybe?

Don't think we have and websockets tests for graphql subscriptions..

hjoliver:
Integration test failures:
```
FAILED tests/integration/test_resolvers.py::test_get_nodes_all - assert 1 == 0
FAILED tests/integration/test_scheduler.py::test_illegal_config_load[illegal item-True]
FAILED tests/integration/test_scheduler.py::test_illegal_config_load[illegal cycle point-True]
```

The last two are fixed by #4794

The first one is presumably genuine.

dwsutherland:
> Integration test failures:
> 
> ```
> FAILED tests/integration/test_resolvers.py::test_get_nodes_all - assert 1 == 0
> FAILED tests/integration/test_scheduler.py::test_illegal_config_load[illegal item-True]
> FAILED tests/integration/test_scheduler.py::test_illegal_config_load[illegal cycle point-True]
> ```
> 
> The last two are fixed by #4794
> 
> The first one is presumably genuine.

Yeah, genuine.. I fixed the filter, we shouldn't really have any nodes (with state field) that have empty state, but possible if the node is added but not updated yet.. so handle this now.

cylc/flow/network/resolvers.py
```diff
@@ -329,14 +329,33 @@ async def get_workflows(self, args):
...
+            return None
+        # Don't lookup data-store state in getattr, as it might not be there.
+        # (args are evaluated first)
+        state = getattr(node, 'state', None)
+        # If node has no state, it must be in the data-store.
+        if not state:
+            return self.data_store_mgr.data[
+                Tokens(node.id).workflow_id
+            ][node_type][node.id].state
+        return state
```
hjoliver, line 344:
```suggestion
        # If node has no state, it must be in the data-store.
        return (
            getattr(node, 'state', None) or
            self.data_store_mgr.data[
                Tokens(node.id).workflow_id
            ][node_type][node.id].state
        )
```

hjoliver: (No need for the intermediate variable).

hjoliver:
```
 tests/integration/test_task_pool.py::test_preparing_tasks_on_restart 
Error: The action has timed out.
```
I can't get this to hang locally though.

cylc/flow/network/resolvers.py
dwsutherland: Fixed.

dwsutherland:
> ```
>  tests/integration/test_task_pool.py::test_preparing_tasks_on_restart 
> Error: The action has timed out.
> ```
> 
> I can't get this to hang locally though.

I put an error suppression around accessing the state from the data-store, I think it must either the mock tests or the scenarios causing that point to fail.. it was intermittent though..

dwsutherland:
> > closes [cylc/cylc-uiserver#328](https://github.com/cylc/cylc-uiserver/issues/328)
> 
> Right issue?

It fixes an issue that make subscriptions appear not to work..
node-filters

But probably not the same thing..

cylc/flow/network/resolvers.py
```diff
@@ -194,15 +194,18 @@ def node_ids_filter(tokens, state, items) -> bool:
             )
             # match cycle/task/job state
             and (
-                not get_state_from_selectors(item)
+                not (
+                    state is not None
+                    and get_state_from_selectors(item)
+                )
```
hjoliver, line 200: As written, this logic can be simplified to:
```
(
   state is None
   or get_state_from_selectors(item) is None
)
```
can't it?

(And should it actually be an `and` in the middle?)

cylc/flow/network/resolvers.py
```diff
@@ -216,15 +219,12 @@ def node_filter(node, node_type, args):
...
-            args.get('ghosts')
-            or state
-            or node_type in DEF_TYPES
-        )
-        and (
-            not args.get('states')
+            not (
+                state is not None
+                and args.get('states')
+            )
```
hjoliver, line 227: Ditto the logic question above.

hjoliver:
(Tests as working, but I'm slightly bamboozled by the logic above ... but maybe my brain is fried)

oliver-sanders:
> closes https://github.com/cylc/cylc-uiserver/issues/328
>> Right issue?
>>> It fixes an issue that make subscriptions appear not to work..

GraphiQL still broken for me, could you explain what you mean by "appear not to work" to help me review.

cylc/flow/network/resolvers.py
dwsutherland: `and` means the second line doesn't get evaluated right? (which I thought more efficient)
Otherwise yes it could be written that way; `!(A & B)` is equivalent to `(!A or !B)`

dwsutherland: Nevermind, it works with or too ... 🤦‍♂️

dwsutherland: fixed

cylc/flow/network/resolvers.py
dwsutherland: fixed.

hjoliver (approved):
Thanks @dwsutherland

dwsutherland:
> > closes [cylc/cylc-uiserver#328](https://github.com/cylc/cylc-uiserver/issues/328)
> > > Right issue?
> > > > It fixes an issue that make subscriptions appear not to work..
> 
> GraphiQL still broken for me, could you explain what you mean by "appear not to work" to help me review.

Yours doesn't even get any data.. So I think it's a different issue, as this one you still got sent something but missed deltas.

cylc/flow/network/resolvers.py
oliver-sanders: Resolved?

Isn't this equivalent to:

```suggestion
                (
                    state is None
                    or get_state_from_selectors(item)
                )
```

cylc/flow/network/resolvers.py
```diff
@@ -194,15 +194,18 @@ def node_ids_filter(tokens, state, items) -> bool:
...
+                )
                 or get_state_from_selectors(item) == state
             )
         )
         for item in uniq(items)
     )
 
 
-def node_filter(node, node_type, args):
+def node_filter(node, node_type, args, state):
```
oliver-sanders, line 208: I'm not too clear on which bits belong to the node and which bits belong to the query?

Is the "state" the node state or the query state?

cylc/flow/network/resolvers.py
```diff
@@ -216,20 +219,17 @@ def node_filter(node, node_type, args):
...
         (
-            args.get('ghosts')
-            or state
-            or node_type in DEF_TYPES
-        )
-        and (
-            not args.get('states')
+            (
+                state is None
+                or not args.get('states')
```
oliver-sanders, line 226: So if the node has no state (how does this happen?) then we include it in filter results irrespective of whether a query state was specified? Or am I misunderstanding?

cylc/flow/network/resolvers.py
```diff
@@ -329,14 +329,34 @@ async def get_workflows(self, args):
...
+            return None
+        # Don't lookup data-store state in getattr, as it might not be there.
+        # (args are evaluated first)
+        # If node has no state, it must be in the data-store.
+        with suppress(Exception):
+            return (
+                getattr(node, 'state', None)
+                or self.data_store_mgr.data[
+                    Tokens(node.id).workflow_id
+                ][node_type][node.id].state
```
oliver-sanders, line 344: Could you explain what these two different methods mean?

* The first is getting the state from the node, the node as it exists in the data store?
* The second is getting the state from the data store directly. How does this differ from the node?

cylc/flow/network/schema.py
```diff
@@ -633,7 +632,6 @@ class Meta:
         lambda: TaskProxy,
         description="""Task cycle instances.""",
         args=PROXY_ARGS,
-        ghosts=GHOSTS_DEFAULT,
```
oliver-sanders, line 636: I don't really understand what "ghosts" were, why we had them or why we can dispense of them now?

oliver-sanders (commented):
I don't think I'm able to test this with GraphiQL broken for me (any ideas)? I'm also not sure I understand the problem this is fixing.

I've added a couple of questions in the review, would be really helpful to add a couple of code comments to clarify them.

oliver-sanders:
(will see if anyone else has working GraphiQL to review...)

dwsutherland:
> I don't think I'm able to test this with GraphiQL broken for me (any ideas)? I'm also not sure I understand the problem this is fixing.

It fixes this issue:
https://github.com/cylc/cylc-uiserver/issues/341

And the above gif in the comment;
https://github.com/cylc/cylc-flow/pull/4791#issuecomment-1085765206
Shows the deltas coming through with outputs (which were being filtered out previously, as the delta had to state field)

The fix is viewable in the UI tree view (don't need graphiql), see the above mentioned issue.

cylc/flow/network/resolvers.py
dwsutherland: `args` will have the query state to filter against
`state` is the node state or data-store node state (if state isn't set in the delta, i.e. with outputs).

This ensures we always have a state to filter against (i.e. if we ever wanted to filter deltas by state, then deltas without a state field set can still be filtered by their data-store state)

cylc/flow/network/resolvers.py
dwsutherland: > So if the node has no state (how does this happen?)
It can happen if it's a task definition or something like that.
> then we include it in filter results irrespective of whether a query state was specified?
Correct, as we use the same filter for nodes without states (task defs, family defs)

cylc/flow/network/resolvers.py
dwsutherland: > The first is getting the state from the node, the node as it exists in the data store?

`node` could be from a data-store or a delta-store.

> The second is getting the state from the data store directly. How does this differ from the node?

Yes, if the first returns `None`  or empty string, fetch the state from the data-store... 

If we want only deltas of specified state(s), then we need to filter out those deltas that don't contain state.

cylc/flow/network/schema.py
dwsutherland: Ghosts where part of a crude n-window  (n=1)  prototype... It has long been deprecated, and need to be removed..
It relates do this PR because it was in the part of the filter I had to change.

cylc/flow/network/resolvers.py
oliver-sanders: Cheers, so would this be correct?:

```python
def node_filter(node, node_type, args, state):
    """Filter nodes based on attribute arguments

    Args:
        node: The node to filter (from the data or delta store).
        node_type: The type of the node being filtered (or is it the type to filter by?)
        args: The query arguments.
        state: The state of the node that is being filtered.
    """
```

cylc/flow/network/resolvers.py
oliver-sanders: Ok, that makes sense, is this right?:

```python
            return (
                # try to retrieve the state from the node (could be a delta-store node)
                getattr(node, 'state', None)
                # fall back to the data store (because this could be a delta-store
                # node which might not have the state field set)
                or self.data_store_mgr.data[
                    Tokens(node.id).workflow_id
                ][node_type][node.id].state
```

cylc/flow/network/schema.py
oliver-sanders: Ok, so I can test this is still working correctly just be running the UI on a workflow and making sure the N=1 window shows as expected.

hjoliver: (Note the original movtivation for this PR was https://github.com/cylc/cylc-uiserver/issues/341 - i.e., custom task outputs not showing up at the UI in real time... which this PR definitely fixes.)

cylc/flow/network/resolvers.py
dwsutherland: Correct 👍

cylc/flow/network/schema.py
dwsutherland: > Ok, so I can test this is still working correctly just be running the UI on a workflow and making sure the N=1 window shows as expected.

The current n-window implementation is orthogonal to that old prototype. These Ghosts literally weren't used anywhere, but the args were still in the API and being filtered for.
But sure, test `n=1` if like :)

oliver-sanders:
Replicated https://github.com/cylc/cylc-uiserver/issues/341, fixed the problem for me, 👍.

oliver-sanders (approved):

oliver-sanders:
(btw tested Tui too, all good)

cylc/flow/network/resolvers.py
dwsutherland: Yes, that's right.

PR #4793, opened 2022-03-31
scheduler: log command failures at error level

oliver-sanders:
Spotted whilst reviewing: https://github.com/cylc/cylc-flow/pull/4720

Command errors are logged at the warning level rather than error. These errors aren't `CRITICAL` (they don't tear down the scheduler), but they are a bit more important than the `WARNING` level which makes it harder to pick them out in the logs.

Also combine two log entries into one for clarity.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Does not need tests (why?).
- [x] No change log entry required (why? e.g. invisible to users).
- [x] No documentation update required.

PR #4794, opened 2022-04-01
Fix integration test.

hjoliver:
Small follow-up fix for #4793


dwsutherland (approved):
Looks good, tests passing (MACOS one is a reoccuring intermitted timeout) 👍

Issue #4802, opened 2022-04-06
Tidy of log directory


datamel:
Discussed in UK Team meeting: 
The log directory has some discrepancies which should be tidied before release.
Renaming:
- log/workflow => log/scheduler
- log/flow-config => log/config
- file-installation-log => remote-install-log
 
Timestamps should be consistently at the beginning or end of the log (to be decided) and for file installation and config should include <start/restart/reload> as appropriate.

The config files end in .cylc currently which is why they currently have the date stamp at the beginning of the file name. 

Note the timestamp also differs between logs, although the log is created before config load so tricky to know what is the preferred time zone.

Note that this change will have implications for cylc-review.

Before:
```
$ tree log
log
├── db
├── flow-config
│   ├── 20220406T124846Z-start.cylc
│   ├── 20220406T124905Z-restart.cylc
│   ├── 20220406T124913Z-reload.cylc
│   └── flow-processed.cylc
├── install
│   └── 20220406T124844Z-install.log
├── job
│   └── 1
│       ├── a
│       │   ├── 01
│       │   │   ├── job
│       │   │   ├── job-activity.log
│       │   │   ├── job.err
│       │   │   ├── job.out
│       │   │   ├── job.status
│       │   │   └── job.xtrace
│       │   └── NN -> 01
│       ├── b
│       │   ├── 01
│       │   │   ├── job
│       │   │   ├── job-activity.log
│       │   │   ├── job.err
│       │   │   ├── job.out
│       │   │   ├── job.status
│       │   │   └── job.xtrace
│       │   └── NN -> 01
│       └── c
│           ├── 01
│           │   ├── job
│           │   ├── job-activity.log
│           │   ├── job.err
│           │   ├── job.out
│           │   ├── job.status
│           │   └── job.xtrace
│           └── NN -> 01
└── workflow
    ├── file-installation-log -> file-installation-log.20220406T134924+01
    ├── file-installation-log.20220406T134850+01
    ├── file-installation-log.20220406T134924+01
    ├── log -> log.20220406T134905+01
    ├── log.20220406T134846+01
    └── log.20220406T134905+01
```
Proposed After:
```
$ tree log
log
├── db
├── config
│   ├── 20220406T124846Z-start.cylc
│   ├── 20220406T124905Z-restart.cylc
│   ├── 20220406T124913Z-reload.cylc
│   └── flow-processed.cylc
├── install
│   └── 20220406T124844Z-install.log
├── job
│   └── 1
[...]

PR #4804, opened 2022-04-06
Close File Installation Log

datamel:
Turns out the file installation log is not closed properly. 

**Requirements check-list**
- [x] Appropriate tests are included (strange one to test, attempted to prove closure by reloading and checking that a new file is opened, which is the desired behaviour.).

### DIFF
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -356,7 +356,8 @@ def scheduler_cli(options: 'Values', workflow_id: str) -> None:
     # NOTE: any threads which include sleep statements could cause
     #       sys.exit to hang if not shutdown properly
     LOG.info("DONE")
-    close_log(LOG)
+    for log in (LOG, RSYNC_LOG):
+        close_log(log)
     sys.exit(ret)
 
 


cylc/flow/task_remote_mgr.py
```diff
@@ -522,6 +523,7 @@ def _file_install_callback(self, ctx, platform, install_target):
             RSYNC_LOG.info(
                 'File installation information for '
                 f'{install_target}:\n{ctx.out}')
+            close_log(RSYNC_LOG)
```
oliver-sanders, line 526: I think _file_install_callback gets called once per remote-init so if we are remote-init'ing multiple platforms the logs will get closed multiple times?

Sadly I think because of that this is hard to do.

E.G. if we have three platforms the order of events would look something like this:

1. open logs
2. remote-init x
3. remote-init y
4. remote-init z
5. close logs

Where we need to wait for all remote-inits to succeed before closing the log.

Best option I can think of is to close the file install logs on shutdown and reload. On reload we would also need to open the next file-install log. Potential problem with that is if a remote-init is in progress when the reload is started it will try to log to a closed handler which could cause issues.

datamel: Yes, good spot. I take it we shouldn't have a file installation log per platform?

oliver-sanders: We could stick them into one file or break them apart into separate ones, both would be valid.

If we were to break them apart into separate files then we no longer need to use `logging` (because we can just `with open(<file>)`, there's only one entry to write per file).

oliver-sanders: (one to raise at the team meet tomorrow)

hjoliver: (one per install target seems a good idea to me)

oliver-sanders: Added this in https://github.com/cylc/cylc-flow/issues/4802#issuecomment-1091424254

datamel:
This log becomes redundant in https://github.com/cylc/cylc-flow/pull/4836. Closing it for now as quick fix for rc3.

hjoliver (approved):

hjoliver:
> This log becomes redundant in https://github.com/cylc/cylc-flow/pull/4836. Closing it for now as quick fix for rc3.

OK cool. I can't see any way for this small change to cause trouble; no need for a second review.

Issue #4796, opened 2022-04-04
ERROR - Workflow shutting down - CylcError: Could not load plugin: "health check"


hiker:
When following the tutorial and trying to `cylc play...`, I get:
``` 
ERROR - Workflow shutting down - CylcError: Could not load plugin: "health check"
$ cylc --version
8.0rc2
```
fresh installed on a system without previous cylc installation, using `pip install`, so I believe I'm on the latest available version.

To reproduce: install a dummy suite and to `cylc play ...`.

I could work around this issue by creating a global config file ($HOME/.cylc/flow/global.cylc) containing:
```
$ cylc config
[scheduler]
    [[main loop]]
        plugins = 
```

The documentation states that the plugin default is 'health, auto restart". Note that `cylc config` did not report any config details, so nothing else was installed.

It would appear that either the default value for plugins should be changed, or a config file installed, or the tutorial needs to be updated to indicate what needs to be done.

hjoliver:
Thanks for the bug report!

That's strange. The core plugins (including health check) should get installed by default.  A quick `pip install cylc-flow` from PyPI, into a clean venv, works for me. I'll take a closer look tomorrow if the UK team doesn't pitch in overnight.

hiker:
I am not using a virtual env, but I also just tried in a clean on an aws instance, and it worked fine there. Please ignore this issue for now, I will try a clean install on my laptop tomorrow and see if I can reproduce this issue. Maybe I did something ... strange :(

hiker:
I uninstalled cylc, and re-installed first 8.0b3, which I then updated to 8.0rc2 - and I still have the problem (not with 8.0b3). I will try to debug this a bit further in the next days.

hjoliver:
Can you try using a virtual env, to see if that gives a clue as to what's going on?

hiker:
OK, found the problem. I had protobuf 3.17 installed in my `$HOME/.local`, and (after updating most of my systems) protobuf 3.20 in `/usr/local/...`. I am not certain why I had two :(

The full error message is actually:
```
2022-04-05T12:11:34+10:00 ERROR - Workflow shutting down - CylcError: Could not load plugin: "health check" - error message: (protobuf 3.20.0 (/usr/local/lib/python3.8/dist-packages), Requirement.parse('protobuf==3.19.*'))

```
[...]

PR #4806, opened 2022-04-06
Comment to explain protobuf version pinning.

hjoliver:

Add an explanatory comment on protobuf version pinning. Somewhat obvious with a moment's thought, but no harm in being clear about it.

Motivated by: https://github.com/cylc/cylc-flow/issues/4796#issuecomment-1088365632

### DIFF
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -10,6 +10,7 @@ dependencies:
   - graphviz  # for static graphing
   - jinja2 ==2.11.0,<2.12
   - metomi-isodatetime >=1!2.0.2, <1!2.1.0
+  # Constrain protobuf version for compatible Scheduler-UIS comms across hosts
   - protobuf >=3.19.0,<3.20.0
   - psutil >=5.6.0
   - python
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -9,12 +9,8 @@ dependencies:
   - graphene >=2.1,<3
   - graphviz  # for static graphing
   - jinja2 ==2.11.0,<2.12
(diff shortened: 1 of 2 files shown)


hjoliver:
Aside: we should add a note about this to the installation docs, because it comes close to a requirement to use virtual environments for Cylc, if protobuf is installed on the system for other reasons. (right?)

hjoliver:
(Cancelled tests - not needed for this!)

hjoliver:
Also, from the linked issue:

> With a bit of reading into protobuf releases we could potentially consider loosening this to protobuf==3.* depending on our confidence in back-compatibility between minor releases.

This sounds a bit difficult and dangerous. Perhaps we could just comment (in installation docs) that users can try this themselves if not using a venv, and see if it works.

oliver-sanders (approved):

conda-environment.yml
```diff
@@ -9,7 +9,12 @@ dependencies:
   - graphene >=2.1,<3
   - graphviz  # for static graphing
   - jinja2 ==2.11.0,<2.12
+<<<<<<< protobuf-version-comment
+  - metomi-isodatetime >=1!2.0.2, <1!2.1.0
+  # Constrain protobuf version for compatible Scheduler-UIS comms across hosts
+=======
   - metomi-isodatetime >=1!3.0.0, <1!3.1.0
+>>>>>>> master
```
hjoliver, line 17:
```suggestion
  - metomi-isodatetime >=1!3.0.0, <1!3.1.0
  # Constrain protobuf version for compatible Scheduler-UIS comms across hosts
```

PR #4820, opened 2022-04-12
Change UserInputError to InputError

hjoliver:

As I noted on Element chat (but got no response) `UserInputError`:
- sounds a bit insulting to the user
- the "User Input" bit is obvious (and therefore unnecessary) when it comes in response to a console command 
  - (Just `Error` would do, but the exception class name always gets printed...)

Just a quick search-and-replace one-liner:
```console
$ find cylc tests -type f | xargs perl -pi -e 's/UserInputError/InputError/g'
```

One 2-minute review will do if no one objects to the change. 

[UPDATE] Went with `InputError` rather than `CommandLineError` as originally suggested.

**Requirements check-list**

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -63,7 +63,7 @@ from cylc.flow.exceptions import (
     IntervalParsingError,
     TaskDefError,
     ParamExpandError,
-    UserInputError
+    InputError
 )
 import cylc.flow.flags
 from cylc.flow.graph_parser import GraphParser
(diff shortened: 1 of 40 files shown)


oliver-sanders:
UserInputError is used more widely than the CLI.

If the "User" part could be considered offensive (which I guess it could) then how 'bout dropping the "User" and going for "InputError"?

hjoliver:
> (note need to check for imports of this exception in cylc-uiserver and cylc-rose)

Checked, no other use found.

hjoliver:
> then how 'bout dropping the "User" and going for "InputError"?

Yep that's fine, and shorter.  Done.

oliver-sanders (approved):

hjoliver:
Rebased and deconflicted. One review will do, IMO.

PR #4821, opened 2022-04-13
Scan: make nested workflows visible.

hjoliver:
This is a small change with no associated Issue. Supersedes #4477 

Allow `cylc scan` to descend into nested workflow directories with names that start with `sub-`.

**Question:** ~~is the `sub-` prefix the best way to prevent a `scan` apocalypse (i.e. to avoid scanning the entire cylc-run tree)?  Another option is a `subworkflows/` sub-directory for all nested workflows,  but that adds an extra level in the hierarchy that the UIs have to deal with.~~

[UPDATE] I'm removing the question label unilaterally so I can finish this off by adding tests. IMO the `sub-` prefix satisfies the following requirements:
- is reasonably descriptive, without being long, which would adversely affect the gscan UI
- and does not add an extra useless level in the hierarchy that the UI has to deal with

(Easily tweaked if others have a better idea, but in lieu of that I'm forging ahead).

NOTE *This PR makes nested subworkflows possible, but it doesn't do anything to specifically encourage or support them*.

```console
~/cylc-src $ tree hydro/   
hydro/
├── flow.cylc
└── sub-wf
    └── flow.cylc
```

**Requirements check-list**

### DIFF
--- a/cylc/flow/network/scan.py
+++ b/cylc/flow/network/scan.py
@@ -80,8 +80,7 @@ from cylc.flow.workflow_files import (
 SERVICE = Path(WorkflowFiles.Service.DIRNAME)
 CONTACT = Path(WorkflowFiles.Service.CONTACT)
 
-FLOW_FILES = {
-    # marker files/dirs which we use to determine if something is a flow
+WORKFLOW_DIR_INDICATORS = {
     WorkflowFiles.Service.DIRNAME,
     WorkflowFiles.FLOW_FILE,  # cylc8 flow definition file name
     WorkflowFiles.LOG_DIR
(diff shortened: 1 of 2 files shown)


hjoliver:
Here's what `cylc scan` sees on this branch, after installing the example above and running it with `cylc play hydro/run1/sub-wf`:

```
$ cylc scan --state=stopped
hydro/run1
hydro/run1/sub-wf
```

oliver-sanders:
Looks fine to me but might as well merge with #4822 (drafted) as I don't think there's any point having it by itself?

hjoliver:
Actually, as discussed on Element chat:
- logically, sub-workflow instances are just internally complex tasks of the main workflow and therefore their run directories should be nested inside the main one
- BUT they should really be nested inside the main task's work directory, which unfortunately makes the nesting structure deeper and uglier

So ... I have to admit that doesn't leave a whole lot of advantage to nesting like this.  Instead, for now, I'll aim for scripted support for sub-worfkows in the old side-by-side structure.

Issue #4582, opened 2022-01-17
xtriggers: use forgign task id for workflow_state xtriggers


oliver-sanders:
`workflow_state` xtriggers allow us to trigger off of outputs in another workflow:

```ini
[scheduling]
    initial cycle point = 2010
    [[xtriggers]]
         upstream = workflow_state(workflow=workflow_id, task=foo, point=%(point)s, \
            message='data ready'):PT10S
   [[graph]]
        P1Y = """
            @upstream => whatever
        """
```

The data store maintains a list of graph edges, these edges have IDs composed of the tasks they involve e.g. `1/@trig|1/b`.

At the moment the edge IDs for `workflow_state` xtriggers reference the xtrigger rather than the task the xtrigger is pointing at.

I think we should change this so the xtrigger acts as an alias pointing at the foreign task using it's absolute ID (e.g. `~user/workflow//cycle/task`). This would allow the UI to follow edges from one workflow to another allowing us to graph a collection of inter-dependent workflows.

> **Note:** Would need to ensure that egde IDs can are safely parsed whether relative to the current workflow (as at the moment) or absolute (as this change would introduce the potential for) in both cylc-flow and cylc-ui.

> **Note:** Edge IDs must be unique to enter in the store so it might be necessary to check for duplication in xtrigger definitions for safety.

**Additional context**

**Pull requests welcome!**

PR #4822, opened 2022-04-13
Support nested subworkflows.

hjoliver:
REQUIRES #4821 
Supersedes #4477

A new job shell function to make it easy for users to run and manage subworkflows defined in sub-directories of the main workflow's source directory.
- these get installed with the main workflow (just like any main workflow source file)
- the installed subworkflow source is a template to create each subworkflow instance on the fly (one for each main workflow cycle point)
- the nesting is convenient for:
   - installing the subworkflow template automatically as part of the main workflow 
   - intuitive grouping under the main workflow
   - targeting main and subworkflows together, with cylc commands
 

NOTE:
- no change to any scheduler code (but relies on #4821)
- now **detaching** subworkflows, so (I think) it can be made to work with workflow migration

**Requirements check-list**

### DIFF
--- a/cylc/flow/etc/job.sh
+++ b/cylc/flow/etc/job.sh
@@ -356,3 +356,41 @@ cylc__job__dummy_result() {
         return 1
     fi
 }
+
+cylc__job__subworkflow() {
+    # Create and play a new instance of a subworkflow, from its installed
+    # source in the main workflow run dir.
+    local NAME="$1"
+    local DONE="$2"  # last task instance in sub-workflow, e.g. "1/foo"
+    if [[ "$NAME" != "sub-"* ]]; then
+        >&2 echo "ERROR: subworkflow name must begin with 'sub-'"
+        exit 1
+    fi
+    if [[ "$DONE" != *"/"* ]]; then
+        >&2 echo "ERROR: subworkflow final task must be 'point/name'"
+        exit 1
+    fi
+    # Set subworkflow paths and ID (includes main cycle point):
+    local SRC_DIR="${CYLC_WORKFLOW_RUN_DIR}/${NAME}"
+    local RUN_DIR="${SRC_DIR}/${CYLC_TASK_CYCLE_POINT}"
+    local ID="${RUN_DIR#*/cylc-run/}"
+    mkdir "$RUN_DIR"
+
+    # Symlink to the installed flow.cylc after renaming it to avoid
+    # detection of the template as a workflow (if not already renamed).
+    mv "${SRC_DIR}/flow.cylc" "${SRC_DIR}/sub-flow.cylc" || true
+    ln -s "${SRC_DIR}/sub-flow.cylc" "${RUN_DIR}/flow.cylc"
+
+    cylc message "Subworkflow: installed $ID from $SRC_DIR"
+
+    # Start the new sub-workflow instance (detaching).
+    cylc play "$ID"
+
+    # Wait for the subworkflow final task (allows subworkflow to stop and restart).
+    cylc message "Subworkflow: waiting for ${ID}//${DONE}"
+    cylc workflow-state \
+        --max-polls=10 --interval=10 \
+	-p "${DONE%/*}" -t "${DONE#*/}" --status succeeded "$ID"
+
+    cylc message "Subworkflow: finished ${ID}//${DONE}"
+}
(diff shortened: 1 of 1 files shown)


hjoliver:
**Working example**

```console
$ tree hydro/
hydro/
├── flow.cylc
└── sub-wf
    └── flow.cylc
```

Main workflow:
```ini
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    [[graph]]
        P1 = "foo => sub => bar"
[runtime]
    [[foo, bar]]
        script = sleep 10
    [[sub]]
        script = cylc__job__subworkflow sub-wf 1/post
```

Sub-workflow:
```ini
[scheduling]
   [[graph]]
       R1 = "pre => proc => post"
[runtime]
   [[pre, post, proc]]
       script = "sleep $((10 + RANDOM % 10))"
```

hjoliver:
Easy management and housekeeping, e.g.:

```
$ cylc stop hydro/run1/*  # stop hydro/run1 and any subworkflows
$ cylc clean hydro/run1/*  # clean up subworkflows of hydro/run1
```

hjoliver:
**Limitations**

Subworkflow instances are not installed with `cylc install` (the template they are created from is, with the main workflow), so:
- their locations aren't individually configurable via symlink-dirs (however, the whole run-directory will be wherever the main run directory is symlinked to)
- installation hooks are bypassed, so subworkflows can't use `rose-suite.conf`...
- other?

oliver-sanders (commented):
Sub workflows are a tricky problem, I completely understand (and support) the desire (I have use cases too), however, I don't think we can "neatly" package them as a solution at this point in time. We would need to develop much more advanced integration until which time users will need to be aware of the implementation details in order to understand how to work with them. For example what do the task statuses mean, how do you shut the workflow(s) down, how to identify/handle graph execution issues in sub-workflows, etc.

So I would be inclined to put this in the docs rather than implement it in the job script until we're ready to make this a formal & supported Cylc feature. Anyone who uses this approach really needs to understand what it is doing. The long-term solution will definitely require a different interface so this isn't an approach we can iterate to perfection.

The difference between `parent, parent/sub` and `parent/controller, parent/sub` is marginal. With UID `'parent/*'` allows them to be controlled as a block either way around so there isn't really any functional difference.

On the implementation side here are some pros and cons for the detatch/no-detach implementation detail:

#### No-Detach

pros:
* task:running means sub-workflow running.
* task:succeed means sub-workflow:succeeded which provides a handy shortcut for sub-workflow:finish-all which is otherwise hard to do (although if "all" tasks are expected to succeed `root:succeed-all => fin` is a valid approach)
* `cylc stop --kill` and `cylc kill` will work on the sub-workflows, kinda (workflow [events] could potentially niceify this sort of thing).
* Potentially compatible-ish with auto retries and `cylc trigger` (which could resume the workflow run).

cons:
* Workflow exit codes don't necessarily match data outcome expectations.
* Not great for branched workflows.
* Blocks workflow migration.

#### Detach:

pros:
* Permits workflow migration.
* Helps prevent workflow servers getting inundated with sub-workflows by allowing load balancing to kick in.

cons:

* task:running means sub-workflow task has not yet reached the desired state, however, does not convey whether the sub-workflow is able to reach that state at all (e.g. it could have stalled, or the task failed, this state will *not* make it back to the parent workflow even if the sub-workflow is configured to shutdown in this eventuality).
* `cylc stop`, `cylc stop --kill` and `cylc kill` don't really do what the user wants/expects.

#### Some problems to consider:

1. How to differentiate "restarting" a sub-workflow from "re-running" a sub-workflow.
2. How to feed sub-workflow info back to the UI (short to mid term), if we use xtriggers then #4582 would work but we can't achieve this with a standalone task (task metadata would be an option but it is static).
3. How to differentiate sub-workflow not is running / stalled / whatever from targetted task(s) have not reached the desired state yet.
4. How to integrate with `cylc install` features, perhaps the sub-workflow installation should be implemented within `cylc install` (special logic for sub-workflows)
   * symlink-dirs
   * run numbers
   * easy reinstallation?

hjoliver:
Note my final commit overlapped your review, and tweaked the model a bit to report not-running+not-finished as failed. 

Some of the concerns above are arguably not important if sub-workflows are (as they really should be) treated as a monolithic task in the main workflow (and of course are appropriate for use in that way) ... in which case, if the sub-workflow fails or stops without finishing for any reason, then it is "failed" and retriggering it in the main workflow should run it again from scratch (although I was experimenting with restarts too without commenting on that).

However, I generally agree. Particularly that anyone using this approach must understand exactly what the implications are) and had also been thinking about the various points above.

Plus: https://github.com/cylc/cylc-flow/pull/4821#issuecomment-1119081989

And: this shell function is not really Cylc-integrated in anyway so I can just go for a workflow bin script for now, for my current sub-workflow users.  So, closing this!!!!

Issue #4786, opened 2022-03-30
terminology: straighten out usage of "flow"


oliver-sanders:
Since we originally renamed this project to "Cylc Flow" our usage of "flow" has changed.

tldr:

* "Workflow" is the thing configured by the `flow.cylc` file.
* "Flows" are logical runs of the workflow graph.

> Some related context:
>
> * https://github.com/cylc/cylc-flow/issues/3817#issuecomment-717206611
> * https://github.com/cylc/cylc-doc/pull/224
> * https://github.com/cylc/cylc-flow/issues/4686

We have mostly straightened out the terminology now but a few things remain:

* [x] The `flow.cylc` file
  * Should really be `workflow.cylc`, we shortened it to `flow.cylc` to be shorter.
  * **Suggest we live with this**, it's not that big an issue.
  * The `flow.cylc` defines the graph the "flows" run so it does make sense.
* [x] `log/flow-config`
  * I think we rejected the name `workflow-config`.
  * Given that the filename is `flow.cylc` perhaps we should **keep this the way it is** too.
* [x] `cylc install --flow-name` (#4823)
  * See https://github.com/cylc/cylc-flow/issues/4741#issuecomment-1082768766
  * This is now quite confusing, especially as we can now add metadata to flows.
  * **Suggest changing to `--id`.**
* [x] `--reflow`
  * **Agreed to retire** term with #4686

Questions:

1) Anything missing from the list?
2) Propose leaving the first two (flow.cylc and flow-config) as is, but changing the latter two. Agreed?

**Pull requests welcome!**

dpmatthews:
> Suggest changing to `--id`

That would be confusing as well, see https://github.com/cylc/cylc-flow/pull/4455.
I'd suggest `--workflow-name` with `-n` as a short form.

oliver-sanders:
> That would be confusing as well, see https://github.com/cylc/cylc-flow/pull/4455.

Yes, the `--id` is only the "full" ID when used in combination with `--no-run-name`, however, we could document that away e.g. `--id, --id-prefix`.

dpmatthews:
> I'd prefer /log/workflow-config for this. We probably rejected that name in the past before the word flow got more specialized? It's not inconsistent with the first point: flow.cylc is the workflow config file; but flow-config is too close to the English (i.e. "flow config" sounds like config for a specific flow, which it isn't). Also, no one will be relying on the current filename yet.

Discussion on this point moved to #4802

[...]

PR #4816, opened 2022-04-11
Disallow direct validation (etc.) of source FILE

hjoliver:
These changes close #4615 (and supersede #4810)

Disallow the following:
```console
$ cylc val ./flow.cylc
$ cylc val ./test/flow.cylc
$ cylc val /path/to/flow.cylc
```

Additionally, I've tried to make the various error messages less confusing:
- check explicitly for a flow config filename on the command line
  -  this will be a common user error, and there's no need to report it as a bad workflow ID
- distinguish between "source path does not exist" and "workflow ID/path does not exist", when that is possible.

```console
$ cylc val hydr/flow.cylc
UserInputError: Not a valid workflow ID or source directory: hydr/flow.cylc

$ cylc val ./hydr/flow.cylc
UserInputError: Not a valid workflow ID or source directory: hydr/flow.cylc

$ cylc val ./hydr  # valid source
Valid for cylc-8.0rc3.dev

$ cylc val hydr  # valid ID
Valid for cylc-8.0rc3.dev

$ cylc val ./no/exist 
UserInputError: Source directory not found: /home/oliverh/cylc-src/no/exist

$ cylc val no/exist 
UserInputError: Workflow ID not found: no/exist
(Directory not found: /home/oliverh/cylc-run/no/exist)
```

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.
- [x] Appropriate tests are included (unit).
- [x] No change log entry required (why? e.g. invisible to users).

- [ ] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/XXXX.

PR #4823, opened 2022-04-13
Remove `--directory` option for `cylc install`

### LINKED ISSUE #4676: Replace `cylc install --directory` with $1
#### Current Cylc install behaviour
```bash
# Look for myflow relative to each path in global.cylc[install]source dirs and install the first match
cylc install myflow
# Install from somewhere else
cylc install --directory <path to flow>
```

#### Proposed change 
- Remove `--directory` as an option.
- Detect what `<flow>` is:
  1. Absolute path
  1. Explicit relative path (e.g. `./foo/bar`) means relative to PWD
  1. Implicit relative path (e.g. `foo/bar`) means relative to `global.cylc[install]source dirs`

#### Reasoning
- `cylc install` works in a way inconsistent with the rest of the CLI - e.g. `cylc validate .`
- `--directory` is unnecessarily complicated.

#### Questions
1. What is the default flow name for a path that is absolute or explicit relative?
    - Basename (inconsistent with decision to drop #4643 )
    - ~Full path name (Horrible)~
    - ~Just don't allow paths without `--flow-name`~
2. How do we tell apart "relative to `global.cylc[install]source dirs`" and "implicit relative"? What happens if both are present?
    - Answer: if an implicit relative path is supplied, take that to mean relative to `global.cylc[install]source dirs`. Don't even check in PWD

[edit: Crossed out alternative based on https://matrix.to/#/!hwZqSYihGPuhDdIzIP:matrix.org/$1644571596108594MybII:matrix.org?via=matrix.org]

#### Unchanged behaviour

```
$ cylc install
```
should still work (installs from PWD)

#### Additional context
From a discussion on https://github.com/cylc/cylc-flow/issues/4643

**Pull requests welcome!**

MetRonnie:
These changes close #4676 and partially address #4786

Remove `--directory` option for `cylc install` and instead allow any directory in the source arg (`cylc install [SOURCE]`):
- Implicit relative path (e.g. `foo/bar`): look for source relative to each path in `global.cylc[install]source dirs` and install first match
- Explicit relative path (e.g. `./foo/bar`): source is relative to PWD
- Absolute path (e.g. `/home/user/foo/bar`): source is at that path
- No path specified: source is PWD 

Rename `--flow-name` option to `--workflow-name` and add `-n` for short (the consensus in #4786 seemed to be `--workflow-name` is preferable to `--id`).

**Requirements check-list**

### DIFF
--- a/cylc/flow/pathutil.py
+++ b/cylc/flow/pathutil.py
@@ -28,9 +28,19 @@ from cylc.flow.exceptions import (
 )
 from cylc.flow.platforms import get_localhost_install_target
 
+
 # Note: do not import this elsewhere, as it might bypass unit test
 # monkeypatching:
 _CYLC_RUN_DIR = os.path.join('$HOME', 'cylc-run')
+
+EXPLICIT_RELATIVE_PATH_REGEX = re.compile(
+    rf'''
+    ^({re.escape(os.curdir)}|{re.escape(os.pardir)})
+    ({re.escape(os.sep)}|$)
+    ''',
+    re.VERBOSE
+)
+"""Matches relative paths that are explicit (starts with ./)"""
 
 
 def expand_path(*args: Union[Path, str]) -> str:
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -562,7 +562,7 @@ def get_flow_file(reg: str) -> Path:
 def get_workflow_source_dir(
     run_dir: Union[Path, str]
 ) -> Union[Tuple[str, Path], Tuple[None, None]]:
-    """Get the source directory path of the workflow in directory provided.
+    """Get the source directory path for a given workflow run directory.
 
     Args:
         run_dir: directory to check for an installed flow inside.
@@ -1453,19 +1453,20 @@ def get_rsync_rund_cmd(src, dst, reinstall=False, dry_run=False):
     return rsync_cmd
 
 
-def reinstall_workflow(named_run, rundir, source, dry_run=False):
+def reinstall_workflow(
+    source: Path,
+    named_run: str,
+    rundir: Path,
+    dry_run: bool = False
+) -> None:
     """Reinstall workflow.
 
     Args:
-        named_run (str):
-            name of the run e.g. my-flow/run1
-        rundir (path):
-            run directory
-        source (path):
-            source directory
-        dry_run (bool):
-            if True, will not execute the file transfer but report what would
-            be changed.
+        source: source directory
+        named_run: name of the run e.g. my-flow/run1
+        rundir: run directory
+        dry_run: if True, will not execute the file transfer but report what
+            would be changed.
     """
     validate_source_dir(source, named_run)
     check_nested_dirs(rundir)
@@ -1489,12 +1490,11 @@ def reinstall_workflow(named_run, rundir, source, dry_run=False):
     reinstall_log.info(f'REINSTALLED {named_run} from {source}')
     print(f'REINSTALLED {named_run} from {source}')
     close_log(reinstall_log)
-    return
 
 
 def install_workflow(
+    source: Path,
     workflow_name: Optional[str] = None,
-    source: Optional[Union[Path, str]] = None,
     run_name: Optional[str] = None,
     no_run_name: bool = False,
     cli_symlink_dirs: Optional[Dict[str, Dict[str, Any]]] = None
@@ -1506,8 +1506,8 @@ def install_workflow(
     work, log, share, share/cycle directories.
 
     Args:
+        source: absolute path to workflow source directory.
         workflow_name: workflow name, default basename($PWD).
-        source: directory location of flow.cylc file, default $PWD.
         run_name: name of the run, overrides run1, run2, run 3 etc...
             If specified, cylc install will not create runN symlink.
         rundir: for overriding the default cylc-run directory.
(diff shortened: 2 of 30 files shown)


cylc/flow/pathutil.py
```diff
@@ -28,10 +28,20 @@
...
 _CYLC_RUN_DIR = os.path.join('$HOME', 'cylc-run')
 
+EXPLICIT_RELATIVE_PATH_REGEX = re.compile(
+    rf'''
+    ^({re.escape(os.curdir)}|{re.escape(os.pardir)})
+    ({re.escape(os.sep)}|$)
+    ''',
+    re.VERBOSE
+)
+"""Matches relative paths that are explicit (starts with ./)"""
```
hjoliver, line 43: Just curious, is this path format officially called "explicit relative"? If you invented the term, I quite like it :+1:  Although arguably it's really an absolute path.

cylc/flow/workflow_files.py
```diff
@@ -1527,13 +1528,10 @@ def install_workflow(
             Another workflow already has this name (unless --redirect).
             Trying to install a workflow that is nested inside of another.
     """
-    if not source:
-        source = Path.cwd()
-    elif Path(source).name == WorkflowFiles.FLOW_FILE:
-        source = Path(source).parent
-    source = Path(expand_path(source))
+    if source.name == WorkflowFiles.FLOW_FILE:
+        source = source.parent
```
hjoliver, line 1532: In light of #4816 this should be treated as an error?

hjoliver (approved):
Looks good :+1:

cylc/flow/pathutil.py
MetRonnie: I can't seem to find any terminology for relative paths for differentiating between whether they start with `./` or not. Explicit and implicit seems to be something one of us (maybe Tim, as author of the issue #4676) came up with.

Maybe we should avoid calling relative paths that don't start with `./` as relative paths at all; maybe calling them workflow source names would be less confusing? The arg could then be `SOURCE_NAME | PATH` where `PATH` can be relative or absolute but must start with `./` if relative.

tests/functional/cylc-install/00-simple.t
```diff
@@ -148,21 +130,3 @@ INSTALLED $RND_WORKFLOW_NAME/run2 from ${RND_WORKFLOW_SOURCE}
...
-__OUT__
-TEST_NAME="${TEST_NAME_BASE}-install-C-twice-2"
-run_ok "${TEST_NAME}" cylc install -C "${RND_WORKFLOW_NAME}"
-contains_ok "${TEST_NAME}.stdout" <<__OUT__
-INSTALLED $RND_WORKFLOW_NAME/run2 from ${RND_WORKFLOW_NAME}
-__OUT__
-purge_rnd_workflow
-
-
-exit
```
datamel, line 168: Is this exit not needed? Seems like the bulk of our tests (but not all) have it, guessing it is to ensure if error occurs then it exits properly.

datamel (approved):
Thanks @MetRonnie.  One small query from me.

MetRonnie:
Pinging @cylc/core for any input on the unresolved reviewers' questions above

cylc/flow/workflow_files.py
MetRonnie: It might make sense to do so in #4816. Whichever PR goes in first will cause a merge conflict for the second (so it will be up to the author of the second to fix).

cylc/flow/pathutil.py
hjoliver: The trouble is that conflicts with standard unix terminology, where a relative path doesn't have to start with `./`.  

IMO `./blah` and `../blah` should really be considered absolute paths, where `.` is short for `$PWD` which expands to an absolute path. Like `~/blah` is an absolute path even though it is relative to `$HOME`.  But, that doesn't seem to be the prevailing opinion out there.

tests/functional/cylc-install/00-simple.t
hjoliver: I'm pretty sure an exit statement with no argument at the end of a script is useless. And annoying.  If no one can think of a good reason for them, let's remove them.

cylc/flow/workflow_files.py
hjoliver: Addressed by last commit in #4816

hjoliver:
Two approvals, and the unresolved questions above are either now resolved or don't require changes to this branch.

cylc/flow/pathutil.py
oliver-sanders: > If you invented the term

MB but I think it's valid and helpful here.

> The trouble is that conflicts with standard unix terminology
> IMO ./blah and ../blah should really be considered absolute paths

I don't think that's correct, technically `.` and `..` *are* relative paths implemented at the FS level and visible in FS listings. So `.` is *not* short for `$PWD` though it happens to *resolve* to the same thing.

* `$PWD/foo` - the `$PWD` is *expanded* by the shell.
* `./foo` - the `./` is *resolved* by the FS.

hjoliver: Technically true, but I doubt that matters to users.

BTW I didn't mean that `.` is *literally* short for `$PWD` and, like the shell variable, expanded by the shell ... just that it is short for the present working directory.  A true relative path is implicit. Presumably the main reason for the existence of `..` is to make changing up a level easier, and `.`, to force the behaviour of programs that don't interpret (implicit) relative paths properly.  So I'd still argue that `rel/path` is a proper relative path, but `./rel/path` is kind of an absolute path masquerading as a relative path.

oliver-sanders: Explicit relative and absolute paths are handled the same by the CLI so this disparity is not user facing. This is internal technical documentation in a docstring.

hjoliver: (Which still needs to be understood at a glance by future developers).

PR #4824, opened 2022-04-13
schema: fix formatting in docstring

oliver-sanders:
The docstring matters here because it gets fed through to the UI via API-On-The-Fly.

**Requirements check-list**


hjoliver (approved):
:+1:

MetRonnie:
This isn't working for me

hjoliver:
Huh. I recall doing a manual test that implied this would work...

...yep:
```console
(venv) foo@bar $ python -c 'from cylc.flow.network.schema import Flow; print(Flow.__doc__); print(Flow().__doc__)'
An integer or one of all, new or none.
An integer or one of all, new or none.
```

hjoliver:
Same result as you in the UI though.

PR #4827, opened 2022-04-19
Move processing of stop cycle point to config.py

### LINKED ISSUE #4637: Giving an invalid `--stopcp` on start "corrupts" database
If you have a not yet run workflow and give an invalid value for `--stopcp` when starting it for the first time, after the workflow shuts down due to the error, you are left with a workflow database with empty tables.

Because the `workflow_params` table is empty, if you try to restart the workflow, Cylc thinks the database is corrupted and refuses to play ball.

```
ERROR - Workflow shutting down - ServiceFileError: Cannot restart - Workflow database is incompatible with Cylc 8.0rc1.dev, or is corrupted
```
And you can't clean the database because of #4450 
```
$ cylc clean myflow/ --rm .service
ServiceFileError: Cannot clean - Workflow database is incompatible with Cylc 8.0rc1.dev, or is corrupted
```

**Pull requests welcome!**

### LINKED ISSUE #4709: Fix stop after cycle point inconsistency with other cycle point options
(Re-written)

The final cycle point is only stored in the DB if specified via command line option `--fcp`, whereas the stop after cycle point is stored in the DB regardless of whether it was specified via command line option or flow.cylc.

This means changing the final cycle point in flow.cylc and reinstalling + restarting will use the new point, but doing the same for stop after cycle point will ignore the new point. Why should this be the case?

MetRonnie:
These changes close:
- #4709 
- #4637

**Requirements check-list**
- [x] Appropriate tests are included (unit, integration and functional!)
- [x] Docs: https://github.com/cylc/cylc-doc/pull/459


cylc/flow/scheduler_cli.py
```diff
@@ -147,7 +147,6 @@ def get_option_parser(add_std_opts=False):
             "Set the final cycle point. "
             "This command line option overrides the workflow "
             "config option '[scheduling]final cycle point'. "
-            "Use a value of 'reload' to reload from flow.cylc in a restart."
```
MetRonnie, line 150: Removed this as it implies changes to flow.cylc will not be picked up normally; not the case; see https://github.com/cylc/cylc-doc/pull/459

wxtim (approved):
- [x] Read
- [x] Manually tested locally

There are a couple of Codecov holes. They aren't too bad IMO, but it would be nice if you could fix them.

cylc/flow/config.py
```diff
@@ -693,10 +695,9 @@ def process_final_cycle_point(self) -> None:
         Raises:
             WorkflowConfigError - if it fails to validate
         """
-        if (
-            self.cfg['scheduling']['final cycle point'] is not None and
-            not self.cfg['scheduling']['final cycle point'].strip()
-        ):
+        if self.cfg['scheduling']['final cycle point'] == '':
```
oliver-sanders, line 698: The `.strip()` may or may not have been helpful, if you're totally sure it wasn't needed that's fine, otherwise perhaps:

```suggestion
        if not self.cfg['scheduling']['final cycle point'].strip():
```

cylc/flow/cycling/iso8601.py
```diff
@@ -897,13 +897,13 @@ def _interval_parse(interval_string):
...
-def point_parse(point_string):
+def point_parse(point_string: str) -> 'TimePoint':
     """Parse a point_string into a proper TimePoint object."""
-    return _point_parse(point_string)
+    return _point_parse(point_string, WorkflowSpecifics.DUMP_FORMAT)
 
 
 @lru_cache(10000)
-def _point_parse(point_string):
+def _point_parse(point_string, _dump_fmt):
```
oliver-sanders, line 906: `_dump_fmt` arg does nothing?

tests/unit/test_config.py
```diff
@@ -473,19 +457,30 @@ def test_process_startcp(
...
+            {
+                'initial cycle point': '2021',
+                'final cycle point': '',
+            },
+            None,
+            None,
+            None,
+            id="Empty fcp in cfg"
+            # Need to check this as fcp treated as string by parsec, unlike
+            # other cycle point settings
```
oliver-sanders, line 477: TODO?

oliver-sanders (approved):
Looks good, yet to manually test start/restart/reload'ing the stop cycle manully but looks like you have done that extensively in the automated tests.

cylc/flow/config.py
MetRonnie: I tested out
```
final cycle point = "       "
```
and put a breakpoint at this line; the value of `self.cfg['scheduling']['final cycle point']` was just `''`. Presumably parsec  handles stripping (as I would hope so)

cylc/flow/cycling/iso8601.py
MetRonnie: It is needed for the caching to be valid, seeing as we rely on the `WorkflowSpecifics.DUMP_FORMAT` inside the function. This is how it is done in isodatetime too: https://github.com/metomi/isodatetime/blob/abec7c8ff05ab19d06f5eda790422df54ca72a8b/metomi/isodatetime/data.py#L2174-L2184

tests/unit/test_config.py
MetRonnie: "Need to check this as" == "This test is needed because"

cylc/flow/cfgspec/workflow.py
```diff
@@ -695,6 +690,11 @@ def get_script_common_text(this: str, example: Optional[str] = None):
                The runahead limit may be automatically raised if this is
                necessary to allow a future task to be triggered, preventing
                the workflow from stalling.
+
+            .. versionchanged:: 8.0.0
+
+               The deprecated ``[scheduling]max active cycle points`` setting
```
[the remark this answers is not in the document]
MetRonnie: But it is deprecated? https://github.com/cylc/cylc-flow/blob/9ed6035f2c8f02cf09a5672692b25ebf604cd7e3/cylc/flow/cfgspec/workflow.py#L1709-L1714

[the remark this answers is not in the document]
oliver-sanders: The setting is deprecated, however, the functionality is not.

```
# deprecated
max active cycle points = 4

# fine
runahead limit = P4
```

Because we had previously marked `runahead limit` as deprecated this has caused some confusion.

cylc/flow/cycling/iso8601.py
oliver-sanders: But this argument is not used within the function.

oliver-sanders: Dunno if `_dump_fmt=WorkflowSpecifics.DUMP_FORMAT` would work with this mutable attribute?

MetRonnie: If my understanding is correct, `dump_fmt=WorkflowSpecifics.DUMP_FORMAT` gets evaluated at time of function definition (when it is still `None`), not at time of call, so that would not be helpful.

IMHO, it's a "private" (leading underscore) argument in a "private" function, so it's an acceptable solution for the time being

datamel (approved):
Checked out and tested. No problems spotted with the implementation. Thanks @MetRonnie.

hjoliver (approved):

cylc/flow/config.py
```diff
@@ -693,10 +695,9 @@ def process_final_cycle_point(self) -> None:
...
+            # (Unlike other cycle point settings in config, fcp is treated as
+            # a string by parsec)
```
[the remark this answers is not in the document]
MetRonnie: Done

PR #4828, opened 2022-04-20
scan CLI: handle bad contact files.

hjoliver:

`cylc scan` currently fails if it encounters a corrupt contact file.

To test this, e.g.:
```console
$ touch ~/cylc-run/blah/.service/contact
$ cylc scan  # ... all hell breaks loose
```

**Requirements check-list**

### DIFF
--- a/cylc/flow/scripts/scan.py
+++ b/cylc/flow/scripts/scan.py
@@ -405,9 +405,17 @@ async def _tree(pipe, formatter, opts, write):
     async for flow in pipe:
         ret.append(flow)
 
-    # construct tree
     tree = {}
-    for flow in sorted(ret, key=lambda f: f['name']):
+    _construct_tree(ret, tree, formatter, opts, write)
+
+    # print tree
+    ret = get_tree(tree, '', sort=False, use_unicode=True)
+    write('\n'.join(ret))
+
+
+def _construct_tree(flows, tree, formatter, opts, write):
+    """Construct the tree, for given flows and formatter."""
+    for flow in sorted(flows, key=lambda f: f['name']):
         parts = Path(flow['name']).parts
         pointer = tree
         for part in parts[:-1]:
@@ -422,10 +430,6 @@ async def _tree(pipe, formatter, opts, write):
         if len(parts) > 1:
             item = f' {item}'
         pointer[item] = ''
-
-    # print tree
-    ret = get_tree(tree, '', sort=False, use_unicode=True)
-    write('\n'.join(ret))
 
 
 def get_pipe(opts, formatter, scan_dir=None):
--- a/cylc/flow/scripts/scan.py
+++ b/cylc/flow/scripts/scan.py
@@ -51,7 +51,6 @@ Examples:
 
 import asyncio
 import json
-import sys
 from pathlib import Path
 from typing import Callable, Optional, TYPE_CHECKING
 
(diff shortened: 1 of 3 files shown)


hjoliver:
Small bug fix, so assigning to the rc3 milestone.

cylc/flow/scripts/scan.py
```diff
@@ -278,6 +282,23 @@ def state_totals_key():
...
 
+def format_and_write(write, formatter, *f_args, **f_kwargs):
+    """Wrapper for the various stdout formatters.
+
+    Format and write out scanned info, or warn of corrupted contact files
+    (which have correct name and path, but are missing required content).
+    """
+    try:
+        write(formatter(*f_args, **f_kwargs))
+    except KeyError:
```
oliver-sanders, line 293: This is catching a bit more than just missing contact file fields, I think the formatters only use host & port from the contact file, might worth shifting the try/except down a level or possibly even `.get(..., '?')`.

oliver-sanders (approved):
Good catch, how did you end up with a corrupted contact file?

Might be a good idea to make the error catching more targetted, otherwise ok.

Looks like there is a similar issue with the UIS :( - https://github.com/cylc/cylc-uiserver/issues/343

hjoliver:
> Good catch, how did you end up with a corrupted contact file?

I tried `touch`ing a contact file while experimenting with nested worklows.

cylc/flow/scripts/scan.py
```diff
@@ -395,14 +416,25 @@ async def _tree(pipe, formatter, opts, write):
...
         for part in parts[:-1]:
             if part not in pointer:
                 pointer[part] = {}
             pointer = pointer[part]
         flow['name'] = parts[-1]
-        item = formatter(flow, opts)
+        try:
+            item = formatter(flow, opts)
+        except KeyError:
+            print(
```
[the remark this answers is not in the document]
hjoliver: Will do.

cylc/flow/scripts/scan.py
hjoliver: I was trying to avoid doing this in each individual formatter.

Do you mean the `write` function could potentially raise `KeyError`?  Or, I suppose the "rich" formatter could potentially raise `KeyError` for non contact-file reasons (missing data fields from the server).

oliver-sanders: > for non contact-file reasons (missing data fields from the server).

Yeah, that sort of thing, in which case the real error would get swallowed and masked by a contact file error.

hjoliver:
OK, I've moved the error detection to a lower level.  

I haven't upped the test coverage much as I ran out of time after backing out of some other changes in the `network.scan` module.

hjoliver:
@datamel - the last commit ups the test coverage quite a bit, good enough for this change, I think.

datamel (approved):
I have read the code, checked out, manually tested and run automated tests locally.
My only query is if we want to remove the contact file, having established that it is a dud?
Approving in any case, since this can be done at a later stage.

oliver-sanders:
> My only query is if we want to remove the contact file, having established that it is a dud?

These "duds" are something which shouldn't be possible normally (we haven't seen them in the wild), however, there is one likely way in which we could get them, if we go fiddling with the contact file in a future version of Cylc. In which case we really wouldn't want the older version of Cylc to go and remove them for us.

The contact file does contain the API version number for the ZMQ server, however, that's no use if you can't load the file in the first place.

PR #4829, opened 2022-04-20
config: suppress config deprecation warnings in compat mode

oliver-sanders:
Quickie spotted during support:

* Configuration deprecation warnings are currently displayed in Cylc 7 compatibility mode.
* Compatibility mode is expected to be used by Cylc 7/8 compatible workflows during this transition period.
* If users were to follow these warnings they would make their workflows incompatible with Cylc 7.
* This change suppresses these deprecation (and only deprecation) warnings in Cylc 7 compat mode.

**Requirements check-list**
- [x] Documentation: cylc/cylc-doc#460


cylc/flow/config.py
```diff
@@ -786,7 +786,6 @@ def _check_implicit_tasks(self) -> None:
             not cylc.flow.flags.cylc7_back_compat
         ):
             raise WorkflowConfigError(msg)
-        LOG.warning(msg)
```
oliver-sanders, line 789: This log message says "to allow implicit tasks ..." which isn't appropriate in Cylc 7 compat mode (as implicit tasks are already allowed by exception), if users followed this advice they would make the workflow Cylc 7 incompatible.

MetRonnie (approved):
I think this needs a docs update, I am happy to work on that

MetRonnie:
(flake8 failing)

MetRonnie (changes requested):
Ah wait, will need to do something about
https://github.com/cylc/cylc-flow/blob/0c6af2437aa4c0a6e556bb0e2c747032f6f66830/cylc/flow/workflow_files.py#L342-L346

https://github.com/cylc/cylc-flow/blob/88b880ba5b43c3cf554de60718e4d149173f6196/cylc/flow/cfgspec/workflow.py#L108-L113

oliver-sanders:
Ach dammit, sorry, should read things better.

oliver-sanders:
FYI this changes the advice from "address warnings in compat mode" to "address warnings/errors after upgrading".

I think this makes more sense, it is hard (and possibly not really possible) to upgrade the graphing in compat mode because the graph execution logic is different?

cylc/flow/workflow_files.py
```diff
@@ -340,9 +340,7 @@ class RemoteCleanQueueTuple(NamedTuple):
 """
 
 SUITERC_DEPR_MSG = (
-    f"Backward compatibility mode ON for CYLC 7 '{WorkflowFiles.SUITE_RC}'"
-    " files: please address deprecation warnings and upgrade to Cylc 8 graph"
-    f" syntax BEFORE renaming the file to '{WorkflowFiles.FLOW_FILE}'.\n"
+    f"Backward compatibility mode ON for CYLC 7 '{WorkflowFiles.SUITE_RC}'."
```
hjoliver, line 343:
```suggestion
    f"Backward compatibility mode ON for CYLC 7 '{WorkflowFiles.SUITE_RC}' config files. "
    f"When ready to upgrade, rename the file to {WorkflowFiles.FLOW_FILE} then address "
    "any resulting validation errors and warnings."
```

hjoliver (approved):
Makes sense, but I think the back-compat mode warning should still explain how to upgrade. (A bigger warning is also a bit harder to ignore forever).

cylc/flow/workflow_files.py
MetRonnie: Maybe "any resulting validation ~errors and~ warnings" because there shouldn't be any errors, only warnings?

cylc/flow/cfgspec/workflow.py
```diff
@@ -107,9 +108,10 @@ def get_script_common_text(this: str, example: Optional[str] = None):
         .. versionchanged:: 8.0.0
 
            The configuration file was previously named ``suite.rc``, but that
-           name is now deprecated. Please take action on any deprecation
-           warnings before renaming ``suite.rc`` configuration files
-           to ``flow.cylc``.
+           name is now deprecated.
+
+           The ``suite.rc`` file name now activates :ref:`cylc_7_compat_mode`,
+           to turn off compatibility mode rename to ``flow.cylc``.
```
MetRonnie, line 114: Wording suggestion
```suggestion
           The ``suite.rc`` file name now activates :ref:`cylc_7_compat_mode`.
           Rename to ``flow.cylc`` to turn off compatibility mode.
```

MetRonnie:
I have opened a PR against this PR: https://github.com/oliver-sanders/cylc-flow/pull/54

> Is it worth warning about recurrence format 1 in Cylc 7 back compat mode, seeing as it's easy to do?

cylc/flow/workflow_files.py
oliver-sanders: > because there shouldn't be any errors

I think errors are quite likely with optional outputs e.g. `a => b; a:fail > c`.

Good enough?:

```suggestion
    f"Backward compatibility mode ON for Cylc 7 '{WorkflowFiles.SUITE_RC}'."
    f"\nTo upgrade to Cylc 8, rename {WorkflowFiles.SUITE_RC}"
    f" to {WorkflowFiles.FLOW_FILE}."
```

MetRonnie: Ah ok. I would go for Hilary's suggestion, as I think it's useful to give notice that the user will get warnings/errors

MetRonnie:
flake8 failing
```
./cylc/flow/workflow_files.py:343:80: E501 line too long (90 > 79 characters)
./cylc/flow/workflow_files.py:344:80: E501 line too long (88 > 79 characters)
```

cylc/flow/workflow_files.py
```diff
@@ -340,9 +340,9 @@ class RemoteCleanQueueTuple(NamedTuple):
 """
 
 SUITERC_DEPR_MSG = (
-    f"Backward compatibility mode ON for CYLC 7 '{WorkflowFiles.SUITE_RC}'"
-    " files: please address deprecation warnings and upgrade to Cylc 8 graph"
-    f" syntax BEFORE renaming the file to '{WorkflowFiles.FLOW_FILE}'.\n"
+    f"Backward compatibility mode ON for CYLC 7 '{WorkflowFiles.SUITE_RC}' config files. "
+    f"When ready to upgrade, rename the file to {WorkflowFiles.FLOW_FILE} then address "
```
oliver-sanders, line 344:
```suggestion
    "Backward compatibility mode ON for CYLC 7"
    f" '{WorkflowFiles.SUITE_RC}' config files."
    " When ready to upgrade, rename the file to"
    f" {WorkflowFiles.FLOW_FILE} then address "
```

oliver-sanders:
sorted (squash-merge me)

MetRonnie (approved):

PR #4832, opened 2022-04-21
Update `cylc install` usage text

MetRonnie:
This is a small change follow-up to #4823 

The usage text makes reference to "implicit" and "explicit" relative paths (e.g. `foo/bar` and `./foo/bar` respectively). These are self-coined and not standard terms, so I have removed reference to them in the user-facing text and split the `SOURCE` arg into two: `SOURCE_NAME | PATH`. Hopefully this will avoid confusion.

Pinging @cylc/core in case anyone wants to scream

**Requirements check-list**
- [x] Does not need tests
- [x] No change log entry required (follow-up to PR on same milestone)

### DIFF
--- a/cylc/flow/scripts/install.py
+++ b/cylc/flow/scripts/install.py
@@ -18,7 +18,7 @@
 
 """cylc install [OPTIONS] ARGS
 
-Install a new workflow.
+Install a workflow into ~/cylc-run.
 
 The workflow can then be started, stopped, and targeted by name.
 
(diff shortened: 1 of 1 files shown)


cylc/flow/scripts/install.py
```diff
@@ -30,19 +30,17 @@
...
+If SOURCE_NAME is supplied, Cylc will search for the workflow source in the
+list of directories given by "global.cylc[install]source dirs", and install
+the first match.
 
-If the SOURCE argument is not supplied, Cylc will install the workflow from
-the source in the current working directory.
+If PATH is supplied, Cylc will install the workflow from the source given
+by the path. Relative paths must start with "./" to avoid ambiguity with
+SOURCE_NAME (i.e. "foo/bar" will be interpreted as a source name, whereas
+"./foo/bar" will be interpreted as a path).
```
hjoliver, line 40: :+1: that's a good way to describe it.

hjoliver (approved):

PR #4837, opened 2022-04-25
cylc wrapper: Provide legacy support for rosie

dpmatthews:
The `rosie` command is not currently functional using Rose 2 so use use Rose 2019.01 for the moment


wxtim (approved):
- [x] Read
- [x] Tested locally

oliver-sanders (approved):

PR #4320, opened 2021-07-22
Better handling of Cylc 8 workflows

MetRonnie:
These changes close #4315

Give better errors and be safer with the contact file and DB when trying to use Cylc 7 commands on Cylc 8 workflows.

#### Manual tests I have tried

First create a **suite.rc** workflow that is compatible with both Cylc 7 and 8:
```ini
# ~/cylc-run/myflow/suite.rc
[cylc]
    cycle point format = %Y
[scheduling]
    initial cycle point = 2012
    [[dependencies]]
        [[[P1Y]]]
            graph = foo
[runtime]
    [[foo]]
```

Run in Cylc 8:
```shell
# cylc8 conda env
$ cylc play myflow -n --pause
```
Then open a new terminal for Cylc 7:

1. `cylc hold myflow`
2. `cylc scan --debug`
3. `cylc run myflow`
4. `cylc restart myflow`
5. `cylc stop myflow`
6. `cylc get-suite-contact myflow` (Should this work?)
7. `cylc get-suite-version` (I think this ought to work, but it doesn't and I'm not sure it's worth getting it to work. At least the error prints the version anyway!)
8. `cd some_other_flow && cylc register myflow`

(have I missed any important ones?)

Now stop the workflow:
```shell
# cylc8 conda env
$ cylc stop myflow
```

Then run those same Cylc 7 commands again.

#### Requirements check-list
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] (Some) appropriate tests are included (unit & functional).
- [x] Appropriate change log entry included.
- [x] No documentation update required.

PR #4838, opened 2022-04-27
Prevent Cylc 7 from running suite.rc workflow that was previously run with Cylc 8

### LINKED ISSUE #4835: cylc 7: check cylc 8 safety check still works
Double check that Cylc 7 won't mess with Cylc 8 backwards-compatible (`suite.rc`) workflows since the removal of the `flow.cylc -> suite.rc`.

In particular we don't want `cylc run` or `cylc restart` to mess with a Cylc 8 workflow.

Another command that could potentially cause issues is `cylc reg`.

MetRonnie:
These changes close #4835.

Fix bug where Cylc 7 would happily run a `suite.rc` workflow that was previously run with Cylc 8. (This bug came after we stopped doing the `flow.cylc -> suite.rc` symlink)

Do not run if:
- `flow.cylc` detected in run dir
- `_cylc-install` dir detected in run dir or 1 above
- Cylc 8 workflow DB detected

This PR does not do anything about:
- `cylc register` or the auto-register that happens on start. All this does is create a `.service/source` symlink, which is relatively harmless
- Creation of auth files on start. This is also relatively harmless.

**Testing**

Pytest

Functional tests I have run (`cylc test-battery`):
- `tests/cylc8`
- `tests/database`
- `tests/registration`
- `tests/restart`
- `tests/startup`

Manual tests to run (from #4320):

> First create a **suite.rc** workflow that is compatible with both Cylc 7 and 8
> 
> Run in Cylc 8:
> ```shell
> # cylc8 conda env
> $ cylc play myflow -n --pause
> ```
> Then open a new terminal for Cylc 7:
> 
> 1. `cylc hold myflow`
> 2. `cylc scan --debug`
> 3. `cylc run myflow`
> 4. `cylc restart myflow`
> 5. `cylc stop myflow`
> 6. `cylc get-suite-contact myflow`
> 7. `cylc get-suite-version` 
> 
> Now stop the workflow:
> ```shell
> # cylc8 conda env
> $ cylc stop myflow
> ```
> 
> Then run those same Cylc 7 commands again.

**Requirements check-list**
- [x] Appropriate tests are included
- [x] No docs update needed


wxtim (approved):
- [x] read.
- [x] Manually tested.

oliver-sanders (approved):
LGTM.

One thing I noticed is that Cylc 7 let me run a workflow which had been "installed" by Cylc 8.

Perhaps this should also be blocked, could/should we add a check for the existence of `_cylc-install`, `../_cylc-install`?

MetRonnie:
What if a user is trying to make use of the back-compat interoperability by installing from a newfangled source dir with different runs for Cylc 8 and 7? E.g. `--run-name c8`, `--run-name c7`

oliver-sanders:
The back compat is with the workflow definition not with the install process.

The `cylc register` and `rose suite-run` commands both worked in very different ways to `cylc install`. Running a workflow that was installed by Cylc 8 under Cylc 7 could be dangerous, especially for Rose workflows as the Rose config will be completely ignored by Rose 2019 (it was only loaded during installation) and remote tasks will fail or pick up an older installation (Rose performed remote installation during workflow installation).

PR #4842, opened 2022-04-28
jinja2: improve error context information

oliver-sanders:
Our Jinja2 error context reporting gets stuck on include statements.

A couple of our Cylc 8 trialers have hit Python 2->3 issues (e.g. `dict.keys()`) that have been obscured by this. I suspect a few more will run into similar issues. It's quite hard to debug these issues when all you see is `TypeError` and the line the `{% include` statement is on.

Took a look, seems an easy fix.

### Example

Should validate with Cylc 7 but fail with Cylc 8 due to Python the 2->3 change in `dict.keys` return type:

(flow.cylc)
```jinja2
#!Jinja2

{% include "foo.cylc" %}
```

(foo.cylc)
```jinja2
{% set foo = {"a": 1} %}
{% for key in foo.keys() + ["b"] %}
#    {{ key }}
{% endfor %}
```

Before this change only the `{% include` line would be highlighted, after this change the real error should also appear in the error.

**Requirements check-list**


cylc/flow/parsec/exceptions.py
```diff
@@ -94,10 +94,11 @@ def __str__(self):
         if self.line:
             msg += ":\n   " + self.line.strip()
         if self.lines:
-            msg += "\nContext lines:\n" + "\n".join(self.lines)
-            msg += "\t<--"
-            if self.err_type:
-                msg += ' %s' % self.err_type
+            for filename, lines in self.lines.items():
```
hjoliver, line 97: I think `self.lines` is still a list in the Jinja2 case (as opposed to empy)? Maybe you forgot to push up the next commit...

oliver-sanders: I replaced it with a dict in `cylc/flow/parsec/jinja2support.py`, it should now print both the `{% include` line and the "included" line.

hjoliver: (When I first looked, there were Jinja2 functional test errors relating to `self.lines` type here; but nevermind, you've clearly fixed it one way or another).

MetRonnie (approved):
Tested out. Before:
```
Jinja2Error: Jinja2 Error: some error
Context lines:

# line before include
{% include "foo.cylc" %}	<-- Exception
```
After:
```
Jinja2Error: Jinja2 Error: some error
File /home/h04/rdutta/cylc-run/GAH/foo.cylc
  # line before error
  {{ raise('some error') }}
  # line after error	<-- Exception
File /home/h04/rdutta/cylc-run/GAH/flow.cylc
  #!Jinja2
  
  # line before include
  {% include "foo.cylc" %}	<-- Exception
```
Minor niggle is that the traceback is reversed compared to what you would expect i.e. `foo.cylc` comes before `flow.cylc`

cylc/flow/parsec/fileparse.py
```diff
@@ -31,11 +31,10 @@
 """
 
 import os
+from pathlib import Path
 import re
 import sys
-
-from pathlib import Path
-from typing import Any, Dict, List, Optional
+import typing as t
```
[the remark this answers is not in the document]
oliver-sanders: It's a pain in managing the list of imported types, `typing` types conflict with `graphene` types and `traitlet` types etc. Saw some other projects are doing `import typing as t`.

hjoliver (approved):
Nice. Just needs a change log entry?

MetRonnie (approved):

PR #4457, opened 2021-10-07
Cylc 8 timezone defaults to UTC

wxtim:
These changes close #2324 

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.py` and
  `conda-environment.yml`.
- [x] Appropriate tests are included (unit and/or functional).
- [x] Appropriate change log entry included.
- [x] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/306

PR #4846, opened 2022-05-03
Improve `flow.cylc` reference docs

MetRonnie:

**Requirements check-list**
- [x] Does not need tests 
- [x] No change log entry required (docs/invisible to users).
- [x] No cylc-doc PR needed.


cylc/flow/cfgspec/workflow.py
```diff
@@ -177,15 +176,20 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
-            logging timestamps.
+            If ``True``, the workflow will use UTC as the time zone for
+            cycle points and timestamps in the logs.
+
+            This may also be set in the global config:
+            :cylc:conf:`global.cylc[scheduler]UTC mode`.
 
+            If instead you only want to set a time zone for cycle points,
+            leaving log timestamps to use the local/system time zone,
+            see :cylc:conf:`flow.cylc[scheduler]cycle point time zone`.
```
wxtim, line 187:
```suggestion
            see :cylc:conf:`flow.cylc[scheduler]cycle point time zone`.
            (`cycle point time zone` defaults to UTC.)
```

cylc/flow/cfgspec/workflow.py
```diff
@@ -196,24 +200,25 @@ def get_script_common_text(this: str, example: Optional[str] = None):
             ``False`` after finishing the :cylc:conf:`flow.cylc[runtime]`
             section.
 
-            In :ref:`Cylc 7 backward compatibility mode <Cylc_7_compat_mode>`,
-            implicit tasks are still allowed unless you explicitly set
-            this to ``False``.
+            .. admonition:: Cylc 7 compatibility mode
+
+               In :ref:`Cylc_7_compat_mode`, implicit tasks are still
```
wxtim, line 205: Have I missed a discussion about this?
There's a slight inconsistency in that I thought that we were going to allow implicit tasks in compatibility mode, unless a rose-suite.conf is present (since Rose 2019 had behaviour equivalent to `implicit tasks = False`).

wxtim (approved):
- [x] Jotted some thoughts down.
- [x] Built docs locally.

cylc/flow/cfgspec/workflow.py
MetRonnie: Good point, that was never updated in the docs

cylc/flow/cfgspec/workflow.py
```diff
@@ -177,15 +176,20 @@ def get_script_common_text(this: str, example: Optional[str] = None):
         {REPLACES} ``[cylc]``
     '''):
         Conf('UTC mode', VDR.V_BOOLEAN, desc='''
-            If true, workflow will use UTC as the time zone and for
-            logging timestamps.
+            If ``True``, the workflow will use UTC as the time zone for
+            cycle points and timestamps in the logs.
```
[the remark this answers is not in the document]
MetRonnie: Actually when I think about it, UTC mode only affects the log timestamps ever since #4457. Updated accordingly

hjoliver (approved):

PR #4849, opened 2022-05-04
Keep prerequisites for the datastore.

### LINKED ISSUE #4036: compute prerequisites for tasks outside n=0

hjoliver:
These changes close #4036 by retaining task prerequisite history in the DB (at the point of triggering), because we can't in general infer prerequisite satisfaction of completed tasks from task state, or use other data in the DB to compute it.

Main use cases for querying task prerequisites:
 1. what prerequisites does a future `n>0` task have?
     - (:+1: works already: we know what they are, and by definition they have not been satisfied)
 1. which prerequisites of a current `n=0` task remain unsatisfied?
     - (:+1: works already - stored in DB in case of restart)
 1. which prerequisites of a past `n>0` task were satisfied, i.e. why exactly did it trigger?
     - :-1:  we can't in general infer this from task state because of:
        - **manual triggering** (task triggered even though prerequisites were unsatisfied)
        - **conditional prerequisites** (task triggered even though some component prerequisites were not satisfied)
     - given the use case (which upstream events caused this task to trigger?) we need to know exactly which prerequisites were satisfied.
  1. (also: what prerequisites does a past task have regardless of state? no need to consider this because we always know what they are; it's just their state that is the problem)
  
Therefore, we need to use the DB to determine exactly which prerequisites were satisfied when past tasks triggered.

The information in the prerequisites table is not redundant with other DB data, because we need to know the prerequisite state *at trigger time*, NOT the eventual final state of all the upstream tasks that contribute to the prerequisites.  

This means the `task_prerequisites` table necessarily grows forever, but it's no worse than the existing history tables.

The only real solution to unconstrained history growth is to provide a housekeeping option. (Main loop plugin?)

**Requirements check-list**

### DIFF
--- a/cylc/flow/workflow_db_mgr.py
+++ b/cylc/flow/workflow_db_mgr.py
@@ -452,8 +452,6 @@ class WorkflowDatabaseManager:
         relevant insert statements for the current tasks in the pool.
         """
         self.db_deletes_map[self.TABLE_TASK_POOL].append({})
-        # We can comment this out to keep prereq history for the data-store:
-        self.db_deletes_map[self.TABLE_TASK_PREREQUISITES].append({})
         # No need to do:
         # self.db_deletes_map[self.TABLE_TASK_ACTION_TIMERS].append({})
         # Should already be done by self.put_task_event_timers above.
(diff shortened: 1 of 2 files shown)


hjoliver:
Project meeting discussion:
- The use case for viewing prerequisites of past tasks is mainly to see exactly why a conditionally-triggered task triggered.
- This is probably too niche to justify the additional DB use.
- Instead, we can close the issue by reporting "N/A" (or similar) in response to such a query.

PR #4853, opened 2022-05-04
bump deprecation warnings from Cylc 9 to 8.x

oliver-sanders:
[meeting 2022-05-04]

* Logically Cylc 9 should come at the next configuration breaking change, however, that isn't planned for some time.
  (note the Python API should not break flow.cylc compatibility)
* So we will remove back-support in the 8.x series rather than waiting for Cylc 9.

**Requirements check-list**

### DIFF
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -51,7 +51,8 @@ DEPRECATION_WARN = '''
 
    Deprecated section kept for compatibility with Cylc 7 workflow definitions.
 
-   **It will not be available at Cylc 9**.
+
+   This will be removed in a future version of Cylc 8.
 
    Use :cylc:conf:`flow.cylc[runtime][<namespace>]platform` instead.
 '''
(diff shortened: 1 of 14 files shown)


hjoliver:
> Logically Cylc 9 should come at the next configuration breaking change,

Historically, I bumped the major version along with "major" (and often internal) changes. However, the new way is better!

cylc/flow/cfgspec/workflow.py
```diff
@@ -1628,7 +1629,7 @@ def get_script_common_text(this: str, example: Optional[str] = None):
 
                 This was done to allow users to control the order of
                 definition of the variables. This section will be removed
-                in Cylc 9.
+                in a later version of Cylc 8.
```
hjoliver, line 1632:
```suggestion
                in a future version of Cylc 8.
```

(Might as well use the same term everywhere)

hjoliver (approved):

oliver-sanders:
> Historically, I bumped the major version along with "major" (and often internal) changes. However, the new way is better!

6, 7 & 8 all had network layer changes which I think were also a valid reason to bump the major version.

datamel (approved):
Approving. No problems found for me. Thanks @oliver-sanders.

PR #4854, opened 2022-05-04
Add warning if a platform name regex matches localhost

### LINKED ISSUE #4845: `[[localhost]]` platform selects from top of platforms list
**Describe the bug**
Localhost doesn't seem to follow the same platform selection rules as other platforms - it appears to select the top regex in the list, and not the bottom one!

Given the `global.cylc`:
```ini
[platforms]
    [[localhost|foo|bar]]
        global init-script = "echo 'defined  first'"
    [[localhost]]
        global init-script = "echo 'defined last'"
    [[foo|bar|baz]]
        global init-script = "echo 'defined first'"
    [[baz]]
        global init-script = "echo 'defined last'"
```

and the `flow.cylc`:
```ini
[scheduling]
    initial cycle point = 19831213T0600Z
    [[dependencies]]
        R1 = Quokka => Numbat
[runtime]
    [[Quokka]]
    [[Numbat]]
        platform = baz
```

then testing for the global init-script in the job script
```bash
grep "defined [first|last]" ~/cylc-run/localhost-investigation/runN/log/job/19831213T0600Z/*/NN/job
```

One would expect both to return "defined last" but in fact this is not the case for localhost.

**Pull requests welcome!**

wxtim:
These changes close #4845 

[post-hoc edit to ticket for the benefit of users clicking on changelog]
### Expand Platform definitions:
Allow platform definitions to use comma separated lists and be expanded as task definitions are. E.g

```ini
[[foo, bar]]
  hosts = host1, host2
[[foo]]
  install target = host2
```
is now equivelant to 
```ini
[[foo]]
  hosts = host1, host2
  install target = host2
[[bar]]
  hosts = host1, host2
```

### Fail if "localhost" matches a platform definition Regex.
Because "localhost" platform is a default item it is always the last platform to be checked. If any platform definition is a regex which matches "localhost" then that definition will be used instead. To prevent this, bad such platform definitions.

### Example

In the following example `localhost|foo|bar...` would have always been selected before this change, but will now be rejected.

```ini
[platforms]
  [[localhost|foo|bar...]]
    # some settings
  [[localhost]]
    # other settings
```

**Requirements check-list**
- [x] Change log entry added.


tests/functional/platforms/08-warn-if-regex-matches-localhost.t
```diff
@@ -0,0 +1,69 @@
...
+    "${TEST_NAME}.stderr")
+TEST_NAME="${TEST_NAME_BASE}-only-1-warn"
+if [[ "${NO_OF_WARNS}" == 1 ]]; then
+    ok "${TEST_NAME}"
+else
+    fail "${TEST_NAME}"
+fi
+
+
+# purge_rnd_workflow
```
datamel, line 68: Think this needs uncommenting.

datamel (approved):
Working as expected for me. Checked out and test run. Can't spot any implementation problems. 
Approving on proviso that the purge is sorted in the test :) Thanks @wxtim

wxtim:
Make it an exception.

tests/functional/platforms/08-warn-if-regex-matches-localhost.t
wxtim:
```suggestion
purge_rnd_workflow
```

wxtim:
@oliver-sanders Do you want me to count having reviewed your parts of the code as review 1 and you as having reviewed mine  as review 1 so we only want 1 further reviewer, or should I/we tag Ronnie?

@datamel This is not the PR you approved any more.

oliver-sanders:
> Do you want me to count having reviewed your parts of the code as review 1 and you as having reviewed mine

Sounds good.

oliver-sanders:
Shellcheck is complaining about syntax in a comment (false positive), otherwise fine.

tests/functional/platforms/08-warn-if-regex-matches-localhost.t
```diff
@@ -0,0 +1,60 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Check that ``[platforms][localhost]`` is only set automatically if it
+# not set in ``global.cylc``.
+
+export REQUIRE_PLATFORM=''
```
oliver-sanders, line 21: No need to set this if a platform is not needed (this might possibly have caused the test to be skipped).

```suggestion
```

tests/functional/platforms/08-warn-if-regex-matches-localhost.t
```diff
@@ -0,0 +1,60 @@
...
+
+create_test_global_config "" "
+    [platforms]
+        [[localh...]]
+        # This shouldn't override `localh...` in this one case, because the
+        # localhost default pins it to the top of the list.
+        [[localhost]]
+            [[[meta]]]
+                foo = 'foo'
+"
```
oliver-sanders, line 34: Quoting error which should cause the test to fail:

```suggestion
create_test_global_config '' '
    [platforms]
        [[localh...]]
        # This shouldn't override `localh...` in this one case, because the
        # localhost default pins it to the top of the list.
        [[localhost]]
            [[[meta]]]
                foo = 'foo'
'
```

oliver-sanders: (unresolving)

oliver-sanders:
Ooh good catch by shellcheck - SC1011: This apostrophe terminated the single quoted string!

cylc/flow/platforms.py
```diff
@@ -184,6 +184,16 @@ def platform_from_name(
...
+    for platform_name_re in list(platforms):
+        if (
+            # If the platform_name_re contains special regex chars
+            re.escape(platform_name_re) != platform_name_re
+            and re.match(platform_name_re, 'localhost')
+        ):
+            raise PlatformLookupError(
+                f'"localhost" settings will be defined by '
+                f'global.cylc[platforms][{platform_name_re}]'
+            )
```
oliver-sanders, line 196: This message is a little cryptic, how about something like:

```
The "localhost" platform cannot be defined using a regular expression.
See the documentation for "global.cylc[platforms][localhost]" for more information.
```

With an accompanying note in `global.cylc[platforms][localhost]` containing an example like this:

```
[platforms]
    [[localhost|cylc-server-\d]]  # error
    [[localhost, cylc-server-\d]]  # ok
```

datamel (approved):
I've checked this out and read the code. Tests pass locally for me. No problems found from me. Thanks @wxtim

PR #4856, opened 2022-05-04
contact: process check should handle dirty JSON

### LINKED ISSUE #4855: contact: process check should handle dirty JSON
> Reported in https://cylc.discourse.group/t/critical-disk-i-o-error/471/8

The process check run to determine whether a scheduler process is still running does not tolerate "dirty" output. This "dirty" output describes when other things write to stdout during the command execution polluting the command output.

This is typically caused by shell configuration files e.g. Bashrc:

```bash
# ~/.bashrc
echo "Hello $USER"  #  this will pollute command output
```

The solution is to remove or suppress the polluting command or alternatively to stick it in a branch which only runs in interactive mode so it can't mess with non-interactive commands:

```bash
# ~/.bashrc
if [[ $- == *i* ]]; then
    # this will only run in interactive mode
    echo "Hello $USER"
fi
```

We see this issue quite often so have built in protection against it here - `cylc.flow.terminal.parse_dirty_json`

Somebody (yours truely, sorry) forgot to add this into the process check when it moved over to the `cylc psutil` implementation (`cylc.flow.workflow_files._is_process_running`) :cry:.

**Pull requests welcome!**

oliver-sanders:
* Closes #4855
* Protects the scheduler process check call against stdout pollution.

**Requirements check-list**

### DIFF
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -21,7 +21,6 @@ from contextlib import suppress
 from enum import Enum
 from functools import partial
 import glob
-import json
 import logging
 import os
 from pathlib import Path
@@ -50,6 +49,7 @@ from cylc.flow.exceptions import (
     WorkflowFilesError,
     handle_rmtree_err,
 )
+from cylc.flow.loggingutil import CylcLogFormatter, close_log
 from cylc.flow.pathutil import (
     expand_path,
     get_cylc_run_dir,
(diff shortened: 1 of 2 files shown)


hjoliver (approved):
Looks good.

cylc/flow/workflow_files.py
```diff
@@ -435,17 +435,27 @@ def _is_process_running(
...
         LOG.debug(
             f'$ {cli_format(cmd)}  # returned {proc.returncode}\n{err}'
         )
+        error = True
+    else:
+        try:
+            process = parse_dirty_json(out)[0]
+        except ValueError:
+            # the JSON cannot be parsed, log it
+            LOG.warn(f'Could not parse JSON:\n{out}')
```
datamel, line 450:
```suggestion
            LOG.warning(f'Could not parse JSON:\n{out}')
```

datamel (commented):
Small comment about log level used. 
If left as `warn` you get a deprecation notice when running tests.
Otherwise, no problems found.

datamel (approved):
Thanks. Think this can be merged as change doesn't affect the CI.

oliver-sanders:
(passed)

PR #4860, opened 2022-05-06
Raise an error if the user tries to set owner.

wxtim:
(Owner setting no longer does anything meaningful at Cylc 8)

These changes address an issue with no Cylc-flow ticket.
They relate to https://github.com/cylc/cylc-doc/pull/464.

## Requirements check-list
- [x] Covered by @oliver-sanders documentation PR at cylc/cylc-doc/pull/464.


cylc/flow/config.py
```diff
@@ -2341,3 +2342,25 @@ def process_metadata_urls(self):
...
+    def check_for_owner(tasks: Dict) -> None:
+        """Raise exception if [runtime][task][remote]owner
+        """
+        owners = {}
+        for task, tdef in tasks.items():
+            remote = tdef.get('remote', None)
+            if remote:
+                owner = remote.get('owner', None)
+                if owner:
+                    owners[task] = owner
```
MetRonnie, line 2356:
```suggestion
            owner = tdef.get('remote', {}).get('owner', None)
            if owner:
                owners[task] = owner
```

tests/unit/test_config.py
```diff
@@ -1390,3 +1390,49 @@ def test_zero_interval(
         assert logged
     else:
         assert not logged
+
+
+@pytest.mark.parametrize(
+    'tasks',
```
MetRonnie, line 1396: Slightly more accurate name might be `runtime_cfg` instead of `tasks`

cylc/flow/config.py
```diff
@@ -2341,3 +2342,25 @@ def process_metadata_urls(self):
...
+        if owners:
+            msg = (
+                '"[runtime][task][remote]owner" is deprecated at Cylc 8.'
```
MetRonnie, line 2359: Obsolete seems more accurate?
```suggestion
                '"[runtime][task][remote]owner" is obsolete at Cylc 8.'
```

cylc/flow/config.py
```diff
@@ -2341,3 +2342,25 @@ def process_metadata_urls(self):
...
+                '\nsee https://cylc.github.io/cylc-doc/latest/'
+                'html/7-to-8/index.html'
```
MetRonnie, line 2361:
```suggestion
                '\nsee https://cylc.github.io/cylc-doc/latest/'
                'html/7-to-8/major-changes/remote-owner.html'
```

MetRonnie:
Another thing: is it possible to do this before https://github.com/cylc/cylc-flow/blob/24b49946e1dc43195b81f8e0b782586e1a997479/cylc/flow/config.py#L353

MetRonnie:
Something went wrong in the last commit, it reverted the previous changes. I have unresolved my suggestions that got reverted

wxtim:
> Something went wrong in the last commit, it reverted the previous changes. I have unresolved my suggestions that got reverted

Thanks for spotting that.

MetRonnie:
Should there be a warning in 
https://github.com/cylc/cylc-flow/blob/3af18367c3bf61250444293c4d439ee37e086e44/cylc/flow/cfgspec/workflow.py#L1342

that it is obsolete, for the sake of the docs? That warning could link to https://cylc.github.io/cylc-doc/latest/html/7-to-8/major-changes/remote-owner.html [^1]

[^1]: Link won't work until 8.0rc3 docs deployed; currently works if you replace `latest` with `nightly`

wxtim:
@MetRonnie - Have I missed something that you wanted me to do on review?

cylc/flow/cfgspec/workflow.py
```diff
@@ -1339,7 +1339,17 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
-                Conf('owner', VDR.V_STRING)
+                # TODO: Convert URL to a stable or latest release doc after 8.0
+                # https://github.com/cylc/cylc-flow/issues/4663
+                Conf('owner', VDR.V_STRING, desc="""
+                    This setting is obsolete at Cylc 8.
+
+                    .. seealso::
+
+                        `Documentation on changes to remote owner
+                        <https://cylc.github.io/cylc-doc/nightly/html/
```
MetRonnie, line 1350:
```suggestion
                        <https://cylc.github.io/cylc-doc/latest/html/
```

cylc/flow/config.py
```diff
@@ -2349,16 +2349,16 @@ def check_for_owner(tasks: Dict) -> None:
...
+                owners[task] = owner
         if owners:
+            # TODO: Convert URL to a stable or latest release doc after 8.0
+            # https://github.com/cylc/cylc-flow/issues/4663
             msg = (
-                '"[runtime][task][remote]owner" is deprecated at Cylc 8.'
-                '\nsee https://cylc.github.io/cylc-doc/latest/'
-                'html/7-to-8/index.html'
+                '"[runtime][task][remote]owner" is obsolete at Cylc 8.'
+                '\nsee https://cylc.github.io/cylc-doc/nightly/'
```
MetRonnie, line 2360:
```suggestion
                '\nsee https://cylc.github.io/cylc-doc/latest/'
```

oliver-sanders: (explained in comment) won't work because the page isn't live yet

cylc/flow/cfgspec/workflow.py
oliver-sanders: (explained in comment) won't work because the page isn't live yet

MetRonnie: @oliver-sanders I can't tell if you're justifying using `latest` or justifying using `nightly`

oliver-sanders: latest won't work because the page is not published yet.

MetRonnie: But surely it doesn't matter that it doesn't work because we haven't released cylc-flow 8.0rc3 yet? Unless this for the benefit of internal MO testers?

MetRonnie: Ideally we don't want to leave it as nightly indefinitely because any changes we make to the docs on a daily basis could break doc links in the current cylc-flow release.

In fact, pinning to the release version would be better than using latest or nightly. If we could centralise all documentation links in a dict that could be achievable as part of the automated release process, maybe even linkchecking could be implemented

oliver-sanders: > In fact, pinning to the release version would be better than using latest or nightly. If we could centralise all documentation links in a dict that could be achievable as part of the automated release process, maybe even linkchecking could be implemented

Yes, but a topic for another issue. I took a look at doing this a while back but it's trickier than it seems.

oliver-sanders: I've written it up so we remember - https://github.com/cylc/cylc-flow/issues/4867

There is a nice solution, however, it involves work in both cylc-doc and cylc-flow which we don't have time for ATM.

oliver-sanders: Since this didn't go into RC3 all this is now irrelevant!

hjoliver: So, I guess the suggested change here should be committed before merging this PR?

wxtim:
Sort out conflict post release -will need RC4 Header!

wxtim:
@MetRonnie  Can you do a quick final eyeball and then we can merge?

MetRonnie (approved):

PR #4863, opened 2022-05-09
Remove setting of execution timeout based on execution time limit

### LINKED ISSUE #4859: Investigate how execution timeout is set
As noted in #3706, execution timeout gets set if you define `execution time limit` but this doesn't appear to be documented and the way it is done probably doesn't make much sense.

dpmatthews:
These changes close #4859

**Requirements check-list**

### DIFF
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -1510,7 +1510,6 @@ class TaskEventsManager():
                     'execution time limit polling intervals')
                 if not time_limit_delays:
                     time_limit_delays = [60, 120, 420]
-                timeout = time_limit + sum(time_limit_delays)
                 if sum(delays) > time_limit:
                     # Remove excessive polling before time limit
                     while sum(delays) > time_limit:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -45,6 +45,10 @@ workflow source argument), and rename the `--flow-name` option to
 
 ### Fixes
 
+[#4863](https://github.com/cylc/cylc-flow/pull/4863) - Execution timeout is no
+longer set based on execution time limit. Fixes bug where execution timeout
+would get overridden.
+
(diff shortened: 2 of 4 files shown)


cylc/flow/task_events_mgr.py
```diff
@@ -1510,7 +1510,6 @@ def _reset_job_timers(self, itask):
                     'execution time limit polling intervals')
                 if not time_limit_delays:
                     time_limit_delays = [60, 120, 420]
-                timeout = time_limit + sum(time_limit_delays)
```
oliver-sanders, line 1513: Using the `execution time limit` as a default for the `execution timeout` makes sense I think. Allowing this to overwrite the configured value is a bug.

How about:

```python
if not timeout:
    timeout = time_limit + P1M
```

?

dpmatthews: I thought about this. However, what is the purpose of the timeout? If the task has reached it's time limit then you should expect to receive a message that the task has completed so the timeout shouldn't happen (unless you're using polling in which case the timeout is fairly meaningless).
Now that we have `execution time limit` I think the main use case for `execution timeout` is to set a time shorter than the time limit - e.g. this task is taking longer than normal.

oliver-sanders: These timeout events could be passed on to handler scripts and ingested into monitoring systems or whatnot, but yes, I wouldn't *expect* the timeout event to occur at all for a default configuration.

dpmatthews: I don't think setting the timeout to a fixed +PT1M offset makes sense. If we really think there is a requirement for this then I'd propose we change `execution timeout` to accept offsets to the `execution time limit` as well as absolute periods (but that's beyond the scope of this PR).

oliver-sanders (approved):

hjoliver (approved):
Looks fine.

But I'll hold off merging till you've considered this: https://github.com/cylc/cylc-flow/issues/4859#issuecomment-1124511815

PR #4869, opened 2022-05-12
Assorted tutorial fixes against Cylc 8

### LINKED ISSUE #4660: Tutorial workflow: messed up wind plot
~/cylc-run/cylc-forecasting-workflow/run1/work/20210723T0300Z/consolidate_observations/wind.png
wind

wxtim:
Correct get-wind-data

These changes close #4660

**Requirements check-list**
- [x] Not currently tested - manually tested.
- [x] No change log entry required (tutorial workflow content).


cylc/flow/etc/tutorial/cylc-forecasting-workflow/flow.cylc
```diff
@@ -43,6 +47,11 @@
             # The area to generate forecasts for (lng1, lat1, lng2, lat2)
             DOMAIN = -12,48,5,61  # Do not change!
 
+            # These environment variables ensure that tasks can
+            # run in the same environment as the workflow:
+            PYTHONPATH = {{':'.join(path)}}
+            PATH = $(dirname {{executable}}):$PATH
+
```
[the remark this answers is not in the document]
hjoliver: (It pays to assume this - and use it - for consistent local and remote job behaviour).

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/consolidate-observations
```diff
@@ -81,8 +81,8 @@ def get_wind_fields():
     data.sort(key=lambda x: x[1])
 
     # Convert data into (lat, lng, (wind_x, wind_y)) components.
-    x_coords = [datum[0] for datum in data]
-    y_coords = [datum[1] for datum in data]
+    x_coords = [datum[1] for datum in data]
+    y_coords = [datum[0] for datum in data]
```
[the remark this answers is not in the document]
wxtim: I wanted to add a map layer but couldn't find a nice way of doing it without adding another dependency. No thanks.

cylc/flow/etc/tutorial/cylc-forecasting-workflow/flow.cylc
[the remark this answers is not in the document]
oliver-sanders: > for consistent local and remote job behaviour

We've configured localhost to use a "clean" env which also requires this.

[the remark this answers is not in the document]
oliver-sanders: This looks good to me. I would maybe tweak the comment as this isn't an approach we want users going down.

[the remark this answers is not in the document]
MetRonnie: In fact, could we hide this further, seeing as we're already using jinja2 at this point? Perhaps as an include file or something? (P.S. this path-wrangling will need to be done for some of the other tutorials too)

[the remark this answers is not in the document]
wxtim:
```suggestion

            # These environment variables ensure that tasks can
            # run in the same environment as the workflow:
            # This is not normally good practice, but required to make
            # the tutorial run out-of-the-box.
            PYTHONPATH = {{':'.join(path)}}
            PATH = $(dirname {{executable}}):$PATH

```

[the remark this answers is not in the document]
wxtim: > In fact, could we hide this further, seeing as we're already using jinja2 at this point? Perhaps as an include file or something? (P.S. this path-wrangling will need to be done for some of the other tutorials too)

I think that this is a reasonable plan - it puts a barrier, however weak, between users and the practice we don't want them copying.

cylc/flow/etc/tutorial/runtime-tutorial/etc/python-job.settings
```diff
@@ -0,0 +1,12 @@
+# WARNING: This file contains an anti-pattern
```
wxtim, line 1: @oliver-sanders  Can I use a symlink instead of copying all of these?

oliver-sanders: Yep, we already do this for other tutorial files.

cylc/flow/etc/tutorial/consolidation-tutorial/flow.cylc
```diff
@@ -100,3 +100,5 @@
             RESOLUTION = 0.2
             # The area to generate forecasts for (lng1, lat1, lng2, lat2)
             DOMAIN = -12,48,5,61  # Do not change!
+
+{% include 'etc/python-job.settings' %}
```
MetRonnie, line 104: Shebang needed

wxtim: This file has been replaced by a symlink to a file which _does_ have the required shebang.

hjoliver:
Failing test: 
```
FAILED tests/unit/test_resources.py::test_get_resources_all[tutorial/runtime-tutorial]
```

cylc/flow/etc/tutorial/consolidation-tutorial/flow.cylc
MetRonnie: But the flow.cylc files will need the Jinja2 shebang

cylc/flow/etc/tutorial/runtime-tutorial/etc/python-job.settings
```diff
@@ -0,0 +1 @@
+cylc/flow/etc/tutorial/cylc-forecasting-workflow/etc/python-job.settings
```
wxtim, line 1:
```suggestion
../../cylc-forecasting-workflow/etc/python-job.settings
```

cylc/flow/etc/tutorial/consolidation-tutorial/etc/python-job.settings
```diff
@@ -0,0 +1 @@
+cylc/flow/etc/tutorial/cylc-forecasting-workflow/etc/python-job.settings
```
wxtim, line 1:
```suggestion
../../cylc-forecasting-workflow/etc/python-job.settings
```

cylc/flow/etc/tutorial/runtime-introduction/etc/python-job.settings
```diff
@@ -0,0 +1 @@
+cylc/flow/etc/tutorial/cylc-forecasting-workflow/etc/python-job.settings
```
wxtim, line 1:
```suggestion
../../cylc-forecasting-workflow/etc/python-job.settings
```

cylc/flow/etc/tutorial/runtime-introduction/etc/python-job.settings
```diff
@@ -0,0 +1,12 @@
+# WARNING: This file contains an anti-pattern
```
MetRonnie, line 1: Any reason this one has not been made into a symlink?

wxtim: It didn't pass the tests when it was a symlink.

MetRonnie: Then it seems suspicious the others are passing

MetRonnie: The problems:
- The symlink points outside of the workflow source so gets broken on install
- There is inconsistency in the test scripts between the three:

  https://github.com/cylc/cylc-flow/blob/47ec638f85ee7a762c8a13736b204e64d2ec6330/cylc/flow/etc/tutorial/runtime-introduction/.validate#L18-L23

  https://github.com/cylc/cylc-flow/blob/9ad4c07f8daa6e339ed43ac388b45e53b0972b69/cylc/flow/etc/tutorial/consolidation-tutorial/.validate#L19-L21  

  (`runtime-tutorial` does not even have a `.validate` script)

wxtim: Yes. Might as well fix it in this PR, it's mission-creeped enow already.

wxtim: > I think the symlinks will need to be replaced by copies of the `python-job.settings`, because if the user hasn't yet installed `cylc-forecasting-workflow` and doesn't use `--no-run-name` when installing the workflows, the symlinks will be broken?

I don't think that this check is in scope, happy to check later because I'm fairly unhappy about what's going - but I'd like to check with @oliver-sanders  because he said that symlinks were ok - (No, I'm not convinced either.)

oliver-sanders: Are the links not working in the other tutorials? Perhaps we have a relative/absolute issue, there are some `rsync` options which configure how `rsync` behaves when it finds links.

Might need to get rsync to `--copy-links`, can we do that with existing configuration?

oliver-sanders: Note we have been using symlinks with the tutorials for years when they were in Rose, using `rsync` to install. Since `cylc install` is `rsync` underneath I would *expect* it to be the same deal?

wxtim: I'll have a look. I guess this is urgent - it ought to go in before the seminars

wxtim: This seems to have sorted itself out. I don't know what was going on, but it no longer appears to be going.

MetRonnie: Ah got it - `cylc get-resources tutorial` turns the symlinks into files

MetRonnie (approved):

hjoliver (approved):

PR #4877, opened 2022-05-16
Update Jinja2 to 3.0.*

### LINKED ISSUE #4702: jinja2: move to version 3
Jinja2 has made it out of the 2.x series into the 3.x series (but is still called Jinja2).

Historically we have had to pin cylc-flow to exact versions of Jinja2 due to a series of highly disruptive breaking changes made during the 2.x series in order to give our users a fighting chance.

We should consider moving to version 3, however, do so cautiously. Suggest rendering a particularly complex workflow configuration with Jinja2 2.11 then with 3.? and comparing the output for piece of mind.

Questions:

* When to update, 8.0.0, 8.1.0, 8.x? Dependent on the presence or absence of breaking changes?
* Should we continue to pin to minor releases?

**Pull requests welcome!**

MetRonnie:
These changes close #4702

**Requirements check-list**


cylc/flow/parsec/jinja2support.py
```diff
@@ -207,25 +207,25 @@ def get_error_lines(
...
             start_line = max(lineno - CONTEXT_LINES, 0)
-            if filename == '<template>':
+            if filename in {'<template>', '<unknown>'}:
                 filename = base_template_file
                 lineno += 1  # shebang line ignored by jinja2
                 lines = template_lines[start_line:lineno]
-            else:
-                with suppress(FileNotFoundError):  # noqa: SIM117
-                    with open(filename, 'r') as jinja2_file:
-                        jinja2_file.seek(start_line, 0)
```
MetRonnie, line 222: `.seek(n)` goes to byte n, not line n
https://docs.python.org/3/library/io.html#io.IOBase.seek

MetRonnie:
Do we want to include a complex jinja2 workflow example that we diff against a reference file, as a functional test?

oliver-sanders:
> Do we want to include a complex jinja2 workflow example that we diff against a reference file, as a functional test?

I don't think so and we wouldn't be able to capture all uses and patterns. Hopefully future upgrades won't be so complicated. The timing of this one with Cylc 8 and the need for 7/8 interoperability has made this tricky.

MetRonnie:
functional tests / macos failed due to Codecov upload failure only

hjoliver (approved):

PR #4881, opened 2022-05-18
Fix task matching bug

### LINKED ISSUE #4843: Task matching fails silently when the cycle point does not exactly follow `[scheduler]cycle point format`
#### Steps to reproduce the bug

```
# flow.cylc
[scheduler]
    UTC mode = True
    allow implicit tasks = True
[scheduling]
    runahead limit = P1
    initial cycle point = 2000
    [[graph]]
        P1Y = foo
```
(The default cycle point format is `CCYYMMDDThhmmZ`)

```console
$ cylc play myflow --pause
$ cylc hold myflow//20000101T00Z  # Note the minute is not specified
```
The workflow log says
```
INFO - Command succeeded: release(['20000101T00Z/foo'])
```
as if it worked. And TUI shows the task as held.

But the subtle giveaway that it did not work is that you don't get
```
INFO - [20000101T0000Z/foo waiting(held)(runahead) job:00 flows:1] => waiting(held)
```

#### Additional context

The string match here looks like the problem. 
https://github.com/cylc/cylc-flow/blob/aa01d96121e4d23c99a0d45262b8e9fe1846938a/cylc/flow/id_match.py#L177-L178

I think we should try matching `ISO8601Point`s or `metomi.isodatetime.data.TimePoint`s (but need to take care to still match globs).

This bug does not occur for holding future cycle points

MetRonnie:
- These changes close #4843 

Can manually test by following steps to reproduce in the issue above

**Requirements check-list**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,24 +29,35 @@ creating a new release entry be sure to copy & paste the span tag with the
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
 -------------------------------------------------------------------------------
+## __cylc-8.0rc4 (<span actions:bind='release-date'>Upcoming</span>)__
+
+Fourth Release Candidate for Cylc 8 suitable for acceptance testing.
+
+### Fixes
+
+[#4881](https://github.com/cylc/cylc-flow/pull/4881) - Fix bug where commands
+targeting a specific cycle point would not work if using an abbreviated
+cycle point format.
+
+-------------------------------------------------------------------------------
 ## __cylc-8.0rc3 (<span actions:bind='release-date'>Released 2022-05-19</span>)__
 
 Third Release Candidate for Cylc 8 suitable for acceptance testing.
 
 ### Enhancements
 
-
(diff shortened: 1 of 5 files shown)


oliver-sanders (approved):

CHANGES.md
```diff
@@ -28,25 +28,36 @@ creating a new release entry be sure to copy & paste the span tag with the
...
 
 ### Enhancements
 
-
 [#4738](https://github.com/cylc/cylc-flow/pull/4738) and
 [#4739](https://github.com/cylc/cylc-flow/pull/4739) - Implement `cylc trigger
 [--flow=] [--wait]` for manual triggering with respect to active flows (the
-default), specific flows, new flows, or one-off task runs.
+default), specific flows, new flows, or one-off task runs. This replaces
+the `--reflow` option from earlier pre-release versions.
```
hjoliver, line 53: :+1:

hjoliver (approved):
Looks good.

wxtim (approved):

PR #4887, opened 2022-05-19
Raise error if there is a relative path in `global.cylc[install]source dirs`

### LINKED ISSUE #4815: `global.cylc[install]source dirs` should not accept relative paths
`global.cylc[install]source dirs` should only accept absolute paths. It should be a global config error if you specify a relative path.

No error is raised for
```
[install]
    source dirs = foo/bar
```
e.g. if you validate a random workflow, the global config validation step should fail first of all, but it doesn't.

**Additional context/questions**

~When checking if a path is absolute or relative, make sure to call `cylc.flow.pathutil.expand_path()` on the path first, because env vars or tilde prefixes might be present.~ (Edit: don't do this - see [Dave's reponse](https://github.com/cylc/cylc-flow/issues/4815#issuecomment-1126017981) to the question below)
- What should happen if an env var fails to expand (i.e. non-existent env var)?

MetRonnie:
These changes close #4815

**Requirements check-list**


cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1487,22 +1498,39 @@ def load(self):
...
-                    LOG.error(f'bad {conf_type} {fname}')
-                    raise
+                self._load(fname, conf_type)
 
         # Expand platforms needs to be performed first because it
         # manipulates the sparse config.
         self._expand_platforms()
+
+        # Flesh out with defaults
+        self.expand()
```
MetRonnie, line 1508: This doesn't change the code actually, just makes it clearer - `self._set_default_editors()` was doing the expansion all this time

cylc/flow/parsec/config.py
```diff
@@ -111,27 +122,32 @@ def expand(self):
...
             cfg = self.dense
 
-        parents = []
+        parents: List[str] = []
         if keys:
             for key in keys:
                 try:
                     cfg = cfg[key]
                 except (KeyError, TypeError):
-                    if parents in self.manyparents or key in self.get(parents):
```
MetRonnie, line 134: This `self.get()` was expanding the config this whole time

datamel (approved):
Code read. Manually tested installing workflows with various source dir settings. Code read. Workflows are running smoothly for me and the error is raised appropriately.

PR #4890, opened 2022-05-20
Check symlink target.

### LINKED ISSUE #4888: symlink dirs: check for existing link target?
```ini
# global.cylc
[install]
   [[symlink dirs]]
      [[[localhost]]]
         run = /nesi/nobackup/$PROJECT/$USER
```

Now:
- install a new workflow from source (to `blah/run1`)
- mess with the installed files (by running the workflow, or otherwise)
- remove the run directory with `rm -rf ~/cylc-run` or `rm -rf ~/cylc-run/blah`
- install the workflow again from source (to `blah/run1`)

Result:
- the original messed-with files (those not overwritten by the new install) will exist in the new run directory (because `rm` does not follow symlinks).

Of course, this is one of the two reasons we should use `cylc clean` and not `rm -rf` (clean up job hosts, and clean up symlinked run dirs). But some users won't be aware of that, or will likely revert to using `rm -rf` sometimes without being aware of the consequences.

I think `cylc install` should abort if it finds the target of a new symlinked dir already exists, to ensure that run directories  contain only the expected files (and those generated at run time).

(I can't think of any good reason why an existing target should be allowed - agree?)

hjoliver:
These changes close #4888  - abort `cylc install` if symlink dirs target already exists.

```console
(venv) oliverh@w-nwp01:~/cylc-src/foo$ cylc install
INSTALLED foo/run1 from /scale_wlg_persistent/filesets/home/oliverh/cylc-src/foo

(venv) oliverh@w-nwp01:~/cylc-src/foo$ ls -lrt ~/cylc-run/foo
total 2
lrwxrwxrwx 1 oliverh oliverh   50 May 20 02:50 run1 -> /nesi/nobackup/niwa99999/oliverh/cylc-run/foo/run1
lrwxrwxrwx 1 oliverh oliverh    4 May 20 02:50 runN -> run1
drwxrwxr-x 2 oliverh oliverh 4096 May 20 02:50 _cylc-install

(venv) oliverh@w-nwp01:~/cylc-src/foo$ rm -rf ~/cylc-run

(venv) oliverh@w-nwp01:~/cylc-src/foo$ cylc install
WorkflowFilesError: Symlink target already exists: /nesi/nobackup/niwa99999/oliverh/cylc-run/foo/run1.
```

**Requirements check-list**


hjoliver:
Confirmed: this fix also covers job platform init, at run time:

Screenshot from 2022-05-24 18-30-53

Tested using a fresh swarm build using this branch, to get a remote host with independent FS:
- set up a symlinked `run` dir for the `_remote_background_indep_tcp` host, in global config
- install and run a small workflow that submits a job to the remote host
- remove `cylc-run` on the remote host
- install and run again: the new run should abort as above during host init

hjoliver:
I'll add or tweak tests tomorrow, before undrafting this.

hjoliver:
Thanks @MetRonnie - if you wouldn't mind. I'm currently a bit swamped with local setup stuff, for Cylc 8 on NIWA and NeSI HPC platforms.

MetRonnie:
> Confirmed: this fix also covers job platform init, at run time...
> 
> Tested with a fresh swarm build using this branch, to get a remote host with independent FS:
> 
> * set up a symlinked `run` dir for the `_remote_background_indep_tcp` host, in global config
> * install and run a small workflow that submits a job to the remote host
> * remove `cylc-run` on the remote host
> * install and run again: the new run should abort as above during host init

Ah, do you want me to create a functional test that does this?

hjoliver:
> Ah, do you want me to create a functional test that does this?

I suppose that would be good. I can't think of an easier (unit or integration) test for this.

wxtim (approved):

oliver-sanders:
> the workflow only stalls instead of shutting down after the remote init failure. Is that how it should be?

Remote init happens when the first task on an install target tries to submit to a platform on that install target. If the remote init fails then the task goes into the submit-failed state (as remote-init is kinda like the first stage of job submission). Tasks for platforms on other install targets might still be able to submit.

PR #4894, opened 2022-05-23
Fix bug in remote tidy error handling

MetRonnie:
This is a small change with no associated Issue. Fix an `AttributeError` that could occur if a remote tidy subprocess returned with a non-zero exit code.

I have checked and this bug is not present in Cylc 8 (fixed in https://github.com/cylc/cylc-flow/commit/a917a6f1629ec53c34988a0410b28fab1dc65d10 if anyone was interested)

**Requirements check-list**
- [x] No tests needed (Meh).
- [x] No change log entry required (Tiny).

### DIFF
--- a/lib/cylc/task_remote_mgr.py
+++ b/lib/cylc/task_remote_mgr.py
@@ -275,7 +275,7 @@ class TaskRemoteMgr(object):
                     LOG.warning(TaskRemoteMgmtError(
                         TaskRemoteMgmtError.MSG_TIDY,
                         (host, owner), ' '.join(quote(item) for item in cmd),
-                        proc.ret_code, out, err))
+                        proc.returncode, out, err))
         # Terminate any remaining commands
         for (host, owner), (cmd, proc) in procs.items():
             try:


lib/cylc/task_remote_mgr.py
```diff
@@ -275,7 +275,7 @@ def remote_tidy(self):
                     LOG.warning(TaskRemoteMgmtError(
                         TaskRemoteMgmtError.MSG_TIDY,
                         (host, owner), ' '.join(quote(item) for item in cmd),
-                        proc.ret_code, out, err))
+                        proc.returncode, out, err))
```
MetRonnie, line 278: `proc` is a `subprocess.Popen` instance, not a `SubProcContext` or whatever the in-house Cylc thing is called. So the correct attribute is `.returncode` - https://docs.python.org/2.7/library/subprocess.html#subprocess.Popen.returncode

hjoliver (approved):

wxtim (approved):

PR #4901, opened 2022-05-28
protobuf major version update (4.21.*)

### LINKED ISSUE #4884: deps: bump protobuf
From element (@dwsutherland)

> Protobuf new release claims significant performance boost: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
>
> just a release candidate at the moment (I think)
> Changes made in May, 2022  |  Protocol Buffers  |  Google Developers - Google Developers
> Protocol Buffers Language English Bahasa Indonesia Deutsch Español Français Português – Brasil Русский
> might be some jupyter that uses protobuf  .. so will need to check if any conflicts (or maybe that was just zmq between hub and server)

Look into bumping the protobuf release, be careful to avoid creating version conflicts with Jupyter projects.

Note the reason for our strict version dependence is that as a distributed system different Cylc installations need to be compatible. Look into protobuf back compat support to determine how strict/loose we need to be when pinning protobuf.

dwsutherland:
These changes close #4884

**Requirements check-list**


dwsutherland:
Conda package needs to become available for `4.21.1`

hjoliver:
> Conda package needs to become available for `4.21.1`

So this PR is blocked till then?

dwsutherland:
> > Conda package needs to become available for `4.21.1`
> 
> So this PR is blocked till then?

Yeah, I guess so:
```
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... failed

ResolvePackageNotFound: 
  - protobuf[version='>=4.21.1']
```

I guess technically we could run with an older version of protobuf with the conda install ... but would be better to keep them inline.

dwsutherland:
might want to wait till this one is resolved too:
https://github.com/protocolbuffers/protobuf/issues/10088

hjoliver:
> I had to take a second look at that too .. It's not deleted, however, it has been drastically reduced (it's what the updated code generator spat out)..

A-ha, that's cool!

oliver-sanders:
Linked issue is still open so it's too late to get this in for `8.0.0`. Will have to look into any inter-version compatibility issues later to ensure the UIS can work with older and newer versions side-by-side or work out a migration if not.

oliver-sanders:
v4.21.2 is out, unblocking this...

dwsutherland:
> v4.21.2 is out, unblocking this...

Not sure where you went to find this.. not here:
https://anaconda.org/anaconda/protobuf

But, the install works 👍  .. PR updated

Tested `flow --> uis --> ui` .. apparently messages should be back compatible with `3.19.4`

dwsutherland:
> Not sure where you went to find this.. not here:
https://anaconda.org/anaconda/protobuf

Ah, the forge:
https://anaconda.org/conda-forge/protobuf

dwsutherland:
That macos test keeps failing/timing-out... the tests work for me (ubuntu).. 

@oliver-sanders - could you check on your system? (since it's `macos`)

oliver-sanders:
I ran the tests which showed up as running after the timeout locally, they all passed quickly.

Running the full test battery locally to see if anything hangs but I think you just got unlucky, these timeouts do happen, from the timestamps the tests were running right up to the timeout so no indication of hanging.

As the number of functional tests continues to increase we will have to increase the timeout for Mac OS. Passionate plea from me to use integration tests more. Lots of functional tests don't actually need to be functional tests. I can run all doctests, unit and integration tests on my machine in just 20 seconds!

conda-environment.yml
```diff
@@ -12,7 +12,7 @@ dependencies:
   - jinja2 >=3.0,<3.1
   - metomi-isodatetime >=1!3.0.0, <1!3.1.0
   # Constrain protobuf version for compatible Scheduler-UIS comms across hosts
-  - protobuf >=3.19.0,<3.20.0
+  - protobuf >=4.21.2
```
oliver-sanders, line 15: Are we ok to un-pin the protobuf version.

If cylc-flow uses protobuf `4.21` and the UIS uses `4.42` would that be a problem? If the UIS uses `5.10` would that be a problem?

dwsutherland: Maybe just do `>=4.21.1,<4.30.0`   (can't do `4.21.0` as I raised an issue with that one, and they fixed it)

hjoliver (approved):
Looks fine to me.

setup.cfg
```diff
@@ -69,7 +69,7 @@ install_requires =
     jinja2==3.0.*
     metomi-isodatetime==1!3.0.*
     # Constrain protobuf version for compatible Scheduler-UIS comms across hosts
-    protobuf==3.19.*
+    protobuf==4.21.*
```
oliver-sanders, line 72: Should this be `>=4.21.2,<4.22.0` too? Or is the issue with `4.21.0`, `4.21.1` just with the conda-forge packages?

(if so worth sticking a comment in here so we don't get confused when updating the conda-forge feedstock at release time).

oliver-sanders (commented):
LGTM, one comment to resolve then ready to merge.

oliver-sanders (approved):

oliver-sanders:
BTW: I've checked and it looks like the UIS is able to handle the version change ok so we don't need to go bumping the API number to exclude 8.0 workflows from the 8.1 UIS (or vice versa).

This suggests we may be able to relax the protobuf version range but without a better understanding of protobuf compatibility I can't say.

Issue #4873, opened 2022-05-13
auto restart plugin issues


dpmatthews:
There are several problems with the auto restart main loop plugin:

1) It's default interval is not documented. True of other plugins as well (only health check and reset bad hosts are documented).

2) The default interval is every main loop! At Cylc 7 I think it was part of the health check (i.e. every PT10M by default).

3) If you modify your global config to contain an invalid setting (after workflow startup) you get a traceback in the workflow log each time the plugin is run.

4) It's not enabled by default (because it's fairly heavyweight). However, there is no mention of this in the documentation for `[scheduler]auto restart delay`, `[scheduler][run hosts]condemned` or `[scheduler][run hosts]available`; nor in the section on Workflow Migration.

oliver-sanders:
> you get a traceback in the workflow log each time the plugin is run.

The issue here being the "traceback" part not the "each time the plugin is run".

PR #4904, opened 2022-06-06
main loop: improve defaults, docs and error handling

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/4873
  (in combination with https://github.com/cylc/cylc-doc/pull/475)
* There is now a default main loop plugin interval of `PT10M`.
  This prevents any plugins without a hardcoded default from being run with every main loop cycle.
* Explicitly set a default for the `auto restart` plugin (also `PT10M`).
* Added a note on enabling plugins.
* Suppress traceback from reloading the global config in the `auto restart` plugin.
  This is now logged nicely.
* Prevent plugins from being loaded twice when specified both on the CLI
  (--main-loop) and in the config (global.cylc[scheduler][main loop]plugins).

**Requirements check-list**
- [x] No change log entry required (why? docs and minor improvements).

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -652,7 +652,7 @@ with Conf('global.cylc', desc='''
                     starting new workflows.
 
                     Only enabled plugins are loaded, plugins can be enabled
-                    in three ways:
+                    in two ways:
 
                     Globally:
                        To enable a plugin for all workflows add it to:
@@ -676,7 +676,7 @@ with Conf('global.cylc', desc='''
                 .. note::
 
                    Only the configured list of :cylc:conf:`[..]plugins`
-                   are loaded when a scheduler is started.
+                   is loaded when a scheduler is started.
             ''') as MainLoopPlugin:
                 Conf('interval', VDR.V_INTERVAL, DurationFloat(600), desc='''
                     :Default For: :cylc:conf:`flow.cylc \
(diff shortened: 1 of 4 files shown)


cylc/flow/cfgspec/globalcfg.py
```diff
@@ -643,19 +643,42 @@
...
+            Conf(
+                'plugins',
+                VDR.V_STRING_LIST,
+                ['health check', 'reset bad hosts'],
+                desc='''
+                    Configure the default main loop plugins to use when
+                    starting new workflows.
+
+                    Only enabled plugins are loaded, plugins can be enabled
+                    in three ways:
```
hjoliver, line 655: By my count, that's two ways!

tests/unit/main_loop/test_auto_restart.py
```diff
@@ -280,3 +281,30 @@ def workflow_select_pass(**_):
...
+    )
+
+    # call the auto_restart plugin, the error should be caught
+    caplog.clear()
+    assert await auto_restart(None, None) is False
+
+    # the error should have been logged
+    assert len(caplog.messages) == 1
+    assert 'an error in the global config' in caplog.messages[0]
+    assert 'something event more bizarrely inexplicable' in caplog.messages[0]
```
hjoliver, line 310:
```suggestion
    assert 'something even more bizarrely inexplicable' in caplog.messages[0]
```

hjoliver (approved):
(Some minor comment and doc suggestions).

cylc/flow/cfgspec/globalcfg.py
oliver-sanders: Derp, I forgot that enabling per-workflow wasn't an option.

```suggestion
                    Only enabled plugins are loaded, plugins can be enabled
                    in two ways:
```

datamel (approved):
Code read, docs built and checked. Checked tests locally too. All good!

PR #4909, opened 2022-06-10
Fix task filtering.

hjoliver:
@oliver-sanders - I think this might be a bug in your task filtering function, found too late in the day to fully investigate - can you take a quick look, and if you agree it's a bug I'll finish off the PR?

If both the hidden pool (partially satisfied prereqs) and main pool are passed to the function - which they are, from `TaskPool.filter_task_proxies()` - it seems to return two copies of matched tasks in cycle points that appear in both pools.

These changes partially address #xxxx
These changes close #xxxx

**Requirements check-list**

### DIFF
--- a/cylc/flow/id_match.py
+++ b/cylc/flow/id_match.py
@@ -145,6 +145,8 @@ def filter_ids(
             if tokens.get(lowest_token.value):
                 break
 
+        # This needs to be a set to avoid getting two copies of matched tasks
+        # in cycle points that appear in both pools:
         cycles = set()
         tasks = []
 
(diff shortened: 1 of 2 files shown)


oliver-sanders:
Looks sensible to me, thanks.

hjoliver:
Added a test that fails on master.

cylc/flow/id_match.py
```diff
@@ -213,7 +215,7 @@ def filter_ids(
             if warn:
                 LOG.warning(f"No active tasks matching: {id_}")
         else:
-            _cycles.extend(cycles)
+            _cycles.extend(list(cycles))
```
oliver-sanders, line 218: The intermediate list is not necessary:

```suggestion
            _cycles.extend(cycles)
```

oliver-sanders (approved):
LGTM, cheers.

Issue #4910, opened 2022-06-10
CHECK_CIRCULAR_LIMIT


oliver-sanders:
The cyclic dependency check and by extension the un-satisfyable dependency check (which is invoked via `get_graph_raw`) are currently only performed when the number of tasks in the workflow is less than 100.

https://github.com/cylc/cylc-flow/blob/e4659ad20d713858d3c9802715e9b20bdc80a27e/cylc/flow/config.py#L164

The reason for this is that we had previously encountered workflows which took a long time to complete these checks, the number 100 is completely arbitrary. Note in a validate-install-play user pattern the circular dependency check is currently run twice (during validate & play). These slow validation results were spotted with older versions of Cylc 7 and have not been reproduced with Cylc 8 / Python 3.

We should test `cylc validate` against the workflows we can get hold of to find out how much resource the most demanding workflows are currently taking. Toggling `--check-circular` on and off should give us an idea how much burden it is. We can most likely bump the number up from 100 or make this test looser in some way. Note I would expect `get_graph_raw` to scale with the dependency density within the graph window evaluated, the number of tasks is just an easier metric.

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

hjoliver:
> Note I would expect get_graph_raw to scale with the dependency density within the graph window evaluated, the number of tasks is just an easier metric.

As I recall, `get_graph_raw` returns the whole graph between two cycle points, Cylc 7 stylz.

For the Cylc 8 graph view we'll need to generate n-window graphs around active tasks, regardless of cycle point.

In current validation we use `get_graph_raw` to generate the first 3 cycle points of the graph, and check that for circular dependencies. So here the error gets detected if `final cycle point = 3` but not if `final cycle point = 4`:

```ini
[scheduler]
   allow implicit tasks = True
[scheduling]
   cycling mode = integer
   initial cycle point = 1
   final cycle point = 3
   [[graph]]
      P1 = "a => b => c => d => e"
      R1/$ = "e => d"  # circular at $
```

What we should do, for the circular check, is generate the whole graph for the first point of every recurrence??

PR #4911, opened 2022-06-10
config: detect un-satisfiable dependencies

### LINKED ISSUE #4514: Task spawning and invalid cycle points
```ini
P2 = "foo[-P1] => bar"  # (sequence P2, with erroneous P1 offset)
```
^ this fails validation in both 7 and 8 with *no cycling sequences defined for foo*

But this:
```ini
P2 = """
   foo
   foo[-P1] => bar
"""
```
^ at Cylc 7:
- passes validation (a cycling sequence is defined for `foo`)
- `foo.1` and `bar.1` run (pre-initial ignore)
- `foo` runs at every `P2` cycle ... 
- ... until suite stalls with prespawned `bar`s waiting on non-existent off-sequence `foo`s

^ at Cylc 8:
- passes validation
- `foo.1` and `bar.1` run (pre-initial ignore)
- `foo` runs at every `P2` cycle
- it does not stall, and no `bar`s run

At first glance this is the expected result. Cylc 8 doesn't pre-spawn waiting task so we should not expect a stall here.

But on closer inspection...

At Cylc 7 the graph above is interpreted as :
- `foo` is valid at every `P2` point
- `bar` at every `P2` point depends on `foo` at non-existent off-sequence points

At Cylc 8, the same is true for prerequisite definition, but for spawning purpose (spawn on demand) the second line gets interpreted as:
- `foo => bar[+P1]`

So `bar` doesn't run (at Cylc 8) because `foo` on `P2` points tries to spawn `bar` at off-sequence points, which is technically not correct.

I don't think that matters in practice, because:
- it works correctly when there isn't an offset error in the graph
- the result is the same (albeit for a bass-ackward reason) when there is an offset error: the child `bar` doesn't run

And `bar` prerequisites are still correct if they get spawned bother another task:
```
P2 = """
    foo
    foo[-P1] => bar
    qux => bar
"""
^ This correctly stalls at Cylc 8, with `bar` on `P2` points waiting on non-existent off-sequence `foo`.
 
Still, it raises some questions:
- can we detect these problems at validation?
- can we detect that `foo[-P1] => bar` is invalid if `foo` only exists at `P2` points, to avoid attempting to interpret it incorrectly as `foo => bar[+P1]`

### LINKED ISSUE #4638: cylc graph: display "ghost" tasks
**Describe the bug**

The way Cylc Graph displays dependencies on tasks that don't exist has changed:

Cylc 7 - non-existant task appeared, but grey.
Now - non-existant task not plotted at all.

This workflow
```
[scheduler]
    UTC mode = True
    allow implicit tasks = True
[scheduling]
    initial cycle point = 20000101T00Z
    [[graph]]
        +PT6H/PT6H = """
            forecast[-PT6H] => forecast
        """
```
when graphed by `cylc graph . 2000 20000101T12Z` should display the following "ghost" dependency:

image

But doesn't.

**Pull requests welcome!**

**Cylc Doc Change Required**

May require re-write of exercise in `cylc-doc/src/tutorial/scheduling/datetime-cycling`

### PR DESCRIPTION
**Pull request by oliver-sanders**
get-graph-raw: check for unsatisfiable dependencies:

* Closes #4514 (@hjoliver to confirm)
* Re-purpose an unused field i `get_graph_raw` output to mark dependencies which cannot be satisfied due to the prerequisite task not existing in the graph at the required point.    
* This can happen due to typos or when an inter-cycle dependence is misaligned with the task's recurrence.    
* This will now warn by default (validate, play, etc), note, however, that this check will be skipped for workflows with >100 tasks (`CHECK_CIRCULAR_LIMIT`) see #4910.    
      
graph: render un-satisfiable dependencies as ghosts:

* Closes #4638    
* Fixes a regression where `cylc graph` was no longer rendering un-satisfiable dependencies differently (see #2346 for original implementation).               
* Restores the missing functionality to `cylc graph` and also adds warnings so un-satisfiable dependencies can be detected more easily.    
* Refactors `cylc graph` a little, main change is that graphviz output is no longer achieved by re-processing the reference output using regexes. It's now generated from the graph directly.   

**Requirements check-list**
- [x] (master branch) Documentation issue - https://github.com/cylc/cylc-doc/issues/477

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -1749,7 +1749,12 @@ class WorkflowConfig:
         return stop_point
 
     def get_graph_raw(
-            self, start_point_str=None, stop_point_str=None, grouping=None):
+        self,
+        start_point_str=None,
+        stop_point_str=None,
+        grouping=None,
+        warn=True,
+    ):
         """Return concrete graph edges between specified cycle points.
 
         Return a family-collapsed graph if the grouping arg is not None:
@@ -1876,9 +1881,12 @@ class WorkflowConfig:
                     else:
                         lstr, rstr = self._close_families(l_id, r_id, clf_map)
                         gr_edges[point].append(
-                            (lstr, rstr, None, suicide, cond))
+                            (lstr, rstr, True, suicide, cond))
                 # Increment the cycle point.
                 point = sequence.get_next_point_on_sequence(point)
+
+        # mark unsatisfyable gr_edges
+        self._detect_unsatisfiable_deps(gr_edges, is_validate, warn=warn)
 
         del clf_map
         del start_point_offset_cache
@@ -1896,12 +1904,82 @@ class WorkflowConfig:
         graph_raw_edges.sort(key=lambda x: [y if y else '' for y in x[:2]])
         return graph_raw_edges
 
+    @staticmethod
+    def _detect_unsatisfiable_deps(gr_edges, is_validate, warn=True):
+        """Detect un-satisfiable tasks in a graph.
+
+        If not `is_validate` the gr_edge will be modified to mark it as
+        un-satisfiable. This transformation is used by cylc graph to change the
+        nodes appearance.
+        """
+        un_satisfiable_edges = []
+        for edges in gr_edges.values():
+            for ind, edge in enumerate(edges):
+                if not edge:
+                    continue
+                if is_validate:
+                    # gr_edges are in the is_validate format
+                    (
+                        (left_task, left_point),
+                        (right_task, right_point)
+                    ) = edge
+                    if not any(
+                        # is the task present in the graph at this cycle?
+                        (left_task, left_point) in {left, right}
+                        for left, right in gr_edges.get(left_point, [])
+                    ):
+                        # ... no, add it to the list
+                        un_satisfiable_edges.append((
+                            Tokens(
+                                cycle=str(left_point),
+                                task=left_task,
+                            ).relative_id,
+                            Tokens(
+                                cycle=str(right_point),
+                                task=right_task,
+                            ).relative_id,
+                        ))
+
+                else:
+                    # gr_edges are in the regular format
+                    left_id, right_id, _, *tail = edge
+                    ltokens = Tokens(left_id, relative=True)
+                    left_task = get_point(ltokens['task'])
+                    left_point = get_point(ltokens['cycle'])
+                    if not any(
+                        # is the task present in the graph at this cycle?
+                        left_id in {_left_id, _right_id}
+                        for _left_id, _right_id, *_ in gr_edges.get(
+                            left_point, []
+                        )
+                    ):
+                        # ... no, add it to the list
+                        un_satisfiable_edges.append((left_id, right_id))
+                        # mark the dependency as un-satisfiable
+                        # (used by cylc graph to change the nodes appearance)
+                        edges[ind] = (
+                            left_id,
+                            right_id,
+                            False,
+                            *tail
+                        )
+
+        if warn and un_satisfiable_edges:
+            LOG.warning(
+                'Detected un-satisfiable dependencies:\n* '
+                + '\n* '.join(
+                    f'{left_id} => {right_id}'
+                    for left_id, right_id
+                    in sorted(un_satisfiable_edges)
+                )
+            )
+
     def get_node_labels(self, start_point_str=None, stop_point_str=None):
         """Return dependency graph node labels."""
         ret = set()
         for edge in self.get_graph_raw(
-                start_point_str,
-                stop_point_str,
+            start_point_str,
+            stop_point_str,
         ):
             left, right = edge[0:2]
             if left:
--- a/cylc/flow/scripts/graph.py
+++ b/cylc/flow/scripts/graph.py
@@ -40,7 +40,7 @@ from shutil import which
 from subprocess import Popen, PIPE
 import sys
 from tempfile import NamedTemporaryFile
-from typing import Dict, List, Optional, TYPE_CHECKING, Tuple
+from typing import Dict, List, Optional, TYPE_CHECKING, Tuple, Callable
 
 from cylc.flow.config import WorkflowConfig
 from cylc.flow.exceptions import InputError, CylcError
@@ -141,6 +141,9 @@ def _get_graph_nodes_edges(
         return [], []
 
     # set sort keys based on cycling mode
+    # (note sorting is very important for the reference format)
+    node_sort: Optional[Callable]
+    edge_sort: Optional[Callable]
     if config.cfg['scheduling']['cycling mode'] == 'integer':
         # integer sorting
         node_sort = sort_integer_node
--- a/cylc/flow/scripts/graph.py
+++ b/cylc/flow/scripts/graph.py
@@ -30,18 +30,20 @@ Examples:
 
     # display the difference between the flows one and two
     $ cylc graph one --diff two
+
+    # render the graph to a svg file
+    $ cylc graph one -o 'one.svg'
 """
 
 from difflib import unified_diff
-import re
 from shutil import which
 from subprocess import Popen, PIPE
 import sys
 from tempfile import NamedTemporaryFile
-from typing import List, Optional, TYPE_CHECKING, Tuple
+from typing import Dict, List, Optional, TYPE_CHECKING, Tuple
 
 from cylc.flow.config import WorkflowConfig
-from cylc.flow.exceptions import InputError
+from cylc.flow.exceptions import InputError, CylcError
 from cylc.flow.id import Tokens
 from cylc.flow.id_cli import parse_id
 from cylc.flow.option_parsers import (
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: cylc/flow/config.py [hunk 1]
```diff
@@ -1896,12 +1904,82 @@ def get_graph_raw(
...
+                        # (used by cylc graph to change the nodes appearance)
+                        edges[ind] = (
+                            left_id,
+                            right_id,
+                            False,
+                            *tail
+                        )
+
+        if warn and un_satisfiable_edges:
+            LOG.warning(
```
wxtim, line 1968: Is there any way of avoiding expanding this warning for every recurrence - IMO users would find
```bash
WARNING - Detected un-satisfiable dependencies:
    * foo[-PT1H] => foo [foo occurs at PT3H intervals]
```

more useful than

```bash
WARNING - Detected un-satisfiable dependencies:
    * 19831213T0800Z/foo => 19831213T0900Z/foo
    * 19831213T1100Z/foo => 19831213T1200Z/foo
```

tests/functional/graphing/12-unsatisfiable.t
```diff
@@ -0,0 +1,94 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test un-satisfiable dependencies are correctly identified in graphs
+. "$(dirname "$0")/test_header"
+#-------------------------------------------------------------------------------
+set_test_number 9
+init_workflow "${TEST_NAME_BASE}" <<'__FLOW_CONFIG__'
+[scheduler]
```
wxtim, line 23: Trivial, but manually testing I wondered whether I could be sure that ISOdatetime cycle points are handled by this logic. Manually tested.

tests/functional/graphing/12-unsatisfiable.t
```diff
@@ -0,0 +1,94 @@
...
+# cylc validate should log warnings for the un-satisfiable dependencies
+TEST_NAME="${TEST_NAME_BASE}-validate"
+run_ok "${TEST_NAME}" cylc validate "${WORKFLOW_NAME}"
+grep_ok "Detected un-satisfiable dependencies" "${TEST_NAME}.stderr"
+grep_ok "1/a => 3/b" "${TEST_NAME}.stderr"
+grep_ok "2/b => 4/a" "${TEST_NAME}.stderr"
+
+#-------------------------------------------------------------------------------
+# cylc graph should log warnings for the un-satisfiable dependencies
+TEST_NAME="${TEST_NAME_BASE}-graph"
```
wxtim, line 45: IMO this should be a separate test - you are checking the output as a dot file. 

I presume that the other two output formats in the `--help` docs are potentially dependent on the environment, and that tests would realistically be limited to checking that the files are a png/svg. If this isn't the case can we have a test, especially for the SVG, which is after all a text file.

wxtim (commented):
- [x] Read code
- [x] Did some manual checks

cylc/flow/scripts/graph.py
```diff
@@ -96,22 +98,47 @@ def sort_datetime_edge(item):
     return (item[0], item[1] or '')
 
 
-def graph_workflow(
+# node/edge types
+Node = str  # node ID
+Edge = Tuple[str, str, bool]   # left, right, left_node_exists
+
+
+def get_nodes_and_edges(
```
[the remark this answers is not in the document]
oliver-sanders: The coverage is low because a couple of the tests have failed.

Because this loads workflow configs it's not suitable for a doctest/example.

cylc/flow/config.py
oliver-sanders: > Is there any way of avoiding expanding this warning for every recurrence

The exact cycle points might be useful debugging information to help the user determine which recurrence is causing the problem. E.G. if the problem cycle points are PT6H apart then the user knows to look for a recurrence with a PT6H pattern.

tests/functional/graphing/12-unsatisfiable.t
oliver-sanders: It's checking the same functionality against the same config so I think these two should be in together.

I haven't tested formats like SVG and PNG because that conversion is done by graphviz and testing graphviz output feels out of scope.

oliver-sanders:
I don't think codecov is working correctly, take a look at the coverage report for this PR, definitely wrong!

Screenshot 2022-06-20 at 11 00 47

oliver-sanders:
Added unit tests for output formats, code will already be covered by `cylc graph` tests but doesn't hurt.

dwsutherland (approved):
Looks good 👍 
**before:**
```
(flow) sutherlander@graphic-vbox:one$ cylc validate one
Valid for cylc-8.0rc4.dev
```
**after:**
```
(flow) sutherlander@graphic-vbox:one$ cylc validate one
WARNING - Detected un-satisfiable dependencies:
    * 1/a => 3/b
    * 2/b => 4/a
    * 3/a => 5/b
    * 4/b => 6/a
Valid for cylc-8.0rc4.dev
```
image

image

dwsutherland:
Overall coverage looks positive;
https://app.codecov.io/gh/cylc/cylc-flow/compare/4911/overview
Mainly just flagged things in the graph script:
https://github.com/cylc/cylc-flow/pull/4911/checks?check_run_id=7015941151

tests/functional/graphing/12-unsatisfiable.t
```diff
@@ -0,0 +1,94 @@
...
+run_ok "${TEST_NAME}" cylc graph "${WORKFLOW_NAME}" 1 5 -o "${TEST_NAME}.dot"
+grep_ok "Detected un-satisfiable dependencies" "${TEST_NAME}.stderr"
```
hjoliver, line 47:
```suggestion
grep_ok "Detected unsatisfiable dependencies" "${TEST_NAME}.stderr"
```

cylc/flow/config.py
```diff
@@ -1876,10 +1881,13 @@ def get_graph_raw(
                     else:
                         lstr, rstr = self._close_families(l_id, r_id, clf_map)
                         gr_edges[point].append(
-                            (lstr, rstr, None, suicide, cond))
+                            (lstr, rstr, True, suicide, cond))
                 # Increment the cycle point.
                 point = sequence.get_next_point_on_sequence(point)
 
+        # mark unsatisfyable gr_edges
```
hjoliver, line 1888:
```suggestion
        # mark unsatisfiable gr_edges
```

hjoliver (approved):
Looks good.

Minor issue: pretty sure "unsatisfiable" (like "unsatisfied", "unjustifiable" etc.) shouldn't be hyphenated.

hjoliver (changes requested):
Actually, sorry, problem struck.

It's similar but different to the circular dependency check problem I commented on here: https://github.com/cylc/cylc-flow/issues/4910#issuecomment-1158442772

This branch works with the first `WorklowConfig.VIS_N_POINTS` (3) points of each recurrence. 

Unfortuately that means it can "detect" spurious unsatisfiable dependencies if the inter-cycle dependence doesn't span that gap.

Example:
```ini
[scheduler]                                                                                                                                                                                                          
   allow implicit tasks = True                                                                                                                                                                                       
[scheduling]                                                                                                                                                                                                         
   cycling mode = integer                                                                                                                                                                                            
   initial cycle point = 1
   final cycle point = 6                                                                                                                                                                                           
   [[graph]]                                                                                                                                                                                                         
      P2 = "a => b => c => d => e"  # 1, 3, 5, 7, 9, ...    
      R1/$ = "e[-P1] => f"  # valid for any EVEN final point
```
This incorrectly flags an unsatisfiable dependency for any even final point `>= 8`:
```
WARNING - Detected un-satisfiable dependencies:
    * 7/e => 8/f
```

It's only a warning, but still, not really acceptable to warn about a non-existent problem.

hjoliver:
**Alternative approach?**

For the first valid point in every recurrence, try to instantiate the parent of any intercycle trigger. If that fails, the parent is not valid at that cycle point and we have an unsatisfiable dependency.

oliver-sanders:
> any inter-cycle trigger that doesn't span that gap

I think that should be solvable.

> For the first valid point in every recurrence, try to instantiate the parent task in any intercycle trigger

I think this would miss a lot of the sort of problems we need it to help diagnose. Misaligned recurrences, etc. Awkward example:

```
1/P2!4  = a
2/P2   = a[-P1] => b
```

Having a head scratch on this one ATM. I don't like the limitation that it can only detect these dependencies within the window get-graph-raw constructs. It feels like there should be some mathematical solution.

For example it's easy to prove this example:

```ini
1/P2 = a => b         # R1
2/P2 = a => c         # R2
1/P1 = a[-P1] => z    # R3
```
The inter-cycle dependency spans from `2/P1` to `(2 - P1)/P1 = 1/P1`.
We know this is fine since `1/P2 ∪ 2/P2 = 1/P1`.

I.E. we need to determine whether the following expression resolves to an empty set, we don't need to exhaustively trawl through each cycle point to do so:

```
# the set of cycle points which belong to R1 but not to the others
R1 \ ( R2 ∪ R2 ∪ ... RN )
```

Problem being that `R1..RN` are sequences, with anything other than with trivial examples deriving a simplified union of two sequences is beyond my ability. E.G. consider the following:

```ini
P1W
P1DT6H
min(T00, T06)
P1D ! 20000101
T00 ! (P3D, P5D)  # yikes who thought excluded sequences were a good idea?!
```

hjoliver:
> I think this would miss a lot of the sort of problems we need it to help diagnose. Misaligned recurrences, etc. 

It would catch all that affect the first point of a recurrence, which might be the majority in practice (if exclusions aren't involved, at least). But yeah, it wouldn't catch everything. That would be better than nothing if a general solution proves difficult though.

oliver-sanders:
> It would catch all that affect the first point of a recurrence, which might be the majority in practice 

I think it's quite common for the error *not* to be the first point of the recurrence, e.g. the example which started this:

http://metomi.github.io/rose/doc/html/tutorial/cylc/scheduling/datetime-cycling.html#admonition-4

hjoliver:
> Sadly I've gotta feeling that the general solution would be tough enough to prove by hand and fiendish to automate.

Yeah I agree, unfortunately.

How about this, as an imperfect solution that is nevertheless useful and won't generate false positives:

It is easy to find unsatisfiable dependencies by trying to instantiate a task (or graph node) instance at a specific cycle point, because tasks know the recurrences they belong to and can plug in their own cycle point to determine validity. So:

- find the intercycle triggers in each graph string
- attempt to instantiate the tasks involved, for some range of cycle points

By default, we could use:
- the first 5 cycle points in the workflow
- or the first 5 points in each recurrences
- or whatever

But allow users to specify a different range of points (as a validation arg) for a more complete but potentially slower check.

hjoliver:
Not also #4515 - we should still log problems of this kind at run time, unless we can come up a full solution that prevents it from ever happening.

hjoliver:
Bumping this back to 8.1, since we don't have an agreed solution yet. Probably worth getting #4515 in for 8.0 though, to help users with runtime diagnosis at least.

oliver-sanders:
Ditching this attempt for now...

PR #4921, opened 2022-06-21
cli: updated cylc term logo

oliver-sanders:
Attempt at improving the readability of the in-terminal Cylc logo using unicode "block elements". I think these chars are the only dependably full width/height chars.

### Before
Screenshot from 2022-06-21 18-00-19

### After

Screenshot from 2022-06-21 18-00-47

Might try to improve the kerning on the "L".

Thoughts?

### DIFF
--- a/cylc/flow/scripts/common.py
+++ b/cylc/flow/scripts/common.py
@@ -16,7 +16,6 @@
 
 """Shared utilities for Cylc scripts."""
 
-from re import sub
 from itertools import zip_longest
 from ansimarkup import strip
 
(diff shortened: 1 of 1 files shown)


hjoliver:
Symbolic logos might resolve the impasse - PR posted to your branch.

I tried two approaches, not sure which I liked best so I've gone for a random choice for each run.

Cylc logo (kind of):
Screenshot from 2022-07-04 12-31-54

Job and task logos:
shot-1

MetRonnie:
Imho the 4-block logo isn't bad

hjoliver:
(Another thought, for added interest - if we kept the array of job/task icons, we could randomly assign their grid positions each time.)

datamel:
This one is appealing to me as the circles are very google assistant. I'm sure Cylc was first though but I can't help the association.

hjoliver:
I think I prefer the squares too. (Maybe we should switch the official logo to that, in light of the google problem).

hjoliver:
@oliver-sanders - By my reading of the responses above the block logo wins. I'll close this as superseded by #4985, to simplify process a bit (no need to merge my side-PR and strip out the no-go options).

PR #4924, opened 2022-06-22
uid: warn if the run number is provided as a cycle

### LINKED ISSUE #4923: Detection of worklfow//task ID errors
This from an early Cylc 8 user (no training yet):

> I think your `cat-log` command is broken for printing out outputs/errors from a particular task. It has:
>
```console
$ cylc cat-log -f o waikanae//run1/setup
file not found: /home/xxx/cylc-run/waikanae/run1/log/job/run1/setup/NN/job.out
```
> when it should be
```console
/home/xxx/cylc-run/waikanae/run1/log/job/1/setup/NN/job.out
```
> At least based on the folders named - i.e. the second `run1` should just be `1`

We should be able to detect erroneous positioning of the double `//` in the command line above, since cycle points can't (currently at least!) be called `run1`, and print an informative error message.

oliver-sanders:
* E.G. workflow//run/cycle rather than workflow/run//cycle.
* Closes #4923

Screenshot from 2022-06-22 13-12-58

**Requirements check-list**
- [x] No change log entry required (minor change)

### DIFF
--- a/cylc/flow/id_cli.py
+++ b/cylc/flow/id_cli.py
@@ -14,13 +14,13 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-
 import asyncio
 import fnmatch
 from pathlib import Path
 import re
 from typing import Optional, Dict, List, Tuple, Any
 
+from cylc.flow import LOG
 from cylc.flow.exceptions import (
     InputError,
     WorkflowFilesError,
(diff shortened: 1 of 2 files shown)


hjoliver (approved):
Nice, thanks @oliver-sanders :+1:

dwsutherland:
Did we not generalize to checking whether it is a valid cycle point because it would involve parsing in the workflow def/config?

hjoliver:
Yeah, but I guess without parsing the workflow config we could check for valid integer or ISO8601 datetime (assuming those are the only valid cycle point types, but that's fine for now). Might be worth it as a follow-up @oliver-sanders ?

oliver-sanders:
> Did we not generalize to checking whether it is a valid cycle point

Invalid cycle points will get detected when they are passed through the internal `get_point` interface.

The UUID system does not attempt to validate cycle points, partly because it's a simple string parser, string in, string out, but also because:

* We don't know what cycle mode to use without parsing the workflow configuration.
* We may support more cycle formats one day, we could potentially support arbitrary parameters as cycle points.

Issue #4902, opened 2022-06-01
Improve schema docs for `broadcast` mutation


MetRonnie:
- [ ] The description for the `broadcast` mutation in `cylc.flow.network.schema` seems copied from `cylc.flow.scripts.broadcast` without much modification. It ought to be refined, seeing as how the user interacts with the mutation form in the UI is quite different from using `cylc broadcast`. At present, it is unclear how the `Clear` and `Expire` modes work in the UI.
- [ ] It is unclear how to use `settings` in the mutation form in the UI

**Pull requests welcome!**

oliver-sanders:
Moving the UI aspects of this to cylc-ui (cylc-flow defines the schema not the forms) into umbrella issue for improving UI forms: https://github.com/cylc/cylc-ui/issues/647

oliver-sanders:
cylc-flow actually has quite good documentation for BroadcastSettings, however, this is not displaying in the UI help.

https://github.com/cylc/cylc-flow/blob/4a1ed03fcd5e1223127bfa4cd11fe29360675c20/cylc/flow/network/schema.py#L1284-L1298

oliver-sanders:
For info, here's how the API-On-The-Fly (aotf) system works.

We define GraphQL types in cylc-flow:

```
# cylc.flow.network.schema

class CyclePoint(String):                                                          
    """An integer or date-time cyclepoint."""
```

These docstrings become the description field of the type:

```graphql
query {
  __schema {
    types {
      name
      description
    }
  }
}
```
```json
{
  "data": {
    "__schema": {
      "types": [
        {
          "name": "CyclePoint",
          "description": "An integer or date-time cyclepoint."
        },
```

(run this introspection query in GraphiQL to find out more).

The aotf system pulls these descriptions into the help icons:

![Screenshot from 2022-06-23 16-28-35](https://user-images.githubusercontent.com/16705946/175337008-8c124c15-944f-44cf-abca-2c8c8407b656.png)

PR #4933, opened 2022-06-23
schema: small docs changes

oliver-sanders:
Small changes to the broadcast mutation docs.

Closes the cylc-flow side of #4902, the (larger) cylc-ui portion has been moved to https://github.com/cylc/cylc-ui/issues/647.

**Requirements check-list**

### DIFF
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -1415,7 +1415,7 @@ class Broadcast(Mutation):
 
             A broadcast can set/override any `[runtime]` configuration for all
             cycles or for a specific cycle. If a task is affected by
-            specific-cycle and an all-cycle broadcasts at the same time, the
+            specific-cycle and all-cycle broadcasts at the same time, the
             specific takes precedence.
 
             Broadcasts can also target all tasks, specific tasks or families of
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -1275,16 +1275,22 @@ class BroadcastMode(graphene.Enum):
         if self == BroadcastMode.Set:
             return 'Create a new broadcast.'
         if self == BroadcastMode.Clear:
-            return 'Revoke an existing broadcast.'
+            return (
+                'Clear existing broadcasts globally or for specified'
+                ' cycle points and namespaces if provided.'
+            )
         if self == BroadcastMode.Expire:
-            return 'Expire an existing broadcast.'
+            return (
+                'Clear all broadcasts for cycle points earlier than the'
+                ' provided cutoff.'
+            )
         return ''
 
 
 class BroadcastSetting(GenericScalar):
-    """A [runtime] key=value configuration for a namespace.
+    """A `[runtime]` configuration for a namespace.
 
-    Should be a key=value pair where sections are wrapped with square
+    Should be a `key=value` pair where sections are wrapped with square
     brackets.
 
     Examples:
@@ -1398,45 +1404,38 @@ class Flow(String):
 
 # Mutations:
 
-# TODO: re-instate:
-# - get-broadcast (can just use GraphQL query BUT needs CLI access too)
-# - expire-broadcast
-
 class Broadcast(Mutation):
     class Meta:
         description = sstrip('''
-            Override or add new `[runtime]` configurations in a running
-            workflow.
+            Override `[runtime]` configurations in a running workflow.
 
             Uses for broadcast include making temporary changes to task
             behaviour, and task-to-downstream-task communication via
             environment variables.
 
-            A broadcast can target any [runtime] namespace for all cycles or
-            for a specific cycle. If a task is affected by specific-cycle and
-            all-cycle broadcasts at once, the specific takes precedence. If
-            a task is affected by broadcasts to multiple ancestor
-            namespaces, the result is determined by normal [runtime]
-            inheritance. In other words, it follows this order:
+            A broadcast can set/override any `[runtime]` configuration for all
+            cycles or for a specific cycle. If a task is affected by
+            specific-cycle and an all-cycle broadcasts at the same time, the
+            specific takes precedence.
 
-            `all:root -> all:FAM -> all:task -> tag:root -> tag:FAM ->
-            tag:task`
+            Broadcasts can also target all tasks, specific tasks or families of
+            tasks. If a task is affected by broadcasts to multiple ancestor
+            namespaces (tasks it inherits from), the result is determined by
+            normal `[runtime]` inheritance.
 
-            Broadcasts persist, even across restarts, until they expire
-            when their target cycle point is older than the oldest current in
-            the workflow, or until they are explicitly cancelled with this
-            command.  All-cycle broadcasts do not expire.
+            Broadcasts are applied at the time of job submission.
 
-            For each task the final effect of all broadcasts to all namespaces
-            is computed on the fly just prior to job submission.  The
-            `--cancel` and `--clear` options simply cancel (remove) active
-            broadcasts, they do not act directly on the final task-level
-            result. Consequently, for example, you cannot broadcast to "all
-            cycles except Tn" with an all-cycle broadcast followed by a cancel
-            to Tn (there is no direct broadcast to Tn to cancel); and you
-            cannot broadcast to "all members of FAMILY except member_n" with a
-            general broadcast to FAMILY followed by a cancel to member_n (there
-            is no direct broadcast to member_n to cancel).
+            Broadcasts persist, even across restarts. Broadcasts made to
+            specific cycle points will expire when the cycle point is older
+            than the oldest active cycle point in the workflow.
+
+            Active broadcasts can be revoked using the "clear" mode.
+            Any broadcasts matching the specified cycle points and
+            namespaces will be revoked.
+
+            Note: a "clear" broadcast for a specific cycle or namespace does
+            *not* clear all-cycle or all-namespace broadcasts.
+
         ''')
         resolver = partial(mutator, command='broadcast')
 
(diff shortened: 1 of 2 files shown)


cylc/flow/network/schema.py
```diff
@@ -1429,7 +1429,7 @@ class Meta:
 
             For each task the final effect of all broadcasts to all namespaces
             is computed on the fly just prior to job submission.  The
-            `--cancel` and `--clear` options simply cancel (remove) active
+            "clear" mode cancels (removes) active
             broadcasts, they do not act directly on the final task-level
             result. Consequently, for example, you cannot broadcast to "all
```
MetRonnie, line 1434:
```suggestion
            broadcasts, it cannot act directly on the final task-level
            result. Consequently, for example, you cannot broadcast to "all
```

MetRonnie (commented):
How does the "expire" mode work?

cylc/flow/network/schema.py
```diff
@@ -1282,9 +1282,9 @@ def description(self):
 
 
 class BroadcastSetting(GenericScalar):
```
MetRonnie, line 1284: Wouldn't making this a `String` type fix the lack of ❔ button?
```suggestion
class BroadcastSetting(String):
```

oliver-sanders: No, the reason the help button isn't showing is because the UI has a special component for broadcast settings which itself does not support the help button, easy fix - https://github.com/cylc/cylc-ui/pull/1058

oliver-sanders:
Added more explanatory docs to the broadcast modes and rebased.

cylc/flow/network/schema.py
oliver-sanders: The subject is "broadcasts" not "Broadcast" so the plural is correct?

MetRonnie: I thought the subject was "the clear mode"

hjoliver: I think "they" referred to "the --cancel and --clear options", which you've replaced with a single "clear mode", so I think @MetRonnie is right.

oliver-sanders: Ok.

On second reading this seems a bit dense so I've re-worked it.

cylc/flow/network/schema.py
```diff
@@ -1398,45 +1404,43 @@ class Flow(String):
...
-            result. Consequently, for example, you cannot broadcast to "all
-            cycles except Tn" with an all-cycle broadcast followed by a cancel
-            to Tn (there is no direct broadcast to Tn to cancel); and you
-            cannot broadcast to "all members of FAMILY except member_n" with a
-            general broadcast to FAMILY followed by a cancel to member_n (there
-            is no direct broadcast to member_n to cancel).
+            A broadcast can set/override any `[runtime]` configuration for all
+            cycles or for a specific cycle. If a task is affected by a
+            specific-cycle and an all-cycle broadcasts at the same time, the
+            specific takes precedence.
```
MetRonnie, line 1419:
```suggestion
            specific-cycle broadcast and an all-cycle broadcast at the same time,
            the specific-cycle one takes precedence.
```

cylc/flow/network/schema.py
```diff
@@ -1398,45 +1404,43 @@ class Flow(String):
...
+
+            Broadcasts can also target all tasks, specific tasks or families of
+            tasks. If a task is affected by broadcasts to multiple ancestor
+            namespaces (tasks it inherits from), the result is determined by
+            normal `[runtime]` inheritance. In other words, it follows this
+            order:
+
+            ```
+            all:root -> all:FAM -> all:task -> tag:root -> tag:FAM -> tag:task
+            ```
```
MetRonnie, line 1429: Unfortunately this overflows the mutation form: 



```suggestion
            `all:root -> all:FAM -> all:task -> tag:root -> tag:FAM ->
            tag:task`
```

cylc/flow/network/schema.py
```diff
@@ -1398,45 +1404,43 @@ class Flow(String):
...
+            specific cycle points will expire when the cycle point is older
+            than the oldest active cycle point in the workflow.
+
+            Active broadcasts can be revoked using the "clear" mode.
+            Any broadcasts matching the specified cycle points and
+            namespaces will be revoked.
+
+            Note you can't broadcast to all cycles or namespaces *except* X
+            by broadcasting to all cycles and then clearing that broadcast for
+            cycle X.
```
MetRonnie, line 1443: I still find this a bit confusing.
```suggestion
            namespaces will be revoked. This means you cannot remove
            individual cycles/namespaces from an all-cycle/namespace broadcast.
```

cylc/flow/network/schema.py
```diff
@@ -1453,24 +1457,28 @@ class Arguments:
...
-                The cylc namespace for the setting to modify.
+                The `[runtime]` configuration to override with this broadcast
+                (e.g. `script = true` or `[environment]ANSWER=42`).
             '''),
             # e.g. `{environment: {variable_name: "value",. . .}. . .}`.
         )
         cutoff = CyclePoint(
-            description='Clear broadcasts earlier than cutoff cycle point.'
+            description=(
+                'Clear broadcasts earlier than cutoff cycle point.'
```
MetRonnie, line 1479:
```suggestion
                'Clear broadcasts earlier than cutoff cycle point'
```

cylc/flow/scripts/broadcast.py
```diff
@@ -18,34 +18,39 @@
...
+all-cycle broadcasts at the same time, the specific takes precedence.
+
+Broadcasts can also target all tasks, specific tasks or families of tasks. If a
+task is affected by broadcasts to multiple ancestor namespaces (tasks it
+inherits from), the result is determined by normal `[runtime]` inheritance. In
+other words, it follows this order:
+
+```
 all:root -> all:FAM -> all:task -> tag:root -> tag:FAM -> tag:task
+```
```
MetRonnie, line 39: This is for the CLI so markdown doesn't work
```suggestion
all:root -> all:FAM -> all:task -> tag:root -> tag:FAM -> tag:task
```

cylc/flow/network/schema.py
hjoliver: The original wording suggests you can do that, by broadcasting to all, and then clearing one.

Maybe phrase it positively? Something like this: *to exclude specific cycles or namespaces from a general broadcast, target them all then explicitly clear the specific ones*

MetRonnie: > ... _to exclude specific cycles or namespaces from a general broadcast, target them all then explicitly clear the specific ones_

My reading was that you **cannot** do that

cylc/flow/network/schema.py
oliver-sanders: Decided to remove this block.

cylc/flow/network/schema.py
oliver-sanders: > My reading was that you cannot do that

Correct, but since it caused confusion I've reworded more clearly.

hjoliver: Oops, you're right. Actually I don't think it was confusing, but my brain evidently reordered the original text on the fly to get *you can't do blah X except by blah* rather than *you can't do "blah except X" by blah* (more or less).

cylc/flow/network/schema.py
```diff
@@ -1275,16 +1275,22 @@ def description(self):
...
         if self == BroadcastMode.Clear:
-            return 'Revoke an existing broadcast.'
+            return (
+                'Clear existing broadcasts globally or for specified'
+                ' cycle points and namespaces if provided.'
+            )
         if self == BroadcastMode.Expire:
-            return 'Expire an existing broadcast.'
+            return (
+                'Clear all broadcasts for cycle points earlier the provided'
```
hjoliver, line 1284:
```suggestion
                'Clear all broadcasts for cycle points earlier than the provided'
```

cylc/flow/scripts/broadcast.py
```diff
@@ -18,34 +18,33 @@
...
+Broadcasts persist, even across restarts. Broadcasts made to specific cycle
+points will expire when the cycle point is older than the oldest active cycle
+point in the workflow.
+
+Active broadcasts can be revoked using the "--clear" option. Any broadcasts
+matching the specified cycle points and namespaces will be revoked. The
+"--cancel" option can cancel only specific configurations from a broadcast.
+
+Note: a "clear" broadcast for a specific cycle or namespace does *not* clear
+all-cycle or all-namespace broadcasts.
```
hjoliver, line 47: Should reuse the schema text here? (with a few tweaks via string formatting)

hjoliver (approved):
Looks good, couple of minor changes suggested. The broadcast text is definitely clearer :+1:

cylc/flow/network/schema.py
MetRonnie: then
```suggestion
            specific-cycle and all-cycle broadcasts at the same time, the
            specific takes precedence.
```

cylc/flow/network/schema.py
```diff
@@ -1398,45 +1404,38 @@ class Flow(String):
...
+            Broadcasts persist, even across restarts. Broadcasts made to
+            specific cycle points will expire when the cycle point is older
+            than the oldest active cycle point in the workflow.
+
+            Active broadcasts can be revoked using the "clear" mode.
+            Any broadcasts matching the specified cycle points and
+            namespaces will be revoked.
+
+            Note: a "clear" broadcast for a specific cycle or namespace does
+            *not* clear all-cycle or all-namespace broadcasts.
```
MetRonnie, line 1437: Maybe 
```suggestion
            *not* affect all-cycle or all-namespace broadcasts.
```

cylc/flow/scripts/broadcast.py
```diff
@@ -18,34 +18,32 @@
...
+Broadcasts are applied at the time of job submission.
+
+Broadcasts persist, even across restarts. Broadcasts made to specific cycle
+points will expire when the cycle point is older than the oldest active cycle
+point in the workflow.
+
+Active broadcasts can be revoked using the "clear" mode. Any broadcasts
+matching the specified cycle points and namespaces will be revoked.
+
+Note: a "clear" broadcast for a specific cycle or namespace does *not* clear
```
MetRonnie, line 45: If using the corresponding suggestion above
```suggestion
Note: a "clear" broadcast for a specific cycle or namespace does *not* affect
```

wxtim:
Replaced by https://github.com/cylc/cylc-doc/pull/498

oliver-sanders:
(wrong PR)

PR #4937, opened 2022-06-27
bash completion

### LINKED ISSUE #4566: CLI: Upgrade auto-completion for Cylc 8
CLI auto-complete is broken.

UID makes it possible for us to complete much more including hierarchical workflow IDs, cycles, tasks and jobs.

Should be relatively straight-forward (famous last words).

See the cylc-admin proposal for details:

https://github.com/cylc/cylc-admin/blob/master/docs/proposal-universal-id.md

Quick notes:
* The previous auto-complete suffered from performance issues as it used the filesystem to list Cylc sub-commands. We can now load the commands via Python. We should investigate cashing these to improve performance.
* Where  patterns are used (e.g. `*`) it would be good if the completion could auto-quote IDs to ensure the matching is being performed server-side.
* Suggest only implementing Bash for now 🐟 🐠 🎣.

**Pull requests welcome!**

oliver-sanders:
Closes #4566 

Surprisingly responsive auto-completion for Bash with scope for extension to other shells.

bash-completion

> The POC was so straight forward it only took a few lines to turn it into a fully-functioning prototype!

**A new Bash completion script which completes:**
* Sub-commands.
* Command options.
* Workflow IDs.
* Cycle points.
* Tasks.
* Jobs.
* A small selection of other things...

```console
$ cylc c<tab>
cat-log            clean              completion-server  cycle-point
check-versions     client             config             
$ cylc cat-log --<tab>
--file           --geditor        --mode           --rotation
--force-remote   --help           --remote-arg     --submit-number
$ cylc cat-log --mode=cat <tab><tab><tab>
code-tests/r12345/run1//     cylctb-xatzd3w/run1//
code-tests/r12345/run2//     cylctb-xxF2lYD//
code-tests/r23456/run1//     mi-aa001/aircraft_global//
code-tests/r34567/run1//     mi-aa001/aircraft_ukv//
code-tests/r34567/run2//     mi-aa001/land_surface_ukv//
code-tests/r34567/run3//     mi-aa001/marine_global//
cylctb-xCgu_Hq//             one//
cylctb-xIYJwFM//             
$ cylc cat-log --mode=cat one//1/foo/<tab>
one//1/foo/01/  one//1/foo/NN/  
$ cylc cat-log --mode=cat one//1/foo/NN  # woohoo!
```

**How it works:**
* The "completions" are provided by the `cylc completion-server`.
* When you first type `cylc<tab>` a completion server is started up as a Bash "coprocess".
* Coprocesses are process you can write to and read from at the same time.
* When you press `<tab>` the CLI string is sent to the server (by writing to it).
* The completion script then dumps its responses from the server into the format Bash is expecting.
* The first completion is slow (because you have to wait for the server to start).
* All subsequent completions are fast (because the server is already running).

**Advantages of the coprocess approach:**
* Write the business logic in Python not Bash!
* Use existing Cylc interfaces in the completion (e.g. `cylc scan`, ID parsing, run dir location)
* Test completions using `pytest`.
* Scanning is quick without the need to start a new Python process / load the Cylc codebase.
* The same completion server can provide completions for different shells (Bash, zsh, Fish, etc).

> Note: Bash 4.2+

**Requirements check-list**
- [x] Appropriate tests are included (100% coverage)


cylc/flow/scripts/completion_server.py
```diff
@@ -0,0 +1,673 @@
...
+        once:
+            Return after sending the first response.
+        write:
+            The function which sends the response back to the originator.
+        timeout:
+            Max time to wait for new requests.
+
+    Note:
+        The server will not play nicely with completions which contain spaces
+        (they will show as separate compleions).
```
hjoliver, line 131:
```suggestion
        (they will show as separate completions).
```

hjoliver (commented):
Took it for a test drive - brilliant!

A couple of initial minor thoughts: 
- completion of task IDs goes down to submit number, although that usually isn't valid
- can it be made to handle latest run inference?  If it type `cylc trigger foo/run2//` it works, but not `foo//`
- could it print a message for commands than can take arguments beyond what the completion suggests (e.g. trigger completes for existing cycle points and tasks, but the command can target any cycle point and task). Or does everything it prints have to be a potential command line?

oliver-sanders:
> completion of task IDs goes down to submit number, although that usually isn't valid

This would require the completion to introspect Cylc commands to see what kind of IDs they are expecting (i.e. workflow, cycle, task or job). Possible but tricky (and a bit of a flaw in the existing CLI ATM), for the moment it just completes the ID as far as the user takes it.

> can it be made to handle latest run inference? If I type cylc trigger foo/run2// it works, but not foo//

Hmm, does this mean the trailing `//` in `foo//`disables run inferrence? If so that's probably something to fix in the ID parser.

> could it print a message for commands than can take arguments beyond what the completion suggests

The completion server does not "print", it returns a list of possibilities to the shell which the shell then formats for the user. So every item we return must be a valid completion.

That said we can print to stderr and the completion will do this on startup e.g. if a new version of the completion script is available, however, I've kept this minimal.

oliver-sanders:
One issue left, getting the cylc-completion.bash script to handle the case where the `cylc completion-server` command returns 1 undetectably (e.g. if your wrapper script is pointing at Cylc 7 rather than Cylc 8).

oliver-sanders:
Tests seem to be failing for arbitrary sorting reasons I can't reproduce locally (probably relying on fs listing order?), will sort that out.

oliver-sanders:
(rebased)

hjoliver:
(I'll take another look at this soon - I'm keen to use it)

oliver-sanders:
(ditto, major timesaver)

tests/functional/rnd/06-completion-server.py
```diff
@@ -0,0 +1,47 @@
...
+run_ok "${TEST_NAME}" cylc completion-server --once <<< 'cylc|trigg'
+cmp_ok "${TEST_NAME}.stdout" << __HERE__
+trigger
+__HERE__
+
+# Make sure the server exits timeout when trying to read from stdin
+# (Note the completion server exits 0 on timeout)
+
+TEST_NAME="${TEST_NAME_BASE}-timeout"
+run_ok "${TEST_NAME}" timeout 5 cylc completion-server --timeout=1
```
datamel, line 45: I take it the difference in the timeouts is necessary to provide sufficient buffer to avoid this being a flaky test?

datamel (commented):
No major problems found by me, checked out and heavily manually tested. A couple of small comments. Great change and very fancy implementation!

tests/functional/rnd/06-completion-server.py
oliver-sanders: * The `--timeout=1` is to test that the server timeout works correctly.
* The `timeout 5` is to prevent the test from hanging if the `--timeout=1` doesn't timeout.

tests/functional/rnd/06-completion-server.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
oliver-sanders: Yes, not sure what I did there, should be a `.t` file.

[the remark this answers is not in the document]
oliver-sanders: Done.

oliver-sanders:
Addressed last review point, rebased.

hjoliver:
> [edit] I think inferrence is working correctly for me, what are you seeing?

If I deliberately terminate the commandline with `//` after the workflow name and then hit TAB, the server doesn't suggest completions based on a `runN` inference.

```console
$ cylc scan --state=all -n bug
bug/run1
bug/run2 niwa-1007885.niwa.local:43018

$ cylc trigger bug/run2//<TAB>
bug/run2//1/  bug/run2//3/  bug/run2//5/  bug/run2//7/  
bug/run2//2/  bug/run2//4/  bug/run2//6/  

$ cylc trigger bug//<TAB>
<NOTHING>
```
Not a big deal at all, since starting to TAB at the name forces me down the explicit run-number route.  But a possible small follow-up perhaps.

hjoliver (approved):
Works great; code LGTM; performance is snappy. :tada:

@datamel - I think this can go in if you're happy now. It's languished long enough, and it can't break anything else so we don't need to be too fussy about code review.

I'll comment below on several small points, but they can be addressed as follow-ups if valid.

hjoliver:
> I'll comment below on several small points, 

So, I think my original points are still valid:

- Completion of task IDs goes down to submit number, although that usually isn't valid

> This would require the completion to introspect Cylc commands to see what kind of IDs they are expecting (i.e. workflow, cycle, task or job). Possible but tricky (and a bit of a flaw in the existing CLI ATM), for the moment it just completes the ID as far as the user takes it.

Fair enough, but I think users could reasonably expect the completions to be valid command lines even if they hit TAB one too many times, so it could be confusing. Introspection would be ideal, but I don't think it would be too kludgy in this case to simply categorize commands according to ID type in the completion server code. There are only a few types and this information will change very infrequently if at all.

- can it be made to handle latest run inference? If I type cylc trigger foo/run2// it works, but not foo//

(I explained this just above)

>  could it print a message for commands than can take arguments beyond what the completion suggests (e.g. trigger completes for existing cycle points and tasks, but the command can target any cycle point and task). Or does everything it prints have to be a potential command line?

For the few commands that can target tasks outside of n=0 can we print a warning to stderr to indicate that the provided completions are not the only valid values in this case?

hjoliver:
A final small thing: `cylc scan` shouldn't get completed with IDs.

oliver-sanders:
> If I deliberately terminate the commandline with // after the workflow name and then hit TAB, the server doesn't suggest completions based on a runN inference.

I don't think it's possible to do this.

Shell completions can only add characters, not remove them. So if you type:

```
cylc trigger foo//<TAB>
```

The completion cannot replace `foo//` with `foo/run1//` because that would involve replacing existing characters.

Also it would involve mixing workflow ID suggestions with cycle point suggestions.

datamel (commented):
Functional test failing on GH actions but I've double checked it is passing locally for me.

oliver-sanders:
> A final small thing: cylc scan shouldn't get completed with IDs.

Have added exceptions for `scan`, `cycle-point` and `message` to prevent argument completion for these commands.

hjoliver:
> The completion cannot replace foo// with foo/run1// because that would involve replacing existing characters.

That's not what I meant.

In principle the server could figure out that `runN` is implied, and list completions appropriate for the actual run, but without making it explicit:

```
cylc blah foo/run1//<TAB>  ---> foo/run1//1   
# or
cylc blah foo//<TAB> ---> foo//1  
```

oliver-sanders:
Right, sorted.

cylc/flow/scripts/completion_server.py
```diff
@@ -0,0 +1,686 @@
...
+        ['walrus', 'wellington']
+
+    """
+    if partial is None:
+        return list(words)
+    return [
+        word
+        for word in words
+        if word.startswith(partial)
+    ]
```
hjoliver, line 232: This is preventing the inferred-run completions getting through. 

For example, `partial foo//`  now correctly infers `words foo/run4//2, ...`, but they get dropped here because they don't start with `foo//` (two `/`s) 

Interestingly if change line 231 to ` if word.startswith(partial[:-1])` the first TAB results in `foo// ---> foo/run4//` which implies the partial arg is actually replaced rather than appended to by the completion. If so, I suppose we could just make the run number explicit here (if the user types `foo//<TAB>`) and take the explicit name for subsequent completion steps (in which case your new infer_run code is currently doing some unnecessary work by finding cycle points?). Hopefully that makes sense :face_with_spiral_eyes:  

:arrow_up: makes it work, and makes the run number explicit - which I think is a good thing.  i.e. `cylc blah foo//<TAB> ---> cylc blah foo/run4//`

oliver-sanders: When we infer the run number we must remove it from the results returned.

oliver-sanders:
Sorted and tested:

```console
$ cylc trigger generic/run2//20191209T
generic/run2//20191209T0900Z/  generic/run2//20191209T1200Z/
$ cylc trigger generic//20191209T
generic//20191209T0900Z/  generic//20191209T1200Z/  
```

hjoliver (approved):
Reapproving!

datamel (approved):
Re-checked this out this afternoon and double-checked code, tests and manually test-drove it. 
It is working really well for me.
One small note, cylc gui auto tab completes workflow name, which will be grand once the ui-server workflow arg pr is in. These are both features which will be released at the same time so I suspect this wouldn't be a problem.

hjoliver:
in that case ... :boom:

PR #4938, opened 2022-06-28
Validate platform settings.

### LINKED ISSUE #4922: background job runner and multiple hosts?
This is currently legal:

```ini
[platforms]
   [[bg-jobs]]
        hosts = foo, bar, baz  # OOPS
        job runner = background
```

However it doesn't make sense for the background job runner to allow multiple interchangeable hosts, because background jobs have to be polled and killed by PID on the exact host they were submitted to. (Similarly `at`, presumably).

So I guess we should either:
1. make it an error to list multiple hosts for the background job runner
2. or handle background jobs differently: allow multiple job submission hosts but always interact with jobs via the specific host they were submitted to

(And either way, document that background platforms are different in this respect).

What do you think @wxtim ?  (I don't think we're already doing 2. above?)

hjoliver:
These changes close #4922 

**Requirements check-list**


hjoliver:
(This broke some functional tests that actually use the background job runner with multiple hosts. I have a fix in the works, will push it up later; out of time for now).

cylc/flow/platforms.py
```diff
@@ -658,3 +662,30 @@ def get_localhost_install_target() -> str:
...
+
+
+def _validate(platforms) -> None:
+    """Check for invalid or inconsistent platforms config.
+
+    Some job runners require a single host (where job ID is only valid on the
+    specific submission host.)
+    """
+    bad_platforms = []
+    for name, deets in platforms.items():
```
wxtim, line 674: `deets` non obvious IMO - have used `spec` elsewhere, `config` would also be cool.

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1559,6 +1567,7 @@ def load(self) -> None:
 
         self._set_default_editors()
         self._no_platform_group_name_overlap()
+        validate_platforms()
```
oliver-sanders, line 1570: It would be better to pass the configuration object through to ensure that we are validating the correct configuration (e.g. the validation is effectively disabled for `cached = False` ATM)

cylc/flow/platforms.py
hjoliver: :+1: haha, I didn't intend to leave that there - it was an on-the-fly sub for "details"

tests/unit/test_platforms.py
```diff
@@ -171,6 +189,11 @@ def test_platform_not_there():
         platform_from_name('moooo', PLATFORMS)
 
 
+def test_invalid_platforms():
```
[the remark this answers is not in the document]
hjoliver: Done.

oliver-sanders (approved):

PR #4939, opened 2022-06-28
play: use full workflow ID when re-invoking

oliver-sanders:
* The Cylc CLI can "infer" run numbers e.g. `cylc play foo` might run `foo/run1`.
* This makes monitoring difficult because you can't see what workflow run a Cylc process is running.
* So when/if we re-invoke the `cylc play` command on a remote host we use the full ID (including the run number) not the abbreviated version.

**Requirements check-list**


tests/functional/restart/44-reinvoke.t
```diff
@@ -0,0 +1,68 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+export REQUIRE_PLATFORM='loc:remote fs:shared runner:background'
+. "$(dirname "$0")/test_header"
+set_test_number 8
```
wxtim, line 20: I make it 3: Fails Locally.

wxtim (commented):
Doesn't pass correctly on my local machine - simple issue.

Otherwise looks reasonable.

cylc/flow/scheduler_cli.py
```diff
@@ -360,24 +360,39 @@ def scheduler_cli(options: 'Values', workflow_id: str) -> None:
...
-def _distribute(host):
-    """Re-invoke this command on a different host if requested."""
+    """
     # Check whether a run host is explicitly specified, else select one.
     if not host:
         host = select_workflow_host()[0]
     if is_remote_host(host):
-        cmd = _protect_remote_cmd_args(sys.argv[1:])
+        # Protect command args from second shell interpretation
+        cmd = list(map(quote, sys.argv[1:]))
```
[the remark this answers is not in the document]
oliver-sanders: This line is tested (see the new test), however, that test doesn't run on GHA.

tests/functional/restart/44-reinvoke.t
oliver-sanders: Give me a call tomorrow, will take a look in.

datamel (approved):
Checked out and tested locally. 
No problems spotted from me. Thanks @oliver-sanders.

wxtim (approved):

PR #4942, opened 2022-06-29
reinstall: don't delete unless asked

hjoliver:
Currently `cylc reinstall` deletes files outside of the standard run sub-di.rectories (share, log, work) that do not exist in the source directory (including files files that get there via `rose-suite.conf` file creation mode, but that's OK because the `cylc-rose` plugin gets re-run at reinstall time, to recreate those).

This was reported to me as a serious bug by a NIWA user migrating to Cylc 8.

Marking as a QUESTION because, going back to the `rose suite-run` migration proposal, this is a design feature, not a bug. 

As it happens, typical practice at NIWA is to use an "install task" early in the run to extract, build, and deploy external code into the run directory (under `bin`, but potentially elsewhere too) so `cylc reinstall` nukes all of that.

This PR makes the rsync `--delete` setting an opt-in option, which I think is the right thing to do if you agree that deploying (or generating) files outside of the share directory at run time is not a fundamentally stupid thing to do. (It seems perfectly reasonable to me ... and I would have thought many MO workflows would do the same thing?)

**Requirements check-list**

### DIFF
--- a/cylc/flow/scripts/reinstall.py
+++ b/cylc/flow/scripts/reinstall.py
@@ -20,6 +20,13 @@
 
 Reinstall a previously installed workflow.
 
+WARNING: this uninstalls previously-installed files that have since been
+removed from the source directory, by detecting files outside of the standard
+run sub-directory locations (share, work, log) that are not present in the
+source directory. If your workflow tasks write to the top level run directory,
+use the --no-delete option or (preferably) configure your .cylcignore file
+appropriately, to protect these files from deletion.
+
 Examples:
   # Having previously installed:
   $ cylc install myflow
@@ -61,12 +68,18 @@ def get_option_parser() -> COP:
     )
 
     parser.add_option(
-        "--delete", "-d",
-        help="Delete previously installed files that have been removed from"
-        " the source directory.",
+        "--no-delete", "-n",
+        help=(
+            "Do not delete files outside of the standard run sub-directories"
+            " (work, share, etc.) that are not present in the source"
+            " directory. NOTE: we recommend use of a .cylcignore file in the"
+            " source directory rather than this option, because that does"
+            " not prevent removal of previously-installed files that are"
+            " no longer needed."
+        ),
         action='store_true',
         default=False,
-        dest="delete"
+        dest="no_delete"
     )
 
     parser.add_cylc_rose_options()
@@ -131,7 +144,7 @@ def main(
         source=Path(source),
         named_run=workflow_id,
         rundir=run_dir,
-        delete=opts.delete,
+        delete=not opts.no_delete,
         dry_run=False  # TODO: ready for dry run implementation
     )
 
--- a/cylc/flow/scripts/reinstall.py
+++ b/cylc/flow/scripts/reinstall.py
@@ -60,6 +60,15 @@ def get_option_parser() -> COP:
         __doc__, comms=True, argdoc=[WORKFLOW_ID_ARG_DOC]
     )
 
+    parser.add_option(
+        "--delete", "-d",
+        help="Delete previously installed files that have been removed from"
+        " the source directory.",
+        action='store_true',
+        default=False,
+        dest="delete"
+    )
+
     parser.add_cylc_rose_options()
     try:
         # If cylc-rose plugin is available
@@ -122,6 +131,7 @@ def main(
         source=Path(source),
         named_run=workflow_id,
         rundir=run_dir,
+        delete=opts.delete,
         dry_run=False  # TODO: ready for dry run implementation
     )
 
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -1426,7 +1426,7 @@ def _open_install_log(rund, logger):
     logger.addHandler(handler)
 
 
-def get_rsync_rund_cmd(src, dst, reinstall=False, dry_run=False):
+def get_rsync_rund_cmd(src, dst, delete=False, dry_run=False):
     """Create and return the rsync command used for cylc install/re-install.
 
     Args:
@@ -1434,8 +1434,8 @@ def get_rsync_rund_cmd(src, dst, reinstall=False, dry_run=False):
             file path location of source directory
         dst (str):
             file path location of destination directory
-        reinstall (bool):
-            indicate reinstall (--delete option added)
+        delete (bool):
+            add --delete option added (for reinstall)
         dry-run (bool):
             indicate dry-run, rsync will not take place but report output if a
             real run were to be executed
@@ -1447,7 +1447,7 @@ def get_rsync_rund_cmd(src, dst, reinstall=False, dry_run=False):
     rsync_cmd = ["rsync"] + DEFAULT_RSYNC_OPTS
     if dry_run:
         rsync_cmd.append("--dry-run")
-    if reinstall:
+    if delete:
         rsync_cmd.append('--delete')
     for exclude in [
         '.git',
@@ -1476,6 +1476,7 @@ def reinstall_workflow(
     source: Path,
     named_run: str,
     rundir: Path,
+    delete: bool = False,
     dry_run: bool = False
 ) -> None:
     """Reinstall workflow.
@@ -1493,7 +1494,7 @@ def reinstall_workflow(
     reinstall_log.info(f"Reinstalling \"{named_run}\", from "
                        f"\"{source}\" to \"{rundir}\"")
     rsync_cmd = get_rsync_rund_cmd(
-        source, rundir, reinstall=True, dry_run=dry_run)
+        source, rundir, delete=delete, dry_run=dry_run)
     proc = Popen(rsync_cmd, stdout=PIPE, stderr=PIPE, text=True)  # nosec
     # * command is constructed via internal interface
     stdout, stderr = proc.communicate()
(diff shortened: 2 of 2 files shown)


oliver-sanders:
I think this is a job for the .cylcignore file.

https://cylc.github.io/cylc-doc/latest/html/user-guide/installing-workflows.html#excluding-items-from-installation

I don't think making `--delete` optional is a good idea.

hjoliver:
OK, I didn't realize `.cylcignore` would handle this. It makes sense in retrospect, given that we use rsync under the hood, but we've only documented it as preventing installation of matched files in the source directory.

But in any case, we still have a problem: it breaks back-compat mode in a particularly insidious way - you won't find out until mid run when you do the first reinstall.

It's not necessarily an easy fix for users with existing Cylc 7 workflows. We give them easy access to the top level run directory, there's nothing to stop them putting whatever they want there, and that wasn't a problem (apart from tidyness!) at Cylc 7. "Relocating these items" might not be so easy because a) tasks depend their locations, and b) tasks could also be generating these items in every cycle - i.e. not just start-up install tasks.

```console
# random NIWA example:
$ ls -d1 ~ecox_oper/cylc-run/tn_nzcsm/*/
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/ancils/                      x
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/app/                         x 
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/bin/                         x
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/cf_netcdf_tools/             x
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/control/                     x
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/include/                     x
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/log/
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/log.20220405T045019Z/
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/meta/                        x
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/R_libs/                      x
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/share/                     
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/suite_utils/                 x
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/um2netcdf/                   x
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/VIS/                         x
/niwa/oper/ecox_oper/cylc-run/tn_nzcsm/work/
```

Also:

> bin/, app/, lib/, work/ or share/

`bin`, `app`, and `lib` are not currently excluded by default.

hjoliver:
Tweaked proposal:

- make "deletion mode" the default, but allow `--no-delete` as an option and turn it off in `suite.rc` back-compat mode 
- in command help for `--no-delete`, advise using `.cylcignore` instead
- document this use of `.cylcignore`
- consider excluding `app/, `bin/`, and `lib/` from deletion, by default

oliver-sanders:
We are aware of these issues. This was one of the major design decisions which went into `cylc install` made right back at the start. I think this was the correct decision and that we should stand by it. I appreciate it was a couple of years back in the "before times" when we discussed this so I'll try and dump as much of the history of this one as I can remember.

#### `rose suite-run` does not delete by default, but does have the `--new` option (which does)

The `--new` option is not well understood and the no-delete behaviour of `rose suite-run` is often reported as a bug.

The presence of a file is as much a part of the configuration as the contents of the file itself. Consider for example the commonly used Rose "optional optional" configurations. These are configuration files which are loaded automatically if present. Not deleting results in installations which neither reflect the source or (earlier) run directory but are a strange mixture of both.

#### `cylc install` uses `rsync --delete`

The deletion issue was one of the problems we set out to resolve with `cylc install`. Cylc has appropriate locations for creating files in, namely `work/` and `share/`. Our migration advice will be to relocate any "floating" files or directories into these locations. If it is difficult for some reason to do so we have provided a workaround in the form of the `.cylcignore` file.

#### `rose suite-run` used to install everything, remotely, at install time

The indiscriminate nature of `rose suite-run` means that it often installs more than necessary on remote install targets. This is half of why it could be so slow, the other half being due to the tar'ing of log files.

#### Remote installation uses `--delete` and a configured list of locations.

Why am I bringing up remote installation, because it was also performed by `rose suite-run` so requires a similar migration for Rose 2019 users. Remote installation uses `--delete`, however, it excludes by default (so less chance of deleting by accident), there is a hardcoded list of locations (which is why I got mixed up earlier with `app/`, `lib/` etc). There's [a configuration](https://cylc.github.io/cylc-doc/latest/html/reference/config/workflow.html#flow.cylc[scheduler]install) to allow additional locations to be included in remote installation.

#### Future gazing

Looking to the future we will want to make the remote installation configurable on a per-install-target basis. Depending on how deeply involved in workflow data management Cylc becomes (there is a strong desire for deeper involvement amongst our users) this could potentially be configured per-task.

#### Tweaked proposal:

> turn deletion off in suite.rc back-compat mode

The `--delete` behaviour is a `cylc install` feature and not related to the workflow configuration. Migrating to the new behaviour is a part of migrating from whatever tooling was being used before `cylc install`. Switching out the behaviour as a part of turning off compatibility mode would be confusing.

> consider allowing --no-delete as an option

No-delete is a bug-prone working practice which we do not want to support going forward. I don't this should be allowed even as an option.

> document this use of .cylcignore

The documentation is definitely lacking on this one.

> consider excluding app/, bin/, and lib/ from deletion, by default

This is dangerous and defeats the object. If you delete a file in the source directory and re-install, you expect that change to be reflected in the run dir.

#### But in any case, we still have a problem: this deletion breaks back-compat mode in a particularly insidious way

Deletion does not break and is not related to back-compat mode. The change in behaviour is due to the installation tool being used. Handling the behaviour change between `rose suite-run` and `cylc install` is a part of migrating from `rose suite-run` to `cylc install`.

#### figuring out exactly what to put in .cylcignore might not be so easy

We should try do more here, especially in the documentation.

* Trouble files will be written to paths defined by Cylc environment variables so it is likely that `s/CYLC_SUITE_RUN_DIR/CYLC_WORKFLOW_SHARE_DIR/` will handle a large proportion of cases.
* The `cylc reinstall` log is captured, this will contain a record of any deleted files. This is the quick and dirty way to get the list to dump into the `.cylcignore` file at the moment.
* We could look at opening up the `rsync --dry-run` option to extract files which would be deleted. This would allow worried users to get a list to add to their `.cylcignore` file before reinstalling for real. We could potentially do this by default but that could get slow. There is also the issue that it would show up files installed by Rose (`[file:*]` configs) but that might be something we can overcome.

hjoliver:
@oliver-sanders - OK, I don't disagree with most of your points, in particular that delete mode IS desirable for good working practices. 

Having considered this - and now that I understand `.cylcignore` is more powerful than currently documented! -  I agree we shouldn't provide `--no-delete` as an option.

However, back-compat is a different matter.

> > But in any case, we still have a problem: this deletion breaks back-compat mode in a particularly insidious way
> 
> Deletion does not break and is not related to back-compat mode. 

It does, and it is! The distinction between workflow configuration and other aspects of the system will not matter to users. The **purpose** of back compat mode is to allow Cylc 8 to run most Cylc 7 workflows out of the box. That is mostly a matter of workflow configuration, but not entirely. Back-compat mode also significantly changes scheduler behaviour (optional outputs, pre-spawned waiting tasks etc.)

NIWA workflows install code (and usually quite a lot of it, as above) from external git repositories, at run time using an in-house deployment tool. We've never used `rose suite-run` for this, because we moved to git a long time ago. So as it stands, every Cylc 7 workflow at NIWA is broken under Cylc 8 back-compat mode, and in quite a nasty way: they won't actually break until mid-run when the owner first does a reinstall, and the fix is a good deal more difficult than most of the configuration changes that back-compat mode protects them from.

So, I think we need to:
- switch off deletion in back-compat mode
- fully document use of `.cylcignore` for installed "external" (non source-dir) files
- and warn that users need to be careful about this when handle 

This will have no negative impact on MO users, because as you noted above no one used `rose suite-run --new` anyway.

OK?

> >    consider excluding app/, bin/, and lib/ from deletion, by default
> 
> This is dangerous and defeats the object. If you delete a file in the source directory and re-install, you expect that change to be reflected in the run dir.

(Agreed, I just thought in your first response above you were saying that `app`, `bin`, `lib` were already excluded; and that for installed external executables most users will likely exclude `bin` with`.cylcignore` as a matter of course anyway. But we can at least recommed that they are more specific than that).

hjoliver:
> The change in behaviour is due to the installation tool being used. Handling the behaviour change between rose suite-run and cylc install is a part of migrating from rose suite-run to cylc install.

That does not argue against back-compat support of current workflows in any way.

What matters is: 
- it was perfectly safe to deploy external files anywhere in the run directory in Cylc 7, and Cylc itself did not provide any reason not do that
- now Cylc 8's built-in install support automatically deletes those files unless users do a non-trivial "upgrade" to use `.cylcignore` appropriately 
- (and it would be easy for us to prevent this in back compat mode, so that Cylc 8 can indeed run most Cylc 7 workflows out of the box).

hjoliver:
> Having considered this - and now that I understand .cylcignore is more powerful than currently documented! - I agree we shouldn't provide --no-delete as an option.

... which does mean we can close this PR

hjoliver:
(Can be closed after reading comments above, which explain why I've superseded this with a smaller back-compat only change)

hjoliver:
(Closing as superseded!)

PR #4836, opened 2022-04-25
Log dir tidy

datamel:
These changes close https://github.com/cylc/cylc-flow/issues/4802

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.
- [x] Appropriate tests are included (unit and/or functional).
- [ ] Appropriate change log entry included.
- [ ] (master branch) I have opened a documentation PR at cylc/cylc-doc/pull/XXXX.
- [ ] (7.8.x branch) I have updated the documentation in this PR branch.
- [ ] No documentation update required.

PR #4946, opened 2022-06-29
glblcfg: increase default rolling archive length

oliver-sanders:
Related to #4836

* Log rollover happens for each restart, if workflows are repeatedly restarted (which often happens when problems occur) the sequence of events can be lost.
* Increase the archive length from 5 to 15.
* Note we have recently reduced the rollover size from 10Mb to 1Mb.

**Question:** Does 15 seem like a sensible number?

**Requirements check-list**


hjoliver:
Seems reasonable to me.

hjoliver (approved):
One review will do, we can always tweak it again later...

PR #4947, opened 2022-06-29
cli: handle UnicodeDecodeError's

oliver-sanders:
* These errors will occur if we try to write UTF-8 chars in terminals which do not support them.
* In the 2020's termials should all support UTF-8 so this is likely caused by the LANG environment variable being set to something that isn't UTF-8.
* Raise a more helpful error message which mentions the LANG variable.

(review tip add `?w=1` to the URL or turn off whitespace in the GH settings)

**Requirements check-list**


hjoliver:
> (review tip add ?w=1 to the URL or turn off whitespace in the GH settings)

(Note there's a handy UI button for that, under the settings cog drop-down)

oliver-sanders:
I guess that makes sense since in GHA it's not running in a "real" terminal?

I'm not sure how to force encoding on stdout to replicate this more generally?

oliver-sanders:
Unfortunately I don't think we can test for this in CI environment so will try to find a way to disable the test there.

oliver-sanders:
We can't test this in CI (requires a real terminal) so have disabled the test in GH actions.

tests/functional/rnd/06-unicode.t
```diff
@@ -0,0 +1,43 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Ensure that UnicodeDecodeError's are caught and handled elegantly
+
```
wxtim, line 19: The PR contains more details of the bug, without opening an unnecessary issue.
```suggestion
See: https://github.com/cylc/cylc-flow/pull/4947

```

wxtim (approved):
All good. Made one suggestion, but only to a comment.

cylc/flow/terminal.py
```diff
@@ -199,59 +199,87 @@ def wrapper(*api_args: str) -> None:
...
-                    # NOTE: sys.exit(a) is equivalent to:
-                    #       print(a, file=sys.stderr); sys.exit(1)
+            except UnicodeEncodeError as exc:
+                # this error can be raised from any code which attempts to
+                # write a UTF-8 character to a terminal which does not or is
+                # not configured to support UTF-8
+                try:
+                    # double check that this error is due to a non-UTF-8
+                    # compatible terminal and not an internal issue...
+                    print('😭')
```
MetRonnie, line 265: 😆

MetRonnie (approved):

MetRonnie (changes requested):
Ah wait, the functional test is not being skipped properly

Issue #4847, opened 2022-05-03
config: make defaults explicit


oliver-sanders:
There are a few configurations for which the default values are not properly centralised in the parsec SPEC. We should have a run through the SPEC's and ensure all defaults are defined in the spec (note they are auto-documented when defined here).

**Pull requests welcome!**

PR #4948, opened 2022-06-29
Fix lack of errors/warnings for deprecated `[runtime][<task>][remote]retrieve job logs *` settings

### LINKED ISSUE #4944: `[runtime][<task>][remote]` whole section deprecation notice?
Should there be deprecation warnings for the `[runtime][<task>][remote]retrieve job logs *` settings, like there are for e.g. `[runtime][<task>][remote]host`?

Unless they are not actually deprecated..?

**Pull requests welcome!**

MetRonnie:
These changes close #4944.

> * Setting `[runtime][<task>][remote]retrieve job logs = false` on a task specified in the old way results in a "No platform found matching your task" error (my platform definition for the matching host has log retrieval enabled) . This appears to working as intended.
> * Setting `[runtime][<task>][remote]retrieve job logs = false` on a task specified using a platform does not result in a validation error so this needs fixing.

**Requirements check-list**
- [x] Updated tests 
- [x] Change log entry
- [x] No cylc-doc PR required.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -54,6 +54,10 @@ default job runner directives for platforms.
 [#4975](https://github.com/cylc/cylc-flow/pull/4975) - Fix selection of
 platforms from `[job]` and `[remote]` configs.
 
+[#4948](https://github.com/cylc/cylc-flow/pull/4948) - Fix lack of
+errors/warnings for deprecated `[runtime][<task>][remote]retrieve job logs *`
+settings.
+
 [#4970](https://github.com/cylc/cylc-flow/pull/4970) - Fix handling of suicide
 triggers in back-compat mode.
 
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -59,15 +59,14 @@ from cylc.flow.task_events_mgr import EventData
 # Regex to check whether a string is a command
 REC_COMMAND = re.compile(r'(`|\$\()\s*(.*)\s*([`)])$')
 
-# Cylc8 Deprecation note.
 REPLACED_BY_PLATFORMS = '''
-.. warning::
+.. deprecated:: 8.0.0
 
-   .. deprecated:: 8.0.0
-
-   This will be removed in a future version of Cylc 8.
-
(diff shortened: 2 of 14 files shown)


cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1398,12 +1406,12 @@ def default_for(
...
                    environment variables``.
             ''')
-            Conf('retrieve job logs', VDR.V_BOOLEAN, desc=f'''
-                .. versionchanged:: 8.0.0
-
-                   {REPLACES}``global.rc[hosts][<host>]retrieve job logs``.
-                   {PLATFORM_REPLACES.format("[remote]retrieve job logs")}
-            ''')
+            Conf(
+                'retrieve job logs', VDR.V_BOOLEAN,
```
MetRonnie, line 1410: Why isn't there a default value for this (`True` or `False`)?

oliver-sanders: Because history I'm afraid. We should make these defaults explicit - https://github.com/cylc/cylc-flow/issues/4847

hjoliver:
(Review pending response to this: https://github.com/cylc/cylc-flow/issues/4944#issuecomment-1170633707 )

hjoliver:
From @dpmatthews explanatory comment on #4944:

> Setting `[runtime][<task>][remote]retrieve job logs = false` on a task specified using a platform does not result in a validation error so this needs fixing.

So I guess you can just fix that here @MetRonnie, and we're good to go.

MetRonnie:
https://github.com/cylc/cylc-flow/blob/19bdbf56e6efdcc8475f156febc03f0783a0f185/tests/functional/events/11-cycle-task-event-job-logs-retrieve.t#L31-L32

Not sure how to achieve this, I think a solution may be:
- Do not deprecate `flow.cylc[runtime][task][remote]retrieve job logs` in favour of platforms
- Instead, move it up a section
- This will override the platform default
- Otherwise, the `retrieve job logs max size` and `retrieve job logs retry delays` should be deprecated in favour of platforms

hjoliver (approved):
:+1:

Issue #4935, opened 2022-06-27
cylc clean: continue if remote clean fails?


hjoliver:
If `cylc clean` can't clean a remote job platform used (or attempted) by the target workflow, it should continue on and clean other job platforms, ~~and locally~~, rather than abort immediately.

I'm reasonably sure this is a bug and not a feature(?)

E.g. install and run a workflow that uses a non-existent job platform, then try to clean it.

hjoliver:
see also #4673

oliver-sanders:
It should continue onto other platforms, yes, however, it **should not** clean locally because doing so would delete the database prohibiting `cylc clean` from being able to remove the remote directory at a later date.

oliver-sanders:
(taking a quick look at the broken platform side of this)

oliver-sanders:
This has now been fixed by other changes, see https://github.com/cylc/cylc-flow/pull/4950#issuecomment-2422302745

PR #4951, opened 2022-06-30
clean: only attempt to clean platforms which have been installed to

oliver-sanders:
Addresses the "broken platform" side of #4935.

* Store the remote init map in the workflow database.
* This table is not in itself housekept because the database as a whole is housekept by `cylc clean`.
* Add a new remote-init status for platforms which have no hosts available (therefor do not need cleaning as we haven't managed to install anything onto them in the first place).

TODO:
* [ ] Don't update the status of an install_target if the new status would not require clean and the old would (e.g. can happen on restart if no hosts are available but the remote-init completed for the previous start)
* [ ] Requires (extremely simple) DB upgrader (i.e. add blank table on restart).
* [ ] Add test.

**Requirements check-list**

### DIFF
--- a/cylc/flow/rundb.py
+++ b/cylc/flow/rundb.py
@@ -188,6 +188,7 @@ class CylcWorkflowDAO:
     TABLE_TASKS_TO_HOLD = "tasks_to_hold"
     TABLE_XTRIGGERS = "xtriggers"
     TABLE_ABS_OUTPUTS = "absolute_outputs"
+    TABLE_REMOTE_INIT = "remote_init"
 
     TABLES_ATTRS = {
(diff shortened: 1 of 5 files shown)


oliver-sanders:
@hjoliver Does this approach make sense to you?

oliver-sanders:
Bumping to 8.x as I think this only really impacts us developers dealing with test battery workflows?

For real platforms users will just wait until the platform becomes available again and retry the clean command.

hjoliver:
Yes, it seems an eminently sensible approach!

oliver-sanders:
May come back to this one day, but not today...

PR #4953, opened 2022-07-01
reinstall: no delete in back compat mode

hjoliver:
Supersede #4942

Based on previous discussion, I agree that `.cylcignore` in its full glory means we should not support a no-delete option for `cylc reinstall`.

However, based on the unassailable logic of https://github.com/cylc/cylc-flow/pull/4942#issuecomment-1171747474 we DO need to turn off deletion in back-compat mode.

Summary:
- Without this we can't safely run many Cylc 7 suites (quite possibly all of them outside of the MO) in back-compat mode, without a potentially non-trivial upgrade process
- This has no negative impact impact on Cylc 7 + `rose suite-run` users, because it's only back-compat mode, and reportedly few if any were using (let alone relying on) the `--new` (deletion) option anyway (so I would argue that this is sensible back-compat support for `rose suite-run` file creation mode too - for default re-installation).
 
Screenshot from 2022-07-01 16-51-45

**Requirements check-list**

### DIFF
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -1433,10 +1433,13 @@ def get_rsync_rund_cmd(src, dst, reinstall=False, dry_run=False):
         list: command to use for rsync.
 
     """
+    check_deprecation(src)
     rsync_cmd = ["rsync"] + DEFAULT_RSYNC_OPTS
     if dry_run:
         rsync_cmd.append("--dry-run")
-    if reinstall:
+    if reinstall and not cylc.flow.flags.cylc7_back_compat:
+        # In back-compat mode don't delete external (non source-dir) files
+        # deployed to the run directory (would break many Cylc 7 suites).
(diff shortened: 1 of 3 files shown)


hjoliver:
[side-meeting]: I've been talked out of this on the basis of remote file install implications (which NIWA doesn't have due to our universal shared FS). Users would have to get to grips with reinstall deletion behaviour pretty soon anyway, so I'll bite the bullet and start evangelizing run directory tidying and use of `.cylcignore` to NIWA users, **before** they run anything with Cylc 8. Hopefully I won't get beaten up in the parking lot :rofl: 

Two of three chunks of the new functional test on this branch can be salvaged though: they test reinstall deletion of externally-installed files, or not if in `.cylcignore`.

So I'll just back out of the 2-line code change, and delete the one bad test chunk...

hjoliver:
(Actually, for transparency, I'll close this PR and cherry-pick the tests to another one...)

Issue #4915, opened 2022-06-15
Ensure that each config item has at least a one line description


MetRonnie:
In the workflow and global config reference docs

wxtim:
```python
>>> from cylc.flow.cfgspec.workflow import SPEC
>>> print('\n- '.join([repr(i[1]) for i in SPEC.walk() if not i[1].desc]))
- flow.cylc[scheduler][events]
- flow.cylc[scheduling][queues][default]
- flow.cylc[scheduling][queues][default]limit
- flow.cylc[runtime][<namespace>][job]batch system
- flow.cylc[runtime][<namespace>][job]batch submit command template
- flow.cylc[runtime][<namespace>][remote]host
- flow.cylc[runtime][<namespace>][remote]retrieve job logs
- flow.cylc[runtime][<namespace>][remote]retrieve job logs max size
- flow.cylc[runtime][<namespace>][remote]retrieve job logs retry delays
- flow.cylc[runtime][<namespace>][events]expired handlers
- flow.cylc[runtime][<namespace>][events]late offset
- flow.cylc[runtime][<namespace>][events]late handlers
- flow.cylc[runtime][<namespace>][events]submitted handlers
- flow.cylc[runtime][<namespace>][events]started handlers
- flow.cylc[runtime][<namespace>][events]succeeded handlers
- flow.cylc[runtime][<namespace>][events]failed handlers
- flow.cylc[runtime][<namespace>][events]submission failed handlers
- flow.cylc[runtime][<namespace>][events]warning handlers
- flow.cylc[runtime][<namespace>][events]critical handlers
- flow.cylc[runtime][<namespace>][events]retry handlers
- flow.cylc[runtime][<namespace>][events]submission retry handlers
- flow.cylc[runtime][<namespace>][events]execution timeout handlers
- flow.cylc[runtime][<namespace>][events]submission timeout handlers
- flow.cylc[runtime][<namespace>][events]custom handlers
- flow.cylc[runtime][<namespace>][parameter environment templates]<parameter>
```

datamel:
@wxtim Something like:

```
diff --git a/cylc/flow/cfgspec/globalcfg.py b/cylc/flow/cfgspec/globalcfg.py
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -792,7 +792,10 @@ with Conf('global.cylc', desc='''
 
             .. versionadded:: 8.0.0
         """):
[...]

PR #4954, opened 2022-07-01
Ensure that all globalconfig items have a description.

wxtim:
Partially address #4915 

**Requirements check-list**
- [x] Does not need tests (doc change - tested by ).
- [x] No change log entry required (doc change only).

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -985,8 +985,7 @@ with Conf('global.cylc', desc='''
             .. versionadded:: 8.0.0
         """):
             with Conf('<install target>', desc="""
-                :ref:`Install target <Install targets>` where directories to
-                be symlinked are located.
+                :ref:`Install target <Install targets>` on which to create the symlinks.
             """):
                 Conf('run', VDR.V_STRING, None, desc="""
                     Alternative location for the run dir.


cylc/flow/cfgspec/globalcfg.py
```diff
@@ -984,7 +984,9 @@ def default_for(
 
             .. versionadded:: 8.0.0
         """):
-            with Conf('<install target>'):
+            with Conf('<install target>', desc="""
+                Install target where directories to be symlinked are located.
```
hjoliver, line 988: Symlinking is optional, and file installation is the important thing?

```suggestion
                Where to install workflow files. A platform with a shared filesystem
                needs only one install target.
```

hjoliver: Oops, sorry, I just saw `install target` and didn't check the context!!

hjoliver: Although, we probably should refer to the platforms docs on what an install target is, or explain it here like I did. (The item name is already "install target", so the value obviously must be a particular "install target" ... in which case we should either link or explain what that means).

hjoliver:
> > These changes partially address #4915
> 
> Can this PR extend to close this issue?

From a look through the file, all items are covered now. I'll the description above to close the issue.

hjoliver:
This is a small change, and we're in a hurry (8.0) so I'm going to  commit my suggested change above (it can always be reworded later, if you disagree) and build on top of this to close the issue).

hjoliver (approved):
:+1:

PR #4961, opened 2022-07-05
Test reinstall deletion

hjoliver:
Tests salvaged from #4953 - to test that `cylc reinstall` uses rsync delete mode as expected. Probably still useful to have?

Note the small change to the unit test on this branch is not strictly required, but still makes the code a bit clearer.


datamel (approved):
Tested locally. Running smoothly for me. Thanks!

oliver-sanders (approved):

PR #4964, opened 2022-07-05
cylc reinstall --dry-run

### LINKED ISSUE #4960: reinstall: dry run
Implement the `--dry-run` option for `cylc reinstall`.

Not 8.0.0 blocking but easy to do.

**Pull requests welcome!**

oliver-sanders:
Closes #4960

Not 8.0.0 blocking but an easy win.

Add a `--dry-run` option for `cylc reinstall`.

**Requirements check-list**
- [x] Docs (I don't think changes are needed ATM, should improve the reinstall migration page soon)

### DIFF
--- a/cylc/flow/scripts/clean.py
+++ b/cylc/flow/scripts/clean.py
@@ -18,7 +18,10 @@
 
 """cylc clean [OPTIONS] ARGS
 
-Remove a stopped workflow from the local scheduler filesystem and remote hosts.
+Delete a stopped workflow.
+
+Remove workflow files from the local scheduler filesystem and any remote hosts
+the workflow was installed on.
 
 NOTE: this command is intended for workflows installed with `cylc install`. If
 this is run for a workflow that was instead written directly in ~/cylc-run and
@@ -54,7 +57,6 @@ Examples:
 
   # Only remove the workflow on remote install targets
   $ cylc clean foo/bar --remote-only
-
 """
 
 import asyncio
--- a/cylc/flow/scripts/cylc.py
+++ b/cylc/flow/scripts/cylc.py
@@ -429,6 +429,22 @@ def cli_help():
     # modern terminal)
     from colorama import init as color_init
     color_init(autoreset=True, strip=False)
+    commands = [
+        'clean',
+        'hold',
+        'install',
+        'kill',
+        'pause',
+        'play',
+        'release',
+        'scan',
+        'stop',
+        'trigger',
+        'tui',
+        'validate',
+    ]
+    if 'gui' in COMMANDS:
+        commands.append('gui')
     print(USAGE)
     sys.exit(0)
 
--- a/cylc/flow/scripts/cylc.py
+++ b/cylc/flow/scripts/cylc.py
@@ -26,7 +26,10 @@ from ansimarkup import parse as cparse
 import pkg_resources
 
 from cylc.flow import __version__, iter_entry_points
-from cylc.flow.option_parsers import format_shell_examples
+from cylc.flow.option_parsers import (
+    format_help_headings,
+    format_shell_examples,
+)
 from cylc.flow.scripts.common import cylc_header
 
 
@@ -173,8 +176,7 @@ Filters
 
 # because this command is not served from behind cli_function like the
 # other cylc commands we have to manually patch in colour support
-USAGE = format_shell_examples(USAGE)
-USAGE = cparse(USAGE)
+USAGE = cparse(format_help_headings(format_shell_examples(USAGE)))
 
 # all sub-commands
 # {name: entry_point}
@@ -429,22 +431,6 @@ def cli_help():
     # modern terminal)
     from colorama import init as color_init
     color_init(autoreset=True, strip=False)
-    commands = [
-        'clean',
-        'hold',
-        'install',
-        'kill',
-        'pause',
-        'play',
-        'release',
-        'scan',
-        'stop',
-        'trigger',
-        'tui',
-        'validate',
-    ]
-    if 'gui' in COMMANDS:
-        commands.append('gui')
     print(USAGE)
     sys.exit(0)
 
--- a/cylc/flow/scripts/reinstall.py
+++ b/cylc/flow/scripts/reinstall.py
@@ -30,32 +30,74 @@ Examples:
   # Or, to reinstall a specific run:
   $ cylc reinstall myflow/run2
 
-  # View the changes reinstall would make:
-  $ cylc reinstall myflow --dry-run
+  # If the workflow is running:
+  $ cylc reinstall myflow  # reinstall as usual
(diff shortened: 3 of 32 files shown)


oliver-sanders:
(currently in draft whilst I add the interactive mode stuff...)

cylc/flow/scripts/clean.py
```diff
@@ -18,7 +18,10 @@
 
 """cylc clean [OPTIONS] ARGS
 
-Remove a stopped workflow from the local scheduler filesystem and remote hosts.
+Deleted a stopped workflow.
+
+Remove workflows files from the local scheduler filesystem and and remote hosts
+the workflow was installed on.
```
hjoliver, line 24:
```suggestion
Delete a stopped workflow.

Remove workflow files from the local scheduler filesystem and remote hosts
the workflow was installed on.
```

hjoliver (commented):
Quick premature review. Thanks @oliver-sanders (and @datamel for laying the groundwork) - this will ease the pain a little!

Can we explicitly warn, in `cylc reinstall` command help, that it will delete any non-source-dir files installed or generated at run time, that aren't in `share` etc. or protected by `.cylcignore`.

oliver-sanders:
> Can we explicitly warn, in cylc reinstall command help, that it will delete any non-source-dir files installed or generated at run time, that aren't in share etc. or protected by .cylcignore.

I was just quickly adding the `--dry-run` option, but the interactive mode stuff should be quick to do so I'm currently working on that. Will make the reinstall command document reinstall behaviour in the process.

oliver-sanders:
@hjoliver Have added interactive mode and some, hopefully helpful messages. Need to straighten out the tests with this change. WDYT?

cylc/flow/scripts/clean.py
```diff
@@ -18,7 +18,10 @@
 
 """cylc clean [OPTIONS] ARGS
 
-Remove a stopped workflow from the local scheduler filesystem and remote hosts.
+Delete a stopped workflow.
+
+Remove workflow files from the local scheduler filesystem and and remote hosts
```
hjoliver, line 23:
```suggestion
Remove workflow files from the local scheduler filesystem and remote hosts
```

hjoliver (approved):
(Premature review: all good so far :+1: )

hjoliver:
> WDYT?

Excellent, thanks.

hjoliver:
Maybe don't prompt if there are no changes at all.

cylc/flow/scripts/cylc.py
```diff
@@ -430,24 +433,28 @@ def cli_help():
     # modern terminal)
     from colorama import init as color_init
     color_init(autoreset=True, strip=False)
+    commands = [
+        'clean',
+        'hold',
+        'install',
```
datamel, line 439: Is reinstall important enough to be added here?

datamel (commented):
I've had a really good play around with this. One thing I have noticed is that in "--debug" mode, the rsync command is printed to the terminal but not the log. I think (if it is desirable for the command to be printed to the log, which I think it is) the log level of the log needs dropping to debug, when --debug in cli opts, and the command is currently written to `LOG`, which would need to be changed to `reinstall_log`? It looks a bit like a conscious choice so happy to be overruled if you think otherwise, I am possibly not thinking of a reason for this. 
Really nice cli interface, with the tip.

cylc/flow/scripts/cylc.py
oliver-sanders: Fair point, subjective, I didn't add `reload` which was silly, if we are having `reinstall` we need `reload` too.

oliver-sanders:
@datamel Were you suggesting logging the `rsync` command to the reinstall log or to the scheduler log?

(The LOG.debug I added doesn't go to either ATM, only to stderr)

datamel:
> @datamel Were you suggesting logging the `rsync` command to the reinstall log or to the scheduler log?
> 
> (The LOG.debug I added doesn't go to either ATM, only to stderr)

I was suggesting the reinstall log. Just somewhere for us to be able to check the command in the case of debugging problems.

cylc/flow/scripts/cylc.py
hjoliver: (That change has been merged)

cylc/flow/workflow_files.py
```diff
@@ -1480,29 +1496,52 @@ def reinstall_workflow(
...
+        f'Reinstalling "{named_run}", from "{source}" to "{rundir}"'
+    )
     rsync_cmd = get_rsync_rund_cmd(
-        source, rundir, reinstall=True, dry_run=dry_run)
+        source,
+        rundir,
+        reinstall=True,
+        dry_run=dry_run,
+    )
+    reinstall_log.info(cli_format(rsync_cmd))
```
oliver-sanders, line 1524: > I was suggesting the reinstall log. Just somewhere for us to be able to check the command in the case of debugging problems.

Now done here.

hjoliver:
(Pushed a style fix to allow tests to passed, because I thought @datamel had approved this already ... but it seems not, so not merging just yet).

cylc/flow/scripts/clean.py
```diff
@@ -18,7 +18,10 @@
 
 """cylc clean [OPTIONS] ARGS
 
-Remove a stopped workflow from the local scheduler filesystem and remote hosts.
+Delete a stopped workflow.
+
+Remove workflow files from the local scheduler filesystem and any remote hosts
```
datamel, line 23: Not sure on this one, maybe filesystem is one word, just flagging in case it needs a space.

```suggestion
Remove workflow files from the local scheduler file system and any remote hosts
```

datamel (commented):
All good. Just a small change I think to make on the help docs.

cylc/flow/scripts/reinstall.py
```diff
@@ -23,37 +23,82 @@
...
+  $ cylc reinstall myflow  # reinstall as usual
+  $ cylc reload myflow     # pick up changes in the workflow config
+
+What reinstall does:
+  Reinstall synchronises files between the workflow source and the specified
+  run directory.
+
+  Any files which have been added, updated or removed in the source directory
+  will be added, updated or removed in the run directory. Cylc uses "rsync"
+  to do this (run in "--debug" mode to see the exact command used).
```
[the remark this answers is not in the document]
oliver-sanders: The command is printed to stdout in `--debug` mode which is useful to a user using `--dry-run` (why I added it). The reinstall log doesn't get created in `--dry-run` mode so isn't of much help.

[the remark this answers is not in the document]
datamel: Ah ok. It is also printed to the reinstall log on a successful reinstall, I guess that does not need saying.

datamel (approved):

PR #4967, opened 2022-07-06
Back-compat mode tweak.

### LINKED ISSUE #4965: `:finish` trigger difference in stall behaviour between c7 & c8 back-compat mode
**Describe the bug / steps to reproduce**

```ini
# suite.rc
[scheduling]
    initial cycle point = 2002
    [[dependencies]]
        [[[P1Y]]]
            graph = foo:finish => cleanup
[runtime]
    [[foo]]
        script = false
    [[cleanup]]
```
Running with Cylc 7, this keeps going indefinitely.

Running with Cylc 8 (in back compat mode), this stalls:
```log
WARNING - Incomplete tasks:
      * 20020101T0000Z/foo did not complete required outputs: ['succeeded']
      * 20030101T0000Z/foo did not complete required outputs: ['succeeded']
      * 20040101T0000Z/foo did not complete required outputs: ['succeeded']
      * 20050101T0000Z/foo did not complete required outputs: ['succeeded']
      * 20060101T0000Z/foo did not complete required outputs: ['succeeded']
```

**Expected behavior**

Stall behaviour should be the same in Cylc 7 and Cylc 8 back-compat mode?

**Additional context**

Running with Cylc 8 after renaming to `flow.cylc`, it continues indefinitely like it did with Cylc 7.

hjoliver:
Ignore incomplete tasks in computing the runahead limit, in back compat mode.
(Because: Cylc 7 ignores failed tasks in computing the limit).

These changes close #4965

From comments in the new func test workflow:
```
# When a task with a finish trigger fails:
# Cylc 7:
#   No runahead stall (failed tasks are ignored in computing runahead) but
#   stall when nothing else to do (e.g. final cycle point) due to failed tasks
#   in the pool.
# Cylc 8 back-compat mode:
#   Replicate Cylc 7 behaviour by making success outputs required and ignoring
#   incomplete tasks in runahead computation.
# Cylc 8 non-back-compat mode:
#   No runahead stall and no stall at final cycle point, because finish
#   triggers imply success is optional (i.e. no incomplete tasks created). 
```

**Requirements check-list**

### DIFF
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -288,7 +288,14 @@ class TaskPool:
                         TASK_STATUS_SUCCEEDED,
                         TASK_STATUS_EXPIRED
                     )
-                    or itask.state.outputs.is_incomplete()
+                    or (
+                        # For Cylc 7 back-compat, ignore incomplete tasks.
+                        # (All :succeed is required in back-compat mode, so
+                        # failed tasks end up incomplete; and Cylc 7 ignores
+                        # failed tasks in computing the runhead limit).
+                        itask.state.outputs.is_incomplete()
+                        and not cylc.flow.flags.cylc7_back_compat
+                    )
                     for itask in itasks
                 )
             ):
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -290,9 +290,9 @@ class TaskPool:
                     )
                     or (
                         # For Cylc 7 back-compat, ignore incomplete tasks.
-                        # (All :succeed is required in back-compat mode, so
-                        # failed tasks end up incomplete; and Cylc 7 ignores
-                        # failed tasks in computing the runhead limit).
+                        # (Success is required in back-compat mode, so failed
+                        # tasks end up as incomplete; and Cylc 7 ignores
+                        # failed tasks when computing the runahead limit).
                         itask.state.outputs.is_incomplete()
                         and not cylc.flow.flags.cylc7_back_compat
                     )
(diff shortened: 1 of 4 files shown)


hjoliver:
NB, if you add a final cycle point to the example in #4965 it stalls at the fcp due to the failed/incomplete tasks, in both C7 and C8 back compat mode.

hjoliver:
Note CI passed. Final trivial commit has skip-ci.

cylc/flow/task_pool.py
```diff
@@ -288,7 +288,14 @@ def compute_runahead(self):
                         TASK_STATUS_SUCCEEDED,
```
dpmatthews, line 288: Is this test still relevant (other than in back-compat mode)?
Any task that has failed or succeeded will only be in the pool if it is incomplete?
Expired tasks immediately get removed from the pool?

cylc/flow/task_pool.py
```diff
@@ -288,7 +288,14 @@ def compute_runahead(self):
...
                         TASK_STATUS_EXPIRED
                     )
```
dpmatthews, line 290: Is this test still relevant (other than in back-compat mode)?
Any task that has failed or succeeded will only be in the pool if it is incomplete?
Expired tasks immediately get removed from the pool?

hjoliver: I think you're right, but it's off-topic for this change, and now that you've raised it I want to re-check that expired tasks work as expected.  #4971

oliver-sanders (approved):

PR #4975, opened 2022-07-08
Fix platform selection bug.

### LINKED ISSUE #4979: Localhost selected from old settings which ought to return a non-local platform
**Describe the bug**
The following config

```ini
# Part of suite.rc[runtime][<namespace>]
    [[remote]]
      host = my_hpcA
    [[job]]
      batch system = pbs
```
Is expected to select platform `my_hpc` - but instead selects localhost.

On inspection of the platform code it looks like `OrderedDictWithDefaults` does not over-ride the `get` method.

as a result the attempt to use

```python
            task_job_section = task_conf.get("job", {})
            task_remote_section = task_conf.get("remote", {})
```
To retrieve the task job and task remote sections to be fed to the platform guessing function is returning `{}`.

**Pull requests welcome!**

wxtim:
`OrderedDictWithDefaults.__getitem__` will get the default, if it exists.
`OrderedDictWithDefaults.get` does not get the default and was causing
the platform selection to fail quietly and select localhost.

This change closes #4979 

**Requirements check-list**

### DIFF
--- a/tests/unit/test_platforms.py
+++ b/tests/unit/test_platforms.py
@@ -609,6 +609,7 @@ def test_get_platform_from_OrderedDictWithDefaults(mock_glbl_cfg):
     Most tests use dictionaries to check platforms functionality.
     This one was added to catch an issue where the behaviour of
     dict.get != OrderedDictWithDefaults.get.
+    See - https://github.com/cylc/cylc-flow/issues/4979
     """
     mock_glbl_cfg(
         'cylc.flow.platforms.glbl_cfg',


tests/unit/test_platforms.py
```diff
@@ -34,6 +35,8 @@
     PlatformLookupError,
     GlobalConfigError
 )
+from cylc.flow.parsec.OrderedDict import OrderedDictWithDefaults
```
oliver-sanders, line 38: Duplicate import?

tests/unit/test_platforms.py
```diff
@@ -599,3 +602,33 @@ def test_is_platform_definition_subshell(
...
     else:
         assert is_platform_definition_subshell(plat_val) is expected
+
+
+def test_get_platform_from_OrderedDictWithDefaults(mock_glbl_cfg):
+    """Get platform works with OrderedDictWithDefaults.
+
+    Most tests use dictionaries to check platforms functionality.
+    This one was added to catch an issue where the behaviour of
+    dict.get != OrderedDictWithDefaults.get.
```
oliver-sanders, line 612: Good idea to link to the issue in the test docstring.

wxtim: > This is a small change with no associated Issue.

Do I need to create an issue?

wxtim:
```suggestion
    dict.get != OrderedDictWithDefaults.get.
    See - https://github.com/cylc/cylc-flow/issues/4979
```

oliver-sanders (approved):

datamel (approved):
I have checked this out and verified that it performs as expected. Checked tests also pass locally. Thanks @wxtim.

MetRonnie:
Why was `OrdererdDictWithDefaults.get()` not fixed?

MetRonnie:
Answer: because we don't want to mess up anything that already relies on the current behaviour.

Note that `.get()` still works after you've got past the "`defaults_` barrier", i.e.

```python-console
>>> a = OrderedDictWithDefaults()
>>> a.defaults_ =  OrderedDictWithDefaults([('job', OrderedDictWithDefaults([('batch sys', 'pbs')]))])
>>> a.get('job', {})
{}
```
doesn't work, but this does:
```python-console
>>> a['job'].get('batch sys', None)
'pbs'
```
because at that point you're accessing the primary dict of an OrderedDictWithDefaults as opposed to its defaults

PR #4978, opened 2022-07-11
Fix failure to clean on remote host with leftover contact file

MetRonnie:

From Teams:

>Cylc clean bug report... it looks like it is trying to perform the whole "is this workflow running" check on the remote side as well as the invocation side which isn't right:
>```
>$ cylc clean u-co906/run1
>INFO - Cleaning u-co906/run1 on install target: ab
>ERROR - Could not clean u-co906/run1 on install target: ab
>    platform: abc - clean up did not complete
>    COMMAND:
>        ssh CYLC_VERSION=8.0rc4.dev CYLC_ENV_NAME=cylc-8.0rc4.dev-3 \
>            bash --login -c 'exec "$0" "$@"' timeout 120 cylc clean \
>            --local-only u-co906/run1
>    RETURN CODE:
>        1
>    STDOUT:
>    STDERR:
>        CylcError: Cannot determine whether workflow is running on <scheduler-host>.
>        ~fcm/cylc-8.0rc4.dev-3/bin/python3.9 ~fcm/cylc-8.0rc4.dev-3/bin/cylc play u-co906 --host=localhost
>
>CylcError: Remote clean failed for u-co906/run1
>```
>Note: We cannot necessarily SSH back from remote platforms to the scheduler host.

**Solution**

~If there us a leftover contact file when doing `cylc clean`, if the host field gives a host that is not the local host, it means we are doing re-invoked `cylc clean --local-only` on the remote host and we should not check if the scheduler process is still running (it can't be).~

The workflow DB only exists on the scheduler host. So if we are doing `cylc clean --local-only` and there is a leftover contact file but no DB, it probably means this is `cylc clean` being re-invoked on a remote host and we can ignore the contact.

Even if the fix gets hit when someone genuinely someone runs `cylc clean --local-only` on the scheduler host, it shouldn't matter.

**Requirements check-list**

### DIFF
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -499,7 +499,7 @@ def detect_old_contact_file(reg: str, contact_data=None) -> None:
     # NOTE: can raise CylcError
     process_is_running = _is_process_running(old_host, old_pid, old_cmd)
 
-    fname = get_contact_file(reg)
+    fname = get_contact_file_path(reg)
     if process_is_running:
         # ... the process is running, raise an exception
         raise ServiceFileError(
@@ -541,7 +541,7 @@ def dump_contact_file(reg, data):
     # The double fsync logic ensures that if the contact file is written to
     # a shared file system e.g. via NFS, it will be immediately visible
     # from by a process on other hosts after the current process returns.
-    with open(get_contact_file(reg), "wb") as handle:
+    with open(get_contact_file_path(reg), "wb") as handle:
         for key, value in sorted(data.items()):
             handle.write(("%s=%s\n" % (key, value)).encode())
         os.fsync(handle.fileno())
@@ -550,7 +550,7 @@ def dump_contact_file(reg, data):
     os.close(dir_fileno)
 
 
-def get_contact_file(reg):
+def get_contact_file_path(reg: str) -> str:
     """Return name of contact file."""
     return os.path.join(
         get_workflow_srv_dir(reg), WorkflowFiles.Service.CONTACT)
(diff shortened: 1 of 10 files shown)


oliver-sanders (approved):

wxtim (approved):
Happy with this - add a changelog entry before merging.

MetRonnie:
Changed approach, relying on the fact that the workflow DB only exists on the scheduler host.

wxtim (approved):

cylc/flow/workflow_files.py
```diff
@@ -751,6 +751,16 @@ def _clean_check(opts: 'Values', reg: str, run_dir: Path) -> None:
         raise ServiceFileError(
             f"Cannot clean running workflow {reg}.\n\n{exc}"
         )
+    except CylcError:
+        db_path = (
+            run_dir / WorkflowFiles.Service.DIRNAME / WorkflowFiles.Service.DB
+        )
```
oliver-sanders, line 757: I think this is catching the SSH error, I think we should really prevent it from happening.

```python
def _clean_check(opts: 'Values', reg: str, run_dir: Path) -> None:
    validate_workflow_name(reg)
    reg = os.path.normpath(reg)
    if get_contact_path(reg).exists() and not get_db_path(reg).exists():
       ...
```

MetRonnie:
Re-done as per Oliver's suggestion

hjoliver (approved):
(A very quick review but LGTM; I haven't tested it).

wxtim (approved):
I'm OK with this. @oliver-sanders  - do you want to do a final check?

oliver-sanders:
LGTM

Issue #4974, opened 2022-07-08
stuck in preparing state


oliver-sanders:
Tasks can get "stuck" in the preparing state, seen in the wild, twice, but struggling to reproduce.

Possibly related to reloads, possibly caused by a mismatch of submission number (we increment sub num on submission so a subsequent preparing job will have the same sub num as a previous instance), dunno.

Order of events:
* submission 01 preparing
* submission 01 submit-failed
* reload  # no change
* submission 02 preparing
* unhandled jobs-submit output ... same-task/02 ...
* [jobs-submit cmd] ... same-task/02
  [jobs-submit ret_code] 1
  [jobs-submit out] ...|cycle/task/01|1
* submission 01 (polled-ignored) submission failed  # i.e. previous job showed up but was ignored

At the end of this, job 02 is still marked as preparing, however, is no longer preparing anything, submission seemingly having failed long ago just after the unhandled jobs-submit output. The task is not killable (because the preparing state is not killable).

oliver-sanders:
It looks like the cause is a mix up of submit numbers.

When we perform reload we make a copy of the task:

https://github.com/cylc/cylc-flow/blob/8ff0c804980efd406069db9d26d4e155bacf6aca/cylc/flow/task_pool.py#L845-L848

The copy is then triggered incrementing its submit number. If something out there somewhere is maintaining a reference to the original task, however, ...

oliver-sanders:
The issue was resolved via a restart which suggests that the issue was due to scheduler state not mirrored in the DB so the `pre_prep_tasks` theory is likely, any easy fix if so...

oliver-sanders:
Haven't managed to reproduce, however, this test shows how reload can interfere with the `pre_prep_tasks` pipeline:

```python
async def test_pre_prep_reload(one, start):
    """Ensure reload does not interfere with the task preparation pipeline."""                         
    async with start(one):
        one.resume_workflow()
        one.pool.release_runahead_tasks()
        one.release_queued_tasks()
        assert len(one.pre_prep_tasks) == 1
        one.command_reload_workflow()           
        assert one.pool.get_tasks()[0] == one.pre_prep_tasks   
```

oliver-sanders:
[...]

PR #4984, opened 2022-07-15
reload: fix submission errors for jobs awaiting preparation

oliver-sanders:
Closes https://github.com/cylc/cylc-flow/issues/4974    

When we reload a workflow we create new TaskProxy instances for all non-active tasks in the pool and replaces the previous instances with these new ones. The job submission pipeline maintains references to the pre-reload TaskProxy's which means if the workflow is reloaded whilst a tasks is going through job prep the pre-reload task gets submitted but the post-reload task is left in the pool. This results in submission issues as the submit number between the two instances are one off.

For busy workflows where the job submission pipeline is constantly churning away this bug is almost guaranteed to strike with any reload operation.

This PR attempts to resolve this issue by:
1. Ensuring job submission is fed the same TaskProxy instance as present in the task pool.
2. Incrementing the job submission number at preparation time (rather than after job submit).

Explanation:

1. scheduler: re-compute pre_prep_tasks for each iteration    
       
   * Addresses https://github.com/cylc/cylc-flow/issues/4974    
   * Tasks which are awaiting job preparation used to be stored in    
     `Scheduler.pre_prep_tasks`, however, this effectively created an    
     intermediate "task pool" which had nasty interactions with reload.    
   * This commit removes the pre_prep_tasks list by merging the listing    
     of these tasks in with TaskPool.release_queued_tasks (to avoid    
     unnecessary task pool iteration).    
   * `waiting_on_job_prep` now defaults to `False` rather than `True`.    
       
2. job: increment the submission number at preparation time                           
          
   * Addresses https://github.com/cylc/cylc-flow/issues/4974      
   * Job submission number used to be incremented *after* submission      
     (i.e. only once there is a "submission" of which to speak).      
   * However, we also incremented the submission number if submission      
     (or preparation) failed (in which cases there isn't really a      
     "submission" but we need one for internal purposes).      
   * Now the submission number is incremented when tasks enter the      
     "preparing" state.      
   * This resolves an issue where jobs which were going through the      
     submission pipeline during a reload got badly broken in the scheduler      
     (until restarted).   
   * It also makes the scheduler logs nicer as the submission number for
     preparing tasks now matches that of the subsequent submitted or
     submit-failed outcome.

Testing this one is a nightmare, I can't really see a way to do it meaningfully. Look at the issue for example tests which demonstrates how tasks could get 

**Requirements check-list**
- [ ] Already covered by existing tests, cannot really test this


oliver-sanders:
This leaves one remaining question of reload safety:

https://github.com/cylc/cylc-flow/blob/d4938dd40320b423241b1bd3feaa693da8d043ac/cylc/flow/task_pool.py#L858-L861

Should preparing tasks be included here? I think there is the potential for preparation to begin with pre-reload settings but possibly to receive post-reload settings later on?

hjoliver:
> It also makes the scheduler logs nicer as the submission number for
preparing tasks now matches that of the subsequent submitted or
submit-failed outcome.

:+1: must admit this annoyed me!

hjoliver:
Your explanation sounds reasonable, but the functional tests aren't happy.

> This leaves one remaining question of reload safety:...

While thinking about how to test this, I discovered this bug: #4987

hjoliver:
(Agreed good to have this fix in 8.0 if we can nail it in time).

oliver-sanders:
> the functional tests aren't happy.

Couple of small breaks, will investigate, expect I can shift them quickly.

> (Agreed good to have this fix in 8.0 if we can nail it in time).

Unfortunately I think this is 8.0 essential as at the moment `cylc reload` is almost guaranteed to break workflows every time.

oliver-sanders:
I had got my if/elif branches the wrong way around, [fixed](https://github.com/cylc/cylc-flow/compare/9adab1f761d9c568dfe2179e3de1c1fadb60759f..56f9bf1fd954f241637ba456e684b5e03b07418d), think that should hold it.

hjoliver:
> This leaves one remaining question of reload safety:
> 
> https://github.com/cylc/cylc-flow/blob/d4938dd40320b423241b1bd3feaa693da8d043ac/cylc/flow/task_pool.py#L858-L861
> 
> Should preparing tasks be included here? I think there is the potential for preparation to begin with pre-reload settings but possibly to receive post-reload settings later on?
 
I think we include `preparing` here because pre-reload settings are fixed once written to the job file. Presumably it's possible to reload a preparing task before it writes the job file, though.  Maybe we should have a separate warning for `preparing` tasks: *may be active with pre-reload settings*.

hjoliver (approved):
Makes sense to me, and the tests are happy now.

hjoliver:
(The failed functional test is just codecov upload)

dwsutherland (approved):
Works, tests passed, code makes sense.. 
From what I can see, the order of events in the main loop is:
- **Do reload if flagged** (reloads task pool at this point (along with DB and store etc))
- **Process the command queue:**
 -- reload command sets up new config, then reconfigures broadcasts, pool (just config), tasks events, and DB... Which flags reload.
- Releases/computes runahead (**uses new pre-reload pool with post reload config**)
- Proc Pool Processed
- Checks triggers (xtrigger etc), and sets expired tasks
- **Releases queued tasks** (`self.release_queued_tasks()` changed with this PR) (still pre-reload pool)
- . . .
- Processes queued task messages 
- **Processes the command queue (Again)**
- Processes task events
- DB, data-store, health checks ...etc

END LOOP  

Probably a bigger issue to look into the ideal order of events, however there's (intentionally or otherwise) a number of things that happen between processing a reload command (reconfiguring) and reloading the task pool.. 

Would it be better to do the reload straight after processing the reload command (setting up the new config)? 
Because at present tasks are potentially released/prepped between setting up a new config and reloading the task pool. Perhaps the first two steps can be switched? (or does it not matter?)

oliver-sanders:
> Probably a bigger issue to look into the ideal order of events

Definitely, this code is fragile as the order will be critical to particular niche behaviours so would need some time to unravel :(

wxtim (approved):
Looks legit and much nicer than my original solution 👍🏼

PR #4989, opened 2022-07-19
Checks tutorials for consistency with cylc docs

wxtim:
...and added validation 

These changes partially address https://github.com/cylc/cylc-doc/issues/470, by adding tests to the following tutorial workflows:
- message triggers
- retries

**Requirements check-list**
- [x] Does not need tests (Change adds testing without adding functionality).


oliver-sanders:
I don't think copying the tutorial content here helps with QA as the two copies are free to drift apart.

E.G. if there is a config change and this test breaks the author will likely update the cylc-flow files to make the test pass (likely `sed -i` style) leaving the cylc-doc files broken.

wxtim:
> I don't think copying the tutorial content here helps with QA as the two copies are free to drift apart.
> 
> E.G. if there is a config change and this test breaks the author will likely update the cylc-flow files to make the test pass (likely `sed -i` style) leaving the cylc-doc files broken.

Nevermind then.

PR #4993, opened 2022-07-21
Remove useless text editor config.

hjoliver:
This is a small change with no associated Issue: 
- remove "your favourite text editor" config items, and associated code.

The primary reasons for having a configurable editor in Cylc were the `cylc trigger --edit` and `cylc edit` commands, neither of which exist any more. (Although the config reference docs still say it is to configure "your favourite editor for editing workflow definitions").

The only remaining use of the configured editor is:
 - `cylc cat-log --mode=edit`
 - `cylc view`  (which weirdly defaults to edit instead of stdout)

These are essentially pointless - if you want to view output in your editor, direct the stdout to a file first, or to your editor's stdin.

NB: the config change proposal notes that we agreed in the workshop to keep the editor config but discourage its use. Maybe we had not decided the fate of `cylc edit` at that point though.

**Requirements check-list**
- [x] Does not need tests (code removal only).


hjoliver:
(Not a blocker for 8.0 obviously, but it removes some crud and should be a trivial review-and-merge)

oliver-sanders:
> or to your editor's stdin

Valid advice providing you `s/emacs/vim`.

hjoliver:
Tests all passing now. One review will do.

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1050,55 +1050,6 @@ def default_for(
 
                     .. versionadded:: 8.0.0
                 """)
-
-    with Conf('editors', desc='''
```
wxtim, line 1054: Do you want to provide obseletion notices for these in the upgrader?

cylc/flow/scripts/cat_log.py
```diff
@@ -264,11 +248,6 @@ def get_option_parser() -> COP:
         help="Job submit number (default=%s, i.e. latest)." % NN,
         metavar="INT", action="store", dest="submit_num", default=NN)
 
-    parser.add_option(
```
wxtim, line 267: Not worth keeping and plugging in to a deprecation warning?

cylc/flow/cfgspec/globalcfg.py
hjoliver: Good point, I suppose I should!

cylc/flow/scripts/cat_log.py
hjoliver: Unlike the config, I don't think that's needed. We've changed a lot of commands and command options without deprecations. It's pretty easy to change your command line on the fly, if you invoke an obsolete option.

wxtim (changes requested):
- [x] Mostly happy
- [x] Played with locally

Asking for addition of obsoletion messages so that anyone with an older config gets a useful message.

cylc/flow/cfgspec/globalcfg.py
hjoliver: Done.

wxtim (approved):
Happy with that.

Perhaps another day we could implement an additional upgrader method, something like `quiet`, which works the same as obsolete, but only logs at debug?

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1,6 +1,4 @@
-# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
-# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
-#
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.  # Copyright (C) NIWA & British Crown (Met Office) & Contributors.  #
```
oliver-sanders, line 1:
```suggestion
# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
#
```

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1050,55 +1048,22 @@ def default_for(
...
-
-            Examples::
-
-               atom --wait
-               code --new-window --wait
-               emacs
-               gedit -s
-               gvim -fg
-               nedit
+            Graphical text editor (deprecated).
```
oliver-sanders, line 1065: I thought this was getting removed & obsoleted?

cylc/flow/cfgspec/globalcfg.py
hjoliver: Oops, weird keyboard fail.

cylc/flow/cfgspec/globalcfg.py
hjoliver: It was, then @wxtim suggested this:
> Asking for addition of obsoletion messages so that anyone with an older config gets a useful message.

... and we both temporarily forgot that we're not doing that for the global config :rofl:

cylc/flow/cfgspec/globalcfg.py
hjoliver: fixed

hjoliver:
(Spurious test failure - artifact upload timed out)

oliver-sanders (approved):

oliver-sanders:
(not worried about the changelog)

PR #4996, opened 2022-07-21
fix flake8 bug which crept into master

wxtim:

A small flake8 failure appears to have crept into master.

- [x] Tiny style fix - skipping normal checklist.


MetRonnie:
Will review after the tests pass

MetRonnie:
Functional tests timed out but should be ok

PR #4998, opened 2022-07-21
scan: update workflow_params to use public db

oliver-sanders:
Spotted by Dave whilst reviewing DB use:

* The workflow_params function was using the private DB to list workflow params.
* Really it should be using the public DB as only the scheduler process/thread should be accessing the private DB to avoid the potential for DB locking.

Note this functionality was prepared ahead of use in the UIS (https://github.com/cylc/cylc-uiserver/issues/371) but this work has been pushed back so it's currently unused.

So this code is a bug "in waiting", if this interface were to be called it would cause DB locking, similar to the `rose_prune` issue.

**Requirements check-list**
- [ ] [if bugfix] PRs raised to both master and the relevant bugfix branch.

### DIFF
--- a/cylc/flow/network/scan.py
+++ b/cylc/flow/network/scan.py
@@ -67,7 +67,10 @@ from cylc.flow.network.client import (
     ClientTimeout,
     WorkflowRuntimeClient,
 )
-from cylc.flow.pathutil import get_cylc_run_dir
+from cylc.flow.pathutil import (
+    get_cylc_run_dir,
+    get_workflow_run_dir,
+)
 from cylc.flow.rundb import CylcWorkflowDAO
 from cylc.flow.workflow_files import (
     ContactFileFields,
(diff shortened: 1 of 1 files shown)


datamel (approved):

cylc/flow/network/scan.py
```diff
@@ -528,7 +531,9 @@ def _callback(_, entry):
         key, value = entry
         params[key] = value
 
-    db_file = flow['path'] / SERVICE / 'db'
+    # NOTE: use the public DB for reading
+    # (only the scheduler process/thread should access the private database)
+    db_file = Path(get_workflow_run_dir(flow['name'], 'log', 'db'))
     if db_file.exists():
         dao = CylcWorkflowDAO(db_file, is_public=False)
         dao.connect()
```
MetRonnie, line 539: I would stick a `dao.close()` at the end for safety
```diff
diff --git a/cylc/flow/network/scan.py b/cylc/flow/network/scan.py
index 717f93372..201ee4f14 100644
--- a/cylc/flow/network/scan.py
+++ b/cylc/flow/network/scan.py
@@ -536,8 +536,11 @@ async def workflow_params(flow):
     db_file = Path(get_workflow_run_dir(flow['name'], 'log', 'db'))
     if db_file.exists():
         dao = CylcWorkflowDAO(db_file, is_public=False)
-        dao.connect()
-        dao.select_workflow_params(_callback)
-        flow['workflow_params'] = params
+        try:
+            dao.connect()
+            dao.select_workflow_params(_callback)
+            flow['workflow_params'] = params
+        finally:
+            dao.close()
 
     return flow
```

MetRonnie:
Sorry, accidentally deleted 8.0.x!

MetRonnie (approved):

PR #5009, opened 2022-07-24
Add new job var CYLC_WORKFLOW_NAME_BASE.

### LINKED ISSUE #4977: CYLC_WORKFLOW_NAME: use the "flat" name
Many Cylc 7 workflows have been assuming a flat workflow ID (despite hierarchical registration going back to Cylc 6). With Cylc 8 hierarchical workflow IDs are now the default because run numbers are turned on by default.

To assist with the transition we added the `CYLC_WORKFLOW_NAME` environment variable (not to be mistaken with `CYLC_WORKFLOW_ID`) which is the ID with the run name/number stripped off.

Since https://github.com/cylc/cylc-flow/issues/4643 we have introduced another mechanism by which you can get hierarchical IDs "by default" in that we now transfer the directory structure from the src to the run dir.

Given its name you might expect the `CYLC_WORKFLOW_NAME` environment variable to reflect the "base name" of the workflow and not reflect its "path" either src or run. E.G for an ID of `a/b/run1` the "name" part might be considered to be `b` rather than `a/b`, especially since the src dir structure is likely to reflect project organisation.

**Question:**
* [ ] Should `CYLC_WORKFLOW_NAME` be "flat" or "hierarchical"?

**Pull requests welcome!**

hjoliver:
These changes close #4977 

As requested: did not add `CYLC_WORKFLOW_NAME_DIR`

Small compromise: I did not remove `CYLC_WORKFLOW_NAME` because  taken together it's immediately clear what the new variable is the basename of.

So now we have:
- `CYLC_WORKFLOW_ID` - e.g. `a/b/c/run1`
- `CYLC_WORKFLOW_NAME` - e.g. `a/b/c`
- `CYLC_WORKFLOW_NAME_BASE` - e.g. `c`

**Requirements check-list**


wxtim:
I've assigned myself to do a code review so that @dpmatthews doesn't have to, and so that @oliver-sanders  only has to do a second review.

tests/functional/restart/50-two-flows/flow.cylc
```diff
@@ -21,8 +21,9 @@
     [[c]]
        script = """
            if ((CYLC_TASK_FLOW_NUMBERS == 1)); then
-               cylc trigger --flow=new --meta="cheese wizard" "$CYLC_WORKFLOW_NAME//1/a"
+               cylc trigger --flow=new --meta="cheese wizard"
+               "$CYLC_WORKFLOW_ID//1/a"
```
wxtim, line 25:
```suggestion
               cylc trigger --flow=new --meta="cheese wizard" \
                   "$CYLC_WORKFLOW_ID//1/a"
```

tests/unit/test_job_file.py
```diff
@@ -279,6 +279,8 @@ def test_write_prelude(monkeypatch, fixture_get_platform, set_CYLC_ENV_NAME):
         expected += '\nexport CYLC_ENV_NAME=\'myenv\''
     expected += '\nexport CYLC_WORKFLOW_ID="test_write_prelude"'
     expected += '\nexport CYLC_WORKFLOW_INITIAL_CYCLE_POINT=\'20200101T0000Z\''
+    #expected += '\nexport CYLC_WORKFLOW_NAME="test_write_prelude"'
+    #expected += '\nexport CYLC_WORKFLOW_NAME_BASE="test_write_prelude"'
```
wxtim, line 283: confusing as comments - recommend removal

```suggestion
```

wxtim (commented):
- [x] Read
- [x] Checked out locally, and successfully run the changed tests.
- [ ] 1 broken test - suggested fix in comment.
- Thank you for getting rid of all those `${CYLC_WORKFLOW_NAMES}`.

wxtim (changes requested):
See comments.

tests/functional/restart/50-two-flows/flow.cylc
hjoliver: Ah, thanks @wxtim ; looks like I had an editor auto line break fail there after changing NAME to ID in the test.

tests/unit/test_job_file.py
hjoliver: Oops, I did not intend to leave those there - a left-over from my first attempt at adding the new variables to the test.

wxtim (approved):

hjoliver:
Docs PR merged. One review OK for this.

PR #5018, opened 2022-07-27
cli: minor docs changes for 8.0.0 release

oliver-sanders:
Had a quick review of the CLI docs, found a couple of small things.

Reformatted a couple of paragraphs which were unnecessarily "narrow" due to being copied from the schema docs?

**Requirements check-list**

### DIFF
--- a/cylc/flow/scripts/check_versions.py
+++ b/cylc/flow/scripts/check_versions.py
@@ -26,7 +26,7 @@ a workflow matches the version this script is run with.
 Note:
   Cylc supports multiple parallel installations at different versions via the
   Cylc wrapper script. For more information see the installation section of the
-  documemtation.
+  documentation.
 
 Use -v/--verbose to see the command invoked to determine the remote version
 (all remote Cylc command invocations will be of the same form, which may be
--- a/cylc/flow/scripts/check_versions.py
+++ b/cylc/flow/scripts/check_versions.py
@@ -20,17 +20,16 @@
 
 Check that Cylc versions match on different platforms.
 
-Check the version of cylc invoked on each of WORKFLOW's task host platforms
-when CYLC_VERSION is set to *the version running this command line tool*.
+Check that the remote versions of Cylc invoked on each of the platforms used by
+a workflow matches the version this script is run with.
 
-It is recommended that cylc versions be installed in parallel and access
-configured via the cylc version wrapper as described in the cylc INSTALL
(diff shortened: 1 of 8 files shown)


wxtim (approved):

cylc/flow/scripts/check_versions.py
```diff
@@ -20,17 +20,17 @@
...
 
-It is recommended that cylc versions be installed in parallel and access
-configured via the cylc version wrapper as described in the cylc INSTALL
-file and User Guide. This must be done on workflow and task hosts. Users then
-get the latest installed version by default, or (like tasks) a particular
-version if $CYLC_VERSION is defined.
+Note:
+  Cylc supports multiple parallel installations at different versions via the
+  Cylc wrapper script. For more information see the installation section of the
+  documemtation.
```
MetRonnie, line 30:
```suggestion
  documentation.
```

cylc/flow/scripts/check_versions.py
```diff
@@ -20,17 +20,17 @@
 
 Check that Cylc versions match on different platforms.
 
-Check the version of cylc invoked on each of WORKFLOW's task host platforms
-when CYLC_VERSION is set to *the version running this command line tool*.
+Check the version of Cylc invoked on each of workflow's task host platforms
+when CYLC_VERSION is set to <bold>the version running this command line
+tool</bold>.
```
MetRonnie, line 25: I am a bit unclear on what it means by "check the version... when CYLC_VERSION is set to..."

cylc/flow/scripts/validate.py
```diff
@@ -20,9 +20,9 @@
 
 Validate a workflow configuration.
 
-If the workflow definition uses include-files reported line numbers
-will correspond to the inlined version seen by the parser; use
-'cylc view -i,--inline WORKFLOW' for comparison.
+If the workflow definition uses include-files reported line numbers will
```
MetRonnie, line 23:
```suggestion
If the workflow definition uses include-files, reported line numbers will
```

cylc/flow/scripts/check_versions.py
oliver-sanders: It is checking that the remote versions are set to the local version as defined via CYLC_VERSION

oliver-sanders: Reworded.

PR #5026, opened 2022-07-29
Improve task pool integration tests

### LINKED ISSUE #5012: Upgrade test_test_pool integration tests
E.g. see https://github.com/cylc/cylc-flow/pull/5008#pullrequestreview-1049110371

MetRonnie:
These changes close #5012

- Add task - that does not get spawned at start - to task pool integration tests
- Fix duplicate results in `WorkflowConfig.find_taskdefs()` method
- Fix spurious warnings when using `TaskPool.match_taskdefs()` with a glob pattern for task names

**Requirements check-list**
- [x] Appropriate tests are included 
- [x] No change log entry required (minor).

### DIFF
--- a/cylc/flow/taskdef.py
+++ b/cylc/flow/taskdef.py
@@ -361,3 +361,6 @@ class TaskDef:
             or all(x < self.start_point for x in parent_points)
             or self.has_only_abs_triggers(point)
         )
+
+    def __repr__(self) -> str:
+        return f"<TaskDef {self.name}>"
--- a/cylc/flow/taskdef.py
+++ b/cylc/flow/taskdef.py
@@ -17,6 +17,7 @@
 """Task definition."""
 
 from collections import deque
+from typing import TYPE_CHECKING
 
 import cylc.flow.flags
 from cylc.flow.exceptions import TaskDefError
@@ -27,8 +28,10 @@ from cylc.flow.task_state import (
     TASK_OUTPUT_SUCCEEDED,
     TASK_OUTPUT_FAILED
 )
-from cylc.flow import LOG
 from cylc.flow.task_outputs import SORT_ORDERS
+
+if TYPE_CHECKING:
+    from cylc.flow.cycling import PointBase
 
 
 def generate_graph_children(tdef, point):
@@ -119,7 +122,6 @@ class TaskDef:
 
     # Store the elapsed times for a maximum of 10 cycles
     MAX_LEN_ELAPSED_TIMES = 10
-    ERR_PREFIX_TASK_NOT_ON_SEQUENCE = "Invalid cycle point for task: "
 
     def __init__(self, name, rtcfg, run_mode, start_point, initial_point):
         if not TaskID.is_valid_name(name):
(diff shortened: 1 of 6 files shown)


cylc/flow/taskdef.py
```diff
@@ -361,3 +358,6 @@ def is_parentless(self, point):
             or all(x < self.start_point for x in parent_points)
             or self.has_only_abs_triggers(point)
         )
+
+    def __repr__(self) -> str:
```
wxtim, line 362:
```suggestion
    def __repr__(self) -> str:
        """
        Examples:
            >>> from types import SimpleNamespace
            >>> x = SimpleNamespace(name="foo")
            >>> TaskDef.__repr__(x)
            '<TaskDef foo>'
        """

or add `# pragma nocover`.

IMO any test is essentially testing f-strings, so not worth it.

File: tests/integration/test_task_pool.py [hunk 2]
```diff
@@ -62,7 +63,7 @@
 def get_task_ids(
     name_point_list: Iterable[Tuple[str, Union[PointBase, str, int]]]
 ) -> List[str]:
-    """Helper function to return sorted task identities ("{name}.{point}")
+    """Helper function to return sorted task identities
     from a list of  (name, point) tuples."""
```
wxtim, line 67: I actually think that the original form was helpful. How about:?

```suggestion
    """Helper function to return sorted task identities
    from a list of  (name, point) tuples.
    
    Returns: name.point
    """
```

MetRonnie: It was out of date since universal ID. Do we really need an explanation of IDs every time they are used?

cylc/flow/taskdef.py
MetRonnie: Or just accept a miniscule drop in coverage?

tests/unit/test_config.py
```diff
@@ -1439,3 +1448,56 @@ def test_check_for_owner(runtime_cfg):
     else:
         # Assert function doesn't raise if no owner set:
         assert WorkflowConfig.check_for_owner(runtime_cfg) is None
+
+
+@pytest.fixture(scope='module')
+def awe_config(mod_tmp_flow_config: Callable) -> WorkflowConfig:
```
wxtim, line 1454: what does "awe" stand for/mean?

wxtim (commented):
Some really nice logic re-working in `config.py`, and much easier to read.

I've made a couple of small comments:
- I'm not sure what awe is.
- Test coverage has a couple of possible holes. Not very significant, but might be nice to plug.

tests/unit/test_config.py
MetRonnie: Sorry, [video game reference](https://en.m.wikipedia.org/wiki/Control_(video_game)). Just means the name of the example workflow is "awe", much like we have "one" in the integration tests

tests/integration/test_task_pool.py
wxtim: No.

cylc/flow/taskdef.py
wxtim: Whatever

wxtim (approved):
All my thoughts responded to.

cylc/flow/taskdef.py
oliver-sanders: Or, crazy idea, test it!

wxtim: Given the content that's perilously close to testing a builtin.

MetRonnie: But I don't want to test it 😥

cylc/flow/taskdef.py
```diff
@@ -361,3 +358,13 @@ def is_parentless(self, point):
...
+        """
+        >>> TaskDef(
+        ...     name='oliver', rtcfg={}, run_mode='fake', start_point='1',
+        ...     initial_point='1'
+        ... )
+        <TaskDef 'oliver'>
```
oliver-sanders, line 368: :banana:

oliver-sanders (approved):

PR #5031, opened 2022-08-01
Fix chained offset bug

### LINKED ISSUE #4908: final cycle point offset giving wrong result
> **Note:** Reported with Cylc 7, replicated with Cylc 8.

In this example the final cycle point ends up on the 30th of December rather than the 31st as expected:

```ini
[scheduling]
    cycling mode        = gregorian
    initial cycle point = 20210701T0000Z
    final cycle point   = +P6M-P1D
    [[dependencies]]
        [[[R1/^]]]
            graph = start    # 2021-07-01
        [[[R1/$]]]
            graph = stop     # 2021-12-30
```

This is surprising given isodatetime yields the 31st:

```console
$ rose date 20210701T0000Z --offset '+P6M' --offset '-P1D'  # rose 2019
20211231T0000Z
$ isodatetime 20210701T0000Z --offset '+P6M' --offset '-P1D'
20211231T0000Z
```

Swap months for years and the calculation comes out correctly:

```ini
[scheduling]                                                                         
    cycling mode        = gregorian        
    initial cycle point = 2000             
    final cycle point = +P1Y-P1D           
    [[dependencies]]                       
        [[[R1/^]]]                         
            graph = start  # 2000-01-01    
        [[[R1/$]]]  
            graph = stop   # 2000-12-31  
```

**Pull requests welcome!**

MetRonnie:
These changes close #4908

Fix bug where specifying multiple datetime offsets would not obey the given order.

E.g. `final cycle point = +P1M-P1D` would lead to it calculating `icp - P1D + P1M` (smallest unit to largest unit order) instead of `icp + P1M - P1D` (given order).

**Requirements check-list**
- [ ] [if bugfix] PRs raised to both master and the relevant bugfix branch.

### DIFF
--- a/cylc/flow/time_parser.py
+++ b/cylc/flow/time_parser.py
@@ -30,7 +30,7 @@ time for inter-cycle task references such as "foo[-P6Y] => foo".
 """
 
 import re
-from typing import TYPE_CHECKING, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, List, Optional, Pattern, Tuple, Union, cast
 
 from metomi.isodatetime.data import Duration, TimeRecurrence
 from metomi.isodatetime.exceptions import IsodatetimeError
@@ -75,7 +75,7 @@ class CylcTimeParser:
           "'T00'?"))
     ]
 
-    RECURRENCE_FORMAT_REGEXES = [
+    RECURRENCE_FORMAT_REGEXES: List[Tuple[Pattern, int]] = [
         (re.compile(r"^(?P<start>[^PR/][^/]*)$"), 3),
         (re.compile(r"^R(?P<reps>\d+)?/(?P<start>[^PR/][^/]*)/(?P<end>[^PR/]"
                     "[^/]*)$"), 1),
@@ -222,8 +222,8 @@ class CylcTimeParser:
             repetitions = result.groupdict().get("reps")
             if repetitions is not None:
                 repetitions = int(repetitions)
-            start = result.groupdict().get("start")
-            end = result.groupdict().get("end")
+            start: Optional[str] = result.groupdict().get("start")
+            end: Optional[str] = result.groupdict().get("end")
             start_required = (format_num in {1, 3})
             end_required = (format_num in {1, 4})
             start_point, start_offset = self._get_point_from_expression(
@@ -290,7 +290,9 @@ class CylcTimeParser:
                 # isodatetime only reverses bounded end-point recurrences.
                 # This is unbounded, and will come back in reverse order.
                 # We need to reverse it.
-                start_point = end_point
+                start_point = cast(  # (end pt can't be None if start is None)
+                    'TimePoint', end_point
+                )
                 repetitions = 1
                 while start_point > context_start_point:
                     start_point -= interval
@@ -355,7 +357,7 @@ class CylcTimeParser:
     def _get_min_from_expression(
         self,
         expr: str,
-        context: 'TimePoint'
+        context: Optional['TimePoint']
     ) -> str:
         points: List[str] = [
             x.strip() for x in re.findall(self.MIN_REGEX, expr)[0].split(",")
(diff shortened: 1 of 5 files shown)


cylc/flow/time_parser.py
```diff
@@ -130,12 +130,12 @@ def __init__(
         )
 
         if isinstance(context_start_point, str):
-            context_start_point = self._get_point_from_expression(
-                context_start_point, None)[0]
+            context_start_point, _ = self._get_point_from_expression(
+                context_start_point, None)
```
oliver-sanders, line 134: FYI: This is subjective, the way this was done before saved an unnecessary assignment.

cylc/flow/time_parser.py
```diff
@@ -178,23 +178,23 @@ def parse_timepoint(
...
-        that supplies the missing information for truncated
-        expressions. For example, context_point should be the task
-        cycle point for inter-cycle dependency expressions. If
-        context_point is None, self.context_start_point is used.
+        expr: a string such as 20010205T00Z, or a truncated/abbreviated format
+            string such as T06 or -P2Y.
+        context_point: supplies the missing information for truncated
+            expressions. E.g. for inter-cycle dependency expressions,
+            it should be the task cycle point.
+            If None, self.context_start_point is used.
```
oliver-sanders, line 186:
```suggestion
        Args:
            expr: a string such as 20010205T00Z, or a truncated/abbreviated format
                string such as T06 or -P2Y.
            context_point: supplies the missing information for truncated
                expressions. E.g. for inter-cycle dependency expressions,
                it should be the task cycle point.
                If None, self.context_start_point is used.
```

cylc/flow/time_parser.py
```diff
@@ -355,78 +358,69 @@ def _get_interval_from_expression(
...
-                if offset is not None:
+                    cpoint += context  # type: ignore[operator]
+                for offset in offsets:
                     cpoint += offset
-            ptslist.append(cpoint)
-            if cpoint == min(ptslist):
-                min_entry = point
+                ptslist.append(cpoint)
+                if cpoint == min(ptslist):
+                    min_entry = point
```
oliver-sanders, line 379:
```python
>>> min([1, None])
Traceback ...
TypeError
```

How has this one not blown up before?

Hole in the tests?

MetRonnie: Not sure. Perhaps it is very unlikely to be hit. It's difficult to tell because the `self._get_point_from_expression()` method probably does too much for a single method.

As for Mypy not flagging the appending `None` to `List[TimePoint]`, I think that might be due to the fact we're missing `py.typed` in isodatetime (added in https://github.com/metomi/isodatetime/pull/203)

MetRonnie:
oops

MetRonnie:
(Fast test failure on MacOS due to Codecov uploader failing again...)

cylc/flow/time_parser.py
```diff
@@ -355,78 +359,69 @@ def _get_interval_from_expression(
     def _get_min_from_expression(
         self,
         expr: str,
-        context: 'TimePoint'
+        context: Optional['TimePoint']
     ) -> str:
```
wxtim, line 363: A docstring might have been nice?

MetRonnie: (Sighing) Yes it would have. When someone figures out what this is doing it would also be a good idea to add a unit test

tests/unit/test_config.py
```diff
@@ -1395,6 +1397,35 @@ def test_zero_interval(
...
+    """Test a "chain expression" final cycle point offset works in the
+    given order."""
+    reg = 'osgiliath'
+    flow_file: Path = tmp_flow_config(reg, f"""
+        [scheduler]
+            UTC mode = True
+            allow implicit tasks = True
+        [scheduling]
+            initial cycle point = {icp}
+            final cycle point = {fcp_expr}
```
wxtim, line 1421: I'm not sure if `_parse_chain_expression` is used elsewhere, but if it is, it's worth noting that this only works if the sum of the durations > 0.

Also, should
```python
        ('2000-01-01T00:00', '+PT6H-PT1S', '2000-01-01T05:59'),
```
Work?

cylc/flow/time_parser.py
wxtim: Out of scope to be fair.

tests/unit/test_config.py
MetRonnie: - I've added more tests
- I've opened an issue for the `+PT6H+PT1S` bug: #5047 
- `parse_chain_expression()` works for net < 0 durations but that does not produce a valid final cycle point, because it is before the initial cycle point (already a tested scenario)

wxtim (approved):

oliver-sanders (approved):

PR #5034, opened 2022-08-02
Update README.md

oliver-sanders:
* Remove Cylc 7 section from the README.
* Add a quickstart with a simple example.

Thoughts, suggestions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,9 @@
 Cylc (pronounced silk) is a general purpose workflow engine
 that specialises in cycling workflows and has strong scaling characteristics.
 
-Cylc was originally developed to meet the challenges of production weather
-forecasting - which is notorious for the size and complexity of its workflows.
+Cylc is a general purpose workflow engine that also orchestrates cycling systems very efficiently.
+It is used in production weather, climate, and environmental forecasting on HPC, but is
+not specialized to those domains.
 
 ### Quick Start
 
--- a/README.md
+++ b/README.md
@@ -20,42 +20,42 @@ that specialises in cycling workflows and has strong scaling characteristics.
 Cylc was originally developed to meet the challenges of production weather
 forecasting - which is notorious for the size and complexity of its workflows.
 
+### Quick Start
+
+
+[Installation](https://cylc.github.io/cylc-doc/stable/html/installation.html) |
+[Documentation](https://cylc.github.io/cylc-doc/stable/html/index.html)
+
+```bash
+# install cylc
+conda install cylc-flow
+
+# write your first workflow
+mkdir -p ~/cylc-src/example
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,42 +20,42 @@ that specialises in cycling workflows and has strong scaling characteristics.
...
+conda install cylc-flow
+
+# write your first workflow
+mkdir -p ~/cylc-src/example
+cat > ~/cylc-src/example/flow.cylc <<__CONFIG__
+[scheduling]
+    initial cycle point = 1
+    cycling mode = integer
+    [[graph]]
+        P1 = a => b[-P1] => c & d
```
MetRonnie, line 40: `b` isn't valid here?

MetRonnie (changes requested):

README.md
oliver-sanders, line 40:
```suggestion
        P1 = """
            a => b => c & d
            b[-P1] => b
```

hjoliver (commented):
Looks good.

I'd suggest changing the description slightly to match that on the web site:

> Cylc is a general purpose workflow engine that also orchestrates cycling systems very efficiently. It is used in production weather, climate, and environmental forecasting on HPC, but is not specialized to those domains. 

(The "also" bit is intended to allay concerns that you can only use Cylc for cycling workflows).

README.md
```diff
@@ -20,42 +20,45 @@ that specialises in cycling workflows and has strong scaling characteristics.
 Cylc was originally developed to meet the challenges of production weather
 forecasting - which is notorious for the size and complexity of its workflows.
```
oliver-sanders, line 21:
```suggestion
Cylc is a general purpose workflow engine that also orchestrates cycling systems very efficiently.
It is used in production weather, climate, and environmental forecasting on HPC, but is
not specialized to those domains.
```

README.md
```diff
@@ -17,45 +17,49 @@
 Cylc (pronounced silk) is a general purpose workflow engine
 that specialises in cycling workflows and has strong scaling characteristics.
 
-Cylc was originally developed to meet the challenges of production weather
-forecasting - which is notorious for the size and complexity of its workflows.
+Cylc is a general purpose workflow engine that also orchestrates cycling systems very efficiently.
+It is used in production weather, climate, and environmental forecasting on HPC, but is
+not specialized to those domains.
```
MetRonnie, line 22: A bit of repetition in these two paragraphs

hjoliver: Better to replace the 1st with the 2nd (maybe keep "pronounced silk" though).

hjoliver: I pushed a commit to do this.

hjoliver (approved):

Issue #4903, opened 2022-06-01
Multiple distinct graphs


hjoliver:
This comes up fairly regularly, but we don't seem to have an issue for it.

Many (most?) workflows have "prep" tasks in the first cycle that need to finish before any future-cycle tasks begin. And similarly, there's often a need for special clean up tasks in the final cycle point, after everything else has finished.

We can do this easily enough (use prep tasks as an example) with perpetual dependence on the initial cycle point:
```ini
R1 = "prep"
P1 = "prep[^] => foo => bar"
```
or with artificial dependencies to force the first tasks in each cycle to start in order:
```ini
R1 = "prep => foo"
P1 = "foo[-P1]:submit => foo => bar"
```
However, neither of these solutions is very elegant. The first is correct in principle, but it seems a bit silly retain the `^` dependency forever, and it has performance implications for long-running workflows due to increasingly long datetime sequence computations. The second is artificial and unnecessarily prevents cycles from running out of order.

What users really want here is just a separate "start-up graph" that runs to completion before the main cycling graph starts.

### Proposal A: allow multiple sequential `graph`s under the scheduling section

```ini
[scheduling]
   cycling mode = integer
   initial cycle point = 1
   [[graph_1]]
        R1 = "prep"
   [[graph_2]]
        P1 = "foo => bar" 
   [[graph_3]]
        R1 = "cleanup"
```

### Proposal B: allow multiple sequential `[scheduling]` sections

Multiple `graph`s would probably suffice for most use cases, but for more power and flexibility we should have separate control of cycling,  queues, etc., for each of the distinct workflow graphs.

```ini
[scheduling_1]
   [[graph]]
        R1 = "prep"

[scheduling_2]
   cycling mode = integer
   initial cycle point = 1
   [[graph]]
        P1 = "foo => bar" 

[scheduling_3]
   [[graph]]
        R1 = "cleanup"
```

hjoliver:
I think this would be quite easy to implement. The scheduler only needs to manage one graph at a time. If there is nothing more to do (and we're not stalled) shut down if there are no more graphs, or start running the next graph if there is one.

hjoliver:
(Context: in my experience, this is a common problem for new users: how do I stop cycle 2 from starting before cycle 1 is finished doing its prep work?)

oliver-sanders:
[...]

PR #5036, opened 2022-08-02
Isolate ICP and FCP graphs using the runahead limit

### LINKED ISSUE #4912: Variable runahead limit, for start-up graphs.
Supercedes #4903 (via an @oliver-sanders suggestion there).

Optionally tie the runahead limit down to `P1` until the first cycle has finished, to avoid the ugliness of making tasks in every subsequent cycle depend explicitly on the first (if they aren't otherwise constrained by dependencies).

hjoliver:
A bit of late night fun (sad!).

Implement "start-up and shutdown graphs" (effectively isolated from the main graph so that perpetual dependence on R1 & $ tasks can be avoided) ... by manipulating the runahead limit. 

These changes close #4912

Imagine a task `prep` is supposed to prepare the run directory for all tasks; and `clean` is supposed to tidy up at the end after all tasks have finished:
```ini
[scheduling]
    cycling mode = integer
    final cycle point = 5
    isolate initial cycle point = True
    isolate final cycle point = True
    [[graph]]
        R1 = "prep => foo"
        P1 = "foo"
        R1/$ = "foo => clean"
[runtime]
    [[prep, clean, foo]]
        script = sleep 10
```

With all tasks having the same run length, on master (without the new config items) `1/prep` runs at the same time as `2/foo, 3/foo, 4/foo, 5/foo`, then `1/foo` runs at the same time as `5/clean`.  Which is clearly not the intention, so we'd need additional ugly dependencies to make it work.

Or ... on this PR branch, `1/prep => 1/foo` runs to completion, then `2/foo, 3/foo, 4/foo` run to completion, then `5/foo => 5/clean` runs. :boom: 

**Requirements check-list**

- [x] (master branch) I have opened a documentation Issue at https://github.com/cylc/cylc-doc/issues/511


hjoliver:
Assigning you as first reviewer @oliver-sanders, since it is was your idea to get "start-up graphs" cheaply and safely by manipulating the runahead limit (and a fine idea it was).

dwsutherland (approved):
Good feature, I might implore our operations team to switch to this.
Would be good if there was a short hand way of specifying the ICP without knowing it via the CLI (i.e. if we had to reflow the prep tasks).

Tested locally before and after (True/False), and found working 👍

dpmatthews:
```
    [[graph]]
        R1 = "prep => foo"
        P1 = "foo"
        R1/$ = "foo => clean"
```
Not sure I like this. One of the key features of Cylc is the support for parallel cycles. In this example, if `foo` is a very long task (or series of tasks) and you want to run lots of cycles in parallel, it's going to be very frustrating if you have to spend a long waiting for the first cycle to complete before the workflow can start running cycles in parallel as intended.

One way around this is to make the start and end tasks run on separate cycles:
```
    [[graph]]
        R1 = "prep"
        P1!(^,$) = "foo" # or R/+P1/P1!$
        R1/$ = "clean"
```
(syntax provided by @oliver-sanders !)
However, the syntax is horrible and you also have to adjust the start and stop points.

It would be much nicer if we could do something like:
```
    [[graph]]
        start = "prep"
        P1 = "foo"
        finish = "clean"
```
I guess that's a lot more work?

hjoliver:
@dpmatthews - 

> It would be much nicer if we could do something like:

Haha, that was my original suggestion, albeit explicitly restricted to start and finish graphs) - you might have chimed in back on #4903!

> I guess that's a lot more work?

Well, I can think about it, if you and @oliver-sanders agree - he didn't like the separate graphs approach, which is how we ended up here. But maybe the restriction to "start" and "end" graphs would make that acceptable - what do think Oliver?

I think the runahead limitation is OK. It's opt-in, we can explain the potential downside and that `prep[^] => foo`) is an alternative. 

Also, worth noting that in a workflow with many cycles, less than optimal scheduling in the first might not matter much.

>  However, the syntax is horrible and you also have to adjust the start and stop points.

inter-cycle triggers would make this way even worse (the offset would break pre-initial ignore behaviour).

hjoliver:
Long story short, I also think this is better:
```
    [[graph]]
        start = "prep"
        P1 = "foo"
        finish = "clean"
```
If you look at the description of #4903 that's exactly the use case, but maybe I shot myself in the foot there by making it look more general than that.

dpmatthews:
> you might have chimed in back on https://github.com/cylc/cylc-flow/issues/4903!!

Sorry!

> But inter-cycle triggers would make that even worse (the offset would break pre-initial ignore).

Yuck - hadn't thought of that.

oliver-sanders:
> @oliver-sanders agree - he didn't like the separate graphs approach

I'm ok with separate graphs, if we could handle true sub-graphs in Cylc that would be the ultimate solution but that's going to have to be a longer term plan.

> Long story short, I also think this is better:
>
>   [[graph]]
>       start = "prep"
>        P1 = "foo"
>        finish = "clean"

It's definitely better, the issue is trying to work out how to implement it! The "start" and "finish" cycles need to be assigned a "cycle point" but what should that be? We could do something like subtract `PT1S` from the ICP at configuration time, but that's a bit icky (and won't necessarily work with `cycle point format`). Or we could implement `start` and `finish` as special cycle points, but that requires patching all cycling interfaces to handle these special keywords.

hjoliver:
> Or we could implement start and finish as special cycle points, but that requires patching all cycling interfaces to handle these special keywords.

I was thinking along those lines, but I haven't thought through it in depth yet.

hjoliver:
Superseded by #5090

Issue #4866, opened 2022-05-11
reload: xtriggers should be preserved by reloads


oliver-sanders:
At the moment if you reload a workflow any `execution retry delays` (probably the same for submission too) get wiped.

E.G. if you run the following that `PT1H` retry delay will be ignored after the reload and the task will run right away:

```
[scheduling]                                                                      
    [[graph]]                                                                     
        R1 = """                                                                  
            flaky:fail? => reload                                                                                              
        """                                                                       
                                                                                  
[runtime]                                                                         
    [[flaky]]                                                                     
        script = """                                                              
            # REPLACE                                                             
            cylc trigger "${CYLC_WORKFLOW_ID}//${CYLC_TASK_CYCLE_POINT}/reload"
            false                                                                 
        """                                                                       
        execution retry delays = PT1H                                                             
                                                                                  
    [[reload]]                                                                    
        script = """         
            sleep 2  # let flaky/01 fail    
            sed -i 's/# REPLACE/exit 0/' "${CYLC_WORKFLOW_RUN_DIR}/flow.cylc"    
            cylc reload "${CYLC_WORKFLOW_ID}"    
        """    
```

Of course the retry delays can change as a result of the reload which is probably the reason for the current behaviour? However, maybe it can be done more nicely?

What do we think the correct behaviour is here?

**Pull requests welcome!**

dpmatthews:
I think reload shouldn't affect any existing retry delay (i.e. an existing xtrigger) but should affect any subsequent retry delay.

oliver-sanders:
This turns out to be a symptom of a larger issue with xtriggers and reloads.

The cause is the same as for #4984. When we perform a reload we create a new task proxy and transfer the attributes of the old onto it. For some reason, xtriggers are not currently transferred.

This is likely a Cylc 7 issue which has only been spotted at Cylc 8 because we've converted the retry mechanism to use xtriggers.

oliver-sanders:
> This is likely a Cylc 7 issue which has only been spotted at Cylc 8 because we've converted the retry mechanism to use xtriggers.

Cannot replicate at Cylc 7 or 8 with pre-configured xtriggers e.g:

```
@mytrig => foo
```

I think it's only triggers which were dynamically added at runtime which pose an issue.

PR #5040, opened 2022-08-03
reload: preserve retry xtriggers on reload

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/4866
* Reload causes new instances of each task to be created.
* The attributes of the old instances are copied onto the new ones.
* Retry xtriggers were not copied so were lost.
* This means that all retry pending tasks would run immediately on reload.

**Requirements check-list**

### DIFF
--- a/tests/functional/reload/25-xtriggers.t
+++ b/tests/functional/reload/25-xtriggers.t
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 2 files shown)


tests/functional/reload/25-xtriggers.t
```diff
@@ -0,0 +1,72 @@
...
+            broken
+            reload
+        """
+
+[runtime]
+    [[broken]]
+        script = false
+        # should be long enough for the reload to complete
+        # (annoyingly we can't make this event driven)
+        execution retry delays = PT1M
```
MetRonnie, line 39: Would it be worth moving this test into `tests/flakyfunctional` and possibly cutting down this delay, in order that it doesn't hog time in the `tests/functional` runs on GH Actions?

MetRonnie (approved):
LGTM

tests/functional/reload/25-xtriggers.t
MetRonnie: I know, just conscious of the fact we are approaching the timeout limit on GH Actions. May have to either increase the limit or add an extra chunk soon

hjoliver (approved):
Good catch :+1:

PR #5056, opened 2022-08-11
Add workflow field to `ClientError`, `ClientTimeout`

MetRonnie:
Needed for https://github.com/cylc/cylc-uiserver/pull/379

**Check List**

- [x] Tests are not needed.
- [x] No changelog entry required
- [x] No docs update needed.

### DIFF
--- a/cylc/flow/exceptions.py
+++ b/cylc/flow/exceptions.py
@@ -254,9 +254,17 @@ class XtriggerConfigError(WorkflowConfigError):
 
 class ClientError(CylcError):
 
-    def __init__(self, message: str, traceback: Optional[str] = None):
+    def __init__(
+        self,
+        message: str,
+        traceback: Optional[str] = None,
+        workflow: Optional[str] = None
+    ):
         self.message = message
         self.traceback = traceback
+        # Workflow not included in string representation but useful bit of
+        # info to attach to the exception object
+        self.workflow = workflow
 
     def __str__(self) -> str:
         ret = self.message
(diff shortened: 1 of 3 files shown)


wxtim (approved):
Yep

oliver-sanders:
The UIS PR is up against 1.2.0 which will use `8.1.0` so this should go against `8.1.0` too.

MetRonnie:
It depends on if we want to do a maintenance release for the UIS? (No milestone for that currently)

oliver-sanders:
I don't think we need a UIS maintenance release for this.

cylc/flow/exceptions.py
```diff
@@ -254,9 +254,17 @@ def __str__(self):
 
 class ClientError(CylcError):
 
-    def __init__(self, message: str, traceback: Optional[str] = None):
+    def __init__(
+        self,
+        message: str,
+        traceback: Optional[str] = None,
+        workflow: Optional[str] = None
```
oliver-sanders, line 261: Might be a good idea to call this `workflow_id` as we also have the tokenised variant.

oliver-sanders (commented):
LGTM, gotta couple of unexpected commits here.

MetRonnie:
> LGTM, gotta couple of unexpected commits here.

Ah, that would be because the base was changed from 8.0.x to master, so it wouldn't have mattered (I'll rebase anyway)

cylc/flow/exceptions.py
MetRonnie: Don't think it matters much for these exceptions. The more important one would be `WorkflowRuntimeClient`
https://github.com/cylc/cylc-flow/blob/af64ace2f078b7d656c83c09db483a531f43b98f/cylc/flow/network/client.py#L64-L66

however that's probably out of scope

oliver-sanders (approved):

PR #5028, opened 2022-07-29
scan: switch from pyuv to os.listdir in asyncio executor

oliver-sanders:
* Closes #5006
* We were using pyuv to perform directory listing.
* Unfortunately pyuv is no longer actively maintained.
* Switching to the simpler method of running `os.listdir` in an executor
  ala aiofiles.
* Unblocks #4524

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.
- [x] Does not need tests (why?).
- [x] No change log entry required (why? e.g. invisible to users).
- [x] No documentation update required.

PR #5065, opened 2022-08-12
scan: ignore FileNotFoundError

oliver-sanders:
* During a scan directories are listed asynchronously and recursively.
* If a directory is queued for listing, but then deleted before the
  listing takes place a FileNotFoundError will occur.
* This is highly unlikely IRL, however, in the integration tests, where run
  dirs are created and destroyed at scale and in parallel this is
  actually quite likely.

Example fail: https://github.com/cylc/cylc-flow/runs/7805964418?check_suite_focus=true

Likely caused by #5028, not sure why, perhaps it didn't raise the error, perhaps it was more eager with its scheduling.

Tested with 110X parallelism (`pytest -n 110 tests/integration`), couldn't reproduce post-fix.

**Check List**

### DIFF
--- a/cylc/flow/network/scan.py
+++ b/cylc/flow/network/scan.py
@@ -250,7 +250,12 @@ async def scan(
             return_when=asyncio.FIRST_COMPLETED
         )
         for task in done:
-            path, depth, contents = task.result()
+            try:
+                path, depth, contents = task.result()
+            except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


MetRonnie (approved):

wxtim (approved):
I can't mathematically prove that this is the fix. But empirically it looks like the fix.

PR #5066, opened 2022-08-14
fix install w/ .cylcignore

hjoliver:
A `.cylcignore` file works as expected if you do `cylc install .` from within the source dir. But otherwise, this happens:
 
```console
WorkflowFilesError: No flow.cylc or suite.rc in /home/oliverh/cylc-run/bug/run8
```
and the install log says:
```console
WARNING - An error occurred when copying files from /home/oliverh/cylc-src/bug to /home/oliverh/cylc-run/bug/run8
WARNING -  Warning: rsync: failed to open exclude file .cylcignore: No such file or directory (2)
    rsync error: error in file IO (code 11) at exclude.c(1207) [client=3.1.2]
```
**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.
- [x] Tests are included (or explain why tests are not needed).
- [ ] `CHANGES.md` entry included if this is a change that can affect users
- [ ] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [ ] If this is a bug fix, PRs raised to both master and the relevant maintenance branch.

PR #5068, opened 2022-08-15
Hilary.fix .cylcignore

wxtim:
copy of https://github.com/cylc/cylc-flow/pull/5066 against 8.0.x


oliver-sanders:
Branch would require rebasing onto 8.0.x to strip out 8.1.0 commits. Bugfixes should be raised against the 8.0.x branch, then we can simply merge `8.0.x` into `master`.

PR #5074, opened 2022-08-16
play: check Cylc version on restart

### LINKED ISSUE #4039: handle version change on restart (resume)
`cylc restart` (soon to be renamed `cylc play`) currently [8.0a3.dev] restarts a workflow using whatever version of Cylc is selected by the current environment.

For many this is a departure from `rose suite-restart` which sets the `CYLC_VERSION` to the version of the previous run. This old behaviour has an upside and a downside in that whist it ensures a compatible Cylc version is selected, it prohibits restarts from benefiting from bug fixes in maintainance releases.

> **Note:** This issue does not cover the configuration of a Cylc version or compatible version range for a workflow
> which is a different matter (#3780 may provide a mechanism for this).

Proposal:

> **Note:** for version 8.1.2 (major=8, minor=1, maintainance=2)

> **Edit 2021-11-29:** Re worded to avoid confusion.

* Restart normally if the restart would:
  * Not change the version.
  * Increase the maintenance number.
* Prompt (override with `--yes`) for confirmation if the restart would:
  * Reduce the maintenance number.
  * Increase the minor number.
* Exit with nasty error (override with `--force` or `--yes --yes`?) if the restart would:
  * Decrease the minor number.
  * Change the major number.

Thoughts?

**Pull requests welcome!**

oliver-sanders:
* Closes #4039
* When restarting a workflow, check the version it previously ran under
  against the current version in order to determine if the change is
  sensible.
  * Restart normally if the restart would:
    * Not change the version.
    * Increase the maintenance number.
  * Prompt (override with --yes) for confirmation if the restart would:
    * Reduce the maintenance number.
    * Increase the minor number.
  * Exit with nasty error (override with --force or --yes --yes?) if
    the restart would:
    * Decrease the minor number.
    * Change the major number.
* This also moves the pre-existing workflow DB check from post-detatch
  to pre-detach.

**Check List**

### DIFF
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -18,6 +18,7 @@
 from ansimarkup import parse as cparse
 import asyncio
 from functools import lru_cache
+from itertools import zip_longest
 from pathlib import Path
 from shlex import quote
 import sys
@@ -55,6 +56,7 @@ from cylc.flow.workflow_files import (
 )
 from cylc.flow.terminal import (
     cli_function,
+    is_terminal,
     prompt,
 )
 
@@ -240,12 +242,22 @@ def get_option_parser(add_std_opts: bool = False) -> COP:
     )
 
     parser.add_option(
-        '--force',
+        '--downgrade',
         help=(
-            'By default Cylc prevents you from restarting a workflow with an'
+            'Allow the workflow to be restarted with an older version of Cylc,'
+            ' NOT RECOMMENDED.'
+            ' By default Cylc prevents you from restarting a workflow with an'
             ' older version of Cylc than it was previously run with.'
-            ' Use this flag to disable this check, note, this is not'
-            ' recommended!'
+            ' Use this flag to disable this check.'
+        ),
+        action='store_true',
+        default=False
+    )
+
+    parser.add_option(
+        '--upgrade',
+        help=(
+            'Allow the workflow to be restarted with an newer version of Cylc.'
         ),
         action='store_true',
         default=False
@@ -316,7 +328,11 @@ def scheduler_cli(options: 'Values', workflow_id_raw: str) -> None:
 
     # check the workflow can be safely restarted with this version of Cylc
     db_file = Path(get_workflow_srv_dir(workflow_id), 'db')
-    if not _version_check(db_file, options.force):
+    if not _version_check(
+        db_file,
+        options.upgrade,
+        options.downgrade,
+    ):
         sys.exit(1)
 
     # re-execute on another host if required
@@ -381,7 +397,11 @@ def _resume(workflow_id, options):
         sys.exit(0)
 
 
-def _version_check(db_file: Path, force: bool) -> bool:
+def _version_check(
+    db_file: Path,
+    can_upgrade: bool,
+    can_downgrade: bool
+) -> bool:
     """Check the workflow can be safely restarted with this version of Cylc."""
     if not db_file.is_file():
         # not a restart
@@ -390,17 +410,15 @@ def _version_check(db_file: Path, force: bool) -> bool:
     this_version = parse_version(__version__)
     last_run_version = wdbm.check_workflow_db_compatibility()
 
-    for itt, (this, that) in enumerate(zip(
+    for itt, (this, that) in enumerate(zip_longest(
         this_version.release,
         last_run_version.release,
+        fillvalue=-1,
     )):
-        if this is None:
-            this = -1
-        if that is None:
-            that = -1
         if this < that:
             # restart would REDUCE the Cylc version
-            if force:
+            if can_downgrade:
+                # permission to downgrade given in CLI flags
                 LOG.warning(
                     'Restarting with an older version of Cylc'
                     f' ({last_run_version} -> {__version__})'
@@ -416,7 +434,7 @@ def _version_check(db_file: Path, force: bool) -> bool:
                 f' <green>{last_run_version}</green>.'
                 f'\n* This version of Cylc is <red>{__version__}</red>.'
 
-                '\nUse --force to disable this check (not recommended!) or'
+                '\nUse --downgrade to disable this check (NOT RECOMMENDED!) or'
                 ' use a more recent version e.g:'
                 '<blue>'
                 f'\n$ CYLC_VERSION={last_run_version} {" ".join(sys.argv[1:])}'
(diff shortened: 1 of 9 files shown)


hjoliver (changes requested):
Prompt format should be of the form:
```
Question blah? [a](b/c)_
# or
Question blah [a](b/c)?_ 
```
(ends with a blank space; choices in parentheses; no colon `?:`)

oliver-sanders:
(I failed to `git add` the unit tests for this, will push up tomorrow)

hjoliver:
Looks like the prompt is messing with the functional tests:
```
This workflow was previously run with Cylc 8.0rc1.dev0.
This is Cylc 8.1.0.dev.
Are you sure you want to upgrade from Cylc 8.0rc1.dev0 to 8.1.0.dev: y,n
```
(huh?!)

oliver-sanders:
Makes sense, in which case we need to have a CLI override to handle this.

* `--force` Permit downgrade [implemented].
* `--yes, -y` Answer "yes" to prompts (permits upgrade) [suggested].

Thoughts?

hjoliver:
> --yes, -y Answer "yes" to prompts (permits upgrade) [suggested].
> Thoughts?

What's the reasoning behind `--force` vs `--yes` for upgrade vs downgrade?  Presumably we could have `--yes` handle both cases, which would be slightly more convenient than `--force` which requires re-running the command. But maybe I'm missing something?

oliver-sanders:
> But maybe I'm missing something?

Upgrading is safe, downgrading is dangerous. Implementing the two via the same option conflates the two and makes situations where `cylc play` is called in scripts awkward.

hjoliver:
Makes sense. In that case, I guess we do need `--yes` as well.

oliver-sanders:
Alternatively we could make `--yes` the default behaviour if the command is run non-interactively?

oliver-sanders:
Added `--upgrade` (safe'ish) and `--downgrade` (nasty) options.

* In interactive mode user's are prompted for permission to upgrade.
* Otherwise use `--upgrade` and `--downgrade`.

cylc/flow/scheduler_cli.py
```diff
@@ -360,6 +376,120 @@ def scheduler_cli(options: 'Values', workflow_id_raw: str) -> None:
...
+                '<red>'
+                'It is not advisible to restart a workflow with an older'
+                ' version of Cylc than it was previously run with.'
+                '</red>'
+
+                '\n* This workflow was previously run with'
+                f' <green>{last_run_version}</green>.'
+                f'\n* This version of Cylc is <red>{__version__}</red>.'
+
+                '\nUse --force to disable this check (not recommended!) or'
```
wxtim, line 439: I don't think you allow this - isn't it `--downgrade`?

cylc/flow/scheduler_cli.py
```diff
@@ -360,6 +376,120 @@ def scheduler_cli(options: 'Values', workflow_id_raw: str) -> None:
...
+                f'\n$ CYLC_VERSION={last_run_version} {" ".join(sys.argv[1:])}'
+                '</blue>'
+            ))
+            return False
+        elif itt < 2 and this > that:
+            # restart would INCREASE the Cylc version in a big way
+            if can_upgrade:
+                # permission to upgrade given in CLI flags
+                LOG.warning(
+                    'Restarting with an newer version of Cylc'
```
wxtim, line 451:
```suggestion
                    'Restarting with a newer version of Cylc'

File: cylc/flow/scheduler_cli.py [hunk 3]
```diff
@@ -360,6 +376,120 @@ def scheduler_cli(options: 'Values', workflow_id_raw: str) -> None:
...
+) -> bool:
+    """Check the workflow can be safely restarted with this version of Cylc."""
+    if not db_file.is_file():
+        # not a restart
+        return True
+    wdbm = WorkflowDatabaseManager(db_file.parent)
+    this_version = parse_version(__version__)
+    last_run_version = wdbm.check_workflow_db_compatibility()
+
+    for itt, (this, that) in enumerate(zip(
```
wxtim, line 412: easier to read if `this, that` were `now, before` IMO, or even:

```python
for level, (now, before) in zip(['major', 'minor', 'maint'], zip(x.release, y.release))]:
```

It would make the conditions slightly harder to write, but much easier to read, and more self documenting.

cylc/flow/scheduler_cli.py
```diff
@@ -360,6 +376,120 @@ def scheduler_cli(options: 'Values', workflow_id_raw: str) -> None:
...
+        this_version.release,
+        last_run_version.release,
+    )):
+        if this is None:
```
wxtim, line 416: I think codecov is correct - ensure that you test these.

cylc/flow/terminal.py
```diff
@@ -285,3 +288,43 @@ def wrapper(*api_args: str) -> None:
                 raise exc from None
         return wrapper
     return inner
+
+
+def prompt(message, options, default=None, process=None):
+    """Dead simple CLI textural prompting.
```
wxtim, line 294:
```suggestion
    """Dead simple CLI textual prompting.
```

https://www.dictionary.com/browse/textural

tests/unit/test_terminal.py
```diff
@@ -152,3 +155,50 @@ def test_cli(
...
+    # test incorrect input (should re-prompt until it gets a valid response)
+    stdinput('40', '41', '42')
+    assert prompt('what is the answer', ['42']) == '42'
+
+    # test a prompt with a default
+    stdinput('')
+    assert prompt('whatever', ['x'], default='x') == 'x'
+
+    # test a prompt with an input pre-process method thinggy
+    stdinput('YES')
```
wxtim, line 203: You missed "What is the airspeed velocity of an unladen sparrow?"

wxtim (changes requested):
- [x] Checked out and tried.
- [x] Read code.

I've made a couple of suggestions about variable naming - I'm particularly unhappy about `this` and `that` which I find very unhelpful.

cylc/flow/scheduler_cli.py
oliver-sanders: IMO "this version" and "that version" make more sense than "now version" and "before version".

wxtim: Happy to leave it to your judgement.

wxtim (approved):
You have answered all my earlier review comments.

oliver-sanders:
The DB upgrader test is now broken due to the upgrading being performed in the `cylc play` script rather than via the Scheduler object. Needs resolving.

cylc/flow/scheduler_cli.py
oliver-sanders: Done.

PR #5077, opened 2022-08-16
rundb: use sqlite3.Cursor.execute not Connection.execute.

### LINKED ISSUE #5005: rundb: use Cursor.execute* not Connection.execute*
Spotted whilst reviewing the DB code, Cylc uses the `sqlite3.Connection.execute*` methods rather than the documented approach which uses the `sqlite3.Cursor.execute*` methods.

#### Full Explanation

The Python sqlite3 docs call the `Cursor.execute*` methods e.g:

```python
conn = sqlite3.connect('/tmp/example')
c = conn.cursor()
c.execute('''create table stocks
```

This advice goes back at least as far at [Python 2.6](https://docs.python.org/2.6/library/sqlite3.html?highlight=sqlite3#module-sqlite3), however, Cylc calls the `Connection.execute*` methods instead e.g:

https://github.com/cylc/cylc-flow/blob/e92f7e8a9a2ebbc5c60e80c79fa24912719a2a80/cylc/flow/rundb.py#L487-L488

(where `self.conn` is a `sqlite3.Connection` object).

According to the docs `Connection.execute*` involves the creation of an intermediate `Cursor` object:

> execute(sql[, parameters])[¶](https://docs.python.org/3.7/library/sqlite3.html#sqlite3.Connection.execute)
>
>    This is a nonstandard shortcut that creates a cursor object by calling the [cursor()](https://docs.python.org/3.7/library/sqlite3.html#sqlite3.Connection.cursor) method, calls the cursor’s [execute()](https://docs.python.org/3.7/library/sqlite3.html#sqlite3.Cursor.execute) method with the parameters given, and returns the cursor.
>
> -- https://docs.python.org/3.7/library/sqlite3.html#sqlite3.Connection.commit

I'm not too keen on the word "nonstandard" here, but something which might be a larger concern is that due to the way we perform DB writes, we may be creating multiple intermediate cursor objects:

https://github.com/cylc/cylc-flow/blob/e92f7e8a9a2ebbc5c60e80c79fa24912719a2a80/cylc/flow/rundb.py#L423-L425

(where `_execute_stmt` calls `Connection.executemany`).

#### Motive

I don't know if this is an issue, probably not, we've used this approach for years without issue, however, unless there's a good reason not to I would prefer to follow the documented approach for piece of mind.

It's a quick piece of work to change approach, however, this should go into a major version if attempted.

**Pull requests welcome!**

oliver-sanders:
* Closes #5005
* Follow the documented approach, use one cursor per transaction.

**Check List**

### DIFF
--- a/cylc/flow/rundb.py
+++ b/cylc/flow/rundb.py
@@ -324,7 +324,8 @@ class CylcWorkflowDAO:
         """
         self.db_file_name = expandvars(db_file_name)
         self.is_public = is_public
-        self.conn = None
+        self.con = None  # connection
+        self.cur = None  # cursor
         self.n_tries = 0
 
         self.tables = {}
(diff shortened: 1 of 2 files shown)


oliver-sanders:
No time for this right now.

Issue #3647, opened 2020-06-04
Cleaner log messages.


hjoliver:
At INFO level, we should just log the timestamped (optionally: #3646) sequence of events in the simplest way possible, unless something goes wrong. There's too much usually-irrelevant crap in there.

E.g. (with two timestamps removed):
```
INFO - [zz.1] status=running: (received)succeeded for job(01)
```
This should be something like:
```
INFO - zz.1 (job 01) succeeded
```
No need to log current status unless it is not the expected status.  

And no need to log "received" either, just modify slightly for the polled case:
```
INFO - zz.1 (job 01) polled succeeded
```

**Pull requests welcome!**

hjoliver:
With timestamps, these log lines are really verbose:
```
2020-05-29T12:24:34+12:00 INFO - [y.1] status=running: (received)succeeded at 2020-05-29T12:24:33+12:00 for job(02)
```

We could at least put the second timestamp ("recieved at"), which is normally the least relevant bit of info, at the end of the line:
```
2020-05-29T12:24:34+12:00 INFO - foo.1 (job 01) succeeded at 2020-05-29T12:24:33+12:00 
```
And maybe even omit it if it is sufficiently similar to the log timestamp (within a few seconds, say):
```
2020-05-29T12:24:34+12:00 INFO - foo.1 (job 01) succeeded
```

oliver-sanders:
We should definitely clean up log messages, the only real barrier to doing so has been the functional tests which make exact comparisons to log messages and often do nasty things like `awk` and `sed` which makes tests hard to fix.

Here are my thoughts on cleaning the log files:

### 1) State change messages

* remove time information
  * It's already in the timestamp
  * It's also recorded in the database
* Remove pervious state
  * Unless changed unexpectedly
* Remove task communication method
  * Unless state changed due to poll and `task communication method != poll`
* Use same `#NN` job number format as we use in the GUI & TUI.

```diff
- 2020-05-29T12:24:34+12:00 INFO - [foo.1] status=running: (received)succeeded at 2020-05-29T12:24:33+12:00 for job(01)
+ 2020-05-29T12:24:34+12:00 INFO - foo.1 #01 succeeded
```

### 2) Make submitting messages comprehendible

* Remove `<user>` from `user@host`
  * We will no longer support this functionality and the user that Cylc reports might not be the same as the remote user anyway (due to SSH config).

```diff
[...]

PR #5081, opened 2022-08-17
Log tidy

MetRonnie:
Partially addresses https://github.com/cylc/cylc-flow/issues/3647#issuecomment-638763568

> **1) State change messages**
> ...

Before this PR we were logging most task state changes twice:
- once for the message sent by the task's job script (which calls `cylc message` to inform the scheduler when it starts, succeeds or fails)
- once when setting the `TaskProxy` object's state to the new value

This PR:
- Demotes scheduler log severity level of `(received)started` (was `INFO`), `(received)succeeded` (was `INFO`), `(received)failed/ERR` (was `CRITICAL`) and `(internal)submission failed` (was `CRITICAL`) all to `DEBUG`
- Likewise demotes poll results from `INFO` to `DEBUG` for tasks whose platform's communication method is `poll` (but manually polling a task whose communication method is not `poll` will still show the result at `INFO` level
- Demotes a few other `INFO` level log calls to `DEBUG`

```diff
--- before
+++ after
@@ -9,21 +9,16 @@
 INFO - Final point: None
 INFO - Cold start from 1
 INFO - New flow: 1 (original flow from 1) 2022-08-17 16:30:57
-INFO - Runahead limit: 1
 INFO - [1/foo waiting(runahead) job:00 flows:1] spawned
 INFO - [1/foo waiting(runahead) job:00 flows:1] => waiting
 INFO - [1/foo waiting job:00 flows:1] => waiting(queued)
 INFO - [1/foo waiting(queued) job:00 flows:1] => waiting
 INFO - [1/foo waiting job:01 flows:1] => preparing
-INFO - [1/foo preparing job:01 flows:1] host=hosty.hostface.co.uk
-INFO - [1/foo preparing job:01 flows:1] (internal)submitted at 2022-08-17T16:33:47+01:00
 INFO - [1/foo preparing job:01 flows:1] submitted to localhost:background[64299]
 INFO - [1/foo preparing job:01 flows:1] => submitted
 INFO - [1/foo submitted job:01 flows:1] health: submission timeout=P1D, polling intervals=PT1H,...
-INFO - [1/foo submitted job:01 flows:1] (received)started at 2022-08-17T16:33:49+01:00
 INFO - [1/foo submitted job:01 flows:1] => running
 INFO - [1/foo running job:01 flows:1] health: execution timeout=None, polling intervals=PT1H,...
-INFO - [1/foo running job:01 flows:1] (received)succeeded at 2022-08-17T16:33:51+01:00
 INFO - [1/foo running job:01 flows:1] => succeeded
```
```diff
--- before
+++ after
@@ -1,12 +1,9 @@
 INFO - [command] poll_tasks
-INFO - Processing 1 queued command(s)
 INFO - Command succeeded: poll_tasks(['1/foo'])
-INFO - [1/foo submitted job:01 flows:1] (received)started at 2022-08-17T16:45:43+01:00
 INFO - [1/foo submitted job:01 flows:1] => running
 INFO - [1/foo running job:01 flows:1] health: execution timeout=None, polling intervals=PT1H,...
 INFO - [1/foo running job:01 flows:1] (polled)started at 2022-08-17T16:45:43+01:00
-CRITICAL - [1/foo running job:01 flows:1] (received)failed/ERR at 2022-08-17T16:45:46+01:00
 INFO - [1/foo running job:01 flows:1] => failed
 CRITICAL - [1/foo failed job:01 flows:1] failed
-INFO - Processing 1 queued command(s)
 INFO - Command succeeded: stop(mode=Clean)
```

**Check List**

- [x] Tests not needed
- [x] No docs needed.
- [x] master branch only

### DIFF
--- a/cylc/flow/network/resolvers.py
+++ b/cylc/flow/network/resolvers.py
@@ -28,6 +28,7 @@ from typing import (
     Dict,
     Iterable,
     List,
+    NamedTuple,
     Optional,
     Tuple,
     TYPE_CHECKING,
@@ -36,8 +37,8 @@ from typing import (
 from uuid import uuid4
 
 from graphene.utils.str_converters import to_snake_case
-from cylc.flow import LOG
 
+from cylc.flow import LOG
 from cylc.flow.data_store_mgr import (
     EDGES, FAMILY_PROXIES, TASK_PROXIES, WORKFLOW,
     DELTA_ADDED, create_delta_store
@@ -57,6 +58,14 @@ if TYPE_CHECKING:
     from cylc.flow.data_store_mgr import DataStoreMgr
     from cylc.flow.scheduler import Scheduler
     from cylc.flow.workflow_status import StopMode
+
+
+class TaskMsg(NamedTuple):
+    """Tuple for Scheduler.message_queue"""
+    job_id: Union[Tokens, str]
+    event_time: str
+    severity: Union[str, int]
+    message: str
 
 
 logger = logging.getLogger(__name__)
@@ -732,35 +741,32 @@ class Resolvers(BaseResolvers):
 
     def put_messages(
         self,
-        task_job=None,
-        event_time=None,
-        messages=None
-    ):
+        task_job: str,
+        event_time: str,
+        messages: List[list]
+    ) -> Tuple[bool, str]:
         """Put task messages in queue for processing later by the main loop.
 
         Arguments:
-            task_job (str, optional):
-                Task job in the format ``CYCLE/TASK_NAME/SUBMIT_NUM``.
-            event_time (str, optional):
+            task_job:
+                Job ID in the format ``CYCLE/TASK_NAME/SUBMIT_NUM``.
+            event_time:
                 Event time as an ISO8601 string.
-            messages (list, optional):
+            messages:
                 List in the format ``[[severity, message], ...]``.
 
         Returns:
-            tuple: (outcome, message)
-
-            outcome (bool)
+            outcome:
                 True if command successfully queued.
-            message (str)
+            message:
                 Information about outcome.
 
         """
-        #  TODO: standardise the task_job interface to one of the other
-        #        systems
         for severity, message in messages:
             self.schd.message_queue.put(
-                (task_job, event_time, severity, message))
-        return (True, 'Messages queued: %d' % len(messages))
+                TaskMsg(task_job, event_time, severity, message)
+            )
+        return (True, f'Messages queued: {len(messages)}')
 
     def set_graph_window_extent(self, n_edge_distance):
         """Set data-store graph window to new max edge distance.
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -19,7 +19,6 @@ import asyncio
 from contextlib import suppress
 from collections import deque
 from dataclasses import dataclass
-import logging
 from optparse import Values
 import os
 from pathlib import Path
@@ -82,9 +81,14 @@ from cylc.flow.loggingutil import (
 from cylc.flow.timer import Timer
 from cylc.flow.network import API
 from cylc.flow.network.authentication import key_housekeeping
+from cylc.flow.network.resolvers import TaskMsg
 from cylc.flow.network.schema import WorkflowStopMode
 from cylc.flow.network.server import WorkflowRuntimeServer
-from cylc.flow.option_parsers import verbosity_to_env, verbosity_to_opts
+from cylc.flow.option_parsers import (
+    log_level_to_verbosity,
+    verbosity_to_env,
+    verbosity_to_opts,
+)
 from cylc.flow.parsec.exceptions import ParsecError
 from cylc.flow.parsec.OrderedDict import DictTree
 from cylc.flow.parsec.validate import DurationFloat
@@ -204,7 +208,7 @@ class Scheduler:
 
     # queues
     command_queue: 'Queue[Tuple[str, tuple, dict]]'
-    message_queue: Queue
+    message_queue: 'Queue[TaskMsg]'
     ext_trigger_queue: Queue
 
     # configuration
@@ -785,24 +789,23 @@ class Scheduler:
         except (KeyError, ValueError, AttributeError):
             return
 
-    def process_queued_task_messages(self):
+    def process_queued_task_messages(self) -> None:
         """Handle incoming task messages for each task proxy."""
-        messages = {}
+        messages: Dict[str, List[Tuple[Optional[int], TaskMsg]]] = {}
         while self.message_queue.qsize():
             try:
-                task_job, event_time, severity, message = (
-                    self.message_queue.get(block=False))
+                task_msg = self.message_queue.get(block=False)
             except Empty:
                 break
             self.message_queue.task_done()
-            tokens = Tokens(task_job, relative=True)
+            tokens = Tokens(task_msg.job_id, relative=True)
             # task ID (job stripped)
             task_id = tokens.duplicate(job=None).relative_id
             messages.setdefault(task_id, [])
             # job may be None (e.g. simulation mode)
             job = int(tokens['job']) if tokens['job'] else None
             messages[task_id].append(
-                (job, event_time, severity, message)
+                (job, task_msg)
             )
         # Note on to_poll_tasks: If an incoming message is going to cause a
         # reverse change to task state, it is desirable to confirm this by
@@ -813,10 +816,11 @@ class Scheduler:
             if message_items is None:
                 continue
             should_poll = False
-            for submit_num, event_time, severity, message in message_items:
+            for submit_num, tm in message_items:
                 if self.task_events_mgr.process_message(
-                        itask, severity, message, event_time,
-                        self.task_events_mgr.FLAG_RECEIVED, submit_num):
+                    itask, tm.severity, tm.message, tm.event_time,
+                    self.task_events_mgr.FLAG_RECEIVED, submit_num
+                ):
                     should_poll = True
             if should_poll:
                 to_poll_tasks.append(itask)
@@ -841,7 +845,7 @@ class Scheduler:
         LOG.info(f"Processing {qsize} queued command(s)")
         while True:
             try:
-                command = (self.command_queue.get(False))
+                command = self.command_queue.get(False)
                 name, args, kwargs = command
             except Empty:
                 break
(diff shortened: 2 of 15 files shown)


cylc/flow/scheduler.py
```diff
@@ -843,9 +849,6 @@ def process_command_queue(self) -> None:
             try:
                 n_warnings: Optional[int] = self.get_command_method(name)(
                     *args, **kwargs)
-            except SchedulerStop:
-                LOG.info(f"Command succeeded: {cmdstr}")
-                raise
```
MetRonnie, line 848: This didn't seem possible, and indeed does not have any coverage hits

hjoliver (approved):
Looks good. I agree it's an improvement :+1:

cylc/flow/etc/job.sh
```diff
@@ -127,7 +127,7 @@ cylc__job__main() {
     export CYLC_TASK_WORK_PATH="${CYLC_TASK_WORK_DIR}"
 
     # Send task started message
-    cylc message -- "${CYLC_WORKFLOW_ID}" "${CYLC_TASK_JOB}" 'started' &
+    cylc message -- "${CYLC_WORKFLOW_ID}" "${CYLC_TASK_JOB}" 'DEBUG:started' &
```
oliver-sanders, line 130: This addresses submitted/succeeded but not failed or custom messages.

(note there is additional logging for tasks which complete with unsatisfied required outputs so we don't need to go to extra effort to draw attention to the failed state any more)

Could we change the default for all `cylc message` calls rather than hardcoding just these two?

cylc/flow/network/resolvers.py
```diff
@@ -59,6 +60,14 @@
     from cylc.flow.workflow_status import StopMode
 
 
+class TaskMsg(NamedTuple):
+    """Tuple for Scheduler.message_queue"""
+    task_job: Union[Tokens, str]
```
oliver-sanders, line 65: Is this "task or job" or "job [associated with task]"?

FYI I'd like to retire the term "task-job" in favour of either "task" or "job".

https://github.com/cylc/cylc-doc/issues/352

cylc/flow/task_pool.py
```diff
@@ -395,7 +400,7 @@ def compute_runahead(self, force=False) -> bool:
         if self.stop_point and limit_point > self.stop_point:
             limit_point = self.stop_point
             LOG.debug(f"{pre_adj_limit} -> {limit_point} (stop point)")
-        LOG.info(f"Runahead limit: {limit_point}")
```
oliver-sanders, line 398: Can we just remove this one?

The processed runahead limit can be found in the processed configuration file. I don't think this message is adding anything?

cylc/flow/scheduler.py
oliver-sanders: :+1: I think this may have been legacy from how `cylc stop` was once implemented.

oliver-sanders (commented):
:+1:

cylc/flow/network/resolvers.py
MetRonnie: Inspected the variable in the debugger and it is just the job ID e.g. `1/foo/01`. Will push a change

cylc/flow/etc/job.sh
MetRonnie: You mean change the default in `cylc/flow/scripts/message.py`?

oliver-sanders: Or wherever it's set, that way any user defined `cylc message` commands will also go to DEBUG level (unless they configure them otherwise).

MetRonnie: Keen to get @hjoliver 's thoughts on this

hjoliver: Not sure about that.

It definitely makes sense for the standard job status messages, because they should (pretty much) always be followed by a logged status change.

But part of the intention for "custom messages" was to allow arbitrary non-triggering info to be sent to the scheduler log by tasks. (Whether or not anyone uses it that way, I'm not sure).  Debug would effectively turn that off for normal use.

And I think it's OK to log task failures even when they're handled by the graph. Failure is still, by definition, the task not doing what it is primarily supposed to do.

cylc/flow/task_pool.py
hjoliver: Correct, that logs the new concrete runahead limit when it changes. I suppose that could go to DEBUG though.

cylc/flow/network/resolvers.py
```diff
@@ -762,11 +762,11 @@ def put_messages(
                 Information about outcome.
 
         """
-        #  TODO: standardise the task_job interface to one of the other
+        #  TODO: standardise the job_id interface to one of the other
         #        systems
```
MetRonnie, line 766: What does this mean/is this valid anymore?

cylc/flow/network/schema.py
```diff
@@ -1547,7 +1547,7 @@ class Meta:
 
     class Arguments:
         workflows = graphene.List(WorkflowID, required=True)
-        task_job = String(required=True)
+        job_id = String(required=True)
```
MetRonnie, line 1550: Note change in GraphQL interface, I assume this is ok since this is a minor rather than maintenance release?

MetRonnie:
(Test failures due to swarm build failure)

cylc/flow/network/schema.py
oliver-sanders: Unfortunately we do not have ANY strategy for handling GraphQL schema changes at the moment.

Because the UI Server uses a single GraphQL endpoint for all workflows the schemas must be consistent between all running workflows (or the resulting errors must be handled).

MetRonnie: Removing the schema change, only keeping the internal python variables rename...

oliver-sanders:
> (Test failures due to swarm build failure)

Dammit, should look into migrate from miniconda to miniforge.

cylc/flow/etc/job.sh
oliver-sanders: Right, wasn't aware custom task messages were considered for non-triggering purposes.

Given the verbosity of workflow logs I highly doubt anyone is really using them in this way. Worth noting that custom outputs are now visible in the GUI. Also know at least one site is using custom messages to communicate model time-steps which would bloat the log somewhat. But can leave at INFO, I guess these messages wouldn't get logged otherwise so fair enough?

For failed tasks we would still get the task status changed message (`=> failed`) message right? So I don't think we need an additional log item telling you that the failed status was sent by `cylc message` (as opposed to being polled)?

hjoliver: > Right, wasn't aware custom task messages were considered for non-triggering purposes.

Well if a message is received, we check to see if it corresponds to a task output or not. If not, log it anyway.

> For failed tasks we would still get the task status changed  ...

Agreed.

>  But can leave at INFO, I guess these messages wouldn't get logged otherwise so fair enough?

:+1: 

> Given the verbosity of workflow logs I highly doubt anyone is really using them in this way.

Yeah probably not, but there might still be potential use cases.

cylc/flow/network/resolvers.py
hjoliver: `git blame` flags @dwsutherland as the author of that comment. Do you recall what it means David?

dwsutherland: I'm afraid I don't quite remember, however, it may have been pre-universal ID.. or perhaps there was an inconsistency between mutations and hence resolvers..

Can probably remove it.

cylc/flow/etc/job.sh
oliver-sanders: > This addresses submitted/succeeded but not failed or custom messages.
>> Not sure about that.
>>> For failed tasks we would still get the task status changed ...
>>>> Agreed.

Are we agreed to remove this duplicate "failed" log entry then?

IMO each state transition should be logged once and only once as duplicate logging is making debugging quite difficult ATM.

Looks like the state transition message we are aiming to preserve are the `=> <state>` ones. So all others (excluding custom outputs) should go or be demoted to DEBUG if there's a good case to keep them.

cylc/flow/network/resolvers.py
oliver-sanders: kill it.

cylc/flow/etc/job.sh
hjoliver: Yes agreed!

MetRonnie: It seems we can't demote the "failed" `cylc message` in the job script because then the `critical handlers` don't run (also we don't get `CRITICAL - failed/ERR` printed in `job.err`)

oliver-sanders: Good point, that presumably also means the succeeded message would be missing too? Will have to handle server side rather than setting the log level explicitly in the job.sh.

MetRonnie: For
```
[runtime]
    [[foo]]
        script = echo Data
```
and `cylc play` in normal verbosity, I get this change in `job.out`:
```diff
 Data
-INFO - started
+DEBUG - started
-INFO - succeeded
+DEBUG - succeeded
```
So the succeeded message is still getting through.

> Will have to handle server side rather than setting the log level explicitly in the job.sh.

I think that would be fiddly, probably not going to attempt that in this PR

oliver-sanders: I think it's probably just adding an `if message in {'started', 'running', 'succeeded', 'failed'}: severity = logging.DEBUG` before this block:

https://github.com/cylc/cylc-flow/blob/f88f32e2bf14568664d27f9452c655baa45eb9b7/cylc/flow/task_events_mgr.py#L811-L814

cylc/flow/task_events_mgr.py
```diff
@@ -807,10 +807,13 @@ def _process_message_check(
                 )
             return False
 
-        LOG.log(
-            LOG_LEVELS.get(severity, INFO),
-            f"[{itask}] {flag}{message}{timestamp}"
-        )
+        severity_lvl = LOG_LEVELS.get(severity, DEBUG)
```
oliver-sanders, line 810: I think Hillary made a good case for the `INFO` default with custom outputs as I don't think these would get logged elsewhere.

MetRonnie: Custom outputs are still logged at `INFO` by default

oliver-sanders (approved):
:+1:

@hjoliver do you want to take a quick look at the recent changes and whack the merge button if happy?

hjoliver (approved):

PR #5082, opened 2022-08-17
Do not raise `FileNotFoundError` out of `scandir()`

MetRonnie:
This matches the previous pyuv implementation.

Fixes bug introduced by #5028 where trying to call `cylc clean` with a workflow that doesn't exist would throw traceback.

Follow-up to #5065

**Check List**

- [x] Tests are included
- [x] No changelog entry (follow-up to unreleased PR)
- [x] No docs needed
- [x] Does not need 8.0.x PR as this bug is on master only


tests/unit/test_async_util.py
```diff
@@ -238,3 +240,20 @@ async def test_asyncqgen():
...
+
+    assert sorted(await scandir(tmp_path)) == [
+        Path(tmp_path, 'a'),
+        Path(tmp_path, 'b'),
+        Path(tmp_path, 'c')
+    ]
+
+    # Should not raise if the given path doesn't exist. This matches the
+    # previous pyuv implementation.
+    assert await scandir(tmp_path / 'HORSE') == []
```
wxtim, line 259: Not willing to go too far with this one, but IMO smaller tests make debugging easier, esp when there's no reason to put them in a single place.

```suggestion
async def test_scandir_non_exist(tmp_path: Path):
    """It returns empty list if target dir doesn't exist

    This matches the previous pyuv implementation.
    """
    assert await scandir(tmp_path / 'HORSE') == []
```

wxtim (approved):
- [x] Tests fail as expected on master
- [x] Read code changes
- [x] Run tests
- [x] Played with code locally.

👍🏼 

Made a comment, but it's only a suggestion.

oliver-sanders:
I think it makes sense for `scandir` to raise `FileNotFoundError` for a directory which doesn't exist.

Could this error be caught in `cylc clean`?

MetRonnie:
It could be, but then we're catching it every time we use it. It looks like the scan apparatus was designed around pyuv's behaviour of not raising?

oliver-sanders:
The error is now caught for all sub-directory scanning:

https://github.com/cylc/cylc-flow/blob/312a9de2aa6be8f23702e307050d6e18255e9f23/cylc/flow/network/scan.py#L256-L260

So this issue can only occur if the top-level directory (`~/cylc-run` by default) is missing:

https://github.com/cylc/cylc-flow/blob/312a9de2aa6be8f23702e307050d6e18255e9f23/cylc/flow/network/scan.py#L241

MetRonnie:
superseded by #5087

PR #5087, opened 2022-08-18
Fix `FileNotFoundError` when scanning path that doesn't exist

MetRonnie:
Fixes bug introduced by #5028 where trying to call `cylc clean` with a workflow that doesn't exist would throw traceback. Also running `cylc scan` if you didn't have a `~/cylc-run` dir would also throw traceback.
 
Follow-up to #5065, supersedes #5082

**Check List**

- [x] Tests are included
- [x] No changelog entry (follow-up to unreleased PR)
- [x] No docs needed
- [x] Does not need 8.0.x PR as this bug is on master only


oliver-sanders (approved):
Cheers.

hjoliver (approved):

PR #5090, opened 2022-08-22
No-cycle start-up and shut-down graphs.

### LINKED ISSUE #4912: Variable runahead limit, for start-up graphs.

hjoliver:
Close #4912 
Supersede #5036
(Really addresses the original #4903 but specifically just for start-up and shut-down graphs)

**Check List**


hjoliver:
[POC works for a test case with a start-up and shut-down graph, but a lot of details need attending to, and not surprisingly it breaks a bunch of other stuff - ignore while Draft]

hjoliver:
```ini
[scheduling]
    cycling mode = integer 
    initial cycle point = 1
    final cycle point = 2
    [[graph]]
        startup = "prep1 => prep2"
        P1 = "foo => bar"
        shutdown = "clean1 => clean2"
[runtime]
    [[prep1, prep2]]
    [[foo, bar]]
    [[clean1, clean2]]
```
Result:
```console
$ cylc cat-log test | grep -E '=> (running|succeeded)'
[startup/prep1 submitted job:01 flows:1] => running
[startup/prep1 running job:01 flows:1] => succeeded
[startup/prep2 submitted job:01 flows:1] => running
[startup/prep2 running job:01 flows:1] => succeeded
[1/foo submitted job:01 flows:2] => running
[2/foo submitted job:01 flows:2] => running
[1/foo running job:01 flows:2] => succeeded
[2/foo running job:01 flows:2] => succeeded
[1/bar submitted job:01 flows:2] => running
[1/bar running job:01 flows:2] => succeeded
[2/bar submitted job:01 flows:2] => running
[2/bar running job:01 flows:2] => succeeded
[shutdown/clean1 submitted job:01 flows:3] => running
[shutdown/clean1 running job:01 flows:3] => succeeded
[shutdown/clean2 submitted job:01 flows:3] => running
[shutdown/clean2 running job:01 flows:3] => succeeded
```

hjoliver:
I'm tentatively putting this one back to 8.1.0.
- It's ready to go and shouldn't be hard to review (and the underlying principles were discussed at length and agreed)
- It's a nice feature that many users will like (it makes special behaviour at start much easier to understand and configure)
- It has some nasty conflict potential if left too long 

(The only controversy, I think, may be what to call the two isolated graph sections).

hjoliver:
Merged master and deconflicted.

hjoliver:
@dpmatthews - pinging you for a big-picture review (given your involvement in the discussions that led to this).

hjoliver:
Squashed and rebased.

hjoliver:
(Rebased)

hjoliver:
(Rebased)

hjoliver:
(Rebased)

dwsutherland:
I'm annoyed I didn't review this one earlier.. A welcome feature for NIWA Ops.

oliver-sanders:
8.4.0 is intended to be a small release filling in the remaining intervention holes:

* Cylc remove proposal.
* Skip mode proposal.

This is a large low-priority feature that requires some thought.

Cheekily bumping this back until we are able to give it the time it requires.

hjoliver:
I agree the new intervention stuff is higher priority. However, I'd really like to prioritize this soon after.

This has been requested by many users for a long time. It looks "bigger" than it really is, in that it doesn't mess with the nasty bits of scheduler internals. And, we did discuss and agree it several years back.

hjoliver:
(Updated from master, which was pretty gnarly after all the changes since last update, couple more broken tests to deal with...)

hjoliver:
(Back on track now. Once this hits the priority list I'll up coverage and document it).

hjoliver:
(squashed, rebased, and deconflicted)

dwsutherland:
Having a little play:
```
[meta]
    title = "alpha & omega graph sections"
[scheduler]
    UTC mode = True
    allow implicit tasks = True
[scheduling]
    initial cycle point = 20250301T00
    final cycle point = 20250310T00
    [[graph]]
        alpha = "prep => unpack"
        PT12H = "foo[-PT12H] => foo => bar"
        omega = "pack => unprep"
[runtime]
    [[root]]
        script = sleep $((3 + $RANDOM % 7))
    [[prep, unprep]]
        script = sleep 30
```
Nice that the UI doesn't care what a cycle looks like:
image


Might want to gracefully handle the mixing of tasks in different sections, i.e. if I change `alpha` to contain `foo`
```
        alpha = "prep => unpack => foo"
```
```
    packages/metomi/isodatetime/parsers.py", line 435, in get_info
        keys, date_info = self.get_date_info(date)
      File "/home/sutherlander/.envs/flow/lib/python3.10/site-
    packages/metomi/isodatetime/parsers.py", line 395, in get_date_info
        raise ISO8601SyntaxError("date", date_string)
    metomi.isodatetime.exceptions.ISO8601SyntaxError: Invalid ISO 8601 date
    representation: alpha
CRITICAL - Workflow shutting down - Invalid ISO 8601 date representation:
    alpha
```
(I guess there just needs to be a validation catch)

dwsutherland:
Also, as expected, if you use clock triggers in this section:
```
    [[xtriggers]]
        clock = wall_clock(offset=PT5M)
    [[graph]]
        alpha = """
@clock => prep
prep => unpack
"""
```
```
    packages/metomi/isodatetime/parsers.py", line 395, in get_date_info
        raise ISO8601SyntaxError("date", date_string)
    metomi.isodatetime.exceptions.ISO8601SyntaxError: Invalid ISO 8601 date representation: alpha
CRITICAL - Workflow shutting down - Invalid ISO 8601 date representation: alpha
```

let's see what happens if I use workflow_state:
```
    [[xtriggers]]
        poll_prep = workflow_state(%(workflow)s//%(point)s/prep)
    [[graph]]
        alpha = """
prep
@poll_prep => unpack
"""
```
image

dwsutherland:
> Are these tasks even recorded in the task table, and with alpha as the cycle point?

Well stop/restart works, so must be

dwsutherland:
Should `alpha` be translated under the hood to ICP?
I'm trying to think of cold-start tasks that need to work with cycle offsets to setup data in `share/cycle`...
Maybe the tasks will just have to be aware of the ICP via a env var (if not already)

hjoliver:
> Should alpha be translated under the hood to ICP?

Probably not. The `alpha` cycle is appropriate for stuff that doesn't care about the cycle point value, such as workspace setup and code deployment tasks (but they still have access to the ICP value via `$CYLC_WORKFLOW_INITIAL_CYCLE_POINT` if they need it). 

The main cycling section still has an ICP for all the usual purposes.

hjoliver:
(rebased onto current master)

hjoliver:
Note on graph names: we need to avoid confusion with the existing concepts of:
- initial cycle point
- start cycle point
- final cycle point
- stop cycle point

However, on coming back to this, "alpha" and "omega" possibly sound a bit too quirky, and it shoudln't be confusing if we consistently at "up" and "down" as in:
- startup graph
- shutdown graph

hjoliver:
I initially called the new isolated graphs "alpha" and "omega" to avoid any confusion with the concepts of initial, start, final, stop, cycle points. However, that sounds a bit too quirky and I've switch over to **"startup graph"** and **"shutdown graph"**, which is descriptive, and it shouldn't be confusing if we consistently use "startup" and "shutdown" (not "start" and "stop"). ... updated issue description.

oliver-sanders:
> I don't want to throw a curve-ball in

Here's a curveball loosely based on a real example:

```cylc
[scheduling]
    [[graph]]
        R1 = install => spinup => run => report

        [[[install]]]
            R1 = install<model>

        [[[spinup]]]
            cycling mode = integer
            initial cycle point = 1
            final cycle point = {{ SPINUP_LENGTH }}
            R1 = spinup<model>_cold => spinup<model>
            P1 = spinup<model> => spinup<model + 1>
            R1/$ = spinup<model> => make_bc

        [[[run]]]
            initial cycle point = {{ START_CYCLE }}
            final cycle point = {{ END_CYCLE }}
            P1D = foo & bar => run<model> => baz

        [[[report]]]
            R1 = pub => qux => report
```

hjoliver:
I quite like David's suggestion, but the problem with it is it *suggests* that non-R1 (i.e., cycling) is possible in the startup and shutdown graphs.

(That is perfectly possible and not much of a change to this PR, but I either decided or was persuaded against it two years ago ... I don't recall which at this point).

oliver-sanders:
(Note, my example above also supported both integer and datetime cycling)

hjoliver:
@oliver-sanders -

> I'm not sure I'm following on the "priority" options above? Is that an alternative way of specifying the order in which the graph sections run, or a way of controlling the order if multiple graph sections were permitted to run in parallel?

I think it's the latter, as one important use case for this is retriggering code deployment tasks (in the startup graph) mid-run. 

So for manual triggering purposes I guess we have to be able to:
 - manage tasks from multiple separate graphs at once in the same task pool
 - or manage multiple separate task pools at once
 - or suspend the current task pool, replace it with another, then restore it again after

The same-task-pool model works fine on this branch, but only because the startup and shutdown graphs are non-cycling, and I have to carefully exclude those special cycle points from comparison operations and the like (which is not great).

dwsutherland:
> I'm not sure I'm following on the "priority" options above? Is that an alternative way of specifying the order in which the graph sections run, or a way of controlling the order if multiple graph sections were permitted to run in parallel?

Yes, both really, this was just a way of implementing the section run order, i.e.:
```
[scheduling]
    R1 = "deploy => graph => final"
    [[deploy]]
        R1 = "a => b"
    [[graph]]
        R1 = "foo => bar"
    [[final]]
        R1/$ = "x => y"
```
(the under the hood priority would be 1, 2, 3 in this case)

I'd support bespoke graph section, and freeform cycling..
However, as a first implementation we just need to agree on the config structure/syntax.. 
Then future improvements like local cycling modes and ICP/FCP can be implemented on top.. i.e.
```
[scheduling]
    R1 = "deploy => spinup => graph => final"
    [[deploy]]
        R1 = "a => b"
    [[spinup]]
        cycling mode = integer
        initial cycle point = 1
        final cycle point = {{ SPINUP_LENGTH }}
        R1 = spinup<model>_cold => spinup<model>
        P1 = spinup<model> => spinup<model + 1>
    [[graph]]
        R1 = "foo => bar"
    [[final]]
        R1/$ = "x => y"
```
So the rest of the features will be mostly backend implementation..

The implementation I proposed (using existing pool) will allow us to get underway quickly, because we can then re-architect under the hood are release features piece wise after.

> I think it's the latter, as one important use case for this is retriggering code deployment tasks (in the startup graph) mid-run.

Exactly, it's really a requirement that we're able to rerun these sections and have downstream/lower-priority section essentially freeze while it runs.

> The runahead limit itself would almost do the job

Well we could just use the queuing system, so any actively waiting tasks get held even if they are released.
(over and above the existing queue)


> How should we expose these graph sections to the CLI / GUI?

Perhaps - with only one graph section:
```
1
    . . .  
2
    . . .
```
with multiple:
```
startup
    1
        . . .
run
    1
        . . .  
    2
        . . .
final
    2
        . . .
```

PR #5091, opened 2022-08-22
Fix tutorial workflows

MetRonnie:
- Map HTML file was broken due to the Jinja2 (?) double curly brackets being broken over lines
- `urllib3` requests no longer working with Met Office proxy, used `requests` instead, which is nicer actually

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Dependency changes only apply to `setup.cfg`, not `conda-environment.yml` (optional dependency `urllib3` removed)
- [x] No tests needed.
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [x] No docs needed
- [ ] If this is a bug fix, PRs raised to both master and the relevant maintenance branch.

PR #5100, opened 2022-08-25
Fix tutorial workflows (#5091)

oliver-sanders:
* Fix broken templates in tutorial map file
* Switch from `urllib3` to `requests` for tutorials


oliver-sanders:
Required deconflicting so needed a branch-of-a-branch, re raised.

PR #5104, opened 2022-08-26
Fix post-reload trigger.

### LINKED ISSUE #5102: Reload broken in compatibility mode
With this suite.rc:
```
[scheduling]
    [[graph]]
        R1 = "foo => bar"
[runtime]
    [[foo]]
        script = "echo Hello; false"
    [[bar]]
```
Play the workflow and let it stall.
Then reload.
Then trigger `foo`.
I get the following in the scheduler log:
```
2022-08-25T18:57:43+01:00 INFO - Processing 1 queued command(s)
2022-08-25T18:57:43+01:00 INFO - [1/foo failed job:01 flows:1] => waiting
2022-08-25T18:57:43+01:00 INFO - Command succeeded: force_trigger_tasks(['1/foo'], flow=['all'], flow_wait=False, flow_descr=None)
2022-08-25T18:57:43+01:00 INFO - [1/foo waiting job:02 flows:1] => preparing
2022-08-25T18:57:43+01:00 INFO - [1/foo preparing job:02 flows:1] host=XXX
2022-08-25T18:57:43+01:00 WARNING - stall timer stopped
2022-08-25T18:57:44+01:00 WARNING - Unhandled jobs-submit output: 2022-08-25T18:57:44+01:00|1/foo/02|0|78545
2022-08-25T18:57:44+01:00 WARNING - ('1', 'foo', '02')
2022-08-25T18:57:44+01:00 WARNING - Unhandled jobs-submit output: 2022-08-25T18:57:44+01:00|1/foo/02|[STDOUT] 78545
2022-08-25T18:57:44+01:00 WARNING - ('1', 'foo', '02')
2022-08-25T18:57:44+01:00 ERROR - [jobs-submit cmd] cylc jobs-submit --clean-env --path=/bin --path=/usr/bin --path=/usr/local/bin --path=/sbin --path=/usr/sbin --path=/usr/local/sbin -- '$HOME/cylc-run/test/reload/run1/log/job' 1/foo/02
    [jobs-submit ret_code] 1
    [jobs-submit out] 2022-08-25T18:57:44+01:00|1/foo/01|1
2022-08-25T18:57:44+01:00 INFO - [1/foo failed job:01 flows:1] (internal)submission failed at 2022-08-25T18:57:44+01:00
2022-08-25T18:57:44+01:00 CRITICAL - [1/foo failed job:01 flows:1] submission failed
2022-08-25T18:57:44+01:00 INFO - [1/foo failed job:01 flows:1] => submit-failed
2022-08-25T18:57:46+01:00 INFO - [1/foo preparing job:02 flows:1] (received)started at 2022-08-25T18:57:45+01:00
```
Something is going badly wrong with the the job submission.
Despite the warnings, job 02 does get submitted and runs.
However, if I modify the task config and reinstall before doing the reload, job 02 does not use the updated config so it's very broken.

The problem does not occur if I:
1. rename `suite.rc` to `flow.cylc`
2. Change the graph to ` R1 = "foo"`

hjoliver:
Close #5102  - retriggering a failed task after reload should not cause weird job submission errors.

**Check List**


hjoliver:
Works fine for the test case in back-compat and normal mode.  I'm not sure why it doesn't affect both modes, on master.

MetRonnie (commented):
This does not currently fix the problem of the second submitted job having the old pre-reload config

cylc/flow/task_job_mgr.py
```diff
@@ -767,7 +767,10 @@ def _manip_task_jobs_callback(
         tasks = {}
         for itask in itasks:
             while itask.reload_successor is not None:
```
MetRonnie, line 769: At this point:
- `suite.rc`: `itask.reload_successor` is a `TaskProxy` object
- `flow.cylc`: `itask.reload_successor` is `None`

hjoliver:
> This does not currently fix the problem of the second submitted job having the old pre-reload config

Sorry, there was a second small change I'd neglected to commit and push to GH.  It fixes what I think is probably the bug causing mysterious reload issues. 

(Disclaimer: not extensively tested beyond the original example yet).

cylc/flow/task_job_mgr.py
hjoliver: I'll look at this more closely tomorrow, post webinar. But not that back-compat mode does have different task pool content: the downstream task in the example gets pre-spawned as waiting.

cylc/flow/task_pool.py
```diff
@@ -153,10 +153,12 @@ def stop_task_done(self):
...
     def _swap_out(self, itask):
         """Swap old task for new, during reload."""
-        if itask.point in self.hidden_pool:
-            if itask.identity in self.hidden_pool[itask.point]:
-                self.hidden_pool[itask.point][itask.identity] = itask
-                self.hidden_pool_changed = True
+        if (
+            itask.point in self.hidden_pool
+            and itask.identity in self.hidden_pool[itask.point]
+        ):
```
MetRonnie, line 159: Could reduce this to one expression (and same for the main pool `elif` branch)
```
if itask.identity in self.hidden_pool.get(itask.point, set()):
```

MetRonnie (commented):
Ok, have confirmed Dave's example now works as expected

hjoliver:
@oliver-sanders - I think @MetRonnie is away - can you review or reassign, and consider getting this into 8.0.2?

tests/functional/reload/27-stall-retrigger.t
```diff
@@ -0,0 +1,26 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+
+# Test retriggering a failed task after fixing the bug and reloading.
+# It should run correctly with the updated settings.
```
MetRonnie, line 20: This test passes for me on 8.0.x

oliver-sanders:
I'm worried that this big will keep resurfacing in different places since it's already cropped up a few times.

I took a look into re-implementing reload to update the tdef and modify the TaskProxy in place. This seems promising, however, is risky, too risky for 8.0.2, requires more thought.

I think this fix makes sense for now, from Ronnie's comment it looks like the test is not capturing the issue though?

tests/functional/reload/27-stall-retrigger.t
hjoliver: Huh, definitely tried that myself at the time, taking a look ...

hjoliver:
> I'm worried that this big will keep resurfacing in different places since it's already cropped up a few times.

It's possible that other bugs will crop up that affect reload, of course. But this PR definitely fixes this particular bug.

cylc/flow/task_job_mgr.py
hjoliver: (This is all fine, a consequence of the aforementioned spawning difference).

tests/functional/reload/27-stall-retrigger/bin/stall-handler.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Change "script = false" -> "true" in 1/foo, then reload and retrigger it.
+
+if grep "\[command\] reload_workflow" "${CYLC_WORKFLOW_LOG_DIR}/log" >/dev/null; then
+    # Abort if not the first call (avoid an endless loop if the reload does not
+    # have the intended effect).
+    cylc stop --now --now "${CYLC_WORKFLOW_ID}"
+    exit 1
```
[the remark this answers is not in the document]
hjoliver: Done.

MetRonnie (approved):

oliver-sanders:
> It's possible that other bugs will crop up that affect reload, of course. But this PR definitely fixes this particular bug.

FYI the source of this bug is exactly the same as for previous reload fixes (i.e. two copies of the task, one pre and one post reload, one modified, one not) - https://github.com/cylc/cylc-flow/issues/5102#issuecomment-1228453991. This issue has already surfaced and been fixed in a couple of other places, my concern is that it could continue to crop up in more contexts.

For 8.1.0 we should consider changing the reload implementation to modify the TaskProxy in place which would resolve the source of this issue.

oliver-sanders (approved):

hjoliver:
> For 8.1.0 we should consider changing the reload implementation to modify the TaskProxy in place which would resolve the source of this issue.

Yes, point taken, but this PR fixes what is the fundamental bug in the current implementation: it sometimes caused the old task proxy not to be swapped out for the new one. 

If we get a chance to reimplement for 8.1.0, good - otherwise this'll have to do.

I need to remind myself of why it is not done in place ...

oliver-sanders:
From experimentation I can't see any reason for it not to be done in place, it's actually simpler.

The main thing to watch out for is cases where derived values are computed from the `TaskDef` and ensure they are updated by the reload.

PR #5110, opened 2022-08-30
Fix repeated stall when reloading a stalled workflow

### LINKED ISSUE #5103: Reload causes repeated stall events
With this `flow.cylc`:
```
[scheduler]
    [[events]]
        mail events = stall
[scheduling]
    [[graph]]
        R1 = "foo"
[runtime]
    [[foo]]
        script = "false"
```
Play the workflow and let it stall.
Then reload.
I get the following in the scheduler log:
```
2022-08-25T19:48:52+01:00 INFO - [command] reload_workflow
2022-08-25T19:48:52+01:00 INFO - Processing 1 queued command(s)
2022-08-25T19:48:52+01:00 INFO - Reloading the workflow definition.
2022-08-25T19:48:52+01:00 INFO - LOADING workflow parameters
2022-08-25T19:48:52+01:00 INFO - + workflow UUID = 562aeff9-ec05-478d-86ed-6bc9dba43b68
2022-08-25T19:48:52+01:00 INFO - + UTC mode = False
2022-08-25T19:48:52+01:00 INFO - + cycle point time zone = Z
2022-08-25T19:48:52+01:00 INFO - Command succeeded: reload_workflow()
2022-08-25T19:48:52+01:00 WARNING - stall timer stopped
2022-08-25T19:48:53+01:00 WARNING - Incomplete tasks:
2022-08-25T19:48:53+01:00 CRITICAL - Workflow stalled
2022-08-25T19:48:53+01:00 WARNING - P3D stall timer starts NOW
2022-08-25T19:48:53+01:00 INFO - Reloading task definitions.
2022-08-25T19:48:53+01:00 INFO - [1/foo failed job:01 flows:1] reloaded task definition
2022-08-25T19:48:53+01:00 INFO - LOADING job data
2022-08-25T19:48:53+01:00 INFO - Reload completed.
2022-08-25T19:48:53+01:00 INFO - Runahead limit: 1
2022-08-25T19:48:53+01:00 WARNING - stall timer stopped
2022-08-25T19:48:53+01:00 WARNING - Incomplete tasks:
2022-08-25T19:48:53+01:00 CRITICAL - Workflow stalled
2022-08-25T19:48:53+01:00 WARNING - P3D stall timer starts NOW
2022-08-25T19:48:53+01:00 INFO - [('workflow-event-handler', 'stall') cmd]
    mail -s ...
2022-08-25T19:48:54+01:00 INFO - [('workflow-event-handler', 'stall') cmd]
    mail -s ....
```
The reload causes lots of repeated warnings.
It also causes 2 stalled events, each of which triggers the stall event handler.
If a workflow has stalled, I don't think a reload should trigger another stall event.
It certainly shouldn't trigger 2!

MetRonnie:
These changes close #5103

Reloading a stalled workflow was causing the workflow to stall again (twice). This PR prevents reloads from unstalling the workflow, so the subsequent stalls do not happen

**Check List**

- [x] No dependency changes
- [x] Tests are included
- [x] No docs needed

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1036,6 +1036,7 @@ class Scheduler:
         self.workflow_db_mgr.put_runtime_inheritance(self.config)
         self.workflow_db_mgr.put_workflow_params(self)
         self.is_updated = True
+        self.is_reloaded = True
 
     def get_restart_num(self) -> int:
         """Return the number of the restart, else 0 if not a restart.
@@ -1598,8 +1599,8 @@ class Scheduler:
             # Update state summary, database, and uifeed
             self.workflow_db_mgr.put_task_event_timers(self.task_events_mgr)
             has_updated = await self.update_data_structure()
-            if has_updated:
-                # Workflow can't be stalled, so stop the stalled timer.
+            if has_updated and not self.is_stalled:
+                # Stop the stalled timer.
                 with suppress(KeyError):
                     self.timers[self.EVENT_STALL_TIMEOUT].stop()
 
@@ -1655,11 +1656,11 @@ class Scheduler:
         updated_tasks = [
             t for t in self.pool.get_tasks() if t.state.is_updated]
         has_updated = self.is_updated or updated_tasks
+        reloaded = self.is_reloaded
         # Add tasks that have moved moved from runahead to live pool.
         if has_updated or self.data_store_mgr.updates_pending:
             # Collect/apply data store updates/deltas
-            self.data_store_mgr.update_data_structure(
-                reloaded=self.is_reloaded)
+            self.data_store_mgr.update_data_structure(reloaded=reloaded)
             self.is_reloaded = False
             # Publish updates:
             if self.data_store_mgr.publish_pending:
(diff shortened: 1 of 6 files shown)


cylc/flow/scheduler.py
```diff
@@ -1500,6 +1512,8 @@ async def main_loop(self):
             # Useful for debugging core scheduler issues:
             # self.pool.log_task_pool(logging.CRITICAL)
 
+            self.process_command_queue()
```
oliver-sanders, line 1515: I'm not sure about moving this before the reload as I think it could cause commands to be run out of order.

E.G. Consider this sequence of events

```bash
cylc reload x
cylc trigger x//y/z
```

I think before the trigger would happen *after* the reload but after the change it would happen *before* it?

MetRonnie: Why do we process the command queue twice in 1 iteration?

hjoliver: Hmm, I don't recall if there is a good reason for that.

oliver-sanders: There will be a reason for this, buried long ago in the evolution of the main loop. The order of events in the main loop is quite sensitive and can impact stability and untested functionality. I wouldn't spend time trying to work out why it's the way it is just now, we will re-jig this when we make Cylc commands asynchronous (which would allow the CLI to wait for command completion before exiting).

oliver-sanders (approved):

hjoliver (approved):
:+1:

PR #5111, opened 2022-08-31
Fix assorted cylc review bugs

### LINKED ISSUE #5108: `cylc review` - Cylc 8 Support
**Describe the bug**
rose-suite.conf doesn't appear to be accessible in either case

The Rose Files Menu is totally disabled for Cylc 8 - I don't remember doing this deliberately, but I'm not sure it's a bad idea. It looks like I have enabled the viewing of some rose files (opt/rose-suite-cylc-install.conf) from the Cylc Files menu.

Cylc install allows you to install a workflow and as far as I can tell the lack of tasks causes a failure when you go to the task list page

**Pull requests welcome!**

wxtim:
Closes #5108 

@datamel tagged for code review.
@dpmatthews tagged for functional review as originator of the bug report.

### Broadcast States pages

Installing a workflow (Cylc 7 or 8) without starting it causes no `broadcast_events` or `broadcast_states` tables to be created in the suite database, but I don't see why this should lead to error messages in the table. I suspect it's not come up before because I doubt anyone has run `rose suite-run -i` and then moved straight to Cylc Review.

### Show `rose-suite.conf` in the log files menu

The log files menus have become a touch confused. To keep things simple I have set up Cylc review to put all log items in the "cylc" menu if it's a Cylc 8 workflow, but to maintain existing behavior otherwise. If the reviewer insists I'll create a new issue to tidy up the menu items. One could even split up the different log directories into their own menus quite easily. Now the logic is there to distinguish Cylc 8 workflows we can do what we like.

**Check List**

- [x] Testing for Cylc Review is manual because the product is deprecated.

### DIFF
--- a/lib/cylc/review.py
+++ b/lib/cylc/review.py
@@ -744,24 +744,29 @@ class CylcReviewService(object):
         data = {"files": {}}
         user_suite_dir = self._get_user_suite_dir(user, suite)  # cylc files
 
+        rose_logs_dest = "rose"
+        if self.suite_dao.is_cylc8(user, suite):
+            rose_logs_dest = "cylc"
+
         # Rose files: to recognise & group, but not process, standard formats
-        data["files"]["rose"] = {}
+        data["files"][rose_logs_dest] = {}
 
         # Rosie suite info
-        info_name = os.path.join(user_suite_dir, "rose-suite.info")
-        if os.path.isfile(info_name):
-            stat = os.stat(info_name)
-            data["files"]["rose"]["rose-suite.info"] = {
-                "path": "rose-suite.info",
-                "mtime": stat.st_mtime,
-                "size": stat.st_size}
+        for fname in ['rose-suite.info', 'rose-suite.conf']:
+            info_name = os.path.join(user_suite_dir, fname)
+            if os.path.isfile(info_name):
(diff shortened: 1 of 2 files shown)


lib/cylc/review.py
```diff
@@ -704,6 +706,7 @@ def view(self, user, suite, path, path_in_tar=None, mode=None,
         # Log files with +TZ in name end up with space instead of plus sign, so
         # put plus sign back in (https://github.com/cylc/cylc-flow/issues/4260)
         path = re.sub(r"(log\.\S+\d{2})\s(\d{2,4})$", r"\1+\2", path)
+        path = re.sub(r"(log\/config\/\d*T?\d*)\s(\d*-rose-suite.conf)", r"\1+\2", path)
```
wxtim, line 710: Adds + back into url - cherrypy apparently drops them - but I can't shed any light on the reasons.

dpmatthews: I think this explains it: https://stackoverflow.com/questions/14307387/navigate-to-file-with-a-plus-in-the-url

dpmatthews (approved):
I've confirmed this addresses the issues

datamel (approved):
I've read the code and am happy with the changes. Thanks @wxtim.

PR #5114, opened 2022-09-02
Revert order of scheduler `configure()` & `log_start()`

### LINKED ISSUE #4982: Invalid workflows fail post-detach
When `[scheduler][run hosts]` are used to run scheduler on, validation errors occur post-detatch which makes the failure invisible.

E.G. when you run an invalid workflow:

```console
$ cylc play <id>
...
<id>: <server> PID=<pid>
```

It looks like it worked, but it didn't:

```console
$ cylc cat-log <id>
...
ERROR - Workflow shutting down - WorkflowConfigError: This workflow requires an initial cycle point.
```

This didn't use to be the case, the change was likely caused by recent startup reordering.

Note: #3896 will reduce the pressure on this, however, ideally we would fix this in the `cylc play` command properly.

**Pull requests welcome!**

MetRonnie:
Closes #4982, follow-up to #4836

Fix bug where a validation error on workflow startup was not being printed before demonization (aka detach), which made it look like nothing was wrong. Also the return code was zero instead of non-zero as it should be.

This partially reverts some of the logging re-ordering from #4836

**Check List**

- [x] No dependency changes
- [x] Tests are included
- [x] No docs needed

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,6 +44,10 @@ taskdefs removed before restart.
 [#5091](https://github.com/cylc/cylc-flow/pull/5091) - Fix problems with
 tutorial workflows.
 
+[#5114](https://github.com/cylc/cylc-flow/pull/5114) - Fix bug where
+validation errors during workflow startup were not printed to stderr before
+daemonisation.
+
 -------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


datamel (approved):
I've checked this out and have tested that it solves the issue and double checked the logging is still correct, thanks @MetRonnie.

hjoliver (approved):
Looks good. I like `patch_log_level`

PR #5125, opened 2022-09-08
reinstall changes to `rose-suite.conf`

wxtim:
Fixes two separate, but related bugs (one of which was partially protecting us from the other, except when using rose stem):

## 1. Rsync Output
`cylc reinstall` relied on the rsync stdout being empty to choose not to reinstall. Under some circumstances messages such as `cannot delete dirctory: opt` were causing re-installations unnecessarily.

## 2. `rose-suite.conf` reinstallation
Under many circumstances the first bug was allowing re-installation to be triggered when only the `rose-suite.conf` had changed, despite `rose-suite.conf` being on the list of files excluded from the rsync.

With the first bug fixed users could change the source `rose-suite.conf` and have Cylc tell them that nothing required re-installing.

**Check List**

- [x] Tests will be in a Cylc Rose PR https://github.com/cylc/cylc-rose/pull/178

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,9 @@ Maintenance release.
 
 ### Fixes
 
+[#5125](https://github.com/cylc/cylc-flow/pull/5125) - Allow rose-suite.conf
+changes to be considered by ``cylc reinstall``.
+
 [#5023](https://github.com/cylc/cylc-flow/pull/5023) - tasks force-triggered
 after a shutdown was ordered should submit to run immediately on restart.
 
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -1464,7 +1464,8 @@ def get_rsync_rund_cmd(src, dst, reinstall=False, dry_run=False):
         rsync_cmd.append("--dry-run")
     if reinstall:
         rsync_cmd.append('--delete')
-    for exclude in [
+
+    exclusions = [
         '.git',
         '.svn',
         '.cylcignore',
@@ -1475,7 +1476,15 @@ def get_rsync_rund_cmd(src, dst, reinstall=False, dry_run=False):
         WorkflowFiles.SHARE_DIR,
         WorkflowFiles.Install.DIRNAME,
         WorkflowFiles.Service.DIRNAME
-    ]:
+    ]
+
+    # This is a hack to make sure that changes to rose-suite.conf
+    # are considered when re-installing.
+    # It should be removed after https://github.com/cylc/cylc-rose/issues/149
+    if not dry_run:
+        exclusions.append('rose-suite.conf')
+
+    for exclude in exclusions:
         if (Path(src).joinpath(exclude).exists() or
                 Path(dst).joinpath(exclude).exists()):
             rsync_cmd.append(f"--exclude={exclude}")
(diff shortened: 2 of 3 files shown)


cylc/flow/workflow_files.py
```diff
@@ -1525,12 +1533,18 @@ def reinstall_workflow(
...
+    # Run rsync command:
     reinstall_log.info(cli_format(rsync_cmd))
     LOG.debug(cli_format(rsync_cmd))
     proc = Popen(rsync_cmd, stdout=PIPE, stderr=PIPE, text=True)  # nosec
     # * command is constructed via internal interface
     stdout, stderr = proc.communicate()
-    stdout = stdout.strip()
+
+    # Strip output of marker.
+    stdout = stdout = '\n'.join(re.findall(r'\+\+\+(.*)\+\+\+', stdout))
```
oliver-sanders, line 1547:
```suggestion
    stdout = '\n'.join(re.findall(r'\+\+\+(.*)\+\+\+', stdout))
```

cylc/flow/workflow_files.py
```diff
@@ -1525,12 +1533,18 @@ def reinstall_workflow(
...
+
+    # Run rsync command:
     reinstall_log.info(cli_format(rsync_cmd))
     LOG.debug(cli_format(rsync_cmd))
     proc = Popen(rsync_cmd, stdout=PIPE, stderr=PIPE, text=True)  # nosec
     # * command is constructed via internal interface
     stdout, stderr = proc.communicate()
-    stdout = stdout.strip()
+
+    # Strip output of marker.
```
oliver-sanders, line 1546:
```suggestion
    # Strip unwanted output.
```

tests/integration/test_reinstall.py
```diff
@@ -244,7 +244,7 @@ def test_rose_warning(one_src, one_run, capsys, interactive, monkeypatch):
 
     # reinstall (no rose-suite.conf file)
     reinstall_cli(opts=ReInstallOptions(), args=one_run.id)
-    assert rose_message not in capsys.readouterr().err
+    assert rose_message not in capsys.readouterr().out
```
oliver-sanders, line 247: This was correct before, it is tested in line 252.

oliver-sanders:
Couple of small comments, otherwise LGTM, tested as working.

tests/integration/test_reinstall.py
wxtim:
```suggestion
    assert rose_message not in capsys.readouterr().err
```

datamel (approved):
Tested as working. A couple of observations which may be irrelevant...

- upon reinstalling a workflow that has had no changes, it thinks the `rose-suite.conf` has changed. This is due to the difference in the added  `# Config Options ' (cylc-install)' from CLI appended to options already in `rose-suite.conf`.. opts=(cylc-install))`. Also, do we need to adapt this to note that it is reinstall rather than install? 
- Cylc (re)install does not copy a nested `rose-suite.conf` file. I don't think this is an issue for now, since support for sub-workflows is limited (I believe best practice has not been agreed), although this may need consideration in the future.

oliver-sanders:
> Cylc (re)install does not copy a nested rose-suite.conf file. I don't think this is an issue for now

ATM there is no support for nesting so this is ok.

CHANGES.md
```diff
@@ -48,6 +48,9 @@ Maintenance release.
 
 ### Fixes
 
+[#5125](https://github.com/cylc/cylc-flow/pull/5125) - Allow rose-suite.conf
+changes to be considered by ``cylc reinstall``.
+
```
MetRonnie, line 53: Now need to move to 8.0.3

oliver-sanders:
@wxtim could you rebase onto 8.0.x and fix the changelog.

oliver-sanders (approved):
LGTM, tested as working.

Ensured the cylc-rose test fails on 8.0.x and passes on this branch.

PR #5138, opened 2022-09-16
add runtime fields/deltas to def, proxy, job data elements

### LINKED ISSUE #5054: graphql: add [runtime] information to schema
We should expose the `[runtime]` section of the workflow configuration via the GraphQL schema.

This is needed for:
* Trigger edit (#3751).
* Displaying `[runtime]` in the "info" view (https://github.com/cylc/cylc-ui/issues/1071)
* Extraction of task configuration for workflow analysis.

This should be exposed for one-off queries but we don't need to (and probably wouldn't want to) expose this for subscriptions using deltas as that would involve maintaining the entire config in the data store.

Technical barriers:
* The config values need to be serialised:
  * (I.E. The configurations should be exposed as they were written in the `flow.cylc` file).
  * Parsec "coerces" values from their written format into the desired types (e.g. `PT6H` is converted into `Duration(hours=6)`).
  * We would need to "uncoerse" (i.e. serialise) these values back to strings. Parsec does not have an interface for this :(
  * Note we have to do this because GraphQL/JSON do not support things like isodatetime types.
* Broadcasts should be applied:
  * For trigger edit (and likely other uses) we would need to extract the `[runtime]` section for a given task instance (i.e. `TaskProxy`) rather than an abstract task (i.e. `TaskDef`).
  * This means we need to apply broadcasts to the `[runtime]` section before dumping it.
* We will need to decide what format to expose the `[runtime]` section in:
  * I.E. expose it as a multiline string or as a nested dictionary.
  * See also https://github.com/cylc/cylc-flow/issues/3751#issuecomment-789749725.
* Future tasks / historical jobs:
  * The main use cases here are for querying the `[runtime]` config of **tasks** which have not been run yet (i.e. future tasks).
  * Long term it would be good to query historical **jobs**.
  * E.G. I've trigger edited a task a couple of times, can I go back and see what config they ran with?

Extension:
* Add `[runtime]` config to `cylc show`?

**Pull requests welcome!**

dwsutherland:
closes #5054 

Includes runtime field for def, proxy, and jobs .. with:

- def the task/family original workflow definition
- proxy the task/family point version, that changes (deltas available) as soon as you broadcast
- jobs, a live record of what they ran with.

**example:**
image
broadcasting:
```
cylc broadcast -s '[environment]GREETING = Broadcasting hello' '~sutherlander/linear/run1'
```
and then running `foo` again results in:
```
          {
            "id": "~sutherlander/linear/run1//20220901T00/foo",
            "state": "succeeded",
            "runtime": {
              "platform": "",
              "script": "sleep 2; echo \"$GREETING\"",
              "initScript": "echo 'Me first'",
              "envScript": "echo \"Hi first, I'm second\"",
              "errScript": "echo 'Boo!'",
              "exitScript": "echo 'Yay!'",
              "preScript": "sleep 1",
              "postScript": "sleep 1",
              "workSubDir": "",
              "executionTimeLimit": 0,
              "directives": {},
              "environment": {
                "GREETING": "Broadcasting hello"
              },
              "outputs": {}
            },
            "jobs": [
              {
                "id": "~sutherlander/linear/run1//20220901T00/foo/02",
                "state": "succeeded",
                "jobLogDir": "/home/sutherlander/cylc-run/linear/run1/log/job/20220901T00/foo/02",
                "runtime": {
                  "platform": "localhost",
                  "script": "sleep 2; echo \"$GREETING\"",
                  "initScript": "echo 'Me first'",
                  "envScript": "echo \"Hi first, I'm second\"",
                  "errScript": "echo 'Boo!'",
                  "exitScript": "echo 'Yay!'",
                  "preScript": "sleep 1",
                  "postScript": "sleep 1",
                  "workSubDir": "",
                  "executionTimeLimit": 0,
                  "directives": {},
                  "environment": {
                    "GREETING": "Broadcasting hello"
                  },
                  "outputs": {}
                }
              },
              {
                "id": "~sutherlander/linear/run1//20220901T00/foo/01",
                "state": "succeeded",
                "jobLogDir": "/home/sutherlander/cylc-run/linear/run1/log/job/20220901T00/foo/01",
                "runtime": {
                  "platform": "localhost",
                  "script": "sleep 2; echo \"$GREETING\"",
                  "initScript": "echo 'Me first'",
                  "envScript": "echo \"Hi first, I'm second\"",
                  "errScript": "echo 'Boo!'",
                  "exitScript": "echo 'Yay!'",
                  "preScript": "sleep 1",
                  "postScript": "sleep 1",
                  "workSubDir": "",
                  "executionTimeLimit": 0,
                  "directives": {},
                  "environment": {
                    "GREETING": "Hello from foo!"
                  },
                  "outputs": {}
                }
              }
            ]
          }
```

Cancel broadcast will create deltas on effected proxy nodes also.

**Check List**


hjoliver:
Did you mean to remove `cylc/flow/data_messages_pb2.py` ?

hjoliver:
Did you mean to remove `cylc/flow/data_messages_pb2.py` ?

dwsutherland:
> Did you mean to remove `cylc/flow/data_messages_pb2.py` ?

It's not removed (nothing would work otherwise), and I used the corresponding `protoc`:
```
$ protoc --version
libprotoc 3.19.4
```
It's just massively reduced.. like in:
https://github.com/cylc/cylc-flow/pull/4901

(but not using that new version, so must have happened in `    protobuf==3.19.*`)..

cylc/flow/data_messages.proto
```diff
@@ -117,25 +140,13 @@ message PbJob {
     optional string finished_time = 8;
     optional string job_id = 9;
     optional string job_runner_name = 10;
-    optional string env_script = 11;
-    optional string err_script = 12;
-    optional string exit_script = 13;
-    optional float execution_time_limit = 14;
     optional string platform = 15;
```
dwsutherland, line 143: Left this in, so current API doesn't break.

cylc/flow/task_job_mgr.py
```diff
@@ -617,40 +617,6 @@ def _create_job_log_path(workflow, itask):
                 exc.filename = target
             raise exc
 
-    @staticmethod
-    def _get_job_scripts(itask, rtconfig):
```
dwsutherland, line 621: This seemed to be a double up of what's already done in the `config.py` .. The only difference is the omission of environment `CYLC_TASK_CYCLE_POINT` for `str(itask.point)` (which isn't available earlier):
```
            comstr = (
                "cylc workflow-state "
                + " --task=" + itask.tdef.workflow_polling_cfg['task']
                + " --point=" + str(itask.point)
            )
```
This doesn't appear to have an impact, however, I could be wrong.. (is the env variable ever not available?)

cylc/flow/data_messages.proto
```diff
@@ -106,6 +106,29 @@ message PbWorkflow {
...
+    optional string init_script = 3;
+    optional string env_script = 4;
+    optional string err_script = 5;
+    optional string exit_script = 6;
+    optional string pre_script = 7;
+    optional string post_script = 8;
+    optional string work_sub_dir = 9;
+    repeated string execution_polling_intervals = 10;
+    repeated string execution_retry_delays = 11;
+    optional float execution_time_limit = 12;
```
MetRonnie, line 122: I think this is a `PT4S` style value so would that not have to be a string?
https://cylc.github.io/cylc-doc/stable/html/reference/config/workflow.html#flow.cylc[runtime][%3Cnamespace%3E]execution%20time%20limit

cylc/flow/data_store_mgr.py
```diff
@@ -188,6 +192,30 @@ def task_mean_elapsed_time(tdef):
     return tdef.rtconfig.get('execution time limit', None)
 
 
+def runtime_from_config(rtconfig):
+    """Populate runtime object from config."""
+    try:
+        platform = rtconfig['platform']['name']
+    except (KeyError, TypeError):
```
MetRonnie, line 199: Shouldn't `rtconfig['platform']` just be a string of the platform name? In what circumstance would it be a dict containing a `name` key?

cylc/flow/network/schema.py
```diff
@@ -703,6 +703,39 @@ class Meta:
...
+Subset of runtime fields.
+
+Existing on 3 different node types:
+- Task/family Definition
+- Task/family cycle instance
+- Job
+
+Definition is the task/family hierarchy processed base.
+Task/famlily cycle instance is the broadcast modified (if any) of the base.
+Job is a record of what was run for a particular submit.
```
MetRonnie, line 718: I think this is clearer:
```suggestion
- Task/family definition (from the flow.cylc file)
- Task/family cycle instance (includes any broadcasts on top of the definition)
- Job (a record of what was run for a particular submit)
```

cylc/flow/scripts/broadcast.py
```diff
@@ -421,20 +421,24 @@ async def run(options: 'Values', workflow_id):
...
+            if modified_settings:
+                ret['stdout'].append(
+                    get_broadcast_change_report(
+                        modified_settings,
+                        is_cancel=report_cancel,
+                    )
                 )
-            )
+        bad_result = report_bad_options(bad_options, is_set=report_set)
+    except TypeError:
```
MetRonnie, line 436: Is it necessary to catch `TypeError`s for all these lines? Could it catch more than anticipated?

cylc/flow/data_messages.proto
dwsutherland: That's what I thought (and originally had it set to string), however, it appears to be converted into seconds somewhere... (and I believe is used to calculate the progress dial)

dwsutherland: The string version would be preferable when presenting to the user (for trigger-edit or otherwise), but it's not what we have right at present.

cylc/flow/data_store_mgr.py
dwsutherland: The confusing thing is.. Platform gets changed before job submission, both exist.

The dict version appears to be more common..

cylc/flow/scripts/broadcast.py
dwsutherland: Well, I believe it's the `for` line that's the issue.. When results isn't a dictionary.. so I could test for that..

dwsutherland: Confirmed, when results is an error list and not a dictionary it fails (other errors still caught normally):
image
This is for internal errors that would otherwise be lost.

I could test that results is a dictionary, instead, but if somehow it catches more than anticipated at least the end user would have something to report back to us (instead of just a `TypeError` error in the script)..

cylc/flow/data_messages.proto
dwsutherland: I can't find anywhere internally that the string version is used even the `TaskProxy`s define it as float:
```
            execution_time_limit (float):
                Execution time limit of latest job.
```
It's as if it gets converted when the config is read in... maybe a follow-on PR.

cylc/flow/network/schema.py
```diff
@@ -703,6 +703,35 @@ class Meta:
...
+    err_script = String(default_value=None)
+    exit_script = String(default_value=None)
+    pre_script = String(default_value=None)
+    post_script = String(default_value=None)
+    work_sub_dir = String(default_value=None)
+    execution_polling_intervals = graphene.List(String)
+    execution_retry_delays = graphene.List(String)
+    execution_time_limit = Float(default_value=None)
+    submission_polling_intervals = graphene.List(String)
+    submissino_retry_delays = graphene.List(String)
```
MetRonnie, line 729:
```suggestion
    submission_retry_delays = graphene.List(String)
```

MetRonnie (commented):
Tested out in graphiql, looks good. Just noticed 1 typo

dwsutherland:
This will need to be rebased once #4901 goes in (or visa versa)

cylc/flow/data_store_mgr.py
hjoliver: That is confusing. Is it just because one of your calls to `runtime_from_config` passes in the job config instead of the `taskdef.rtconfig`?

MetRonnie:
Another problem Oliver has just realised: the order of the `[environment]` variables needs to be preserved, which we don't think it currently is by the GraphQL mechanism if using a dictionary as you've done at present

oliver-sanders:
I think we'll need a list of environment variables to get around this e.g:

```json
environment: [
    {"key": "answer", "value": 42}
]
```

dwsutherland:
Investigating.. it appears the job still gets it:
```
          {
            "name": "foo",
            "runtime": {
              "environment": {}
            },
            "jobs": [
              {
                "runtime": {
                  "platform": "localhost",
                  "script": "sleep 60",
                  "initScript": "",
                  "envScript": "",
                  "errScript": "",
                  "exitScript": "",
                  "preScript": "",
                  "postScript": "",
                  "workSubDir": "",
                  "executionTimeLimit": 0,
                  "directives": {},
                  "environment": {
                    "root": "b0",
                    "FOO": "1",
                    "foo": "2"
                  },
                  "outputs": {}
                }
              },
```
(just did one broadcast)
So something to do with data-store.. (happens with both integer and non-integer cycling as expected)

dwsutherland:
Ok I've fixed the problem.. I was using `get()` with ordered default dict, I changed to that as a workaround for some tests (so they might fail, and I'll just need to adjust the test(s))..

WRT the environment order:

> Another problem Oliver has just realised: the order of the [environment] variables needs to be preserved, which we don't think it currently is by the GraphQL mechanism if using a dictionary as you've done at present

Dictionaries are ordered, and I dump them as string fields:
```
PbRuntime(
        platform=platform,
        script=rtconfig['script'],
        init_script=rtconfig['init-script'],
        env_script=rtconfig['env-script'],
        err_script=rtconfig['err-script'],
        exit_script=rtconfig['exit-script'],
        pre_script=rtconfig['pre-script'],
        post_script=rtconfig['post-script'],
        work_sub_dir=rtconfig['work sub-directory'],
        execution_time_limit=rtconfig['execution time limit'],
        directives=json.dumps(directives),
        environment=json.dumps(environment),
    )
```
The resolver does:
```
def resolve_json_dump(root, info, **args):
    field = getattr(root, to_snake_case(info.field_name), '{}') or '{}'
    return json.loads(field)
```

So unless something odd happens between this and firing it off to the client, the fields should be fine..

MetRonnie:
I've had a look at `[environment]` order and confirmed the response order is identical to that which you get in `log/config/NN-start-MM.cylc`, including when inheritance is involved.

MetRonnie:
Getting a KeyError on `tests/functional/retries/01-submission-retry.t`
```
ERROR - 'directives'
  File "~/cylc-flow-8.1/cylc/flow/scheduler.py", line 621, in run_scheduler
    await self.main_loop()
  File "~/cylc-flow-8.1/cylc/flow/scheduler.py", line 1580, in main_loop
    self.release_queued_tasks()
  File "~/cylc-flow-8.1/cylc/flow/scheduler.py", line 1303, in release_queued_tasks
  File "~/cylc-flow-8.1/cylc/flow/task_job_mgr.py", line 481, in submit_task_jobs
    self._prep_submit_task_job_error(
  File "~/cylc-flow-8.1/cylc/flow/task_job_mgr.py", line 1235, in _prep_submit_task_job_error
    self.task_events_mgr.process_message(
  File "~/cylc-flow-8.1/cylc/flow/task_events_mgr.py", line 652, in process_message
    if self._process_message_submit_failed(
  File "~/cylc-flow-8.1/cylc/flow/task_events_mgr.py", line 1186, in _process_message_submit_failed
    self._insert_task_job(itask, event_time, self.JOB_SUBMIT_FAIL_FLAG)
  File "~/cylc-flow-8.1/cylc/flow/task_events_mgr.py", line 1264, in _insert_task_job
    self.data_store_mgr.insert_job(
  File "~/cylc-flow-8.1/cylc/flow/data_store_mgr.py", line 1250, in insert_job
    j_buf.runtime.CopyFrom(runtime_from_config(j_cfg))
  File "~/cylc-flow-8.1/cylc/flow/data_store_mgr.py", line 201, in runtime_from_config
    directives = rtconfig['directives']
KeyError: 'directives'
```

dwsutherland:
> Getting a KeyError on `tests/functional/retries/01-submission-retry.t`

Sorted now.. It's because some submission failures happen before full job construction.
(this is why I had changed it before, causing the field deletion due to parsec dictionary behavior with `get()`)

cylc/flow/data_store_mgr.py
hjoliver: Ping @dwsutherland - one question to respond to here. As I recall, I didn't think it should be like this, hence my suggested reason for it.

dwsutherland: Yes, the platform gets fully resolved/loaded on job construction.. however it's just a string in the workflow config runtime.

oliver-sanders:
> WRT the environment order:
>
>> Another problem Oliver has just realised: the order of the [environment] variables needs to be preserved, which we don't think it currently is by the GraphQL mechanism if using a dictionary as you've done at present
>>
>>> Dictionaries are ordered, and I dump them as string fields:

Dictionaries are ordered in Python land, however, they are not in either JSON or JS. Because JSON is the transport format GraphQL uses and the UI code is JS the order provided by the Python code will not be preserved.

So we need a JSON structure like this to preserve order:

```json
[
  {
    "key": "FOO",
    "value": "42"
  },
  {
    "key": "BAR",
    "value": "answer"
  }
]
```

oliver-sanders:
Here's the problem:

> An object is an unordered set of name/value pairs.
>
> -- https://www.json.org/json-en.html

Some implementations of JSON writers/parsers might decide to preserve order, but it's completely chance if they do.

dwsutherland:
> Some implementations of JSON writers/parsers might decide to preserve order, but it's completely chance if they do.

So even if the Python json library writes them in order, the parser(s) at the client end we can't guarantee?

dwsutherland:
I suspect the python side is ordered according to the dict (`json` has a `sort_keys` option but we don't use it),  because the graphql tests would be very flaky otherwise.. 

Yeah they are:
https://docs.python.org/3/library/json.html

> **Note:** This module’s encoders and decoders preserve input and output order by default. Order is only lost if the underlying containers are unordered.

oliver-sanders:
The Python end is fine, it's the JS side which is problematic.

MetRonnie:
I don't seem to be getting the list-type fields.
```ini
[runtime]
    [[foo]]
        execution retry delays = 5*PT2M, PT5M
```
```js
{
  "data": {
    "taskProxy": {
      "runtime": {
        "executionRetryDelays": [], // empty
        ...
```

cylc/flow/data_store_mgr.py
```diff
@@ -188,6 +192,30 @@ def task_mean_elapsed_time(tdef):
...
+        env_script=rtconfig['env-script'],
+        err_script=rtconfig['err-script'],
+        exit_script=rtconfig['exit-script'],
+        pre_script=rtconfig['pre-script'],
+        post_script=rtconfig['post-script'],
+        work_sub_dir=rtconfig['work sub-directory'],
+        execution_time_limit=rtconfig['execution time limit'],
+        directives=json.dumps(directives),
+        environment=json.dumps(environment),
+    )
```
MetRonnie, line 216: Isn't `outputs` missing here?

dwsutherland: Outputs are at the taskProxy level, but could be added here.. They are less static though.

dwsutherland:
> Just because we push them out in order doesn't mean they'll be read back in order, or preserve their order when operated on by JS.

Yeah, but, before I go an change it.. Does the JS json library read it in order? (regardless of the `1999` spec)

oliver-sanders:
> Or is there no order guarantee post read (in memory) also?

tldr; JSON doesn't guarantee order, JS doesn't guarantee order, ECMA2015/ES6 does guarantee order in some cases, for some, but not all interfaces. So order is largely down to the implementation and likely to be unstable.

Searching around I found this on ECMA2015/ES6:

> Conclusion: Even in ES2015 you shouldn't rely on the property order of normal objects in JavaScript. It is prone to errors. If you need ordered named pairs, use [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) instead, which purely uses insertion order. If you just need order, use an array or [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) (which also uses purely insertion order).
>
> -- https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order/38218582#38218582

Generally JS things seem to use Arrays of Objects where order matters.

dwsutherland:
> > So I think we need a JSON structure like this to preserve order:
> > ```json
> > [
> >   {
> >     "key": "FOO",
> >     "value": "42"
> >   },
> >   {
> >     "key": "BAR",
> >     "value": "answer"
> >   }
> > ]
> > ```
> 
> I think it would also make sense to make `directives` and `outputs` use that structure, even if they don't need to preserve order. (It makes it easier to handle in cylc-ui if they're consistent with `environment`)

Why not just:
```
[
  {
    "FOO": "42"
  },
  {
    "BAR": "answer"
  }
]
```
Any reason for the `"key"` `"value"`?

dwsutherland:
Have changed it to:
```
[
  {
    "FOO": "42"
  },
  {
    "BAR": "answer"
  }
]
```
and added outputs.

MetRonnie:
Still have this problem where e.g. `execution retry delays` doesn't come through
 
> ```ini
> [runtime]
>     [[foo]]
>         execution retry delays = 5*PT2M, PT5M
> ```
> 
> ```js
> {
>   "data": {
>     "taskProxy": {
>       "runtime": {
>         "executionRetryDelays": [], // empty
>         ...
> ```

dwsutherland:
> > Any reason for the `"key"` `"value"`?
> 
> It would make it easier to use in the Vue component.

So you don't have to iterate the dict? Ok will change..

dwsutherland:
Fixed .. however, like the execution timeout, the string values are converted on config parse:
```
              "submissionRetryDelays": [
                "120.0"
              ],
              "executionRetryDelays": [
                "15.0",
                "600.0",
                "3600.0",
                "10800.0"
              ]
```
so that's a wider issue of not having the original string available.

oliver-sanders:
> so that's a wider issue of not having the original string available.

Parsec can convert these values back to the original form which is what happens when we run `cylc config <id>` so we should be able to handle it.

If we were to `cylc broadcast` these number values back to the workflow it would rejected as invalid. For the "broadcast edit" work we'll have to check whether this applies to any other fields.

dwsutherland:
I've "stringif"d the float lists, and the numerical types already have a `str` method to convert into appropriate list:
<details>
  <summary>graphql output</summary>
```
          {
            "name": "foo",
            "runtime": {
              "environment": [
                {
                  "key": "GREETING",
                  "value": "Hello from foo!"
                }
              ],
              "outputs": [
                {
                  "key": "OUTPUT1",
                  "value": "yes"
                },
                {
                  "key": "OUTPUT2",
                  "value": "no"
                }
              ],
              "submissionRetryDelays": "PT2M",
              "submissionPollingIntervals": "PT1M, PT2M",
              "executionTimeLimit": "PT1H",
              "executionRetryDelays": "PT15S, PT10M, PT1H, PT3H",
              "executionPollingIntervals": "PT1M, PT10M, PT10M, PT10M, PT10M, PT10M, PT1M"
            },
            "outputs": [
              {
                "label": "expired",
                "message": "expired",
                "satisfied": false,
                "time": 1667363549.6685379
              },
              {
                "label": "submitted",
                "message": "submitted",
                "satisfied": true,
                "time": 1667363558.671344
              },
              {
                "label": "submit-failed",
                "message": "submit-failed",
                "satisfied": false,
                "time": 1667363549.6685379
              },
              {
                "label": "started",
                "message": "started",
                "satisfied": true,
                "time": 1667363559.6712284
              },
              {
                "label": "succeeded",
                "message": "succeeded",
                "satisfied": false,
                "time": 1667363549.6685379
              },
              {
                "label": "failed",
                "message": "failed",
                "satisfied": false,
                "time": 1667363549.6685379
              },
              {
                "label": "OUTPUT1",
                "message": "yes",
                "satisfied": false,
                "time": 1667363549.6685379
              },
              {
                "label": "OUTPUT2",
                "message": "no",
                "satisfied": false,
                "time": 1667363549.6685379
              }
            ],
            "jobs": [
              {
                "executionTimeLimit": 3600,
                "runtime": {
                  "platform": "localhost",
                  "script": "sleep 2; echo \"$GREETING\"",
                  "initScript": "echo 'Me first'",
                  "envScript": "echo \"Hi first, I'm second\"",
                  "errScript": "echo 'Boo!'",
                  "exitScript": "echo 'Yay!'",
                  "preScript": "sleep 1",
                  "postScript": "sleep 1",
                  "workSubDir": "",
                  "executionTimeLimit": "PT1H",
                  "directives": [],
                  "environment": [
                    {
                      "key": "GREETING",
                      "value": "Hello from foo!"
                    }
                  ],
                  "outputs": [
                    {
                      "key": "OUTPUT1",
                      "value": "yes"
                    },
                    {
                      "key": "OUTPUT2",
                      "value": "no"
                    }
                  ]
                }
              }
            ]
          }
```
</details>
One thing, they obviously don't get grouped and/or replicate the original exactly:
```
    [[FAM]]
        execution time limit = PT1H
        execution polling intervals = PT1M, 5*PT10M, PT1M
        execution retry delays = PT15S, PT10M, PT1H, PT3H
        submission polling intervals = PT1M, PT2M
        submission retry delays = PT2M

```
(the function I'm calling does attempt to group for other things)

Also when doing a broadcast;
`cylc broadcast linear -n FAM -s 'execution retry delays=PT25S'`
it doesn't get converted into that shorthand time format:
```
              "submissionRetryDelays": "PT2M",
              "submissionPollingIntervals": "PT1M, PT2M",
              "executionTimeLimit": "PT1H",
              "executionRetryDelays": "25.0",
              "executionPollingIntervals": "PT1M, PT10M, PT10M, PT10M, PT10M, PT10M, PT1M"
```
This is because the broadcasts use regular objects and not the same as our config..
This is will be my next commit, to change how broadcasts are assimilated and produce overrides.

MetRonnie (approved):
I think this is looking good now. Will probably need to use some custom types for the fields of `runtime` in the schema, for the benefit of the [upcoming Edit-Runtime form UI](https://github.com/cylc/cylc-ui/issues/1104), but that can come later.

dwsutherland:
The latest commit fixes the incorrect displayed format:
**before**
```
$ cylc broadcast linear -n FAM -s 'execution retry delays=PT35S, PT60S, PT130M'
Broadcast set:
+ [*/FAM] execution retry delays=[35.0, 60.0, 7800.0]
```
**after**
```
$ cylc broadcast linear -n FAM -s 'execution retry delays=PT35S, PT60S, PT130M'
Broadcast set:
+ [*/FAM] execution retry delays=PT35S, PT60S, PT130M
```

It does this by keeping the raw `.cylc` format external, which includes:
- Arguments from broadcast command/mutation
- Saved DB format
- GraphQL/API runtime fields

And coercing this external format to internal objects (i.e. duration fields (`execution retry delays`...) to `DurationFloat` values) when assimilating DB loads or CLI/API broadcast put mutations.

Previously the internal format was created externally then read in like this.. 
Which meant we would get `float` instead of `DurationFloat` internally (as `DurationFloat` isn't passed over json)... This is a bug (resulting in the CLI-response and API-field being in seconds).

Have added to an existing test, to cover these fields..
(there maybe be some test failures from DB loads that have broadcast settings saved in the `Float` format.. will address)

dwsutherland:
hmm...

> Error: Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v3/dist/codecov' failed with exit code 255

for both "failed" tests (2nd attempt) ..  the actual tests passed

MetRonnie:
Unfortunately this breaks restart compatibility with 8.0.x workflows, as if you had done a broadcast for e.g. `execution time limit` you would get an `IllegalValueError` on restart with 8.1

dwsutherland:
> Unfortunately this breaks restart compatibility with 8.0.x workflows, as if you had done a broadcast for e.g. `execution time limit` you would get an `IllegalValueError` on restart with 8.1

It's done the wrong way at the moment.. So there'll need to be an upgrade somehow.

dwsutherland:
Ok, I've put a workaround in that will check and alter `8.0.x` broadcast db loads by;
- creating a `DurationFloat` out of the seconds string.
- strip list strings of `[` `]` if they both exist on the respective ends. (this is narrowed down to list settings in the config only)

here's an old and new side by side 
```
2022-11-11T00:44:45Z INFO - + [*/foo] execution retry delays=[120.0, 180.0]
2022-11-11T00:44:45Z INFO - + [*/foo] submission retry delays=PT2M, PT3M
```

This workaround can be removed some time in the future...

dwsutherland:
(most the coverage failures are pre-existing, and just replicated-in/copied-to the workaround)

MetRonnie:
I have opened a PR against this with a functional test for the restart: https://github.com/dwsutherland/cylc-flow/pull/10

cylc/flow/data_messages.proto
oliver-sanders: Makes sense to keep this as the job platform is chosen based on the runtime configuration (e.g. `platform` could be set to a platform-group) causing these two values to differ.

cylc/flow/task_job_mgr.py
oliver-sanders: `CYLC_TASK_CYCLE_POINT` should always be available in job execution environments so should be good.

Took a look and I think we are safe to remove this, cheers.

tests/functional/database/08-broadcast-upgrade/flow.cylc
```diff
@@ -0,0 +1,5 @@
+[scheduler]
+    allow implicit tasks = True
+[scheduling]
+    [[graph]]
+        R1 = foo
```
oliver-sanders, line 5: This test confirms that the workflow can restart, but doesn't confirm that the broadcasted execution retry delays actually work.

Re-using the script example from `database/10` here would confirm the broadcast worked e.g:

```ini
[runtime]
    [[foo]]
        script = """
            if (($CYLC_TASK_TRY_NUMBER < 2 )); then
                false
            fi
        """
        # NOTE: "execution retry delays" set by broadcast
```

cylc/flow/parsec/validate.py
```diff
@@ -1047,6 +1045,64 @@ def _coerce_type(cls, value):
         return val
 
 
+# TODO:- Remove at some suitable time post 8.1.0,
+# as DB broadcast loading on restart from 8.0.x may break.
+# The DB at 8.0.x stores Interval values as neither ISO8601 duration string
+# or DurationFloat. This has been fixed at 8.1.0, and the following class
+# acts as a bridge between fixed and broken.
```
oliver-sanders, line 1052: Should stick this comment into the `BACK COMPAT` format so that we pick this up when we next review deprecated code.

```suggestion
# BACK COMPAT: post_load_db_coerce
# The DB at 8.0.x stores Interval values as neither ISO8601 duration
# string or DurationFloat. This has been fixed at 8.1.0, and
# the following class acts as a bridge between fixed and broken.
# url:
#     https://github.com/cylc/cylc-flow/pull/5138
# from:
#    8.0.x
# to:
#    8.1.x
# remove at:
#    8.x
```

tests/functional/database/08-broadcast-upgrade/flow.cylc
MetRonnie: I've added this in https://github.com/dwsutherland/cylc-flow/pull/11

cylc/flow/parsec/validate.py
MetRonnie: I'ev added this in https://github.com/dwsutherland/cylc-flow/pull/11

tests/functional/database/08-broadcast-upgrade/flow.cylc
dwsutherland: I also modified an existing test that did a similar thing.

oliver-sanders (commented):
LGTM, tested runtime fields on the following objects:

* `Task` (task config as defined in flow.cylc, no broadcasts applied)
* `TaskProxy` (task config + broadcasts)
* `Family` (family config as defined in flow.cylc, no broadcasts applied)
* `FamilyProxy` (family config + broadcasts)
* `Job` (task config + broadcasts at the time of job submission, no backsies)

All worked as expected :rocket:

I spotted two minor issues during the course of testing...

#### 1) Duration lists get formatted as integers when cleared.

When I create a broadcast it is presented to me in native-ISO8601 format:

```console
$ cylc broadcast broad -s 'execution retry delays = PT1M'
Broadcast set:
+ [*/root] execution retry delays=PT1M
```

But when I clear the broadcast it is presented to me in integer format:

```console
$ cylc broadcast --clear broad
Broadcast cancelled:
- [*/root] execution retry delays=[60.0]
```

#### 2) Updated deltas contain all fields

E.G. for this GraphQL query:

```graphql
subscription {
  deltas {
    updated(stripNull: true) {
      taskProxies {
        id
        runtime {
          script
          executionRetryDelays
        }
      }
    }
  }
}
```

And the following broadcast:

```console
$ cylc broadcast broad -s 'execution retry delays = PT10M'
Broadcast set:
+ [*/root] execution retry delays=PT10M
```

I get an updated delta like this:

```
{
  "data": {
    "deltas": {
      "updated": {
        "taskProxies": [
          {
            "id": "~osanders/broad//20191209T1200Z/pub",
            "runtime": {
              "script": "sleep 1",
              "executionRetryDelays": "PT10M"
            }
          },
          {
            "id": "~osanders/broad//20191209T1200Z/wipe_bar",
            "runtime": {
              "script": "sleep 1",
              "executionRetryDelays": "PT10M"
            }
          },
```

Which contains the `execution retry delays` which have changed but also the `script` which hasn't.

As long as this doesn't cause broadcasted runtime fields to get re-sent for things like task state changes this is probably harmless.

MetRonnie (approved):

oliver-sanders (approved):
Thanks @dwsutherland :rocket:

PR #5139, opened 2022-09-16
`log_vc_info`: Redirect diff straight to file to avoid blocking pipe

### LINKED ISSUE #5117: log_vc_info: support large diffs
I've seen a case where `svn diff --internal-diff --non-interactive` was hanging for some reason. Not sure why just yet but it might be worth putting a default timeout on `svn` commands called by the `log_vc_info` plugin to be safe.

Tagged against 8.0.2 but not a critical priority unless others run into it.

When the process was killed the only logged error was:

```
PluginError: Error in plugin cylc.post_install.log_vc_info
OSError: [Errno -9]
```

The `-9` presumably coming from `kill -9` looking at the code.

**Pull requests welcome!**

MetRonnie:
Closes #5117. Fixes bug where `cylc install` would hang if the uncommitted diff of the source dir repo was large.

Instead of using `subprocess.PIPE` for the output of git/svn diff commands, use the file object itself. This avoids blocking the pipe if the diff is large, and also has the benefit of not loading large amounts of text into memory.

**Check List**

- [x] No dependency changes 
- [x] Tests updated
- [x] No docs change needed

### DIFF
--- a/cylc/flow/install_plugins/log_vc_info.py
+++ b/cylc/flow/install_plugins/log_vc_info.py
@@ -62,7 +62,9 @@ Any uncommitted changes will also be saved as a diff in
 import json
 from pathlib import Path
 from subprocess import Popen, DEVNULL, PIPE
-from typing import Any, Dict, Iterable, List, Optional, TYPE_CHECKING, Union
+from typing import (
+    Any, Dict, Iterable, List, Optional, TYPE_CHECKING, TextIO, Union, overload
+)
 
 from cylc.flow import LOG
 from cylc.flow.exceptions import CylcError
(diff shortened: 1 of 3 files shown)


cylc/flow/install_plugins/log_vc_info.py
```diff
@@ -275,41 +302,40 @@ def _parse_svn_info(info_text: str) -> Dict[str, Any]:
...
+        f.write(
+            "# Auto-generated diff of uncommitted changes in the Cylc "
+            "workflow repository:\n"
+            f"#   {repo_path}\n"
+        )
+        f.flush()
+        try:
+            _run_cmd(vcs, args, repo_path, stdout=f)
+        except VCSMissingBaseError as exc:
+            f.write(f"# No diff - {exc}")
```
oliver-sanders, line 337: I think this is more "could not extract diff" than "no diff"? Should it say "ERROR"?

oliver-sanders (commented):
LGTM

cylc/flow/install_plugins/log_vc_info.py
MetRonnie: Not so much an error, it's basically "Git diff does not produce an output when there is no base commit"

oliver-sanders (approved):

wxtim (approved):

Issue #5135, opened 2022-09-14
review: add version/* files to dropdown menus


oliver-sanders:
The log/version files are not presently visible in Cylc Review, the Rose 2019 equivalents were.

Given the number of back-and-forths here it might be worth changing tack on how we list these files, e.g an approach along the lines of this might be better:

```
find log/ -maxdepth 2 -type f
```

**Pull requests welcome!**

wxtim:
> The log/version files are not presently visible in Cylc Review, the Rose 2019 equivalents were.
> 
> Given the number of back-and-forths here it might be worth changing tack on how we list these files, e.g an approach along the lines of this might be better:
> 
> ```
> find log/ -maxdepth 2 -type f
> ```
> 
> **Pull requests welcome!**

I'd be happy to expose everything in log which is not in log/job - @dpmatthews  Can we agree on this approach?

datamel:
To summarize discussion this morning with @dpmatthews the decision has been made to go with the one line fix of adding the version dir. This is to minimize interference with existing logic.
If changes are subsequently needed, a search type fix would be the preferred solution.

Note that while testing the fix, another issue has cropped up with SQLite errors and traceback visible, investigation is ongoing with this.

datamel:
This was closed by: https://github.com/cylc/cylc-flow/pull/5140

PR #5140, opened 2022-09-16
Cylc review update

datamel:
Closes https://github.com/cylc/cylc-flow/issues/5135

I managed to get SQLite3 traceback with a workflow where the `flow.cylc` had been deleted (for me, a cylc test battery test had not fully cleaned up correctly). This check was added in  https://github.com/cylc/cylc-flow/pull/5111. I think if someone deletes the suite.rc/flow.cylc files they should be able to continue to use cylc review without traceback, unless I am missing something?   

Investigation continues into `ProgrammingError: SQLite objects created in a thread can only be used in that same thread` and `CylcReviewService object is not callable` errors.

**Check List**

- [x] Tests are included (or explain why tests are not needed - legacy).
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request not required.


datamel:
Update to this, there has been another bug found in cylc review. While viewing a workflow that had not yet been run, an old database file `cylc-suite.db` has been created. This renders the workflow unable to play at cylc8. This is also causing these workflows that have not run before to appear at the top of the list of running workflows, since they have been 'updated' recently.

hjoliver:
Roger that, still holding off on the "final" cylc 7 release :grin:

datamel:
Another Cylc review bug found today, workflows which have used the flow feature to run tasks multiple times are causing duplicate entries.

datamel:
Update, last of the bugs fixed 
- problems with threading of database object, this resulted in a sqlite ProgrammingError. This is difficult to reproduce, no clear steps and was seemingly random but occurred fairly frequently on master
- some db _exec calls were not closed. 
- version files now should be visible
- this last one is undocumented as an issue, summary to reproduce here for reviewers:
Run a cylc 8 workflow and trigger an already run task with `--flow=new`
On `7.8.x` duplicate task entries are found, in this branch there should only be the correct number showing.

This is now ready for review.

lib/cylc/review_dao.py
```diff
@@ -218,8 +221,12 @@ def get_suite_job_entries(
                 return ([], 0)
         except sqlite3.Error:
             return ([], 0)
-
-        if self.is_cylc8(user_name, suite_name):
+        from cylc.review import CylcReviewService
```
datamel, line 224: This is imported down here due to a circular import problem. Unsure if there is a better way of solving this.

lib/cylc/review.py
```diff
@@ -836,6 +839,19 @@ def _get_suite_logs_info(self, user, suite):
 
         return data
 
+    @staticmethod
+    def is_cylc8(user_suite_dir):
```
wxtim, line 843: Nice. Since it's a staticmethod couple of unit tests might be nice. You can probably mock the results of os.path.exists so that you don't have to touch the file system.

lib/cylc/review.py
```diff
@@ -748,9 +750,9 @@ def _get_suite_logs_info(self, user, suite):
         """Return a dict of suite-related files including suite logs."""
         data = {"files": {}}
         user_suite_dir = self._get_user_suite_dir(user, suite)  # cylc files
-
         rose_logs_dest = "rose"
-        if self.suite_dao.is_cylc8(user, suite):
+        cylc_8 = self.is_cylc8(user_suite_dir)
```
wxtim, line 754: Why not just have the method in the condition? Do you re-use `cylc_8`?

lib/cylc/review_dao.py
wxtim: Possible alternative - make `is_cylc8` a module function (not a class method) and using
`from cylc.review import is_cylc8`. Might look more elegant.

wxtim:
I've done a code review, nothing that I'm going to demand you change, although I think some suggestions worth considering.

I would like to see a unit test of the `is_cylc8` function.

I'm going to do a practical review.

lib/cylc/review_dao.py
datamel: This import is also needed for the call to `_get_user_home`, not just for `is_cylc8` so making it a module function won't help with that unfortunately. I did look at moving this out but I don't think a big refactor is a good idea.

wxtim (approved):
I've had a fair poke around the Cylc codebase looking for files we've missed.

wxtim (changes requested):
I've commented.

datamel:
> I've commented.

Test added, responded to all your comments I think. I can't spot anything you have said following your second review where you have requested changes though, maybe github is being funny for me.

lib/cylc/tests/test_review.py
```diff
@@ -0,0 +1,59 @@
...
+    if cylc_8:
+        workflow_file = os.path.join(temp_workflow_dir, 'flow.cylc')
+    else:
+        workflow_file = os.path.join(temp_workflow_dir, 'suite.rc')
+    if log_dir:
+        workflow_log_dir = os.path.join(temp_workflow_dir, "log", "scheduler")
+        os.makedirs(workflow_log_dir)
+    with open(workflow_file, 'w'):
+        pass
+    assert CylcReviewService.is_cylc8(temp_workflow_dir) == expected
```
wxtim:
```suggestion
    assert CylcReviewService.is_cylc8(temp_workflow_dir) == expected

```

lib/cylc/tests/test_review.py
```diff
@@ -0,0 +1,59 @@
...
+        ),
+        pytest.param(
+            False, False, False,
+            id="False when suite.rc and no log dir"
+        ),
+    ],
+)
+def test_is_cylc8(cylc_8, log_dir, expected):
+    """Check is_cylc8 returns"""
+    temp_workflow_dir = mkdtemp()
```
wxtim, line 49: Why not just tmp_path fixture?

lib/cylc/tests/test_review.py
```diff
@@ -0,0 +1,59 @@
...
+    if cylc_8:
+        workflow_file = os.path.join(temp_workflow_dir, 'flow.cylc')
+    else:
```
wxtim, line 52: IMO also a case for `cylc_8 is None` - i.e. there is neither a suite.rc or a flow.cylc.

datamel: Added.

lib/cylc/tests/test_review.py
datamel: I had thought it was neater this way, since you have to cast tmp_path to a string, but I've changed it anyway.

wxtim (approved):

lib/cylc/review.py
```diff
@@ -263,6 +263,7 @@ def cycles(
             return self.template_env.get_template("cycles.html").render(**data)
         except jinja2.TemplateError:
             traceback.print_exc()
+        self._db_close(user, suite)
```
oliver-sanders, line 266: Now that you've switched over to the simpler non-db way of telling Cylc 8 from 7 I think we might be best to revert these DB closes for now in-case they have unintended side effects.

oliver-sanders (commented):
Sorry, I keep getting distracted half-way through reviewing this.

LGTM but I think it would be safer to revert the db closes. Whilst I think it's the right thing to do, Cylc Review has worked for a long time without them. There's a risk we don't understand the interactions fully which I'd rather not take with this "final" Cylc 7 release :grin:!

lib/cylc/review.py
datamel: I totally agree, I've removed them now.

oliver-sanders (approved):

datamel:
@hjoliver, this pr covers the last lot of cylc review bugs that surfaced over the last week or so. We should be good to go for the next Cylc 7 release (when convenient).

hjoliver:
> which I'd rather not take with this "final" Cylc 7 release :grin: !

Hey - no jokes about "final" releases! :sob:

PR #5148, opened 2022-09-21
Assume Jinja2 used in global-tests.cylc.

hjoliver:
The test function `create_test_global_config` should create a valid global config file even if `global-tests.cylc` contains Jinja2 code.

This only affects tests, and is a trivial fix, so I don't think it needs a test itself, or a change log entry.

**Check List**

### DIFF
--- a/tests/functional/lib/bash/test_header
+++ b/tests/functional/lib/bash/test_header
@@ -143,8 +143,9 @@
 #         Expect 1 OK test.
 #     create_test_global_config [PRE [POST]]
 #         Create a new global config file $PWD/etc from global-tests.cylc
-#         with PRE and POST pre- and ap-pended (PRE for e.g. jinja2 shebang).
+#         with PRE and POST pre- and ap-pended.
 #         PRE and POST are strings.
+#         The global config starts with #!Jinja2 in case Jinja2 is used.
 #     localhost_fqdn
 #         Get the FQDN of the current host using the same mechanism Cylc uses.
 #     get_fqdn [TARGET]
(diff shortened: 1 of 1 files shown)


oliver-sanders (approved):

oliver-sanders:
(one review should be ok)

oliver-sanders:
Dammit I missed a test failure - https://github.com/cylc/cylc-flow/actions/runs/3094570352/jobs/5008086584

oliver-sanders:
The source of the test error is here:

https://github.com/cylc/cylc-flow/blob/ce00060e818d14b4ba7ab9194675be01ca6a52af/cylc/flow/parsec/jinja2support.py#L174

(note the test deliberately unsets `$HOME` to ensure we can work without it)

hjoliver:
> More curiously, apparently we support user-level Jinja2 filters, but not site-level. News to me, doesn't sound like a good idea as means Jinja2 could behave differently for different user accounts? 

The intention was to allow users to make their own filters, for use in multiple workflows. That's OK isn't it?

> , and in .cylc not .cylc/flow.

That probably reflects the fact that we forgot we had this (and hence didn't move it down into the new flow sub-dir).

> We don't have any users of this at our end.

I haven't checked, but I kinda doubt we do either.

oliver-sanders:
> That's OK isn't it?

I advise people to move user-level configuration that's required for a workflow to run (e.g.`~/bin/`, `~/.bashrc`, `~/.config`) down into the workflow itself.

Otherwise we end up in a situation where the user account setup is a required part of the workflow itself hindering collaboration. So I would consider user-level Jinja2 modules to be bad practice.

hjoliver:
Fair enough, I agree.

I was thinking above it's really a working practice issue, not something we should force on users. 

However, not actively supporting user-level filters is not a draconian move like (say) deliberately cleaning `sys.path` of external locations :grin: 

So, I'd be happy to remove this from Cylc.

PR #5151, opened 2022-09-21
host select: allow unary operators in ranking expressions

oliver-sanders:
Currently we can configure Cylc to select the host with the LEAST available memory:

```
virtual_memory().available
```

But not the most:

```
-1 * virtual_memory().available
```

Because the required operator is not on the whitelist. This PR adds the required operator to the whitelist allowing us to use it and adds more logging / error output for host selection to make it more transparent when things go wrong.

**Check List**

### DIFF
--- a/cylc/flow/host_select.py
+++ b/cylc/flow/host_select.py
@@ -27,10 +27,13 @@ from tokenize import tokenize
 
 from cylc.flow import LOG
 from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
-from cylc.flow.exceptions import CylcError, HostSelectException
+from cylc.flow.exceptions import GlobalConfigError, HostSelectException
 from cylc.flow.hostuserutil import get_fqdn_by_host, is_remote_host
 from cylc.flow.remote import run_cmd, cylc_server_cmd
 from cylc.flow.terminal import parse_dirty_json
+
+
+GLBL_CFG_STR = 'global.cylc[scheduler][run hosts]ranking'
 
 
 def select_workflow_host(cached=True):
@@ -77,10 +80,10 @@ def select_workflow_host(cached=True):
 
 
 def select_host(
-        hosts,
-        ranking_string=None,
-        blacklist=None,
-        blacklist_name=None
+    hosts,
+    ranking_string=None,
+    blacklist=None,
+    blacklist_name=None
 ):
     """Select a host from the provided list.
 
@@ -165,6 +168,7 @@ def select_host(
     if ranking_string:
         # parse rankings
         rankings = list(_get_rankings(ranking_string))
+        data['ranking'] = ranking_string
 
     if not rankings:
         # no metrics or ranking required, pick host at random
@@ -202,10 +206,10 @@ def select_host(
 
 
 def _filter_by_hostname(
-        hosts,
-        blacklist,
-        blacklist_name=None,
-        data=None
+    hosts,
+    blacklist,
+    blacklist_name=None,
+    data=None
 ):
     """Filter out any hosts present in `blacklist`.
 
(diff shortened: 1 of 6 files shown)


cylc/flow/host_select.py
```diff
@@ -332,11 +344,11 @@ def visit(self, node):
         # variables
         ast.Name, ast.Load, ast.Attribute, ast.Subscript, ast.Index,
         # opers
-        ast.BinOp, ast.operator,
+        ast.BinOp, ast.operator, ast.UnaryOp, ast.unaryop,
```
oliver-sanders, line 347: This adds the required oppererators.

See [ast](https://docs.python.org/3/library/ast.html?highlight=ast#abstract-grammar) for more info.

cylc/flow/scripts/psutil.py
```diff
@@ -48,10 +48,17 @@ def get_option_parser():
...
+    try:
+        methods = [
+            getattr(psutil, key[0])
+            for key in metrics
+        ]
+    except AttributeError as exc:
+        # error obtaining interfaces from psutil e.g:
+        # * requesting a method which does not exist
+        print(exc, file=sys.stderr)
+        sys.exit(2)
```
oliver-sanders, line 61: The `2` return code is used to help differentiate errors in the ranking expression from unexpected errors.

cylc/flow/host_select.py
MetRonnie: What is the difference between `UnaryOp` and `unaryop`?

MetRonnie (approved):
LGTM, I have tested out `-1 * virtual_memory().available`

tests/unit/test_exceptions.py
```diff
@@ -14,9 +14,67 @@
...
+    HostSelectException,
+)
 
 
 def test_cylc_error_str():
     error = CylcError("abcd")
     assert str(error) == "abcd"
+
+
+def test_host_select_exception():
```
wxtim, line 30: IMO this ought to be one test per assert.

tests/unit/test_exceptions.py
```diff
@@ -14,9 +14,67 @@
...
+    assert 'Cylc could not establish SSH connection to the run hosts.' in (
+        str(HostSelectException({
+            'ranking': 'virtual_memory().available > 1',
+            'host-1': {
+                'returncode': 255
+            },
+            'host-2': {
+                'returncode': 255
+            },
+        })))
```
wxtim, line 80:
```suggestion
@pytest.mark.parameterize(
    'ret_code, expect',
    [
        # it should give a useful hint for exit code "2"
        # (error in the selection ranking expression)
        2, 'This is likely an error in the ranking expression',
        # it should give a useful hint for the exit code "255"
        # (ssh error)
        255, 'Cylc could not establish SSH connection to the run hosts.'
    ]
)
def test_host_select_exception_returncodes(ret_code, expect):
    assert expect in (
        str(HostSelectException({
            'ranking': 'virtual_memory().available > 1',
            'host-1': {
                'returncode': ret_code
            },
            'host-2': {
                'returncode': ret_code
            },
        })))

```

wxtim (approved):
Seems legit - I have a suggestion about the test, but it's at the "feel free to reject level"

tests/unit/test_exceptions.py
oliver-sanders: I think it's ok to have multiple asserts in a test, the error comes back with a line number so I don't think there's really anything to gain by splitting them arbitrarily.

wxtim: Happy.

MetRonnie (approved):

Issue #5118, opened 2022-09-05
parameters: complete support for alternative syntax


oliver-sanders:
> [edit] Read down, this syntax was not intended to work, but does work in all other contexts.

Cylc allows us to inherit from parametrised families:

```ini
[runtime]
    [[X<foo>]]
    [[x<foo>]]
        inherit = X<foo>
```

However, for some reason this only works when there is a single parameter involved, this example errors:

```ini
# suite.rc
[cylc]
    [[parameters]]
        foo = 1..2
        bar = 3..4

[scheduling]
    [[dependencies]]
        [[[R1]]]
            graph = x<foo><bar>

[runtime]
    [[X<foo><bar>]]
    [[x<foo><bar>]]
        inherit = X<foo><bar>
```

Reported with Cylc 7.

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

hjoliver:
I think we can close this as invalid, possibly after adding a warning to the documentation and improved validation.

`foo<m><n>` is not documented as [valid syntax](https://cylc.github.io/cylc-doc/stable/html/user-guide/writing-workflows/parameterized-tasks.html) and I'm surprised it works even without inheritance.

It should be `foo<m,n>`, and that does work with inheritance:

```ini
[task parameters]
    m = 1,2
    n = a,b
[scheduling]
    [[graph]]
        R1 = "FOO<m,n>"
[runtime]
    [[FOO<m,n>]]
    [[foo<m,n>]]
        inherit = "FOO<m,n>"
```

![fam](https://user-images.githubusercontent.com/2362137/188566796-83111e65-6e20-4376-a993-63a6a64dd88f.png)

oliver-sanders:
> foo<m><n> is not documented as [valid syntax](https://cylc.github.io/cylc-doc/stable/html/user-guide/writing-workflows/parameterized-tasks.html)

And yet it works fine in `[runtime]` sections. Given this I expected it to work elsewhere.

In this particular example there is a semi-legit use case for separating the two:

```
   abc_<foo>_def_ghi_<bar>
```

oliver-sanders:
I ran a quick scan for `<.*>.*<.*>` and found this syntax is fairly widely used, here are some examples I found:

```
run_pointstat_1<model><fcstr_pt> => check_metlogs_pointstat<model>
[...]

PR #5162, opened 2022-09-27
Parsec: Don't require all items in list to be quoted if one must

### LINKED ISSUE #1520: #1475 user auth (step 2 extended) - user privilege groups
#1480 distinguishes between two user types: "authenticated" (with passphrase) and "public" (no passphrase).  Global/suite config can set the privilege level for public users, and authenticated users get full control privileges.  Public read-only is now an option, but is probably not advisable at a large site because of the potential for hammering suite daemons with GUI info requests.

The new connection validator receives the client user and host names (and UUID) as well as passphrase, so we easily could split both public and authenticated users into multiple groups according to users and/or hosts specified under the new config [authentication] sections.  So for instance, you could allow a specified subset of all users, or of users on a particular host, to have read-only access to a suite, and allow only "identity" information to others.  Or you could give a subset of authenticated (with passphrase) users read-only access.

This would be almost as good as the proposed final "user account" solution (step 3 of #1475) but it could be implemented very quickly (est. 1 hour after #1480) and would give us some freedom to think through the final solution carefully.

### LINKED ISSUE #5120: parsec: comma separated list items containing commas
Spotted in relation to #5118, the "supported" syntax for multiple parameters doesn't play nicely with the inherit configuration either.

For example this workflow:

```ini
[task parameters]                                                                    
    x = 1..2           
    y = 3..4           
                       
[scheduling]           
    [[dependencies]]   
        [[[R1]]]       
            graph = foo
                       
[runtime]              
    [[I, J]]           
    [[A<x,y>]]         
    [[a1<x,y>, a2<x,y>, a3<x,y>]]
        inherit = I, "A<x,y>", J 
```

Looks fine, but doesn't validate:

```
ListValueError: (type=list) [runtime][a1<x,y>, a2<x,y>, a3<x,y>]inherit = I, "A<x,y>", J - (names containing commas must be quoted(e.g. 'foo<m,n>'))
```

The parametrised item *has* been quoted as suggested by the error message, but it still fails validation.

I found a workaround, for some reason if one of the items in the list is quoted, all of the items must be, whether they need to be or not:

```diff
     [[a1<x,y>, a2<x,y>, a3<x,y>]]
-        inherit = I, "A<x,y>", J 
+        inherit = "I", "A<x,y>", "J" 
```

Should be a simple fix, probably in the "coerser".

wxtim:
Closes #5120

**Check List**

  - [x] Unit tests added and refined.
  - [ ] Run `cylc config` against lots of workflows locally


cylc/flow/parsec/validate.py
```diff
@@ -582,27 +585,43 @@ def _unquoted_list_parse(cls, keys, value):
...
+            >>> list(ParsecValidator._unquoted_list_parse(
+            ...     None, 'I, "A<x,y>", J, "<a, b>B"'))
+            ['I', 'A<x,y>', 'J', '<a, b>B']
         """
-        # http://stackoverflow.com/questions/4982531/
-        # how-do-i-split-a-comma-delimited-string-in-python-except-
-        # for-the-commas-that-are
 
-        # First detect multi-parameter lists like <m,n>.
-        if cls._REC_MULTI_PARAM.search(value):
```
wxtim, line 592: Running this regex before attempting to parse the list will mean we either reject or pass the entire list.

cylc/flow/parsec/validate.py
```diff
@@ -582,27 +584,42 @@ def _unquoted_list_parse(cls, keys, value):
         Examples:
             >>> list(ParsecValidator._unquoted_list_parse(None, '"1", 2'))
             ['1', '2']
```
MetRonnie, line 586: 😤
```
>>> list(ParsecValidator._unquoted_list_parse(None, '1,"2",3'))
['1', '2', '3']

>>> list(ParsecValidator._unquoted_list_parse(None, '1, "2", 3'))
['1', '"2"', '3']
```

MetRonnie (commented):
Sorry, these 2 comments are nothing to do with your changes but I have been baffled by these

cylc/flow/parsec/validate.py
hjoliver: We've discussed this somewhere recently, just trying to find evidence...

hjoliver: Probably in relation to #5118

PR #5165, opened 2022-09-27
move references to sub-folders into workflow files

wxtim:
Rationalize where constants are kept. Came out of reviewing #5140 

**Check List**

- [x] Tests not required because there has been not functional change

### DIFF
--- a/cylc/flow/install_plugins/log_vc_info.py
+++ b/cylc/flow/install_plugins/log_vc_info.py
@@ -112,8 +112,6 @@ SVN_INFO_KEYS: List[str] = [
     'revision', 'url', 'working copy root path', 'repository uuid'
 ]
 
-
-LOG_VERSION_DIR = Path(WorkflowFiles.LOG_DIR, 'version')
 DIFF_FILENAME = 'uncommitted.diff'
 INFO_FILENAME = 'vcs.json'
 JSON_INDENT = 4
@@ -265,7 +263,12 @@ def write_vc_info(
     """
     if not info:
         raise ValueError("Nothing to write")
-    info_file = Path(run_dir, LOG_VERSION_DIR, INFO_FILENAME)
+    info_file = Path(
+        run_dir,
+        WorkflowFiles.LogDir.DIRNAME,
+        WorkflowFiles.LogDir.VERSION,
+        INFO_FILENAME
+    )
     info_file.parent.mkdir(exist_ok=True, parents=True)
     with open(info_file, 'w') as f:
         f.write(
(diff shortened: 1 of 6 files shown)


cylc/flow/workflow_files.py
```diff
@@ -193,6 +193,12 @@ class WorkflowFiles:
     LOG_DIR = 'log'
     """Workflow log directory."""
 
+    LOG_INSTALL_DIR = os.path.join(LOG_DIR, 'install')
+    """Install log dir"""
+
+    LOG_VERSION_DIR = os.path.join(LOG_DIR, 'version')
+    """Version control log dir"""
+
```
MetRonnie, line 201: Would it make sense to move these to a subclass `Log` á la `Service` and `Install` below?

MetRonnie (approved):

tests/unit/post_install/test_log_vc_info.py
```diff
@@ -167,7 +166,8 @@ def test_write_diff_git(git_source_repo: Tuple[str, str], tmp_path: Path):
     """Test write_diff() for a git repo"""
     source_dir, _ = git_source_repo
     run_dir = tmp_path / 'run_dir'
-    (run_dir / WorkflowFiles.LOG_DIR).mkdir(parents=True)
+    breakpoint()
```
datamel, line 169: Breakpoint needs removing.
```suggestion
```

datamel (commented):
Looks good, just debug removing needed from me.

MetRonnie (commented):
Got some functional test failures

cylc/flow/install_plugins/log_vc_info.py
```diff
@@ -265,7 +263,8 @@ def write_vc_info(
     """
     if not info:
         raise ValueError("Nothing to write")
-    info_file = Path(run_dir, LOG_VERSION_DIR, INFO_FILENAME)
+    info_file = Path(
+        run_dir, WorkflowFiles.LogDir.VERSION, INFO_FILENAME)
```
datamel, line 267: This isn't the same as the original path. We need `run_dir/log/version`, at the moment it is making `run_dir/version`.

datamel (changes requested):

cylc/flow/install_plugins/log_vc_info.py
wxtim: Thanks Mel - this has turned out to be a bigger refactor than intended

datamel:
Looks ready, is re-review needed @wxtim?

datamel (approved):
👍

PR #5023, opened 2022-07-28
Db store force triggered

hjoliver:
These changes close #5020

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.
- [x] Appropriate tests are included (unit and/or functional).
- [x] Appropriate change log entry included.
- [x] No documentation update required.
- [ ] [if bugfix] PRs raised to both master and the relevant bugfix branch.

PR #5173, opened 2022-10-03
Revert "Db store force triggered (#5023)"

MetRonnie:
This reverts commit 20bc9a097668fd365163a9c87d03b6e5ee3c54c5.

As commented in #5023, it broke the ability to restart workflows run in earlier versions of Cylc 8.0


oliver-sanders:
@hjoliver We need a DB upgrader for this, I don't know if we can reliably determine a value for the missing field. I guess a default of `False` is probably as safe as it was before so good enough?

hjoliver:
Just saw this at end of the day (still catching up post holiday). I'll prioritize it tomorrow.

hjoliver:
Superseded!

PR #5182, opened 2022-10-07
cache get_platform

oliver-sanders:
Spotted that the `get_platform_from_name` function was using a lot of CPU.

<details>
<summary>Turns out that it was called 7662 times for this sample workflow:</summary>

```ini
[task parameters]
    a = 1..5
    b = 1..5
    c = 1..5

[scheduling]
    initial cycle point = 2000
    final cycle point = 20000110T00Z
    runahead limit = P5
    [[queues]]
        [[[default]]]
            limit = 25
    [[graph]]
        P1D = """
            <a> => <b> => <c>
            <b>[-P1D] => <b>
        """

[runtime]
    [[<a>, <b>, <c>]]
        script = true
```
</details>

Embarrassingly this workflow only uses the localhost platform.

Added `lru_cache` to the function to reduce the impact:

Before:
Screenshot from 2022-10-07 15-19-27

After:
Screenshot from 2022-10-07 15-19-38

Caching does have an overhead, but inspecting the call tree it would appear this saved ~3 seconds of a ~35s run, which is ~10% of CPU!

**Check List**

### DIFF
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -166,7 +166,6 @@ def get_platform(
 @lru_cache(100)
 def platform_from_name(
     platform_name: Optional[str] = None,
-    platforms: Optional[Dict[str, Dict[str, Any]]] = None,
     bad_hosts: Optional[Set[str]] = None
 ) -> Dict[str, Any]:
     """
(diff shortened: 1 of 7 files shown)


oliver-sanders:
Dammit the `bad_hosts` make a mess of this.

oliver-sanders:
Closing for now pending two changes:

1. A reduction in the number of `get_platform` calls (done).
2. A refactor of the `get_platform` logic to separate the cachable and non-cachable parts.

PR #5184, opened 2022-10-10
Scan workflow name during install.

### LINKED ISSUE #5127: install: report if previous run is still running
If a previous instance of a workflow is still running when a new one is installed it would be good for `cylc install` to detect this and inform the user.

E.G. something along the lines of:

```
INSTALLED ....

whatever/run1 is still running, to shut it down (killing active tasks)
$ cylc stop whatever/run1 --kill
```

I don't think this should block the new installation, just a quick message after it has completed.

**Pull requests welcome!**

hjoliver:
Close #5127

Scan and report any running instances of a workflow at install time.

Screenshot from 2022-10-10 15-26-43

(Note I've used the WARNING color, but without saying "WARNING:" on grounds that already-running instances of an installed  workflow might possibly indicate user error, but probably do not ... but I'm not particularly invested in this coloring!).

**Check List**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,9 @@ ones in. -->
 
 ### Enhancements
 
+[#5284](https://github.com/cylc/cylc-flow/pull/5184) - scan for active
+workflows at install time.
+
 [#5032](https://github.com/cylc/cylc-flow/pull/5032) - set a default limit of
 100 for the "default" queue.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,8 +16,8 @@ ones in. -->
 
 ### Enhancements
 
-[#5284](https://github.com/cylc/cylc-flow/pull/5184) - scan for active
-workflows at install time.
+[#5184](https://github.com/cylc/cylc-flow/pull/5184) - scan for active
+runs of the same workflow at install time.
 
 [#5032](https://github.com/cylc/cylc-flow/pull/5032) - set a default limit of
 100 for the "default" queue.
--- a/cylc/flow/scripts/install.py
+++ b/cylc/flow/scripts/install.py
@@ -89,11 +89,16 @@ from typing import Optional, Dict, Any
 from cylc.flow.scripts.scan import (
     get_pipe,
     _format_plain,
+    FLOW_STATE_SYMBOLS,
+    FLOW_STATE_CMAP
 )
 from cylc.flow import iter_entry_points
 from cylc.flow.exceptions import PluginError, InputError
 from cylc.flow.loggingutil import CylcLogFormatter
-from cylc.flow.option_parsers import CylcOptionParser as COP
+from cylc.flow.option_parsers import (
+    CylcOptionParser as COP,
+    Options
+)
 from cylc.flow.pathutil import EXPLICIT_RELATIVE_PATH_REGEX, expand_path
 from cylc.flow.workflow_files import (
     install_workflow, search_install_source_dirs, parse_cli_sym_dirs
@@ -155,6 +160,16 @@ def get_option_parser() -> COP:
         default=False,
         dest="no_run_name")
 
+    parser.add_option(
+        "--no-ping",
+        help=(
+            "When scanning for active instances of the workflow, "
+            "do not attempt to contact the schedulers to get status."
+        ),
+        action="store_true",
+        default=False,
+        dest="no_ping")
+
     parser.add_cylc_rose_options()
 
     return parser
@@ -167,7 +182,7 @@ def get_source_location(path: Optional[str]) -> Path:
     """
     if path is None:
         return Path.cwd()
-    path = path.strip()
+    path = str(path).strip()
     expanded_path = Path(expand_path(path))
     if expanded_path.is_absolute():
         return expanded_path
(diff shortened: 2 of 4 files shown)


hjoliver:
(Considering an integration test for this).

oliver-sanders (approved):
LGTM. Perhaps add in a templated `cylc stop` command to stop these additional runs.

```
Stop them with `cylc stop <id> ...`
```

MetRonnie (commented):
Tested, looking good. probably could do with a test as you've said.

Is there an easy way to include the state (running, paused etc) next to the listed workflows? Might help clarify what is meant by "active" in each case

hjoliver:
Integration test added. Workflow status icons added.

shot

Note this method doesn't catch instances installed with `--workflow-name=other`.

cylc/flow/scripts/install.py
```diff
@@ -171,14 +191,76 @@ def get_source_location(path: Optional[str]) -> Path:
...
+    if active:
+        n = len(active)
+        grammar = (
+            ["s", "are", "them"]
+            if n > 1 else
+            ["", "is", "it"]
+        )
+        print(
+            CylcLogFormatter.COLORS['WARNING'].format(
+                f'WARNING: {n} run%s of "{wf_name}"'
```
MetRonnie, line 214: What do you think of "NOTE" instead? Perhaps less aggressive
```suggestion
                f'NOTE: {n} run%s of "{wf_name}"'
```

cylc/flow/scripts/install.py
```diff
@@ -171,14 +191,76 @@ def get_source_location(path: Optional[str]) -> Path:
     return search_install_source_dirs(expanded_path)
 
 
+async def scan(wf_name: str, ping: bool = True) -> None:
+    """Print any instances of wf_name that are already active."""
+    opts = Values({
+        'name': [f'{wf_name}/*'],
```
oliver-sanders, line 197: Might want to restrict this to numbered runs?

```suggestion
        'name': [f'{wf_name}/run*'],
```

I think this approach will actually list all workflows on the filesystem, then filter by name. There is a way to restrict scanning to just the "install dir" we are interested in which should be a good bit faster, will take a look.

tests/integration/test_install.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+SRV_DIR = Path(WorkflowFiles.Service.DIRNAME)
+CONTACT = Path(WorkflowFiles.Service.CONTACT)
+RUN_N = Path(WorkflowFiles.RUN_N)
+INSTALL = Path(WorkflowFiles.Install.DIRNAME)
+
+
+@pytest.fixture()
+def src_run_dirs(mock_glbl_cfg, monkeypatch):
```
MetRonnie, line 41: I think you can use the `tmp_path` fixture to simplify this a bit
```suggestion
def src_run_dirs(mock_glbl_cfg, monkeypatch, tmp_path: Path):
```

tests/integration/test_install.py
```diff
@@ -0,0 +1,88 @@
...
+    Source dirs:
+      <tmp-src>/w1
+      <tmp-src>/w2
+
+    Run dir:
+      <tmp-run>/w1/run1
+
+    """
+    tmp_src_path = Path(TemporaryDirectory().name)
+    tmp_run_path = Path(TemporaryDirectory().name)
```
MetRonnie, line 53:
```suggestion
    tmp_src_path = tmp_path / 'cylc-src'
    tmp_run_path = tmp_path / 'cylc-run'
```

tests/integration/test_install.py
```diff
@@ -0,0 +1,88 @@
...
+        f'''
+            [install]
+                source dirs = {tmp_src_path}
+        '''
+    )
+    monkeypatch.setattr('cylc.flow.pathutil._CYLC_RUN_DIR', tmp_run_path)
+
+    yield tmp_src_path, tmp_run_path
+    rmtree(tmp_src_path)
+    rmtree(tmp_run_path)
```
MetRonnie, line 74:
```suggestion
    return tmp_src_path, tmp_run_path
```

tests/integration/test_install.py
```diff
@@ -0,0 +1,88 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Test cylc install."""
+
+from pathlib import Path
+from shutil import rmtree
+from tempfile import TemporaryDirectory
```
MetRonnie, line 21:
```suggestion
```

MetRonnie (approved):

cylc/flow/scripts/install.py
hjoliver: :+1: yep (I originally left WARNING off, then back tracked!)

cylc/flow/scripts/install.py
hjoliver: Regarding your suggested change, I was trying to catch named runs as well. Is there some reason not to do that?

> There is a way to restrict scanning to just the "install dir" we are interested in which should be a good bit faster, will take a look.

Did you take a look?

oliver-sanders: > Is there some reason not to do that?

I expect having multiple named runs of a workflow running simultaneously to be a likely pattern.

E.G. our post-processing is defined in a single workflow which is generic, when it is installed, it becomes specialised to it's usage via Jinja2. So one workflows, multiple concurrent named runs.

But meh, maybe leave it as is and adapt later if desired.

> Did you take a look?

Use the `scan_dir` argument:

https://github.com/cylc/cylc-flow/blob/79721142e872fb6fde97a79ef7a6f05e9332384c/cylc/flow/network/scan.py#L188-L215

Here's an example of an integration test which uses this argument to scan a single directory:

https://github.com/cylc/cylc-flow/blob/79721142e872fb6fde97a79ef7a6f05e9332384c/tests/integration/test_scan.py#L319-L333

oliver-sanders: Scanning all workflows can get very slow (especially on NFS) so this is quite important for us.

hjoliver: > 13 users with >500 with the prize going to a user with a whopping 2021 workflows. 

:exploding_head: 

Anyhow, great, I'll tweak it to restrict the directory...

hjoliver: > I expect having multiple named runs of a workflow running simultaneously to be a likely pattern.

Me too, which is why I was trying to alert the user to those too.  So I still don't quite understand your point.  Maybe it was that WARNING (which I've backed out of) is too strong when those runs are deliberate?  But the same goes for `runN` runs, in principle.

hjoliver: scan now restricted to the one workflow

tests/integration/test_install.py
```diff
@@ -0,0 +1,140 @@
...
+WF_ACTIVE_MSG = '1 run of "{wf}" is already active:'
+BAD_CONTACT_MSG = "Bad contact file:"
+
+
+class graphql_query_mock:
+    """Context manager to mock network.scan.graphql_query().
+
+    (In the scripts.scan namepace, where it is used below).
+
+    """
```
hjoliver, line 48: I couldn't figure out how to patch a decorated (`@pipe`) async function by standard means, so I resorted to this. Mocking and patching is still a bit of a black art, to me.

hjoliver:
(Tutorial test failure unrelated)

tests/integration/test_install.py
```diff
@@ -0,0 +1,140 @@
...
+
+    opts = InstallOptions()
+    opts.no_ping = True
+
+    install_cli(opts, reg='w1')
+    out = capsys.readouterr().out
+    assert INSTALLED_MSG.format(wfrun='w1/run2') in out
+    assert WF_ACTIVE_MSG.format(wf='w1') in out
+    assert f"{BAD_CONTACT_MSG} w1/run1" in caplog.text
```
MetRonnie, line 112: Why is this bad contact file message expected?

CHANGES.md
```diff
@@ -16,6 +16,9 @@ ones in. -->
 
 ### Enhancements
 
+[#5284](https://github.com/cylc/cylc-flow/pull/5184) - scan for active
+workflows at install time.
```
MetRonnie, line 20: How about
```suggestion
[#5184](https://github.com/cylc/cylc-flow/pull/5184) - scan for active
runs of the same workflow at install time.
```

tests/integration/test_install.py
oliver-sanders: I think we could do it be replacing the function in the place where it is being assembled into a pipe:

```python
def patch_graphql(monkeypatch):
    # define a mocked graphql pipe function
    @pipe
    def _graphql(*,args, **kwargs):
        pass

    # swap the graphql function that cylc.flow.scripts.scan sees for the patched version
    monkeypatch.setattr(
        'cylc.flow.scripts.graphql',
        _graphql,
    )

tests/integration/test_install.py
hjoliver: Because the contact file has no content (it's faked with `touch`). I'll add a comment.

tests/integration/test_install.py
hjoliver: Brilliant, with a couple of tweaks that worked, and it's cleaner than my approach so I've switched over. Cheers.

tests/integration/test_install.py
hjoliver: Done.

MetRonnie (approved):

hjoliver:
Two approvals, going in...

PR #5201, opened 2022-10-17
Make shutdown more robust

MetRonnie:
Fixes a problem where an unexpected exception raised during the scheduler shutdown - before it gets to closing the workflow runtime server thread - would cause Cylc to hang because Python doesn't like that.

**Check List**

- [x] No dep changes
- [x] No tests are included
- [x] No changelog entry as it's pretty hard to hit
- [x] No docs needed

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1731,45 +1731,25 @@ class Scheduler:
 
     async def _shutdown(self, reason: BaseException) -> None:
         """Shutdown the workflow."""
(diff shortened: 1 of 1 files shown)


hjoliver (commented):
LGTM. (Are you planning to add or tweak a test?)

MetRonnie:
I think I won't include a test as this is a bit tricky

oliver-sanders (approved):

hjoliver (approved):

PR #5189, opened 2022-10-11
Get Cylc template variables from database for validation.

wxtim:
Allowed "revalidation" of cylc workflows where template variables are collected from

Probably closes #4740 

database of already played workflows.
- Added `--revaladidate` as an option to the following scripts (Made async to allow testing of changes):
  - cylc validate
  - cylc view
  - cylc graph
  - cylc config

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Tests are included.
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [ ] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.

Issue #3896, opened 2020-10-27
cli: command aliases


oliver-sanders:
Consider supporting aliases for common command combinations e.g:

* install + play
* reinstall + reload

For inspiration look at `git` "transactions" e.g:

* `git pull` (`git fetch` + `git merge`).

And `git` "aliases" e.g:

* `git config --global alias.co checkout`

Also consider ways of exposing these aliases via the UIS so they can be used in the UI as GraphQL mutations.

oliver-sanders:
### Proposal

1. Make `cylc clean` an option of `cylc install`.
   * I.E. `cylc install foo --clean`.
   * Don't support any `cylc clean` options via `cylc install` for now, default behaviour only.
2. Make the "Review" category options of `cylc play`
   * Initially `cylc play --tui` (easy).
   * Near future add `cylc play --gui` (requires UIS work).
   * Don't worry about `cylc review`.
   * These options are incompatible with `--no-detach`.
3. Kick `rose macro` down the line for now, it's tricker to solve.
   * `rose macro` has a fair few options, the command won't necessarily do what you want it to do without specifying some of these.
   * We could allow defaults to be configured in the `rose-suite.conf`.
   * After that we can consider making `rose macro` an option of `cylc validate` via a cylc-rose interface. 

Implement the following commands:

1. `vip`
   * **validate-install-play**
   * Works with source directories in the same way as `cylc install`.
2. `vrp`
   * **validate-reinstall-[play/reload]**
   * Consider calling `vra` (`a` for apply) or something like that to convey the play/reload abstraction.
   * Works with run IDs the same way as `cylc play`.
   * Uses play if the workflow is stopped, uses reload if the workflow is running.
   * Support options of `cylc play` but not `cylc reload` (which doesn't have any options anyway).
3. `sp` (implement in cylc-rose)
   * **stem-play**
   * Works with source directories in the same way as `rose stem`

Intermingle the CLI options e.g:

```bash
# -s goes to validate & play
# --pause goes to play
cylc vip -s ANSWER=42 --pause
```

Attempt to do this in a way which minimises duplication, drop any incompatible arguments. Dump all of the options into a single object and provide it to each of the commands called.

Examples:

```bash
# validate "foo" from "[install]source dirs" then install and play it
$ cylc vip foo

# validate the source of bar/run1 (using CLI options scraped from run dir)
# then reinstall and play/reload it depending on workflow state
cylc vrp "bar/run1"

[...]

PR #5214, opened 2022-10-31
Get template variables set using previous `cylc play`

### LINKED ISSUE #4740: config: template variables specified to play should be loaded
Template variables can be provided to `cylc play`, they are then stored in the DB so they can be used with future restarts.

However, the `cylc config` command is currently unaware of this:

```console
$ cylc install one
$ cylc config
$ cylc play one -s "FOO='bar'"
$ cylc config  # should load FOO='bar' from the workflow DB
```

Note that Cylc-Rose can also be used to specify template variables (at install time), these variables are store in a file which **is loaded** when `cylc config` is invoked.

**Pull requests welcome!**

wxtim:
### Closes #4740 

Partially addresses #3896 

This is an alternative implementation of those features of #5189 which pertain to #4740 - This PR allows cylc validate (and graph, list, view, and config) to use template variables retrieved from the workflow database, should you be running them against locations where a database containing these is available.

Originally I implemented this as a Cylc plugin, but I think it's better like this, because the template variables logic is called at the top level in the CLI scripts making it easier to get the template variables in use in revalidation.

### Milestone towards VRO
This PR also now includes a mechanism to revalidate/validate installed workflows against source in Cylc Validate. This is designed to allow Cylc VRO to validate workflows before re-installation.

**Check List**

- [x] Tests are included.
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) Pull request not opened - will open a single docs PR after VRO is in.
- [x] ~If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.~ Not a bug fix


oliver-sanders:
@wxtim, I think you've won the prize for the most confusing PR title?!

wxtim:
> @wxtim, I think you've won the prize for the most confusing PR title?!

Yes.

oliver-sanders:
(one small conflict)

cylc/flow/pathutil.py
```diff
@@ -456,3 +456,32 @@ def get_workflow_name_from_id(workflow_id: str) -> str:
...
+
+    # Test whether there is a config file at source:
+    if (src / WorkflowFiles.FLOW_FILE).exists():
+        return src.resolve() / WorkflowFiles.FLOW_FILE
+    elif (src / WorkflowFiles.SUITE_RC).exists():
+        return src.resolve() / WorkflowFiles.SUITE_RC
+    else:
+        raise WorkflowFilesError(
+            f'Source file not present at: {src.resolve()}'
+        )
```
oliver-sanders, line 487: There is already a `get_workflow_source_dir(run_dir)` function in `workflow_files`.

Suggest doing:

```python
get_workflow_source_dir(get_workflow_run_dir(workflow_id))[0]
```

cylc/flow/scripts/validate.py
```diff
@@ -37,20 +38,44 @@
...
 from cylc.flow.profiler import Profiler
 from cylc.flow.task_proxy import TaskProxy
 from cylc.flow.templatevars import get_template_vars
 from cylc.flow.terminal import cli_function
 
 
+def _revalidate(
+    cfg, flow_file, options, workflow_id, mem_log_func=None
+):
+    # Check if this is a rundir - revalidate not allowed otherwise:
```
oliver-sanders, line 60: Should rename "revalidate" to something more informative e.g. "validate_against_installation"

cylc/flow/scripts/validate.py
```diff
@@ -80,6 +105,16 @@ def get_option_parser():
         default="live", dest="run_mode",
         choices=['live', 'dummy', 'simulation'])
 
+    parser.add_option(
+        '--reinstall-validate', '--revalidate', '--against-source',
```
oliver-sanders, line 109: The one option should be enough:

```suggestion
        '--against-source',
```

cylc/flow/scripts/validate.py
```diff
@@ -80,6 +105,16 @@ def get_option_parser():
...
+        help=(
+            "Validate as if for reinstallation. If this option is selected"
+            " the argument can only be a workflow id."
+        ),
+        dest='revalidate',
```
oliver-sanders, line 114: `s/revalidate/against_source`

cylc/flow/scripts/view.py
```diff
@@ -25,9 +25,11 @@
   configuration (as Cylc would see it).
 """
 
+import asyncio
+
```
oliver-sanders, line 29:
```suggestion
```

cylc/flow/scripts/validate.py
```diff
@@ -37,20 +38,44 @@
...
+        workflow_id,
+        get_source_conf_from_id(workflow_id),
+        options,
+        get_template_vars(options, flow_file),
+        output_fname=options.output,
+        mem_log_func=mem_log_func
+    )
+
+    # Update the in-place config with source config items.
+    cfg.pcfg.sparse.update(source_cfg.pcfg.sparse)
```
oliver-sanders, line 76: Why is this function implemented differently for `validate` to other commands e.g. `config`?

oliver-sanders (commented):
Got an empty file: tests/unit/scripts/test_validate.py

cylc/flow/scripts/validate.py
wxtim: This PR contains two different, but related things 
- Run Cylc Validate/View/Graph/Config on played workflows, retrieving template variables from database.
- Validate against source.

I should probably add validate against source to all four scripts, but those are not stepping stones to VRO.

datamel (commented):
I have checked this out and run manual and CI tests locally it is looking good to me. My only observation is that I think the newly added scripts integration tests need a teardown added to them as I've some directories left over.

datamel (commented):
I've re-checked this out. Still directories left, maybe your branch needs re-pushing?

wxtim:
> I've re-checked this out. Still directories left, maybe your branch needs re-pushing?

Apparently so!

tests/integration/conftest.py
```diff
@@ -404,3 +420,67 @@ def run_job_cmd(
...
+                },
+                'scheduling': {
+                    'initial cycle point': '1854',
+                    'graph': {
+                        'R1': 'foo'
+                    }
+                }
+            })
+        else:
+            setup.src = flow_src(cfg)
```
oliver-sanders, line 477: For a "default" config tests can use `one_conf` which is a little more explicit.

E.G. here's how we might create a simple workflow where we don't care about the config:

```python
def test_something(flow, one_conf):
    id_ = flow(one_conf)
```

Suggest making `src_flow` work the same way (i.e. making `cfg` mandatory).

tests/integration/conftest.py
```diff
@@ -404,3 +420,67 @@ def run_job_cmd(
...
+
+        # Setup Opts:
+        setup.opts = install_gop().get_default_values()
```
oliver-sanders, line 480: The approach used by other scripts for integration testing is to use `cylc.flow.option_parsers.Options` to create a `SomethingOptions` object.

This handles the default for you and validates that any specified options are valid options.

tests/integration/conftest.py
```diff
@@ -404,3 +420,67 @@ def run_job_cmd(
...
+        setup.opts.no_run_name = True
+        setup.opts.workflow_name = 'cit-' + str(setup.src.name)
```
oliver-sanders: To make this play nicely with the `cit*` directory structure (with it's automatic housekeeping) use the `run_dir` fixture.

E.G:

```python
# this pre-sets the --workflow-name option to install this
# workflow into a cit... run dir, filed under the test name
uuid = ...  # TODO
setup.opts.workflow_name = str(run_dir / uuid)
```

NOTE: I've added a uuid to avoid conflicts in case a test wants to install multiple workflows.

tests/integration/scripts/test_validate_integration.py
```diff
@@ -0,0 +1,97 @@
...
+    _source_workflow, capsys, _setup_validate_cli
+):
+    """Validation fails if revalidating with broken config.
+    """
+    wf = _source_workflow()
+
+    setup = _setup_validate_cli({'against_source': True})
+
+    # Check that the original installation validates OK:
+    await validate(setup.parser, setup.opts, wf.opts.workflow_name)
```
oliver-sanders, line 37: E.G. this could become (DON'T APPLY):

```suggestion
from cylc.flow.parsec.exceptions import IllegalItemError
import pytest
from cylc.flow.parsec.exceptions import Jinja2Error


async def test_revalidate_checks_source(
    _source_workflow, capsys, _setup_validate_cli
):
    """Validation fails if revalidating with broken config.
    """
    wf = _source_workflow()
    await validate(fs.path)
```

oliver-sanders (commented):
Few comments on the integration test harness.

I think it should be easy to get it to install into the same `cit*` directory structure as the `flow` fixture installs into. This would make life a lot easier, especially as the installed flows would come under the regular housekeeping so be left if the test fails, or be removed if it passes, recursively.

tests/integration/conftest.py
```diff
@@ -421,70 +413,46 @@ def run_job_cmd(
...
-        # Carry out the installation:
-        install(setup.opts, setup.src)
-        setup.flow_file = setup.src / 'flow.cylc'
-        return setup
 
+@pytest.fixture
+def cylc_install(test_dir, run_dir):
+    """Install a workflow from source"""
+    def _inner(source, **kwargs):
+        Opts = Options(install_gop())
```
oliver-sanders, line 452: This will create `Opts` every time that `_inner` is called, which could be a lot.

Should create a constant to hold this value so it doesn't have to be re-computed e.g:

```python
InstallOptions = Options(install_gop())

...

@pytest.fixture
def cylc_install(test_dir, run_dir):
    ...
```

tests/integration/utils/flow_tools.py
```diff
@@ -39,15 +39,27 @@
...
+        conf = {
+            'scheduler': {
+                'allow implicit tasks': True
+            },
+            'scheduling': {
+                'graph': {
+                    'R1': 'one'
+                }
+            }
+        }
```
oliver-sanders, line 54: This is a copy of `one_conf`?

I don't think we need an implicit configuration?

tests/integration/conftest.py
```diff
@@ -421,70 +413,46 @@ def run_job_cmd(
...
-        setup.parser = validate_gop()
-        setup.opts = setup.parser.get_default_values()
-        setup.opts.templatevars = []
-        setup.opts.templatevars_file = []
-        if setup.opts is not None:
-            for key, val in opts.items():
-                setattr(setup.opts, key, val)
-        return setup
+@pytest.fixture(scope='module')
+def mod_workflow_source(mod_flow, mod_one_conf, tmp_path_factory):
```
oliver-sanders, line 417: `mod_one_conf` not used.

tests/integration/conftest.py
```diff
@@ -421,70 +413,46 @@ def run_job_cmd(
...
+        return src_dir
     yield _inner
 
 
 @pytest.fixture
-def source_workflow(run_dir, flow_src):
-    """Create a workflow in source directory and install it.
-
-    Useful for testing interaction between installed and source directories.
+def workflow_source(mod_flow, mod_one_conf, tmp_path):
```
oliver-sanders, line 433: `mod_one_conf` not used.

tests/integration/scripts/test_validate_integration.py
```diff
@@ -25,34 +25,32 @@
...
-    _source_workflow, capsys, _setup_validate_cli
+    capsys, validate, workflow_source, cylc_install
 ):
     """Validation fails if revalidating with broken config.
     """
-    wf = _source_workflow()
-
-    setup = _setup_validate_cli({'against_source': True})
+    src_dir = workflow_source()
+    run_dir = cylc_install(src_dir)
```
oliver-sanders, line 33: I think it might make more sense to return the workflow ID rather than the workflow run-dir from `cylc_install`.

That way `cylc_install` matches the return value of `flow` and the `id` is the thing you need to provide to Cylc commands e.g. `cylc play`.

tests/integration/conftest.py
```diff
@@ -421,70 +413,46 @@ def run_job_cmd(
...
+        return src_dir
+    yield _inner
 
-        # Carry out the installation:
-        install(setup.opts, setup.src)
-        setup.flow_file = setup.src / 'flow.cylc'
-        return setup
 
+@pytest.fixture
+def cylc_install(test_dir, run_dir):
```
oliver-sanders, line 449: Can drop the `cylc_` prefix (we don't add this for other fixures e.g. `validate`)?

oliver-sanders:
Test harness looking much simplified, thanks.

tests/integration/conftest.py
```diff
@@ -421,70 +413,46 @@ def run_job_cmd(
...
-        setup = SimpleNamespace()
-        setup.parser = validate_gop()
-        setup.opts = setup.parser.get_default_values()
-        setup.opts.templatevars = []
-        setup.opts.templatevars_file = []
-        if setup.opts is not None:
-            for key, val in opts.items():
-                setattr(setup.opts, key, val)
-        return setup
+@pytest.fixture(scope='module')
```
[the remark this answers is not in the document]
oliver-sanders: Might come in handy for `cylc install` integration tests.

oliver-sanders:
> 98.79% of diff hit

:+1:

tests/integration/conftest.py
```diff
@@ -439,20 +442,19 @@ def workflow_source(mod_flow, mod_one_conf, tmp_path):
...
     def _inner(source, **kwargs):
-        Opts = Options(install_gop())
-        opts = Opts(**kwargs)
+        opts = InstallOpts(**kwargs)
         opts.workflow_name = str(test_dir.relative_to(run_dir))
-        install(opts, str(source))
+        workflow_id = cylc_install(opts, str(source))
 
-        return test_dir
+        return workflow_id
```
oliver-sanders, line 459: I think this will break if you try to install two workflows in the same test (as they will try to install into the same run dir).

I previously suggested adding a UUID to the workflow ID. On reflection it would probably make more sense to just add `basename source` onto the end something like this:

```python
opts.workflow_name = str(test_dir.relative_to(run_dir) / run_dir.name)
```

E.G. consider the test:

```python
def test_scan(src_dir, one_conf, install):
    foo_src = src_dir(one_conf)
    bar_src = src_dir(one_conf)
    foo_id = install(foo)
    bar_id = install(bar)
    assert listify(scan()) == {foo_id, bar_id}
```

tests/integration/scripts/test_validate_integration.py
```diff
@@ -25,27 +25,27 @@
...
+    src_dir = workflow_source(one_conf)
+    workflow_id = install(src_dir)
 
     # Check that the original installation validates OK:
-    validate(run_dir, against_source=True)
+    validate(workflow_id, against_source=True)
 
     # Break the source config:
     with open(src_dir / 'flow.cylc', 'a') as handle:
         handle.write('\n[runtime]\n[[foo]]\nAgrajag = bowl of petunias')
```
oliver-sanders, line 40: How true, how very true.

tests/integration/scripts/test_validate_integration.py
```diff
@@ -68,25 +68,26 @@ async def test_revalidate_gets_old_tvars(
...
+    installed_dir = run_dir / wf_id / 'runN'
 
     # Check that the original installation validates OK:
-    validate(wf_run_dir, against_source=True)
+    validate(installed_dir)
 
     # # Start a scheduler with tvars option:
     schd = scheduler(
-        str(wf_run_dir.relative_to(run_dir)),
+        wf_id + '/runN',
```
oliver-sanders, line 79: I think `install` should return the full workflow ID (which would include the run number) to save tests having to manually bodge it like this.

oliver-sanders (commented):
LGTM, just need to do a bit of testing.

cylc/flow/scripts/validate.py
```diff
@@ -89,8 +93,14 @@
...
     ),
+    OptionSettings(
+        ["-u", "--run-mode"], help="Validate for run mode.", action="store",
+        default="live", dest="run_mode",
+        choices=['live', 'dummy', 'simulation']
+    ),
     VALIDATE_RUN_MODE,
-    VALIDATE_ICP_OPTION
+    VALIDATE_ICP_OPTION,
+    VALIDATE_AGAINST_SOURCE_OPTION,
```
datamel, line 103: Not sure this is playing nicely with vip. I'm getting `ParsecError: Cannot validate .../cylc-src/workflow-name against source: it is not an installed workflow.` running vip with this `--against-source` option. Should we be removing it from the vip command?

wxtim: Probably worth a check - the against-source option is fundamentally antithetical to VIP. We should probably catch this error earlier.

cylc/flow/parsec/fileparse.py
```diff
@@ -343,6 +347,43 @@ def merge_template_vars(
...
+    sourcedir.
+    Else return fpath unchanged.
+
+    Raises:
+        ParsecError: If validate against source is set and this is not
+        an installed workflow then we don't want to continue.
+    """
+    if getattr(opts, 'against_source', False):
+        thispath = Path(fpath)
+        source_dir = get_workflow_source_dir(thispath.parent)[0]
```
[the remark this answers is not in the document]
wxtim: This works, and is expected to. `parent` is chopping off the `flow.cylc`, not the `runX` from the given for a path, and `get_workflow_source_dir` is designed to handle either case.

cylc/flow/scripts/validate.py
```diff
@@ -89,8 +93,14 @@
         dest="profile_mode",
         sources={'validate'}
     ),
+    OptionSettings(
+        ["-u", "--run-mode"], help="Validate for run mode.", action="store",
+        default="live", dest="run_mode",
+        choices=['live', 'dummy', 'simulation']
+    ),
```
oliver-sanders, line 100: Now `cylc validate` has two copies of this option :confused:

```console
$ cylc val --help
...
  -u RUN_MODE, --run-mode=RUN_MODE
                        Validate for run mode.
  -m STRING, --mode=STRING
                        Run mode: live, dummy, simulation (default live).
...
```

> **Context:** I think you were changing this because the previous value conflicted with another option?

wxtim: Yes, that's wrong! Merge/rebase error.

cylc/flow/templatevars.py
```diff
@@ -17,11 +17,39 @@
...
+        them to self.template_vars.
+        """
+        self.template_vars[row[0]] = eval_var(row[1])
+
+    def _get_db_template_vars(self):
+        if (self.run_dir / self.DB).exists():
+            dao = CylcWorkflowDAO(str(self.run_dir / self.DB))
+            dao.select_workflow_template_vars(self._callback)
+
+
```
oliver-sanders, line 52: I think this is another rebase error, this was replaced by `_prepend_old_templatevars`.

cylc/flow/parsec/fileparse.py
```diff
@@ -343,6 +347,43 @@ def merge_template_vars(
...
+
+    Args:
+        fpath: filepath of workflow config file.
+        template_vars: Template vars to prepend old tvars to.
+    """
+    rundir = Path(fpath).parent
+    old_tvars = OldTemplateVars(rundir).template_vars
+    old_tvars.update(template_vars)
+    template_vars = old_tvars
+    return template_vars
```
oliver-sanders, line 361: Put this in [cylc/flow/templatevars.py](https://github.com/cylc/cylc-flow/pull/5214/files#diff-489533e95911c3f09c836f60cfb0a618358460ddd1b626b91e154d5b6789b986)?

oliver-sanders (commented):
I think there's been a branch mixup here, this branch contains a mixture of changes before/after review.

Need to go through the diff, pull out the old bits and make sure the new bits are all there. I've flagged the bits I've noticed...

This bit has disappeared:

https://github.com/cylc/cylc-flow/compare/5d89fdc91b271ebf01f478f2df28ea607f4fcb17..582956a65f2bdf31f8ab40730930a62228f6191d#diff-d975bf659606195d2165918f93e1cf680ef68ea3c9cab994f033705fea8238b2R16-R21

wxtim:
I've branched at https://github.com/cylc/cylc-flow/commit/360349697b01f151b72e8fe3d368d5490d368ec3 and manually re-instated the subsequent response to review. Hopefully cleaner now.

foo
```diff
@@ -0,0 +1,2905 @@
+diff --git a/.github/workflows/2_auto_publish_release.yml b/.github/workflows/2_auto_publish_release.yml
```
datamel, line 1: I think this file needs deleting.

wxtim:
Rolled into #5229

PR #5219, opened 2022-11-08
Allow for MyPy making --no-implicit-optional default

wxtim:
Fixes MyPy failing on master because `--no-implicit-options=True` is not default. See https://github.com/python/mypy/pull/13401

I think that this is a single reviewer change.

**Check List**

- [x] Change to code to handle update to MyPy - testing not required..
- [x] `CHANGES.md`/ Documentation - not edited - not user facing

### DIFF
--- a/cylc/flow/task_state.py
+++ b/cylc/flow/task_state.py
@@ -17,7 +17,7 @@
 """Task state related logic."""
 
 
-from typing import List
+from typing import List, Optional
 from cylc.flow.prerequisite import Prerequisite
 from cylc.flow.task_outputs import (
     TaskOutputs,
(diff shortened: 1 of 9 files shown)


cylc/flow/task_state.py
```diff
@@ -231,8 +231,8 @@ def __init__(self, tdef, point, status, is_held):
         self._suicide_is_satisfied = None
 
         # Prerequisites.
-        self.prerequisites: List[Prerequisite] = []
-        self.suicide_prerequisites: List[Prerequisite] = []
+        self.prerequisites: [List[Prerequisite]] = []
+        self.suicide_prerequisites: Optional[List[Prerequisite]] = []
```
MetRonnie, line 235: I think these are spurious

MetRonnie (changes requested):

cylc/flow/task_state.py
wxtim:
```suggestion
        self.prerequisites: List[Prerequisite] = []
        self.suicide_prerequisites: List[Prerequisite] = []
```

wxtim:
```suggestion
        self.prerequisites: List[Prerequisite] = []
        self.suicide_prerequisites: List[Prerequisite] = []
```

PR #5224, opened 2022-11-14
install: disallow reserved names only at top level source dir

### LINKED ISSUE #5223: Cylc install does not copy file called log in bin
**Describe the bug**
When trying to run a program called `log` I get a "no such program" error. Upon investigation I found that the file had *not* been copied from `cylc-src/workflow/bin` to `cylc-run/workflow/run1/bin` when I ran `cylc install workflow`. Files with other names were copied just fine.

Cylc 8.0.3 installed at Met Office.

Make a test workflow that runs a program called "log".
Minimal flow.cylc:
```cylc
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    final cycle point = 1
    [[graph]]
        R1 = """
            test_log
        """
[runtime]
    [[test_log]]
        script = "log"
```
bin/log: (Make executable)
```bash
#! /bin/bash
echo "Test"
```

**Expected behaviour**
The `log` file to be copied, and for the workflow to run the program.

**Additional context**
Is this an intentional design feature to avoid copying logs? If so it should probably be documented somewhere.

hjoliver:
Close #5223 

(Title is description)

**Check List**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,9 @@ Maintenance release.
 
 ### Fixes
 
+[#5224](https://github.com/cylc/cylc-flow/pull/5225) - workflow installation:
+disallow reserved names only in the top level source directory.
+
 [#5211](https://github.com/cylc/cylc-flow/pull/5211) - Provide better
 explanation of failure if `icp = next (T-02, T-32)` when list should be
 semicolon separated.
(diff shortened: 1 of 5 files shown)


wxtim (approved):
- [x] Code is sane
- [x] Tried out locally.
- [ ] 1 test fails on CI - kicking it.

datamel (approved):
I have reproduced the original issue and tested that this fixes it, which it does. 
The CI failure is not related, codecov upload problem, so merging with two approvals.

PR #5230, opened 2022-11-16
completion: increase timeout

oliver-sanders:
A couple of small things:

* Increase the timeout of the completion server from 5 to 15 mins.
  * Server timeouts are a little annoying when they happen so often.
  * Could potentially increase further? We can always change this later (no reinstallation required).
* ~Rename "cat-log" to "log":
  * ~"log" used to be an alias for "cat-log".~
  * ~But "log" seems clearer and is easier to type.~
  * ~Maintain "cat-log" as an alias.~

~Will `s/cat-log/log` in the docs if approved.~

**I have bailed out of the `cat-log` change as the way aliases are handled changed whilst this PR was under review. Can come back to that later**

**Check List**

### DIFF
--- a/cylc/flow/scripts/completion_server.py
+++ b/cylc/flow/scripts/completion_server.py
@@ -647,7 +647,7 @@ def get_option_parser() -> COP:
     parser.add_option(
         '--timeout',
         type='int',
-        default='300',
+        default=900,  # PT15M
         help=(
             'The maximum idle time before the server shuts down in seconds.'
         )


hjoliver:
Re: `cat-log`

The name of course referenced the unix `cat` command, which is overwhelming used to print a single file to stdout rather than concatenate multiple files. However, I'm fine with `log`, makes sense.

hjoliver (approved):

wxtim (approved):
LGTM

I took the liberty of fixing the confiict -feel free to merge @oliver-sanders

oliver-sanders:
I'm hanging on to this one for a sec because:

1. I think you may have changed the alias logic with the `cylc vip` PR which would mean that `cylc cat` would no longer be a valid alias for `cylc cat-log` which could trip up users. If so would need to put an explicit alias in for that.
2. Need to quickly `s/cat-log/log/` in the code examples and in `cylc-doc`.

Issue #5236, opened 2022-11-22
7-8 `cylc suite/workflow-state` compat


hjoliver:
Via @dwsutherland today:

The DB polling commands don't work in either direction, and they'll need to as sites transition between Cylc 7 and 8.

(I reported that they did work some time ago, but it seems I had only tested the most basic full-state query)

hjoliver:
8.1.0
and possibly another Cylc 7 release, although I suppose we could document how to patch the last one for those that need it.

dwsutherland:
Thanks Hilary, I've had a look at both 👍  .. Let me know when you'd like them reviewed

MetRonnie:
Closed by #5237

PR #5237, opened 2022-11-23
workflow-state command back compat.

hjoliver:
Address #5236 
Sibling of #5238 

**Check List**

### DIFF
--- a/cylc/flow/rundb.py
+++ b/cylc/flow/rundb.py
@@ -179,6 +179,7 @@ class CylcWorkflowDAO:
     TABLE_BROADCAST_STATES = "broadcast_states"
     TABLE_INHERITANCE = "inheritance"
     TABLE_WORKFLOW_PARAMS = "workflow_params"
+    TABLE_SUITE_PARAMS = "suite_params"  # back compat
     TABLE_WORKFLOW_FLOWS = "workflow_flows"
     TABLE_WORKFLOW_TEMPLATE_VARS = "workflow_template_vars"
     TABLE_TASK_JOBS = "task_jobs"
--- a/cylc/flow/rundb.py
+++ b/cylc/flow/rundb.py
@@ -179,7 +179,13 @@ class CylcWorkflowDAO:
     TABLE_BROADCAST_STATES = "broadcast_states"
     TABLE_INHERITANCE = "inheritance"
     TABLE_WORKFLOW_PARAMS = "workflow_params"
-    TABLE_SUITE_PARAMS = "suite_params"  # back compat
+    # BACK COMPAT: suite_params
+    # This Cylc 7 DB table is needed to allow workflow-state
+    # xtriggers (and the `cylc workflow-state` command) to
+    # work with Cylc 7 workflows.
+    # url: https://github.com/cylc/cylc-flow/issues/5236
+    # remove at: 8.x
+    TABLE_SUITE_PARAMS = "suite_params"
     TABLE_WORKFLOW_FLOWS = "workflow_flows"
     TABLE_WORKFLOW_TEMPLATE_VARS = "workflow_template_vars"
     TABLE_TASK_JOBS = "task_jobs"
--- a/cylc/flow/scripts/workflow_state.py
+++ b/cylc/flow/scripts/workflow_state.py
@@ -117,8 +117,11 @@ class WorkflowPoller(Poller):
         if connected and self.args['cycle']:
             try:
                 fmt = self.checker.get_remote_point_format()
-            except sqlite3.OperationalError:
-                fmt = self.checker.get_remote_point_format_back_compat()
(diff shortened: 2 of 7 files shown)


cylc/flow/scripts/workflow_state.py
```diff
@@ -115,7 +115,10 @@ def connect(self):
             sys.stderr.write('\n')
 
         if connected and self.args['cycle']:
-            fmt = self.checker.get_remote_point_format()
+            try:
+                fmt = self.checker.get_remote_point_format()
+            except sqlite3.OperationalError:
+                fmt = self.checker.get_remote_point_format_back_compat()
```
MetRonnie, line 121: Perhaps
```suggestion
            except sqlite3.OperationalError as exc:
                try:
                    fmt = self.checker.get_remote_point_format_back_compat()
                except sqlite3.OperationalError:
                    raise exc  # original error
```

cylc/flow/rundb.py
```diff
@@ -179,6 +179,7 @@ class CylcWorkflowDAO:
     TABLE_BROADCAST_STATES = "broadcast_states"
     TABLE_INHERITANCE = "inheritance"
     TABLE_WORKFLOW_PARAMS = "workflow_params"
+    TABLE_SUITE_PARAMS = "suite_params"  # back compat
```
MetRonnie, line 182: Consider moving this to end and adding a comment like
```python
# BACK COMPAT: suite_params
# This Cylc 7 DB table is needed to allow workflow-state
# xtriggers (and the `cylc workflow-state` command) to
# work with Cylc 7 workflows.
# url: https://github.com/cylc/cylc-flow/issues/5236
# remove at: 8.x
TABLE_SUITE_PARAMS = "suite_params"
```

hjoliver:
Codecov patch fail is because I'm not testing the `cylc workflow-state` script directly. I've added a unit test for the underlying functionality though; I think that's sufficient in this case (making a functional test for the script, with old and new DBs, seems more painful than it's worth).

cylc/flow/rundb.py
hjoliver: Done.

hjoliver:
(Reviewers: last commit is comment-only, so skip-ci added. All tests passed previously except slightly low on patch coverage as per my previous comment).

hjoliver:
Actually, I'll re-base this to 8.x. It's arguably a bug fix, and needed ASAP at NIWA.

MetRonnie (approved):
LGTM but haven't tested

dwsutherland (approved):
Changes look good to me 👍 
(I would have to build an older VM or use work machines to test)

dwsutherland:
The sibling is probably equally important (as some cylc8 workflows may be upstream of cylc7 on migration).

MetRonnie (changes requested):
Ok managed to test out but it's not working. And the error message only shows up in `--debug` 

```
DEBUG - [xtrigger-func cmd] cylc function-run workflow_state '[]' '{"workflow": "cylc7-complex", "task": "brief_wagtail_forecast_landsim", "point": "20150219T0000Z"}' ~/cylc-
    run/workflow-state
    [xtrigger-func ret_code] 1
    [xtrigger-func err]
      File "~/.conda/envs/cylc8/bin/cylc", line 8, in <module>
        sys.exit(main())
      File "~/cylc-flow/cylc/flow/scripts/cylc.py", line 660, in main
      File "~/cylc-flow/cylc/flow/scripts/cylc.py", line 286, in execute_cmd
        entry_point.resolve()(*args)
      File "~/cylc-flow/cylc/flow/scripts/function_run.py", line 41, in main
        run_function(args[1], args[2], args[3], args[4])
      File "~/cylc-flow/cylc/flow/subprocpool.py", line 115, in run_function
        res = func(*func_args, **func_kwargs)
      File "~/cylc-flow/cylc/flow/xtriggers/workflow_state.py", line 93, in workflow_state
        fmt = checker.get_remote_point_format()
      File "~/cylc-flow/cylc/flow/dbstatecheck.py", line 80, in get_remote_point_format
        for row in self.conn.execute(
    sqlite3.OperationalError: no such table: workflow_params
```

MetRonnie (commented):
I've pushed a fix. May be sensible to add a functional test with a minimal Cylc 7 workflow DB, I can have a look at doing this if you're short on time

hjoliver:
> I've pushed a fix. May be sensible to add a functional test with a minimal Cylc 7 workflow DB, I can have a look at doing this if you're short on time

Argh, last minute non-careful function name change by me, damn it. Thanks for the fix.

MetRonnie (approved):

MetRonnie:
@hjoliver Do you want to merge if happy with the test I've added

hjoliver:
Yes, very happy - thanks @MetRonnie

PR #5238, opened 2022-11-23
suite-state command forward compat.

hjoliver:
Address #5236 
Sibling of #5237 

**Check List**

### DIFF
--- a/lib/cylc/dbstatecheck.py
+++ b/lib/cylc/dbstatecheck.py
@@ -67,6 +67,14 @@ class CylcSuiteDBChecker(object):
                 ['cycle_point_format']):
             return row[0]
 
+    def get_remote_point_format_forward_compat(self):
+        """Query a remote suite database for a 'cycle point format' entry"""
+        for row in self.conn.execute(
+                r"SELECT value FROM " + CylcSuiteDAO.TABLE_WORKFLOW_PARAMS +
+                r" WHERE key==?",
+                ['cycle_point_format']):
+            return row[0]
+
     def state_lookup(self, state):
         """allows for multiple states to be searched via a status alias"""
         if state in self.STATE_ALIASES:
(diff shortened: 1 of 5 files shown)


hjoliver:
Not sure if we really need this one (it'd be nice to avoid yet another final Cylc 7 release). 

#5237 is the critical one, for Cylc 8 research users triggering tasks off of Cylc 7 operational suites that haven't been migrated yet.

dwsutherland:
For operations, this is important for migration, otherwise we'll be forced to upgrade downstream first.. Which we could do I suppose...

hjoliver:
Can you give this a quick review and test @dwsutherland ? 

(Since we need it soon!)

hjoliver:
(New tests are not needed at this point, for Cylc 7 IMO)

MetRonnie (changes requested):
Missing the xtrigger part here?
https://github.com/cylc/cylc-flow/blob/e60e8f7a6d77256af0e312237c6c8f7bf4fd38f8/lib/cylc/xtriggers/suite_state.py#L49

hjoliver:
Good spot @MetRonnie, thanks. 

Turns out I hadn't quite finished this off - another change was need to handled the change in task_outputs table format too.  (Well, 11 months is a long time...)

hjoliver:
Testing:

Run with Cylc 8:
```ini
[scheduler]
    cycle point format = %Y
[scheduling]
    initial cycle point = 2010
    final cycle point = 2012
   [[graph]]
      P1Y = "foo"
[runtime]
   [[foo]]
      script = "cylc message x"
      [[[outputs]]]
          x = x
```
Now with Cylc 7, both of these work:
```console
cylc-7 $ cylc suite-state c8/runN --task=foo --point=2011 --message=x
polling for 'output: x': satisfied

cylc-7 $ cylc suite-state c8/runN --task=foo --point=2011 --status=succeeded
polling for 'succeeded': satisfied
```

And to test xtriggers polling the above Cylc 8 workflow:
```ini
# suite.rc
[cylc]
  cycle point format = %Y
[scheduling]
    initial cycle point = 2010
    final cycle point = 2012
    [[xtriggers]]
         upstream = suite_state(suite=rtx/c8/runN, task=foo, point=%(point)s, \
            message=x):PT10S
   [[dependencies]]
        [[[P1Y]]]
           graph = "@upstream => foo"
[runtime]
    [[foo]]
```

hjoliver:
Should be good to go now.

lib/cylc/dbstatecheck.py
```diff
@@ -123,11 +131,12 @@ def task_state_met(self, task, cycle, status=None, message=None):
...
-                for value in json.loads(outputs_str).values():
-                    if message == value:
-                        return True
-            return False
+            # This works for Cylc 7 and Cylc 8 outputs table format:
+            return any(
+                message == value
+                for outputs_str, in res
+                for value in json.loads(outputs_str)
+            )
```
MetRonnie, line 139: Are you saying removing the `.values()` makes it work for Cylc 8 DBs? Surely that breaks Cylc 7?

MetRonnie: Or was that just a mistake?
```suggestion
            # This works for Cylc 7 and Cylc 8 outputs table format:
            return any(
                message == value
                for outputs_str, in res
                for value in json.loads(outputs_str).values()
            )
```

hjoliver: Damn it. I owe you a chocolate fish 🐟 

It wasn't a mistake - I took it from master, and I tested it. 
However, I tested it with our classic shortcut `x = x` outputs - i.e. with message and label the same 🤦 

So this is a bug on master too. Patch coming...

hjoliver:
@MetRonnie - now it's done!!

lib/cylc/dbstatecheck.py
MetRonnie: > So this is a bug on master too.

You mean, only a bug on master/8.x? (Though nearly introduced to Cylc 7 in this PR)

MetRonnie (approved):
Tested

PR #5245, opened 2022-11-28
config: descendants list -> set

oliver-sanders:
More low hanging fruit spotted whilst profiling a large workflow.

If you want a unique list, so long as order doesn't matter it's faster to use a set because the lookup time is so quick.

* Change the type of `descendants[<namespace>]` and `first-parent descendants[<namespace>]` fields from `list` to `set`.
* Set has faster lookup `O(1)` which makes additions faster (because the checking is faster).
* Lookups are common with these sets in `cylc.flow.config` so this may speed up a few methods.
* The method which gains the most is `compute_family_tree` where I've seen a reduction of ~34s down to ~1.75s for a workflow with a large number of task proxies.

**Check List**

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -284,10 +284,10 @@ class WorkflowConfig:
             'linearized ancestors': {},
             # lists of first-parent ancestor namespaces
             'first-parent ancestors': {},
-            # lists of all descendant namespaces
+            # sets of all descendant namespaces
             # (not including the final tasks)
             'descendants': {},
-            # lists of all descendant namespaces from the first-parent
+            # sets of all descendant namespaces from the first-parent
             # hierarchy (first parents are collapsible in visualization)
             'first-parent descendants': {},
         }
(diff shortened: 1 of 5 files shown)


oliver-sanders:
Patching a couple of tests which were sensitive to iteration order.

oliver-sanders:
Needed to sort results for a couple of DB tests. Turns out that `descendants` iteration order impacts DB insertion order, I think due to `descendants` iteration in the queueing logic?

DB insertion/listing order is stable but unpredictable (as well as irrelevant) so we should probably run these things through sort anyway.

cylc/flow/config.py
```diff
@@ -1179,16 +1179,12 @@ def compute_family_tree(self):
         for name in self.cfg['runtime']:
             ancestors = self.runtime['linearized ancestors'][name]
             for p in ancestors[1:]:
-                if p not in self.runtime['descendants']:
-                    self.runtime['descendants'][p] = []
-                if name not in self.runtime['descendants'][p]:
-                    self.runtime['descendants'][p].append(name)
+                self.runtime['descendants'].setdefault(p, set()).add(name)
```
wxtim, line 1182: I probably spent too much time looking into this... but speedup looks impressive : https://github.com/wxtim/Notes/blob/master/Olivers_opt.ipynb

wxtim (approved):

MetRonnie (approved):

tests/functional/triggering/16-fam-expansion.t
```diff
@@ -31,7 +31,7 @@ workflow_run_ok "${TEST_NAME}" \
     cylc play --debug --no-detach --set="SHOW_OUT='$SHOW_OUT'" "${WORKFLOW_NAME}"
 #-------------------------------------------------------------------------------
 contains_ok "$SHOW_OUT" <<'__SHOW_DUMP__'
-  + (((1 | 0) & (3 | 2) & (5 | 4)) & (0 | 2 | 4))
+  + (((3 | 2) & (5 | 4) & (1 | 0)) & (2 | 4 | 0))
```
MetRonnie, line 34: Ah wait, it seems this order is not guaranteed?

oliver-sanders: On it.

oliver-sanders: This should really be integration tested which would handing unstable sorting easier.

I've added a sort to stabilise family iteration order #5279 to sort these functional tests out for now.

Issue #5235, opened 2022-11-21
CLI via Hub/proxy and/or UIS


dwsutherland:
Long awaited, and relates to:
#2123
#3005
#3528
#4742

The hub started UIS negotiates an API authentication token that can be used via the hub proxy or directly with the UIS, and in the absence of the hub (`cylc gui`) the UIS generates it's own token.

The scheduler currently has two comms methods `tcp` and `ssh`, and now that the client has been abstracted somewhat (#4742), we could add an `http` option/client that picks up the API token. All the CLI GraphQL queries/mutations should work as is.

Tasks:
- [ ] - `cylc-uiserver`, make the API info (token, URL ...etc) available to the user (read-only)
- [ ] - `cylc-flow`,  create a new client and comms method to send commands via the Hub/UIS using this API info.

**Sample**
(from the UIS task)
```
#!/usr/bin/env python3

import json
import requests

from cylc.uiserver.app import API_INFO_FILE

f = open(API_INFO_FILE, "r")
api_info = json.loads(f.read())
f.close()

query = '''
query {
  workflows {
    id
    stateTotals
  }
}
'''

r = requests.post(api_info["url"] + 'cylc/graphql',
    headers={
        'Authorization': f'token {api_info["token"]}',
    },
    json={'query': query}
)


r.raise_for_status()
data = r.json()

print(json.dumps(data, indent=4))

```
```
$ ./uis_api.py
{
    "data": {
        "workflows": [
            {
                "id": "~sutherlander/linear/run1",
                "stateTotals": {
                    "waiting": 2,
                    "expired": 0,
                    "preparing": 0,
                    "submit-failed": 1,
                    "submitted": 0,
                    "running": 0,
                    "failed": 0,
                    "succeeded": 0
                }
            }
        ]
    }
}
```

dwsutherland:
going via the UIS:

```
(uiserver) sutherlander@graphic-vbox:~$ cylc ping --debug linear
2022-12-07T21:18:59+13:00 DEBUG - Loading site/user config
    files
2022-12-07T21:18:59+13:00 DEBUG - Reading file
    /home/sutherlander/.cylc/flow/8.1.0.dev/global.cylc
2022-12-07T21:18:59+13:00 DEBUG - http:send {'command':
    'graphql', 'args': {'request_string': '\nquery ($wFlows: [ID]) {\n
    workflows(ids: $wFlows) {\n    id\n    name\n    port\n    pubPort\n
[...]

Issue #5265, opened 2022-12-07
client: comms method configuration


oliver-sanders:
Currently the communications method can be configured for job -> scheduler communication, however, this config does not apply to any other connections.

> **How it works:** When a job is submitted the comms method for that platform is extracted from the global config on the Scheduler host and written into the job script as an environment variable. When we get the communication client instance this environment variable is checked and an instance of the relevant client implementation is returned.

### The problem

So currently we can configure the comms method for:

* job-scheduler

But not for:

1. local -> scheduler (shared filesystem to scheduler e.g. `cylc trigger`)
2. remote -> scheduler (remote job platform to scheduler but not from a job script)
3. really remote -> scheduler (somewhere where there isn't a contact file to read)

We need to be able to configure the comms method for (1) & (2) for https://github.com/cylc/cylc-flow/issues/5235#issuecomment-1340604947. We may one day wish to support (3) too.

### Proposal

If `CYLC_TASK_COMMS_METHOD` is not set, use the `[platforms][localhost]communications method` configuration from the global config.

With the exception that if we are already on the scheduler host we should always use the TCP client.

> **Efficiency:** Note that the global configuration would NOT need to be read for `cylc message` commands from jobs because the `CYLC_TASK_COMMS_METHOD` would overrule the config anyway.

### Questions

1) Ok with this?
2) The `localhost` platform already configures:
    * Jobs on the scheduler host.
    * Jobs on hosts which share a FS with the scheduler host.
    * The scheduler host itself (e.g. auto-restart uses the `ssh command` from `localhost`).

   Adding this fourth meaning does convolute this a little, but it was convoluted already so does it matter?
    Can we thing of something better?

**Pull requests welcome!**

dwsutherland:
Yes, agreed. and partially addressed in #5235 .. The difference being, job communications can include `ssh` and `poll` where the latter doesn't appear to make sense for non-job CLI usage.. 

So I found it hard to justify using `[platforms][localhost]communications method` for CLI, if it was intentionally set to `poll`

oliver-sanders:
In practice sites would want to set `[platforms][localhost]communications method` to match `[CLI]communications method` so I don't think there is any benefit to separating the two.

oliver-sanders:
I don't think `poll` makes any sense as a `localhost` configuration. All of the other comms methods ultimately require local TCP so this it is an installation requirement that `[platform][localhost]communications method = tcp` works on the Scheduler hosts.

PR #5267, opened 2022-12-09
Add CLI http comms method

dwsutherland:
Partially address https://github.com/cylc/cylc-flow/issues/5235
Sibling to https://github.com/cylc/cylc-uiserver/pull/396

Description to come.

**Check List**

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -29,7 +29,7 @@ from cylc.flow import __version__ as CYLC_VERSION
 from cylc.flow.platforms import validate_platforms
 from cylc.flow.exceptions import GlobalConfigError
 from cylc.flow.hostuserutil import get_user_home
-from cylc.flow.network.client_factory import CommsMeth
+from cylc.flow.network.client_factory import CommsMeth, CLICommsMeth
 from cylc.flow.parsec.config import (
     ConfigNode as Conf,
     ParsecConfig,
@@ -1204,6 +1204,9 @@ with Conf('global.cylc', desc='''
                    The workflow polls for task status (no task messaging)
                 ssh
                    Use non-interactive ssh for task communications
+                http
+                   Via the running Hub proxy and/or UI-Server (requires UI
+                   Server installation)
 
                 .. versionchanged:: 8.0.0
 
@@ -1773,6 +1776,27 @@ with Conf('global.cylc', desc='''
                 "[runtime][<namespace>][events]submission timeout"
             )
         ))
+    with Conf('CLI', desc='''
+        Configure command line interface options.
+
+        .. versionadded:: 8.1.0
+    '''):
+        Conf('communication method',
+             VDR.V_STRING, 'zmq',
+             options=[meth.value for meth in CLICommsMeth], desc='''
+            The communication method used to interact with a running workflow
+            via the command line interface.
+
+            Options:
+
+            zmq
+               Direct client-server TCP communication via network ports
+            http
+               Via the running Hub proxy and/or UI-Server (requires UI
+               Server installation)
+
+            .. versionchanged:: 8.1.0
+        ''')
 
 
 def upg(cfg, descr):
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -29,7 +29,7 @@ from cylc.flow import __version__ as CYLC_VERSION
 from cylc.flow.platforms import validate_platforms
 from cylc.flow.exceptions import GlobalConfigError
 from cylc.flow.hostuserutil import get_user_home
-from cylc.flow.network.client_factory import CommsMeth, CLICommsMeth
+from cylc.flow.network.client_factory import CommsMeth, LocalCommsMeth
 from cylc.flow.parsec.config import (
     ConfigNode as Conf,
     ParsecConfig,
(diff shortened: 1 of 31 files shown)


cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1767,6 +1770,27 @@ def default_for(
...
+            Options:
+
+            zmq
+               Direct client-server TCP communication via network ports
+            http
+               Via the running Hub proxy and/or UI-Server (requires UI
+               Server installation)
+
+            .. versionchanged:: 8.1.0
+        ''')
```
oliver-sanders, line 1793: I think we might use `[platforms][localhost]communication method` for this configuration, see https://github.com/cylc/cylc-flow/issues/5265

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1198,6 +1198,9 @@ def default_for(
                    The workflow polls for task status (no task messaging)
                 ssh
                    Use non-interactive ssh for task communications
+                http
+                   Via the running Hub proxy and/or UI-Server (requires UI
+                   Server installation)
```
oliver-sanders, line 1203: How does cylc-flow know which port to contact the hub on when it doesn't share a filesystem with the hub? Do we need to configure this?

oliver-sanders (commented):
Nice.

We should be able to add the HTTPS comms method to the existing test battery. I've had a crack at the config here but haven't tried it out yet: https://github.com/oliver-sanders/cylc-flow/pull/new/swarm-https

Due to the way the test battery works there are two hosts:

* The one where tests run:
  * This is either your own machine or the GitHub actions runner.
  * It has cylc-flow installed.
* The job host:
  * This is the docker image, pretending to be another host.
  * It will have the UIS installed and the hub running.

So existing tests would try submitting jobs to the host which has the UIS running on it. This isn't how HTTPS comms would be used IRL as we would expect it to be installed on a third machine, however, this should be good enough for testing purposes. Jobs will contact the hub (on the same host) which will then TCP back to the test machine.

Any tests which define `REQUIRE_PLATFORM` that matches `_remote_background_indep_https` (e.g. `*` & `tcp:*`) would be run against HTTPS mode comms.

To write tests specifically for HTTPS comms use `REQUIRE_PLATFORM='comms:https'`. You can `ssh $CYLC_TEST_HOST` to test communication the more natural way around.

cylc/flow/cfgspec/globalcfg.py
dwsutherland: I think the user might need to copy/replicate the `~/.cylc/uiserver/api_info.json` file (if we don't figure out a way to automate it), because the server in question might not have ssh access..

cylc/flow/cfgspec/globalcfg.py
dwsutherland: Ok, wasn't sure if that was job specific or not.

dwsutherland:
> We should be able to add the HTTPS comms method to the existing test battery. I've had a crack at the config here but haven't tried it out yet: https://github.com/oliver-sanders/cylc-flow/pull/new/swarm-https

Should we merge that branch into this?

Question:
- Use `http` or `https` label?

`HTTPS` is just the secure version or `HTTP` but both are still know as `HTTP`, so given people can use `HTTP` versions of UIS/hub-proxy, I stuck with `http`... Or is the point to encourage `S` use?

oliver-sanders:
I would label it `https` just to avoid people thinking it might be unsecured.

cylc/flow/cfgspec/globalcfg.py
oliver-sanders: ATM clients are generally short lived (with the exception of the `cylc tui` client) so this isn't a biggie.

I guess reload this file if we get a comms failure and retry (once) if the file contents has changed.

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1204,8 +1204,11 @@ def default_for(
                    The workflow polls for task status (no task messaging)
                 ssh
                    Use non-interactive ssh for task communications
+                https
+                   Via the running Hub proxy and/or UI-Server (requires UI
+                   Server installation)
```
oliver-sanders, line 1209: Can put an added label just on the `https` method:

```suggestion
                   Server installation)
                   
                   .. versionadded:: 8.2.0
```

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1204,8 +1204,11 @@ def default_for(
...
 
-                .. versionchanged:: 8.0.0
+                .. versionchanged:: 8.2.0
```
oliver-sanders, line 1211: This version changed annotation was because the config had moved.

```suggestion
                .. versionchanged:: 8.0.0
```

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1646,6 +1649,22 @@ def default_for(
...
+                Options:
+
+                zmq
+                   Direct client-server TCP communication via network ports
+                https
+                   Via the running Hub proxy and/or UI-Server (requires UI
+                   Server installation)
+
+                .. versionchanged:: 8.2.0
+            ''')
```
oliver-sanders, line 1667: Can't remember where we got to with this, but I think we can use the `[localhost]` platform to configure the CLI since the config applies the CLI as much as it does job communication (which is over the same CLI).

dwsutherland: Yeah, this is done

oliver-sanders: There are still two `communication method` sections in the global config though?

dwsutherland: Yes, still in both sections. I don't want to rule out using `https` for remote jobs.

oliver-sanders: As, sorry, hadn't spotted that this section was just the `localhost` default.

PR #5259, opened 2022-12-05
db:  add flow nums to task jobs table

wxtim:
Closes #5252 

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.
- [x] Tests are included (or explain why tests are not needed).
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [ ] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #5271, opened 2022-12-12
play: upgrade DB *after* user upgrade confirmation 

### LINKED ISSUE #5270: play: upgrade DB after the user has confirmed upgrade not before
The DB gets upgraded before the user is even presented with the upgrade prompt :(

Also `cylc clean` can trigger DB upgrade for some reason? Can't see any reason for it to try this, we don't have any relevant upgraders to run.

**Pull requests welcome!**

oliver-sanders:
* Closes #5270
* Only upgrade the database after the user has answered th prompt saying that they want to upgrade!
* Remove the DB upgrade from the logic used by `cylc clean` as this is currently pointless.
* Add notes and comments to ensure we preserve clean compatibility with newer versions of Cylc (or at least so we know when we break compat).

**Check List**

- [x] No changelog (unreleased bug)

### DIFF
--- a/tests/functional/cylc-clean/03-multi.t
+++ b/tests/functional/cylc-clean/03-multi.t
@@ -75,13 +75,13 @@ ${HOME}/cylc-run/${CYLC_TEST_REG_BASE}
 __TREE__
 
 mkdir "${WORKFLOW_RUN_DIR}/run1/.service"
-touch "${WORKFLOW_RUN_DIR}/run1/.service/db"  # corrupted db!
+echo 'x' > "${WORKFLOW_RUN_DIR}/run1/.service/db"  # corrupted db!
 
 TEST_NAME="${TEST_NAME_BASE}-yes-no"
 run_fail "${TEST_NAME}" \
     cylc clean -y "$WORKFLOW_NAME/run1" "$WORKFLOW_NAME/run2"
 
-grep_ok "Cannot clean .*/run1" "${TEST_NAME}.stderr" -e
+grep_ok "file is not a database" "${TEST_NAME}.stderr" -e
 
 TEST_NAME="tree-post-clean-2"
 run_ok "${TEST_NAME}" tree --noreport --charset=ascii -L 4 "${HOME}/cylc-run/${CYLC_TEST_REG_BASE}"
--- a/cylc/flow/rundb.py
+++ b/cylc/flow/rundb.py
@@ -232,6 +232,7 @@ class CylcWorkflowDAO:
             ["delay"],
             ["timeout"],
         ],
+        # NOTE: this table is used by `cylc clean`, don't rename me!
         TABLE_TASK_JOBS: [
             ["cycle", {"is_primary_key": True}],
             ["name", {"is_primary_key": True}],
@@ -246,6 +247,7 @@ class CylcWorkflowDAO:
             ["time_run_exit"],
             ["run_signal"],
             ["run_status", {"datatype": "INTEGER"}],
+            # NOTE: this field is used by `cylc clean` don't rename me!
             ["platform_name"],
             ["job_runner_name"],
             ["job_id"],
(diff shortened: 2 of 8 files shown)


tests/functional/cylc-clean/03-multi.t
```diff
@@ -75,13 +75,13 @@ ${HOME}/cylc-run/${CYLC_TEST_REG_BASE}
 __TREE__
 
 mkdir "${WORKFLOW_RUN_DIR}/run1/.service"
-touch "${WORKFLOW_RUN_DIR}/run1/.service/db"  # corrupted db!
+echo 'x' > "${WORKFLOW_RUN_DIR}/run1/.service/db"  # corrupted db!
```
MetRonnie, line 78: I don't think that is a valid way to imitate a corrupted DB. Instead it makes it a non-DB file, whereas the way it was before (a blank file) is accepted as a DB by sqlite3

MetRonnie (commented):
I have tested this with an incompatible workflow DB
```
UPDATE workflow_params SET value = "8.0rc1" WHERE key == "cylc_version"
```
Cylc clean used to fail if you tried it on an incompatible workflow but now it goes ahead. Is this intended?

oliver-sanders:
> Cylc clean used to fail if you tried it on an incompatible workflow but now it goes ahead. Is this intended?

The way compatibility was tested before wasn't quite right. Now`cylc clean` will still fail if it encounters a clean-incompatible database (i.e. one without the platform field), but it will no longer fail if it encounters a restart-incompatible database (i.e. one we can't `cylc play`).

Restart-compatibility is not relevant for `cylc clean` and does not guarantee clean-compatibility. E.G. if a workflow is run with a newer version of Cylc which changes the platform field, then an older version of `cylc clean` would consider this DB to be restart-compatible, but the clean command would fail with `sqlite3.OperationalError` due to the missing field.

The best way to test DB compatibility for `cylc clean` is to try to access the platform field, if we get `sqlite3.OperationalError` then it ain't compatible.

tests/functional/cylc-clean/03-multi.t
oliver-sanders: I'm not sure a blank file is a legit way to make a blank DB as SQLITE will happily take a blank file and turn it into a DB.

MetRonnie: Ok. I guess I was thinking more in terms of the DB contents being "corrupted" (e.g. empty tables which has happened before in #4637) rather than the file itself becoming unreadable

MetRonnie:
> **Built on:** #5270

I think you mean #5259, may be worth rebasing onto master now that's been merged

MetRonnie:
I've spotted a problem with `tests/functional/cylc-clean/03-multi.t` which was failing every time it was run the first time round due to picking up other test files. Fix here: https://github.com/oliver-sanders/cylc-flow/pull/61

MetRonnie:
I can't reproduce the remaining functional test failures locally 🤔

oliver-sanders:
Maybe jam another directory into the tree and move the clean down one level to avoid the clash.

MetRonnie:
Ok, got another PR to fix it: https://github.com/oliver-sanders/cylc-flow/pull/62

MetRonnie:
Is it worth adding a test that the DB upgrades have not taken place if the user chooses not to upgrade?

oliver-sanders:
Flaky test ` tests/f/cylc-show/03-clock-triggered-non-utc-mode.t` fixed on another brnach.

MetRonnie:
I think a functional test could do:
- Include a DB dump from an 8.0.2 workflow
- Do `cylc play` which will not run because it won't automatically upgrade in non-interactive mode
- Query the DB to check the `is_manual_submit` column has not been added to the `task_states` table

oliver-sanders:
Yes, but that DB upgrader won't be there for ever, or even necessarily for much longer, at which point the test becomes dead.

MetRonnie (approved):

wxtim (approved):
- [x] Code read
- [x] Tried it out.

PR #5280, opened 2022-12-22
coverage: update config

oliver-sanders:
This might help with a coverage error [reported on Element](https://matrix.to/#/!hwZqSYihGPuhDdIzIP:matrix.org/$167164934811296mZdsf:matrix.org?via=matrix.org)

According to the [changelog](https://coverage.readthedocs.io/en/stable/changes.html#version-7-0-0b1-2022-12-03):

> Changes to file pattern matching, which might require updating your configuration:
>
> Previously, * would incorrectly match directory separators, making precise matching difficult. This is now fixed, closing [issue 1407](https://github.com/nedbat/coveragepy/issues/1407).
>
> Now ** matches any number of nested directories, including none.

So I think our use of `*` is now broken? Not sure.

**Check List**

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -13,9 +13,9 @@ disable_warnings =
     no-data-collected
     module-not-measured
 omit =
-    tests/*
-    */cylc/flow/*_pb2.py
-    cylc/flow/etc/*
+    tests/**
+    **_pb2.py
+    cylc/flow/etc/**
 parallel = True
 source = ./cylc
 timid = False
@@ -41,9 +41,9 @@ exclude_lines =
 fail_under=0
 ignore_errors = False
 omit =
-    tests/*
-    */cylc/flow/*_pb2.py
-    cylc/flow/etc/*
+    tests/**
+    **_pb2.py
+    cylc/flow/etc/**
 precision = 2
 show_missing = False
 skip_covered = False
--- a/.coveragerc
+++ b/.coveragerc
@@ -14,7 +14,7 @@ disable_warnings =
     module-not-measured
 omit =
     tests/**
-    **_pb2.py
+    ./**_pb2.py
     cylc/flow/etc/**
 parallel = True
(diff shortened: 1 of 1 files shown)


oliver-sanders:
The error message has now changed to:

> coverage.exceptions.ConfigError: File pattern can't include '**/**'

`**/**` does not appear in the config file either.

MetRonnie:
That looks like it's coming from `**_pb2.py`:
https://github.com/nedbat/coveragepy/blob/31513b4b0559fdc7578aa1a30e048b928dc46188/coverage/files.py#L336-L337

oliver-sanders:
So `./**_pb2.py` would probably work then.

.coveragerc
```diff
@@ -13,9 +13,9 @@ disable_warnings =
     no-data-collected
     module-not-measured
 omit =
-    tests/*
-    */cylc/flow/*_pb2.py
-    cylc/flow/etc/*
+    tests/**
+    **_pb2.py
```
oliver-sanders, line 17:
```suggestion
    ./**_pb2.py
```

.coveragerc
```diff
@@ -41,9 +41,9 @@ exclude_lines =
 fail_under=0
 ignore_errors = False
 omit =
-    tests/*
-    */cylc/flow/*_pb2.py
-    cylc/flow/etc/*
+    tests/**
+    **_pb2.py
```
oliver-sanders, line 45:
```suggestion
    ./**_pb2.py
```

MetRonnie:
I don't think anything we do in the config will fix this. It appears to be a bug https://github.com/nedbat/coveragepy/issues/1513. A solution could be to pin coverage to `6.*`

nedbat:
The coverage.py fix is now on master in GitHub, I'll probably be publishing it today.

MetRonnie:
The test is now passing on master

PR #5229, opened 2022-11-16
`cylc vro`

wxtim:
Partially addresses #3896 

Branch based upon a rebase of #5214  onto #5121 .

Depending on what Reviewers thing it may replace #5214 once #5121 has been merged.

## Purpose

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.
- [ ] Tests are included (or explain why tests are not needed).
- [ ] `CHANGES.md` entry included if this is a change that can affect users
- [ ] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [ ] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #5297, opened 2023-01-12
Restrict `--set` & `--set-file`  options in `cylc vr`

wxtim:
This PR documents the bug as well as fixing it. There is no related issue. Follow-up to #5229

## Synopsis

`--set` and `--set-file` were included by mistake in `cylc vr` because I copied the general form of VR from `cylc vip` where it makes sense. Reinstall doesn't use these CLI variables so they won't change the workflow. This is potentially misleading to users.

## Code to try

```ini
#!jinja2
# Cylc Workflow
[scheduler]
    allow implicit tasks = ture
[scheduling]
    [[graph]]
        {{CYCLE}}= foo => bar => baz
```

then,

On the command line

```bash
cylc vip --set 'CYCLE="PT1H"'
...

# Does not change the running workflow!
cylc vr --set 'CYCLE="PT2H"' 
```

**Check List**

- [x] No tests, change is specification rather than tests.
- [x] No `CHANGES.md` entry - this effectively fixes a recently added PR before relase.
- [x] NO [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.

### DIFF
--- a/cylc/flow/scripts/validate_reinstall.py
+++ b/cylc/flow/scripts/validate_reinstall.py
@@ -106,13 +106,11 @@ def check_tvars_and_workflow_stopped(
         tvars: options.tvars, from `--set`
         tvars_file: options.tvars_file, from `--set-file`
     """
-    if is_running and (tvars or [] or tvars_file or []):
-        msg = (
-            'Template variables can only be changed if a'
-            ' workflow is stopped. You tried to change:')
-        for tvar in tvars or [] + tvars_file or []:
-            msg += f'\n * {tvar}'
-        LOG.warning(msg)
+    if is_running and (tvars or tvars_file):
+        LOG.warning(
+            'Template variables (from --set/--set-file) can '
+            'only be changed if the workflow is stopped.'
+        )
         return False
     return True
 
--- a/cylc/flow/scripts/validate_reinstall.py
+++ b/cylc/flow/scripts/validate_reinstall.py
@@ -92,6 +92,31 @@ def get_option_parser() -> COP:
     return parser
 
 
+def check_tvars_and_workflow_stopped(
+    is_running: bool, tvars: list, tvars_file: list
+) -> bool:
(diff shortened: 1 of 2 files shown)


cylc/flow/scripts/validate_reinstall.py
```diff
@@ -92,6 +92,31 @@ def get_option_parser() -> COP:
...
+    Template vars set by --set (options.templatevars) or --set-file
+    (optiions.templatevars_file) are only valid if the workflow is stopped
+    and vr will play it.
+
+    args:
+        is_running: Is workflow running?
+        tvars: options.tvars, from `--set`
+        tvars_file: options.tvars_file, from `--set-file`
+    """
+    if is_running and (tvars or [] or tvars_file or []):
```
oliver-sanders, line 109:
```suggestion
    if is_running and (tvars or tvars_file):
```

MetRonnie, line 109:
```suggestion
    if is_running and (tvars or tvars_file):
```

oliver-sanders (approved):
a

cylc/flow/scripts/validate_reinstall.py
```diff
@@ -92,6 +92,31 @@ def get_option_parser() -> COP:
...
+        msg = (
+            'Template variables can only be changed if a'
+            ' workflow is stopped. You tried to change:')
+        for tvar in tvars or [] + tvars_file or []:
+            msg += f'\n * {tvar}'
+        LOG.warning(msg)
```
MetRonnie, line 115: I don't think it needs to list the template variables back to the user
```suggestion
        LOG.warning(
            'Template variables (from --set/--set-file) can '
            'only be changed if the workflow is stopped.'
        )
```

MetRonnie (approved):

Issue #5272, opened 2022-12-13
pyzmq: upgrade


oliver-sanders:
Our pyzmq dependency conflicts with jupyter_server v2's requirements (`22.*` and `>=24` respectively at the time of writing) 

See:

https://github.com/conda-forge/jupyter_server-feedstock/blob/91f6e6f5b9cd24b8825e71e09941fc61536ecd5b/recipe/meta.yaml#L39

We should bump to a more recent version of pyzmq.

With a little investigation we may be able to relax our version range for pyzmq.

> **Context:**
>
> Because Cylc is a distributed system we have to consider that different installations of Cylc on different platforms must work together seamlessly. We have historically pinned pyzmq to a major release to ensure all distributed components have matching versions, however, pyzmq has a rapid major release cycle which causes issues here. Suggest taking a dive into the zmq docs to find out what their back/forward compatibility policies are and whether it is sensible to push out our version range.

**Pull requests welcome!**

PR #5307, opened 2023-01-17
deps: loosen pyzmq dependency

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/5272
* Closes https://github.com/cylc/cylc-flow/issues/5311
* Unblocks https://github.com/cylc/cylc-uiserver/issues/411

tldr; We should be good to loosen off on this requirement, full rationale in the issue.

* The ZMTP protocol provides robust forward & backward compatibilit from libzmq version 3.2.2 onwards.
* Since this is the minimum version compatible with pyzmq, we can safely loosen this version coupling.
* For more information see the rationale in https://github.com/cylc/cylc-flow/issues/5311#issuecomment-1414073472

### DIFF
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -14,7 +14,7 @@ dependencies:
   - protobuf >=4.21.2,<4.22.0
   - psutil >=5.6.0
   - python
-  - pyzmq >=22,<23
+  - pyzmq >=22
   - setuptools >=49, <67
   - urwid >=2,<3
   # Add # [py<3.11] for tomli once Python 3.11 Released
--- a/cylc/flow/network/client.py
+++ b/cylc/flow/network/client.py
@@ -305,7 +305,9 @@ class WorkflowRuntimeClient(  # type: ignore[misc]
         if msg['command'] in PB_METHOD_MAP:
             response = {'data': res}
         else:
-            response = decode_(res.decode())
+            response = decode_(
+                res.decode() if isinstance(res, bytes) else res
+            )
         LOG.debug('zmq:recv %s', response)
 
         try:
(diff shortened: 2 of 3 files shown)


wxtim:
The Python 3.10 failure is a codecov upload failure, but the other two look like they might be real failures.

oliver-sanders:
The pyzmq package on conda-forge appears to be missing Python 3.7 builds, looking in...

oliver-sanders:
Conda-forge has dropped 3.7 - https://conda-forge.org/docs/user/announcements.html#dropping-python-3-7

This means new releases will not be built against 3.7.

oliver-sanders:
Options:

1. Build the swarm against "3.8" for "3.7" testing.
   Fine, but it does mean that the remote code won't get tested with 3.7.
2. Drop "3.7".
   Fine but it is a little early, 3.7 doesn't hit end of life until June.

conda-environment.yml
```diff
@@ -15,11 +15,11 @@ dependencies:
   - protobuf >=4.21.2,<4.22.0
   - psutil >=5.6.0
   - python
-  - pyzmq >=22,<23
+  - pyzmq >=25,<26
   - setuptools >=49
   - urwid >=2,<3
   # Add # [py<3.11] for tomli once Python 3.11 Released
-  - tomli >=2.0.0
+  - tomli >=2.0.0,<3
```
MetRonnie, line 22: This `<3` doesn't match `setup.cfg`

oliver-sanders:
(rebased)

MetRonnie:
Want me to have a look at the mypy failures?

oliver-sanders:
Presumably not related to this branch?

MetRonnie:
Looks like it's type annotations added to newer versions of pyzmq

MetRonnie:
Actually I think I already did this earlier, must have got lost in one of the force pushes, luckily I still had the branch locally

wxtim (approved):

MetRonnie (approved):

PR #5312, opened 2023-01-19
treat items in queues as implicit tasks; queues to be list separated type

### LINKED ISSUE #5260: Space-separated list of queue members ignored without error
**Describe the bug**
See [discussion on Element](https://matrix.to/#/!hwZqSYihGPuhDdIzIP:matrix.org/$1670255552138977aoIyT:matrix.org?via=matrix.org
)

If a space separated list is given for `flow.cylc[scheduling][queues][<queue name>]members`, rather than the required comma-separated list, cylc interprets the value as a single-element list containing a string with spaces. `cylc validate` reports the flow is valid and no errors appear in `cylc play` or `cylc log`, but cylc behaves as though the internal queue defined in `flow.cylc[scheduling][queues][<queue name>]` did not exist, ignoring any limits defined in it.

Have tested 8.0.1 and 8.0.3 and saw this behavior in both versions.

Create a flow that defines an internal queue, and put a space-delimited list in the `members` field, e.g.:
```
[scheduling]
  cycling mode = integer
  initial cycle point = 1
  runahead limit = P10
  [[queues]]
    [[[my_queue]]]
      limit = 2
      members = task1 task2
  [[graph]]
    P1 = "task1 => task2"
[runtime]
  [[task1]]
    script = "sleep 5; echo done>status"
  [[task2]]
    script = "sleep 5; echo done>status"
```

**Expected behavior**
`cylc validate` and `cylc play` should exit with an error on the above flow, since the `members` list should be comma delimited and 'task1 task2' is not a valid task name.

**Actual behavior**
`cylc validate` reports the above flow is valid, and `cylc play` executes the flow, ignoring the queue limit.

**Additional context**
Per @oliver-sanders in the [Element thread](https://matrix.to/#/!hwZqSYihGPuhDdIzIP:matrix.org/$1670256094139282eKgpQ:matrix.org?via=matrix.org):

> members is currently a "V_STRING_LIST" which is a comma-separated list which tolerates spaces (hence no error) We could try "V_SPACELESS_STRING_LIST" which forbids spaces (creating an error).

> Suggesting:
> 
>    1. Switch to spaceless string list since spaces are not valid in task names.
>    2. Validate queue members to ensure that all listed members exist in the [runtime] section.
> 
> Unfortunately we may need to do a bit of testing to ensure we don't break too many workflows with these changes.

wxtim:
Closes #5260 

**Check List**


wxtim:
Currently running local tests to see if this branch causes breakages.

cylc/flow/config.py
```diff
@@ -532,6 +532,16 @@ def __init__(
...
+        for queue in self.cfg["scheduling"]["queues"]:
+            for name in self.cfg["scheduling"]["queues"][queue][
+                "members"
+            ]:
+                if (
+                    name not in self.taskdefs
+                    and name not in self.cfg['runtime']
+                ):
+                    self.implicit_tasks.add(name)
+
```
oliver-sanders, line 544: I'm not sure that **failing** for implicit tasks in the queues is a good idea. It is common to have Jinja2 switches which turn bits of graph on and off. This change would mandate matching switches to turn off members in queues too.

```ini
[scheduling]                        
    [[queues]]                      
        [[[myq]]]                   
            members = foo, bar, baz    
    [[graph]]                       
        R1 = """                    
            foo     
            bar     
{% if BAZ %}    
            baz    
{% endif %}    
        """    
     
[runtime]    
    [[foo, bar]]    
{% if BAZ %}    
    [[baz]]    
{% endif %}  
```

There may be value in warning about implicit tasks in queues e.g. to help spot typos.

Maybe bump this bit into a "question" issue for "8.2.0".

oliver-sanders (commented):
The spacelist list change works a treat.

I'm not so sure about checking for implicit tasks in the queues as this will break a lot of existing workflows and makes more programmatic workflows trickier to implement.

oliver-sanders:

CHANGES.md
```diff
@@ -11,6 +11,14 @@ updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
 
 -------------------------------------------------------------------------------
+## __cylc-8.1.1 (<span actions:bind='release-date'>Coming Soon</span>)__
+
+### Fixes
+
+[#5312](https://github.com/cylc/cylc-flow/pull/5312) - Force queues to be
+comma separated lists only. Tasks listed in queues will be treated as implicit
+in the same way as tasks in listed in runtime.
```
hjoliver, line 20: I think this description is a bit misleading. Is this more accurate?:
```suggestion
[#5312](https://github.com/cylc/cylc-flow/pull/5312) - task names must be comma-separated in queue member lists. Any implicit tasks (i.e. with no task definition under runtime) assigned to a queue will generate a warning.
```
(Taking into account the suggestion below not to fail implicit tasks assigned to a queue).

cylc/flow/config.py
hjoliver: Agreed, a warning could be helpful, but failing implicit tasks here would create some pain.

tests/integration/test_config.py
```diff
@@ -282,3 +283,39 @@ def test_parse_special_tasks_families(flow, scheduler, validate, section):
             'foo(P1D)',
             'foot(P1D)'
         }
+
+
+def test_queue_treated_as_implicit(flow, validate):
+    """Tasks listed in queue should be regarded as implicit
```
hjoliver, line 289: I'm not sure this docstring makes sense. It's more that any implicit task - i.e. a task name with no definition under runtime - assigned to a queue will cause validation failure (or, better, just a warning as suggested above).

hjoliver (changes requested):

cylc/flow/config.py
wxtim: I thought this was overkill at the moment, but the ticket makes it look like it's a should, not a  "could sometime down the line".

oliver-sanders (approved):
:+1:

tests/integration/test_config.py
```diff
@@ -282,3 +283,42 @@ def test_parse_special_tasks_families(flow, scheduler, validate, section):
...
+    assert (
+        'Queues contain tasks not defined in runtime'
+        in caplog.records[0].message
+    )
+
+
+def test_queue_treated_as_comma_separated(flow, validate):
+    """Tasks listed in queue should be separated with commas, not spaces.
+
+    https://github.com/cylc/cylc-flow/issues/5260
```
datamel, line 312: Not sure both these links to the issue are needed?

datamel (approved):
I have reproduced the issue on master and checked out this and can confirm it fixes both of the problems. Confirmed tests are also passing locally for me. Thanks @wxtim.

datamel (commented):
Actually, just to confirm the behaviour of `cylc play`... before I had manually tested only `validate`. Cylc play is continuing for me when I have a space separated list, including an item which isn't in the runtime. I do get the warning in the log, but it keeps going. Are we expecting a failure here instead? 
Removing my approval for now until I understand desired behaviour.

oliver-sanders:
Hmmm, that doesn't sound right.

oliver-sanders:
I can't get it to actually play the workflow, however, the play error only comes out after the scheduler has started due to the way validation is performed:

```
[qval] $ cylc val .                                                                
ListValueError: (type=list) [scheduling][queues][myq]members = foo bar - (list item   "foo bar" cannot contain a space character)
[qval] $ cylc vip .    
$ cylc validate ~/cylc-tmp/qval
ListValueError: (type=list) [scheduling][queues][myq]members = foo bar - (list item   "foo bar" cannot contain a space character)
[qval] $ cylc install .
INSTALLED qval/run1 from ~/cylc-tmp/qval
[qval] $ cylc play qval/run1 --no-detach
                  
   ▪ ■  Cylc Workflow Engine 8.1.1.dev
                  
2023-01-25T10:16:46Z INFO - Extracting job.sh to ~/cylc-
    run/qval/run1/.service/etc/job.sh
2023-01-25T10:16:46Z INFO - Workflow: qval/run1
2023-01-25T10:16:46Z ERROR - Workflow shutting down - ListValueError:
    (type=list) [scheduling][queues][myq]members = foo bar - (list item "foo bar"
    cannot contain a space character)
2023-01-25T10:16:46Z INFO - DONE
```

datamel:
I'll try to reproduce, I may have done something wrong.

datamel (approved):
I must have been mistaken, working as expected! Sorry!!

oliver-sanders:
Changing PR base to 8.1.x, rebase may be required

oliver-sanders:
Got away without rebase, please merge when the test pass.

tests/integration/test_config.py
wxtim: I think it's worth adding them to the tests they go with.

PR #5313, opened 2023-01-19
Fix Validation using old tvars fails for Cylc 7 workflows

wxtim:
Allow parsec to fail gracefully when trying to retrieve template variables from a workflow database if it's a Cylc 7 DB, by doing the same DB compatibility check that's done on restart.

**Fixes two small issues not recorded elsewhere**

- I had failed to use the `is_public` kwarg on `CylcWorkflowDAO.__init__` , without which the function tries to write to the database, which we don't want to do for the `parsec.fileparse` use of database access. _(Now sorted out by #5330)_
- Pre Cylc 8 workflows don't have the database field under the same name - have made `cylc play` and `cylc vr` fail gracefully when trying to run on such workflows. 

Also closes https://github.com/cylc/cylc-flow/issues/5317

**Check List**


cylc/flow/templatevars.py
```diff
@@ -20,17 +20,19 @@
...
-
 from cylc.flow.rundb import CylcWorkflowDAO
+from cylc.flow.workflow_files import check_deprecation
 
 
 def get_template_vars_from_db(run_dir):
     """Get template vars stored in a workflow run database.
     """
     template_vars = {}
+    if check_deprecation(run_dir):
```
oliver-sanders, line 31: I don't think back-compat mode is relevant here.

The Cylc database should be the same for Cylc 8 workflows whether in compat-mode or not.

We cannot restart Cylc 7 workflows at Cylc 8, `cylc play` protects against this, we need to ensure this check is performed earlier on for `cylc vr`.

oliver-sanders:
The check which prevents Cylc 8 from restarting Cylc 7 workflows is this:

```
  ...
  File "~/cylc-flow/cylc/flow/workflow_db_mgr.py", line 799, in check_workflow_db_compatibility
cylc.flow.exceptions.ServiceFileError: Workflow database is incompatible with Cylc 8.2.0.dev, or is corrupted.
```

So all that's needed is to add the `check_workflow_db_compatibility` check at the start of `cylc vr`.

cylc/flow/templatevars.py
wxtim: > The Cylc database should be the same for Cylc 8 workflows whether in compat-mode or not.

Yes, but what stops Cylc 8 `validate` being used on workflows run using Cylc 7?

MetRonnie:
https://github.com/cylc/cylc-flow/blob/06b40795c24f5de5d4e5c5740d3e75f7b709618a/cylc/flow/rundb.py#L338-L339

IMO whether to create the tables or not should be handled by a different arg, called `create_tables` instead of `is_public`, because that really isn't clear otherwise

cylc/flow/workflow_files.py
```diff
@@ -1226,6 +1226,8 @@ def check_deprecation(path, warn=True):
         cylc.flow.flags.cylc7_back_compat = True
         if warn:
             LOG.warning(SUITERC_DEPR_MSG)
+        return True
+    return False
```
datamel, line 1230: Can't see how these returns are being used. I'm probably missing something though.

datamel (commented):
I have checked out this branch. I have run workflows under cylc7 and then tried validate with this branch and master. 
I get a nice error on this branch when the workflow has run and validating at cylc 8, master will state the workflow is valid at cylc 8.
If I validate an actively running (with cylc 7) workflow on this branch, it still pass validation. This is probably a niche case? EDIT: I see Oliver has this covered and it it ok in this case.
When I `play` an actively running cylc 7 workflow, I get traceback with `KeyError: 'CYLC_WORKFLOW_PID'` at cylc8, again probs niche and maybe a separate issue?

cylc/flow/workflow_files.py
wxtim: @`validate_reinstall.py::147`

cylc/flow/templatevars.py
```diff
@@ -20,17 +20,19 @@
 from typing import Any, Dict
 
 from cylc.flow.exceptions import InputError
-
-
 from cylc.flow.rundb import CylcWorkflowDAO
+from cylc.flow.workflow_db_mgr import WorkflowDatabaseManager
+from cylc.flow.workflow_files import check_deprecation
```
datamel, line 25: Unused import.

wxtim:
> IMO whether to create the tables or not should be handled by a different, new arg, called create_tables instead of is_public, because that really isn't clear otherwise

It might well have saved me from this error, but unless you insist I think it's out of scope.

cylc/flow/templatevars.py
wxtim:
```suggestion
```

cylc/flow/workflow_files.py
datamel: Hmm, maybe you have not pushed changes? I have that line as checking... `if not workflow_running:` and I have this as adding returns to `check_deprecation`. I have pulled changes and nothing coming through for me.

wxtim, line 1230:
```suggestion
```

oliver-sanders:

cylc/flow/scripts/validate_reinstall.py
```diff
@@ -135,12 +137,18 @@ def vro_cli(parser: COP, options: 'Values', workflow_id: str):
...
-        # Workflow is definately running:
+        # Workflow is definitely running:
         workflow_running = True
     else:
-        # Workflow is definately stopped:
+        # Workflow is definitely stopped:
         workflow_running = False
 
+    # We won't be able to restart stopped 7.x workflows. Check for this NOW:
+    if not workflow_running:
```
oliver-sanders, line 147: From experimentation we don't get this far for Cylc 7 workflows because `detect_old_contact_file` fails.

Methodology to test:

```console
$ # Write a Cylc 7 compatible workflow in `~/cylc-src`.

$ # install it with Cylc 8
$ cylc install .

$ # make it look like it *wasn't* installed with Cylc 8
$ (cd ~/cylc-run/<id>; mv _cylc-install _cylc-install-2)

$ # run it with Cylc 7
$ CYLC_VERSION=7 cylc run <id>` --hold

$ # make it look like it *was* installed with Cylc 8
$ (cd ~/cylc-run/<id>; mv _cylc-install-2 _cylc-install)

$ # make it look like there are changes to reload & vr'em
$ echo -e '\n' >>> suite.rc
$ cylc vr <id>
```

oliver-sanders: Suggestion:

```python
try:
    detect_old_contact_file(workflow_id)
except ServiceFileError:
    ...
except Exception:
    if looks_cylc_7ey():
        raise CylcError('Cylc 8 cannot restart or reload Cylc 7 workflows')
```

Where:

```
def looks_cylc_7ey():
    with open(path_to_contact_file, 'r') as f:
        for line in f:
            if 'CYLC_VERSION=7' in line:
                return True
    return False
```

---

Or simply:


```python
try:
    detect_old_contact_file(workflow_id)
except ServiceFileError:
    ...
except Exception:
    LOG.critical(
        'Cannot tell if the workflow is running'
        '\nNote, Cylc 8 cannot restart Cylc 7 workflows.'
    )
    raise
```

oliver-sanders:
Ran through the methodology from here - https://github.com/cylc/cylc-flow/pull/5313#discussion_r1084285970

```
2023-01-24T10:10:23Z CRITICAL - Cannot tell if the workflow is running
    Note, Cylc 8 cannot restart Cylc 7 workflows.
  File "/home/h06/osanders/mambaforge/envs/cylc.dev/bin/cylc", line 33, in <module>
    sys.exit(load_entry_point('cylc-flow', 'console_scripts', 'cylc')())
  File "/net/home/h06/osanders/cylc-flow/cylc/flow/scripts/cylc.py", line 653, in main
  File "/net/home/h06/osanders/cylc-flow/cylc/flow/scripts/cylc.py", line 286, in execute_cmd
    entry_point.resolve()(*args)
  File "/net/home/h06/osanders/cylc-flow/cylc/flow/terminal.py", line 232, in wrapper
  File "/net/home/h06/osanders/cylc-flow/cylc/flow/scripts/validate_reinstall.py", line 120, in main
    sys.exit(vro_cli(parser, options, workflow_id))
  File "/net/home/h06/osanders/cylc-flow/cylc/flow/scripts/validate_reinstall.py", line 152, in vro_cli
    workflow_running, options.templatevars, options.templatevars_file
UnboundLocalError: local variable 'workflow_running' referenced before assignment
```

cylc/flow/templatevars.py
```diff
@@ -20,17 +20,18 @@
...
 
 
 def get_template_vars_from_db(run_dir):
     """Get template vars stored in a workflow run database.
     """
     template_vars = {}
     if (run_dir / 'log/db').exists():
-        dao = CylcWorkflowDAO(str(run_dir / 'log/db'))
+        WorkflowDatabaseManager(
+            str(run_dir / 'log')).check_workflow_db_compatibility()
```
oliver-sanders, line 33: Should this be using the private DB?

MetRonnie: There is a problem with this - `WorkflowDatabaseManager().check_workflow_db_compatibility()` will try to create the tables. I have just realised I had commented this in the code in a "tidy" branch but never created a PR

oliver-sanders: I think:

* We need to add a switch to allow `check_workflow_db_compatibility` to run on the public DB?
  * Or are we really really sure the workflow is not running at this point?
* We should avoid that `CREATE TABLES` junk for this as it's just another [unnecessary] thing to go wrong.

cylc/flow/workflow_db_mgr.py
```diff
@@ -194,19 +194,24 @@ def delete_workflow_stop_task(self):
         """Delete workflow stop task from workflow_params table."""
         self.delete_workflow_params(self.KEY_STOP_TASK)
 
-    def _get_pri_dao(self) -> CylcWorkflowDAO:
+    def _get_pri_dao(self, use_pri_dao=True) -> CylcWorkflowDAO:
```
oliver-sanders: Rename `_get_dao`?

cylc/flow/templatevars.py
wxtim: I've made some changes which ought to fix this - 
Essentially allowing the `check_workflow_db_compatibility` method to specify whether use public or private databases.

cylc/flow/scripts/validate_reinstall.py
```diff
@@ -135,10 +135,16 @@ def vro_cli(parser: COP, options: 'Values', workflow_id: str):
...
         detect_old_contact_file(workflow_id)
     except ServiceFileError:
-        # Workflow is definately running:
+        # Workflow is definitely running:
         workflow_running = True
+    except Exception:
+        LOG.critical(
+            'Cannot tell if the workflow is running'
+            '\nNote, Cylc 8 cannot restart Cylc 7 workflows.'
+        )
```
datamel, line 144: After discussion in another pr...
https://github.com/cylc/cylc-flow/pull/5323#issuecomment-1403731534

This error message needs to be raised on other commands when interacting with running cylc 7 workflow.

datamel: From my investigation it is only `cylc play` which doesn't raise this error.

wxtim:
I'm actually concerned that this is still not OK, because if I do
```bash
$ cylc7 stop workflow
$ cylc8 play workflow
.... correct error
$ cylc7 run workflow
ERROR: Suite Cylc version (unknown) is incompatible with Cylc 7
```

cylc/flow/scripts/validate_reinstall.py
datamel: I don't think this is resolved? `cylc play` needs to inform users of this message when playing a running Cylc 7 workflow. I checked the other commands and they fail correctly.

datamel (approved):
I've checked this out and heavily manually tested various scenarios with the interactions between cylc 7 workflows and cylc 8. I can't spot any more issues, nice work @wxtim.

datamel (commented):
Note the failing test is just a url problem, unrelated issue.

cylc/flow/scripts/report_timings.py
```diff
@@ -132,7 +132,8 @@ def main(parser: COP, options: 'Values', workflow_id: str) -> None:
         # No output specified - choose summary by default
         options.show_summary = True
 
-    with _get_dao(workflow_id) as dao:
+    db_file = get_workflow_run_pub_db_path(workflow_id)
+    with CylcWorkflowDAO(db_file) as dao:
```
MetRonnie, line 136: Sorry, just spotted something I missed
```suggestion
    with CylcWorkflowDAO(db_file, is_public=True) as dao:
```

tests/unit/test_templatevars.py
```diff
@@ -141,3 +149,10 @@ def test_get_old_tvars(key, expect, _setup_db):
     """It can extract a variety of items from a workflow database.
     """
     assert _setup_db[key] == expect
+
+
+def test_dont_get_old_tvars_back_compat(tmp_path):
+    """It won't even try to extract workflow_template_vars in compat mode.
+    """
+    tmp_path.touch('suite.rc')
+    assert get_template_vars_from_db(tmp_path) == {}
```
MetRonnie, line 158: I don't think this test is valid - why shouldn't it extract template vars in back compat mode?

However, I think we should test that parsing the config fails if a Cylc 7 DB is present

MetRonnie (changes requested):

cylc/flow/templatevars.py
MetRonnie: Resolved by #5330

cylc/flow/workflow_db_mgr.py
MetRonnie: Resolved by #5330

tests/unit/test_templatevars.py
oliver-sanders: This will result in an SQL error (table was renamed).

tests/unit/test_templatevars.py
```diff
@@ -114,6 +115,13 @@ def _setup_db(tmp_path_factory):
     logfolder.mkdir()
     db_path = logfolder / WorkflowFiles.LogDir.DB
     with CylcWorkflowDAO(db_path, create_tables=True) as dao:
+        dao.connect().execute(
+            rf'''
+                INSERT INTO workflow_params
+                VALUES
+                    ("cylc_version", "{cylc_version}")
```
oliver-sanders, line 122: Should this be:

```suggestion
                    ("cylc_version", cylc_version)
```

oliver-sanders: Is this needed, I think the tests pass as it was, I don't think the Cylc version should matter with template variable loading?

MetRonnie: The reason it matters is that Cylc 7 workflows do not have the `workflow_template_vars` table.

wxtim: This is an f-string after all!

tests/unit/test_templatevars.py
wxtim: I have now added a test where a completely blank file is checked.

MetRonnie:
Added some tests in https://github.com/wxtim/cylc/pull/52

cylc/flow/scripts/report_timings.py
MetRonnie: ^

cylc/flow/scripts/validate_reinstall.py
```diff
@@ -135,10 +135,16 @@ def vro_cli(parser: COP, options: 'Values', workflow_id: str):
     try:
         detect_old_contact_file(workflow_id)
     except ServiceFileError:
-        # Workflow is definately running:
+        # Workflow is definitely running:
         workflow_running = True
+    except Exception:
```
MetRonnie, line 140: For debugging purposes, don't swallow the original exception
```suggestion
    except Exception as exc:
        LOG.error(exc)
```

oliver-sanders:
Re-reviewed code, LGTM, moving onto functional testing...

MetRonnie:
Regression in latest rebase. Would not recommend rebasing at this point due to how out of date the earlier commits are, just whack squash 'n merge at the end

MetRonnie (approved):

MetRonnie:
Have tested for both a running and stopped Cylc 7 workflow:
- `cylc play`
- `cylc validate`
- `cylc vr` (Produces traceback for a running Cylc 7 workflow but the correct error message is now printed at the top so who cares)
- `cylc vip`

And ensured no tables are created by these commands

datamel (approved):
Tested as working. Thanks @wxtim

Issue #5315, opened 2023-01-20
efficiency: increment_graph_window


oliver-sanders:
I've just tried out an old Cylc 7 scaling example and found out that Cylc 8 scales disastrously against this benchmark.

The issue occurs when pushing the number of edges in a workflow. This causes the `increment_graph_window` function to do a lot of work, which causes a vast number of non-functional `TaskProxy` objects to be created (and destroyed).

Here's some `--profile` results showing the `increment_graph_window` call stack soaking up CPU:

![Screenshot from 2023-01-20 10-50-20](https://user-images.githubusercontent.com/16705946/213677830-2c9f87a4-20c8-4b95-86b9-157e55ba6cff.png)

And here's the example workflow that generated it (I used `-s TASKS=50`):

```ini
#!Jinja2                                                                                                   
                                                                                                           
{% set TASKS = TASKS | default(100) %}                                                                     
{% set CYCLES = CYCLES | default(1) %}                                                                     
{% set SLEEP = SLEEP | default(1) %}                                                                       
                                                                                                           
[meta]                                                                                                     
    description = """                                                                                      
        Example scaling workflow with {{ (TASKS * 2) + 2 }} tasks and {{ (TASKS ** 2) + (TASKS * 2) }}                             
        dependencies per cycle which runs for {{ CYCLES }} cycles.                                         
    """                                                                                                    
                                                                                                           
[task parameters]                                                                                          
    x = 1..{{ TASKS }}                                                                                     
    y = 1..{{ TASKS }}                                                                                     
                                                                                                           
[scheduling]                                                                                               
    cycling mode = integer                                                                                 
    initial cycle point = 1                                                                                
    final cycle point = {{ CYCLES }}                                                                       
    [[graph]]                                                                                              
        P1 = """                                                                                           
            d[-P1] => a => b<x> => c<y> => d                                                               
        """                                                                                                
                                                                                                           
[runtime]                                                                                                  
    [[b<x>, c<y>]]                                                                                         
        script = sleep {{ SLEEP }}                                                                         
    [[a, d]]    
```

The underlying culprits are the platforms and tokens methods:

![Screenshot from 2023-01-19 17-56-34](https://user-images.githubusercontent.com/16705946/213523464-fe89b032-5cc9-4437-ac3e-bf8e2b08b63f.png)

There are things we can do to reduce the overheads of these:

* https://github.com/cylc/cylc-flow/pull/5182
* https://github.com/cylc/cylc-flow/pull/5183

[...]

PR #5318, opened 2023-01-20
remove use of get platform at itask.__init__

wxtim:
A partial response to #5315 

Calling `get_platforms` when task proxies are created should not be necessary. That there are one or two places where it is implies some level of bug.

This PR attempts to deal with both these bugs and to provide a performance boost by not calling `get_platform` as often.

* If platform selection fails, `task_events_mgr._insert_task_job` and `task_job_mgr._prep_submit_task_job_error` want the platform name to put in the database, but if we haven't got a platform selected they need to get it from the task definition not itask.platform.
* Platform needs setting to localhost if in simulation mode.
* Platforms need to be got on restart.

Testing by @oliver-sanders  suggests that a branch like this has the potential to reduce CPU time:
Figure_1

**Check List**

- [x] `CHANGES.md` unchanged

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -1324,7 +1324,7 @@ class DataStoreMgr:
             name=tproxy.name,
             cycle_point=tproxy.cycle_point,
             execution_time_limit=job_conf.get('execution_time_limit'),
-            platform=job_conf.get('platform')['name'],
+            platform=job_conf.get('platform name'),
             job_runner_name=job_conf.get('job_runner_name'),
         )
         # Not all fields are populated with some submit-failures,
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -1285,6 +1285,7 @@ class DataStoreMgr:
         Args:
             name (str): Corresponding task name.
             point_string (str): Cycle point string
+            status: ??
             job_conf (dic):
                 Dictionary of job configuration used to generate
                 the job script.
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -1285,7 +1285,7 @@ class DataStoreMgr:
         Args:
             name (str): Corresponding task name.
             point_string (str): Cycle point string
-            status: ??
+            status: Job-state - i.e. running, submitted, failed, submit failed
             job_conf (dic):
                 Dictionary of job configuration used to generate
                 the job script.
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -751,10 +751,10 @@ class Scheduler:
         """Remote init for all submitted/running tasks in the pool."""
         self.task_job_mgr.task_remote_mgr.is_restart = True
         distinct_install_target_platforms = []
-        for itask in self.pool.get_tasks():
 
+        for itask in self.pool.get_tasks():
             # get a platform if the task proxy hasn't got one:
-            if itask.platform is None:
+            if not itask.platform:
                 itask.platform = get_platform(
                     itask.tdef.rtconfig, bad_hosts=self.bad_hosts
                 )
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -754,6 +754,11 @@ class Scheduler:
 
         for itask in self.pool.get_tasks():
 
+            if not itask.platform:
+                itask.platform = get_platform(
+                    itask.tdef.rtconfig, bad_hosts=self.bad_hosts
+                )
+
             itask.platform['install target'] = (
                 get_install_target_from_platform(itask.platform))
             if (
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -752,6 +752,13 @@ class Scheduler:
         self.task_job_mgr.task_remote_mgr.is_restart = True
         distinct_install_target_platforms = []
         for itask in self.pool.get_tasks():
+
+            # get a platform if the task proxy hasn't got one:
+            if itask.platform is None:
+                itask.platform = get_platform(
+                    itask.tdef.rtconfig, bad_hosts=self.bad_hosts
+                )
+
             itask.platform['install target'] = (
                 get_install_target_from_platform(itask.platform))
             if (
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -753,11 +753,6 @@ class Scheduler:
         distinct_install_target_platforms = []
 
         for itask in self.pool.get_tasks():
-            # get a platform if the task proxy hasn't got one:
-            if not itask.platform:
-                itask.platform = get_platform(
-                    itask.tdef.rtconfig, bad_hosts=self.bad_hosts
-                )
 
             itask.platform['install target'] = (
                 get_install_target_from_platform(itask.platform))
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -1284,6 +1284,13 @@ class TaskEventsManager():
         # not see previous submissions (so can't use itask.jobs[submit_num-1]).
         job_conf = itask.jobs[-1]
 
+        # Extract a resolved platform name if one is available from the itask
+        # else get an unresolved platform name from the task def:
+        if itask.platform:
+            platform_name = itask.platform['name']
+        else:
+            platform_name = itask.tdef.rtconfig['platform']
+
         # insert job into data store
         self.data_store_mgr.insert_job(
             itask.tdef.name,
@@ -1294,16 +1301,12 @@ class TaskEventsManager():
                 # NOTE: the platform name may have changed since task
                 # preparation started due to intelligent host (and or
                 # platform) selection
-                'platform': itask.platform,
+                'platform': platform_name,
             },
         )
         # update job in database
         # NOTE: the job must be added to the DB earlier so that Cylc can
         # reconnect with job submissions if the scheduler is restarted
-        if itask.platform:
-            platform_name = itask.platform['name']
-        else:
-            platform_name = itask.tdef.rtconfig['platform']
         self.workflow_db_mgr.put_update_task_jobs(
             itask,
             {
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -1300,6 +1300,10 @@ class TaskEventsManager():
         # update job in database
         # NOTE: the job must be added to the DB earlier so that Cylc can
         # reconnect with job submissions if the scheduler is restarted
+        if itask.platform:
+            platform_name = itask.platform['name']
+        else:
+            platform_name = itask.tdef.rtconfig['platform']
         self.workflow_db_mgr.put_update_task_jobs(
             itask,
             {
@@ -1309,7 +1313,7 @@ class TaskEventsManager():
                 # NOTE: the platform name may have changed since task
                 # preparation started due to intelligent host (and or
                 # platform) selection
-                'platform_name': itask.platform['name'],
+                'platform_name': platform_name,
             }
         )
 
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -969,7 +969,7 @@ class TaskJobManager:
 
         submit_delays = (
             rtconfig['submission retry delays']
-            or itask.platform and itask.platform['submission retry delays']
+            or itask.platform.get('submission retry delays')
         )
 
         for key, delays in [
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -985,10 +985,8 @@ class TaskJobManager:
 
     def _simulation_submit_task_jobs(self, itasks, workflow):
         """Simulation mode task jobs submission."""
-        sim_platform = get_platform()
-        sim_platform['name'] = 'SIMULATION'
         for itask in itasks:
-            itask.platform = sim_platform
+            itask.platform = {'name': 'SIMULATION'}
             itask.waiting_on_job_prep = False
             itask.submit_num += 1
             self._set_retry_timers(itask)
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -969,7 +969,7 @@ class TaskJobManager:
 
         submit_delays = (
             rtconfig['submission retry delays']
-            or itask.platform['submission retry delays']
+            or itask.platform and itask.platform['submission retry delays']
         )
 
         for key, delays in [
@@ -985,11 +985,13 @@ class TaskJobManager:
 
     def _simulation_submit_task_jobs(self, itasks, workflow):
         """Simulation mode task jobs submission."""
+        sim_platform = get_platform()
+        sim_platform['name'] = 'SIMULATION'
         for itask in itasks:
+            itask.platform = sim_platform
             itask.waiting_on_job_prep = False
             itask.submit_num += 1
             self._set_retry_timers(itask)
-            itask.platform = {'name': 'SIMULATION'}
             itask.summary['job_runner_name'] = 'SIMULATION'
             itask.summary[self.KEY_EXECUTE_TIME_LIMIT] = (
                 itask.tdef.rtconfig['job']['simulated run length']
@@ -1229,7 +1231,12 @@ class TaskJobManager:
             'submit_num': itask.submit_num,
             'try_num': itask.get_try_num(),
         })
+
         # create a DB entry for the submit-failed job
+        if itask.platform:
+            platform_name = itask.platform['name']
+        else:
+            platform_name = itask.tdef.rtconfig['platform']
         self.workflow_db_mgr.put_insert_task_jobs(
             itask,
             {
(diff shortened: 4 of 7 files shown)


cylc/flow/data_store_mgr.py
```diff
@@ -1233,6 +1233,7 @@ def insert_job(self, name, point_string, status, job_conf):
         Args:
             name (str): Corresponding task name.
             point_string (str): Cycle point string
+            status: ??
```
wxtim, line 1236: Not really me - it's been added to the list of args at somepoint without documentation.

oliver-sanders: It's the job state i.e. `running`, `submitted`, `failed`, `submit-failed`.

MetRonnie:
Needs rebasing onto 8.1.x?

cylc/flow/task_job_mgr.py
```diff
@@ -969,7 +969,7 @@ def _set_retry_timers(
 
         submit_delays = (
             rtconfig['submission retry delays']
-            or itask.platform['submission retry delays']
+            or itask.platform and itask.platform['submission retry delays']
```
MetRonnie, line 972:
```suggestion
            or itask.platform.get('submission retry delays')
```

cylc/flow/data_store_mgr.py
wxtim:
```suggestion
status: Job-state - i.e. running, submitted, failed, submit failed
```

cylc/flow/task_job_mgr.py
wxtim: What does `None.get` do? If the answer isn't safe then this is not right.

MetRonnie:
Conflicting

tests/functional/platforms/08-warn-if-regex-matches-localhost.t
```diff
@@ -46,16 +46,13 @@ __HEREDOC__
 ERR_STR='cannot be defined using a regular expression'
 
 TEST_NAME="${TEST_NAME_BASE}-validate"
-run_fail "${TEST_NAME}" cylc validate "${RND_WORKFLOW_SOURCE}"
-grep_ok "${ERR_STR}" \
-    "${TEST_NAME}.stderr" -F
+run_ok "${TEST_NAME}" cylc validate "${RND_WORKFLOW_SOURCE}"
```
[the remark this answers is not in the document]
wxtim: Because platform selection only happens at job submission time.

cylc/flow/task_proxy.py
```diff
@@ -236,7 +236,10 @@ def __init__(
 
         self.local_job_file_path: Optional[str] = None
 
-        self.platform = None
+        if data_mode:
+            self.platform = {}
+        else:
+            self.platform = None
```
MetRonnie: This looks redundant to me. I think it should be either `None` or `{}` in any case

wxtim: I agree.

cylc/flow/scheduler.py
```diff
@@ -751,7 +751,14 @@ def restart_remote_init(self):
         """Remote init for all submitted/running tasks in the pool."""
         self.task_job_mgr.task_remote_mgr.is_restart = True
         distinct_install_target_platforms = []
+
         for itask in self.pool.get_tasks():
+            # get a platform if the task proxy hasn't got one:
+            if not itask.platform:
+                itask.platform = get_platform(
+                    itask.tdef.rtconfig, bad_hosts=self.bad_hosts
+                )
```
oliver-sanders, line 760: I think that this is a possible bug.

If a running task does *not* have a platform set, we cannot safely guess it from the configuration because the configuration may have changed during restart. Also if the platform is a `platform-group` we may assign the wrong platform to the task. Either case would result in a polling failure.

It's ok for tasks not to have a platform set on restart provided they are in the waiting state (because we wouldn't expect the platform to be set at that point).

cylc/flow/task_job_mgr.py
```diff
@@ -985,11 +985,13 @@ def _set_retry_timers(
...
     def _simulation_submit_task_jobs(self, itasks, workflow):
         """Simulation mode task jobs submission."""
+        sim_platform = get_platform()
+        sim_platform['name'] = 'SIMULATION'
         for itask in itasks:
+            itask.platform = sim_platform
             itask.waiting_on_job_prep = False
             itask.submit_num += 1
             self._set_retry_timers(itask)
-            itask.platform = {'name': 'SIMULATION'}
```
oliver-sanders, line 992: If this is all that's needed for simulation mode, I don't think we need to run `get_platform` at all?

cylc/flow/task_events_mgr.py
```diff
@@ -1284,6 +1284,13 @@ def _insert_task_job(
         # not see previous submissions (so can't use itask.jobs[submit_num-1]).
         job_conf = itask.jobs[-1]
 
+        # Extract a resolved platform name if one is available from the itask
+        # else get an unresolved platform name from the task def:
+        if itask.platform:
+            platform_name = itask.platform['name']
+        else:
+            platform_name = itask.tdef.rtconfig['platform']
```
oliver-sanders, line 1292: In what case was `itask.platform` not set?

Logically the platform must have been set before the job was submission was made.

cylc/flow/task_job_mgr.py
```diff
@@ -1229,7 +1231,12 @@ def _prep_submit_task_job_error(self, workflow, itask, action, exc):
             'submit_num': itask.submit_num,
             'try_num': itask.get_try_num(),
         })
+
         # create a DB entry for the submit-failed job
+        if itask.platform:
+            platform_name = itask.platform['name']
+        else:
+            platform_name = itask.tdef.rtconfig['platform']
```
oliver-sanders, line 1239: In what case was `itask.platform` not set?

Logically the platform must have been set before the job was submission could fail? Unless maybe platform selection errored?

cylc/flow/task_job_mgr.py
```diff
@@ -985,11 +985,13 @@ def _set_retry_timers(
 
     def _simulation_submit_task_jobs(self, itasks, workflow):
         """Simulation mode task jobs submission."""
+        sim_platform = get_platform()
+        sim_platform['name'] = 'SIMULATION'
```
wxtim, line 989:
```suggestion
```

cylc/flow/task_job_mgr.py
```diff
@@ -985,11 +985,13 @@ def _set_retry_timers(
...
         for itask in itasks:
+            itask.platform = sim_platform
```
wxtim, line 991:
```suggestion
            itask.platform = {'name': 'SIMULATION'}
```

cylc/flow/task_events_mgr.py
wxtim: When it's called by `_process_message_submit_failed`.

cylc/flow/task_job_mgr.py
wxtim: Yes.

cylc/flow/task_job_mgr.py
wxtim: When the submit task job error is a caused by a failure to select a platform - see line 1132

cylc/flow/scheduler.py
wxtim, line 760:
```suggestion
```

cylc/flow/task_events_mgr.py
oliver-sanders: The platform gets set during submission:

https://github.com/cylc/cylc-flow/blob/a52943a034b195c08ffd0195636a80d94e7401b1/cylc/flow/task_job_mgr.py#L279-L280

So it should be set for submit-failed jobs. This functionality is tested so it definitely works.

Have you identified a scenario where the submission fails before the platform is set there?

oliver-sanders: Proved to myself that the platform is set with this example:

```diff
diff --git a/cylc/flow/task_events_mgr.py b/cylc/flow/task_events_mgr.py
index ea2c7c1be..3b177cb50 100644
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -1207,6 +1207,7 @@ class TaskEventsManager():
 
         # Register newly submit-failed job with the database and datastore.
         job_d = itask.tokens.duplicate(job=str(itask.submit_num)).relative_id
+        print('#', itask.platform)
         self._insert_task_job(itask, event_time, self.JOB_SUBMIT_FAIL_FLAG)
         self.data_store_mgr.delta_job_state(job_d, TASK_STATUS_SUBMIT_FAILED)
 ```

```
[scheduling]    
    [[graph]]
        R1 = foo

[runtime]
    [[foo]]
        platform = foo
```

```console
(cylc8) [cylc-run] $ cylc play submit-failed --no-detach

 ▪ ■  Cylc Workflow Engine 8.1.1.dev

...

2023-01-25T15:52:37Z INFO - [1/foo waiting(runahead) job:00 flows:1] => waiting
2023-01-25T15:52:37Z INFO - [1/foo waiting job:00 flows:1] => waiting(queued)
2023-01-25T15:52:37Z INFO - [1/foo waiting(queued) job:00 flows:1] => waiting
2023-01-25T15:52:37Z INFO - [1/foo waiting job:01 flows:1] => preparing
2023-01-25T15:52:37Z ERROR - [jobs-submit cmd] (platform not defined)
    [jobs-submit ret_code] 1
    [jobs-submit err] No matching platform "foo" found
2023-01-25T15:52:37Z CRITICAL - [1/foo preparing job:01 flows:1] submission
    failed
2023-01-25T15:52:37Z INFO - [1/foo preparing job:01 flows:1] => submit-failed
# OrderedDictWithDefaults([('meta', OrderedDictWithDefaults()), ('hosts', ['localhost']), ('job runner', 'background'), ('job runner command template', None), ('shell', '/bin/bash'), ('communication method', 'zmq'), ('submission polling intervals', [900.0]), ('submission retry delays', None), ('execution polling intervals', [900.0]), ('execution time limit polling intervals', [60.0, 120.0, 420.0]), ('ssh command', 'ssh -oBatchMode=yes -oConnectTimeout=10'), ('rsync command', 'rsync'), ('use login shell', True), ('cylc path', None), ('global init-script', None), ('copyable environment variables', ''), ('retrieve job logs', None), ('retrieve job logs command', 'rsync -a'), ('retrieve job logs max size', None), ('retrieve job logs retry delays', []), ('tail command template', 'tail -n +1 -F %(filename)s'), ('err tailer', None), ('out tailer', None), ('err viewer', None), ('out viewer', None), ('job name length maximum', None), ('install target', None), ('clean job submission environment', False), ('job submission environment pass-through', []), ('job submission executable paths', []), ('max batch submit size', 100), ('selection', OrderedDictWithDefaults([('method', 'definition order')])), ('directives', OrderedDictWithDefaults()), ('name', 'localhost')])
2023-01-25T15:52:37Z ERROR - Incomplete tasks:
2023-01-25T15:52:37Z CRITICAL - Workflow stalled
2023-01-25T15:52:37Z WARNING - PT1H stall timer starts NOW
```

cylc/flow/task_job_mgr.py
oliver-sanders: The platform is set for me using the same example as in the previous comment and this diff:

```diff
 diff --git a/cylc/flow/task_events_mgr.py b/cylc/flow/task_events_mgr.py
index ea2c7c1be..3b177cb50 100644
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -1207,6 +1207,7 @@ class TaskEventsManager():
 
         # Register newly submit-failed job with the database and datastore.
         job_d = itask.tokens.duplicate(job=str(itask.submit_num)).relative_id
+        print('#', itask.platform)
         self._insert_task_job(itask, event_time, self.JOB_SUBMIT_FAIL_FLAG)
         self.data_store_mgr.delta_job_state(job_d, TASK_STATUS_SUBMIT_FAILED)
 
diff --git a/cylc/flow/task_job_mgr.py b/cylc/flow/task_job_mgr.py
index da995bac8..1f000531d 100644
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -1229,6 +1229,7 @@ class TaskJobManager:
             'submit_num': itask.submit_num,
             'try_num': itask.get_try_num(),
         })
+        print('%', itask.platform)
         # create a DB entry for the submit-failed job
         self.workflow_db_mgr.put_insert_task_jobs(
             itask,
```

cylc/flow/scheduler.py
wxtim: If you don't have it the next line down will break:
```
itask.platform['install target'] = (
                get_install_target_from_platform(itask.platform))
 ```

File: cylc/flow/task_events_mgr.py (hunk 7 above)
Comment (wxtim) [reply cylc/flow/task_events_mgr.py:1292]: With your example, and your global config I get this total failure:

```
2023-01-25T16:16:16Z ERROR - name 'platform_name' is not defined
      File "/net/home/h02/tpilling/metomi/cylc-flow/cylc/flow/task_job_mgr.py", line 1165, in _prep_submit_task_job
        platform = get_platform(
      File "/net/home/h02/tpilling/metomi/cylc-flow/cylc/flow/platforms.py", line 143, in get_platform
        return platform_from_name(task_conf['platform'], bad_hosts=bad_hosts)
      File "/net/home/h02/tpilling/metomi/cylc-flow/cylc/flow/platforms.py", line 249, in platform_from_name
        raise PlatformLookupError(
    cylc.flow.exceptions.PlatformLookupError: No matching platform "foo" found
      File "/net/home/h02/tpilling/metomi/cylc-flow/cylc/flow/scheduler.py", line 634, in run_scheduler
        await self.main_loop()
      File "/net/home/h02/tpilling/metomi/cylc-flow/cylc/flow/scheduler.py", line 1617, in main_loop
        self.release_queued_tasks()
      File "/net/home/h02/tpilling/metomi/cylc-flow/cylc/flow/scheduler.py", line 1334, in release_queued_tasks
      File "/net/home/h02/tpilling/metomi/cylc-flow/cylc/flow/task_job_mgr.py", line 270, in submit_task_jobs
        prepared_tasks, bad_tasks = self.prep_submit_task_jobs(
      File "/net/home/h02/tpilling/metomi/cylc-flow/cylc/flow/task_job_mgr.py", line 242, in prep_submit_task_jobs
        prep_task = self._prep_submit_task_job(
      File "/net/home/h02/tpilling/metomi/cylc-flow/cylc/flow/task_job_mgr.py", line 1174, in _prep_submit_task_job
        self._prep_submit_task_job_error(
      File "/net/home/h02/tpilling/metomi/cylc-flow/cylc/flow/task_job_mgr.py", line 1247, in _prep_submit_task_job_error
        'platform_name': platform_name,
    NameError: name 'platform_name' is not defined

```

Comment (wxtim) [reply cylc/flow/task_events_mgr.py:1292]: On consideration of the following example:

```ini
#flow.cylc 
[runtime]
    [[mytask]]
        platform = foobarbaz
```

```ini
# global.cylc
[platforms]
        [[platform1]]
            ... some settings

# Definitely no regex matching "foo bar baz"
```

When mytask fails to submit because no matching platform can be found - how does the platform definition get populated?

cylc/flow/task_proxy.py
```diff
@@ -237,10 +236,7 @@ def __init__(
 
         self.local_job_file_path: Optional[str] = None
 
-        if data_mode:
```
dwsutherland, line 240: You'd probably want to remove from `__init__` args, and use in the data-store manager.

cylc/flow/scheduler.py
oliver-sanders: If a task doesn't have a platform, that means it hasn't been run yet, declaring a platform would be incorrect.

So I think this should be:

```
if not itask.platform:
    continue
```

cylc/flow/task_events_mgr.py
oliver-sanders: >> Have you identified a scenario where the submission fails before the platform is set there?
>
> Yes - if submission fails because a platform cannot be selected.

I think this is correct, if the job failed to submit then it doesn't have a platform.

wxtim: Put back - lets see what happens on testing!

hjoliver:
Getting a platforms-related traceback in several functional tests:
```
      File "/home/runner/work/cylc-flow/cylc-flow/cylc/flow/task_job_mgr.py", line 1239, in _prep_submit_task_job_error
        'platform_name': itask.platform['name'],
    KeyError: 'name'
```

wxtim:
> Getting a platforms-related traceback in several functional tests:
> 
> ```
>       File "/home/runner/work/cylc-flow/cylc-flow/cylc/flow/task_job_mgr.py", line 1239, in _prep_submit_task_job_error
>         'platform_name': itask.platform['name'],
>     KeyError: 'name'
> ```

job-submission/19-platform_select - fixed by https://github.com/cylc/cylc-flow/pull/5395

cylc/flow/data_store_mgr.py
```diff
@@ -1366,7 +1367,7 @@ def insert_job(self, name, point_string, status, job_conf):
             name=tproxy.name,
             cycle_point=tproxy.cycle_point,
             execution_time_limit=job_conf.get('execution_time_limit'),
-            platform=job_conf.get('platform')['name'],
+            platform=job_conf.get('platform name'),
```
oliver-sanders, line 1370: I don't think `platform name` is set in the job conf.

E.G. take a look at the simulation conf:

https://github.com/cylc/cylc-flow/blob/master/cylc/flow/task_job_mgr.py#L1346

So I'm guessing this is always evaluating as `None`?

The `job_conf` thinngy is now documented (recent addition) here:

https://cylc.github.io/cylc-doc/nightly_8.2/html/plugins/job-runners/index.html

wxtim:
Stale

Issue #5317, opened 2023-01-20
Raise Error when playing a running Cylc 7 workflow


datamel:
Niche bug but one that may crop up with users.

**Release version(s) and/or repository branch(es) affected?**

Current `Master` branch

**Steps to reproduce the bug**

Run a suite/workflow at Cylc 7.

Play this running workflow at Cylc 8
Traceback occurs: 

```
Traceback (most recent call last):
  File "…/cylc/flow/workflow_files.py", line 507, in detect_old_contact_file
    old_pid: str = contact_data[ContactFileFields.PID]
KeyError: 'CYLC_WORKFLOW_PID'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "…/cylc", line 33, in <module>
    sys.exit(load_entry_point('cylc-flow', 'console_scripts', 'cylc')())
  File "…/cylc/flow/scripts/cylc.py", line 653, in main
    execute_cmd(command, *cmd_args)
  File "…/cylc/flow/scripts/cylc.py", line 286, in execute_cmd
    entry_point.resolve()(*args)
  File "…/cylc/flow/terminal.py", line 232, in wrapper
    wrapped_function(*wrapped_args, **wrapped_kwargs)
  File "…/cylc/flow/scheduler_cli.py", line 624, in play
    return _play(parser, options, id_)
  File "…/cylc/flow/scheduler_cli.py", line 634, in _play
    return scheduler_cli(options, id_)
  File "…/cylc/flow/scheduler_cli.py", line 374, in scheduler_cli
    _resume(workflow_id, options)
  File "…/cylc/flow/scheduler_cli.py", line 433, in _resume
    detect_old_contact_file(workflow_id)
  File "…/cylc/flow/workflow_files.py", line 512, in detect_old_contact_file
    raise Exception(f'Found contact file with incomplete data:\n{exc}.')
Exception: Found contact file with incomplete data:
'CYLC_WORKFLOW_PID'.
```

**Expected behavior**

Expect clean error raised

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

oliver-sanders:
Closed by #5313

PR #5323, opened 2023-01-23
Ensure error is raised for playing running Cylc 7 workflow

datamel:
Closes https://github.com/cylc/cylc-flow/issues/5317

**Check List**

### DIFF
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -44,6 +44,7 @@ from cylc.flow import LOG
 from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
 from cylc.flow.exceptions import (
     CylcError,
+    CylcVersionError,
     PlatformError,
     PlatformLookupError,
     ServiceFileError,
@@ -502,11 +503,15 @@ def detect_old_contact_file(
             return
 
     try:
+        old_version: str = contact_data[ContactFileFields.VERSION]
         old_host: str = contact_data[ContactFileFields.HOST]
         old_port: str = contact_data[ContactFileFields.PORT]
         old_pid: str = contact_data[ContactFileFields.PID]
         old_cmd: str = contact_data[ContactFileFields.COMMAND]
     except KeyError as exc:
+        # scenario - playing a workflow which is already running with Cylc 7:
+        if old_version and int(old_version.split('.')[0]) < 8:
+            raise CylcVersionError(version=old_version, status="Running")
         # this shouldn't happen
         # but if it does re-raise the error as something more informative
         raise Exception(f'Found contact file with incomplete data:\n{exc}.')
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,8 @@ ones in. -->
 
 ### Fixes
 
+[#5314](https://github.com/cylc/cylc-flow/pull/5323) - Raise error when playing a workflow which is actively running with Cylc 7.
+
 [#5314](https://github.com/cylc/cylc-flow/pull/5314) - Fix broken
 command option: `cylc vip --run-name`.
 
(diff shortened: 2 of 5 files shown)


oliver-sanders:

wxtim:
I don't think that you've rebased onto 8.1.x correctly - you will need to do `cylc rebase upstrea/8.1.x -i` and remove the commits unrelated to the PR.

datamel:
> * I don't think that you've rebased onto 8.1.x correctly - you will need to do `cylc rebase upstrea/8.1.x -i` and remove the commits unrelated to the PR.

I think once the merge to 8.1.x has been sorted the commits should go

> Can you do a changelog entry?

Yes but I'll wait until someone can reproduce. No point in fixing something that may not be broken.
To reproduce I set off a long-ish running workflow with Cylc 7, then activated Cylc 8 env (I pip installed too), the played the workflow that was running with cylc 7 but this time with Cylc 8. I got KeyError traceback, this just raises a slightly neater error.

oliver-sanders:
When you rebase back from master to an earlier branch (e.g. 8.1.x) you may drag some unwanted commits with you. There are four commits on this branch:

https://github.com/cylc/cylc-flow/pull/5323/commits

The unwanted commits can be safely removed with `git rebase -i`.

cylc/flow/workflow_files.py
```diff
@@ -44,6 +44,7 @@
 from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
 from cylc.flow.exceptions import (
     CylcError,
+    CylcVersionError,
```
datamel, line 47:
```suggestion
```

oliver-sanders:
I think `detect_old_contact` file fails for real Cylc 7 contact files.

https://github.com/cylc/cylc-flow/pull/5313#discussion_r1084285970

The solution on #5313 was to catch the error, log a warning about Cylc 8 not being able to start Cylc 7 workflows, then raise the error.

datamel:
> I think `detect_old_contact` file fails for real Cylc 7 contact files.

I am getting traceback after running with cylc 7:
```
  File "~/cylc-flow/cylc/flow/workflow_files.py", line 507, in detect_old_contact_file
    old_pid: str = contact_data[ContactFileFields.PID]
KeyError: 'CYLC_WORKFLOW_PID'
 ```

My cylc 7 contact file has a field `CYLC_SUITE_PROCESS`, not the  'CYLC_WORKFLOW_PID' this method is looking for, hence the key error.

> 
> [#5313 (comment)](https://github.com/cylc/cylc-flow/pull/5313#discussion_r1084285970)
> 
> The solution on #5313 was to catch the error, log a warning about Cylc 8 not being able to start Cylc 7 workflows, then raise the error.

Happy to close this and have it fixed on #5313, I  picked up the bug while reviewing it so makes sense to put the fix in there. 
For me, this traceback is still occurring on #5313 in this scenario. 

@wxtim was unable to reproduce this bug so can you confirm this is happening for you @oliver-sanders?

datamel:
I think the error is only being raised on `vr`, not just play on https://github.com/cylc/cylc-flow/pull/5313. Whereas this bug I am referring to is when you just `play` a running cylc 7 workflow.

oliver-sanders:
> Still happy for this to be fixed on @wxtim's branch.

Tim's branch fixes it for `cylc vr` but I'm not sure if it fixes it for `cylc play`.

oliver-sanders:
> Here is the traceback from the vr command on https://github.com/cylc/cylc-flow/pull/5313 with a running cylc7 workflow.

IMO this traceback is a good enough solution. So long as the "Note Cylc 8 can't restart Cylc 7 workflows" notice is in there in bold red. This traceback could result from other things and it's probably more bother than it's worth trying to tell the difference between a Cylc 7/8 contact file and all that junk.

So probably just need to make sure other commands i.e. `play` (and possibly others?) aren't affected. (note `cylc scan` has its own logic for filtering out Cylc 7 workflows).

datamel:
Closing this after the above discussion.

PR #5321, opened 2023-01-23
Optimize ID/tokens code

MetRonnie:
Profiling revealed `Tokens` computed properties and `detokenise()` were being called many times and so was acting as a bit of a bottleneck (within the bottleneck of data store manager's `increment_graph_window()` - #5315).

- Remove a couple of unnecessary nested comprehensions
- Precompute certain sets

Partially addresses #5315 & #5183

#### Context

<details>

<summary>Click to see `flow.cylc`</summary>

```
#!Jinja2

{% set TASKS = TASKS | default(100) %}
{% set CYCLES = CYCLES | default(1) %}
{% set SLEEP = SLEEP | default(1) %}

[meta]
    description = """
        Example scaling workflow with {{ (TASKS * 2) + 2 }} tasks and {{ (TASKS ** 2) + (TASKS * 2) }}
        dependencies per cycle which runs for {{ CYCLES }} cycles.
    """

[task parameters]
    x = 1..{{ TASKS }}
    y = 1..{{ TASKS }}

[scheduling]
    cycling mode = integer
    initial cycle point = 1
    final cycle point = {{ CYCLES }}
    [[graph]]
        P1 = """
            d[-P1] => a => b<x> => c<y> => d
        """

[runtime]
    [[b<x>, c<y>]]
        script = sleep {{ SLEEP }}
    [[a, d]]
```

</details>

```
$ cylc play efficiency/ --profile -s 'TASKS=16'
```

c1

Expanded to `edge_id`:

c2

#### After

c3

#### Check list

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] No dependency changes 
- [x] No new tests needed
- [x] No changelog entry - presumably will be part of a bunch of efficiency improvements for 8.1.1
- [x] No docs
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #5325, opened 2023-01-24
tokens: speed up duplication of Tokens objects

oliver-sanders:
Sometimes we need to duplicate a `Tokens` object.

Currently this works by doing `tokenise(detokenise(tokens))` which is really inefficient (I should know better, I've moaned about this in the cycler classes before!).

This PR switches to using the already parsed values held in the `tokens` object.

Using the example from #5315 with `-s TASKS=25 -s CYCLES=1` this reduces the time taken by `increment_graph_window` by ~50% *after* the changes from #5319 and #5321 have been applied.

**Before (with #5319 & #5321)**

Screenshot from 2023-01-24 11-32-10

**After (still with #5319 & #5321)**

Screenshot from 2023-01-24 11-32-21

**Check List**


cylc/flow/id.py
```diff
@@ -107,7 +107,10 @@ def __init__(
         if args:
             if len(args) > 1:
                 raise ValueError()
-            kwargs = tokenise(str(args[0]), relative)
+            if isinstance(args[0], str):
+                kwargs = tokenise(str(args[0]), relative)
```
MetRonnie, line 111:
```suggestion
                kwargs = tokenise(args[0], relative)
```

oliver-sanders: Derp.

hjoliver:
Nice!

datamel (commented):
In the integration tests, the `id` is currently a posix path which is breaking them with this change. I'm guessing mypy doesn't check tests.

hjoliver:
That's a genuine test fail, in `tests/integration/test_reinstall.py`:

```python
        if args:
            if len(args) > 1:
                raise ValueError()
            if isinstance(args[0], str):
                kwargs = tokenise(args[0], relative)
            else:
>               kwargs = dict(args[0])
E               TypeError: 'PosixPath' object is not iterable

cylc/flow/id.py:113: TypeError
```

hjoliver:
Fix?

```diff
diff --git a/tests/integration/test_reinstall.py b/tests/integration/test_reinstall.py
index 02be406bc..8b13c5574 100644
--- a/tests/integration/test_reinstall.py
+++ b/tests/integration/test_reinstall.py
@@ -80,7 +80,7 @@ def one_run(one_src, test_dir, run_dir):
     )
     return SimpleNamespace(
         path=w_run_dir,
-        id=w_run_dir.relative_to(run_dir),
+        id=str(w_run_dir.relative_to(run_dir)),
     )
```

hjoliver:
Argghhh, @datamel bet me to it!!! (she must have been working late!)

oliver-sanders:
Ta for the fix both :)

oliver-sanders:
Added a collective changelog for all three efficiency changes.

MetRonnie (approved):
(only linkcheck failing)

PR #5326, opened 2023-01-24
8.1.x Sync

datamel:


cylc/flow/__init__.py
```diff
@@ -46,7 +46,7 @@ def environ_init():
 
 environ_init()
 
-__version__ = '8.1.1.dev'
+__version__ = '8.2.0.dev'
```
wxtim, line 49: Surely not?

datamel:
```suggestion
__version__ = '8.1.1.dev'
```

oliver-sanders:
Because this PR is raised from the `8.1.x` branch, suggestions applied here are applied to the `8.1.x` branch.

To get around this, if there are conflicts to be resolved with a sync PR we have to create an intermediate branch like this:

```console
$ git checkout -b sync upstream/master
$ git merge upstream/8.1.x
$ git push origin sync
```

MetRonnie:
The problem here was the PR is master into 8.1.x - the wrong way round! Also it seems the branch protection rules are blocking force pushes but allowing normal pushes for some reason! Looking into it

oliver-sanders:
Cheers.

We could do with automating this at some point.

PR #5329, opened 2023-01-25
data_store: remove unnecessary pdeepcopy

oliver-sanders:
I think this `pdeepcopy` call is not needed because it is already being performed lower down in the stack:

Here's the call being removed:

https://github.com/cylc/cylc-flow/blob/a52943a034b195c08ffd0195636a80d94e7401b1/cylc/flow/data_store_mgr.py#L1279-L1282

And here's what is was being called on the output of:

https://github.com/cylc/cylc-flow/blob/a52943a034b195c08ffd0195636a80d94e7401b1/cylc/flow/data_store_mgr.py#L1222-L1228

Here's a `--profile` result from the same workflow used in #5328 which shows the `pdeepcopy` to be about half the cost of `insert_job`:

Screenshot 2023-01-25 at 13 39 54

So worth stripping out if it's not needed.

Some of the Parsec methods are pretty expensive, particularly replicate unfortunately.

**Check List**

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -1328,10 +1328,10 @@ class DataStoreMgr:
         )
         # Not all fields are populated with some submit-failures,
         # so use task cfg as base.
-        j_cfg = pdeepcopy(self._apply_broadcasts_to_runtime(
+        j_cfg = self._apply_broadcasts_to_runtime(
             tp_tokens.relative_id,
             self.schd.config.cfg['runtime'][tproxy.name]
-        ))
+        )
         for key, val in job_conf.items():
             j_cfg[key] = val
         j_buf.runtime.CopyFrom(runtime_from_config(j_cfg))


hjoliver:
This does seem to break some stuff.

Maybe because the lower-level deepcopy is only done if there are active broadcasts?

dwsutherland:
The tests failed cause I actually use `pdeepcopy` to set default values by transformation to `OrderedDictWithDefaults`.. so instead of `None`:
```
2023-01-25T20:30:12+05:30 ERROR - argument of type 'NoneType' is not iterable
```
 the default will be according to the workflow cfg default..

dwsutherland:
To avoid `pdeepcopy` we would need to set our own defaults if the field doesn't exist, by using `.get()` for the dict look ups here:
https://github.com/cylc/cylc-flow/blob/a52943a034b195c08ffd0195636a80d94e7401b1/cylc/flow/data_store_mgr.py#L196

oliver-sanders:
Cheers for looking in, I'll try and see if there's a way around this if I get a chance.

Cylc 8's job overheads are a bit steep causing poor throuput, we could do with cutting back a bit. One of the most substantial overheads is the Bash syntax checking in the job file submission and job file writing both of which are done synchronously.

oliver-sanders:
Closing, capturing the issue in - https://github.com/cylc/cylc-flow/issues/5373

PR #5330, opened 2023-01-25
DB interface & safety improvements

MetRonnie:
Partially addresses same issue as #5313

- **Do not create tables by default**, only if specified via kwarg
- Provide context manager functionality to automatically close DB connection
- A `WorkflowDatabaseManager` should only be created by `Scheduler`; just use `CylcWorkflowDAO` elsewhere, or certain static methods on `WorkflowDatabaseManager`

**Check List**

- [x] No changelog entry needed (invisible to users)
- [x] No docs entry needed

### DIFF
--- a/cylc/flow/network/scan.py
+++ b/cylc/flow/network/scan.py
@@ -545,14 +545,12 @@ async def workflow_params(flow):
 
     # NOTE: use the public DB for reading
     # (only the scheduler process/thread should access the private database)
-    db_file = Path(get_workflow_run_dir(flow['name'], 'log', 'db'))
+    db_file = Path(get_workflow_run_dir(
+        flow['name'], WorkflowFiles.LogDir.DIRNAME, WorkflowFiles.LogDir.DB
+    ))
     if db_file.exists():
-        dao = CylcWorkflowDAO(db_file, is_public=False)
-        try:
-            dao.connect()
+        with CylcWorkflowDAO(db_file, is_public=True) as dao:
             dao.select_workflow_params(_callback)
             flow['workflow_params'] = params
-        finally:
-            dao.close()
 
     return flow
--- a/cylc/flow/rundb.py
+++ b/cylc/flow/rundb.py
@@ -21,11 +21,14 @@ from os.path import expandvars
 from pprint import pformat
 import sqlite3
 import traceback
-from typing import List, Tuple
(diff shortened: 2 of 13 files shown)


cylc/flow/rundb.py
```diff
@@ -318,26 +321,44 @@ class CylcWorkflowDAO:
...
         """Initialise database access object.
 
+        An instance of this class can also be opened as a context manager
+        which will automatically close the DB connection.
+
         Args:
-            db_file_name (str): Path to the database file.
-            is_public (bool): If True, allow retries, etc.
+            db_file_name: Path to the database file.
+            is_public: If True, allow retries, etc.
```
wxtim, line 337: Grammatical peeve - if it does something else we should document it.
```suggestion
    is_public: If True, allow retries.
```

wxtim (approved):
I've learned a fair bit about SQLite this morning. It Looks sane - will wait for tests  to   pass on CI before downloading and playing with branch.

MetRonnie:
(Test failure is just linkcheck)

cylc/flow/network/scan.py
```diff
@@ -545,14 +545,12 @@ def _callback(_, entry):
 
     # NOTE: use the public DB for reading
     # (only the scheduler process/thread should access the private database)
-    db_file = Path(get_workflow_run_dir(flow['name'], 'log', 'db'))
+    db_file = Path(get_workflow_run_dir(
+        flow['name'], WorkflowFiles.LogDir.DIRNAME, WorkflowFiles.LogDir.DB
+    ))
     if db_file.exists():
-        dao = CylcWorkflowDAO(db_file, is_public=False)
```
hjoliver, line 550: Huh, scan was reading private DBs in spite of the comment above? 😬

cylc/flow/rundb.py
hjoliver: I think that (retries) is the main consequence (along with associated different reporting of db write errors).

hjoliver: I'll commit the suggestion in anticipation of approving this PR...

hjoliver (approved):
Looks good.

cylc/flow/network/scan.py
MetRonnie: It was until https://github.com/cylc/cylc-flow/commit/b338a91d5db0e1e60ca7ca16c3be28c86c630693

PR #5333, opened 2023-01-30
fix to make setup.cfg work with setuptools v67.0.0

wxtim:
A very small change to the format of the setup.cfg to make it work with the latest version of setuptools.

Non fast-tests cancelled.

See also https://github.com/pypa/setuptools/blob/v67.0.0/CHANGES.rst

(Still digging to find out what part of PEP508 or 440 was the problem)

Caused by [this PR](https://github.com/pypa/packaging/pull/634) fixing.

[Reference: Using setuptools comments to make requirements py version dependent](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -78,7 +78,7 @@ install_requires =
     rx
     promise
     # Once Python 3.11 released -'; python_version<"3.11"'
-    tomli>=2.*
+    tomli>=2
 
 [options.packages.find]
 include = cylc*
--- a/setup.cfg
+++ b/setup.cfg
@@ -72,13 +72,12 @@ install_requires =
     protobuf>=4.21.2,<4.22.0
     psutil>=5.6.0
     pyzmq==22.*
-    setuptools>=49
+    setuptools>=49, <67
     urwid==2.*
     # unpinned transient dependencies used for type checking
     rx
     promise
-    # Once Python 3.11 released -'; python_version<"3.11"'
-    tomli>=2
+    tomli>=2; python_version < "3.11"
 
 [options.packages.find]
 include = cylc*
--- a/setup.cfg
+++ b/setup.cfg
@@ -72,6 +72,7 @@ install_requires =
     protobuf>=4.21.2,<4.22.0
     psutil>=5.6.0
     pyzmq==22.*
+    # https://github.com/pypa/setuptools/issues/3802
     setuptools>=49, <67
     urwid==2.*
     # unpinned transient dependencies used for type checking
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -16,7 +16,7 @@ dependencies:
   - psutil >=5.6.0
   - python
   - pyzmq >=22,<23
-  - setuptools >=49
+  - setuptools >=49, <67
   - urwid >=2,<3
   # Add # [py<3.11] for tomli once Python 3.11 Released
   - tomli >=2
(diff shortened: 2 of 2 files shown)


MetRonnie:
> pkg_resources.DistributionNotFound: The 'metomi-isodatetime==1!3.0.*' distribution was not found and is required by cylc-flow

https://github.com/cylc/cylc-flow/actions/runs/4042658912/jobs/6950630795#step:10:358

Seems to be a bug in `setuptools` or `packaging` for Python 3.7?

wxtim:
> > pkg_resources.DistributionNotFound: The 'metomi-isodatetime==1!3.0.*' distribution was not found and is required by cylc-flow
> 
> https://github.com/cylc/cylc-flow/actions/runs/4042658912/jobs/6950630795#step:10:358
> 
> Seems to be a bug in `setuptools` or `packaging` for Python 3.7?

Looks like it :(

setup.cfg
```diff
@@ -67,7 +67,8 @@ install_requires =
     graphene>=2.1,<3
     # Note: can't pin jinja2 any higher than this until we give up on Cylc 7 back-compat
     jinja2==3.0.*
-    metomi-isodatetime==1!3.0.*
+    metomi-isodatetime==1!3.0.*; python_version > "3.7"
+    metomi-isodatetime==1!3; python_version == "3.7"
```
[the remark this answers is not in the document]
wxtim: OK - good point - that works!

setup.cfg
```diff
@@ -72,13 +72,13 @@ install_requires =
     protobuf>=4.21.2,<4.22.0
     psutil>=5.6.0
     pyzmq==22.*
-    setuptools>=49
+    setuptools>=49; python_version > "3.7"
+    setuptools>=49, <67; python_version == "3.7"
```
MetRonnie, line 76: Ah actually I have just looked at each fast tests' logs and found that for Python 3.8-10, it's using a cached version of setuptools that is lower than 67 anyway, so maybe the problem will still be present for those versions too

MetRonnie: Yep, problem present on all versions of Python unfortunately: https://github.com/MetRonnie/cylc-flow/actions/runs/4043433284/jobs/6952341119

wxtim, line 76:
```suggestion
    setuptools>=49, <67
```

setup.cfg
```diff
@@ -72,13 +72,12 @@ install_requires =
     protobuf>=4.21.2,<4.22.0
     psutil>=5.6.0
     pyzmq==22.*
-    setuptools>=49
+    setuptools>=49, <67
```
oliver-sanders, line 75: Please apply this to the `conda-environment.yml` file two (they should always match!)

setup.cfg
```diff
@@ -72,13 +72,12 @@ install_requires =
     protobuf>=4.21.2,<4.22.0
     psutil>=5.6.0
     pyzmq==22.*
-    setuptools>=49
+    setuptools>=49, <67; python_version == "3.7"
```
oliver-sanders, line 75: I think the conclusion was that this issue is generic to all Python versions? @MetRonnie

oliver-sanders (approved):

setup.cfg
MetRonnie, line 75:
```suggestion
    # https://github.com/pypa/setuptools/issues/3802
    setuptools>=49, <67
```

oliver-sanders:
This PR was raised against master but it should have been "8.1.x", I didn't spot that.

Please go through the check-list with PRs, the last check covers this.

PR #5334, opened 2023-01-30
Apply prerequisite changes to spawned tasks, on reload & restart

hjoliver:
On restart we assume that all prerequisites of a spawned task are recorded in the DB. This will not be the case on restarting after adding new dependencies to that task.

Example:

```ini
[scheduling]
    [[graph]]
        R1 = """
            a => z
            b => z
            # c => z
        """
[runtime]
    [[a]]
         script = "sleep 20"
    [[b, c, z]]
```

To reproduce the bug:

- install and play
- do `cylc stop --now` when `a` is still running and `z` has been spawned by `b:succeeded`
- uncomment `# c => z` to add a new prerequisite (`c`) task for `z`
- reinstall and play: scheduler will crash

The fix: if a prerequisite does not exist in the DB, assume that it is unsatisfied.

**Check List**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -69,6 +69,15 @@ Rose options (`-O`, `-S` & `-D`) with `cylc view`.
 for `cylc lint`.
 
 -------------------------------------------------------------------------------
+## __cylc-8.1.2 (<span actions:bind='release-date'>Upcoming</span>)__
+
+### Fixes
+
+[#5334](https://github.com/cylc/cylc-flow/pull/5334) - Fix to prevent scheduler
+crash if an already-spawned tasks has new prerequisites added before restart.
+
+
+-------------------------------------------------------------------------------
 ## __cylc-8.1.1 (<span actions:bind='release-date'>Released 2023-01-31</span>)__
 
 ### Fixes
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -74,7 +74,7 @@ for `cylc lint`.
 ### Fixes
 
 [#5334](https://github.com/cylc/cylc-flow/pull/5334) - Fix to prevent scheduler
-crash if an already-spawned tasks has new prerequisites added before restart.
+crash if already-spawned tasks have new prerequisites added before a restart.
 
 
 -------------------------------------------------------------------------------
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -73,8 +73,8 @@ for `cylc lint`.
 
 ### Fixes
 
-[#5334](https://github.com/cylc/cylc-flow/pull/5334) - Fix to prevent scheduler
-crash if already-spawned tasks have new prerequisites added before a restart.
+[#5334](https://github.com/cylc/cylc-flow/pull/5334) - Apply graph prerequisite 
+changes to already-spawned tasks after reload or restart.
 
 
 -------------------------------------------------------------------------------
--- a/tests/integration/test_task_pool.py
+++ b/tests/integration/test_task_pool.py
@@ -17,6 +17,7 @@
 from cylc.flow import CYLC_LOG
 from copy import deepcopy
 import logging
+from pathlib import Path
 import pytest
 from pytest import param
 from typing import AsyncGenerator, Callable, Iterable, List, Tuple, Union
@@ -33,6 +34,9 @@ from cylc.flow.task_state import (
     TASK_STATUS_RUNNING,
     TASK_STATUS_SUCCEEDED,
 )
+
+from cylc.flow.pathutil import get_cylc_run_dir, get_workflow_run_dir
+from .utils.flow_tools import _make_flow
 
 
 # NOTE: foo and bar have no parents so at start-up (even with the workflow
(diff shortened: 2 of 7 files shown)


hjoliver:
(Rebased to pick up #5336)

hjoliver:
Thanks for the fix @oliver-sanders

hjoliver:
> I haven't looked into removing prereqs ...

This works already, for restart. Checking on tests ...

hjoliver:
I expect the equivalent reload will be fine. Already-spawned task proxies do have prerequisites updated on reload, and it was just the DB check on restart that caused the problem.

hjoliver:
(Reload fixed, still working on the integration test for that ...)

hjoliver:
After considerable pfaffing about, this should be good to go now.

hjoliver:
@dwsutherland - could you review at this end? It touches the datastore a bit.

oliver-sanders:
I tried testing with this example:

```
[scheduler]                    
    allow implicit tasks = True    
                               
[scheduling]                   
    initial cycle point = 1    
    cycling mode = integer    
    [[graph]]               
        P1 = """            
            a => b          
            a => c          
        """
```

Then added this dependency:

```
            a[-P1] => a    
```

And reloaded.

Cylc didn't crash, but the workflow became stuck (no stall event detected):

Screenshot from 2023-02-16 15-42-43

I think this is because `34/a` had succeeded and left the pool before the reload happened.

Rather than defaulting to `False` should we run it through the DB or re-spawn the dependency (which would run it through the DB)?

oliver-sanders:
The issue I mentioned above isn't great, but that said, it is a lot better than crashing. If it's difficult we could bump the DB stuff to a later PR as the pressure on release is quite high due to the GUI bug.

dwsutherland (approved):
Looks good.. not sure how I missed this one (as it seems like a common operation)

DB search and/or unsatisfied prereq spawn on follow-up? (as per Oliver's findings)

oliver-sanders:
I think I have a solution to the issue I reported, will put something up shortly.

oliver-sanders:
Gotta fix here: https://github.com/hjoliver/cylc-flow/pull/31

However, we still need to work out how to square this with the data store which still defaults to `False` after my change.

hjoliver:
(Thanks, just back to this now, I'll take a look at the problem you found, and your fix ...)

hjoliver:
OK what's going on is, `a` is initially parent-less so it gets spawned to the runahead limit. Then after the reload it has a parent (it's own previous instance) and we update the already-spawned instances to give them the new prerequisite. The new prerequisite should be satisfied for the first instance, but it isn't - because under SoD prerequisite status only gets updated on demand, at the moment the corresponding upstream output gets completed.

hjoliver:
Agreed with your fix @oliver-sanders ; merged it to this branch.

> One small issue, I don't think we can provide the prerequsite satisfied message accurately i.e. we can't tell the difference between succeeded naturally and whatever the alternative is. However, this is minor.

Yeah, minor. I think "satisfied naturally" is the only one used at the moment in Cylc 8 anyhow. Although that will change with https://cylc.github.io/cylc-admin/proposal-cylc-set.html

hjoliver:
From the side-PR:

>   I'm not sure how to square this with the data store without duplicating the database calls.

Presumably I just need to call `delta_task_prerequisite(task)` in the datastore manager.

>    This PR marks all satisfied prerequisites as satisfied naturally. I think the information is lost by this point, is there a more suitable value e.g satisfied from DB?

(See above) This isn't used for anything at the moment. For now I'll just make something up to distinguish the two options.

**TODO**
- update datastore
- new test for this case
- "satisfied by DB" (or similar)

hjoliver:
> Gotta solution to the last three bullet points here - https://github.com/hjoliver/cylc-flow/pull/32

Great, thanks.

hjoliver:
@oliver-sanders - applied your tweaks and rebased. 

`_remote_background` functional tests failing unrelated infrastructure reasons.

CHANGES.md
```diff
@@ -68,6 +68,15 @@ Rose options (`-O`, `-S` & `-D`) with `cylc view`.
 [#5363](https://github.com/cylc/cylc-flow/pull/5363) Improvements and bugfixes
 for `cylc lint`.
 
+-------------------------------------------------------------------------------
+## __cylc-8.1.2 (<span actions:bind='release-date'>Upcoming</span>)__
```
oliver-sanders, line 72: Should be 8.1.3.

tests/integration/test_task_pool.py
```diff
@@ -593,3 +593,368 @@ async def test_load_db_bad_platform(
...
+        (itask.tokens['cycle'], itask.tokens['task'], itask.state.status)
+        for itask in schd.pool.get_all_tasks()
+    )
+
+
+@pytest.mark.parametrize(
+    'graph_1, graph_2, '
+    'expected_1, expected_2, expected_3, expected_4',
+    [
+        param(  # Restart after adding a prerequisite to task z
```
wxtim, line 617: IMO these should have an `id=` kwarg - 

Imagine trying to run

```bash
pytest 'tests/i/test_task_pool.py::test_restart_prereqs[a => z\n               b => z-a => z\n               b => z\n               c => z-expected_10-expected_20-expected_30-expected_40]'

```

tests/integration/test_task_pool.py
```diff
@@ -593,3 +593,368 @@ async def test_load_db_bad_platform(
...
+            '''a => z
+               b => z''',
+            '''a => z
+               b => z
+               c => z''',
+            [
+                ('1', 'a', 'running'),
```
wxtim, line 624: Given that everything is at CP = 1, could we just test `output[1:] = expect` and not include the '1'? 

It's marginal, but I think it cuts down on the amount written. Don't care that much.

wxtim (approved):
Approving - It looks good - I've made a couple of stylistic points, but

- [x] Reproduced the bug, failed to reproduce it on branch.
- [x] Read Code

PR #5336, opened 2023-01-30
Fix setup.cfg for setuptools 67.

hjoliver:
`pip install .` results in the latest `setuptools` (version 67) being installed, which rejects bad syntax in `setup.cfg`: `foo>=3.*` (the wildcard is only allowed with `==`).

This is calling CI to fail on GitHub, and reproduced locally.

**Check List**


hjoliver:
(Doh, beaten too it. Duplicate of #5333)

PR #5343, opened 2023-02-02
Fix localhost platform matching

### LINKED ISSUE #5342: platform name can't start with "localhost"?
We have a platform defined with name `localhost_slurm` that is supposed to submit jobs to Slurm locally (i.e. on the scheduler host) ... but it ignores the job runner setting and just runs local background jobs.

If I change the platform name to (e.g.) `localhxst_slurm` it works as intended.

wxtim:
Closes #5342

**Check List**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,9 @@ Fixes `cylc set-verbosity`.
 absence of `job name length maximum` in PBS platform settings would cause
 Cylc to crash when preparing the job script.
 
+[#5343](https://github.com/cylc/cylc-flow/pull/5343) - Fix a bug causing
+platform names to be checked as if they were hosts.
+
 -------------------------------------------------------------------------------
 ## __cylc-8.1.2 (<span actions:bind='release-date'>Released 2023-02-20</span>)__
 
@@ -37,12 +40,6 @@ Rose options (`-O`, `-S` & `-D`) with `cylc view`.
 
 [#5363](https://github.com/cylc/cylc-flow/pull/5363) Improvements and bugfixes
 for `cylc lint`.
-
--------------------------------------------------------------------------------
-## __cylc-8.1.2 (<span actions:bind='release-date'>Coming Soon</span>)__
-
-[#5343](https://github.com/cylc/cylc-flow/pull/5343) - Fix a bug causing
-platform names to be checked as if they were hosts.
 
 -------------------------------------------------------------------------------
 ## __cylc-8.1.1 (<span actions:bind='release-date'>Released 2023-01-31</span>)__
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,6 +39,12 @@ Rose options (`-O`, `-S` & `-D`) with `cylc view`.
 for `cylc lint`.
 
 -------------------------------------------------------------------------------
+## __cylc-8.1.2 (<span actions:bind='release-date'>Coming Soon</span>)__
+
+[#5343](https://github.com/cylc/cylc-flow/pull/5343) - Fix a bug causing
+platform names to be checked as if they were hosts.
+
+-------------------------------------------------------------------------------
 ## __cylc-8.1.1 (<span actions:bind='release-date'>Released 2023-01-31</span>)__
 
 ### Fixes
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -1125,7 +1125,9 @@ class TaskJobManager:
                 )
             else:
                 platform_name = self.task_remote_mgr.subshell_eval(
-                    rtconfig['platform'], PLATFORM_REC_COMMAND
+                    rtconfig['platform'],
+                    PLATFORM_REC_COMMAND,
+                    host_check=False
                 )
         except PlatformError as exc:
             itask.waiting_on_job_prep = False
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -58,8 +58,6 @@ from cylc.flow.parsec.util import (
 )
 from cylc.flow.pathutil import get_remote_workflow_run_job_dir
 from cylc.flow.platforms import (
-    HOST_REC_COMMAND,
-    PLATFORM_REC_COMMAND,
     get_host_from_platform,
     get_install_target_from_platform,
     get_localhost_install_target,
@@ -1120,14 +1118,12 @@ class TaskJobManager:
         host_n, platform_name = None, None
         try:
             if rtconfig['remote']['host'] is not None:
-                host_n = self.task_remote_mgr.subshell_eval(
-                    rtconfig['remote']['host'], HOST_REC_COMMAND
+                host_n = self.task_remote_mgr.eval_host(
+                    rtconfig['remote']['host']
                 )
             else:
-                platform_name = self.task_remote_mgr.subshell_eval(
-                    rtconfig['platform'],
-                    PLATFORM_REC_COMMAND,
-                    host_check=False
+                platform_name = self.task_remote_mgr.eval_platform(
+                    rtconfig['platform']
                 )
         except PlatformError as exc:
             itask.waiting_on_job_prep = False
--- a/tests/functional/platforms/10-do-not-host-check-platforms.t
+++ b/tests/functional/platforms/10-do-not-host-check-platforms.t
@@ -15,7 +15,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #-------------------------------------------------------------------------------
-# Check that we don't check whether the platform name is a valid host.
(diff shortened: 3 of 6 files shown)


datamel (commented):
This hasn't fixed the problem for me unfortunately. Here is my test:
```
# global.cylc
[platforms]
    [[localhost]]
    [[localhost_test]]
        hosts = <a remote platform>
        install target = <remote platform install target>
```
The play a workflow e.g:
```
# flow.cylc
[scheduling]
    [[graph]]
        R1 = """
        hello => goodbye
    """

[runtime]
    [[root]]
        script = "sleep 1"
        platform = localhost_test
    [[hello]]
    [[goodbye]]
```

`hello` and `goodbye` both run locally, not respecting the settings in `global.cylc`.

As suggested, changing `localhost_test` to `localhxst_test` works fine.
I'll have a look to see if I can spot the problem.

wxtim:
> This hasn't fixed the problem for me unfortunately. Here is my test:
> 
> ```
> # global.cylc
> [platforms]
>     [[localhost]]
>     [[localhost_test]]
>         hosts = <a remote platform>
>         install target = <remote platform install target>
> ```
> 
> The play a workflow e.g:
> 
> ```
> # flow.cylc
> [scheduling]
>     [[graph]]
>         R1 = """
>         hello => goodbye
>     """
> 
> [runtime]
>     [[root]]
>         script = "sleep 1"
>         platform = localhost_test
>     [[hello]]
>     [[goodbye]]
> ```
> 
> `hello` and `goodbye` both run locally, not respecting the settings in `global.cylc`.
> 
> As suggested, changing `localhost_test` to `localhxst_test` works fine. I'll have a look to see if I can spot the problem.

I can't reproduce this.

oliver-sanders:
There are other uses of this interface with `host_check=True` so this error might be able to crop up in other places?

I'm not sure I understand the logic of `host_check` in `subshell_eval`, seems a little strange:

https://github.com/cylc/cylc-flow/blob/c817dd9c5b50cba36232d3bef9add8dbf4e46bac/cylc/flow/task_remote_mgr.py#L170-L176

Return localhost if the host is localhost???

I would have thought the `subshell_eval` should look something like this:

```python
match = command_pattern.match(command) 
if not match:
    # this isn't a command, no subshell evaluation to perform
    return command
```

Although note that later on we do:

```python
command = os.path.expandvars(command)
```

Which might need to be moved further up if we intend to support configuring the platform by an environment variable (e.g. `platform = $MY_PLATFORM`) which I don't think we do?

datamel (commented):
> I can't reproduce this.

Sorry @wxtim my remote testing was broken.  I have run this with a fresh environment and this is now working perfectly. 

I will hold off approval as I think Oliver has made some suggestions about host checking which I think sound sensible to address in this pr.

wxtim:
> Which might need to be moved further up if we intend to support configuring the platform by an environment variable (e.g. platform = $MY_PLATFORM) which I don't think we do?

No.

wxtim:
> I'm not sure I understand the logic of host_check in subshell_eval, 

This function can be called to find the value of a `platform` _or_ a `host`. If it's a host_check indicates we've called it from the compatibility mode part of the code, and that it should retain Cylc 7 behavior - that is to run host check on command, which should, confusingly, be an evaluated host name at this point. Since a platform isn't a host this check makes no sense if the command is getting a value for platform. 

I can't say for sure without looking more whether the Cylc 7 behavior is sane. But surely we'd have noticed if unreachable hosts caused jobs to run on local?

cylc/flow/task_job_mgr.py
```diff
@@ -1125,7 +1125,9 @@ def _prep_submit_task_job(
                 )
             else:
                 platform_name = self.task_remote_mgr.subshell_eval(
-                    rtconfig['platform'], PLATFORM_REC_COMMAND
+                    rtconfig['platform'],
+                    PLATFORM_REC_COMMAND,
+                    host_check=False
```
MetRonnie, line 1130: I'm concerned this will have unintended effects.

Does the bug not actually lie here? https://github.com/cylc/cylc-flow/blob/79e324df6519d1a007165e4878dd325dd7a1c4a5/cylc/flow/hostuserutil.py#L192-L200

MetRonnie: Ok makes sense. The `subshell_eval` method is somewhat confusing to figure out.

Also I'm still wondering about https://github.com/cylc/cylc-flow/blob/79e324df6519d1a007165e4878dd325dd7a1c4a5/cylc/flow/hostuserutil.py#L200

If "localhost.localdomain[.whatever]" is the only correct pattern then it should be
```diff
-name.split(".")[0].startswith("localhost")
+name.split(".")[0] == "localhost"
```

MetRonnie: cc @oliver-sanders

oliver-sanders: There can be many different variants of `localhost` e.g. `localhost.localdomain` etc, which is why the test is `startswith('localhost')` rather than `== 'localhost'`.

oliver-sanders: I'm confused by the `host_check` stuff in the first place though - https://github.com/cylc/cylc-flow/pull/5343#issuecomment-1433059203

MetRonnie: > There can be many different variants of `localhost` e.g. `localhost.localdomain` etc, which is why the test is `startswith('localhost')` rather than `== 'localhost'`.

But I'm asking if those variants will always follow the pattern of `locahost.something`. Because if not, then the `.split(".")[0]` is pointless

wxtim: I too am curious - is `localhost-foo.localdomain` a real possibility - and if it is, do we want it to match?

datamel: I think this is a possibility, looking at /etc/hosts file where a similar pattern is used seems like we should cater for it?

wxtim: I think we had some examples of cases where the logic _is_ necessary - can someone remind me what they are, and I'll add a comment to this to avoid this debate happening again?

MetRonnie: My suggestion to avoid future confusion
```diff
-if not name or name.split(".")[0].startswith("localhost"): 
-    # e.g. localhost.localdomain 
+if not name or name.startswith("localhost"):
+    # e.g. localhost4.localdomain4
```
and actually unit test this function!

wxtim:
> I'm not sure I understand the logic of host_check in subshell_eval, seems a little strange:

Only being run if subshell eval is supposed to be returning a host.

MetRonnie:
I have put up a draft PR to improve the readability of the subshell eval code, let me know thoughts: https://github.com/wxtim/cylc/pull/53

CHANGES.md
```diff
@@ -38,6 +38,12 @@ Rose options (`-O`, `-S` & `-D`) with `cylc view`.
 [#5363](https://github.com/cylc/cylc-flow/pull/5363) Improvements and bugfixes
 for `cylc lint`.
 
+-------------------------------------------------------------------------------
+## __cylc-8.1.2 (<span actions:bind='release-date'>Coming Soon</span>)__
+
+[#5343](https://github.com/cylc/cylc-flow/pull/5343) - Fix a bug causing
+platform names to be checked as if they were hosts.
+
```
MetRonnie, line 46: Wrong section

wxtim:
@MetRonnie  & @hjoliver  - Poke

hjoliver:
I think this is still awaiting merge of @MetRonnie 's side PR to your branch @wxtim ?

Tested as working however. (And I agree the affected code is a bit confusing).

wxtim:
> I think this is still awaiting merge of @MetRonnie 's side PR to your branch @wxtim ?
> 
> Tested as working however. (And I agree the affected code is a bit confusing).

Yes, but I'm assuming that @MetRonnie has marked it draft for a reason

tests/functional/platforms/10-do-not-host-check-platforms.t
```diff
@@ -0,0 +1,52 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Check that we don't check whether the platform name is a valid host.
```
MetRonnie, line 18:
```suggestion
# Check that platform names are not treated as host names. E.g. a platform
# name starting with "localhost" should not be treated as localhost.
# https://github.com/cylc/cylc-flow/issues/5342
```

MetRonnie (approved):

hjoliver (approved):
👍

PR #5346, opened 2023-02-06
data store: support unsatisfied ext_trigger

oliver-sanders:
Something I spotted whilst doing #5328.

This doesn't matter at the moment as the only use of this function calls it with `satisfied=True`.

**Check List**


dwsutherland (approved):
Crazy, it's a compulsory arg.. (should be in the docstring I suppose)

Is it even called if `False`? harmless either way I suppose
(Ah, the discription!)

oliver-sanders:
Yeah I think this was harmless as it's not presently called with `False`, all good.

PR #5349, opened 2023-02-07
Fix a bug preventing `cylc vip --workflow-name=foo` from working.

wxtim:
`--workflow-name foo` was fine.
Caused by removal of unwanted sys.argv for Cylc VIP not removing items not exactly matching items in the Command line options arguments.

This is a small fix for a bug without an issue.

**Check List**

- [x] ~[Cylc-Doc](https://github.com/cylc/cylc-doc)~ not req'd

### DIFF
--- a/cylc/flow/option_parsers.py
+++ b/cylc/flow/option_parsers.py
@@ -821,6 +821,7 @@ def cleanup_sysargv(
         for x in compound_script_opts
     }
     # Filter out non-cylc-play options.
+    args = [i.split('=')[0] for i in sys.argv]
     for unwanted_opt in (set(options.__dict__)) - set(script_opts_by_dest):
         for arg in compound_opts_by_dest[unwanted_opt].args:
             if arg in sys.argv:
@@ -831,8 +832,8 @@ def cleanup_sysargv(
                     not in ['store_true', 'store_false']
                 ):
                     sys.argv.pop(index)
-            elif arg in [i.split('=')[0] for i in sys.argv]:
-                index = [i.split('=')[0] for i in sys.argv].index(arg)
+            elif arg in args:
+                index = args.index(arg)
                 sys.argv.pop(index)
 
     # replace compound script name:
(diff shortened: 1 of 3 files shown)


datamel (commented):
I can't reproduce this bug, --worfklow-name is working fine for me with ` cylc vip`. Can you give me steps to follow?

wxtim:
> I can't reproduce this bug, --worfklow-name is working fine for me (with `=` and without) with ` cylc vip`. Can you give me steps to follow?

1. Ensure that you are using a Cylc remote host (i.e. Cylc server) - this bug is not visible if you use `--host=localhost` or if you don't have `run hosts` setup in your global config

```
cylc vip ~/metomi/workflows/simplest --workflow-name=dfgaiufhn
```

datamel (commented):
Thanks @wxtim. Bug now reproduced and various manual testing completed.  I think a rebase should pickup the mypy fix to let the CI run.

datamel (approved):
Thanks Tim, this is working well for me.

cylc/flow/option_parsers.py
```diff
@@ -831,6 +831,9 @@ def cleanup_sysargv(
                     not in ['store_true', 'store_false']
                 ):
                     sys.argv.pop(index)
+            elif arg in [i.split('=')[0] for i in sys.argv]:
+                index = [i.split('=')[0] for i in sys.argv].index(arg)
```
oliver-sanders, line 835: Could you compute this list outside of the nested for loop (i.e. at line 823) to avoid repeated recomputation.

oliver-sanders (commented):
LGTM

oliver-sanders:
Reproduced the bug and confirmed that this fixes it.

cylc/flow/option_parsers.py
wxtim: Yes.

oliver-sanders (approved):

MetRonnie:
Is it worth editing the changelog entry to make users aware of `-n` -> `-N` for `--no-detach`? This just caught me out.

PR #5350, opened 2023-02-07
fix mypy fail caused by https://github.com/python/mypy/pull/13969

wxtim:
Small fix to ensure mypy works.


datamel (approved):
Codecov upload error unrelated.

Issue #5341, opened 2023-02-01
cat-log: use --follow=name


oliver-sanders:
There's a potential issue where tail processes started by `cylc cat-log` can cause network file systems to keep files open for reading even after they have been deleted. These tempfiles can cause `cylc clean` to fail because the filesystem won't let it delete the file.

The `tail` command has a `--follow` option which can be set to `name` (as opposed to descriptor) which may solve this problem.

If it works this is a more elegant solution to [#417](https://github.com/cylc/cylc-uiserver/issues/417) which would only work at the UIS layer.

**Pull requests welcome!**

datamel:
I have tried this out and it does not fix the `cylc clean` error in the gui for me.

There are two scenarios that cause `cylc clean` to error, both occur when the log view is up and the subscription is active:
- the workflow is running and the log is still writing, this results in a `ServiceFileError`
- the log is closed (e.g. stopped workflow), this results in a `FileRemovalError`

Note, we can change the tail command in the  [global config](https://github.com/cylc/cylc-flow/blob/9e042645c92b871a07e07c3a90a73bb2ff5c99e8/cylc/flow/cfgspec/globalcfg.py#L1428) which may have been helpful as an immediate fix.

I think the way forward with this problem is to close the subscription on clean. This could be done from the ui or ui-server. I'll have a look into which one works best.

datamel:
As an aside, I think it sensible to have the clean button in the ui disabled if the workflow is still running.

oliver-sanders:
> As an aside, I think it sensible to have the clean button in the ui disabled if the workflow is still running.

https://github.com/cylc/cylc-flow/issues/4726

oliver-sanders:
Closed by #5359

Issue #5324, opened 2023-01-23
Move error information out of printing/logging and into exception messages


MetRonnie:
**Describe exactly what you would like to see in an upcoming release**

Example: 
```console
$ cylc clean running_workflow
WARNING - Cannot clean running workflow Mist.
    workflow contact file exists: ...
    blah blah...
    
CylcError: Clean failed: Mist
```
The problem with this is it's very CLI-centric and so if you try to clean a running workflow from the GUI you only get an alert saying "Clean failed: Mist" with no further context

_Optimistically tagged against 8.1.1 but can be bumped_

MetRonnie:
Here is an example UIS log snippet - need to figure out how to condense the relevant error messages down and include the final exception:
```
[I 2025-04-28 11:13:59.024 CylcUIServer] Cleaning ~ronnie.dutta/Simple
[I 2025-04-28 11:13:59.056 cylc] Cleaning Simple on install target: borked
[E 2025-04-28 11:13:59.334 cylc] Could not clean Simple on install target: borked
    platform: borked - clean up did not complete
    COMMAND:
        ssh -oBatchMode=yes -oConnectTimeout=10 bogus env \ 
    
            CYLC_VERSION=8.4.2 CYLC_ENV_NAME=cylc-8.4.2-1 bash --login \ 
    
            -c 'exec "$0" "$@"' timeout 600 cylc clean --local-only \ 
    
            --no-scan Simple --rm .service:log
    RETURN CODE:
        255
    STDOUT:
    STDERR:
        ssh: Could not resolve hostname bogus: Name or service not known
    
[E 2025-04-28 11:13:59.335 cylc] Failed to clean Simple
    Error: Remote clean failed for Simple
[E 2025-04-28 11:13:59.342 CylcUIServer] CylcError: Clean failed
    concurrent.futures.process._RemoteTraceback: 
    """
    Traceback (most recent call last):
      File ".../python3.9/concurrent/futures/process.py", line 246, in _process_worker
        r = call_item.fn(*call_item.args, **call_item.kwargs)
      File ".../python3.9/site-packages/cylc/uiserver/resolvers.py", line 202, in _clean
        return loop.run_until_complete(
      File ".../python3.9/asyncio/base_events.py", line 647, in run_until_complete
        return future.result()
      File ".../python3.9/site-packages/cylc/flow/scripts/clean.py", line 240, in run
        raise CylcError("Clean failed")
[...]

PR #5359, opened 2023-02-09
Fix cylc clean while cat-log running

datamel:
`cylc clean` will fail when a cat-log process is keeping one of the log files open. This is a more pressing issue now that the log view is available in the gui.
This is related to an nfs issue and may not be possible to reproduce on all systems.

The solution involves 
- changing the cat-log command to include `--follow=name`, changed in the default.  
- changing `rmtree` behaviour, by retrying on error.

closes https://github.com/cylc/cylc-flow/issues/5341 

Note for debugging this issue, I changed the error message to be passed into the gui.

To test, open the gui and open a log file. With the log view still open, clean the workflow.
This will error on master but not on this branch.

In the gui, I think we should maybe close any open views on clean, perhaps even redirect to the dashboard? But this would be a separate issue.

**Check List**


oliver-sanders:
> If this is a bug fix, PR should be raised against the relevant ?.?.x branch.

Needs rebasing onto 8.1.x.

MetRonnie:
> Note for debugging this issue, I changed the error message to be passed into the gui, so this (if reviewers are happy with the approach) will also close #5324.

Thanks, however this isn't sufficient to close that issue. We need to check for other instances of logging that should be moved into the exception message

datamel:
> > Note for debugging this issue, I changed the error message to be passed into the gui, so this (if reviewers are happy with the approach) will also close #5324.
> 
> Thanks, however this partially addresses that issue. We need to check for other instances of logging that should be moved into the exception message

Okay, I've removed the close from the description.

cylc/flow/pathutil.py
```diff
@@ -297,15 +298,27 @@ def remove_dir_and_target(path: Union[Path, str]) -> None:
...
-        rmtree(path, onerror=handle_rmtree_err)
+        _rmtree(path)
+
+
+def _rmtree(target, retries=5, sleep_time=0.5):
+    for _try_num in range(retries):
+        try:
+            rmtree(target)
+            return
+        except OSError as exc:
```
oliver-sanders, line 317: To avoid retrying the `shutil.rmdirs` when it is not appropriate to do so we should filter for "directory not empty" errors. The `errno` module is the tool for the job (error numbers are OS dependent):

```python
        except OSError as exc:
            if exc.errno == errno.ENOTEMPTY:
                sleep(sleep_time)
            else:
                raise
```

https://docs.python.org/3/library/errno.html?highlight=errno#errno.ENOTEMPTY

MetRonnie: That's how `handle_rmtree_err` works

MetRonnie: Why the decision to retry 5 times? (There is no docstring for this function)

cylc/flow/scripts/clean.py
```diff
@@ -193,15 +193,18 @@ async def run(*ids: str, opts: 'Values') -> None:
...
 
-    failed = []
+    failed = {}
     for workflow in sorted(workflows):
         try:
             init_clean(workflow, opts)
         except Exception as exc:
-            failed.append(workflow)
+            failed[workflow] = exc
             LOG.warning(exc)
```
MetRonnie, line 202: I don't think we'll need the warning as its a duplicate of the error below
```suggestion
```

MetRonnie (approved):

MetRonnie:
Ah, just realised a slight flaw with only testing for `ENOTEMPTY` - the second time it tries it, the error will be `EBUSY` instead, so it will only retry once before failing. (Assuming it's a "sillyrenamed" `.nfsXXXX...` file causing the problem)

image

Also annoying is how `rmtree` only gives the filename, not path. If it's a job log file that's being kept open by a running job, there could be hundreds of job log dirs and you don't know which one contains the `.nfsXXXX...`.

Anyway, we can sort these out in a follow-up issue.

oliver-sanders:
> Ah, just realised a slight flaw with only testing for ENOTEMPTY - the second time it tries it, the error will be EBUSY instead

Dammit, confirmed with this script:

```bash
#!/usr/bin/env bash                                                      
                                                                         
set -eux                                  
                                          
TD="$HOME/cylc-run/foo/log/scheduler"    
mkdir -p "$TD"         
TF="$TD/log"           
                       
echo -e 'a\nb\nc' > "$TF"    
tree "$TD"             
                       
tail --follow=name -f "$TF" & PROC="$!"    
     
sleep 2    
     
# this will likely fail due to FS locking stuffs    
rm -r "$TD" && echo 'REMOVE PASSED' || echo 'REMOVE FAILED'    
rm -r "$TD" && echo 'REMOVE PASSED' || echo 'REMOVE FAILED'    
rm -r "$TD" && echo 'REMOVE PASSED' || echo 'REMOVE FAILED'                                                                        
                                             
# tidy up after                              
kill "$PROC"                                 
rm -r "$TD" || true     
```

Which reliably gives:

* Directory not empty
* Device or resource busy
* Device or resource busy

cylc/flow/pathutil.py
```diff
@@ -297,15 +299,35 @@ def remove_dir_and_target(path: Union[Path, str]) -> None:
...
+    """Make rmtree more robust to nfs issues.
+
+    shutil.rmtree will be retried (default 5 times)
+    """
+    for _try_num in range(retries):
+        try:
+            rmtree(target)
+            return
+        except OSError as exc:
+            if exc.errno == errno.ENOTEMPTY:
```
oliver-sanders, line 325:
```suggestion
            if exc.errno in {errno.ENOTEMPTY, errno.EBUSY}:
```

oliver-sanders:
I used this script to double-check the premise that the tail process dies when the file is removed and to confirm the error message(s) returned:

```bash
#!/usr/bin/env bash                                                      
                                                                         
set -eu                                                            
                                                 
TD="$HOME/cylc-run/foo/log/scheduler"            
mkdir -p "$TD"                                   
TF="$TD/log"                                     
                                                 
echo -e 'a\nb\nc' > "$TF"                        
                                                 
# NOTE: for some reason the tail command doesn't die when launched
# as a background process                        
echo 'Start you tail process now:'               
echo "i.e. tail -f --follow=name '$TF'"          
sleep 5                                          
                                                 
# retry removal a few times                      
SLEEP=1                                          
for try in $(seq 1 10); do                       
    sleep "$SLEEP"                               
    echo "Try: $try"                             
    rm -r "$TD" || continue                      
    echo "Success after $((try * SLEEP)) seconds."
    break            
done                 
                     
# tidy up after JIC
rm -r "$TD" 2>/dev/null || true 
```

This is hard to test, would need NFS in a container and whatnot so commenting it here for the record, will link this comment into the docstring.

cylc/flow/pathutil.py
```diff
@@ -297,15 +299,35 @@ def remove_dir_and_target(path: Union[Path, str]) -> None:
...
-        rmtree(path, onerror=handle_rmtree_err)
+        _rmtree(path)
+
+
+def _rmtree(target: Union[Path, str],
+            retries: int = 5,
+            sleep_time: float = 0.5):
+    """Make rmtree more robust to nfs issues.
+
+    shutil.rmtree will be retried (default 5 times)
```
oliver-sanders, line 318:
```suggestion
    """Make rmtree more robust to nfs issues.
    
    If a file is deleted which is being held open for reading by another process
    NFS will create a ".nfs" file in the containing directory to handle this.

    If you try to delete the directory which contains these files you will get either
    a ENOTEMPTY or EBUSY error.
    
    A likely cause of open file handles in cylc-run directories is `cylc cat-log -m t`.
    If the file being cat-log'ged is removed, the command will fail on its next poll.
    The default poll interval is one second, so if we wait a couple of seconds and
    retry the removal it will likely work.

    This command retries removal a specified number of times at a specified
    interval before failing to give cat-log process a chance to die gracefully and
    release their filesystem locks. For more info see:
    https://github.com/cylc/cylc-flow/pull/5359#issuecomment-1479989975
```

cylc/flow/pathutil.py
```diff
@@ -297,15 +299,49 @@ def remove_dir_and_target(path: Union[Path, str]) -> None:
...
+    A likely cause of open file handles in cylc-run directories is `cylc cat-log -m t`.
+    If the file being cat-log'ged is removed, the command will fail on its next poll.
+    The default poll interval is one second, so if we wait a couple of seconds and
+    retry the removal it will likely work.
+
+    This command retries removal a specified number of times at a specified
+    interval before failing to give cat-log process a chance to die gracefully and
+    release their filesystem locks. For more info see:
+    https://github.com/cylc/cylc-flow/pull/5359#issuecomment-1479989975
+    """
```
oliver-sanders, line 333:
```suggestion
def _rmtree(
    target: Union[Path, str],
    retries: int = 5,
    sleep_time: float = 0.5,
):
    """Make rmtree more robust to nfs issues.

    If a file is deleted which is being held open for reading by
    another process. NFS will create a ".nfs" file in the
    containing directory to handle this.

    If you try to delete the directory which contains these
    files you will get either a ENOTEMPTY or EBUSY error.

    A likely cause of open file handles in cylc-run directories
    is `cylc cat-log -m t`. If the file being cat-log'ged is removed,
    the command will fail on its next poll. The default poll
    interval is one second, so if we wait a couple of seconds and
    retry the removal it will likely work.

    This command retries removal a specified number
    of times at a specified interval before failing to
    give cat-log process a chance to die gracefully and
    release their filesystem locks. For more info see:
    https://github.com/cylc/cylc-flow/pull/5359#issuecomment-1479989975
    """
```

oliver-sanders:
Found a `rmdir` call which might need to be converted to `_rmtree`?

```diff
diff --git a/cylc/flow/pathutil.py b/cylc/flow/pathutil.py
index 3acde3700..173a82753 100644
--- a/cylc/flow/pathutil.py
+++ b/cylc/flow/pathutil.py
@@ -397,8 +397,8 @@ def remove_empty_parents(
         if not parent.is_dir():
             continue
         try:
-            parent.rmdir()
             LOG.info(f'Removing directory: {parent}')
+            _rmtree(path)
         except OSError:
             break
```

MetRonnie (approved):
I've tested it out. Added a changelog entry

PR #5360, opened 2023-02-10
clock-expire: experimental main-loop implementation

oliver-sanders:
PR for the purpose of debate...

* The current clock-expire implementation uses task pool logic.
* This means that tasks can only become expired when they enter the pool.
* Consequently expire events are retroactive and might not occur at all. This means they are not useful for triggering in the way they were at Cylc 7.
* This alternative implementation maintains a moving window for detecting clock expired tasks which is separate to the task_pool so is not dependent on tasks being in the task_pool for expired events to be generated.
* It works by using a "periodic" main loop plugin to extend the expiry window, adding any clock-expire tasks within that window into a list of timeouts.
* It then checks that list of timeouts via the same "periodic" main loop plugin to see if any tasks have expired.
* To avoid expiring tasks which have submitted successfully it uses a [new] "submit" main loop plugin to remove timeouts for tasks which have submitted.

Advantages:
* Expired events are raised close to the time of expiry (according to the interval the main loop plugin is configured to run on).
* Expired outputs are generated whether the workflkow has made it as far as the task or not, allowing for effective triggering e.g: `foo:expire? => special_action`.
* The same methodology could also be applied to late tasks.

Challenges:
* Removing this logic from the task_pool makes interfacing with multiple flows tricky.
* What flow(s) do we expire the tasks in?
* Graph branching!!!

TODO:
* Activate the clock_expire plugin if there are clock-expired tasks in the workflow.
* Make the logic more efficient.
* Fix old & write new tests, tidy, etc.

**Check List**

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -270,7 +270,7 @@ class WorkflowConfig:
         self.start_point: 'PointBase'
         self.stop_point: Optional['PointBase'] = None
         self.final_point: Optional['PointBase'] = None
-        self.sequences: List['SequenceBase'] = []
+        self.sequences: Dict['SequenceBase', List[str]] = {}
         self.actual_first_point: Optional['PointBase'] = None
         self._start_point_for_actual_first_point: Optional['PointBase'] = None
 
@@ -1661,6 +1661,7 @@ class WorkflowConfig:
             elif not suicide:
                 # "foo => !bar" does not define a sequence for bar
                 taskdef.add_sequence(seq)
+                self.sequences[seq].append(name)
 
     def generate_triggers(self, lexpression, left_nodes, right, seq,
                           suicide, task_triggers):
@@ -2095,7 +2096,7 @@ class WorkflowConfig:
                 if isinstance(exc, CylcError):
                     msg += ' %s' % exc.args[0]
                 raise WorkflowConfigError(msg)
-            self.sequences.append(seq)
+            self.sequences.setdefault(seq, [])
             parser = GraphParser(
                 family_map,
                 self.parameters,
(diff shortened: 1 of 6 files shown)


hjoliver:
From discussion on Element chat a while back:

### Use cases for clock-expire before submit time?

1. in catch-up mode, a task that normally fetches real-time data might instead get already-fetched data from archive

However, this sort of task is likely to be solely clock-triggered, so it will be spawned out to the runahead limit, and expire checking will happen ahead of time.

### Cylc 7 clock-expire?

Cylc 7 clock-expire was also task-pool based. Earlier spawning of waiting tasks into the pool meant that expiry could often be detected ahead of time. However, the time of detection depended entirely on spawning logic and was unrelated to the actual moment of expiry.

### Look ahead of the pool to expire tasks early (e.g. this PR)?

Graph branching is a problem, because we can't know in advance which branch will be taken.

- we shouldn't expire a task that is on a branch not taken
   - expiry doesn't matter if the task wasn't going to run anyway
- the multi-flow problem seems gnarly
  - we will have to look ahead for expiry again, every time a new flow is started
  - and each time, we have the above branching problem

### Conclusion: stick with task-pool based expiry??

- it's uncomplicated
- it achieves the primary purpose of not running the expired task if too far behind the clock
- flows aren't a problem, - spawned tasks already belong to a flow
- branching isn't a problem  - tasks (in Cylc 8) are only spawned on "taken" branches

The only problem is, expiry on non-parentless tasks won't be detected until they are spawned into the task pool.

Workaround for that: add an artificial dependency to cause spawning at a specific earlier point in the workflow, after which expiry can be detected.

PR #5368, opened 2023-02-16
graphql: stabilise ID iteration order

### LINKED ISSUE #5366: linear ancestors not so linear
A nasty bug in the ancestor ordering which causes whole bunches of tasks to disappear from the UI!

For the following workflow:

```ini
[scheduling]             
    [[graph]]            
        R1 = A & B       
                         
[runtime]                
    [[A]]                
    [[A1]]               
        inherit = A      
    [[A2]]    
        inherit = None, A    
    [[a11]]    
        inherit = A1    
    [[a22]]    
        inherit = A2    
    [[B]]    
    [[B1]]    
        inherit = B    
    [[b]]    
        inherit = B1  
```

And this GraphQL request (run through GraphiQL):

```graphql
query {
  workflows {
    familyProxies {
      id
      ancestors {
        id
      }
    }
  }
}
```

I get this response:

```json
{
  "data": {
    "workflows": [
      {
        "id": "~/fam-test/run1",
        "familyProxies": [
          {
            "id": "~/fam-test/run1//1/B",
            "ancestors": [
              {
                "id": "~/fam-test/run1//1/root"
              }
            ]
          },
          {
            "id": "~/fam-test/run1//1/A",
            "ancestors": [
              {
                "id": "~/fam-test/run1//1/root"
              }
            ]
          },
          {
            "id": "~/fam-test/run1//1/B1",
            "ancestors": [
              {
                "id": "~/fam-test/run1//1/root"
              },
              {
                "id": "~/fam-test/run1//1/B"
              }
            ]
          },
          {
            "id": "~/fam-test/run1//1/A1",
            "ancestors": [
              {
                "id": "~/fam-test/run1//1/A"
              },
              {
                "id": "~/fam-test/run1//1/root"
              }
            ]
          },
          {
            "id": "~/fam-test/run1//1/A2",
            "ancestors": [
              {
                "id": "~/fam-test/run1//1/root"
              }
            ]
          },
          {
            "id": "~/fam-test/run1//1/root",
            "ancestors": []
          }
        ]
      }
    ]
  }
}
```

What's interesting here is the difference between the ancestor order of A1 and B1:

```json
          {
            "id": "~/fam-test/run1//1/A1",
            "ancestors": [
              {
                "id": "~/fam-test/run1//1/A"
              },
              {
                "id": "~/fam-test/run1//1/root"
              }
            ]
          },
```

```json
          {
            "id": "~/fam-test/run1//1/B1",
            "ancestors": [
              {
                "id": "~/fam-test/run1//1/root"
              },
              {
                "id": "~/fam-test/run1//1/B"
              }
            ]
           }
```

Despite having identical inheritance structures, the inheritance order is reversed with the `root` family at the bottom of one list and the top of the other.

Strangely it seems to be the family `A2` which introduces this behaviour with the `inherit = None` bit.

Because the UI relies on this field to compute the family tree things get confused and the tree ends up with three "root" notes, `A`, `B` & `root` which should be impossible (because there should only be one "root" node, and it should be called `root`!).

**Pull requests welcome!**

oliver-sanders:
> Sibling: https://github.com/cylc/cylc-ui/pull/1230

* A couple of ID lists were being converted to sets which caused the return order to become jumbled.
* Switch to a function which removes duplicates whilst maintaining order
* Closes #5366

**Check List**

- [x] `CHANGES.md` Changelog entry added in cylc-ui.

### DIFF
--- a/cylc/flow/network/resolvers.py
+++ b/cylc/flow/network/resolvers.py
@@ -385,7 +385,7 @@ class BaseResolvers(metaclass=ABCMeta):  # noqa: SIM119
 
     async def get_nodes_by_ids(self, node_type, args):
         """Return protobuf node objects for given id."""
-        nat_ids = set(args.get('native_ids', []))
+        nat_ids = uniq(args.get('native_ids', []))
         # Both cases just as common so 'if' not 'try'
         if 'sub_id' in args and args['delta_store']:
             flow_data = [
@@ -447,7 +447,7 @@ class BaseResolvers(metaclass=ABCMeta):  # noqa: SIM119
 
     async def get_edges_by_ids(self, args):
         """Return protobuf edge objects for given id."""
-        nat_ids = set(args.get('native_ids', []))
+        nat_ids = uniq(args.get('native_ids', []))
         if 'sub_id' in args and args['delta_store']:
             flow_data = [
                 delta[args['delta_type']]
@@ -470,6 +470,8 @@ class BaseResolvers(metaclass=ABCMeta):  # noqa: SIM119
         nodes = root_nodes
         node_ids = {n.id for n in root_nodes}
         edges = []
+        # NOTE: we don't expect edges to be returned in definition order
+        # so it is ok to use `set` here
         edge_ids = set()
         # Setup for edgewise search.
         new_nodes = root_nodes
@@ -492,6 +494,8 @@ class BaseResolvers(metaclass=ABCMeta):  # noqa: SIM119
             edges += new_edges
             # Gather nodes.
             # One of source or target will be in current set of nodes.
+            # NOTE: we don't expect node-edges to be returned in definition
+            # order so it is ok to use `set` here
             new_node_ids = set(
                 [e.source for e in new_edges]
                 + [e.target for e in new_edges]).difference(node_ids)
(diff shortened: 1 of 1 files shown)


wxtim (commented):
It looks sane enough, but could you not use dictionary keys (ordered from 3.7) instead of a set?

Profiling makes `uniq` written this way look a bit faster, although it's only a comparison of 2 linears

image
image


I tried following the instructions in #5336 but couldn't re-create the original bug.

oliver-sanders:
> but could you not use dictionary keys

Like this?

```python
ids = {key: None for key in ids}.keys()
```

oliver-sanders:
Yes, that is remarkably efficient, I'll need to check through the code as `keys()` objects are a bit different to the standard containers, but I expect it will be fine.

oliver-sanders:
> Went through the uses which are all iteration so dict.keys should be just fine, nice.

Sorry, the tests flagged that the items in the list aren't necessarily strings, or even `__hash__`able so the `dict.keys` approach don't work for all :(

I don't think the iteration time spent here is too big an issue compared to the graphql overheads, nice suggestion though.

wxtim:
> 

Sorry, should have spotted that caveat .
I think I read something like at Python 2 `set([1, 2, 3])` is implementation-wise almost exactly the same as `{i: None for i in [1, 2, 3]}`, but they've diverged at Python.

dwsutherland (approved):
LGTM 👍

wxtim (approved):
- [x] Read Code
- [x] Accept that tests are in the UI in effect
- [x] Manual testing

PR #5369, opened 2023-02-16
workflow_files: split overgrown module

### LINKED ISSUE #4414: Refactor workflow_files
`cylc.flow.workflow_files` has ballooned in size; some parts could do with splitting off into their own modules, e.g. clean, install, etc.

Also, some parts of `cylc.flow.pathutil` are perhaps better suited to `workflow_files`.

**Pull requests welcome!**

oliver-sanders:
Closes #4414

(the remaining aspects of #4414 to be solved in a future refactor of cylc.flow.pathutil)

The `workflow_files` module has ballooned to accommodate install and clean functionality.

Pure refactor with no functionality changes.

Refactor the code to split the `clean` and `install` parts into their own modules. This code is now only loaded by the `clean`, `install` and `reinstall` commands respectively reducing the import overheads on other functionality (including `cylc message` which currently imports it via `cylc.flow.network.client`.

Tidier and probably marginally more efficient. Cylc's codebase and dependencies take up a fair bit of memory and contribute to startup overheads. Further tweaking of this nature aught to reduce the impact of this by straightening out the import tree.

**Check List**


wxtim (approved):
Nice - Additional benefit - makes the code easier to understand if you don't know the history of how it got like this.

cylc/flow/network/__init__.py
```diff
@@ -22,6 +22,7 @@
 
 import zmq
 import zmq.asyncio
+import zmq.auth
```
datamel, line 25: Just wondering why this import was added? zmq is already imported, is it more efficient to pull both in? I'm not fully sure of the intricacies of imports.

oliver-sanders: zmq.auth is a module so not ordinarally available by `import zmq`.

```python
>>> import zmq
>>> zmq.auth
AttributeError: module 'zmq' has no attribute 'auth'
>>> import zmq.auth
>>> zmq.auth
<module 'zmq.auth' from '/.../site-packages/zmq/auth/__init__.py'>

```

I think this worked before because zmq.auth was imported in another file so the namespace was already loaded into the `zmq` namespace when we imported `zmq` in this module? Not sure. Python's import system is pretty nasty.

datamel: Ahhh ok, thanks, that's a bit strange - I imagine I've been importing incorrectly all this time 😆

oliver-sanders: NP, easy to do, I didn't spot it before, neither did flake8 which is disappointing.

datamel (commented):
Much better structured code. I have checked this out, run tests and workflows/install/clean. Working smoothly, thanks @oliver-sanders .

datamel (approved):

datamel:
I will leave the merge to you @oliver-sanders as I'm not sure if you would rather this went into 8.1.x or master? I expect as just infrastructure master would be fine but I'm not 100% sure.

oliver-sanders:
Safer to keep refactors away from maintenance releases to give us longer to spot any issues.

Issue #5383, opened 2023-02-28
CLI: VIP and VR should accurately report options for components


wxtim:
### Problem

The current VIP and VR scripts tell users what atomic scripts would be being called. That is, 

```bash
cylc vip /loc/myworkflow --workflow-name foo -s='BAZ=42'
...
$ cylc validate /loc/myworkflow
...
$ cylc install /loc/myworkflow
...
$ cylc play foo -s='BAZ=42'
```

These reports should also include a full list of options relevant to that command, as is effectively done for Cylc Play in https://github.com/cylc/cylc-flow/pull/5377:

That is, when called the output should be
```
...
$ cylc validate /loc/myworkflow -s='BAZ=42'
...
$ cylc install /loc/myworkflow --workflow-name foo
...
$ cylc play foo -s='BAZ=42'
```

oliver-sanders:
I did something similar for argparse, easy conversion.

oliver-sanders:
Don't have the time to push this one through ATM.

Closing this PR: https://github.com/cylc/cylc-flow/pull/5385

PR #5381, opened 2023-02-27
tui: add vr

oliver-sanders:
Response to user feedback:

* Add a `--yes` option to `cylc reinstall` / `cylc vr`.
* Add `cylc vr` into `cylc tui`.

This also presents a box which tells the user that a command is running to avoid confusion with longer-running commands e.g. `play`, `clean`, `validate-reinstall`.

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [ ] There is no interactive test harness for `cylc tui` at present.
- [ ] `CHANGES.md` entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #5385, opened 2023-02-28
Combined command logging

oliver-sanders:
**Closes:** #5383 
**Built On:**: https://github.com/cylc/cylc-flow/pull/5381
**Sibling PR:** https://github.com/cylc/cylc-rose/pull/214 (for cylc-rose functionality)

To help clarify what the combined commands are doing, this PR logs CLI strings which would reproduce the actions performed if run manually.

Screenshot from 2023-02-28 17-31-48

Needs close checking, the last thing we want is to introduce a new error breaking these scripts in niche cases. Could also do with more testing.

**Check List**

### DIFF
--- a/cylc/flow/option_parsers.py
+++ b/cylc/flow/option_parsers.py
@@ -42,6 +42,8 @@ from cylc.flow.loggingutil import (
     CylcLogFormatter,
     setup_segregated_log_streams,
 )
+from cylc.flow.util import format_parsed_opts
+
 
 WORKFLOW_ID_ARG_DOC = ('WORKFLOW', 'Workflow ID')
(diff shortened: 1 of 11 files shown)


cylc/flow/option_parsers.py
```diff
@@ -530,7 +532,7 @@ def get_cylc_rose_options():
                 help=(
                     "Use optional Rose Config Setting"
                     " (If Cylc-Rose is installed)"),
-                action="append", default=[], dest="opt_conf_keys",
+                action="append", default=None, dest="opt_conf_keys",
```
oliver-sanders, line 535: These `default=[]` statements appear to have been passing a mutable object through a mechanism which mutated them?!

To prove this, try printing `option.default` after option parsing, you'll find it matches the value?!

Dunno, simple fix, don't pass in something mutable.

PR #5391, opened 2023-03-02
Fix possible IndexError during shutdown

### LINKED ISSUE #5390: scheduler: traceback on shutdown
```
      File "/cylc/flow/scheduler.py", line 1751, in shutdown
        await self._shutdown(reason)
      File "/cylc/flow/scheduler.py", line 1836, in _shutdown    
        self.run_event_handlers(self.EVENT_SHUTDOWN, reason.args[0])
    IndexError: tuple index out of range
```

The reason being that `reason.args` has length `0`. This logic should be made more robust.

MetRonnie:
Closes #5390

**Check List**

- [x] No dependency changes
- [x] No tests needed - cba
- [x] No changelog entry needed (very minor)
- [x] No docs needed

### DIFF
--- a/cylc/flow/exceptions.py
+++ b/cylc/flow/exceptions.py
@@ -19,6 +19,7 @@ import errno
 from textwrap import wrap
 from typing import (
     Callable,
+    Dict,
     Iterable,
     NoReturn,
     Optional,
@@ -366,11 +367,10 @@ class PlatformLookupError(CylcConfigError):
 class HostSelectException(CylcError):
     """No hosts could be selected from the provided configuration."""
 
-    def __init__(self, data):
+    def __init__(self, data: Dict[str, dict]):
         self.data = data
-        CylcError.__init__(self)
 
-    def __str__(self):
+    def __str__(self) -> str:
         ret = 'Could not select host from:'
         for host, data in sorted(self.data.items()):
             if host != 'ranking':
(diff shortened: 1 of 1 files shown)


cylc/flow/exceptions.py
```diff
@@ -430,7 +430,6 @@ class NoHostsError(CylcError):
     """None of the hosts of a given platform were reachable."""
     def __init__(self, platform):
         self.platform_name = platform['name']
-        super().__init__()
```
oliver-sanders, line 433: Ok, funky.

It looks like `HostSelectException` might be similar?

oliver-sanders (approved):

cylc/flow/exceptions.py
MetRonnie: Good spot, I have fixed that too now

oliver-sanders (approved):

hjoliver:
Maybe I'm being dense but how do the code changes on this branch fix the IndexError during shutdown?

MetRonnie:
The traceback:
```
    self.run_event_handlers(self.EVENT_SHUTDOWN, reason.args[0])
IndexError: tuple index out of range
```
The code:
```py
class SomethingError(CylcError):
    def __init__(self, platform):
        self.platform_name = platform['name']
        super().__init__()
```
Calling `super().__init__()` means the exception has an empty list for the `args` property

hjoliver (approved):

PR #5393, opened 2023-03-02
remote: add note about sys.exit behaviour

oliver-sanders:
Add a note about `sys.exit` behaviour in the remote interfaces.

### DIFF
--- a/cylc/flow/remote.py
+++ b/cylc/flow/remote.py
@@ -78,7 +78,7 @@ def run_cmd(
     manage=False,
     text=True,
 ):
-    """Run a given cylc command on another account and/or host.
+    """Run a given cylc command on another host.
 
     Arguments:
         command (list):
@@ -107,6 +107,10 @@ def run_cmd(
         * Else True if the remote command is executed successfully, or
           if unsuccessful and capture_status=True the remote command exit code.
         * Otherwise exit with an error message.
+
+    Exits:
+        1: In the event of certain command errors.
+
     """
     # CODACY ISSUE:
     #   subprocess call - check for execution of untrusted input.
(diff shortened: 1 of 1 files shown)


cylc/flow/remote.py
```diff
@@ -107,6 +107,10 @@ def run_cmd(
         * Else True if the remote command is executed successfully, or
           if unsuccessful and capture_status=True the remote command exit code.
         * Otherwise exit with an error message.
+
+    Exits:
+        1: In the event of certain command errors.
```
MetRonnie, line 112: I think `Exits with code 1 in the event of certain command errors.` might be a little clearer

MetRonnie (approved):

hjoliver (approved):

cylc/flow/remote.py
hjoliver: And/or "Exit codes" rather than "Exits".

oliver-sanders: All yours!

MetRonnie:
```suggestion
    Exits with code 1 in the event of certain command errors.
```

cylc/flow/remote.py
```diff
@@ -385,6 +388,9 @@ def remote_cylc_cmd(
     Raises:
         NoHostsError: If the platform is not contactable.
 
+    Exits:
+        1: In the event of certain command errors.
```
MetRonnie, line 392:
```suggestion
    Exits with code 1 in the event of certain command errors.
```

cylc/flow/remote.py
```diff
@@ -426,6 +432,9 @@ def cylc_server_cmd(cmd, host=None, **kwargs):
...
```
MetRonnie, line 436:
```suggestion
    Exits with code 1 in the event of certain command errors.
```

PR #5395, opened 2023-03-03
Fix workflow shutdown on no platforms error

### LINKED ISSUE #5392: NoPlatformsError
Seen in the wild with Cylc 8.1.2:

```
ERROR - Workflow shutting down - NoPlatformsError: Unable to find a platform from group <group>
```

One of these `NoPlatformsError`s managed to slip the net. Annoyingly it wasn't raised as traceback so the only evidence is this one line :angry:!

Just before this error the user had attempted to trigger a task. The task entered the preparing state `=> preparing`, falied to contact a couple of login nodes, then the above message was logged.

wxtim:
closes #5392 

**Check List**


MetRonnie:
> `NoPlatformsError` comes from `get_platform` and will only occur if get platform is given an argument (i.e. it **isn't** just getting the localhost platform).

I'm not yet convinced of this.

But while investigating I've noticed a slight flaw where this variable is basically unused:

https://github.com/cylc/cylc-flow/blob/dcd1290bdefa1a7e00a4c2998b1c22d54b0983c7/cylc/flow/platforms.py#L193

The only reference to it:

https://github.com/cylc/cylc-flow/blob/dcd1290bdefa1a7e00a4c2998b1c22d54b0983c7/cylc/flow/platforms.py#L245-L246

I'm sure this is unrelated to the bug, but worth sorting out while we're at it

wxtim:
> I'm not yet convinced of this.

My reasoning for this in detail:

- `raise NoPlatformsError` occurs once at `platforms.py:290` in function `get_platform_from_group`.
- `get_platform_from_group` is called once, in `platform_from_name`.
- `platform_from_name` occurs 
    - 4 times in `get_platform`
    - Once in `get_platform_from_group`, which is then used by `get_platform` - this sounds circular, except that in `get_platform_from_group` we ask for a `platform_from_name` for a platform within a group, so we know that `get_platform` shouldn't send us back to `get_platform_from_group`. I should probably check what happens if I have recursive platform groups though. 
    - Once in `get_install_target_to_platforms_map`. This is being fed a list of platform names by `get_install_target_to_platforms_map`, which gets them from `workflow_files.remote_clean`, which is called by `init_clean`. That call gets platform names for tasks from the database, which I think therefore have to be platforms and not groups. I still need to check that supposition though.

wxtim:
> But while investigating I've noticed a slight flaw where this variable is basically unused:

Yes, I've had a check, and I think you're right. I'll remove it.

wxtim:
ME
> What's the difference between `NoPlatformsError` and `PlatformLookupError` 

Answer: No Platforms Error is when the platform group definition exists, but the hosts aren't reachable for any platform in group. PlatformLookupError indicates a problem with the `global.cylc` or `flow.cylc` - i.e. user has asked for an undefined platform.

cylc/flow/platforms.py
```diff
@@ -190,14 +198,14 @@ def platform_from_name(
...
 
-    platform_group = None
     for platform_name_re in reversed(list(platform_groups)):
         # Platform is member of a group.
         if re.fullmatch(platform_name_re, platform_name):
             platform_name = get_platform_from_group(
                 platform_groups[platform_name_re], group_name=platform_name,
                 bad_hosts=bad_hosts
             )
+            break
```
oliver-sanders, line 208: I think this will change the order in which platforms match.

Please check whether this matches the order specified in the platforms proposal. If this doesn't break tests we could do with adding one. Might be easier to push to another PR if unrelated.

oliver-sanders:
> @oliver-sanders any idea how to test this in an automated way? Can I create an SSH config for the test shell?

```python
def test_whatever(monkeypatch):
    def mock_c():
        raise NoPlatformError(...)
    monkeypatch.setattr(
         'a.b.c',
         mock_c,
    )
    c()
```

Mangling SSH settings on the fly nasty, I wouldn't functional test this.

cylc/flow/platforms.py
MetRonnie: See below https://github.com/cylc/cylc-flow/blob/dcd1290bdefa1a7e00a4c2998b1c22d54b0983c7/cylc/flow/platforms.py#L213-L216

If we are happy the same principle should be applied to platform groups (in which case it was wrong all this time) then a similar comment should be added above this for loop.

wxtim: I'm pretty confident that the same principle should be applied to platform groups, and for the same reason: We want user created platform groups to take precedence over site-admin set ones.

wxtim:
```suggestion
```

oliver-sanders (commented):
The `get_platforms` interface can raise three possible errors:

* `PlatformLookupError`
* `NoPlatformsError`
* `NoHostsError`

For safety I think every call of `get_platform` which could be called by the scheduler (cat-log etc, don't matter) should catch all three.

Note that we can't combine them at the moment because `NoHostsError` may require special handling.

oliver-sanders:
I've taken a look into more radical change, however, this is going to be a large refactor so not suitable for this release.

cylc/flow/subprocpool.py
```diff
@@ -487,7 +488,14 @@ def _run_callback(callback, args_=None):
         # that you can use that platform's ssh command.
         platform = None
         if isinstance(ctx.cmd_key, TaskJobLogsRetrieveContext):
-            platform = get_platform(ctx.cmd_key.platform_name)
+            try:
```
wxtim, line 491: I've kept this one reasonably simple. I've struggled to recreate this error by changing externals and in the end was only able to manually test it by adding

```python
               raise PLATFORM_LOOKUP_ERRORS[2]('foobar')
```

This isn't really very satisfactory, but I don't think that this likely to cause a real world bug, and that this case is just covering ourselves.

cylc/flow/task_job_mgr.py
```diff
@@ -1166,7 +1166,7 @@ def _prep_submit_task_job(
                     rtconfig, itask.tdef.name, bad_hosts=self.bad_hosts
                 )
 
-            except PlatformLookupError as exc:
+            except (PlatformLookupError, NoPlatformsError) as exc:
```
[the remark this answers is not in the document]
wxtim: `NoHostsError` comes from `get_hosts_from_platform` and not from `get_platform`.

cylc/flow/platforms.py
```diff
@@ -122,6 +122,11 @@ def get_platform(
         platform: A platform definition dictionary. Uses either
             get_platform() or platform_name_from_job_info(), but to the
             user these look the same.
+
+    Raises:
+        NoPlatformsError: Platform group has no platforms with usable hosts.
+        This should be caught if this function is used on a raw on config,
+        or in any other context where a platform group might be selected.
```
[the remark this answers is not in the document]
wxtim: Only `PlatformLookupError`, not `NoHostsError`

cylc/flow/task_pool.py
```diff
@@ -461,6 +461,9 @@ def load_db_task_pool_for_restart(self, row_idx, row):
                     TASK_STATUS_SUCCEEDED
             ):
                 # update the task proxy with platform
+                # Do not capture failure here as we don't want to default
+                # to localhost platform: If a platform cannot be found
+                # Here restart _should_fail.
                 itask.platform = get_platform(platform_name)
```
wxtim, line 467: We could also set the task state to submit failed?

cylc/flow/task_pool.py
```diff
@@ -461,7 +462,12 @@ def load_db_task_pool_for_restart(self, row_idx, row):
                     TASK_STATUS_SUCCEEDED
             ):
                 # update the task proxy with platform
-                itask.platform = get_platform(platform_name)
+                # If we get a failure from the platform selection function
```
wxtim, line 465: Again a case where I'm not convinced platform name could ever be a platform group. I think it's safe to say that any platform which has become inaccessible/been removed from the config (I think that the latter is much more likely here) that the task's submission has failed. This then triggers whatever retry logic the user wants.

cylc/flow/platforms.py
wxtim, line 1: Changes in this file are documentation updates + the removal of an unwanted variable (`'platform group'`) being added to platform object.

cylc/flow/task_events_mgr.py
```diff
@@ -961,6 +964,13 @@ def _process_job_logs_retrieval(self, schd, ctx, id_keys):
                 for id_key in id_keys:
                     self.unset_waiting_event_timer(id_key)
                 return
+        except PlatformLookupError:
```
wxtim, line 967: Again, part of the job-log-retrieval, so the same approach of log and forget.

cylc/flow/task_job_mgr.py
```diff
@@ -310,7 +309,7 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
                                 itask.tdef.rtconfig['platform'],
                                 bad_hosts=self.bad_hosts
                             )
-                        except NoPlatformsError:
+                        except PlatformLookupError:
```
wxtim, line 312: This case has only changed to make the error catching more general - the response to an error is unchanged.

cylc/flow/task_job_mgr.py
```diff
@@ -912,7 +911,13 @@ def _run_job_cmd(
 
         # Go through each list of itasks and carry out commands as required.
         for platform_name, itasks in sorted(auth_itasks.items()):
-            platform = get_platform(platform_name)
+            try:
```
wxtim, line 914: Here I think that the correct choice is to try this and move on logging an error if the command fails.

We might want "try N times" or "try until we get the response we want", or some combo of the two, but I think that needs to be handled separately.

cylc/flow/task_job_mgr.py
```diff
@@ -1165,7 +1170,6 @@ def _prep_submit_task_job(
                 platform = get_platform(
```
wxtim, line 1170: Already Handled, with the change to sub-classing in `exceptions.py`

tests/unit/test_subprocpool.py
```diff
@@ -273,6 +273,20 @@ def test__run_command_exit_no_255_callback(caplog, mock_ctx):
     assert 'callback called' in caplog.records[0].msg
 
 
+def test__run_command_exit_no_gettable_platform(caplog, mock_ctx):
```
wxtim, line 276: I considered using monkeypatch to have `get_platform` to return both exceptions, but decided that was re-testing basic python Exception class inheritances.

cylc/flow/platforms.py
```diff
@@ -190,10 +209,6 @@ def platform_from_name(
     if platform_name is None:
         platform_name = 'localhost'
 
-    platform_group = None
-    # The list is reversed to allow user-set platform groups (which are loaded
-    # later than site set platforms) to be matched first and override site
-    # defined platforms.
```
MetRonnie, line 196: Useful comment deleted?

cylc/flow/platforms.py
```diff
@@ -214,9 +229,7 @@ def platform_from_name(
                 'regular expression. See the documentation for '
                 '"global.cylc[platforms][localhost]" for more information.'
             )
-    # The list is reversed to allow user-set platforms (which are loaded
-    # later than site set platforms) to be matched first and override site
-    # defined platforms.
```
MetRonnie, line 219: Ditto

cylc/flow/platforms.py
wxtim: Useful comment moved to function docstring. I'm open to moving it back.

cylc/flow/platforms.py
wxtim: Will do it for both if I do it for either :. resolving

cylc/flow/platforms.py
MetRonnie: Imho it's better next to the line that is in question

cylc/flow/task_pool.py
oliver-sanders: This should hopefully be a really niche error only really possible due to global config changes since restart. The most likely case for this error is probably a really old workflow being restarted after a global config change.

E.G. we currently have two Cylc platforms representing a single real platform just with different configuration for back-compat reasons. One day we'll remove the back-compat platform, this error could then ensue, yikes!

Submit-failed is definitely wrong because the last recorded status was submitted or greater.
* In the case of succeeded and failed tasks the platform shouldn't really matter any more but I guess it might still be used by some command or another?! Or possibly recorded back to the DB with a task pool dump!?
* For submitted / running tasks it's more of a problem as Cylc will try to poll them.

Perhaps the correct behaviour here is to raise an exception :grin:! We could potentially collect any `PlatformLookupError`s, list the tasks in the error with a suggestion to hack the DB to allow restart to happen.

cylc/flow/task_job_mgr.py
wxtim: Worth handling the two separately?

cylc/flow/task_job_mgr.py
```diff
@@ -912,7 +911,14 @@ def _run_job_cmd(
...
         for platform_name, itasks in sorted(auth_itasks.items()):
-            platform = get_platform(platform_name)
+            try:
+                platform = get_platform(platform_name)
+            except PlatformLookupError:
+                LOG.error(
+                    f'Unable to run command {cmd_key}: Unable to find'
+                    f' platform {platform_name}.'
+                )
+                break
```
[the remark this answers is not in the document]
wxtim:
```suggestion
                break
```

cylc/flow/task_pool.py
```diff
@@ -461,7 +463,18 @@ def load_db_task_pool_for_restart(self, row_idx, row):
...
+                # set task status to submit-failed.
+                try:
+                    itask.platform = get_platform(platform_name)
+                except PlatformLookupError as exc:
+                    LOG.critical(
+                        f'Unable to find platform {platform_name}.'
+                        ' This is probably caused by changes to your'
+                        ' site or personal global.cylc since you last played'
+                        ' this workflow.'
+                    )
```
oliver-sanders, line 476: Suggest printing a message something along the lines of:

```
The following platforms are not defined in the global.cylc file:
* platform_1
* platform_2
* ...

To restart this workflow, please restore the definitions of these platforms in
the Cylc global configuration.
```

wxtim:
> Code LGTM, error handling makes sense, just need to test.

Is this "You (Tim) need to write more tests" or "I (Oliver) want more time to try and break it?

tests/functional/platforms/10-platform-gone-on-restart.t
```diff
@@ -0,0 +1,69 @@
...
+    initial cycle point = 2934
+    [[graph]]
+        R1 = foo
+
+[runtime]
+    [[foo]]
+        script = """
+            # Make the workflow stop, but keep the task running.
+            cylc stop ${CYLC_WORKFLOW_ID} --now --now
+            sleep 1000
```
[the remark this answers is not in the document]
wxtim: No
```suggestion
            cylc stop ${CYLC_WORKFLOW_ID} --now **--now**
```

tests/integration/test_task_events_mgr.py
```diff
@@ -0,0 +1,44 @@
...
+from cylc.flow.scheduler import Scheduler
+
+from typing import Any as Fixture
+
+
+async def test_process_job_logs_retrieval_warns_no_platform(
+    one_conf: Fixture, flow: Fixture, scheduler: Fixture, run: Fixture,
+    db_select: Fixture, caplog: Fixture
+):
+    """Test the paused status is unset on normal shutdown."""
```
MetRonnie, line 27: ?

tests/integration/test_task_job_mgr.py
```diff
@@ -104,3 +107,24 @@ async def test_run_job_cmd_no_hosts_error(
             log,
             contains='No available hosts for no-host-platform',
         )
+
+
+async def test__run_job_cmd_logs_platform_lookup_fail(
+    one_conf: Fixture, flow: Fixture, scheduler: Fixture, run: Fixture,
+    db_select: Fixture, caplog: Fixture
+) -> None:
+    """Test the paused status is unset on normal shutdown."""
```
MetRonnie, line 116: ?

MetRonnie (approved):
I have followed your manual test recipe

hjoliver:
Flake8 reporting a few of these in the CI fast tests:
> C419 Unnecessary list comprehension passed to any() prevents short-circuiting - rewrite as a generator.

MetRonnie:
Actually all that was needed was closing + reopening. This causes GH Actions to create a new merge commit that it uses at checkout

MetRonnie (approved):

cylc/flow/platforms.py
```diff
@@ -117,11 +117,22 @@ def get_platform(
...
 
     Returns:
         platform: A platform definition dictionary. Uses either
             get_platform() or platform_name_from_job_info(), but to the
             user these look the same.
+
+    Raises:
+        NoPlatformsError:
+            Platform group has no platforms with usable hosts.
+            This should be caught if this function is used on a raw on config,
```
hjoliver, line 131:
```suggestion
            This should be caught if this function is used on a raw config,
```

hjoliver (approved):
From a quick pass-through this looks good to me. I'll leave @oliver-sanders for the merge though, since he has reviewed it pretty extensively.

oliver-sanders (approved):

PR #5400, opened 2023-03-08
8.1.x

oliver-sanders:


MetRonnie (approved):
I have noticed this fast test flakiness on MacOS before, shouldn't affect real workflows however
```
FAILED tests/unit/cycling/test_iso8601.py::test_validate_fails_comma_sep_offset_list[next (T-00, T-30)-T-00;T-30] -
    AssertionError: Regex pattern did not match.
 Regex: 'T-00;T-30'
 Input: "type object 'WorkflowSpecifics' has no attribute 'point_parser'"
FAILED tests/unit/cycling/test_iso8601.py::test_validate_fails_comma_sep_offset_list[next (wildebeest)-Invalid ISO 8601 date] -
    AssertionError: Regex pattern did not match.
 Regex: 'Invalid ISO 8601 date'
 Input: "type object 'WorkflowSpecifics' has no attribute 'point_parser'"
```

oliver-sanders:
This test fails on Linux too, it only passes if it is run after a prior test so is vulnerable to sort order.

PR #5402, opened 2023-03-08
update Gedit syntax highlighting instructions

wxtim:
The "2.0" referred to by the removed sentence refers to the XML version, not the GTK version, as implied.

https://gitlab.gnome.org/GNOME/gtksourceview/-/blob/master/docs/lang-tutorial.md#a-language-definition-for-the-c-language

**Check List**

- [x] No tests, a documentation change.

### DIFF
--- a/tests/functional/platforms/02-host-to-platform-upgrade.t
+++ b/tests/functional/platforms/02-host-to-platform-upgrade.t
@@ -45,6 +45,7 @@ grep_ok "\[T2\]\[remote\]host = ${CYLC_TEST_HOST}"\
     "${TEST_NAME_BASE}-run.stderr"
 
 # Run the workflow
+echo $CYLC_TEST_HOST >&2
 workflow_run_ok "${TEST_NAME_BASE}-run" \
     cylc play --debug --no-detach --reference-test \
     -s CYLC_TEST_HOST="'$CYLC_TEST_HOST'" \
--- a/cylc/flow/etc/syntax/cylc.lang
+++ b/cylc/flow/etc/syntax/cylc.lang
@@ -10,8 +10,7 @@
  /usr/share/gtksourceview-2.0/language-specs/
 
  If your installation uses GNOME 3, the '2.0' in the paths will need
- to be '3.0', and the version="2.0" string below will need changing
- as well.
+ to be '3.0'.
 
  If using gedit, gedit will need to be totally closed down and reloaded.
 


tests/functional/platforms/02-host-to-platform-upgrade.t
```diff
@@ -45,6 +45,7 @@ grep_ok "\[T2\]\[remote\]host = ${CYLC_TEST_HOST}"\
     "${TEST_NAME_BASE}-run.stderr"
 
 # Run the workflow
+echo $CYLC_TEST_HOST >&2
```
MetRonnie, line 48: 🙅‍♂️

MetRonnie:
> The "2.0" referred to by the removed sentence refers to the XML version, not the GTK version, as implied.

I don't think that's correct. The XML version is in the top of the file and is 1.0

tests/functional/platforms/02-host-to-platform-upgrade.t
wxtim:
```suggestion
```

wxtim:
The document linked says

> The attribute version specifies the version of the xml syntax used in your language definition file, so it should always be 2.0.

Having said which it looks mega inconsistent:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<language id="c" name="C" version="2.0" _section="Source">
```

I went looking but can't find any more info. Not willing to devote a whole lot of time. Happy to write it off as bottom right of 

```
Worth knowing
              ^
              |
              |
--------------+-----------------> difficult to understand
              |
              |
```

MetRonnie (approved):

mikemarsh-mo:
> The document linked says
> 
> > The attribute version specifies the version of the xml syntax used in your language definition file, so it should always be 2.0.
> 
> Having said which it looks mega inconsistent:
> 
> ```
> <?xml version="1.0" encoding="UTF-8"?>
> <language id="c" name="C" version="2.0" _section="Source">
> ```
> 
> I went looking but can't find any more info. Not willing to devote a whole lot of time. Happy to write it off as bottom right of
> 
> ```
> Worth knowing
>               ^
>               |
>               |
> --------------+-----------------> difficult to understand
>               |
>               |
> ```

I think where the above tutorial says:
> version of the xml syntax used in your language definition file
I think what it means is the version of the gtksourceview language file syntax rather than XML version. It's worded a bit clumsily but I think it means (gtksourceview language file syntax) version="2.0" whilst <?xml version="1.0"...>

PR #5405, opened 2023-03-13
Tweak scan CLI help.

hjoliver:
- Improve `cylc scan` command help
- add scheduler PID to the output
- print a note **to stderr** to remind that scan only reads contact files by default

Pingless:
```console
$ cylc scan 
demo/run9 niwa-1007885.niwa.local:43070 21212
demo/run7 niwa-1007885.niwa.local:43010 21132
demo/run8 niwa-1007885.niwa.local:43069 21173

Use `--ping` to attempt to connect to schedulers and clean 
up after any that were killed or did not shut down cleanly.
```

Pingful:
```console
$ cylc scan 
demo/run9 niwa-1007885.niwa.local:43070 21212
demo/run7 niwa-1007885.niwa.local:43010 21132
demo/run8 niwa-1007885.niwa.local:43069 21173
```

**Check List**

- [x] Tests are included (or explain why tests are not needed). [Not needed: minor documentation only]


hjoliver:
As an aside,  `cylc scan -t json` just prints contact file content plus a little more static data, e.g. contact file location.

@oliver-sanders was that the intention for `-t json`? If so, we should document that and disallow use with other options e.g. `-t rich -t json` which currently doesn't have the expected effect.

oliver-sanders:
> @oliver-sanders was that the intention for -t json?

I think this was inherited from Cylc 7. It's just a better interface for any systems which want to scape this information.

> If so, we should document that and disallow use with other options e.g. -t rich -t json which currently doesn't have the expected effect.

???

`cylc scan --format=json --format=rich` results in `--format=rich`

In the same way that:

`git commit --author=one --author=two` results in `--author=two`

If you repeat an argument, the latter definition is used.

hjoliver:
> If you repeat an argument, the latter definition is used.

Duh :facepalm: 

Well, not exclusively true but fair enough. 

```
cylc config -i foo -i bar
```

What I was getting at was: was [arrggh so many "was"s] the intention to enable return of kinds of `cylc scan` outputs in json format? 

It would be reasonable to expect that `-t json` does that. But what it actually does is return the contents of the contact file (albeit in json format). We should probably rename the option to `--contact-file` ...

hjoliver:
I've stuck with just documenting what `-t json` does better:
```
json: full contact data in JSON format
```

cylc/flow/scripts/scan.py
```diff
@@ -568,7 +585,21 @@ async def main(
...
+        or opts.source
+        or not nflows
+    ):
+        return
+
+    print(
+        "\nUse `--ping` to attempt to connect to schedulers and clean "
+        "\nup after any that were killed or did not shut down cleanly.",
+        file=stderr
+    )
```
oliver-sanders, line 602: I think printing this message every time is a bit too obstructive.

hjoliver: Well, it's only printed for non scheduler-contacting cases, and only to stderr.  And I got here via a bunch of confused-user questions (*"`cylc scan` says my workflow is running, but stuff isn't working!"*).

oliver-sanders: > Well, it's only printed for non scheduler-contacting cases

So the default i.e. 99.9% of cases.

oliver-sanders: I feel that if it is really this important that we add a message to every command, then it should really be the default.

But we don't want to do that in order to keep `cylc scan` fast.

#5511 possible alternative?

hjoliver: This situation is particularly confusing for users when it does happen, because they have a reasonable expectation that `cylc scan` does not report stopped workflows as running workflows.

However, I'll remove the message.  I quite like your proposed #5511 solution.

hjoliver:
@oliver-sanders - this now just clarifies the command help, and adds scheduler PID to the basic scan output. Which makes it easier for users to see if a scheduler really is running, if they need to do that.

One review should do now.

PR #5412, opened 2023-03-15
Fix expire triggers.

### LINKED ISSUE #5354: graph parser: expire-all ?
The `expire-all` trigger (valid at Cylc 7) appears to be missing from Cylc 8:

https://github.com/cylc/cylc-flow/blob/76762bbfc1ca83be88ed8b0e0a7224e6af520964/cylc/flow/graph_parser.py#L105-L116

I'm not sure whether `expire-all` was "officially supported" or just a side-effect of the way the logic worked, however, it seems like a sensible trigger condition to me. I'm not sure whether its removal was intentional, nothing springs to mind.

At Cylc 8 such a trigger would probably required a `?` modifier (as succeeded and expired are orthogonal), but should still work.

**Pull requests welcome!**

### LINKED ISSUE #5364: :expire trigger broken
By broken I mean, it has no effect.

When implementing the spawn-on-demand scheduler, I think I must have forgotten about the `:expired` trigger, and hence removed expired tasks before triggering on that "output" (because an expired task does not have any *other* outputs).

### LINKED ISSUE #5447: expiry: broken interaction with runahead limit
There is a bug in the interaction between runahead limit and task expiry.

In the following example, Cylc spawns the task out to the runahead limit, these tasks then clock-expire. The task pool is now empty so the workflow shuts down, job done, despite there still being un-expired tasks left to run before the FCP.

```
[scheduling]                                         
    initial cycle point = previous(T-00) - PT5H    
    final cycle point = +PT10H    
    runahead limit = P3                                                                                                            
    [[special tasks]]                                
        clock-expire = task1(PT0H)                   
    [[dependencies]]                                 
        [[[R//PT1H]]]                                
            graph = "@wall_clock => task1"           
                                                     
[runtime]                                            
    [[task1]]
```

hjoliver:
- fix expire trigger: close #5364 
- fix expire-all/any triggers: close #5354  
- fix #5447

And require `?` (optional) on all expire triggers.

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/596

### DIFF
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -686,6 +686,7 @@ class TaskPool:
             if not self.hidden_pool[itask.point]:
                 del self.hidden_pool[itask.point]
             LOG.debug(f"[{itask}] {msg}")
+            self.task_queue_mgr.remove_task(itask)
             return
 
         try:
@@ -696,9 +697,9 @@ class TaskPool:
             self.main_pool_changed = True
             if not self.main_pool[itask.point]:
                 del self.main_pool[itask.point]
-                self.task_queue_mgr.remove_task(itask)
-                if itask.tdef.max_future_prereq_offset is not None:
-                    self.set_max_future_offset()
+            self.task_queue_mgr.remove_task(itask)
+            if itask.tdef.max_future_prereq_offset is not None:
+                self.set_max_future_offset()
 
             # Notify the data-store manager of their removal
             # (the manager uses window boundary tracking for pruning).
@@ -815,8 +816,8 @@ class TaskPool:
 
         for itask in released:
             itask.state_reset(is_queued=False)
-            itask.waiting_on_job_prep = True
             self.data_store_mgr.delta_task_queued(itask)
+            itask.waiting_on_job_prep = True
 
             if cylc.flow.flags.cylc7_back_compat:
                 # Cylc 7 Back Compat: spawn downstream to cause Cylc 7 style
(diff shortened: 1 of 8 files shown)


hjoliver:
Converted back to draft temporarily, in light of today' project meeting.

I think have nailed the expire stall/no-stall situation, ran out of time to write it up this afternoon though.

cylc/flow/task_pool.py
```diff
@@ -686,6 +686,7 @@ def remove(self, itask, reason=""):
             if not self.hidden_pool[itask.point]:
                 del self.hidden_pool[itask.point]
             LOG.debug(f"[{itask}] {msg}")
+            self.task_queue_mgr.remove_task(itask)
```
oliver-sanders, line 689: Unrelated bug?

hjoliver: It is a bug fix, but not unrelated. Expired tasks need to be removed as complete (possibly still subject to debate!) but the "remove" process was not removing the target task from the queues.

hjoliver:
Note this PR is largely orthogonal to the additional and currently unresolved discussion of what we should do in response to an expiration event beyond just triggering off of it (e.g. stall as well if the graph doesn't explicitly handle the expiration). However it might make sense to bundle the result of that with this PR.

cylc/flow/task_pool.py
hjoliver: Having run into this bug again, I'll cherry-pick it to a separate bug fix PR. (As task expiry is a bigger discussion).

hjoliver: Having run into this bug again, I'll cherry-pick it to a separate bug fix PR. (As task expiry is a bigger discussion).... #5573

hjoliver: ... #5573

hjoliver:
Closing as superseded by #5658

PR #5422, opened 2023-03-21
tui: sort negitive numbers

### LINKED ISSUE #5421: TUI natural sort - negative numbers
_[Reported on Discourse](https://cylc.discourse.group/t/cylc-tui-display-order/611)_

TUI should sort negative numbers before positive ones even if they occur midway through the string. However, the current behaviour is:

TUI screenshot

See https://github.com/cylc/cylc-flow/blob/6fc3c58b7798e95f0aa6b0ed18b70724a2309449/cylc/flow/tui/util.py#L192-L193

oliver-sanders:
* Closes #5421

**Check List**

### DIFF
--- a/cylc/flow/tui/util.py
+++ b/cylc/flow/tui/util.py
@@ -211,6 +211,12 @@ class NaturalSort:
         >>> N('10') < N('9')
         False
 
+        Negative integers should sort correctly:
+        >>> N('-1') < N('0')
+        True
+        >>> N('1') == N('-1')
+        False
+
         Integers rank higher than strings:
         >>> N('1') < N('a')
         True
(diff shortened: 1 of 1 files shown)


hjoliver:
Presumably this reverses the sort order for your other example on the issue? https://github.com/cylc/cylc-flow/issues/5421#issuecomment-1478099669

I prefer hyphens over underscores in task names, so I don't think we can assume that "-\d" inside a string is always a negative number.

oliver-sanders:
I'm not convinced this is a good idea, it may be helpful in some cases but has the potential to backfire.

oliver-sanders:
See https://github.com/cylc/cylc-flow/issues/5421#issuecomment-1484914905

Issue #5435, opened 2023-03-28
efficiency: increment_graph_window


oliver-sanders:
Re-issue of: https://github.com/cylc/cylc-flow/issues/5315

The scaling of `increment_graph_window` has improved, but is still severe enough to prevent some Cylc 7 workflows from being able to run with Cylc8.

See Element for discussion:

https://matrix.to/#/!gMuCeOJBhYqgWDGwnR:matrix.org/$1679999387176196Xelev:matrix.org?via=matrix.org

### Description

For the workflow `a => b<x>` there are `x + 1` tasks and `x` dependencies.

However, for `x=700` the `increment_graph_window` function is called in excess of 3'000'000 times.

### Reproducible Example

```ini
[task parameters]
   m = 0..7000
[scheduling]
   [[queues]]
      [[[default]]]
         limit = 4
   [[graph]]
      R1 = "a => b<m>"
[runtime]
   [[a, b<m>]]
```

```console
$ cylc play <id> --profile --no-detach
```

Run this for a few mins, when you get bored ctrl+c it, then:

```console
$ pip install snakeviz
$ snakeviz ~/cylc-run/<id>/log/scheduler/profile.prof
```

### Expected Behaviour

It should take seconds, not minutes to spawn the 700 tasks in this workflow.

oliver-sanders:
I left this example running for ~20 mins, then ctrl+c'ed it. The `increment_graph_window` function was called 30'276'650 times in this period.

![Screenshot from 2023-03-28 15-41-03](https://user-images.githubusercontent.com/16705946/228274997-a100b1d2-14dc-42f9-a2d5-7527e92dd270.png)

Naively, I would guess that this should be ok?

```diff
diff --git a/cylc/flow/data_store_mgr.py b/cylc/flow/data_store_mgr.py
index f47cfcbbb..8d13cf598 100644
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -741,6 +741,15 @@ class DataStoreMgr:
                 source_tokens.id
             )
 
+        if self.n_window_nodes.get(active_id):
+            # we've already walked this node
+            if edge_distance == 0:
+                if active_id in self.prune_trigger_nodes:
+                    self.prune_flagged_nodes.update(
+                        self.prune_trigger_nodes[active_id])
+                    del self.prune_trigger_nodes[active_id]
+            return
+
         # Setup and check if active node is another's boundary node
[...]

PR #5436, opened 2023-03-28
task_pool: avoid listing tasks where not necessary

oliver-sanders:
Some low-hanging fruit spotted whilst investigating #5435

The task_pool code was repeatedly requesting the same sorted list of tasks which was starting to bite for a 7001 task workflow.

* Re-use the cached task list where appropriate.
* Switch to using the task dictionary for "in" comparisons for faster performance.

I profiled this using the example in #5435 and the [flawed] patch from https://github.com/cylc/cylc-flow/issues/5435#issuecomment-1487031362. I pressed ctrl+c once the first `b<x>` task has submitted.

This diff brings the time taken by the `can_spawn` function down from 16.93s to 0.2046s which is a saving of ~5% of the overall runtime!

~Question: 8.1.3 or 8.2.0?~

**Check List**

### DIFF
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -128,7 +128,7 @@ class TaskPool:
         self.task_name_list = self.config.get_task_name_list()
         self.task_queue_mgr = IndepQueueManager(
             self.config.cfg['scheduling']['queues'],
-            self.config.get_task_name_list(),
+            self.task_name_list,
             self.config.runtime['descendants']
         )
         self.tasks_to_hold: Set[Tuple[str, 'PointBase']] = set()
@@ -137,7 +137,7 @@ class TaskPool:
         """Set stop after a task."""
         tokens = Tokens(task_id, relative=True)
         name = tokens['task']
-        if name in self.config.get_task_name_list():
+        if name in self.config.taskdefs:
             task_id = TaskID.get_standardised_taskid(task_id)
             LOG.info("Setting stop task: " + task_id)
             self.stop_task_id = task_id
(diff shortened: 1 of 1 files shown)


oliver-sanders:
### Before

Screenshot from 2023-03-28 16-50-03

### After

Screenshot from 2023-03-28 16-50-17

hjoliver:
Nice.

Shows there's plenty of scope for optimizing the new system yet.

Test fails are just codecov upload.

> Question: 8.1.3 or 8.2.0?

I'd vote for getting these kinds of changes in as quickly as possible, so 8.1.3 if we do it.

This is not strictly a bug fix, but then it's not really a user-facing enhancement or new feature either unless you're running certain kinds of nasty workflow.

hjoliver (approved):

oliver-sanders:
Agreed, rebased onto 8.1.x.

oliver-sanders:
@wxtim please check you're happy that this will work correctly post-reload.

cylc/flow/task_pool.py
```diff
@@ -128,7 +128,7 @@ def __init__(
         self.task_name_list = self.config.get_task_name_list()
         self.task_queue_mgr = IndepQueueManager(
             self.config.cfg['scheduling']['queues'],
-            self.config.get_task_name_list(),
+            self.task_name_list,
```
wxtim, line 131: what's the difference between this and `self.config.taskdefs` - is this the cached version?

How do you know when to use each?

wxtim (approved):

Issue #5437, opened 2023-03-29
Spawning + datastore with thousands of "ready" tasks


hjoliver:
Not strictly a bug, but some extreme optimization is required in Cylc 8 for pathological workflows where a huge number of tasks hit the active window at once.

Here the scheduler has to spawn 7000 tasks all at once, off of `a:succeed`, which takes ~5 minutes on my fairly powerful laptop, during which time the scheduler is unresponsive.
```ini
[task parameters]
   m = 0..6999  # !!! 
[scheduling]
   [[queues]]
      [[[default]]]
         limit = 4
   [[graph]]
      R1 = "a => b<m>"  # !!!
[runtime]
   [[a]]
      script = sleep 10
   [[b<m>]]
```

Initial profiling results from @oliver-sanders shows:
1. primarily, the datastore n-window computation is responsible
2. secondarily (much less time), each spawned tasks needs a database read to get submit number and flow info

> Ok, I got bored after 20mins or so and cut the run off at that point. FYI, if you ctrl+c your workflow, the [profile.prof](http://profile.prof/) file still gets generated.
>
> -    The spawn_on_output function itself took 0.1562s.
> -    The increment_graph_window function in the data store took 1139s (including its resulting calls).
>
> So it's the data store not the task pool. The increment_graph_window function was called 4'325 times, but called itself recursively 30'276'650 times which is where the CPU gets soaked up.
>
> This is more-or-less as expected, we knew this function was being called more times than necessary, see this comment - https://github.com/cylc/cylc-flow/pull/5319#issuecomment-1400528271
[data-store graph window efficiency/refactor by dwsutherland · Pull Request #5319 · cylc/cylc-flow](https://github.com/cylc/cylc-flow/pull/5319#issuecomment-1400528271) - GitHub
Partially addresses #5315 Supersedes #5316 Check List I have read [CONTRIBUTING.md](http://contributing.md/) and added my name as a Code Contributor. Contains logically grouped changes (else tidy your branch by rebase). ...
>
> Two suggestions:
>
> -    If possible, batch together the increment_graph_window / task spawning to reduce the number of top-level calls to increment_graph_window.
>   -   Would require heavy refactoring, the function is designed to expand the graph around one task at a time.
>   -   Potential savings ~4000x
>
> Come up with a more efficient approach to increment_graph_window.
> -   I.E. Remember which nodes we have already visited to avoid repeat visits.
[...]

PR #5438, opened 2023-03-29
Batch spawn POC

hjoliver:
Zero-th level workaround for #5437 

(We still need to optimize the problematic code, but this approach might be useful in the interim, and possibly in the long run as well).

UPDATE: the main problem (n-window computation) was fixed by #5475. But it may still be worth doing this as well.

When an output gets completed, instead of spawning all children into the task pool at once, record what needs to be spawned, and spawn them batch-wise via the main loop.

This plays well with queuing, because queues work with what they've got, so tasks can be released to run throughout the long spawning period.

The example from #5437 is quite usable on this branch, although CPU remains high till spawning is complete, and only the GUI table view is workable (and that with filtering for active tasks):

```ini
[task parameters]
   m = 0..6999
[scheduling]
   [[queues]]
      [[[default]]]
         limit = 4
   [[graph]]
      R1 = "a => b<m>"
[runtime]
   [[a]]
      script = sleep 10
   [[b<m>]]
```

**Check List**

### DIFF
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1221,26 +1221,25 @@ class TaskPool:
             if forced:
                 self.tasks_to_spawn_forced[
                     (itask, output)
-                ] = itask.graph_children[output]
+                ] = list(itask.graph_children[output])
             else:
                 self.tasks_to_spawn[
                     (itask, output)
-                ] = itask.graph_children[output]
+                ] = list(itask.graph_children[output])
         except KeyError:
             # No children depend on this output
             pass
         else:
             self.spawn_children()
 
-	# Remove after spawning, to avoid briefly emptying the task pool
-	# in simple cases (foo[-P1] => foo) - which can lead to shutdown.
+        # Remove after spawning, to avoid briefly emptying the task pool
+        # in simple cases (foo[-P1] => foo) - which can lead to shutdown.
         if not forced and output in [
             TASK_OUTPUT_SUCCEEDED,
             TASK_OUTPUT_EXPIRED,
             TASK_OUTPUT_FAILED
         ]:
             self.remove_if_complete(itask)
-
 
     def spawn_children(self):
         self._spawn_children(self.tasks_to_spawn)
@@ -1305,8 +1304,7 @@ class TaskPool:
                             (str(itask.point), itask.tdef.name, output)
                         })
                         self.data_store_mgr.delta_task_prerequisite(t)
-                        # Add it to the hidden pool or move it to the main
-                        # pool.
+                        # Add it to hidden pool or move it to main pool.
                         self.add_to_pool(t)
 
                         if t.point <= self.runahead_limit_point:


hjoliver:
(Tests not happy but I'm pretty sure it'll be something trivial...)

cylc/flow/scheduler.py
```diff
@@ -1645,6 +1645,7 @@ async def main_loop(self) -> None:
             # Shutdown workflow if timeouts have occurred
             self.timeout_check()
 
+            self.pool.spawn_children()
```
oliver-sanders, line 1648: I think a lot of the functional tests can be sensitive to the order of main loop events. Maybe try to relocate this where spawning happened before (process_queued_task_messages?).

cylc/flow/task_pool.py
```diff
@@ -1214,70 +1216,112 @@ def spawn_on_output(self, itask, output, forced=False):
...
+            TASK_OUTPUT_EXPIRED,
+            TASK_OUTPUT_FAILED
+        ]:
+            self.remove_if_complete(itask)
+
         try:
-            children = itask.graph_children[output]
+            if forced:
+                self.tasks_to_spawn_forced[
+                    (itask, output)
```
oliver-sanders, line 1230: `(itask, output, forced)` to avoid needing two dicts?

Possibly consider using a deque or list rather than dict unless there's a need to perform `in` comparisons.

oliver-sanders (commented):
This seems like the best option to me.

Tried it out, the workflow is able to make progress as expected.

I found that client connections were taking longer than expected. Not sure why, a bit of debugging showed that once the requests were received, they were processed quickly by the workflow so it's the connection/security bit which is slow?

For the most part connections worked fine, but requests with larger payloads e.g. `cylc dump` and `cylc tui` don't work without a major increase in `--comms-timeout`. Assuming we can reduce the number of `increment_graph_window` calls this problem should go away and we should be able to up the max number of tasks spawned per main-loop cycle.

oliver-sanders:
FYI: I've bumped into a workflow today which would benefit from this change even after the `increment_graph_window` efficiency improvements.

hjoliver:
OK, I'll resurrect this and try to finish it off soon.

hjoliver:
Damn it, I've just realized a problem with this approach. 😡 

Sometimes we need to update the prerequisites of already-spawned tasks. On this branch, it's possible for those tasks to queued for spawning by main loop. So for this to work, we'll need a more complex data-structure that records tasks-to-be-spawned **and** prerequisites to update immediately when they are spawned.

hjoliver:
> I think the advantages of this have been wiped out by the n-window changes.

I'm not sure of that.  Extreme mass-spawning cases (`foo => bar<0...N>` for large enough `N`) likely still need this, to avoid scheduler lock-up. It's unfortunate that I haven't got back to it in 2 years, but we've had higher priorities...

oliver-sanders:
> I'm not sure of that. Extreme mass-spawning cases (foo => bar<0...N> for large enough N) likely still need this

Do they? From the `TaskProxy` perspective, queues should protect us?

hjoliver:
> Do they? From the TaskProxy perspective, queues should protect us?

From running all at once, yes, but not from having to spawn a gazillion proxies at once.  I'm not sure, I suppose we need to do some experiments with the current system.

But your opinion seems to have changed, since your comment above? https://github.com/cylc/cylc-flow/pull/5438#issuecomment-1536337684

PR #5439, opened 2023-03-29
CLI: short option changes

### LINKED ISSUE #5227: play: rename -n short option
> Follow-on to #5121

Both `--workflow-name` (`cylc install`) and `--no-detach` (`cylc play`) share the `-n` short option. The option-merging functionality of `cylc vip` removes this conflicting short option so it can be used for neither.

Since `cylc vip --workflow-name` is a common use case and a pain to type we should consider renaming the `--no-detach` short option to avoid this conflict allowing the `-n` short option to be used for `--workflow-name`.

> **Note:** `git grep -- -n` in the functional tests *before* running them after making this change, otherwise functional tests will timeout all over the place!

If the usage of `-n` for `--no-detach` is to engrained we could consider changing the `cylc vip` logic to allow the short opt to be hardcoded for `--workflow-name`.

**Pull requests welcome!**

oliver-sanders:
Closes #5227

* Rename `-n` for `--no-detach` to `-N`. This removes the conflict with `--workflow-name` allowing the option to be used in `cylc vip`.
* Add `-r` as a short option for `--run-name`.

Note, `-r` is used as a short option for the `--raw` output option, however, this doesn't conflict with install/play/clean/reload/reinstall/tui so shouldn't be a conflict risk with any future compound commands.

**Check List**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,15 @@ creating a new release entry be sure to copy & paste the span tag with the
 `actions:bind` attribute, which is used by a regex to find the text to be
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
+
+-------------------------------------------------------------------------------
+## __cylc-8.2.0 (<span actions:bind='release-date'>Upcoming</span>)__
+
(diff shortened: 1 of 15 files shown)


oliver-sanders:
(checked that the functional test failures were all codecov)

PR #5441, opened 2023-03-29
efficiency: pass tokens instances to get_broadcast

oliver-sanders:
A small efficiency improvement which reduces the number of times the tokenise/detokenise interfaces get hit by passing the parsed objects around as arguments.

Profiling results for a workflow with a single task:

| Method | Before (calls) | After (calls) |
| --- | --- | --- |
| tokenise | 21 | 6 |
| detokenise | 37 | 36 |

So that's a reduction of 15 tokenise and 1 detokenise call per task. The tokenise calls take 7.095e-06s so not a big deal, but collectively they can add up.

**Check List**

### DIFF
--- a/cylc/flow/broadcast_mgr.py
+++ b/cylc/flow/broadcast_mgr.py
@@ -28,7 +28,6 @@ from cylc.flow.broadcast_report import (
     get_broadcast_bad_options_report,
 )
 from cylc.flow.cfgspec.workflow import SPEC
-
 from cylc.flow.cycling.loader import get_point, standardise_point_string
 from cylc.flow.exceptions import PointParsingError
 from cylc.flow.parsec.util import listjoin
--- a/cylc/flow/broadcast_mgr.py
+++ b/cylc/flow/broadcast_mgr.py
@@ -18,6 +18,7 @@
 import re
 from copy import deepcopy
 from threading import RLock
+from typing import Optional, TYPE_CHECKING
 
 from cylc.flow import LOG
 from cylc.flow.broadcast_report import (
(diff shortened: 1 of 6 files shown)


hjoliver:
Looks good.

cylc/flow/broadcast_mgr.py
```diff
@@ -156,28 +161,21 @@ def expire_broadcast(self, cutoff=None, **kwargs):
...
-        if not task_id:
+        if tokens is None or tokens == 'None':
             # all broadcasts requested
             return self.broadcasts
-        try:
-            tokens = Tokens(task_id, relative=True)
-            name = tokens['task']
-            point_string = tokens['cycle']
-        except ValueError:
-            raise Exception("Can't split task_id %s" % task_id)
```
oliver-sanders, line 171: I've removed this because tokens objects are pre-parsed so this error cannot occur here any more.

cylc/flow/broadcast_mgr.py
```diff
@@ -27,12 +28,16 @@
     get_broadcast_bad_options_report,
 )
 from cylc.flow.cfgspec.workflow import SPEC
-from cylc.flow.id import Tokens
+
```
wxtim, line 31: Delete entire line?

```suggestion
```

cylc/flow/id.py
```diff
@@ -152,12 +152,16 @@ def __repr__(self):
...
 
     def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
         return all(
             self[key] == other[key]
             for key in self._KEYS
         )
 
     def __ne__(self, other):
```
wxtim, line 162: Can you not just do
```
return not self.__eq__(other) and save yourself 5 lines of code?

wxtim (approved):
Nice optimization

cylc/flow/id.py
oliver-sanders: More efficient.

hjoliver (approved):

PR #5445, opened 2023-03-31
Fix.platform is regex remote tidy fail

### LINKED ISSUE #5429: Remote Tidy: Cannot get platform from install target if platform config is regex
### Synopsis

Remote tidy logic gets a platform from install targets recorded in `TaskRemoteMgr.remote_init_map`.

Currently it uses the `globalconfig.get(['platforms'])` and attempts to do an exact match, so if platform is a regex and install target is not (it shouldn't be) then it's possible not to find a matching platform.

The platform name regex in the global config is not the platform name, and unless it has exactly one possible expansion we cannot infer the platform from an install target - we'd need to know the platform name as well.

### Reproducible Example

```
# global.cylc
[platforms]
  [[someremote0.]]
```

```
# chunk of flow.cylc
[runtime]
  [[task]]
    platform = someremote01
```

Will result in  `UnboundLocalError: local variable 'platform' referenced before assignment`.

### Where to look in the codebase

- `cylc.flow.task_remote_mgr.py` ~ line 300, method `remote_tidy`. https://github.com/cylc/cylc-flow/blob/8d7cfcbeba961dda951c18873fd561573ccc7e15/cylc/flow/task_remote_mgr.py#L304-L306
- `cylc.flow.platforms.py` ~ line 670, method `get_all_platforms_for_install_target`. https://github.com/cylc/cylc-flow/blob/8d7cfcbeba961dda951c18873fd561573ccc7e15/cylc/flow/platforms.py#L652

### Also Check

- Are there other instances of `globalconfig.get(['platforms'])` as we cannot be sure they are safe.
- Are there other uses of `get_all_platforms_for_install_target`

### Expected Behaviour

- Not attempting to get platform from install target.
- Only getting a platform from the install target if the platform name regex in the global config has one and only one possible match.

wxtim:
Closes #5429 

**Check List**


tests/integration/test_task_remote_mgr.py
```diff
@@ -0,0 +1,104 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from cylc.flow.task_remote_mgr import (
+    REMOTE_FILE_INSTALL_DONE,
+    REMOTE_FILE_INSTALL_FAILED
+)
+
+
+async def test_remote_tidy(
```
wxtim, line 23: I don't love this test. Suggestions for the streamlining of it would be gratefully accepted.

cylc/flow/platforms.py
```diff
@@ -649,32 +650,52 @@ def is_platform_with_target_in_list(
...
+        except PlatformLookupError:
+            # We only need one working platform per install target:
+            continue
+        else:
+            # Install Target
+            if 'install target' in platform and platform['install target']:
+                install_targets_map[platform['install target']].append(
+                    platform)
+            else:
+                install_targets_map[platform['name']].append(platform)
```
oliver-sanders, line 686: I think there is a potential `KeyError` here if a platform's install target isn't in the list.

```suggestion
    install_targets_map: Dict[str, List] = {}
    for platform_name in platform_names:
        try:
            platform = get_platform(platform_name)
        except PlatformLookupError:
            # We only need one working platform per install target:
            continue
        install_targets_map.setdefault(
            platform.get('install target', platform['name']), []
        ).append(platform)
```

cylc/flow/task_remote_mgr.py
```diff
@@ -293,19 +295,39 @@ def remote_tidy(self) -> None:
...
+            self.db_mgr.get_pri_dao().select_platforms_used())
+        # For each install target compile a list of platforms:
+        install_targets = {
+            target for target, msg
+            in self.remote_init_map.items()
+            if msg == REMOTE_FILE_INSTALL_DONE
+        }
+        install_targets_map, unreachable_targets = (
+            map_platforms_used_for_install_targets(
+                platforms_used, install_targets))
```
oliver-sanders, line 309: Need to handle the case where `platforms_used` is `None`.

cylc/flow/task_remote_mgr.py
```diff
@@ -293,19 +295,39 @@ def remote_tidy(self) -> None:
...
+
         # Issue all SSH commands in parallel
         queue: Deque[RemoteTidyQueueTuple] = deque()
-        for install_target, message in self.remote_init_map.items():
-            if message != REMOTE_FILE_INSTALL_DONE:
-                continue
+        for install_target in install_targets_map.keys():
             if install_target == get_localhost_install_target():
                 continue
+            platform = random.choice(list(install_targets_map[install_target]))
```
oliver-sanders, line 327: Whilst we're here we might as well IHS this logic i.e:

```python
for platform in install_targets_map[install_target]:
    try:
        ...
    except NoHostsError as exc:
        ...
        # move onto the next platform
    else:
        break  # done
```

tests/integration/test_task_remote_mgr.py
```diff
@@ -0,0 +1,104 @@
...
+
+        # We can get foo bar baz, and we try to remote tidy them.
+        # (This will ultimately fail, but past the point we are testing).
+        for target in ['foo', 'bar9', 'baz']:
+            msg = f'platform: {target} - remote tidy (on {target})'
+            assert msg in records
+
+        # We haven't done anything with Quiz because we're only looking
+        # at cases where platform == REMOTE_FILE_INSTALL_DONE
+        assert not [r for r in records if 'quiz' in r]
```
oliver-sanders, line 104: Can dedent this bit (don't need the scheduler to be running any more:

```suggestion
    records = [str(r.msg) for r in log.records]

    # We can't get qux, no defined platform has a matching install target:
    qux_msg = (
        'Unable to tidy the following install targets because no'
        ' matching platform was found:\n * qux')
    assert qux_msg in records

    # We can get foo bar baz, and we try to remote tidy them.
    # (This will ultimately fail, but past the point we are testing).
    for target in ['foo', 'bar9', 'baz']:
        msg = f'platform: {target} - remote tidy (on {target})'
        assert msg in records

    # We haven't done anything with Quiz because we're only looking
    # at cases where platform == REMOTE_FILE_INSTALL_DONE
    assert not [r for r in records if 'quiz' in r]
```

tests/integration/test_task_remote_mgr.py
oliver-sanders: Looks ok to me TBH. How do you prevent remote-tidy from actually trying to run on the remote platforms?

cylc/flow/platforms.py
wxtim: That is a safer and more elegant thing in general, but we need to keep the `install_targets` in the `install_targets_map` to ensure that we can identify which install targets have been used but we cannot find platforms for.

```suggestion
    install_targets_map: Dict[str, List] = {
        target: [] for target in install_targets}
    for platform_name in platform_names:
        try:
            platform = get_platform(platform_name)
        except PlatformLookupError:
            # We only need one working platform per install target:
            continue
        install_targets_map.setdefault(
            platform.get('install target', platform['name']), []
        ).append(platform)
```

tests/integration/test_task_remote_mgr.py
wxtim: I don't. I just don't worry about it when they fail because that's after the chunk of code I'm trying to test

I considered making the hosts `uuids` to ensure that they remote clean would fail reliably. I got strange results when I had `foo` in my `~/.ssh/config`

cylc/flow/platforms.py
```diff
@@ -649,32 +650,52 @@ def is_platform_with_target_in_list(
...
+
+    # Look for unreachable install targets and report them:
+    unreachable_targets = {
+        target for target, platforms
+        in install_targets_map.items()
+        if not platforms
+    }
```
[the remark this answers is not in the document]
wxtim: Perhaps I don't need this function at all and can just use `get_install_target_to_platforms_map`.

tests/integration/test_task_remote_mgr.py
oliver-sanders: Maybe just monkeypatch it out?

cylc/flow/platforms.py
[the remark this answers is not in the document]
oliver-sanders: Makes sense.

tests/integration/test_task_remote_mgr.py
wxtim: Good idea, now done

cylc/flow/task_remote_mgr.py
```diff
@@ -286,44 +287,91 @@ def construct_remote_tidy_ssh_cmd(
...
+        unreachable_targets = {
+            t for t in install_targets if not install_targets_map.get(t)
+        }
+        if unreachable_targets:
+            msg = 'No platforms available to remote tidy install targets:'
+            for unreachable_target in unreachable_targets:
+                msg = (
+                    'Unable to tidy the following install targets'
+                    ' because no matching platform was found:'
+                )
```
oliver-sanders, line 322: Repeating this whole message for each install target is a bit wordy.

cylc/flow/task_remote_mgr.py
```diff
@@ -286,44 +287,91 @@ def construct_remote_tidy_ssh_cmd(
...
         """
+        # Get a list of all platforms used from workflow database:
+        platforms_used = (
+            self.db_mgr.get_pri_dao().select_platforms_used())
+        # For each install target compile a list of platforms:
+        install_targets = {
+            target for target, msg
+            in self.remote_init_map.items()
+            if msg == REMOTE_FILE_INSTALL_DONE
+        }
```
oliver-sanders, line 343: minor: Might make more sense to move this into _get_remote_tidy_targets?

cylc/flow/task_remote_mgr.py
```diff
@@ -286,44 +287,91 @@ def construct_remote_tidy_ssh_cmd(
...
+        install_targets_map = self._get_remote_tidy_targets(
+            platforms_used, install_targets)
+
         # Issue all SSH commands in parallel
         queue: Deque[RemoteTidyQueueTuple] = deque()
-        for install_target, message in self.remote_init_map.items():
-            if message != REMOTE_FILE_INSTALL_DONE:
-                continue
+        for install_target in install_targets_map.keys():
```
oliver-sanders, line 349: minor: Can avoid a lookup with this:

```suggestion
        for install_target, platforms in install_targets_map.items():
```

...

cylc/flow/task_remote_mgr.py
```diff
@@ -286,44 +287,91 @@ def construct_remote_tidy_ssh_cmd(
...
-                platform = get_random_platform_for_install_target(
-                    install_target
-                )
-                cmd, host = self.construct_remote_tidy_ssh_cmd(platform)
-            except (NoHostsError, PlatformLookupError) as exc:
-                LOG.warning(
-                    PlatformError(
-                        f'{PlatformError.MSG_TIDY}\n{exc}',
-                        platform['name'],
+            for platform in install_targets_map[install_target]:
```
oliver-sanders, line 352: ... and this:

```suggestion
            for platform in platforms:
```

oliver-sanders, line 352: Should really have an `else:` on this `for:` to log that `remote-tidy` failed completely.

oliver-sanders (commented):
LGTM, couple of minor points.

oliver-sanders:
Gotta small lint failure.

cylc/flow/task_remote_mgr.py
wxtim: By doing it this way I allow `_get_remote_tidy_targets` to be a staticmethod, else I must pass self to it. Not the end of the world, but I like it like this.

cylc/flow/task_remote_mgr.py
wxtim: Yes. I wonder what I can have been thingking

cylc/flow/task_remote_mgr.py
```diff
@@ -286,43 +288,90 @@ def construct_remote_tidy_ssh_cmd(
...
+                            Popen(  # nosec
+                                cmd, stdout=PIPE, stderr=PIPE, stdin=DEVNULL,
+                                text=True
+                            )  # * command constructed by internal interface
+                        )
                     )
+                    break
+            else:
+                LOG.error(
+                    NoPlatformsError(install_target, 'install target')
```
oliver-sanders, line 374: Could do with a bit of context in this error message?

oliver-sanders (approved):
Tested with this workflow:

```ini
[scheduling] 
    [[graph]]           
        R1 = x => y     
        
[runtime]               
    [[x]]               
        platform = myhost01       
        script = cylc stop "${CYLC_WORKFLOW_ID}" --now --now    
    [[y]] 
```

and this global config:
```ini
[platforms]
  [[myhost0.]]
```

Where "myhost01" exists on my network (obviously edit this to match an actual host).

Before:

```
INFO - Workflow shutting down - REQUEST(NOW-NOW)
WARNING - Orphaned tasks:
    * 1/x (running)
ERROR - Error during shutdown
ERROR - local variable 'platform' referenced before assignment
      File "~/cylc-flow/cylc/flow/scheduler.py", line 1756, in shutdown
        await self._shutdown(reason)
      File "~/cylc-flow/cylc/flow/scheduler.py", line 1834, in _shutdown
        self.task_job_mgr.task_remote_mgr.remote_tidy()
      File "~/cylc-flow/cylc/flow/task_remote_mgr.py", line 312, in remote_tidy
        platform['name'],
    UnboundLocalError: local variable 'platform' referenced before assignment
INFO - DONE
```

After:

```
INFO - Workflow shutting down - REQUEST(NOW-NOW)
WARNING - Orphaned tasks:
    * 1/x (running)
INFO - platform: exvcylcdev01 - remote tidy (on exvcylcdev01)
INFO - DONE
```

cylc/flow/exceptions.py
```diff
@@ -436,12 +436,20 @@ def __str__(self):
...
 
 class NoPlatformsError(CylcError):
-    """None of the platforms of a given group were reachable."""
-    def __init__(self, platform_group):
-        self.platform_group = platform_group
+    """None of the platforms of a given set were reachable.
+
+    Args:
+        identity: The name of the platform group or install target
+        _type: Whether the set of platforms is a platform group or an
```
MetRonnie, line 443:
```suggestion
        set_type: Whether the set of platforms is a platform group or an
```

cylc/flow/rundb.py
```diff
@@ -684,6 +684,14 @@ def select_task_job(self, cycle, name, submit_num=None):
...
             return None
 
+    def select_platforms_used(self):
+        """Get a list of platforms used by tasks."""
+        stmt = r'SELECT platform_name FROM task_jobs'
+        try:
+            return {r[0] for r in self.connect().execute(stmt)}
+        except sqlite3.DatabaseError:
+            return None
+
```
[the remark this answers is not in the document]
wxtim: It appears to be. I thought I'd looked for the functionality existing already, but I missed it. Thank you for catching!

tests/unit/test_platforms.py
```diff
@@ -450,14 +450,18 @@ def test_get_install_target_to_platforms_map(
...
     """Test that get_install_target_to_platforms_map works as expected."""
     monkeypatch.setattr('cylc.flow.platforms.platform_from_name',
                         lambda x: platform_from_name(x, PLATFORMS_TREK))
 
-    if expected_err:
+    if expected_err and not quiet:
         with pytest.raises(expected_err):
             get_install_target_to_platforms_map(platform_names)
+    elif expected_err and quiet:
+        result = get_install_target_to_platforms_map(platform_names, quiet)
```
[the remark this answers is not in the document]
wxtim: Nor should it be - what we are testing here is that `quiet=True` doesn't cause an error to be raised.

I've streamlined this and will push.

cylc/flow/task_remote_mgr.py
wxtim: The error does this 
```
>>> from cylc.flow import LOG
>>> from cylc.flow.exceptions import NoPlatformsError
>>> print(NoPlatformsError('x', 'y'))
Unable to find a platform from y x.
```

wxtim: I've modified this to include a context in the error message

tests/unit/test_task_remote_mgr.py
```diff
@@ -127,3 +127,111 @@ def test_get_log_file_name(tmp_path: Path,
...
+            None,
+            ['mountain_railway'],
+            '''
+            [platforms]
+                [[peter_sam, sir_handel]]
+                    install target = mountain_railway
+            ''',
+            {
+                'targets': {'mountain_railway': []},
+                'unreachable': 'mountain_railway'
```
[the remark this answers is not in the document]
wxtim: I made a mistake.

cylc/flow/task_remote_mgr.py
```diff
@@ -300,44 +301,93 @@ def construct_remote_tidy_ssh_cmd(
...
+        else:
+            install_targets_map = get_install_target_to_platforms_map(
+                platform_names, quiet=True)
+
+        # If we couldn't find a platform for a target, we cannot tidy it -
+        # raise an Error:
+        unreachable_targets = [
+            t for t, v in install_targets_map.items()
+            if not v
+        ] + list(install_targets.difference(install_targets_map))
```
MetRonnie, line 331: This seems more complicated than it need be
```suggestion
            install_targets_map: Dict[str, List[Dict[str, Any]]] = {
                t: [] for t in install_targets
            }
            unreachable_targets = install_targets
        else:
            install_targets_map = get_install_target_to_platforms_map(
                platform_names, quiet=True)
            unreachable_targets = install_targets.union(
                t for t, v in install_targets_map.items()
                if not v
            )
```
However I'm not sure the `unreachable_targets` in the second branch is correct - can I talk to you about this?

cylc/flow/task_remote_mgr.py
```diff
@@ -300,44 +301,93 @@ def construct_remote_tidy_ssh_cmd(
...
     def remote_tidy(self) -> None:
         """Remove workflow contact files and keys from initialised remotes.
 
         Call "cylc remote-tidy".
         This method is called on workflow shutdown, so we want nothing to hang.
         Timeout any incomplete commands after 10 seconds.
         """
+        # Get a list of all platforms used from workflow database:
+        platforms_used = (
+            self.db_mgr.get_pri_dao().select_task_job_platforms())
```
MetRonnie, line 349:
```suggestion
        with self.db_mgr.get_pri_dao() as dao:
            platforms_used = dao.select_task_job_platforms())
```

cylc/flow/task_remote_mgr.py
```diff
@@ -300,44 +301,93 @@ def construct_remote_tidy_ssh_cmd(
...
+        install targets.
+
+        logs:
+            A list of install targets where no platform can be found.
+
+        returns:
+            A mapping of install targets to valid platforms only where
+            platforms are available.
+        """
+        if not platform_names and install_targets:
```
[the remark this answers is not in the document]
hjoliver: Or:
```python
if install_targets and not platform_names:
```

cylc/flow/task_remote_mgr.py
wxtim: sure, not sure how much I can remember of this...

wxtim: In words - it's an unreachable target if:
- it's in the install targets list by not the install targets map
- it's in the install targets map and has no platforms.

cylc/flow/task_remote_mgr.py
```diff
@@ -316,19 +316,18 @@ def _get_remote_tidy_targets(
...
             install_targets_map: Dict[str, List[Dict[str, Any]]] = {
                 t: [] for t in install_targets}
+            unreachable_targets = install_targets
         else:
             install_targets_map = get_install_target_to_platforms_map(
                 platform_names, quiet=True)
+            # If we couldn't find a platform for a target, we cannot tidy it -
+            # raise an Error:
+            unreachable_targets = install_targets.difference(
+                install_targets_map)
```
MetRonnie: Ok, I'm happy with this as `get_install_target_to_platforms_map()` does not produce any empty values in the dictionary 👍

MetRonnie (approved):
Tested out using a regex pattern platform section in global.cylc

wxtim:
@oliver-sanders  - Ronnie's had me make some changes since last you looked.. mind giving it another quick look?

oliver-sanders:
@wxtim, some minor conflicts with the other branch.

oliver-sanders:
Gotta couple of test failures :(

oliver-sanders:
I've re-run the `f/rnd` test and it's failed again (so it's failed four times in a row due to retries) so it looks like it's genuine.

cylc/flow/exceptions.py
```diff
@@ -435,13 +435,30 @@ def __str__(self):
         return f'Unable to find valid host for {self.platform_name}'
 
 
-class NoPlatformsError(PlatformLookupError):
-    """None of the platforms of a given group were reachable."""
-    def __init__(self, platform_group):
-        self.platform_group = platform_group
+class NoPlatformsError(CylcError):
```
MetRonnie, line 438: This regression was the reason for some of the failing tests
```suggestion
class NoPlatformsError(PlatformLookupError):
```

oliver-sanders:
Merging with an unrelated test failure fixed in #5479

PR #5455, opened 2023-04-06
Improve documentation for retrieve job logs command

### LINKED ISSUE #5289: Use scheduler default permissions for log retrieval
The default for `retrieve job logs command` is currently `rsync -a`.
`-a` implies `-p` which means that the files get the same permissions as on the remote host:
https://explainshell.com/explain?cmd=rsync+-p
This can cause problems.
In my case, the default permissions on my scheduler host give world read access.
However, my remote host is setup to give no access to world.
As a result, when the log files are retrieved to the scheduler host they have no world access which is not what I expect on this host.
Also, the way the rsync is constructed means that the permissions of the `log/job` directory get changed to match the remote host so world access to the entire `job/log` directory gets disabled after the first log retrieval.
Our cylc review relies on world read access so it was very confused as to why I couldn't see any job logs (even for local jobs).

In order for the retrieved log files to get the default permissions used on the scheduler host you can use:
`retrieve job logs command = rsync -a --no-p --no-g --chmod=ugo=rwX`
Should we set this as the default?

dpmatthews:
Closes #5289

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -1399,6 +1399,17 @@ with Conf('global.cylc', desc='''
                  desc=f'''
                 {LOG_RETR_SETTINGS['retrieve job logs command']}
 
+                .. note::
+                   The default command (``rsync -a``) means that the retrieved
+                   files (and the directories above including ``job/log``) get
+                   the same permissions as on the remote host. This can cause
+                   problems if the remote host uses different permissions to the
+                   scheduler host (e.g. no world read access). To avoid this
+                   problem you can set the command to
+                   ``rsync -a --no-p --no-g --chmod=ugo=rwX`` which means the
+                   retrieved files get the default permissions used on the
+                   scheduler host.
+
                 .. versionchanged:: 8.0.0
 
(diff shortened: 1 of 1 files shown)


cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1399,6 +1399,17 @@ def default_for(
                  desc=f'''
                 {LOG_RETR_SETTINGS['retrieve job logs command']}
 
+                .. note::
+                   The default command (``rsync -a``) means that the retrieved
+                   files (and the directories above including ``job/log``) get
+                   the same permissions as on the remote host. This can cause
+                   problems if the remote host uses different permissions to the
+                   scheduler host (e.g. no world read access). To avoid this
+                   problem you can set the command to
```
oliver-sanders, line 1408: Exceeded max line length.

```suggestion
                   problems if the remote host uses different permissions to
                   the scheduler host (e.g. no world read access). To avoid
                   this problem you can set the command to
```

oliver-sanders (approved):
Advice matches `man rsync`, LGTM.

Can merge once the fast tests pass.

oliver-sanders:
Test failures all codecov.

PR #5458, opened 2023-04-11
Stop Cylc Reinstall modifying global variables

wxtim:
This caused the same variable to be added multiple times causing errors from clashing options.

Related to questions asked by @MetRonnie in https://github.com/cylc/cylc-rose/pull/219

**Check List**

- [x] Tests are not included because:
    - In the case of the change to `get_option_parser` we would largely be testing Python functionality.
    - In the case of moving ReinstallOptions we haven't done anything except move it to the test where it's used.
    - You can manually test before/after with `cylc reinstall --help`.
- [x] This is a bug-fix for a bug existing only on master

### DIFF
--- a/tests/integration/test_reinstall.py
+++ b/tests/integration/test_reinstall.py
@@ -28,14 +28,18 @@ from cylc.flow.exceptions import (
 from cylc.flow.install import (
     reinstall_workflow,
 )
+from cylc.flow.option_parsers import Options
 from cylc.flow.scripts.reinstall import (
-    ReInstallOptions,
+    get_option_parser as reinstall_gop,
     reinstall_cli,
 )
+from cylc.flow.terminal import cli_function
 from cylc.flow.workflow_files import (
     WorkflowFiles,
 )
 
+
+ReInstallOptions = Options(reinstall_gop())
 
 # cli opts
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,13 +17,15 @@ ones in. -->
 [#5291](https://github.com/cylc/cylc-flow/pull/5291) - re-implement old-style
 clock triggers as wall_clock xtriggers.
 
--------------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


tests/integration/test_reinstall.py
```diff
@@ -28,15 +28,19 @@
...
+    get_option_parser as reinstall_gop,
     reinstall_cli,
 )
+from cylc.flow.terminal import cli_function
 from cylc.flow.workflow_files import (
     WorkflowFiles,
 )
 
 
+ReInstallOptions = Options(reinstall_gop())
```
MetRonnie, line 42: Why move this here?

MetRonnie (commented):
Looks like it does the trick

tests/integration/test_reinstall.py
wxtim: 1. It's not used anywhere else.
2. The reason `options = REINSTALL_OPTS; options.extend(OTHER_OPTS)` was being run multiple times was the presence of this line. Let us not do it if it's not necessary.

oliver-sanders: (This object is only used by the tests so we don't need to initiate it as a global variable in the reinstall code)

MetRonnie: > 2\. The reason `options = REINSTALL_OPTS; options.extend(OTHER_OPTS)` was being run multiple times was the presence of this line. Let us not do it if it's not necessary.

The problem was not that `get_option_parser()` was being called multiple times, that should be able to happen without issue. The problem was `options = REINSTALL_OPTS; options.extend(OTHER_OPTS)` causing `REINSTALL_OPTS` to be mutated

MetRonnie (approved):

oliver-sanders (approved):

tests/integration/test_reinstall.py
wxtim: > that should be able to happen without issue

Yes, but also see point 1

PR #5466, opened 2023-04-14
Avoid duplicate prerequisites from multiple recurrences.

### LINKED ISSUE #5465: Task prerequisites duplication
### Description

This was something I noticed in Cylc7, and just checked if it was still the case in Cylc 8.1.0. It is.

When you do certain graph constructs, task prerequisites can be duplicated. It's not optimal to look at, but I'm going to guess, its not optimal for Cylc internals either. Are the prerequisites stored as a list instead of a set? The off-topic discussion in [this threaad](https://cylc.discourse.group/t/cylc-reinstall-auto-yes/610/18) made me think to test this item.

### Reproducible Example

Here is a contrived example

```
[scheduler]
   allow implicit tasks=True

[task parameters]
  run = 1..2

[scheduling]
  initial cycle point = 20200101
  final cycle point = 20200103
  [[ graph]]
    P1D = """
      a<run> => c
      b[-P1D] => b
      b => c
    """
    PT12H = """
      a<run> => c
      b[-P1D] => b
      b => c
    """

[runtime]
  [[root]]
    script = echo hello
```

Launching that and doing `cylc show workflow//20200102T0000Z/c`, we get this list of prerequisites.

```
prerequisites: ('-': not satisfied)
  - 20200102T0000Z/a_run2 succeeded
  - 20200102T0000Z/b succeeded
  - 20200102T0000Z/a_run1 succeeded
  - 20200102T0000Z/a_run2 succeeded
  - 20200102T0000Z/b succeeded
  - 20200102T0000Z/a_run1 succeeded
```

So there are 6 listed, but really it is only 3, its just been duplicated.

Similarly, if I do `cylc show workflow//20200102T0000Z/b` we see two prerequisites instead of one.

```
prerequisites: ('-': not satisfied)
  - 20200101T0000Z/b succeeded
  - 20200101T0000Z/b succeeded
```

### Expected Behaviour

As described above, no duplication in task prerequisites.

hjoliver:
- Avoid duplicate prerequisites due to other recurrences with coincident points.

Close #5465

**Check List**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,9 @@ configuration files.
 [#5334](https://github.com/cylc/cylc-flow/pull/5334) - Apply graph prerequisite
 changes to already-spawned tasks after reload or restart.
 
+[5466](https://github.com/cylc/cylc-flow/pull/5466) - Don't generate duplicate
+prerequisites from recurrences with coincident points.
+
 [5450](https://github.com/cylc/cylc-flow/pull/5450) - Validation provides
 better error messages if [sections] and settings are mixed up in a
 configuration.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,10 @@ ones in. -->
 
 ### Enhancements
 
+[#5475](https://github.com/cylc/cylc-flow/pull/5475) - much faster computation
+of the visualization window around active tasks (at the cost, for now, of not
+showing non-active "cousin" nodes).
+
 [#5453](https://github.com/cylc/cylc-flow/pull/5453) - `cylc cat-log` can now
 list and view workflow log files including install logs and workflow
 configuration files.
--- a/cylc/flow/prerequisite.py
+++ b/cylc/flow/prerequisite.py
@@ -42,7 +42,7 @@ class Prerequisite:
 
     # Memory optimization - constrain possible attributes to this list.
     __slots__ = ["satisfied", "_all_satisfied",
-                 "target_point_strings", "start_point",
+                 "target_point_strings",
                  "conditional_expression", "point"]
 
     # Extracts T from "foo.T succeeded" etc.
@@ -53,14 +53,10 @@ class Prerequisite:
     DEP_STATE_OVERRIDDEN = 'force satisfied'
     DEP_STATE_UNSATISFIED = False
 
-    def __init__(self, point, start_point=None):
+    def __init__(self, point):
         # The cycle point to which this prerequisite belongs.
         # cylc.flow.cycling.PointBase
         self.point = point
-
-        # Start point for prerequisite validity.
-        # cylc.flow.cycling.PointBase
-        self.start_point = start_point
 
         # List of cycle point strings that this prerequisite depends on.
         self.target_point_strings = []
@@ -83,7 +79,6 @@ class Prerequisite:
         """Test equality of prerequisites. Ignore derived attributes."""
         return (
             self.point == other.point
-            and self.start_point == other.start_point
             and self.conditional_expression == other.conditional_expression
             and self.satisfied == other.satisfied
         )
--- a/cylc/flow/prerequisite.py
+++ b/cylc/flow/prerequisite.py
@@ -75,10 +75,6 @@ class Prerequisite:
         # * `False` (prerequisite unsatisfied).
         self._all_satisfied = None
 
-    def __eq__(self, other):
-        """Test equality of prerequisites. Ignore derived attributes."""
-        return self.hash() == other.hash()
-
     def instantaneous_hash(self):
         """Generate a hash of this prerequisite in its current state.
 
--- a/cylc/flow/prerequisite.py
+++ b/cylc/flow/prerequisite.py
@@ -79,6 +79,15 @@ class Prerequisite:
         # * `False` (prerequisite unsatisfied).
         self._all_satisfied = None
 
+    def __eq__(self, other):
+        """Test equality of prerequisites. Ignore derived attributes."""
+        return (
+            self.point == other.point
+            and self.start_point == other.start_point
+            and self.conditional_expression == other.conditional_expression
+            and self.satisfied == other.satisfied
+        )
+
     def add(self, name, point, output, pre_initial=False):
         """Register an output with this prerequisite.
 
@@ -190,7 +199,7 @@ class Prerequisite:
         return res
 
     def satisfy_me(self, all_task_outputs):
-        """Evaluate pre-requisite against known outputs.
+        """Evaluate prerequisite against known outputs.
 
         Updates cache with the evaluation result.
 
--- a/cylc/flow/prerequisite.py
+++ b/cylc/flow/prerequisite.py
@@ -77,11 +77,19 @@ class Prerequisite:
 
     def __eq__(self, other):
         """Test equality of prerequisites. Ignore derived attributes."""
-        return (
-            self.point == other.point
-            and self.conditional_expression == other.conditional_expression
-            and self.satisfied == other.satisfied
-        )
+        return self.hash() == other.hash()
+
+    def instantaneous_hash(self):
+        """Generate a hash of this prerequisite in its current state.
+
+        Note not using `__hash__` because Prerequisite objects are mutable.
+        """
+        return hash((
+            self.point,
+            self.conditional_expression,
+            tuple(self.satisfied.keys()),
+            tuple(self.satisfied.values()),
+        ))
 
     def add(self, name, point, output, pre_initial=False):
         """Register an output with this prerequisite.
--- a/cylc/flow/prerequisite.py
+++ b/cylc/flow/prerequisite.py
@@ -84,7 +84,6 @@ class Prerequisite:
             self.point,
             self.conditional_expression,
             tuple(self.satisfied.keys()),
-            tuple(self.satisfied.values()),
         ))
 
     def add(self, name, point, output, pre_initial=False):
--- a/cylc/flow/task_state.py
+++ b/cylc/flow/task_state.py
@@ -468,17 +468,20 @@ class TaskState:
         self._is_satisfied = None
         self._suicide_is_satisfied = None
 
+        # Use dicts to avoid generating duplicate prerequisites from sequences
+        # with coincident cycle points.
+        prerequisites = {}
+        suicide_prerequisites = {}
+
         for sequence, dependencies in tdef.dependencies.items():
             if not sequence.is_valid(point):
                 continue
             for dependency in dependencies:
                 cpre = dependency.get_prerequisite(point, tdef)
                 if dependency.suicide:
-                    if cpre not in self.suicide_prerequisites:
-                        self.suicide_prerequisites.append(cpre)
+                    suicide_prerequisites[cpre.instantaneous_hash()] = cpre
                 else:
-                    if cpre not in self.prerequisites:
-                        self.prerequisites.append(cpre)
+                    prerequisites[cpre.instantaneous_hash()] = cpre
 
         if tdef.sequential:
             # Add a previous-instance succeeded prerequisite.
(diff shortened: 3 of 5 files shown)


hjoliver:
TODO: initial quick fix with explicit test for presence in list ... might be more efficient to convert to a set, if possible

hjoliver:
A simpler version of @ColemanTom 's example on #5465 : 

```ini
[scheduling]
   initial cycle point = 20200101
   [[ graph]]
      R1 = "foo => bar"
      R2/^/PT12H = "foo => bar"
[runtime]
  [[foo]]
      script = false  # stall the workflow
  [[bar]]
```

Run this until stalled, then check the prerequisites of `20200101T0000Z/bar`:

On master:
```console
$ cylc show cot//20200101T0000Z/bar
prerequisites: ('-': not satisfied)
  - 20200101T0000Z/foo succeeded
  - 20200101T0000Z/foo succeeded
```

On this branch:
```console
$ cylc show cot//20200101T0000Z/bar
prerequisites: ('-': not satisfied)
  - 20200101T0000Z/foo succeeded
```

cylc/flow/prerequisite.py
```diff
@@ -79,6 +79,25 @@ def __init__(self, point, start_point=None):
         # * `False` (prerequisite unsatisfied).
         self._all_satisfied = None
 
+    def __str__(self):
```
oliver-sanders, line 82: This will help with prerequisite sorting.

There are two places in the data store where we iterate over prereqs:

```python
for prereq in itask.state.prerequisites:
```

With this __str__ method, if we did this:

```python
for prereq in sorted(itask.state.prerequisites, key=str):
```

They should sort ok. Be aware we build this by default, not just on request though so efficiency will matter.

hjoliver:
(NB Tom C requested prereq sorting too).

cylc/flow/prerequisite.py
hjoliver: I've gone for sorting once at task proxy state initialization. The order is preserved by the data store.

hjoliver:
(Still DRAFT because prerequisite sorting is somewhat complicated when conditional expressions are involved).

cylc/flow/prerequisite.py
hjoliver: This is now off-topic for this branch. I'll do a separate follow-up for sorting.

CHANGES.md
```diff
@@ -15,12 +15,19 @@ ones in. -->
 
 ### Enhancements
 
+[#5475](https://github.com/cylc/cylc-flow/pull/5475) - much faster computation
+of the visualization window around active tasks (at the cost, for now, of not
+showing non-active "cousin" nodes).
```
hjoliver, line 20: ??? What the hell. I don't know why this is appearing here. (No time to investigate now).

hjoliver:
@oliver-sanders - I've backed out of prerequisite and output sorting changes on this branch. Now it's just a simple change to avoid duplicates. Sorting will be a follow-up.

cylc/flow/prerequisite.py
```diff
@@ -42,7 +42,7 @@ class Prerequisite:
 
     # Memory optimization - constrain possible attributes to this list.
     __slots__ = ["satisfied", "_all_satisfied",
-                 "target_point_strings", "start_point",
+                 "target_point_strings",
```
oliver-sanders, line 45: Good spot, this appears to have been unused. Potentially a reasonable efficiency enhancement.

oliver-sanders (changes requested):
LGTM, however, the method should be tweaked for efficiency.

oliver-sanders:
Because `Prereusite` objects are mutable (i.e. non-hashable) the best I could come up with is this:

```diff
diff --git a/cylc/flow/prerequisite.py b/cylc/flow/prerequisite.py
index af95ba6f1..81c86565e 100644
--- a/cylc/flow/prerequisite.py
+++ b/cylc/flow/prerequisite.py
@@ -77,11 +77,19 @@ class Prerequisite:
 
     def __eq__(self, other):
         """Test equality of prerequisites. Ignore derived attributes."""
-        return (
-            self.point == other.point
-            and self.conditional_expression == other.conditional_expression
-            and self.satisfied == other.satisfied
-        )
+        return self.hash() == other.hash()
+
+    def hash(self):
+        """Generate a hash of this prerequisite in its current state.
+
+        Note `hash` not `__hash__` because Prerequisite objects are mutable.
+        """
+        return hash((
+            self.point,
+            self.conditional_expression,
+            tuple(self.satisfied.keys()),
+            tuple(self.satisfied.values()),
+        ))
 
     def add(self, name, point, output, pre_initial=False):
         """Register an output with this prerequisite.
diff --git a/cylc/flow/task_state.py b/cylc/flow/task_state.py
index b05d9537a..e870b7720 100644
--- a/cylc/flow/task_state.py
+++ b/cylc/flow/task_state.py
@@ -468,17 +468,17 @@ class TaskState:
         self._is_satisfied = None
         self._suicide_is_satisfied = None
 
+        prerequisites = {}
+        suicide_prerequisites = {}
         for sequence, dependencies in tdef.dependencies.items():
             if not sequence.is_valid(point):
                 continue
             for dependency in dependencies:
                 cpre = dependency.get_prerequisite(point, tdef)
                 if dependency.suicide:
-                    if cpre not in self.suicide_prerequisites:
-                        self.suicide_prerequisites.append(cpre)
+                    suicide_prerequisites[cpre.hash()] = cpre
                 else:
-                    if cpre not in self.prerequisites:
-                        self.prerequisites.append(cpre)
+                    prerequisites[cpre.hash()] = cpre
 
         if tdef.sequential:
             # Add a previous-instance succeeded prerequisite.
@@ -494,8 +494,10 @@ class TaskState:
                 cpre.add(tdef.name, p_prev, TASK_STATUS_SUCCEEDED,
                          p_prev < tdef.start_point)
                 cpre.set_condition(tdef.name)
-                if cpre not in self.prerequisites:
-                    self.prerequisites.append(cpre)
+                prerequisites[cpre.hash()] = cpre
+
+        self.suicide_prerequisites = list(suicide_prerequisites.values())
+        self.prerequisites = list(prerequisites.values())
 
     def add_xtrigger(self, label, satisfied=False):
         self.xtriggers[label] = satisfied
```

hjoliver:
> Because Prereusite objects are mutable (i.e. non-hashable) the best I could come up with is this:

Thanks for that, I'll take a look tomorrow.

I actually had something similar before backing out of the sorting changes (you might have seen it pre force-push) but I wasn't sure of myself on the hash front (I tried to define `__hash__()` and use a set of prerequisites).

oliver-sanders:
Ah right. The `__hash__` method is supposed to be reserved for non-mutable objects but that doesn't matter here as the set will be destroyed moments after it's created.

CHANGES.md
MetRonnie: Added in c17529a282bae4e88c4373c35516c09f826c95a8. IMHO it's a good addition as #5475 was missing a changelog entry

hjoliver:
> Ah right. The __hash__ method is supposed to be reserved for non-mutable objects

Yes, that's the bit I wasn't sure about.

CHANGES.md
hjoliver: Hmm, I thought I had pushed that as a hot fix to the 8.1.x branch. Must have got my wires crossed somewhere.

cylc/flow/prerequisite.py
hjoliver: Yes, I checked the spec etc.

cylc/flow/task_state.py
```diff
@@ -474,9 +474,11 @@ def _add_prerequisites(self, point, tdef):
             for dependency in dependencies:
                 cpre = dependency.get_prerequisite(point, tdef)
                 if dependency.suicide:
-                    self.suicide_prerequisites.append(cpre)
+                    if cpre not in self.suicide_prerequisites:
```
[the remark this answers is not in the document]
hjoliver: Cool to see the numbers. 

As noted below I had used the set method but ditched it when I backed out of the sorting changes, because I was unsure if it was legit to define `__hash__()` for the mutable Prerequisites type.

hjoliver:
Bumping back to 8.1.4 to avoid blocking the 8.1.3 release tomorrow...

hjoliver:
Actually, I just looked more closely at your suggested solution:

```
                prerequisites[cpre.hash()] = cpre
```

That's clever!  Avoids use of `__hash__`.

CHANGES.md
hjoliver: A-ha, I had added the commit to my local 8.1.x branch but forgot to push it to GitHub. Then I rebased this branch on to 8.1.x locally and pushed this branch to GitHub.  That explains it 😌

hjoliver:
@oliver-sanders - I've put in your suggested fix, because it works and it seems more legit than a somewhat dodgy use of `__hash__` and `set`.  I haven't had time to actually compare performance with a set though.

hjoliver:
Functional test fail unrelated ? 

I'm seeing it locally on both 8.1.x and master.  Weird.

```
Test Summary Report
-------------------
tests/f/platforms/10-platform-gone-on-restart.t          (Wstat: 0 Tests: 3 Failed: 2)
  Failed tests:  2-3
```

cylc/flow/prerequisite.py
```diff
@@ -79,6 +75,22 @@ def __init__(self, point, start_point=None):
         # * `False` (prerequisite unsatisfied).
         self._all_satisfied = None
 
+    def __eq__(self, other):
+        """Test equality of prerequisites. Ignore derived attributes."""
+        return self.hash() == other.hash()
+
```
MetRonnie, line 81: There's a problem with this, but it isn't actually used so might as well remove?
```suggestion
```

cylc/flow/prerequisite.py
```diff
@@ -79,6 +75,22 @@ def __init__(self, point, start_point=None):
...
+    def instantaneous_hash(self):
+        """Generate a hash of this prerequisite in its current state.
+
+        Note not using `__hash__` because Prerequisite objects are mutable.
+        """
+        return hash((
+            self.point,
+            self.conditional_expression,
+            tuple(self.satisfied.keys()),
+            tuple(self.satisfied.values()),
```
MetRonnie, line 91: Is `self.satisfied` actually relevant to `TaskState._add_prerequisites()`? And is `self.target_point_strings` not relevant?

cylc/flow/prerequisite.py
oliver-sanders: What's the problem?

MetRonnie: Should be `.instantaneous_hash()` and also it doesn't account for testing equality against non-`Prerequisite` objects (although this is unlikely to happen)

hjoliver: You're right that it isn't used now (it was when using the list-based duplicate checking) so it could be removed.

Happy for you and Oliver to take over the branch and tweak as needed to get it in!

cylc/flow/prerequisite.py
hjoliver: `self.satisfied` is poorly named, but it's a dict of all prerequisites, so it's really the most important thing (the boolean satisfied/not-satisfied bit is in there, but incidental to this PR).

`self.target_point_strings` just repeats info held in `self.satisfied` so it's not needed in the hash.

oliver-sanders: So I think we only need `satisfied.keys()`.

cylc/flow/prerequisite.py
hjoliver: (oops, I accidentally hit commit on that - but it needed to be done anyway)

cylc/flow/prerequisite.py
hjoliver: Yes actually.

hjoliver: (Signing off for now - over to you guys 👋 )

oliver-sanders: Confirmed that `target_point_strings` is the same as `{message[0] for message in self.satisfied}`.

We don't need to store both so remove `target_point_strings` here: https://github.com/cylc/cylc-flow/pull/5496

oliver-sanders:
```suggestion
            tuple(self.satisfied.keys()),
```

oliver-sanders (approved):
Confirmed this fixes the duplicate prereqs reported in `cylc show` output.

Need to do a quick profiling check to make sure the scaling is good.

MetRonnie (approved):

PR #5475, opened 2023-04-20
data-store graph walk restricted

### LINKED ISSUE #5435: efficiency: increment_graph_window

dwsutherland:
closes #5435

**Check List**

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -725,6 +725,7 @@ class DataStoreMgr:
             None
 
         """
+        is_active = not (descendant or is_parent)
         # ID passed through recursion as reference to original/active node.
         if active_id is None:
             source_tokens = self.id_.duplicate(source_tokens)
@@ -738,7 +739,7 @@ class DataStoreMgr:
 
         # Setup and check if active node is another's boundary node
         # to flag its paths for pruning.
-        if edge_distance == 0:
+        if is_active:
             self.n_window_edges[active_id] = set()
             self.n_window_boundary_nodes[active_id] = {}
             self.n_window_nodes[active_id] = set()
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Have repeated the reported result, this reduces the number of calls of `increment_graph_window` from `49021002/7001` to `14002/7001` for the example in #5435 :tada:. Doubtless the single biggest complexity reduction for cylc-flow in a single change ever! There are 7001 tasks in the workflow causing the 7001 original calls, each makes two recursive calls giving the 14002 which makes sense.

The revised algorithm seems to be producing the same results, these screenshots come from the same workflow, installed & run under `upstream/8.1.x` and `dwsutherland/graph-walk-efficiency` with the same operations performed on both. The graph view displays the same results:

Screenshot from 2023-04-20 11-49-25
Screenshot from 2023-04-20 11-49-38

dwsutherland:
The only down side is; increasing the window size will no longer flesh out the entire workflow graph, it will only show the the direct ancestors/descendants of active nodes.

But obviously, no down sides with `n=1` .. And all the tests passed, so nothing missing there.

If we're happy with this trade-off, perhaps we should get this one in and work on the no-rewalk change separately.

oliver-sanders:
Ah, right, this has removed the ability to generate the n>=1 window?

oliver-sanders:
From a bit of experimentation with this pattern:

```
a => b<N> => c
```

The number of times `increment_graph_window` is called (recursive) scales according to:

| Before | After
| --- | --- |
| **2N^2** + 3N + 1 | 5N + 1 |

This change removes the N^2 term which is the nasty one.

The factor of 5 in the new solution is a little odd, not sure where that's coming from. Would be great to reduce it if possible, but linear scaling is the important bit for now.

The lack of cousins in the graph window isn't great, however, this doesn't affect the N=1 window (default) and we don't currently make it easy for users to switch the n-window in the GUI so this shouldn't have any impact right now.

oliver-sanders (approved):
I think the performance is well worth the caveats, we can always improve later, thanks @dwsutherland!

oliver-sanders:
(can drop in a changelog later)

hjoliver:
> we don't currently make it easy for users to switch the n-window in the GUI so this shouldn't have any impact right now.

Ha, note https://github.com/cylc/cylc-ui/issues/1281

hjoliver:
>  I think the performance is well worth the caveats, we can always improve later, thanks @dwsutherland!

Agreed, fantastic.

> The only down side is; increasing the window size will no longer flesh out the entire workflow graph, it will only show the the direct ancestors/descendants of active nodes.

This may be OK for the n-window.  For n=cycle (or cycle-window?) we shouldn't need to walk out from active tasks.

hjoliver:
On my laptop, it took 15s to spawn 7000 tasks on this branch vs 400s on master :tada:

hjoliver (approved):
Great, merging this now to get it in 8.1.3

@dwsutherland - as a small follow-up could you just add some code comments to explain the current functionality and the "cousins problem"?

hjoliver:
(I'll push a hot-fix change log entry for this)

PR #5483, opened 2023-04-24
task_job_mgr: move compute out of open call

oliver-sanders:
Spotted whilst reviewing #5475, writing to the activity log takes ~0.035 seconds per task which is ~8% of the `main_loop` runtime.

This PR reduces that to ~0.025s (~28% saving), not entirely sure why this is faster, but, meh.

I profiled this with:

```
# b = 1..100
a => <b> => c
```

Although you don't need 100 tasks and you could just time the method call.

**Before:**

Screenshot from 2023-04-24 12-43-38

**After:**

Screenshot from 2023-04-24 12-43-52

**Check List**


hjoliver (approved):
Somewhat surprising that this makes a difference, but it's nicer this way anyway.

PR #5486, opened 2023-04-24
platforms: use cached global config

oliver-sanders:
Use the cached global config for platform selection.

Spotted whilst profiling the "complex" workflow as `get_platform` accounted for ~19% of the main_loop runtime.

We are currently reloading the global config for platform selection which wasn't the plan. I'm not sure why we're using the cached config, the change was made a while back in a commit with "fixed caching problem" in the description:

https://github.com/cylc/cylc-flow/commit/91be2ffb1c61e616c09fa01d177be325b0d272fd#diff-f2c9506863bbbf888bfe92fe41bcd6f87f64f6a7818180cc3b7ace81848ba6b1R449

But that doesn't make much sense to me? The tests seem to agree.

**Check List**

### DIFF
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -166,7 +166,7 @@ def get_platform(
                 task_conf['remote'] if 'remote' in task_conf else {})
             return platform_from_name(
                 platform_name_from_job_info(
-                    glbl_cfg(cached=False).get(['platforms']),
+                    glbl_cfg().get(['platforms']),
                     task_job_section,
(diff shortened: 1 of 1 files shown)


wxtim (approved):
- [x] Unable to discover/remember why it was ever like this
- [x] Tests are passing
- [x] I cannot think of any reason why this is thus.

I think the test failures are unrelated, but I've refrained from hitting merge because of them.

hjoliver (approved):
Good spot :tada:

PR #5493, opened 2023-04-26
GH Actions: auto sync 8.1.x -> master

MetRonnie:
Add workflow for automatically creating sync PRs on push to `8.*.x` branches (or by `workflow_dispatch`). This will not create a PR if one is already open.

**Check List**

- [x] No dependency changes
- [x] Tests N/A
- [x] Changelog N/A
- [x] Docs N/A
- [x] This GH Actions workflow only needs to be on master

### DIFF
--- a/.github/workflows/branch_sync.yml
+++ b/.github/workflows/branch_sync.yml
@@ -1,39 +1,37 @@
 name: Sync PR
 
 on:
-  schedule:
-    - cron: '33 15 * * 1-5'  # 15:33 UTC Mon-Fri
+  push:
+    branches:
+      - '8.*.x'
   workflow_dispatch:
     inputs:
       branch:
         description: Branch to merge into master
-        required: false
-  # push:
-  #   branches:
-  #     - master
-  #     - '8.*.x'
+        required: true
 
 jobs:
   sync:
     runs-on: ubuntu-latest
     timeout-minutes: 5
     env:
-      BRANCH: ${{ inputs.branch || '8.1.x' }}
+      BRANCH: ${{ inputs.branch || github.ref_name }}
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
-
-      - name: Configure git
-        uses: cylc/release-actions/configure-git@v1
-
       - name: Check branch name
+        shell: python
         run: |
-          git show-branch "origin/${BRANCH}"
+          import os
+          import sys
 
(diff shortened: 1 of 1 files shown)


oliver-sanders:
We are going to want to use this on most of our repositories so could we locate this in cylc/release-actions?

.github/workflows/branch_sync.yml
```diff
@@ -0,0 +1,103 @@
...
+    - cron: '33 15 * * 1-5'  # 15:33 UTC Mon-Fri
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: Branch to merge into master
+        required: false
+  # push:
+  #   branches:
+  #     - master
+  #     - '8.*.x'
```
oliver-sanders, line 14:
```suggestion
  push:
    branches:
      - master
  #     - '8.*.x'
```

.github/workflows/branch_sync.yml
```diff
@@ -0,0 +1,103 @@
+name: Sync PR
+
+on:
+  schedule:
+    - cron: '33 15 * * 1-5'  # 15:33 UTC Mon-Fri
```
oliver-sanders, line 5: If running on push removing cron would avoid the need to hardcode the 8.*.x branch name in the workflow and update it with each minor release.

```suggestion
```

.github/workflows/branch_sync.yml
```diff
@@ -0,0 +1,103 @@
...
+  #   branches:
+  #     - master
+  #     - '8.*.x'
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    env:
+      BRANCH: ${{ inputs.branch || '8.1.x' }}
```
oliver-sanders, line 21: Otherwise we could potentially retrieve the 8.*.x branch name(s) from https://github.com/cylc/cylc-admin/blob/master/docs/status/branches.json

MetRonnie:
> We are going to want to use this on most of our repositories so could we locate this in cylc/release-actions?

We can switch over to doing this after a while. For now I think it's easier to start with cylc-flow before figuring out hot to use reusable workflows

.github/workflows/branch_sync.yml
MetRonnie: Done (just without master as that's not needed)

oliver-sanders:
Ok, fair enough.

oliver-sanders (approved):
Lets give it a go, there's currently need for a sync PR so this can go into action right away.

(leaving you to merge when convenient)

PR #5494, opened 2023-04-26
swarm: fix docker builds

oliver-sanders:
> This fixes the remote CI tests, it's not strictly necessary for 8.1.3 so don't let it get in the way of release.

We were using the minconda Docker image which is now dead it seams.

Switch to the superior and less deceased Mamba image.

This also switches up the build process:

* Before we started with a Conda image, created the Cylc environment and committed it as a layer on that image.
* Now we start with a Mamba image, create the Cylc environment, then conda-pack it onto a base image of our choice.

This means we have free choice of base image and we don't have to have Mamba installed on the final image which slims it down a little (no more base env).

**Check List**

### DIFF
--- a/dockerfiles/cylc-dev/Dockerfile
+++ b/dockerfiles/cylc-dev/Dockerfile
@@ -1,51 +1,49 @@
-# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
-# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
-#
-# This program is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


dockerfiles/cylc-dev/Dockerfile
```diff
@@ -1,51 +1,49 @@
...
     echo "export TMPDIR=/tmp" >> $HOME/.bashrc
 
-# run setup logic
-RUN . $HOME/.bashrc && \
-    pip install -e ./cylc[all] && \
-    cylc version
+# unpack the Mamba environment, then install Cylc into it from source
+RUN \
+    /venv/bin/conda-unpack && \
+    pip install --no-deps -e /cylc && \
```
oliver-sanders, line 41:
```suggestion
    pip install --no-deps -e /cylc coverage && \
```

hjoliver (approved):
LGTM (but @MetRonnie will know better).

MetRonnie (approved):
LGTM (not sure I know any better than @hjoliver)

But tests now run much quicker and seem not to be flaky anymore 🎉

PR #5495, opened 2023-04-26
Ignore off-sequence parents, for datastore.

### LINKED ISSUE #5490: generate_graph_parents: tasks appearing at the wrong cycle point
Reported by two users.

### Description

Under some circumstances tasks which should only be carried out once, at the initial cycle point appear in the GUI at Cycle point other than the ICP.

In both cases the following pattern of dependencies has been seen - 

```
R1 = run_once_task
TX = run_once_task[-PTX] => task_for_this_cycle
```

### Reproducible Example

```ini
[scheduler]
    allow implicit tasks = True
[scheduling]
    initial cycle point = 2023
    final cycle point = +PT9H
    [[graph]]
        R1 = run_once_at_midnight
        T00 = run_once_at_midnight[-PT0H] => every_cycle
        T03 = run_once_at_midnight[-PT3H] => every_cycle
        T06 = run_once_at_midnight[-PT6H] => every_cycle
[runtime]
    [[every_cycle]]
        script = sleep 60  # Added to stop the bug being transitory.

```

In the GUI we get the following 

image

(and the equivalent wrong dependency in the graph view)

### Expected Behaviour

The graph should look thus:

image

### Additional information

In the example above looking at the database and log files strongly suggests that the workflow is behaving as intended:

```sqlite
sqlite> SELECT cycle, name FROM task_jobs;
20230101T0000Z|every_cycle
20230101T0000Z|run_once_at_midnight
20230101T0300Z|every_cycle
20230101T0600Z|every_cycle
```

```bash
> ls ~/cylc-run/example/runN/log/job/20230101T0*/
~/cylc-run/example/runN/log/job/20230101T0000Z/:
every_cycle  run_once_at_midnight
~/cylc-run/example/runN/log/job/20230101T0300Z/:
every_cycle
~/cylc-run/example/runN/log/job/20230101T0600Z/:
every_cycle
```

hjoliver:
Close #5490 

We weren't correctly ignoring parents from triggers belonging to recurrences that are not relevant to the current child point.

**Check List**


cylc/flow/taskdef.py
```diff
@@ -85,11 +92,7 @@ def generate_graph_parents(tdef, point):
             if is_abs and parent_point != point:
                 # If 'foo[^] => bar' only spawn off of '^'.
                 continue
-            if seq.is_valid(parent_point):
```
hjoliver, line 88: The bug was use `parent_point` here instead of child `point`.  So can move the test up higher, should be a small efficiency improvement.

hjoliver:
The bug was hiding another bug!

Main bug fix:
- infer parents only from triggers belonging to recurrences that include the **child point**

Secondary bug fix:
- for inter-cycle triggers check the validity of inferred parents with respect to their own recurrences. We currently ignore invalid parents, which include those arising from pre-initial triggers. 

Note we don't yet detect and fail erroneous offset triggers (i.e. where the inferred parent point is not on a valid point for the parent task).

hjoliver:
Targeting 8.1.4 as this turned out to be less trivial than I initially thought. But feel free to bump it forward if happy with the fix.

cylc/flow/taskdef.py
```diff
@@ -73,26 +73,42 @@ def generate_graph_children(tdef, point):
...
-                # E.g.: foo should trigger only on T06:
-                #   PT6H = "waz"
-                #   T06 = "waz[-PT6H] => foo"
-                graph_parents[seq].append((name, parent_point, is_abs))
+            graph_parents[seq].append((parent_name, parent_point, is_abs))
 
     if tdef.sequential:
-        # Add prev-instance parent.
+        # Add implicit prevous-instance parent.
```
hjoliver, line 111:
```suggestion
        # Add implicit previous-instance parent.
```

(Add with skip-ci once tests all passing)

tests/unit/test_taskdef.py
```diff
@@ -0,0 +1,100 @@
...
+                    ISO8601Point('20230101T0000Z'),
+                    False
+                )
+            ]
+        ]
+
+
+def test_generate_graph_parents_2(tmp_flow_config):
+    """Test inferred parents are valid w.r.t to their own recurrences."""
+    reg = 'gargle-blaster'
```
oliver-sanders, line 68: ...wrapped around a large gold brick.

tests/unit/test_taskdef.py
```diff
@@ -0,0 +1,100 @@
...
+from cylc.flow.taskdef import generate_graph_parents
+from cylc.flow.cycling.iso8601 import ISO8601Point
+from cylc.flow.cycling.integer import IntegerPoint
+
+from .test_config import tmp_flow_config
+
+
+def test_generate_graph_parents_1(tmp_flow_config):
+    """Test that parents are only generated from valid recurrences."""
+    reg = 'pan-galactic'
```
oliver-sanders, line 27: The effect of which is like having your brain slapped by a lemon...

tests/unit/test_taskdef.py
```diff
@@ -0,0 +1,100 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from cylc.flow.config import WorkflowConfig
+from cylc.flow.taskdef import generate_graph_parents
+from cylc.flow.cycling.iso8601 import ISO8601Point
+from cylc.flow.cycling.integer import IntegerPoint
+
+from .test_config import tmp_flow_config
```
oliver-sanders, line 22:
```suggestion
```

oliver-sanders (approved):

tests/unit/test_taskdef.py
oliver-sanders: Ok, this fixture hasn't been properly registered for some reason, putting the import back, will sort out post-release.

```suggestion

from .test_config import tmp_flow_config
```

cylc/flow/taskdef.py
```diff
@@ -73,26 +73,44 @@ def generate_graph_children(tdef, point):
     return graph_children
 
 
-def generate_graph_parents(tdef, point):
-    """Determine graph parents of this task."""
+def generate_graph_parents(tdef, point, taskdefs):
+    """Determine concrent graph parents of task tdef at point.
```
MetRonnie, line 77: "concrent"?

cylc/flow/taskdef.py
```diff
@@ -73,26 +73,44 @@ def generate_graph_children(tdef, point):
...
+    Infer parents be reversing upstream triggers that lead to point/task.
+    """
     graph_parents = {}
-    for seq, ups in tdef.graph_parents.items():
+    for seq, triggers in tdef.graph_parents.items():
+        if not seq.is_valid(point):
+            # Don't infer parents if the trigger belongs to a sequence that
+            # does not include the child point. E.g.:
+            #   T06 = "waz[-PT6H] => foo"
+            # here waz[-PT6H] is a parent of T06/foo but not of T12/foo.
```
MetRonnie, line 87: This example is not making a lot of sense to me

oliver-sanders: It's essentially the same as the example in the issue:

waz doesn't exist on the sequence T06m, it is a parent (i.e. upstream) of foo at T06

PR #5496, opened 2023-04-27
prerequsite: remove target_point_strings attribute

oliver-sanders:
The `Prerequisite` class contains two attributes with overlapping information, `target_point_strings` and `satisfied`.

See https://github.com/cylc/cylc-flow/pull/5466#discussion_r1178887978

Remove `target_point_strings` (only used internally) to save memory.

**Check List**

### DIFF
--- a/cylc/flow/prerequisite.py
+++ b/cylc/flow/prerequisite.py
@@ -44,7 +44,6 @@ class Prerequisite:
     __slots__ = (
         "satisfied",
         "_all_satisfied",
-        "target_point_strings",
         "start_point",
         "conditional_expression",
         "point",
--- a/cylc/flow/prerequisite.py
+++ b/cylc/flow/prerequisite.py
@@ -41,9 +41,14 @@ class Prerequisite:
     """
 
     # Memory optimization - constrain possible attributes to this list.
-    __slots__ = ["satisfied", "_all_satisfied",
-                 "target_point_strings", "start_point",
-                 "conditional_expression", "point"]
+    __slots__ = (
+        "satisfied",
+        "_all_satisfied",
+        "target_point_strings",
+        "start_point",
+        "conditional_expression",
+        "point",
+    )
 
     # Extracts T from "foo.T succeeded" etc.
     SATISFIED_TEMPLATE = 'bool(self.satisfied[("%s", "%s", "%s")])'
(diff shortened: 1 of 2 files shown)


tests/unit/test_prerequisite.py
```diff
@@ -0,0 +1,116 @@
...
+        ('2000', 'b', 'succeeded'): 'satisfied naturally',
+        ('2000', 'c', 'succeeded'): 'satisfied naturally',
+        # the remaining dependency should be marked as forse-satisfied
+        ('2001', 'd', 'custom'): 'force satisfied',
+    }
+    # mark all prereqs as unsatisfied
+    prereq.set_not_satisfied()
+    assert prereq.satisfied == {
+        # all prereqs INCLUDING the pre-initial should be marked as nnot
+        # satisfied
```
oliver-sanders, line 95: Should the pre-initial dependency be set to `False` here?

It is initialised as `satisfied naturally` so it seems odd for this method to undo that.

cylc/flow/prerequisite.py
```diff
@@ -41,9 +41,14 @@ class Prerequisite:
     """
 
     # Memory optimization - constrain possible attributes to this list.
-    __slots__ = ["satisfied", "_all_satisfied",
-                 "target_point_strings", "start_point",
-                 "conditional_expression", "point"]
+    __slots__ = (
+        "satisfied",
+        "_all_satisfied",
+        "target_point_strings",
```
MetRonnie, line 47:
```suggestion
```

oliver-sanders: Derp.

MetRonnie (approved):
Will conflict with #5504

tests/unit/test_prerequisite.py
hjoliver: Maybe not, but I checked and it's actually not used anywhere in Cylc 8.  We should either remove it, or fix it (pre-initial) depending on the resolution of this open question: https://cylc.github.io/cylc-admin/proposal-cylc-set.html#un-setting-prerequisites

hjoliver (approved):

tests/unit/test_prerequisite.py
hjoliver: I'll put up a new issue for this, as a reminder.

PR #5497, opened 2023-04-27
actions: test up to python 3.11

oliver-sanders:
Add Python 3.11 support.

**Check List**

### DIFF
--- a/.github/workflows/1_create_release_pr.yml
+++ b/.github/workflows/1_create_release_pr.yml
@@ -32,8 +32,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          # return to 3.x once compatible with 3.10+ (pyuv)
-          python-version: '3.9'
+          python-version: '3.x'
 
       - name: Create & checkout PR branch
         uses: cylc/release-actions/stage-1/checkout-pr-branch@v1
--- a/.github/workflows/2_auto_publish_release.yml
+++ b/.github/workflows/2_auto_publish_release.yml
@@ -29,8 +29,7 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@v4
       with:
-        # return to 3.x once compatible with 3.10+ (pyuv)
-        python-version: '3.9'
+        python-version: '3.11'
 
     - name: Get the version number
       uses: cylc/release-actions/stage-2/get-version-from-pr@v1
--- a/.github/workflows/2_auto_publish_release.yml
+++ b/.github/workflows/2_auto_publish_release.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@v4
       with:
-        python-version: '3.11'
+        python-version: '3.x'
 
     - name: Get the version number
       uses: cylc/release-actions/stage-2/get-version-from-pr@v1
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -18,7 +18,6 @@
 import asyncio
 from contextlib import suppress
 from collections import deque
-from dataclasses import dataclass
 from optparse import Values
 import os
 from pathlib import Path
@@ -161,7 +160,6 @@ class SchedulerError(CylcError):
     """Scheduler expected error stop."""
 
 
-@dataclass
 class Scheduler:
     """Cylc scheduler server."""
 
@@ -1803,6 +1801,9 @@ class Scheduler:
         sys.stdout.flush()
         sys.stderr.flush()
 
+        if self.contact_data and self.task_job_mgr:
+            self.task_job_mgr.task_remote_mgr.remote_tidy()
+
         try:
             # Remove ZMQ keys from scheduler
             LOG.debug("Removing authentication keys from scheduler")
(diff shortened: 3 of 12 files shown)


cylc/flow/scripts/lint.py
```diff
@@ -38,7 +38,15 @@
...
-import tomli
+try:
+    # BACK COMPAT: tomli
+    #   Support for Python versions before tomllib was added to the
+    #   standard library.
+    # FROM: Python 3.7
+    # TO: Python: 3.10
+    from tomli import loads as toml_loads, TOMLDecodeError
+except ImportError:
+    from tomllib import loads as toml_loads, TOMLDecodeError
```
oliver-sanders, line 49: The tomli library changed its namespace when entering the standard library (same as importlib_metadata) so needed this try/except.

cylc/flow/task_remote_mgr.py
```diff
@@ -344,8 +344,9 @@ def remote_tidy(self) -> None:
         Timeout any incomplete commands after 10 seconds.
         """
         # Get a list of all platforms used from workflow database:
-        platforms_used = (
-            self.db_mgr.get_pri_dao().select_task_job_platforms())
+        with self.db_mgr.get_pri_dao() as pri_dao:
+            platforms_used = pri_dao.select_task_job_platforms()
```
oliver-sanders, line 348: Ensure the DB gets closed.

.github/workflows/1_create_release_pr.yml
```diff
@@ -32,8 +32,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          # return to 3.x once compatible with 3.10+ (pyuv)
-          python-version: '3.9'
+          python-version: '3.x'
```
MetRonnie, line 35: 3.x here but 3.11 in stage 2?

.github/workflows/2_auto_publish_release.yml
```diff
@@ -29,8 +29,7 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@v4
       with:
-        # return to 3.x once compatible with 3.10+ (pyuv)
-        python-version: '3.9'
+        python-version: '3.11'
```
MetRonnie, line 32: 3.11 here but 3.x in stage 1?

cylc/flow/scheduler.py
```diff
@@ -161,7 +160,6 @@ class SchedulerError(CylcError):
     """Scheduler expected error stop."""
 
 
-@dataclass
```
MetRonnie, line 164: Why remove this?

cylc/flow/task_remote_mgr.py
MetRonnie: Hmm, I suggested this twice in my review of #5445, both times it was marked as resolved but never went in!

cylc/flow/scheduler.py
```diff
@@ -1803,6 +1801,9 @@ async def _shutdown(self, reason: BaseException) -> None:
         sys.stdout.flush()
         sys.stderr.flush()
 
+        if self.task_job_mgr:
+            self.task_job_mgr.task_remote_mgr.remote_tidy()
```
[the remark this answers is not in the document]
MetRonnie: Is this still necessary in conjunction with the [change to `task_remote_mgr.py`](https://github.com/cylc/cylc-flow/pull/5497/files#r1182511940)?

MetRonnie (commented):
I have some questions, otherwise broadly happy

.github/workflows/2_auto_publish_release.yml
oliver-sanders: Not sure why I did that?

```suggestion
        python-version: '3.x'
```

cylc/flow/scheduler.py
oliver-sanders: This was causing incompatibility with Python 3.11 because some of the attributes are mutable which dataclass now rejects outright.

There wasn't an especially good reason for using dataclass. The main reason I put it in was to more explicitly define the scheduler's attributes (I think before type checking syntax came along, can't remember?).

oliver-sanders:
> Is this still necessary in conjunction with the [change to task_remote_mgr.py](https://github.com/cylc/cylc-flow/pull/5497/files#r1182511940)?

Seems to be.

It really shouldn't have been where it was before either way, because it was happening *after* `workflow_db_mgr.on_workflow_shutdown` when we wouldn't expect a new database connection to be made.

cylc/flow/task_remote_mgr.py
oliver-sanders: I'm going to just mark this as resolved :see_no_evil:

cylc/flow/scheduler.py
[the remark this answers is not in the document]
oliver-sanders: https://github.com/cylc/cylc-flow/pull/5497#issuecomment-1532824684

.github/workflows/1_create_release_pr.yml
oliver-sanders: Changed the other to 3.x

MetRonnie (approved):

PR #5513, opened 2023-05-03
use latest version of python + 3.7 for tutorial workflow test

### LINKED ISSUE #5512: Run the tutorial against the newest Python (as well as 3.7)
Currently the Github Actions test for the tutorial workflow is only run for Python 3.7. It would be a good idea to run it for multiple versions of Python.

wxtim:
Closes #5512 

**Check List**

- [x] This is a testing change - no updates to documentation/changelog are required.


.github/workflows/test_tutorial_workflow.yml
```diff
@@ -19,13 +19,16 @@ concurrency:
 
 jobs:
   test:
+    strategy:
+      matrix:
+        python-version: ['3.7', '3.11']
```
oliver-sanders, line 24: Consider using `3.x` to avoid having to update this number.

```suggestion
        python-version: ['3.7', '3.x']
```

wxtim: I changed this for a reason, but I think that reason related to an unrelated test failure.

.github/workflows/test_tutorial_workflow.yml
```diff
@@ -19,13 +19,16 @@ concurrency:
 
 jobs:
   test:
+    strategy:
+      matrix:
+        python-version: ['3.7', '3.x']
```
wxtim, line 24:
```suggestion
        python-version: ['3.7', '3.11']
```

.github/workflows/test_tutorial_workflow.yml
wxtim, line 24:
```suggestion
        python-version: ['3.7', '3']
```

MetRonnie: That error is coming from here https://github.com/cylc/cylc-doc/blob/151dc587f1d338f6a1c3730fb51100948168c7a4/.github/actions/install-dependencies/action.yml#L31

What happens if you just specify `3` instead of `3.x`?

wxtim: That's the current state of the PR
QED

MetRonnie: Hmm, conda is installing Python 3.10: https://github.com/cylc/cylc-flow/actions/runs/4881596596/jobs/8710692879?pr=5513#step:4:491

I'm not sure which version is being used by the tests

oliver-sanders:
Setting `python-version = 3` worked:

```
Successfully set up CPython (3.11.3)
```

.github/workflows/test_tutorial_workflow.yml
MetRonnie: @wxtim Click the link I sent, it's the same as yours but with a link to the line number confirming conda is installing python 3.10

oliver-sanders: Ah, didn't realise we had another environment, I think that's the version of Python which is used to run the tutorial scripts?

Conda/Mamba uses a flexible solver so may pick a lower version of Python to allow other dependencies to be bumped.

So if we definitely want 3.11 we will have to explicitly state it, soz.

.github/workflows/test_tutorial_workflow.yml
```diff
@@ -19,13 +19,16 @@ concurrency:
 
 jobs:
   test:
+    strategy:
+      matrix:
+        python-version: ['3.7', '3']
```
wxtim, line 24:
```suggestion
python-version: ['3.7', '3.11']
```

.github/workflows/test_tutorial_workflow.yml
```diff
@@ -19,13 +19,16 @@ concurrency:
 
 jobs:
   test:
+    strategy:
+      matrix:
+python-version: ['3.7', '3.11']
```
wxtim, line 24:
```suggestion
          python-version: ['3.7', '3.11']
```

MetRonnie:
Strangely conda is still installing Python 3.10 even when 3.11 is specified: https://github.com/cylc/cylc-flow/actions/runs/4883032421/jobs/8713848794?pr=5513#step:4:491

.github/workflows/test_tutorial_workflow.yml
MetRonnie, line 24: I've tested the python version after setup-miniconda has run, it's showing the same version as setup-python. So this should be fine:
```suggestion
        python-version: ['3.7', '3']
```

https://github.com/MetRonnie/cylc-flow/actions/runs/4883788233/jobs/8715616835 (not sure why it failed there)

MetRonnie (approved):
(For the benefit of Oliver)

I've tested the python version after setup-miniconda has run, it's showing the same version as setup-python. So `python-version: 3` should be fine
 
https://github.com/MetRonnie/cylc-flow/actions/runs/4883788233/jobs/8715616835 (CI failed here because it's missing #5497)

PR #5524, opened 2023-05-05
Turn log timestamps back on before running Cylc Play in VIP & VR

### LINKED ISSUE #5505: vip: --no-timestamp default
`cylc play` and `cylc vip` seem to have different defaults for the `--no-timestamp` option.

Also check `cylc vr`.

wxtim:
Closes #5505

Cylc Validate disables the use of timestamps in logging. I've added a command to allow it to be restarted before these commands play a workflow.

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) entry not req'd

### DIFF
--- a/cylc/flow/loggingutil.py
+++ b/cylc/flow/loggingutil.py
@@ -333,11 +333,11 @@ def re_formatter(log_string):
     return log_string
 
 
-def disable_timestamps(logger: logging.Logger) -> None:
-    """For readability omit timestamps from logging."""
+def set_timestamps(logger: logging.Logger, enable: bool) -> None:
+    """Enable or disable logging timestamps."""
     for handler in logger.handlers:
         if isinstance(handler.formatter, CylcLogFormatter):
-            handler.formatter.configure(timestamp=False)
+            handler.formatter.configure(timestamp=enable)
 
 
 def setup_segregated_log_streams(
--- a/cylc/flow/scripts/validate_install_play.py
+++ b/cylc/flow/scripts/validate_install_play.py
@@ -35,7 +35,9 @@ from cylc.flow.scripts.validate import (
 from cylc.flow.scripts.install import (
     INSTALL_OPTIONS, install_cli as cylc_install, get_source_location
 )
+from cylc.flow import LOG
 from cylc.flow.scheduler_cli import PLAY_OPTIONS
+from cylc.flow.loggingutil import set_timestamps
 from cylc.flow.option_parsers import (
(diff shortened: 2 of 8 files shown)


hjoliver (changes requested):
Looks good. One minor issue, it would be slightly cleaner to have a single `toggle_timestamps` function?

wxtim:
> Looks good. One minor issue, it would be slightly cleaner to have a single `toggle_timestamps` function?

Sort of - my only concern is that for me the word "toggle" means "get the state and change it to the other state". Whilst this function could do this, I think it's better if it specifies the desired state to avoid ambiguity. To this end I've made the `enable` argument compulsory, because I think defaulting to `True` is misleading.

cylc/flow/loggingutil.py
```diff
@@ -333,11 +333,11 @@ def re_formatter(log_string):
     return log_string
 
 
-def disable_timestamps(logger: logging.Logger) -> None:
-    """For readability omit timestamps from logging."""
+def toggle_timestamps(logger: logging.Logger, enable: bool) -> None:
```
MetRonnie, line 336: `set_timestamps` would be a more appropriate name imho

cylc/flow/scripts/validate_install_play.py
```diff
@@ -109,5 +110,6 @@ def main(parser: COP, options: 'Values', workflow_id: Optional[str] = None):
         source=orig_source,
     )
 
+    toggle_timestamps(LOG, True)
```
MetRonnie, line 113: Still need to respect CLI option
```suggestion
    toggle_timestamps(LOG, options.log_timestamp)
```

cylc/flow/scripts/validate_reinstall.py
```diff
@@ -177,6 +178,7 @@ def vro_cli(parser: COP, options: 'Values', workflow_id: str):
 
     # run play anyway, to play a stopped workflow:
     else:
+        toggle_timestamps(LOG, True)
```
MetRonnie, line 181:
```suggestion
        toggle_timestamps(LOG, options.log_timestamp)
```

hjoliver:
> Sort of - my only concern is that for me the word "toggle" means "get the state and change 

Yes, badly chosen word sorry. I really meant something like `set_timestamps(state=bool)` - i.e. one function but you do have to specify the desired state, not just "toggle" to the other state whatever that is.

hjoliver (approved):
👍

PR #5528, opened 2023-05-09
schema: enable selected mutations for the stopping workflow state

oliver-sanders:
* Some mutations e.g. "stop" remain applicable even when a workflow is shutting down. E.G. you could tell a workflow to shut down, then upgrade that to "--now" with a subsequent mutation.
* The UI now uses metadata stored in mutations "description" field to filter out mutations which are not applicable, causing these to be disabled at present.

To test, try starting a workflow like this:

```
[scheduling]
  [[graph]]
    R1 = a

[runtime]
  [[a]]
    script = sleep 1000
```

* Then, opening this workflow in the GUI and pressing the stop button.
* The workflow should now show as stopping, but not actually stop due to the running task.
* Then open the mutations menu for the workflow and edit the stop mutation. (stop was disabled in the menu before)
* Change the mode to NOW_NOW and press submit.
* The workflow should now shut down immediately.

**Check List**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,14 @@ creating a new release entry be sure to copy & paste the span tag with the
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
 -------------------------------------------------------------------------------
+## __cylc-8.1.4 (<span actions:bind='release-date'>Awaiting Release</span>)__
+
+### Fixes
+
+[#5228](https://github.com/cylc/cylc-flow/pull/5228) -
+Enabled the "stop", "poll", "kill" and "message" commands to be issued from
+the UI whilst the workflow is in the process of shutting down.
+
 ## __cylc-8.1.4 (<span actions:bind='release-date'>Released 2023-05-04</span>)__
 
 ### Fixes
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -10,6 +10,14 @@ creating a new release entry be sure to copy & paste the span tag with the
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
 -------------------------------------------------------------------------------
+## __cylc-8.1.4 (<span actions:bind='release-date'>Awaiting Release</span>)__
```
MetRonnie, line 13:
```suggestion
## __cylc-8.1.5 (<span actions:bind='release-date'>Awaiting Release</span>)__
```

hjoliver (approved):
This is just doc strings. Presumably the fix will go into cylc-ui shortly?

wxtim:
> This is just doc strings. Presumably the fix will go into cylc-ui shortly?

Functional testing suggests that these apparent docstrings are functional - I've had a quick grep of cylc-ui and uiserver but cannot see the mechanism - Perhaps @oliver-sanders could explain for our general edification?

wxtim (approved):
- [x] Works functionally - have checked that all three commands are still available for stopping workflows.
- [x] Have read the code. I have the same concern as Hilary that I don't understnad the mechanism by which this change works. Yet it moves!

oliver-sanders:
> This is just doc strings. Presumably the fix will go into cylc-ui shortly?

API-on-the-fly!

The GUI reads this information and uses it to decide which mutations to enable, testing instructions in the OP.

PR #5460, opened 2023-04-12
validate-reinstall: don't print reload message

oliver-sanders:
* The reload message is not applicable when using `cylc vr` because it does it for you.
* Closes #5344

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [ ] Tests are included (or explain why tests are not needed).
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #5529, opened 2023-05-10
validate-reinstall: don't print reload message [skip ci]

hjoliver:
Cherry-picked #5460 to 8.1.x to get it released sooner (this small bug is already causing some confusion out in the wild). 

One review will do.

* The reload message is not applicable when using `cylc vr` because it does it for you.


wxtim (approved):
Yes.

PR #5532, opened 2023-05-11
GH Actions patch

MetRonnie:
Follow-up to #5493

### DIFF
--- a/.github/workflows/branch_sync.yml
+++ b/.github/workflows/branch_sync.yml
@@ -84,7 +84,7 @@ jobs:
         id: diff
         if: steps.merge.outcome == 'success'
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/branch_sync.yml
```diff
@@ -84,7 +84,7 @@ jobs:
         id: diff
         if: steps.merge.outcome == 'success'
         run: |
-          if [[ "$(git rev-parse HEAD)" == "$(git rev-parse master)" ]]; then
+          if [[ "$(git rev-parse HEAD)" == "$(git rev-parse origin/master)" ]]; then
```
MetRonnie, line 87: Oops, as master is currently checked out this was always true.

(Didn't pick up in my testing as I was using a different branch to master

PR #5534, opened 2023-05-11
Logging: say command actioned instead of succeeded

MetRonnie:
```
INFO - Command succeeded: poll_tasks(['1/foo'])
```

"Succeeded" is perhaps too strong a word. What's happened is the command finished without any uncaught exceptions but it doesn't mean that it necessarily did what the user expected (e.g. in this case there were no available hosts which is logged in a separate debug message).

**Check List**

- [x] No dependency changes
- [x] Tests updated
- [x] No changelog entry as tiny change
- [x] No docs update needed

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -898,10 +898,11 @@ class Scheduler:
             else:
                 if n_warnings:
                     LOG.info(
-                        'Command succeeded with %s warning(s): %s' %
-                        (n_warnings, cmdstr))
+                        f"Command actioned with {n_warnings} warning(s): "
+                        f"{cmdstr}"
+                    )
                 else:
-                    LOG.info(f"Command succeeded: {cmdstr}")
+                    LOG.info(f"Command actioned: {cmdstr}")
                 self.is_updated = True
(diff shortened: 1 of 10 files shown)


oliver-sanders (approved):
Might lean towards "completed" rather than "done".

MetRonnie:
Actually I might lean towards "processed" or "submitted" because some commands e.g. reload don't truly finish until the next main loop iteration

oliver-sanders:
We used to say "command queued", not sure what happened to that.

MetRonnie:
That also gets logged but was demoted to debug a while ago

oliver-sanders:
> some commands e.g. reload don't truly finish until the next main loop iteration

Note the proper solution to this will come with https://github.com/cylc/cylc-flow/issues/3329

hjoliver (approved):
Good, but for the wording, how about "Command actioned".

That says the command action was started but does not necessarily imply it finished yet. 

(Whereas "done" and "completed" imply it finished; and "processed" doesn't necessarily imply it even started).

hjoliver (approved):

PR #5535, opened 2023-05-12
`cylc play`: add suggestion for `--upgrade` in non-interactive terminal

wxtim:
Companion to https://github.com/cylc/cylc-uiserver/pull/455

**Check List**

- [x] ~Tests are included (or explain why tests are not needed).~ One line message for Cylc GUI to pick up.
- [x] ~`CHANGES.md` entry included if this is a change that can affect users~ Covered by CHANGES.md in https://github.com/cylc/cylc-uiserver/pull/455
- [x] ~[Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.~ N/A

### DIFF
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -523,6 +523,7 @@ def _version_check(
                     process=str.lower,
                 )
             # we are in non-interactive mode, abort abort abort
+            print(cparse('Use "--upgrade" to upgrade the workflow.'))
             return False
         elif itt > 2 and this > that:
             # restart would INCREASE the Cylc version in a little way
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -523,8 +523,7 @@ def _version_check(
                     process=str.lower,
                 )
             # we are in non-interactive mode, abort abort abort
-            else:
-                print('Use "--upgrade" to upgrade the workflow.', file=sys.stderr)
+            print('Use "--upgrade" to upgrade the workflow.', file=sys.stderr)
             return False
         elif itt > 2 and this > that:
             # restart would INCREASE the Cylc version in a little way
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -523,7 +523,8 @@ def _version_check(
                     process=str.lower,
                 )
             # we are in non-interactive mode, abort abort abort
-            print(cparse('Use "--upgrade" to upgrade the workflow.'))
+            else:
+                print('Use "--upgrade" to upgrade the workflow.', file=sys.stderr)
             return False
         elif itt > 2 and this > that:
             # restart would INCREASE the Cylc version in a little way


MetRonnie (approved):

cylc/flow/scheduler_cli.py
```diff
@@ -523,6 +523,7 @@ def _version_check(
                     process=str.lower,
                 )
             # we are in non-interactive mode, abort abort abort
+            print(cparse('Use "--upgrade" to upgrade the workflow.'))
```
oliver-sanders, line 526: * This message should not appear in interactive mode (covered by the prompts above).
* No need for `cparse` unless ANSI colour is used in the message.

```suggestion
            else:
                print('Use "--upgrade" to upgrade the workflow.')
```

cylc/flow/scheduler_cli.py
```diff
@@ -523,6 +523,8 @@ def _version_check(
                     process=str.lower,
                 )
             # we are in non-interactive mode, abort abort abort
+            else:
+                print('Use "--upgrade" to upgrade the workflow.', file=sys.stderr)
```
wxtim, line 527:
```suggestion
            print('Use "--upgrade" to upgrade the workflow.', file=sys.stderr)
```

wxtim: The else isn't actually necessary, since the if contains a return.

cylc/flow/scheduler_cli.py
```diff
@@ -523,6 +523,7 @@ def _version_check(
                     process=str.lower,
                 )
             # we are in non-interactive mode, abort abort abort
+            print('Use "--upgrade" to upgrade the workflow.', file=sys.stderr)
```
MetRonnie, line 526: There is a bit of a mismatch printing this to stderr when just above
```
This workflow was previously run with <yellow>{last_run_version}</yellow>.
This version of Cylc is <green>{__version__}</green>.
```
was printed to stdout

wxtim: @oliver-sanders  - Any reason not to print the earlier msg to stderr?

MetRonnie: Unfortunately the GUI and TUI are only regurgitating stderr. We should find a way for them to include the useful info about which versions are flying about IMHO

oliver-sanders: Fair point, but a widespread issue not related to this PR.

I don't think stdout/err are a great interface for Tui/GUI, but short of a Python API (that allows Cylc version changing) I can't think of a better option that isn't tons of work.

In this case, the version compat issue *should* also be handled better in at the UI end as we intend to implement upgrade logic there.

oliver-sanders (approved):

cylc/flow/scheduler_cli.py
oliver-sanders: Moved to #5554

PR #5537, opened 2023-05-15
Support expanding `<param1><param2>` in inherited family names

### LINKED ISSUE #5118: parameters: complete support for alternative syntax

wxtim:
Closes #5118 

The form `FAM<foo,bar>` is already supported in graph strings and `[runtime][task]inherit`, while the functionally equivalent `FAM<foo><bar>` is only supported in graph strings. This PR adds full support for the latter

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/611.
- [x] ~If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.~ N/A

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,9 @@ ones in. -->
 
 ### Enhancements
 
+[#5537](https://github.com/cylc/cylc-flow/pull/5537) - Allow parameters
+in family names to be split, e.g. `<foo>FAM<bar>`.
+
 [#5405](https://github.com/cylc/cylc-flow/pull/5405) - Improve scan command
 help, and add scheduler PID to the output.
 
--- a/cylc/flow/param_expand.py
+++ b/cylc/flow/param_expand.py
@@ -58,6 +58,7 @@ foo_m1=>bar_m1_n2
 
 from contextlib import suppress
 import re
+from typing import List, Tuple
 
 from cylc.flow.exceptions import ParamExpandError
 from cylc.flow.task_id import TaskID
@@ -70,9 +71,6 @@ REC_NAMES = re.compile(r'(?:[^,<]|<[^>]*>)+')
 REC_P_ALL = re.compile(r"(%s)?(?:<(.*?)>)?(.+)?" % TaskID.NAME_RE)
 # To extract all parameter lists e.g. 'm,n,o' (from '<m,n,o>').
 REC_P_GROUP = re.compile(r"<(.*?)>")
-# As REC_P_ALL, but retaining <> so that we can tell which bits of re.split
-# are templates, and which are plain text:
-REC_P_MATCH = re.compile(r'(<[^>]*>)')
 # To extract parameter name and optional offset or value e.g. 'm-1'.
 REC_P_OFFS = re.compile(
     r'(\w+)\s*([\-+]\s*\d+|=\s*%s)?' % TaskID.NAME_SUFFIX_RE)
@@ -205,7 +203,7 @@ class NameExpander:
                 self._expand_name(results, tmpl, params[1:], spec_vals)
 
     @staticmethod
-    def _parse_task_name_string(parent):
+    def _parse_task_name_string(task_str: str) -> Tuple[List[str], str]:
         """Takes a parent string and returns a list of parameters and a
         template string.
 
@@ -240,34 +238,32 @@ class NameExpander:
             >>> this('FAM<i = cat ,j=3>')
             (['i = cat', 'j=3'], 'FAM{i}{j}')
         """
-        tmpl_list = REC_P_MATCH.split(parent)
         param_list = []
-        for template in tmpl_list:
-            group = REC_P_GROUP.findall(template)
-            if group:
-                param = group[0]
-                if ',' in param:
-                    # parameter syntax `<foo, bar>`
-                    replacement = ''
-                    for sub_param in param.split(','):
-                        sub_param = sub_param.strip()
-                        param_list.append(sub_param)
-                        if '=' in sub_param:
-                            sub_param = sub_param.split('=')[0].strip()
-                        replacement += '{' + sub_param + '}'
+
+        def _parse_replacement(match: re.Match) -> str:
+            nonlocal param_list
+            param = match.group(1)
+            if ',' in param:
+                # parameter syntax `<foo, bar>`
+                replacement = ''
+                for sub_param in param.split(','):
+                    sub_param = sub_param.strip()
+                    param_list.append(sub_param)
+                    if '=' in sub_param:
+                        sub_param = sub_param.split('=')[0].strip()
+                    replacement += '{' + sub_param + '}'
+            else:
+                # parameter syntax: `<foo><bar>`
+                param_list.append(param)
+                if '=' in param:
+                    replacement = '{' + param.split('=')[0] + '}'
                 else:
-                    # parameter syntax: `<foo><bar>`
-                    param_list.append(param)
-                    if '=' in param:
-                        replacement = '{' + param.split('=')[0] + '}'
-                    else:
-                        replacement = '{' + param + '}'
+                    replacement = '{' + param + '}'
+            return replacement
 
-                # Replace param in template list with template.
-                if f'<{param}>' in tmpl_list:
-                    tmpl_list[tmpl_list.index(f'<{param}>')] = replacement
+        replacement = REC_P_GROUP.sub(_parse_replacement, task_str)
 
-        return param_list, ''.join(tmpl_list)
+        return param_list, replacement
 
     def expand_parent_params(self, parent, param_values, origin):
         """Replace parameters with specific values in inherited parent names.
--- a/cylc/flow/param_expand.py
+++ b/cylc/flow/param_expand.py
@@ -70,6 +70,9 @@ REC_NAMES = re.compile(r'(?:[^,<]|<[^>]*>)+')
 REC_P_ALL = re.compile(r"(%s)?(?:<(.*?)>)?(.+)?" % TaskID.NAME_RE)
 # To extract all parameter lists e.g. 'm,n,o' (from '<m,n,o>').
 REC_P_GROUP = re.compile(r"<(.*?)>")
+# As REC_P_ALL, but retaining <> so that we can tell which bits of re.split
+# are templates, and which are plain text:
+REC_P_MATCH = re.compile(r'(<[^>]*>)')
 # To extract parameter name and optional offset or value e.g. 'm-1'.
 REC_P_OFFS = re.compile(
     r'(\w+)\s*([\-+]\s*\d+|=\s*%s)?' % TaskID.NAME_SUFFIX_RE)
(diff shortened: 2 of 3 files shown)


cylc/flow/param_expand.py
```diff
@@ -201,6 +201,64 @@ def _expand_name(self, results, tmpl, params, spec_vals=None):
...
+
+            # Parent contains 2 parameters, once with an = sign in it.
+            >>> this('something<foo=42>middlebit<bar>other')
+            (['foo=42', 'bar'], 'something{foo}middlebit{bar}other')
+
+            # Parent contains 2 parameters in 2 <>:
+            >>> this('something<foo,bar=99>other')
+            (['foo', 'bar=99'], 'something{foo}{bar}other')
+        """
+        params = re.findall(r'<.*?>', parent)
```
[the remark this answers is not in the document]
wxtim: I wanted this! And missed it in the docs!

hjoliver:
> I'm not clear if we need to document it since many (most?) users seem to think that it exists regardless?

I think it need to be documented now that's it's becoming officially supported syntax. All examples in the docs will (presumably) be using the original syntax.

wxtim:
Converted to draft until I've:

- Responded to comments
- Written a docs PR

cylc/flow/param_expand.py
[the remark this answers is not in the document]
wxtim: REC_P_GROUP doesn't work straightforwardly, because we are replacing the entire parameter including the angle brackets, not just the parameter name.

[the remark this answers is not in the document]
oliver-sanders: (Unresolving)

I think you can swap this out for `REC_P_GROUP`.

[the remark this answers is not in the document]
oliver-sanders: Please move the "split" regex into a constant using `re.compile`.  Otherwise, Python is having to compile the regex every time this function is called.

cylc/flow/param_expand.py
```diff
@@ -201,6 +201,73 @@ def _expand_name(self, results, tmpl, params, spec_vals=None):
...
+
+            # Parent contains spaces around = sign:
+            >>> this('FAM<i = cat ,j=3>')
+            (['i = cat', 'j=3'], 'FAM{i}{j}')
+        """
+        tmpl_list = re.split(r'(<[^>]*>)', parent)
+        raw_param_list = [
+            REC_P_GROUP.findall(i)[0]
+            for i in tmpl_list
+            if REC_P_GROUP.findall(i)
```
oliver-sanders, line 244: This is calling the same regex twice per item.

Given that we are already iterating using a for loop below we may as well combine this list comprehension with the for loop to save unnecessary iteration.

```python
for item in re.split(r'(<[^>]*>)', parent):
    param_match = REC_P_GROUP.findall(i)[0]
    if param_match:
        ...
```

cylc/flow/param_expand.py
```diff
@@ -201,6 +201,73 @@ def _expand_name(self, results, tmpl, params, spec_vals=None):
                 spec_vals[params[0][0]] = param_val
                 self._expand_name(results, tmpl, params[1:], spec_vals)
 
+    @staticmethod
+    def _parse_parent_string(parent):
```
oliver-sanders, line 205: parent?

graph string, task name, item, string.

cylc/flow/param_expand.py
```diff
@@ -201,6 +201,73 @@ def _expand_name(self, results, tmpl, params, spec_vals=None):
...
+                # parameter syntax `<foo, bar>`
+                sub_params = [
+                    i.strip() for i in param.split(',')]
+                for sub_param in sub_params:
+                    param_list.append(sub_param)
+                    if '=' in sub_param:
+                        sub_params[
+                            sub_params.index(sub_param)
+                        ] = sub_param.split('=')[0].strip()
+                replacement = '{' + '}{'.join(sub_params) + '}'
```
oliver-sanders, line 258: This logic must already be happening somewhere else in the code so is presumably happening twice with this addition?

Looks like it might be covered by `REC_P_OFFS`?

tests/unit/test_param_expand.py
```diff
@@ -344,3 +347,122 @@ def test_parameter_graph_mixing_offset_and_conditional(self):
...
+        ).get(),
+        myParam(
+            raw_str='foo<bar=3>baz',
+            parameter_values={'bar': [1, 2]},
+            raises=(ParamExpandError, '^illegal'),
+            id='illegal-value'
+        ).get(),
+        myParam(
+            expect=({'bar': 1}, 'foo_bar1_baz'),
+            raw_str='foo<bar=3>baz',
```
oliver-sanders, line 449: Would be worth testing param indexing e.g. `<foo - 1, bar + 2>`

cylc/flow/param_expand.py
wxtim: Agreed.

tests/unit/test_param_expand.py
wxtim: See line 429?

cylc/flow/param_expand.py
wxtim: It's happening in the `expand` method. Unfortunately I'm struggling to get the interface to work elegantly - I think it might be worth thoroughly testing and a refactor at some point. I strongly suspect  that `tests/functional/param-expand` is a prime candidate for replacement (or at least pruning) with unit tests and functional tests..

CHANGES.md
```diff
@@ -15,6 +15,9 @@ ones in. -->
 
 ### Enhancements
 
+[#5537](https://github.com/cylc/cylc-flow/pull/5537) - Allow parameters
+in the form `<foo>word<bar>` which were not previously fully supported.
```
MetRonnie, line 19: I think this would be a bit clearer
```suggestion
in the form `task<foo><bar>` which were not previously fully supported.
```

tests/unit/test_param_expand.py
```diff
@@ -344,3 +347,122 @@ def test_parameter_graph_mixing_offset_and_conditional(self):
...
+        """Ease of reading wrapper for pytest.param
+
+        Args:
+            expect:
+                Output of expand_parent_params()
+            raw_str:
+                The parent_params input string.
+            parameter_values
+        """
+        self.xfail = xfail
```
MetRonnie, line 372: I don't think this xfail stuff is being used?

CHANGES.md
wxtim: `task<foo><bar>` illustrates the new syntax, but it doesn't illustrate that the new syntax allows one to have parameters in arbitrary places in the text.

tests/unit/test_param_expand.py
wxtim: Nope 

```suggestion
```

CHANGES.md
MetRonnie:
```suggestion
in family names to be split, e.g. `<foo>FAM<bar>`.
```

tests/unit/test_param_expand.py
```diff
@@ -344,3 +347,121 @@ def test_parameter_graph_mixing_offset_and_conditional(self):
...
+            expect=({'bar': 1}, 'foo_bar1_baz'),
+            raw_str='foo<bar=3>baz',
+            raises=(ParamExpandError, '^parameter \'bar\' undefined'),
+            id='parameter-undefined'
+        ).get(),
+    )
+)
+def test_expand_parent_params(param):
+    if param.xfail:
+        pytest.skip('Functionality yet to be added')
```
MetRonnie, line 456: Please remove all references to xfail in that case

MetRonnie (commented):
I've unresolved some of Oliver's comments which have not been addressed

tests/unit/test_param_expand.py
```diff
@@ -337,7 +337,7 @@ def test_parameter_graph_mixing_offset_and_conditional(self):
             self.assertEqual(
                 len(expanded_values),
                 len(expanded),
-                f"Invalid length for expected {expanded_values} and "
+                f"Invalid_ length for expected {expanded_values} and "
```
MetRonnie, line 340: ?

MetRonnie (commented):
I've noticed that `<x>foo<y>` ends up like `_x1foo_y1`, though I imagine this is a low priority issue

cylc/flow/param_expand.py
```diff
@@ -70,6 +70,7 @@ def expand(template, params, results, values=None):
 REC_P_ALL = re.compile(r"(%s)?(?:<(.*?)>)?(.+)?" % TaskID.NAME_RE)
 # To extract all parameter lists e.g. 'm,n,o' (from '<m,n,o>').
 REC_P_GROUP = re.compile(r"<(.*?)>")
+REC_P_MATCH = re.compile(r'(<[^>]*>)')
```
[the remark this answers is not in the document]
wxtim: Yes - for what I'm doing I need to have the `<>` to allow me to identify which parts are templates.
```
>>> REC_P_GROUP.split('foo<bar>baz')
['', 'foo', 'bar', 'baz', '']          # Which one is a parameter?
>>> REC_P_MATCH.split('foo<bar>baz')
['', 'foo', '<bar>', 'baz', '']        # Bar is a parameter
```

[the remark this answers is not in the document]
MetRonnie: I have been able to do away with `REC_P_MATCH`: https://github.com/wxtim/cylc/pull/55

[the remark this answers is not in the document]
wxtim: Nice.

MetRonnie (approved):

cylc/flow/param_expand.py
```diff
@@ -240,34 +238,32 @@ def _parse_task_name_string(parent):
...
-                    # parameter syntax `<foo, bar>`
-                    replacement = ''
-                    for sub_param in param.split(','):
-                        sub_param = sub_param.strip()
-                        param_list.append(sub_param)
-                        if '=' in sub_param:
-                            sub_param = sub_param.split('=')[0].strip()
-                        replacement += '{' + sub_param + '}'
+
+        def _parse_replacement(match: re.Match) -> str:
```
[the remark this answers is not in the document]
MetRonnie: Little bit tricky to refactor due to the nonlocal `param_list` variable. But performance-wise I think the reduced number of operations here wins out over the nested function.

Before:
```
%timeit NameExpander._parse_task_name_string('something<foo>middlebit<bar=1>other')
3.47 µs ± 285 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

Nested function approach:
```
%timeit NameExpander._parse_task_name_string('something<foo>middlebit<bar=1>other')
2.05 µs ± 174 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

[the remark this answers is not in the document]
oliver-sanders: I think it could just be done with a simple for loop right?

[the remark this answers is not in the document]
MetRonnie: Ok there's this
```diff
diff --git a/cylc/flow/param_expand.py b/cylc/flow/param_expand.py
index 828192d26..a75c5c363 100644
--- a/cylc/flow/param_expand.py
+++ b/cylc/flow/param_expand.py
@@ -241,6 +241,5 @@ class NameExpander:
         param_list = []
 
-        def _parse_replacement(match: re.Match) -> str:
-            nonlocal param_list
+        for match in REC_P_GROUP.finditer(task_str):
             param = match.group(1)
             if ',' in param:
@@ -260,9 +259,8 @@ class NameExpander:
                 else:
                     replacement = '{' + param + '}'
-            return replacement
 
-        replacement = REC_P_GROUP.sub(_parse_replacement, task_str)
+            task_str = task_str.replace(match.group(0), replacement, 0)
 
-        return param_list, replacement
+        return param_list, task_str
 
     def expand_parent_params(self, parent, param_values, origin):
```

but it's not actually any faster according to my testing

MetRonnie (changes requested):
https://github.com/cylc/cylc-flow/pull/5537#discussion_r1246366358

MetRonnie (approved):

oliver-sanders (approved):

PR #5538, opened 2023-05-15
GH Actions: Cherry pick branch sync workflow

MetRonnie:
Ugh, can never grok which branch a workflow needs to be on in order to run

Cherry pick of #5493 + #5532


MetRonnie:
That sorted it

Issue #4917, opened 2022-06-16
linter: additional rules


oliver-sanders:
#4900 has gone beyond a basic `sed` file and is now the basis for a more generic linter :tada:.

Some follow-on ideas:

### Editor integration:

(See #5014)

### 7 to 8 migration rules

> This linter could be a highly valuable tool in the 7->8 upgrade. Also include a couple of Python 2->3 pointers.

* [x] Missing shebang
  * If Jinja2 or EmPy syntax is detected in the `suite.rc` or `flow.cylc` file and the first line of the file is not a matching shebang.
  * (`rose suite-run` used to tolerate missing shebang lines)
* [ ] Module name changes
  * `cylc` -> `cylc.flow`
  * `rose` -> `metomi.rose`
  * `isodatetime` -> `metomi.isodatetime`
  * In Python files `from _ import` and `import _`.
  * In `.cylc` & `.rc` files `from "_" import`
* [ ] Tabs in Python files (e.g. Jinja2 filters)
  * Use spaces to indent Python 3 files, mix of tabs and spaces is a syntax error.
* [x] Zero-prefixed integers in `.cylc`, `.rc` files.
  * E.G. `05` rather than `5`, previously accepted, now a Jinja2 syntax error.
* [x] Suicide triggers
  * No longer needed, use optional outputs (for `.cylc` files only).
* [x] platform = $(rose host-select)
  * Advise using built-in host selection via configured platforms rather than `$(rose host-select)`.
  * Note there may be legit reasons to use subshells here (platforms does not offer a `filter` interface yet).
  * But `rose host-select` is definitely unnecessary.
* [x] Deprecated env vars (all files in project).
* [x] Check for deprecated commands (documented in `cylc.flow.scripts.cylc.DEAD_ENDS`).
* [x] Parameter simplification
  * `<x><y>` is valid but can be simplified `<x, y>`
  * See #5537
* [ ] Deprecated [task](https://cylc.github.io/cylc-doc/stable/html/user-guide/writing-workflows/runtime.html#task-event-template-variables) and workflow event template variables
  * e.g. `%(suite)` = https://github.com/cylc/cylc-flow/pull/5879
* [x] Use of `rose suite-hook` (command removed at Rose 2)
* [ ] `CYLC_VERSION={{CYLC_VERSION}}` (same for `ROSE_` and `FCM_`). https://github.com/cylc/cylc-flow/pull/5890
[...]

PR #5549, opened 2023-05-17
Ensure that lint codes are unique

### LINKED ISSUE #5540: cylc lint - S008 is used twice
### Description

If you look at [the `cylc lint` docs](https://cylc.github.io/cylc-doc/8.1.4/html/user-guide/writing-workflows/configuration.html#s008), S008 is used twice. Once for the shebang line missing and once for line length.

image

I'm raising the issue here as it is the code here which controls those documents. 

### Reproducible Example

See documentation.

### Expected Behaviour

S008 should only be for one item, not two. I'm guessing line length was 8, then it was commented out, then uncommented out but another entry had re-used S008. 

image

### LINKED ISSUE #5541: cylc lint - missing jinja2 shebang excessive warnings?
### Description

If you run `cylc lint` and it finds multiple `.cylc` files, it will want a `#!jinja2` line at the top of each of them if it detects jinja2 in there. Is this required, or should it only be required in the `flow.cylc` file? Looking at https://github.com/cylc/cylc-flow/issues/4917 - that reads as though it should be like I suggest, only for the top level file.

### Reproducible Example

```bash
touch flow.cylc
echo '{{a}}' > include.cylc
cylc lint
[S008] include.cylc:1: jinja2 found: no shebang (#!jinja2)
```
### Expected Behaviour

No warning as it is not the `flow.cylc` file.

wxtim:
Closes #5540 (After a docs rebuild)
Closes #5541 

- (Main issue) Ensure that Max line length (which rule is added to the list of checks dynamically is given a different number to checks in the rest of the list.
- Test that check numbers are unique and that the list of error codes has no gaps in it.
- Allow arbitrary functions as well as regexes for checks (return True if there is a problem)
- Adds a number of refinements from https://github.com/cylc/cylc-flow/issues/4917

**Check List**

- [x] `CHANGES.md` entry included
- [x] ~[Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.~ Fix - N/A


tests/unit/scripts/test_lint.py
```diff
@@ -533,6 +534,24 @@ def test_invalid_tomlfile(tmp_path):
...
+
+
+def test_checks_are_sorted():
+    checks = {}
+    for i in range(2):
+        for origin in [STYLE_CHECKS, UPG_CHECKS]:
+            checks[[k for k in origin][i]] = [v for v in origin.values()][i]
+    sorted_checks = sorted(checks.items(), key=sort_checks)
+    result = [i[1]['purpose'] + str(i[1]['index']) for i in sorted_checks]
+    assert result == ['S1', 'S2', 'U1', 'U2']
```
MetRonnie, line 549: This is a bit gnarly, may I suggest
```suggestion
    sorted_checks = sorted(
        [*UPG_CHECKS.items(), *STYLE_CHECKS.items()],
        key=sort_checks
    )
    assert [
        f"{v['purpose']}{v['index']}" for _, v in sorted_checks
    ] == [
        *[f"S{n}" for n in range(1, len(STYLE_CHECKS) + 1)],
        *[f"U{n}" for n in range(1, len(UPG_CHECKS) + 1)]
    ]
```

tests/unit/scripts/test_lint.py
```diff
@@ -533,6 +534,24 @@ def test_invalid_tomlfile(tmp_path):
...
+
+
+@pytest.mark.parametrize(
+    'test_set', [STYLE_CHECKS, UPG_CHECKS]
+)
+def test_lint_codes_are_sequential(test_set):
+    checks = [i['index'] for i in test_set.values()]
+    assert list(range(1, len(checks) + 1)) == checks
```
MetRonnie, line 557: A Yoda expression, this is
```suggestion
    assert checks == list(range(1, len(checks) + 1))
```

MetRonnie (changes requested):
Another reference to 8 here
https://github.com/cylc/cylc-flow/blob/a4afb422b29ca70fe5d9de9d279f12ce224a98f5/cylc/flow/scripts/lint.py#L449-L450

cylc/flow/scripts/lint.py
```diff
@@ -546,6 +538,12 @@ def get_cylc_files(
                 yield path
 
 
+def sort_checks(check):
```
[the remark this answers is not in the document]
wxtim: Yes.

wxtim:
Converting to draft pending my having a think about keeping linter codes consisitent.

cylc/flow/scripts/lint.py
```diff
@@ -592,30 +627,40 @@ def get_reference_text(checks):
...
             title = CHECKS_DESC[meta["purpose"]]
             output += f'\n{title}\n{"-" * len(title)}\n\n'
 
+            if current_checkset == 'A':
+                output += (
+                    'U998 and U999 represent automatically generated'
+                    ' sets of deprecations and upgrades.'
+                    '\nBecause of the automation they cannot be'
+                    ' consistently numbered.\n\n'
+                )
```
MetRonnie, line 671: As above

tests/unit/scripts/test_lint.py
```diff
@@ -6,7 +6,8 @@
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
 # (at your option) any later version.
-#
+#PID: 12632 Threads: 1 RSS: 1044
```
MetRonnie, line 9: ?
```suggestion
```

tests/unit/scripts/test_lint.py
```diff
@@ -379,13 +367,13 @@ def test_get_upg_info(fixture_get_deprecations, findme):
     n.b this is just sampling to ensure that the test it getting items.
     """
     if findme.startswith('!'):
-        assert findme[1:] not in str(fixture_get_deprecations)
-    elif findme.startswith('['):
-        pattern = f'\\[\\s*{findme.strip("[").strip("]")}\\s*\\]\\s*$'
-        assert pattern in [i.pattern for i in fixture_get_deprecations.keys()]
+        result = ([
+            i['function'](findme) for i in fixture_get_deprecations.values()])
+        assert not any(result)
```
MetRonnie, line 373:
```suggestion
        assert not any(
            i['function'](findme) for i in fixture_get_deprecations.values()
        )
```

tests/unit/scripts/test_lint.py
```diff
@@ -379,13 +367,13 @@ def test_get_upg_info(fixture_get_deprecations, findme):
...
     else:
-        pattern = f'^\\s*{findme}\\s*=\\s*.*'
-        assert pattern in [i.pattern for i in fixture_get_deprecations.keys()]
+        results = (
+            [i['function'](findme) for i in fixture_get_deprecations.values()])
+        assert any(results)
```
MetRonnie, line 377:
```suggestion
        assert not any(
            i['function'](findme) for i in fixture_get_deprecations.values()
        )
```

MetRonnie (changes requested):

tests/unit/scripts/test_lint.py
wxtim: I suspect that I may need to fiddle with my mouse-wheel-button settings... I think I'm pasting whilst scrolling...

cylc/flow/scripts/lint.py
```diff
@@ -54,8 +55,142 @@
...
+    'CYLC_SUITE_RUN_DIR': 'CYLC_WORKFLOW_RUN_DIR',
+}
+
+OBSOLETE_ENV_VARS = {
+    'CYLC_SUITE_DEF_PATH',
+    'CYLC_SUITE_DEF_PATH_ON_SUITE_HOST'
+}
+
+
+def check_jinja2_no_shebang(line, **kwargs):
```
MetRonnie, line 83: Why use a `kwargs` dict instead of explicitly specifying the args?
```suggestion
def check_jinja2_no_shebang(line, file, jinja_shebang=False, fallback=None):
```

cylc/flow/scripts/lint.py
```diff
@@ -555,31 +579,42 @@ def get_reference_rst(checks):
...
-            url = meta['url']
+            if current_checkset == 'A':
+                output += (
+                    '\n.. note::\n'
+                    '\n   U998 and U999 represent automatically generated'
+                    'sets of deprecations and upgrades'
+                    '\n   Because of the automation they cannot be'
+                    ' consistently numbered.\n\n\n'
+                )
+
```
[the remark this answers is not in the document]
oliver-sanders: I don't think we should "automatically generate" these codes, just use one code for all upgrades.

cylc/flow/scripts/lint.py
wxtim: Because these functions are called generically with a larger set of kwargs - some of which this one doesn't need.

cylc/flow/scripts/lint.py
wxtim: Sorry, what _was_ above?

cylc/flow/scripts/lint.py
```diff
@@ -378,14 +382,16 @@ def get_upgrader_info():
...
+    for upgrades_for_version in upgrades.values():
+        for index, upgrade in enumerate(upgrades_for_version):
             # Set a flag indicating that a variable has been moved.
+            is_dep, is_upg = False, False
             if upgrade['new'] is None:
                 section_name = list_to_config(
                     upgrade["old"], upgrade["is_section"])
                 short = f'{section_name} - not available at Cylc 8'
                 rst = f'``{section_name}`` is not available at Cylc 8'
+                is_upg = True
```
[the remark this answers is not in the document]
wxtim: Ay.

cylc/flow/scripts/lint.py
MetRonnie: https://github.com/cylc/cylc-flow/pull/5549#discussion_r1202501487

cylc/flow/scripts/lint.py
MetRonnie:
```suggestion
def check_jinja2_no_shebang(line, file, jinja_shebang=False, fallback=None, **kwargs):
```

cylc/flow/scripts/lint.py
```diff
@@ -88,156 +223,175 @@
...
-    re.compile(r'(?<!{)#.*?{[{%]'): {
-        'short': 'Cylc comments (#) do not prevent Jinja2 processing.',
+    "S011": {
+        'short': 'Cylc will process commented Jinja2!',
         'url': '',
-        'index': 11
+        'kwargs': True,
+        'evaluate commented lines': True,
+        FUNCTION: check_if_jinja2,
+        'fallback': re.compile(r'(?<!{)#.*?{[{%]').findall,
```
MetRonnie, line 364: Instead of this "fallback" approach could you not use `functools.partial`?
```suggestion
        FUNCTION: partial(
            check_if_jinja2,
            pattern=re.compile(r'(?<!{)#.*?{[{%]')
        ),
```

MetRonnie (approved):

hjoliver (approved):
👍

PR #5552, opened 2023-05-17
Use more sensible exception class for `detect_old_contact_file()`

MetRonnie:
This is a small change with no associated issue.

**Check List**

- [x] No dependency changes 
- [x] Tests are updated
- [x] Changelog not needed (pretty insignificant to users)
- [x] No docs update
- [x] Refactor on master branch only

### DIFF
--- a/cylc/flow/clean.py
+++ b/cylc/flow/clean.py
@@ -47,6 +47,7 @@ from typing import (
 from cylc.flow import LOG
 from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
 from cylc.flow.exceptions import (
+    ContactFileExists,
     CylcError,
     InputError,
     PlatformError,
@@ -124,7 +125,7 @@ def _clean_check(opts: 'Values', reg: str, run_dir: Path) -> None:
         return
     try:
         detect_old_contact_file(reg)
-    except ServiceFileError as exc:
+    except ContactFileExists as exc:
         raise ServiceFileError(
             f"Cannot clean running workflow {reg}.\n\n{exc}"
         )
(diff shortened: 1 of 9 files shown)


cylc/flow/network/client.py
```diff
@@ -147,7 +147,7 @@ def timeout_handler(self) -> None:
         # behind a contact file?
         try:
             detect_old_contact_file(self.workflow)
-        except (AssertionError, ServiceFileError):
+        except ContactFileExists:
```
hjoliver, line 150: Any idea why `AssertionError` was there?

cylc/flow/clean.py
```diff
@@ -124,7 +125,7 @@ def _clean_check(opts: 'Values', reg: str, run_dir: Path) -> None:
         return
     try:
         detect_old_contact_file(reg)
-    except ServiceFileError as exc:
+    except ContactFileExists as exc:
```
hjoliver, line 128: Not sure what we were thinking, using ServiceFileError in this context!

hjoliver (approved):

cylc/flow/network/client.py
MetRonnie: Did a quick git blame time travel, there used to be an `AssertionError` raised in `detect_old_contact_file()` but no more

cylc/flow/workflow_files.py
```diff
@@ -456,7 +457,7 @@ def detect_old_contact_file(
         CylcError:
             If it is not possible to tell for sure if the workflow is running
             or not.
-        ServiceFileError(CylcError):
+        ContactFileExists:
             If old contact file exists and the workflow process still alive.
```
oliver-sanders, line 462:
```suggestion
            If old contact file exists and the workflow process still alive.
        ServiceFileError:
            For corrupt / incompatible contact files.

```

oliver-sanders (approved):

PR #5556, opened 2023-05-24
lint: fix spurious lowercase family style warnings

### LINKED ISSUE #5558: cylc lint - S007 and task parameters
### Problem

I haven't checked, but I'm guessing the current implementation for S007 (lowercase family names) does not handle task parameters? My question is, should it gain the intelligence to expand task parameters and raise warnings if they are lowercase?

e.g. the below family `FOO` contains lowercase characters after expansion.

```
[task parameters]
    # parameters: "ship", "buoy", "plane"
    # default task suffixes: _ship, _buoy, _plane
    obs = ship, buoy, plane

[scheduling]
    [[graph]]
        R1 = "FOO<obs>:succeed-all => a"
```

### Proposed Solution

1. Look for FOO<obs> type syntax
2. Check what that expands to
3. If lowercase, raise a warning

But, with this approach, it means, if that task parameter is used in a task name, then the task would be a mixture of lowercase and uppercase letters. Typically I believe people tend to try to make tasks lowercase. So perhaps, task parameters could use an extension. I don't know the best way to do it, but something like

FOO<u:obs> would convert `obs` to uppercase, and `foo<l:obs>` could convert it to lowercase?

Or perhaps this is all not really worth pursuing as it raises these new areas of syntax complexity?

oliver-sanders:
Closes #5558

This rule should now permit:
* Parameters.
* Jinja2.
* Empy.
* `none`, `None`, `root`.

~Regex playground: https://regex101.com/r/gyRD8h/1~ outdated

Worth trying this out with some real workflows to see if there are any other surprises.

**Check List**

### DIFF
--- a/cylc/flow/scripts/lint.py
+++ b/cylc/flow/scripts/lint.py
@@ -129,10 +129,50 @@ STYLE_CHECKS = {
         'url': STYLE_GUIDE + 'trailing-whitespace',
         'index': 6
     },
-    # Look for families both from inherit=FAMILY and FAMILY:trigger-all/any.
-    # Do not match inherit lines with `None` at the start.
+    # Look for families in inherit= configurations.
     re.compile(
-        r'(inherit\s*=(?!\s*None\s*(?!.*[a-z])))|(\w[a-z]\w:\w+?-a(ll|ny))'
+        r'''
+          # match all inherit statements
+          ^\s*inherit\s*=
+          # filtering out those which match only valid family names
+          (?!
+            \s*
+            # none, None and root are valid family names
+            (none|None|root|
+              (
+                # as are families named with capital letters
+                [A-Z0-9_-]+
+                # which may include Cylc parameters
+                | (<[^>]+>)
+                # or Jinja2
+                | ({[{%].*[%}]})
+                # or EmPy
+                | (@[\[{\(]).*([\]\}\)])
+              )+
+            )
+            # this can be a comma separated list
+            (
+              \s*,\s*
+              # none, None and root are valid family names
+              (none|None|root|
+                (
+                  # as are families named with capital letters
+                  [A-Z0-9_-]+
+                  # which may include Cylc parameters
+                  | (<[^>]+>)
+                  # or Jinja2
+                  | ({[{%].*[%}]})
+                  # or EmPy
+                  | (@[\[{\(]).*([\]\}\)])
+                )+
+              )
+            )*
+            # allow trailing commas and whitespace
+            \s*,?\s*
+            $
+          )
+        ''',
+        re.X
     ): {
         'short': 'Family name contains lowercase characters.',
         'url': STYLE_GUIDE + 'task-naming-conventions',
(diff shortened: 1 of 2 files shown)


cylc/flow/scripts/lint.py
```diff
@@ -129,10 +129,50 @@
         'url': STYLE_GUIDE + 'trailing-whitespace',
         'index': 6
     },
-    # Look for families both from inherit=FAMILY and FAMILY:trigger-all/any.
```
oliver-sanders, line 132: Supporting this test in triggers turned out to be a little tricky.

cylc/flow/scripts/lint.py
```diff
@@ -129,10 +129,50 @@
...
+                (
+                  # as are families named with capital letters
+                  [A-Z0-9_-]+
+                  # which may include Cylc parameters
+                  | (<[^>]+>)
+                  # or Jinja2
+                  | ({[{%].*[%}]})
+                  # or EmPy
+                  | (@[\[{\(]).*([\]\}\)])
+                )+
```
oliver-sanders, line 167: Duplicate of the above logic (Regex doesn't support variables).

If re-used elsewhere we can refactor this out using string templating. Example here https://github.com/cylc/cylc-flow/blob/c5f177024b15e392fa1a0ba0f5c1d908fa6dc4f6/cylc/flow/id.py#L500

wxtim:
I  have found two places where this check returns a false positive:

```ini
# Presence of quotation marks negates other checks (occurs  3000 times)
inherit = 'SOMETHING<foo, bar>'

# Doesn't handle inline comments (< 100 instances of any given example of this)
inherit = FAM1 # Some comment
```

oliver-sanders:
Good spots!

Digging, I also found `inherit =` was being flagged erroneously too. Latest changes should fix this, added your cases to the tests.

wxtim (approved):
You've fixed my cases.

cylc/flow/scripts/lint.py
oliver-sanders: Not really helpful.

However, this is now outdated as these two patterns don't match.

MetRonnie (approved):
I have tested out and it successfully catches this case
```
inherit = FOO, \
          bar
```
but not
```
inherit = """
    FOO, 
    bar
"""
```
Although I don't think that matters much

oliver-sanders:
Catching the latter form via regex is going to be hard without parsing the whole document with parsec which would be too slow for text-editor integration and would likely error whilst the document is being written.

cylc/flow/scripts/lint.py
MetRonnie: These two are identical aren't they?

https://github.com/cylc/cylc-flow/blob/e540410d7dd1084d338206e07698d85771462ba5/cylc/flow/scripts/lint.py#L143-L154

https://github.com/cylc/cylc-flow/blob/e540410d7dd1084d338206e07698d85771462ba5/cylc/flow/scripts/lint.py#L161-L172

oliver-sanders: It's the line above containing none, None and root which now differs.

Issue #5551, opened 2023-05-17
job.sh: scaling limit with CYLC_TASK_DEPENDENCIES


oliver-sanders:
There is an artificial scaling limit to the number of dependencies that a task can have.

The limit is fairly high, however, is dependent on the length of task names / cycle point format so hard to quantise.

### Description

There is an upper limit to how long an environment variable can be after which point random commands will start failing with the cryptic error message "Argument list too long".

https://stackoverflow.com/questions/28865473/setting-environment-variable-to-a-large-value-argument-list-too-long

Because Cylc lists all dependencies (or should that be prerequisites?) in the environment variable `CYLC_TASK_DEPENDENCIES`, we can hit this limit where scaling gets crazy.

### Reproducible Example

```ini
abcdefghijklmnopqrstuvwxyz_1234567890<a> => b
```

Increase `a` until failures occur.

Test with `cylc play <id> --mode=simulation`.

### Solution

This variable is required for normal function, fairly new and has limited usage so we have various options here.

1. Don't set this variable when the list of dependencies is too high.
2. Don't provide this information in an environment variable in the first place.
3. Only provide this environment variable when referenced in the *script setting.
4. Only provide this environment variable when explicitly requested.
5. ...

hjoliver:
> This variable is required for normal function, ...

Do you mean NOT required for normal function?

PR #5560, opened 2023-05-30
job: redact the CYLC_TASK_DEPENDENCIES variable if too long

oliver-sanders:
* Long environment variables can eat into the shell's address space causing subsequent commands to fail.
* Redact the `CYLC_TASK_DEPENDENCIES` variable when there are more than 50 upstream tasks.
* If a task cycle point is ~25 chars and the name up to ~75 chars then that's 100 chars per task x 50 tasks => 50'000 chars which should be within the limit (~126'000 on my system).
* Closes https://github.com/cylc/cylc-flow/issues/5551

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc/pull/613) pull request opened if required at [cylc/cylc-doc/pull/613](https://github.com/cylc/cylc-doc/pull/613).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,13 @@ ones in. -->
 -------------------------------------------------------------------------------
 ## __cylc-8.1.5 (<span actions:bind='release-date'>Upcoming</span>)__
 
+### Breaking Changes
+
+[#5600](https://github.com/cylc/cylc-flow/pull/5600) -
+The `CYLC_TASK_DEPENDENCIES` environment variable will no longer be exported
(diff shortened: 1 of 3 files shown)


hjoliver (approved):

hjoliver:
LGTM, and I guess it is necessary at the moment (and as you noted on the issue, as a new variable in Cylc 8 no one will be relying on it yet). Maybe we should think about what to do if anyone does want to rely on it in the future though.

wxtim (approved):

Issue #5215, opened 2022-11-03
data store: task outputs are not re-loaded on reload


oliver-sanders:
(requires confirmation) steps to reproduce:

* Write a workflow with a task which produces a custom output.
* Run the workflow, holding a downstream task of the one which produces the custom output (to keep it in the n=1 window).
* Inspect the tasks outputs, they are displayed in the job row in the tree view and can be queried from graphiql.
* Reload the workflow.

Lower priority bug, hence the milestone, can slip if needed.

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

dwsutherland:
Yes, makes sense this is the case.. I assume the job is no longer in the active pool, so it probably needs loaded from the DB?

oliver-sanders:
Makes sense, could we avoid wiping the store on reload or at least just avoid wiping `n=1` completed tasks that remain `n=1` post-reload to avoid needing to go back to the DB?

dwsutherland:
> Makes sense, could we avoid wiping the store on reload or at least just avoid wiping `n=1` completed tasks that remain `n=1` post-reload to avoid needing to go back to the DB?

I had toyed with this idea, but we have to be happy with the fact the outputs (and/or other fields) may have been the subject of the reload.

I think I'd have to move the `self.data[w_id]` to `self.pre_reload[w_id]` temporarily and only target finished/completed and/or non-active parents..

oliver-sanders:
> the outputs (and/or other fields) may have been the subject of the reload.

Ah yeah, that makes sense.

Issue #2788, opened 2018-10-09
Traceback restarting simulation mode suite


oliver-sanders:
Suites run in simulation mode will crash if stopped with `--now --now` and restarted:

```
cylc run $SUITE_NAME --mode=simulation
sleep 5
cylc stop $SUITE_NAME --now --now
sleep 5
cylc restart $SUITE_NAME --mode=simulation
```

Example traceback:

```
2018-10-09T08:29:30Z CRITICAL - Traceback (most recent call last):
	  File "cylc/lib/cylc/scheduler.py", line 243, in start
	    self.configure()
	  File "cylc/lib/cylc/scheduler.py", line 394, in configure
	    self.load_tasks_for_restart()
	  File "cylc/lib/cylc/scheduler.py", line 462, in load_tasks_for_restart
	    self.pool.load_db_task_pool_for_restart, self.options.checkpoint)
	  File "cylc/lib/cylc/rundb.py", line 745, in select_task_pool_for_restart
	    callback(row_idx, list(row))
	  File "cylc/lib/cylc/task_pool.py", line 367, in load_db_task_pool_for_restart
	    itask.task_owner, itask.task_host = user_at_host.split(
	AttributeError: 'NoneType' object has no attribute 'split'
```

Sadly it's not as trivial as just fixing `lib/cylc/task_pool.py#367` as `NoneType` errors will crop up elsewhere in the `TaskPool` as well.

hjoliver:
Hmm, possibly never tried restarting a sim mode suite (my fault, I expect).

matthewrmshin:
Looking at the logic. This is the only value from table `task_jobs` that is used in such a way. All others are protected by `if value:` blocks. The block surrounding the problematic value already catches `ValueError`, so it can catch `AttributeError` as well.

(The key to trigger this bug is to have a task in the submitted or running state on shutdown.)

matthewrmshin:
(Very easy to fix, but not easy to write a test.)

PR #2157, opened 2017-02-10
scheduler, task_pool, task_proxy refactor

matthewrmshin:
The main drives of this change are:
* Efficiency.
* Reduce strongly coupled logic where possible.

Some headlines:
* Reduce number of attributes in `cylc.task_proxy` and related objects.
* Reduce line counts in `cylc.scheduler`, `cylc.task_pool` and `cylc.task_proxy`.
* Remove `get_task_proxy`.
  * Replaced by `SuiteConfig.get_taskdef` and constructor of `TaskProxy`.
* Singletons related to job management are now instantiated normally. These include:
  * Batch system manager.
  * Job file writer.
  * Job host initialisation manager.
  * Multi-processing pool for job commands, suite events and task events.
* New module `cylc.task_job_mgr`: job management logic from `cylc.task_pool` and `cylc.task_proxy`.
* New module `cylc.suite_events`: suite events logic from `cylc.scheduler`.
* New module `cylc.task_events_mgr`:
  * Task message and task event handler logic in `cylc.scheulder` and `cylc.task_proxy`.
  * Manage task event handlers: these no longer hang onto their originating task proxy objects. The new logic allows task pool to clean up tasks without having to worry about their incomplete event handlers.
* New module `cylc.suite_db_mgr`:
  * Database management logic from `cylc.scheduler`, `cylc.task_pool` and `cylc.task_proxy`.
  * Manage queueing of database operations - improve efficiency by using a single set of queues.
  * Manage database recovery, clean up, etc.
* Rationalise task state reset logic, by reducing the number of entry points to the `reset_state` and `set_state` methods. Apart from `hold` and `release`, all state change will now call `reset_state`.
* Close #1550. Rewrite `cylc.task_outputs` to allow manual reset of individual outputs to complete/incomplete. (Outputs can be referenced by either the trigger names or by the message strings.)
* Close #1182. `cylc submit` can now submit multiple families and tasks in a single command.

[HJO: Close #2228.]

PR #5562, opened 2023-06-01
Prevent task events manager logging the same message more than once

### LINKED ISSUE #5526: Polling causes duplicate messages
### Description

Polling a task that uses `cylc message` will result in those messages coming through again even if they have already been sent.

This is most obviously seen in the UI tree view.

### Expected Behaviour

(Need to define properly)

wxtim:
Closes #5526 

Deals with a bug was returning every message ever sent by a task on each poll.

(Draft pending further checks, to ensure that I haven't broken anything)

**Check List**

- [x] Tests are included 
- [x] ~Update docs~  - this is a change to Cylc's logging.


wxtim:
@metronnie - I think that I fixed it.

wxtim:
The original approach (adding messages to the activity log file and checking for them later seemed to break some of the workflow restart tests in ways that make me un-confident that they are safe changes.

Scraping the log is horrible.

Next thing to try: using the database.

cylc/flow/task_events_mgr.py
```diff
@@ -819,6 +829,31 @@ def _process_message_check(
...
+            self.workflow,
+            list(itask.tokens.values())[0],
+            list(itask.tokens.values())[1],
+            submit_num
+        ))
+        if not job_activity_log.exists():
+            job_activity_log.parent.mkdir(exist_ok=True, parents=True)
+            job_activity_log.touch()
+
+        # Record recieved messages in the job activity log.
```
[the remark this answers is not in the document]
wxtim: I had thought that this PR had resolved all doubts in my mind about this word... 😞

cylc/flow/task_events_mgr.py
```diff
@@ -819,6 +829,31 @@ def _process_message_check(
...
+        # Do not log it a second time if Polling reveals the same
+        # message has been logged already by reciept method:
+        loggable_message = f'{flag} {itask.identity} {message} {event_time}'
+        if flag == self.FLAG_RECEIVED:
+            with open(os.path.expandvars(job_activity_log), "ab") as handle:
+                handle.write((loggable_message + '\n').encode())
```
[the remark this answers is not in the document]
wxtim: Yes

cylc/flow/task_events_mgr.py
```diff
@@ -587,27 +588,36 @@ def process_message(
...
-        self.data_store_mgr.delta_job_msg(
-            itask.tokens.duplicate(job=str(submit_num)).relative_id,
-            new_msg
-        )
 
         # Satisfy my output, if possible, and spawn children.
         # (first remove signal: failed/EXIT -> failed)
-
         msg0 = message.split('/')[0]
         completed_trigger = itask.state.outputs.set_msg_trg_completion(
```
oliver-sanders, line 609: I don't think this should be run for duplicate messages.

cylc/flow/task_events_mgr.py
```diff
@@ -587,27 +588,36 @@ def process_message(
...
+            # haven't already:
+            self.data_store_mgr.delta_job_msg(
+                itask.tokens.duplicate(job=str(submit_num)).relative_id,
+                new_msg
+            )
+
+            self.data_store_mgr.delta_task_output(itask, msg0)
 
         # Check the `started` event has not been missed e.g. due to
         # polling delay
```
oliver-sanders, line 623: I don't think this should be run for duplicate messages either.

cylc/flow/task_events_mgr.py
```diff
@@ -819,6 +830,31 @@ def _process_message_check(
...
+        # message has been logged already by reciept method:
+        loggable_message = f'{flag} {itask.identity} {message} {event_time}'
+        if flag == self.FLAG_RECEIVED:
+            with open(os.path.expandvars(job_activity_log), "a") as handle:
+                handle.write(loggable_message + '\n')
+        elif flag == self.FLAG_POLLED:
+            with open(os.path.expandvars(job_activity_log), "r") as handle:
+                log = handle.read()
+                if f'{itask.identity} {message} {event_time}' in log:
+                    return 'quiet'
```
oliver-sanders, line 856: Should this be `return None`?

cylc/flow/task_events_mgr.py
wxtim: The whole of the rest of the method or just this check?

cylc/flow/task_events_mgr.py
MetRonnie: I think he's asking if the behaviour around L600 could be controlled by a value of `None` rather than a string

wxtim: Well, yes, it could be, but I think the logic makes more sense if we return an explict (and truthy) value.

oliver-sanders: If this returns `None` then the rest of the previous function would be skipped by the `return` which feels like the behaviour we are after here.

cylc/flow/task_events_mgr.py
```diff
@@ -819,6 +830,31 @@ def _process_message_check(
...
+
+        # Get job activity log to check whether a message is a repeat.
+        job_activity_log = Path(get_task_job_activity_log(
+            self.workflow,
+            itask.tokens['cycle'],
+            itask.tokens['task'],
+            submit_num
+        ))
+        if not job_activity_log.exists():
+            job_activity_log.parent.mkdir(exist_ok=True, parents=True)
```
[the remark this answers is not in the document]
wxtim: It happens in `tests/functional/restart/33*.t`:  I'm not sure I understand the scenario under test well enough to summarize it - but it references https://github.com/cylc/cylc-flow/issues/2788

cylc/flow/task_events_mgr.py
oliver-sanders: Order of events:

* Task submitted (received)
* Task running (received)
* Workflow shut down
* Task hits ETL
* Workflow restarted
* Task failed (polled)

The task failed message is only received once (on restart) so this shouldn't be related?

oliver-sanders:
> @oliver-sanders - I've left this in a draft state - could I have your thoughts on the issue around restart?

I don't see what the issue is with restart? The intent of this PR is to prevent task messages being counted more than once. It shouldn't matter whether a restart happened in between two messages.

wxtim:
> > @oliver-sanders - I've left this in a draft state - could I have your thoughts on the issue around restart?
> 
> I don't see what the issue is with restart? The intent of this PR is to prevent task messages being counted more than once. It shouldn't matter whether a restart happened in between two messages.

No, I agree now, I just forgot to un-draft after I got it working.

oliver-sanders:
Good news, messages are stored in the DB along with timestamps so we don't need to use the job-activity.log to hold this info.

Just make sure that queued database operations get processed between calls of whatever calls this code (presumably `Scheduler.process_queued_task_messages`) so that task messages are committed to the DB from the last batch before we process the next batch. This is probably already the case, but well worth checking.

cylc/flow/rundb.py
```diff
@@ -1077,3 +1077,19 @@ def select_jobs_for_datastore(
...
+            r'SELECT * FROM task_events WHERE'
+            fr' name="{name}"'
+            fr' AND cycle="{cycle}"'
+            fr' AND message="{message}"'
+            fr' AND submit_num="{submit}"'
+            fr' AND time="{timestamp}"'
+        )
+        conn = self.connect()
+        previous_message = any(i for i in conn.execute(stmt))
+        conn.close()
```
oliver-sanders, line 1094: I'm not sure we should close the DB here?

Follow the pattern of the "select_*" methods in this file.

cylc/flow/scheduler.py
```diff
@@ -1623,6 +1623,7 @@ async def main_loop(self) -> None:
             self.broadcast_mgr.expire_broadcast(self.pool.get_min_point())
             self.late_tasks_check()
 
+            self.process_workflow_db_queue()
```
MetRonnie, line 1626: Why this extra call?

cylc/flow/task_events_mgr.py
```diff
@@ -587,8 +587,16 @@ def process_message(
         # Any message represents activity.
         self.reset_inactivity_timer_func()
 
-        if not self._process_message_check(
-                itask, severity, message, event_time, flag, submit_num):
+        with self.workflow_db_mgr.get_pri_dao() as pri_dao:
+            message_already_received = pri_dao.message_in_db(
+                submit_num, message, event_time, itask)
```
MetRonnie, line 592: This context manager automatically closes the DB. But do we need to do that here? cc @oliver-sanders 

If not it can just be accessed like
```suggestion
        message_already_received = self.workflow_db_mgr.pri_dao.message_in_db(
            submit_num, message, event_time, itask)
```

cylc/flow/task_proxy.py
```diff
@@ -263,10 +263,18 @@ def __repr__(self) -> str:
...
     def __str__(self) -> str:
-        """Stringify with tokens, state, submit_num, and flow_nums."""
+        return self.get_str()
+
+    def get_str(self, state=True) -> str:
+        """Stringify with tokens, state, submit_num, and flow_nums.
+
+        Args:
+            state: Include/omit state from string.
+        """
```
MetRonnie, line 273: This looks unused other than in `__str__()` from a quick search

tests/integration/test_task_events_mgr.py
```diff
@@ -42,3 +43,33 @@ async def test_process_job_logs_retrieval_warns_no_platform(
...
         assert 'Unable to retrieve' in warning.msg
+
+
+async def test_process_message_no_repeat(
+    one_conf: Fixture, flow: Fixture, scheduler: Fixture, run: Fixture
+):
+    """Don't log received messages if they are found again when."""
+    reg: str = flow(one_conf)
+    schd: 'Scheduler' = scheduler(reg, paused_start=True)
+    message: str = 'The dead swans lay in the stagnant pool'
```
MetRonnie, line 54: Charming 😆

wxtim: (Hitchikers Reference - The worst poetry in the Universe)

cylc/flow/rundb.py
```diff
@@ -1077,3 +1077,19 @@ def select_jobs_for_datastore(
...
+        stmt = r"""
+            SELECT * FROM task_events WHERE
+            name == ?
+            AND cycle == ?
+            AND message == ?
+            AND submit_num == ?
+            AND time == ?
+        """
+        stmt_args = [task_name, cycle, message, submit, timestamp]
+        return any(i for i in self.connect().execute(stmt, stmt_args))
```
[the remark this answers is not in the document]
wxtim: Really Nice!

cylc/flow/scheduler.py
wxtim: Need to ensure we've processed anything pending in the workflow DB queue otherwise this logic might not work if two messages are close together.

cylc/flow/task_proxy.py
wxtim: Well noticed - this is garbage left over from the earlier approach of scraping `job-activity.log`.

cylc/flow/task_events_mgr.py
oliver-sanders: Yes, we should not be closing the DB here.

No need to use the context manager when accessing the DB from the Scheduler process itself, the Scheduler has built-in logic to manage DB connections.

cylc/flow/rundb.py
```diff
@@ -1085,3 +1085,21 @@ def select_jobs_for_datastore(
...
+                SELECT 1 FROM task_events WHERE
+                    name == ?
+                    AND cycle == ?
+                    AND message == ?
+                    AND submit_num == ?
+                    AND time == ?
+            )
+        """
+        stmt_args = [task_name, cycle, message, submit, timestamp]
+        return bool(self.connect().execute(stmt, stmt_args).fetchone())
```
MetRonnie, line 1105: If using `SELECT EXISTS` it's returning either `(1,)` or `(0,)`
```suggestion
        return bool(self.connect().execute(stmt, stmt_args).fetchone()[0])
```

MetRonnie (commented):
Last thing is that the `started (polled)` still piles up in the UI if you poll repeatedly
image

wxtim:
> Last thing is that the `started (polled)` still piles up in the UI if you poll repeatedly image

Rah!

cylc/flow/rundb.py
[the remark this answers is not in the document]
wxtim: For future reference 
```sql
SELECT 1 FROM task_events
```
always returns "1"!

Correct answer is 

```sql
SELECT * FROM table WHERE conditions LIMIT 1
```

cylc/flow/rundb.py
```diff
@@ -1085,3 +1085,27 @@ def select_jobs_for_datastore(
...
+                SELECT * FROM task_events WHERE
+                    name == ?
+                    AND cycle == ?
+                    AND (message == ? OR event == ?)
+                    AND submit_num == ?
+                LIMIT 1
+            )
+        """
+        stmt_args = [task_name, cycle, message, message, submit, timestamp]
+        return bool(self.connect().execute(stmt, stmt_args[:-1]).fetchone()[0])
```
MetRonnie, line 1111: Wasn't my suggestion that was the problem was it? Only the missing `OR event == ?`?
```suggestion
                SELECT 1 FROM task_events WHERE
                    name == ?
                    AND cycle == ?
                    AND (message == ? OR event == ?)
                    AND submit_num == ?
            )
        """
        stmt_args = [task_name, cycle, message, message, submit]
        return bool(self.connect().execute(stmt, stmt_args).fetchone()[0])
```

MetRonnie (commented):
I've noticed that the polled messages on restart do not show anymore (screenshot is from master branch)

image

oliver-sanders:
For the tests/f/restart/58-removed-task.t test failure you're clear to bodge the test to accept the non-polled message as it isn't trying to test polling logic, it's trying to test task_pool logic.

cylc/flow/rundb.py
MetRonnie: Working as expected for me

oliver-sanders:
> I've noticed that the polled messages on restart do not show anymore (screenshot is from master branch). Is that desirable?

That is not desirable, however, it's a different problem - https://github.com/cylc/cylc-flow/issues/5215

The data store needs to load outputs from the database on restart/reload.

cylc/flow/rundb.py
```diff
@@ -1081,3 +1081,26 @@ def select_jobs_for_datastore(
...
+            SELECT EXISTS (
+                SELECT 1 FROM task_events WHERE
+                    name == ?
+                    AND cycle == ?
+                    AND (message == ? OR event == ?)
+                    AND submit_num == ?
+            )
+        """
+        stmt_args = [task_name, cycle, message, message, submit, timestamp]
+        return bool(self.connect().execute(stmt, stmt_args[:-1]).fetchone()[0])
```
MetRonnie, line 1106: Timestamp no longer used? (Needs removing in the function definition + calls also)
```suggestion
        stmt_args = [task_name, cycle, message, message, submit]
        return bool(self.connect().execute(stmt, stmt_args).fetchone()[0])
```

wxtim: That shouldn't have happened - and I think that there's another problem I've not sorted out.

cylc/flow/task_events_mgr.py
```diff
@@ -843,10 +857,10 @@ def _custom_handler_callback(self, ctx, schd, id_key):
         else:
             self.unset_waiting_event_timer(id_key)
 
-    def _db_events_insert(self, itask, event="", message=""):
+    def _db_events_insert(self, itask, event_time, event="", message=""):
         """Record an event to the DB."""
         self.workflow_db_mgr.put_insert_task_events(itask, {
-            "time": get_current_time_string(),
+            "time": event_time,
```
wxtim, line 863: I can't see any reason that this should be got at the time of execution and not fed down from the calling method. Without this change, however, the event time in the table may not match the event time returned by poll.

oliver-sanders: IMO this is a bugfix!

oliver-sanders: What is `event_time`, is it an ISO8601 string, or an integer epoch time? I've gotta feeling that it might be both (search for now in the diff).

cylc/flow/rundb.py
```diff
@@ -1081,3 +1081,27 @@ def select_jobs_for_datastore(
     def vacuum(self):
         """Vacuum to the database."""
         return self.connect().execute("VACUUM")
+
+    def message_in_db(self, submit, message, timestamp, itask):
```
oliver-sanders, line 1085: [minor] For consistency with other methods consider:

```
def message_in_db(self, itask, event_time, submit, message):
```

cylc/flow/task_events_mgr.py
wxtim: It _is_ _either_ - 

> IMO this is a bugfix!

I was hoping that was how you'd see it.

cylc/flow/rundb.py
```diff
@@ -1081,3 +1081,26 @@ def select_jobs_for_datastore(
...
+            SELECT EXISTS (
+                SELECT 1 FROM task_events WHERE
+                    name == ?
+                    AND cycle == ?
+                    AND (message == ? OR event == ?)
+                    AND submit_num == ?
+            )
+        """
+        stmt_args = [task_name, cycle, message, message, submit, event_time]
+        return bool(self.connect().execute(stmt, stmt_args[:-1]).fetchone()[0])
```
MetRonnie, line 1106: `event_time` is unused

cylc/flow/task_events_mgr.py
```diff
@@ -587,8 +589,18 @@ def process_message(
         # Any message represents activity.
         self.reset_inactivity_timer_func()
```
MetRonnie, line 591: Suggest that the `event_time` arg is made non-optional (not a kwarg)

cylc/flow/rundb.py
wxtim: this is a mistaken regression, and it should be - I've put it back.

cylc/flow/task_events_mgr.py
wxtim: Sounds legit - .

cylc/flow/rundb.py
MetRonnie: Why didn't the tests fail if `event_time` is needed?

cylc/flow/task_events_mgr.py
wxtim: I've tried it, but actually, not a good idea, because it involved moving a call to `get_current_time_string` in many, many places.

cylc/flow/rundb.py
wxtim: The tests are not as thorough as they could be. I tested several cases, what I did not do is 
```
for arg in args:
  arg = 'garbage'
  assert function(*args) == False
```

I'll have a poke at that.

cylc/flow/task_events_mgr.py
MetRonnie: Ah my bad I missed the
```py
        if event_time is None:
            event_time = get_current_time_string()
```
above

tests/unit/test_rundb.py
```diff
@@ -175,3 +176,81 @@ def callback(index, row):
...
+        VALUES
+            (
+                "qux", "-191",
+                "morning", "1", "started", ""),
+            (
+                "qux", "-190",
+                "afternoon", "1", "message critical", "Hello Rome");
+    """
+    dao.connect().execute(setup_stmt)
+    yield dao
```
MetRonnie, line 198: Closing DB may avoid any possible flakiness
```suggestion
    setup_stmt = r"""
        INSERT INTO task_events
        VALUES
            (
                "qux", "-191",
                "morning", "1", "started", ""),
            (
                "qux", "-190",
                "afternoon", "1", "message critical", "Hello Rome");
    """
    with CylcWorkflowDAO(db_file, create_tables=True) as dao:
        dao.connect().execute(setup_stmt)
        yield dao
```

tests/functional/restart/28-execution-timeout.t
```diff
@@ -17,17 +17,12 @@
...
 
 run_ok "${TEST_NAME_BASE}-validate" cylc validate "${WORKFLOW_NAME}"
 workflow_run_ok "${TEST_NAME_BASE}-run" cylc play "${WORKFLOW_NAME}" --no-detach
 workflow_run_ok "${TEST_NAME_BASE}-restart" \
     cylc play "${WORKFLOW_NAME}" --no-detach --reference-test
-contains_ok "${WORKFLOW_RUN_DIR}/log/job/1/foo/NN/job-activity.log" <<'__LOG__'
-[(('event-handler-00', 'execution timeout'), 1) cmd] echo 1/foo 'execution timeout'
-[(('event-handler-00', 'execution timeout'), 1) ret_code] 0
-[(('event-handler-00', 'execution timeout'), 1) out] 1/foo execution timeout
-__LOG__
```
MetRonnie, line 31: Why is this removed from the test?

wxtim: Because it no longer gets logged. According to @oliver-sanders that's OK.

cylc/flow/rundb.py
```diff
@@ -1081,3 +1081,27 @@ def select_jobs_for_datastore(
     def vacuum(self):
         """Vacuum to the database."""
         return self.connect().execute("VACUUM")
+
+    def message_in_db(self, itask, event_time, submit, message):
```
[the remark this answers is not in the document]
wxtim: Your view is the evidence I need to make up my mind for `submit_num`.

CHANGES.md
```diff
@@ -114,6 +114,10 @@ resuming a workflow.
 [#5625](https://github.com/cylc/cylc-flow/pull/5625) - Exclude `setuptools`
 version (v67) which results in dependency check failure with editable installs.
 
+[#5652](https://github.com/cylc/cylc-flow/pull/5562) - Prevent the same
+message being logged multiple times when polled.
+
+
```
MetRonnie, line 120: Needs to be moved now. I would suggest writing the description to `changes.d/5562.fix.md` in anticipation of #5634

wxtim: It's waiting on Hilary, who will presumably want to have a think about town crier (unless he doesn't). I'll update in this format for now.

MetRonnie: IMO it's easier to do towncrier fragments now and switch back to CHANGES.md later in the chance we don't proceed with towncrier.

MetRonnie:
Actually I've just tested this with the UI and unfortunately no messages come through while the workflow is running, but when you stop it, duplicate messages can come through

wxtim:
> Actually I've just tested this with the UI and unfortunately no messages come through while the workflow is running, but when you stop it, duplicate messages can come through

As it currently is, or with your changes?

CHANGES.md
MetRonnie: #5634 has been merged anyway so please do convert to a towncrier fragment

cylc/flow/__init__.py
```diff
@@ -53,7 +53,7 @@ def environ_init():
 
 environ_init()
 
-__version__ = '8.2.1.dev'
+__version__ = '8.3.0.dev'
```
wxtim, line 56:
```suggestion
__version__ = '8.2.1.dev'
```

MetRonnie:
As it is currently. I have been testing with this workflow
```
[scheduling]
    [[graph]]
        R1 = foo
[runtime]
    [[foo]]
        script = sleep 15; cylc message -- 'MEOW: Hello world'; sleep 600
```
1. Play
2. After 15 secs "Hello World" chip should show in tree view
3. Poll a couple of times to see if message is repeated
4. Stop `--now`
5. Restart, message should show up again (once) as polled
6. Poll again
7. Stop `--kill`

wxtim:
@MetRonnie 

> As it currently is. 

With all the changes since we started this convo?

MetRonnie:
Not since you've rebased (btw if you could avoid rebasing it would make it easier to tell what changes you're making in between my reviews)

oliver-sanders:
(needs rebase onto 8.2.x)

MetRonnie:
> 5\. Restart, message should show up again (once) as polled

Not happening at the moment

wxtim:
> > 5. Restart, message should show up again (once) as polled
> 
> Not happening at the moment

I didn't think we were bothered about _how_ the message was transmitted, so we don't want a second chip if polled.

MetRonnie:
But we do want *one* chip, which is not the case on restart for a ongoing running task?

Also I've noticed this PR does not stop duplicate entries piling up in the DB when polling
```
$ select * from task_events
foo5|1|2023-08-04T10:43:29+01:00|1|submitted|
foo5|1|2023-08-04T10:43:31+01:00|1|started|
foo5|1|2023-08-04T10:43:46+01:00|1|message meow|Hello world
foo5|1|2023-08-04T10:43:46+01:00|1|message meow|Hello world
foo5|1|2023-08-04T10:43:46+01:00|1|message meow|Hello world
```

oliver-sanders:
> But we do want one chip, which is not the case on restart for a ongoing running task?

I've lost track of the example this comment is coming from, but note that this PR *will* result in some message chips disappearing from the GUI on restart/realod. This isn't desired behaviour, but it is the correct logic!

The true source of the problem is here: https://github.com/cylc/cylc-flow/issues/5215, fixing this issue will restore the chip.

MetRonnie:
Ok, other than duplicate entries in `task_events` DB table, I think this is good to go

cylc/flow/task_events_mgr.py
```diff
@@ -819,17 +838,21 @@ def _process_message_check(
             self.EVENT_SUBMIT_FAILED, f'{FAIL_MESSAGE_PREFIX}ERR'
         }:
             severity = DEBUG
-        LOG.log(severity, f"[{itask}] {flag}{message}{timestamp}")
+
+        if not quiet:
+            LOG.log(severity, f"[{itask}] {flag}{message}{timestamp}")
```
MetRonnie, line 843: Can move the severity calculation just above inside this `if not quiet` block

MetRonnie:
After discussing with Oliver we think there is still some unnecessary processing of already-received messages going on. I've opened a PR against this at https://github.com/wxtim/cylc/pull/57

wxtim:
> Update: closed that as it does not seem feasible without a major refactor

Not entirely surprised, but thank you for attempting it.

cylc/flow/task_events_mgr.py
wxtim: Done locally - will push

tests/functional/cylc-poll/15-job-st-file-no-batch.t
```diff
@@ -30,5 +30,5 @@ run_ok "${TEST_NAME_BASE}-log-1" \
 run_ok "${TEST_NAME_BASE}-log-2" \
     grep -E '1/t1 running .*\(polled\)failed' "${LOG}"
 
-purge
+# purge
 exit
```
MetRonnie, line 34:
```suggestion
purge
```

tests/functional/restart/58-removed-task.t
```diff
@@ -39,10 +39,9 @@ workflow_run_ok "${TEST_NAME}" cylc play --no-detach "${WORKFLOW_NAME}"
 TEST_NAME="${TEST_NAME_BASE}-restart"
 workflow_run_ok "${TEST_NAME}" cylc play --set="INCL_B_C=False" --no-detach "${WORKFLOW_NAME}"
 
-grep_workflow_log_ok "grep-3" "\[1/a running job:01 flows:1\] (polled)started"
-grep_workflow_log_ok "grep-4" "\[1/b failed job:01 flows:1\] (polled)failed"
+grep_workflow_log_ok "grep-3" "=> failed"
```
MetRonnie, line 42: Having read what the test is trying to do I think this would be more explicit
```suggestion
grep_ok '\[jobs-poll out\] .* "run_status": 0'  "${WORKFLOW_RUN_DIR}/log/job/1/a/NN/job-activity.log"
grep_ok '\[jobs-poll out\] .* "run_status": 1'  "${WORKFLOW_RUN_DIR}/log/job/1/b/NN/job-activity.log"
grep_workflow_log_ok "grep-3" "\[1/b .* => failed"
```

tests/functional/restart/58-removed-task.t
```diff
@@ -39,10 +39,9 @@ workflow_run_ok "${TEST_NAME}" cylc play --no-detach "${WORKFLOW_NAME}"
...
 
 # Failed (but not incomplete) task c should not have been polled.
-grep_fail "\[1/c failed job:01 flows:1\] (polled)failed" "${WORKFLOW_RUN_DIR}/log/scheduler/log"
+grep_fail "\[1/c failed job:01 flows:1\] failed" "${WORKFLOW_RUN_DIR}/log/scheduler/log"
```
MetRonnie, line 45:
```suggestion
grep_fail "jobs-poll" "${WORKFLOW_RUN_DIR}/log/job/1/c/NN/job-activity.log"
```

tests/functional/restart/58-removed-task.t
```diff
@@ -22,7 +22,7 @@
 
 . "$(dirname "$0")/test_header"
 
-set_test_number 7
+set_test_number 6
```
MetRonnie, line 25:
```suggestion
set_test_number 8
```

MetRonnie (commented):
Got some suggestions for the tests and then I think this is good to go!

wxtim:
@oliver-sanders  are you waiting on me to do something with this?

oliver-sanders:
Kinda, but I need to take a closer look. I think this isn't skipping enough of the method, e.g. it appears to running event handlers and spawning logic for duplicate messages? But I know that when you tried moving the `return` higher up the function it caused problems.

wxtim:
I've had another look at this recently, and the notes on Matt Shin's original logic says

>            # Unhandled messages. These include:
>            #  * general non-output/progress messages
>            #  * poll messages that repeat previous results
>            # Note that all messages are logged already at the top.

_https://github.com/cylc/cylc-flow/pull/2157/files_

So it looks like Matt intended to have it work more simply than this and it's broken at some point. I'm going to reexamine this. :)

wxtim:
Very stale. Link to issue remains. Closing for PR management.

PR #5570, opened 2023-06-06
Fix runtime environment (scheduler, validation, etc.)

hjoliver:
Currently in `config.py` we infer a run directory, and work and share sub-directories, and export associated environment variables, even when parsing a source workflow (which may not be installed) or when parsing an installed workflow but not running the scheduler (in which case work etc. may not exist).

The original motivation for this PR was:
 *There are use cases where src-dir needs to be distinguished from run-dir at config parsing time.  Example: dependency graph constructed on the fly according to the content of some workflow sub-dir.  For validation of the source workflow, the location of this content is relative to the source dir; for validation of the installed workflow (or at scheduler start-up) it must be relative to the run directory.*

#5589 now provides a better solution to that problem, but in any case the exported environment should be consistent with the workflow location (as it is accessible to the template processor during parsing).

**Check List**

- [x] ~Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).~
- [ ] ~If this is a bug fix, PR should be raised against the relevant `?.?.x` branch~. (we're currently aiming all changes at 8.2.0)

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -74,13 +74,7 @@ from cylc.flow.param_expand import NameExpander
 from cylc.flow.parsec.exceptions import ItemNotFoundError
 from cylc.flow.parsec.OrderedDict import OrderedDictWithDefaults
 from cylc.flow.parsec.util import replicate
-from cylc.flow.pathutil import (
-    get_workflow_run_dir,
-    get_workflow_run_scheduler_log_dir,
-    get_workflow_run_share_dir,
-    get_workflow_run_work_dir,
-    get_workflow_name_from_id
-)
+from cylc.flow.pathutil import get_workflow_name_from_id
 from cylc.flow.platforms import FORBIDDEN_WITH_PLATFORM
 from cylc.flow.print_tree import print_tree
 from cylc.flow.subprocctx import SubFuncContext
@@ -106,6 +100,7 @@ from cylc.flow.workflow_files import (
     NO_TITLE,
     WorkflowFiles,
     check_deprecation,
+    is_installed,
 )
 from cylc.flow.xtrigger_mgr import XtriggerManager
 
@@ -247,13 +242,12 @@ class WorkflowConfig:
         self.mem_log("config.py:config.py: start init config")
         self.workflow = workflow  # workflow id
         self.workflow_name = get_workflow_name_from_id(self.workflow)
-        self.fpath = str(fpath)  # workflow definition
-        self.fdir = os.path.dirname(fpath)
-        self.run_dir = run_dir or get_workflow_run_dir(self.workflow)
-        self.log_dir = (log_dir or
-                        get_workflow_run_scheduler_log_dir(self.workflow))
-        self.share_dir = share_dir or get_workflow_run_share_dir(self.workflow)
-        self.work_dir = work_dir or get_workflow_run_work_dir(self.workflow)
+        self.fpath: Path = Path(fpath)  # workflow definition
+        self.fdir = str(self.fpath.parent)
+        self.run_dir = run_dir
+        self.log_dir = log_dir
+        self.share_dir = share_dir
+        self.work_dir = work_dir
         self.options = options
         self.implicit_tasks: Set[str] = set()
         self.edges: Dict[
@@ -890,7 +884,9 @@ class WorkflowConfig:
             )
         # Allow implicit tasks in back-compat mode unless rose-suite.conf
         # present (to maintain compat with Rose 2019)
-        elif not Path(self.run_dir, 'rose-suite.conf').is_file():
+        elif (
+            not (self.fpath.parent / Path('rose-suite.conf')).is_file()
+        ):
             LOG.debug(msg)
             return
 
@@ -1491,18 +1487,39 @@ class WorkflowConfig:
         print_tree(tree, padding=padding, use_unicode=pretty)
 
     def process_workflow_env(self):
-        """Workflow context is exported to the local environment."""
+        """Export Workflow context to the local environment.
+
+        A source workflow has only a name.
+        Once installed it also has an ID and a run directory.
+        And at scheduler start-up it has work, share, and log sub-dirs too.
+        """
         for key, value in {
             **verbosity_to_env(cylc.flow.flags.verbosity),
-            'CYLC_WORKFLOW_ID': self.workflow,
             'CYLC_WORKFLOW_NAME': self.workflow_name,
             'CYLC_WORKFLOW_NAME_BASE': str(Path(self.workflow_name).name),
-            'CYLC_WORKFLOW_RUN_DIR': self.run_dir,
-            'CYLC_WORKFLOW_LOG_DIR': self.log_dir,
-            'CYLC_WORKFLOW_WORK_DIR': self.work_dir,
-            'CYLC_WORKFLOW_SHARE_DIR': self.share_dir,
         }.items():
             os.environ[key] = value
+
+        if is_installed(self.fdir):
+            # This is an installed workflow.
+            #  - self.run_dir is only defined by the scheduler
+            #  - but the run dir exists, created at installation
+            #  - run sub-dirs may exist, if this installation was run already
+            #    but if the scheduler is not running they shouldn't be used.
+            for key, value in {
+                'CYLC_WORKFLOW_ID': self.workflow,
+                'CYLC_WORKFLOW_RUN_DIR': str(self.fdir),
+            }.items():
+                os.environ[key] = value
+
+        if self.run_dir is not None:
+            # This is the scheduler; run sub-dirs must exist.
+            for key, value in {
+                'CYLC_WORKFLOW_LOG_DIR': str(self.log_dir),
+                'CYLC_WORKFLOW_WORK_DIR': str(self.work_dir),
+                'CYLC_WORKFLOW_SHARE_DIR': str(self.share_dir),
+            }.items():
+                os.environ[key] = value
 
     def process_config_env(self):
         """Set local config derived environment."""
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -74,7 +74,11 @@ from cylc.flow.param_expand import NameExpander
 from cylc.flow.parsec.exceptions import ItemNotFoundError
 from cylc.flow.parsec.OrderedDict import OrderedDictWithDefaults
 from cylc.flow.parsec.util import replicate
-from cylc.flow.pathutil import get_workflow_name_from_id
+from cylc.flow.pathutil import (
+    get_workflow_name_from_id,
+    get_cylc_run_dir,
+    is_relative_to,
+)
 from cylc.flow.platforms import FORBIDDEN_WITH_PLATFORM
 from cylc.flow.print_tree import print_tree
 from cylc.flow.subprocctx import SubFuncContext
@@ -100,7 +104,6 @@ from cylc.flow.workflow_files import (
     NO_TITLE,
     WorkflowFiles,
     check_deprecation,
-    is_installed,
 )
 from cylc.flow.xtrigger_mgr import XtriggerManager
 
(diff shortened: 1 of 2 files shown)


cylc/flow/config.py
```diff
@@ -249,11 +243,10 @@ def __init__(
...
         self.fdir = os.path.dirname(fpath)
-        self.run_dir = run_dir or get_workflow_run_dir(self.workflow)
-        self.log_dir = (log_dir or
-                        get_workflow_run_scheduler_log_dir(self.workflow))
-        self.share_dir = share_dir or get_workflow_run_share_dir(self.workflow)
-        self.work_dir = work_dir or get_workflow_run_work_dir(self.workflow)
+        self.run_dir = run_dir
+        self.log_dir = log_dir
+        self.share_dir = share_dir
+        self.work_dir = work_dir
```
hjoliver, line 249: If there is no run-dir, these should be left as `None`.

hjoliver:
Note the use case mentioned above will be handled better by #5589.

But regardless of that, the environment during file parsing should correctly reflect the source vs installed status of the workflow directory (which is this PR).

hjoliver:
This is ready to go. @oliver-sanders and @wxtim - plz review or reassign at your end.

hjoliver:
(Apologies reviewers - some stray debug code had broken the tests at the last minute ... should be fixed now!)

cylc/flow/config.py
```diff
@@ -247,13 +242,12 @@ def __init__(
...
         self.workflow = workflow  # workflow id
         self.workflow_name = get_workflow_name_from_id(self.workflow)
-        self.fpath = str(fpath)  # workflow definition
-        self.fdir = os.path.dirname(fpath)
-        self.run_dir = run_dir or get_workflow_run_dir(self.workflow)
-        self.log_dir = (log_dir or
-                        get_workflow_run_scheduler_log_dir(self.workflow))
-        self.share_dir = share_dir or get_workflow_run_share_dir(self.workflow)
-        self.work_dir = work_dir or get_workflow_run_work_dir(self.workflow)
+        self.fpath: Path = Path(fpath)  # workflow definition
```
wxtim, line 245: Would it be unreasonable to ask you for a docstring for the init method? I'd particularly like context for the non-keyword arguments, which I feel are more obscure/vague/generalized than the kwargs.

cylc/flow/config.py
```diff
@@ -890,7 +884,9 @@ def _check_implicit_tasks(self) -> None:
             )
         # Allow implicit tasks in back-compat mode unless rose-suite.conf
         # present (to maintain compat with Rose 2019)
-        elif not Path(self.run_dir, 'rose-suite.conf').is_file():
+        elif (
```
wxtim, line 887: What was wrong with this before?

tests/unit/test_config.py
```diff
@@ -1565,3 +1566,84 @@ def test__warn_if_queues_have_implicit_tasks(caplog):
...
+            False,  # not installed (parsing a source dir)
+            None,  # no run directory passed to config object by scheduler
+            {
+                'CYLC_WORKFLOW_NAME': True,  # expected environment variables
+                'CYLC_WORKFLOW_ID': False,
+                'CYLC_WORKFLOW_RUN_DIR': False,
+                'CYLC_WORKFLOW_WORK_DIR': False,
+                'CYLC_WORKFLOW_SHARE_DIR': False,
+                'CYLC_WORKFLOW_LOG_DIR': False,
+            }
```
wxtim, line 1583: This

```suggestion
            },
        id="source-dir"
```

might make the purpose of the test easier to understand than

```bash
tests/unit/test_config.py::test_cylc_env_at_parsing[False-None-cylc_vars0] 
```

tests/unit/test_config.py
```diff
@@ -1565,3 +1566,84 @@ def test__warn_if_queues_have_implicit_tasks(caplog):
...
+            True,
+            None,
+            {
+                'CYLC_WORKFLOW_NAME': True,
+                'CYLC_WORKFLOW_ID': True,
+                'CYLC_WORKFLOW_RUN_DIR': True,
+                'CYLC_WORKFLOW_WORK_DIR': False,
+                'CYLC_WORKFLOW_SHARE_DIR': False,
+                'CYLC_WORKFLOW_LOG_DIR': False,
+            }
```
wxtim, line 1595:
```suggestion
            },
        id="run-dir"

File: tests/unit/test_config.py [hunk 6]
```diff
@@ -1565,3 +1566,84 @@ def test__warn_if_queues_have_implicit_tasks(caplog):
...
+            True,
+            "/some/path",
+            {
+                'CYLC_WORKFLOW_NAME': True,
+                'CYLC_WORKFLOW_ID': True,
+                'CYLC_WORKFLOW_RUN_DIR': True,
+                'CYLC_WORKFLOW_WORK_DIR': True,
+                'CYLC_WORKFLOW_SHARE_DIR': True,
+                'CYLC_WORKFLOW_LOG_DIR': True,
+            }
```
wxtim, line 1607:
```suggestion
            },
        id="run-dir-from-scheduler"
```

wxtim (approved):
Nothing sufficient to mark other than "Approve". Have made some suggestions.

cylc/flow/config.py
oliver-sanders: TBH, for this piece of work, yes.

cylc/flow/config.py
oliver-sanders: It was always pointing at the run dir, the point of this PR is point this variable at the workflow configuration dir, which is sometimes the run dir, but could also be the src dir.

cylc/flow/config.py
oliver-sanders: Context for the non-keyword args:

* `workflow: str` is the workflow ID
* `fpath: Path` is the workflow file path
* `options` are the cli options

cylc/flow/config.py
```diff
@@ -1491,19 +1487,40 @@ def print_first_parent_tree(self, pretty=False, titles=False):
...
             'CYLC_WORKFLOW_NAME': self.workflow_name,
             'CYLC_WORKFLOW_NAME_BASE': str(Path(self.workflow_name).name),
-            'CYLC_WORKFLOW_RUN_DIR': self.run_dir,
-            'CYLC_WORKFLOW_LOG_DIR': self.log_dir,
-            'CYLC_WORKFLOW_WORK_DIR': self.work_dir,
-            'CYLC_WORKFLOW_SHARE_DIR': self.share_dir,
         }.items():
             os.environ[key] = value
 
+        if is_installed(self.fdir):
```
oliver-sanders, line 1503: `is_installed` is the wrong check here as `cylc install` is supposed to be optional.

The test should be if the workflow file is relative to the cylc-run directory.

cylc/flow/config.py
oliver-sanders: Oh heck, is this fixing a bug?!

cylc/flow/config.py
```diff
@@ -1491,19 +1487,40 @@ def print_first_parent_tree(self, pretty=False, titles=False):
...
+            #  - run sub-dirs may exist, if this installation was run already
+            #    but if the scheduler is not running they shouldn't be used.
+            for key, value in {
+                'CYLC_WORKFLOW_ID': self.workflow,
+                'CYLC_WORKFLOW_RUN_DIR': str(self.fdir),
+            }.items():
+                os.environ[key] = value
+
+        if self.run_dir is not None:
+            # This is the scheduler; run sub-dirs must exist.
```
oliver-sanders, line 1516: This is the scheduler?

hjoliver: From some debug investigations, `self.run_dir` is only defined by the scheduler (i.e., if the scheduler, as opposed to validation etc., is parsing the config file).

cylc/flow/config.py
hjoliver: Even though it is unreasonable 😁  I've added one with basic positional arg docs.

cylc/flow/config.py
hjoliver: (I'll make that comment clearer)

cylc/flow/config.py
hjoliver: On this branch, `self.run_dir` is only defined by the scheduler. (i.e., not if we're parsing the config file validation etc.). So this change just makes the `rose-suite.conf` check continue to work in the source directory as well as the run directory.

cylc/flow/config.py
hjoliver: Good point. I kinda forgot `cylc install` is optional. I'll revert to a `cylc-run` check.

cylc/flow/config.py
hjoliver: Fair play, I actually didn't realize `/` coerces its args like that.  I'll make that change.

cylc/flow/config.py
```diff
@@ -240,20 +238,27 @@ def __init__(
         work_dir: Optional[str] = None,
         share_dir: Optional[str] = None
     ) -> None:
+        """
+        Initialize the workflow config object.
+
+        Positional args:
+            workflow: workflow ID
+            fpath: workflow config file path
+             options: CLI options
```
oliver-sanders, line 247: It's just "Args" or "Arguments":

```suggestion
        Args:
            workflow: workflow ID
            fpath: workflow config file path
            options: CLI options
```

tests/unit/test_config.py
```diff
@@ -1565,3 +1566,88 @@ def test__warn_if_queues_have_implicit_tasks(caplog):
...
+            "/some/path",
+            {
+                'CYLC_WORKFLOW_NAME': True,
+                'CYLC_WORKFLOW_ID': True,
+                'CYLC_WORKFLOW_RUN_DIR': True,
+                'CYLC_WORKFLOW_WORK_DIR': True,
+                'CYLC_WORKFLOW_SHARE_DIR': True,
+                'CYLC_WORKFLOW_LOG_DIR': True,
+            },
+        id="run-dir-from-scheduler"
```
oliver-sanders, line 1610:
```suggestion
            id="source-dir"
        ),
        pytest.param(
            True,
            None,
            {
                'CYLC_WORKFLOW_NAME': True,
                'CYLC_WORKFLOW_ID': True,
                'CYLC_WORKFLOW_RUN_DIR': True,
                'CYLC_WORKFLOW_WORK_DIR': False,
                'CYLC_WORKFLOW_SHARE_DIR': False,
                'CYLC_WORKFLOW_LOG_DIR': False,
            },
            id="run-dir"
        ),
        pytest.param(
            True,
            "/some/path",
            {
                'CYLC_WORKFLOW_NAME': True,
                'CYLC_WORKFLOW_ID': True,
                'CYLC_WORKFLOW_RUN_DIR': True,
                'CYLC_WORKFLOW_WORK_DIR': True,
                'CYLC_WORKFLOW_SHARE_DIR': True,
                'CYLC_WORKFLOW_LOG_DIR': True,
            },
            id="run-dir-from-scheduler"
```

tests/unit/test_config.py
```diff
@@ -1565,3 +1566,88 @@ def test__warn_if_queues_have_implicit_tasks(caplog):
...
+        allow implicit tasks = True
+    [scheduling]
+        [[graph]]
+            R1 = 'foo'
+    """
+
+    flow_file.write_text(flow_config)
+
+    # Make it looks as if path is relative to cylc-run (i.e. installed).
+    monkeypatch.setattrt (
```
oliver-sanders, line 1636:
```suggestion
    monkeypatch.setattr(
```

hjoliver: (damn it, typo in last commit just pushed up to GH)

cylc/flow/config.py
hjoliver: Fair enough (the intention was to indicate I had not attempted to document all the args, just the positional ones as singled out by @wxtim).

wxtim:
Changes still look good.

PR #5571, opened 2023-06-06
Add CYLC_ variables to template engine globals.

hjoliver:
What it says.  Make this sort of thing work during `flow.cylc` parsing:
```
{{ CYLC_WORKFLOW_RUN_DIR | default("not defined") }}
``` 

See explanation in https://github.com/cylc/cylc-flow/pull/5570#issue-1744687086

If agreed, I'll document this, and do the same for Empy.

Will need to document that this should not be used in place of the associated environment variables in task definitions (they get evaluated at job run time and will therefore always be defined).

[UPDATE: for the linked use case, #5589 makes this PR redundant. But this might be useful for other reasons]

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/631.

### DIFF
--- a/cylc/flow/parsec/fileparse.py
+++ b/cylc/flow/parsec/fileparse.py
@@ -105,6 +105,27 @@ _UNCLOSED_MULTILINE = re.compile(
 )
 
 
+def get_cylc_env_vars() -> t.Dict[str, str]:
+    """Return a restricted dict of CYLC_ environment variables for templating.
+
+    The following variables are ignored because the do not necessarily reflect
+    the running code version (I might not use the "cylc" wrapper, or it might
+    select a different version):
+
+    CYLC_VERSION
+        Set as a template variable elsewhere, from the hardwired code version.
+
+    CYLC_ENV_NAME
+        Providing it as a template variable would just be misleading.
+    """
+    return {
+        key: val
+        for key, val in os.environ.items()
+        if key.startswith('CYLC_')
+        if key not in ["CYLC_VERSION", "CYLC_ENV_NAME"]
+    }
+
+
 def _concatenate(lines):
     """concatenate continuation lines"""
     index = 0
@@ -446,10 +467,9 @@ def read_and_proc(
         flines = inline(
             flines, fdir, fpath, viewcfg=viewcfg)
 
+    # Add the hardwired code version to template vars as CYLC_VERSION
     template_vars['CYLC_VERSION'] = __version__
-
     template_vars = merge_template_vars(template_vars, extra_vars)
-
     template_vars['CYLC_TEMPLATE_VARS'] = template_vars
 
     # Fail if templating_detected ≠ hashbang
--- a/cylc/flow/parsec/jinja2support.py
+++ b/cylc/flow/parsec/jinja2support.py
@@ -39,6 +39,7 @@ from jinja2 import (
 
 from cylc.flow import LOG
 from cylc.flow.parsec.exceptions import Jinja2Error
+from cylc.flow.parsec.fileparse import get_cylc_env_vars
 
 TRACEBACK_LINENO = re.compile(
     r'\s+?File "(?P<file>.*)", line (?P<line>\d+), in .*template'
@@ -186,11 +187,15 @@ def jinja2environment(dir_=None):
                     envnsp[fname] = getattr(module, fname)
 
     # Import WORKFLOW HOST USER ENVIRONMENT into template:
-    # (usage e.g.: {{environ['HOME']}}).
+    # (Usage e.g.: {{environ['HOME']}}).
     env.globals['environ'] = os.environ
     env.globals['raise'] = raise_helper
     env.globals['assert'] = assert_helper
 
+    # Add `CYLC_` environment variables to the global namespace.
+    env.globals.update(
+        get_cylc_env_vars()
+    )
     return env
 
 
@@ -269,7 +274,6 @@ def jinja2process(
     # CALLERS SHOULD HANDLE JINJA2 TEMPLATESYNTAXERROR AND TEMPLATEERROR
     # AND TYPEERROR (e.g. for not using "|int" filter on number inputs.
     # Convert unicode to plain str, ToDo - still needed for parsec?)
-
     try:
         env = jinja2environment(dir_)
         template = env.from_string('\n'.join(flines[1:]))
@@ -300,16 +304,5 @@ def jinja2process(
             lines=get_error_lines(fpath, flines),
         )
 
-    flow_config = []
-    for line in lines:
-        # Jinja2 leaves blank lines where source lines contain
-        # only Jinja2 code; this matters if line continuation
-        # markers are involved, so we remove blank lines here.
-        if not line.strip():
-            continue
-            # restoring newlines here is only necessary for display by
-        # the cylc view command:
-        # ##flow_config.append(line + '\n')
-        flow_config.append(line)
-
-    return flow_config
+    # Ignore blank lines (lone Jinja2 statements leave blank lines behind)
+    return [line for line in lines if line.strip()]
--- a/cylc/flow/scripts/view.py
+++ b/cylc/flow/scripts/view.py
@@ -20,9 +20,17 @@
 
 Print a processed workflow configuration.
 
-Note:
-  This is different to `cylc config` which displays the parsed
-  configuration (as Cylc would see it).
+Print workflow configurations as processed before full parsing by Cylc. This
+includes Jinja2 or Empy template processing, and inlining of include-files.
+Some explanatory markup may also be requested.
+
+Warning:
+  This command will fail if `CYLC_` template variables are referenced
+  without default values, because they are only defined for full parsing.
+  E.g. (Jinja2): `{{CYLC_WORKFLOW_ID | default("not defined")}}`.
+
+See also `cylc config`, which displays the fully parsed configuration.
+
 """
 
 import asyncio
@@ -115,20 +123,21 @@ async def _main(options: 'Values', workflow_id: str) -> None:
         constraint='workflows',
     )
     # read in the flow.cylc file
-    viewcfg = {
-        'mark': options.mark,
-        'single': options.single,
-        'label': options.label,
-        'empy': options.empy or options.process,
-        'jinja2': options.jinja2 or options.process,
-        'contin': options.cat or options.process,
-        'inline': (options.inline or options.jinja2 or options.empy
-                   or options.process),
-    }
     for line in read_and_proc(
         flow_file,
         get_template_vars(options),
-        viewcfg=viewcfg,
+        viewcfg={
+            'mark': options.mark,
+            'single': options.single,
+            'label': options.label,
+            'empy': options.empy or options.process,
+            'jinja2': options.jinja2 or options.process,
+            'contin': options.cat or options.process,
+            'inline': (
+                options.jinja2 or options.empy or
+                options.inline or options.process
+            ),
+        },
         opts=options,
     ):
         print(line)
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
+import sys
 from optparse import Values
 from typing import Any, Callable, Dict, List, Optional, Tuple, Type
 from pathlib import Path
@@ -34,6 +35,7 @@ from cylc.flow.exceptions import (
     WorkflowConfigError,
     XtriggerConfigError,
 )
+from cylc.flow.parsec.exceptions import Jinja2Error, EmPyError
 from cylc.flow.scheduler_cli import RunOptions
 from cylc.flow.scripts.validate import ValidateOptions
 from cylc.flow.workflow_files import WorkflowFiles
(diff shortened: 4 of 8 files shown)


oliver-sanders:
(remote tests unhappy)

cylc/flow/parsec/jinja2support.py
```diff
@@ -186,11 +186,17 @@ def jinja2environment(dir_=None):
...
+
     env.globals['raise'] = raise_helper
     env.globals['assert'] = assert_helper
 
+    # And all `CYLC_` variables (this will include CYLC_WORKFLOW environment
+    # variables exported by `cylc` commands before config file parsing.
+    for key, val in os.environ.items():
+        if key.startswith('CYLC_'):
+            env.globals[key] = val
     return env
```
oliver-sanders, line 200: Note this will include `CYLC_VERSION` and `CYLC_ENV_NAME` if using the wrapper script.

I think `CYLC_VERSION` is already set elsewhere?

hjoliver: Does that matter?

cylc/flow/scripts/view.py
```diff
@@ -18,11 +18,16 @@
 
 """cylc view [OPTIONS] ARGS
 
-Print a processed workflow configuration.
+Print workflow configurations as processed before full parsing by Cylc. This
```
oliver-sanders, line 21: IMO, each command/config/whatever should have a concise simple description.

The first line of the CLI help is also used by `cylc help all` so needs to be concise.

```suggestion
Print a processed workflow configuration.

Print workflow configurations as processed before full parsing by Cylc. This
```

cylc/flow/scripts/view.py
```diff
@@ -18,11 +18,16 @@
...
+includes Jinja2 or Empy template processing, and inlining of include-files.
+Some explanatory markup may also be requested.
+
+Warning: this command will fail if Jinja2 `CYLC_` variables are referenced
+without default values, because they are only defined during full parsing.
+E.g., `{{CYLC_WORKFLOW_NAME | default("not defined")}}`.
```
oliver-sanders, line 27: Headings ending in `:` get put in bold which will help make this warning stand out from the surrounding text.

Swapped `CYLC_WORKFLOW_NAME` for `ID` as I can't really think of legit uses of `CYLC_WORKFLOW_NAME` so would rather keep the visibility of this low (we only added it to support MASS use cases at the MO).

```suggestion
Warning:
  This command will fail if Jinja2 `CYLC_` variables are referenced
  without default values, because they are only defined during full parsing.
  E.g., `{{CYLC_WORKFLOW_ID | default("not defined")}}`.
```

cylc/flow/scripts/view.py
hjoliver: > IMO, each command/config/whatever should have a concise simple description.

I agree. 

Problem was the original didn't say what "processed" means.  I suspect most users would equate that with "parsed".

But moving the fuller explanation down a bit is all good. (And I'd forgotten we were using the first line like that).

oliver-sanders (approved):

cylc/flow/parsec/jinja2support.py
hjoliver: Hmm, yeah the "truth" `CYLC_VERSION` is inserted there in `parsec.fileparse`.

I'll deal with this one tomorrow. And I'll move the env export into parse.fileparse to catch empy as well.

oliver-sanders:
Hmm, found this in the debug section for the failed test:

```
Can't find:
===========
CYLC_JOB_RUNNER_NAME=background
===========
in:
===========
CYLC_JOB_RUNNER_NAME=at
...
```

???

tests/unit/test_config.py
```diff
@@ -1007,6 +1008,32 @@ def test_rsync_includes_will_not_accept_sub_directories(tmp_flow_config):
     assert "Directories can only be from the top level" in str(exc.value)
 
 
+@pytest.mark.parametrize(
+    'cylc_var, expected_err',
+    [
+        ["CYLC_WORKFLOW_NAME", None],
+        ["CYLC_BEEF_WELLINGTON", (Jinja2Error, "is undefined")],
```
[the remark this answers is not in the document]
wxtim: https://media.giphy.com/media/3o85xG3XDsgCLuGC52/giphy.gif

tests/unit/test_config.py
```diff
@@ -1007,6 +1008,32 @@ def test_rsync_includes_will_not_accept_sub_directories(tmp_flow_config):
...
+    [scheduling]
+        [[graph]]
+            R1 = foo
+    """)
+    if expected_err is None:
+        WorkflowConfig(workflow=reg, fpath=flow_file, options=Values())
+    else:
+        with pytest.raises(expected_err[0]) as exc:
+            WorkflowConfig(workflow=reg, fpath=flow_file, options=Values())
+        assert expected_err[1] in str(exc)
```
wxtim, line 1034: I think this is just a smidge faster.

```suggestion
    try:
        WorkflowConfig(workflow=reg, fpath=flow_file, options=Values())
    except expected_err[0] as exc:
        assert expected_err[1] in str(exc)
```

wxtim:
I've not put anything demanding radical changes in my comments. Will approve once I've done a bit of manual testing and conflicts are resolved.

tests/unit/test_config.py
wxtim: @oliver-sanders  - I'm not care that much about this change - we can ignore it safely.

wxtim (approved):
@oliver-sanders 

I've done a deconflict merge.

Could you do a final check that my merge hasn't broken anything?

oliver-sanders:
> Could you do a final check that my merge hasn't broken anything?

This PR will still have tests failing, see above.

MetRonnie:
The test that is failing had this in it:
https://github.com/cylc/cylc-flow/blob/6f8a201b9ffd67d0293d80b47445ed3b753b501c/tests/functional/cylc-cat-log/01-remote/flow.cylc#L19

Before this PR I think `CYLC_TEST_PLATFORM` was always an undefined Jinja2 variable so the workflow was not actually running remotely and the test was not doing what it was supposed to be doing (it was testing local cat-log not remote cat-log)

oliver-sanders:
That doesn't make sense to me, there are other tests which follow this pattern:

https://github.com/cylc/cylc-flow/blob/6f8a201b9ffd67d0293d80b47445ed3b753b501c/tests/functional/remote/00-basic.t#L19

But don't provide a default:

https://github.com/cylc/cylc-flow/blob/6f8a201b9ffd67d0293d80b47445ed3b753b501c/tests/functional/remote/basic/flow.cylc#L9

So would get a Jinja2Undefined error for these cases?

oliver-sanders:
Oh, sorry, didn't read that properly, it's a Jinja2 variable :facepalm:

oliver-sanders:
Here's a test that runs in CI and uses the Jinja2 variable:

https://github.com/cylc/cylc-flow/blob/6f8a201b9ffd67d0293d80b47445ed3b753b501c/tests/functional/cylc-kill/01-multi-hosts/flow.cylc#L23

MetRonnie:
In that one it uses `cylc play -s "CYLC_TEST_PLATFORM=$CYLC_TEST_PLATFORM"` but not in the failing one

tests/unit/test_config.py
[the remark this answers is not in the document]
hjoliver: A frightening sight for a vegetarian!

tests/unit/test_config.py
hjoliver: (I've taken your suggestion)

tests/unit/test_config.py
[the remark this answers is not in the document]
wxtim: Sorry

cylc/flow/parsec/jinja2support.py
hjoliver: Addressing ...

tests/unit/test_config.py
hjoliver: >    `except expected_err[0] as exc:`

Huh, flake8 no like this: *B030 Except handlers should only be names of exception classes*

MetRonnie: I don't like the `try...except` - we should use `pytest.raises()` to assert that a raise happens according to plan IMO

cylc/flow/parsec/jinja2support.py
hjoliver: I'm now not taking CYLC_VERSION or CYLC_ENV_NAME from the environment, as template variables.

cylc/flow/parsec/jinja2support.py
```diff
@@ -300,16 +305,4 @@ def jinja2process(
...
-        # markers are involved, so we remove blank lines here.
-        if not line.strip():
-            continue
-            # restoring newlines here is only necessary for display by
-        # the cylc view command:
-        # ##flow_config.append(line + '\n')
-        flow_config.append(line)
-
-    return flow_config
+    return [line for line in lines if line.strip()]
```
hjoliver, line 308:
```suggestion
    # Ignore blank lines (lone Jinja2 statements leave blank lines behind)
    return [line for line in lines if line.strip()]
```

tests/unit/test_config.py
hjoliver: Yeah, agreed. I'll revert.

hjoliver:
Addressed review comments.
Extended to empy.

cylc/flow/parsec/jinja2support.py
```diff
@@ -269,11 +274,11 @@ def jinja2process(
     # CALLERS SHOULD HANDLE JINJA2 TEMPLATESYNTAXERROR AND TEMPLATEERROR
     # AND TYPEERROR (e.g. for not using "|int" filter on number inputs.
     # Convert unicode to plain str, ToDo - still needed for parsec?)
-
     try:
         env = jinja2environment(dir_)
         template = env.from_string('\n'.join(flines[1:]))
         lines = str(template.render(template_vars)).splitlines()
+        print()
```
MetRonnie, line 281: Leftover debugging?
```suggestion
```

CHANGES.md
```diff
@@ -38,6 +38,9 @@ in family names to be split, e.g. `<foo>FAM<bar>`.
 directory during file parsing, to give the template processor access to
 workflow files.
 
+[#5571](https://github.com/cylc/cylc-flow/pull/5571) - Make workflow `CYLC_`
+variables available to the template processor during parsing.
+
```
MetRonnie, line 43: Will need to be moved. If #5634 goes in then the sentence should be written to `changes.d/5571.feat.md`.

cylc/flow/parsec/jinja2support.py
hjoliver: Oops

hjoliver: (was causing functional test failures)

CHANGES.md
hjoliver: moved

hjoliver:
> In that one it uses `cylc play -s "CYLC_TEST_PLATFORM=$CYLC_TEST_PLATFORM"` but not in the failing one

And fixed the failing test (it now runs remote jobs, as advertised).

hjoliver:
Should be good to go now. Cylc-doc PR up too.

cylc/flow/parsec/fileparse.py
```diff
@@ -105,6 +105,27 @@
...
+        Set as a template variable elsewhere, from the hardwired code version.
+
+    CYLC_ENV_NAME
+        Providing it as a template variable would just be misleading.
+    """
+    return {
+        key: val
+        for key, val in os.environ.items()
+        if key.startswith('CYLC_')
+        if key not in ["CYLC_VERSION", "CYLC_ENV_NAME"]
```
wxtim, line 125: Not militant, but this seems eminently unit-testable?

CHANGES.md
```diff
@@ -10,6 +10,13 @@ creating a new release entry be sure to copy & paste the span tag with the
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
 
+
+## __cylc-8.2.1 (<span actions:bind='release-date'>Pending</span>)__
+
+[#5571](https://github.com/cylc/cylc-flow/pull/5571) - Make workflow `CYLC_`
+variables available to the template processor during parsing.
+
+
```
MetRonnie, line 19: Unless any objections to #5634 then I suggest the description should instead be written to `changes.d/5571.feat.md`.

MetRonnie (approved):

cylc/flow/parsec/fileparse.py
MetRonnie: Good point, I agree a unit test is called for

CHANGES.md
hjoliver: No objections. I've just merged that PR.  I'll draft this briefly until I've added a unit test and rejigged the change log.

hjoliver: Done.

cylc/flow/parsec/fileparse.py
hjoliver: Done.

hjoliver:
Right, all done.

I think `tests/functional/broadcast/00-simple.t` is going to fail, but is also failing on 8.2.x ... I've run out of time to investigate for now.

MetRonnie:
Regression in `tests/f/cylc-cat-log/01-remote.t` failing again

Btw as this is an enhancement should it be on the 8.3.0 milestone (and base changed to master)?

hjoliver:
> Regression in tests/f/cylc-cat-log/01-remote.t failing again

Damn it, problem was I stupidly force-pushed from my outdated branch on the laptop at home today after yesterday's fixes at work. I had to redo some changes, but forgot this one.

And yes, it should be rebased to master.

MetRonnie:
I've cherry picked the missing commit. And changed the base + milestone; no need to rebase however

MetRonnie (approved):

hjoliver:
Great, thanks!

wxtim (approved):

PR #5574, opened 2023-06-09
log_vc_info: catch info command errors and present them as warnings

MetRonnie:
This fixes a scary traceback when installing a workflow from an untracked subdirectory of an SVN repo. In general, it presents any non-zero exit from a VCS command as a warning rather than an error.

**Check List**

- [x] No dependency changes 
- [x] Tests are included
- [x] No `CHANGES.md` entry included as this is minor
- [x] No docs update needed

### DIFF
--- a/cylc/flow/install_plugins/log_vc_info.py
+++ b/cylc/flow/install_plugins/log_vc_info.py
@@ -169,7 +169,7 @@ def get_vc_info(path: Union[Path, str]) -> Optional[Dict[str, Any]]:
             ):
                 LOG.debug(f"Source dir {path} is not a {vcs} repository")
             elif cylc.flow.flags.verbosity > -1:
-                LOG.warning(exc)
+                LOG.warning(f"$ {vcs} {' '.join(args)}\n{exc}")
             continue
 
         info['version control system'] = vcs
--- a/cylc/flow/install_plugins/log_vc_info.py
+++ b/cylc/flow/install_plugins/log_vc_info.py
@@ -66,13 +66,15 @@ from typing import (
     Any, Dict, Iterable, List, Optional, TYPE_CHECKING, TextIO, Union, overload
 )
 
-from cylc.flow import LOG
+from cylc.flow import LOG as _LOG, LoggerAdaptor
 from cylc.flow.exceptions import CylcError
 import cylc.flow.flags
 from cylc.flow.workflow_files import WorkflowFiles
 
 if TYPE_CHECKING:
     from optparse import Values
+
+LOG = LoggerAdaptor(_LOG, {'prefix': __name__})
 
 
 SVN = 'svn'
--- a/cylc/flow/install_plugins/log_vc_info.py
+++ b/cylc/flow/install_plugins/log_vc_info.py
@@ -68,6 +68,7 @@ from typing import (
 
 from cylc.flow import LOG
 from cylc.flow.exceptions import CylcError
+import cylc.flow.flags
 from cylc.flow.workflow_files import WorkflowFiles
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


tests/unit/post_install/test_log_vc_info.py
```diff
@@ -272,3 +273,14 @@ def test_no_base_commit_git(tmp_path: Path):
...
     for line in diff_file.read_text().splitlines():
         assert line.startswith('#')
+
+
+@require_svn
+def test_untracked_svn_subdir(
+    svn_source_repo: Tuple[str, str, str], caplog, log_filter
+):
+    repo_dir, *_ = svn_source_repo
+    source_dir = Path(repo_dir, 'jar_jar_binks')
```
wxtim, line 283: How wooud!

cylc/flow/scripts/install.py
```diff
@@ -285,6 +285,7 @@ def install_cli(
 def install(
     opts: 'Values', reg: Optional[str] = None
 ) -> Tuple[str, str]:
+    set_timestamps(LOG, opts.verbosity > 1)
```
wxtim, line 288: I think that you need this:

```suggestion
    set_timestamps(LOG, opts.verbosity > 1 and opts.log_timestamp)
```

If you don't I think that

```bash
cylc install --debug --no-timestamp
```

Gets a timestamp.

cylc/flow/install_plugins/log_vc_info.py
```diff
@@ -162,14 +163,14 @@ def get_vc_info(path: Union[Path, str]) -> Optional[Dict[str, Any]]:
...
-            if not any(
+            if any(
                 exc.strerror.lower().startswith(err)
                 for err in NOT_REPO_ERRS[vcs]
             ):
-                raise exc
-            else:
                 LOG.debug(f"Source dir {path} is not a {vcs} repository")
-                continue
+            elif cylc.flow.flags.verbosity > -1:
```
wxtim, line 171: This seems to deal with the problem.

In my head there is a question about why we handle some SVN exc and not others - Why is `svn: warning: W155007` more worthy of capture and re-interpretation than `svn: warning: W155010`? I'm happy that this question only tangentially related to this PR.

wxtim (commented):
Broadly happy with this, except the bit about timestamps.

cylc/flow/scripts/install.py
oliver-sanders: (would need to test this with `vip`)

wxtim (approved):
I'm happy with this.

* I have checked VIP with `--no-timestamp` and `-vvv` and got no unexpected behaviour.
* We discussed error code capture in person with @oliver-sanders  and @dpmatthews  and I'm happy to pass this as is.

oliver-sanders (commented):
LGTM, works fine.

The warnings are a bit confusing though because log-vc-info is a post-install plugin so the warning gets output *after* the `INSTALLED` message.

This makes it look like the command as a whole failed:

Screenshot from 2023-06-13 10-07-15

Suggest moving this CLI print statement out of the library code, into the CLI script and calling it after the post-install plugins:

```diff
diff --git a/cylc/flow/scripts/install.py b/cylc/flow/scripts/install.py
index b2b424291..e6813b641 100755
--- a/cylc/flow/scripts/install.py
+++ b/cylc/flow/scripts/install.py
@@ -337,4 +337,6 @@ def install(
                 exc
             ) from None
 
+    print(f'INSTALLED {workflow_id} from {source_dir}')
+
     return workflow_name, workflow_id
diff --git a/cylc/flow/workflow_files.py b/cylc/flow/workflow_files.py
index e51036da3..f74bed8f7 100644
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -1753,7 +1753,6 @@ def install_workflow(
         install_log.info(
             f'Symlink from "{source_link}" to "{source}" in place.')
     install_log.info(f'INSTALLED {named_run} from {source}')
-    print(f'INSTALLED {named_run} from {source}')
     close_log(install_log)
     return source, rundir, workflow_name, named_run
```

PR #5582, opened 2023-06-13
Allow pre configure plugin to access compat mode info in cylc install.

wxtim:
(Show "real" args used by Cylc Play in VIP)

Part fixes a bug spotted by a user and not documented as a separate ticket: Cylc install behaviour was inconsistent between `cylc install` and `cylc vip` in the example described in https://github.com/cylc/cylc-rose/issues/230 because `cylc install` did not set the `cylc7_compat_mode` until after the pre_configure plugin ran.

**Check List**

### DIFF
--- a/cylc/flow/scripts/validate_install_play.py
+++ b/cylc/flow/scripts/validate_install_play.py
@@ -28,6 +28,8 @@ This script is equivalent to:
 
 """
 
+import sys
+
 from cylc.flow.scripts.validate import (
     VALIDATE_OPTIONS,
     _main as validate_main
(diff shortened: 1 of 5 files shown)


cylc/flow/scripts/validate_install_play.py
```diff
@@ -111,5 +113,5 @@ def main(parser: COP, options: 'Values', workflow_id: Optional[str] = None):
     )
 
     set_timestamps(LOG, options.log_timestamp)
-    log_subcommand('play', workflow_id)
+    log_subcommand(*sys.argv[1:])
```
wxtim, line 116: Technically off-topic, but it's a very small enhancement.

PR #5585, opened 2023-06-14
Graduated force-triggering (prereqs, xtriggers, queues).

### LINKED ISSUE #5572: Spawning parentless clock-xtriggered task
Reported by @hjoliver on Element:

Consider a cycling graph with a parentless clock-xtriggered task foo at the top of each cycle. Currently (perhaps pending a future capability for clock triggers to spawn dependent tasks on demand):

* the clock-xtrigger requires a waiting foo in the active pool, to wait on it
* successive instances of foo get auto-spawned out to the runahead limit (there are no parent tasks to do It)

At normal start-up, the first instance of foo gets spawned by the scheduler, then (having no parents to do so) it spawns future instances of itself.

However, when triggering a new flow, or when starting up with `--start-tasks`, at some other point in the graph, we likely need to arrange for the "next" instance of foo to be spawned as waiting on the clock-xtrigger. At the moment this can't be done (since `cylc spawn` was removed):
* triggering a previous instance of foo will spawn the correct "next" one, but it will cause unwanted action downstream of the previous instance
* triggering the "next" instance will cause it to run before it's clock-xtrigger is satisfied
* setting outputs or prerequisites won't help

hjoliver:
Close #5572 (and extends it to push ext-triggers as well)

- If a task has unsatisfied prerequisites, `cylc trigger` will satisfy them
- Otherwise, it will satisfy any unsatisfied xtriggers
- Otherwise, it will submit the task to run despite queue limiting

The runahead limit is ignored (i.e., you can trigger tasks beyond the limit).

Reviewers - this is an important part of required intervention capability. E.g. if starting a flow in the middle of a cycle point, we need to be able to set up clock-triggered tasks in subsequent cycles to start checking their clock-triggers now, but not to run immediately.

@dpmatthews - pinging you for a quick look, as the brains behind the xtrigger tweak on this branch.

**Check List**

- [ ] ~Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).~
- [ ] ~Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.~
- [ ] ~If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.~

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,10 @@ ones in. -->
 
 ### Enhancements
 
+[#5585](https://github.com/cylc/cylc-flow/pull/5585) - Make `cylc trigger` 
+act incrementally with each call: first runahead release and satisfy
+prerequisites, then xtriggers, then queue release.
+
 [#5405](https://github.com/cylc/cylc-flow/pull/5405) - Improve scan command
 help, and add scheduler PID to the output.
 
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1604,8 +1604,7 @@ class Scheduler:
                     if self.xtrigger_mgr.check_xtriggers(
                             itask, self.workflow_db_mgr.put_xtriggers):
                         housekeep_xtriggers = True
-                        if all(itask.is_ready_to_run()):
-                            self.pool.queue_task(itask)
+                        self.pool.queue_if_ready(itask)
 
                 # Check for satisfied ext_triggers, and queue if ready.
                 if (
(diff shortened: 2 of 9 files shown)


hjoliver:
For the record, the way users force-spawn an xtriggered task such that it start checking its xtrigger now, as opposed to actually running now, falls into the `cylc set` proposal discussion. But we need this particularly capability fairly urgently, so temporarily homing it here in `cylc trigger` would be better than delaying it for long.  [Discussed more on #5572]

hjoliver:
(Probably superseded by the newly agreed intervention plans,)

hjoliver:
Superseded by #5658 

(I'll close this once I get a chance to do a quick salvage investigation).

oliver-sanders:
@hjoliver, ready to close this one?

oliver-sanders:
Poke

PR #5589, opened 2023-06-19
cd to workflow dir during file parsing.

hjoliver:
Allow the template processor to access workflow files - in the source dir for a source workflow, or the run dir for installed workflows.

This is the most straightforward way to cater to [this kind of use case](https://github.com/cylc/cylc-flow/pull/5570#issue-1744687086)

(I can't think of any downside...)

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) issue opened if required at cylc/cylc-doc/issues/622

### DIFF
--- a/cylc/flow/parsec/fileparse.py
+++ b/cylc/flow/parsec/fileparse.py
@@ -401,6 +401,12 @@ def read_and_proc(
     fpath = _get_fpath_for_source(fpath, opts)
     fdir = os.path.dirname(fpath)
 
+    odir = os.getcwd()
+
+    # Move to the file location to give the template processor easy access to
+    # other files in the workflow directory (whether source or installed).
+    os.chdir(fdir)
+
     # Allow Python modules in lib/python/ (e.g. for use by Jinja2 filters).
     workflow_lib_python = os.path.join(fdir, "lib", "python")
     if (
@@ -493,6 +499,8 @@ def read_and_proc(
     # concatenate continuation lines
     if do_contin:
         flines = _concatenate(flines)
+
+    os.chdir(odir)
 
     # return rstripped lines
     return [fl.rstrip() for fl in flines]
--- a/tests/unit/parsec/test_fileparse.py
+++ b/tests/unit/parsec/test_fileparse.py
@@ -16,6 +16,7 @@
 
 import tempfile
 
+from pathlib import Path
 import os
 import pytest
 from pytest import param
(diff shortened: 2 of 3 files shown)


oliver-sanders (approved):

cylc/flow/parsec/fileparse.py
```diff
@@ -401,6 +401,12 @@ def read_and_proc(
     fpath = _get_fpath_for_source(fpath, opts)
     fdir = os.path.dirname(fpath)
 
+    odir = os.getcwd()
```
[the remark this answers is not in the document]
oliver-sanders: It doesn't actually matter if we don't `os.chdir` back afterwards, this only applies to the Python session itself not the parent shell.

tests/unit/parsec/test_fileparse.py
```diff
@@ -357,6 +358,31 @@ def test_read_and_proc_jinja2():
...
+            tf.write(
+"""#!Jinja2
+{% from "os" import listdir %}
+{% for f in listdir("sub") %}
+{{f}}
+{% endfor %}
+""")
+            tf.flush()
+            r = read_and_proc(fpath=tf.name, viewcfg=viewcfg)
+            assert sorted(r) == ['a', 'b', 'c']
```
wxtim, line 385: Possible alternative

- Slightly faster
- Less code
- Uses builtin `test_path` fixture which clears files up when the test passes.
- Use implied type forcing of Path's `/` method.

```suggestion

def test_read_and_proc_cwd(tmp_path):
    """The template processor should be able to read workflow files.

    This relies on moving to the config dir during file parsing.
    """
    sdir = tmp_path / "sub"
    sdir.mkdir()
    [(sdir / s).touch() for s in ["a", "b", "c"]]
    viewcfg = {
        'empy': False, 'jinja2': True,
        'contin': False, 'inline': False
    }
    tf = tmp_path / "a.conf"
    tf.write_text(
        '#!Jinja2\n'
        '{% from "os" import listdir %}\n'
        '{% for f in listdir("sub") %}\n'
        '{{f}}\n'
        '{% endfor %}'
    )
    with open(tf, 'r') as fh:
        r = read_and_proc(fpath=fh.name, viewcfg=viewcfg)
    assert sorted(r) == ['a', 'b', 'c']

```

wxtim (commented):
- [x] Read code
- [x] Tried code

I'm generally happy, but I've made some suggestions about the test.

tests/unit/parsec/test_fileparse.py
oliver-sanders: FYI, this may be considered an anti-pattern, I think some linters will flag it.

```diff
-       for sub in ["a", "b", "c"]:
-           (sdir / Path(sub)).touch()
+      [(sdir / s).touch() for s in ["a", "b", "c"]]
```

The reason being that list-comprehensions are not expected to have side-effects (touching files in this case).

wxtim: That seems reasonable.

hjoliver: It also creates a list `[None, None, None]` for no reason.

hjoliver: @wxtim - is there anything else in your suggestions above that you particularly think we should keep?

Note I wrapped it all in the `tempfile.TemporaryDirectory()` context manager (which you removed) for the automatic cleanup.

cylc/flow/parsec/fileparse.py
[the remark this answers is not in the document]
hjoliver: Note my change does `os.chdir` back afterwards. I agree it should not matter, but to explicitly avoid any side-effects of the dir change that I might not be aware of.

tests/unit/parsec/test_fileparse.py
wxtim: Pytest does automatic cleanup on tmp_path, which is why that's the thing I'd keep out of that lot. Also, if you need to debug this later you know the files will be in `/var/tmp/pytest-of-${USER}/pytest-current`

hjoliver: Done.

hjoliver:
@wxtim - I tweaked the new test as you suggested.  A final commit just cleans up the file to pass flake8.

wxtim (approved):

Issue #5107, opened 2022-08-26
Making reload safer


dpmatthews:
re https://cylc.discourse.group/t/task-wrangling-issues/517/8
We need to make it obvious when reloads fail.

I think reload should:

* pause the workflow (and possibly wait for any tasks in the preparing state to submit?)
* reload the config (log what has changed?)
* if the reload succeeds, release the workflow (unless the workflow was previously held)
* if the reload fails, leave the workflow paused so the user has the choice to resume with the existing workflow definition or fix the problem and reload again

Also, what about restarts? If a server dies and you restart a workflow it's meant to continue unchanged. However, what happens if you've done a reinstall but not reloaded (and perhaps even have a workflow config which does not validate)?
I think restart should always use the previous config.
If this differs from the installed config this should be logged (if possible) but not loaded.
We could add a --reload option to cylc play to support the old behaviour.

oliver-sanders:
> and possibly wait for any tasks in the preparing state to submit?

We would need to flush through all tasks in the preparing state (otherwise we can't say for sure whether the job file was written with the old or new settings) whilst the workflow is paused (otherwise tasks will continue to enter the preparing state).

One downside of this is that a remote operation (e.g. remote-init) running during the reload process would cause the workflow to become unresponsive for that period. Without major changes there would be no way to "cancel" a reload once actioned.

oliver-sanders:
> I think restart should always use the previous config.

I don't think this is really possible because the config isn't just the `flow.cylc` file, it's also the `bin/` dir and `rose-app.conf` files. If we restarted the workflow with a previous Cylc configuration we could be launching it with a mixed configuration that was never intended to be run.

dpmatthews:
> I think restart should always use the previous config.

It should also not repeat any remote file installation.

hjoliver:
Well done, good to have that info laid out in full :+1: 

> Main takeaway is that prereqs and outputs can't be changed by a reload.

Worth pointing out, for anyone reading this, we're only talking about reloading existing task proxies here - everything can be changed by reload if the task hasn't been spawned yet.

> Since SoD this should probably be possible now. It makes sense to add or remove a prerequisite of a waiting task or add an output to a failed one?

Yeah, we should be able to do that.

hjoliver:
What should be preserved:
- any attribute that relates to the history of this particular task proxy (jobs, submit number etc.)

What should be reloaded:
- everything else

Any exceptions to that?

[...]

PR #5592, opened 2023-06-19
reload: wait for pending tasks to submit and pause the workflow

### LINKED ISSUE #5579: reload: stuck preparing task
Seen in the wild.

* A task was configured to run on a remote platform.
* The task entered the preparing state just before a reload was actioned.
* The reload subsequently triggered remote-init (tasks had submitted to this platform pre-reload).
* The task never left the preparing state.

Investigate what happens to preparing tasks if remote-init is performed whilst they are "preparing".

See #5107, waiting for preparing tasks to clear before pausing the workflow and attempting reload seems a sensible awy forward.

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/5107
* Reload now waits for pending tasks to submit before attempting to reload the config itself.
* Reload now also puts the workflow into the paused state during the reload process. This doesn't actually achieve anything as the reload command is blocking in the main loop, but it does help to communicate that the workflow will not de-queue or submit and new tasks during this process.
* The workflow status message is now updated to reflect the reload progress.

This should also close #5579 by removing any potential for interaction between reload and preparing tasks. But note, I've not actually reproduced #5579.

**Needs extensive functional testing.**

The second commit straightens out the reload logic which is currently split between four routines:
* Scheduler.command_reload_workflow
* Scheduler.main_loop
* Pool.set_do_reload
* Pool.reload_taskdefs

With `command_reload_workflow` calling `set_do_reload` somewhere in the middle of the routine, then handing back over to the main loop which does its part before calling `reload_taskdefs`. I think this is just Cylc 7 hangover, removing the main-loop interaction from reload should reduce the chance of unexpected state changes occurring during the process.

**Check List**


cylc/flow/task_pool.py
```diff
@@ -913,25 +911,7 @@ def set_max_future_offset(self):
...
-
-        # find any old tasks that have been removed from the workflow
-        old_task_name_list = self.task_name_list
-        self.task_name_list = self.config.get_task_name_list()
-        for name in old_task_name_list:
-            if name not in self.task_name_list:
-                self.orphans.append(name)
-        for name in self.task_name_list:
-            if name in self.orphans:
-                self.orphans.remove(name)
```
oliver-sanders, line 930: The list of orphans used to be persisted between reloads. Newly bereaved tasks were added to the list and only removed if a matching task was added back in, otherwise they were left in libo. However, they could be pulled back by a subsequent reload if the task was added back into the definition.

I think this is another piece of Cylc 7 hangover as at Cylc 8, orphans are swiftly destroyed unless ready to run.

cylc/flow/scheduler.py
```diff
@@ -1048,35 +1053,89 @@ def command_remove_tasks(self, items) -> int:
...
     def command_reload_workflow(self) -> None:
         """Reload workflow configuration."""
+        # pause the workflow if not already
+        was_paused = self.is_paused
+        if not was_paused:
+            self.pause_workflow('Reloading workflow defintion')
+            self.process_workflow_db_queue()  # see #5593
+
+        # flush out preparing tasks before attempting reload
+        self.reload_pending = 'waiting for pending tasks to submit'
```
oliver-sanders, line 1063: This gets used in the workflow status so progress will be indicated in the GUI.

tests/functional/logging/01-basic.t
```diff
@@ -31,14 +31,18 @@ init_workflow "${TEST_NAME_BASE}" <<'__FLOW_CONFIG__'
         R1 = reloader1 => stopper => reloader2
 [runtime]
     [[reloader1, reloader2]]
-        script = cylc reload "${CYLC_WORKFLOW_ID}"
+        script = """
+            cylc reload "${CYLC_WORKFLOW_ID}"
+            # wait for the command to complete
+            cylc__job__poll_grep_workflow_log 'Reload completed'
```
oliver-sanders, line 37: Orthogonal change: This test was timing dependent because it didn't wait for reload to complete.

tests/functional/logging/01-basic.t
```diff
@@ -31,14 +31,18 @@ init_workflow "${TEST_NAME_BASE}" <<'__FLOW_CONFIG__'
...
+        """
     [[stopper]]
         script = cylc stop --now --now "${CYLC_WORKFLOW_ID}"
 __FLOW_CONFIG__
 
 run_ok "${TEST_NAME_BASE}-validate" cylc validate "${WORKFLOW_NAME}"
 
-workflow_run_ok "${TEST_NAME_BASE}-run" cylc play --no-detach "${WORKFLOW_NAME}"
+workflow_run_ok "${TEST_NAME_BASE}-run" cylc play --no-detach "${WORKFLOW_NAME}" --main-loop='log db'
```
oliver-sanders, line 45: Debug I forgot to remove (tomorrow problem)

tests/integration/test_scheduler.py
```diff
@@ -241,80 +241,6 @@ async def test_no_poll_waiting_tasks(
...
-                    'graph': {'R1': "a => b"}
-                }
-            },
-            ("IllegalValueError: (type=cycle point) "
-             "[scheduling]initial cycle point = 2k22 - (Invalid cycle point)"),
-            id="illegal cycle point"
-        )
-    ]
-)
-async def test_illegal_config_load(
```
oliver-sanders, line 266: Covered by the new test.

oliver-sanders:
To test, you can simulate a slow remote-init and a slow config load by jamming in sleeps like so:

```diff
diff --git a/cylc/flow/scheduler.py b/cylc/flow/scheduler.py
index d3cab5225..28bd79ac8 100644
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1096,6 +1096,7 @@ class Scheduler:
         LOG.info("Reloading the workflow definition.")
         try:
             config = self.load_flow_file(is_reload=True)
+            sleep(20)
         except (ParsecError, CylcConfigError) as exc:
             if cylc.flow.flags.verbosity > 1:
                 # log full traceback in debug mode
diff --git a/cylc/flow/scripts/remote_init.py b/cylc/flow/scripts/remote_init.py
index 7706cecfa..9c52ae8af 100755
--- a/cylc/flow/scripts/remote_init.py
+++ b/cylc/flow/scripts/remote_init.py
@@ -60,7 +60,8 @@ def get_option_parser() -> COP:
 
 @cli_function(get_option_parser)
 def main(parser, options, install_target, rund, *dirs_to_be_symlinked):
-
+    from time import sleep
+    sleep(20)
     remote_init(
         install_target,
         rund,
```

oliver-sanders:
Got a couple of functional tests to fix, but otherwise the functionality should be good.

One question, in order to keep the scheduler responsive whilst we're waiting for pending tasks to submit, I added `process_command_queue` into the mini event-loop in `command_reload_workflow`. This means any arbitrary command could be actioned during this period.

This is probably fine, the point of this loop is just to wait for the workflow to enter a safe state for reload to be performed. It's probably even possible to trigger tasks at this stage if you want them to be submitted with the pre-reload config! But there's the potential for unplanned interactions so we may want to consider whitelisting this somehow.

cylc/flow/scheduler.py
```diff
@@ -1046,37 +1056,112 @@ def command_remove_tasks(self, items) -> int:
...
+    async def command_reload_workflow(self) -> None:
         """Reload workflow configuration."""
-        LOG.info("Reloading the workflow definition.")
-        old_tasks = set(self.config.get_task_name_list())
-        # Things that can't change on workflow reload:
-        self.workflow_db_mgr.pri_dao.select_workflow_params(
-            self._load_workflow_params
-        )
+        # pause the workflow if not already
+        was_paused = self.is_paused
```
wxtim, line 1062: This feels like it could be a context manager:

```python
async with pause:
    await self.command_reload_workflow()
```

wxtim (commented):
- [x] Code read

## Functional Review

- [ ] Try this with a variety of workflows
- [ ] Check that this works the same way with `cylc vr` and `cylc reinstall`
- [ ] Check what reloading with these changes looks like in the GUI.
- [ ] Try reloading a change which breaks the workflow.
- [ ] Try to get tasks in horrible states to break the logic at line 1418 of `scheduler.py`

cylc/flow/scheduler.py
oliver-sanders: For this usage, we don't always want to unpause the workflow, e.g. if the reload goes wrong. But otherwise, good shout!

oliver-sanders:
The two test failures have flagged a legitimate issue with reload causing duplicate stall events which I'll sort out soon, will require a line or two of change.

hjoliver (approved):
Some trivial, and possibly annoying, comments.  Code LGTM.  I'll do some more brutal functional testing on Monday, but I'm pretty confident it's all good 🎉

cylc/flow/scheduler.py
```diff
@@ -1312,29 +1399,43 @@ def release_queued_tasks(self) -> None:
...
+                self.should_auto_restart_now()
+                and self.auto_restart_mode == AutoRestartMode.RESTART_NORMAL
+            ) or (
+                # Need to get preparing tasks to submit before reload
+                self.reload_pending
+            )
         ):
-            # Need to get preparing tasks to submit before auto restart
+            # don't release queued tasks, just finish processing any preparing
+            # ones
```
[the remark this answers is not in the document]
wxtim: just -> but if wanting to shorten by 1 extra char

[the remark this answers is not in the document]
oliver-sanders: > (Apologies for even suggesting this, but I hate dangling comments grimacing ).

Oh wow!

https://xkcd.com/276/

wxtim:
Using  a super-basic workflow I tried:

```
$ cylc vip workflow
$ sed 's@allow implicit tasks = True@allow implicit tasks = False@' -i ~/cylc-src/workflow/flow.cylc
$ cylc vr workflow
#  ... Fails horribly - it's now broken
$ cylc reinstall workflow
#  ... succeeds, does no validation checks
$ cylc reload workflow
Done
$ cylc log workflow 
# Contains the evidence that the reload did not in fact work.
```

Is that what you want? I think it's dangerous/misleading...

oliver-sanders:
Yes:

* `cylc vr` protects us from configuration issues.
* If users decide to bypass `cylc vr` and a config error is present then reload will log the error and leave the workflow paused.
* Cylc commands are async so don't return their completion status.

cylc/flow/scheduler.py
```diff
@@ -1945,8 +2044,14 @@ def resume_workflow(self, quiet: bool = False) -> None:
...
+                Whether to log anything in the event the workflow is not
+                paused.
+
         """
+        if self.reload_pending:
+            LOG.warning('Cannot resume - workflow is reloading')
+            return
         if not self.is_paused:
             if not quiet:
                 LOG.warning("Cannot resume - workflow is not paused")
```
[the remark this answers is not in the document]
oliver-sanders: I don't understand what the confusion is here. The workflow cannot be resumed right now, it is in the process of resuming.

oliver-sanders:
Fixed the test failures, the problem was that `update_workflow_structure` has a role to play in stall event detection, so I've come up with a cut-down version of the method which is enough to update the workflow status and status-message so that users can see what's going on, but doesn't try to update the task pool or mess with stall detection.

https://github.com/cylc/cylc-flow/compare/d5f331fba83b67acd5e12613e786051020859aa7..e7545d86ee27c0f6e51a226297299bbf5b1656d5

cylc/flow/scheduler.py
```diff
@@ -1931,12 +2020,22 @@ def check_auto_shutdown(self):
...
 
-    def pause_workflow(self) -> None:
-        """Pause the workflow."""
+    def pause_workflow(self, msg: Optional[str] = None) -> None:
+        """Pause the workflow.
+
+        Args:
+            msg:
+                A user-facing string explaining why the workflow was paused if
+                helpful.
```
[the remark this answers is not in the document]
hjoliver: Why not? Isn't the "if helpful" redundant? (We would put it there if it not helpful!)

CHANGES.md
```diff
@@ -21,6 +21,10 @@ issue which could cause jobs to fail if this variable became too long.
 
 ### Enhancements
 
+[#5992](https://github.com/cylc/cylc-flow/pull/5992) -
+The scheduler will now wait for preparing tasks to submit before attempting
+to perform a reload and will also pause/unpause the workflow.
```
wxtim, line 26: Suggestion: I think that `when you do x => then the scheduler will do y` is more readable.

```suggestion
Before trying to reload the workflow definition, the scheduler will
wait for preparing tasks to submit, and pause the workflow. 
After reload the scheduler will unpause the workflow.
```

cylc/flow/scheduler.py
```diff
@@ -897,8 +903,12 @@ def process_command_queue(self) -> None:
             sep = ', ' if kwargs_string and args_string else ''
             cmdstr = f"{name}({args_string}{sep}{kwargs_string})"
             try:
-                n_warnings: Optional[int] = self.get_command_method(name)(
-                    *args, **kwargs)
+                fcn = self.get_command_method(name)
+                n_warnings: Optional[int]
+                if inspect.iscoroutinefunction(fcn):
+                    n_warnings = await fcn(*args, **kwargs)
```
wxtim, line 909: Can we test this line?

oliver-sanders: Yes and are doing!

Without it `async def command_reload_workflow` wouldn't work.

wxtim (approved):
Happy

Issue #5598, opened 2023-06-21
DB: `task_pool` table does not get updated upon stall


MetRonnie:
### Description

Running tasks that succeed before the scheduler realises that it has stalled at this point do not get updated in the `task_pool` table.

### Reproducible Example

```ini
[scheduling]
    [[graph]]
        R1 = foo & bar
[runtime]
    [[foo]]
        script = sleep 3
    [[bar]]
        script = false
```

After running this workflow you will see in the log
```
INFO - [1/bar running job:01 flows:1] => failed
WARNING - [1/bar failed job:01 flows:1] did not complete required outputs: ['succeeded']
INFO - [1/foo running job:01 flows:1] => succeeded
ERROR - Incomplete tasks:
      * 1/bar did not complete required outputs: ['succeeded']
CRITICAL - Workflow stalled
```

But
```console
$ sqlite3 .service/db 'select * from task_pool'
1|bar|[1]|failed|0
1|foo|[1]|running|0
```

If you issue a command e.g. poll `1/bar` then the `task_pool` table gets updated the `1/foo` disappears as expected.

### Additional context

A similar thing was reported in #5239:
> the DB revealed an inconsistency with a task marked as running in the `task_pool table`, despite the succeeded output being present in the `task_outputs` table

however in that case the stall was supposedly erroneous

oliver-sanders:
Snippet obtained by `--main-loop='log db'` which shows that the `task_events` table is being updated, but the `task_pool` table isn't.

Probably just missing a DB call on task removal.

```sql
DELETE    
  FROM task_pool    
     
INSERT    
    OR REPLACE INTO task_pool    
VALUES('1', 'bar', '[1]', 'failed', 0)    
     
INSERT    
    OR REPLACE INTO task_pool    
VALUES('1', 'foo', '[1]', 'running', 0)    
     
DELETE    
  FROM task_prerequisites    
     
UPDATE task_states    
   SET time_updated = '2023-07-05T11:52:47+01:00',    
       submit_num = 1,    
       status = 'failed'    
 WHERE name == 'bar'    
   AND CYCLE == '1'    
   AND flow_nums == '[1]'    
     
DELETE    
  FROM task_timeout_timers    
     
COMMIT    
     
BEGIN    
     
INSERT    
    OR REPLACE INTO task_events    
VALUES('foo', '1', '2023-07-05T11:52:48+01:00', 1, 'succeeded', '')    
     
UPDATE task_jobs    
[...]

Issue #5593, opened 2023-06-19
workflow_params: potential bug when rapidly toggling parameters


oliver-sanders:
Possible bug spotted whilst working on #5592.

When we change a workflow parameter such as `is_paused`, we must update the DB, this involves inserting or deleting an entry as appropriate. We queue this operation for processing in the next main-loop cycle (or later on in this cycle). This means that inserts and deletes can occur in the same transaction. Problem is, they can interact!

Discovered in #5592 where `pause_workflow` and `resume_workflow` were happening in the same operation (reload) so were getting committed in the same transaction. The workflow ended up paused in the DB, but unpaused in the workflow state.

Integration test to prove the bug:

```python
async def test_is_paused(one_conf, flow, scheduler, run):
    def check_paused(_, row):
        key, value = row
        if key == 'is_paused' and value == '1':
            raise Exception('Workflow paused!')

    id_ = flow(one_conf)
    schd = scheduler(id_, paused_start=False)
    async with run(schd):
        assert schd.is_paused is False
        schd.pause_workflow()
        schd.resume_workflow()
        schd.process_workflow_db_queue()
        assert schd.is_paused is False

        schd.workflow_db_mgr.pri_dao.select_workflow_params(check_paused)
```

oliver-sanders:
For clarity as much as safety, if rather than deleting the `is_paused` entry, we set it to False (i.e. `'0'`).

This applies for all workflow parameters.

PR #5599, opened 2023-06-21
pool-process-queued-ops

oliver-sanders:
This `process_queued_ops` is called inside a `for` loop, it should probably be called afterwards.

But more importantly, `remove_if_complete`, which is called at the end of the function, calls `remove` which queues a DB operation which wasn't getting processed until later.

The main loop calls `process_queued_ops`:

https://github.com/cylc/cylc-flow/blob/4d787f445d5d9cfa950fbd5568e03af9f630e486/cylc/flow/scheduler.py#L1646

So this could cause the task to stick around in the pool table longer than intended, but with the correct task state so probably isn't related to #5598.

**Check List**

### DIFF
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1283,7 +1283,6 @@ class TaskPool:
                     (str(itask.point), itask.tdef.name, output))
                 self.workflow_db_mgr.put_insert_abs_output(
                     str(itask.point), itask.tdef.name, output)
-                self.workflow_db_mgr.process_queued_ops()
 
             c_taskid = Tokens(
                 cycle=str(c_point),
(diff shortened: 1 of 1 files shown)


hjoliver (approved):
Good spot. Pretty sure putting that inside the loop was a mistake.

hjoliver:
Huh, but there is a genuine integration test failure on this branch!

oliver-sanders:
Possibly related to #5593, I'm not sure the way we queue DB updates is completely safe.

oliver-sanders:
Now an investigation issue to understand why this change is causing issues which is likely a bug in itself.

oliver-sanders:
Replaced with issue: https://github.com/cylc/cylc-flow/issues/5800

PR #5600, opened 2023-06-23
dump: restrict window to n=0

oliver-sanders:
* This restores the old `cylc dump` behaviour of displaying only the pool contents.

This also fixes a bug where many n=0 tasks were reported as being in the n=1 window.

**Check List**

### DIFF
--- a/changes.d/5600.break.md
+++ b/changes.d/5600.break.md
@@ -1,0 +1,3 @@
+The `cylc dump` command now only shows active tasks (e.g. running & queued
+tasks). This restores its behaviour of only showing the tasks which currently
+exist in the pool as it did in Cylc 7 and earlier versions of Cylc 8.
--- a/cylc/flow/scripts/dump.py
+++ b/cylc/flow/scripts/dump.py
@@ -20,6 +20,9 @@
 
 Print information about a running workflow.
 
+This command can provide information about active tasks, e.g. running or queued
+tasks. For more detailed view of the workflow see `cylc tui` or `cylc gui`.
+
 For command line monitoring:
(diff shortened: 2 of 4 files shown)


hjoliver:
(Looks good, thanks. )

oliver-sanders:
Unfortunately, I spotted that the `depth` field which carries the n-window number for a task in the schema, doesn't appear to be giving the right answer.

I managed to fix it for many cases, but a few remain, especially when pushing out the graph-extent. I've written this up as a new issue here - #5609

So sadly, `cylc dump` isn't quite the task-pool reporting tool it once was, but we need to get this bug fixed because it will be needed for n-window filtering in the GUI so we aught to get this sorted out soon.

cylc/flow/data_store_mgr.py
```diff
@@ -937,8 +938,9 @@ def generate_ghost_task(
         tokens: Tokens,
         point,
         flow_nums,
+        depth,
```
MetRonnie, line 941: Docstring entry?

MetRonnie (commented):
In #5609 there seems to be some confusion about whether "depth" refers to number of graph edges or family inheritance depth, is it the same here?

oliver-sanders:
> In https://github.com/cylc/cylc-flow/issues/5609 there seems to be some confusion about whether "depth" refers to number of graph edges or family inheritance

Yep, drafting this for now.

oliver-sanders:
Blocked by #5609

MetRonnie (approved):

wxtim (approved):

Issue #5603, opened 2023-06-28
pool: infinite loop on startup


oliver-sanders:
This workflow:

```
[scheduler]                    
    allow implicit tasks = True    
                               
[scheduling]                   
    initial cycle point = 2000    
    runahead limit = P2Y       
    [[graph]]                  
        R1/2000 = a            
        P1Y = b[-P1Y] => b 
```

When run like this:

```console
cylc play --start-cycle-point=2005
```

Will get stuck in an infinite loop on startup during which it will not log any messages. This happens before the contact file is written making it look like the scheduler isn't running which can result in multiple schedulers running for the same workflow.

hjoliver:
Yikes, good catch.

PR #5604, opened 2023-06-28
pool: fix infinite loop with --start-cycle-point

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/5603
* Fixes an infinite loop which could occur where one or more task sequences terminate before the start point.
* This triggered a bug whereby the runahead limit was disabled on startup causing an infinite spawning bug.

**Check List**

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -699,8 +699,8 @@ class Scheduler:
             self.server.thread.start()
             barrier.wait()
 
-            await self.configure()
             self._configure_contact()
+            await self.configure()
         except (KeyboardInterrupt, asyncio.CancelledError, Exception) as exc:
             await self.handle_exception(exc)
 
@@ -1807,7 +1807,11 @@ class Scheduler:
         sys.stdout.flush()
         sys.stderr.flush()
 
-        if self.contact_data and self.task_job_mgr:
+        if (
+            self.workflow_db_mgr.pri_path
+            and Path(self.workflow_db_mgr.pri_path).exists()
+        ):
+            # only attempt remote tidy if the workflow has been started
             self.task_job_mgr.task_remote_mgr.remote_tidy()
 
         try:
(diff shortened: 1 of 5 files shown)


cylc/flow/scheduler.py
```diff
@@ -699,8 +699,8 @@ async def start(self):
             self.server.thread.start()
             barrier.wait()
 
-            await self.configure()
             self._configure_contact()
+            await self.configure()
```
oliver-sanders, line 703: Unrelated: Swap the order of these two so that the contact file gets written earlier.

This reduces the window of opportunity for two schedulers to start up for the same workflow. Taking a look at the contact file fields, I can't see anything set during `Scheduler.configure` which is used in `_configure_contact` so the order should be arbitrary.

cylc/flow/task_pool.py
```diff
@@ -312,12 +312,17 @@ def compute_runahead(self, force=False) -> bool:
...
-                seq.get_first_point(self.config.start_point)
-                for seq in self.config.sequences
-            })
+            points = [
+                point
+                for point in {
+                    seq.get_first_point(self.config.start_point)
+                    for seq in self.config.sequences
+                }
+                if point is not None
```
oliver-sanders, line 324: The problem was caused by a `None` value in this list. Due to the logic lower down, this caused the runahead limit point to be set to `None`.

cylc/flow/scheduler.py
```diff
@@ -1807,7 +1807,11 @@ async def _shutdown(self, reason: BaseException) -> None:
         sys.stdout.flush()
         sys.stderr.flush()
 
-        if self.contact_data and self.task_job_mgr:
```
oliver-sanders, line 1810: Related to the previous, because `contact_data` is set a little earlier, this test needed bodging a little as we could get into the situation where the workflow failed to startup, but did configure its contact file triggering remote-tidy to run on shutdown which caused the database to be created when it shouldn't have been.

Test changed to only run when the database is present, which is a better test, as remote-tidy needs database info to function.

hjoliver (approved):
All good.

wxtim (approved):
Well that's a horrid bug!

PR #5605, opened 2023-06-28
Added -z option

### LINKED ISSUE #5433: template vars: simpler syntax for defining a list of strings
### Template Variables

Currently we would define a list of strings like this:

```
-s 'MY_LIST=["foo", "bar", "baz"]'
```

This approach is highly flexible and supports other types e.g:

```
-s MIXED_LIST=["foo",42,True]
-s INT_SET={2,5,9}
-s DICT='{"foo": 1, "bar": 2, "baz": 3}'
```

### Rose Stem

The `rose stem` system has two options called `--tasks` and `--groups` which confusingly do the same thing. They set a Jinja2 variable called `RUN_NAMES` to a list of strings defined in comma separated format.

E.G. This:

```
--tasks=foo,bar,baz
```

Is actually equivalent to this:

```
-s RUN_NAMES='["foo", "bar", "baz"]'
```

So, the `--tasks` and `--groups` options are actually nothing to do with `rose stem` at all, they are really just a shorthand for specifying template variables to Cylc. This is a bit obstructive as these arguments are only supported by the `rose stem` command itself, which only runs `cylc install`.

### Future Rose Stem Interface?

It would make things a lot if we could pull the `--tasks` option out of `rose stem` and have it specified directly to Cylc. However, this option is a bit lengthy to write out in full.

We could make this easier by specifying a shortcut for defining lists of comma separated strings.

```
+s RUN_NAMES=foo,bar,baz
```

This might just be enough to allow us to deprecate the `--tasks` and `--groups` options (leaving them in as a wrapper of course), opening up this functionality to other Cylc commands e.g. `cylc validate`, `cylc graph` and `cylc vr`.

wxtim:
Closes #5433
Allow -z simple template variable string lists.

As part of this work I have replaced all instances of `load_template_vars` (which takes 3 keyword args) with `get_template_vars` (which takes a CLI options object and pulls the args for `load_template_vars` out). I have left the separation of the two functions because it reduces testing.

Additionally I have added error handling for this case:

`cylc vip -s "FOO"` - which previously failed with traceback. It now raises a CylcError subclass which is handled nicely.

**Check List**

- [x] https://github.com/cylc/cylc-doc/pull/618

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,10 @@ issue which could cause jobs to fail if this variable became too long.
 
 ### Enhancements
 
+-[#5605](https://github.com/cylc/cylc-flow/pull/5605) - A shorthand for defining
+-a list of strings - Before: `cylc command -s "X=['a', 'bc', 'd']"` - After:
+-`cylc command -z X=a,bc,d`.
+
 [#5405](https://github.com/cylc/cylc-flow/pull/5405) - Improve scan command
 help, and add scheduler PID to the output.
 
--- a/cylc/flow/option_parsers.py
+++ b/cylc/flow/option_parsers.py
@@ -22,7 +22,7 @@ from optparse import (
     OptionParser,
     Values,
     Option,
-    IndentedHelpFormatter
+    IndentedHelpFormatter,
 )
 import os
 import re
@@ -363,6 +363,15 @@ class CylcOptionParser(OptionParser):
             See `cylc help id` for more details.
     '''))
 
+    CAN_BE_USED_MULTIPLE = (
+        " This option can be used multiple times on the command line.")
+
+    NOTE_PERSIST_ACROSS_RESTARTS = (
+        " NOTE: these settings persist across workflow restarts,"
+        " but can be set again on the \"cylc play\""
+        " command line if they need to be overridden."
+    )
+
     STD_OPTIONS = [
         OptionSettings(
             ['-q', '--quiet'], help='Decrease verbosity.',
@@ -403,13 +412,27 @@ class CylcOptionParser(OptionParser):
                 " Values should be valid Python literals so strings"
                 " must be quoted"
                 " e.g. 'STR=\"string\"', INT=43, BOOL=True."
-                " This option can be used multiple "
-                " times on the command line."
-                " NOTE: these settings persist across workflow restarts,"
-                " but can be set again on the \"cylc play\""
-                " command line if they need to be overridden."
+                + CAN_BE_USED_MULTIPLE
+                + NOTE_PERSIST_ACROSS_RESTARTS
             ),
-            action='append', default=[], dest='templatevars', useif='jset'),
+            action='append', default=[], dest='templatevars', useif='jset'
+        ),
+        OptionSettings(
+            ['-z', '--set-list', '--template-list'],
+            metavar='NAME=VALUE1,VALUE2,...',
+            help=(
+                'Set the value of a Jinja2 template variable in the'
+                ' workflow definition as a comma separated'
+                ' list of Python strings.'
+                ' Values containing commas must be quoted.'
+                " e.g. '+s STR=a,b,c' => ['a', 'b', 'c']"
+                " or '+ s STR=a,\"b,c\"' => ['a', 'b,c']"
+                + CAN_BE_USED_MULTIPLE
+                + NOTE_PERSIST_ACROSS_RESTARTS
+            ),
+            action='append', default=[], dest='templatevars_lists',
+            useif='jset'
+        ),
         OptionSettings(
             ['--set-file'], metavar='FILE',
             help=(
@@ -418,9 +441,7 @@ class CylcOptionParser(OptionParser):
                 " pairs (one per line)."
                 " As with --set values should be valid Python literals "
                 " so strings must be quoted e.g. STR='string'."
-                " NOTE: these settings persist across workflow restarts,"
-                " but can be set again on the \"cylc play\""
-                " command line if they need to be overridden."
+                + NOTE_PERSIST_ACROSS_RESTARTS
             ),
             action='store', default=None, dest='templatevars_file',
             useif='jset'
--- a/cylc/flow/templatevars.py
+++ b/cylc/flow/templatevars.py
@@ -17,15 +17,21 @@
 
 from ast import literal_eval
 from optparse import Values
-from typing import TYPE_CHECKING, Any, Dict
+from typing import Any, Dict, List, Optional, Set, Union
 
+from cylc.flow import LOG
 from cylc.flow.exceptions import InputError
 from cylc.flow.rundb import CylcWorkflowDAO
 from cylc.flow.workflow_db_mgr import WorkflowDatabaseManager
 from cylc.flow.workflow_files import WorkflowFiles
 
-if TYPE_CHECKING:
-    from pathlib import Path
+from pathlib import Path
+
+
+OVERWRITE_WARNING = (
+    'Template variable {} redefined:'
+    ' the previous value will be overwritten.'
+)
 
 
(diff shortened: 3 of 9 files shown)


cylc/flow/templatevars.py
```diff
@@ -17,15 +17,21 @@
...
 from cylc.flow.workflow_db_mgr import WorkflowDatabaseManager
 from cylc.flow.workflow_files import WorkflowFiles
 
-if TYPE_CHECKING:
-    from pathlib import Path
+from pathlib import Path
+
+
+OVERWRITE_WARNING = (
+    'Template variable {} defined redefined:'
```
oliver-sanders, line 32: "defined redefined"!

cylc/flow/templatevars.py
```diff
@@ -77,22 +83,104 @@ def eval_var(var):
...
+
+
+def load_template_vars(
+    template_vars: Optional[List[str]] = None,
+    template_vars_file: Union[Path, str, None] = None,
+    templatevars_lists: Optional[List[str]] = None
+) -> Dict[str, Any]:
     """Load template variables from key=value strings."""
-    res = {}
+    keys: set = set()
```
oliver-sanders, line 118: If defining the type, might as well go for it:

```
keys: Set[str] = set()
```

cylc/flow/templatevars.py
```diff
@@ -77,22 +83,104 @@ def eval_var(var):
...
+    invalid_vars: str = ''
+
+    # Parse Template vars set by file (-S)
+    file_tvars: Dict = {}
```
oliver-sanders, line 122:
```
    file_tvars: Dict[str, Any] = {}
```

cylc/flow/templatevars.py
```diff
@@ -77,22 +83,104 @@ def eval_var(var):
...
+                try:
+                    key, val = line.split("=", 1)
+                except ValueError:
+                    invalid_vars += f'\n * {line}'
+                    continue
+                file_tvars[key.strip()] = eval_var(val.strip())
+                keys.add(key.strip())
+
+    cli_tvars: Dict = {}
+    tvars_lists: Dict = {}
```
oliver-sanders, line 138:
```
    cli_tvars: Dict[str, Any] = {}
    tvars_lists: Dict[str, str] = {}
```

tests/unit/test_loggingutil.py
```diff
@@ -34,7 +34,7 @@
     set_timestamps,
 )
 
-
+@pytest.mark.skip
```
oliver-sanders, line 37:
```suggestion
```

oliver-sanders (commented):
LGTM

hjoliver:
(Corrected linked Issue in description from 5543 to #5433)

cylc/flow/option_parsers.py
```diff
@@ -403,13 +412,28 @@ class CylcOptionParser(OptionParser):
...
+        ),
+        OptionSettings(
+            ['-z', '--set-list', '--template-list'],
+            metavar='NAME=VALUE1,VALUE2,...',
+            help=(
+                'Set the value of a Jinja2 template variable in the'
+                ' workflow definition as a comma separated'
+                ' list of Python strings.'
+                ' Values containing commas must be quoted, but this is'
+                ' not otherwise necessary.'
```
hjoliver, line 428:
```suggestion
                ' Values containing commas must be quoted.'
```

cylc/flow/templatevars.py
hjoliver, line 32:
```suggestion
    'Template variable {} redefined:'
```

cylc/flow/templatevars.py
```diff
@@ -17,15 +17,21 @@
...
 from cylc.flow.workflow_files import WorkflowFiles
 
-if TYPE_CHECKING:
-    from pathlib import Path
+from pathlib import Path
+
+
+OVERWRITE_WARNING = (
+    'Template variable {} defined redefined:'
+    ' previous definitions will be overwritten.'
```
hjoliver, line 33:
```suggestion
    ' the previous value will be overwritten.'
```

cylc/flow/templatevars.py
```diff
@@ -77,22 +83,104 @@ def eval_var(var):
         ) from None
 
 
-def load_template_vars(template_vars=None, template_vars_file=None):
+def parse_string_list(stringlist: str) -> List:
+    """Parse a string comma separated list into a python list of strings.
```
hjoliver, line 87:
```suggestion
    """Parse a comma separated string list into a Python string list.
```

cylc/flow/templatevars.py
```diff
@@ -77,22 +83,104 @@ def eval_var(var):
...
+        else:
+            res[key] = cli_tvars.get(
+                key, tvars_lists.get(
+                    key, file_tvars.get(key)))
+
+    # Raise an error if there are any key-clashes between tvars and tvars_list
+    if badkeys:
+        raise InputError(
+            'Setting the same variable using -s and -z is not permitted -'
+            '\nThe following clashing template variables:\n * '
```
hjoliver, line 180:
```suggestion
            "You can't set the same template variable with both -s and -z:"
```
Statements like "not permitted" not really necessary; that's implied by the error message itself. And the second sentence kind of restates the first.

cylc/flow/templatevars.py
```diff
@@ -77,22 +83,104 @@ def eval_var(var):
...
+            if key in result:
+                LOG.warning(OVERWRITE_WARNING.format(key))
+            result[key] = func(val)
+            keys.add(key)
+
+    # Raise an error if there are any args without the form key=value.:
+    if invalid_vars:
+        raise InputError(
+            'Template variables should be in the form key=value(s) -'
+            '\nThe following arguments did not validate:'
```
hjoliver, line 160:
```suggestion
            'Template variables must be set with key=value(s):'
```
Second sentence kind of restates the first.

hjoliver (changes requested):
👍 a few trivial changes suggested

tests/unit/test_loggingutil.py
wxtim: My current GPG setting breaks this test. :(

CHANGES.md
```diff
@@ -41,6 +41,21 @@ in `share/bin` and Python modules in `share/lib/python`.
...
+
+### Breaking Changes
+
+[#5600](https://github.com/cylc/cylc-flow/pull/5600) -
+The `CYLC_TASK_DEPENDENCIES` environment variable will no longer be exported
+in job environments if there are more than 50 dependencies. This avoids an
+issue which could cause jobs to fail if this variable became too long.
+
+### Enhancements
+
```
oliver-sanders, line 58: There is no 8.1.5

tests/unit/test_templatevars.py
```diff
@@ -161,3 +162,61 @@ def test_get_old_tvars_fails_if_cylc_7_db(tmp_path):
...
+            templatevars_lists=['BAZ', 'QUX']
+        )
+
+
+def test_load_template_vars_ValueError_fm_file(monkeypatch):
+    from io import StringIO
+    handle = StringIO('Hello')
+    monkeypatch.setattr('builtins.open', lambda *_: handle)
+    with pytest.raises(InputError, match='Hello'):
+        load_template_vars(template_vars_file='foo')
```
oliver-sanders, line 222: \n

oliver-sanders (approved):

tests/unit/test_templatevars.py
wxtim:
```suggestion
        load_template_vars(template_vars_file='foo')

```

wxtim:
I'm not sure what's going on with the Diff of CHANGES.md

hjoliver (approved):
👍 thanks @wxtim

PR #5617, opened 2023-07-05
ensure output from xtrigger scripts are logged

wxtim:
This closes a small issue not logged elsewhere:

Due to a change in the `__str__` method of the subprocess context object the results of printing to stdout/err in extrigger scripts was no longer logged.

To replicate  the problem on master try the [first Toy example of an xtrigger from the docs](https://cylc.github.io/cylc-doc/stable/html/user-guide/writing-workflows/external-triggers.html#echo)

**Check List**

- [x] ~`CHANGES.md`~ This is a really, really small change.
- [x] ~[Cylc-Doc](https://github.com/cylc/cylc-doc)~ Brings code back into line with docs.

### DIFF
--- a/cylc/flow/subprocctx.py
+++ b/cylc/flow/subprocctx.py
@@ -158,12 +158,3 @@ class SubFuncContext(SubProcContext):
         args = self.func_args + [
             "%s=%s" % (i, self.func_kwargs[i]) for i in skeys]
         return "%s(%s)" % (self.func_name, ", ".join([str(a) for a in args]))
-
-    def __str__(self):
-        return (
-            f'{self.func_name}('
-            f'{", ".join(map(str, self.func_args + list(self.func_kwargs)))}'
-            f'):{self.intvl}'
-        )
-
-    __repr__ = __str__
--- a/tests/integration/test_subprocctx.py
+++ b/tests/integration/test_subprocctx.py
@@ -20,13 +20,15 @@ import asyncio
 from cylc.flow import logging
 
 
-async def test_log_xtrigger_stdout(flow, scheduler, run_dir, run):
+async def test_log_xtrigger_stdout(
+    flow, scheduler, run_dir, start, log_filter
+):
     """Output from xtriggers should appear in the scheduler log:
 
     (As per the toy example in the Cylc Docs)
     """
     # Setup a workflow:
-    reg = flow({
+    id_ = flow({
         'scheduler': {'allow implicit tasks': True},
         'scheduling': {
             'graph': {'R1': '@myxtrigger => foo'},
@@ -34,7 +36,7 @@ async def test_log_xtrigger_stdout(flow, scheduler, run_dir, run):
         }
     })
     # Create an xtrigger:
-    xt_lib = run_dir / reg / 'lib/python/myxtrigger.py'
+    xt_lib = run_dir / id_ / 'lib/python/myxtrigger.py'
     xt_lib.parent.mkdir(parents=True, exist_ok=True)
     xt_lib.write_text(
         "from sys import stderr\n\n\n"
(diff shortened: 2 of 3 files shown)


tests/integration/test_subprocctx.py
```diff
@@ -0,0 +1,62 @@
...
+        xt_complete = False
+        while not xt_complete:
+            xt_complete = schd.xtrigger_mgr._get_xtrigs(
+                schd.pool.get_tasks()[0])[0][3]
+            await asyncio.sleep(1)
+
+        # Assert that both stderr and out from the print statement
+        # in our xtrigger appear in the log.
+        for expected in ['Hello World', 'Hello Hades']:
+            assert any(expected in i.message for i in log.records)
```
oliver-sanders, line 62: [FYI] There's a `log_filter` fixture which can help with this sort of thing.

tests/integration/test_subprocctx.py
```diff
@@ -0,0 +1,62 @@
...
+from cylc.flow import logging
+
+
+async def test_log_xtrigger_stdout(flow, scheduler, run_dir, run):
+    """Output from xtriggers should appear in the scheduler log:
+
+    (As per the toy example in the Cylc Docs)
+    """
+    # Setup a workflow:
+    reg = flow({
```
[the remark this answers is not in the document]
wxtim: Might want to check [your examples](https://github.com/cylc/cylc-flow/blob/5108f2555daac0023b4327106a266bc34de59743/tests/integration/test_examples.py#L34)? ;p

tests/integration/test_subprocctx.py
```diff
@@ -0,0 +1,62 @@
...
+    xt_lib.parent.mkdir(parents=True, exist_ok=True)
+    xt_lib.write_text(
+        "from sys import stderr\n\n\n"
+        "def myxtrigger():\n"
+        "    print('Hello World')\n"
+        "    print('Hello Hades', file=stderr)\n"
+        "    return True, {}"
+    )
+    schd = scheduler(reg)
+    async with run(schd, level=logging.DEBUG) as log:
```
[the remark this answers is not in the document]
wxtim: I tried that and couldn't get it to work. I probably need to process some messages or something·

tests/integration/test_subprocctx.py
```diff
@@ -0,0 +1,61 @@
...
+        schd.xtrigger_mgr.call_xtriggers_async(task)
+
+        # while not schd.xtrigger_mgr._get_xtrigs(task):
+        while schd.proc_pool.is_not_done():
+            schd.proc_pool.process()
+
+        # Assert that both stderr and out from the print statement
+        # in our xtrigger appear in the log.
+        for expected in ['Hello World', 'Hello Hades']:
+            assert log_filter(log, contains=expected)
```
MetRonnie, line 61: Do you want to test that this is at an expected logging level?

cylc/flow/subprocctx.py
```diff
@@ -158,12 +158,3 @@ def get_signature(self):
...
         return "%s(%s)" % (self.func_name, ", ".join([str(a) for a in args]))
-
-    def __str__(self):
-        return (
-            f'{self.func_name}('
-            f'{", ".join(map(str, self.func_args + list(self.func_kwargs)))}'
-            f'):{self.intvl}'
-        )
-
-    __repr__ = __str__
```
MetRonnie, line 169: I'm struggling to understand how removing this fixes the problem

MetRonnie (commented):
.

cylc/flow/subprocctx.py
wxtim: The `__str__` method in the class this one inherits from prints out stdout/err from the xtrigger script (stored in `self.err` and `self.out`). By overriding this we lost that information to the logging.

tests/integration/test_subprocctx.py
wxtim: Yes

cylc/flow/subprocctx.py
MetRonnie: Looks like it's been this way for 4 years according to git blame

MetRonnie (approved):

cylc/flow/subprocctx.py
MetRonnie: Don't know if @oliver-sanders would like to keep the `__repr__` method however?

oliver-sanders: Here's the commit which added the `__str__` method: https://github.com/cylc/cylc-flow/commit/c971b9605977f53c989cbfdedb06ce4f1d19ae72

This overrode `SubProcContext.__str__` which was previously being called and includes logic for including stdout/err.

I've got a feeling that the old `__str__` method causes traceback in some situations or something like that. I clearly didn't spot that it was being used in some `LOG.debug` statement and assumed it was an internal interface.

The `__repr__` method was for use in doctests.

tests/integration/test_subprocctx.py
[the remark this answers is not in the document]
oliver-sanders: Hence the FYI, try to avoid using `reg` when writing new code.

oliver-sanders (approved):

PR #5618, opened 2023-07-05
DB `workflow_params` table: store `None` instead of deleting/not storing params

### LINKED ISSUE #5593: workflow_params: potential bug when rapidly toggling parameters

MetRonnie:
Closes #5593

> Discovered in https://github.com/cylc/cylc-flow/pull/5592 where pause_workflow and resume_workflow were happening in the same operation (reload) so were getting committed in the same transaction. The workflow ended up paused in the DB, but unpaused in the workflow state.

So instead of deleting or not storing params, store them as `None` (which is an empty string for the `TEXT` type in sqlite). In the case of `is_paused` store it as a boolean (`'0'` or `'1'`).

Full backward compatibility is maintained.

There is no change to the DB structure so the https://github.com/cylc/cylc-flow/labels/db%20change label doesn't apply.

**Check List**

- [x] No dependency changes 
- [x] Tests are included
- [x] No docs change needed


cylc/flow/rundb.py
```diff
@@ -545,22 +545,18 @@ def select_broadcast_states(self, callback, sort=None):
         for row_idx, row in enumerate(self.connect().execute(stmt)):
             callback(row_idx, list(row))
 
-    def select_workflow_params(self, callback):
-        """Select from workflow_params.
+    def select_workflow_params(self) -> Iterable[Tuple[str, str]]:
```
oliver-sanders: :+1:

cylc/flow/scheduler.py
```diff
@@ -1214,62 +1214,66 @@ def _load_workflow_params(self, row_idx, row):
...
-            if self.is_restart and self.options.stopcp == 'reload':
-                LOG.debug(f"- stop point = {value} (ignored)")
-            elif self.options.stopcp is None:
-                self.options.stopcp = value
-                LOG.info(f"+ stop point = {value}")
-        elif key == self.workflow_db_mgr.KEY_RUN_MODE:
-            if self.options.run_mode is None:
+        LOG.info('LOADING workflow parameters')
+        for key, value in params:
+            if not value:
```
oliver-sanders:
```suggestion
            if value is None:
```

cylc/flow/scheduler.py
```diff
@@ -1214,62 +1213,66 @@ def _load_workflow_params(self, row_idx, row):
...
-            LOG.info('+ workflow UUID = %s', value)
-        elif key == self.workflow_db_mgr.KEY_PAUSED:
-            if self.options.paused_start is None:
-                self.options.paused_start = bool(value)
-                LOG.info(f'+ paused = {bool(value)}')
-        elif key == self.workflow_db_mgr.KEY_HOLD_CYCLE_POINT:
-            if self.options.holdcp is None:
+            elif key == self.workflow_db_mgr.KEY_UUID_STR:
+                self.uuid_str = value
+                LOG.info(f"+ workflow UUID = {value}")
```
oliver-sanders, line 1248: FYI, this is a negative change.

The way it was before, the string formatting only happened if the message was actually logged. With this change the string formatting happens, even if the message is not needed.

This is why the variables are provided as arguments to the logging function rather than `'+ workflow UUID = %s' % value`.

% formatting is still the documented advice.

https://docs.python.org/3/howto/logging.html#optimization

oliver-sanders (approved):

cylc/flow/scheduler.py
MetRonnie: Interesting to know 👍  However this method is only called once per start and reload so I don't think it is significant in this case.

oliver-sanders: I'm not asking for changes, but please leave these out of future tidying.

MetRonnie:
Ah, if you try to restart a workflow ran with this version of Cylc in an earlier version, you're going to get
> CRITICAL - Workflow shutting down - int() argument must be a string, a bytes-like object or a real number, not 'NoneType'

plus traceback

PR #5606, opened 2023-06-28
config: validate task messages and outputs

oliver-sanders:
At this moment users can do things like this:

```
[runtime]
  [[task]]
    [[[outputs]]]
      succeeded = failed  # shadowing builtin output, undefined behaviour
      a:b = abc  # not compatible with graph syntax
      required = required  # conflicts with cylc terminology
      BE $$ EF ££ WELL )(*& TON =  ?  # just nasty
```

* Task messages & outputs should be validated to avoid users shadowing built-in messages, outputs or qualifiers.
* Reserve the `_cylc` prefix save we need it later.

Also reserve the outputs `required`, `optional` and `all` as keywords for future use by `cylc set`, see [cylc set proposal](https://github.com/cylc/cylc-admin/pull/173). This proposal isn't approved yet and won't be implemented until Cylc 8.3.0, however, we may as well reserve these keywords now as the proposal will not change their meaning.

I've checked workflows at our end and have found shockingly little usage of custom outputs and no obvious breakages as a result of these changes.

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/617.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #5621, opened 2023-07-07
Added a :noindex: to an autodoc command

wxtim:
The addition of `.. autoclass:: cylc.flow.unicode_rules.TaskMessageValidator` in https://github.com/cylc/cylc-flow/pull/5606/files broke Cylc Doc because `TaskMessageValidator` referenced in autodocs twice. I've undone this, which should allow the nightly build against `master` to succeed. The problem with the 8.1.x branch is different.

Reviewer should build docs locally to check.

**Check List**
- [x] This is a small PR enabling the documentation build to work. As such I believe that it's reasonable to replace the normal checklist.


wxtim:
Not required - duplicates work by @oliver-sanders

PR #5623, opened 2023-07-10
Ensure that UUID is retrieved from database on workflow restart.

### LINKED ISSUE #5615: Workflow contact UUID changes on restart
When restarting a workflow, the contact file UUID does not match the database UUID. But it is supposed to preserve the UUID from the DB.

A new UUID is always created at `Scheduler.__init__()` which seems dodgy:
https://github.com/cylc/cylc-flow/blob/125c4434f863cdcf3cabf0d088931684ac5a001a/cylc/flow/scheduler.py#L279

(This compounded with https://github.com/cylc/cylc-uiserver/pull/463, causes workflows played in the UI to appear remaining as stopped when they have started)

wxtim:
Closes #5615 

**Check List**

- [x] ~`CHANGES.md` entry included if this is a change that can affect users~ Not currently a  user facing change.

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -700,7 +700,10 @@ class Scheduler:
         Lightweight wrapper for testing convenience.
 
         """
-        params = self.load_workflow_params_and_tmpl_vars()
+        if self.is_restart:
+            params = self.load_workflow_params_and_tmpl_vars()
+        else:
+            params = []
 
         try:
             await self.initialise()
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -424,7 +424,7 @@ class Scheduler:
         self._check_startup_opts()
 
         if self.is_restart:
-            self._parse_workflow_params(params)
+            self._set_workflow_params(params)
 
         self.profiler.log_memory("scheduler.py: before load_flow_file")
         try:
@@ -1134,7 +1134,7 @@ class Scheduler:
             self.reload_pending = 'applying the new config'
             old_tasks = set(self.config.get_task_name_list())
             # Things that can't change on workflow reload:
-            self._parse_workflow_params(
+            self._set_workflow_params(
                 self.workflow_db_mgr.pri_dao.select_workflow_params()
             )
             self.apply_new_config(config, is_reload=True)
@@ -1315,10 +1315,10 @@ class Scheduler:
             'CYLC_WORKFLOW_FINAL_CYCLE_POINT': str(self.config.final_point),
         })
 
-    def _parse_workflow_params(
+    def _set_workflow_params(
         self, params: Iterable[Tuple[str, Optional[str]]]
     ) -> None:
-        """Load a row in the "workflow_params" table in a restart/reload.
+        """Set workflow params on restart/reload.
 
         This currently includes:
         * Initial/Final cycle points.
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -716,9 +716,9 @@ class Scheduler:
             self.server.thread.start()
             barrier.wait()
 
+            self._configure_contact()
             await self.configure()
             self.task_events_mgr.uuid_str = self.uuid_str
-            self._configure_contact()
         except (KeyboardInterrupt, asyncio.CancelledError, Exception) as exc:
             await self.handle_exception(exc)
 
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -406,7 +406,7 @@ class Scheduler:
 
         self.profiler = Profiler(self, self.options.profile_mode)
 
-    async def configure(self):
+    async def configure(self, params):
         """Configure the scheduler.
 
         * Load the flow configuration.
@@ -424,9 +424,7 @@ class Scheduler:
         self._check_startup_opts()
 
         if self.is_restart:
-            self.load_workflow_params_and_tmpl_vars()
-        else:
-            self.uuid_str = str(uuid4())
+            self._parse_workflow_params(params)
 
         self.profiler.log_memory("scheduler.py: before load_flow_file")
         try:
@@ -548,15 +546,6 @@ class Scheduler:
         self.process_workflow_db_queue()
 
         self.profiler.log_memory("scheduler.py: end configure")
-
-    def load_workflow_params_and_tmpl_vars(self) -> None:
-        """Load workflow params and template variables"""
-        with self.workflow_db_mgr.get_pri_dao() as pri_dao:
-            # This logic handles lack of initial cycle point in flow.cylc and
-            # things that can't change on workflow restart/reload.
-            self._load_workflow_params(pri_dao.select_workflow_params())
-            pri_dao.select_workflow_template_vars(self._load_template_vars)
-            pri_dao.execute_queued_items()
 
     def log_start(self) -> None:
         """Log headers, that also get logged on each rollover.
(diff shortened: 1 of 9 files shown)


oliver-sanders:
It's a little more complex than this. The UUID is already being loaded on restart:

https://github.com/cylc/cylc-flow/blob/master/cylc/flow/scheduler.py#1330-L1333

But it's not happening at the right time (search for `.uuid_str`) e.g:

https://github.com/cylc/cylc-flow/blob/6b4055d0e3eb454971ae90c8e8a0386fedd76ced/cylc/flow/scheduler.py#L386-L407

To close #5615, we need to remove the UUID which gets set in `Scheduler.__init__` (only required on first run) and ensure that the UUID gets set on the `Scheduler` before it gets accessed.

This is kinda similar to the `TaskProxy.platform` bug where having an erroneous default set by `__init__` has covered over runtime errors disguising a problem until we spotted behaviour differences.

cylc/flow/scheduler.py
```diff
@@ -360,6 +364,9 @@ async def initialise(self):
         * Initialise managers.
 
         """
+        if self.is_restart:
+            self.load_workflow_params_and_tmpl_vars()
+
```
MetRonnie, line 369: I think moving this will cause problems - and it looks like it's resulted in several test failures

cylc/flow/scheduler.py
```diff
@@ -307,6 +306,11 @@ def __init__(self, reg: str, options: Values) -> None:
             pub_d=os.path.join(self.workflow_run_dir, 'log')
         )
         self.is_restart = Path(self.workflow_db_mgr.pri_path).is_file()
+
+        # In the case of a restart get UUID from DB:
+        if not self.is_restart:
+            self.uuid_str = str(uuid4())
```
MetRonnie, line 312: Possibly we will just have to load the UUID from the DB here if it's a restart

wxtim: That's what I did originally, and Oliver Objected to.

oliver-sanders: > That's what I did originally, and Oliver Objected to.

Oliver objected to it being done two ways in two places with potential for mixups.

> But still need to carefully check the implications of moving self.load_workflow_params_and_tmpl_vars()

Yes, very careful!

tests/integration/test_scheduler.py
```diff
@@ -292,3 +294,26 @@ def mock_auto_restart(*a, **k):
...
+       * Create a new scheduler for the same workflow and check that it has
+         retrieved the UUID from the Daatabase.
+    """
+    uuid_re = re.compile('CYLC_WORKFLOW_UUID=(.*)')
+    contact_file = Path(one.workflow_run_dir) / '.service/contact'
+    async with start(one):
+        original_uuid = uuid_re.findall(contact_file.read_text())[0]
+    schd = scheduler(one.workflow_name, paused_start=True)
+    async with start(schd):
+        assert schd.uuid_str == original_uuid
```
oliver-sanders, line 319: This test passes on master!

The issue is more nuanced, the UUID str *is* getting loaded correctly on master, but *after* the point when the contact file is written.

To detect the old bug you would need to compare the contact file of the second run, maybe something like this:

```python
num_starts = 3
uuids = []
for _ in range(num_starts):
    schd = scheduler(one.workflow_name, paused_start=True)
    async with start(schd):
        uuids.append(schd.uuid_str)
        uuids.append(uuid_re.findall(contact_file.read_text())[0])

assert len(uuids) == num_starts
assert len(set(uuids)) == 1
```

Note there are some other things which use the UUID too, we should probably test those too.

cylc/flow/scheduler.py
oliver-sanders: That would not be surprising, there is a lot of order sensitivity here.

tests/integration/test_scheduler.py
wxtim: !  And I spent ages branch flicking to try and avoid that!

MetRonnie (commented):
Looks good, I have some suggestions at https://github.com/wxtim/cylc/pull/56

MetRonnie (approved):

cylc/flow/scheduler.py
```diff
@@ -716,8 +716,9 @@ async def start(self):
             self.server.thread.start()
             barrier.wait()
 
-            self._configure_contact()
             await self.configure()
+            self.task_events_mgr.uuid_str = self.uuid_str
+            self._configure_contact()
```
oliver-sanders, line 721: `_configure_contact` should happen before `configure`.

https://github.com/cylc/cylc-flow/commit/0aadc909aae6a687545e665854340ef1cb794a54

tests/integration/test_scheduler.py
```diff
@@ -296,6 +298,34 @@ def mock_auto_restart(*a, **k):
...
+
+    async with start(one):
+        pass
+
+    schd = scheduler(one.workflow_name, paused_start=True)
+    async with start(schd):
+        # UUID in contact file should be the same as that set in the database
+        # and the scheduler.
+        cf_uuid = uuid_re.findall(contact_file.read_text())[0]
+        assert schd.uuid_str == cf_uuid
```
oliver-sanders, line 326:
```suggestion
        cf_uuid = uuid_re.findall(contact_file.read_text())
        assert cf_uuid == [schd.uuid_str]
```

cylc/flow/scheduler.py
oliver-sanders: It's nasty to do it this way around because another scheduler can startup for the same workflow whilst Cylc is loading the configuration which lead to a reported bug where three schedulers were running for the same workflow simultaneously.

oliver-sanders: * Various template variables are required at various stages of startup.
* `Scheduler` logic is likely sensitive to the timing of `load_workflow_params_and_tmpl_vars` during startup.

> Can we accept the extra database call that I proposed in an earlier iteration to simply get the UUID at this point?

Does this mean we need two database calls to fetch these variables, no.

E.G:

```
def load_db_params(self):
    params = []
    self.workflow_db_mgr.select_workflow_params(params.append)
    return params

def start(self):
    ...
    params = self.load_db_params()
    ...
    self._configure_contact(..., params)
    self.configure(..., params)

def load_workflow_params_and_tmpl_vars(self, params):
    for key, value in params:
        ...
```

cylc/flow/scheduler.py
```diff
@@ -426,7 +424,7 @@ async def configure(self):
         self._check_startup_opts()
 
         if self.is_restart:
-            self.load_workflow_params_and_tmpl_vars()
+            self._parse_workflow_params(params)
```
oliver-sanders, line 427:
```suggestion
            self._set_workflow_params(params)
```

cylc/flow/scheduler.py
```diff
@@ -1124,7 +1134,7 @@ async def command_reload_workflow(self) -> None:
             self.reload_pending = 'applying the new config'
             old_tasks = set(self.config.get_task_name_list())
             # Things that can't change on workflow reload:
-            self._load_workflow_params(
+            self._parse_workflow_params(
```
oliver-sanders, line 1137:
```suggestion
            self._set_workflow_params(
```

cylc/flow/scheduler.py
```diff
@@ -1305,7 +1315,7 @@ def apply_new_config(self, config, is_reload=False):
             'CYLC_WORKFLOW_FINAL_CYCLE_POINT': str(self.config.final_point),
         })
 
-    def _load_workflow_params(
+    def _parse_workflow_params(
         self, params: Iterable[Tuple[str, Optional[str]]]
     ) -> None:
         """Load a row in the "workflow_params" table in a restart/reload.
```
oliver-sanders, line 1321:
```suggestion
    def _set_workflow_params(
        self, params: Iterable[Tuple[str, Optional[str]]]
    ) -> None:
        """Set workflow params on restart/reload.
```

oliver-sanders (approved):
:+1:

wxtim:
@MetRonnie  - It's three weeks since you approved - want to have a double check?

MetRonnie:
Can you get the tests to run? The last commit included `[skip ci]` in the description so either add a changelog commit or change the description and force push

wxtim:
> Can you get the tests to run? The last commit included `[skip ci]` in the description so either add a changelog commit or change the description and force push

Yes, barring CodeCov.

MetRonnie (approved):

PR #5624, opened 2023-07-11
misc

oliver-sanders:
Misceallaneous stuff from recent reviews...

terminology: workflow-name => workflow-id                                                                                          
    
* Workflow name still exists as an install concept in relation to the source    
  directory, and in the run directory where it is the workflow ID with    
  the run name removed.    
    
terminology: reg->id                               
                                                            
* Reg was a Cylc <= 6 concept relating to the registration DB.    
* Beyond Cylc 6 we didn't have a better name for this.    
* With the universal ID, workflow reg became workflow ID.    
    
tests/integration: set run_mode=simulation as the default    
    
* Integration tests shouldn't go running workflows for real.    
* Set the simulation run mode as the default, this avoids shell    
  interaction and reduces the number of moving parts.  

config: fill small coverage gap

**Check List**

### DIFF
--- a/cylc/flow/install.py
+++ b/cylc/flow/install.py
@@ -583,7 +583,7 @@ def check_nested_dirs(
             install dirs.
 
     Raises:
-        WorkflowFilesError if reg dir is nested inside a run dir, or an
+        WorkflowFilesError if id_ dir is nested inside a run dir, or an
             install dirs are nested.
     """
     if install_dir is not None:
--- a/cylc/flow/install.py
+++ b/cylc/flow/install.py
@@ -547,8 +547,8 @@ def detect_flow_exists(
     """Returns True if installed flow already exists.
 
     Args:
-        run_path_base: Absolute path of workflow directory,
-            i.e ~/cylc-run/<workflow_id>
+        run_path_base: Absolute path of the parent of the workflow's run dir,
+            i.e ~/cylc-run/<workflow_name>
         numbered: If True, will detect if numbered runs exist. If False, will
             detect if non-numbered runs exist, i.e. runs installed
             by --run-name.
@@ -583,8 +583,8 @@ def check_nested_dirs(
             install dirs.
 
     Raises:
-        WorkflowFilesError if id_ dir is nested inside a run dir, or an
-            install dirs are nested.
+        WorkflowFilesError if run_dir is nested inside an existing run dir,
+            or install dirs are nested.
     """
     if install_dir is not None:
         install_dir = Path(os.path.normpath(install_dir))
--- a/cylc/flow/install.py
+++ b/cylc/flow/install.py
@@ -548,7 +548,7 @@ def detect_flow_exists(
 
     Args:
         run_path_base: Absolute path of workflow directory,
-            i.e ~/cylc-run/<workflow_name>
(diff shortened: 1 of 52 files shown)


wxtim (approved):
Looks good. Mostly simple, but with the test changes needs a second review IMO.

cylc/flow/scripts/install.py
```diff
@@ -27,13 +27,13 @@
 
 If a SOURCE_NAME is supplied, Cylc will search for the workflow source in the
 list of directories given by "global.cylc[install]source dirs", and install
-the first match. The installed workflow name will be the same as SOURCE_NAME,
+the first match. The installed workflow ID will be the same as SOURCE_NAME,
```
MetRonnie, line 30: Not this one I think

```suggestion
the first match. The installed workflow name will be the same as SOURCE_NAME,
```
or perhaps
```suggestion
the first match. The installed workflow ID will start with SOURCE_NAME,
```

cylc/flow/pathutil.py
```diff
@@ -157,7 +157,7 @@ def make_localhost_symlinks(
     """Creates symlinks for any configured symlink dirs from glbl_cfg.
     Args:
         rund: the entire run directory path
-        named_sub_dir: e.g workflow_name/run1
+        named_sub_dir: e.g workflow_id/run1
```
MetRonnie, line 160:
```suggestion
        named_sub_dir: e.g workflow_name/run1
```

cylc/flow/pathutil.py
```diff
@@ -204,7 +204,7 @@ def get_dirs_to_symlink(
 
     Args:
         install_target: Symlinks to be created on this install target
-        flow_name: full name of the run, e.g. myflow/run1
+        flow_id: full id of the run, e.g. myflow/run1
```
MetRonnie, line 207:
```suggestion
        workflow_id: full id of the run, e.g. myflow/run1
```

cylc/flow/install.py
```diff
@@ -583,7 +583,7 @@ def check_nested_dirs(
             install dirs.
 
     Raises:
-        WorkflowFilesError if reg dir is nested inside a run dir, or an
+        WorkflowFilesError if id_ dir is nested inside a run dir, or an
             install dirs are nested.
```
MetRonnie, line 587:
```suggestion
        WorkflowFilesError if run_dir is nested inside an existing run dir,
            or install dirs are nested.
```

cylc/flow/install.py
```diff
@@ -548,7 +548,7 @@ def detect_flow_exists(
 
     Args:
         run_path_base: Absolute path of workflow directory,
-            i.e ~/cylc-run/<workflow_name>
+            i.e ~/cylc-run/<workflow_id>
```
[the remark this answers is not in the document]
oliver-sanders: Ah right, I'll change the arg description to avoid "workflow directory".

cylc/flow/install.py
```diff
@@ -548,7 +548,7 @@ def detect_flow_exists(
 
     Args:
         run_path_base: Absolute path of workflow directory,
```
oliver-sanders, line 550:
```suggestion
        run_path_base: Absolute path of the parent of the workflow's run dir,
```

PR #5631, opened 2023-07-18
Fix `cylc clean` remote re-invocation bug

### LINKED ISSUE #5628: `cylc clean` remote re-invocation fails if `flow.cylc` detected in sub dir
### Description

A user encountered this case: when running `cylc clean` without using the `--yes`/`-y` option, the re-invocation on the remote host exited 1 because it detected a `flow.cylc` file somewhere in the `work` directory and required the `-y` option to continue.

We need to somehow not scan in the remote re-invocation. Maybe add a `--no-scan` option to the `cylc clean` command?

Re-invoking with `-y` is not enough as it would have lead to an incomplete clean (only the `work` directory would have been removed).

### Reproduceable example

```ini
# flow.cylc
[scheduling]
    [[graph]]
        R1 = foo
[runtime]
    [[foo]]
        platform = <remote-platform>
        script = touch flow.cylc # created in work dir
```
After playing, run `cylc clean`

MetRonnie:
Closes #5628

**Check List**

- [x] No dependency changes 
- [x] Tests are updated
- [x] Changelog entry included (as a towncrier fragment - see #5634)

### DIFF
--- a/cylc/flow/scripts/clean.py
+++ b/cylc/flow/scripts/clean.py
@@ -121,7 +121,7 @@ def get_option_parser():
     parser.add_option(
         '--timeout',
         help=("The number of seconds to wait for cleaning to take place on "
-              "remote hosts before cancelling."),
+              r"remote hosts before cancelling. Default: %default."),
         action='store', default='120', dest='remote_timeout'
     )
 
--- a/cylc/flow/scripts/clean.py
+++ b/cylc/flow/scripts/clean.py
@@ -60,6 +60,7 @@ Examples:
 """
 
 import asyncio
+from optparse import SUPPRESS_HELP
 import sys
 from typing import TYPE_CHECKING, Iterable, List, Tuple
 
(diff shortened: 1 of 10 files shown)


cylc/flow/scripts/clean.py
```diff
@@ -120,10 +121,17 @@ def get_option_parser():
     parser.add_option(
         '--timeout',
         help=("The number of seconds to wait for cleaning to take place on "
-              "remote hosts before cancelling."),
+              r"remote hosts before cancelling. Default: %default."),
```
oliver-sanders, line 124: Nice to document the defaults, I have no idea why optparse does not do this by "default".

We are already defining our own "help formatter", looks like we could override `expand_default` to add a "Default: %default" bit onto the end of every option where a default is set rather than having to add it to each option manually. (for future work).

oliver-sanders (approved):

wxtim (approved):

PR #5634, opened 2023-07-20
Use towncrier for changelog generation

MetRonnie:
https://towncrier.readthedocs.io/en/stable/index.html

If we go for this it means no more changelog conflicts 🎉 Instead you either
- write your changelog entries to a file at `changes.d/<PR-number>.<break|feat|fix>.md`
- or run `towncrier create <PR-number>.<break|feat|fix>.md --content "<insert content here>"`
 
and before the release they get collated into `CHANGES.md`.

Also it guarantees changelog entries go under the right release section despite any milestone bump-backs or cherry-picks that happen.

**Check List**

- [x] Dependency changes do not apply to `conda-environment.yml`
- [x] Tests N/A
- [x] No cylc-doc change needed

### DIFF
--- a/.github/workflows/1_create_release_pr.yml
+++ b/.github/workflows/1_create_release_pr.yml
@@ -43,14 +43,13 @@ jobs:
           init-file: 'cylc/flow/__init__.py'
           pypi-package-name: 'cylc-flow'
 
-      - name: Update "released on" date in changelog
-        continue-on-error: true
-        uses: cylc/release-actions/stage-1/update-changelog-release-date@v1
-        with:
-          changelog-file: 'CHANGES.md'
-
       - name: Test build
         uses: cylc/release-actions/build-python-package@v1
+
+      - name: Generate changelog
+        run: |
+          python3 -m pip install -q towncrier
+          towncrier build --yes
 
       - name: Create pull request
         uses: cylc/release-actions/stage-1/create-release-pr@v1
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,23 @@
+[tool.towncrier]
+directory = "changes.d"
+name = "Cylc"
+package = "cylc.flow"
+filename = "CHANGES.md"
+template = "changes.d/changelog-template.jinja"
+underlines = ["", "", ""]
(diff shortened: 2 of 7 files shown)


MetRonnie:
Here is the `towncrier build` generated changelog section from a bunch of example fragment files I added

image

MetRonnie:
In response to a question Dave asked: although these changelog fragments will be included into the master branch in sync PRs, once we do a maintenance release and subsequent sync PR, the deleted fragments will also be removed on master so you won't get duplicates

.github/workflows/1_create_release_pr.yml
```diff
@@ -43,15 +43,14 @@ jobs:
...
-        with:
-          changelog-file: 'CHANGES.md'
-
       - name: Test build
         uses: cylc/release-actions/build-python-package@v1
 
+      - name: Generate changelog
+        run: |
+          python3 -m pip install -q towncrier
+          towncrier build --yes
```
oliver-sanders, line 52: Should we also run this on fast tests to avoid letting broken fragments getting to release?

pyproject.toml
```diff
@@ -0,0 +1,23 @@
...
+directory = "break" # NB this is just the filename not directory e.g. 123.break.md
+name = "Breaking Changes"
+showcontent = true
+[[tool.towncrier.type]]
+directory = "feat"
+name = "Enhancements"
+showcontent = true
+[[tool.towncrier.type]]
+directory = "fix"
+name = "Fixes"
```
oliver-sanders, line 22: Daft suggestion:

```suggestion
name = "⚠ Breaking Changes"
showcontent = true
[[tool.towncrier.type]]
directory = "feat"
name = "🚀 Enhancements"
showcontent = true
[[tool.towncrier.type]]
directory = "fix"
name = "🐛 Fixes"
```

oliver-sanders (approved):
Willing to give it a go.

pyproject.toml
MetRonnie: I would prefer 🔧 for fixes personally

hjoliver (approved):
Nice 👍

PR #5639, opened 2023-07-25
Tweak swarm configure. [skip ci]

hjoliver:
It silently fails to modify `.ssh/config` if the file doesn't already exist, or if it exists but is empty.

(One review will do; no need for tests or change log entry).


hjoliver:
Cancelled pointless CI run.

oliver-sanders (approved):

oliver-sanders:
Not an entirely pointless CI run, I've triggered one of the swarm jobs just to make sure (I can't access Docker ATM).

MetRonnie:
Style test failing

PR #5647, opened 2023-07-25
Bump dev version 8.3.0.dev

MetRonnie:


hjoliver:
(Do we need to update the release action to get this right automatically?)

oliver-sanders (approved):

MetRonnie:
We could potentially have it create the maintenance series branch and bump the dev versions differently for each of that and master...

PR #5648, opened 2023-07-25
Fix style

MetRonnie:
Follow-up to #5639

### DIFF
--- a/etc/bin/swarm
+++ b/etc/bin/swarm
@@ -204,7 +204,7 @@ append_config () {
         if prompt "Write \"$LINE\" to \"$LOC\"?"; then
             if [[ "$POS" == top ]]; then
                 # ... at the top of the file
-                echo -e "${LINE}\n$(cat ${LOC})" > "$LOC"
+                echo -e "${LINE}\n$(cat "$LOC")" > temp && mv temp "$LOC"
             elif [[ "$POS" == bottom ]]; then
                 # ... at the bottom of the file
                 echo -e "\n$LINE" >> "$LOC"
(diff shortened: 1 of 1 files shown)


hjoliver:
Sorry, didn't realize we were style checking under `etc`!  🤦

etc/bin/swarm
```diff
@@ -204,7 +204,7 @@ append_config () {
         if prompt "Write \"$LINE\" to \"$LOC\"?"; then
             if [[ "$POS" == top ]]; then
                 # ... at the top of the file
-                echo -e "${LINE}\n$(cat ${LOC})" > "$LOC"
+                echo -e "${LINE}\n$(cat "$LOC")" > temp && mv temp "$LOC"
```
hjoliver, line 207: I get the quoting, but is the temp file necessary?

hjoliver, line 207:
```suggestion
                echo -e "${LINE}\n$(cat "$LOC")" > "$LOC"
```

hjoliver: @MetRonnie I'm gonna merge this without the temp file, to stop other CI runs failing on the shellcheck. You can always propose another one-liner if you disagree!

hjoliver (approved):

etc/bin/swarm
MetRonnie: Funny, shellcheck was complaining about SC2094 locally

> Make sure not to read and write the same file in the same pipeline

But apparently not on GH Actions?

hjoliver: Ah OK, my local shellcheck must be outdated, didn't see that.  However, going by my logic above, SC2094 does not apply for this particular construct.

Issue #5595, opened 2023-06-20
data store: task state count too high


oliver-sanders:
The task state count appears to include tasks which are outside of the n=1 window.

E.G. for the workflow defined here:

https://github.com/cylc/cylc-ui/issues/1322#issuecomment-1594819686

Cylc TUI shows one succeeded task in its tree view, but the state counts show four succeeded tasks:

![Screenshot from 2023-06-20 11-27-50](https://github.com/cylc/cylc-flow/assets/16705946/a9ed41e7-e140-4dbd-8f7e-aff59f9a39b8)

Cylc dump agress with the Tui tree view:

![Screenshot from 2023-06-20 11-28-06](https://github.com/cylc/cylc-flow/assets/16705946/5795e717-2891-413e-8859-f1eaec889bd6)

PR #5231, opened 2022-11-16
Timeout on restarting completed workflows

hjoliver:
**Check List**

- [ ] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [ ] Contains logically grouped changes (else tidy your branch by rebase).
- [ ] Does not contain off-topic changes (use other PRs for other changes).
- [ ] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.
- [ ] Tests are included (or explain why tests are not needed).
- [ ] `CHANGES.md` entry included if this is a change that can affect users
- [ ] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [ ] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #5650, opened 2023-07-27
Attempt to undo a small logic change preventing the data store being …

wxtim:
…tidied

**Check List**


oliver-sanders:
@dwsutherland any thoughts on how this should work?

dwsutherland:
> @dwsutherland any thoughts on how the `has_updated` variable should play with this data store method?

I think is because the `update_data_structure` method also includes non-data-store processing:
```
        # Database update
        self.workflow_db_mgr.put_task_pool(self.pool)
        self.update_data_store()
```

cylc/flow/scheduler.py
```diff
@@ -1846,18 +1847,16 @@ def _update_workflow_state(self):
...
+        if self.data_store_mgr.publish_pending:
+            self.data_store_mgr.publish_pending = False
+            self.server.publish_queue.put(
+                self.data_store_mgr.publish_deltas)
+            # Non-async sleep - yield to other threads rather
+            # than event loop
+            sleep(0)
         # Database update
         self.workflow_db_mgr.put_task_pool(self.pool)
         self.update_data_store()
```
dwsutherland, line 1862: There might be an issue here... that method sets `self.data_store_mgr.updates_pending` to `True` .. Which seems kinda circular if there's no criteria other than `update_data_structure` running..

wxtim: I'm not sure I understand - what symptoms of this might I see? Are you saying that I should keep `if self.data_store_mgr.updates_pending:` here too, or that the way it was before my change also seems sketchy? If so, could you have a look into it.

dwsutherland: Because `self.update_data_store()` sets `self.data_store_mgr.updates_pending` to `True`, then:
```
 if has_updated or self.data_store_mgr.updates_pending:
```
will run `update_data_structure` every main loop now including the DB update (which isn't a data-store thing)..

dwsutherland: (actually not sure why `self.update_data_store()` is even called in `update_data_structure`)

oliver-sanders: > will run update_data_structure every main loop now

Yikes!

wxtim: @hjoliver  - do you know why you added `self.update_data_store` in #5231 [link to diff](https://github.com/cylc/cylc-flow/commit/4a6b447fdbd4c28e6bdecaf3814a4d4cdda9652b#r123264068)? 

I don't think there's any need to call it, but I'm going to try some in-practice tests with it locally.

dwsutherland:
Think I reproduced this problem.. by re-triggering a succeeded task:
image

Just doing this:
```
(flow) sutherlander@cortex:cylc-flow$ git diff

index 3e1b40403..20c6c2d4d 100644

--- a/cylc/flow/scheduler.py

+++ b/cylc/flow/scheduler.py

@@ -1755,7 +1755,7 @@ class Scheduler:

                     self.timers[self.EVENT_RESTART_TIMEOUT].stop()

                     self.is_restart_timeout_wait = False

 

-            if has_updated:

+            if has_updated or self.data_store_mgr.updates_pending:

                 # Update the datastore.

                 await self.update_data_structure(self.is_reloaded)

 

@@ -1860,7 +1860,6 @@ class Scheduler:

                 sleep(0)

         # Database update

         self.workflow_db_mgr.put_task_pool(self.pool)

-        self.update_data_store()

 

     def check_workflow_timers(self):

         """Check timers, and abort or run event handlers as configured."""

```
appeared to fix it

dwsutherland:
But I think you have it now.. will review.

cylc/flow/scheduler.py
```diff
@@ -1774,6 +1775,8 @@ async def main_loop(self) -> None:
                     with suppress(KeyError):
                         self.timers[self.EVENT_STALL_TIMEOUT].stop()
 
+                self.update_data_store()
```
dwsutherland, line 1778: Don't think this is needed.. As there's nothing in this block that requires data-store change, and if something else changes downstream it should message the data-store with a delta.

cylc/flow/scheduler.py
```diff
@@ -1846,21 +1849,18 @@ def _update_workflow_state(self):
...
+        if self.data_store_mgr.publish_pending:
+            self.data_store_mgr.publish_pending = False
+            self.server.publish_queue.put(
+                self.data_store_mgr.publish_deltas)
+            # Non-async sleep - yield to other threads rather
+            # than event loop
+            sleep(0)
         # Database update
         self.workflow_db_mgr.put_task_pool(self.pool)
-        self.update_data_store()
```
dwsutherland, line 1863: Don't think this should have been here in the first place.

dwsutherland (changes requested):
Probably just remove that `self.update_data_store()` call under `has_updated` ..
This actually may have effected the state-totals (..etc), i.e. #5595   (but I have a PR for that)

cylc/flow/scheduler.py
```diff
@@ -1755,10 +1755,11 @@ async def main_loop(self) -> None:
                     self.timers[self.EVENT_RESTART_TIMEOUT].stop()
                     self.is_restart_timeout_wait = False
 
-            if has_updated:
+            if has_updated or self.data_store_mgr.updates_pending:
                 # Update the datastore.
                 await self.update_data_structure(self.is_reloaded)
 
+            if has_updated:
```
dwsutherland, line 1762: Good idea to separate this block from data store update 👍

dwsutherland:
This problem was quite pernicious, I think it made some tests flakier too (by creating a race condition with state updates .. i.e. `tests/f/cylc-show/06-past-present-future.t`)

dwsutherland (approved):
LGTM 👍

oliver-sanders (approved):
LGTM:

* Reproduced the original issue, confirmed this fixed it.
* Jammed print statements in to check that the data store update methods were being called as expected.

Issue #5640, opened 2023-07-25
optional outputs: implement new proposal


oliver-sanders:
Supersedes #5423 
Supersedes #5361 

Implement the extended [optional outputs proposal](https://cylc.github.io/cylc-admin/proposal-optional-output-extension.html)

* ~Require at least one optional output to be generated in order for the task to be considered "completed" by default.~
  * ~Update docs.~
  * ~There is a breaking change (see proposal)~
* Add the ability for users to manually specify the completion expression.
  * Advertise on Discourse.
* Bring task expiry into the optional outputs model.
  * Document in changelog.

PR #5651, opened 2023-07-27
optional outputs extension

oliver-sanders:
**Implements:** https://cylc.github.io/cylc-admin/proposal-optional-output-extension.html
**Closes:** #5640

* Require at least one optional output to be generated in order for the task to be considered "completed" by default.
* Add the ability for users to manually specify the completion expression.
* Bring task expiry into the optional outputs model. Document in changelog.

**Check List**


tests/unit/test_graph_parser.py
```diff
@@ -840,8 +841,33 @@ def test_fail_family_triggers_on_tasks(ftrig):
     gp = GraphParser()
     with pytest.raises(GraphParseError) as cm:
         gp.parse_graph(f"foo:{ftrig} => bar")
-        assert (
```
oliver-sanders, line 843: [unrelated] Broken assertion.

cylc/flow/host_select.py
oliver-sanders, line 1: Tangentially related.

We were already using restricted Python evaluation to run the host selection ranking expressions, so I generalised this interface to allow for new uses going forward. This logic in this file should be unchanged.

cylc/flow/task_outputs.py
oliver-sanders, line 1: The actual logic of the change lies here, most of the other changes are validation.

cylc/flow/task_pool.py
```diff
@@ -1374,12 +1384,11 @@ def remove_if_complete(self, itask):
...
                 self.release_runahead_tasks()
         else:
-            incomplete = itask.state.outputs.get_incomplete()
-            if incomplete:
+            if itask.state.outputs.is_incomplete():
                 # Retain as incomplete.
                 LOG.warning(
                     f"[{itask}] did not complete required outputs:"
-                    f" {incomplete}"
+                    f" {itask.state.outputs.get_incomplete()}"
```
oliver-sanders, line 1391: After these changes `bool(get_incomplete()) != is_complete()`.

E.G. if `completion = succeeded and (x or y)` and `succeeded` and `x` have been generated, `get_incomplete()` will return `y`.

tests/unit/test_task_outputs.py
```diff
@@ -13,30 +13,254 @@
...
+):
+    """Return a TaskDef-like object for completion expression purposes.
 
+    Args:
+        opt_outputs: List of optional outputs or None.
+        req_outputs: List of required outputs or None.
+        completion: Custom completion expression or None.
+        set_implicit: Set implicit outputs (not done for removed tasks).
 
-class TestMessageSorting(unittest.TestCase):
```
oliver-sanders, line 22: [unrelated] converted from unit test.

hjoliver:
(That was quick!!)

oliver-sanders:
The functional bit of the code is shockingly simple, validation proved trickier.

cylc/flow/task_outputs.py
```diff
@@ -48,6 +54,119 @@
...
+    ands.sort()
+    ors.sort()
+
+    # join the lists of "ands" and "ors" into statements
+    _ands = ''
+    if ands:
+        _ands = ' and '.join(ands)
+    _ors = ''
+    if ors:
+        _ors = ' or '.join(ors)
```
MetRonnie, line 119: More simply
```suggestion
    _ands = ' and '.join(ands)
    _ors = ' or '.join(ors)
```

cylc/flow/task_outputs.py
```diff
@@ -48,6 +54,119 @@
...
 
 
+# this evaluates task completion expressions
+CompletionEvaluator = restricted_evaluator(
+    # expressions
+    ast.Expression,
+    # variables
+    ast.Name, ast.Load,
+    # operations
+    ast.BoolOp, ast.And, ast.Or, ast.BinOp,
```
MetRonnie, line 64: Why `BinOp`?

MetRonnie (commented):
Partial review

cylc/flow/config.py
```diff
@@ -1127,6 +1131,100 @@ def check_param_env_tmpls(self):
...
+            if not expr:
+                # no expression to validate
+                continue
+
+            if 'submit-failed' in expr:
+                raise WorkflowConfigError(
+                    f'Error in [runtime][{name}]completion:'
+                    f'\nUse "submit_failed" rather than "submit-failed"'
+                    ' in completion expressions.'
+                )
```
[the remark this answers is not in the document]
oliver-sanders: An error will be raised if `-` chars appear in the expression as these nodes are not whitelisted.

I added this to provide a more helpful error message for the one built-in output which contains a hyphen character.

cylc/flow/task_outputs.py
oliver-sanders: Try taking it out, you'll get:

```
"BoolOp" not permitted
```

cylc/flow/config.py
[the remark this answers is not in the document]
MetRonnie: I see but the error message is
```
Invalid expression: succeeded and x-y
"Sub" not permitted
```
which is fairly unclear IMO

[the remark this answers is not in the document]
oliver-sanders: Which is why I added this message!

tests/integration/test_optional_outputs.py
```diff
@@ -0,0 +1,476 @@
...
+        assert task_a.state.outputs._completion_expression == (
+            'succeeded or failed'
+        )
+
+        # so the workflow should not stall if 1/a fails
+        task_a.state_reset(TASK_OUTPUT_FAILED)
+        assert not schd.pool.is_stalled()
+
+
+async def test_completion_expiression(
```
MetRonnie, line 165:
```suggestion
async def test_completion_expression(
```

tests/integration/test_optional_outputs.py
```diff
@@ -0,0 +1,476 @@
...
+    monkeypatch.setattr('cylc.flow.flags.cylc7_back_compat', False)
+    validate(id_)
+
+    # it should fail if we reference outputs which haven't been registered
+    config['runtime']['a']['completion'] = 'succeeded or (failed and x)'
+    id_ = flow(config)
+    with pytest.raises(WorkflowConfigError) as exc_ctx:
+        validate(id_)
+    assert '[runtime][a]completion' in str(exc_ctx.value)
+    assert "Input 'x' is not defined" in str(exc_ctx.value)
```
MetRonnie, line 253: Worth testing the case of a typo? E.g. `suceeded`

cylc/flow/config.py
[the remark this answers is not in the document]
MetRonnie: Can we not just check if the string contains `-` and give an error saying use underscore instead?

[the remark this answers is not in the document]
oliver-sanders: This example should error because subtraction is not an allowed operation:

```
succeeded - failed
```

This example should fail because they should replace `-` with `_`:

```
my-output
```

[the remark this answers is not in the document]
MetRonnie: IMO it's less confusing overall if we always error with "`-` invalid, use underscore instead", even if the user meant to do subtraction (whatever that would even mean to them)

tests/integration/test_optional_outputs.py
MetRonnie: Actually that's the same as this test

tests/integration/test_config.py
```diff
@@ -239,7 +239,7 @@ def test_parse_special_tasks_families(flow, scheduler, validate, section):
                 section: 'FOO(P1D)',
             },
             'graph': {
-                'R1': 'foo & foot',
+                'R1': 'foo? & foot?',
```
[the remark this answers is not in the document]
MetRonnie: Is this clock-expired? Is that what the special tasks section indicates? (sorry not too familiar with this aspect of Cylc)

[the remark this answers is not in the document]
oliver-sanders: > Is this clock-expired

Yes, it's defined in the pytest params above.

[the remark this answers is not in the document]
MetRonnie: Oh of course 🤦‍♂️

oliver-sanders:
(FYI, I'll apply cosmetic suggestions once functional review is complete)

hjoliver:
(I'll look at this ASAP...)

cylc/flow/util.py
```diff
@@ -134,3 +136,204 @@ def serialise(flow_nums: set):
...
+
+    Raised in the event non-whitelisted syntax is detected in an expression.
+    """
+
+    def __init__(self, node):
+        self.node = node
+
+
+class RestrictedNodeVisitor(ast.NodeVisitor):
+    """AST node visitor which error on non-whitelisted syntax.
```
hjoliver, line 152:
```suggestion
    """AST node visitor which errors on non-whitelisted syntax.
```

cylc/flow/util.py
```diff
@@ -134,3 +136,204 @@ def serialise(flow_nums: set):
...
+
+    Raises _RestrictedEvalError if a non-whitelisted node is visited.
```
hjoliver, line 154: No biggie, but lines 152 and 154 are pretty close to redundant - combine them?

cylc/flow/util.py
```diff
@@ -134,3 +136,204 @@ def serialise(flow_nums: set):
...
+) -> Callable:
+    """Returns a Python eval statement restricted to whitelisted operations.
+
+    The "eval" function can be used to run arbitrary code. This is useful
+    but presents security issues. This returns an "eval" method which will
+    only allow whitelisted operations to be performed allowing it to be used
+    safely with user-provided input.
+
+    The code passed into the evaluator will be parsed into an abstract syntax
+    tree AST), then that tree will be executed using Python's internal logic.
```
hjoliver, line 180:
```suggestion
    tree (AST), then that tree will be executed using Python's internal logic.
```

tests/unit/test_graph_parser.py
hjoliver: Am I blind or have you replaced it with an identical one?

tests/integration/test_optional_outputs.py
hjoliver: I really wish "expiression" was a word though.

tests/integration/test_optional_outputs.py
```diff
@@ -0,0 +1,476 @@
...
+        itask.state.outputs.set_completion(TASK_OUTPUT_SUCCEEDED, True)
+
+        # the task should *not* be removed from the pool as complete
+        schd.pool.remove_if_complete(itask)
+        assert mock_remove.call_count == 0
+
+        # complete the "x" output
+        itask.state.outputs.set_completion('x', True)
+
+        # the task *should* not be removed from the pool as complete
```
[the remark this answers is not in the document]
hjoliver: Or, "the task should *be* removed from..." 🤣

tests/integration/test_optional_outputs.py
```diff
@@ -0,0 +1,476 @@
...
+        'Use "submit_failed" rather than "submit-failed"'
+    ) in str(exc_ctx.value)
+
+    # it should accept "submit_failed"
+    config['runtime']['a']['completion'] = 'submit_failed'
+    id_ = flow(config)
+    validate(id_)
+
+
+def test_clock_expire_infers_optional_expiry(
```
hjoliver, line 294:
```suggestion
def test_clock_expire_implies_optional_expiry(
```

hjoliver (commented):
Very nice code (as usual 😩 ). Not tested yet, I'll approve tomorrow if I fail to break it!

tests/unit/test_graph_parser.py
oliver-sanders: I've just dedented the assertion so that it gets run.

```python
with pytest.raises(Exception):
    raise Exception()
    print('this code will never get called')
print('this code will always get called')
```

tests/integration/test_optional_outputs.py
oliver-sanders: Should have got this testing expiry!

cylc/flow/cfgspec/workflow.py
```diff
@@ -961,6 +961,115 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+            Conf('completion', VDR.V_STRING, desc='''
+                Define the condition for task completion.
+
+                The completion condition is evaluated when a task is finished,
+                it is a validation check which confirms that the task has
+                generated the outputs it was expected to.
+
+                If the task fails this check it is considered
+                :term:`incomplete` and will cause the workflow to
+                :term:`stall` altering you that something has gone wrong which
```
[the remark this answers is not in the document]
oliver-sanders: > if there is nothing else to run

OR if the task causes the workflow to hit the runahead limit.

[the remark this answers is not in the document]
oliver-sanders: s/will/may/?

cylc/flow/config.py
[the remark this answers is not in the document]
oliver-sanders: Ok, added a catch all for this.

hjoliver:
Testing all good.

One question occurred to me though.  It should have done earlier when reviewing the proposal, sorry. Still, better late than never. I'll illustrate with the xy(z) case:

```
foo:x? => x
foo:y? => y
```
If that's the entire graph off of `foo`, it makes pretty good sense that at least one of x or y must be generated by default, even though we've said they're both optional. We can document that that's for safety reasons, to avoid unplanned early shutdown of the workflow.

But what about this:
```
foo:x? => x
foo:y => y
```
or this:
```
foo:x? => x
foo => bar
```
In these cases early shutdown is not on the cards, and we have said that `x` is optional, but if `foo` does not generate `x` we get:
```
ERROR - Incomplete tasks:
      * 1/foo did not complete required outputs: ['x']
```
IMO this is overkill and confusing when we do not need to protect against early shutdown.

So perhaps we should change tack slightly: *if a task has **only** optional outputs in the graph, then by default at least one of them must be generated*.

oliver-sanders:
```
foo:x? => x
foo:y => y
```

```
completion = succeeded and y
```

---

```
foo:x? => x
foo => bar
```

```
completion = succeeded
```

oliver-sanders:
> if a task has only optional outputs in the graph, then by default at least one of them must be generated.

That wouldn't work as `:succeeded` is almost always explicitly specified in order to trigger the task e.g:

```
a => b
b:x? => x
b:y? => y
```

Making chains is a more common use case than breaking chains, we must prioritise the more likely use case at the expense of one line of configuration for the less likely. Note we've yet to encounter a use case for early shutdown.

hjoliver:
> Note we've yet to encounter a use case for early shutdown, we should support it, but it shouldn't be the default, 

I agree (since the proposal discussions) - note I was not suggesting this above. I was suggesting we should take the optional output declaration seriously (by default) IF there required outputs to prevent early shutdown.

>  Do this:
>    completion = succeeded 

I know that, as I said above, I'm talking about the default situation. 

> That wouldn't work as :succeeded is almost always explicitly specified in order to trigger the task e.g:

Fair point. Our foo:xyz examples tend to have foo as the first task in the chain.

But that still leaves the (default!) single optional output case looking illogical:

```
foo:x? => x
foo => bar
```

By default this is exactly equivalent to:
```
foo:x => x
foo => bar
```

i.e. the optional out is by default not optional. Does that not bother you?

We could avoid that with a small tweak: 
- *if there are MULTIPLE optional outputs then (by default) at least one must be generated*

(Equivalently: if there are any optional outputs - including just one - there will be some optionality by default).

This is not a massive deal as it is only the default; but before we lock it in we should make sure we don't have default behavour that seems surprising or illogical to users.

hjoliver:
A small follow-up. 

Note I'm fully on board now (since the proposal discussion) with requiring at least one optional output to be generated *to prevent early shutdown*.

But I'm still not convinced on the "chain breaking" argument, if early shut down is prevented by other (required) outputs.

> Making chains is a more common use case than breaking chains, we must prioritise the more likely use case at the expense of one line of configuration for the less likely.

Here the graph won't end if none of the optional outputs are generated because another path extends from `foo:succeed`:
```
foo:x? => chain-x
foo:y? => chain-y
foo:z? => chain-z
...
foo => chain-bar
```

So by requiring one of them, we are arbitrarily allowing ALL but ONE of the chains to be broken.  If there are 26 chains, we allow (by default) 25 to be broken.  That's hardly a chain breaking safety net. 

So ... I think the most logically justifiable defaults would be EITHER:

  1.  do not require any optional outputs to be generated, UNLESS that could could result in early shutdown
OR
  2. require ALL optional outputs to be generated unless a completion condition is supplied 

In the second case, the optional syntax `foo:x?` becomes an indicator that the output may be optional - to see exactly in what sense it is optional you have to look for the completion condition.

oliver-sanders:
> require all chains to be unbroken, to prevent any inadvertent chain breaking (and as a side-effect, inadvertent shutdown)
>
> i.e. require ALL optional outputs to be generated, unless a completion condition explicitly states their optionality
> here the ? syntax becomes an indicator that the output can be optional

This approach effectively shifts optional output logic into the "runtime" section completely, the question marks in the graph would become superfluous, only validation would keep them in sync with the completion expression so they could be removed completely. Not necessarily a bad approach, however, this would be a substantial change which would break all graph branching in existing Cylc 8 workflows.

> How is that to be justified exactly

It's the safest default which provides the safest behaviour in the most common scenarios. More complex requirements require more complex configuration, but the simple remains simple until you break it.

hjoliver:
> This approach effectively shifts optional output logic into the "runtime" section completely, the question marks in the graph would become superfluous, only validation would keep them in sync with the completion expression so they could be removed completely. 

Maybe not. We could keep them in the graph to signify that there should be a corresponding completion condition. Outputs with no `?`  are definitively required.

> Not necessarily a bad approach, however, this would be a substantial change which would break all graph branching in existing Cylc 8 workflows.

A safe breaking change though: the workflow would carry on as normal from the generated output; the task would just be marked incomplete. (And, it is already a breaking change without this).

> > 
> 
>    How is that to be justified exactly
> 
> It's the safest default which provides the safest behaviour in the most common scenarios. More complex requirements require more complex configuration, but the simple remains simple until you break it.

My two suggestions are equally simple, but without the downsides:
- by default, at least one used-in-the-graph output (optional or required) must be generated - to prevent inadvertent early shutdown
- or, all used-in-the-graph outputs must be generated - to prevent any inadvertent chain breaking (and early shutdown)

Otherwise, use a completion condition.

What's not simple about those?

MetRonnie:
Would it be least confusing to users if we *don't* mandate at least 1 optional output be generated, instead leaving it up to them to write a completion condition if needed?

An exception could be made for if `succeeded` is optional, in which case we might want to mandate at least 1 optional output be generated, to ensure one path is taken?

hjoliver:
> Would it be least confusing to users if we don't mandate at least 1 optional output be generated, instead ...
> An exception could be made for if succeeded is optional ...  to ensure one path is taken?

Yes, that's exactly what I'm suggesting as my preferred approach.

The other option (no-chains-broken-by-default) is more heavy handed, and not my preference, but needs to be considered.

oliver-sanders:
> Would it be least confusing to users if we don't mandate at least 1 optional output be generated, instead leaving it up to them to write a completion condition if needed?
>
>> Yes, that's exactly what I'm suggesting as my preferred approach.

Note, Hillary's suggestion was "one or more utilised outputs must be generated", which isn't quite the same and does have it's own drawbacks e.g. for this graph `a:x? => x`, the optional output `a:x` would still be required.

IMO, the proposed approach is nicer (the scope is restricted to optional outputs, the behaviour more closely caters to the principle "switch" use cases for graph branching).

The runtime only approach described above as "no-chains-broken-by-default" (i.e. if you want something to be optional you must explicitly configure it in the completion expression) has merit. We did consider going this way back at the beginning. The separation of `scheduling` and `runtime` is kinda a core principle of Cylc so it feels a bit wrong in some ways but makes things very explicit which would be beneficial. [Un]fortunately [depending on position] this would break all existing Cylc 8 graph branching which wouldn't make it the most popular change.

oliver-sanders:
Either way, this PR implements the accepted proposal, so I think review should continue whilst discussion goes on (in a proposal?) so we're in the place to release this when we've decided.

hjoliver:
> Note, Hillary's suggestion was "one or more utilised outputs must be generated", which isn't quite the same 

Sorry, my reading of @MetRonnie's question was a bit off. Yes, my suggestion was "we don't mandate at least 1 optional output be generated" **but also**, as Oliver says, "one or more utilized outputs must be generated" - in order to prevent early shutdown if a task "accidentally" fails to generate any optional outputs.

> and does have it's own drawbacks e.g. for this graph `a:x? => x` [if there are no other outputs used from `a`] the optional output `a:x` would still be required.

That's not a drawback or an inconsistency, because the rule is *"at least one utilized output must be generated"* (by default; and any output not just optional) *"in order to prevent early shutdown"*.

On this branch, the interpretation of this graph `a:x? => x; a => b` is a drawback, because it makes `:x?` NOT optional (by default) despite that fact that if a task has only one optional output is **must be** truly optional in principle.

hjoliver:
> The runtime only approach described above as "no-chains-broken-by-default" (i.e. if you want something to be optional you must explicitly configure it in the completion expression) has merit.

I think we had better make a decision on this one way or the other, very soon. It would be too disruptive to switch to that approach after wide adoption of Cylc 8.

Pros:
- no ambiguity
- very safe: any/all chain-breaking has to be explicit
- some users have found the `?` notation difficult to understand?

Cons:
- arguably the `?` notation would not be needed
   - this will be jarring to those who have already adopted it
   - IMO the notation is good, and helpful in most cases
- heavy handed 
   -  any optionality requires an explicit completion expression 
   -  prioritizing safety over power, by default, is not necessarily desirable (`rm -i` is not the default, etc.)

I only raised this for completeness, as one of the possible approaches that don't result in the problem optional-but-not-optional inconsistency.  IMO it is probably unnecessarily heavy handed, so my vote is NO. But not a hard NO.

hjoliver:
Assuming we do dump the heavy-handed solution from consideration, here's a tweak that I think could resolve the impasse!

```
foo:x? => x
foo:y? => y
```
We say that the above notation means "restricted optional outputs", with the restriction being:
- `foo` must generate at least one of its optional outputs (unless you supply a completion condition)

The justification for this is that *optional outputs are usually used to direct the flow one way or another, in which case no outputs at all indicates something means something must be wrong*.

So far so good - that is this branch 🎉 

The problem with that is it forces you to write a completion condition for *truly optional* outputs, and in particular the default looks downright stupid for the single-optional output case where the "one way or another argument" is irrelevant. 

So we could support a variant notation for *unrestricted optional outputs*:
```
foo:x?? => x
```
This notation means:
- the scheduler does not care one way or another if `:x` is generated
- `x` is not allowed to appear in completion conditions 
- we document (and even warn in validation) that this means the branch can end and potentially the workflow can shut down early if nothing else is going on
- this allows deliberate branch-ending without a completion condition, but now it is explicit in the graph (which is even more explicit, than hidden under runtime in a completion condition)

oliver-sanders:
Blocked whilst the proposal is under re-review.

PR #5655, opened 2023-07-28
event handlers: improve docs

oliver-sanders:
Goes with https://github.com/cylc/cylc-doc/pull/632

Better linking into docs, straighten out similar doc labels.


wxtim (approved):
Looks good.

Issue #5609, opened 2023-06-29
data_store_mgr: n-window depth field


oliver-sanders:
# Edit

I had misinterpreted what the `depth` field meant, turns out it was inheritance depth not window depth, have added a description into the schema to clarify.

This issue now relates to the problem of reporting the n-window depth via GraphQL.

This will be required in order to filter tasks by different n-window levels e.g:

* #5600
* https://github.com/cylc/cylc-ui/issues/864

----

# Original Issue (invalid)

Discovered whilst working on https://github.com/cylc/cylc-flow/pull/5600, the `depth` field (which holds the n-window value of a task) is giving some funky results.

> **Note:** Results in this issue were obtained on #5600 which includes a fix where n=0 tasks were reporting as depth=1.

### Reproducible Example:

In this example:

```ini
[scheduling]                   
    runahead limit = P2
    initial cycle point = 1
    cycling mode = integer                                                      
    [[graph]]                  
        P1 = """               
            a => b => c => d    
            b[-P1] => b        
        """                    
                               
[scheduler]                    
    allow implicit tasks = True  
```

Then set the window extent to, say, 5.

Several tasks have the wrong depth number.

E.G. `2/c` has `depth=3`, but it should actually be `depth=2`:

* `2/a` is n=0
* `2/b` is n=1
* `2/c` is n=2  # we're getting n=3
* `2/d` is n=3  # we're getting n=4

### Problem

The issue is that the same task can have different n-window numbers when the graph is traced from different n=0 nodes.

E.G. There are three routes from the n=0 window to the task `2/c`:

* Via `1/a`:
  * `1/a` n=0
  * `1/b` n=1
  * `2/b` n=2
  * `2/c` n=3  # Reported
* Via `2/a`:
  * `2/a` n=0
  * `2/b` n=1
  * `2/c` n=2  # Correct
* Via `3/a` (runahead)
  * `3/a` n=0
  * `3/b` n=1
  * `2/b` n=2
  * `2/c` n=3  # Reported

At present the algorithm is iterating in the following order (window-extent=5):

* `1/a`  # iterate from n=0 to n=1
* `1/b`
* `2/a`
* `2/b`
* `3/a` 
* `3/b`
* `4/a` 
* `4/b`
* `1/a`  # then go back to n=0 and flesh things out from there
* `1/b`
* `1/c`
* `1/d`
* `2/b`
* `2/c`
* `2/d`
* `3/b`
* `3/c`
* `3/d`

[...]

Issue #5485, opened 2023-04-24
increment_graph_window: don't revisit nodes


oliver-sanders:
Avoid re-visiting nodes in the `increment_graph_window` algorithm.

This will help to reduce the number of recursive calls and will allow us to efficiently include "cousins" in the n>1 window.

> **Note:** Cousins
> For the graph `a => <b> => c`, if `b1` is active, then `b2` is a cousin which should be included in the `n=2` window.

See this comment and the larger issue for discussion:

> Not revisiting nodes
>
> we can do this by storing graph relations in an easily referenced way (if we don't already)
> adopting graph sections visited, and skipping walks to head of adopted section (if required)
>
> --- https://github.com/cylc/cylc-flow/issues/5435#issuecomment-1515703855

MetRonnie:
Closed by #5600

PR #5660, opened 2023-08-01
Data store avoid graph re-walk

dwsutherland:
addresses #5485, #5609

At present we've resolved an efficiency problem at n=1 window size by restricting the graph walk to only parents or only children (i.e. 'c' 'p', or 'ccc' 'ppp' for n=3) .. However this excludes cousins, which is evident at n>1 (i.e. 'cp' or 'pc' at n=2)

This restriction was a cheap way to avoid graph re-walks.. #5475

Here the problem is actually fixed at any window size, by changing the graph walk algorithm from a recursive to a walk location approach.. From the method's description:

> Fills in graph walk from existing walks if possible, otherwise expands
the graph front from whereever hasn't been walked.
Walk nodes are grouped into locations which are tag according to
parent child path, i.e. 'cpc' would be children-parents-children away
from active/start task. Which not only provide a way to cheaply rewalk,
but also the edge distance from origin.
The futherest child boundary nodes are registered as prune triggers for
the origin's parents, so when they become active the parents are
assessed for pruning eligibility.

Window resizing will now happen as soon as it's set (not as the graph progresses)

N-window edge depth of nodes is now found/updated with new field `graph_depth`.. A good example workflow:
```
[scheduling]
    initial cycle point = 2020
   [[graph]]
      P1Y = """
         a => b => c => d => e => f => g
         b => h => i => f
         g[-P1Y] => a
      """

[runtime]
    [[root]]
        script = sleep 2

    [[b, c, d, e, f, i]]

    [[a]]
        script = """
set +e

read -r -d '' gqlDoc <<_DOC_
{"request_string": "
mutation {
  setGraphWindowExtent (
    workflows: [\"${CYLC_WORKFLOW_ID}\"],
    nEdgeDistance: 4) {
    result
  }
}",
"variables": null}
_DOC_

echo "${gqlDoc}"

cylc client "$CYLC_WORKFLOW_ID" graphql < <(echo ${gqlDoc}) 2>/dev/null


set -e
"""

    [[g]]
        script = """
# Slow 2nd branch down
sleep 20
"""

    [[h]]
        script = """
# Slow 2nd branch down
sleep 20

"""
```
image
Gives:
```
{
  "data": {
    "workflows": [
      {
        "id": "~sutherlander/pug/run1",
        "taskProxies": [
          {
            "id": "~sutherlander/pug/run1//20200101T0000Z/a",
            "state": "succeeded",
            "graphDepth": 2
          },
          {
            "id": "~sutherlander/pug/run1//20200101T0000Z/b",
            "state": "succeeded",
            "graphDepth": 1
          },
          {
            "id": "~sutherlander/pug/run1//20200101T0000Z/h",
            "state": "running",
            "graphDepth": 0
          },
          {
            "id": "~sutherlander/pug/run1//20200101T0000Z/c",
            "state": "running",
            "graphDepth": 0
          },
          {
            "id": "~sutherlander/pug/run1//20200101T0000Z/i",
            "state": "waiting",
            "graphDepth": 1
          },
          {
            "id": "~sutherlander/pug/run1//20200101T0000Z/d",
            "state": "waiting",
            "graphDepth": 1
          },
          {
            "id": "~sutherlander/pug/run1//20200101T0000Z/f",
            "state": "waiting",
            "graphDepth": 2
          },
          {
            "id": "~sutherlander/pug/run1//20200101T0000Z/e",
            "state": "waiting",
            "graphDepth": 2
          },
          {
            "id": "~sutherlander/pug/run1//20200101T0000Z/g",
            "state": "waiting",
            "graphDepth": 3
          },
          {
            "id": "~sutherlander/pug/run1//20210101T0000Z/a",
            "state": "waiting",
            "graphDepth": 4
          }
        ]
      }
    ]
  }
}
```

**Performance**
Taken from #5435

```ini
[task parameters]
   m = 0..7000
[scheduling]
   [[queues]]
      [[[default]]]
         limit = 4
   [[graph]]
      R1 = "a => b<m>"
[runtime]
   [[a, b<m>]]
```
```console
$ cylc play <id> --profile --no-detach
```
Run this for a few mins, when you get bored ctrl+c it, then:
```console
$ pip install snakeviz
$ snakeviz ~/cylc-run/<id>/log/scheduler/profile.prof
```
This PR is even faster than current (#5475) at `n=1`:
```
     7001    0.156    0.000    1.738    0.000 /home/sutherlander/cylc/cylc-flow/cylc/flow/data_store_mgr.py:695(increment_graph_window)
```
image

At `n=2` there's a bit more load:
```
     7001    0.185    0.000    4.161    0.001 /home/sutherlander/cylc/cylc-flow/cylc/flow/data_store_mgr.py:695(increment_graph_window)
```
but clearly not squared or cubed.

**Check List**

- [x] Does not contain off-topic changes.
- [x] Tests are included.


dwsutherland:
Coverage should improve with a n=2 window test.

oliver-sanders:
I've been having a play with this branch with the n=1 window in `cylc tui`, looks good.

I've been playing with the following diff in order to track which children/parents are being walked over:

```diff
diff --git a/cylc/flow/data_store_mgr.py b/cylc/flow/data_store_mgr.py
index b6a2960fb..e78cd767c 100644
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -732,6 +732,7 @@ class DataStoreMgr:
             None
 
         """
+        print(f'# increment_graph_window({source_tokens.relative_id})')
 
         # common refrences
         active_id = source_tokens.id
diff --git a/cylc/flow/taskdef.py b/cylc/flow/taskdef.py
index a4ae681e2..d2517383b 100644
--- a/cylc/flow/taskdef.py
+++ b/cylc/flow/taskdef.py
@@ -36,6 +36,7 @@ if TYPE_CHECKING:
 
 def generate_graph_children(tdef, point):
     """Determine graph children of this task at point."""
+    print(f'# generate_graph_children({point}/{tdef.name})')
     graph_children = {}
     for seq, dout in tdef.graph_children.items():
         for output, downs in dout.items():
@@ -78,6 +79,7 @@ def generate_graph_parents(tdef, point, taskdefs):
 
     Infer parents be reversing upstream triggers that lead to point/task.
     """
+    print(f'# generate_graph_parents({point}/{tdef.name})')
     graph_parents = {}
     for seq, triggers in tdef.graph_parents.items():
         if not seq.is_valid(point):
```

Using the graph from your example above with the default n=1 window: <details>
<summary>flow.cylc</summary>

```ini
[scheduler]
    cycle point format = CCYY

[scheduling]
    runahead limit = P1
    initial cycle point = 2020
   [[graph]]
      P1Y = """
         a => b => c => d => e => f => g
         b => h => i => f
         g[-P1Y] => a
      """

[runtime]
    [[root]]
        script = sleep 2
    [[b, c, d, e, f, i]]
    [[a]]
    [[g,h]]
```

</details>

It looks like the algorithm is looking one edge further infront (`c`) than behind (`p`).

E.G. here's a snippet from `--no-detach` output:

```
2023-08-10T09:33:54+01:00 INFO - New flow: 1 (original flow from 2020) 2023-08-10 09:33:54
# generate_graph_children(2020/a)
# increment_graph_window(2020/a)
# generate_graph_children(2020/a)
# generate_graph_children(2020/b)    <= b is n=1 so its children would be n=2
# generate_graph_parents(2020/a)
2023-08-10T09:33:54+01:00 INFO - [2020/a waiting(runahead) job:00 flows:1] => waiting
2023-08-10T09:33:54+01:00 INFO - [2020/a waiting job:00 flows:1] => waiting(queued)
2023-08-10T09:33:55+01:00 INFO - [2020/a waiting(queued) job:00 flows:1] => waiting
2023-08-10T09:33:55+01:00 INFO - [2020/a waiting job:01 flows:1] => preparing
2023-08-10T09:33:56+01:00 INFO - [2020/a preparing job:01 flows:1] submitted to localhost:background[34318]
2023-08-10T09:33:56+01:00 INFO - [2020/a preparing job:01 flows:1] => submitted
2023-08-10T09:33:56+01:00 INFO - [2020/a submitted job:01 flows:1] health: submission timeout=P1D, polling intervals=PT1H,...
2023-08-10T09:33:58+01:00 INFO - [2020/a submitted job:01 flows:1] => running
2023-08-10T09:33:58+01:00 INFO - [2020/a running job:01 flows:1] health: execution timeout=None, polling intervals=PT1H,...
2023-08-10T09:34:00+01:00 INFO - [2020/a running job:01 flows:1] => succeeded
# generate_graph_children(2020/b)
# increment_graph_window(2020/b)
# generate_graph_children(2020/b)   <= duplicate method call from previous walk
# generate_graph_children(2020/h)
# generate_graph_children(2020/c)   <= c is n=1 so its children would be n=2
# generate_graph_parents(2020/b)
```

E.G. in the above output, when the n=0 window was `{1/a}` the algorithm iterated over the graph children of `1/b` which are outside the n-window. These edges aren't needed for the n-window but might be needed for the algorithm?

When the children of `1/b` entered the n=1 window the iteration was repeated which might mean the cache was removed too early?

oliver-sanders:
Moving onto the n=2 window (and dumping out the n-window after each call) with this diff:

```diff
diff --git a/cylc/flow/data_store_mgr.py b/cylc/flow/data_store_mgr.py
index b6a2960fb..3a21bbbea 100644
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -473,7 +473,7 @@ class DataStoreMgr:
         self.updated_state_families = set()
         # Update workflow state totals once more post delta application.
         self.state_update_follow_on = False
-        self.n_edge_distance = 1
+        self.n_edge_distance = 2
         self.next_n_edge_distance = None
         self.latest_state_tasks = {
             state: deque(maxlen=LATEST_STATE_TASKS_QUEUE_SIZE)
@@ -732,6 +732,7 @@ class DataStoreMgr:
             None
 
         """
+        print(f'# increment_graph_window({source_tokens.relative_id})')
 
         # common refrences
         active_id = source_tokens.id
@@ -1018,6 +1019,10 @@ class DataStoreMgr:
                 self.prune_trigger_nodes[active_id])
             del self.prune_trigger_nodes[active_id]
 
+        task_proxies = self.data[self.workflow_id][TASK_PROXIES]
+        for tp_id in sorted(list(task_proxies) + list(self.added[TASK_PROXIES])):
+            print(f'$ {tp_id}')
+
     def generate_edge(
         self,
         parent_tokens: Tokens,
diff --git a/cylc/flow/taskdef.py b/cylc/flow/taskdef.py
index a4ae681e2..d2517383b 100644
--- a/cylc/flow/taskdef.py
+++ b/cylc/flow/taskdef.py
@@ -36,6 +36,7 @@ if TYPE_CHECKING:
 
 def generate_graph_children(tdef, point):
     """Determine graph children of this task at point."""
+    print(f'# generate_graph_children({point}/{tdef.name})')
     graph_children = {}
     for seq, dout in tdef.graph_children.items():
         for output, downs in dout.items():
@@ -78,6 +79,7 @@ def generate_graph_parents(tdef, point, taskdefs):
 
     Infer parents be reversing upstream triggers that lead to point/task.
     """
+    print(f'# generate_graph_parents({point}/{tdef.name})')
     graph_parents = {}
     for seq, triggers in tdef.graph_parents.items():
         if not seq.is_valid(point):
```

Here's the log:

```
2023-08-10T10:17:30+01:00 INFO - New flow: 1 (original flow from 2020) 2023-08-10 10:17:30
# generate_graph_children(2020/a)
# increment_graph_window(2020/a)
# generate_graph_children(2020/a)
# generate_graph_children(2020/b)
# generate_graph_parents(2020/a)
# generate_graph_children(2020/b)   <= duplicate call
# generate_graph_children(2020/h)   <= h is n=2 so its children would be n=3
# generate_graph_children(2020/c)   <= c is n=2 so its children would be n=3
# generate_graph_parents(2020/b)
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/a
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/b
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/c
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/h
2023-08-10T10:17:30+01:00 INFO - [2020/a waiting(runahead) job:00 flows:1] => waiting
2023-08-10T10:17:30+01:00 INFO - [2020/a waiting job:00 flows:1] => waiting(queued)
2023-08-10T10:17:30+01:00 INFO - [2020/a waiting(queued) job:00 flows:1] => waiting
2023-08-10T10:17:30+01:00 INFO - [2020/a waiting job:01 flows:1] => preparing
2023-08-10T10:17:32+01:00 INFO - [2020/a preparing job:01 flows:1] submitted to localhost:background[71918]
2023-08-10T10:17:32+01:00 INFO - [2020/a preparing job:01 flows:1] => submitted
2023-08-10T10:17:32+01:00 INFO - [2020/a submitted job:01 flows:1] health: submission timeout=P1D, polling intervals=PT1H,...
2023-08-10T10:17:34+01:00 INFO - [2020/a submitted job:01 flows:1] => running
2023-08-10T10:17:34+01:00 INFO - [2020/a running job:01 flows:1] health: execution timeout=None, polling intervals=PT1H,...
2023-08-10T10:17:36+01:00 INFO - [2020/a running job:01 flows:1] => succeeded
# generate_graph_children(2020/b)
# increment_graph_window(2020/b)
# generate_graph_children(2020/h)   <= duplicate call
# generate_graph_children(2020/i)   <= i is n=2 so its children would be n=3
# generate_graph_parents(2020/h)
# generate_graph_children(2020/c)   <= duplicate call
# generate_graph_children(2020/d)   <= d is n=2 so its children would be n=3
# generate_graph_parents(2020/c)   <= duplicate call
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/a
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/b
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/c
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/d
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/h
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/i
2023-08-10T10:17:36+01:00 INFO - [2020/b waiting(runahead) job:00 flows:1] => waiting
2023-08-10T10:17:36+01:00 INFO - [2020/b waiting job:00 flows:1] => waiting(queued)
2023-08-10T10:17:37+01:00 INFO - [2020/b waiting(queued) job:00 flows:1] => waiting
2023-08-10T10:17:37+01:00 INFO - [2020/b waiting job:01 flows:1] => preparing
2023-08-10T10:17:38+01:00 INFO - [2020/b preparing job:01 flows:1] submitted to localhost:background[72234]
2023-08-10T10:17:38+01:00 INFO - [2020/b preparing job:01 flows:1] => submitted
2023-08-10T10:17:38+01:00 INFO - [2020/b submitted job:01 flows:1] health: submission timeout=P1D, polling intervals=PT1H,...
2023-08-10T10:17:40+01:00 INFO - [2020/b submitted job:01 flows:1] => running
2023-08-10T10:17:40+01:00 INFO - [2020/b running job:01 flows:1] health: execution timeout=None, polling intervals=PT1H,...
2023-08-10T10:17:42+01:00 INFO - [2020/b running job:01 flows:1] => succeeded
# generate_graph_children(2020/h)
# increment_graph_window(2020/h)
# generate_graph_children(2020/d)   <= duplicate call
# generate_graph_children(2020/e)   <= d is n=2 so its children would be n=3
# generate_graph_parents(2020/d)
# generate_graph_children(2020/i)   <= duplicate call
# generate_graph_children(2020/f)   <= d is n=2 so its children would be n=3
# generate_graph_parents(2020/i)
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/a
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/b
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/c
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/d
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/e
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/f
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/h
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/i
2023-08-10T10:17:42+01:00 INFO - [2020/h waiting(runahead) job:00 flows:1] => waiting
2023-08-10T10:17:42+01:00 INFO - [2020/h waiting job:00 flows:1] => waiting(queued)
# generate_graph_children(2020/c)
# increment_graph_window(2020/c)   <= fully cached graph walk, nice!
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/a
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/b
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/c
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/d
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/e
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/f
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/h
$ ~osanders/tmp.h3r2rW6eFX/run12//2020/i
```

Looks good, gives the right result, some possible savings to be made avoiding duplicate `generate_graph_children` calls.

oliver-sanders:
There are duplicate `generate_ghost_task` calls, but this isn't a problem because the method returns early if the task already exists in the store.

I spotted a [very] small optimisation moving that return a little higher up in the code:

```diff
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -1108,11 +1108,15 @@ class DataStoreMgr:
         Orphan tasks with no children return (True, None) respectively.
 
         """
+        tp_id = tokens.id
+        task_proxies = self.data[self.workflow_id][TASK_PROXIES]
+        if tp_id in task_proxies or tp_id in self.added[TASK_PROXIES]:
+            # ghost task already created
+            return
+
         name = tokens['task']
         point_string = tokens['cycle']
         t_id = self.definition_id(name)
-        tp_id = tokens.id
-        task_proxies = self.data[self.workflow_id][TASK_PROXIES]
 
         is_orphan = False
         if name not in self.schd.config.taskdefs:
@@ -1120,8 +1124,6 @@ class DataStoreMgr:
 
         if itask is None:
             itask = self.schd.pool.get_task(point_string, name)
-        if tp_id in task_proxies or tp_id in self.added[TASK_PROXIES]:
-            return
 
         if itask is None:
             itask = TaskProxy(
```

oliver-sanders:
Tried out the new graphDepth field, the numbers came out as expected so this should close #5609 which will allow us to unlock the higher n-window views in the GUI!

dwsutherland:
> It looks like the algorithm is looking one edge further infront (`c`) than behind (`p`).

Yes, I rely on all but the first node in a branch should have `c` and `p` already generated  (because the first walk generates the walk stub for all walked nodes in it's graph)

> There are duplicate `generate_ghost_task` calls, but this isn't a problem because the method returns early if the task already exists in the store. 

I've been relying on `generate_ghost_task` to return early (it has the same block at the top), but to be honest, the fact a walk stub exists would be enough to not call that initial one too (if it's that expensive).


> Is the graph window that the data store is currently configured to track available in the schema? That would be handy for the UI.

I think we can add that as a workflow field.
Also I think instead of changing the window size for a request (as that would change all views), we need to have those min/max depth args, so we can just filter the request/sub by depth.

I've also thought of an optimization for the depth finder, so will put that up in the morning.

dwsutherland:
I changed it to reuse the `TaskProxy` call to `generate_graph_children`:
```
2023-08-11T17:00:37+12:00 INFO - New flow: 1 (original flow from 2020) 2023-08-11 17:00:37
# Children of 2020/a - Called by __init__
# Active ID: ~sutherlander/zig/run1//2020/a
# Generate Children/Parents of 2020/a
# Children of 2020/b - Called by __init__
# parents of 2020/a - Called by increment_graph_window
2023-08-11T17:00:37+12:00 INFO - [2020/a waiting(runahead) job:00 flows:1] => waiting
2023-08-11T17:00:37+12:00 INFO - [2020/a waiting job:00 flows:1] => waiting(queued)
2023-08-11T17:00:37+12:00 INFO - [2020/a waiting(queued) job:00 flows:1] => waiting
2023-08-11T17:00:37+12:00 INFO - [2020/a waiting job:01 flows:1] => preparing
2023-08-11T17:00:38+12:00 INFO - [2020/a preparing job:01 flows:1] submitted to localhost:background[16606]
2023-08-11T17:00:38+12:00 INFO - [2020/a preparing job:01 flows:1] => submitted
2023-08-11T17:00:38+12:00 INFO - [2020/a submitted job:01 flows:1] health: submission timeout=None, polling intervals=PT15M,...
2023-08-11T17:00:39+12:00 INFO - [2020/a submitted job:01 flows:1] => running
2023-08-11T17:00:39+12:00 INFO - [2020/a running job:01 flows:1] health: execution timeout=None, polling intervals=PT15M,...
2023-08-11T17:00:41+12:00 INFO - [2020/a running job:01 flows:1] => succeeded
# Children of 2020/b - Called by __init__
# Active ID: ~sutherlander/zig/run1//2020/b
# Generate Children/Parents of 2020/b
# Children of 2020/c - Called by __init__
# Children of 2020/h - Called by __init__
# parents of 2020/b - Called by increment_graph_window
2023-08-11T17:00:41+12:00 INFO - [2020/b waiting(runahead) job:00 flows:1] => waiting
2023-08-11T17:00:41+12:00 INFO - [2020/b waiting job:00 flows:1] => waiting(queued)
2023-08-11T17:00:42+12:00 INFO - [2020/b waiting(queued) job:00 flows:1] => waiting
2023-08-11T17:00:42+12:00 INFO - [2020/b waiting job:01 flows:1] => preparing
2023-08-11T17:00:43+12:00 INFO - [2020/b preparing job:01 flows:1] submitted to localhost:background[16631]
2023-08-11T17:00:43+12:00 INFO - [2020/b preparing job:01 flows:1] => submitted
2023-08-11T17:00:43+12:00 INFO - [2020/b submitted job:01 flows:1] health: submission timeout=None, polling intervals=PT15M,...
2023-08-11T17:00:44+12:00 INFO - [2020/b submitted job:01 flows:1] => running
2023-08-11T17:00:44+12:00 INFO - [2020/b running job:01 flows:1] health: execution timeout=None, polling intervals=PT15M,...
2023-08-11T17:00:46+12:00 INFO - [2020/b running job:01 flows:1] => succeeded
# Children of 2020/c - Called by __init__
# Active ID: ~sutherlander/zig/run1//2020/c
# Generate Children/Parents of 2020/c
# Children of 2020/d - Called by __init__
# parents of 2020/c - Called by increment_graph_window
2023-08-11T17:00:46+12:00 INFO - [2020/c waiting(runahead) job:00 flows:1] => waiting
2023-08-11T17:00:46+12:00 INFO - [2020/c waiting job:00 flows:1] => waiting(queued)
# Children of 2020/h - Called by __init__
# Active ID: ~sutherlander/zig/run1//2020/h
# Generate Children/Parents of 2020/h
# Children of 2020/i - Called by __init__
# parents of 2020/h - Called by increment_graph_window
2023-08-11T17:00:46+12:00 INFO - [2020/h waiting(runahead) job:00 flows:1] => waiting
```

dwsutherland:
There were still some duplicate `generate_graph_children` calls (as shown above) from the creation of `TaskProxy` objects at n > 0  (which are used for data store info).. I've excluding generating children for these (re-using the `data_mode` option)

dwsutherland:
API fields and args now added:
image

oliver-sanders:
Looking good, neat optimisation grabbing the graph_children from the already generated TaskProxy in the n=1 window.

Tested with the workflow from earlier:

<details>
<summary>flow.cylc</summary>

```ini[scheduler]
    cycle point format = CCYY

[scheduling]
    runahead limit = P1
    initial cycle point = 2020
   [[graph]]
      P1Y = """
         a => b => c => d => e => f => g
         b => h => i => f
         g[-P1Y] => a
      """

[runtime]
    [[root]]
        script = sleep 2
    [[b, c, d, e, f, i]]
    [[a]]
    [[g,h]]


```

</details>

Using this patch to see what `increment_graph_window` is doing:

<details>
<summary>patch</summary>

```diff
diff --git a/cylc/flow/data_store_mgr.py b/cylc/flow/data_store_mgr.py
index 68ede96ae..9a5709982 100644
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -473,7 +473,7 @@ class DataStoreMgr:
         self.updated_state_families = set()
         # Update workflow state totals once more post delta application.
         self.state_update_follow_on = False
-        self.n_edge_distance = 1
+        self.n_edge_distance = 2
         self.next_n_edge_distance = None
         self.latest_state_tasks = {
             state: deque(maxlen=LATEST_STATE_TASKS_QUEUE_SIZE)
@@ -733,6 +733,7 @@ class DataStoreMgr:
             None
 
         """
+        print(f'# increment_graph_window({source_tokens.relative_id})')
 
         # common refrences
         active_id = source_tokens.id
@@ -904,6 +905,7 @@ class DataStoreMgr:
                     # TODO: xtrigger is workflow_state edges too
                     # Reference set for workflow relations
                     if itask is not None and n_depth == 1:
+                        # print(f'  * get-cashed-graph-children({itask.tokens.relative_id})')
                         graph_children = itask.graph_children
                     else:
                         graph_children = generate_graph_children(
@@ -1038,6 +1040,11 @@ class DataStoreMgr:
                 self.prune_trigger_nodes[active_id])
             del self.prune_trigger_nodes[active_id]
 
+        print('$ Active Window:')
+        task_proxies = self.data[self.workflow_id][TASK_PROXIES]
+        for tp_id in sorted(list(task_proxies) + list(self.added[TASK_PROXIES])):
+            print(f'  * {Tokens(tp_id).relative_id}')
+
     def generate_edge(
         self,
         parent_tokens: Tokens,
@@ -1128,6 +1135,7 @@ class DataStoreMgr:
         Orphan tasks with no children return (True, None) respectively.
 
         """
+        print(f'  * generate_ghost_task({tokens.relative_id})')
         tp_id = tokens.id
         if (
             tp_id in self.data[self.workflow_id][TASK_PROXIES]
diff --git a/cylc/flow/task_pool.py b/cylc/flow/task_pool.py
index 7b21d7203..bf1261e40 100644
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1482,6 +1482,7 @@ class TaskPool:
 
         Force arg used in manual triggering.
         """
+        print(f'# spawn_task({point}/{name})')
         if not self.can_spawn(name, point):
             return None
 
diff --git a/cylc/flow/task_proxy.py b/cylc/flow/task_proxy.py
index 5e550c04a..b6c30a042 100644
--- a/cylc/flow/task_proxy.py
+++ b/cylc/flow/task_proxy.py
@@ -196,7 +196,6 @@ class TaskProxy:
         flow_wait: bool = False,
         data_mode: bool = False,
     ) -> None:
-
         self.tdef = tdef
         if submit_num is None:
             submit_num = 0
@@ -213,6 +212,7 @@ class TaskProxy:
             cycle=str(self.point),
             task=self.tdef.name,
         )
+        # print(f'  * TaskProxy.__init__({self.tokens.relative_id}, data_mode={data_mode})')
         self.identity = self.tokens.relative_id
         self.reload_successor: Optional['TaskProxy'] = None
         self.point_as_seconds: Optional[int] = None
diff --git a/cylc/flow/taskdef.py b/cylc/flow/taskdef.py
index a4ae681e2..f191f15f7 100644
--- a/cylc/flow/taskdef.py
+++ b/cylc/flow/taskdef.py
@@ -36,6 +36,7 @@ if TYPE_CHECKING:
 
 def generate_graph_children(tdef, point):
     """Determine graph children of this task at point."""
+    print(f'  * generate_graph_children({point}/{tdef.name})')
     graph_children = {}
     for seq, dout in tdef.graph_children.items():
         for output, downs in dout.items():
@@ -78,6 +79,7 @@ def generate_graph_parents(tdef, point, taskdefs):
 
     Infer parents be reversing upstream triggers that lead to point/task.
     """
+    print(f'  * generate_graph_parents({point}/{tdef.name})')
     graph_parents = {}
     for seq, triggers in tdef.graph_parents.items():
         if not seq.is_valid(point):
```

</details>

The n=1 window looks great, no duplicate graph_(children|parents) calls, only looks ahead as far as it needs to.

I think I may have spotted something in the n=2 window (I hacked this as the default in the above patch).

When task `b` succeeds, and task `h` is spawned, I would expect the n=2 window to be:

* a  n=2
* b  n=1
* c   n=2
* h  n=0
* i  n=1
* f  n=2




However, the routine fetches the graph children of `1/c` which is n=2, then generates a ghost task for `1/d` which is n=3:

```
2023-08-22T16:58:59+01:00 INFO - [2020/b running job:01 flows:1] => succeeded
# spawn_task(2020/h)
  * generate_graph_children(2020/h)
# increment_graph_window(2020/h)
  * generate_ghost_task(2020/h)
  * generate_graph_children(2020/i)
  * generate_ghost_task(2020/f)
  * generate_graph_parents(2020/i)
  * generate_ghost_task(2020/h)
  * generate_graph_children(2020/d)   # n=3 !!!
  * generate_ghost_task(2020/e)    # n=3 !!!
  * generate_graph_parents(2020/d)  # n=3 !!!
  * generate_ghost_task(2020/c)
* Active Window:
  * 2020/a  # n=2
  * 2020/b  # n=1
  * 2020/c  # n=2
  * 2020/d  # n=3 !!!
  * 2020/e  # n=3 !!!
  * 2020/f  # n=2
  * 2020/h # n=0
  * 2020/i  # n=1
2023-08-22T16:58:59+01:00 INFO - [2020/h waiting(runahead) job:00 flows:1] => waiting
2023-08-22T16:58:59+01:00 INFO - [2020/h waiting job:00 flows:1] => waiting(queued)
```

Miliseconds later it also spawns `c` (`c` and `h` are siblings with the same parent), at which point `d` and `e` enter the n=2 window:

```
2023-08-22T16:58:59+01:00 INFO - [2020/h waiting(runahead) job:00 flows:1] => waiting
2023-08-22T16:58:59+01:00 INFO - [2020/h waiting job:00 flows:1] => waiting(queued)
# spawn_task(2020/c)
  * generate_graph_children(2020/c)
# increment_graph_window(2020/c)
  * generate_ghost_task(2020/c)
* Active Window:
  * 2020/a
  * 2020/b
  * 2020/c
  * 2020/d  # now n=1
  * 2020/e  # now n=2
  * 2020/f
  * 2020/h
  * 2020/i
2023-08-22T16:58:59+01:00 INFO - [2020/c waiting(runahead) job:00 flows:1] => waiting
```

I'm not quite sure what's going on here, is `increment_graph_window` being clever and realising that `c` is about to be spawned so `d` and `e` are about to drift into the window?

oliver-sanders:
I think I'm also getting an `n=3` task when `e` gets spawned:

Screenshot from 2023-08-22 17-29-37

I think 2023/a should be n=3?

```
2023-08-22T17:28:26+01:00 INFO - [2020/i running job:01 flows:1] => succeeded
# spawn_task(2020/f)
  * generate_graph_children(2020/f)
2023-08-22T17:28:26+01:00 INFO - [2020/f waiting(runahead) job:00 flows:1] => waiting
2023-08-22T17:28:26+01:00 INFO - [2020/d running job:01 flows:1] => succeeded
# spawn_task(2020/e)
  * generate_graph_children(2020/e)
# increment_graph_window(2020/e)
  * generate_ghost_task(2020/e)
  * generate_graph_children(2020/g)
  * generate_ghost_task(2021/a)
  * generate_graph_parents(2020/g)
  * generate_ghost_task(2020/f)
$ Active Window:
  * 2020/b
  * 2020/c
  * 2020/d
  * 2020/e
  * 2020/f
  * 2020/g
  * 2020/h
  * 2020/i
  * 2021/a  # n=3 !!!
```

oliver-sanders:
I managed to get the UI to colour code tasks according to their n-depth:

Screenshot from 2023-08-24 14-56-04

Ugly but functional, note, only works if you have the Tree view open.

<details>
<summary>patch</summary>

```diff
diff --git a/src/components/cylc/SVGTask.vue b/src/components/cylc/SVGTask.vue
index c242a32c..dc2d505d 100644
--- a/src/components/cylc/SVGTask.vue
+++ b/src/components/cylc/SVGTask.vue
@@ -37,6 +37,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
       queued: task.isQueued && !task.isHeld,
       runahead: task.isRunahead && !(task.isHeld || task.isQueued),
     }"
+    :style="`
+      --foreground: ${foreground};
+      --background: ${background};
+    `"
   >
     <!-- status
 
@@ -279,6 +283,22 @@ export default {
       // otherwise the progress indicator may end up in the wrong place.
       type: Number,
       default: 0
+    },
+    colourMap: {
+      default: {
+        undefined: 'grey',
+        0: 'blue',
+        1: 'green',
+        2: 'yellow'
+      }
+    }
+  },
+  computed: {
+    foreground () {
+      return this.colourMap[this.task.graphDepth]
+    },
+    background () {
+      return 'white'
     }
   },
   methods: {
@@ -342,19 +362,21 @@ export default {
 </script>
 
 <style lang="scss">
-  $foreground: rgb(90,90,90);
-  $background: rgb(255,255,255);
+  :root {
+    --foreground: rgb(90,90,90);
+    --background: rgb(255,255,255);
+  }
 
   .c8-task {
     .status {
       .outline {
         // NOTE: ensure the outline is filled so that it can be clicked
-        fill: $background;
-        stroke: $foreground;
+        fill: var(--background);
+        stroke: var(--foreground);
       }
       .progress {
         fill: transparent;
-        stroke: $foreground;
+        stroke: var(--foreground);
         transform-origin: 50% 50%;
         opacity: 0.4;
         /* 0% progress */
@@ -397,75 +419,75 @@ export default {
     }
 
     &.preparing .status .dot {
-      fill: $foreground;
+      fill: var(--foreground);
     }
 
     &.submitted .status .dot {
-      fill: $foreground;
+      fill: var(--foreground);
     }
 
     &.running .status .hub {
-      fill: $foreground;
+      fill: var(--foreground);
     }
     &.running .status .progress {
-      fill: $foreground;
+      fill: var(--foreground);
     }
 
     &.succeeded .status .outline {
-      fill: $foreground;
+      fill: var(--foreground);
     }
 
     &.failed .status {
       .outline {
-        fill: $foreground;
+        fill: var(--foreground);
       }
       .cross rect {
-        fill: $background;
+        fill: var(--background);
       }
     }
 
     &.submit-failed .status {
       .outline {
-        fill: $background;
+        fill: var(--background);
       }
       .cross rect {
-        fill: $foreground;
+        fill: var(--foreground);
       }
     }
 
     &.expired .status {
       .outline {
-        fill: $foreground;
+        fill: var(--foreground);
       }
       .dot {
-        fill: $background;
+        fill: var(--background);
       }
       .expired rect {
-        fill: $background;
+        fill: var(--background);
       }
     }
 
     &.held .modifier {
       .outline {
-        stroke: $foreground;
+        stroke: var(--foreground);
       }
       .held rect {
-        fill: $foreground;
+        fill: var(--foreground);
       }
     }
 
     &.queued .modifier {
       .queued rect {
-        fill: $foreground;
+        fill: var(--foreground);
       }
     }
 
     &.runahead .modifier {
       .outline {
-        stroke: $foreground;
+        stroke: var(--foreground);
       }
       .runahead circle {
-        fill: $foreground;
+        fill: var(--foreground);
       }
     }
   }
diff --git a/src/graphql/queries.js b/src/graphql/queries.js
index 86db792e..2bb5e0b3 100644
--- a/src/graphql/queries.js
+++ b/src/graphql/queries.js
@@ -75,6 +75,7 @@ const TASK_PROXY_DATA_FRAGMENT = `
 fragment TaskProxyData on TaskProxy {
   id
   state
+  graphDepth
   isHeld
   isQueued
   isRunahead
```

</details>

oliver-sanders:
There's something funny going on with the graph depth in some cases, here the tasks d, e, f and i all have a graph depth of 1, but only a & f should be n=1.

n=1 - blue
n=2 - green
n=3 - yellow

Screenshot from 2023-08-24 15-00-07

<details>
<summary>graphql response</summary>

```graphql
query {
  workflows {
    taskProxies {
      id
      graphDepth
    }
  }
}
```

```json
{
  "data": {
    "workflows": [
      {
        "taskProxies": [
          {
            "id": "~osanders/david//2020/h",
            "graphDepth": 2
          },
          {
            "id": "~osanders/david//2020/c",
            "graphDepth": 2
          },
          {
            "id": "~osanders/david//2020/i",
            "graphDepth": 1
          },
          {
            "id": "~osanders/david//2020/d",
            "graphDepth": 1
          },
          {
            "id": "~osanders/david//2020/f",
            "graphDepth": 1
          },
          {
            "id": "~osanders/david//2020/e",
            "graphDepth": 1
          },
          {
            "id": "~osanders/david//2020/g",
            "graphDepth": 0
          },
          {
            "id": "~osanders/david//2021/a",
            "graphDepth": 1
          },
          {
            "id": "~osanders/david//2021/b",
            "graphDepth": 2
          }
        ]
      }
    ]
  }
}
```

</details>

oliver-sanders:
We've deduced that one of the reported GUI bugs is actually a data store housekeeping bug which appears to be fixed on this branch. So this PR should close #5707 :tada:. This bug does increase the urgency somewhat so we may want to look at bumping this forward to 8.2.x when we're closer to merge.

dwsutherland:
> It looks like d might be lingering from when the n=0 window was {b}? I'm not quite sure what's going on here, is increment_graph_window being clever and realising that c is about to be spawned so d and e are about to drift into the window? If so how?!

I think this might be (without looking) because (to avoid blinking), the pruning is batched to prevent blinking (being pruned and then having to recreate (and load from DB)) when then next task spawns (and it knows they aren't end of branch)..

> There's something funny going on with the graph depth in some cases, here the tasks d, e, f and i all have a graph depth of 1, but only a & f should be n=1.
.
.
.
Really not sure what's going on here, possibly restoring nodes from old graph walks?

I'll have a look, must be missing something in the depth finder. . .

dwsutherland:
Ok, I've pushed a fix and changed the base to `8.2.x`.

There was actually a major issue with the node associations I was making in the walk (fixed).

There was also an issue where pruned nodes were still in the walks of some nodes, so when nodes would disappear then reappear, i.e.:
image
image

(as only `e` is active, with `f` having partially met prerequisites (it is `n=1`))

but when they reappeared (in the UI) they would have no edges or state (...etc).. This is because the depth was being sent to the UI.. 
This has been fixed, by making sure they get fully regenerated/rewalked when the associated node becomes active.

dwsutherland:
(there's a bug at `n=1` I'm trying to solve)

oliver-sanders:
These tasks which leave, then re-enter the n-window seem to be causing issues.

So, for testing, I wrote a workflow in which the task "blink" should disappear three times, re-appearing and running at the end.

Set the `window` Jinja2 variable to whatever graph window you want to test and it will jam in enough tasks to ensure that `blink` drifts in and out of the window.

(bin/set-workflow-extent)

```bash
#!/bin/bash                                 

# sets the graph window, taken from the example in the OP
                                            
set -u                                                       
                                            
read -r -d '' gqlDoc <<_DOC_                
{"request_string": "                        
mutation {                                  
  setGraphWindowExtent (                    
    workflows: [\"${CYLC_WORKFLOW_ID/\//\\/}\"],                                                                                   
    nEdgeDistance: $1                       
  ) {                                       
    result                                  
  }                                         
}",                                         
"variables": null}                                        
_DOC_                                       
                                            
echo "${gqlDoc}"                            
                                            
cylc client "$CYLC_WORKFLOW_ID" graphql < <(echo ${gqlDoc}) 2>/dev/null
                                            
echo DONE
```

(flow.cylc)

```ini
#!Jinja2                                    
                                            
{% set letters = 'abcdefghijklmnopqrstuvwxyz' %}
{% set window = window | default(1) %}      
{% set repititions = repititions | default(3) %}
{% set state = namespace(index=0) %}        
                                            
[scheduler]                                 
    allow implicit tasks = True             
    [[events]]
        startup handlers = set-window-extent {{ window }}
                                                                                                                                   
[scheduling]                                
    [[graph]]                               
        R1 = """                            
{% for rep in range(repititions) %}                       
    {{ letters[state.index] }} => blink     
    {% for ind in range((2 * window)) %}    
        {{ letters[state.index] }} => {{ letters[state.index + 1] }}
        {% set state.index = state.index + 1 %}
    {% endfor %}
{% endfor %}                                
    {{ letters[state.index] }} => blink     
        """
 
[runtime]
    [[root]]
        script = sleep 5
```

oliver-sanders:
Now to try out [that workflow](https://github.com/cylc/cylc-flow/pull/5660#issuecomment-1713967538) with the colour-coded graph view thinggy:

<details>
<summary>cylc-ui patch</summary>

```pre
diff --git a/src/components/cylc/SVGTask.vue b/src/components/cylc/SVGTask.vue
index c242a32c..b46e4581 100644
--- a/src/components/cylc/SVGTask.vue
+++ b/src/components/cylc/SVGTask.vue
@@ -37,6 +37,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
       queued: task.isQueued && !task.isHeld,
       runahead: task.isRunahead && !(task.isHeld || task.isQueued),
     }"
+    :style="`
+      --foreground: ${foreground};
+      --background: ${background};
+    `"
   >
     <!-- status
 
@@ -279,6 +283,25 @@ export default {
       // otherwise the progress indicator may end up in the wrong place.
       type: Number,
       default: 0
+    },
+    colourMap: {
+      default: {
+        undefined: 'grey',
+        0: 'blue',
+        1: 'green',
+        2: 'yellow',
+        3: 'orange',
+        4: 'red',
+        5: 'purple'
+      }
+    }
+  },
+  computed: {
+    foreground () {
+      return this.colourMap[this.task.graphDepth]
+    },
+    background () {
+      return 'white'
     }
   },
   methods: {
@@ -342,19 +365,21 @@ export default {
 </script>
 
 <style lang="scss">
-  $foreground: rgb(90,90,90);
-  $background: rgb(255,255,255);
+  :root {
+    --foreground: rgb(90,90,90);
+    --background: rgb(255,255,255);
+  }
 
   .c8-task {
     .status {
       .outline {
         // NOTE: ensure the outline is filled so that it can be clicked
-        fill: $background;
-        stroke: $foreground;
+        fill: var(--background);
+        stroke: var(--foreground);
       }
       .progress {
         fill: transparent;
-        stroke: $foreground;
+        stroke: var(--foreground);
         transform-origin: 50% 50%;
         opacity: 0.4;
         /* 0% progress */
@@ -397,75 +422,75 @@ export default {
     }
 
     &.preparing .status .dot {
-      fill: $foreground;
+      fill: var(--foreground);
     }
 
     &.submitted .status .dot {
-      fill: $foreground;
+      fill: var(--foreground);
     }
 
     &.running .status .hub {
-      fill: $foreground;
+      fill: var(--foreground);
     }
     &.running .status .progress {
-      fill: $foreground;
+      fill: var(--foreground);
     }
 
     &.succeeded .status .outline {
-      fill: $foreground;
+      fill: var(--foreground);
     }
 
     &.failed .status {
       .outline {
-        fill: $foreground;
+        fill: var(--foreground);
       }
       .cross rect {
-        fill: $background;
+        fill: var(--background);
       }
     }
 
     &.submit-failed .status {
       .outline {
-        fill: $background;
+        fill: var(--background);
       }
       .cross rect {
-        fill: $foreground;
+        fill: var(--foreground);
       }
     }
 
     &.expired .status {
       .outline {
-        fill: $foreground;
+        fill: var(--foreground);
       }
       .dot {
-        fill: $background;
+        fill: var(--background);
       }
       .expired rect {
-        fill: $background;
+        fill: var(--background);
       }
     }
 
     &.held .modifier {
       .outline {
-        stroke: $foreground;
+        stroke: var(--foreground);
       }
       .held rect {
-        fill: $foreground;
+        fill: var(--foreground);
       }
     }
 
     &.queued .modifier {
       .queued rect {
-        fill: $foreground;
+        fill: var(--foreground);
       }
     }
 
     &.runahead .modifier {
       .outline {
-        stroke: $foreground;
+        stroke: var(--foreground);
       }
       .runahead circle {
-        fill: $foreground;
+        fill: var(--foreground);
       }
     }
   }
diff --git a/src/graphql/queries.js b/src/graphql/queries.js
index 86db792e..2bb5e0b3 100644
--- a/src/graphql/queries.js
+++ b/src/graphql/queries.js
@@ -75,6 +75,7 @@ const TASK_PROXY_DATA_FRAGMENT = `
 fragment TaskProxyData on TaskProxy {
   id
   state
+  graphDepth
   isHeld
   isQueued
   isRunahead
```

</details>

The results are looking more stable than before, but the algorithm has problems around the `blink` task when it re-enters the n-window...

### n=1

(blink should be n=2)

Screenshot from 2023-09-11 15-10-21

### n=2

(`i` and `m` should be n=3)

Screenshot from 2023-09-11 13-48-46

### n=3

(`f`, `g`, `a` and `b` should be n>3)

Screenshot from 2023-09-11 15-34-03

In the n>1 window, we start to see some random graph isolates too. I though this might be a graph view bug, but the tasks also appear in the tree view?

### n=2

Screenshot from 2023-09-11 15-17-58

### n=3

Screenshot from 2023-09-11 15-34-16

oliver-sanders:
The graph is very satisfying when the workflow completes:

Screenshot from 2023-09-11 15-42-49

dwsutherland:
> In the n>1 window, we start to see some random graph isolates too. I though this might be a graph view bug, but the tasks also appear in the tree view?

Refresh the browser and they disappear, so probably not the data-store.

dwsutherland:
It looks like all those nodes with the wrong depth shouldn't even been in the graph... Will try to find out why.

dwsutherland:
Ok, I've fixed the problems (I believe).

To keep things simple, I only work out depths for the depths available..
So anything outside the window (that would have a greater depth) has now been pruned properly.

(that was a fun workflow to play with 👍 )

oliver-sanders:
Re-ran the [last set of tests](https://github.com/cylc/cylc-flow/pull/5660#issuecomment-1714030929).

It works flawlessly for the n=1 and n=2 windows, the graph isolates have disappeared and the window is exactly as it should be.

In the n=3 window, I'm getting some odd behaviour with a few n=4 tasks being shown as n=3.

I've turned the n=3 example into an integration test which checks that the window that the data store generates is correct. The test passes so `increment_graph_window` is definitely doing the right thing, I think the issues I'm seeing in the GUI are probably caused by a missing "pruned" delta, or possibly a GUI bug, I haven't confirmed which one yet.

Here's the integration test: https://github.com/dwsutherland/cylc-flow/pull/16.

I couldn't work out how to test the deltas in this integration test, "added" and "updated" seem to be empty, I probably need to call something to populate them?

Here are screenshots from the n=3 example using the GUI patch above. The issues are only found around the "blink" task:

Screenshot from 2023-09-15 10-07-33
Screenshot from 2023-09-15 10-01-20
Screenshot from 2023-09-15 10-04-20

dwsutherland:
>  I think the issues I'm seeing in the GUI are probably caused by a missing "pruned" delta, or possibly a GUI bug, I haven't confirmed which one yet.

The fact the data-store is ok, means the pruned deltas are being applied there.. should be able to query the data-store to confirm it's ok at the UIS.

dwsutherland:
I don't know why (and maybe it's nothing to do with the new test), but mypy tests are now failing:
```
/opt/hostedtoolcache/Python/3.10.[13](https://github.com/cylc/cylc-flow/actions/runs/6203402090/job/16871285396?pr=5660#step:9:14)/x64/lib/python3.10/site-packages/matplotlib/pyplot.py:2700: error: Positional-only parameters are only supported in Python 3.8 and greater  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```
kinda weird, as the python versions are 3.10 3.11 that are failing on this.. (maybe I don't understand it)

dwsutherland:
> In the n=3 window, I'm getting some odd behaviour with a few n=4 tasks being shown as n=3.

This should be fixed "once and for all", I've included all walk ids as candidates for pruning (when triggered)..
Most will be excluded (when in the active window also), but it should catch all branches..

oliver-sanders:
Great, I'll have a play today.

Any idea how to check the deltas in that integration test? Would be good to ensure we don't mess up this code later by accident.

oliver-sanders:
Manual testing going well.

Did a bit more integration testing to nail down those deltas: https://github.com/dwsutherland/cylc-flow/pull/17

tests/functional/n-window/02-big-window.t
```diff
@@ -0,0 +1,57 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+
+# Test large window size using graphql and find tasks in window.
+# This is helpful with coverage by using most the no-rewalk mechanics.
```
oliver-sanders, line 20: These show tests are integration testable which removes the potential for race conditions and gives us more flexibility to push tasks into whatever state we choose.

```python
# untested
from cylc.flow.scripts import show
from cylc.flow.option_parsers import Options
from cylc.flow.cyclers.integer import IntegerPoint
from cylc.flow.task_status import TASK_STATUS_SUCCEEDED

ShowOptions = Options(get_option_parser())

def show_task(schd, task):
        return await show(
            schd.tokens.id,
            schd.tokens.duplicate(cycle='1', task=task),
            ShowOptions(json=True)
        )

async def test_whatevs(...):
   ...
    async with start(schd):
        itask = schd.pool.get_task(IntegerPoint(1), 'task-name')
        await schd.update_data_structure()

        assert await show_task(schd, 'task-name') == {...}

        itask.reset_state(TASK_STATUS_SUCCEEDED)
        schd.data_store_mgr.delta_task_status(itask)
        assert await show_task(schd, 'task-name') == {...}

File: tests/integration/test_increment_graph_window.py [hunk 2]
```diff
@@ -0,0 +1,308 @@
...
+                    key: value
+                    for key, value in _n_window.items()
+                    if key not in previous_n_window
+                }
+                # TODO: updated deltas don't seem to be coming through as expected
+                # assert {
+                #     key: value
+                #     for key, value in _n_window.items()
+                #     if previous_n_window.get(key, value) != value
+                # } == updated
```
oliver-sanders, line 256: Any idea why the updated deltas are off?

dwsutherland: Will have a look

oliver-sanders:
@hjoliver, there's one pending question about testing the updated deltas, but this is ready for review and quite high priority as we're getting a lot of people hitting this one.

It tests correctly for the examples I've put through it and the GUI is showing the right things.

dwsutherland:
>  there's one pending question about testing the updated deltas

I think this is `update_data_structure` mopping up deltas.. but will look again

> I still need to check the caches are properly housekept to avoid memory leak issues, but haven't figured out a way to test this yet...

Think we did have a plugin for this, or we've done some testing before..

oliver-sanders:
> I still need to check the caches are properly housekept to avoid memory leak issues, but haven't figured out a way to test this yet...
>
>> Think we did have a plugin for this, or we've done some testing before..

The "log memory" main loop plugin is good at tracking the top-level memory usage. With a bit of fiddling with the source code we can get it to track specific targets in the code.

Just counting the number of objects in the cache would probably be enough though. I guess in a cycling workflow, I should expect to see the number of walks in the cache increase, then stabilise, possibly with some periodic fluctuation?

oliver-sanders:
I checked the caches today and confirmed they were being cleared as expected.

Added an integration test to cover this: https://github.com/dwsutherland/cylc-flow/pull/18

tests/integration/test_increment_graph_window.py
oliver-sanders: Any ideas? It looks on the surface like the updated deltas are completely wrong, but I'm not sure whether that's genuine or whether I'm just accessing them wrong.

dwsutherland:
Fixed the updated deltas test:
https://github.com/cylc/cylc-flow/pull/5660/commits/6dd1b96988ea5df3ea7e0b1f7d3b3de472c70f84

tests/integration/test_increment_graph_window.py
dwsutherland: fixed, needed to call `window_depth_finder`

oliver-sanders:
I've been hammering this PR with tests, it seems to be working pretty well :+1:, but I did manage to find a couple of errors, sorry!

n-window colours:
* 0=blue
* 1=teal
* 2=green
* 3=yellow
* 4=orange
* 5=red

#### 1. From the test workflow above in the n=3 window

Going back through the test workflow above, the test workflow worked fine in the n=0-2 range.

At n=3, there is one error when the workflow gets to task "c":



g, m & s are n=4 so shouldn't appear in this window.

I do occasionally see things like this occur in the graph view momentarily during an update, but this is only a momentary flash as a result of UI update order, not something to worry about. However, in this case the n=4 tasks don't disappear promptly. Inspecting the workflow in GraphiQL, the tasks blink, g, m & s all appear with `graphDepth: 3`.

#### 2. Complex Workflow n=5

I had a go at running the complex workflow in the n=5 window and spotted a similar error in the graph. I've pulled out the edges leading in/out of the node and highlighted them in red. The task before is n=5, the tasks after are also n=5 but strangelt the task in-between is n=4?

Screenshot from 2023-10-25 15-58-00

cylc/flow/network/schema.py
```diff
@@ -1072,6 +1078,11 @@ class Meta:
     depth = Int(
         description='The family inheritance depth',
     )
+    graph_depth = Int(
+        description=sstrip('''
+            The n-window graph edge depth from closet active task(s).
+        '''),
+    )
```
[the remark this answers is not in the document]
dwsutherland: Because this is the depth of the task proxy, the other is the max graph depth (the n in n-window).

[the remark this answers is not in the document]
oliver-sanders: (this is the field that the cylc-ui diff is using to colour in tasks by n-window depth, the other field will be used by the ui to display the configured n-window depth so that the user can change the window they are viewing)

cylc/flow/data_store_mgr.py
```diff
@@ -744,146 +741,343 @@ def increment_graph_window(
...
+                len(loc)
+                for loc in active_locs
+                if p_tag not in loc
+            )
+            # add the most distant child as a trigger to prune it.
+            boundary_nodes.update(*(
+                active_locs[loc]
+                for loc in active_locs
+                if p_tag not in loc and len(loc) >= max_level
+            ))
```
MetRonnie, line 1051: Where is the `ValueError` expected to occur here?

cylc/flow/data_store_mgr.py
```diff
@@ -964,29 +1161,31 @@ def generate_ghost_task(
                 Used to determine whether to load DB state.
             itask (cylc.flow.task_proxy.TaskProxy):
                 Update task-node from corresponding task proxy object.
+            n_depth (int): n-window graph edge distance.
 
         Returns:
             (is_orphan, graph_children)
 
         Orphan tasks with no children return (True, None) respectively.
```
MetRonnie, line 1170: out of date
```suggestion
```

cylc/flow/data_store_mgr.py
```diff
@@ -1550,6 +1758,82 @@ def update_workflow_states(self):
...
+        """Recalculate window depths, creating depth deltas."""
+        # Setup new window depths
+        n_window_depths: Dict(int, Set(str)) = {
+            0: set().union(self.all_task_pool)
+        }
+
+        depth = 1
+        # Since starting from smaller depth, exclude those whose depth has
+        # already been found.
+        depth_found_tasks: Set(str) = set().union(self.all_task_pool)
```
MetRonnie, line 1802:
```suggestion
        depth_found_tasks: Set[str] = self.all_task_pool.copy()
```

cylc/flow/data_store_mgr.py
```diff
@@ -1550,6 +1758,82 @@ def update_workflow_states(self):
...
+                for k, v in self.n_window_nodes.items()
+                if k in self.all_task_pool
+            ))
+        )
+
+    def window_depth_finder(self):
+        """Recalculate window depths, creating depth deltas."""
+        # Setup new window depths
+        n_window_depths: Dict(int, Set(str)) = {
+            0: set().union(self.all_task_pool)
```
MetRonnie, line 1796:
```suggestion
        n_window_depths: Dict[int, Set[str]] = {
            0: self.all_task_pool.copy()
```

cylc/flow/data_store_mgr.py
```diff
@@ -1584,15 +1868,18 @@ def prune_data_store(self):
...
-            if tp_id in self.n_window_edges:
-                del self.n_window_edges[tp_id]
             if tp_id in tp_data:
                 node = tp_data[tp_id]
             elif tp_id in tp_added:
                 node = tp_added[tp_id]
             else:
                 node_ids.remove(tp_id)
                 continue
+            self.n_window_edges.difference_update(set(node.edges))
```
MetRonnie, line 1878: `set` methods can be called straight on iterables
```suggestion
            self.n_window_edges.difference_update(node.edges)
```

tests/integration/test_increment_graph_window.py
```diff
@@ -0,0 +1,410 @@
...
+            # add the next task to the pool
+            add_task(schd, active_task)
+            # run the graph window algorithm
+            increment_graph_window(schd, active_task)
+            # get the deltas which increment_graph_window created
+            added, updated, pruned = get_deltas(schd)
+
+            # compare the n-window in the store to what we were expecting
+            _n_window = await get_n_window(schd)
+            assert _n_window == n_window
```
MetRonnie, line 266: I think it would be clearer if `n_window -> expected_n_window` and `_n_window -> n_window` or just
```suggestion
            assert (await get_n_window(schd)) == n_window
```

tests/integration/test_increment_graph_window.py
```diff
@@ -0,0 +1,410 @@
...
+
+            # compate the deltas to what we were expecting
```
MetRonnie, line 268:
```suggestion
            # compare the deltas to what we were expecting
```

tests/functional/n-window/02-big-window.t
MetRonnie: Is this functional test still needed now that integration tests have been added? Or do they cover different things?

cylc/flow/data_store_mgr.py
dwsutherland: This is from the `max`:
```
>>> max(blah for blah in set())
ValueError: max() arg is an empty sequence
```
which can happen with isolates and end-of-branch nodes.

cylc/flow/data_store_mgr.py
```diff
@@ -744,146 +741,343 @@ def increment_graph_window(
...
+                    ):
+                        active_walk['done_locs'].add(w_loc[:-1])
+                        active_walk['done_ids'].update(
+                            active_locs[w_loc[:-1]]
+                        )
+            new_locs = []
+            for loc in working_locs:
+                if loc in active_locs and len(loc) < self.n_edge_distance:
+                    new_locs.extend((loc + c_tag, loc + p_tag))
+            working_locs = new_locs
```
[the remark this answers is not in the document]
dwsutherland: That (`List[Tuple[str, str]]`) isn't quite what we want, this works though:
```
            working_locs = [
                new_loc
                for loc in working_locs
                if loc in active_locs and len(loc) < self.n_edge_distance
                for new_loc in (loc + c_tag, loc + p_tag)
            ]
```

tests/functional/n-window/02-big-window.t
dwsutherland: At min tests API and cylc-show on large windows, also larger window size than integration tests (I think)

dwsutherland:
> I did manage to find a couple of errors, sorry!

No problem, 1st problem is fixed, it was the pruning at the beginning of the workflow run.

Haven't tested the 2nd problem, can check to see this isn't also fixed now? (just in case it was downstream)

cylc/flow/data_store_mgr.py
```diff
@@ -1550,6 +1758,82 @@ def update_workflow_states(self):
...
+                v
+                for k, v in self.n_window_nodes.items()
+                if k in self.all_task_pool
+            ))
+        )
+
+    def window_depth_finder(self):
+        """Recalculate window depths, creating depth deltas."""
+        # Setup new window depths
+        n_window_depths: Dict(int, Set(str)) = {
```
MetRonnie, line 1795:
```suggestion
        n_window_depths: Dict[int, Set[str]] = {
```

cylc/flow/data_store_mgr.py
MetRonnie: I would suggest splitting it to avoid potentially catching a ValueError in the next bit
```suggestion
        try:
            max_level = max(
                len(loc)
                for loc in active_locs
                if p_tag not in loc
            )
        except ValueError:
            pass
        else:
            # add the most distant child as a trigger to prune it.
            boundary_nodes.update(*(
                active_locs[loc]
                for loc in active_locs
                if p_tag not in loc and len(loc) >= max_level
            ))
```

oliver-sanders:
Done some more testing...

#### Using longest rather than shortest n-window distance

Spotted whilst testing this example in the n=10 window:

```ini
[scheduler]
    allow implicit tasks = True
[scheduling]
    initial cycle point = 1
    cycling mode = integer
    [[graph]]
        P1 = """
            a => b => c
            b[-P1] => b
        """
```

The task "c" should be n=2 for the first 6 cycles, but I actually got:

* 6/c - n=2
* 5/c - n=3
* 4/c - n=4
* 3/c - n=5
* 2/c - n=6
* 1/c - n=7

Screenshot from 2023-10-30 16-15-46

Note, it doesn't reproduce the same way each time?!


#### Reloading resets graph window

Not strictly related to the changes on this PR. Would be good to squeeze in now if easy, otherwise can bump to an issue.

Reloading the workflow appears to reset the n-window. E.G. if I set the window to n=2, then reload, it goes back to n=1.



#### Graph errors

I've seen a couple of bits of traceback in the graph view with `bbox` undefined (note the error in the console won't reference the variable name properly due to minification). I think this can only happen if the edges reference nodes which are not present in the nodes.

I.E. if there are sources or targets in `edges { source, target }` which are not in `TaskProxies { id }`.

I've seen this when the workflow is paused which is surprising as all deltas should have been applied (rules out race conditions in delta batching).

I can't get a reproducible example at the moment so can't confirm the cause. Could be unrelated.

dwsutherland:
> The task "c" should be n=2 for the first 6 cycles, but I actually got:

This one turned out to be and easy fix:
```
@@ -1788,6 +1788,7 @@ class DataStoreMgr:
                 if k in self.all_task_pool
             ))
         )
+        self.update_window_depths = True
 
     def window_depth_finder(self):
         """Recalculate window depths, creating depth deltas."""
```
(needed to flag depth to be recalculated after window size change)
image

dwsutherland:
Don't know what's with the test failures .. they are working locally (unit and integration testing at Python 3.10.12)

oliver-sanders:
(something odd is going on with tests, it's not just this branch)

dwsutherland:
> Don't know what's with the test failures .. they are working locally (unit and integration testing at Python 3.10.12)

Happening in other PRs so 🤷

MetRonnie:
Tests fixed in #5797

MetRonnie (approved):
Worked well in my testing

oliver-sanders:
Been back through the test workflow above from n=1 to n=4, no problems!

Re-ran the one which showed the longest-path issues before, perfect (and kinda pretty):

Screenshot from 2023-11-02 10-04-53

Tried a bunch of workflows from my cylc-src dir, the algorithm is now producing the correct results for all of them :tada:!

oliver-sanders:
Profiling results for the following workflow in simulation mode:

```ini
[task parameters]
    x = 1..100
    y = 1..100
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    final cycle point = 1
    [[graph]]
        P1 = d[-P1] => a => b<x> => c<y> => d
```

* The number of `increment_graph_window` calls is down from 50501, to 201!
* The total time taken by `increment_graph_window` is down from 7.6s to 4.3s (10% of total runtime down to 6%).
* The number of `generate_graph_(parents|children)` calls are down from 10301 to 201.
* Total Cylc runtime is down by 5%.
* The number of calls to `Prerequisite.api_dump` (now the biggest offender) is unchanged.

Good positive change, increases functionality whilst improving performance :100:!

oliver-sanders (approved):
Thanks for all your hard work @dwsutherland, this looked like a tricky one!

PR #5661, opened 2023-08-01
Succeeded state totals fix

### LINKED ISSUE #5595: data store: task state count too high

dwsutherland:
closes #5595 

~~This improves the issue, but there's momentary higher `succeeded` counts while pruning is immanent (which is the underlying cause of the problem).~~
#5650 may have improved this too, by updating the store more frequently (as flagged by the store). 

**Check List**

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -469,6 +469,8 @@ class DataStoreMgr:
         self.parents = {}
         self.state_update_families = set()
         self.updated_state_families = set()
+        # Update workflow state totals once more post delta application.
+        self.state_update_follow_on = False
         self.n_edge_distance = 1
         self.next_n_edge_distance = None
         self.latest_state_tasks = {
@@ -496,6 +498,7 @@ class DataStoreMgr:
         self.publish_deltas = []
         # internal n-window
         self.all_task_pool = set()
+        self.all_n_window_nodes = set()
         self.n_window_nodes = {}
         self.n_window_edges = {}
         self.n_window_boundary_nodes = {}
@@ -505,6 +508,7 @@ class DataStoreMgr:
         self.prune_flagged_nodes = set()
         self.pruned_task_proxies = set()
         self.updates_pending = False
+        self.updates_pending_follow_on = False
         self.publish_pending = False
 
     def initiate_data_model(self, reloaded=False):
@@ -1488,7 +1492,6 @@ class DataStoreMgr:
 
     def update_data_structure(self, reloaded=False):
         """Workflow batch updates in the data structure."""
-
         # load database history for flagged nodes
         self.apply_task_proxy_db_history()
 
@@ -1497,22 +1500,21 @@ class DataStoreMgr:
             self.n_edge_distance = self.next_n_edge_distance
             self.next_n_edge_distance = None
 
+        self.updates_pending_follow_on = False
         self.prune_data_store()
-        if self.state_update_families:
+        if self.updates_pending:
+            # update
             self.update_family_proxies()
 
-        next_update_pending = False
-        if self.updates_pending:
             # Update workflow statuses and totals if needed
             self.update_workflow()
 
             # Don't process updated deltas of pruned nodes
-            if self.pruned_task_proxies:
-                next_update_pending = True
             self.prune_pruned_updated_nodes()
 
-            # Apply current deltas
+            # Gather deltas
             self.batch_deltas()
+            # Apply all deltas
             self.apply_delta_batch()
 
         if reloaded:
@@ -1524,7 +1526,7 @@ class DataStoreMgr:
             # Gather this batch of deltas for publish
             self.publish_deltas = self.get_publish_deltas()
 
-        self.updates_pending = next_update_pending
+        self.updates_pending = self.updates_pending_follow_on
 
         # Clear deltas
         self.clear_delta_batch()
@@ -1539,21 +1541,21 @@ class DataStoreMgr:
             return
 
         # Keep all nodes in the path of active tasks.
-        in_paths_nodes = set().union(*[
+        self.all_n_window_nodes = set().union(*(
             v
             for k, v in self.n_window_nodes.items()
             if k in self.all_task_pool
-        ])
+        ))
         # Gather all nodes in the paths of tasks flagged for pruning.
-        out_paths_nodes = self.prune_flagged_nodes.union(*[
+        out_paths_nodes = self.prune_flagged_nodes.union(*(
             v
             for k, v in self.n_window_nodes.items()
             if k in self.prune_flagged_nodes
-        ])
+        ))
         # Trim out any nodes in the runahead pool
         out_paths_nodes.difference(self.all_task_pool)
         # Prune only nodes not in the paths of active nodes
-        node_ids = out_paths_nodes.difference(in_paths_nodes)
+        node_ids = out_paths_nodes.difference(self.all_n_window_nodes)
         # Absolute triggers may be present in task pool, so recheck.
         # Clear the rest.
         self.prune_flagged_nodes.intersection_update(self.all_task_pool)
@@ -1593,6 +1595,7 @@ class DataStoreMgr:
         if node_ids:
             self.pruned_task_proxies.update(node_ids)
             self.updates_pending = True
+            self.updates_pending_follow_on = True
 
     def _family_ascent_point_prune(
             self, fp_id, node_ids, parent_ids, checked_ids, prune_ids):
@@ -1607,12 +1610,9 @@ class DataStoreMgr:
         if fp_id in fp_data:
             fam_node = fp_data[fp_id]
             # Gather child families, then check/update recursively
-            child_fam_nodes = [
-                n_id
-                for n_id in fam_node.child_families
-                if n_id not in checked_ids
-            ]
-            for child_id in child_fam_nodes:
+            for child_id in fam_node.child_families:
+                if child_id in checked_ids:
+                    continue
                 self._family_ascent_point_prune(
                     child_id, node_ids, parent_ids, checked_ids, prune_ids)
             child_tasks = set(fam_node.child_tasks)
@@ -1623,12 +1623,15 @@ class DataStoreMgr:
                     child_tasks.update(fp_updated[fp_id].child_tasks)
                 if fp_updated[fp_id].child_families:
                     child_families.update(fp_updated[fp_id].child_families)
-            # if any child tasks or families are active, don't prune.
+            # if any child tasks or families are in window, don't prune.
             if (
-                    child_tasks.difference(node_ids)
-                    or child_families.difference(prune_ids)
+                child_tasks.difference(node_ids)
+                or child_families.difference(prune_ids)
             ):
-                if fp_id in prune_ids:
+                if (
+                    child_tasks.intersection(node_ids)
+                    or child_families.intersection(prune_ids)
+                ):
                     self.state_update_families.add(fp_id)
             else:
                 if fam_node.first_parent:
@@ -1684,7 +1687,7 @@ class DataStoreMgr:
             self._family_ascent_point_update(
                 next(iter(self.state_update_families)))
         if self.updated_state_families:
-            self.updates_pending = True
+            self.state_update_follow_on = True
 
     def _family_ascent_point_update(self, fp_id):
         """Updates the given family and children recursively.
@@ -1695,6 +1698,7 @@ class DataStoreMgr:
         ancestor to the set of families flagged for update.
 
         """
+        all_nodes = self.all_n_window_nodes
         fp_added = self.added[FAMILY_PROXIES]
         fp_data = self.data[self.workflow_id][FAMILY_PROXIES]
         if fp_id in fp_data:
@@ -1709,12 +1713,9 @@ class DataStoreMgr:
                 self.state_update_families.remove(fp_id)
             return
         # Gather child families, then check/update recursively
-        child_fam_nodes = [
-            n_id
-            for n_id in fam_node.child_families
-            if n_id not in self.updated_state_families
-        ]
-        for child_fam_id in child_fam_nodes:
+        for child_fam_id in fam_node.child_families:
+            if child_fam_id in self.updated_state_families:
+                continue
             self._family_ascent_point_update(child_fam_id)
         if fp_id in self.state_update_families:
             fp_updated = self.updated[FAMILY_PROXIES]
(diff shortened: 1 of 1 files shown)


cylc/flow/data_store_mgr.py
```diff
@@ -505,6 +505,7 @@ def __init__(self, schd):
         self.prune_flagged_nodes = set()
         self.pruned_task_proxies = set()
         self.updates_pending = False
+        self.next_update_pending = False
         self.publish_pending = False
```
oliver-sanders, line 509: Could do with a comment explaining these three flags.

cylc/flow/data_store_mgr.py
```diff
@@ -1685,6 +1689,12 @@ def update_family_proxies(self):
                 next(iter(self.state_update_families)))
         if self.updated_state_families:
             self.updates_pending = True
+            if (
+                self.deltas[FAMILY_PROXIES].pruned
+                or self.deltas[TASK_PROXIES].pruned
+            ):
+                self.state_update_families.update(self.updated_state_families)
+                self.next_updates_pending = True
```
oliver-sanders, line 1697:
```suggestion
                self.next_update_pending = True
```

cylc/flow/data_store_mgr.py
```diff
@@ -1524,7 +1524,8 @@ def update_data_structure(self, reloaded=False):
             # Gather this batch of deltas for publish
             self.publish_deltas = self.get_publish_deltas()
 
-        self.updates_pending = next_update_pending
+        self.updates_pending = self.next_update_pending
```
oliver-sanders, line 1527: If `self.next_update_pending` is only used to set `self.updates_pending`, can we merge these variables.

oliver-sanders (commented):
I tested this against the original example with https://github.com/cylc/cylc-flow/pull/5650 merged in.

The number of succeeded tasks has dropped from 4 to 3 but still doesn't drop down to 1:

Screenshot from 2023-08-01 16-36-13

Could we hook this update into the add/update/prune functions directly?

cylc/flow/data_store_mgr.py
dwsutherland: The problem is; the `update_data_structure` does things in a specific order (which we can change):
- Prune data-store (remove from internal sets and create pruned deltas for application to the store)
- Update family proxies (states and state totals updated (including root), **based on current DB**
- Update the workflow fields (including state and state totals, by collating all the states of the cycle roots)
- Prune deltas of pruned nodes (so they don't get applied)
- Batch all deltas
- Apply deltas to data-store
- publish deltas

So the issue is that the state_totals get applied before the prune deltas get applied to the data-store (before the old data gets removed, hence why it's only succeeded counts).. 
So what I was trying to do with `self.next_update_pending` is run the update to redo the state-totals (..etc) after a pruning has been applied.

So a couple of options would be:
- Apply the pruning to the data-store first then do the state totals (and just make sure we manage the deltas properly)
- Ignore those-to-be-pruned in the state-totals calculations

The latter is proving to be a little tricky.. so I might try the former ..

dwsutherland: Actually strike that .. I've found the issue

dwsutherland:
Fixed now:
image

cylc/flow/data_store_mgr.py
```diff
@@ -1735,7 +1742,11 @@ def _family_ascent_point_update(self, fp_id):
                     state_counter += Counter(dict(child_node.state_totals))
             # Gather all child task states
             task_states = []
-            for tp_id in fam_node.child_tasks:
+            for tp_id in [
+                ct_id
+                for ct_id in fam_node.child_tasks
+                if not all_nodes or ct_id in all_nodes
+            ]:
```
MetRonnie, line 1749: Can avoid intermediate list comprehension
```py
for tp_id in fam_node.child_tasks:
    if not all_nodes or tp_id in all_nodes:
```

cylc/flow/data_store_mgr.py
```diff
@@ -504,6 +507,7 @@ def __init__(self, schd):
         self.prune_trigger_nodes = {}
         self.prune_flagged_nodes = set()
         self.pruned_task_proxies = set()
+        self.prune_pending = False
```
MetRonnie, line 510: I think this is unnecessary
```suggestion
```

cylc/flow/data_store_mgr.py
```diff
@@ -1497,22 +1500,21 @@ def update_data_structure(self, reloaded=False):
             self.n_edge_distance = self.next_n_edge_distance
             self.next_n_edge_distance = None
 
+        self.prune_pending = False
```
MetRonnie, line 1503:
```suggestion
```

cylc/flow/data_store_mgr.py
```diff
@@ -1524,7 +1526,7 @@ def update_data_structure(self, reloaded=False):
             # Gather this batch of deltas for publish
             self.publish_deltas = self.get_publish_deltas()
 
-        self.updates_pending = next_update_pending
+        self.updates_pending = self.prune_pending
```
MetRonnie, line 1529:
```suggestion
        self.updates_pending = False
```

cylc/flow/data_store_mgr.py
```diff
@@ -1593,6 +1595,7 @@ def prune_data_store(self):
         if node_ids:
             self.pruned_task_proxies.update(node_ids)
             self.updates_pending = True
+            self.prune_pending = True
```
MetRonnie, line 1598:
```suggestion
```

MetRonnie (commented):
Looks like it fixes the problem

cylc/flow/data_store_mgr.py
dwsutherland: Not quite, the extra flag actually provides an additional call to `update_data_structure` after pruning (which is needed because pruning effects state totals).

cylc/flow/data_store_mgr.py
dwsutherland: see below

cylc/flow/data_store_mgr.py
dwsutherland: see below

cylc/flow/data_store_mgr.py
dwsutherland: see above

cylc/flow/data_store_mgr.py
dwsutherland: Which is faster?

dwsutherland: I mean I suppose not creating a list comprehension .. if the evaluation of the if block outside and a `continue` is faster

oliver-sanders: I prefer the comprehension, but for performance swap the square brackets for round ones when only using the comprehension for iteration purposes, this avoides creating an intermediate list i.e:

```python
for tp_id in (
    ct_id
    ...
```

oliver-sanders:
Tried this out in `cylc tui` with the original example, the state totals always matched the tree view perfectly :1st_place_medal:

cylc/flow/data_store_mgr.py
```diff
@@ -1539,7 +1541,7 @@ def prune_data_store(self):
             return
 
         # Keep all nodes in the path of active tasks.
-        in_paths_nodes = set().union(*[
+        self.all_n_window_nodes = set().union(*[
```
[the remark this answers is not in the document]
dwsutherland: v is a set

cylc/flow/data_store_mgr.py
MetRonnie: I now see `prune_data_store()` may set `self.prune_pending` to `True`. But it also sets `self.updates_pending` to `True`, which leads me to believe this line can be removed altogether.

```suggestion
```

cylc/flow/data_store_mgr.py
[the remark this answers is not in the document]
MetRonnie: Not quite, I had look into this and I think that `v` is a tuple so 
```python-console
>>> set().union((0, 6), (1, 6), (2, 6))
{0, 1, 2, 6}
```
whereas just doing a set comprehension would result in
```
{(0, 6), (1, 6), (2, 6)}
```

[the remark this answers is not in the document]
oliver-sanders: In that case, one comprehension with two `for`'s will do it:

```python
>>> d = {('a', 'b'): 'x', ('c', 'd'): 'y'}
>>> { i for k in d for i in k }
{'b', 'd', 'c', 'a'}
```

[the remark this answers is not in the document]
MetRonnie: I think the `set().union(*...)` approach, while maybe unusual, is clearer actually, and seems to be faster from a bit of playing about with `%timeit`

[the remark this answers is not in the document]
dwsutherland: Not sure having to iterate out elements would be faster than a union of sets?

cylc/flow/data_store_mgr.py
dwsutherland: Looks like you win Ronnie XD

cylc/flow/data_store_mgr.py
dwsutherland: I don't think so, as that's what flags `update_data_structure` to run again (it's checked in the scheduler main loop)

dwsutherland: It also sets `self.updates_pending`  to true for the `if` block below `prune_data_store()`  ... Probably a better way of doing it, but the justification is there.

dwsutherland: Essentially when `update_data_structure` is called .. If pruning was done, not only run all the other updating (including delta application and publishing), but run the `update_data_structure` again (after pruning has been applied) to recalculate state totals etc for application and publishing.

dwsutherland: If we apply the pruning beforehand, we may loose intermediate states.

dwsutherland: Also, `update_data_structure` can be called without `self.updates_pending`  otherwise we could get rid of that `if self.updates_pending` block  and set it to `False` at the top .. 
If we broke up `update_data_structure` into two separate functions, maybe, but what if there are also updates pending for the alternate call?

dwsutherland: I've renamed it `self.updates_pending_follow_on`  .. Just for clarity.

MetRonnie: Sorry but unless I'm missing something, `self.updates_pending_follow_on` is not actually used. It already matches the value of `self.updates_pending` here in `prune_data_store()`
https://github.com/cylc/cylc-flow/blob/54f4756a4c3fe80dd8edc8a95f5ca78878274896/cylc/flow/data_store_mgr.py#L1597-L1598

If you are saying that the values of `self.updates_pending_follow_on` and `self.updates_pending` can diverge in between here
https://github.com/cylc/cylc-flow/blob/54f4756a4c3fe80dd8edc8a95f5ca78878274896/cylc/flow/data_store_mgr.py#L1503-L1504

and this line that this comment thread is on
https://github.com/cylc/cylc-flow/blob/54f4756a4c3fe80dd8edc8a95f5ca78878274896/cylc/flow/data_store_mgr.py#L1529

?

MetRonnie: Ah I guess if `self.updates_pending` is `True` to begin with it may not get set to `False`.

I think this would simplify it a bit
```diff
diff --git a/cylc/flow/data_store_mgr.py b/cylc/flow/data_store_mgr.py
index a15341d8f..982fe046d 100644
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -509,5 +509,4 @@ class DataStoreMgr:
         self.pruned_task_proxies = set()
         self.updates_pending = False
-        self.updates_pending_follow_on = False
         self.publish_pending = False
 
@@ -1501,6 +1500,5 @@ class DataStoreMgr:
             self.next_n_edge_distance = None
 
-        self.updates_pending_follow_on = False
-        self.prune_data_store()
+        follow_on = self.prune_data_store()
         if self.updates_pending:
             # update
@@ -1527,5 +1525,5 @@ class DataStoreMgr:
             self.publish_deltas = self.get_publish_deltas()
 
-        self.updates_pending = self.updates_pending_follow_on
+        self.updates_pending = follow_on
 
         # Clear deltas
@@ -1533,11 +1531,14 @@ class DataStoreMgr:
         self.clear_delta_store()
 
-    def prune_data_store(self):
-        """Remove flagged nodes and edges not in the set of active paths."""
+    def prune_data_store(self) -> bool:
+        """Remove flagged nodes and edges not in the set of active paths.
+
+        Returns whether updates are pending after this.
+        """
 
         self.family_pruned_ids.clear()
 
         if not self.prune_flagged_nodes:
-            return
+            return False
 
         # Keep all nodes in the path of active tasks.
@@ -1595,6 +1596,7 @@ class DataStoreMgr:
         if node_ids:
             self.pruned_task_proxies.update(node_ids)
-            self.updates_pending = True
-            self.updates_pending_follow_on = True
+            return True
+
+        return False
 
     def _family_ascent_point_prune(
```

dwsutherland: No, it wouldn't .. If there was no pruning, then it would get set to `False`..
The reason I would keep it as a separate variable is, if another operation requires a follow-on update.

dwsutherland: > Unless you are saying that the values of `self.updates_pending_follow_on` and `self.updates_pending` can diverge in between here

Exactly, no pruning, they diverge.. All this:
```
        if node_ids:
            self.pruned_task_proxies.update(node_ids)
            self.updates_pending = True
            self.updates_pending_follow_on = True
```
Says is (the `if node_ids:` is key) if pruning was actually done, update all the other stuff (states ..etc)..
(note: running `self.prune_data_store()` doesn't imply it actually found something to prune, but if it did, do an update, and a follow on update)

MetRonnie: Ok I think I'm convinced now 😁

MetRonnie (approved):

oliver-sanders:
Had a look at testing, it's a pain in the neck.

PR #5679, opened 2023-08-10
Docs: fix hidden note

MetRonnie:
I did a grep for other possible unintentionally hidden admonitions and couldn't find any others.


oliver-sanders (approved):
Good spot!

PR #5693, opened 2023-08-16
Fix command logging

### LINKED ISSUE #5691: not all commands logged with actor
When commands are actioned by another user via the uiserver, the user who actioned the command is supposed to be logged.

For some reason this is not the case for the `stop` command.

E.G. here's some log showing `other-user` actioning `set_verbosity` followed by `stop`:

```
2023-08-15T13:44:51Z INFO - [command] set_verbosity (issued by other-user)
2023-08-15T13:44:51Z INFO - Command actioned: set_verbosity(10)
2023-08-15T13:44:53Z DEBUG - Processing 1 queued command(s)
2023-08-15T13:44:53Z INFO - Command actioned: stop(mode=REQUEST(CLEAN))
```

Their user name was recorded for the first command but not the second.

Inspect the logging to ensure that the user name is logged for all commands (unless they were actioned by the workflow owner).

hjoliver:
Close #5691 

Log issuer name, if not owner, for all commands.

Question: this now logs `[command] <command-name>` for all commands, with or without `issuer != owner`. Arguably that is duplication (with the later "command actioned" message) but it is more consistent than before. Any objections?

**Check List**

### DIFF
--- a/cylc/flow/network/resolvers.py
+++ b/cylc/flow/network/resolvers.py
@@ -686,19 +686,22 @@ class Resolvers(BaseResolvers):
         self, command: str, kwargs: Dict[str, Any], meta: Dict[str, Any]
     ) -> Optional[Tuple[bool, str]]:
         """Map between GraphQL resolvers and internal command interface."""
+
+        log_msg = f"[command] {command}"
+        user = meta.get('auth_user', self.schd.owner)
+        if user != self.schd.owner:
+            log_msg += (f" (issued by {user})")
+        LOG.info(log_msg)
+
         method = getattr(self, command, None)
         if method is not None:
             return method(**kwargs)
+
         try:
             self.schd.get_command_method(command)
         except AttributeError:
             raise ValueError(f"Command '{command}' not found")
-        if command != "put_messages":
-            log_msg = f"[command] {command}"
-            user = meta.get('auth_user', self.schd.owner)
-            if user != self.schd.owner:
-                log_msg += (f" (issued by {user})")
-            LOG.info(log_msg)
+
         self.schd.queue_command(
             command,
             kwargs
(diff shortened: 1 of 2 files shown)


cylc/flow/network/resolvers.py
```diff
@@ -686,19 +686,22 @@ async def _mutation_mapper(
         self, command: str, kwargs: Dict[str, Any], meta: Dict[str, Any]
     ) -> Optional[Tuple[bool, str]]:
         """Map between GraphQL resolvers and internal command interface."""
+
+        log_msg = f"[command] {command}"
+        user = meta.get('auth_user', self.schd.owner)
+        if user != self.schd.owner:
+            log_msg += (f" (issued by {user})")
+        LOG.info(log_msg)
```
oliver-sanders, line 694: This message was previously only logged if:

* The command was not issued by the user.
* The command was not `put_messages` (#3647)

hjoliver: Yes that statement is correct. (But what's you're point?)

Now it is logged:

- regardless of user, but "issued by {user}" is added if not the owner
  - this is more consistent: log the message, but add the extra info is necessary
    
- for all commands including `put_message`
  - were we singling this command out because (a) we don't want to log it because it is so frequent; and/or (b) because it is "normal" for the job owner to be different due to the owner simply having a different user name on the job platform? 
  - my thinking was, other users could in principle issue `cylc message` commands and so it should be logged, and all commands should be logged equally
  
(However, I did pose the question in the issue description, in case discussion was needed).

oliver-sanders: Fair enough, how about only log `put_message` commands if they were issued from another user account?

hjoliver: Note this fix goes beyond `put_messages`. Previously this message was not logged at all for commands (such as `stop`) that don't have wrapper methods in the resolver.

But I'll tweak the fix for the `put_message` case...

hjoliver: Done.

oliver-sanders (approved):

PR #5658, opened 2023-07-31
Implement "cylc set" command

hjoliver:
Implement accepted proposal [A New Command for Manual Task Forcing](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-cylc-set.md)
- Close #5642 
- Close #5638
- Close #5572 
- Close #5882 

Supersede #5412 (fix expire triggers)
- Close #5364
- Close #5354
- Close #5447
- Close #4527 

TODO
- [x] Rename `set-outputs` to `set` 
- [x] Update the CLI and GraphQL schema 
- [x]  Custom outputs: prohibit "all", "required", "optional" keywords, "," character, and "_cylc" prefix. 
- [x] Reproduce old set-outputs behaviour with new command options and query
- [x] Implement the new logic for setting prerequisites [FIRST CUT done]
- [x] Implement the new logic for setting outputs
- [x] Test new schema works as expected in the UI
- [x] Tests for all new functionality
- [x] Update documentation -> Punted to https://github.com/cylc/cylc-doc/issues/688
- [x] This is a breaking change, add an entry into the cylc-doc changes page -> Punted to https://github.com/cylc/cylc-doc/issues/688

Side effects:
- rename `cylc set-verbosity` to `cylc verbosity` to avoid command ambiguity apocalypse
- command UUIDs 
- cleaner log message prefix for tasks

Follow-up:
-   `cylc set --pre=cycle`

**PR Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [x] `CHANGES.md` entry included if this is a change that can affect users

PR #5698, opened 2023-08-17
Flow-specific hold/release

### LINKED ISSUE #4277: Task hold needs to be flow-specific
Follow-up to #4238

hjoliver:
Close #4277 

Hold/release tasks in a specific flow (or any flow, by default)

- [x] Implement `cylc hold --flow=n`
- [x] Same for `cylc release --flow=n`

If `--flow` is not used, matching tasks will be held or released regardless of flow number.
  - (most of the time there will only be one flow)

If `--flow=n` is specified, matching tasks will be held or released if their flow numbers include `n`
  - (if a merged task `foo` has flow numbers `{2,3}` it belongs to flow `2` (and `3`), so if I ask to hold `foo` in flow `2` (or `3`) it should be held. Ditto for release).

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/646
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch. (not a bug fix)

### DIFF
--- a/cylc/flow/flow_mgr.py
+++ b/cylc/flow/flow_mgr.py
@@ -38,7 +38,7 @@ def validate_flow_opt(val):
             int(val)
         except ValueError:
             raise InputError(f"--flow={val}: value must be integer.")
- 
+
 
 class FlowMgr:
     """Logic to manage flow counter and flow metadata."""
--- a/cylc/flow/flow_mgr.py
+++ b/cylc/flow/flow_mgr.py
@@ -20,6 +20,7 @@ from typing import Dict, Set, Optional
 import datetime
 
 from cylc.flow import LOG
+from cylc.flow.exceptions import InputError
 from cylc.flow.workflow_db_mgr import WorkflowDatabaseManager
 
 
@@ -29,6 +30,15 @@ FLOW_ALL = "all"
 FLOW_NEW = "new"
 FLOW_NONE = "none"
 
+
+def validate_flow_opt(val):
+    """Validate command line --flow opions."""
+    if val is not None:
+        try:
+            int(val)
+        except ValueError:
+            raise InputError(f"--flow={val}: value must be integer.")
+ 
 
 class FlowMgr:
     """Logic to manage flow counter and flow metadata."""
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -2052,6 +2052,11 @@ class Release(Mutation, TaskMutation):
         ''')
         resolver = mutator
 
+    class Arguments(TaskMutation.Arguments):
+        flow_num = Int(
+            description='Number of flow to release.'
+        )
+
 
 class Kill(Mutation, TaskMutation):
     # TODO: This should be a job mutation?
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -2035,6 +2035,11 @@ class Hold(Mutation, TaskMutation):
         ''')
         resolver = mutator
 
+    class Arguments(TaskMutation.Arguments):
+        flow_num = Int(
+            description='Number of flow to hold.'
+        )
+
 
 class Release(Mutation, TaskMutation):
     class Meta:
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -1735,7 +1735,9 @@ class SetHoldPoint(Mutation):
             description='Hold all tasks after the specified cycle point.',
             required=True
         )
-
+        flow_num = Int(
+            description='Number of flow to hold.'
+        )
     result = GenericScalar()
 
 
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1070,14 +1070,14 @@ class Scheduler:
         self.pause_workflow()
 
     @staticmethod
-    def command_set_verbosity(lvl: Union[int, str]) -> None:
+    def command_set_verbosity(level: Union[int, str]) -> None:
         """Set workflow verbosity."""
         try:
-            lvl = int(lvl)
-            LOG.setLevel(lvl)
+            level = int(level)
+            LOG.setLevel(level)
         except (TypeError, ValueError) as exc:
             raise CommandFailedError(exc)
-        cylc.flow.flags.verbosity = log_level_to_verbosity(lvl)
+        cylc.flow.flags.verbosity = log_level_to_verbosity(level)
 
     def command_remove_tasks(self, tasks) -> int:
         """Remove tasks."""
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1044,6 +1044,11 @@ class Scheduler:
                     self.data_store_mgr.delta_task_state(itask)
             return len(bad_items)
         self.task_job_mgr.kill_task_jobs(self.workflow, itasks)
+        # killed tasks get held
+        # TODO TAKE TASK HOLD OUT OF JOB MGR?
+        for itask in itasks:
+            if itask.state.is_held:
+                self.pool.hold_mgr.hold_active_task(itask, check=False)
         return len(bad_items)
 
     def command_hold(self, tasks: Iterable[str], flow_num=None) -> int:
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1043,12 +1043,10 @@ class Scheduler:
                     itask.state_reset(TASK_STATUS_FAILED)
                     self.data_store_mgr.delta_task_state(itask)
             return len(bad_items)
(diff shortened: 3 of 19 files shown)


cylc/flow/scheduler.py
```diff
@@ -905,7 +905,8 @@ def get_command_method(self, command_name: str) -> Callable:
     def queue_command(self, command: str, kwargs: dict) -> None:
         self.command_queue.put((
             command,
-            tuple(kwargs.values()), {}
+            [],
+            kwargs,
```
hjoliver, line 909: Bug fix: this was turning command kwargs into args, which could result in arg order errors.

hjoliver:
Test case:
```ini
[scheduling]
    [[graph]]
        R1 = "a => b => c => d"
[runtime]
    [[a,b,c,d]]
        pre-script = "sleep 2"
    [[a]]
        script = """
            if (( CYLC_TASK_SUBMIT_NUMBER == 1)); then
                #cylc hold         ${CYLC_WORKFLOW_ID}//1/b
                cylc hold --flow=2 ${CYLC_WORKFLOW_ID}//1/b
            fi
        """
    [[c]]
        script = """
            if (( CYLC_TASK_SUBMIT_NUMBER == 1)); then
                cylc trigger --flow=new ${CYLC_WORKFLOW_ID}//1/a
            fi
        """
```
- at start-up, `a` puts a future hold on `b` in flow 2
- flow 1 runs `b` and `c`, which starts flow 2 at `a`
- flow 2 gets held at `b`

hjoliver:
(Almost done, needs tidying and tests).

hjoliver:
One thing to note, on "hold after cycle point" functionality:

This branch makes it flow-specific, in that you can start a new flow and tell just that flow to hold after a given point:
```
cylc trigger --flow=new ...
cylc hold --after=4 --flow=2
```

However, there is still only one global hold-after point, whether it is for all tasks or for a single flow, so canceling the hold point doesn't need to be flow-specific:
```
cylc release --all ...
```

In future we should probably allow multiple flow-specific hold points, but I don't think that's a high priority. This branch is still a significant enhancment as-is.

hjoliver:
This bad boy is ready for review. I just need to up the test coverage a bit.

cylc/flow/scheduler.py
hjoliver: (And I've ended up having to fix it #5658 too; conflicts coming damn it).

hjoliver:
Test coverage good without any functional tests, although that makes me a bit nervous for this sort of thing. Any opinions on that?

dwsutherland:
With the example, does the 2nd flow get merged into the first flow?
image

It ends, after release;
```
(flow) sutherlander@cortex:cylc-flow$ cylc release --flow=2 fhold/run1//1/b

Done
```
(confirmed working)
 with flow 2 coming out the other end:
```
{
  "data": {
    "workflows": [
      {
        "id": "~sutherlander/fhold/run1",
        "taskProxies": [
          {
            "id": "~sutherlander/fhold/run1//1/c",
            "state": "succeeded",
            "flowNums": "[1]",
            "isHeld": false
          },
          {
            "id": "~sutherlander/fhold/run1//1/d",
            "state": "running",
            "flowNums": "[2]",
            "isHeld": false
          }
        ],
        "familyProxies": [
          {
            "id": "~sutherlander/fhold/run1//1/root",
            "state": "running"
          }
        ]
      }
    ]
  }
}
```

Kind of beyond the scope of this PR, but I guess I'll need to look into whether the flow numbers are correct in the data-store (perhaps no-merge is default but I would have expected flow number `"[1, 2]"` coming out of it otherwise).

Looks good.

dwsutherland (approved):

hjoliver:
> With the example, does the 2nd flow get merged into the first flow?

No, merging only occurs if the same task from different flows meet in n=0. Here, flow=1 finishes before flow-2 can catch it up:

```console
cylc log exa --no-timestamp | grep "=> succeeded" | sed -e 's/^.*-
//g'
 [1/a running job:01 flows:1] => succeeded
 [1/b running job:01 flows:1] => succeeded
 [1/c running job:01 flows:1] => succeeded
 [1/a running job:02 flows:2] => succeeded
 [1/d running job:01 flows:1] => succeeded
 [1/b running job:02 flows:2] => succeeded
 [1/c running job:02 flows:2] => succeeded
 [1/d running job:02 flows:2] => succeeded
```

cylc/flow/scripts/hold.py
```diff
@@ -114,6 +120,11 @@ def get_option_parser() -> COP:
         help="Hold all tasks after this cycle point.",
         metavar="CYCLE_POINT", action="store", dest="hold_point_string")
 
+    parser.add_option(
+        "--flow",
+        help="Hold tasks that belong to a specific flow.",
+        metavar="INT", action="store", dest="flow_num")
```
oliver-sanders, line 126: Presumably `--flow=all` is the default?

cylc/flow/network/schema.py
```diff
@@ -1735,7 +1735,9 @@ class Arguments:
             description='Hold all tasks after the specified cycle point.',
             required=True
         )
-
+        flow_num = Int(
+            description='Number of flow to hold.'
+        )
```
oliver-sanders, line 1740: Should the default be `flow=all`?

The pattern used in other mutations is a field called `flow` of type `Flow` which accepts either an integer or a string. See `FlowMutationArguments` in this file.

In this case `new` and `none` don't make sense, so it might need a new type.

cylc/flow/flow_mgr.py
```diff
@@ -30,6 +31,15 @@
 FLOW_NONE = "none"
 
 
+def validate_flow_opt(val):
```
wxtim, line 34: Doesn't this semi duplicate the logic in cylc trigger? Surely all use-cases of `--flow` should behave roughly the same, and probably be centralized in `option_parser.py`.

cylc/flow/flow_mgr.py
```diff
@@ -30,6 +31,15 @@
...
+    """Validate command line --flow opions."""
+    if val is not None:
```
wxtim, line 36: How will the parser allow this value to be `None`?

cylc/flow/task_pool.py
```diff
@@ -85,6 +86,154 @@
...
+
+    def flag_future_task(
+        self,
+        name: str,
+        point: 'PointBase',
+        flow_num: Optional[int] = None
+    ) -> None:
+        """Flag that we should hold a future task."""
+        self.hold[(name, point)] = flow_num
+        self._update_stores((name, point, True))
```
wxtim, line 165: I'm concerned that this means that you can only flag a future task for one flow at a time: 

```console
> cylc hold flows-hold//10100101T0000Z/a --flow 1
Done
> sqlite3 ~/cylc-run/flows-hold/runN/log/db 'SELECT * FROM tasks_to_hold'
a|10100101T0000Z|1
> cylc hold flows-hold//10100101T0000Z/a --flow 2
Done
> sqlite3 ~/cylc-run/flows-hold/runN/log/db 'SELECT * FROM tasks_to_hold'
a|10100101T0000Z|2
```

I would expect to get either
```
a|10100101T0000Z|1,2
# or
a|10100101T0000Z|1
a|10100101T0000Z|2
```

I also think that this means that:
```
> cylc hold flows-hold//10100101T0000Z/a --flow 1 --flow 2
Done
> sqlite3 ~/cylc-run/flows-hold/runN/log/db 'SELECT * FROM tasks_to_hold'
a|10100101T0000Z|2
```
Is not what I as a user would expect: I would expect this to either return an error ("don't use --flow twice please") or set that task as held in both flows as above.

oliver-sanders: I think it would be `--flow=1,2` rather than `--flow=1 --flow=2` but we could extend the CLI to allow this.

wxtim (commented):
I have had a good look at this. I think it mostly works as advertised, but I have a couple of questions.

cylc/flow/task_pool.py
wxtim: Ok, in that case it's only the `Reuse the option` that I have a problem with.

hjoliver:
[Update]: I think this is basically done, I just have to come back to respond to some review comments. It's possible I can do that in time for 8.3.0, let's see.

hjoliver:
(Damn it, I need to revive this one)

oliver-sanders:
This draft is way out of sync now.

Probs easier to close this PR (leaving the issue open)?

hjoliver:
Lemme take a look...

hjoliver:
Closing this as too out of date.  The need for this is is reduced now thanks to group trigger reducing the need for new flows, but it is still needed and the work was already done ... so checking if a quick redo on top of current master is easy.

PR #5704, opened 2023-08-24
Fix upgrader for max active cylc points to runahead limit.

wxtim:
Closes a bug noted locally by a customer.

Previously the converted was `x -> 'P{x}'`, but the runahead limit should be `'P{x-1}'`.

Additional feature - allow the converter description to include the templates for old and new values.

**Check List**

- [x] ~Cylc doc update~ is not required.

### DIFF
--- a/changes.d/5704.fix.md
+++ b/changes.d/5704.fix.md
@@ -1,0 +1,2 @@
+Fix bug where runahead limit was set to max active cycle points by
+the configuration converter.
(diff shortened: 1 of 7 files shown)


changes.d/5704.fix.md
```diff
@@ -0,0 +1,2 @@
+Fix bug where runahead limit was set to max active cycle points by
+the configuration converter.
```
hjoliver, line 2: Maybe be more specific so that users don't have to figure out this is Cylc 7 -> 8 conversion (if they ever read change logs 🤣 ). E.g.
```suggestion
Fix off-by-one error in automatic upgrade of Cylc 7 "max active cycle points" to Cylc 8 "runahead limit".
```

MetRonnie (approved):

PR #5708, opened 2023-08-26
Fix initial runahead limit, for offset recurrence start

### LINKED ISSUE #5705: Monthly task in an otherwise hourly workflow has unexpected scheduling runahead limit behavior
### Description

We have been working to move from Cylc 7 to Cylc 8.  Currently using Cylc 8.2.1

Our workflow has multiple jobs occurring every hour.  Also in that workflow we have a job that is scheduled to only run on the 10th of every month.  Our workflow has a high volume of parallel tasks each cycle, some of which will go into a retry state until data is available, to help manage performance we implemented a `[scheduling]runahead limit = PT6H`.  However, it seems the when loading the `10T00` entry from the graph it causes all hourly cycles/tasks from the `initial cycle point` until the monthly task to be generated and tracked causing significant delay to play the workflow or pull up in the UI.

### Reproducible Example

Relevant portion of a simplified `flow.cylc` to reproduce the problem:
```
[scheduling]
  initial cycle point = 20230812T000000Z
  runahead limit = PT4H

  [[xtriggers]]
    clock_1 = wall_clock():PT1H

  [[graph]]
    PT1H = """
        @clock_1 => TASK_A1 => TASK_A2 => TASK_A3
        @clock_1 => TASK_B1 => TASK_B2
        @clock_1 => TASK_C1
    """
    10T00 = TASK_A3 => MONTHLY_TASK
```

Will result in all cycles between 2023-08-12 and 2023-09-10 to be tracked and all tasks for those cycles to be tracked in a waiting state.  I've only put 3 sets of tasks here but our production workflow can be much larger on the cardinal hours.

### Expected Behaviour

Don't know exactly the expectation here , other than not having such an impact on performance.  If this is expected behavior, looking for recommendations for a work-around.  Only solution I've thought of is to create a separate workflow for our monthly tasks and have one workflow signal the other.

hjoliver:
Close #5705 

When computing the runahead limit at start-up (before the task pool is loaded), recurrence start points beyond the limit have to be ignored. 

Getting this right for "max cycles" limits (e.g. `P3` in integer or datetime cycling, as opposed to `PT3H` in datetime cycling) requires counting sequence points for all sequences.

**Check List**

### DIFF
--- a/changes.d/5708.fix.md
+++ b/changes.d/5708.fix.md
@@ -1,0 +1,1 @@
+Fix runahead limit at start-up, with recurrences that start beyond the limit.
--- a/cylc/flow/cycling/__init__.py
+++ b/cylc/flow/cycling/__init__.py
@@ -320,8 +320,8 @@ class SequenceBase(metaclass=ABCMeta):
     They should also provide get_async_expr, get_interval,
     get_offset & set_offset (deprecated), is_on_sequence,
     get_nearest_prev_point, get_next_point,
-    get_next_point_on_sequence, get_first_point, and
-    get_stop_point.
+    get_next_point_on_sequence, get_first_point
+    get_start_point, and get_stop_point.
 
     They should also provide a self.__eq__ implementation
     which should return whether a SequenceBase-derived object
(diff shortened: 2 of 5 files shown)


hjoliver:
@oliver-sanders - this is as per your suggested fix, but I combined the two start-up blocks into one, and made it work for integer cycling too.

oliver-sanders (approved):

PR #5710, opened 2023-08-29
Hide autoclass objects from `flow.cylc` reference

MetRonnie:
Sibling to https://github.com/cylc/cylc-doc/pull/637

We don't want to show these in the TOC tree

image

### DIFF
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -1715,10 +1715,12 @@ with Conf(
                     Custom outputs must satisfy these rules:
 
                     .. autoclass:: cylc.flow.unicode_rules.TaskOutputValidator
+                       :noindex:
 
                     Task messages must satisfy these rules:
 
                     .. autoclass:: cylc.flow.unicode_rules.TaskMessageValidator
+                       :noindex:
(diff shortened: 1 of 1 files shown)


cylc/flow/cfgspec/workflow.py
```diff
@@ -1715,10 +1715,12 @@ def get_script_common_text(this: str, example: Optional[str] = None):
                     Custom outputs must satisfy these rules:
 
                     .. autoclass:: cylc.flow.unicode_rules.TaskOutputValidator
+                       :no-index:
```
wxtim, line 1718: I don't think it works with the hyphen.

The [documentation](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoclass) is confusing IMO - it looks like it should be `:no-index:` but a search for [`:noindex:`](https://www.sphinx-doc.org/en/master/search.html?q=%3Anoindex%3A) suggests otherwise.

MetRonnie: Ok, changed to `:noindex` and it still works on 7.2

oliver-sanders:
Branch and milestone don't match :(

PR #5712, opened 2023-08-30
Test the simulation mode code.

### LINKED ISSUE #5654: Test Simulation Mode
## Synopsis

Simulation Mode is currently largely untested. To ensure that https://github.com/cylc/cylc-flow/issues/5641 doesn't break existing interfaces I intend to write a thorough set of tests for similar existing functionality. This ticket is a way of keeping notes about what I need to test.

Although the target of the work in #5641  is a new feature, fixing testing for existing features should be counted as a bug-fix, hence the 8.2.1 milestone.

## Current state of play

There are functional tests in `tests/f/modes` for dummy mode, but not for simulation mode, at least on the surface. On closer insepection it looks like they rely on settings in `[runtime][namespace]simulation` which describes itself as 

> task configuration for workflow *simulation* and *dummy* run modes.

## What _should_ simulation mode do?

- [ ] Prevent a run directory being created for tasks.
- [ ] Take a time specified by:
  - [ ]  `[simulation]speedup factor` * `execution time limit` if both are set.
  - [ ]  else `[simulation]default run length`.
- [ ] Config item `time limit buffer` should not affect simulation mode (Docs claim that it is only used in dummy mode).
- [ ] Tasks should fail at cycle points specified by `[simulation]fail cycle points`.
- [ ] Tasks should fail on the first attempt, but not retries if `[simulation]fail try 1 only` set.
- [ ] Task event handers will only be called if `disable task event handlers` is set True.
- [ ]

wxtim:
Closes #5654 

Pre-requisite work for Skip mode 

- Refactor simulation mode functions.
- Improve documentation of sim mode settings.
- Ensure 100% test coverage of sim mode.

On an initial self-review I noticed that `configure_sim_modes` might as well be in the `simulation` module. I've moved it, but not squashed that commit down, because that way reviewers can see how I refactored it in a separate commit to the one I move it in.

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/641.


wxtim:
Changed to draft because a final sanity check/attempt to write an example for the docs showed up at least 2 exciting bugs.

changes.d/5712.fix.md
```diff
@@ -0,0 +1 @@
+Test the simulation mode code.
```
oliver-sanders, line 1: not a user facing change.

cylc/flow/task_job_mgr.py
```diff
@@ -1000,16 +1000,17 @@ def _simulation_submit_task_jobs(self, itasks, workflow):
             itask.waiting_on_job_prep = False
             itask.submit_num += 1
             self._set_retry_timers(itask)
-            itask.platform = {'name': 'SIMULATION'}
+            itask.platform['name'] = 'SIMULATION'
```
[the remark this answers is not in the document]
wxtim: We don't want the platform config in it's entirety replaced by `{'name': 'SIMULATION'}`. When we come around to retry with that config we have no info about the host &c to use. This is a bug!

cylc/flow/task_job_mgr.py
```diff
@@ -1000,16 +1000,17 @@ def _simulation_submit_task_jobs(self, itasks, workflow):
...
             itask.summary['job_runner_name'] = 'SIMULATION'
             itask.summary[self.KEY_EXECUTE_TIME_LIMIT] = (
-                itask.tdef.rtconfig['job']['simulated run length']
+                itask.tdef.rtconfig['simulation']['simulated run length']
             )
             itask.jobs.append(
                 self.get_simulation_job_conf(itask, workflow)
             )
+            itask.state.status = TASK_STATUS_PREPARING
```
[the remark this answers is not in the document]
wxtim: Otherwise it sticks in the waiting state on retry.

tests/functional/modes/03-simulation.t
```diff
@@ -0,0 +1,31 @@
...
+. "$(dirname "$0")/test_header"
+set_test_number 2
+
+install_workflow "${TEST_NAME_BASE}" "${TEST_NAME_BASE}"
+run_ok "${TEST_NAME_BASE}-validate" cylc validate "${WORKFLOW_NAME}"
+workflow_run_ok "${TEST_NAME_BASE}-run" \
+    cylc play \
+        --no-detach \
+        --mode=simulation \
+        --reference-test "${WORKFLOW_NAME}" 2> sim.log
```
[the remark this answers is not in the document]
wxtim: No - it's the stub of an extra test I decided was overkill. Will Remove.

tests/integration/test_simulation.py
```diff
@@ -0,0 +1,143 @@
...
+
+def test_false_if_not_running(sim_time_check_setup, monkeypatch):
+    schd, itasks, msg_q = sim_time_check_setup
+
+    # False if task status not running:
+    assert sim_time_check(msg_q, itasks) is False
+
+
+def test_sets_started_time(sim_time_check_setup, monkeypatch):
+    """But that the method still returns False"""
```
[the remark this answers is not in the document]
wxtim: I've refactored - the idea is to read the name of the test and the first line of the doc as complimentary clauses. 🤶🏼

cylc/flow/task_job_mgr.py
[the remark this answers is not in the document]
oliver-sanders: I think this was correct before.

Sure this is modifying the `localhost` platform that each new `TaskProxy` object is (incorrectly) assigned when it is initiated?

cylc/flow/task_job_mgr.py
```diff
@@ -1000,16 +1000,17 @@ def _simulation_submit_task_jobs(self, itasks, workflow):
             itask.waiting_on_job_prep = False
             itask.submit_num += 1
             self._set_retry_timers(itask)
-            itask.platform = {'name': 'SIMULATION'}
+            itask.platform['name'] = 'SIMULATION'
             itask.summary['job_runner_name'] = 'SIMULATION'
             itask.summary[self.KEY_EXECUTE_TIME_LIMIT] = (
-                itask.tdef.rtconfig['job']['simulated run length']
+                itask.tdef.rtconfig['simulation']['simulated run length']
```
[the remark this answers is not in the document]
oliver-sanders: I'm not sure this was wrong, because if it was then we would have got a `KerError` whenever this code was run, so `[job][simulated run length]` must have existed right?

cylc/flow/task_job_mgr.py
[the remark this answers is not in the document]
oliver-sanders: This wasn't needed before right?

This gets done somewhere else.

tests/integration/test_simulation.py
```diff
@@ -0,0 +1,143 @@
...
+
+class PrintLog():
+    def __init__(self):
+        self.old_records = []
+
+    def log(self, records):
+        for record in records:
+            if record not in self.old_records:
+                print(record.message)
+                self.old_records.append(record)
```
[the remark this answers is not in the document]
oliver-sanders: (unresolved, still appears to be unused)

cylc/flow/simulation.py
```diff
@@ -0,0 +1,220 @@
...
+    rtc: Dict[str, Any]
+) -> None:
+    """Force platform = localhost
+
+    Remove legacy sections [job] and [remote], which would conflict
+    with setting platforms.
+
+    TODO (Cylc 9): This function can probably be removed
+    when the items in FORBIDDEN_WITH_PLATFORM with platforms
+    are removed from the config - it will become trivially simple.
```
oliver-sanders, line 125:
```suggestion
    This can be removed when support for the FORBIDDEN_WITH_PLATFORM
    configurations is dropped.
```

cylc/flow/task_job_mgr.py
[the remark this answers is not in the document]
wxtim: Simulated run length isn't in the config spec. AFAICT Cylc just uses it as a convenient place to dump this value. We could make it part of an object on the scheduler if you want? Like `sim_mode_manager`

cylc/flow/task_job_mgr.py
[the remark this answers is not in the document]
wxtim: Before what? No, resubmit was broken!

cylc/flow/simulation.py
```diff
@@ -0,0 +1,220 @@
...
+        ).get_seconds()
+
+    return sleep_sec
+
+
+def build_simulation_script(rtc: Dict[str, Any], sleep_sec: int) -> str:
+    """Create fake scripting for simulation modes.
+
+    This is for Dummy mode only.
+    """
```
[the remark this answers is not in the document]
wxtim: Probably not great terminology - willfix.

cylc/flow/task_job_mgr.py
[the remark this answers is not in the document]
oliver-sanders: At the moment (master), I don't think that simulated jobs even go through the preparing state in the first place (which makes sense), we send them straight to submitted:

https://github.com/cylc/cylc-flow/blob/63ceffb21badc2d634a43241ae9a4da75e24189d/cylc/flow/task_job_mgr.py#L997-L1013

I don't think we should change this.

cylc/flow/task_job_mgr.py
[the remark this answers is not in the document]
oliver-sanders: My point is that this appears to have been working before (otherwise we would have got traceback right?) so I'm not sure why it has changed here?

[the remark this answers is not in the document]
wxtim: > Did you change the output of the config stuff to move this from job to simulation?

No!

cylc/flow/task_job_mgr.py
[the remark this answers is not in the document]
wxtim: It does however, need the platform default for `submission retry delays` added to this "platform" to prevent a failure in `_set_retry_timers`  on the retry.

cylc/flow/task_job_mgr.py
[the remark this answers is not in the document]
wxtim: At the moment on master the line

```python
            self.task_events_mgr.process_message(
                itask, INFO, TASK_OUTPUT_SUBMITTED
            )
```
just below this chunk converts a task in the waiting state to a task in the running state on the first pass, but doesn't on retry - so that sim mode hangs in waiting state. I'll find out more about why and get back to you.

cylc/flow/simulation.py
```diff
@@ -0,0 +1,219 @@
...
+    now = time()
+    for itask in itasks:
+        if itask.state.status != TASK_STATUS_RUNNING:
+            continue
+        # Started time is not set on restart
+        if itask.summary['started_time'] is None:
+            itask.summary['started_time'] = now
+        timeout = (
+            itask.summary['started_time'] +
+            itask.tdef.rtconfig['simulation']['simulated run length']
```
wxtim, line 179: Not a real config item, added to the `rtconfig['simulation']` dictionary at line 54.

cylc/flow/task_events_mgr.py
```diff
@@ -591,6 +592,16 @@ def process_message(
                 itask, severity, message, event_time, flag, submit_num):
             return None
 
+        severity = cast(int, LOG_LEVELS.get(severity, INFO))
```
wxtim, line 595: Moved out of `_process_message_check` because they only run if it is true.

cylc/flow/task_events_mgr.py
```diff
@@ -777,24 +787,23 @@ def _process_message_check(
             return False
 
         if (
-                itask.state(TASK_STATUS_WAITING)
-                and
+            itask.state(TASK_STATUS_WAITING)
+            and itask.tdef.run_mode == 'live'   # Polling in live mode only.
```
wxtim, line 791: The code within this section is designed to avoid problems with polling - but there will be no polling in simulation mode.

cylc/flow/task_job_mgr.py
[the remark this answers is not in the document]
wxtim: Closing this after in person convo with Oliver - taking a different approach to the resubmit problem.

oliver-sanders:
> codecov/patch Successful in 1s — 100.00% of diff hit (target 95.00%) 

:tada:

tests/functional/modes/03-simulation.t
```diff
@@ -0,0 +1,31 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test that simulation mode runs, and reruns a failed task successfully.
```
oliver-sanders, line 18:
```suggestion
# Test that simulation mode runs, and reruns a failed task successfully
# when execution retry delays is configured
```

oliver-sanders (approved):

MetRonnie (commented):
Partial review

cylc/flow/task_events_mgr.py
wxtim: It's not meant to change anything.

cylc/flow/simulation.py
```diff
@@ -0,0 +1,219 @@
...
+
+from queue import Queue
+from typing import Any, Dict, List, Optional, Union
+from time import time
+
+from cylc.flow.cycling import PointBase
+from cylc.flow.cycling.loader import get_point
+from cylc.flow.network.resolvers import TaskMsg
+from cylc.flow.platforms import FORBIDDEN_WITH_PLATFORM
+from cylc.flow.task_proxy import TaskProxy
```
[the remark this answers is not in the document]
wxtim: #WellThatsHorrid

cylc/flow/task_events_mgr.py
MetRonnie: What did you mean by this then:

> Moved out of `_process_message_check` because they only run if it is true.

wxtim: It only runs in `process_message_check` if that method is going to return true. Here it only runs after process message check returns true and not otherwise. I thought it was more helpful to understanding the flow of the program here.

wxtim: Turns out you were right - I shouldn't have moved this. I was wrong to thing that `event_time` in process_event == `timestamp` in process message check and a couple of tests were broken by this. Please hit resolve if you are satisfied.

tests/unit/test_config.py
```diff
@@ -1741,3 +1743,32 @@ def test_cylc_env_at_parsing(
...
+        'fail cycle points': '',
+    }
+    rtconfig_1 = {
+        'execution time limit': '',
+        'simulation': sim_section,
+        'job': job_section,
+        'outputs': {},
+    }
+    rtconfig_2 = deepcopy(rtconfig_1)
+    rtconfig_2['job'] = job_section
```
MetRonnie, line 1764: (Not needed)
```suggestion
```

tests/unit/test_simulation.py
```diff
@@ -0,0 +1,172 @@
...
+    result = build_dummy_script(rtc, 60)
+    expect = [
+        'sleep 60',
+        "cylc message '1'",
+        "cylc message '2'",
+        f"cylc__job__dummy_result {str(fail_one_time_only).lower()}"
+        " 1 || exit 1"
+    ]
+    for i, line in enumerate(result.split('\n')):
+        assert expect[i] == line
```
MetRonnie, line 78:
```suggestion
    assert result.split('\n') == [
        'sleep 60',
        "cylc message '1'",
        "cylc message '2'",
        f"cylc__job__dummy_result {str(fail_one_time_only).lower()}"
        " 1 || exit 1"
    ]
```
This is easier to debug IMO

tests/unit/test_simulation.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    disable_platforms(rtc)
+    assert rtc['platform'] == expect
+    subdicts = [v for v in rtc.values() if isinstance(v, dict)]
+    for val in [
+        v for subdict in subdicts
+        for k, v in subdict.items()
+        if k != 'platform'
+    ]:
+        assert val is None
```
MetRonnie, line 100:
```suggestion
    for subdict in subdicts:
        for k, val in subdict.items():
            if k != 'platform':
                assert val is None
```

cylc/flow/simulation.py
```diff
@@ -0,0 +1,219 @@
...
+    for section, keys in FORBIDDEN_WITH_PLATFORM.items():
+        if section in rtc:
+            for key in keys:
+                if key in rtc[section]:
+                    rtc[section][key] = None
+    rtc['platform'] = 'localhost'
+
+
+def parse_fail_cycle_points(
+    f_pts_orig: list
```
MetRonnie, line 135:
```suggestion
    f_pts_orig: List[str]
```

tests/unit/test_simulation.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def test_parse_fail_cycle_points(monkeypatch):
+    before = ['2']
+    monkeypatch.setattr(
+        'cylc.flow.cycling.loader.get_point_cls',
+        lambda cycling_type: IntegerPoint)
+    result = parse_fail_cycle_points(before)[0]
+    assert isinstance(result, IntegerPoint)
+    assert int(result) == 2
```
MetRonnie, line 110:
```suggestion
def test_parse_fail_cycle_points(set_cycling_type):
    before = ['2', '4']
    set_cycling_type()
    assert parse_fail_cycle_points(before) == [
        IntegerPoint(i) for i in before
    ]
```

wxtim:
@oliver-sanders  - accidentally re-tagged you for review. Happy for you to re-approve

MetRonnie (approved):

cylc/flow/task_job_mgr.py
```diff
@@ -1002,14 +1002,7 @@ def _simulation_submit_task_jobs(self, itasks, workflow):
...
 
-            # We largely want a clean "platform" and not localhost,
-            # but we need to get the existing submission retry delays
-            # to stop _set_retry_delays missing it on retry.
-            itask.platform = {
-                'name': 'SIMULATION',
-                'submission retry delays':
-                    itask.platform['submission retry delays']
-            }
+            itask.platform = {'name': 'SIMULATION'}
```
oliver-sanders, line 1004: I thought this was needed for submission retry delays?

MetRonnie:
While you're here Tim, could you check that restarting a workflow that has been running in simulation mode will restart it in simulation mode? (If get round to testing it I'll comment below)

cylc/flow/task_job_mgr.py
wxtim: https://github.com/cylc/cylc-flow/pull/5712#discussion_r1360622309

wxtim:
> check that restarting a workflow that has been running in simulation mode will restart it in simulation mode without having to specify it? 

Yes, if you restart you do not have to specify mode to get the same mode.

> Oliver said trying to restart in simulation mode when it had been running in live mode merely restarted in live mode.

When I tried this on master it looks like it restarts in simulation mode (on master & my branch). Ick.

That's a bug.

cylc/flow/task_job_mgr.py
oliver-sanders: You added `submission retry delays` here in an earlier commit, then removed them later.

oliver-sanders: See this: https://github.com/cylc/cylc-flow/pull/5712#discussion_r1322582113

wxtim: Are submission retry delays supposed to impact simulation mode?

No

    Can simulation tasks submit-fail?

No

Nor can I recreate the scenario where there was a problem. Have removed this.

hjoliver:
This has two approvals, but I'm not sure if all questions addressed yet? If so, let's merge it.

wxtim:
> This has two approvals, but I'm not sure if all questions addressed yet? If so, let's merge it.

Agreed, but I think @MetRonnie  or @oliver-sanders  should press the button here.

MetRonnie (approved):
Pending resolution of Oliver's comment

oliver-sanders (approved):

PR #5721, opened 2023-09-06
Feature: Sim mode at runtime

wxtim:
Recheck simulation mode settings at runtime if a broadcast might have changed them. If it has, update these settings.

A waypoint on the road to [skip mode](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-skip-mode.md).

~Based upon https://github.com/cylc/cylc-flow/pull/5712. Drafted until that is in.~

**Check List**


cylc/flow/simulation.py
```diff
@@ -0,0 +1,305 @@
...
+    this = dict_
+    for i in range(len(path)):
+        if isinstance(this[path[i]], dict):
+            this = this[path[i]]
+        else:
+            this[path[i]] = value
+    return dict_
+
+
+def update_nested_dict(rtc: NestedDict, dict_: NestedDict) -> None:
```
wxtim, line 221: This and the two functions above feel like I've replicated functionality from somewhere else (parsec?, OrderedDictWithDefaults?), but I cannot for the life of me work out how to use existing functions.

cylc/flow/simulation.py
```diff
@@ -0,0 +1,305 @@
...
+    TASK_STATUS_RUNNING,
+    TASK_STATUS_FAILED,
+    TASK_STATUS_SUCCEEDED,
+)
+from cylc.flow.wallclock import get_current_time_string
+
+from metomi.isodatetime.parsers import DurationParser
+
+
+# Exotic: Recursive Type hint.
```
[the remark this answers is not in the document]
wxtim: This is in theory self documenting, but it's so unusual that the comment seems useful.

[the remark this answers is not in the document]
wxtim: Is what I'm trying to do possible?

cylc/flow/simulation.py
```diff
@@ -37,59 +37,86 @@
...
+    rtc['submission retry delays'] = [1]
+
+    if dummy_mode:
         # Generate dummy scripting.
         rtc['init-script'] = ""
         rtc['env-script'] = ""
         rtc['pre-script'] = ""
         rtc['post-script'] = ""
         rtc['script'] = build_dummy_script(
             rtc, sleep_sec) if dummy_mode else ""
```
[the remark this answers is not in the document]
wxtim: Agreed. Happy to just do it now if agreed.

cylc/flow/simulation.py
```diff
@@ -37,59 +37,86 @@
...
+    else:
+        rtc['script'] = ""
```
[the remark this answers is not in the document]
wxtim: I don't see why it should be.

cylc/flow/simulation.py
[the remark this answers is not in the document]
wxtim: Actually, NVM - I don't need it!

cylc/flow/simulation.py
```diff
@@ -250,16 +195,22 @@ def sim_time_check(
...
             broadcast = broadcast_mgr.get_broadcast(itask.tokens)
             if broadcast:
-                update_nested_dict(
-                    itask.tdef.rtconfig, broadcast)
+                for config in SIMULATION_CONFIGS:
+                    if (
+                        config in broadcast
+                        and isinstance(broadcast[config], dict)
+                    ):
+                        itask.tdef.rtconfig[config].update(broadcast[config])
```
oliver-sanders, line 211: This looks like it's modifying the task defs runtime config object, this would persist broadcasts after they've expired or been cancelled.

wxtim: Yes. Looks like I might have some re-thinking to do.

oliver-sanders: Maybe take a look at how it's done for tasks run in live mode and see if we can use the same approach for simulation/dummy mode.

cylc/flow/simulation.py
```diff
@@ -166,46 +220,52 @@ def sim_time_check(
...
-    now = time()
+    sim_task_state_changed: bool = False
+    now: float = time()
     for itask in itasks:
         if itask.state.status != TASK_STATUS_RUNNING:
             continue
         # Started time is not set on restart
         if itask.summary['started_time'] is None:
             itask.summary['started_time'] = now
+        if itask.mode_settings is None:
```
wxtim, line 231: @MetRonnie

MetRonnie: I don't think Mypy supports type narrowing for dictionary values.
```diff
diff --git a/cylc/flow/simulation.py b/cylc/flow/simulation.py
index edc693825..b7f74823c 100644
--- a/cylc/flow/simulation.py
+++ b/cylc/flow/simulation.py
@@ -227,18 +227,12 @@ def sim_time_check(
             continue
         # Started time is not set on restart
-        if itask.summary['started_time'] is None:
-            itask.summary['started_time'] = now
+        started_time = itask.summary['started_time']
+        if started_time is None:
+            itask.summary['started_time'] = started_time = now
         if itask.mode_settings is None:
             itask.mode_settings = ModeSettings()
             itask.mode_settings.update(
                 itask, broadcast_mgr)
-        # Ugly hack to make MyPy Happy - It seems quite clear to
-        # me that this is ensured 5-6 lines up ↑
-        if TYPE_CHECKING:
-            assert itask.summary['started_time'] is not None
-        timeout = (
-            itask.summary['started_time'] +
-            itask.mode_settings.simulated_run_length
-        )
+        timeout = started_time + itask.mode_settings.simulated_run_length
         if now > timeout:
             job_d = itask.tokens.duplicate(job=str(itask.submit_num))
```

cylc/flow/simulation.py
```diff
@@ -37,6 +41,56 @@
...
+
+        Example:
+            >>> this = ModeSettings()
+            >>> print(this)
+            <flow.simulation.ModeSettings object at ...
+                simulated_run_length  : 0.0
+                sim_task_fails        : False
+        """
+        repr_ = [super().__repr__()]
+        maxwidth = max([len(i) for i in self.__slots__]) + 2
```
oliver-sanders, line 69:
```suggestion
        maxwidth = max(len(i) for i in self.__slots__) + 2
```

cylc/flow/simulation.py
```diff
@@ -46,23 +100,17 @@ def configure_sim_modes(taskdefs, sim_mode):
...
         rtc['submission retry delays'] = [1]
 
-        # Generate dummy scripting.
-        rtc['init-script'] = ""
-        rtc['env-script'] = ""
-        rtc['pre-script'] = ""
-        rtc['post-script'] = ""
-        rtc['script'] = build_dummy_script(
-            rtc, sleep_sec) if dummy_mode else ""
+        if dummy_mode:
```
oliver-sanders, line 106: Note, remember not to permit `[runtime][<namespace>]run mode = dummy`.

cylc/flow/simulation.py
```diff
@@ -91,6 +139,11 @@ def get_simulated_run_len(rtc: Dict[str, Any]) -> int:
             str(rtc['simulation']['default run length'])
```
oliver-sanders, line 139: s/get_simulated_run_len/set_simulated_run_len

cylc/flow/simulation.py
```diff
@@ -157,7 +210,8 @@ def parse_fail_cycle_points(
 
 
 def sim_time_check(
-    message_queue: 'Queue[TaskMsg]', itasks: 'List[TaskProxy]'
+    message_queue: 'Queue[TaskMsg]', itasks: 'List[TaskProxy]',
+    broadcast_mgr,
```
oliver-sanders, line 214: Sorry, mixed horizontal/vertical arrangement drives me nuts.

```suggestion
    message_queue: 'Queue[TaskMsg]',
    itasks: 'List[TaskProxy]',
    broadcast_mgr,
```

cylc/flow/simulation.py
```diff
@@ -37,6 +41,56 @@
     from cylc.flow.task_proxy import TaskProxy
 
 
+class ModeSettings:
```
oliver-sanders, line 44: Neat, but probably gives the same result as a `NamedTuple` or a `@dataclass`?

cylc/flow/simulation.py
```diff
@@ -166,46 +220,46 @@ def sim_time_check(
...
-            itask.tdef.rtconfig['simulation']['simulated run length']
-        )
+        started_time = itask.summary['started_time']
+        if started_time is None:
+            itask.summary['started_time'] = started_time = now
+        if itask.mode_settings is None:
+            itask.mode_settings = ModeSettings()
+            itask.mode_settings.update(
+                itask, broadcast_mgr)
+        timeout = started_time + itask.mode_settings.simulated_run_length
```
oliver-sanders, line 236: These values should get set the first time `sim_time_check` is called on a simulated job.

On each subsequent call the values are retrieved from the itask so should stay the same.

Worth considering setting these values during simulated job submission for clearer code.

cylc/flow/simulation.py
```diff
@@ -215,5 +269,5 @@ def sim_task_failed(
         sim_conf['fail cycle points'] is None  # i.e. "all"
         or point in sim_conf['fail cycle points']
     ) and (
-        try_num == 1 or not sim_conf['fail try 1 only']
+        submit_num == 0 or not sim_conf['fail try 1 only']
```
oliver-sanders, line 272: Behaviour change?

cylc/flow/task_job_mgr.py
```diff
@@ -997,14 +998,18 @@ def _set_retry_timers(
     def _simulation_submit_task_jobs(self, itasks, workflow):
         """Simulation mode task jobs submission."""
         for itask in itasks:
+            if not itask.mode_settings:
+                itask.mode_settings = ModeSettings()
+            itask.mode_settings.update(
+                itask, self.task_events_mgr.broadcast_mgr)
```
oliver-sanders, line 1004: This can either be set at simulation time or at sim_time_check time?

cylc/flow/task_proxy.py
```diff
@@ -178,6 +179,7 @@ class TaskProxy:
         'tokens',
         'try_timers',
         'waiting_on_job_prep',
+        'mode_settings',
```
oliver-sanders, line 182: Make sure you preserve this variable in the `copy_to_reload_successor` method, otherwise it will get wiped on reload.

cylc/flow/simulation.py
wxtim: I need to have a look at whether we have the memory reserving properties of `__slots__`, but I bet that you are right!

cylc/flow/simulation.py
wxtim: Yes. I'd not have done that deliberately.

cylc/flow/simulation.py
wxtim: I shouldn't have been using `try_num`, but `submit_num`, so yes. It's the correct behaviour though.

cylc/flow/simulation.py
wxtim: This method is no longer being called at runtime in this version - some of the methods it uses are called to recalculate values in the `mode_settings` object.

cylc/flow/simulation.py
wxtim: > This now sets the sim time

Where?

wxtim: Where does it set the sim_time?

cylc/flow/task_job_mgr.py
wxtim: I don't think they should be no, but I see that's what I've done. I'll assume that I'm wrong and remove them.

cylc/flow/simulation.py
wxtim: What happens if this is a restart?

cylc/flow/task_proxy.py
wxtim: Yes - That's a key thing that will make the code simpler.

cylc/flow/simulation.py
oliver-sanders: It sets the execution time limit a couple of lines down.

On the whole we would expect a `get_` method to return a value, but not to make changes to things.

cylc/flow/simulation.py
oliver-sanders: You tell me!

cylc/flow/simulation.py
wxtim: The `execution time limit` is not the simulation run time though. It's just a side effect.

MetRonnie:
Kicking tests

MetRonnie:
```
tests/integration/tui/test_updater.py::test_filters - TypeError: 'NoneType' object is not subscriptable
```
Looks genuine?

cylc/flow/rundb.py
```diff
@@ -251,6 +251,7 @@ class CylcWorkflowDAO:
             ["flow_nums"],
             ["is_manual_submit", {"datatype": "INTEGER"}],
             ["try_num", {"datatype": "INTEGER"}],
+            # This is used to store simulation task start time across restarts.
             ["time_submit"],
```
MetRonnie, line 255: ?

cylc/flow/simulation.py
```diff
@@ -166,46 +200,63 @@ def sim_time_check(
...
+    sim_task_state_changed: bool = False
     for itask in itasks:
-        if itask.state.status != TASK_STATUS_RUNNING:
+        if (
+            itask.state.status != TASK_STATUS_RUNNING
+            or any([
+                itask.state.is_queued,
+                itask.state.is_held,
+                itask.state.is_runahead
+            ])
```
MetRonnie, line 211: No real gain to using `any()` here, and disadvantage of a performance penalty
```suggestion
            or itask.state.is_queued
            or itask.state.is_held
            or itask.state.is_runahead
```

cylc/flow/simulation.py
```diff
@@ -76,7 +102,7 @@ def configure_sim_modes(taskdefs, sim_mode):
         )
 
 
-def get_simulated_run_len(rtc: Dict[str, Any]) -> int:
+def set_simulated_run_len(rtc: Dict[str, Any]) -> int:
     """Get simulated run time.
 
     rtc = run time config
```
MetRonnie, line 108: Docstring could be clearer

cylc/flow/simulation.py
```diff
@@ -157,7 +188,10 @@ def parse_fail_cycle_points(
 
 
 def sim_time_check(
-    message_queue: 'Queue[TaskMsg]', itasks: 'List[TaskProxy]'
+    message_queue: 'Queue[TaskMsg]',
+    itasks: 'List[TaskProxy]',
+    broadcast_mgr,
+    db_mgr,
```
MetRonnie, line 194: Type annotations? (seeing as this is already an annotated function)

cylc/flow/simulation.py
```diff
@@ -29,14 +34,41 @@
...
+            poverride(rtconfig, overrides, prepend=True)
+        else:
+            rtconfig = itask.tdef.rtconfig
+        self.simulated_run_length = (
+            set_simulated_run_len(rtconfig))
+        self.sim_task_fails = sim_task_failed(
+            rtconfig['simulation'],
+            itask.point,
+            itask.submit_num
+        )
```
MetRonnie, line 69: This could be in the `__init__()` method?

tests/functional/modes/04-simulation-runtime.t
```diff
@@ -0,0 +1,63 @@
...
+
+workflow_run_ok "${TEST_NAME_BASE}-unpause" \
+    cylc play "${WORKFLOW_NAME}"
+
+# Wait for the workflow to finish (it wasn't run in no-detach mode):
+poll_grep "INFO - DONE" "${SCHD_LOG}"
+
+# If we hadn't changed the speedup factor using broadcast
+# The workflow timeout would have been hit:
+grep_fail "WARNING - Orphaned tasks" "${SCHD_LOG}"
```
MetRonnie, line 60: This is possibly fragile. Is there a more direct way to test this?

cylc/flow/rundb.py
wxtim: Oliver and I identified that this is no longer used for whatever it may have been used for in the past - so we don't have to make db changes. However, since it's not obvious that this is now used for an entirely different function a comment seems warranted.

cylc/flow/simulation.py
wxtim: No.

We might wish to broadcast a change to the outcome (probably only for skip mode really) between simulated job submissions, in which case we want to modify rather than create anew.

cylc/flow/simulation.py
wxtim: How does "Calculate task run time in simulation mode from the simulation config"?

cylc/flow/simulation.py
MetRonnie: The 2 instances where it's used both do
```
itask.mode_settings = ModeSettings()
itask.mode_settings.update(...)
```

tests/functional/modes/04-simulation-runtime/flow.cylc
```diff
@@ -0,0 +1,20 @@
+[scheduler]
+    [[events]]
+        # workflow timeout = PT30S
+
```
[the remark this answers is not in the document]
wxtim: Thanks

cylc/flow/simulation.py
MetRonnie: Fair enough, but I believe it would still have the same behaviour if you replaced it with
```diff
-if not itask.mode_settings:
-    itask.mode_settings = ModeSettings()
-itask.mode_settings.update(
+itask.mode_settings = ModeSettings(
     itask, self.task_events_mgr.broadcast_mgr)
```

tests/integration/test_simulation.py
```diff
@@ -30,96 +30,260 @@ def get_msg_queue_item(queue, id_):
...
+    itask = schd.pool.get_task(
+        ISO8601Point(point), itask)
+
+    for result in results:
+        schd.task_job_mgr._simulation_submit_task_jobs(
+            [itask], schd.workflow)
+        assert itask.mode_settings.sim_task_fails is result
+
+
+async def test_sim_time_check_sets_started_time(
```
MetRonnie, line 127:
```suggestion
def test_sim_time_check_sets_started_time(
```

tests/integration/test_simulation.py
```diff
@@ -30,96 +30,260 @@ def get_msg_queue_item(queue, id_):
...
-    # Setup a task to fail
+    # Setup a task to fail, submit it.
     fail_all_1066 = schd.pool.get_task(ISO8601Point('1066'), 'fail_all')
     fail_all_1066.state.status = 'running'
-    fail_all_1066.try_timers = {'execution-retry': SimpleNamespace(num=0)}
+    fail_all_1066.state.is_queued = False
+    schd.task_job_mgr._simulation_submit_task_jobs(
+        [fail_all_1066], schd.workflow)
+
+    # For the purpose of the test delete the started time set but
```
MetRonnie, line 168: Is this meant to be?
```suggestion
    # For the purpose of the test delete the started time set by
```

cylc/flow/simulation.py
```diff
@@ -38,8 +38,10 @@
 
 if TYPE_CHECKING:
     from queue import Queue
+    from cylc.flow. mgr import BroadcastMgr
```
MetRonnie, line 41:
```suggestion
    from cylc.flow.broadcast_mgr import BroadcastMgr
```

cylc/flow/simulation.py
```diff
@@ -51,9 +53,7 @@ class ModeSettings:
     simulated_run_length: float = 0.0
     sim_task_fails: bool = False
 
-    def update(self, itask, broadcast_mgr):
-        """Update an instance of this object.
-        """
+    def __init__(self, itask, broadcast_mgr):
```
MetRonnie, line 56:
```suggestion
    def __init__(self, itask: 'TaskProxy', broadcast_mgr: 'BroadcastMgr'):
```

cylc/flow/simulation.py
```diff
@@ -103,12 +103,13 @@ def configure_sim_modes(taskdefs, sim_mode):
 
 
 def set_simulated_run_len(rtc: Dict[str, Any]) -> int:
-    """Get simulated run time.
+    """Calculate simulation run time from a task's config.
```
MetRonnie, line 106: This no longer sets anything

wxtim:
@oliver-sanders

cylc/flow/simulation.py
```diff
@@ -76,13 +102,14 @@ def configure_sim_modes(taskdefs, sim_mode):
         )
 
 
-def get_simulated_run_len(rtc: Dict[str, Any]) -> int:
-    """Get simulated run time.
+def set_simulated_run_len(rtc: Dict[str, Any]) -> int:
```
MetRonnie, line 105: I would change this back to `get_simulated_run_len` as there is no longer any change to the code

cylc/flow/simulation.py
```diff
@@ -166,46 +196,59 @@ def sim_time_check(
...
+                .parse(
+                    db_mgr.pub_dao.select_task_job(
+                        *itask.tokens.relative_id.split("/")
+                    )["time_submit"]
+                )
+                .seconds_since_unix_epoch
+            )
+            itask.summary['started_time'] = started_time
+        if itask.mode_settings is None:
+            itask.mode_settings = ModeSettings(itask, broadcast_mgr)
```
oliver-sanders, line 223: The `ModeSettings` gets set whenever it isn't set:

* This means it should persist for the life of a job which is correct.
  * E.G. a broadcast made whilst the simulated job is running won't effect the job
* This means it should persist for future submissions of the job which is incorrect.
  *  E.G. a broadcast made whilst a simulated job is running or after it has completed will not affect future submissions (i.e. auto-retries or manual triggers)

Could do with tests for reload/restart interactions to ensure behaviour is as expected.

Note: On restart, I think we re-submit the simulated job with the new config as we don't currently have the ability to "continue" it with the original config, this is fine.

cylc/flow/simulation.py
```diff
@@ -166,46 +196,59 @@ def sim_time_check(
...
+
+        timeout = started_time + itask.mode_settings.simulated_run_length
+        if time() > timeout:
```
oliver-sanders, line 226: Cache the value of `time()` to avoid repeated calls.

cylc/flow/simulation.py
```diff
@@ -166,46 +196,59 @@ def sim_time_check(
...
+                        *itask.tokens.relative_id.split("/")
+                    )["time_submit"]
+                )
+                .seconds_since_unix_epoch
+            )
+            itask.summary['started_time'] = started_time
+        if itask.mode_settings is None:
+            itask.mode_settings = ModeSettings(itask, broadcast_mgr)
+
+        timeout = started_time + itask.mode_settings.simulated_run_length
```
oliver-sanders, line 225: If we're using `RunMode` objects to cache values derived from the runtime config, we might as well cache this little bit of maths there to avoid re-computation with each loop.

cylc/flow/simulation.py
wxtim: > This means it should persist for future submissions of the job which is incorrect

Bother. Will fix.

> Could do with tests for reload/restart interactions to ensure behaviour is as expected.

_**YES**_

> Note: On restart, I think we re-submit the simulated job with the new config as we don't currently have the ability to "continue" it with the original config, this is fine.

I'm not convinced, will investigate.

cylc/flow/simulation.py
wxtim: You can cache more than a little bit.

cylc/flow/simulation.py
wxtim: > Note: On restart, I think we re-submit the simulated job with the new config as we don't currently have the ability to "continue" it with the original config, this is fine.

I was initially thinking about the fact that we use the database to get the start time back, but that's literally the only thing we retrieve, everything else is reloaded from config, so I think that you are correct, and that this needs to be a resubmission. 

Are we OK with it using bog-standard re-submission behavior where it increments the job number? This may cause confusion if users have set 'fail try 1 only' or were expecting the job to succeed first time. (I'll make sure I put in some explanatory Logging.)

Should we handle the abandoned submission?  Is there anything in the database/datastore which is likely to look confusingly incomplete from such a task?

oliver-sanders: * When you restart a workflow, any simulated jobs which were running at the time the workflow stopped should continue to run (i.e. they shouldn't fail as the result of the restart).
* Ideally they would continue with the config they started with (i.e. the simulated job should run for the length of the configured runtime it was submitted with), but it's not a biggie if they continue with the config the workflow was restarted with (call it a documented limitation).
* The submit number should not change for a job which has "continued" in this way (i.e. it should behave just like a background task would if it was left running whilst the workflow restarted).

I think this is how it's currently implemented (this branch).

wxtim: It is, and If you're happy with the limitation you've described in this. Having said which, what if we put the timeout into the database as the end time? Then we can get the timeout on restart. That just leaves the intended target task state.

wxtim: I think that https://github.com/cylc/cylc-flow/pull/5721/commits/35b637782ab6754153700d025cab078e5bb7a10d answers this issue, both in the change and the testing. Will wait for the CI before I un-draft. Will spin the conversation into a new ticket.

wxtim: https://github.com/cylc/cylc-flow/issues/5942

cylc/flow/simulation.py
oliver-sanders: I'm not sure submit_num is correct here?

cylc/flow/simulation.py
```diff
@@ -29,12 +34,61 @@
...
+            itask.point,
+            itask.submit_num
+        )
+
+        # itask.summary['started_time'] and mode_settings.timeout need
+        # repopulating from the DB on workflow restart:
+        started_time = itask.summary['started_time']
+        if started_time is None and db_mgr:
+            started_time = int(
+                TimePointParser()
```
oliver-sanders, line 81: Initialising the `TimePointParser` for every task is going to get expensive.

We carefully cache the instance elsewhere in the code. Maybe try using the `WorkflowSpecifics` cache?

cylc/flow/simulation.py
wxtim: Not sure. Putting it back as it was.

cylc/flow/simulation.py
wxtim: Turned out to be a bit trickier to fix this than expected, because I wasn't putting the time into the DB in a format which the `get_unix_time_from_time_string` was happy with.

cylc/flow/simulation.py
oliver-sanders: Tis not how it was.

cylc/flow/simulation.py
```diff
@@ -166,39 +213,44 @@ def sim_time_check(
...
     now = time()
+    sim_task_state_changed: bool = False
+
     for itask in itasks:
-        if itask.state.status != TASK_STATUS_RUNNING:
+        if (
+            itask.state.status != TASK_STATUS_RUNNING
+            or itask.state.is_queued
+            or itask.state.is_held
+            or itask.state.is_runahead
```
oliver-sanders, line 224: I don't think this is right? Why the change?

wxtim: It's not, I've removed it. 👍🏼

MetRonnie: Still not pushed?

cylc/flow/task_job_mgr.py
```diff
@@ -1020,8 +1020,12 @@ def _simulation_submit_task_jobs(self, itasks, workflow):
...
             )
             self.workflow_db_mgr.put_insert_task_jobs(
-                itask, {'time_submit': now_str})
-
+                itask, {
+                    'time_submit': now_str,
+                    'try_num': itask.get_try_num(),
+                }
+            )
+        self.workflow_db_mgr.process_queued_ops()
```
oliver-sanders, line 1028: This will get called automatically by the main loop.

```suggestion
```

tests/integration/test_simulation.py
```diff
@@ -28,98 +29,395 @@ def get_msg_queue_item(queue, id_):
...
+            raise UsageError(
+                'run_simjob requires either a task and point, or neither.')
+        else:
+            itasks = schd.pool.get_tasks()
+            if len(itasks) != 1:
+                raise UsageError(
+                    'run_simjob cannot needs a task and point if more '
+                    'than one task is in the task pool.')
+            else:
+                itask, = itasks
```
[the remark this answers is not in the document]
wxtim: Yes.

tests/functional/modes/04-simulation-runtime.t
```diff
@@ -0,0 +1,62 @@
...
+run_ok "second-task-broadcast-too-long" \
+    cylc broadcast "${WORKFLOW_NAME}" \
+        -n second_task \
+        -s 'execution time limit = P1D'
+
+# Test cancelling broadcast changes the task config back.
+run_ok "cancel-second-task-broadcast" \
+    cylc broadcast "${WORKFLOW_NAME}" \
+        -n second_task\
+        --clear
```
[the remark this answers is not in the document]
wxtim: Yes. Probably ought to check the log.

MetRonnie (commented):
Getting some odd behaviour on this branch with this workflow

```
[scheduler]
    allow implicit tasks = True
[scheduling]
    [[graph]]
        R1 = foo => bar
[runtime]
    [[foo]]
        [[[simulation]]]
            fail cycle points = all
            default run length = PT1S
```

When `foo` fails I trigger it again:

```
CRITICAL - [1/foo running job:01 flows:1] (received)failed
INFO - [1/foo running job:01 flows:1] => failed
WARNING - [1/foo failed job:01 flows:1] did not complete required outputs: ['succeeded']
ERROR - Incomplete tasks:
CRITICAL - Workflow stalled
WARNING - P3D stall timer starts NOW
INFO - [command] force_trigger_tasks
INFO - [1/foo failed job:01 flows:1] => waiting
INFO - Command actioned: force_trigger_tasks(['1/foo'], flow=['all'], flow_wait=False, flow_descr=None)
INFO - [1/foo waiting job:01 flows:1] => waiting(queued)
INFO - [1/foo waiting(queued) job:01 flows:1] => waiting
INFO - [1/foo waiting job:02 flows:1] => running
WARNING - stall timer stopped
WARNING - [1/foo running job:02 flows:1] (received-ignored)failed for job(01) != job(02)
WARNING - [1/foo running job:02 flows:1] (received-ignored)failed for job(01) != job(02)
WARNING - [1/foo running job:02 flows:1] (received-ignored)failed for job(01) != job(02)
WARNING - [1/foo running job:02 flows:1] (received-ignored)failed for job(01) != job(02)
WARNING - [1/foo running job:02 flows:1] (received-ignored)failed for job(01) != job(02)
WARNING - [1/foo running job:02 flows:1] (received-ignored)failed for job(01) != job(02)
WARNING - [1/foo running job:02 flows:1] (received-ignored)failed for job(01) != job(02)
...
```

Every 1 second it keeps printing that line.

Compared to the expected behaviour on master:

```
CRITICAL - [1/foo running job:01 flows:1] (received)failed
INFO - [1/foo running job:01 flows:1] => failed
WARNING - [1/foo failed job:01 flows:1] did not complete required outputs: ['succeeded']
ERROR - Incomplete tasks:
CRITICAL - Workflow stalled
WARNING - P3D stall timer starts NOW
INFO - [command] force_trigger_tasks
INFO - [1/foo failed job:01 flows:1] => waiting
INFO - Command actioned: force_trigger_tasks(['1/foo'], flow=['all'], flow_wait=False, flow_descr=None)
INFO - [1/foo waiting job:01 flows:1] => waiting(queued)
INFO - [1/foo waiting(queued) job:01 flows:1] => waiting
INFO - [1/foo waiting job:02 flows:1] => running
WARNING - stall timer stopped
CRITICAL - [1/foo running job:02 flows:1] (received)failed
INFO - [1/foo running job:02 flows:1] => failed
WARNING - [1/foo failed job:02 flows:1] did not complete required outputs: ['succeeded']
ERROR - Incomplete tasks:
CRITICAL - Workflow stalled
```

tests/functional/modes/05-sim-trigger.t
```diff
@@ -0,0 +1,44 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test that we can broadcast an alteration to simulation mode.
```
MetRonnie, line 18: Copy-paste needs updating?

MetRonnie (commented):
With my earlier example
```
[scheduler]
    allow implicit tasks = True
[scheduling]
    [[graph]]
        R1 = foo => bar
[runtime]
    [[foo]]
        [[[simulation]]]
            fail cycle points = all
            default run length = PT1S
```
When I play the workflow it stalls, I then do
```
$ cylc broadcast sim-mode -n foo -s '[simulation]fail try 1 only=True'
Broadcast set:
+ [*/foo] [simulation]fail try 1 only=True

$ cylc trigger sim-mode//1/foo
```
However the task fails again

tests/functional/modes/05-sim-trigger.t
wxtim:
```suggestion
# Test that we can re-trigger a task in sim mode
```

wxtim:
> However `foo` fails again

`fail try 1 only` != `fail submit 1 only`: It only plays with `execution retry delays`, not manual resubmits. 

I'm not sure if that's OK, but that's the current behaviour. I'm considering raising a "Question" issue.

MetRonnie:
That's OK, but the docs could do with that detail

image

cylc/flow/simulation.py
```diff
@@ -157,7 +211,9 @@ def parse_fail_cycle_points(
 
 
 def sim_time_check(
-    message_queue: 'Queue[TaskMsg]', itasks: 'List[TaskProxy]'
+    task_events_manager,  #: 'TaskEventsMgr',
```
MetRonnie, line 214: ?
```suggestion
    task_events_manager: 'TaskEventsMgr',
```

wxtim:
> That's OK, but the docs could do with that detail
> 
> image

Agreed - see the commit immediately above this comment. 😄

cylc/flow/cfgspec/workflow.py
```diff
@@ -1271,6 +1271,12 @@ def get_script_common_text(this: str, example: Optional[str] = None):
 
                     Task instances must be set to fail by
                     :cylc:conf:`[..]fail cycle points`.
+
+                    .. warning::
```
MetRonnie, line 1275: (Warning is a bit much I think)
```suggestion
                    .. note::
```

tests/flakyfunctional/cylc-get-config/04-dummy-mode-output.t
```diff
@@ -47,8 +47,8 @@ delete_db
 workflow_run_ok "${TEST_NAME_BASE}-run-simulation" \
     cylc play -m 'simulation' --reference-test --debug --no-detach "${WORKFLOW_NAME}"
 LOG="$(cylc log -m p "$WORKFLOW_NAME")"
-count_ok '(received)meet' "${LOG}" 1
-count_ok '(received)greet' "${LOG}" 1
+count_ok '(internal)meet' "${LOG}" 1
+count_ok '(internal)greet' "${LOG}" 1
 
-purge
+# purge
```
MetRonnie, line 53:
```suggestion
purge
```

MetRonnie (approved):

oliver-sanders (commented):
Some minor comments and a couple of issues:

**Restart traceback**

I tried running a workflow in simulation mode with Cylc 8.2. I then restarted it on this branch and got the following traceback:

```
CRITICAL - An uncaught error caused Cylc to shut down.
    If you think this was an issue in Cylc, please report the following traceback to the developers.
    https://github.com/cylc/cylc-flow/issues/new?assignees=&labels=bug&template=bug.md&title=;
2024-02-26T11:54:14Z ERROR - 'NoneType' object is not subscriptable
      File "/cylc-flow/cylc/flow/scheduler.py", line 654, in run_scheduler
        await self.main_loop()
      File "ylc-flow/cylc/flow/scheduler.py", line 1744, in main_loop
        and sim_time_check(
      File "cylc-flow/cylc/flow/simulation.py", line 232, in sim_time_check
        itask.mode_settings = ModeSettings(
      File "cylc-flow/cylc/flow/simulation.py", line 74, in __init__
        db_info["time_submit"])
    TypeError: 'NoneType' object is not subscriptable
2024-02-26T11:54:14Z CRITICAL - Workflow shutting down - 'NoneType' object is not subscriptable
```

**"fail cycle points" not broadcastable:**

In this example, the task "a" is configured to fail on the first cycle:

```ini
[scheduling]
    initial cycle point = 1
    final cycle point = 3
    cycling mode = integer
    [[graph]]
        P1 = """
            a:succeeded? => s
            a:failed? => f
            s | f => b
            b[-P1] => a?
        """

[runtime]
    [[root]]
        [[[simulation]]]
            default run length = PT5S
    [[a]]
        [[[simulation]]]
            fail cycle points = 1
    [[b,f,s]]
```

I ran this and waited for `1/a` to submit, then broadcasted a change to make `2/a` fail also:

```
cylc broadcast tmp.tqqU6eDW6W -n a -p 2 -s '[simulation]fail cycle points = 2'
```

But it didn't appear to take effect:

```
2024-02-26T12:44:35Z INFO - [1/a waiting job:01 flows:1] => running
2024-02-26T12:44:36Z INFO - [command] broadcast
2024-02-26T12:44:36Z INFO - Broadcast set:
    + [2/a] [simulation]fail cycle points=2
2024-02-26T12:44:40Z CRITICAL - [1/a running job:01 flows:1] (internal)failed at 2024-02-26T12:44:40Z
2024-02-26T12:44:40Z INFO - [1/a running job:01 flows:1] => failed
...
2024-02-26T12:44:52Z INFO - [2/a waiting(runahead) job:00 flows:1] => waiting
2024-02-26T12:44:53Z INFO - [2/a waiting job:01 flows:1] => running
2024-02-26T12:44:58Z INFO - [2/a running job:01 flows:1] => succeeded
...
2024-02-26T12:45:10Z INFO - Broadcast cancelled:
    - [2/a] [simulation]fail cycle points=2
```
Are the mode_settings are being created when the task is created (which would be immediately after `1/a` submits in this case) rather than when the task is submitted (which is after the broadcast)?

cylc/flow/task_job_mgr.py
```diff
@@ -996,23 +998,33 @@ def _set_retry_timers(
 
     def _simulation_submit_task_jobs(self, itasks, workflow):
         """Simulation mode task jobs submission."""
+        now = time()
+        now_str = get_time_string_from_unix_time(now)
         for itask in itasks:
+            itask.summary['started_time'] = now
+            self._set_retry_timers(itask, itask.tdef.rtconfig)
+            itask.mode_settings = ModeSettings(
+                itask, self.task_events_mgr.broadcast_mgr)
```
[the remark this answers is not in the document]
wxtim: But I'm using the presence of the db manager as part of the test for the workflow being restarted!

wxtim:
> Are the mode_settings are being created when the task is created (which would be immediately after 1/a submits in this case) rather than when the task is submitted (which is after the broadcast)?

No.

It turns out that this problem is caused by `sim_task_failed` expecting a type of `Union[PointBase, Literal["all"]]` for `sim_conf['fail cycle points']`

cylc/flow/task_job_mgr.py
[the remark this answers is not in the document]
wxtim: Not if the `mode_settings` object is being created by the first job submission as is happening here. I think you are right though, it'd make sense to have an explicit flag rather than using the presence of the db_mgr object as an implicit flag, which is what I'm doing.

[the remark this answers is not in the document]
wxtim: Yes.

cylc/flow/network/resolvers.py
```diff
@@ -767,7 +767,9 @@ def broadcast(
                 cycle_points, namespaces, settings)
         if mode == 'clear_broadcast':
             return self.schd.task_events_mgr.broadcast_mgr.clear_broadcast(
-                cycle_points, namespaces, settings)
+                point_strings=cycle_points,
+                namespaces=namespaces,
```
wxtim, line 771:
```suggestion
                namespaces=namespaces,
```

changes.d/5721.feat.md
```diff
@@ -0,0 +1 @@
+Allow users to broadcast simulation mode settings to tasks.
```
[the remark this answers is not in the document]
MetRonnie: I don't think that's correct?? Run mode applies to the scheduler, not tasks

cylc/flow/broadcast_mgr.py
```diff
@@ -83,7 +83,12 @@ def check_ext_triggers(self, itask, ext_trigger_queue):
         return self._match_ext_trigger(itask)
 
     def clear_broadcast(
-            self, point_strings=None, namespaces=None, cancel_settings=None):
+        self,
+        point_strings=None,
+        namespaces=None,
+        fail_points=None,
```
MetRonnie, line 89: I'm struggling to understand this handling of `fail_points`. Why isn't it handled in `cancel_settings`?

cylc/flow/broadcast_mgr.py
```diff
@@ -265,9 +274,23 @@ def put_broadcast(
         modified_settings = []
         bad_point_strings = []
         bad_namespaces = []
-
+        bad_fail_points = []
         with self.lock:
             for setting in settings:
+                fail_cycle_points = setting.get(
+                    'simulation', {'fail cycle points': None}).get(
+                        'fail cycle points', None)
```
MetRonnie, line 282:
```suggestion
                    'simulation', {}).get('fail cycle points', None)
```

cylc/flow/broadcast_mgr.py
```diff
@@ -359,6 +392,11 @@ def _point_string_in_prunes(prunes, point_string):
         """Is point_string pruned?"""
         return point_string in [prune[0] for prune in prunes]
 
+    @staticmethod
+    def _fail_points_in_prunes(prunes, fail_points):
+        """Is point_string pruned?"""
```
MetRonnie, line 397:
```suggestion
        """Is fail cycle points pruned?"""
```

changes.d/5721.feat.md
```diff
@@ -1 +1 @@
-Allow users to broadcast simulation mode settings to tasks.
\ No newline at end of file
+Allow task's run mode (i.e. "live", "simulation" or "dummy") to be changed dynamically at runtime using `cylc broadcast`.
```
MetRonnie: https://github.com/cylc/cylc-flow/pull/5721#discussion_r1507499124

> I don't think that's correct?? Run mode applies to the scheduler, not tasks

cylc/flow/simulation.py
```diff
@@ -16,25 +16,116 @@
...
+        db_mgr: 'WorkflowDatabaseManager',
+    ):
+
+        # itask.summary['started_time'] and mode_settings.timeout need
+        # repopulating from the DB on workflow restart:
+        started_time = itask.summary['started_time']
+        try_num = None
+        if started_time is None:
+            # Get DB info
+            db_info = db_mgr.pub_dao.select_task_job(
```
MetRonnie, line 90: Ooh, is this supposed to be the public database actually?

tests/integration/test_simulation.py
```diff
@@ -14,112 +14,416 @@
...
+        schd.task_events_mgr.broadcast_mgr.put_broadcast(
+            ['1066'], ['one'], [{
+                'simulation': {'fail cycle points': 'higadfuhasgiurguj'}
+            }])
+        schd.task_events_mgr.broadcast_mgr.put_broadcast(
+            ['1066'], ['one'], [{
+                'simulation': {'fail cycle points': '1'}
+            }])
+        assert {'fail cycle points': []} == (
+            schd.broadcast_mgr.broadcasts['1066']['one']['simulation'])
```
MetRonnie, line 412: Yoda expression
```suggestion
        assert schd.broadcast_mgr.broadcasts['1066']['one'][
            'simulation'
        ] == {'fail cycle points': []}
```

changes.d/5721.feat.md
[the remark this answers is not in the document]
wxtim: In the past this has been true, but the point of this work is to make Cylc ready for Skip mode, and that requires the run mode to be applied on a task-by-task basis.

I think we should probably add the word "settings" - because these definately are changed on a task by task basis.

changes.d/5721.feat.md
```diff
@@ -0,0 +1 @@
...
```
wxtim, line 1:
```suggestion
Allow task's run mode settings to be changed dynamically at runtime using `cylc broadcast`.
```

cylc/flow/simulation.py
wxtim: I'm sorry, I don't understand what you are asking.

tests/integration/test_simulation.py
wxtim: > Yoda expression

And I suddenly understood what you meant!

cylc/flow/broadcast_mgr.py
wxtim: Because I honestly don't understand how this all fits together with the calling of the mutation by the CLI, and I'm not sure how to do this. I think `cancel_keys_list` is there for the cli rather than internal interfaces.

cylc/flow/simulation.py
MetRonnie: > Do you think I _should_ be consulting the private database?

Yes I would have thought so

changes.d/5721.feat.md
```diff
@@ -0,0 +1 @@
+Allow task's run mode settings to be changed dynamically at runtime using `cylc broadcast`.
```
MetRonnie, line 1: > ...the point of this work is to make Cylc ready for Skip mode, and that requires the run mode to be applied on a task-by-task basis. Having said which, simulation mode will continue to be a whole workflow setting, at least by default, so you're probably right.
> 
> I think we should probably add the word "settings" - because these definately are changed on a task by task basis.

_https://github.com/cylc/cylc-flow/pull/5721#discussion_r1510912272_

But this PR only enables changing simulation mode settings, right?
```suggestion
Allow tasks' simulation mode settings to be changed dynamically at runtime using `cylc broadcast`.
```

cylc/flow/simulation.py
oliver-sanders: Yes, this should use the private DB, good spot.

changes.d/5721.feat.md
oliver-sanders: Yes, the next, hopefully smaller step, is to allow the run mode to be broadcasted at runtime.

cylc/flow/simulation.py
wxtim:
```suggestion
            db_info = db_mgr.pri_dao.select_task_job(
```

cylc/flow/broadcast_mgr.py
oliver-sanders: We shouldn't require any changes to the broadcast code for simulation mode.

Try casting the cycle point in the simulation code.

tests/flakyfunctional/cylc-get-config/04-dummy-mode-output.t
```diff
@@ -47,8 +47,8 @@ delete_db
 workflow_run_ok "${TEST_NAME_BASE}-run-simulation" \
     cylc play -m 'simulation' --reference-test --debug --no-detach "${WORKFLOW_NAME}"
 LOG="$(cylc log -m p "$WORKFLOW_NAME")"
-count_ok '(received)meet' "${LOG}" 1
-count_ok '(received)greet' "${LOG}" 1
+count_ok '(internal)meet' "${LOG}" 1
```
oliver-sanders, line 50: ???

wxtim: NVM. Will revert. Easy to fix.

cylc/flow/broadcast_mgr.py
wxtim: > We shouldn't require any changes to the broadcast code for simulation mode.

Seems reasonable - I've removed it, just one test failure as a result.

> Try casting the cycle point in the simulation code rather than the broadcast code.

Sorry, not sure I understand

> The simulation code should not need to call the clear_broadcast interface?

What should the sim mode code do if the Fail CP is garbage.

wxtim: I've dealt with this after a conversation on teams with @oliver-sanders . :)

MetRonnie (commented):
With this
```
[runtime]
    [[foo]]
        execution retry delays = 100*PT10S
        [[[simulation]]]
            fail cycle points = all
            default run length = PT10S
            fail try 1 only = False
```

I tried setting `fail try 1 only=True` and got this
```
INFO - Broadcast set:
    + [20240201T0000Z/foo] [simulation]fail try 1 only=True
INFO - [20240201T0000Z/foo waiting(queued) job:15 flows:1] => waiting
CRITICAL - An uncaught error caused Cylc to shut down.
    If you think this was an issue in Cylc, please report the following traceback to the developers.
    https://github.com/cylc/cylc-flow/issues/new?assignees=&labels=bug&template=bug.md&title=;
ERROR - argument of type 'NoneType' is not iterable
      File "~/cylc-flow/cylc/flow/scheduler.py", line 654, in run_scheduler
        await self.main_loop()
      File "~/cylc-flow/cylc/flow/scheduler.py", line 1745, in main_loop
        self.release_queued_tasks()
      File "~/cylc-flow/cylc/flow/scheduler.py", line 1496, in release_queued_tasks
      File "~/cylc-flow/cylc/flow/task_job_mgr.py", line 266, in submit_task_jobs
        return self._simulation_submit_task_jobs(itasks, workflow)
      File "~/cylc-flow/cylc/flow/task_job_mgr.py", line 1006, in _simulation_submit_task_jobs
        itask.mode_settings = ModeSettings(
      File "~/cylc-flow/cylc/flow/simulation.py", line 118, in __init__
        ] = parse_fail_cycle_points(
      File "~/cylc-flow/cylc/flow/simulation.py", line 249, in parse_fail_cycle_points
        if 'all' in f_pts_orig:
    TypeError: argument of type 'NoneType' is not iterable
CRITICAL - Workflow shutting down - argument of type 'NoneType' is not iterable
```

cylc/flow/simulation.py
```diff
@@ -16,25 +16,129 @@
...
+
+            try:
+                rtconfig["simulation"][
+                    "fail cycle points"
+                ] = parse_fail_cycle_points(
+                    rtconfig["simulation"]["fail cycle points"]
+                )
+            except PointParsingError as exc:
+                # Broadcast Fail CP didn't parse
+                LOG.error(
```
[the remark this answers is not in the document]
MetRonnie: I think it should be fine because integer points are valid datetime points too. I think the cycle point coercers already rely on this principle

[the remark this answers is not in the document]
oliver-sanders: Presumably not *all* integer points are valid datetime points and most datetime points are not valid integer points?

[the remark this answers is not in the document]
MetRonnie: Ah yeah I was being dim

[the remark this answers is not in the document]
wxtim: I can probably make the warning pop up only once, but I'm not sure that's a reasonable thing to do. If you have a live, but invalid broadcast which Cylc is attempting bootlessly to apply to every task then it probably _should_ warn every time.

wxtim:
> ```
> 
> ```
> 
> 
>     
>   
> 
> I tried setting `fail try 1 only=True` after `foo` had already retried a few times and got this

Think that I've fixed this.

cylc/flow/simulation.py
```diff
@@ -145,19 +244,23 @@ def parse_fail_cycle_points(
...
-    if 'all' in f_pts_orig:
+    f_pts: 'Optional[List[PointBase]]' = []
+    if f_pts_orig and 'all' in f_pts_orig:
         f_pts = None
-    else:
+    elif f_pts_orig:
         f_pts = []
         for point_str in f_pts_orig:
             f_pts.append(get_point(point_str).standardise())
     return f_pts
```
MetRonnie, line 257:
```diff
diff --git a/cylc/flow/simulation.py b/cylc/flow/simulation.py
index 2c733f705..fc80640d6 100644
--- a/cylc/flow/simulation.py
+++ b/cylc/flow/simulation.py
@@ -232,5 +232,5 @@ def disable_platforms(
 
 def parse_fail_cycle_points(
-    f_pts_orig: List[str]
+    f_pts_orig: Optional[List[str]]
 ) -> 'Union[None, List[PointBase]]':
     """Parse `[simulation][fail cycle points]`.
@@ -248,12 +248,9 @@ def parse_fail_cycle_points(
         []
     """
-    f_pts: 'Optional[List[PointBase]]' = []
-    if f_pts_orig and 'all' in f_pts_orig:
-        f_pts = None
-    elif f_pts_orig:
-        f_pts = []
-        for point_str in f_pts_orig:
-            f_pts.append(get_point(point_str).standardise())
-    return f_pts
+    if not f_pts_orig:
+        return []
+    if 'all' in f_pts_orig:
+        return None
+    return [get_point(point_str).standardise() for point_str in f_pts_orig]
```

MetRonnie (commented):
```
[runtime]
    [[foo]]
        execution retry delays = 100*PT10S
        [[[simulation]]]
            fail cycle points = all
            default run length = PT10S
            fail try 1 only = False
```

Broadcasting `[simulation]default run length=PT1S` causes the task to succeed 😕

wxtim:
> Broadcasting `[simulation]default run length=PT1S` causes the task to succeed 😕

Got it - side effect of earlier change.

MetRonnie:
- Broadcasting `execution retry delays` does not cause a change when running in simulation mode
- Also, in the Edit Runtime form, `submission retry delays` shows as `1` for some reason even though it is not set in `flow.cylc`

wxtim:
> * Broadcasting `execution retry delays` does not cause a change when running in simulation mode

I cannot replicate this on the command line. Do you mean when operating from the GUI (looking at the context of your other comment)?

cylc/flow/simulation.py
```diff
@@ -46,23 +150,17 @@ def configure_sim_modes(taskdefs, sim_mode):
...
         rtc = tdef.rtconfig
-        sleep_sec = get_simulated_run_len(rtc)
-
-        rtc['execution time limit'] = (
-            sleep_sec + DurationParser().parse(str(
-                rtc['simulation']['time limit buffer'])).get_seconds()
-        )
 
-        rtc['simulation']['simulated run length'] = sleep_sec
         rtc['submission retry delays'] = [1]
```
wxtim, line 154: @MetRonnie  Here is the cause of the value in the GUI's submission retry delays box. I will think about whether this is required. But it's a change from master, not a fault in my branch...

MetRonnie:
> > * Broadcasting `execution retry delays` does not cause a change when running in simulation mode
> 
> I cannot replicate this on the command line. Given the context of your comment I tried it from the gui and still couldn't replicate.
> 
> In fact some of the tests I've written should fail if that were the case.
> 
> Can you come up with a concrete example?

```
[scheduler]
    allow implicit tasks = True
[scheduling]
    initial cycle point = 2024
    runahead limit = P0
    [[graph]]
        P1M = foo => bar

[runtime]
    [[foo]]
        execution retry delays = 100*PT10S
        script = false
        [[[simulation]]]
            fail cycle points = all
            default run length = PT10S
            fail try 1 only = False
```
```
cylc broadcast sim-mode -n foo -s 'execution retry delays=20*PT1S'
```
But it still keeps saying
```
WARNING - [20240101T0000Z/foo waiting job:06 flows:1] retrying in PT10S
```

wxtim:
@MetRonnie - This is a good spot. I think that the broadcasts probably need to be applied in the `TaskJobManager._simulation_submit_task_jobs` method to ensure that we are broadcasting all task settings, not just those in `[simulation]`. I think you might also have found problems clearing broadcasts with things as they were.

tests/integration/test_simulation.py
```diff
@@ -14,112 +14,431 @@
...
+        # Broadcast tasks will reparse correctly:
+        schd.broadcast_mgr.put_broadcast(
+            ['1066'], ['one'], [{
+                'simulation': {'fail cycle points': '1945, 1977, 1066'},
+                'execution retry delays': '5*PT2S'
+            }])
+        schd.task_job_mgr._simulation_submit_task_jobs(
+            [itask], schd.workflow)
+        assert itask.mode_settings.sim_task_fails is True
+        assert itask.tdef.rtconfig['execution retry delays'] == [5.0, 5.0]
```
MetRonnie, line 444: Shouldn't this be?
```suggestion
        assert itask.tdef.rtconfig['execution retry delays'] == [2, 2, 2, 2, 2]
```

cylc/flow/simulation.py
```diff
@@ -16,25 +16,118 @@
...
+    """
+    simulated_run_length: float = 0.0
+    sim_task_fails: bool = False
+    timeout: float = 0.0
+
+    def __init__(
+        self,
+        itask: 'TaskProxy',
+        db_mgr: 'WorkflowDatabaseManager',
+        rtconfig
```
MetRonnie, line 78:
```suggestion
        rtconfig: Dict[str, Any]
```

tests/integration/test_simulation.py
wxtim: Yes. :(

MetRonnie (approved):

cylc/flow/simulation.py
```diff
@@ -166,38 +261,42 @@ def sim_time_check(
...
+                    flag=task_events_manager.FLAG_RECEIVED
                 )
+            # Simulate message outputs.
+            for msg in itask.tdef.rtconfig['outputs'].values():
+                task_events_manager.process_message(
+                    itask, 'DEBUG', msg,
+                    flag=task_events_manager.FLAG_RECEIVED
+                )
+
+            # We've finished this psuedojob, so delete all the mode settings.
```
hjoliver, line 298:
```suggestion
            # We've finished this pseudo job, so delete all the mode settings.
```

changes.d/5721.feat.md
```diff
@@ -0,0 +1 @@
+Allow tasks' simulation mode settings to be changed dynamically at runtime using `cylc broadcast`.
```
hjoliver, line 1:
```suggestion
Allow task simulation mode settings to be changed dynamically using `cylc broadcast`.
```

wxtim:
> Conflict resolution looks good

Thank you @hjoliver

MetRonnie:
Genuine test failure presumably: `tests/f/modes/05-sim-trigger.t`

hjoliver:
I'm looking at the broken test -it's because of the `cylc set` merge.

hjoliver (approved):

PR #5727, opened 2023-09-11
Cylc to ignore PYTHONPATH: Use CYLC_PYTHONPATH for custom Python for Cylc

### LINKED ISSUE #5124: wrapper script: consider clearing PYTHONPATH
We advise that Cylc is installed into an isolated environment (e.g. conda / virtualenv). Unfortunately the `PYTHONPATH` environment variable can corrupt these.

Illustrative example (less extreme cases can cause problems too):

```bash
export PYTHONPATH='/path/to/python3.6/site-packages'
cylc version
Traceback ... not quite so isolated after all
```

This is very inconvenient for users who rely on `PYTHONPATH` (e.g. to load their own Python environment) as they have to unload this environment before they can run Cylc commands. If Cylc is installed in an isolated environment, then it should be isolated from stuff like this.

We might be able to do something like this:

```
PYTHONPATH= exec "${CYLC_HOME}/bin/${0##*/}" "$@"
```

Otherwise, there is the nuclear [Python] option:

```
-E     : ignore PYTHON* environment variables (such as PYTHONPATH)
```

Probably a good idea? But it does have the unwelcome side-effect of disabling `PYTHONBREAKPOINT`.

And also:

```
-s     : don't add user site directory to sys.path; also PYTHONNOUSERSITE
```

Could do with a bit of investigation to work out what the consequences of these approaches are.

**Pull requests welcome!**

wxtim:
Closes #5124 

> * Remove from `sys.path` anything in `PYTHONPATH`.
>   
>   * This prevents `PYTHONPATH` from breaking Cylc.
> * Add to `sys.path` anything in `CYLC_PYTHONPATH`.
>   
>   * This allows us to continue to install Cylc plugins via environment variable.

**Check List**

- [x] Tests this change needs to be manually tested alongside https://github.com/metomi/rose/pull/2725 . I've tested it with this workflow: [tmp.pythonpath.tar.gz](https://github.com/cylc/cylc-flow/files/12575203/tmp.pythonpath.tar.gz)
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/665.

### DIFF
--- a/changes.d/5727.break.md
+++ b/changes.d/5727.break.md
@@ -1,2 +1,1 @@
-Cylc now ignores `PYTHONPATH` to make it more robust to task environments which set this value.
-If you want to add to the Cylc environment itself, e.g. to install a Cylc extension, use `CYLC_PYTHONPATH`.
+Cylc now ignores `PYTHONPATH` to make it more robust to task environments which set this value. If you want to add to the Cylc environment itself, e.g. to install a Cylc extension, use `CYLC_PYTHONPATH`.
--- a/changes.d/5727.break.md
+++ b/changes.d/5727.break.md
@@ -1,0 +1,2 @@
+Cylc now ignores `PYTHONPATH` to make it more robust to task environments which set this value.
+If you want to add to the Cylc environment itself, e.g. to install a Cylc extension, use `CYLC_PYTHONPATH`.
--- a/cylc/flow/etc/tutorial/cylc-forecasting-workflow/etc/python-job.settings
+++ b/cylc/flow/etc/tutorial/cylc-forecasting-workflow/etc/python-job.settings
@@ -8,5 +8,4 @@
             # These environment variables ensure that tasks can
             # run in the same environment as the workflow:
             {% from "sys" import executable %}
-            PYTHONPATH = {{executable}}/../lib/python:$PYTHONPATH
             PATH = $(dirname {{executable}}):$PATH
--- a/cylc/flow/etc/tutorial/cylc-forecasting-workflow/etc/python-job.settings
+++ b/cylc/flow/etc/tutorial/cylc-forecasting-workflow/etc/python-job.settings
@@ -7,6 +7,6 @@
         [[[environment]]]
             # These environment variables ensure that tasks can
             # run in the same environment as the workflow:
-            {% from "sys" import path, executable %}
-            PYTHONPATH = {{':'.join(path)}}
+            {% from "sys" import executable %}
+            PYTHONPATH = {{executable}}/../lib/python:$PYTHONPATH
             PATH = $(dirname {{executable}}):$PATH
--- a/cylc/flow/scripts/cylc.py
+++ b/cylc/flow/scripts/cylc.py
@@ -16,10 +16,35 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """cylc main entry point"""
 
-import argparse
-from contextlib import contextmanager
 import os
 import sys
+
+
+def pythonpath_manip():
+    """Stop PYTHONPATH contaminating the Cylc Environment
+
+    * Remove PYTHONPATH items from sys.path to prevent PYTHONPATH
+      contaminating the Cylc Environment.
+    * Re-add items from CYLC_PYTHONPATH to sys.path.
+
+    See Also:
+        https://github.com/cylc/cylc-flow/issues/5124
+    """
+    if 'CYLC_PYTHONPATH' in os.environ:
+        for item in os.environ['CYLC_PYTHONPATH'].split(os.pathsep):
+            abspath = os.path.abspath(item)
+            sys.path.insert(0, abspath)
+    elif 'PYTHONPATH' in os.environ:
+        for item in os.environ['PYTHONPATH'].split(os.pathsep):
+            abspath = os.path.abspath(item)
+            if abspath in sys.path:
+                sys.path.remove(abspath)
+
+
+pythonpath_manip()
+
+import argparse
+from contextlib import contextmanager
 from typing import Iterator, NoReturn, Optional, Tuple
 
 from ansimarkup import parse as cparse
--- a/cylc/flow/scripts/cylc.py
+++ b/cylc/flow/scripts/cylc.py
@@ -25,7 +25,7 @@ def pythonpath_manip():
 
     * Remove PYTHONPATH items from sys.path to prevent PYTHONPATH
       contaminating the Cylc Environment.
-    * Re-add items from CYLC_PYTHONPATH to sys.path.
+    * Add items from CYLC_PYTHONPATH to sys.path.
 
     See Also:
         https://github.com/cylc/cylc-flow/issues/5124
@@ -34,7 +34,7 @@ def pythonpath_manip():
         for item in os.environ['CYLC_PYTHONPATH'].split(os.pathsep):
             abspath = os.path.abspath(item)
             sys.path.insert(0, abspath)
-    elif 'PYTHONPATH' in os.environ:
+    if 'PYTHONPATH' in os.environ:
         for item in os.environ['PYTHONPATH'].split(os.pathsep):
             abspath = os.path.abspath(item)
             if abspath in sys.path:
--- a/tests/unit/scripts/test_cylc.py
+++ b/tests/unit/scripts/test_cylc.py
@@ -15,14 +15,16 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
 import pkg_resources
+import sys
 from types import SimpleNamespace
 from typing import Callable
 from unittest.mock import Mock
 
 import pytest
 
-from cylc.flow.scripts.cylc import iter_commands
+from cylc.flow.scripts.cylc import iter_commands, pythonpath_manip
 
 from ..conftest import MonkeyMock
 
@@ -136,3 +138,25 @@ def test_execute_cmd(
     # the "bad" entry point should raise an exception
     with pytest.raises(ModuleNotFoundError):
         execute_cmd('bad')
+
+
+def test_pythonpath_manip(monkeypatch):
+    """pythonpath_manip removes items in PYTHONPATH from sys.path
+
+    and adds items from CYLC_PYTHONPATH
+    """
+    # If PYTHONPATH is set...
+    monkeypatch.setenv('PYTHONPATH', '/remove-from-sys.path')
+    monkeypatch.setattr('sys.path', ['/leave-alone', '/remove-from-sys.path'])
+    pythonpath_manip()
+    # ... we don't change PYTHONPATH
+    assert os.environ['PYTHONPATH'] == '/remove-from-sys.path'
+    # ... but we do remove PYTHONPATH items from sys.path, and don't remove
+    # items there not in PYTHONPATH
+    assert sys.path == ['/leave-alone']
+
+    # If CYLC_PYTHONPATH is set we retrieve its contents and
+    # add them to the sys.path:
+    monkeypatch.setenv('CYLC_PYTHONPATH', '/add-to-sys.path')
+    pythonpath_manip()
+    assert sys.path == ['/add-to-sys.path', '/leave-alone']


cylc/flow/etc/cylc
```diff
@@ -96,6 +96,12 @@ CYLC_HOME_ROOT="${CYLC_HOME_ROOT:-/opt}"
 # export ROSE_SITE_CONF_PATH="${ROSE_SITE_CONF_PATH:-/etc/rose}"
 ###############################################################################
 
+# Prevent user's PyhtonPath interfering with Cylc:
+if [[ -n "${PYTHONPATH}" ]]; then
```
dpmatthews, line 100: It's more complicated than this. We need to know what we are running.
I think we need:
- Cylc 8: unset PYTHONPATH (no need to save but no harm if we do)
- Rose 2: unset PYTHONPATH and save in _PYTHONPATH
- Cylc 7: do nothing (might be OK to unset but safer not to)
- Rose 2019: do nothing (would break Rose apps if we unset)

dpmatthews (dismissed):

cylc/flow/etc/cylc
dpmatthews: Not sure if this is sufficient (moved further down the script):

`if [[ -n "${PYTHONPATH}" && ${CYLC_ENV_NAME:-} =~ ^cylc-8 ]]; then`

wxtim: Thanks @dpmatthews : The list of conditions is probably what I needed to get further.

wxtim:
I've put this back to draft - I have something else to try!

oliver-sanders:
LGTM; the tests are failing as a few `PYTHONPATH` variables will need to be updated to `CYLC_PYTHONPATH` in the test code / workflows.

Manual testing is ok for this, but a basic automated test is possible, put this code in a function, then monkeypatch `sys.path`.

```
import sys

def patch_pythonpath():
    ...

def test_patch_pythonpath(monkeypatch):
     monkeypatch.setattr('sys.path', [])
     monkeypatch.setenv('PYTHONPATH', ...)
     monkeypatch.setenv('CYLC_PYTHONPATH', ...)
     patch_pythonpath()
     assert sys.path == [...]
```

wxtim:
> Manual testing is ok for this, but a basic automated test is possible, put this code in a function, then monkeypatch sys.path.

I had considered this - I'll do it.

Existing test failures seem to be caused by the print statements I left in the new code.

wxtim:
@hjoliver Please confirm that you are happy with this approach. 😄

cylc/flow/etc/tutorial/cylc-forecasting-workflow/etc/python-job.settings
```diff
@@ -8,5 +8,5 @@
             # These environment variables ensure that tasks can
             # run in the same environment as the workflow:
             {% from "sys" import path, executable %}
-            PYTHONPATH = {{':'.join(path)}}
+            CYLC_PYTHONPATH = {{':'.join(path)}}
```
oliver-sanders, line 11: I don't think this is right. The lib/python dir is for use by task scripts, not Cylc itself.

oliver-sanders:
Hmmm:

```
ModuleNotFoundError: No module named 'argparse'
```

https://github.com/cylc/cylc-flow/actions/runs/6315139062/job/17147002345?pr=5727#step:7:19

This is in a task's job.err section where it was raised from `cylc/flow/scripts/cylc.py`, so presumably a `cylc message` call from the job script.

I think this suggests that the logic is removing the standard library from `sys.path` internally?

The previous diff (replacing `PYTHONPATH` with `CYLC_PYTHONPATH`) may have suppressed this by bypassing the removal logic?

cylc/flow/etc/tutorial/cylc-forecasting-workflow/etc/python-job.settings
oliver-sanders: This line was removed, then added back in again. I can see that this gets the tests passing, but I don't think it makes sense so may be highlighting a valid bug.

The `CYLC_PYTHONPATH` variable should only impact the Cylc application, but this is the environment section for a job. If `CYLC_PYTHONPATH` is required here, then `module load` commands won't work (because they set `PYTHONPATH`).

wxtim: This file deliberately describes itself as an anti-pattern - we're going out of our way to use the Cylc environment, not trying to avoid it.

oliver-sanders (approved):
Here's how I tested:

```console
$ # create a virtual environment with an incompatible version of psutils
$ virtualenv venv
$ . venv/bin/activate
$ pip install psutils==4
```

```console
$ # replicate the bug
$ conda activate cylc.dev
$ PYTHONPATH=venv/lib/python3.9/site-packages/ cylc vip generic -N
...
2023-10-16T12:02:07Z ERROR - Workflow shutting down - PluginError: Error in plugin cylc.main_loop.auto_restart
    VersionConflict: (psutil 4.4.2 (/var/tmp/tmp.f8Khmjq2Dp/create/lib/python3.9/site-packages),
    Requirement.parse('psutil>=5.6.0'))
$ # verify the fix
$ git -C ~/cylc-flow checkout wxtim/feature.cache_pythonpath
$ PYTHONPATH=venv/lib/python3.9/site-packages/ cylc vip generic -N
...
...
$ # all good!
```

oliver-sanders:
We could do with an entry in the references/changes section of cylc-doc explaining the problem and introducing the CYLC_PYTHONPATH workaround.

changes.d/5727.feat.md
```diff
@@ -0,0 +1,2 @@
+Cylc now ignores PYTHONPATH. If you want to provide custom Python
+to Cylc use CYLC_PYTHONPATH.
```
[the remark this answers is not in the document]
MetRonnie: (which means renaming to `*.break.md`)

cylc/flow/etc/tutorial/cylc-forecasting-workflow/etc/python-job.settings
```diff
@@ -7,6 +7,6 @@
         [[[environment]]]
             # These environment variables ensure that tasks can
             # run in the same environment as the workflow:
-            {% from "sys" import path, executable %}
-            PYTHONPATH = {{':'.join(path)}}
+            {% from "sys" import executable %}
+            PYTHONPATH = {{executable}}/../lib/python:$PYTHONPATH
```
MetRonnie, line 11: Are you sure this works?

```
~/.conda/envs/cylc8/bin/python3.10/../lib: Not a directory
```

MetRonnie (commented):
Does this close #5124?

cylc/flow/etc/tutorial/cylc-forecasting-workflow/etc/python-job.settings
oliver-sanders: Try it out! Note, it may behave differently in your interactive shell.

MetRonnie: I'm not sure setting `PYTHONPATH` is needed? The workflow works for me without it (it seems only the setting of `PATH` below is needed)

MetRonnie (approved):

MetRonnie:
> Will need docs too, but should run this by @hjoliver first,

Has this happened yet?

hjoliver:
> > Will need docs too, but should run this by @hjoliver first,
> 
> Has this happened yet?

Apologies, happening now!

cylc/flow/scripts/cylc.py
```diff
@@ -16,10 +16,36 @@
...
+    * Re-add items from CYLC_PYTHONPATH to sys.path.
+
+    See Also:
+        https://github.com/cylc/cylc-flow/issues/5124
+    """
+    if 'CYLC_PYTHONPATH' in os.environ:
+        for item in os.environ['CYLC_PYTHONPATH'].split(os.pathsep):
+            abspath = os.path.abspath(item)
+            sys.path.insert(0, abspath)
+    elif 'PYTHONPATH' in os.environ:
```
hjoliver, line 37: Shouldn't we be removing PYTHONPATH items regardless of whether or not CYLC_PYTHONPATH is defined?

```suggestion
    if 'PYTHONPATH' in os.environ:
```

hjoliver (changes requested):
Agree with the approach. One question.

wxtim:
> > Will need docs too, but should run this by @hjoliver first,
> 
> Has this happened yet?

There's a Cylc Docs PR open...

cylc/flow/scripts/cylc.py
```diff
@@ -34,7 +34,7 @@ def pythonpath_manip():
         for item in os.environ['CYLC_PYTHONPATH'].split(os.pathsep):
             abspath = os.path.abspath(item)
             sys.path.insert(0, abspath)
-    elif 'PYTHONPATH' in os.environ:
+    if 'PYTHONPATH' in os.environ:
         os.environ['CYLC_PYTHONPATH'] = os.environ['PYTHONPATH']
```
MetRonnie, line 38: This now clobbers `CYLC_PYTHONPATH` if it was already set

wxtim:
```suggestion
    elif 'PYTHONPATH' in os.environ:
```

cylc/flow/scripts/cylc.py
```diff
@@ -16,10 +16,36 @@
...
+
+    See Also:
+        https://github.com/cylc/cylc-flow/issues/5124
+    """
+    if 'CYLC_PYTHONPATH' in os.environ:
+        for item in os.environ['CYLC_PYTHONPATH'].split(os.pathsep):
+            abspath = os.path.abspath(item)
+            sys.path.insert(0, abspath)
+    if 'PYTHONPATH' in os.environ:
+        os.environ['CYLC_PYTHONPATH'] = os.environ['PYTHONPATH']
```
wxtim, line 38:
```suggestion
        if 'CYLC_PYTHONPATH' in os.environ:
            os.environ['CYLC_PYTHONPATH'] = ':'.join([
                os.environ['PYTHONPATH'], os.environ['CYLC_PYTHONPATH']
            ])
        else:
            os.environ['CYLC_PYTHONPATH'] = os.environ['PYTHONPATH']
```

wxtim: @MetRonnie  - Does this make sense?

MetRonnie: Why is this needed at all? Are we not ignoring `PYTHONPATH` now?

wxtim: Yes, I think that the original elif is fine.

@hjoliver  - see this>

cylc/flow/scripts/cylc.py
```diff
@@ -16,10 +16,36 @@
...
+
+    See Also:
+        https://github.com/cylc/cylc-flow/issues/5124
+    """
+    if 'CYLC_PYTHONPATH' in os.environ:
+        for item in os.environ['CYLC_PYTHONPATH'].split(os.pathsep):
+            abspath = os.path.abspath(item)
+            sys.path.insert(0, abspath)
+    elif 'PYTHONPATH' in os.environ:
+        os.environ['CYLC_PYTHONPATH'] = os.environ['PYTHONPATH']
```
MetRonnie, line 38: I think this is what should be happening?
```suggestion
    if 'PYTHONPATH' in os.environ:
```

MetRonnie: Why?

cylc/flow/scripts/cylc.py
hjoliver: Not sure if I got my wires crossed this morning, or if the code has changed so I can't see what I thought I was querying anymore.  Sorry if the former!

cylc/flow/scripts/cylc.py
hjoliver: I thought it was supposed to be this:
- if PYTHONPATH exists, it might interfere with the Cylc environment, so nuke it
- legit Cylc-specific libs have to be put in CYLC_PYTHONPATH by users instead

Is that right?  If so, this seems wrong:
> No, we definitely want to save the PYTHONPATH as CYLC_PYTHONPATH.

hjoliver: i.e.,:
- if PYTHONPATH exists, delete it
- if CYLC_PYTHONPATH exists, put its elements in `sys.path`

(2 ifs, no elsif -- this is what I thought I was suggesting above, in my review comment).

cylc/flow/scripts/cylc.py
```diff
@@ -16,10 +16,35 @@
...
+    * Re-add items from CYLC_PYTHONPATH to sys.path.
+
+    See Also:
+        https://github.com/cylc/cylc-flow/issues/5124
+    """
+    if 'CYLC_PYTHONPATH' in os.environ:
+        for item in os.environ['CYLC_PYTHONPATH'].split(os.pathsep):
+            abspath = os.path.abspath(item)
+            sys.path.insert(0, abspath)
+    if 'PYTHONPATH' in os.environ:
```
wxtim, line 37: I'm trying this - I can't remember why I did this any other way - I think that there may be a reason - I'll investigate after I've had lunch.

MetRonnie:
There is no specification for this behaviour written in the PR which is making it difficult to understand what is supposed to be happening

wxtim:
@hjoliver , and

> There is no specification for this behaviour written in the PR which is making it difficult to understand what is supposed to be happening

Fair Point:
I think this is what is making me somewhat uncertain. I'll ask @oliver-sanders  to check that the following is the aim:

1. If the user modifies `PYTHONPATH` do not allow these items to interfere with Cylc by removing them from `sys.path`. Leave the environment `PYTHONPATH` unchanged.
2. If `CYLC_PYTHONPATH` is set add items in that variable to `sys.path` for Cylc to use.

What I am now unclear on is whether I want to save the original `PYTHONPATH` to `CYLC_PYTHONPATH` - I don't thing I do?

oliver-sanders:
Yes.

Very simple:

* Remove from `sys.path` anything in `PYTHONPATH`.
* Add to `sys.path` anything in `CYLC_PYTHONPATH`.

cylc/flow/scripts/cylc.py
```diff
@@ -16,10 +16,35 @@
...
 import os
 import sys
+
+
+def pythonpath_manip():
+    """Stop PYTHONPATH contaminating the Cylc Environment
+
+    * Remove PYTHONPATH items from sys.path to prevent PYTHONPATH
+      contaminating the Cylc Environment.
+    * Re-add items from CYLC_PYTHONPATH to sys.path.
```
hjoliver, line 28:
```suggestion
    * Add items from CYLC_PYTHONPATH to sys.path.
```

cylc/flow/scripts/cylc.py
hjoliver: @wxtim - I think this is still wrong, and by the 👍 @oliver-sanders seems to agree. It should be:
- if PYTHONPATH exists, remove its items from sys.path
- if CYLC_PYTHONPATH exists, add its items to sys.path

(not elif).

(And maybe do it in that order too - remove before add - for a tiny efficiency gain)

hjoliver (changes requested):

changes.d/5727.break.md
```diff
@@ -0,0 +1,2 @@
+Cylc now ignores `PYTHONPATH` to make it more robust to task environments which set this value.
+If you want to add to the Cylc environment itself, e.g. to install a Cylc extension, use `CYLC_PYTHONPATH`.
```
MetRonnie, line 2: Either

```suggestion
Cylc now ignores `PYTHONPATH` to make it more robust to task environments which set this value. If you want to add to the Cylc environment itself, e.g. to install a Cylc extension, use `CYLC_PYTHONPATH`.
```

or 

```suggestion
Cylc now ignores `PYTHONPATH` to make it more robust to task environments which set this value.
- If you want to add to the Cylc environment itself, e.g. to install a Cylc extension, use `CYLC_PYTHONPATH`.
```

cylc/flow/etc/tutorial/cylc-forecasting-workflow/etc/python-job.settings
MetRonnie: Why is this back?

MetRonnie (approved):

hjoliver (approved):
🎉

Issue #5716, opened 2023-09-04
tui: filter bug


oliver-sanders:
## Update

Filtering by task states does not work properly. Sometimes tasks are omitted, sometimes empty families/cycles are left in.

The approach Tui is using attempts to filter out cycles/families/tasks by state in the GraphQL query. This makes sense for tasks, but the approach will never work correctly for cycles/families (as they use an aggregate state representing a number of tasks [and or nested families]).

## OP

Sometimes, when you filter by task state in Tui, you don't see complete results.

It would appear that this is caused by missing "root" families in the data store, demonstrated by this diff:

```diff
diff --git a/cylc/flow/network/resolvers.py b/cylc/flow/network/resolvers.py
index b6effe143..319851e75 100644
--- a/cylc/flow/network/resolvers.py
+++ b/cylc/flow/network/resolvers.py
@@ -183,7 +183,10 @@ def get_state_from_selectors(tokens):
 
 def node_ids_filter(tokens, state, items) -> bool:
     """Match id arguments with node attributes."""
-    return any(
+    print()
+    for item in items:
+        print(tokens, item)
+    ret = any(
         (
             # don't match an empty string (globs should be implicit)
             not item.is_null
@@ -195,6 +198,7 @@ def node_ids_filter(tokens, state, items) -> bool:
             # match namespace name
             and (
                 not item['task']
+                # or item['task'] == 'root'
                 or fnmatchcase(tokens['task'], item['task'])
             )
             # match job
@@ -213,6 +217,8 @@ def node_ids_filter(tokens, state, items) -> bool:
         )
         for item in uniq(items)
     )
+    print('$$$', ret)
+    return ret
 
 
 def node_filter(node, node_type, args, state):
@@ -239,6 +245,7 @@ def node_filter(node, node_type, args, state):
     else:
         # live objects can be represented by a universal ID
         tokens = Tokens(node.id)
+
     return (
         (
             (

```

Which when run against this query:

```graphql
query($taskStates:[String]) {
  workflows(ids:["*"]) {
    test1: familyProxies(ids: ["*"], states: $taskStates) {
      id
      state
    }
  }
}
```

Produces this result:

```json
{
  "data": {
    "workflows": [
      {
        "test1": [
          {
            "id": "~osanders/generic/run1//20191218T1200Z/BAR",
            "state": "succeeded"
          }
        ]
      }
    ]
  }
}
```

With this stdout:

```
[...]

PR #5731, opened 2023-09-14
Tui 1.0

### LINKED ISSUE #3464: tui: list all suites
Follow-on from #3463
Attempt after: #3615 

Good low-priority rainy-day/jetlag activity.

Hook up Tui to "scan" results so that we can see all workflows (both stopped and running).

The idea is that the tree will look something like this:

* workflows
  * foo
      * task1
   * bar
      * task2

When a user expands a workflow we start requesting full data from the workflow. When the user collapses a workflow we request only the state totals.

**Pull requests welcome!**

### LINKED ISSUE #3527: tui: improve performance

### LINKED ISSUE #3530: tui: testing
Use html output from `cylc tui` to write regression tests.

**Pull requests welcome!**

oliver-sanders:
Remove the "Tui is experimental" notice :tada:

* Move the data update into another process: Closes #3527
  * Tui used to contact workflows in the same process it used to run the application.
  * This meant that the app was held in lockstep with the update. If the update took longer than the refresh time it would fall over.
  * This PR puts the updater into another process to break this lockstep so that the application and updater refresh rates can differ.
* Add multi-workflow capability: Closes #3464
  * Use the `scan` pipe to list workflows in Tui the same way we use it in the CylcUIServer.
  * The workflow refresh rate is different to the scan refresh rate.
  * Each workflow node has one child under it, the "spring" node. When you expand this "spring" node, Tui will connect to the workflow and will replace the "spring" node with the workflow tree once the data arrives.
  * When you collapse the workflow node, Tui will disconnect from the workflow.
  * You can expand multiple workflows.
* Add visual regression testing framework: Closes #3530
  * Add a test fixture which launches Tui in a "headless" mode where it's not rendering to a terminal.
  * The fixture has a method to provide user input allowing you to press keys as if there was a terminal running.
  * The fixture has a method for generating HTML screenshots which are then used for visual regression.
  * I.E. if anything changes in the output HTML, then test will fail.
  * This allows us to test navigation, filtering, help screens and more.
  * Tui coverage is now up to ~80%.
* Add some new functionalities.
  * Log view: support for `cylc cat-log`.
  * Show view: support for `cylc show`.

https://github.com/cylc/cylc-flow/assets/16705946/ff7d491d-9df4-4f66-9cd5-935d8d080c13

In the process I spotted an issue with task filtering which looks like it might be a data store thing but requires more investigation, lower priority as filtering is not an issue for the GUI. Until then the task filtering won't work right: https://github.com/cylc/cylc-flow/issues/5716

TODO:

* [x] Niceify the code.
* [x] Complete restart/reconnect test.
* [x] Do something better with the Tui header (which is now blank).
* [x] Investigate test flakyness (workflow states test quite flaky).

Tui integration test failures result in HTML files like this being written into the test directory in `~cylc-run/`. When tests fail in CI, these HTML files will be bundled into the artefact so can be inspected locally to determine the cause of the failure:

Screenshot from 2023-09-26 13-17-55

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at [cylc/cylc-doc/pull/663.](https://github.com/cylc/cylc-doc/pull/663)


tests/integration/tui/screenshots/test_navigation.cursor-at-bottom-of-screen.html
```diff
@@ -0,0 +1,31 @@
+<pre><span style="color:#000000;background:#e5e5e5">                                                                                </span>
```
oliver-sanders, line 1: I was really hoping GitHub would render this HTML which would make it much easier to review screenshot updates in PRs. Any ideas?

hjoliver:
Nice, looking forward to this!

tests/integration/tui/screenshots/test_navigation.cursor-at-bottom-of-screen.html
oliver-sanders: The best thing I can come up with is to use some Python lib to render them into png derived files.

cylc/flow/scripts/tui.py
```diff
@@ -15,34 +15,37 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-"""cylc tui WORKFLOW
+"""cylc tui [WORKFLOW]
```
oliver-sanders, line 18: If you give a workflow, tui will startup, pre-filtered to just that workflow. Press "E" to unset this filtering an view all workflows.

cylc/flow/scripts/tui.py
```diff
@@ -55,57 +58,36 @@
...
     parser = COP(
         __doc__,
-        argdoc=[WORKFLOW_ID_ARG_DOC],
+        argdoc=[OPT_WORKFLOW_ID_ARG_DOC],
         # auto_add=False,  NOTE: at present auto_add can not be turned off
         color=False
     )
 
-    parser.add_option(
-        '--display',
```
oliver-sanders, line 64: Removed the HTML testing mode from the Tui script as this is now handled by an integration test fixture.

cylc/flow/tui/app.py
```diff
@@ -206,16 +246,25 @@ class TuiApp:
 
     """
 
-    UPDATE_INTERVAL = 1
-    CLIENT_TIMEOUT = 1
+    # the UI update interval
+    # NOTE: this is different from the data update interval
+    UPDATE_INTERVAL = 0.1
```
oliver-sanders, line 251: This means that the UI will look for updates every 0.1 seconds, but the frequency of these updates is determined by the updater which is decoupled and runs to its own clock.

cylc/flow/tui/data.py
```diff
@@ -154,23 +155,20 @@ def cli_cmd(*cmd):
         raise ClientError(f'Error in command {" ".join(cmd)}\n{err}')
 
 
-def _clean(workflow):
-    # for now we will exit tui when the workflow is cleaned
-    # this will change when tui supports multiple workflows
-    cli_cmd('clean', workflow)
-    sys.exit(0)
```
oliver-sanders, line 161: Tui used to exit when you cleaned the workflow, but now it can stay open.

cylc/flow/tui/updater.py
```diff
@@ -0,0 +1,311 @@
...
+class Updater():
+    """The bit of Tui which provides the data.
+
+    It lists workflows using the "scan" interface, and provides detail using
+    the "GraphQL" interface.
+
+    """
+
+    # the maximum time to wait for a workflow update
+    CLIENT_TIMEOUT = 2
```
oliver-sanders, line 92: If the workflow takes more than 2 seconds to send its response, then it will time out. Tui will put a message under the workflow saying that it has timed out, but will retry with the next update cycle.

Previously both the application and updater used the same update period of 1 second. If it took longer to get the response and do the required processing, then Tui would crash.

cylc/flow/tui/updater.py
```diff
@@ -0,0 +1,311 @@
...
+
+    # the interval between workflow listing scans
+    BASE_SCAN_INTERVAL = 20
+
+    # the interval between workflow data updates
+    BASE_UPDATE_INTERVAL = 1
```
oliver-sanders, line 98: This is the desired update interval. If it takes longer than 1 second to update the workflow, no problem, the updates just come through slower, no more crashes.

oliver-sanders:
With Tui now better tested, this brings the project coverage up above 90%!

We still can get more of Tui covered now there's a testing framework for it.

dwsutherland:
Amazing work 🚀

oliver-sanders:
(added some extra tests to fill some coverage holes)

wxtim:
Sorry, I think that I found a bug on pressing `p` in the main view I get this error (sorry, my terminal isn't letting me copy/paste it:

image

oliver-sanders:
Good spot, the tests cover toggling workflow states using `W` and filling out the form, but not using the shortcuts. Will fix.

wxtim:
And from `user>stop-all`

```
Cylc Tui   tasks filtered (F - edit, R - re┌────────────────────────────────────────────────┐
                                           │                                                │
- ~tim                                     │  Error in command cylc stop *                  │
                                           │  ClientError: Failed to find user's private    │
                                           │  key, so cannot connect.                       │
                                           │  Task exception was never retrieved            │
                                           │  future: <Task finished name='Task-124'        │
                                           │  coro=<run() done, defined at /home/tim/tims_  │
                                           │  code/metomi/cylc-flow/cylc/flow/scripts/stop  │
                                      ┌────│  .py:206> exception=ClientError("Failed to     │
                                      │  id│  find user's private key, so cannot            │
                                      │    │  connect.")>                                   │
                                      │  Ac│  Traceback (most recent call last):            │
                                      │  < │    File "/home/tim/tims_code/metomi/cylc-flow  │
                                      │    │  /cylc/flow/network/__init__.py", line 237,    │
                                      │  < │  in _socket_connect                            │
                                      │    │      client_public_key, client_priv_key =      │
                                      │    │  zmq.auth.load_certificate(                    │
                                      │    │                                                │
                                      │    │  ^^^^^^^^^^^^^^^^^^^^^^^^^^                    │
                                      │    │    File "/home/tim/mambaforge/envs/cylc.dev/l  │
                                      │    │  ib/python3.11/site-packages/zmq/auth/certs.p  │
                                      │    │  y", line 105, in load_certificate             │
                                      │    │      raise OSError(f"Invalid certificate       │
                                      │    │  file: {filename}")                            │
                                      │    │  OSError: Invalid certificate file: /home/tim  │
                                      │    │  /cylc-run/flows-hold/run16/.service/client.k  │
                                      │ q t│  ey_secret                                     │
                                      └────│                                                │
                                           │  During handling of the above exception,       │
                                           │  another exception occurred:                   │
                                           │                                                │
                                           │  Traceback (most recent call last):            │
                                           │    File "/home/tim/tims_code/metomi/cylc-flow  │
                                           │  /cylc/flow/scripts/stop.py", line 223, in     │
                                           │                                                │
quit: q  help: h  context: enter  tree: - ←│ q to close                                     │T f s r R  filter workflows: W E p
                                           └────────────────────────────────────────────────┘

```

wxtim:
Would `clean-all-stopped` be unreasonable to ask for?

oliver-sanders:
> And from user>stop-all

I think that's correct behaviour. The `clean` command failed, so Tui put the stderr in a box for you.

oliver-sanders:
> Would clean-all-stopped be unreasonable to ask for?

Not unreasonable, I think that's just `cylc clean '*:stopped'` or since clean auto filters out running workflows, just `cylc clean '*'`.

It might be considered a little dangerous having a button like that kicking about, so we might want to put an "are you sure" check on that which Tui doesn't currently have an interface for.

oliver-sanders:
Fixed the "p" shortcut and added it to the tests: https://github.com/cylc/cylc-flow/compare/8e371542150b9792a8593399cefc14f96be5c1df..54b7e609af8676c7fcb3b4e8723a50d92ad9c3a7

oliver-sanders:
Fixed an issue with "E" not working properly: https://github.com/cylc/cylc-flow/compare/54b7e609af8676c7fcb3b4e8723a50d92ad9c3a7..6f3f1701c1106ecd779b11b08ad4293186678da0

cylc/flow/scripts/tui.py
```diff
@@ -55,57 +57,36 @@
...
-
     return parser
 
 
-@cli_function(get_option_parser)
-def main(_, options: 'Values', workflow_id: str) -> None:
-    workflow_id, *_ = parse_id(
-        workflow_id,
-        constraint='workflows',
+def configure_screenshot(v_term_size):
```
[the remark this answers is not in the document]
oliver-sanders: I'll move it in with the test code.

wxtim (commented):
- [x] Functional testing of PR.
- [x] Read function code.
- [x] Read the testing.

tests/integration/tui/conftest.py
```diff
@@ -0,0 +1,266 @@
...
+from cylc.flow.id import Tokens
+from cylc.flow.tui.app import TuiApp
+from cylc.flow.scripts.tui import configure_screenshot
+from cylc.flow.tui.overlay import _get_display_id
+
+
+SCREENSHOT_DIR = Path(__file__).parent / 'screenshots'
+
+
+def format_test_failure(expected, got, description):
```
wxtim, line 20: Can we have a docstring? Why are we turning the diff into an HTML doc?

tests/integration/tui/screenshots/test_auto_expansion.later-time.html
```diff
@@ -0,0 +1,21 @@
+<pre><span style="color:#ff0000;background:#e5e5e5;font-weight:bold">C</span><span style="color:#ffff00;background:#e5e5e5;font-weight:bold">y</span><span style="color:#00ff00;background:#e5e5e5;font-weight:bold">l</span><span style="color:#5c5cff;background:#e5e5e5;font-weight:bold">c</span><span style="color:#000000;background:#e5e5e5;font-weight:bold"> Tui</span><span style="color:#7f7f7f;background:#e5e5e5">   workflows filtered (</span><span style="color:#7f7f7f;background:#e5e5e5;font-weight:bold">W</span><span style="color:#7f7f7f;background:#e5e5e5"> - edit, </span><span style="color:#7f7f7f;background:#e5e5e5;font-weight:bold">E</span><span style="color:#7f7f7f;background:#e5e5e5"> - reset)</span><span style="color:#000000;background:#e5e5e5">                             </span>
```
wxtim, line 1: Are you OK with this offest?

Untitled


(it appears in lots of your images)

tests/integration/tui/test_app.py
```diff
@@ -0,0 +1,378 @@
...
+        )
+
+        # "enter" again should close it via the "cancel" button
+        rk.user_input('enter')
+        rk.compare_screenshot(
+            'test-raikura',
+            'the context menu should have closed',
+        )
+
+        # "ctrl d" should exit Tui
```
wxtim, line 83: Ought you test `ctrl c` and `ctrl z` cases too?

wxtim (approved):

tests/integration/tui/screenshots/test_auto_expansion.later-time.html
oliver-sanders: The reason for this is the way your browser handles the queued modifier.

In XTerm, the modifier applies to the task icon with a character width of 1. In FireFox the modifier comes before the task icon with a character width of 2.

tests/integration/tui/test_app.py
oliver-sanders: `ctrl z` can't be reasonably tested (you need a genuine interactive terminal).

`ctrl c` doesn't yield the exception.

oliver-sanders:
Training day came around again, added support for `cat-log` and `show`.

@wxtim, could you check the last three commits.

cylc/flow/data_store_mgr.py
```diff
@@ -1380,7 +1380,7 @@ def insert_job(self, name, cycle_point, status, job_conf):
             name=tproxy.name,
             cycle_point=tproxy.cycle_point,
             execution_time_limit=job_conf.get('execution_time_limit'),
-            platform=job_conf.get('platform')['name'],
+            platform=job_conf['platform']['name'],
```
oliver-sanders, line 1383: Unrelated, spotted during testing.

The `.get` here wasn't protecting against anything as the default return value of `None` would cause the `['name']` bit to error anyway.

wxtim (approved):
This is great. Please stop with the mission creep so that we can merge. Pretty Please.

markgrahamdawson (approved):

PR #5732, opened 2023-09-15
parentless sequential clock trigger spawns

dwsutherland:
superseded by #5738 

Spawning parentless clock triggered tasks out to the run ahead limit is unnecessary, and creates a lot of UI clutter.

Naturally clock trigger tasks are sequential in time, so it makes sense to only spawn the next parentless clock-trigger task when the current clock trigger is satisfied.

This PR attempts this.. In the simplest case;
```
[scheduling]
    initial cycle point = 20230915T0140Z
    [[xtriggers]]
        clock_1 = wall_clock()
    [[graph]]
        PT1M = """
@clock_1 => a
a => b
"""
[runtime]
    [[root]]
        script = sleep 5
    [[a,b]]
```
cutting down this;
image

to this:
image

(would only show one cycle, but obviously the trigger is satisfied)

Real workflows often have a much greater chain than `a => b` (and possibly much greater run-ahead limits), so you can imagine the clutter. (such as those at NIWA)

One thing to note; if your task has multiple xtriggers including a clock trigger, i.e.;
```
    [[xtriggers]]
        clock_1 = wall_clock()
        up_1 = workflow_state(workflow=clockspawn, task=b, point=%(point)s, offset=-P1Y):PT10S
    [[graph]]
        PT1M = """
@clock_1 => a
a => b
"""
        +PT1M/PT1M = """
@up_1 => b
"""
```
Then the task will still be clock blocked from spawning out to the RH (will instead spawn sequentially on trigger satisfaction, as described).
And if for some fringe reason a user wants the non-clock xtrigger to be checked/run out into the future/RH-limit **and** associated task clock bound to when it actually runs, then they can simply split them up:
```
    [[xtriggers]]
        clock_1 = wall_clock()
        up_1 = workflow_state(workflow=clockspawn, task=b, point=%(point)s, offset=-P1Y):PT10S
    [[graph]]
        PT1M = """
@clock_1 => a
a => b
"""
        +PT1M/PT1M = """
@up_1 => up_xtrig
up_xtrig => a
"""
```
(or w/e they desire.. alternatively turning their bespoke xtrigger into a task if that works)

In this way, I believe the behavior introduced here should be the default/natural restriction..

**Check List**

- [x] Appears to be covered by existing Tests.

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -467,6 +467,7 @@ class Scheduler:
             self.config,
             self.workflow_db_mgr,
             self.task_events_mgr,
+            self.xtrigger_mgr,
             self.data_store_mgr,
             self.flow_mgr
         )
@@ -1734,6 +1735,8 @@ class Scheduler:
                     self.pool.queue_task(itask)
 
             if housekeep_xtriggers:
+                # Spawn parentless clock-triggered
+                self.pool.spawn_parentless_clock_triggered()
                 # (Could do this periodically?)
                 self.xtrigger_mgr.housekeep(self.pool.get_tasks())
 
--- a/cylc/flow/xtrigger_mgr.py
+++ b/cylc/flow/xtrigger_mgr.py
@@ -211,6 +211,11 @@ class XtriggerManager:
         # Signatures of active functions (waiting on callback).
         self.active: list = []
 
+        # Gather named wall-clocks closet to current time.
+        # (no need to check or spawn future clocks of tasks respectively).
+        self.wall_clock_labels: set = set()
+        self.wall_clock_spawns: list = []
+
         self.workflow_run_dir = workflow_run_dir
 
         # For function arg templating.
@@ -316,6 +321,8 @@ class XtriggerManager:
         """
         self.validate_xtrigger(label, fctx, fdir)
         self.functx_map[label] = fctx
+        if fctx.func_name == "wall_clock":
+            self.wall_clock_labels.add(label)
 
     def mutate_trig(self, label, kwargs):
         self.functx_map[label].func_kwargs.update(kwargs)
@@ -380,7 +387,7 @@ class XtriggerManager:
 
         args = []
         kwargs = {}
-        if ctx.func_name == "wall_clock":
+        if label in self.wall_clock_labels:
             if "trigger_time" in ctx.func_kwargs:
                 # Internal (retry timer): trigger_time already set.
                 kwargs["trigger_time"] = ctx.func_kwargs["trigger_time"]
@@ -420,10 +427,12 @@ class XtriggerManager:
             itask: task proxy to check.
         """
         for label, sig, ctx, _ in self._get_xtrigs(itask, unsat_only=True):
-            if sig.startswith("wall_clock"):
+            if label in self.wall_clock_labels:
                 # Special case: quick synchronous clock check.
                 if wall_clock(*ctx.func_args, **ctx.func_kwargs):
                     itask.state.xtriggers[label] = True
+                    if itask.tdef.is_parentless(itask.point):
+                        self.wall_clock_spawns.append(itask)
                     self.sat_xtrig[sig] = {}
                     self.data_store_mgr.delta_task_xtrigger(sig, True)
                     LOG.info('xtrigger satisfied: %s = %s', label, sig)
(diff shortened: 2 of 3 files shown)


dwsutherland:
Actually, I might broaden this to optionally specify other xtriggers as `sequential-spawn`, e.g:
```
    [[xtriggers]]
        up_a = workflow_state(workflow=aflow, task=a, point=%(point)s):PT10S
        [[[settings]]]
            sequential-spawn = up_a
    [[graph]]
        PT1M = """
@up_1 => b
"""
```

Because, for some workflows, I wouldn't want `workflow_state` xtriggers to flood the UI  (not to mention the process pool)..

Even though you could add a clock-trigger to make them sequential, this would clutter the workflow definition and be too narrow (i.e. you may want them to run sequential but not bound by real time)..

Wall clock xtriggers should still be sequential.

cylc/flow/scheduler.py
```diff
@@ -1734,6 +1735,8 @@ async def main_loop(self) -> None:
                     self.pool.queue_task(itask)
 
             if housekeep_xtriggers:
+                # Spawn parentless clock-triggered
+                self.pool.spawn_parentless_clock_triggered()
                 # (Could do this periodically?)
                 self.xtrigger_mgr.housekeep(self.pool.get_tasks())
```
hjoliver, line 1741: Is this the right place to handle the spawning? The reason for the `if` block here is to "delete satisfied xtriggers no longer needed by any tasks" (from the `xtrigger_mgr.housekeep()` docstring).

cylc/flow/xtrigger_mgr.py
```diff
@@ -211,6 +211,11 @@ def __init__(
         # Signatures of active functions (waiting on callback).
         self.active: list = []
 
+        # Gather named wall-clocks closet to current time.
+        # (no need to check or spawn future clocks of tasks respectively).
+        self.wall_clock_labels: set = set()
+        self.wall_clock_spawns: list = []
```
hjoliver, line 217: Could this variable be named better?  like `clock_spawn_next` or something?

hjoliver (commented):
Had a good play around with this; it appears to do what it says on the tin 👍 

Couple of comments...

hjoliver:
My test case:
```ini
[scheduling]
    initial cycle point = now
    [[xtriggers]]
        c2 = "wall_clock(offset=PT1M)"
        x1 = "xrandom(percent=50, _=%(point)s)"
    [[graph]]
        PT1M = """
            @wall_clock => foo1
            @c2 => foo2
            @x1 => bar
            qux
        """
[runtime]
    [[foo1, foo2, bar, qux]]
        script = sleep 5
```
One weird thing noticed.  Spawned tasks waiting on clock triggers should be visible as waiting in `n=0`. When I run this, `foo1` disappears from Tui after a few cycles, except when it is actually running. But `foo2` is always visible as a waiting task.  Both are present in `cylc dump -t`, however, so presumably that implies a bug in Tui, not the datastore?

hjoliver:
Tested with a parent in one cycle too, all good. As above, with this additional bit of graph:
```ini
       R1/+PT2M/PT0M = "qux => foo2"       
```

hjoliver:
Test coverage passed, but presumably we aren't checking the new spawning behaviour since you haven't modified any of the test files yet.

cylc/flow/scheduler.py
dwsutherland: I noticed that `housekeep_xtriggers` is satisfied if xtriggers are, so just piggy backed off that.. But could just use that spawns list.

cylc/flow/xtrigger_mgr.py
dwsutherland: Yes, will change.. maybe to broaden `sequential_spawn_next`

dwsutherland:
> When I run this, foo1 disappears from Tui after a few cycles, except when it is actually running.

Yes, I found the same.. Will have a look (maybe after next revision)

> Coverage passed, but we can't be verifying the new spawning behavior as you haven't modified any of the test files.

I guess what I mean, is the current tests appear to cover 100% of this patch.. But, yes, will need to add specific tests (especially with expanded to optionally other xtriggers)

dwsutherland:
superseded by #5738

PR #5738, opened 2023-09-20
parentless sequential xtrigger spawning

dwsutherland:
supersedes #5732 

Spawning parentless xtrigger tasks out to the run ahead limit is unnecessary in the majority of cases, and creates a lot of UI clutter.

This PR makes sequential spawning on xtrigger satisfaction for parentless tasks optional, with non-sequential the default (no change). And the default for clock triggers.

Justification:
- Naturally clock trigger tasks are sequential in time, so it makes sense to only spawn the next parentless clock-triggered task when the current clock trigger is satisfied.
- Most workflows are sequential, and so workflows/downstream-workflows usually don't need to check xtriggers out to the runahead limit (i.e. `workflow_state`).
- It makes little difference (split seconds) in "catch-up" mode, where all xtriggers will be satisfied on first check.
- It puts less load on the Scheduler and UI, with `~1/runahead-limit` times less xtriggers to check and tasks to show.
- If the user wants non-sequential xtrigger checking, that xtrigger can be specified as such.

At NIWA we have a modular approach to running our workflows, both in operations  (around ~50 workflows, most interconnected) and in research (commonly downstream of operational workflows). So `wall_clock` and `workflow_state` are by far the most common xtriggers. We do have some requirement for non-sequential `workflow_state` xtriggers (i.e. satellite checking, where data may or may not come in for a cycle), however, nearly all workflows and downstream workflows are sequential.. And many of these require very large runahead limits.

Take the following simple example;
```
[scheduling]
    initial cycle point = 20230920T0530Z
    runahead limit = P5
    [[xtriggers]]
        clock_1 = wall_clock()
        up_1 = workflow_state(workflow=clockspawn, task=b, point=%(point)s):PT7S
    [[graph]]
        PT1M = """
@clock_1 => a
a => b
@up_1 => c
"""

[runtime]
    [[root]]
        script = sleep 2
    [[a,b,c]]
```
Currently that produces:
image

With this default behavior changed;
```
[scheduling]
    initial cycle point = 20230920T0530Z
    runahead limit = P5
    sequential xtriggers = True
    [[xtriggers]]
        clock_1 = wall_clock()
        up_1 = workflow_state(workflow=clockspawn, task=b, point=%(point)s):PT7S
.
.
.
```
it is reduced to:
image

Real workflows commonly have **way more** tasks (branches length and/or parallel) and **way larger** runahead limits, so the efficiency and clutter-less gains are massive.

And if for some reason users want some xtriggers to be checked/run out into the future/RH-limit in a non-sequential manner, then they can override the default with an argument (i.e. `sequential=True/False`):
```
    sequential xtriggers = True
    [[xtriggers]]
        clock_1 = wall_clock()
        up_1 = workflow_state(workflow=clockspawn, task=b, point=%(point)s, sequential=False):PT7S
.
.
.
```
image
(if there's multiple xtriggers on a single task it upholds the sequential setting if any)

This `sequential` argument name is reserved, and not passed onto the xtrigger module/function.
It can be defined in the xtrigger module/function (as in workflow declaration), but must have a Boolean default:
image
(i.e. from modifying `cylc.flow.xtriggers.workflow_state` to include settings arg in function definition)
`cylc.flow.xtriggers.wall_clock.wall_clock` has `sequential=True` set

On xtrigger satisfaction the corresponding parentless sequential xtriggered task (PSXT) will spawn out to the RH limit, or next occurrence of a PSXT, or non-parentless task occurrence.

**Note:**
Tui sometimes shows no tasks before a corresponding xtrigger is satisfied. This problem is not reflected in the data-store or WUI.

**Check List**

### DIFF
--- a/changes.d/5738.feat.md
+++ b/changes.d/5738.feat.md
@@ -1,0 +1,1 @@
+Optionally spawn parentless xtriggered tasks sequentially - i.e., one at a time, after the previous xtrigger is satisfied, instead of all at once out to the runahead limit. The `wall_clock` xtrigger is now sequential by default.
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -814,8 +814,26 @@ with Conf(
 
                 Example::
 
-                ``my_trigger(arg1, arg2, kwarg1, kwarg2):PT10S``
+            .. code-block:: cylc
+
+                [[xtriggers]]
+                    my_trig = my_trigger(arg1, arg2, kwarg1, kwarg2):PT10S
+                    [[[settings]]]
+                        non-sequential xtriggers = my_trig
+
             ''')
+            with Conf('settings', desc='''
+                Section heading for xtrigger behavior settings.
+            ''') as Queue:
+                Conf('non-sequential xtriggers', VDR.V_STRING_LIST, desc='''
+                    A list of xtrigger labels whose xtrigger, of associated
+                    parentless task, be checked out to the runahead limit.
+
+                    This allows for non-sequential xtrigger checking.
+
+                    A task with multiple xtriggers requires all labels
+                    be specified to behave in this way.
+                ''')
 
         with Conf('graph', desc=f'''
             The workflow graph is defined under this section.
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -802,9 +802,21 @@ with Conf(
                     :ref:`SequentialTasks`.
             ''')
 
+        Conf('sequential xtriggers default', VDR.V_BOOLEAN, False, desc='''
+            Set to ``True``, this allows for sequential spawning of associated
+            parentless tasks on xtrigger satisfaction.
+            Instead of out to the runahead limit (default: ``False``).
+
+            This workflow wide default can be overridden by a reserved
+            keyword argument in the xtrigger function declaration
+            (``sequential=True/False``).
+
+            The presence of one sequential xtrigger on a parentless task with
+            multiple xtriggers will cause sequential behavior.
+        ''')
         with Conf('xtriggers', desc='''
-                This section is for *External Trigger* function declarations -
-                see :ref:`Section External Triggers`.
+            This section is for *External Trigger* function declarations -
+            see :ref:`Section External Triggers`.
         '''):
             Conf('<xtrigger name>', VDR.V_XTRIGGER, desc='''
                 Any user-defined event trigger function declarations and
@@ -814,26 +826,8 @@ with Conf(
 
                 Example::
 
-            .. code-block:: cylc
-
-                [[xtriggers]]
-                    my_trig = my_trigger(arg1, arg2, kwarg1, kwarg2):PT10S
-                    [[[settings]]]
-                        non-sequential xtriggers = my_trig
-
+                ``my_trigger(arg1, arg2, kwarg1, kwarg2):PT10S``
             ''')
-            with Conf('settings', desc='''
-                Section heading for xtrigger behavior settings.
-            ''') as Queue:
-                Conf('non-sequential xtriggers', VDR.V_STRING_LIST, desc='''
-                    A list of xtrigger labels whose xtrigger, of associated
-                    parentless task, be checked out to the runahead limit.
-
-                    This allows for non-sequential xtrigger checking.
-
-                    A task with multiple xtriggers requires all labels
-                    be specified to behave in this way.
-                ''')
 
         with Conf('graph', desc=f'''
             The workflow graph is defined under this section.
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -808,8 +808,8 @@ with Conf(
             Instead of out to the runahead limit (default: ``False``).
 
             This workflow wide default can be overridden by a reserved
-            keyword argument in the xtrigger function declaration
-            (``sequential=True/False``).
+            keyword argument in the xtrigger function declaration and/or
+            function (``sequential=True/False``).
 
             The presence of one sequential xtrigger on a parentless task with
             multiple xtriggers will cause sequential behavior.
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -808,9 +808,8 @@ with Conf(
             until their previous (cycle point) instance is satisfied.
             Otherwise, they will all spawn at once out to the runahead limit.
 
-            This setting can be overridden by a reserved keyword argument in
-            individual xtrigger declarations, or in xtrigger function
-            definitions.
+            This setting can be overridden by the reserved keyword argument
+            ``sequential`` in individual xtrigger declarations.
 
             One sequential xtrigger on a parentless task with multiple
             xtriggers will cause sequential behavior.
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -802,18 +802,18 @@ with Conf(
                     :ref:`SequentialTasks`.
             ''')
 
-        Conf('sequential xtriggers default', VDR.V_BOOLEAN, False, desc='''
-            When set to ``True``, parentless tasks that trigger off xtriggers
-            will only spawn sequentially, i.e. on the satisfaction of the
-            xtriggers in order. Otherwise, these tasks will all spawn at the
-            same time up to the runahead limit.
+        Conf('spawn from xtriggers sequentially', VDR.V_BOOLEAN, False,
+             desc='''
+            If ``True``, tasks that only depend on xtriggers will not spawn
+            until their previous (cycle point) instance is satisfied.
+            Otherwise, they will all spawn at once out to the runahead limit.
 
-            This workflow-wide default can be overridden by a reserved
-            keyword argument in the xtrigger function declaration and/or
-            function (``sequential=True/False``).
+            This setting can be overridden by a reserved keyword argument in
+            individual xtrigger declarations, or in xtrigger function
+            definitions.
 
-            The presence of one sequential xtrigger on a parentless task with
-            multiple xtriggers will cause sequential behavior.
+            One sequential xtrigger on a parentless task with multiple
+            xtriggers will cause sequential behavior.
         ''')
         with Conf('xtriggers', desc='''
             This section is for *External Trigger* function declarations -
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -802,17 +802,17 @@ with Conf(
                     :ref:`SequentialTasks`.
             ''')
 
-        Conf('spawn from xtriggers sequentially', VDR.V_BOOLEAN, False,
+        Conf('sequential xtriggers', VDR.V_BOOLEAN, False,
              desc='''
             If ``True``, tasks that only depend on xtriggers will not spawn
-            until their previous (cycle point) instance is satisfied.
+            until the xtrigger of previous (cycle point) instance is satisfied.
             Otherwise, they will all spawn at once out to the runahead limit.
 
             This setting can be overridden by the reserved keyword argument
             ``sequential`` in individual xtrigger declarations.
 
             One sequential xtrigger on a parentless task with multiple
-            xtriggers will cause sequential behavior.
+            xtriggers will cause sequential spawning.
         ''')
         with Conf('xtriggers', desc='''
             This section is for *External Trigger* function declarations -
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -803,11 +803,12 @@ with Conf(
             ''')
 
         Conf('sequential xtriggers default', VDR.V_BOOLEAN, False, desc='''
-            Set to ``True``, this allows for sequential spawning of associated
-            parentless tasks on xtrigger satisfaction.
-            Instead of out to the runahead limit (default: ``False``).
+            When set to ``True``, parentless tasks that trigger off xtriggers
+            will only spawn sequentially, i.e. on the satisfaction of the
+            xtriggers in order. Otherwise, these tasks will all spawn at the
+            same time up to the runahead limit.
 
-            This workflow wide default can be overridden by a reserved
+            This workflow-wide default can be overridden by a reserved
             keyword argument in the xtrigger function declaration and/or
             function (``sequential=True/False``).
 
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -1705,7 +1705,13 @@ class WorkflowConfig:
             self.taskdefs[right].add_dependency(dependency, seq)
 
         validator = XtriggerNameValidator.validate
-        for label in self.cfg['scheduling']['xtriggers']:
+        xtrigs = self.cfg['scheduling']['xtriggers']
+        for label in xtrigs:
+            if (
+                label == 'settings'
+                and not isinstance(xtrigs[label], SubFuncContext)
+            ):
+                continue
             valid, msg = validator(label)
             if not valid:
                 raise WorkflowConfigError(
@@ -1714,7 +1720,7 @@ class WorkflowConfig:
 
         for label in xtrig_labels:
             try:
-                xtrig = self.cfg['scheduling']['xtriggers'][label]
+                xtrig = xtrigs[label]
             except KeyError:
                 if label != 'wall_clock':
                     raise WorkflowConfigError(f"xtrigger not defined: {label}")
@@ -1739,6 +1745,12 @@ class WorkflowConfig:
                 self.xtrigger_mgr.add_trig(label, xtrig, self.fdir)
 
             self.taskdefs[right].add_xtrig_label(label, seq)
+
+        if self.xtrigger_mgr is not None:
+            with suppress(KeyError):
+                self.xtrigger_mgr.non_sequential_labels.update(
+                    set(xtrigs['settings']['non-sequential xtriggers'])
+                )
 
     def get_actual_first_point(self, start_point):
         """Get actual first cycle point for the workflow
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -1715,7 +1715,7 @@ class WorkflowConfig:
 
         if self.xtrigger_mgr is not None:
             self.xtrigger_mgr.sequential_xtriggers_default = (
-                self.cfg['scheduling']['spawn from xtriggers sequentially']
+                self.cfg['scheduling']['sequential xtriggers']
             )
         for label in xtrig_labels:
             try:
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -1707,17 +1707,16 @@ class WorkflowConfig:
         validator = XtriggerNameValidator.validate
         xtrigs = self.cfg['scheduling']['xtriggers']
         for label in xtrigs:
-            if (
-                label == 'settings'
-                and not isinstance(xtrigs[label], SubFuncContext)
-            ):
-                continue
             valid, msg = validator(label)
             if not valid:
                 raise WorkflowConfigError(
                     f'Invalid xtrigger name "{label}" - {msg}'
                 )
 
+        if self.xtrigger_mgr is not None:
+            self.xtrigger_mgr.sequential_xtriggers_default = (
+                self.cfg['scheduling']['sequential xtriggers default']
+            )
         for label in xtrig_labels:
             try:
                 xtrig = xtrigs[label]
@@ -1745,12 +1744,6 @@ class WorkflowConfig:
                 self.xtrigger_mgr.add_trig(label, xtrig, self.fdir)
 
             self.taskdefs[right].add_xtrig_label(label, seq)
-
-        if self.xtrigger_mgr is not None:
-            with suppress(KeyError):
-                self.xtrigger_mgr.non_sequential_labels.update(
-                    set(xtrigs['settings']['non-sequential xtriggers'])
-                )
 
     def get_actual_first_point(self, start_point):
         """Get actual first cycle point for the workflow
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -1715,7 +1715,7 @@ class WorkflowConfig:
 
         if self.xtrigger_mgr is not None:
             self.xtrigger_mgr.sequential_xtriggers_default = (
-                self.cfg['scheduling']['sequential xtriggers default']
+                self.cfg['scheduling']['spawn from xtriggers sequentially']
             )
         for label in xtrig_labels:
             try:
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1750,7 +1750,6 @@ class Scheduler:
                     self.pool.queue_task(itask)
 
             if self.xtrigger_mgr.sequential_spawn_next:
-                # Spawn parentless tasks with sequentially checked xtrigger(s).
                 self.pool.spawn_parentless_sequential_xtriggers()
 
             if self.xtrigger_mgr.do_housekeeping:
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -2151,7 +2151,7 @@ class Scheduler:
     def command_force_trigger_tasks(
         self,
         tasks: Iterable[str],
-        flow: List[str],
+        flow: List[Union[str, int]],
         flow_wait: bool = False,
         flow_descr: Optional[str] = None
     ):
@@ -2162,7 +2162,7 @@ class Scheduler:
     def command_set(
         self,
         tasks: List[str],
-        flow: List[str],
+        flow: List[Union[str, int]],
         outputs: Optional[List[str]] = None,
         prerequisites: Optional[List[str]] = None,
         flow_wait: bool = False,
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -2151,7 +2151,7 @@ class Scheduler:
     def command_force_trigger_tasks(
         self,
         tasks: Iterable[str],
-        flow: List[Union[str, int]],
+        flow: List[str],
         flow_wait: bool = False,
         flow_descr: Optional[str] = None
     ):
@@ -2162,7 +2162,7 @@ class Scheduler:
     def command_set(
         self,
         tasks: List[str],
-        flow: List[Union[str, int]],
+        flow: List[str],
         outputs: Optional[List[str]] = None,
         prerequisites: Optional[List[str]] = None,
         flow_wait: bool = False,
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -481,6 +481,7 @@ class Scheduler:
             self.config,
             self.workflow_db_mgr,
             self.task_events_mgr,
+            self.xtrigger_mgr,
             self.data_store_mgr,
             self.flow_mgr
         )
@@ -1747,6 +1748,10 @@ class Scheduler:
 
                 if all(itask.is_ready_to_run()):
                     self.pool.queue_task(itask)
+
+            if self.xtrigger_mgr.sequential_spawn_next:
+                # Spawn parentless tasks with sequentially checked xtrigger(s).
+                self.pool.spawn_parentless_sequential_xtriggers()
 
             if self.xtrigger_mgr.do_housekeeping:
                 self.xtrigger_mgr.housekeep(self.pool.get_tasks())
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1803,7 +1803,7 @@ class TaskPool:
         items: Iterable[str],
         outputs: List[str],
         prereqs: List[str],
-        flow: List[Union[str, int]],
+        flow: List[str],
         flow_wait: bool = False,
         flow_descr: Optional[str] = None
     ):
@@ -2001,7 +2001,7 @@ class TaskPool:
 
     def _get_flow_nums(
             self,
-            flow: List[Union[str, int]],
+            flow: List[str],
             meta: Optional[str] = None,
     ) -> Optional[Set[int]]:
         """Get correct flow numbers given user command options."""
@@ -2073,7 +2073,7 @@ class TaskPool:
 
     def force_trigger_tasks(
         self, items: Iterable[str],
-        flow: List[Union[str, int]],
+        flow: List[str],
         flow_wait: bool = False,
         flow_descr: Optional[str] = None
     ):
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -480,7 +480,10 @@ class TaskPool:
                 submit_num=submit_num,
                 is_late=bool(is_late),
                 flow_wait=bool(flow_wait),
-                is_manual_submit=bool(is_manual_submit)
+                is_manual_submit=bool(is_manual_submit),
+                sequential_xtrigger_labels=(
+                    self.xtrigger_mgr.sequential_xtrigger_labels
+                ),
             )
 
         except WorkflowConfigError:
@@ -562,15 +565,6 @@ class TaskPool:
                                 itask.flow_nums,
                             )
                         )
-
-            # Set xtrigger checking type, which effects parentless spawning.
-            if (
-                itask.tdef.is_parentless(itask.point)
-                and set(itask.state.xtriggers.keys()).intersection(
-                    self.xtrigger_mgr.sequential_xtrigger_labels
-                )
-            ):
-                itask.is_xtrigger_sequential = True
 
             if itask.state_reset(status, is_runahead=True):
                 self.data_store_mgr.delta_task_runahead(itask)
@@ -712,6 +706,18 @@ class TaskPool:
     ) -> 'Tuple[Optional[TaskProxy], bool, bool]':
         """Return new or existing task point/name with merged flow_nums.
 
+        Returns:
+            tuple - (itask, is_in_pool, is_xtrig_sequential)
+
+            itask:
+                The requested task proxy, or None if task does not
+                exist or cannot spawn.
+            is_in_pool:
+                Was the task found in a pool.
+            is_xtrig_sequential:
+                Is the next task occurance spawned on xtrigger satisfaction,
+                or do all occurances spawn out to the runahead limit.
+
         It does not add a spawned task proxy to the pool.
         """
         ntask = self._get_task_by_id(
@@ -723,19 +729,15 @@ class TaskPool:
             # ntask does not exist: spawn it in the flow.
             ntask = self.spawn_task(tdef.name, point, flow_nums, flow_wait)
             # if the task was found set xtrigger checking type.
+            # otherwise find the xtrigger type if it can't spawn
+            # for whatever reason.
             if ntask is not None:
-                if set(ntask.state.xtriggers.keys()).intersection(
-                    self.xtrigger_mgr.sequential_xtrigger_labels
-                ):
-                    ntask.is_xtrigger_sequential = True
-                    is_xtrig_sequential = True
-            elif {
-                xtrig_label
+                is_xtrig_sequential = ntask.is_xtrigger_sequential
+            elif any(
+                xtrig_label in self.xtrigger_mgr.sequential_xtrigger_labels
                 for sequence, xtrig_labels in tdef.xtrig_labels.items()
                 for xtrig_label in xtrig_labels
                 if sequence.is_valid(point)
-            }.intersection(
-                self.xtrigger_mgr.sequential_xtrigger_labels
             ):
                 is_xtrig_sequential = True
         else:
@@ -743,7 +745,8 @@ class TaskPool:
             is_in_pool = True
             self.merge_flows(ntask, flow_nums)
             is_xtrig_sequential = ntask.is_xtrigger_sequential
-        return ntask, is_in_pool, is_xtrig_sequential  # may be None
+        # ntask may still be None
+        return ntask, is_in_pool, is_xtrig_sequential
 
     def spawn_to_rh_limit(self, tdef, point, flow_nums) -> None:
         """Spawn parentless task instances from point to runahead limit.
@@ -807,6 +810,13 @@ class TaskPool:
             msg = "task completed"
         else:
             msg = f"removed ({reason})"
+
+        if itask.is_xtrigger_sequential:
+            with suppress(ValueError):
+                self.xtrigger_mgr.sequential_spawn_next.remove(
+                    itask.identity
+                )
+
         try:
             del self.active_tasks[itask.point][itask.identity]
         except KeyError:
@@ -1001,20 +1011,15 @@ class TaskPool:
                     itask.point,
                     itask.flow_nums,
                     itask.state.status,
+                    sequential_xtrigger_labels=(
+                        self.xtrigger_mgr.sequential_xtrigger_labels
+                    ),
                 )
                 itask.copy_to_reload_successor(
                     new_task,
                     self.check_task_output,
                 )
                 self._swap_out(new_task)
-                # Set xtrigger checking type for parentless spawning.
-                if (
-                    new_task.tdef.is_parentless(new_task.point)
-                    and set(new_task.state.xtriggers.keys()).intersection(
-                        self.xtrigger_mgr.sequential_xtrigger_labels
-                    )
-                ):
-                    new_task.is_xtrigger_sequential = True
                 self.data_store_mgr.delta_task_prerequisite(new_task)
                 LOG.info(f"[{itask}] reloaded task definition")
                 if itask.state(*TASK_STATUSES_ACTIVE):
@@ -1721,7 +1726,10 @@ class TaskPool:
             flow_wait=flow_wait,
             submit_num=submit_num,
             transient=transient,
-            is_manual_submit=is_manual_submit
+            is_manual_submit=is_manual_submit,
+            sequential_xtrigger_labels=(
+                self.xtrigger_mgr.sequential_xtrigger_labels
+            ),
         )
         if itask is None:
             return None
@@ -2092,6 +2100,9 @@ class TaskPool:
                 flow_nums,
                 flow_wait=flow_wait,
                 submit_num=submit_num,
+                sequential_xtrigger_labels=(
+                    self.xtrigger_mgr.sequential_xtrigger_labels
+                ),
             )
             if itask is None:
                 continue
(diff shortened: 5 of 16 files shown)


oliver-sanders:
Note, the Cylc 7 spawning behaviour can be achieved by adding submit dependencies between the parentless tasks e.g:

```
@up => a
a[-PT1M]:submitted => a
```

This forces the sequentiality of `a` without prohibiting multiple instances of `a` from running in parallel.

dwsutherland:
> Note, the Cylc 7 spawning behaviour can be achieved by adding submit dependencies between the parentless tasks e.g:
> 
> ```
> @up => a
> a[-PT1M]:submitted => a
> ```
> 
> This forces the sequentiality of `a` without prohibiting multiple instances of `a` from running in parallel.

IMO that introduces clutter and false dependencies into the workflow definition.. I would have to do that roughly thousands of times across all 50 workflows..
Also some tasks might be in queues or tight job scheduler queues (and require submission retries)..

I have a change inbound, like what we messaged about in Element, where this will be solved without effecting the current behavior.. Using one line of workflow definition ..

oliver-sanders (commented):
I think this makes sense but haven't thought through the details as yet. Because this alters the spawning logic it does require a good amount of thought to ensure that it doesn't interact with other features. Here are some things this change could potentially interact with off the top of my head:

* Trigger - What happens if you manually trigger a task *before* the xtrigger is satisfied? If we're not careful, could this result in the task being removed from the workflow Cylc 7 style?
* Shutdown - We would also need to make sure that the task pool cant become empty as a result of delayed spawning as this could interact with shutdown logic.
* Remove - Could removing one instance of the task result in the task being removed from the workflow Cylc 7 style? Note `cylc remove` will work on active tasks soon.
* Flows - Could the delayed spawning change the behaviour of flow merging in unexpected ways?

cylc/flow/cfgspec/workflow.py
```diff
@@ -785,9 +785,21 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+            parentless tasks on xtrigger satisfaction.
+            Instead of out to the runahead limit (default: ``False``).
+
+            This workflow wide default can be overridden by a reserved
+            keyword argument in the xtrigger function declaration
+            (``sequential=True/False``).
+
+            The presence of one sequential xtrigger on a parentless task with
+            multiple xtriggers will cause sequential behavior.
+        ''')
```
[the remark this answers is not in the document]
dwsutherland: I reserve and strip `sequential` out of the args when it's read in...
I could undo the reservation and read in the function args (as I did for the reservation), but it would have to be well known, so people don't use `sequential` arg name for some other purpose (I could definitely change the arg type).

And I disagree, on not being able to set a workflow wide default.. Again some sites have hundreds/thousands of `wall_clock` and `workflow_state`.. And I would need to write in the argument to nearly all of them if we have no ability to change the default.. I would rather just

I don't want a global default (in the sense of a user/site setting), because this would be bad for sharing a workflow (as workflows may act differently for different users).

cylc/flow/task_pool.py
```diff
@@ -281,11 +284,12 @@ def release_runahead_tasks(self):
...
-                itask.tdef,
-                itask.tdef.next_point(itask.point),
-                itask.flow_nums
-            )
+            if not itask.is_xtrigger_sequential:
+                self.spawn_to_rh_limit(
+                    itask.tdef,
+                    itask.tdef.next_point(itask.point),
+                    itask.flow_nums
+                )
```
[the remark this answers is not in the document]
dwsutherland: Yes, I think I noticed this.. Will have a look.

cylc/flow/cfgspec/workflow.py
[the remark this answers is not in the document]
dwsutherland: Done, it's now an option to add `sequential=True/False` to the function (as well as declaration) .. It will error if is does not have a default or default is not `bool`:
image

dwsutherland:
Restart/reload and remove are now handled..

Remove will cause the next occurrence to spawn, the argument being:
- the non-sequential case would be spawned, so be consistent.
- otherwise the workflow would shutdown if it's the only task (and we have a stop button for that).

cylc/flow/task_pool.py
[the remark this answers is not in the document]
dwsutherland: Fixed I believe, and restart/reload/remove incorporated into the test.

dwsutherland:
I assume the docs are auto-generated from the config definition.. but I suppose I should build it with this branch to find out.

cylc/flow/task_pool.py
```diff
@@ -719,46 +732,87 @@ def rh_release_and_queue(self, itask) -> None:
...
         )
+        is_in_pool = False
+        is_xtrig_sequential = False
         if ntask is None:
             # ntask does not exist: spawn it in the flow.
-            ntask = self.spawn_task(name, point, flow_nums)
+            ntask = self.spawn_task(tdef.name, point, flow_nums)
+            # if the task was found set xtrigger checking type.
+            if ntask is not None:
+                if set(ntask.state.xtriggers.keys()).intersection(
```
oliver-sanders, line 750:
```suggestion
                if set(ntask.state.xtriggers).intersection(
```

cylc/flow/task_pool.py
```diff
@@ -719,46 +732,87 @@ def rh_release_and_queue(self, itask) -> None:
...
+                self.xtrigger_mgr.sequential_xtrigger_labels
+            ):
+                is_xtrig_sequential = True
         else:
             # ntask already exists (n=0 or incomplete): merge flows.
+            is_in_pool = True
             self.merge_flows(ntask, flow_nums)
-        return ntask  # may be None
+            is_xtrig_sequential = ntask.is_xtrigger_sequential
+        return ntask, is_in_pool, is_xtrig_sequential  # may be None
```
oliver-sanders, line 769:
```suggestion
        return ntask, is_in_pool, is_xtrig_sequential
```

cylc/flow/task_pool.py
```diff
@@ -719,46 +732,87 @@ def rh_release_and_queue(self, itask) -> None:
             self.queue_task(itask)
 
     def _get_spawned_or_merged_task(
-        self, point: 'PointBase', name: str, flow_nums: 'FlowNums'
-    ) -> Optional[TaskProxy]:
+        self, point: 'PointBase', tdef: 'TaskDef', flow_nums: 'FlowNums'
+    ) -> 'Tuple[Optional[TaskProxy], bool, bool]':
         """Return new or existing task point/name with merged flow_nums"""
```
oliver-sanders, line 737: Would be good to have a quick description of the two new return values for future reference:

```suggestion
        """Return new or existing task point/name with merged flow_nums
        
        Returns:
            tuple - (itask, is_in_pool, is_xtrig_sequential)

            itask:
                The requested task proxy, or None if task does not
                exist or is not valid at the specified point.
            is_in_pool:
                TODO
            is_xtrig_sequential:
                TODO

        """
```

oliver-sanders (commented):
Tried a couple of functional tests, works nicely.

When a workflow is rapidly catching up, clock-triggered tasks get pushed through in batches of the runahead limit which is a slight quirk, but nothing we can't live with.

This logic needs more tests, particularly covering interactions with this logic, e.g. trigger, remove, reload, etc. These interactions are easy to get wrong and really easy to break by accident.

I've made a start at some integration tests and commented a couple of suggestions for new ones. Two of my tests currently fail and will need a look in:

```python
# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
#
# This program is free software: you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation, either version 3 of the License, or
# (at your option) any later version.
#
# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.
#

"""Test interactions with sequential xtriggers."""

import pytest

from cylc.flow.cycling.iso8601 import ISO8601Point


@pytest.fixture()
def sequential(flow, scheduler):
    id_ = flow({
        'scheduler': {
            'allow implicit tasks': 'True',
            'cycle point format': 'CCYY',
        },
        'scheduling': {
            'runahead limit': 'P2',
            'initial cycle point': '2000',
            'graph': {
                'P1Y': '@wall_clock => foo',
            }
        }
    })

    sequential = scheduler(id_)

    def list_tasks():
        """List the task instance cycle points present in the pool."""
        nonlocal sequential
        return sorted(itask.tokens['cycle'] for itask in sequential.pool.get_all_tasks())

    sequential.list_tasks = list_tasks

    return sequential


async def test_remove(sequential, start):
    """It should spawn the next instance when a task is removed.

    Ensure that removing a task with a sequential xtrigger does not break the
    chain causing future instances to be removed from the workflow.
    """
    async with start(sequential):
        # the scheduler starts with one task in the pool
        assert sequential.list_tasks() == ['2000']

        # it sequentially spawns out to the runahead limit
        for year in range(2000, 2010):
            foo = sequential.pool.get_task(ISO8601Point(f'{year}'), 'foo')
            if foo.state(is_runahead=True):
                break
            sequential.xtrigger_mgr.call_xtriggers_async(foo)
            sequential.pool.spawn_parentless_sequential_xtriggers()
        assert sequential.list_tasks() == [
            '2000',
            '2001',
            '2002',
            '2003',
        ]

        # remove all tasks in the pool
        sequential.pool.remove_tasks(['*'])

        # the next cycle should be automatically spawned
        assert sequential.list_tasks() == ['2004']

        # TODO: This usually fails because each task that is removed spawns its
        # next instance, so depending on the order they are removed in, some
        # tasks get removed then re-spawned again.
        # NOTE: You won't spot this issue in a functional test because the
        # re-spawned tasks are detected as completed and automatically removed.
        # So ATM not dangerous, but potentially inefficient.


async def test_trigger(sequential, start):
    """It should spawn its next instance if triggered ahead of time.

    If you manually trigger a sequentially spawned task before its xtriggers
    have become satisfied, then the sequential spawning chain is broken.

    The task pool should defend against this to ensure that triggering a task
    doesn't cancel it's future instances.
    """
    async with start(sequential):
        assert sequential.list_tasks() == ['2000']

        foo = sequential.pool.get_task(ISO8601Point('2000'), 'foo')
        sequential.pool.force_trigger_tasks([foo.identity], {1})
        foo.state_reset('succeeded')
        sequential.pool.spawn_on_output(foo, 'succeeded')

        assert sequential.list_tasks() == ['2001']

        # TODO: This test fails with an empty pool


async def test_reload(sequential, start):
    """It should set the is_xtrigger_sequential flag on reload.

    TODO: test that changes to the sequential status in the config get picked
          up on reload
    """
    async with start(sequential):
        # the task should be marked as sequential
        pre_reload = sequential.pool.get_task(ISO8601Point('2000'), 'foo')
        assert pre_reload.is_xtrigger_sequential is True

        # reload the workflow
        sequential.pool.reload_taskdefs(sequential.config)

        # the original task proxy should have been replaced
        post_reload = sequential.pool.get_task(ISO8601Point('2000'), 'foo')
        assert id(pre_reload) != id(post_reload)

        # the new task should be marked as sequential
        assert post_reload.is_xtrigger_sequential is True


# TODO: test that a task is marked as sequential if any of its xtriggers are
# sequential (as opposed to all)?

# TODO: test setting the sequential argument in [scheduling][xtrigger] items
# changes the behaviour

# TODO: test the interaction between "sequential xtriggers default" and the
# sequential argument to [scheduling][xtrigger]
# * Should we be able to override the default by setting sequential=False?
# * Or should that result in a validation error?
```

cylc/flow/task_pool.py
```diff
@@ -1755,6 +1824,22 @@ def force_trigger_tasks(
...
+                or self._get_main_task_by_id(taskid)
+            )
+            # Will spawn out to RH limit or next parentless clock trigger
+            # or non-parentless.
+            self.spawn_to_rh_limit(
+                itask.tdef,
+                itask.tdef.next_point(itask.point),
+                itask.flow_nums
+            )
+
```
oliver-sanders, line 1842: I think we might need to get `cylc remove` to pop tasks off of this list to prevent tasks from being reincarnated?

cylc/flow/task_pool.py
```diff
@@ -1008,6 +1062,14 @@ def reload_taskdefs(self, config: 'WorkflowConfig') -> None:
...
                 )
                 self._swap_out(new_task)
+                # Set xtrigger checking type for parentless spawning.
+                if (
+                    new_task.tdef.is_parentless(new_task.point)
+                    and set(new_task.state.xtriggers.keys()).intersection(
+                        self.xtrigger_mgr.sequential_xtrigger_labels
+                    )
+                ):
+                    new_task.is_xtrigger_sequential = True
```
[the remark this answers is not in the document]
dwsutherland: Yes, granted the info is available at creation, that would simplify it..

dwsutherland:
Think I've made the suggested changes .. however not sure why fast tests are failing

MetRonnie:
TUI integration test failures are unrelated to your changes - see #5849

cylc/flow/task_pool.py
```diff
@@ -719,46 +726,98 @@ def rh_release_and_queue(self, itask) -> None:
...
+        # ntask may still be None
+        return ntask, is_in_pool, is_xtrig_sequential
 
     def spawn_to_rh_limit(self, tdef, point, flow_nums) -> None:
-        """Spawn parentless task instances from point to runahead limit."""
+        """Spawn parentless task instances from point to runahead limit.
+
+        Sequentially checked xtriggers with spawn corresponding task out to the
+        next task with any xtrigger with the same behaviour, or to the runahead
+        limit (whichever occurs first).
```
MetRonnie, line 781: I can't wrap my head around this sentence 😅

cylc/flow/task_pool.py
```diff
@@ -768,6 +827,12 @@ def remove(self, itask, reason=""):
         if reason:
             msg += f" ({reason})"
 
+        if itask.is_xtrigger_sequential:
+            with suppress(ValueError):
+                self.xtrigger_mgr.sequential_spawn_next.remove(
+                    itask.identity
+                )
```
MetRonnie, line 834: Why would a `ValueError` be raised here?
```suggestion
            self.xtrigger_mgr.sequential_spawn_next.remove(itask.identity)
```
or, if you expect a `KeyError` to be raised, use `set.discard()`:
```suggestion
            self.xtrigger_mgr.sequential_spawn_next.discard(itask.identity)
```

cylc/flow/task_pool.py
```diff
@@ -1755,6 +1836,22 @@ def force_trigger_tasks(
 
         return len(unmatched)
 
+    def spawn_parentless_sequential_xtriggers(self):
+        """Spawn successor(s) of parentless wall clock satisfied tasks."""
+        while self.xtrigger_mgr.sequential_spawn_next:
+            taskid = self.xtrigger_mgr.sequential_spawn_next.pop()
```
MetRonnie, line 1842: This is a set, so the order of spawning will be arbitrary. Does that matter?

cylc/flow/task_pool.py
```diff
@@ -1665,6 +1739,13 @@ def remove_tasks(self, items):
         """Remove tasks from the pool."""
         itasks, _, bad_items = self.filter_task_proxies(items)
         for itask in itasks:
+            # Spawn next occurance of xtrigger sequential task.
```
MetRonnie, line 1742:
```suggestion
            # Spawn next occurrence of xtrigger sequential task.
```

cylc/flow/task_pool.py
dwsutherland: Will reword.. Essentially it's saying the parentless task will spawn to it's next occurrence, which may be a task with a sequential xtrigger or a non-parentless task, otherwise all occurrences will spawn to the runahead limit.

cylc/flow/task_pool.py
dwsutherland: I don't think so.. Unless we need predictability .. It would also happen quite fast, anyways.

MetRonnie:
> However there are a few functional test failures here

Looks like it was the `ValueError` vs `KeyError` thing. Functional tests passing now 👍

dwsutherland:
> > However there are a few functional test failures here
> 
> Looks like it was the `ValueError` vs `KeyError` thing. Functional tests passing now 👍

Yeah, I had in my head that it was a list for some reason 🤦‍♂️

cylc/flow/task_pool.py
dwsutherland: Replaced with:

> Sequentially checked xtriggers will spawn the next occurrence of their
corresponding tasks. These tasks will keep spawning until they depend
on any unsatisfied xtrigger of the same sequential behavior, are no
longer parentless, and/or hit the runahead limit.

cylc/flow/xtriggers/wall_clock.py
```diff
@@ -19,11 +19,13 @@
 from time import time
 
 
-def wall_clock(trigger_time=None):
+def wall_clock(trigger_time=None, sequential=True):
```
MetRonnie, line 22: This means `wall_clock` xtrigger does not respect the `[scheduling]sequential xtriggers default` setting. Was that intentional?

dwsutherland: Correct, and intentional (`[scheduling]sequential xtriggers default = False` is the default`).  
However, this can be overruled by defining `sequential=False` in the declaration.. i.e.
```
    sequential xtriggers default = True
    [[xtriggers]]
        clock_1 = wall_clock(sequential=False)
```
or to overwrite the function default
```
    [[xtriggers]]
        clock_1 = wall_clock(sequential=False)
```
(so declaration trumps all)
This respects local trumps global..

For wall_clock, there is no downside of having `sequential=True` the default (with time itself being sequential, and spawning will be nigh instant for historical runs (out to RH))..

dwsutherland: Hmm.. On second thought, it might be clearer if workflow or declaration overwrites the function default   (but not the undeclared  `sequential xtriggers default` if I can distinguish them).. Will have a look

dwsutherland: No, in two minds .. I think if it's defined in the xtrigger function, then it's intentional no matter the workflow-wide default, and should only be explicitly overwritten in the xtrigger declaration

MetRonnie:
Oh I've just spotted earlier up the conversation @oliver-sanders was in the process of writing integration tests - any progress?

oliver-sanders:
I got as far as the code I commented here: https://github.com/cylc/cylc-flow/pull/5738#pullrequestreview-1747695603

But it looks like there were some issues documented in TODOs.

dwsutherland:
I'll add more tests. (note: anything with clocktriggers will now test this behavior implicitly, with clocktriggers are tested)
Think that comment was right around when the our HPC went a bit bonkers..

cylc/flow/xtrigger_mgr.py
```diff
@@ -282,6 +327,27 @@ def validate_xtrigger(
...
+                    fname,
+                    (
+                        f"xtrigger module '{fname}' contains reserved argument"
+                        " name 'sequential' that has no boolean default"
+                    ),
+                )
+            elif 'sequential' not in fctx.func_kwargs:
+                fctx.func_kwargs['sequential'] = x_argspec.defaults[
+                    x_argspec.args.index('sequential')
+                ]
```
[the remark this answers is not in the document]
dwsutherland: Either way really, I've seen a few that this could conflict with.

cylc/flow/cfgspec/workflow.py
```diff
@@ -785,9 +785,22 @@ def get_script_common_text(this: str, example: Optional[str] = None):
                     :ref:`SequentialTasks`.
             ''')
 
+        Conf('sequential xtriggers default', VDR.V_BOOLEAN, False, desc='''
```
hjoliver, line 788: Maybe this item could be more descriptive? Something like?:
```suggestion
        Conf('spawn from xtriggers sequentially', VDR.V_BOOLEAN, False, desc='''
```
(and I'm not sure "default" is needed in the name, as opposed to the description) 

(obviously changing this would have some knock-on effects in the code)

cylc/flow/cfgspec/workflow.py
```diff
@@ -785,9 +785,22 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+            When set to ``True``, parentless tasks that trigger off xtriggers
+            will only spawn sequentially, i.e. on the satisfaction of the
+            xtriggers in order. Otherwise, these tasks will all spawn at the
+            same time up to the runahead limit.
+
+            This workflow-wide default can be overridden by a reserved
+            keyword argument in the xtrigger function declaration and/or
+            function (``sequential=True/False``).
+
+            The presence of one sequential xtrigger on a parentless task with
```
hjoliver, line 798: Could be clearer.  A quick attempt at it:
```suggestion
            If ``True``, tasks that only depend on xtriggers will not spawn until their
            previous (cycle point) instance is satisfied. Otherwise, they will all spawn
            at once out to the runahead limit.

            This setting can be overridden by a reserved keyword argument in
            individual xtrigger declarations, or in xtrigger function definitions.

            One sequential xtrigger on a parentless task with
```

cylc/flow/scheduler.py
```diff
@@ -1733,6 +1734,10 @@ async def main_loop(self) -> None:
                 if all(itask.is_ready_to_run()):
                     self.pool.queue_task(itask)
 
+            if self.xtrigger_mgr.sequential_spawn_next:
+                # Spawn parentless tasks with sequentially checked xtrigger(s).
```
hjoliver, line 1738:
```suggestion
```
(Not a biggie, but with such verbose attribute names that comment is kinda redundant).

cylc/flow/task_pool.py
```diff
@@ -697,46 +704,99 @@ def rh_release_and_queue(self, itask) -> None:
...
 
     def spawn_to_rh_limit(self, tdef, point, flow_nums) -> None:
-        """Spawn parentless task instances from point to runahead limit."""
+        """Spawn parentless task instances from point to runahead limit.
+
+        Sequentially checked xtriggers will spawn the next occurrence of their
+        corresponding tasks. These tasks will keep spawning until they depend
+        on any unsatisfied xtrigger of the same sequential behavior, are no
+        longer parentless, and/or hit the runahead limit.
+
```
hjoliver, line 761:
```suggestion
```

Isn't this describing the spawning of tasks that don't go via this method? (So the comment should really be elsewhere).

cylc/flow/task_pool.py
```diff
@@ -697,46 +704,99 @@ def rh_release_and_queue(self, itask) -> None:
...
+            tuple - (itask, is_in_pool, is_xtrig_sequential)
+
+            itask:
+                The requested task proxy, or None if task does not
+                exist or cannot spawn.
+            is_in_pool:
+                Was the task found in a pool.
+            is_xtrig_sequential:
+                Is the next task occurance spawned on xtrigger satisfaction,
+                or do all occurances spawn out to the runahead limit.
```
hjoliver, line 721:
```suggestion
                Is the next task occurrence spawned on xtrigger satisfaction,
                or do all occurrence spawn out to the runahead limit.
```

hjoliver (commented):
Looks good, from a quick first run-through.

It's going to conflict a bit nastily with #5658 😬

dwsutherland:
I've found some holes in the spawning of parentless tasks with unsatisfied sequential xtriggers (via Oliver's integration test, and replicated manually).. should be easy fix

dwsutherland:
Added the integration tests (thanks Oliver), and have gotten the existing scenarios/tests working after adding more spawning conditions and requirements.
Will add more scenarios next.

dwsutherland:
hmm... all the TUI tests are failing...

oliver-sanders:
> hmm... all the TUI tests are failing...

Kicking off a test run on master to make sure this isn't the result of an external factor: https://github.com/cylc/cylc-flow/actions/runs/7887199508

MetRonnie:
Fixed on master now. Kicking tests

cylc/flow/cfgspec/workflow.py
dwsutherland: Yes, far less clunky than my attempt.

> tasks that only depend on xtriggers

If xtriggers are the only non-graph prerequisites possible (external triggers?), you'd know :)

cylc/flow/task_pool.py
dwsutherland: Well, I added this as a caveat to `spawn_to_rh_limit` .. Because as described, it won't always spawn to the RH limit.

dwsutherland: Oh sorry, misread, this method _is_ used to spawn x-sequentials, it just bails out early.

MetRonnie (commented):
With this example
```
[scheduler]
    allow implicit tasks = True
[scheduling]
    initial cycle point = 2018
    [[xtriggers]]
        clock_1 = wall_clock()
        clock_2 = wall_clock(P10Y)
    [[graph]]
        P1D = @clock_1 & @clock_2 => foo
```
Even though both clock triggers are sequential by default, I get all of the tasks spawning out to the runahead limit ~1s after each other

dwsutherland:
> Even though both clock triggers are sequential by default, I get all of the tasks spawning out to the runahead limit ~1s after each other

I guess the expected behavior would be to only spawn when all sequential clocktriggers are satisfied?  will add that behavior in

dwsutherland:
> > Even though both clock triggers are sequential by default, I get all of the tasks spawning out to the runahead limit ~1s after each other
> 
> I guess the expected behavior would be to only spawn when all sequential clocktriggers are satisfied? will add that behavior in

Yeah, not sure about this .. Maybe xtriggers shouldn't hold each other up, and xtriggers could be doing who knows what..
(and BTW - ~1s spawn out to the RH limit is because `clock_1` is satisfied straight away, so the next is spawned, as expected)

The point of this PR is to introduce a reasonable (and optional) restriction on the unbounded spawning and xtrigger-checking out the the RH limit.

If an xtrigger is satisfied, then the next occurrence (which may or may not be the next task occurrence) can be checked.. 
For example (this could be any xtrigger):
```
[scheduler]
    allow implicit tasks = True
[scheduling]
    initial cycle point = 2018
    [[xtriggers]]
        clock_1 = wall_clock()
        clock_2 = wall_clock(P10Y)
    [[graph]]
        P2D = @clock_1 & @clock_2 => foo
        ^+P1D/P2D = @clock_1 => foo
```
At present it works:
image

If we restrict spawning until both are ready, then the next `foo`s that can run won't.

Of course, having only `clock_2` on the first:
```
[scheduler]
    allow implicit tasks = True
[scheduling]
    initial cycle point = 2018
    [[xtriggers]]
        clock_1 = wall_clock()
        clock_2 = wall_clock(P10Y)
    [[graph]]
        P2D = @clock_2 => foo
        ^+P1D/P2D = @clock_1 => foo
```
Will stop foo from spawning to a cycle that it would be able to run.. But this is in the hands of the workflow designer.


However, I can understand that there is an inconsistency here:

The presence of an unsatisfied sequential xtrigger will prevent the task spawning, and hence, future checking of any non-sequential xtriggers (regardless of their satisfaction).
Yet the presence of a unsatisfied sequential xtrigger does not prevent spawning if satisfied sequential xtrigger exists


So in light of that, I think I'm full circle to maybe the spawning should be held up in the presence of **any** unsatisfied sequential xtrigger..

dwsutherland:
Ok fixed it now:
image

oliver-sanders:
Tui has been broken by an urwid release, don't upgrade beyond 1.6.1 and you'll be fine.

cylc/flow/xtrigger_mgr.py
```diff
@@ -435,18 +509,26 @@ def call_xtriggers_async(self, itask: 'TaskProxy'):
...
                     LOG.info('xtrigger satisfied: %s = %s', label, sig)
+                    if (
+                        itask.is_xtrigger_sequential
+                        and all(
+                            itask.state.xtriggers[seq_label]
+                            for seq_label in itask.state.xtriggers
+                            if seq_label in self.sequential_xtrigger_labels
+                        )
+                    ):
+                        self.sequential_spawn_next.add(itask.identity)
```
MetRonnie, line 531:
```diff
diff --git a/cylc/flow/xtrigger_mgr.py b/cylc/flow/xtrigger_mgr.py
index 90c32b8dc..861fc791e 100644
--- a/cylc/flow/xtrigger_mgr.py
+++ b/cylc/flow/xtrigger_mgr.py
@@ -521,12 +521,5 @@ class XtriggerManager:
                     self.workflow_db_mgr.put_xtriggers({sig: {}})
                     LOG.info('xtrigger satisfied: %s = %s', label, sig)
-                    if (
-                        itask.is_xtrigger_sequential
-                        and all(
-                            itask.state.xtriggers[seq_label]
-                            for seq_label in itask.state.xtriggers
-                            if seq_label in self.sequential_xtrigger_labels
-                        )
-                    ):
+                    if self.all_task_seq_xtriggers_satisfied(itask):
                         self.sequential_spawn_next.add(itask.identity)
                     self.do_housekeeping = True
@@ -548,12 +541,5 @@ class XtriggerManager:
                             xtrigger_env
                         )
-                    if (
-                        itask.is_xtrigger_sequential
-                        and all(
-                            itask.state.xtriggers[seq_label]
-                            for seq_label in itask.state.xtriggers
-                            if seq_label in self.sequential_xtrigger_labels
-                        )
-                    ):
+                    if self.all_task_seq_xtriggers_satisfied(itask):
                         self.sequential_spawn_next.add(itask.identity)
                 continue
@@ -613,2 +599,10 @@ class XtriggerManager:
             self.sat_xtrig[sig] = results
             self.do_housekeeping = True
+
+    def all_task_seq_xtriggers_satisfied(self, itask: 'TaskProxy') -> bool:
+        """Check if all sequential xtriggers are satisfied for a task."""
+        return itask.is_xtrigger_sequential and all(
+            itask.state.xtriggers[label]
+            for label in itask.state.xtriggers
+            if label in self.sequential_xtrigger_labels
+        )
```

dwsutherland: Yeah, thought about doing this.. But wasn't sure it mattered, but that method may potentially get called elsewhere too..
Done now.

MetRonnie: If it's used twice that's enough to make it its own method for me 👍  Also for readability

MetRonnie:
I've sorted out the conflict and added a few tests in https://github.com/dwsutherland/cylc-flow/pull/19

oliver-sanders:
Note, will need a changelog entry (autogenerated these days, see the CONTRIBUTING page) and possibly a small entry in the "External Triggers" section of cylc-doc?

MetRonnie:
Can build off https://github.com/cylc/cylc-doc/pull/686

dwsutherland:
> I've sorted out the conflict and added a few tests in [dwsutherland#19](https://github.com/dwsutherland/cylc-flow/pull/19)

Saw there were a lot of conflicts, so assume you had rebased on master, then rebased this branch on master.
Now all conflicts in your branch are only  in one file (the integration test).

MetRonnie:
I hadn't rebased, I had merged master into this branch for dwsutherland#19 (before you rebased this branch on top of master). There was a merge mistake when you pushed to dwsutherland#19 after this rebase so I've now force-pushed a fix onto this branch. If that makes sense lol

MetRonnie:
Ah I see you've force pushed again - the merge error is here: https://github.com/cylc/cylc-flow/compare/598e0afe6c2e89cbcc6353f48e2363fe2dde34f2..8e7e3f17fc8832aa6ebb9a9dec800ec85d624762#diff-616dbf97f9b89762d7da3666f7ab3abf0cc3f2e6443b5367af9d25439a4321d5R334-R354

dwsutherland:
> Ah I see you've force pushed again - the merge error is here: https://github.com/cylc/cylc-flow/compare/598e0afe6c2e89cbcc6353f48e2363fe2dde34f2..8e7e3f17fc8832aa6ebb9a9dec800ec85d624762#diff-616dbf97f9b89762d7da3666f7ab3abf0cc3f2e6443b5367af9d25439a4321d5R334-R354

Haha.. Apologies, yes, I rebased and did some tests.. will squash as fix to your commit

cylc/flow/cfgspec/workflow.py
dwsutherland: > until their previous (cycle point) instance is satisfied.

Now that I read this again, it's not satisfaction of the previous (cycle point) instance, but the satisfaction of the sequential xtriggers of the previous (cycle point) instance.

dwsutherland:
> @dwsutherland this is now my main focus for review - are you working on conflict resolution, post `cylc set` merge? Ping me if you need any help understanding the changes from that PR.

Yes, working on this now.. thanks will do

cylc/flow/task_pool.py
```diff
@@ -2043,6 +2144,23 @@ def force_trigger_tasks(
...
 
+    def spawn_parentless_sequential_xtriggers(self):
+        """Spawn successor(s) of parentless wall clock satisfied tasks."""
+        while self.xtrigger_mgr.sequential_spawn_next:
+            taskid = self.xtrigger_mgr.sequential_spawn_next.pop()
+            self.xtrigger_mgr.sequential_has_spawned_next.add(taskid)
+            itask = (
+                self._get_task_by_id(taskid)
+                or self._get_task_by_id(taskid)
+            )
```
MetRonnie, line 2155:
```suggestion
            itask = self._get_task_by_id(taskid)
```

MetRonnie (commented):
Just spotted 1 mistake in the conflict resolution but otherwise looks good to me

cylc/flow/task_pool.py
dwsutherland: Ah, silly me.. fixed and amended associated commit

MetRonnie (approved):

changes.d/5738.feat.md
```diff
@@ -0,0 +1 @@
+Added sequentially checked xtriggers option to restrict checking, and the spawning behavior of any associated parentless tasks. `wall_clock` is now sequential by default.
```
hjoliver, line 1: I find that a bit hard to follow. Is this better?:

```suggestion
Optionally spawn parentless xtriggered tasks sequentially - i.e., one at a time, after the previous xtrigger is satisfied, instead of all at once out to the runahead limit.  The `wall_clock` xtrigger is now sequential by default.
```

cylc/flow/cfgspec/workflow.py
```diff
@@ -802,9 +802,21 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+             desc='''
+            If ``True``, tasks that only depend on xtriggers will not spawn
+            until their previous (cycle point) instance is satisfied.
+            Otherwise, they will all spawn at once out to the runahead limit.
+
+            This setting can be overridden by the reserved keyword argument
+            ``sequential`` in individual xtrigger declarations.
+
+            One sequential xtrigger on a parentless task with multiple
+            xtriggers will cause sequential behavior.
```
hjoliver, line 815:
```suggestion
            xtriggers will cause sequential spawning.
```

cylc/flow/cfgspec/workflow.py
```diff
@@ -802,9 +802,21 @@ def get_script_common_text(this: str, example: Optional[str] = None):
                     :ref:`SequentialTasks`.
             ''')
 
+        Conf('spawn from xtriggers sequentially', VDR.V_BOOLEAN, False,
```
hjoliver, line 805: Not sure I like that as a config item name. I understand what it means, but it's long and we haven't described the concept in this way ("spawning *from* an xtrigger") before.  I think a short snappy name is fine, and we can define exactly what it means:
```suggestion
        Conf('sequential xtriggers', VDR.V_BOOLEAN, False,
```

cylc/flow/config.py
```diff
@@ -1705,16 +1705,21 @@ def generate_triggers(self, lexpression, left_nodes, right, seq,
...
+        for label in xtrigs:
             valid, msg = validator(label)
             if not valid:
                 raise WorkflowConfigError(
                     f'Invalid xtrigger name "{label}" - {msg}'
                 )
 
+        if self.xtrigger_mgr is not None:
+            self.xtrigger_mgr.sequential_xtriggers_default = (
+                self.cfg['scheduling']['spawn from xtriggers sequentially']
```
hjoliver, line 1718: (If you agree to my previous suggestion)

```suggestion
                self.cfg['scheduling']['sequential xtriggers']
```

cylc/flow/task_pool.py
```diff
@@ -1913,14 +1978,30 @@ def remove_tasks(self, items):
...
+                )
             self.remove(itask, 'request')
         if self.compute_runahead():
             self.release_runahead_tasks()
         return len(bad_items)
 
     def _get_flow_nums(
             self,
-            flow: List[str],
+            flow: List[Union[str, int]],
```
hjoliver, line 2004: Why is this change needed?

hjoliver (commented):
Part-way through... a few minor comments so far.

cylc/flow/cfgspec/workflow.py
dwsutherland: haha, I love how you critique your own suggestions!

hjoliver: 🤦‍♂️  In that case I prefer my NEW suggestion 😁

cylc/flow/task_pool.py
dwsutherland: Not sure I quite remember, but it may well have been because how it's stored in the DB and how it is internally.. I'll reverse it and see what happens .. pretty sure something failed

hjoliver: It'll need to be reversed in a bunch of places.

hjoliver: Just wondering as I don't think you made any changes to how flow number are handled, on this branch?

cylc/flow/xtrigger_mgr.py
```diff
@@ -186,6 +194,32 @@ class XtriggerManager:
...
+            workflow_x = workflow_state(
+                workflow=other,
+                point=%(task_cycle_point)s,
+                sequential=False
+            ):PT30S
+        [[graph]]
+            PT1H = '''
+                @clock_1 & @workflow_x => foo & bar
+                @wall_clock = baz  # pre-defined zero-offset clock
+            '''
```
hjoliver, line 221: This example is near-identical to the one above. Just delete it and add `sequential=False` to the one above?

cylc/flow/xtrigger_mgr.py
```diff
@@ -217,6 +251,18 @@ def __init__(
...
 
+        # Clock labels, to avoid repeated string comparisons
+        self.wall_clock_labels: Set[str] = set()
+
+        # Workflow wide default, used when not specified in xtrigger kwargs.
+        self.sequential_xtriggers_default = False
+        # Labels whose xtriggers are sequentially checked.
+        self.sequential_xtrigger_labels: Set[str] = set()
+        # Gather parentless tasks whose xtrigger(s) have been satisfied
+        # (these will be used to spawn the next occurance).
```
hjoliver, line 262:
```suggestion
        # (these will be used to spawn the next occurrence).
```

tests/functional/cylc-config/00-simple/section1.stdout
```diff
@@ -6,6 +6,7 @@ hold after cycle point =
 stop after cycle point = 
 cycling mode = integer
 runahead limit = P4
+spawn from xtriggers sequentially = False
```
hjoliver, line 9:
```suggestion
sequential xtriggers = False
```

tests/functional/xtriggers/04-sequential.t
```diff
@@ -0,0 +1,110 @@
...
+
+# Test workflow uses built-in 'echo' xtrigger.
+init_workflow "${TEST_NAME_BASE}" << '__FLOW_CONFIG__'
+[scheduler]
+    cycle point format = %Y
+    allow implicit tasks = True
+[scheduling]
+    initial cycle point = 3000
+    runahead limit = P5
+    spawn from xtriggers sequentially = True
```
hjoliver, line 33:
```suggestion
    sequential xtriggers = True
```

tests/integration/test_sequential_xtriggers.py
```diff
@@ -0,0 +1,250 @@
...
+        async with start(schd):
+            itask = schd.pool.get_task(IntegerPoint('1'), 'foo')
+            assert itask.is_xtrigger_sequential is is_sequential
+
+
+async def test_override(flow, scheduler, start):
+    """Test that the 'sequential=False' arg can override a default of True."""
+    wid = flow({
+        'scheduling': {
+            'spawn from xtriggers sequentially': True,
```
hjoliver, line 227:
```suggestion
            'sequential xtriggers': True,
```

cylc/flow/xtrigger_mgr.py
hjoliver: And this config item doesn't exist (anymore):
`   sequential xtrigger default = True`

dwsutherland: The purpose was to show how you could modify the above example to make the spawning behave differently.. Actually now that `wall_clock` is sequential this doesn't quite achieve that ..

dwsutherland: Have modified it somewhat to emphasize the ways this option can modify the spawning behavior of a similar workflow.

cylc/flow/task_pool.py
dwsutherland: Have reverted everywhere I would have changed it (`scheduler.py`, `task_pool.py`) .. Let's see what happens, fast tests and xtirgger functional passed locally..

hjoliver:
Testing ... looks good so far, nice change.

cylc/flow/task_pool.py
hjoliver: All passed 🎉 - I guess we can go with the reversion and resolve this discussion?

cylc/flow/task_pool.py
```diff
@@ -1665,6 +1727,13 @@ def remove_tasks(self, items):
         """Remove tasks from the pool."""
         itasks, _, bad_items = self.filter_task_proxies(items)
         for itask in itasks:
+            # Spawn next occurance of xtrigger sequential task.
+            if itask.is_xtrigger_sequential:
+                self.spawn_to_rh_limit(
+                    itask.tdef,
+                    itask.tdef.next_point(itask.point),
+                    itask.flow_nums
+                )
```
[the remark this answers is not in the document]
hjoliver: Can this comment be resolved @dwsutherland ?

hjoliver (approved):
Great, thanks @dwsutherland

hjoliver:
@MetRonnie - the only code changes after my review were a type annotation and config item name (which wasn't formally agreed on anyway), so no need for you to re-review.

PR #5740, opened 2023-09-22
client: don't attempt to clean up old contact files in task messages

### LINKED ISSUE #5013: old contact file check
Currently, if a client connection fails we check whether the workflow is still running and remove the "stale" contact file if not. This requires an SSH call.

This check serves two purposes:

* Removes "stale" contact files (which would otherwise trick `cylc scan`).
* Differentiates between communications timeout and workflow stopped. This distinction can be important, e.g. for `cylc tui`.

This check is currently being performed for *all* timeouts including e.g. `cylc message` from a remote host which might not be able to SSH back.

* If the host can SSH back, and determine that the workflow is not running, then it will remove the contact file, removing the source of the problem and preventing the issue from repeating.
* However, if the host cannot SSH back, it cannot determine whether the workflow is still running or not and consequently does not remove the contact file.

As a result, in the following situations *all* client commands will attempt to SSH back home on timeout:

* Workflow has died/shutdown but failed to remove it's remote contact file.
* Network issues are resulting in timeout (i.e. you can ping but not connect).

The SSH'es will fail quickly but may still contribute load and are undesierable. We should find a way to avoid this situation.

**Pull requests welcome!**

oliver-sanders:
Closes #5013

* The `detect_old_contact_file` check serves two purposes:
  1. It removes the contact files of crashed workflows.
  2. It helps the client to differentiate between a communication timeout (e.g. network error or blocked port) and a stopped workflow.
* Remote platforms don't necessarily have SSH access back to the scheduler host which can cause confusing errors.
* Due to the potentially large number of jobs, this check can cause a flurry of SSH'es which isn't great.
* The consequence of this change is that WorkflowStopped errors will be replaced by ClientTimeout errors in job stderr.

Note this applies to the `cylc message` commands built into the job script, but also to any command run within a user's job e.g. `cylc message -- <msg or custom output> `, `cylc broadcast`, etc.

**Check List**

### DIFF
--- a/cylc/flow/network/client.py
+++ b/cylc/flow/network/client.py
@@ -143,6 +143,10 @@ class WorkflowRuntimeClientBase(metaclass=ABCMeta):
                 f'It has moved to {contact_host}:{contact_port}'
             )
 
+        if os.getenv('CYLC_TASK_COMMS_METHOD'):
+            # don't attempt to clean up old contact files in task messages
+            return
+
         # Cannot connect, perhaps workflow is no longer running and is leaving
         # behind a contact file?
         try:


dpmatthews:
This looks sensible to me

hjoliver (approved):
👍 Agreed

PR #5741, opened 2023-09-26
Upper pin urwid

wxtim:
Urwid 2.2.x is causing failures in the remote tests. 

I strongly suspect (but cannot fully evidence) that [this PR](https://github.com/urwid/urwid/commit/8d8e4b678cc0f93149a4a964b26ee11fb136ea0f) is the source of the problem.

**Check List**

- [x] Test dependency change - no tests included.

### DIFF
--- a/.github/workflows/test_functional.yml
+++ b/.github/workflows/test_functional.yml
@@ -62,14 +62,14 @@ jobs:
           # remote platforms
           - name: '_remote_background_indep_poll'
             os: 'ubuntu-latest'
-            python-version: '3.7'
+            python-version: '3.8'
             test-base: 'tests/f tests/k'
             chunk: '1/1'
             platform: '_remote_background_indep_poll _remote_at_indep_poll'
           - name: '_remote_background_indep_tcp'
             os: 'ubuntu-latest'
             test-base: 'tests/f tests/k'
-            python-version: '3.7'
+            python-version: '3.8'
             chunk: '1/1'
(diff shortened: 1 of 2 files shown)


oliver-sanders:
I suspect this is something to do with the way we `pip install coverage after the conda-unpack` which is probably a questionable practice.

oliver-sanders:
Got to the bottom of this!

@wxtim you were so close, it wasn't the version.py file, but is was `setuptools_scm`, the tool which writes the version.py file.

Hopefully they'll release a new build on conda-forge shortly and things will go back to normal.

wxtim:
> Got to the bottom of this!
> 
> @wxtim you were so close, it wasn't the version.py file, but is was `setuptools_scm`, the tool which writes the version.py file.
> 
> Hopefully they'll release a new build on conda-forge shortly and things will go back to normal.

I did mention setuptools_scm - what about it caused the problem?

oliver-sanders:
> what about it caused the problem?

They, hadn't added it as a "host" dependency in the conda-forge recipe. This meant that when the package was installed, the version number was not provided to setuptools, however, the version.py file was present because conda-forge pulls in the released package from pypi. So according to urwid the version was 2.2.2 but according to setuptools the version was None and for some bizzare reason rather than failing on this, they simply set the version to 0.0.0.

The tests now pass with the new conda-forge build pushed out last night:

https://github.com/cylc/cylc-flow/actions/runs/6323378182

PR #5744, opened 2023-09-27
respond to @joehickson comments

wxtim:
Tutorial Changes to go with https://github.com/cylc/cylc-doc/pull/653

- Make order of obs stations match tutorial in docs.
- Make it necessary to carry follow step 2 of the tutorial and insert ICP and FCP.
- Make it necessary to add a jinja2 shebang, as in the tutorial.

IMO we should not merge these PR's until @joehickson and I have had a chance to skim the whole tutorial.

### DIFF
--- a/cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
+++ b/cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
@@ -6,9 +6,9 @@
     # Start the workflow 7 hours before now ignoring minutes and seconds
     # * previous(T-00) takes the current time ignoring minutes and seconds.
     # * - PT7H subtracts 7 hours from the time.
-    initial cycle point = previous(T-00) - PT7H
+    initial cycle point = # Change me
     # Stop the workflow 6 hours after the initial cycle point.
-    final cycle point = +PT6H
+    final cycle point = # Change me
     [[graph]]
         # Repeat every three hours starting at the initial cycle point.
         PT3H = """
@@ -34,10 +34,10 @@
         """
 
 [runtime]
-    [[get_observations_camborne]]
     [[get_observations_heathrow]]
-    [[get_observations_aldergrove]]
+    [[get_observations_camborne]]
     [[get_observations_shetland]]
+    [[get_observations_aldergrove]]
     [[consolidate_observations]]
     [[forecast]]
     [[get_rainfall]]
--- a/cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
+++ b/cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
@@ -3,12 +3,8 @@
     UTC mode = True
 
 [scheduling]
-    # Start the workflow 7 hours before now ignoring minutes and seconds
-    # * previous(T-00) takes the current time ignoring minutes and seconds.
-    # * - PT7H subtracts 7 hours from the time.
-    initial cycle point = # Change me
-    # Stop the workflow 6 hours after the initial cycle point.
-    final cycle point = # Change me
+    # TODO: Set initial cycle point
+    # TODO: Set final cycle point
     [[graph]]
         # Repeat every three hours starting at the initial cycle point.
         PT3H = """
--- a/cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
+++ b/cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
@@ -1,3 +1,4 @@
+#!jinja2
 [scheduler]
     UTC mode = True
 
--- a/cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
+++ b/cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
@@ -29,13 +29,5 @@
         """
 
 [runtime]
-    [[get_observations_heathrow]]
-    [[get_observations_camborne]]
-    [[get_observations_shetland]]
-    [[get_observations_aldergrove]]
-    [[consolidate_observations]]
-    [[forecast]]
-    [[get_rainfall]]
-    [[post_process_exeter]]
 
 %include 'etc/python-job.settings'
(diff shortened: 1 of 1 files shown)


cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
```diff
@@ -6,9 +6,9 @@
     # Start the workflow 7 hours before now ignoring minutes and seconds
     # * previous(T-00) takes the current time ignoring minutes and seconds.
     # * - PT7H subtracts 7 hours from the time.
-    initial cycle point = previous(T-00) - PT7H
+    initial cycle point = # Change me
     # Stop the workflow 6 hours after the initial cycle point.
-    final cycle point = +PT6H
+    final cycle point = # Change me
```
oliver-sanders, line 11: Noooo, this was correct before!

cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
```diff
@@ -34,10 +34,10 @@
         """
 
 [runtime]
-    [[get_observations_camborne]]
     [[get_observations_heathrow]]
-    [[get_observations_aldergrove]]
+    [[get_observations_camborne]]
     [[get_observations_shetland]]
+    [[get_observations_aldergrove]]
```
oliver-sanders, line 40: Random order change?

cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
oliver-sanders, line 11: Looking back at the [Cylc 7 tutorial(https://github.com/metomi/rose/blob/caafa20cb659d3c90000215ddcbc08903a20e211/etc/tutorial/runtime-tutorial/suite.rc#L5-L6) this was supposed to be a couple of TODO comments.

```suggestion
    # TODO - set the initial cycle point.
    # TODO - set the final cycle point.
```

cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
wxtim: No, changed to match the tutorial

cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
wxtim, line 11:
```suggestion
    # TODO: Set initial cycle point
    # TODO: Set final cycle point
```

cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
oliver-sanders: These sections don't serve any purpose (they're empty).

They weren't present in the original tutorial:

https://github.com/metomi/rose/blob/2019.01.x/etc/tutorial/runtime-tutorial/suite.rc

I think they were randomly added when the tutorials were copied over, possibly just to make them validate for the purpose of tests. Suggest reverting to the way it was before and adapting the tests.

oliver-sanders:
(converting to draft until this is completed)

cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
wxtim: I couldn't get the validation tasks to fail when I took them out... so easy!

cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
```diff
@@ -1,14 +1,9 @@
-#!jinja2
```
MetRonnie, line 1: This results in 
```
FileParseError: Invalid line (line 104):
   {% from "sys" import path, executable %}
```

MetRonnie (changes requested):

cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
MetRonnie: Might be worth adding a `.validate` script to `cylc/flow/etc/tutorial/runtime-tutorial` to catch this sort of breakage

wxtim: Sorry @MetRonnie  - this PR ought to be draft - I will come back to it ... eventually ...

wxtim: Sorry, I've just got back to this and I think that this is fine - this config is meant to be broken when it is extracted by students. It's not breaking any automated tests is  it?

oliver-sanders: **Note:** These are tutorial resources, they might not be complete working examples. What's important is that they fit with the context of the tutorials.

This PR reverts a previous attempt to "fix" these resources broke the tutorials.

MetRonnie: The runtime configuration tutorial does not mention adding a Jinja2 shebang. The reason it's needed is because `etc/python-job.settings` contains Jinja2 syntax.

The `.validate` script could just be to run `cylc view` to make sure pre-processing succeeds; it doesn't have to run `cylc validate`

cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
```diff
@@ -1,14 +1,9 @@
...
 [scheduler]
```
wxtim, line 1:
```suggestion
#!jinja2
[scheduler]
```

cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc
wxtim: @MetRonnie  - sorry - you are absolutely right - I had forgotten that reason I'd taken this out was because I had removed the Jinja2 include (replaced with a Cylc include), and not because the tutorial asks students to re-add it.

MetRonnie (approved):

markgrahamdawson (approved):

PR #5747, opened 2023-09-28
Fix.cycle task states query

wxtim:
Fixes a bug reported by a user.

Pagination in Cylc Review works on number of cycles in database up to the limit per page, whether or not those cycles will be displayed!

### Bug description and reproduction

```ini
# Suite.rc
[scheduling]
    cycling mode = integer
    initial cycle point = 0
    runahead limit = P20
    spawn to max active cycle points = true
    [[dependencies]]
        [[[P1]]]
            graph = foo[-P1] => foo
[runtime]
    [[foo]]
        script = sleep 600
```

1. Run this workflow, wait until all the tasks have spawned out to the runahead limit.
2. Open a copy of Cylc Review and click on the link to the Cycles List of this workflow.
3. Try changing the number of entries per page for i in range(21,23). 

### Possible secondary bug (Cylc 8 workflows)

`ready` state has become `preparing`. States `queued`, `runahead`, `retrying` and `submit-retrying` can no longer be identified from the database.

**Check List**

- [x] Tests are not included because this is legacy software. Reviewers should test the main bug as described. n.b. to test explicitly run `/path/to/cylc-7-worktree/bin/cylc review start` to start your own server.

### DIFF
--- a/lib/cylc/review_dao.py
+++ b/lib/cylc/review_dao.py
@@ -117,6 +117,7 @@ class CylcReviewDAO(object):
                 self.daos[key] = CylcSuiteDAO(db_f_name, is_public=True)
                 if os.path.exists(db_f_name):
                     break
+        self.is_cylc8 = self.set_is_cylc8(user_name, suite_name)
         return self.daos[key]
 
     def _db_close(self, user_name, suite_name):
@@ -171,6 +172,15 @@ class CylcReviewDAO(object):
             broadcast_events.append(
                 (time_, change, point, namespace, key, value))
         return broadcast_events
+
+    @staticmethod
+    def set_is_cylc8(user_name, suite_name):
+        from cylc.review import CylcReviewService
+        suite_dir = os.path.join(
+            CylcReviewService._get_user_home(user_name),
+            "cylc-run",
+            suite_name)
+        return CylcReviewService.is_cylc8(suite_dir)
 
     def get_suite_job_entries(
             self, user_name, suite_name, cycles, tasks, task_status,
@@ -228,12 +238,7 @@ class CylcReviewDAO(object):
                 return ([], 0)
         except sqlite3.Error:
             return ([], 0)
-        from cylc.review import CylcReviewService
-        suite_dir = os.path.join(
-            CylcReviewService._get_user_home(user_name),
-            "cylc-run",
-            suite_name)
-        if CylcReviewService.is_cylc8(suite_dir):
+        if self.is_cylc8:
             stmt = (
                 "SELECT" +
                 " task_states.time_updated AS time," +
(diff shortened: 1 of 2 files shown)


lib/cylc/review_dao.py
```diff
@@ -172,6 +173,15 @@ def get_suite_broadcast_events(self, user_name, suite_name):
                 (time_, change, point, namespace, key, value))
         return broadcast_events
 
+    @staticmethod
```
wxtim, line 176: Centralize this logic, we don't want to run it twice if we can help it.

lib/cylc/review_dao.py
```diff
@@ -539,7 +558,11 @@ def get_suite_cycles_summary(
             " sum(" + states_stmt["success"] + ") AS n_success,"
             " sum(" + states_stmt["fail"] + ") AS n_fail"
             " FROM task_states" +
-            " GROUP BY cycle")
+            " GROUP BY cycle" +
+            " HAVING sum(" + states_stmt["active"] + ")>0" +
```
wxtim, line 562: Replaces the logic at line 556 of the old code. Only retrieve info for cycles with non-waiting tasks.

MetRonnie (approved):

lib/cylc/review_dao.py
```diff
@@ -539,7 +558,11 @@ def get_suite_cycles_summary(
...
+            " OR sum(" + states_stmt["success"] + ")>0" +
+            " OR sum(" + states_stmt["fail"] + ")>0"
```
oliver-sanders, line 564: I like the `HAVING`, much nicer than filtering it in the Python layer.

Since we've already computed the counts of active/success/fail task groups above, we might as well reuse them here. I think this does the same thing, but only performs the `sum()` operations once:

```suggestion
            " HAVING n_active > 0" +
            " OR n_success > 0" +
            " OR n_fail >0"
```

oliver-sanders (commented):
Tested with the Cylc 7 example, looks good.

Do you have a Cylc 8 example to trigger this?

oliver-sanders:
Got it:

```diff
     [[dependencies]]
         [[[P1]]]
-            graph = foo[-P1] => foo
+            graph = foo
 [runtime]
     [[foo]]
```

oliver-sanders:
@MetRonnie could you check the latest changes and test with a Cylc 8 workflow (if not already done), subtracting the inter-cycle-dep is enough to replicate. Best avoid needing a final final final final final final Cylc 7 release.

```
export PATH="/path/to/cylc-flow/bin:$PATH"
export PYTHOnPATH="/path/to/cylc-flow/lib"
cylc review start
```

MetRonnie:
Not sure I understand. This is on the deployed version of Cylc Review, can't see the bug here

image

MetRonnie:
I've got this to have lots of waiting tasks
```
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    runahead limit = P26
    [[graph]]
        R1 = foo
        {% for i in range(2, 30) %}
        R1/{{ i }} = foo[-P{{ i - 1 }}] => foo
        {% endfor %}
[runtime]
    [[foo]]
        script = sleep 600
```

But on the current deployed Cylc Review and on this branch I just get the 1 running task showing

MetRonnie:
That setting is obsolete in Cylc 8

oliver-sanders:
Try using Tim's example from the OP for Cylc 7, and remove the inter-cycle dependency for Cylc 8.

You need to use the cycles page and manually specify the number of items to display per page.

MetRonnie (approved):
Ok so this is what I have reproduced in Cylc 7 with your example:
1. Click display options cog and set number of cycles per page to <= 20
2. The running task will no longer be shown

In Cylc 8:
1. Just set the recurrence to `P1 = foo` as Oliver said
2. Set the number of cycles per page to 1
3. No tasks will be shown anymore

Both fixed on this branch

oliver-sanders (approved):

oliver-sanders:
Have been through the same tests, fixed the reported issue, couldn't find anything else amiss.

PR #5748, opened 2023-09-29
clarify SLURM job handler docs

wxtim:
Clarify the formats permitted for directives to SLURM.

Record investigations:

```ini
# part of a flow.cylc file
[[[directives]]]
  -H               # Fails Cylc validation
  -H=              # Works
  --hold           # Fails Cylc validation
  --hold=          # Holds your SLURM job as you'd hope
  -b now+60        # Fails Cylc validation
  -b=now+60        # Sbatch fails with invalid options
  -b now+60=       # Sbatch ... just no
  --begin=now+60   # Works

```

**Check List**

- [x] This is a documentation change only: I have built the documentation locally.

### DIFF
--- a/cylc/flow/job_runner_handlers/slurm.py
+++ b/cylc/flow/job_runner_handlers/slurm.py
@@ -42,7 +42,6 @@ file:
 
    * Cylc requires long form directives (e.g. ``--begin`` not ``-b``).
    * Cylc requires an ``=`` even if the directive does not have a value
-   * Cylc requires an ``=`` even if the directive does not have a value
      (e.g. ``--hold=`` not ``--hold``).
    * If a directive does not have a value you may use the short form
      (e.g. ``-H=``). But the directive must still be suffixed with an ``=``.
--- a/cylc/flow/job_runner_handlers/slurm.py
+++ b/cylc/flow/job_runner_handlers/slurm.py
@@ -40,8 +40,12 @@ file:
 
 .. note::
 
-   Since not all SLURM commands have a short form, cylc requires
-   the long form directives.
+   * Cylc requires long form directives (e.g. ``--begin`` not ``-b``).
+   * Cylc requires an ``=`` even if the directive does not have a value
+   * Cylc requires an ``=`` even if the directive does not have a value
+     (e.g. ``--hold=`` not ``--hold``).
+   * If a directive does not have a value you may use the short form
+     (e.g. ``-H=``). But the directive must still be suffixed with an ``=``.
 
 These are written to the top of the job script like this:
 


cylc/flow/job_runner_handlers/slurm.py
```diff
@@ -40,8 +40,9 @@
 
 .. note::
 
-   Since not all SLURM commands have a short form, cylc requires
-   the long form directives.
+   * Cylc requires long form directives (e.g. ``--begin`` not ``-b``).
```
oliver-sanders, line 43: Is this actually true?

Wouldn't `-b =` work?

wxtim: No, because Slurm won't accept that:

```console

> cat > hello.sh <<__HERE__
+ #!/bin/bash
+ #SBATCH -b=60
+ echo "Hello World"
+ __HERE__
> chmod +x hello.sh 
> sbatch hello.sh 
sbatch: invalid option -- '='
sbatch: invalid option -- '6'
sbatch: invalid option -- '0'
Submitted batch job 3309508
```

`-H =` does work, however.

cylc/flow/job_runner_handlers/slurm.py
```diff
@@ -40,8 +40,9 @@
...
+   * Cylc requires an ``=`` even if the directive does not have a value
+     (e.g. ``--hold=`` not ``--hold``).
```
wxtim, line 45:
```suggestion
   * Cylc requires an ``=`` even if the directive does not have a value
     (e.g. ``--hold=`` not ``--hold``).
   * If a directive does not have a value you may use the short form (e.g. ``-H=``).
     But the directive must still be suffixed with an ``=``.
```

cylc/flow/job_runner_handlers/slurm.py
oliver-sanders: Note, I didn't give an argument with the `-b` option above, just left it empty which should be valid?

E.G:

```
[directrives]
-H =
```
Should result in:
```
#SBATCH -H
```

Which I would expect to work?

oliver-sanders: So Cylc slurm support doesn't support short options which require arguments?

wxtim: > So Cylc slurm support doesn't support short options which require arguments? Because slurm doesn't support the `=` operator with short options?

This

wxtim: > Note, I didn't give an argument with the `-b` option above, just left it empty which should be valid?

Yes, `-b` is, much to my surprise, valid. Pointless, but valid.

cylc/flow/job_runner_handlers/slurm.py
```diff
@@ -40,8 +40,12 @@
 
 .. note::
 
-   Since not all SLURM commands have a short form, cylc requires
-   the long form directives.
+   * Cylc requires long form directives (e.g. ``--begin`` not ``-b``).
+   * Cylc requires an ``=`` even if the directive does not have a value
```
oliver-sanders, line 44:
```suggestion
```

oliver-sanders (approved):

PR #5749, opened 2023-09-30
Kill tasks during job prep.

### LINKED ISSUE #5746: Kill should abort preparing tasks
`cylc kill "foo*"` kills all matching running tasks, but if any `foo*` are preparing they will go on to submit their jobs almost immediately.

This is not what users would expect. Preparing tasks are as good as running. If I issued the kill command 2 seconds later the jobs would have been submitted and then killed.

So, preparing tasks should get "killed" and set to submit-failed (and held, to prevent automatic submit-retries).

It's not clear if this will actually be easy to do.  Several ideas discussed on Element...

hjoliver:
Close #5746 

This PR allows `cylc kill` to abort jobs that are in the job preparation pipeline, resulting in the submit-failed state.

To avoid dealing with the complexities of partial job preparation, I've take the following approach:
- the kill command sets a flag in preparing task proxies
- the job submission process aborts at the last minute, after preparation has completed 
- (so retriggering the task will result in a new submit number).

Tested by deliberately extending job preparation like this (and kill tasks whilst still in the preparing state)

```
platform = $(sleep 10; echo localhost)
```

**Check List**

### DIFF
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -101,6 +101,7 @@ from cylc.flow.task_remote_mgr import (
 from cylc.flow.task_state import (
     TASK_STATUS_PREPARING,
     TASK_STATUS_SUBMITTED,
+    TASK_STATUS_SUBMIT_FAILED,
     TASK_STATUS_RUNNING,
     TASK_STATUS_WAITING,
     TASK_STATUSES_ACTIVE
@@ -187,6 +188,9 @@ class TaskJobManager:
                 itask.state_reset(is_held=True)
                 self.data_store_mgr.delta_task_held(itask)
                 to_kill_tasks.append(itask)
+            elif itask.state(TASK_STATUS_PREPARING):
+                itask.killed_in_job_prep = True
+                LOG.warning(f"[{itask}] killed in prep")
             else:
                 LOG.warning(f"[{itask}] not killable")
         self._run_job_cmd(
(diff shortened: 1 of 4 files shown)


MetRonnie:
After discussing removing a preparing task with Oliver (for #6472), this will probably be even more important as the task might end up submitting and running outside of the task pool? Not sure, haven't tested yet

MetRonnie:
I am happy to take over this if needed, but more likely to end up in 8.4.1 rather than 8.4.0

hjoliver:
Thanks @MetRonnie - yes this one slipped off my radar screen.

IMO this is less important to get into 8.4.0 than trigger-when-paused. It's kind of a bug fix, but fairly niche circumstances.

MetRonnie:
Assigning myself for now. If you get round to re-tackling, then unassign me

cylc/flow/task_job_mgr.py
```diff
@@ -483,6 +487,16 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
                     )
                 continue
 
+            if itask.killed_in_job_prep:
```
MetRonnie, line 490: `itask` is possibly unbound

cylc/flow/task_job_mgr.py
```diff
@@ -483,6 +487,16 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
...
+                itask.waiting_on_job_prep = False
+                itask.killed_in_job_prep = False
+                itask.state_reset(TASK_STATUS_SUBMIT_FAILED)
+                self.data_store_mgr.delta_task_state(itask)
+                itask.local_job_file_path = None  # reset for retry
+                self._prep_submit_task_job_error(
+                    workflow, itask, '(killed in job prep)', '')
```
MetRonnie, line 497: Another problem is this does not get run for tasks that are `cylc remove`d in the preparing state, as they are removed from the pool and do not get to the submission stage

hjoliver:
Superseded by #6535

PR #4844, opened 2022-04-29
Set excecution polling delays correctly

dpmatthews:
These changes partially address #3706

Tested with this workflow:
```
[scheduling]
    [[graph]]
        R1 = nolimit & limit95M & limit1H  & limit10M & limit70S
[runtime]
    [[root]]
        script = "echo Hello"
        execution polling intervals = 3*PT30S, PT10M, PT1H
    [[nolimit]]
    [[limit95M]]
        execution time limit = PT95M
    [[limit1H]]
        execution time limit = PT1H
    [[limit10M]]
        execution time limit = PT10M
    [[limit70S]]
        execution time limit = PT70S
```

Result using master:
```
$ grep "execution timeout" ~/cylc-run/bug.polling-intervals/runN/log/workflow/log
2022-04-29T15:45:33+01:00 INFO - [1/limit10M running job:01 flows:1] health: execution timeout=PT20M, polling intervals=20*PT30S,PT1M,PT2M,PT7M,...
2022-04-29T15:45:33+01:00 INFO - [1/limit1H running job:01 flows:1] health: execution timeout=PT1H10M, polling intervals=3*PT30S,5*PT10M,PT9M30S,PT2M,PT7M,...
2022-04-29T15:45:33+01:00 INFO - [1/limit70S running job:01 flows:1] health: execution timeout=PT11M10S, polling intervals=2*PT30S,PT1M10S,PT2M,PT7M,...
2022-04-29T15:45:33+01:00 INFO - [1/limit95M running job:01 flows:1] health: execution timeout=PT1H45M, polling intervals=3*PT30S,PT10M,PT1H,PT24M30S,PT2M,PT7M,...
2022-04-29T15:45:33+01:00 INFO - [1/nolimit running job:01 flows:1] health: execution timeout=None, polling intervals=3*PT30S,PT10M,PT1H,...
```
e.g. limit10M has `polling intervals=20*PT30S,PT1M,PT2M,PT7M,...` which not what we want.

Result using this branch:
```
$ grep "execution timeout" ~/cylc-run/bug.polling-intervals/runN/log/workflow/log
2022-04-29T15:43:45+01:00 INFO - [1/limit10M running job:01 flows:1] health: execution timeout=PT20M, polling intervals=3*PT30S,PT9M30S,PT2M,PT7M,...
[...]

PR #5751, opened 2023-10-02
Fix execution time limit polling intervals [Cylc 7]

wxtim:
Fixes 2 bugs without issues reports.

1. Backport of [this fix](https://github.com/cylc/cylc-flow/pull/5753) - ticket contains a fuller explanation.
2. Prevent task events manager modifying the list of `execution time limit polling intervals` in `glbl_cfg`, which was causing the first polling interval after the execution time limit to increment every time this code was run.

**Check List**

- [x] Tests are included - I've tested that this now produces more sensible output, but have not formally tested it.

### DIFF
--- a/lib/cylc/task_events_mgr.py
+++ b/lib/cylc/task_events_mgr.py
@@ -993,11 +993,15 @@ class TaskEventsManager(object):
             timeref = itask.summary['started_time']
             timeout_key = 'execution timeout'
             timeout = self._get_events_conf(itask, timeout_key)
+            # delays - All polling times after start
+            # timeout - Total time limit including all polling:
             delays = list(self.get_host_conf(
                 itask, 'execution polling intervals', skey='job',
                 default=[900]))  # Default 15 minute intervals
             if itask.summary[self.KEY_EXECUTE_TIME_LIMIT]:
                 time_limit = itask.summary[self.KEY_EXECUTE_TIME_LIMIT]
+
+                # Get execution time limit polling intervals or set to default:
                 try:
                     host_conf = self.get_host_conf(itask, 'batch systems')
                     batch_sys_conf = host_conf[itask.summary['batch_sys_name']]
@@ -1005,16 +1009,15 @@ class TaskEventsManager(object):
                     batch_sys_conf = {}
                 time_limit_delays = batch_sys_conf.get(
                     'execution time limit polling intervals', [60, 120, 420])
-                timeout = time_limit + sum(time_limit_delays)
-                # Remove excessive polling before time limit
-                while sum(delays) > time_limit:
-                    del delays[-1]
-                # But fill up the gap before time limit
-                if delays:
-                    size = int((time_limit - sum(delays)) / delays[-1])
-                    delays.extend([delays[-1]] * size)
-                time_limit_delays[0] += time_limit - sum(delays)
-                delays += time_limit_delays
+
+                # Total timeout after adding execution time limit polling
+                # intervals:
+                timeout = (time_limit + sum(time_limit_delays))
+
+                delays = self.process_execution_polling_delays(
+                    delays, time_limit_delays, time_limit
+                )
+
         else:  # if itask.state.status == TASK_STATUS_SUBMITTED:
             timeref = itask.summary['submitted_time']
             timeout_key = 'submission timeout'
@@ -1024,13 +1027,16 @@ class TaskEventsManager(object):
                 default=[900]))  # Default 15 minute intervals
         try:
             itask.timeout = timeref + float(timeout)
-            timeout_str = intvl_as_str(timeout)
+            time_limit_str = intvl_as_str(time_limit)
         except (TypeError, ValueError):
             itask.timeout = None
-            timeout_str = None
+            time_limit_str = None
+
         itask.poll_timer = TaskActionTimer(ctx=ctx, delays=delays)
+
         # Log timeout and polling schedule
-        message = 'health check settings: %s=%s' % (timeout_key, timeout_str)
+        message = 'health check settings: %s=%s' % (
+            timeout_key, time_limit_str)
         # Attempt to group identical consecutive delays as N*DELAY,...
         if itask.poll_timer.delays:
             items = []  # [(number of item - 1, item), ...]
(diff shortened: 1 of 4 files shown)


wxtim:
Put back to draft whilst I finish fiddling with the Cylc 8 version. Will have another look after, because I think that I will have a better understanding

lib/cylc/task_events_mgr.py
```diff
@@ -1048,3 +1054,36 @@ def _reset_job_timers(self, itask):
...
+            time_limit: Execution Time Limit.
+
+        Returns: List of delays from start of task.
+        """
+        # Remove excessive polling before time limit
+        while sum(delays) > time_limit:
+            del delays[-1]
+
+        # But fill up the gap before time limit
+        if delays:
```
oliver-sanders, line 1076: In the Cylc 8 PR, this logic was:

```python
        if sum(delays) > time_limit:
            # Remove execution polling which would overshoot the
            # execution time limit:
            while sum(delays) > time_limit:
                del delays[-1]
        elif delays:
```

What's the reason for the difference?

lib/cylc/task_events_mgr.py
```diff
@@ -1048,3 +1054,36 @@ def _reset_job_timers(self, itask):
...
+    def process_execution_polling_delays(
+        delays, time_limit_delays, time_limit
+    ):
+        """Create list of intervals after starting at which to poll a task.
+
+        Args:
+            delays: Input is Execution Polling intervals.
+            time_limit_delays: Execution Time Limit Polling Intervals.
+            time_limit: Execution Time Limit.
+
```
oliver-sanders, line 1068: Would be good to copy across the doctests from the Cylc 8 PR for piece of mind.

lib/cylc/task_events_mgr.py
```diff
@@ -993,28 +993,31 @@ def _reset_job_timers(self, itask):
...
-                # But fill up the gap before time limit
-                if delays:
-                    size = int((time_limit - sum(delays)) / delays[-1])
-                    delays.extend([delays[-1]] * size)
-                time_limit_delays[0] += time_limit - sum(delays)
-                delays += time_limit_delays
+
+                # Total timeout after adding execution time limit polling
+                # intervals:
+                timeout = (time_limit + sum(time_limit_delays))
```
oliver-sanders, line 1015: This line seems to be missing from the Cylc 8 branch?

lib/cylc/task_events_mgr.py
wxtim: Because @dpmatthews  added that logic in Cylc 8, but not Cylc 7 in this pr https://github.com/cylc/cylc-flow/pull/4844/files

I'm going to pull that logic in here.

lib/cylc/task_events_mgr.py
wxtim: It was removed here: https://github.com/cylc/cylc-flow/pull/4863 - @dpmatthews  is it sane to backport this change?

dpmatthews: Actually, this bit is a change in behaviour rather than a bug fix. I suggest we leave it unchanged in Cylc 7.

lib/cylc/task_events_mgr.py
```diff
@@ -1048,3 +1054,63 @@ def _reset_job_timers(self, itask):
...
+            size = int((time_limit - sum(delays)) / delays[-1])
+            delays.extend([delays[-1]] * size)
+
+        # After the last delay before the execution time limit add the
+        # delay to get to the execution_time_limit
+        if len(time_limit_delays) > 1:
+            time_limit_delays[0] += time_limit - sum(delays)
+        else:
+            delays.append(
+                time_limit_delays[0] + time_limit - sum(delays))
```
MetRonnie, line 1113: In the Cylc 8 PR this was simplified to
```suggestion
        if len(time_limit_delays) == 1:
            time_limit_delays.append(time_limit_delays[0])
        time_limit_delays[0] += time_limit - sum(delays)
```

lib/cylc/task_events_mgr.py
```diff
@@ -1024,13 +1027,16 @@ def _reset_job_timers(self, itask):
...
             itask.timeout = None
-            timeout_str = None
+            time_limit_str = None
+
         itask.poll_timer = TaskActionTimer(ctx=ctx, delays=delays)
+
         # Log timeout and polling schedule
-        message = 'health check settings: %s=%s' % (timeout_key, timeout_str)
+        message = 'health check settings: %s=%s' % (
+            timeout_key, time_limit_str)
```
MetRonnie, line 1039: I don't get the reason for this change

MetRonnie (commented):
> 2\. Prevent task events manager modifying the list of `execution time limit polling intervals` in `glbl_cfg`, which was causing the first polling interval after the execution time limit to increment every time this code was run.

I'm not seeing how the changes here address this, could you explain?

lib/cylc/task_events_mgr.py
wxtim, line 1039: Nor do I
```suggestion
        message = 'health check settings: %s=%s' % (timeout_key, time_limit_str)
```

lib/cylc/task_events_mgr.py
```diff
@@ -1013,7 +1014,7 @@ def _reset_job_timers(self, itask):
                 # Total timeout after adding execution time limit polling
                 # intervals:
                 timeout = (time_limit + sum(time_limit_delays))
-
+                import pdb; pdb.set_trace()
```
MetRonnie, line 1017: 🐞

wxtim:
```suggestion
```

wxtim: Oops/

lib/cylc/task_events_mgr.py
MetRonnie: @wxtim You marked this as resolved but didn't elucidate?

wxtim: I changed it back - this diff is marked outdated.

Unfortunately the actual reason for this change is that changing `timeout_str` to `time_limit_str` (personal preference, and timeout being used elsewhere for other purposes) has made this line too long.

MetRonnie: I'm not talking about the line wrapping - I'm talking about `timeout_str -> time_limit_str`

wxtim: I agree - I was trying to change the names to make it more readable, but I think I made a mistake with this.

MetRonnie (approved):
Tested out; fixes both bugs

oliver-sanders (approved):

PR #5752, opened 2023-10-02
Improved `cylc release` task matching.

### LINKED ISSUE #5750: `cylc release`: only need to match in the tasks-to-hold list
Currently the hold/release commands both work like this:

- glob pattern args:
   - find matching task proxies in the task pool
   - add/remove their IDs from the tasks-to-hold list
   - action the hold/release on the matching task proxies
- specific task ID args:
  - add/remove the IDs from tasks-to-hold list
  - action the hold/release for any of those that exist in the task pool

HOWEVER, we can only release tasks that are already held, so "release" actually only needs to look at the tasks-to-hold list:

- glob pattern or specific ID:
  - find matching IDs in the tasks-to-hold list, and remove them
  - action the release for any of those that exist in the task pool

NOTE this allows releasing a glob-hold by using the same glob, even if it no longer matches in the task pool - which can be important in partial graph rewind scenarios. [I have a real example, which motivated this Issue].

hjoliver:
The release command doesn't need to do generic matching of future tasks, because we can only release tasks that are already held and those are all recorded in the tasks-to-hold list.

And that being the case, we can glob-match in both the task pool and the tasks-to-hold list, to identify tasks to release.

Close #5750 

**Check List**


hjoliver:
Reviewers, an easy way to test this branch:
```ini
[scheduling]
   cycling mode = integer
   initial cycle point = 1
   final cycle point = 3
   [[graph]]
      P1 = """
         holder => a & b & c => d & e & f
      """
[runtime]
   [[holder]]
      script = """
         P=${CYLC_TASK_CYCLE_POINT}
         cylc hold ${CYLC_WORKFLOW_ID} //$P/a //$P/b //$P/c //$P/d //$P/e //$P/f
      """
   [[FAM, BAM]]
   [[a,b]]
      inherit = FAM
   [[d,e]]
      inherit = BAM
   [[c,f]]
```
So, `a, b, c` end up held in the task pool, and `d, e, f` flagged as future tasks to be held, in every cycle.  

Now mess about with `cylc release` globs to release 'em.

oliver-sanders:
> Bonus - cylc remove now discards removed tasks from the tasks-to-hold list.

Good spot!

cylc/flow/config.py
```diff
@@ -573,6 +573,10 @@ def __init__(
 
         self.mem_log("config.py: end init config")
 
+        self.namespaces: Dict[str, List[str]] = {}
+        for name in self.taskdefs:
+            self.namespaces[name] = self.taskdefs[name].namespace_hierarchy
+
```
oliver-sanders, line 579: Don't we already have this info in the config somewhere?

cylc/flow/id_match.py
```diff
@@ -234,6 +240,102 @@ def filter_ids(
     return ret, _not_matched
 
 
+def filter_ids_2(
+    xpool: 'Set[Tuple[str, PointBase]]',
```
oliver-sanders, line 244: Two use cases:

* Matching tasks in a pool.
* Matching tasks in a hold list.

Both use cases are essentially looping over a list and looking for matching entries.

If we reformatted the hold-list to match the format of the pool, could we use the one method for both cases to avoid adding a new implementation?

cylc/flow/task_pool.py
```diff
@@ -780,6 +789,8 @@ def remove(self, itask, reason=""):
                 del self.hidden_pool[itask.point]
             LOG.debug(f"[{itask}] {msg}")
             self.task_queue_mgr.remove_task(itask)
+            self.tasks_to_hold.discard((itask.tdef.name, itask.point))
+            LOG.info(f"Releasing held task {itask}")
```
oliver-sanders, line 793: This message does kinda sound like we should expect the task to run.

Perhaps "Removing held task"?

cylc/flow/id_match.py
hjoliver: (Yes, I meant to come back to this myself, will take another look...)

cylc/flow/task_pool.py
hjoliver: Good spot, pretty sure I meant to type "Removing" there!

cylc/flow/config.py
hjoliver: Yes, got it.

hjoliver:
(I've just come back to this, will un-draft when done)

cylc/flow/id_match.py
hjoliver: Done.

hjoliver:
Ready for review (if tests are good). I've pushed it forward to 8.3.0 - not critical but useful and probably to review and test.

cylc/flow/id_match.py
```diff
@@ -35,54 +33,18 @@
...
-#     'Tuple[List[TaskProxy], List[str]]'
-#     ', '
-#     'Tuple[List[PointBase], List[str]]'
-#     ']'
-# )
-
-
 def filter_ids(
-    pool: 'Pool',
+    pool: Union['Pool', Dict['PointBase', Dict[str, 'PseudoTaskProxy']]],
```
oliver-sanders, line 42: I'm not overly keen on letting these posers mingle with the real deal in the pool as they could get sucked through into other code where they don't belong causing runtime errors. Also could they interfere with spawning logic or cause flow merging?

Perhaps we could stick these into a special held pool?

hjoliver: > I'm not overly keen on letting these posers mingle with the real deal in the pool

Firstly, I did that in response to your earlier review comment above 😁 :

> *If we reformatted the hold-list to match the format of the pool*, could we use the one method for both cases to avoid adding a new implementation?

The problem is `filter_ids` works on a task pool, whereas the hold list - because it also holds future tasks - has just `(task-name, cycle-point)` tuples.

Secondly, there's no "mixing" going on. The `filter_ids` method just iterates over the pool to extract information. It can return a list of tasks, the type of which depends on whether we're matching the active pool or future tasks, so the return value has to be treated accordingly. 

But the pseudo task proxies are only generated, for this specific purpose, by an explicit call to the task hold manager, so I don't see any reason to be concerned about them leaking into the real task pool somehow. And they can't interfere with spawning logic or anything else if they're not in the task pool.

cylc/flow/task_hold_mgr.py
```diff
@@ -0,0 +1,171 @@
...
+    def release_future_tasks(
+        self, ftasks: List['PseudoTaskProxy']
+    ) -> None:
+        """Release future tasks from the hold store."""
+        matched = set()
+        for ftask in ftasks:
+            name = ftask.tdef.name
+            cycle = ftask.point
+            matched.add((ftask.tdef.name, ftask.point))
+            self.data_store_mgr.delta_task_held((name, cycle, False))
```
dwsutherland, line 122: Ah, nice, so this is why they now appear released at n>0:
image

dwsutherland (approved):
Yeah, works for me:
**before**
```
2024-04-08T20:22:07+12:00 INFO - Command "release" received. ID=4294af93-ed58-461f-b262-b432641f4fc7
    release(tasks=['1/d'])
2024-04-08T20:22:08+12:00 INFO - Command "release" actioned. ID=4294af93-ed58-461f-b262-b432641f4fc7
2024-04-08T20:22:57+12:00 INFO - Command "release" received. ID=08777829-a3c4-458e-9ef8-b7eeb4fa0529
    release(tasks=['*/d'])
2024-04-08T20:22:58+12:00 WARNING - No active tasks matching: */d
2024-04-08T20:22:58+12:00 INFO - Command "release" actioned with 1 warnings. ID=08777829-a3c4-458e-9ef8-b7eeb4fa0529
```
**after**
```
2024-04-08T20:25:28+12:00 INFO - Command "release" received. ID=0f774513-6307-437d-890a-807c846ea142
    release(tasks=['1/d'])
2024-04-08T20:25:29+12:00 INFO - Task hold list:
     * 3/b
     * 2/f
     * 1/b
     * 2/e
     * 1/e
     * 3/e
     * 3/c
     * 3/a
     * 3/d
     * 3/f
     * 2/c
     * 2/a
     * 2/d
     * 1/c
     * 1/a
     * 1/f
     * 2/b
2024-04-08T20:25:29+12:00 INFO - Command "release" actioned. ID=0f774513-6307-437d-890a-807c846ea142
2024-04-08T20:26:11+12:00 INFO - Command "release" received. ID=256e25a9-ff87-4639-a1d1-3aecc70c1263
    release(tasks=['*/d'])
2024-04-08T20:26:11+12:00 INFO - Task hold list:
     * 3/b
     * 2/f
     * 1/b
     * 2/e
     * 1/e
     * 3/e
     * 3/c
     * 3/a
     * 3/f
     * 2/c
     * 2/a
     * 1/c
     * 1/a
     * 1/f
     * 2/b
```
quite verbose .. but not really an issue I suppose (?)

hjoliver:
> quite verbose .. but not really an issue I suppose, maybe for large workflows it might be a bit much (?)..

We don't have another way to expose the complete list of all held and to-be-held tasks to users yet, so I figured the log was best for the moment. Could put it to DEBUG mode I guess.

PR #5753, opened 2023-10-04
Fix execution time limit polling intervals [cylc 8]

wxtim:
Response to a bug not otherwise recorded:

### Bug description

If the sum of `execution polling intervals` < `execution time limit` the remainder should be added to the first, but only the first `execution time limit polling interval`. Currently the remainder is always added to the first items in the list of `execution time limit polling intervals`.

The last item in the list of `execution time limit polling intervals` is repeated until the final itask timeout. At the moment if we set only one item in `execution time limit polling intervals` then every polling interval is `execution time limit polling intervals[0] + remainder`.

#### Example

```
execution_time_limit = 60
execution_polling_intervals = [50]
execution time limit polling intervals = [10]

list_of_polling_intervals_is = [50, 20] # polls onwards at intervals of 20s ...
list_of_polling_intervals_should_be = [50, 20, 10]   # polls onwards at intervals of 10s ...

```

**Check List**

### DIFF
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -1653,12 +1653,11 @@ class TaskEventsManager():
             # execution time limit:
             while sum(delays) > time_limit:
                 del delays[-1]
-        else:
+        elif delays:
             # Repeat the last execution polling interval up to the execution
             # time limit:
-            if delays:
-                size = int((time_limit - sum(delays)) / delays[-1])
-                delays.extend([delays[-1]] * size)
+            size = int((time_limit - sum(delays)) / delays[-1])
+            delays.extend([delays[-1]] * size)
 
         # After the last delay before the execution time limit add the
         # delay to get to the execution_time_limit
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -1662,11 +1662,11 @@ class TaskEventsManager():
 
         # After the last delay before the execution time limit add the
         # delay to get to the execution_time_limit
-        if len(time_limit_polling_intervals) > 1:
-            time_limit_polling_intervals[0] += time_limit - sum(delays)
-        else:
-            delays.append(
-                time_limit_polling_intervals[0] + time_limit - sum(delays))
+        if len(time_limit_polling_intervals) == 1:
+            time_limit_polling_intervals.append(
+                time_limit_polling_intervals[0]
+            )
+        time_limit_polling_intervals[0] += time_limit - sum(delays)
 
         # After the execution time limit poll at execution time limit polling
         # intervals.
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -33,7 +33,7 @@ import os
 from shlex import quote
 import shlex
 from time import time
-from typing import TYPE_CHECKING, Optional, Union, cast
+from typing import TYPE_CHECKING, List, Optional, Union, cast
 
 from cylc.flow import LOG, LOG_LEVELS
 from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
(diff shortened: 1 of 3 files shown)


cylc/flow/task_events_mgr.py
```diff
@@ -1557,24 +1557,23 @@ def _reset_job_timers(self, itask):
...
-                if sum(delays) > time_limit:
-                    # Remove excessive polling before time limit
-                    while sum(delays) > time_limit:
-                        del delays[-1]
-                else:
-                    # Fill up the gap before time limit
-                    if delays:
-                        size = int((time_limit - sum(delays)) / delays[-1])
-                        delays.extend([delays[-1]] * size)
-                time_limit_delays[0] += time_limit - sum(delays)
```
wxtim, line 1576: The last member of the list is repeated.

if `len(time_limit_delays) == 1` then all subsequent delays are not the set execution time limit polling intervals, but that value + the remainder of the execution polling intervals under the time limit.

cylc/flow/task_events_mgr.py
```diff
@@ -1608,6 +1607,73 @@ def _reset_job_timers(self, itask):
...
+                size = int((time_limit - sum(delays)) / delays[-1])
+                delays.extend([delays[-1]] * size)
+
+        # After the last delay before the execution time limit add the
+        # delay to get to the execution_time_limit
+        if len(time_limit_polling_intervals) > 1:
+            time_limit_polling_intervals[0] += time_limit - sum(delays)
+        else:
+            delays.append(
+                time_limit_polling_intervals[0] + time_limit - sum(delays))
```
wxtim, line 1669: This is the bug fix - the rest is refactoring for easier testing.

```diff
- time_limit_polling_intervals[0] += time_limit - sum(delays)
```

cylc/flow/task_events_mgr.py
```diff
@@ -1608,6 +1607,72 @@ def _reset_job_timers(self, itask):
...
```
MetRonnie, line 1669: A bit simpler to do it this way? Take it or leave it
```suggestion
        if len(time_limit_polling_intervals) == 1:
            time_limit_polling_intervals.append(
                time_limit_polling_intervals[0]
            )
        time_limit_polling_intervals[0] += time_limit - sum(delays)
```

MetRonnie (approved):
I haven't manually tested but I'm happy that the tests are comprehensive enough

cylc/flow/task_events_mgr.py
```diff
@@ -1608,6 +1607,72 @@ def _reset_job_timers(self, itask):
...
+            # Remove execution polling which would overshoot the
+            # execution time limit:
+            while sum(delays) > time_limit:
+                del delays[-1]
+        else:
+            # Repeat the last execution polling interval up to the execution
+            # time limit:
+            if delays:
+                size = int((time_limit - sum(delays)) / delays[-1])
+                delays.extend([delays[-1]] * size)
```
oliver-sanders, line 1661: Might as well collapse these switches:

```suggestion
        elif delays:
            # Repeat the last execution polling interval up to the execution
            # time limit:
            size = int((time_limit - sum(delays)) / delays[-1])
            delays.extend([delays[-1]] * size)
```

cylc/flow/task_events_mgr.py
```diff
@@ -1608,6 +1607,72 @@ def _reset_job_timers(self, itask):
...
+            (execution) time_limit
+            (execution) time_limit_polling_intervals
+
+        Examples:
+
+        >>> this = TaskEventsManager.process_execution_polling_intervals
+
+        # Basic example:
+        >>> this([40, 35], 100, [10])
+        [40, 35, 35, 10]
```
oliver-sanders, line 1629: I'm probably wrong, but I would have expected the output to be:

```
[40, 35, 25, 10]
```

Because:

* The polling interval was set as 40, 35
* To get from the last poll to the ETL we need to add 25 seconds - `100 - (40+35) = 25`
* The post ETL interval was defined as 10.

Which makes sense to me from the definition of execution time limit polling intervals:

> List of intervals after execution time limit to poll jobs.
>
> -- https://cylc.github.io/cylc-doc/stable/html/reference/config/global.html#global.cylc[platforms][%3Cplatform%20name%3E]execution%20time%20limit%20polling%20intervals

wxtim: According to Dave we don't poll on the ETL. I did ask

oliver-sanders: Gotcha.

oliver-sanders (approved):

PR #5755, opened 2023-10-04
opt outputs: submitted is implicitly required

oliver-sanders:
* This fixes a bug where submit-failed tasks were incorrectly identified as complete.
* E.G. `foo:finish => bar`.
* A task is incomplete if:
  * it finished executing without completing all required outputs
  * or if job submission failed and the :submit output was not optional
* See: https://github.com/cylc/cylc-admin/blob/master/docs/proposal-new-output-syntax.md#output-syntax

**Check List**

### DIFF
--- a/cylc/flow/task_outputs.py
+++ b/cylc/flow/task_outputs.py
@@ -79,7 +79,7 @@ class TaskOutputs:
         if (
             # "submitted" is not declared as optional/required
             tdef.outputs[TASK_OUTPUT_SUBMITTED][1] is None
-            # and "submit-failed" is explicitly declared as optional/required
+            # and "submit-failed" is not declared as optional/required
             and tdef.outputs[TASK_OUTPUT_SUBMIT_FAILED][1] is None
         ):
             self._add(
--- a/cylc/flow/task_outputs.py
+++ b/cylc/flow/task_outputs.py
@@ -70,9 +70,23 @@ class TaskOutputs:
         self._by_message = {}
         self._by_trigger = {}
         self._required = set()
+
         # Add outputs from task def.
         for trigger, (message, required) in tdef.outputs.items():
             self._add(message, trigger, required=required)
+
+        # Handle implicit submit requirement
+        if (
+            # "submitted" is not declared as optional/required
+            tdef.outputs[TASK_OUTPUT_SUBMITTED][1] is None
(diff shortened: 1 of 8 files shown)


cylc/flow/task_outputs.py
```diff
@@ -70,10 +70,24 @@ def __init__(self, tdef):
...
         for trigger, (message, required) in tdef.outputs.items():
             self._add(message, trigger, required=required)
 
+        # Handle implicit submit requirement
+        if (
+            # "submitted" is not declared as optional/required
+            tdef.outputs[TASK_OUTPUT_SUBMITTED][1] is None
+            # and "submit-failed" is explicitly declared as optional/required
+            and tdef.outputs[TASK_OUTPUT_SUBMIT_FAILED][1] is None
+        ):
```
hjoliver, line 84: The comments don't match the code here (in the second clause).  The logic seems to work in all cases, from testing, but I'm not quite satisfied that I've thought it through fully yet... [coming back after dinner].

hjoliver: I guess the overall logic should be this:
- if neither `:submitted` nor `:submit-failed` are declared (as required or optional) then `:submitted` (required) is implicit and must be added
- otherwise if any of `:submitted` or `:submitted?` or `:submit-failed` or `:submit-failed?` are declared, then the opposite is implied, and there's no need to add implicit `:submitted`

Is that right?

wxtim: I think that's what the logic now does, but the comments need updating.

oliver-sanders: Yes, sorry, the comment is wrong, but the logic is right.

```suggestion
        if (
            # "submitted" is not declared as optional/required
            tdef.outputs[TASK_OUTPUT_SUBMITTED][1] is None
            # and "submit-failed" is not declared as optional/required
            and tdef.outputs[TASK_OUTPUT_SUBMIT_FAILED][1] is None
        ):
```

| submitted | submit-failed | implicit submit dependence |
| --- | --- | --- |
| optional | optional | False |
| optional |  | False |
|  | optional | False |
|  |  | **True** |
| required | required | False |
| required |  | False |
| | required | False |

Note it is not possible to mix "required" and "optional" because the graph parser will raise an error.

wxtim (approved):

Issue #5566, opened 2023-06-02
event emails should include associated task messages


hjoliver:
REF: https://cylc.discourse.group/t/sending-custom-emails-from-cylc-workflow/665

```ini
[scheduling]
   [[graph]]
      R1 = "foo"
[runtime]
   [[foo]]
       script = """
          ON_THE_FLY=$(date)
          cylc message WARNING:"My birthday is ${ON_THE_FLY}."
       """
       [[[events]]]
           mail events = warning
```

Running this results in an email to the user, for the warning event, but it does not include the associated task message.

The only way to get custom content into the email seems to be `[scheduler][mail]footer` which allows templating of workflow-level variables (so not task-level `%(message)s` etc.

Fixing this is probably complicated by event-batching for emails.

oliver-sanders:
The message is not present in the `id_key` which means this missing context would need to be added. This is a right pain due to the number of functions which use this key with tuple unpacking, so would be broken by any change to its format. 

I think this `id_key` logic should get straightened out either way, suggest using a named tuple to carry the context info and passing the `Tokens` object to avoid repeated ID formatting, then adding the message into the named tuple. This should be an 8.3.0 change as the risk of breakages is quite high.

oliver-sanders:
Assigning myself as I've been having a poke around the id_key problem.

Issue #5183, opened 2022-10-07
cache Tokens methods


oliver-sanders:
The `detokenise` function is fairly quick, ~1e-05 on my machine, however, it does get called an ungodly amount of times, ~1e+05 for a small cycling workflow run for a few cycles.

Part of the solution is to pass around Tokens objects rather than string IDs where possible to avoid tokenising/detokenising where not necessary. This should be a long-term goal.

The qiucker part of the solution is to lazy-cache the three properties of `cylc.flow.id.Tokens` which call `detokenise`. Note that `Tokens` instances are dynamic (i.e. we can set properties) so the cache would need to be wiped whenever any properties are set.

Note the `functools.cached_property` decorator is only avoidable for Python 3.8+ and not a drop-in solution because it's a dynamic object.

**Pull requests welcome!**

oliver-sanders:
It looks like we've managed to avert the scaling issues in #3515 without caching in the tokens layer, so pushing back the milestone as no longer critical. We can bump this forward again if we have a solution ready.

MetRonnie:
With a quick + dirty go at caching
```diff
$ git diff 8.1.x -U1
diff --git a/cylc/flow/id.py b/cylc/flow/id.py
index 9c3a56e30..e7c998deb 100644
--- a/cylc/flow/id.py
+++ b/cylc/flow/id.py
@@ -22,2 +22,3 @@ This module contains the abstract ID tokenising/detokenising code.
 from enum import Enum
+from functools import lru_cache
 import re
@@ -153,13 +154,4 @@ class Tokens(dict):
 
-    def __eq__(self, other):
-        return all(
-            self[key] == other[key]
-            for key in self._KEYS
-        )
-
-    def __ne__(self, other):
-        return any(
-            self[key] != other[key]
-            for key in self._KEYS
-        )
+    def __hash__(self):
+        return hash(tuple(self.get(key) for key in Tokens._KEYS))
 
@@ -552,4 +544,4 @@ def _dict_strip(dictionary):
     return {
-        key: value.strip() if value else None
-        for key, value in dictionary.items()
+        key: value.strip() for key, value in dictionary.items()
+        if value
     }
@@ -603,2 +595,3 @@ def legacy_tokenise(identifier: str) -> Tokens:
 
+@lru_cache(1000)
 def tokenise(
@@ -677,2 +670,3 @@ def tokenise(
 
+@lru_cache(1000)
 def detokenise(
```

```
[...]

PR #5769, opened 2023-10-12
events: refactor key and make the event message available to handlers

oliver-sanders:
#5566 turned out to be remarkably awkward to solve, as a result there are two other changes on this PR.

* Tokens objects are now immutable.
  * Partially addresses #5183
  * This means we can use them as dict keys and in sets which is greatly advantageous.
  * It also opens the door to caching as needed.
* Task events now use a named tuple and the fields have been rationalised.
  * Closes https://github.com/cylc/cylc-flow/issues/5566
  * The separate cycle/task/job fields have been combined into tokens.
  * The `key1` field has been flattened out.
* The message field is now available to event handlers.
  * This means we can get access to task messages.
  * There is a caveat attached to this, I haven't added a message column to the DB. This would be awkward as the table is shared with task action timers, etc.
  * This means that if an event handler is actioned just before a workflow shuts down, the exact message will be lost when the workflow restarts.
  * In this situation, it defaults to the event name, which might be INFO or custom for a task message.
* Fix a remote test which doesn't get run on CI and became broken by another change.
* Fix an email field which was never present.
* Change the event email format a tad to allow multiple event messages per email.

**Check List**

### DIFF
--- a/cylc/flow/id.py
+++ b/cylc/flow/id.py
@@ -74,7 +74,7 @@ class Tokens(dict):
         <id: w//c>
         >>> Tokens(workflow='w', cycle='c')['job']
 
-        # Make a copy (note Tokens are mutable):
+        # Make a copy (note Tokens are immutable):
         >>> tokens.duplicate()
         <id: ~u/w//c/t/01>
         >>> tokens.duplicate(job='02')  # make changes at the same time
@@ -118,9 +118,10 @@ class Tokens(dict):
         dict.__init__(self, **kwargs)
 
     def __setitem__(self, key, value):
-        if key not in self._KEYS:
-            raise ValueError(f'Invalid token: {key}')
-        dict.__setitem__(self, key, value)
+        raise Exception('Tokens objects are not mutable')
+
+    def update(self, other):
+        raise Exception('Tokens objects are not mutable')
 
     def __getitem__(self, key):
         try:
@@ -150,6 +151,9 @@ class Tokens(dict):
         else:
             id_ = self.id
         return f'<id: {id_}>'
+
+    def __hash__(self):
+        return hash(tuple(self.values()))
 
     def __eq__(self, other):
         if not isinstance(other, self.__class__):
@@ -336,11 +340,9 @@ class Tokens(dict):
             >>> tokens = Tokens()
             >>> tokens.is_null
             True
-            >>> tokens['job_sel'] = 'x'
-            >>> tokens.is_null
+            >>> tokens.duplicate(job_sel='x').is_null
             True
-            >>> tokens['job'] = '01'
-            >>> tokens.is_null
+            >>> tokens.duplicate(job='01').is_null
             False
 
         """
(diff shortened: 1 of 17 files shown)


.github/workflows/test_functional.yml
```diff
@@ -108,7 +108,7 @@ jobs:
         run: |
           # install system deps
           brew update
-          brew install bash coreutils gnu-sed
+          brew install bash coreutils gnu-sed grep
```
oliver-sanders, line 111: Install GNU grep.

oliver-sanders:
(draft until I can confirm there are no glaring coverage holes)

cylc/flow/id.py
```diff
@@ -74,7 +74,7 @@ class Tokens(dict):
         <id: w//c>
         >>> Tokens(workflow='w', cycle='c')['job']
 
-        # Make a copy (note Tokens are mutable):
+        # Make a copy (note Tokens are imutable):
```
wxtim, line 77:
```suggestion
        # Make a copy (note Tokens are immutable):
```

.github/workflows/test_functional.yml
wxtim: I think this may not work out of the box: I think you'll want to use `ggrep`: https://xebia.com/blog/learning-grep/

dwsutherland (approved):
LGTM 👍   .. (tests pending)

cylc/flow/id.py
```diff
@@ -159,6 +163,12 @@ def __eq__(self, other):
             for key in self._KEYS
         )
 
+    def __lt__(self, other):
+        return self.id < other.id
```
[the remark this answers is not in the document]
oliver-sanders: I think something was sorting tokens somewhere, possibly to get cycles/tasks in order in emails.

cylc/flow/task_events_mgr.py
```diff
@@ -867,35 +958,44 @@ def _process_event_email(self, schd: 'Scheduler', ctx, id_keys) -> None:
...
+            stdin_str += f'job: {id_key.tokens.relative_id}\n'
+            stdin_str += f'event: {id_key.event}\n'
+            stdin_str += f'message: {id_key.message}\n\n'
+
         # STDIN for mail, event info + workflow detail
         stdin_str += "\n"
-        for key in (
-            WorkflowEventData.Workflow.value,
-            WorkflowEventData.Host.value,
-            WorkflowEventData.Port.value,
```
oliver-sanders, line 884: Fix a small bug where the port was never provided to the event handler as it had changed from `Scheduler.port` to `Scheduler.server.port`.

cylc/flow/task_events_mgr.py
```diff
@@ -104,6 +114,63 @@
...
+    Warning: This information is not currently preserved in the DB so will be
+    lost on restart.
+    """
+    message: str
+
+    """The job tokens."""
+    tokens: 'Tokens'
+
+
+def get_event_id(event: str, itask: 'TaskProxy') -> str:
```
oliver-sanders, line 140: This was pre-existing code which I centralised.

cylc/flow/task_events_mgr.py
```diff
@@ -867,35 +958,44 @@ def _process_event_email(self, schd: 'Scheduler', ctx, id_keys) -> None:
...
         cmd.append(ctx.mail_to)
+
         # STDIN for mail, tasks
         stdin_str = ""
         for id_key in sorted(id_keys):
-            (_, event), point, name, submit_num = id_key
-            stdin_str += "%s: %s/%s/%02d\n" % (event, point, name, submit_num)
+            stdin_str += f'job: {id_key.tokens.relative_id}\n'
+            stdin_str += f'event: {id_key.event}\n'
+            stdin_str += f'message: {id_key.message}\n\n'
```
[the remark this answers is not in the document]
wxtim: Will check, but both the source and the output I'm expecting look more readable.

wxtim:
90% Happy

Discussed example

```
#!jinja2
[meta]
    description = This is a workflow created by a script..
    created on = 2024-1-4T9:58

[scheduler]
    allow implicit tasks = True
    cycle point format = %Y
    [[events]]
        mail events = warning
    [[mail]]
        footer = The event was %(event)s The message was \n%(message)s

[scheduling]
    initial cycle point = 2469
    # final cycle point = 2469
    [[graph]]
        R1 = foo

[runtime]
    [[root]]
        script = """
            cylc message -- WARNING:"some information"
            exit 0
        """
        [[[events]]]
        #     mail events = warning
        #     [scheduler]
        warning handlers = my-handler %(workflow)s %(event)s %(message)s
        # platform


```
with 

```
#!/bin/bash
#workflow/bin/my-handler

touch ${HOME}/info
echo "Workflow $1 event message: $2:$3" >> ${HOME}/info

```

hjoliver:
> 98% Happy

@wxtim & @oliver-sanders - is there still something to finish off on this?

hjoliver (approved):
LGTM

wxtim:
> > 98% Happy
> 
> @wxtim & @oliver-sanders - is there still something to finish off on this?

The example above - Oliver was going to have a think about it. TBH this is a massive improvement.

wxtim (approved):
Looks like the best for the moment.

oliver-sanders:
It's trivial (pass the message with the event), I think we're good.

PR #5772, opened 2023-10-17
Lint: Add a check for indentation being 4N spaces.

wxtim:
Make 'url' key optional in check specification.
Document pyproject.toml fields.

A couple of small fixes I found when looking at Cylc Lint for my talk on Thurs.

**Check List**

### DIFF
--- a/cylc/flow/scripts/lint.py
+++ b/cylc/flow/scripts/lint.py
@@ -66,7 +66,7 @@ except ImportError:
         loads as toml_loads,
         TOMLDecodeError,
     )
-from typing import Callable, Dict, Iterator, List, Union
+from typing import TYPE_CHECKING, Callable, Dict, Iterator, List, Union
 
 from cylc.flow import LOG
 from cylc.flow.exceptions import CylcError
@@ -79,6 +79,10 @@ from cylc.flow.id_cli import parse_id
 from cylc.flow.parsec.config import ParsecConfig
 from cylc.flow.scripts.cylc import DEAD_ENDS
 from cylc.flow.terminal import cli_function
+
+if TYPE_CHECKING:
+    from optparse import Values
+
 
 DEPRECATED_ENV_VARS = {
     'CYLC_SUITE_HOST': 'CYLC_WORKFLOW_HOST',
--- a/cylc/flow/scripts/lint.py
+++ b/cylc/flow/scripts/lint.py
@@ -22,6 +22,8 @@ Checks code style, deprecated syntax and other issues.
 # NOTE: docstring needed for `cylc help all` output
 # (if editing check this still comes out as expected)
 
+LINT_SECTIONS = ['cylc-lint', 'cylclint', 'cylc_lint']
+
 COP_DOC = """cylc lint [OPTIONS] ARGS
 
 Check .cylc and .rc files for code style, deprecated syntax and other issues.
@@ -33,9 +35,19 @@ Commit to version control before using this, in case you want to back out.
 
 A non-zero return code will be returned if any issues are identified.
 This can be overridden by providing the "--exit-zero" flag.
+"""
 
-Configurations for Cylc lint can also be set in a pyproject.toml file.
+TOMLDOC = """
+Configurations for Cylc lint can also be set in a pyproject.toml file using
+the following parameters:
 
+{}
+
+   [cylc-lint]                     # any of {}
+       ignore = ['S001', 'S002]    # List of rules to ignore
+       exclude = ['etc/foo.cylc']  # List of files to ignore
+       rulesets = ['style', '728'] # Sets default rulesets to check
+       max-line-length = 130       # Max line length for linting
 """
 from colorama import Fore
 import functools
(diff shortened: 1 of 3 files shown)


cylc/flow/scripts/lint.py
```diff
@@ -222,13 +234,41 @@ def check_for_obsolete_environment_variables(line: str) -> List[str]:
...
+        >>> check_indentation('baz')
+        False
+        >>> check_indentation('    qux')
+        False
+        >>> check_indentation('   foo')
+        True
+        >>> check_indentation('     bar')
+        True
+    """
+    match = re.findall(r'^(\s*)(.*)', line)[0]
```
oliver-sanders, line 260: May as well `re.compile` this expression and store as a global variable to avoid compiling it with each call.

cylc/flow/scripts/lint.py
```diff
@@ -33,9 +35,19 @@
...
+Configurations for Cylc lint can also be set in a pyproject.toml file using
+the following parameters:
 
+{}
+
+   [cylc-lint]                     # any of {}
+       ignore = ['S001', 'S002]    # List of rules to ignore
+       exclude = ['etc/foo.cylc']  # List of files to ignore
+       rulesets = ['style', '728'] # Sets default rulesets to check
+       max-line-length = 130       # Max line length for linting
```
oliver-sanders, line 50: Is something supposed to be inserted into the `{}`?

This is what I get:

```console
$ cylc lint --help
...
Configurations for Cylc lint can also be set in a pyproject.toml file using
the following parameters:



   [cylc-lint]                     # any of ['cylc-lint', 'cylclint', 'cylc_lint']
       ignore = ['S001', 'S002]    # List of rules to ignore
       exclude = ['etc/foo.cylc']  # List of files to ignore
       rulesets = ['style', '728'] # Sets default rulesets to check
       max-line-length = 130       # Max line length for linting
...
```

cylc/flow/scripts/lint.py
```diff
@@ -398,6 +437,14 @@ def check_for_obsolete_environment_variables(line: str) -> List[str]:
             check_if_jinja2,
             function=re.compile(r'(?<!{)#.*?{[{%]').findall
         )
+    },
+    'S012': {
+        'short': 'placeholder',
+        FUNCTION: lambda *args, **kwargs: True
+    },
```
oliver-sanders, line 444: Eh?

cylc/flow/scripts/lint.py
wxtim: Yes, that's what's supposed to happen. Are you going to quote me the "there should be one and preferably only one obvious way"?

oliver-sanders: > Yes, that's what's supposed to happen.

Then what is the `{}` supposed to do?

wxtim:
@oliver-sanders U001 and U002 being the same is deliberate, but I agree, perhaps a little confusing. They are detecting two different patterns with the same underlying cause (The change from `[scheduling][dependencies][<reccurrance>]graph=`, but one of them is detecting the `graph=` and one the `[[dependencies]]`).

I have added a bracket with a little text to this effect to each message - hopefully that makes it OK for you?

cylc/flow/scripts/lint.py
wxtim: Have the possible toml section headers filled in using str.format()

``` 
        COP_DOC + TOMLDOC.format('', str(LINT_SECTIONS)),
```

cylc/flow/scripts/lint.py
wxtim: Stops future developers accidentally over-writing S012 which is reserved for line length check. I've added a bit more explanation, and removed the FUNCTION, because it's unnecessary.

cylc/flow/scripts/lint.py
oliver-sanders: It doesn't look like anything is being inserted into it, bug?

cylc/flow/scripts/lint.py
oliver-sanders: Ah right, if you could jam in a comment to clarify.

wxtim: I've changed short to make it clearer.

cylc/flow/scripts/lint.py
wxtim: Sorry - there's two `{}` in your snippet, and we're talking about different ones. Sorry.

wxtim: Template is used at line 1116 for the CLI doc and we don't want to insert anything.

It's also used at line 1276 for rst format documentation where an rst directive  is inserted.

oliver-sanders: Gotcha.

It's the blank lines in the CLI output which threw me because it looked like something was missing.

Can shift those newlines into the RST bit:

```diff
​diff --git a/cylc/flow/scripts/lint.py b/cylc/flow/scripts/lint.py
index 1e986fea2..3baa98073 100755
--- a/cylc/flow/scripts/lint.py
+++ b/cylc/flow/scripts/lint.py
@@ -17,7 +17,6 @@
"""Cylc configuration linter.

Checks code style, deprecated syntax and other issues.
-
"""
# NOTE: docstring needed for `cylc help all` output
# (if editing check this still comes out as expected)
@@ -38,11 +37,7 @@ This can be overridden by providing the "--exit-zero" flag.
"""

TOMLDOC = """
-Configurations for Cylc lint can also be set in a pyproject.toml file using
-the following parameters:
-
-{}
-
+pyproject.toml configuration:{}
    [cylc-lint]                     # any of {}
        ignore = ['S001', 'S002]    # List of rules to ignore
        exclude = ['etc/foo.cylc']  # List of files to ignore
@@ -1274,5 +1269,5 @@ def main(parser: COP, options: 'Values', target=None) -> None:
# NOTE: use += so that this works with __import__
# (docstring needed for `cylc help all` output)
__doc__ += TOMLDOC.format(
-    '.. code-block:: toml', str(LINT_SECTIONS)) + get_reference_rst(
+    '\n\n.. code-block:: toml\n', str(LINT_SECTIONS)) + get_reference_rst(
     parse_checks(['728', 'style'], reference=True))
```

oliver-sanders (approved):

cylc/flow/scripts/lint.py
wxtim: Yes. Done.

markgrahamdawson (approved):

markgrahamdawson (approved):

PR #5777, opened 2023-10-19
Add missing test cases for `next()`/`previous()`

MetRonnie:
We were not testing the case of `next()` and `previous()` where the truncated properties were the same as the current datetime. E.g.
```
>>> now = 2020-10-19T13:53Z

>>> next(--10-19)
2020-10-19T00:00Z

>>> previous(--10-19)
2019-10-19T00:00Z
```

This behaviour does not seem to have been rigorously defined before? Adding these test cases ensures we do not break compatibility

**Check List**

- [x] No dependency changes
- [x] Tests are included 
- [x] No docs needed

### DIFF
--- a/tests/unit/cycling/test_iso8601.py
+++ b/tests/unit/cycling/test_iso8601.py
@@ -683,6 +683,7 @@ def test_simple(set_cycling_type):
         ('next(--0325)', '20110325T0000Z'),
         ('next(---10)', '20100810T0000Z'),
         ('next(---05T1200Z)', '20100905T1200Z'),
+        ('next(--08-08)', '20100808T0000Z'),
     ]
 )
 def test_next_simple(value: str, expected: str, set_cycling_type):
@@ -704,6 +705,7 @@ def test_next_simple(value: str, expected: str, set_cycling_type):
         ('previous(--0325)', '20100325T0000Z'),
         ('previous(---10)', '20100710T0000Z'),
         ('previous(---05T1200Z)', '20100805T1200Z'),
+        ('previous(--08-08)', '20090808T0000Z'),
     ]
 )
 def test_previous_simple(value: str, expected: str, set_cycling_type):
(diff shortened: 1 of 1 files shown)


tests/unit/cycling/test_iso8601.py
```diff
@@ -855,63 +829,37 @@ def test_weeks_days(set_cycling_type):
...
+        ('next(---01)', '20180401T0000Z'),
+        ('previous(--1225)', '20171225T0000Z'),
+        ('next(-2006)', '20200601T0000Z'),
+        ('previous(-W101)', '20180305T0000Z'),
+        ('next(-W-1; -W-3; -W-5)', '20180314T0000Z'),
+        ('next(-001; -091; -181; -271)', '20180401T0000Z'),
+        ('previous(-365T12Z)', '20171231T1200Z'),
+    ]
+)
+def test_cug(value: str, expected: str, set_cycling_type):
```
wxtim, line 858: I don't like the acronym.

```suggestion
def test_user_guide_examples(value: str, expected: str, set_cycling_type):
```

Might be a good idea to put a URL to the page of the User Guide in question in the docstring?

wxtim:
It strikes me that we also ought to test for 

```console
>>> now = 2020-10-19T13:53Z

>>> next(T-53)
2020-10-19T13:53Z

>>> previous(T-53)
2019-10-19T12:53Z
```

Although in the latter case I'm really not clear what I'd expect the correct answer to be. I think it probably should be the previous hour.

MetRonnie:
**Hang on a minute**, it strikes me that
> ```
> >>> next(--10-19)
> 2020-10-19T00:00Z
> ```

is a https://github.com/cylc/cylc-flow/labels/bug because it returns a datetime that is in the past!

wxtim:
> is a[ bug](https://github.com/cylc/cylc-flow/labels/bug) because it returns a datetime that is in the past!

That's definitely a bug. 🐞

oliver-sanders:
Yes, I would expect `next` to return a time in the future too.

oliver-sanders:
The one that's in the future ("next year") because just as `T00` is short for `T0000` and `01` is short for `01T0000`, `--10-19` is short for `--10-19T0000`.

wxtim (approved):
Nice use of xfail. Worth putting a comment with a link to the issue?

oliver-sanders (approved):

MetRonnie:
> Worth putting a comment with a link to #5795 in the code nearby?

A curious dev can use git blame

Issue #4579, opened 2022-01-14
reinstall: support reinstalling multiple workflows


oliver-sanders:
Follow on from #3931

The `cylc reinstall` command currently only works on a single workflow.

It should be able to support the multi workflow command syntax.

This might be useful in the context of sub-workflows e.g. `cylc reinstall '<sub_worflow_base>/*'`.

 
**Pull requests welcome!**

PR #5803, opened 2023-11-01
made reinstall work on multiple workflows

markgrahamdawson:
closes https://github.com/cylc/cylc-flow/issues/4579

## Updates to code
To make the reinstall operation work for multiple workflows we need to make the main function asynchronous. To do this we **switch out the `reinstall_cli` function which is called in def `main` for a `call_multi` function which is passed `reinstall_cli` wrapped in a `partial`**. 

We then have to go through the call stack and make sure that async and await keywords are used in the correct places - or that async versions of functions are used. So in this case the `reinstall_cli` function needed the `async` keyword added and `get_option_parser` multiworkflow=True.

That would be job done - however `cylc_reinstall` is also imported in `validate_reinstall` so we have to repeat the process of making everything conform with the asynchronous process here also. This includes:

1. using `wrapped_main` instead of `_main`
2. starting a new event loop for` vro_cli`
3. using `parse_id_async` instead of `parse_id`
4. making `vro_cli` asynchronous with the async keyword
5. await `cylc_reinstall` and `scheduler_cli`

As `scheduler_cli` was now being called asynchronously but was not an async function so this needed to be implemented in `cylc/flow/scheduler.py` which was achieved by:

1. switching out `parse_ids` for `parse_ids_async`
2. ...some other stuff

## Updates to tests

-- updates to tests --

**Check List**

### DIFF
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -37,7 +37,7 @@ import cylc.flow.flags
 from cylc.flow.id import upgrade_legacy_ids
 from cylc.flow.host_select import select_workflow_host
 from cylc.flow.hostuserutil import is_remote_host
-from cylc.flow.id_cli import parse_ids
+from cylc.flow.id_cli import parse_ids_async
 from cylc.flow.loggingutil import (
     close_log,
     RotatingLogFileHandler,
@@ -354,7 +354,11 @@ def _open_logs(id_: str, no_detach: bool, restart_num: int) -> None:
     )
 
 
-def scheduler_cli(options: 'Values', workflow_id_raw: str) -> None:
+async def scheduler_cli(
+    options: 'Values',
+    workflow_id_raw: str,
+    parse_workflow_id: bool = True
+) -> None:
     """Run the workflow.
 
     This function should contain all of the command line facing
@@ -368,15 +372,18 @@ def scheduler_cli(options: 'Values', workflow_id_raw: str) -> None:
     # Parse workflow name but delay Cylc 7 suite.rc deprecation warning
     # until after the start-up splash is printed.
     # TODO: singleton
-    (workflow_id,), _ = parse_ids(
-        workflow_id_raw,
-        constraint='workflows',
-        max_workflows=1,
-        # warn_depr=False,  # TODO
-    )
+    if parse_workflow_id:
+        (workflow_id,), _ = await parse_ids_async(
+            workflow_id_raw,
+            constraint='workflows',
+            max_workflows=1,
+            # warn_depr=False,  # TODO
+        )
+    else:
+        workflow_id = workflow_id_raw
 
     # resume the workflow if it is already running
-    _resume(workflow_id, options)
+    await _resume(workflow_id, options)
 
     # check the workflow can be safely restarted with this version of Cylc
     db_file = Path(get_workflow_srv_dir(workflow_id), 'db')
@@ -400,9 +407,7 @@ def scheduler_cli(options: 'Values', workflow_id_raw: str) -> None:
     # NOTE: asyncio.run opens an event loop, runs your coro,
     #       then shutdown async generators and closes the event loop
     scheduler = Scheduler(workflow_id, options)
-    asyncio.run(
-        _setup(scheduler)
-    )
+    await _setup(scheduler)
 
     # daemonize if requested
     # NOTE: asyncio event loops cannot persist across daemonization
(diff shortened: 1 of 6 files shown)


cylc/flow/scheduler_cli.py
```diff
@@ -419,9 +424,13 @@ def scheduler_cli(options: 'Values', workflow_id_raw: str) -> None:
     )
 
     # run the workflow
-    ret = asyncio.run(
-        _run(scheduler)
-    )
+    # ret = asyncio.new_event_loop().run_until_complete(_run(scheduler))
```
oliver-sanders, line 427:
```suggestion
    # Note: The daemonization messes with asyncio so we have to start a
    # new event loop if detaching
```

tests/integration/test_reinstall.py
```diff
@@ -89,15 +89,17 @@ def one_run(one_src, test_dir, run_dir):
         id=str(w_run_dir.relative_to(run_dir)),
     )
 
+from functools import partial
+from cylc.flow.network.multi import call_multi
```
oliver-sanders, line 93: Can stick these at the top of the file.

oliver-sanders (commented):
Looks good.

cylc/flow/scheduler_cli.py
wxtim: I agree with Oliver on this.

cylc/flow/scripts/validate_reinstall.py
```diff
@@ -124,16 +125,16 @@ def check_tvars_and_workflow_stopped(
 
 @cli_function(get_option_parser)
 def main(parser: COP, options: 'Values', workflow_id: str):
-    sys.exit(vro_cli(parser, options, workflow_id))
+    sys.exit(asyncio.run(vro_cli(parser, options, workflow_id)))
 
 
-def vro_cli(parser: COP, options: 'Values', workflow_id: str):
+async def vro_cli(parser: COP, options: 'Values', workflow_id: str):
```
wxtim, line 131: Tangential (and my fault), but while we're here can we change this for consistency - VRO is dead as a name.
```suggestion
async def vr_cli(parser: COP, options: 'Values', workflow_id: str):
```

wxtim:
I don't think that VR is quite right, sorry:

```console
> # for name in foo foot footling fool; do cylc vip ~/cylc-src/simplest/ -n $name; done
> cylc stop '*'
fool/run1
Command submitted; the scheduler will log any problems.
...
>  # ensure it's stopped
> cylc scan
> # Make some reinstall-worthy change:
> sed -i 's@a => b & c => d@a => b@' ~/cylc-src/simplest/flow.cylc 
> cylc scan
> cylc vr 'foo*'
WorkflowFilesError: invalid workflow name 'foo*' - can only contain: alphanumeric, `/`, `_`, `+`, `-`, `.`, `@`
```

markgrahamdawson:
> I don't think that VR is quite right, sorry - I don't think it will yet take wild cards:
> 
> ```shell
> # for name in foo foot footling fool; do cylc vip ~/cylc-src/simplest/ -n $name; done
> cylc stop '*'
> fool/run1
> Command submitted; the scheduler will log any problems.
>  ...
> # ensure it's stopped
> cylc scan
> # Make some reinstall-worthy change:
> sed -i 's@a => b & c => d@a => b@' ~/cylc-src/simplest/flow.cylc 
> cylc vr 'foo*'
> WorkflowFilesError: invalid workflow name 'foo*' - can only contain: alphanumeric, `/`, `_`, `+`, `-`, `.`, `@`
> ```

Hmm I think this must be something to do with parsing of workflow ids but Im not sure why

oliver-sanders:
@wxtim, this PR is just about `cylc reinstall`, `cylc vr` does not yet accept multiple workflows.

wxtim:
> @wxtim, this PR is just about `cylc reinstall`, `cylc vr` does not yet accept multiple workflows.

So it's out of scope. Agreed.

oliver-sanders (approved):

oliver-sanders:
LGTM, works well.

Tested `cylc play` with and without detaching locally and remotely to ensure this logic is still working as expected.

oliver-sanders:
There are some unrelated test failures on this branch which have been fixed on master.

Suggest rebasing this branch to pull in the updates:

```console
$ git fetch upstream
$ git checkout reinstall-multi-workflows
$ git branch reinstall-multi-workflows-safe  # make a safe copy before performing rebase
$ git rebase upstream/master
$ git push -f origin HEAD
```

wxtim (approved):

cylc/flow/scheduler_cli.py
```diff
@@ -419,9 +424,14 @@ def scheduler_cli(options: 'Values', workflow_id_raw: str) -> None:
...
-        _run(scheduler)
-    )
+    if options.no_detach:
+        ret = await _run(scheduler)
+    else:
+        # Note: The daemonization messes with asyncio so we have to start a
+        # new event loop if detaching
+        ret = asyncio.run(
+            _run(scheduler)
+        )
```
hjoliver, line 434: Suspected cause of #6291

PR #5144, opened 2022-09-19
workflow-state: query latest task instance

hjoliver:
When a task gets updated in the task states table, update the flow numbers too. 

Close #5143

**Check List**

- [ ] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [ ] Contains logically grouped changes (else tidy your branch by rebase).
- [ ] Does not contain off-topic changes (use other PRs for other changes).
- [ ] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.
- [ ] Tests are included (or explain why tests are not needed).
- [ ] `CHANGES.md` entry included if this is a change that can affect users
- [ ] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [ ] If this is a bug fix, PRs raised to both master and the relevant maintenance branch.

PR #5809, opened 2023-11-03
Fix workflow-state command and xtrigger.

### LINKED ISSUE #5143: workflow-state: check latest instance for status
The task_states DB table gets a new entry for a given task ID if its flow numbers change. That's because we use the table to extract flow-wait status for historical tasks that already ran in this flow.

(For a given flow, the submit number of a task ID updates but we don't get a new entry).

`cylc workflow-state` wrongly assumes that any non-null query result indicates status achieved. It should only check the result for the latest matching record in the table (i.e. latest submit number).

hjoliver:
- handle `--status` and `--output` queries the same, with respect to polling vs one-off queries
- unified polling (for db not found, not connected, or condition not satisfied)
- close #5143 (check latest instance of a task)
- make the workflow-state command and xtrigger flow-specific: add `--flow` option
- log warnings for failed xtrigger calls (currently these are silent except in debug mode)
- supersede #5144 
- (alternate run-dir already done #6031 )
- ~~retain `--message` option (task message) as well as `--output` (trigger name) to avoid breaking existing polling workflows~~

---

Minor https://github.com/cylc/cylc-flow/labels/db%20change: Format of `outputs` column in `task_outputs` table has changed from Cylc 8.0.0 where it was a stringified list of task messages, to a stringified dictionary of the format `{<output label>: <task message>}` (like how it was in Cylc 7).

-------------
The command and xtrigger now take `WORKFLOW//CYCLE/TASK:SELECTOR` as an arg. 

`SELECTOR` can be a status or an output. With `--output`, or if not a valid status, it is taken as an output; otherwise a status.

Any query will be polled for if no matching result is found.

A good test case:
```ini
[scheduling]
    cycling mode = integer
    initial cycle point = 2001
    final cycle point = 2003
[scheduling]
    [[xtriggers]]
        ws = workflow_state("%(workflow)s//%(point)s/foo:succeeded", offset="-P1"):PT2S
    [[graph]]
        P1 = """
           foo[-P1] => foo
           foo:x => bar & baz
        """
        R/2002/P1 = "@ws => baz"
[runtime]
    [[foo]]
        script = """
           cylc message -- "xxx"
        """
        [[[outputs]]]
           x = xxx
    [[bar]]
        script = """
          if [[ "$CYLC_TASK_CYCLE_POINT" == "$CYLC_WORKFLOW_FINAL_CYCLE_POINT" \
             && "$CYLC_TASK_SUBMIT_NUMBER" == "1" ]]
          then
              cylc trigger --flow=new $CYLC_WORKFLOW_ID//2002/foo
          fi
        """
    [[baz]]
```

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/727


cylc/flow/dbstatecheck.py
```diff
@@ -145,37 +180,49 @@ def workflow_state_query(
...
             stmt += " where " + (" AND ").join(stmt_wheres)
 
+        # Note we can't use "where outputs==message"; because the outputs
+        # table holds a serialized string of all received outputs.
+
         res = []
         for row in self.conn.execute(stmt, stmt_args):
-            if not all(v is None for v in row):
+            if row[-1] is not None:
+                # not all(v is None for v in row):
```
hjoliver, line 199: Note to self: fix comment. I can't see how the original test makes sense, but it is possible for the task_outputs table (a Cylc 7 one at least) to have a row with outputs value None.

cylc/flow/scripts/workflow_state.py
```diff
@@ -142,51 +159,40 @@ def get_option_parser() -> COP:
...
+        action="store", dest="offset", metavar="OFFSET", default=None)
 
     parser.add_option(
-        "-S", "--status",
-        help="Specify a particular status or triggering condition to "
-             f"check for. {conds}{states}",
+        "-S", "--status", metavar="STATUS",
+        help=f"Task status: {', '.join(TASK_STATUSES_ORDERED)};"
+        " plus: started, finished.",
         action="store", dest="status", default=None)
```
MetRonnie, line 191: Can let optparse handle the choices:
```diff
diff --git a/cylc/flow/scripts/workflow_state.py b/cylc/flow/scripts/workflow_state.py
index fda8e722d..5f8d3e99d 100755
--- a/cylc/flow/scripts/workflow_state.py
+++ b/cylc/flow/scripts/workflow_state.py
@@ -185,9 +185,12 @@ def get_option_parser() -> COP:
         action="store", dest="offset", metavar="OFFSET", default=None)
 
+    status_choices = [
+        *TASK_STATUSES_ORDERED, *CylcWorkflowDBChecker.STATE_ALIASES
+    ]
     parser.add_option(
         "-S", "--status", metavar="STATUS",
-        help=f"Task status: {', '.join(TASK_STATUSES_ORDERED)};"
-        " plus: started, finished.",
-        action="store", dest="status", default=None)
+        help=f"Task status. Choices: {', '.join(status_choices)}",
+        action="store", dest="status", default=None,
+        choices=status_choices)
 
     parser.add_option(
@@ -229,10 +232,4 @@ def main(parser: COP, options: 'Values', workflow_id: str) -> None:
         raise InputError("cannot poll both status and custom output")
 
-    # Exit if an invalid status is requested
-    if (options.status and
-            options.status not in TASK_STATUSES_ORDERED and
-            options.status not in CylcWorkflowDBChecker.STATE_ALIASES):
-        raise InputError(f"invalid status '{options.status}'")
-
     # this only runs locally
     if options.run_dir:
```

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,35 +18,55 @@
...
-  $ cylc workflow-state WORKFLOW_ID --task=TASK --status=STATUS --task-point
-  # uses CYLC_TASK_CYCLE_POINT environment variable as the value for the
-  # CYCLE to poll. This is useful when you want to use cylc workflow-state in a
-  # cylc task.
+
+  # Print the current or latest status of all tasks:
+  $ cylc workflow-state WORKFLOW_ID
+
+  # Print the current or latest status of all tasks named "foo":
+  $ cylc workflow-state --task=foo WORKFLOW_ID
```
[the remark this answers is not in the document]
oliver-sanders: It's more correct to put them before, but either is fine.

changes.d/5809.fix.d
```diff
@@ -0,0 +1,2 @@
+Fix the workflow-state command to handle task statuses and output messages
+consistently.
```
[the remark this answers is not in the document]
hjoliver: Roger that.

cylc/flow/scripts/workflow_state.py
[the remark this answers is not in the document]
hjoliver: Hmmm, yeah sorry, to me, options always go before arguments!

cylc/flow/dbstatecheck.py
```diff
@@ -108,6 +126,10 @@ def get_remote_point_format_compat(self):
         ):
             return row[0]
 
+    def get_point_format(self):
+        """Return the cycle point format of this DB."""
+        return self.point_fmt
+
```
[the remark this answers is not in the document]
hjoliver: I struggle to get over historical arguments (pre-Python possibly) that getter methods should always be used. In this case, fair enough, I'll ditch the method. 

We should probably be using `@property` decorator to kind-of do read-only access. https://docs.python.org/3/library/functions.html?highlight=property#property

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,35 +18,55 @@
...
+Override default polling parameters with --max-polls and --interval.
+
+If the database does not exist at first, polls are consumed waiting for it.
+
+For task outputs, give the task message, not the output label.
+
+For non-cycling workflows, provide --point=1 for specific queries.
+
+This command can be used to make polling tasks that trigger off of tasks in
+other workflows - but see also the built-in workflow_state xtrigger for that.
```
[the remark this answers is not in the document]
hjoliver: I'm more acknowledging the fact that some users already do this, here at least (although it may be a pre-xtrigger hang-over). Maybe there are valid use cases - a task that uses the command could do other things at the same time. 

Your wording isn't quite correct as the task is not a "polling task" anymore if the xtrigger is doing the polling, but I'll change the wording to suggest that this is possible but not necessarily desirable.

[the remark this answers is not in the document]
oliver-sanders: > Not sure it is a good idea to suggest using this command for this purpose at all? Is there any circumstance the xtrigger is not better for it?

Yes, they don't clog up the workflow's process pool, can be more efficient and can be used to poll workflows on remote platforms which xtriggers cannot do.

cylc/flow/dbstatecheck.py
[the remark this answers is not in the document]
oliver-sanders: > (pre-Python possibly) that getter methods should always be used

Java!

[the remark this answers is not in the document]
MetRonnie: > We should probably be using `@property` decorator to kind-of do read-only access. https://docs.python.org/3/library/functions.html?highlight=property#property

Only if it offers a reasonable advantage IMO, because it comes with a slight performance (and complexity) penalty

hjoliver:
I'll punt this back to 8.4.0, not urgent.

hjoliver:
Note also: #6030

hjoliver:
I've realized NIWA's heavy use of cross-workflow triggering means we need the workflow-state command and xtrigger fixed in several ways, for Cylc 8 migration.  Therefore:
- bumping this forward to 8.3.0 
- superseding #5144 
- adding `--flow` option (needed for obvious reasons, with Cylc 8)

hjoliver:
(Don't bother reviewing this till un-drafted - which should be tomorrow - I have some local changes pending)

hjoliver:
Some more tests needed, but otherwise ready for review now.

cylc/flow/dbstatecheck.py
```diff
@@ -151,36 +203,69 @@ def workflow_state_query(
...
+            stmt += dedent("""
+                ORDER BY
+                    submit_num
+            """)
+
+        # idx_status = 2
+        # idx_flow = 3
+
+        # TODO option to list outputs instead of status
+        # TODO if flows not specified, use the most recent for the task
```
MetRonnie, line 220: What's the status with these comments?

cylc/flow/dbstatecheck.py
```diff
@@ -151,36 +203,69 @@ def workflow_state_query(
...
-        return res
+        if output:
+            # Replace res with a task-states like result,
+            # [[foo, 2032, output], [foo, 2033, output]]
+            if self.back_compat_mode:
+                # Cylc 7 DB: list of {output: message}
+                results = [
+                    [item[0], item[1], output]
+                    for item in res
+                    if output in deserialise(item[2]).values()
```
MetRonnie, line 241: For better readability 
```suggestion
                    [name, cycle, output]
                    for name, cycle, outputs_str in res
                    if output in deserialise(outputs_str).values()
```

tests/functional/xtriggers/04-suite_state.t
```diff
@@ -29,7 +29,7 @@ init_workflow "$TEST_NAME_BASE" << __FLOW_CONFIG__
         [[[R1]]]
             graph = @upstream => foo
     [[xtriggers]]
-        upstream = suite_state(suite=thorin/oin/gloin, task=mithril, point=1)
+        upstream = suite_state(suite=thorin/oin/gloin, task=mithril, point=1, status="succeeded")
```
MetRonnie, line 32: I don't think this change is needed?

cylc/flow/xtrigger_mgr.py
```diff
@@ -623,22 +625,35 @@ def callback(self, ctx: 'SubFuncContext'):
...
             ValueError: if the context given is not active
         """
         sig = ctx.get_signature()
         self.active.remove(sig)
+
+        if ctx.ret_code != 0:
+            msg = f"ERROR in xtrigger {sig}"
+            if ctx.err:
+                msg += f"\n{ctx.err}"
+            LOG.warning(msg)
```
MetRonnie, line 643: 👍

MetRonnie:
`tests/k/xtriggers/01-workflow_state.t` failing btw

cylc/flow/scripts/workflow_state.py
```diff
@@ -142,52 +160,52 @@ def get_option_parser() -> COP:
...
+        action="store", dest="status", default=None, choices=statuses)
 
     parser.add_option(
-        "-S", "--status",
-        help="Specify a particular status or triggering condition to "
-             f"check for. {conds}{states}",
-        action="store", dest="status", default=None)
+        "-O", "--output", "--message", metavar="OUTPUT",
+        help="Check for a given task output"
+        " (--message is now aliased to task output)",
```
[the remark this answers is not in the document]
hjoliver: Yes, came to the same conclusion myself and had already done it for the xtrigger side of this.

tests/functional/xtriggers/04-suite_state.t
hjoliver: Good spot - it briefly was, early in branch development.

cylc/flow/scripts/workflow_state.py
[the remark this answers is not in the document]
hjoliver: Done.

hjoliver:
Temporarily converted to draft again - review comments partially addressed, more local changes pending.

cylc/flow/dbstatecheck.py
```diff
@@ -67,34 +69,51 @@ def __init__(self, rund, workflow, db_path=None):
...
     def display_maps(res):
         if not res:
             sys.stderr.write("INFO: No results to display.\n")
         else:
             for row in res:
-                sys.stdout.write((", ").join(row) + "\n")
+                sys.stdout.write((", ").join([str(s) for s in row]) + "\n")
 
-    def get_remote_point_format(self):
+    def _get_pt_fmt(self):
```
wxtim, line 92: How about `_get_scheduler_cycle_point_format` - it's a mouthful, but it then exactly matches the config item, the value of which are are getting from the DB.

wxtim (commented):
Seems reasonable; Given the amount on your plate I will start drafting tests for this so that we can ram it through.

I've added the following:https://github.com/hjoliver/cylc-flow/pull/50 - It's not a complete set of tests, but I've covered some of this. I'll do more tomorrow. If you want to get them in feel free to merge though.

cylc/flow/dbstatecheck.py
hjoliver: As a private method now a self-explanatory name is a bit less import, but I'll consider the options...

cylc/flow/dbstatecheck.py
```diff
@@ -109,39 +128,72 @@ def get_remote_point_format_compat(self):
...
+        """Allows for multiple states to be searched via a status alias."""
         if state in self.STATE_ALIASES:
             return self.STATE_ALIASES[state]
         else:
             return [state]
 
     def workflow_state_query(
-            self, task, cycle, status=None, message=None, mask=None):
-        """run a query on the workflow database"""
+        self,
```
wxtim, line 138: I'm slightly questioning whether this is attached to the correct object: Should this perhaps be attached to the workflow database or something?

cylc/flow/dbstatecheck.py
```diff
@@ -151,36 +203,69 @@ def workflow_state_query(
...
+                    for name, cycle, outputs_str in res
+                    if output in deserialise(outputs_str).values()
+                ]
+            else:
+                # Cylc 8 DB list of [output]
+                results = []
+                for name, cycle, outputs_str, flows_str in res:
+                    flows = deserialise(flows_str)
+                    outputs = deserialise(outputs_str)
+                    if output not in outputs:
```
wxtim, line 249: I think that this is broken (I've included a manual test below, becuase it's possible that I've misunderstood how this command ought to work.

Looking at the db I think that it would be more accurate to call the field in res currently called `outputs_str`, `messages_string`. This condition is therefore only ever false if the output == message. That's fine for the standard internal messages, but probably not for most external ones.

## Example

```toml
# flow.cylc
[scheduler]
    allow implicit tasks = True
    cycle point format = %Y

[scheduling]
    initial cycle point = 2380
    [[graph]]
        R1 = foo:trigger => wonder_horse

[runtime]
    [[foo]]
        script = cylc message -- message
        [[[outputs]]]
            # If you tested with trigger = trigger this will not show
            trigger = 'message'   
```

```bash
cylc vip .
sleep 5

# This is never satisfied
cylc workflow-state icky.lynx --point 2380 --task foo --output trigger

# This is
cylc workflow-state icky.lynx --point 2380 --task foo --output message

```

cylc/flow/dbstatecheck.py
```diff
@@ -151,36 +203,69 @@ def workflow_state_query(
...
-        return res
+        if output:
+            # Replace res with a task-states like result,
+            # [[foo, 2032, output], [foo, 2033, output]]
+            if self.back_compat_mode:
+                # Cylc 7 DB: list of {output: message}
+                results = [
+                    [name, cycle, output]
+                    for name, cycle, outputs_str in res
+                    if output in deserialise(outputs_str).values()
```
wxtim, line 241: See comment at line 249

cylc/flow/dbstatecheck.py
hjoliver: Not sure I understand the question @wxtim ??  `self` is an object that knows which DB to connect to once the query has been formulated based on the command options.

cylc/flow/dbstatecheck.py
hjoliver: What should work there is `--trigger` but not `--message`.

This relates to (a) Ronnie's comment [above](https://github.com/cylc/cylc-flow/pull/5809#discussion_r1583193487) to remove `--message`; and (b) the pending local changes I mentioned (the reason for temporarily converting this PR back to Draft).

Post optional outputs etc. we decided that the graph trigger name is "the output", rather than (as was historically the case) the task message. So all user commands and queries should specify the trigger, not the message. Only the task job sends the message, which can be more verbose for logging purposes.

cylc/flow/dbstatecheck.py
wxtim: Yes. That's correct. I missed that this is semi-standalone.

cylc/flow/dbstatecheck.py
wxtim: > What should work there is `--trigger` but not `--message` (which isn't the case as you've noted).
> 
> Sorry, this was partially borked, hence the pending local changes mentioned as the reason for temporarily converting this PR back to Draft.
> 
> Post optional outputs, we decided that "the outputs" are now the graph trigger names, not the task messages (as was historically the case) . So now all user commands should specify the trigger, not the message. Only the task job sends the message, which can be more verbose for logging purposes.
> 
> My initial thought on this PR was to alias `--message` to `--output` so that existing use cases would continue to work for (the majority of?) cases where trigger and message are identical. Then as per Ronnie's comment above better to just remove `--message` altogther.
> 
> However, I'm rethinking that. It would be easy enough to continue to support both, just in this command, to avoid breaking any existing workflows that use it for triggering (NIWA has many).

This sounds reasonable, especailly if there's a deprecation warning if `--message` is used.

hjoliver:
@wxtim - plz hold off on this till I've pushed up my pending changes (later tonight NZ if poss).

wxtim:
> @wxtim - plz hold off on this till I've pushed up my pending changes (later tonight NZ if poss).

I've re-drafted my tests for this: will rebase them once you have a fix.

hjoliver:
I pushed my local changes, and I think everything works now on this branch, but annoyingly there's a pretty good conflict with master due to merging the optional outputs extension. I've have to leave it for today.  

TODO:
- address a couple of internal TODOs flagged by Ronnie above
- reinstate output message checking (easy) to avoid breaking existing workflow-state triggering use cases (so we'll have `--trigger` (or `--output`) and `--message`).
- fix conflicts with master

hjoliver:
(Actually the conflict is easier than it looked at first glance, having a go at it now).

cylc/flow/dbstatecheck.py
MetRonnie: > especially if there's a deprecation warning if `--message` is used.

👍

hjoliver: Unfortunately we've often used `[outputs] x = x` for convenience (laziness?) in examples and tests. Which of course fails to pick up any confusion between trigger and message in the code.

cylc/flow/dbstatecheck.py
hjoliver: Not sure what that was - is it still relevant?

tests/unit/xtriggers/test_workflow_state.py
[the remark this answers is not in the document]
hjoliver: Yes - currently it's only testing status query, not outputs.

cylc/flow/dbstatecheck.py
hjoliver: (1) done; (2) punted to #6094

hjoliver:
All done, I think @wxtim

hjoliver:
Draft again. It turns out this can't be made a non-breaking change after all, so I'm going all in and ditching the horrible old separate args (--task, --cycle, --status, etc.) and output format (name, cycle, status, etc.), for our nice new IDs. 

There's still the `suite_state` wrapper for xtrigger to Cylc 7 suites.

hjoliver:
(almost done 😬 )

cylc/flow/cfgspec/workflow.py
```diff
@@ -1995,6 +1973,25 @@ def upg(cfg, descr):
...
+        '8.0.0',
+        ['runtime', '__MANY__', 'suite state polling'],
+        ['runtime', '__MANY__', 'workflow state polling'],
+        silent=cylc.flow.flags.cylc7_back_compat,
+        is_section=True,
+    )
+    for item in ('host', 'user'):
+        u.obsolete(
+            '8.0.0', ['runtime', '__MANY__', 'workflow state polling', item],
+        )
```
wxtim, line 1986: IMO for loop obscures this code and lengthens it.

```suggestion
    u.obsolete( '8.0.0', ['runtime', '__MANY__', 'workflow state polling', 'host'])
    u.obsolete( '8.0.0', ['runtime', '__MANY__', 'workflow state polling', 'user'])
```

cylc/flow/cfgspec/workflow.py
```diff
@@ -2195,7 +2184,7 @@ def upgrade_graph_section(cfg: Dict[str, Any], descr: str) -> None:
                         keys.add(key)
                 if keys and not cylc.flow.flags.cylc7_back_compat:
                     msg = (
-                        'deprecated graph items were automatically upgraded '
+                        'graph items were automatically upgraded '
```
wxtim, line 2187: IMO think the word deprecated is useful as part poking users to upgrade.

cylc/flow/config.py
```diff
@@ -1531,24 +1530,37 @@ def configure_workflow_state_polling_tasks(self):
...
+                    ('alt-cylc-run-dir', ' --%s=%s')]:
                 if rtc['workflow state polling'][key]:
                     comstr += fmt % (key, rtc['workflow state polling'][key])
-            if rtc['workflow state polling']['message']:
-                comstr += " --message='%s'" % (
-                    rtc['workflow state polling']['message'])
-            else:
-                comstr += " --status=" + tdef.workflow_polling_cfg['status']
-            comstr += " " + tdef.workflow_polling_cfg['workflow']
             script = "echo " + comstr + "\n" + comstr
```
wxtim, line 1564: Alternative, unimportant
```suggestion
            script = "set +x\n" + comstr
```

cylc/flow/config.py
```diff
@@ -1531,24 +1530,37 @@ def configure_workflow_state_polling_tasks(self):
                 continue
             rtc = tdef.rtconfig
             comstr = (
-                "cylc workflow-state"
-                f" --task={tdef.workflow_polling_cfg['task']}"
-                " --point=$CYLC_TASK_CYCLE_POINT"
+                "cylc workflow-state "
```
wxtim, line 1533: Am I correct in thinking that this loop is only called on the old graph-syntax polling system?

I tried

```
[scheduler]
    allow implicit tasks = True
    cycle point format = %Y
[scheduling]
    initial cycle point = 2010
   [[graph]]
        P1Y = """
            foo<WORKFLOW::TASK:fail> => bar
        """
```

expecting a deprecation warning, but didn't get one.

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,260 @@
...
+        tokens = Tokens(self.id_)
+        self.workflow_id_raw = tokens.workflow_id
+        self.task_sel = tokens["task_sel"] or default_status
+        self.cycle_raw = tokens["cycle"]
+        self.task = tokens["task"]
+
+        self.workflow_id = None
+        self.cycle = None
+        self.results = None
+        self.db_checker = None
```
wxtim, line 138: Why define this twice?

cylc/flow/dbstatecheck.py
```diff
@@ -65,36 +52,99 @@ def __init__(self, rund, workflow, db_path=None):
...
+        except sqlite3.OperationalError as exc:
+            # BACK COMPAT: Cylc 7 DB (see method below).
+            try:
+                self.db_point_fmt = self._get_db_point_format_compat()
+                self.back_compat_mode = True
+            except sqlite3.OperationalError:
+                raise exc  # original error
+
+    def adjust_point_to_db(self, cycle, offset):
+        """Adjust a cycle point (with offset) to the DB point format."""
```
wxtim, line 73:
```suggestion
        """Adjust a cycle point (with offset) to the DB point format.
        
        Used by workflow state to ....
        """
```

cylc/flow/xtriggers/workflow_state.py
```diff
@@ -14,101 +14,93 @@
...
+from cylc.flow.id import tokenise
+from cylc.flow.exceptions import WorkflowConfigError
+from cylc.flow.task_state import TASK_STATUS_SUCCEEDED
 
 
 def workflow_state(
-    workflow: str,
-    task: str,
-    point: str,
+    workflow_task_id: str,
```
wxtim, line 27: This seems like a massive breaking change...

Could you not have something like

```python
def workflow_state(*args, **kwargs):
    if len(args) == 1 or 'workflow_task_id' in kwargs:
        _workflow_state()
    else:
        LOG.warn('using back compatibility version of xtrigger')
        _workflow_state_compat()
```

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,260 @@
...
+
+    def _find_workflow(self) -> bool:
+        """Find workflow and infer run directory, return True if found."""
+        try:
+            self.workflow_id = infer_latest_run_from_id(
+                self.workflow_id_raw,
+                self.alt_cylc_run_dir
+            )
+        except InputError:
+            LOG.debug("Workflow not found")
```
wxtim, line 158: This feels like it might be a warning level error?

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,260 @@
...
+        """
+        if self.workflow_id is None and not self._find_workflow():
+            return False
+
+        if self.db_checker is None and not self._db_connect():
+            return False
+
+        if self.cycle is None:
+            # Adjust target cycle point to the DB format.
+            self.cycle = self.db_checker.adjust_point_to_db(
```
wxtim, line 195: Can we not do this once when we instantiate the object to avoid calling it each time?

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,260 @@
...
-             "Shorthand for --point=$CYLC_TASK_CYCLE_POINT",
-        action="store_true", dest="use_task_point", default=False)
-
-    parser.add_option(
-        "-d", "--run-dir",
-        help="The top level cylc run directory if non-standard. The "
-             "database should be DIR/WORKFLOW_ID/log/db. Use to interrogate "
-             "workflows owned by others, etc.; see note above.",
-        metavar="DIR", action="store", dest="alt_run_dir", default=None)
+        "-d", "--alt-cylc-run-dir",
```
wxtim, line 215: Similar to my other comments - Is it wise to make this a totally breaking change rather than deprecating the older style of xtriggers?

cylc/flow/config.py
```diff
@@ -1531,24 +1530,37 @@ def configure_workflow_state_polling_tasks(self):
...
+            graph_trigger = tdef.workflow_polling_cfg['status']
+            config_trigger = rtc['workflow state polling']['output']
+            if (
+                graph_trigger is not None and
+                config_trigger is not None and
+                graph_trigger != config_trigger
+            ):
+                raise WorkflowConfigError(
+                    f'Polling task "{name}" must configure a target status or'
+                    f' output\nin the graph (:{graph_trigger}) or in its task'
```
MetRonnie, line 1547:
```suggestion
                    f' output in the graph (:{graph_trigger}) or in its task'
```

wxtim:
I enjoyed being able to trigger a Gregorian cycling task off a task in an integer cycling workflow.

cylc/flow/xtrigger_mgr.py
```diff
@@ -159,143 +159,70 @@ class TemplateVariables(Enum):
...
+        """
+        if label in self.functx_map:
+            # we've already seen this one
+            return
+
+        if (
+            not label.startswith('_cylc_retry_') and not
+            label.startswith('_cylc_submit_retry_')
+        ):
+            # (the "_wall_clock" function fails "wall_clock" validation)
```
MetRonnie, line 202: ?

cylc/flow/xtrigger_mgr.py
```diff
@@ -159,143 +159,70 @@ class TemplateVariables(Enum):
...
         fctx: 'SubFuncContext',
         fdir: str,
     ) -> None:
-        """Generic xtrigger validation: check existence, string templates and
-        function signature.
+        """Check xtrigger existence, string templates and function signature.
 
-        Xtrigger modules may also supply a specific `validate` function
-        which will be run here.
+        Also call an xtrigger-specific argument validate() function if defined.
```
MetRonnie, line 226: IMO the previous wording was clearer

tests/functional/workflow-state/05-output.t
```diff
@@ -0,0 +1,32 @@
...
+set_test_number 2
+
+install_workflow "${TEST_NAME_BASE}" output
+
+TEST_NAME="${TEST_NAME_BASE}-run"
+workflow_run_ok "${TEST_NAME}" \
+    cylc play --reference-test --debug --no-detach "${WORKFLOW_NAME}"
+
+TEST_NAME=${TEST_NAME_BASE}-cli-check
+run_ok "${TEST_NAME}" cylc workflow-state "${WORKFLOW_NAME}//20100101T0000Z/t1:out1" --max-polls=1
```
wxtim, line 30:
```suggestion
run_ok "${TEST_NAME}" \
    cylc workflow-state \
    "${WORKFLOW_NAME}//20100101T0000Z/t1:out1" \
    --max-polls=1
```

cylc/flow/parsec/upgrade.py
```diff
@@ -245,9 +245,8 @@ def upgrade(self):
                 # User level configuration, user should be able to fix.
                 # Log at warning level.
                 level = WARNING
-            LOG.log(level,
-                    'deprecated items were automatically upgraded in '
-                    f'"{self.descr}"')
+            LOG.log(
+                level, f"items were automatically upgraded in {self.descr}")
```
MetRonnie, line 249: IMO leave the "deprecated" in there

cylc/flow/workflow_db_mgr.py
```diff
@@ -742,8 +742,8 @@ def upgrade_pre_810(pri_dao: CylcWorkflowDAO) -> None:
 
         # We can't upgrade if the flow_nums in task_states are not
         # distinct.
-        from cylc.flow.util import deserialise
-        flow_nums = deserialise(conn.execute(
+        from cylc.flow.util import deserialise_set
```
MetRonnie, line 745: Already imported
```suggestion
```

tests/integration/scripts/test_validate_integration.py
```diff
@@ -105,7 +105,7 @@ def test_validate_simple_graph(flow, validate, caplog):
     })
     validate(id_)
     expect = (
-        'deprecated graph items were automatically upgraded'
+        'graph items were automatically upgraded'
```
wxtim, line 108: As per my other comment, I think that the use of the word "deprecated" hints at us wanting to remove this in future.

tests/unit/test_config.py
```diff
@@ -181,7 +175,7 @@ def test_xfunction_attribute_error(self, mock_glbl_cfg, tmp_path):
         with pytest.raises(XtriggerConfigError) as excinfo:
             WorkflowConfig(workflow="capybara_workflow", fpath=flow_file,
                            options=SimpleNamespace())
-        assert "not found" in str(excinfo.value)
+        assert "module 'capybara' has no attribute 'capybara'" in str(excinfo.value)
```
wxtim, line 178: 🐀.🐖

cylc/flow/workflow_db_mgr.py
```diff
@@ -632,13 +632,13 @@ def put_update_task_outputs(self, itask):
         """Put UPDATE statement for task_outputs table."""
         set_args = {
             "outputs": json.dumps(
-                list(itask.state.outputs.iter_completed_messages())
+                itask.state.outputs.get_completed_outputs()
```
MetRonnie, line 635: List changed to dictionary.

This looks like a https://github.com/cylc/cylc-flow/labels/db%20change that has not been handled properly. Is it actually needed?

wxtim (commented):
## Breaking change

I'm not sure whether the extent to which this breaks existing practice is great: It instantly breaks the workflows of anyone using the existing xtrigger or directly scripting workflow-state in a task. 

Ironically it handles it if you are using the deprecated `suite_state_check<WORKFLOW::TASK>`, which means that it's easier to deal with if you are using that, now deprecated method.

## Documentation

I think this _really needs_ a companion docs PR as a number of docs pages are quite outdated as a result of this change:

* https://cylc.github.io/cylc-doc/stable/html/user-guide/writing-workflows/external-triggers.html#built-in-workflow-state-triggers

cylc/flow/cfgspec/workflow.py
```diff
@@ -1730,57 +1730,34 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
-                    The maximum number of polls before timing out and entering
-                    the "failed" state.
+                    Maximum number of polls to attempt.
                 ''')
-                Conf('message', VDR.V_STRING, desc='''
-                    Wait for the task in the target workflow to receive a
-                    specified message rather than achieve a state.
+                Conf('output', VDR.V_STRING, desc='''
+                    Target task output (trigger name, not task message).
+                    Alternatively, put this in the graph trigger syntax.
```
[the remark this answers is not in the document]
wxtim: Deprecation is a name change mechanism: So it was called message, and is now called output.

cylc/flow/xtrigger_mgr.py
wxtim: Have a look at `xtriggers/wall_clock.py` - It's wierd.

wxtim (approved):

wxtim (commented):
Un-approving

cylc/flow/xtrigger_mgr.py
MetRonnie: I know, but I don't see what that has to do with:
- validation, as `_wall_clock` does not go through any validation process.
- the `if` condition above.

cylc/flow/xtriggers/workflow_state.py
```diff
@@ -14,101 +14,93 @@
...
-        tuple: (satisfied, results)
-
+        tuple: (satisfied, result)
         satisfied:
             True if ``satisfied`` else ``False``.
-        results:
-            Dictionary containing the args / kwargs which were provided
-            to this xtrigger.
+        result:
+            Dict {workflow_id, task_id, task_selector, flow_number}.
```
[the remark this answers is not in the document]
oliver-sanders: Explanation, if type hints are not present, then Google style docstrings should provide the type (followed by a colon) like so:

```
    Returns:
        bool: The return value. True for success, False otherwise.
```

However, when type hints are used, the type does not need to be specified:

```
    Returns:
        The return value. True for success, False otherwise.
```

Those examples are from this page of examples: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google

Sphinx Napoleon will error when docstrings don't match its expected formats (either Google style or Numpy style).

cylc/flow/cfgspec/workflow.py
```diff
@@ -1730,57 +1730,34 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
-                ''')
+                   Please switch to workflow_state xtriggers.
+            '''):
                 Conf('interval', VDR.V_INTERVAL, desc='''
                     Polling interval.
                 ''')
                 Conf('max-polls', VDR.V_INTEGER, desc='''
-                    The maximum number of polls before timing out and entering
-                    the "failed" state.
+                    Maximum number of polls to attempt.
```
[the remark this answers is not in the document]
hjoliver: The "timing out" bit was not true. I'll restore the "failed state" bit.

cylc/flow/cfgspec/workflow.py
hjoliver: (Yeah, it originally had more items!)

cylc/flow/cfgspec/workflow.py
hjoliver: The message is printed for obsoleted items as well as for deprecated ones.

I'll make the message more instructive.

cylc/flow/config.py
hjoliver: I think you mean `set -x`. But echo is simpler here. It's just intended to provide a record in the job.out of the command that was automatically put in the job script. Using `set -x` writes to stderr, which might make it appear that something went wrong, and it requires turning off again to avoid reporting the job script boiler-plate code as well.

cylc/flow/config.py
hjoliver: You are correct, but we only print deprecation warnings for config items that get automatically upgraded. In this case you don't have a `[[[suite state polling]]]` section that triggers the warning.

I've added an additional deprecation warning at validation time, via the config module:
```
WARNING - Workflow state polling tasks are deprecated, please convert to workflow_state xtriggers:
     * oof
     * woof
```

cylc/flow/scripts/workflow_state.py
hjoliver: Not intentional! fixed.

cylc/flow/dbstatecheck.py
hjoliver: Rephrased, but more context added.

cylc/flow/xtriggers/workflow_state.py
hjoliver: I decided a breaking change was preferable because full backward compatibility is not possible. 

The main reason is the old command allows both `--output` and `--message` but, contrary to the help docs, they both query the task message. Whereas now, post optional outputs, "output" means the trigger name, not the task message.

However, I just checked again and the xtrigger function (unlike its CLI counterpart) doesn't have an alternative `output` arg, so maybe I can do it for that...

cylc/flow/config.py
wxtim: I'm sure it'd not be the first time we've had to write custom config warnings outside the parsec deprecation framework....

cylc/flow/scripts/workflow_state.py
hjoliver: I considered that, but in principle you could query the connected object multiple times with different cycle points.

cylc/flow/scripts/workflow_state.py
hjoliver: c.f. my comment on the xtrigger - the problem is we need the "--output" option for the trigger name now, not the corresponding task message, and on master `--message` and `--output` both query the task message, which makes back-compat difficult. 

However, now that I have managed to achieve back-compat for the xtrigger, I think I can do it for the CLI as well by considering all the command line as a whole.

tests/functional/workflow-state/05-output.t
hjoliver: Does this matter?  We don't enforce our Python line length in these bash scripts.

tests/unit/test_config.py
hjoliver: I presume this is comment rather than critique 😁 ?

cylc/flow/cfgspec/workflow.py
```diff
@@ -1995,6 +1973,25 @@ def upg(cfg, descr):
...
+    u.deprecate(
+        '8.0.0',
+        ['runtime', '__MANY__', 'suite state polling'],
+        ['runtime', '__MANY__', 'workflow state polling'],
+        silent=cylc.flow.flags.cylc7_back_compat,
+        is_section=True,
+    )
+    for item in ('host', 'user'):
+        u.obsolete(
+            '8.0.0', ['runtime', '__MANY__', 'workflow state polling', item],
```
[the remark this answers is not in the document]
hjoliver: Maybe not - host and user were not used in any version of Cylc 8.  So they were technically obsolete despite not been marked as such during parsing.

cylc/flow/xtrigger_mgr.py
hjoliver: The "weirdness" allows us to avoid recomputing the offset from cycle point every time a clock trigger is called. We do it once, and then call `_wall_clock` instead with an absolute time.

> the if condition above.

retries timers are implemented via wall_clock xtriggers

> validation

(This was needed, but I need to remind myself of why...)

cylc/flow/cfgspec/workflow.py
```diff
@@ -1730,57 +1730,34 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+                "user" and "host" config items are not supported.
 
-                The items in this section reflect
-                options and defaults of the ``cylc workflow-state`` command,
-                except that the target workflow ID and the
-                ``--task``, ``--cycle``, and ``--status`` options are
-                taken from the graph notation.
-
-                .. versionchanged:: 8.0.0
+                .. versionchanged:: 8.3.0
```
MetRonnie, line 1737: This was changed back in 8.0.0

```suggestion
                .. versionchanged:: 8.0.0
```

cylc/flow/parsec/upgrade.py
hjoliver: (see other comment on this point)

cylc/flow/cfgspec/workflow.py
```diff
@@ -1730,57 +1730,34 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
-
-                    For your own workflows, there is no need to set this as it
-                    is always ``~/cylc-run/``. But for other workflows,
-                    (e.g those owned by others), or mirrored workflow databases
-                    use this item to specify the location of the top level
-                    cylc run directory (the database should be in a the same
-                    place relative to this location for each workflow).
+                Conf('alt-cylc-run-dir', VDR.V_STRING, desc='''
+                    The cylc-run directory location of the target workflow.
+                    Use to poll workflows owned by other users.
```
MetRonnie, line 1757: This change is missing an upgrader?

cylc/flow/workflow_db_mgr.py
hjoliver: Actually it wasn't already imported. But I'm not sure why I did it at this point in the file!

cylc/flow/workflow_db_mgr.py
hjoliver: (I need to check this)

cylc/flow/cfgspec/workflow.py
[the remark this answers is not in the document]
MetRonnie: At least `message -> output` below should be 8.3.0

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,260 @@
...
+
+    poller = WorkflowPoller(
+        id_,
+        options.offset,
+        options.flow_num,
+        options.alt_cylc_run_dir,
+        None,  # default status
+        options.is_output,
+        options.old_format,
+        f'"{id_}"',
```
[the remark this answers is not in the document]
hjoliver: It gets passed through the super-class init, as a description of the condition being polled for.

tests/unit/test_config.py
oliver-sanders: I think that's mouse - crossproduct - pig

hjoliver:
Most but not all comments addressed - I pushed my changes up just now.

I found to a way to get back-compat on the workflow_state xtrigger via `inspect` binding of args to function signatures. But it took  a bit of working out.

TODO:
- address remaining small comments
- get the back-compat workflow_state xtrigger to query task messages instead of outputs
- having done it with the xtrigger, try to get back-compat for the CLI (and hence automatic polling tasks too)

tests/functional/workflow-state/05-output.t
oliver-sanders: The suggestion is much more readable, but I tend not to raise things like this at review time.

hjoliver:
xtrigger back-compat test case:
```
[scheduling]
    initial cycle point = 2001
    [[xtriggers]]
        new = workflow_state(dem7//2001/foo:succeeded):PT2S  # new-style
        old = workflow_state(dem7, foo, 2001, status=succeeded, message=None):PT2S  # old-style 
    [[graph]]
        R1 = """
           @old => bar
           @new => baz
        """
[runtime]
    [[foo, bar, baz]
```

cylc/flow/xtrigger_mgr.py
```diff
@@ -261,32 +259,61 @@ def _validate(
...
 
         # Validate args and kwargs against the function signature
         try:
-            bound_args = sig.bind(
-                *fctx.func_args, **fctx.func_kwargs
-            )
+            bound_args = sig.bind(*fctx.func_args, **fctx.func_kwargs)
         except TypeError as exc:
-            raise XtriggerConfigError(label, f"{sig_str}: {exc}")
+            # try fname_backcompat
```
MetRonnie, line 278: I don't think this should be done on a general basis, it should only be done for `workflow_state` - I'm happy to write a PR against this one if you'd like

hjoliver: This seems pretty safe to me - can you see it causing trouble somehow? 

Anyway, I don't mind if you want to tweak it to be workflow_state-specific - it's unlikely (if not impossible) to be more widely useful.

(I think we should keep my change to allow different function and module names though).

cylc/flow/config.py
```diff
@@ -1561,7 +1574,7 @@ def configure_workflow_state_polling_tasks(self):
                     ('alt-cylc-run-dir', ' --%s=%s')]:
                 if rtc['workflow state polling'][key]:
                     comstr += fmt % (key, rtc['workflow state polling'][key])
-            script = "echo " + comstr + "\n" + comstr
+            script = "set -x; " + comstr + "\n" + comstr
```
wxtim, line 1577: Sorry? I was happy with keeping the echo in the end; and if you are going to do this, you only want 1 copy of `comstr`.

wxtim:
I ran a workflow on master, stopped it and restarted on your branch - the DB looks like this: Is that OK?  

```
1291|foo|[1]|["submitted", "started", "succeeded", "common message"]
1292|foo|[1]|{"submitted": "submitted", "started": "started", "succeeded": "succeeded", "selector2": "common message"}
```

cylc/flow/xtrigger_mgr.py
MetRonnie: The code will be more straightforward if we only handle the case of `workflow_state`

> (I think we should keep my change to allow different function and module names though).

I don't think this is used though?

cylc/flow/config.py
hjoliver: No I wasn't going to do that. I think I was experimenting with your suggesting and must have committed it 🤦‍♂️

hjoliver: (Fixed)

cylc/flow/xtrigger_mgr.py
hjoliver: It's used for finding the "validate" function in the xtrigger module.

MetRonnie:
```console
$ cylc workflow-state gimli --output 'and my axe'
InputError: Please give a single ID
```

Not an especially clear error message for someone using the old CLI syntax

hjoliver:
> $ cylc workflow-state gimli --output 'and my axe'
> InputError: Please give a single ID
>
> Not an especially clear error message for someone using the old CLI syntax.

That's because the option as become a flag, to interpret the task ID selector as an output (trigger) as opposed to a status or a task message. 

I suppose I could change the names of these flags to `--outputs` and `--messages`, to avoid this. Since multiple outputs will be returned if they match the command line.

MetRonnie:
You mentioned transient statueses in the videoconf today. This currently does not work once the task status is suceeded:
```
workflow_state(gimli//1/foo:submitted)
```

For a Cylc 7 DB we have no way around this other than the user adding `is_output=True` to the function call.

But for Cylc 8 DBs we can just query the `task_outputs` table instead, no need to check the `task_states` table at all, right?

hjoliver:
For consistency with previous usage, and with the command name "workflow-**state**", I've stuck with interpreting standard names ("submitted" etc.) as statuses by default, but now you can choose to query the corresponding output instead (with `--output`) to avoid the transient status problem. If you do that, then the task_states table is not checked at all.   

In an earlier iteration of this branch I did in fact automatically translate transient states to outputs, but bailed out of that for the reasons above, and because it could potentially be confusing in the case of `running` (status) `--> `started` (output) - might look like getting something you didn't ask for.

Is that what you're asking?

MetRonnie:
Ah right, I forgot that statuses include "running", "waiting" etc. Never mind

hjoliver:
Tests all passing again, this time with manual output completion recorded in the DB.  There's still a bit more to do though:
- trawl through the comments above
- see if I can make the CLI back-compatible so as not to break existing hand-rolled polling tasks.

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,404 @@
 
 r"""cylc workflow-state [OPTIONS] ARGS
 
-Retrieve task states from the workflow database.
+Deprecated support for existing scripts:
+    cylc workflow-state --task=NAME --point=CYCLE --status=STATUS
+       --output=MESSAGE --message=MESSAGE WORKFLOW
```
MetRonnie, line 23: This shouldn't be mentioned until later (currently breaks the `cylc help all` description)

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,404 @@
...
+        action="store", type="int", dest="flow_num", default=None)
 
     parser.add_option(
-        "-d", "--run-dir",
-        help="The top level cylc run directory if non-standard. The "
-             "database should be DIR/WORKFLOW_ID/log/db. Use to interrogate "
-             "workflows owned by others, etc.; see note above.",
-        metavar="DIR", action="store", dest="alt_run_dir", default=None)
+        "--triggers",
+        help="Task selector should match output triggers rather than status."
```
MetRonnie, line 298:
```suggestion
        help="Task selector should match output triggers rather than status. "
```

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,404 @@
...
-             ("', '").join(
-                 sorted(CylcWorkflowDBChecker.STATE_ALIASES.keys())[:-1]) +
-             "' and '" + sorted(
-                 CylcWorkflowDBChecker.STATE_ALIASES.keys())[-1] + "'. ")
-    states = ("Valid states to check for include: '" +
-              ("', '").join(TASK_STATUSES_ORDERED[:-1]) +
-              "' and '" + TASK_STATUSES_ORDERED[-1] + "'.")
+        "--messages",
+        help="Task selector should match output messages rather than status."
+             "(For legacy support - better to use --output).",
```
MetRonnie, line 305:
```suggestion
```

cylc/flow/option_parsers.py
```diff
@@ -290,9 +290,15 @@ class CylcOptionParser(OptionParser):
...
-            default=True, useif='all'),
+            ['--timestamp'],
+            help='Add a timestamp to messages logged to the terminal.',
+            action='store_true', dest='log_timestamp',
+            default=False, useif='all'),
+        OptionSettings(
+            ['--no-timestamp'], help="Don't add a timestamp to messages logged"
+            " to the terminal (this does nothing - it is now the default.",
+            action='store_false', dest='_noop',
+            default=False, useif='all'),
```
MetRonnie, line 301: Off-topic?

hjoliver: Arguably yes, but it was motivated by the fact that this PR touches code used by both the CLI and the scheduler, so log messages can appear in the terminal, in which case the long timestamp sticks out like a sore thumb and is really not helpful.

This means `cylc play --no-detach` won't log timestamps to the terminal by default, but I don't see any downside to that. The log file itself still has timestamps.

So I'd prefer to leave this in.

cylc/flow/scripts/workflow_state.py
hjoliver: Oops yes, fixed.

changes.d/5809.break.d
```diff
@@ -0,0 +1,11 @@
...
+instead of separate arguments for cycle point, task name, etc. 
+
+The Cylc 7 suite_state xtrigger is still supported, with separate arguments, but 
+is deprecated.
+
+Automatic workflow state polling tasks (via special graph syntax) are still
+supported, but deprecated: use the workflow_state xtrigger instead.
+
+The owner and host arguments are no longer supported, for workflow state polling.
+To poll another user's workflow you must be able to see their run directory.
```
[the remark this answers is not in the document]
hjoliver: Of the four paragraphs:
- 1. is no longer breaking (removed)
- 2. and 3. were intended to address an obvious question in light of 1., but they don't really need to be mentioned in the change log (removed)
- 4. still valid

cylc/flow/xtriggers/workflow_state.py
hjoliver: This is no longer a breaking change - I worked around it.

cylc/flow/scripts/workflow_state.py
hjoliver: No longer a breaking change. (I was able to do this, because the old `--output` and `--message` args were only a problem for 8.0-8.2.x DBs, which only store the message).

cylc/flow/cfgspec/workflow.py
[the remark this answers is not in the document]
hjoliver: > Maybe not - host and user were not used in any version of Cylc 8. So they were technically obsolete despite not been marked as such during parsing.

Actually, worse than that - if host or user were configured, it would create a broken command line for the auto polling task.

hjoliver:
(Most comments address, but not all yet).

This is no longer a breaking change, even for the CLI - i.e. it won't break existing hand-rolled workflow-state polling tasks.

cylc/flow/option_parsers.py
MetRonnie: It just means it will take longer to review...

hjoliver: It's a tiny change! (But if you really insist, I'll take it out).

cylc/flow/xtrigger_mgr.py
```diff
@@ -159,6 +164,244 @@ class TemplateVariables(Enum):
...
+
+        """
+        if label in self.functx_map:
+            # we've already seen this one
+            return
+
+        if (
+            not label.startswith('_cylc_retry_') and not
+            label.startswith('_cylc_submit_retry_')
+        ):
```
MetRonnie, line 206:
```suggestion
        if fctx.func_name != _wall_clock.__name__:
```

cylc/flow/cfgspec/workflow.py
```diff
@@ -1730,57 +1730,37 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
                 ''')
                 Conf('max-polls', VDR.V_INTEGER, desc='''
-                    The maximum number of polls before timing out and entering
-                    the "failed" state.
+                    Maximum number of polls to attempt before the task fails.
+                ''')
+                Conf('output', VDR.V_STRING, desc='''
+                    Target task output (trigger name, not task message).
+                    Alternatively, put this in the graph trigger syntax.
                 ''')
```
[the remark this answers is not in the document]
hjoliver: I think I was trying to fix this config so that it matches the (old-format CLI args), which it was always supposed to do. But yeah, no one could be relying on this item in existing usage, and we're deprecating the feature, so adding it is arguably a bad idea... I'll remove it.

cylc/flow/option_parsers.py
MetRonnie: Maybe, but these things add up!

changes.d/5809.feat.d
```diff
@@ -0,0 +1,2 @@
+The workflow-state command and xtrigger now take univeral IDs instead of separate
+arguments for cycle point, task name, etc., and are flow-aware.
```
MetRonnie, line 2:
```suggestion
The workflow-state command and xtrigger are now flow-aware and take universal IDs instead of separate arguments for cycle point, task name, etc. (which are still supported, but deprecated). 
```

cylc/flow/cfgspec/workflow.py
```diff
@@ -1730,57 +1731,33 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
-                    run on the remote account.
-                ''')
-                Conf('host', VDR.V_STRING, desc='''
-                    The hostname of the target workflow.
+                .. deprecated:: 8.3.0
 
-                    The polling
-                    ``cylc workflow-state`` command will be run there.
-                ''')
+                   Please switch to workflow_state xtriggers.
```
MetRonnie, line 1744:
```suggestion
                   Please switch to :ref:`workflow_state xtriggers <Built-in Workflow State Triggers>`.
```
Or whatever makes Sphinx and Flake8 happy

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,403 @@
...
-  # cylc task.
+
+  # Print the status of all tasks in WORKFLOW:
+  $ cylc workflow-state WORKFLOW
+
+  # Print the status of all tasks in cycle point 2033:
+  $ cylc workflow-state WORKFLOW//2033
+
+  # Print the status of all tasks named foo:
+  $ cylc workflow-state WORKFLOW//*/foo
```
MetRonnie, line 75:
```suggestion
  $ cylc workflow-state "WORKFLOW//*/foo"
```

changes.d/5809.fix.d
```diff
@@ -0,0 +1 @@
+The Workflow-state command now handles task status and outputs equally.
```
MetRonnie, line 1:
```suggestion
The `cylc workflow-state` command now polls for task status (like it does for outputs).
```

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,403 @@
...
 from cylc.flow.workflow_files import infer_latest_run_from_id
-
-from metomi.isodatetime.parsers import TimePointParser
+from cylc.flow.task_state import TASK_STATUSES_ORDERED
 
 if TYPE_CHECKING:
     from optparse import Values
 
 
+# TODO: flow=all, none?  Useful for CLI if not xrigger, pt format.
```
MetRonnie, line 115: Confusing TODO still here

cylc/flow/option_parsers.py
hjoliver: Thanks (and apologies 😁  - but I don't think there's other off-topic changes in this PR)

changes.d/5809.fix.d
hjoliver: It's really the other way around but not quite as simple as that, so I was trying to avoid the specifics.

cylc/flow/xtriggers/workflow_state.py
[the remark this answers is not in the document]
hjoliver: fixed

cylc/flow/xtrigger_mgr.py
hjoliver: Actually, I had to revert your suggestion - it doesn't cover all the bases:

```
FAILED tests/integration/test_simulation.py::test_settings_reload - cylc.flow.exceptions.XtriggerConfigError: [@_cylc_retry_1066/one] wall_clock(trigger_time=1718159104.6357827)
got an unexpected keyword argument 'trigger_time'
```

cylc/flow/scripts/workflow_state.py
hjoliver: removed

hjoliver:
> $ cylc workflow-state gimli --output 'and my axe'
> InputError: Please give a single ID
> 
> Not an especially clear error message for someone using the old CLI syntax.

Works now.

hjoliver:
DONE

wxtim:
@hjoliver  It's going to want a changes.d entry?

cylc/flow/xtriggers/workflow_state.py
```diff
@@ -14,18 +14,117 @@
...
+            must be an integer
+        alt_cylc_run_dir:
+            must be a valid path
+
+    """
+    try:
+        tokens = tokenise(args["workflow_task_id"])
+    except KeyError:
+        raise WorkflowConfigError(
+            # TODO better message
```
wxtim, line 107: Do you still want this TODO?

cylc/flow/xtriggers/workflow_state.py
```diff
@@ -14,18 +14,117 @@
...
+            must be a valid status
+        flow_num:
+            must be an integer
+        alt_cylc_run_dir:
+            must be a valid path
+
+    """
+    try:
+        tokens = tokenise(args["workflow_task_id"])
+    except KeyError:
```
wxtim, line 105: What input was getting a `KeyError` from this?

wxtim:
I've started a new testing branch at https://github.com/hjoliver/cylc-flow/pull/52 - unfortunately it's failing for a good reason.

cylc/flow/config.py
```diff
@@ -1513,6 +1512,19 @@ def adopt_orphans(self, orphans):
             self.runtime['linearized ancestors'][orphan] = [orphan, 'root']
 
     def configure_workflow_state_polling_tasks(self):
+
+        # Deprecation warning - automatic workflow state polling tasks don't
+        # necessarily have deprecated config items outside the graph string.
+        if (
```
wxtim, line 1518: IMO this should work for Cylc VR and VIP too.

hjoliver:
> @hjoliver It's going to want a changes.d entry?

It's already got two of 'em!

cylc/flow/xtriggers/workflow_state.py
hjoliver: Not really, it'll do. (Note to self - grep branch diffs for "TODO"!)

hjoliver: (removing the TODO now)

cylc/flow/xtriggers/workflow_state.py
hjoliver: Fair question. I think that pre-dates the new function signature introspection. Removing it...

cylc/flow/xtriggers/workflow_state.py
```diff
@@ -14,18 +14,117 @@
...
+        tokens[token] is None
+        for token in ("workflow", "cycle", "task")
+    ):
+        raise WorkflowConfigError(
+            "Full ID needed: workflow//cycle/task[:selector].")
+
+    try:
+        int(args["flow_num"])
+    except ValueError:
+        raise WorkflowConfigError("flow_num must be an integer.")
```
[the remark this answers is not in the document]
hjoliver: But `"3.141492"` (str) would. I think we used to parse all xtrigger args from the config file as strings, and *convert* them to type like this, in which case `isinstance` would fail. `int("1") = 1`. But we seem to be doing better than that now (tested), so probably should do it your way... changing it

hjoliver:
@wxtim - I just merged your small side-PR with tests (all tests passed on it, contrary to what you said above - but maybe that was before you made the isinstance change.

cylc/flow/config.py
hjoliver: I'm bailing out for the night, but if the fix is as small as it seems (on your side-branch) I don't mind if you either:
- merge it to this branch yourself (after testing it)
- or re-raise it against master

... especially if it means this PR can get merged ASAP (for 8.3.0...)

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,402 @@
...
-        "requested state",
-        options.interval,
-        options.max_polls,
-        args=pollargs,
+def main(parser: COP, options: 'Values', *ids: str) -> None:
+
+    if len(ids) != 1:
+        raise InputError("Please give a single ID")
+
+    id_ = ids[0].rstrip('/')  # might get 'id/' due to autcomplete
```
MetRonnie, line 362: Can we use `cylc.flow.id_cli.parse_id` instead?

cylc/flow/xtriggers/workflow_state.py
```diff
@@ -14,18 +14,115 @@
...
+
+    if any(
+        tokens[token] is None
+        for token in ("workflow", "cycle", "task")
+    ):
+        raise WorkflowConfigError(
+            "Full ID needed: workflow//cycle/task[:selector].")
+
+    if not isinstance(args["flow_num"], int):
+        raise WorkflowConfigError("flow_num must be an integer.")
```
MetRonnie, line 113: But `flow_num` can be `None` too?

cylc/flow/config.py
```diff
@@ -1531,25 +1543,42 @@ def configure_workflow_state_polling_tasks(self):
...
-                "cylc workflow-state"
-                f" --task={tdef.workflow_polling_cfg['task']}"
-                " --point=$CYLC_TASK_CYCLE_POINT"
+                "cylc workflow-state "
+                f"{tdef.workflow_polling_cfg['workflow']}//"
+                "$CYLC_TASK_CYCLE_POINT/"
+                f"{tdef.workflow_polling_cfg['task']}"
             )
+            graph_trigger = tdef.workflow_polling_cfg['status']
+            config_trigger = rtc['workflow state polling']['message']
```
MetRonnie, line 1552: `trigger = message` is a bit confusing

cylc/flow/xtrigger_mgr.py
```diff
@@ -159,6 +164,244 @@ class TemplateVariables(Enum):
...
+        self.functx_map[label] = fctx
+
+        if fctx.func_kwargs.pop(
+            'sequential',
+            self.sequential_xtriggers_default
+        ):
+            self.sequential_xtrigger_labels.add(label)
+
+        if fctx.func_name == "wall_clock":
+            self.wall_clock_labels.add(label)
```
MetRonnie, line 219:
```suggestion
        if fctx.func_name == "wall_clock":
            self.wall_clock_labels.add(label)
        else:
            self._validate(label, fctx, fdir)

        self.functx_map[label] = fctx

        if fctx.func_kwargs.pop(
            'sequential',
            self.sequential_xtriggers_default
        ):
            self.sequential_xtrigger_labels.add(label)    
```

cylc/flow/config.py
wxtim: I can't merge it.

cylc/flow/cfgspec/workflow.py
```diff
@@ -1731,57 +1732,34 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
-                ''')
-                Conf('host', VDR.V_STRING, desc='''
-                    The hostname of the target workflow.
+                .. deprecated:: 8.3.0
 
-                    The polling
-                    ``cylc workflow-state`` command will be run there.
-                ''')
+                Please use the workflow_state xtrigger instead:
+                :py:mod:`cylc.flow.xtriggers.workflow_state`.
```
MetRonnie, line 1746: Docs build error (on https://github.com/cylc/cylc-doc/pull/727):
```
Warning, treated as error:
<unknown>:2779:py:mod reference target not found: cylc.flow.xtriggers.workflow_state
```

```suggestion
                Please use the :ref:`workflow_state xtrigger
                <Built-in Workflow State Triggers>` instead:
```

hjoliver: Hmm, thought I had fixed this. I must not have pushed the commit from my laptop (late night).

cylc/flow/config.py
hjoliver: Huh, I thought you should be able to (possibly after "reviewing" it yourself. Anyhow, I'll do it.

cylc/flow/xtriggers/workflow_state.py
hjoliver: Actually no, it is defaulting to 1 despite the misleading type hint. But it should default to None. Fixed.

cylc/flow/config.py
hjoliver: Clarified.

cylc/flow/scripts/workflow_state.py
hjoliver: It would be a bit cleaner to use `parse_ids` but the problem is that aborts if the workflow is not found - i.e., if it is not installed yet.

Here we want to be able to start polling before the upstream workflow is installed - it just makes it a bit easier to get a set of interdependent workflows up and running.

I could probably do it by moving ID parsing into the WorkflowPoller class, but it's not an entirely trivial refactor. So given this works as-is, let's leave it for now. I've added a comment in the code about this.

cylc/flow/xtriggers/workflow_state.py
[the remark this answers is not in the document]
hjoliver: (Now tweaked further to handle the None case, after changing the default to None).

cylc/flow/xtriggers/workflow_state.py
```diff
@@ -14,18 +14,127 @@
...
+    is_output: bool = False,
+    is_message: bool = False,
+    alt_cylc_run_dir: Optional[str] = None,
+) -> Tuple[bool, Dict[str, Any]]:
+    """Connect to a workflow DB and check a task status or output.
+
+    If the status or output has been achieved, return {True, result}.
+
+    Arg:
+        workflow_task_id:
```
wxtim, line 40: Can you describe `is_message` in the docstring?

MetRonnie: And `is_output`

hjoliver: Done. And I changed is_output to is_trigger (everywhere) in light of us recently settling on "output" = "trigger: message".

hjoliver:
Roger that, good job @wxtim (seriously :-) .  I'll come back after feeding the kids...

hjoliver:
One-line fix pushed for the function signature advice (we shouldn't advise adding all the null args with `= None`).

hjoliver:
And just in case, another small commit handles "None" properly in xtrigger defintions, just in case some one does it.

hjoliver:
> and the same as the above but with alt_cylc_run_dir=None added to the xtrigger args causes it the xtrigger to not return True.

That would be because it was interpreting the string "None" as an alternate run-dir.  My previous commit will prevent that too.

wxtim (approved):
I've had a really decent go at breaking this, and run out of ideas.

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,409 @@
...
+                self.result, self.old_format, self.pretty_print)
+
+        return bool(self.result)
 
 
 def get_option_parser() -> COP:
     parser = COP(
         __doc__,
-        argdoc=[WORKFLOW_ID_ARG_DOC]
+        argdoc=[ID_MULTI_ARG_DOC]
```
MetRonnie, line 272: Why multi, when it's only supposed to take 1 ID?

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,409 @@
...
-        raise InputError(
-            "You must target a cycle point to use an offset")
-
-    # Attempt to apply specified offset to the targeted cycle
-    if options.offset:
-        options.cycle = str(add_offset(options.cycle, options.offset))
-
-    # Exit if both task state and message are to being polled
-    if options.status and options.msg:
-        raise InputError("cannot poll both status and custom output")
```
MetRonnie, line 219: This check has been lost

cylc/flow/scripts/workflow_state.py
```diff
@@ -18,252 +18,409 @@
...
-def main(parser: COP, options: 'Values', workflow_id: str) -> None:
-
-    if options.use_task_point and options.cycle:
-        raise InputError(
-            "cannot specify a cycle point and use environment variable")
-
-    if options.use_task_point:
-        if "CYLC_TASK_CYCLE_POINT" not in os.environ:
-            raise InputError("CYLC_TASK_CYCLE_POINT is not defined")
-        options.cycle = os.environ["CYLC_TASK_CYCLE_POINT"]
```
MetRonnie, line 207: This has been lost, `--task-point` doesn't do anything anymore

cylc/flow/scripts/workflow_state.py
hjoliver: Ah, yeah, so I originally removed all of these options, then ended up restoring them for CLI back-compat, to avoid breaking existing hand-rolled polling tasks.  I'll restore the check...

cylc/flow/scripts/workflow_state.py
hjoliver: oops, partial restore. Fixed.

cylc/flow/scripts/workflow_state.py
hjoliver: Because WORKFLOW_ID_ARG_DOC is for workflow, not workflow/task.

I've now created a new doc string, for single ID.

hjoliver:
@MetRonnie - final comments addressed.  Merge? 🙏

cylc/flow/dbstatecheck.py
```diff
@@ -138,49 +255,103 @@ def workflow_state_query(
...
+        results = []
+        for row in db_res:
+            outputs: Union[Dict[str, str], List[str]] = json.loads(row[2])
+            if isinstance(outputs, dict):
+                messages: Iterable[str] = outputs.values()
+            else:
+                # Cylc 8 pre 8.3.0 back-compat: list of output messages
+                messages = outputs
+                if warn_output_fallback:
+                    LOG.warning(output_fallback_msg)
```
MetRonnie, line 325: This message doesn't get through when using the `workflow_state` xtrigger. It appears the logging handler is the null handler in this case, not sure why, here is a workaround
```suggestion
                    print(f"WARNING - {output_fallback_msg}", file=sys.stderr)
```

MetRonnie: I've pushed it

MetRonnie (commented):
Alright, think we're nearly there. However there is this still outstanding? https://github.com/hjoliver/cylc-flow/pull/54

MetRonnie (approved):
💫

wxtim (approved):
🚀

PR #5813, opened 2023-11-09
actions: patch DNS issues on Mac OS runners

oliver-sanders:
There are DNS issues with the GitHub Mac OS runners resulting in reverse lookup failures causing Cylc test failures.

Hopefully the networking will be fixed before long, in the mean time, this workaround was suggested on the ticket.

See https://github.com/actions/runner-images/issues/8649

### DIFF
--- a/.github/workflows/test_fast.yml
+++ b/.github/workflows/test_fast.yml
@@ -51,6 +51,13 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y sqlite3
 
+      - name: Patch DNS
+        if: startsWith(matrix.os, 'macos')
+        run: |
+          # see https://github.com/actions/runner-images/issues/8649#issuecomment-1779548056
+          echo -e "$(ipconfig getifaddr en0) $(hostname -f) $(hostname -s)" | sudo tee -a /etc/hosts
+          dscacheutil -q host -a name $(hostname -f)
+
       - name: Install
         run: |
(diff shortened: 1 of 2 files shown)


oliver-sanders:
This patch fixes most, but not all of the DNS issues.

It would appear the hostname returned by reverse lookup results in "Host key verification failed":

```
2023-11-09T15:36:26+05:30 DEBUG - $ ssh -oBatchMode=yes -oConnectTimeout=10 Mac-1699523276107.local env CYLC_VERSION=8.3.0.dev CYLC_CONF_PATH=/private/var/folders/3s/vfzpb5r51gs6y328rmlgzm7c0000gn/T/tmp.FxeHKbLvTe/etc CYLC_COVERAGE=1 bash --login -c 'exec "$0" "$@"' cylc psutil  # returned 255
    Host key verification failed.
```

There is a Cylc / SSH configuration hack that could potentially be used to work around this.

MetRonnie (approved):
Can we get this merged (into 8.2.x) for the time being?

oliver-sanders:
It doesn't work.

MetRonnie:
IMO it's better than the current situation because it works for the fast tests and cuts down the functional test failures to ~4

oliver-sanders:
For the time being we have worked around the issue by downgrading to the macos 11 images (note 12 is the current default image, the 13 beta image has the same issue as the 12 image).

PR #5819, opened 2023-11-13
fix broken test

wxtim:
**Check List**


.github/workflows/test_functional.yml
```diff
@@ -130,6 +130,7 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y sqlite3 tree at
+          sudo apt-get install -y pcregrep
```
oliver-sanders, line 133: Sorry, I got that wrong, I didn't need to install anything funny on Linux, but I did have to install GNU grep on Mac OS (otherewise you get BSD grep which doesn't have the `-P` option).

Here's the PR where I was making very similar changes to the same tests (but for different reasons).

https://github.com/cylc/cylc-flow/pull/5769/files

conda-environment.yml
```diff
@@ -2,6 +2,7 @@ name: cylc-dev
 channels:
   - conda-forge
 dependencies:
+  - aiosmtpd
```
MetRonnie, line 5: This is only an optional dependency
```suggestion
```

oliver-sanders (approved):

MetRonnie (approved):

PR #5821, opened 2023-11-15
log_vc_info: handle long command output

oliver-sanders:
* Log the commands run by log_vc_info (debug level).
* Handle the risk of large command output clogging up the buffer causing commands to hang.

**Check List**

### DIFF
--- a/cylc/flow/install_plugins/log_vc_info.py
+++ b/cylc/flow/install_plugins/log_vc_info.py
@@ -63,12 +63,22 @@ import json
 from pathlib import Path
 from subprocess import Popen, DEVNULL, PIPE
 from typing import (
-    Any, Dict, Iterable, List, Optional, TYPE_CHECKING, TextIO, Union, overload
+    Any,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    TYPE_CHECKING,
+    TextIO,
+    Union,
+    overload,
 )
 
 from cylc.flow import LOG as _LOG, LoggerAdaptor
 from cylc.flow.exceptions import CylcError
 import cylc.flow.flags
+from cylc.flow.pipe_poller import pipe_poller
+from cylc.flow.util import format_cmd
 from cylc.flow.workflow_files import WorkflowFiles
 
 if TYPE_CHECKING:
@@ -171,7 +181,7 @@ def get_vc_info(path: Union[Path, str]) -> Optional[Dict[str, Any]]:
             ):
                 LOG.debug(f"Source dir {path} is not a {vcs} repository")
             elif cylc.flow.flags.verbosity > -1:
-                LOG.warning(f"$ {vcs} {' '.join(args)}\n{exc}")
+                LOG.warning(f"$ {vcs} {format_cmd(args)}\n{exc}")
             continue
 
         info['version control system'] = vcs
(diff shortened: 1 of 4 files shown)


cylc/flow/pipe_poller.py
```diff
@@ -0,0 +1,75 @@
...
+        # read any data from files
+        nonlocal chunk_size, files
+        for file in select(list(files), [], [], timeout)[0]:
+            buffer = file.read(chunk_size)
+            if len(buffer) > 0:
+                _files[file] += buffer
+
+    while proc.returncode is None:
+        # poll the process (updates the returncode)
+        proc.poll()
```
MetRonnie, line 69:
```suggestion
    while proc.poll() is None:
```

cylc/flow/pipe_poller.py
```diff
@@ -0,0 +1,75 @@
...
+def pipe_poller(proc, *files, chunk_size=4096):
+    """Read from a process without hitting buffer issues.
+
+    Standin for subprocess.Popen.communicate.
+
+    When PIPE'ing from subprocesses, the output goes into a buffer. If the
+    buffer gets full, the subprocess will hang trying to write to it.
+
+    This function polls the process, reading output from the buffers into
+    memory to prevent them from filling up.
```
MetRonnie, line 38: In what cases might you call this function that isn't the case of `pipe_poller(proc, proc.stdout, proc.stderr)`? I.e. when would `files` be something other than the output of `proc`?

MetRonnie (approved):

MetRonnie (commented):
(Oh, still need changelog)

MetRonnie (approved):

cylc/flow/install_plugins/log_vc_info.py
```diff
@@ -63,12 +63,22 @@
...
-    Any, Dict, Iterable, List, Optional, TYPE_CHECKING, TextIO, Union, overload
+    Any,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    TYPE_CHECKING,
+    TextIO,
+    Union,
+    overload,
```
wxtim, line 74: Pedantic
```suggestion
    TYPE_CHECKING,
    Any,
    Dict,
    Iterable,
    List,
    Optional,
    TextIO,
    Union,
    overload,
```

cylc/flow/pipe_poller.py
```diff
@@ -0,0 +1,73 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Utility for preventing pipes from getting clogged up.
+
+If you're reading files from Popen (i.e. to extract command output) where the
+command output has the potential to be long-ish, then you should use this
+function to protect against the buffer filling up.
+
+Note, these is a more advanced version of this baked into the subprocpool.
```
wxtim, line 23: I have no idea which is the more advanced version from this.

```suggestion
Note, there is a more advanced version of this baked into the subprocpool.
```
or
```suggestion
Note, this is a more advanced version of a pipe_poller baked into the subprocpool.
```

wxtim (commented):
Seems generally sane. 

I wonder if it might be good to have a functional test which uses bash `timeout` to demonstrate that this approach is safer (i.e. to get an actual script to hang with proc.communicate).

oliver-sanders:
I don't think using a Bash timeout will replicate the problem being fixed here?

MetRonnie (approved):

wxtim (approved):

wxtim:
Have approved - the MacOs Tests look to be failing in a standard way. I've kicked flaky and the tutorial tests just to be sure.

oliver-sanders:
The tutorials tests are failing here as on master at the install step.

PR #5832, opened 2023-11-21
Fix `IndepQueueManager` test

### LINKED ISSUE #5812: Testing: Mock object will always return True!
Investigate replacing `tests/unit/test_indep_task_queues.py` with an integration test.

### Description

The code at https://github.com/cylc/cylc-flow/blob/36a6ce9fe1ff8030caeb93824959d46616b1ffae/tests/unit/test_indep_task_queues.py#L96 does not work as desired: One may add a line such as

```python
        assert r.state.garbage_method.called_with(TASK_STATUS_PREPARING)
```
Which will also pass!

As explained in https://github.com/python/cpython/issues/100690 this is because up to Python 3.11 

```python
(Pdb) type(r.state.reset.called_with)
<class 'unittest.mock.Mock'>
```
which will return true, and not as the reader would expect, test the truth of that method having been called.

At Python 3.12 additional checks have been added, preventing this issue, but breaking the tests.

Investigate

Check whether the method under test should be called. I don't think that the state of the task should change - it should be waiting before and after so I suspect that this test shouldn't pass, and just needs correcting. We might want to test the task proxy's `is_queued` flag though.

It might be more sensible to run this as an integration test rather than a unit test.

MetRonnie:
Closes #5812

**Check List**

- [x] No dependency changes 
- [x] Tests are included 
- [x] No `CHANGES.md` entry needed
- [x] No docs needed

### DIFF
--- a/tests/unit/test_indep_task_queues.py
+++ b/tests/unit/test_indep_task_queues.py
@@ -21,8 +21,8 @@ from unittest.mock import Mock
 
 import pytest
 
+from cylc.flow.task_proxy import TaskProxy
 from cylc.flow.task_queues.independent import IndepQueueManager
-from cylc.flow.task_state import TASK_STATUS_PREPARING
 
 
 MEMBERS = {"a", "b", "c", "d", "e", "f"}
@@ -61,9 +61,7 @@ READY_TASK_NAMES = ["b3", "s4", "o2", "s3", "b4", "o3", "o4", "o5", "o6", "o7"]
 
 
 @pytest.mark.parametrize(
-    "active,"
-    "expected_released,"
-    "expected_foo_groups",
+    "active, expected_released, expected_foo_groups",
     [
         (
             Counter(["b1", "b2", "s1", "o1"]),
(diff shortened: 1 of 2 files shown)


tests/unit/test_indep_task_queues.py
```diff
@@ -73,28 +71,24 @@
...
         queue_mgr.push_task(itask)
 
     # release tasks, given current active task counter
     released = queue_mgr.release_tasks(active)
-    assert sorted([r.tdef.name for r in released]) == sorted(expected_released)
-
-    # check released tasks change state to "preparing", and not is_queued
-    for r in released:
-        assert r.state.reset.called_with(TASK_STATUS_PREPARING)
-        assert r.state.reset.called_with(is_queued=False)
```
MetRonnie, line 97: `itask.state.reset()` is only ever called by `TaskPool.release_queued_tasks()`, so not applicable for this test

oliver-sanders (approved):

Issue #5825, opened 2023-11-17
reload: reloading shakes out runahead tasks


oliver-sanders:
There's a nasty interaction between reload and runahead.

```ini
[scheduler]                    
    allow implicit tasks = True    
                               
[scheduling]                   
    initial cycle point = 2000
    runahead limit = P5D       
    [[graph]]                  
        P1D = """              
            a => b             
        """                    
                               
[runtime]                      
    [[root]]                   
        script = sleep 500 
```

When this workflow starts, the tasks 2000 through 2005 are submitted:

```
2023-11-17T10:05:39Z INFO - [2000/a submitted job:01 flows:1] => running
2023-11-17T10:05:39Z INFO - [2000/a running job:01 flows:1] health: execution timeout=None, polling intervals=PT1H,...
2023-11-17T10:05:39Z INFO - [2002/a submitted job:01 flows:1] => running
2023-11-17T10:05:39Z INFO - [2002/a running job:01 flows:1] health: execution timeout=None, polling intervals=PT1H,...
2023-11-17T10:05:39Z INFO - [2003/a submitted job:01 flows:1] => running
2023-11-17T10:05:39Z INFO - [2003/a running job:01 flows:1] health: execution timeout=None, polling intervals=PT1H,...
2023-11-17T10:05:39Z INFO - [2004/a submitted job:01 flows:1] => running
2023-11-17T10:05:39Z INFO - [2004/a running job:01 flows:1] health: execution timeout=None, polling intervals=PT1H,...
2023-11-17T10:05:39Z INFO - [2005/a submitted job:01 flows:1] => running
2023-11-17T10:05:39Z INFO - [2005/a running job:01 flows:1] health: execution timeout=None, polling intervals=PT1H,...
```

But if I reload it, the task 2006 submits:

```
2023-11-17T10:05:40Z INFO - Reloading task definitions.
2023-11-17T10:05:40Z INFO - [2000/a running job:01 flows:1] reloaded task definition
2023-11-17T10:05:40Z WARNING - [2000/a running job:01 flows:1] active with pre-reload settings
2023-11-17T10:05:40Z INFO - [2001/a running job:01 flows:1] reloaded task definition
2023-11-17T10:05:40Z WARNING - [2001/a running job:01 flows:1] active with pre-reload settings
2023-11-17T10:05:40Z INFO - [2002/a running job:01 flows:1] reloaded task definition
[...]

PR #5826, opened 2023-11-17
tests/i: add reflog fixture

oliver-sanders:
Add a reflog fixture to the integration tests for trigger testing.

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #5833, opened 2023-11-21
task_pool: fix compute_runahead calculation

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/5825
* Cycles were considered active if they contained runahead limited tasks.
* This could cause the runahead limit to be bumped forwards whenever the limit calculation was forced to update, e.g. on reload.
* This filters out tasks at or beyond the runahead limit and straigntens out the task status checks to match Cylc 7 behaviour in compat mode.

**Check List**

### DIFF
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -51,13 +51,13 @@ from cylc.flow.task_proxy import TaskProxy
 from cylc.flow.task_state import (
     TASK_STATUSES_ACTIVE,
     TASK_STATUSES_FINAL,
-    TASK_STATUS_WAITING,
     TASK_STATUS_EXPIRED,
-    TASK_STATUS_PREPARING,
-    TASK_STATUS_SUBMITTED,
-    TASK_STATUS_RUNNING,
-    TASK_STATUS_SUCCEEDED,
     TASK_STATUS_FAILED,
+    TASK_STATUS_PREPARING,
+    TASK_STATUS_RUNNING,
+    TASK_STATUS_SUBMITTED,
+    TASK_STATUS_SUCCEEDED,
+    TASK_STATUS_WAITING,
     TASK_OUTPUT_EXPIRED,
     TASK_OUTPUT_FAILED,
     TASK_OUTPUT_SUCCEEDED,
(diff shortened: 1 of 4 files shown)


cylc/flow/task_pool.py
```diff
@@ -362,20 +362,35 @@ def compute_runahead(self, force=False) -> bool:
             # Find the earliest point with unfinished tasks.
             for point, itasks in sorted(self.get_tasks_by_point().items()):
                 if (
-                    points  # got the limit already so this point too
```
oliver-sanders, line 365: This could never be true as points is populated on a different branch.

cylc/flow/task_pool.py
```diff
@@ -362,20 +362,35 @@ def compute_runahead(self, force=False) -> bool:
...
-                    or any(
-                        not itask.state(
-                            TASK_STATUS_FAILED,
-                            TASK_STATUS_SUCCEEDED,
-                            TASK_STATUS_EXPIRED
-                        )
```
oliver-sanders, line 371: This counted waiting runahead tasks.

It should be superfluous if `not cylc.flow.flags.cylc7_back_compat` because `is_incomplete` should cover it anyway.

tests/integration/README.md
```diff
@@ -64,3 +64,21 @@ These methods both shut down the workflow / clean up after themselves.
...
+
+You'll find a bunch of module-scoped fixtues prefixed with `mod_`, e.g.
+`mod_start` is the module-scoped version of `start`. When using module-scoped
+fixtures, ensure that tests do not modify the fixture object as this will enable
+tests to interact.
+
+In order to get speedup from module-scoped fixtures when running with
+pytest-xdist, we configure pytest-xdist to run all of the tests in a module in
+series using the same pytest runner. This incentivises breaking up larger test
+modules.
```
wxtim, line 84: Possible alternative
```suggestion
series using the same pytest runner. Consider breaking up
large test modules to allow tests to be run in parallell.
```

wxtim (approved):

MetRonnie:
Got integration test failures

oliver-sanders:
> Got integration test failures

See the OP, caused by the same issue fixed in https://github.com/cylc/cylc-flow/pull/5826

MetRonnie:
Is that the correct PR? It doesn't mention closing an issue, and also is against master rather than 8.2.x

cylc/flow/task_pool.py
hjoliver: No, it's also populated on this branch too, within the same loop. So the logic says to automatically append the next points once we have added the first one based on the state checks. (However, I'm not sure why it wasn't just stopping at the first one, since then we just take the min anyway).

cylc/flow/task_pool.py
```diff
@@ -362,20 +362,35 @@ def compute_runahead(self, force=False) -> bool:
...
-                            itask.state.outputs.is_incomplete()
-                            and not cylc.flow.flags.cylc7_back_compat
+                    any(
+                        # filter out runahead tasks
+                        itask.state(is_runahead=False)
+                        and (
+                            # waiting tasks
+                            # * tasks with partially satisfied prereqs
+                            # * tasks with incomplete xtriggers
+                            # * held tasks
```
hjoliver, line 372:
```suggestion
                            # * held tasks
                            # * queued tasks
```

hjoliver (commented):
I think the new logic is correct, but can we just bail out of the loop once the first point is found? (given the loop is over sorted points, and we only want the min point in the end).

cylc/flow/task_pool.py
oliver-sanders: Since we're iterating in sorted order the earlier cycles should come first, so a `break` should make sense?

oliver-sanders: It's using points later in ways which don't cause test failures, so I'll restore the original to avoid breakages.

pytest.ini
```diff
@@ -34,11 +34,12 @@ testpaths =
     tests/integration/
 env =
     # a weird timezone to check that tests aren't assuming the local timezone
-    TZ=XXX-09:35
+    # Note: this doesn't work properly with pytest-xdist
+    # TZ=XXX-09:35
```
oliver-sanders, line 38: This prevents the new test from failing.

Setting the TZ variable in this way doesn't work as intended.

A solution will arrive in https://github.com/cylc/cylc-flow/pull/5826

MetRonnie:
> Yes, it's the right PR, the issue bring fixed was the way the time zone variable was set in the tests. Will have to pull the fix into 8.2.x.

Why not do it in this PR?

PR #5840, opened 2023-11-22
Clean up built-in xtrigger docstrings

MetRonnie:
Accompanies but does not depend on https://github.com/cylc/cylc-doc/pull/673

**Check List**

- [x] No dependency changes 
- [x] No tests needed
- [x] No changelog needed
- [x] Docs

### DIFF
--- a/cylc/flow/xtriggers/xrandom.py
+++ b/cylc/flow/xtriggers/xrandom.py
@@ -23,9 +23,10 @@ SIZES = ["tiny", "small", "medium", "large", "huge", "humongous"]
 
 
 def xrandom(percent, secs=0, _=None):
+    # NOTE: docstring must be valid rst as this is included in the docs
     """Random xtrigger, with configurable sleep and percent success.
 
-    Sleep for ``sec`` seconds, and report satisfied with ~``percent``
+    Sleep for ``sec`` seconds, and report satisfied with ~ ``percent``
     likelihood.
 
     The ``_`` argument is not used in the function code, but can be used to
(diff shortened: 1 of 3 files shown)


cylc/flow/xtriggers/xrandom.py
```diff
@@ -43,28 +44,37 @@ def xrandom(percent, secs=0, _=None):
     Examples:
         If the percent is zero, it returns that the trigger condition was
         not satisfied, and an empty dictionary.
-        >>> xrandom(0, 0)
-        (False, {})
+
+        .. code-block:: python
```
oliver-sanders, line 48: We shouldn't have to do this.

The Napoleon extension should automatically interpret the examples section as a Python code-block.

oliver-sanders: But we aren't using Napoleon in cylc-doc, only Rose.

Suggest adding this into the `extension` list in cylc-doc:

```
'sphinx.ext.napoleon'
```

It's a built-in plugin, but it's not enabled by default, it understands the stricture of Google style docstring (what we use) and formats them more nicely.

oliver-sanders: I think `py:exc reference target not found` means that it is trying to cross-reference the exception, but can't because the exception hasn't been documented.

Normally this wouldn't matter, but we turn on nit-picky mode so that referencing errors cause failures.

One option would be to auto-document exceptions in the reference section. I guess this could have some value.

oliver-sanders: [unrelated] Skimming the Rose config I spotted we're using `sphinx.ext.viewcode` there which is kinda cool.

MetRonnie:
Superseded by #5955

PR #5843, opened 2023-11-23
Fix.bad toml example in lint docstring

wxtim:
Very small change to make Cylc Doc build correctly.

### DIFF
--- a/cylc/flow/scripts/lint.py
+++ b/cylc/flow/scripts/lint.py
@@ -40,7 +40,7 @@ This can be overridden by providing the "--exit-zero" flag.
 TOMLDOC = """
 pyproject.toml configuration:{}
    [cylc-lint]                     # any of {}
-       ignore = ['S001', 'S002]    # List of rules to ignore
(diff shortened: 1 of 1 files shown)


cylc/flow/scripts/lint.py
MetRonnie, line 1: On master you removed `TOMLDOC` from being appended to the docstring

https://github.com/cylc/cylc-flow/blob/31150c902e2ddc45b3094a823c1177f55ca7ccd6/cylc/flow/scripts/lint.py#L1275-L1277

This makes sense to me as it seems a bit out of place appearing in the docs like this:

image

Note that it still appears in the `--help` text as that is done separately

wxtim: Did we have this discussion before? I'm not sure that I really wanted to do that. Or did we agree to put the Toml reference somewhere else. Perhaps it could do with a new page somewhere?

MetRonnie: There is already a section on `pyproject.toml` at the bottom

image

MetRonnie (approved):

PR #5848, opened 2023-11-24
Improve changelog entries

MetRonnie:


changes.d/5841.fix.md
```diff
@@ -1 +1 @@
-Improve handling of S011 to not warn if the # is '#$' (e.g. shell base arithmetic)
\ No newline at end of file
+`cylc lint`: improved handling of S011 to not warn if the `#` is `#$` (e.g. shell base arithmetic).
```
hjoliver, line 1: Re the grammatical tweak - I've had the impression forever that it was conventional to use the shortened form in change logs, as in "Fix blah", rather than "Fixed blah".  I kind of like that, if indeed it is a commonly used convention (@cylc/core - anyone else know?).  Arguably it's a grammatical error, but change logs aren't exactly normal prose text, and it has a certain snappy air to it. 

Anyhow, not a biggie, but we should be consistent - let's make a team decision either way, then stick to it.

oliver-sanders: Don't mind.

MetRonnie: I follow the convention of present tense for commits and past tense for changelog (or sometimes "X can now do Y"). Most of the time when I see changelog entries written in the present tense, it's because they use an automated system for pulling out the commit messages

wxtim:
Would it be worth adding some more categories to town crier?

changes.d/5841.fix.md
hjoliver: OK I'm fine with that.

hjoliver (approved):

PR #5849, opened 2023-11-28
tests/i: fix flaky tui/test_updater tests

oliver-sanders:
* These tests were trying to be a bit too clever, testing the updater whilst it was running.
* This wasn't really needed for what the tests were tring to assert, there are other tests which run Tui for real (live updater and all), those tests didn't exist when these ones were written.
* This commit freezes the updater so we can run it in lock-step with the tests.

Tests tests passed on the PR (https://github.com/cylc/cylc-flow/pull/5731), but failed when the PR was merged into master, three weeks later. I'm not sure what changed between these points, but they appear to be stable now.

Try me: `while true; do if ! pytest tests/integration/tui/test_updater.py::test_filters -s; then break; fi; done`

**Check List**

### DIFF
--- a/cylc/flow/tui/updater.py
+++ b/cylc/flow/tui/updater.py
@@ -155,7 +155,11 @@ class Updater():
         """
         with suppress_logging():
             self._update_filters(filters)
-            await self._update()
+            while True:
+                ret = await self._update()
+                if ret == self.SIGNAL_TERMINATE:
+                    break
+                self.update_queue.put(ret)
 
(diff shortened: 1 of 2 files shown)


oliver-sanders:
Well, it go rid of the test_updater failures, will follow up on the others later...

MetRonnie (approved):
Can we disable or xfail the other flaky tests for now?

wxtim (approved):

PR #5860, opened 2023-12-04
Test.example replace a reftest

wxtim:
Moved a functional test to the integration battery as part of review of https://github.com/cylc/cylc-flow/pull/5826 . Takes about 2/3 of the time after (~20s from ~30s).

**Check List**


tests/integration/triggering.py
MetRonnie, line 1: This file does not get run, because it doesn't start with `test_`

tests/integration/triggering.py
```diff
@@ -14,9 +13,30 @@
...
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#-------------------------------------------------------------------------------
-# Test fail triggering
-. "$(dirname "$0")/test_header"
-set_test_number 2
-reftest
-exit
+
+async def test_fail(flow, scheduler, run, reflog, complete, validate):
```
MetRonnie, line 17:
```suggestion
async def test_fail(flow, scheduler, run, reflog, complete, validate):
    """Test triggering on :fail"""
```

tests/integration/triggering.py
```diff
@@ -14,9 +13,30 @@
...
+    id_ = flow({
+        'scheduler': {
+            'allow implicit tasks': 'true'
+        },
+        'scheduling': {
+            'graph': {
+                'R1': 'foo:failed => bar'
+            }
+        },
+        'runtime': {
```
MetRonnie, line 27:
```suggestion
        'runtime': {
            'root': {
                'simulation': {'default run length': 'PT0S'}
            },
```
Makes this test run in 2-3s! It might be worth coming up with a fixture that can automatically merge this into the flow config dict?

tests/integration/triggering.py
wxtim: Good point!

tests/integration/triggering.py
MetRonnie: Although https://github.com/cylc/cylc-flow/blob/ec76e2e1a7069f52a5fcc05eef3bb3f4641224e0/tests/integration/test_examples.py#L234

runs in under 3s too without any special config... I'm not sure how that works

wxtim: https://cylc.github.io/cylc-doc/stable/html/reference/config/workflow.html#flow.cylc[runtime][%3Cnamespace%3E][simulation]?

wxtim: Another look suggests that it's the implicit tasks - I tried taking out all the task definition and got the speedup, although of course the task didn't fail when I wanted it to for the test.

MetRonnie: Ah I've got it, the integration tests already run in simulation mode with a default run length of `PT0S`. But it doesn't intelligently merge the config dict so this is lost when you specify the runtime section

MetRonnie: https://github.com/wxtim/cylc/pull/58

tests/integration/test_triggering.py
```diff
@@ -0,0 +1,46 @@
...
+        },
+        'scheduling': {
+            'graph': {
+                'R1': 'foo:failed => bar'
+            }
+        },
+        'runtime': {
+            'root': {
+                'simulation': {'default run length': 'PT0S'}
+            },
```
oliver-sanders, line 31: Note, this is the default for integration tests so does not need to be specified.

tests/integration/conftest.py
```diff
@@ -512,6 +512,7 @@ def reflog():
     """
 
     def _reflog(schd, flow_nums=False):
+
```
oliver-sanders, line 515:
```suggestion
```

tests/integration/test_triggering.py
oliver-sanders, line 31:
```suggestion
```

Note, this is the default for integration tests so does not need to be specified.

oliver-sanders (approved):

tests/integration/test_triggering.py
MetRonnie: See https://github.com/wxtim/cylc/pull/58

tests/integration/triggering.py
wxtim: Thanks!

MetRonnie (approved):

Issue #5835, opened 2023-11-21
config: auto upgrade "suite_state" to "workflow_state" in compat mode


oliver-sanders:
The `suite_state` xtrigger was renamed to `workflow_state`, this breaks Cylc 7-8 interoperability.

Ideally, we would back-support the old name in compat mode.

Suggestions:
* Hack the config.
* Add an xtrigger called suite_state which imports workflow_state and logs a deprecation message.

PR #5452, opened 2023-04-05
Xtrigger function arg validation.

hjoliver:
Close #5448 

~~Targeting master not 8.1.x because this conflicts with #5291~~

This is not just a bug-fix for the `wall_clock` xtrigger - I've gone for a generic approach:
 - if an xtrigger module has a function called `validate_config` it will be called at validation time, with the parsed args.

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [ ] Tests are included (or explain why tests are not needed).
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [ ] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #5864, opened 2023-12-05
added suite state xtrigger for back-compat

markgrahamdawson:
closes https://github.com/cylc/cylc-flow/issues/5835

## The issue
The suite_state xtrigger was renamed to workflow_state, this breaks Cylc 7-8 interoperability.

## The solution
I have just added a new xtrigger called suite_state which imports from workflow_state

**Check List**

### DIFF
--- a/cylc/flow/xtriggers/suite_state.py
+++ b/cylc/flow/xtriggers/suite_state.py
@@ -27,10 +27,49 @@ if not cylc.flow.flags.cylc7_back_compat:
 
 def suite_state(suite, task, point, offset=None, status='succeeded',
                 message=None, cylc_run_dir=None, debug=False):
-    '''The suite_state xtrigger was renamed to workflow_state,
-    this breaks Cylc 7-8 interoperability.
-    This suite_state xtrigger replicates
-    workflow_state - ensuring back-support'''
+    """Suite state xtrigger, required for interoperability with Cylc 7.
+
+    * The suite_state xtrigger was renamed to workflow_state, this breaks Cylc 7-8 interoperability.
+    * This suite_state xtrigger replicates workflow_state - ensuring back-support.
+
+    Arguments:
+        suite:
+            The workflow to interrogate.
+        task:
+            The name of the task to query.
+        point:
+            The cycle point.
+        offset:
+            The offset between the cycle this xtrigger is used in and the one
+            it is querying for as an ISO8601 time duration.
+            e.g. PT1H (one hour).
+        status:
+            The task status required for this xtrigger to be satisfied.
+        message:
+            The custom task output required for this xtrigger to be satisfied.
+            .. note::
+
+               This cannot be specified in conjunction with ``status``.
+
+        cylc_run_dir:
+            The directory in which the workflow to interrogate.
+
+            .. note::
+
+               This only needs to be supplied if the workflow is running in a
+               different location to what is specified in the global
+               configuration (usually ``~/cylc-run``).
+
+    Returns:
+        tuple: (satisfied, results)
+
+        satisfied:
+            True if ``satisfied`` else ``False``.
+        results:
+            Dictionary containing the args / kwargs which were provided
+            to this xtrigger.
+
+    """
     return workflow_state(
         workflow=suite,
         task=task,
--- a/cylc/flow/xtriggers/suite_state.py
+++ b/cylc/flow/xtriggers/suite_state.py
@@ -1,0 +1,42 @@
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from cylc.flow import LOG
+import cylc.flow.flags
+from cylc.flow.xtriggers.workflow_state import workflow_state
+
+if not cylc.flow.flags.cylc7_back_compat:
+    LOG.warning(
+        "The suite_state xtrigger is deprecated. "
+        "Please use the workflow_state xtrigger instead."
+    )
+
+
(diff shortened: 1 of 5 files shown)


tests/unit/xtriggers/test_workflow_state.py
```diff
@@ -80,7 +81,14 @@ def test_back_compat(tmp_run_dir):
...
+    # Test workflow_state function
     satisfied, _ = workflow_state(id_, task='mithril', point='2012')
     assert satisfied
     satisfied, _ = workflow_state(id_, task='arkenstone', point='2012')
     assert not satisfied
+
+    # Test back-compat (old suite_state function)
+    satisfied, _ = suite_state(id_, task='mithril', point='2012')
+    assert satisfied
+    satisfied, _ = suite_state(id_, task='arkenstone', point='2012')
```
MetRonnie, line 93: Worth testing the `suite` kwarg explicitly
```suggestion
    satisfied, _ = suite_state(suite=id_, task='mithril', point='2012')
    assert satisfied
    satisfied, _ = suite_state(suite=id_, task='arkenstone', point='2012')
```

wxtim (commented):
Seems fine. I've made a comment about explicitly testing the warning, but I'm not super fussed if you want to argue that it's unnecessary.

tests/unit/xtriggers/test_workflow_state.py
```diff
@@ -80,7 +81,14 @@ def test_back_compat(tmp_run_dir):
     finally:
         conn.close()
 
+    # Test workflow_state function
     satisfied, _ = workflow_state(id_, task='mithril', point='2012')
     assert satisfied
     satisfied, _ = workflow_state(id_, task='arkenstone', point='2012')
     assert not satisfied
+
```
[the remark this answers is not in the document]
MetRonnie: Might be better as a functional test

tests/unit/xtriggers/test_workflow_state.py
```diff
@@ -87,6 +90,19 @@ def test_back_compat(tmp_run_dir):
...
+    assert (
+        'The suite_state xtrigger is deprecated.'
+        ' Please use the workflow_state xtrigger instead.' in
+        caplog.messages
+    )
+    caplog.clear()
+
+    monkeypatch.setattr('cylc.flow.flags.cylc7_back_compat', True)
+    importlib.reload(cylc.flow.xtriggers.suite_state)
+    assert caplog.messages == []
```
MetRonnie, line 104: I don't think this part of the test makes sense. @wxtim What was the rationale behind this?
```suggestion
```

wxtim: Quite right @MetRonnie  - this is the wrong way around: But we should just as much test that we don't get an error if the flag is False as do get the error if flag is true.

MetRonnie: Monkeypatch definitely works (there are other tests that patch the `cylc7_back_compat` flag). But testing this is a bit icky in pytest due to the warning happening on import. I suggest a functional test would be more appropriate (would have a quick runtime as it would just call validate a couple of times)

wxtim: OK, that's more reasonable. Although we are still starting extra Python instances.

MetRonnie: https://github.com/markgrahamdawson/cylc-flow/pull/4

MetRonnie (changes requested):
Tests are failing for a valid reason
> XtriggerConfigError: [upstream] xtrigger module 'suite_state' not found

This is because of the change introduced in #5831 where built-in xtriggers are now registered by the `cylc.xtriggers` entry point in `setup.cfg`

https://github.com/cylc/cylc-flow/blob/22f41e250e54a92aea375e6692035002e9c7e7ec/setup.cfg#L218-L223

I think the best solution is to just add `suite_state` to the entry point? cc @oliver-sanders

oliver-sanders:
> I think the best solution is to just add suite_state to the entry point

Yes, that will be needed to register this xtrigger.

MetRonnie (approved):
👍

cylc/flow/xtriggers/suite_state.py
```diff
@@ -0,0 +1,42 @@
...
+        "Please use the workflow_state xtrigger instead."
+    )
+
+
+def suite_state(suite, task, point, offset=None, status='succeeded',
+                message=None, cylc_run_dir=None, debug=False):
+    '''The suite_state xtrigger was renamed to workflow_state,
+    this breaks Cylc 7-8 interoperability.
+    This suite_state xtrigger replicates
+    workflow_state - ensuring back-support'''
```
hjoliver, line 33: Not a biggie in this context, but this doc string doesn't conform to our usual conventions: one summary line, then a blank line before the details. And typically double quotes too.

setup.cfg
```diff
@@ -220,6 +220,7 @@ cylc.xtriggers =
     echo = cylc.flow.xtriggers.echo:echo
     wall_clock = cylc.flow.xtriggers.wall_clock:wall_clock
     workflow_state = cylc.flow.xtriggers.workflow_state:workflow_state
+    suite_state = cylc.flow.xtriggers.suite_state:suite_state
```
hjoliver, line 223: Note small conflict with #5452

tests/functional/xtriggers/04-suite_state.t
```diff
@@ -0,0 +1,50 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Test that deprecation warnings are printed appropriately for the suite_state
+# xtrigger.
```
hjoliver, line 19: Does this functional test provide anything more than the unit tests below?

setup.cfg
MetRonnie: (No need to do anything unless #5452 gets merged first)

tests/functional/xtriggers/04-suite_state.t
MetRonnie: It's icky to test the deprecation message logging from within unit tests because it happens at import time, so did this as a functional test. https://github.com/cylc/cylc-flow/pull/5864#discussion_r1424290455

MetRonnie:
Just failing flake8:
```
./cylc/flow/xtriggers/suite_state.py:32:80: E501 line too long (100 > 79 characters)
./cylc/flow/xtriggers/suite_state.py:33:80: E501 line too long (82 > 79 characters)
```

wxtim (approved):
That's fine - Sorry I failed to review this in a timely fashion.

PR #5868, opened 2023-12-06
post_configure: ensure asyncio tasks created by plugins are awaited

oliver-sanders:
This is one of a triplet of PRs which closes https://github.com/cylc/cylc-rose/issues/274
* https://github.com/cylc/cylc-rose/pull/276
* https://github.com/metomi/rose/pull/2744

---

* Await any background tasks started by a plugin before continuing. See https://github.com/cylc/cylc-rose/issues/274
* Centralise the plugin loading/running/reporting logic.
* Fix the installation test for back-compat-mode.

---

#### Reviewers, please check:

* It does what it says on the tin.
* (re)Install logging works as expected.
* Errors in file installation are reported as expected.

---

#### Testing:

This diff will make file installation take longer:

```diff
diff --git a/metomi/rose/config_processors/fileinstall.py b/metomi/rose/config_processors/fileinstall.py
index 937ed746..9797c042 100644
--- a/metomi/rose/config_processors/fileinstall.py
+++ b/metomi/rose/config_processors/fileinstall.py
@@ -399,6 +399,8 @@ class ConfigProcessorForFile(ConfigProcessorBase):
 
     async def process_job(self, job, conf_tree, loc_dao, work_dir):
         """Process a job, helper for "process"."""
+        import asyncio
+        await asyncio.sleep(1)
         for key, method in [
             (Loc.A_INSTALL, self._target_install),
             (Loc.A_SOURCE, self._source_pull),
```

This diff will disable the new guard:

```diff
diff --git a/cylc/flow/async_util.py b/cylc/flow/async_util.py
index 478c41dc29..14ddc26a80 100644
--- a/cylc/flow/async_util.py
+++ b/cylc/flow/async_util.py
@@ -519,8 +519,8 @@ async def async_block():
         # any tasks two() started will have been awaited by now
     """
     # make a list of all tasks running before we enter the context manager
-    tasks_before = asyncio.all_tasks()
+    # tasks_before = asyncio.all_tasks()
     # run the user code
     yield
     # await any new tasks
-    await asyncio.gather(*(asyncio.all_tasks() - tasks_before))
+    # await asyncio.gather(*(asyncio.all_tasks() - tasks_before))
```

Try installing a workflow with `cylc install --debug`, it will now log the time taken by each plugin.

With the `async_block` it should take >1s (due to the sleep) without the `async_block` (i.e. with the above diff applied) it should take <1s because the sleep will not be awaited.

---

#### CI:

Note that the cylc-rose and metomi-rose tests will fail until both the cylc-flow and metomi-rose branches are merged.

I.E. please run the tests manually due review.

**Check List**

- [x] `CHANGES.md` no - unreleased issue

### DIFF
--- a/cylc/flow/async_util.py
+++ b/cylc/flow/async_util.py
@@ -16,6 +16,7 @@
 """Utilities for use with asynchronous code."""
 
 import asyncio
+from contextlib import asynccontextmanager
 from functools import partial, wraps
 from inspect import signature
 import os
@@ -478,3 +479,48 @@ def make_async(fcn):
 
 
 async_listdir = make_async(os.listdir)
+
+
+@asynccontextmanager
+async def async_block():
+    """Ensure all tasks started within the context are awaited when it closes.
+
+    Normally, you would await a task e.g:
+
+    await three()
+
+    If it's possible to await the task, do that, however, this isn't always an
+    option. This interface exists is to help patch over issues where async code
+    (one) calls sync code (two) which calls async code (three) e.g:
+
+    async def one():
+        two()
+
+    def two():
+        # this breaks - event loop is already running
+        asyncio.get_event_loop().run_until_complete(three())
+
+    async def three():
+        await asyncio.sleep(1)
+
+    This code will error because you can't nest asyncio (without nest-asyncio)
+    which means you can schedule tasks the tasks in "two", but you can't await
+    them.
+
+    def two():
+        # this works, but it doesn't wait for three() to complete
+        asyncio.create_task(three())
+
+    This interface allows you to await the tasks
+
+    async def one()
+        async with async_block():
+            two()
+        # any tasks two() started will have been awaited by now
+    """
+    # make a list of all tasks running before we enter the context manager
+    tasks_before = asyncio.all_tasks()
+    # run the user code
+    yield
+    # await any new tasks
+    await asyncio.gather(*(asyncio.all_tasks() - tasks_before))
--- a/cylc/flow/plugins.py
+++ b/cylc/flow/plugins.py
@@ -1,0 +1,154 @@
+#!/usr/bin/env python3
+
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Common functionality related to the loading and calling of plugins."""
+
+import os
+from time import time
+
+from cylc.flow import LOG, iter_entry_points
+from cylc.flow.async_util import async_block as _async_block
+from cylc.flow.exceptions import PluginError
+import cylc.flow.flags
+
+
+async def run_plugins_async(
+    plugin_namespace,
+    *args,
+    async_block=False,
+    **kwargs
+):
+    """Run all installed plugins for the given namespace.
+
+    This runs plugins in series, yielding the results one by one.
+
+    Args:
+        plugin_namespace:
+            The entry point namespace for the plugins to run,
+            e.g. "cylc.post_install".
+        args:
+            Any arguments to call plugins with.
+        async_block:
+            If True, this will wait for any async tasks started by the plugin
+            to complete before moving on to the next plugin.
+        kwargs:
+            Any kwargs to call plugins with.
+
+    Yields:
+        (entry_point, plugin_result)
+
+        See https://github.com/cylc/cylc-rose/issues/274
+
+    """
+    startpoint = os.getcwd()
+    for entry_point in iter_entry_points(plugin_namespace):
+        try:
+            # measure the import+run time for the plugin (debug mode)
+            start_time = time()
+
+            # load the plugin
+            meth = entry_point.load()
+
+            # run the plugin
+            if async_block:
+                # wait for any async tasks started by the plugin to complete
+                async with _async_block():
+                    plugin_result = meth(*args, **kwargs)
+            else:
+                plugin_result = meth(*args, **kwargs)
+
+            # log the import+run time (debug mode)
+            if cylc.flow.flags.verbosity > 1:
+                LOG.debug(
+                    f'ran {entry_point.name} in {time() - start_time:0.05f}s'
+                )
+
+            # yield the result to the caller
+            yield entry_point, plugin_result
+
+        except Exception as exc:  # NOTE: except Exception (purposefully vague)
+            _raise_plugin_exception(exc, plugin_namespace, entry_point)
+
+        finally:
+            # ensure the plugin does not change the CWD
+            os.chdir(startpoint)
+
+
+def run_plugins(plugin_namespace, *args, **kwargs):
+    """Run all installed plugins for the given namespace.
+
+    This runs plugins in series, yielding the results one by one.
+
+    Warning:
+        Use run_plugins_async for "cylc.post_install" plugins.
+        See https://github.com/cylc/cylc-rose/issues/274
+
+    Args:
+        plugin_namespace:
+            The entry point namespace for the plugins to run,
+            e.g. "cylc.post_install".
+        args:
+            Any arguments to call plugins with.
+        kwargs:
+            Any kwargs to call plugins with.
+
+    Yields:
+        (entry_point, plugin_result)
+
+    """
+    startpoint = os.getcwd()
+    for entry_point in iter_entry_points(plugin_namespace):
+        try:
+            # measure the import+run time for the plugin (debug mode)
+            start_time = time()
+
+            # load the plugin
+            meth = entry_point.load()
+
+            # run the plugin
+            plugin_result = meth(*args, **kwargs)
+
+            # log the import+run time (debug mode)
+            if cylc.flow.flags.verbosity > 1:
+                LOG.debug(
+                    f'ran {entry_point.name} in {time() - start_time:0.05f}s'
+                )
+
+            # yield the result to the caller
+            yield entry_point, plugin_result
+
+        except Exception as exc:  # NOTE: except Exception (purposefully vague)
+            _raise_plugin_exception(exc, plugin_namespace, entry_point)
+
+        finally:
+            # ensure the plugin does not change the CWD
+            os.chdir(startpoint)
+
+
+def _raise_plugin_exception(exc, plugin_namespace, entry_point):
+    """Re-Raise an exception captured from a plugin."""
+    if cylc.flow.flags.verbosity > 1:
+        # raise the full exception in debug mode
+        # (this helps plugin developers locate the error in their code)
+        raise
+    # raise a user-friendly exception
+    raise PluginError(
+        plugin_namespace,
+        entry_point.name,
+        exc
+    ) from None
--- a/cylc/flow/scripts/install.py
+++ b/cylc/flow/scripts/install.py
@@ -86,14 +86,8 @@ from optparse import Values
 from pathlib import Path
 from typing import Any, Dict, Optional, Tuple
 
-from cylc.flow.scripts.scan import (
-    get_pipe,
-    _format_plain,
-    FLOW_STATE_SYMBOLS,
-    FLOW_STATE_CMAP
-)
-from cylc.flow import LOG, iter_entry_points
-from cylc.flow.exceptions import PluginError, InputError
+from cylc.flow import LOG
+from cylc.flow.exceptions import InputError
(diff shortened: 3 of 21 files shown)


cylc/flow/scripts/install.py
```diff
@@ -325,23 +318,14 @@ def install(
...
-            )
-        except Exception as exc:
-            # NOTE: except Exception (purposefully vague)
-            # this is to separate plugin from core Cylc errors
-            raise PluginError(
-                'cylc.post_install',
-                entry_point.name,
-                exc
-            ) from None
+    async with async_block():
```
oliver-sanders, line 321: This ensures that any async tasks started by the plugins are awaited before we move on.

Rose fileinstallation will create async tasks but cannot await them (because it's called synchronously) so we have to do it here, otherwise `cylc vip` would start running the workflow before local file installation has finished.

cylc/flow/scripts/reinstall.py
```diff
@@ -295,7 +299,14 @@ def reinstall(
     # run post_install plugins
     if not dry_run:
         # don't run plugins in dry-mode
-        post_install(opts, src_dir, run_dir)
+        async with async_block():
```
oliver-sanders, line 302: This ensures that any async tasks started by the plugins are awaited before we move on.

Rose fileinstallation will create async tasks but cannot await them (because it's called synchronously) so we have to do it here, otherwise `cylc vip` would start running the workflow before local file installation has finished.

tests/integration/test_install.py
```diff
@@ -136,55 +136,60 @@ def test_install_scan_ping(
...
+    def failIfDeprecated(*args, **kwargs):
         """A fake Cylc Plugin entry point"""
-        @staticmethod
-        def load():
-            return failIfDeprecated.raiser
-
-        @staticmethod
-        def raiser(*_, **__):
-            import cylc.flow.flags
-            if cylc.flow.flags.cylc7_back_compat:
```
oliver-sanders, line 171: This test was intended to check that the back-compat flag was set correctly, however, it achieved this by monkeypatching the back-compat flag (i.e. manually setting it to what it expected it to be).

cylc/flow/async_util.py
```diff
@@ -478,3 +479,48 @@ async def _fcn(*args, executor=None, **kwargs):
 
 
 async_listdir = make_async(os.listdir)
+
+
+@asynccontextmanager
+async def async_block():
```
oliver-sanders, line 485: Crude but effective way to allow us to call async -> sync -> async.

This makes a list of async tasks before the plugin is called in order to await any new tasks created by the plugin.

This means that plugins can go async like so:

```python
def plugin(opts):
    asyncio.create_task(run(opts))
    # the plugin will return now, but Cylc will await the "run" for us

async def run(opts):
    ...
```

I think it's a reasonable solution.

Sadly, directly converting the plugin interfaces to `async` isn't an easy option due to the use of plugins within parsec. Making the plugin async would make loading the global config async which would make global config value retrieval (e.g. the line below) async:

```python
glbl_cfg().get(['install', 'symlink dirs'])
```

Which would basically require the entire cylc-flow codebase to be async and prohibit config caching at the module level.

cylc/flow/plugins.py
```diff
@@ -0,0 +1,159 @@
...
+    """Yield all installed plugin entry points which match the given name."""
+    yield from iter_entry_points(plugin_namespace)
+
+
+async def run_plugins_async(
+    plugin_namespace,
+    *args,
+    async_block=False,
+    **kwargs
+):
```
oliver-sanders, line 40: This centralises the plugin running logic to avoid duplicating it in each place.

It now logs each plugin run along with the time taken to run it at the debug level.

The full plugin traceback (not just the cylc-flow side of it) will be raised when in debug mode.

cylc/flow/plugins.py
```diff
@@ -0,0 +1,159 @@
...
+            start_time = time()
+
+            # load the plugin
+            meth = entry_point.load()
+
+            # run the plugin
+            if async_block:
+                # wait for any async tasks started by the plugin to complete
+                async with _async_block():
+                    plugin_result = meth(*args, **kwargs)
```
oliver-sanders, line 76: This ensures that any async tasks started by the plugins are awaited before we move on.

Rose fileinstallation will create async tasks but cannot await them (because it's called synchronously) so we have to do it here, otherwise cylc vip would start running the workflow before local file installation has finished.

cylc/flow/async_util.py
```diff
@@ -478,3 +479,48 @@ async def _fcn(*args, executor=None, **kwargs):
...
+    async def three():
+        await asyncio.sleep(1)
+
+    This code will error because you can't nest asyncio (without nest-asyncio)
+    which means you can schedule tasks the tasks in "two", but you can't await
+    them.
+
+    def two():
+        # this works, but it doesn't wait for three() to complete
+        asyncio.create_task(three())
```
wxtim, line 512: According to the [docs ](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task)this usage may be risky.

wxtim (approved):

cylc/flow/async_util.py
oliver-sanders: What is the safety issue? I can't see anything.

Also, if this is unsafe, then the entire approach is unsafe, not just the docs!

cylc/flow/scripts/install.py
```diff
@@ -297,19 +296,12 @@ def install(
     # for compatibility mode:
     check_deprecation(source)
 
-    for entry_point in iter_entry_points(
-        'cylc.pre_configure'
+    async for _entry_point, _plugin_result in run_plugins_async(
```
wxtim, line 299: Why is this async, since pre-config doesn't need it?

oliver-sanders: Why *should* it be sync?

tests/integration/test_reinstall.py
```diff
@@ -299,3 +300,89 @@ async def test_rsync_fail(one_src, one_run, mock_glbl_cfg, non_interactive):
...
+
+        async def my_async():
+            # the async task
+            nonlocal progress
+            await asyncio.sleep(2)
+            progress.append('end')
+
+        # start the async task
+        progress.append('start')
+        asyncio.get_event_loop().create_task(my_async())
```
wxtim, line 326: According to the docs you should assign this to something: https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

wxtim (changes requested):

tests/integration/test_reinstall.py
```diff
@@ -299,3 +300,89 @@ async def test_rsync_fail(one_src, one_run, mock_glbl_cfg, non_interactive):
...
+    @asynccontextmanager
+    async def async_block():
+        yield
+
+    monkeypatch.setattr(
+        'cylc.flow.plugins._async_block',
+        async_block,
+    )
+
+    # this is would it would have done without async block
```
wxtim, line 382:
```suggestion
    # this is what it would have done without async block
```

cylc/flow/async_util.py
wxtim: https://github.com/python/cpython/issues/88831

tests/integration/test_reinstall.py
oliver-sanders: Hmm... This is just the test which proves the concept, not functional code, so this isn't an issue.

But it is what Rose is doing at the other end, so you might want to stick this comment here: https://github.com/metomi/rose/pull/2744/files#diff-8db25a5509044308d363a425af008c5ba353b7d04f1d78dc08c3a4c5a659afb9R188

This doesn't appear to be an issue, but we should ensure that it can't become one.

The tasks created by Rose *are* referenced because they get added to a list called `running`:

https://github.com/metomi/rose/blob/191ce20b8515e9624a7e9081071a8b6533da770e/metomi/rose/job_runner.py#L223-L227

However, the task which creates those tasks is not referenced:

https://github.com/metomi/rose/blob/191ce20b8515e9624a7e9081071a8b6533da770e/metomi/rose/job_runner.py#L188

Simplified call stack:

* cylc-flow (async)
  * cylc-rose (**sync**)
    * metomi-rose (**sync**)
      * _run (async) - **un**referenced
        * task (async) - referenced 

So I would need to find a way to ensure that task remains referenced. It would be possible to pass it back through the callchain which would remove the need for the `async_block`, but would get messy in other ways.

cylc/flow/plugins.py
wxtim: I really wish I'd done this, back when we were implementing all this.

tests/integration/test_reinstall.py
oliver-sanders: Addressed in the metomi-rose PR: https://github.com/metomi/rose/pull/2744/files#r1473137433

cylc/flow/async_util.py
oliver-sanders: The issue is metomi-rose side, this commit attempts to solve it: https://github.com/metomi/rose/pull/2744/files#r1473137433

wxtim (approved):

wxtim:
- [x] Read code
- [x] Tried out locally
- [x] Assured myself that test failures were unrelated to this change:
  - [x] Rose stem failure caused by a local change
  - [x] MyPy Failure (metomi/rose) caused by MyPy becoming annoying around Python 3.7.
  - [x] Test Tutorials.

tests/integration/test_reinstall.py
```diff
@@ -299,3 +300,89 @@ async def test_rsync_fail(one_src, one_run, mock_glbl_cfg, non_interactive):
...
+        async_block,
+    )
+
+    # this is what it would have done without async block
+    (one_src.path / 'b').touch()  # give it something else to install
+    my_install_plugin.clear()
+    assert my_install_plugin == []
+    await reinstall_cli(opts=ReInstallOptions(), workflow_id=one_run.id)
+    # the absence of "end" means that the task was not awaited
+    assert my_install_plugin == ['start', 'return']
```
MetRonnie, line 388: Not sure I understand why this bit is being tested?

cylc/flow/plugins.py
```diff
@@ -0,0 +1,159 @@
...
+
+    """
+    startpoint = os.getcwd()
+    for entry_point in get_entry_points(plugin_namespace):
+        try:
+            # measure the import+run time for the plugin (debug mode)
+            start_time = time()
+
+            # load the plugin
+            meth = entry_point.load()
```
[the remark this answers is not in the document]
oliver-sanders: Much duller I'm afraid, **meth**od.

tests/integration/test_reinstall.py
oliver-sanders: tldr; the second part of this test confirms that the first part is a valid test.

It's easy to write a test for this sort of thing that passes, even if the related functionality doesn't actually work.

To make the that this functionality *does* work, I wrote two tests:
1. One checks that there would be a problem if we didn't use the fix.
2. The second checks that the fix solves the problem.

By writing them together using the same approach you reduce the likelyhood of the test being broken and producing false negatives in the future.

cylc/flow/plugins.py
```diff
@@ -0,0 +1,159 @@
...
+
+from cylc.flow import LOG, iter_entry_points
+from cylc.flow.async_util import async_block as _async_block
+from cylc.flow.exceptions import PluginError
+import cylc.flow.flags
+
+
+def get_entry_points(plugin_namespace):
+    """Yield all installed plugin entry points which match the given name."""
+    yield from iter_entry_points(plugin_namespace)
```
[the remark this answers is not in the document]
oliver-sanders: I think I had a plan for it a some point but can't remember what it was, will address.

MetRonnie:
Unfortunately `functional tests / py-3-latest` and `test-tutorial-workflow / test (3)` hang at some point

oliver-sanders:
Lucky catch, this PR appears to have broken daemonization, which is a major weak spot in our testing (due to its slippery nature). I have no idea why just that one test fails, but you can easily replicate the issue by configuring `run hosts`.

The cause isn't obvious, even uninstalling `cylc-rose` and commenting out the `async_block` code doesn't change it.

wxtim (approved):
- [x] Read changes
- [x] Run tests locally
- [x] Tried to break various components of this

oliver-sanders:
I think I have a solution to the hanging problem, see [d173c2a](https://github.com/cylc/cylc-flow/pull/5868/commits/d173c2a1aaaae701275bdc18e70c1a11543a8747) for details.

cylc/flow/scripts/validate_install_play.py
```diff
@@ -92,20 +92,19 @@ def get_option_parser() -> COP:
...
 @cli_function(get_option_parser)
 def main(parser: COP, options: 'Values', workflow_id: Optional[str] = None):
-    asyncio.run(run(parser, options, workflow_id))
-
-
-async def run(parser: COP, options: 'Values', workflow_id: Optional[str]):
     """Run Cylc validate - install - play in sequence."""
+    # NOTE: We call each of the stages in its own event loop because there is a
+    # strange interaction whereby the cylc.flow.network.scan coroutine pipe can
+    # cause sys.exit to hang on the original process post-daemonization
```
MetRonnie, line 98: 👍

MetRonnie (commented):
I'll test this out tomorrow

wxtim:
TL;DR - Revoked approval pending me getting tests working locally. :(

I've checked out these branches again enroute to https://github.com/cylc/cylc-flow/issues/5968, and run the tests to ensure I was building on a set of branches where tests were passing, but got failures.

wxtim (approved):

MetRonnie (approved):
Tested

PR #5869, opened 2023-12-07
Fix misspelling of 'Rakiura'.

hjoliver:
I love the TUI test framework, but the name is spelled wrong! 

https://www.doc.govt.nz/parks-and-recreation/places-to-go/southland/places/stewart-island-rakiura/rakiura-national-park/?tab-id=50578

One review will do.


oliver-sanders (approved):
Dangit

PR #5872, opened 2023-12-08
`cylc clean` remote timeout improvements

### LINKED ISSUE #5867: `cylc clean` timeout traceback
> Some folks are hitting the remote timeout for `cylc clean`. The default is 120 seconds which is quite short. Suggest:
>- Increase the default timeout.
>- Better handing of timeout error (exit code 124), this currently results in traceback.

Also:
- Allow ISO 8601 duration format for the timeout option

MetRonnie:
Closes #5867 

- Increase default remote `--timeout` to 5 mins and allow ISO 8601 durations
- Provide a shorter and clearer error message for remote timeouts

**Check List**

- [x] No dependency changes 
- [x] Tests are included 
- [x] No docs needed

### DIFF
--- a/cylc/flow/clean.py
+++ b/cylc/flow/clean.py
@@ -44,9 +44,6 @@ from typing import (
     Union,
 )
 
-from metomi.isodatetime.exceptions import ISO8601SyntaxError
-from metomi.isodatetime.parsers import DurationParser
-
 from cylc.flow import LOG
 from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
 from cylc.flow.exceptions import (
@@ -362,9 +359,6 @@ def remote_clean(
         raise PlatformLookupError(
             f"Cannot clean {id_} on remote platforms as the workflow database "
             f"is out of date/inconsistent with the global config - {exc}")
-
-    with suppress(ISO8601SyntaxError):
-        timeout = str(DurationParser().parse(timeout).get_seconds())
 
     queue: Deque[RemoteCleanQueueTuple] = deque()
     remote_clean_cmd = partial(
--- a/cylc/flow/clean.py
+++ b/cylc/flow/clean.py
@@ -44,6 +44,9 @@ from typing import (
     Union,
 )
 
+from metomi.isodatetime.exceptions import ISO8601SyntaxError
+from metomi.isodatetime.parsers import DurationParser
+
 from cylc.flow import LOG
 from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
 from cylc.flow.exceptions import (
@@ -187,7 +190,7 @@ def init_clean(id_: str, opts: 'Values') -> None:
 
         if platform_names and platform_names != {'localhost'}:
             remote_clean(
-                id_, platform_names, opts.rm_dirs, opts.remote_timeout
+                id_, platform_names, opts.remote_timeout, opts.rm_dirs
             )
 
     if not opts.remote_only:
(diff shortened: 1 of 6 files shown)


cylc/flow/clean.py
```diff
@@ -358,6 +362,10 @@ def remote_clean(
         raise PlatformLookupError(
             f"Cannot clean {id_} on remote platforms as the workflow database "
             f"is out of date/inconsistent with the global config - {exc}")
+
+    with suppress(ISO8601SyntaxError):
+        timeout = str(DurationParser().parse(timeout).get_seconds())
```
oliver-sanders, line 367: Shouldn't this error be raised rather than suppressed?

MetRonnie: I did this to allow users to still use a raw number of seconds. We weren't doing any validation on that before, I suppose your suggestion is a good idea

wxtim: I'm not sure that it's reasonable to use two try excepts here because we can't say for sure which way of setting this value will become the popular one with users:

IMO

```python
if timeout.isdigit():
    timeout = int(timeout)
else:
    try:
        timeout = str(DurationParser().parse(timeout).get_seconds())
    except ISO8601SyntaxError:
        raise InputError(...)
    
```

wxtim: Jupyter notebook testing this.

```python
from metomi.isodatetime.parsers import DurationParser
from metomi.isodatetime.exceptions import ISO8601SyntaxError

def one(timeout):
    try: 
        timeout = str(DurationParser().parse(timeout).get_seconds())
    except ISO8601SyntaxError:
        try:
            timeout = int(timeout)
        except ValueError:
            raise InputError("Hi")


def two(timeout):
    if timeout.isdigit():
        timeout = int(timeout)
    else:
        try:
            timeout = str(DurationParser().parse(timeout).get_seconds())
        except ISO8601SyntaxError:
            raise InputError(...)
        
```


```python
from random import randint, choice
```


```python
%%timeit
one(choice([str(randint(0, 3600))]))
```

    3.53 µs ± 34.1 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)



```python
%%timeit
two(choice([str(randint(0, 3600))]))
```

    1.1 µs ± 13 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)



```python
%%timeit
one(choice([str(randint(0, 3600)), 'PT4S']))
```

    7.61 µs ± 85.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)



```python
%%timeit
two(choice([str(randint(0, 3600)), 'PT4S']))
```

    6.23 µs ± 143 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)



```python
%%timeit
with suppress(InputError):
    one(choice([str(randint(0, 3600)), 'PT4S', 'garbage']))
```

    7.76 µs ± 180 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)



```python
%%timeit
with suppress(InputError):
    two(choice([str(randint(0, 3600)), 'PT4S', 'garbage']))
```

    6.39 µs ± 174 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

wxtim (commented):
I'm fairly happy with this. Action comments and I'll approve.

cylc/flow/clean.py
MetRonnie: I've made the default `PT5M` so I suspect the ISO branch will be hit most of the time. Anyway, a few ms shouldn't matter too much as this will only be executed once per clean (I'll move this validation to `scripts/clean.py`)

oliver-sanders (approved):

wxtim (approved):

PR #5886, opened 2023-12-18
Tidy network code

MetRonnie:
**Check List**

- [x] No dependency changes 
- [x] No tests added
- [x] No changelog needed
- [x] No docs needed.

### DIFF
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -349,8 +349,6 @@ async def get_nodes_all(root, info, **args):
 
     _, field_ids = process_resolver_info(root, info, args)
 
-    if hasattr(args, 'id'):
-        args['ids'] = [args.get('id')]
     if field_ids:
         if isinstance(field_ids, str):
             field_ids = [field_ids]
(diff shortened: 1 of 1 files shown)


cylc/flow/network/schema.py
```diff
@@ -349,8 +349,6 @@ async def get_nodes_all(root, info, **args):
 
     _, field_ids = process_resolver_info(root, info, args)
 
-    if hasattr(args, 'id'):
-        args['ids'] = [args.get('id')]
```
MetRonnie, line 353: This logic is flawed because a dict will never have an attribute `id` (only a key, perhaps)

oliver-sanders: This line doesn't appear to have been hit in coverage so this may be correct: https://app.codecov.io/gh/cylc/cylc-flow/blob/master/cylc%2Fflow%2Fnetwork%2Fschema.py#L352

But this could just be a coverage hole. Dunno, @dwsutherland any ideas.

MetRonnie: It is physically impossible to hit - a dictionary like `foo = {'id': 1}` doesn't have `foo.id`, it has `foo['id']`

oliver-sanders: True if its always a dictionary. By the looks of it, the purpose of this code is to handle an object that isn't a dictionary but supports `__setitem__`.

MetRonnie: `args` is a dict because it's the keyword arguments object from the function definition

oliver-sanders: Can't fault that!

oliver-sanders (approved):

PR #5894, opened 2023-12-20
Move some imports into TYPE_CHECKING block.

hjoliver:
A flake8 update in my environment flagged these changes.

One review will do.

### DIFF
--- a/cylc/flow/exceptions.py
+++ b/cylc/flow/exceptions.py
@@ -26,10 +26,13 @@ from typing import (
     Tuple,
     Type,
     Union,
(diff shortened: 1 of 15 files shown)


oliver-sanders:
That's not a flake8 rule change. It's the result of a new rule set which was added to master (along with the lint fixes)

https://github.com/cylc/cylc-flow/pull/5770

hjoliver:
Duh 🤦 I nominate myself for idiot of the day.

In that case, the warnings suddenly appeared because I hadn't run local tests on 8.2.x for a while.

PR #5874, opened 2023-12-13
Fixed unhelpful upgrade message for `[scheduling][dependencies]graph`

wxtim:
… foo=>bar

Closes https://github.com/cylc/cylc-flow/issues/5810

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [x] ~`CHANGES.md` entry~ improves documentation, doesn't change functioning.
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #5901, opened 2024-01-02
Tests.move deprecations tests to integration battery

wxtim:
Testing only - a couple of easy wins I spotted moving functional tests to the integration battery. Will clash with https://github.com/cylc/cylc-flow/pull/5874


pytest.ini
```diff
@@ -39,3 +39,4 @@ doctest_optionflags =
 asyncio_mode = auto
 markers=
     linkcheck: Test links
+    compatmode: Tests a feature of Cylc 7 compatibility mode
```
oliver-sanders, line 42: I'm not sure how useful this marker is, especially as it's not applied to existing compatmode tests.

tests/integration/scripts/test_validate_integration.py
```diff
@@ -94,3 +94,64 @@ async def test_validate_against_source_gets_old_tvars(
         flow_file.read_text().replace('P1Y = foo', 'P1Y = {{FOO}}'))
     with pytest.raises(Jinja2Error):
         validate(src_dir)
+
+
+@pytest.mark.compatmode
```
oliver-sanders, line 99: I don't think this actually tests Cylc 7 compatibility mode?

tests/integration/scripts/test_validate_integration.py
```diff
@@ -94,3 +94,64 @@ async def test_validate_against_source_gets_old_tvars(
...
+    (
+        " * (7.8.0) [runtime][foo, cat, dog][suite state polling]template - DELETED (OBSOLETE)",
+        " * (7.8.1) [cylc][events]reset timer - DELETED (OBSOLETE)",
+        ' * (7.8.1) [cylc][events]reset inactivity timer - DELETED (OBSOLETE)',
+        ' * (7.8.1) [runtime][foo, cat, dog][events]reset timer - DELETED (OBSOLETE)',
+        ' * (8.0.0) [runtime][foo, cat, dog][suite state polling] -> [runtime][foo, cat, dog][workflow state polling] - value unchanged',
+        ' * (8.0.0) [cylc] -> [scheduler] - value unchanged'
+    )
+)
+@pytest.mark.compatmode
```
oliver-sanders, line 128: I don't think this actually tests Cylc 7 compatibility mode?

tests/integration/scripts/test_validate_integration.py
```diff
@@ -94,3 +94,64 @@ async def test_validate_against_source_gets_old_tvars(
...
+    expect = (
+        'deprecated graph items were automatically upgraded '
+        'in "workflow definition":'
+        '\n * (8.0.0) [scheduling][dependencies][X]graph'
+        ' -> [scheduling][graph]X - for X in:\n       graph'
+    )
+    assert expect in caplog.messages
+
+
+@pytest.mark.parametrize(
```
oliver-sanders, line 117: I don't think this is a valid use case for parametrization, the test only needs to be run once, this runs it once per message to check.

Use a for loop instead:

```python
python
validate(...)
for message in [
    ...
]:
    assert message in ...
```

Note, if a comparison fails, pytest will print the difference, so it's easy to tell which message the test fails on.

tests/integration/scripts/test_validate_integration.py
wxtim: It is!
```
[scheduling]
  [[dependencies]]
    graph = foo
```

tests/integration/scripts/test_validate_integration.py
wxtim: No, testing the upgrading.

oliver-sanders (dismissed):

tests/integration/utils/flow_tools.py
```diff
@@ -74,7 +82,10 @@ def _make_flow(
         .setdefault('default run length', 'PT0S')
     )
     # allow implicit tasks by default:
-    conf.setdefault('scheduler', {}).setdefault('allow implicit tasks', 'True')
+    if defaults:
```
oliver-sanders, line 85: Should add the other default set above into this block.

wxtim: I have changed it for completeness, I don't think that's necessary, because `[runtime][NAMESPACE][simulation]default run length is both unchanged from Cylc 7 and overrideable.

oliver-sanders: (It's not relevant to your use case, but will likely be relevant to other validate/config use cases)

oliver-sanders (approved):

PR #5902, opened 2024-01-03
Made execution time limit unsettable by reload or broadcast.

### LINKED ISSUE #5891: Clearing execution time limit does not work
### Description

Clearing the execution time limit either by editing `flow.cylc` and reloading, or by broadcast/edit-runtime, does not work. The original execution time limit is maintained.

### Reproducible Example

```ini
# flow.cylc
[scheduling]
    cycling mode = integer
    [[graph]]
        R1 = a
[runtime]
    [[a]]
        script = sleep 10
        execution time limit = PT1S
```

```console
$ cylc play workflow
```

Wait for task to fail.

```diff
     [[a]]
         script = sleep 10
-        execution time limit = PT1S
```

```console
$ cylc reload workflow
$ cylc trigger workflow//1/a
```

### Expected Behaviour

The task should succeed now after 10s, but it still fails after 1s

wxtim:
Closes #5891 

Formerly a suppress statement was supressing cases when we wanted to set itask.summary['execution time limit'] = None.

Added tests.

**Check List**

- [x] ~[Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.~ Not req'd

### DIFF
--- a/changes.d/5902.fix.md
+++ b/changes.d/5902.fix.md
@@ -1,2 +1,1 @@
-Fixed a case where reloading or broadcasting execution time limit=
-would have no effect.
+Fixed a bug that prevented unsetting `execution time limit` by broadcast or reload.
--- a/changes.d/5902.fix.md
+++ b/changes.d/5902.fix.md
@@ -1,0 +1,2 @@
+Fixed a case where reloading or broadcasting execution time limit=
+would have no effect.
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -36,7 +36,7 @@ from logging import (
 )
 from shutil import rmtree
 from time import time
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Any, Union, Optional
 
 from cylc.flow import LOG
 from cylc.flow.job_runner_mgr import JobPollContext
@@ -1262,10 +1262,11 @@ class TaskJobManager:
             itask.submit_num] = itask.platform['name']
 
         itask.summary['job_runner_name'] = itask.platform['job runner']
-        with suppress(TypeError):
-            itask.summary[self.KEY_EXECUTE_TIME_LIMIT] = float(
-                rtconfig['execution time limit']
-            )
+
+        # None is an allowed non-float number for Execution time limit.
+        itask.summary[
+            self.KEY_EXECUTE_TIME_LIMIT
+        ] = self.get_execution_time_limit(rtconfig['execution time limit'])
 
         # Location of job file, etc
         self._create_job_log_path(workflow, itask)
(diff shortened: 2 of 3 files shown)


cylc/flow/task_job_mgr.py
```diff
@@ -1281,6 +1282,35 @@ def _prep_submit_task_job_impl(self, workflow, itask, rtconfig):
...
+            This function does not save you from a value error:
+            >>> from pytest import raises
+            >>> with raises(ValueError):
+            ...     this("🇳🇿")
+        """
+        execution_time_limit = None
+        if config_execution_time_limit is not None:
+            with suppress(TypeError):
+                execution_time_limit = float(config_execution_time_limit)
+        return execution_time_limit
```
oliver-sanders, line 1312: I think this does the same?

```suggestion
        with suppress(TypeError):
            return float(config_execution_time_limit)
        return None
```

tests/integration/test_task_job_mgr.py
```diff
@@ -128,3 +129,56 @@ async def test__run_job_cmd_logs_platform_lookup_fail(
...
+        }
+    })
+
+    # Run in live mode - function not called in sim mode.
+    schd = scheduler(id_, run_mode='live')
+    async with start(schd):
+        task_a = schd.pool.get_tasks()[0]
+        # We're not interested in the job file stuff, just
+        # in the summary state.
+        with suppress(FileExistsError):
```
oliver-sanders, line 166: This means that the remainder of the test will be skipped if FileExistsError occurs anywhere in this block which could cause false positives.

oliver-sanders (approved):

changes.d/5902.fix.md
```diff
@@ -0,0 +1,2 @@
+Fixed a case where reloading or broadcasting execution time limit=
+would have no effect.
```
hjoliver, line 2: Could be clearer? 

```suggestion
Fixed a bug preventing update of `execution time limit` by broadcast or reload.
```

cylc/flow/task_job_mgr.py
```diff
@@ -1281,6 +1282,33 @@ def _prep_submit_task_job_impl(self, workflow, itask, rtconfig):
...
 
+    @staticmethod
+    def get_execution_time_limit(
+        config_execution_time_limit: Any
+    ) -> Union[None, float]:
+        """Get execution time limit from config and process it.
+
+        We are making no assumptions about what can be passed in,
+        but passing silently and returning None if the value is
+        not a type convertable to a float:
```
hjoliver, line 1293: Shouldn't it at least log a warning for incoming values that aren't either None or convertible to float, which would clearly be user error. e.g. `12.3x`

wxtim: I was trying to leave well alone and keep it as close to the existing code as possible. But I don't think it's a bad idea to be more specific.

markgrahamdawson (approved):

PR #5908, opened 2024-01-05
xtrigger efficiency fix.

hjoliver:
Update the DB insert-maps for xtriggers when the xtriggers get satisfied, not when the tasks that depend on them get satisfied.

This bug (which causes duplicate entries in lists of DB updates) can cause long delays (minutes) just after initially populating the task pool, if hundreds of tasks depend on the same xtrigger (typically, a clock trigger).

The workflow that revealed this is large with several hundred cycles of runahead, resulting in 1.8 million update entries instead of ~800 🤯 

We should get this into 8.2.4 given that it has caused problems out in the wild already.

**Check List**


hjoliver:
Ready to go. Any reviewers available at your end @oliver-sanders ? (DS is still on leave here).

cylc/flow/workflow_db_mgr.py
```diff
@@ -233,13 +233,15 @@ def process_queued_ops(self) -> None:
                     self.db_deletes_map.items()):
                 while db_deletes:
                     where_args = db_deletes.pop(0)
+                    # LOG.debug(f"DB {table_name} DEL {where_args}")
```
wxtim, line 236: Surely these should either be removed or uncommented?

tests/integration/test_xtrigger_mgr.py
```diff
@@ -65,3 +66,56 @@ async def test_2_xtriggers(flow, start, scheduler, monkeypatch):
...
             'clock_3': False,
         }
+
+
+async def test_1_xtrigger_2_tasks(flow, start, scheduler, monkeypatch, mocker):
+    """
+    If multiple tasks depend on the same satisfied xtrigger, the DB mgr method
+    put_xtriggers should only be called once - when the xtrigger gets satisfied.
+
+    See GitHub #5908
```
wxtim, line 76:
```suggestion
    See [GitHub #5908](https://github.com/cylc/cylc-flow/pull/5908)
```

cylc/flow/workflow_db_mgr.py
oliver-sanders: FYI `cylc play --main-loop='log db'` will write out a log file with all the SQL operations performed on the database in it.

oliver-sanders:
```suggestion
```

cylc/flow/workflow_db_mgr.py
```diff
@@ -233,13 +233,15 @@ def process_queued_ops(self) -> None:
...
                     self.pri_dao.add_delete_item(table_name, where_args)
                     self.pub_dao.add_delete_item(table_name, where_args)
         if any(self.db_inserts_map.values()):
             for table_name, db_inserts in sorted(
                     self.db_inserts_map.items()):
                 while db_inserts:
                     db_insert = db_inserts.pop(0)
+                    # LOG.debug(f"DB {table_name} INS {db_insert}")
```
oliver-sanders, line 244:
```suggestion
```

cylc/flow/workflow_db_mgr.py
```diff
@@ -248,6 +250,8 @@ def process_queued_ops(self) -> None:
                     self.db_updates_map.items()):
                 while db_updates:
                     set_args, where_args = db_updates.pop(0)
+                    # LOG.debug(
+                    #     f"DB {table_name} UPD {set_args} WHERE {where_args}")
```
oliver-sanders, line 254:
```suggestion
```

tests/integration/test_xtrigger_mgr.py
```diff
@@ -65,3 +66,56 @@ async def test_2_xtriggers(flow, start, scheduler, monkeypatch):
...
+        # Call the clock trigger via its dependent tasks, to get it satisfied.
+        for task in schd.pool.get_tasks():
+            # (For clock triggers this is synchronous)
+            schd.xtrigger_mgr.call_xtriggers_async(task)
+
+        # It should now be satisfied.
+        assert task.state.xtriggers == {'clock_1': True}
+
+        # Check one put_xtriggers call only, not two.
+        assert spy.call_count == 1
```
wxtim, line 114:
```suggestion
        spy.assert_called_once_with(
            {'wall_clock(trigger_time=1588636800)': {}})
```

wxtim (approved):
- [x] Read code. Made a few comments, but none are critical to this PR.
- [x] Manually tested bug and fix.

😄

cylc/flow/xtrigger_mgr.py
```diff
@@ -489,23 +511,9 @@ def callback(self, ctx: SubFuncContext):
             return
         LOG.debug('%s: returned %s', sig, results)
         if satisfied:
+            # Newly satisfied
             self.data_store_mgr.delta_task_xtrigger(sig, True)
+            self.workflow_db_mgr.put_xtriggers({sig: results})
```
oliver-sanders, line 516: I'm a bit confused by what should and shouldn't be in the xtriggers table.

The `put_xtriggers` routine wipes the entire table, then adds the new entry. If you call `put_xtriggers` multiple times *before* the workflow DB queue is processed, all of the wiping will happen first, and all the adding second.

https://github.com/cylc/cylc-flow/blob/df350cd5481f8e42f3e236840816ca296d855660/cylc/flow/workflow_db_mgr.py#L406-L412

Practical upshot, anything added to the xtrigger table will get deleted the next time an xtrigger is satisfied in a future main loop cycle?

oliver-sanders (commented):
LGTM, though I'm not sure the DB table is doing what it's supposed to (before or after).

[unrelated] I had a go at running a workflow that calls an xtrigger which returns data (`{'abc': 'def'}`). Taking a look at the result of running the workflow with `--main-loop='log db'`, it looks like there are duplicate SQL calls updating the broadcast tables (before and after). Guessing this ain't meant to happen?


```sql
INSERT                        
    OR REPLACE INTO broadcast_events
VALUES('2024-01-08T10:41:49.690744Z', '+', '20240108T0000Z', 'bar', '[environment]myxtrigger_abc', 'def')
                     
INSERT               
    OR REPLACE INTO broadcast_events
VALUES('2024-01-08T10:41:49.692659Z', '+', '20240109T0000Z', 'bar', '[environment]myxtrigger_abc', 'def')
                     
INSERT                     
    OR REPLACE INTO broadcast_events
VALUES('2024-01-08T10:41:49.694808Z', '+', '20240110T0000Z', 'bar', '[environment]myxtrigger_abc', 'def')
                           
INSERT                          
    OR REPLACE INTO broadcast_events
VALUES('2024-01-08T10:41:49.697189Z', '+', '20240111T0000Z', 'bar', '[environment]myxtrigger_abc', 'def')
                           
INSERT                        
    OR REPLACE INTO broadcast_events
VALUES('2024-01-08T10:41:49.699815Z', '+', '20240112T0000Z', 'bar', '[environment]myxtrigger_abc', 'def')
                              
INSERT                          
    OR REPLACE INTO broadcast_states
VALUES('20240108T0000Z', 'bar', '[environment]myxtrigger_abc', 'def')
                               
INSERT                        
    OR REPLACE INTO broadcast_states
VALUES('20240109T0000Z', 'bar', '[environment]myxtrigger_abc', 'def')
                              
INSERT               
    OR REPLACE INTO broadcast_states
VALUES('20240110T0000Z', 'bar', '[environment]myxtrigger_abc', 'def')
                     
INSERT               
    OR REPLACE INTO broadcast_states
VALUES('20240111T0000Z', 'bar', '[environment]myxtrigger_abc', 'def')
                     
INSERT               
    OR REPLACE INTO broadcast_states                                                                                               
VALUES('20240112T0000Z', 'bar', '[environment]myxtrigger_abc', 'def')
```

cylc/flow/workflow_db_mgr.py
hjoliver: It's a pity Python doesn't have multiple levels of debug logging, so we can have really fine-grained logging on demand without having too much log bloat for less serious debugging.

BTW I had initially thought this bug was causing loads of redundant method calls to update a map of DB inserts, but not actually causing redundant DB updates - hence the debug statements here - but it's a queue not a map, so it would cause both.

cylc/flow/xtrigger_mgr.py
hjoliver: > I'm a bit confused by what should and shouldn't be in the xtriggers table.

It should contain the function signature and result of all satisfied xtriggers - i.e. ones for which the function no longer needs to be called, because we already have the result.

hjoliver: >  it looks like there are duplicate SQL calls updating the broadcast tables (before and after). Guessing this ain't meant to happen?

They're not duplicate, each is for a different cycle point.

If multiple tasks depend on the same xtrigger, we broadcast its result to all of the dependent tasks.

So I don't think there's a problem.

hjoliver: > This looks like a bug to me, e.g. an xtrigger could get re-run on restart, potentially causing different data to be broadcasted into the cycle?

Yikes, good spot!  That's a different bug, although possibly this branch exposed it more.  It seems to me the table delete update does not need to be there. I've removed it, and the table content and sqlite trace looks right now, for a multi-cycle run with variously-specialized xtriggers.

I guess we need a test to ensure that all satisfied xtriggers get stored ...

tests/integration/test_xtrigger_mgr.py
hjoliver: @wxtim - do you want to argue that your version of the above is better, or easier to understand? Checking the specific call args has pros and cons, I think.

cylc/flow/xtrigger_mgr.py
oliver-sanders: > They're not duplicate, each is for a different cycle point.

Derp :facepalm: 

> Yikes, good spot! That's a different bug, although possibly this branch exposed it more

Yep, that was my conclusion.

> the table delete update does not need to be there

Without the delete, this table will accumulate indefinitely, which may be ok, so long as it doesn't mess with reflow use cases.

Presumably we could remove satisfied xtriggers from the table once the task has left the pool?

hjoliver: > I'm guessing this used to work by dumping all xtriggers for all tasks in the pool into the DB for each main loop cycle. So the delete was effectively just removing tasks no longer in the pool?

Yes, I think so. BUT, given that external triggers are "external", their result should in principle be independent of the workflow state, so it seems to me they shouldn't need to be re-run if the dependent tasks get re-triggered after dropping out of the task pool. What do you think?

oliver-sanders: That's fine by me, it will inflate the DB a bit, but it might actually speed up the DB operations (because we're no longer wiping the table every main loop cycle). So long as we store the flow numbers with the xtrigger it should be fine.

However, retries (which are implemented via xtriggers) might be a separate case?

hjoliver: > However, retries (which are implemented via xtriggers) might be a separate case?

Retries are implemented by clock xtriggers, so they have wallclock time-specific function signatures, and so an old task that gets retriggered will still have correct retry delays.

```console
~/c/xtr $ sqlite3 ~/cylc-run/xtr/runN/.service/db 'select * from xtriggers'
wall_clock(trigger_time=1704799660.2086887)|{}  # original job (fail then succeed)
wall_clock(trigger_time=1704799695.2978518)|{}  # retriggered later (fail then succeed).
```

Hmmm. Which does suggest, however: https://github.com/cylc/cylc-flow/issues/5911

oliver-sanders (approved):

PR #5909, opened 2024-01-08
Ensure that sys argv cleaner cleans up --workflow-name=foo

### LINKED ISSUE #5905: VIP: Fails if both `--workflow-name` & `--run-name` are set
`cylc vip` should handle any valid inputs for any of its component scripts but the following command fails:

```
> cylc vip insert-workflow-source --run-name=foo --workflow-name=toy
```

Producing traceback from the sysargv cleaning functions.

wxtim:
as well as --workflow-name foo

Closes #5905 

**Check List**

### DIFF
--- a/cylc/flow/option_parsers.py
+++ b/cylc/flow/option_parsers.py
@@ -33,7 +33,7 @@ from ansimarkup import (
 
 import sys
 from textwrap import dedent
-from typing import Any, Dict, Optional, List, Tuple, Union
+from typing import Any, Dict, Iterable, Optional, List, Tuple, Union
 
 from cylc.flow import LOG
 from cylc.flow.terminal import supports_color, DIM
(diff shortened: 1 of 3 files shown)


cylc/flow/option_parsers.py
```diff
@@ -841,21 +857,31 @@ def cleanup_sysargv(
...
+                    compound_opts_by_dest[unwanted_dest].kwargs['action']
                     not in ['store_true', 'store_false']
                 ):
                     sys.argv.pop(index)
-            elif arg in args:
-                index = args.index(arg)
-                sys.argv.pop(index)
+
+            # Check for `--workflow-name=foo`:
+            elif unwanted_arg in [a.split('=')[0] for a in sys.argv]:
```
oliver-sanders, line 881: Recomputing this list for each "unwanted" option is a bit expensive.

It might be easier to loop over `sys.argv` than the "unwanted_opts" because this allows you to work over the opts/args in order, it also saves popping items off of the list.

I think something like this should work (very untested!):

```python
# options which require a value e.g. "--foo=a" or "--foo a"
unwanted_compound_opts: List[str]
# options which don't require a value e.g. "--foo"
unwanted_simple_opts: List[str]

pop_next = False
args = []
for arg in sys.argv:
    if pop_next:
        pop_next = False
        continue
    parts = arg.split('=', 1)
    elif parts[0] in unwanted_compound_opts:
       # we don't want this option or its value
        if len(parts) == 1:
            # no "=" in arg, skip the next arg
            pop_next = True
    elif parts[0] in unwanted_simple_opts:
       # we don't want this option
        continue
    else:
        # keep this arg
        args.append(arg)

sys.argv = args 
```

oliver-sanders (approved):

markgrahamdawson (approved):

PR #5915, opened 2024-01-11
xtriggers: test db / restart interaction

oliver-sanders:
Follow on to #5908, add an integration test to check that a DB entry is created for the xtrigger and that satisfied xtriggers are not re-run when the workflow is restarted.


hjoliver (approved):
Nice.

PR #5920, opened 2024-01-12
Fix.ensure logger set for all caplog.set level

wxtim:
`caplog.set_level` can conflict with other instances of caplog if the logger isn't set in the tests. This was causing the manylinux tests to fail, and would have caused the same problem if all pytests were run in the same task in the fast tests.

https://github.com/wxtim/cylc/actions/runs/7501317283 (PR's with changes to Python tests only trigger the fast-tests workflow - a neat feature, but somewhat unhelpful in this particular case)

**Check List**

- [x] This is a change to testing, no new test, documentation or changelog entries are reqd


wxtim:
@oliver-sanders  pick a second reviewer if you think it required.

oliver-sanders (approved):

PR #5921, opened 2024-01-12
Cylc Message Help Contains two patterns of use

### LINKED ISSUE #5883: cylc message: three or more outputs cause problems
In a task script you can do this:

```
cylc message -- x
cylc message -- y
```

Or, for efficiency, you could do this:

```
cylc message -- x y
```

Which does the same thing with one less call. But for some reason this:

```
cylc message -- x y z
```

Does not work:

```
InputError: Workflow ID not found: x
```

Yet this usage appears to be supported:

```
Usage: cylc message [OPTIONS] -- [WORKFLOW] [JOB] [[SEVERITY:]MESSAGE ...]
```

No one has reported this, so this functionality probably isn't in use or any users of this functionality are still on Cylc 7.

Note, we have to be very careful not to create any breaking changes in the `cylc message` interface as these calls may be embedded in users scripts/executables where they are hard to change.

wxtim:
Closes #5883 

A single script might have multiple different valid arguments.

for example `cylc message` can be either
```
cylc message -- [WORKFLOW JOB [[SEVERITY:]MESSAGE ...]]] __or__
cylc message --	[[SEVERITY:]MESSAGE ...]]]
```

This PR fixes the Cylc message docs.

- Refactored part of the __init__ method of CylcOptionParser. into a method for ease of testing.
- Added unit tests to get complete coverage of new method.

### Note
The reason for this behaviour is given further down in the doc:

```
For backward compatibility, if number of arguments is less than or equal to 2,
the command assumes the classic interface, where all arguments are messages.
Otherwise, the first 2 arguments are assumed to be workflow ID and job
identifier.
```

**Check List**

- [x] This is effectively a documentation change and doesn't need logging in `CHANGES.md` or in the docs.

### DIFF
--- a/cylc/flow/option_parsers.py
+++ b/cylc/flow/option_parsers.py
@@ -369,6 +369,8 @@ class CylcOptionParser(OptionParser):
         )
     ]
 
+    LINEBREAK = ('\n', '')
+
     def __init__(
         self,
         usage: str,
@@ -407,7 +409,6 @@ class CylcOptionParser(OptionParser):
         if multitask:
             usage += self.MULTITASK_USAGE
 
-        args = ""
         self.n_compulsory_args = 0
         self.n_optional_args = 0
         self.unlimited_args = False
@@ -419,21 +420,7 @@ class CylcOptionParser(OptionParser):
         self.segregated_log = segregated_log
 
         if argdoc:
-            maxlen = max(len(arg) for arg, _ in argdoc)
-            usage += "\n\nArguments:"
-            for arg, descr in argdoc:
-                if arg.startswith('['):
-                    self.n_optional_args += 1
-                else:
-                    self.n_compulsory_args += 1
-                if arg.rstrip(']').endswith('...'):
-                    self.unlimited_args = True
-
-                args += arg + " "
-
-                pad = (maxlen - len(arg)) * ' ' + '               '
-                usage += "\n   " + arg + pad + descr
-            usage = usage.replace('ARGS', args)
+            usage = self.argdoc_parser(argdoc, usage)
 
         OptionParser.__init__(
             self,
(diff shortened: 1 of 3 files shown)


oliver-sanders:
Close, but the broken use case is still covered by the first usage, I think this does it:

```diff
diff --git a/cylc/flow/scripts/message.py b/cylc/flow/scripts/message.py
index 061d842475..b12f0e180f 100755
--- a/cylc/flow/scripts/message.py
+++ b/cylc/flow/scripts/message.py
@@ -109,8 +109,9 @@ def get_option_parser() -> COP:
         argdoc=[
             WORKFLOW_ID_ARG_DOC,
             ('JOB', 'Job ID - CYCLE/TASK_NAME/SUBMIT_NUM'),
+            ('[SEVERITY:]MESSAGE', 'Messages'),
             COP.optional(
-                ('[SEVERITY:]MESSAGE ...', 'Messages')
+                ('[SEVERITY:]MESSAGE', 'Messages')
             ),
             COP.LINEBREAK,
             COP.optional(('[SEVERITY:]MESSAGE ...', 'Messages')),
```

tests/functional/cli/04-kgo.t
MetRonnie, line 1: Is this test really worthwhile? Might end up as a maintenance drag

wxtim: I semi agree. On the other hand, it's easy enough to regenerate the files. I wonder if it might be worth just testing 

- cylc message
- cylc broadcast
- cylc tui 

Which demonstrate both patterns in the code I've changed in this PR, with the benefit of adding a double check on the two scripts which we least want to drift. I added TUI to the list, because it seems to be an outlier. This test would have caught the errors in my PR.

MetRonnie: Might be worth it for `cylc message` and `cylc broadcast` as these have to be stable for task scripts

wxtim: Seems a little non-standard - it broke the original PR>

oliver-sanders:
You might want to have a quick skim over `cylc message --help`...

tests/functional/cli/04-kgo.t
wxtim: The point is to check whether changes to the option parser have caused a change in the help text.

wxtim:
> You might want to have a quick skim over `cylc message --help` bearing in mind #5883...

Not sure what your point is - in https://github.com/cylc/cylc-flow/issues/5883#issuecomment-1886842872 we verbally agreed that the help should look like this. I think that is what it now looks like?

wxtim:
> You might want to have a quick skim over `cylc message --help` bearing in mind #5883...

I'm not sure what your point was here. As far as I can see the behaviour makes sense and is correct but the docs are wrong. What behaviour are you unhappy with?

Issue #5911, opened 2024-01-09
Don't record satisfied clock triggers in the DB 


hjoliver:
Currently all satisfied xtriggers get stored in the DB xtriggers table. This includes clock triggers, and "retry delay" triggers (which are implemented via clock triggers).

These really don't need to be stored in the DB
- clock triggers will likely be the main cause of xtrigger table bloat
- calling a clock trigger again unnecessarily, e.g. after restart, is quick (i.e. it doesn't matter) - a synchronous wall-time check, not an asynchronous sub-process call like for other xtriggers

wxtim:
I think I've done this, and I'm just writing tests for it, but I'll be refraining from pushing a branch util https://github.com/cylc/cylc-flow/pull/5908 is in on master.

PR #5923, opened 2024-01-15
Don't record satisfied wall clock triggers in the DB

wxtim:
Closes https://github.com/cylc/cylc-flow/issues/5911

n.b. I tried to find a more sophisticated method than `startswith('wall_clock(...')` to identify the correct xtriggers, but I think this will do. 

**Check List**

- [x] This is a change to internals, so it doesn't need a docs/changelog update.
- [x] Raised against 8.2.x

### DIFF
--- a/cylc/flow/workflow_db_mgr.py
+++ b/cylc/flow/workflow_db_mgr.py
@@ -405,9 +405,10 @@ class WorkflowDatabaseManager:
     def put_xtriggers(self, sat_xtrig):
         """Put statements to update external triggers table."""
         for sig, res in sat_xtrig.items():
-            self.db_inserts_map[self.TABLE_XTRIGGERS].append({
-                "signature": sig,
-                "results": json.dumps(res)})
+            if not sig.startswith('wall_clock(trigger_time='):
+                self.db_inserts_map[self.TABLE_XTRIGGERS].append({
+                    "signature": sig,
+                    "results": json.dumps(res)})
 
     def put_update_task_state(self, itask):
         """Update task_states table for current state of itask.
--- a/cylc/flow/workflow_db_mgr.py
+++ b/cylc/flow/workflow_db_mgr.py
@@ -405,7 +405,7 @@ class WorkflowDatabaseManager:
     def put_xtriggers(self, sat_xtrig):
         """Put statements to update external triggers table."""
         for sig, res in sat_xtrig.items():
-            if not sig.startswith('wall_clock(trigger_time='):
+            if not sig.startswith('wall_clock('):
                 self.db_inserts_map[self.TABLE_XTRIGGERS].append({
                     "signature": sig,
                     "results": json.dumps(res)})
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -20,11 +20,11 @@ from functools import partial
 from pathlib import Path
 import pytest
 from shutil import rmtree
+from time import time
 from typing import List, TYPE_CHECKING, Set, Tuple, Union
 
 from cylc.flow.config import WorkflowConfig
 from cylc.flow.option_parsers import Options
-from cylc.flow.network.client import WorkflowRuntimeClient
 from cylc.flow.pathutil import get_cylc_run_dir
 from cylc.flow.rundb import CylcWorkflowDAO
 from cylc.flow.scripts.validate import ValidateOptions
@@ -34,6 +34,7 @@ from cylc.flow.scripts.install import (
 )
 from cylc.flow.wallclock import get_current_time_string
 from cylc.flow.workflow_files import infer_latest_run_from_id
+from cylc.flow.workflow_status import StopMode
 
 from .utils import _rm_if_empty
 from .utils.flow_tools import (
@@ -47,6 +48,7 @@ from .utils.flow_tools import (
 if TYPE_CHECKING:
     from cylc.flow.scheduler import Scheduler
     from cylc.flow.task_proxy import TaskProxy
+    from cylc.flow.network.client import WorkflowRuntimeClient
 
 
 InstallOpts = Options(install_gop())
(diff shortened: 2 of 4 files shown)


oliver-sanders:
Execution/submission retries are also implemented via wall_clock xtriggers.

Worth checking/testing that these aren't written to the DB.

tests/integration/test_workflow_db_mgr.py
```diff
@@ -127,3 +128,53 @@ async def test_workflow_param_rapid_toggle(
...
+            },
+            "graph": {
+                "R1": "@another & @wall_clock & @real_wall_clock => foo"
+            }
+        },
+        'runtime': {
+            'foo': {
+                'execution retry delays': 'PT0S',
+                'script': (
+                    'test $CYLC_TASK_SUBMIT_NUMBER == 1 && exit 1 || exit 0')
```
wxtim, line 165: This will be do-able in simulation mode after the simulation mode refactoring, which should speed this test up rather a lot.

tests/integration/test_workflow_db_mgr.py
```diff
@@ -127,3 +128,53 @@ async def test_workflow_param_rapid_toggle(
...
+            }
+        }
+    })
+    # Run workflow unto completion:
+    schd = scheduler(id_, paused_start=False, run_mode='live')
+    async with run(schd) as log:
+        while 'Workflow shutting down - AUTOMATIC' not in log.messages:
+            await asyncio.sleep(0.5)
```
wxtim, line 173: This seems to be the best balance between waiting longer and checking more - but it won't get < 15s.

tests/integration/test_workflow_db_mgr.py
```diff
@@ -127,3 +128,53 @@ async def test_workflow_param_rapid_toggle(
...
+
+    # Get xtriggers db table:
+    info = schd.workflow_db_mgr.get_pri_dao().conn.execute(
+        'SELECT * FROM xtriggers').fetchall()
+
+    # All xtriggers are xrandom: None are wall_clock:
+    assert all(i[0].startswith('xrandom') for i in info)
```
MetRonnie, line 180: Can simply use `db_select` fixture
```suggestion
    assert db_select(schd, False, 'xtriggers', 'signature') == [
        ('xrandom(100)',),
        ('xrandom(100, _=Not a real wall clock trigger)',)
    ]
```

tests/integration/test_workflow_db_mgr.py
oliver-sanders: The `complete` fixture can do this for you. I think it's:

```python
await complete(schd)
```

tests/integration/test_workflow_db_mgr.py
oliver-sanders: Shell interaction is the line in the sand between integration and functional tests.

But I expect it can be dispensed with, here's one idea, get the simulated task to wait so you can feed the failed message into the system manually:

```ini
[runtime]
  [[foo]]
     [[[simulation]]]
       default run length = PT1M
```

```python
# fail a simulated task
schd.task_events_mgr.process_message(itask, 'error', 'failed')
```

cylc/flow/workflow_db_mgr.py
```diff
@@ -405,9 +405,10 @@ def put_task_event_timers(self, task_events_mgr):
     def put_xtriggers(self, sat_xtrig):
         """Put statements to update external triggers table."""
         for sig, res in sat_xtrig.items():
-            self.db_inserts_map[self.TABLE_XTRIGGERS].append({
-                "signature": sig,
-                "results": json.dumps(res)})
+            if not sig.startswith('wall_clock(trigger_time='):
```
oliver-sanders, line 408: Probably safer not to check the kwarg incase it's provided by position:

```suggestion
            if not sig.startswith('wall_clock('):
```

tests/integration/test_workflow_db_mgr.py
wxtim: There's a complete_task fixture?

tests/integration/test_workflow_db_mgr.py
wxtim: @oliver-sanders  - Using simulation mode isn't possible because of the bug outlined in https://github.com/cylc/cylc-flow/issues/5935.

tests/integration/conftest.py
```diff
@@ -473,3 +475,85 @@ def _inner(source, **kwargs):
         workflow_id = infer_latest_run_from_id(workflow_id)
         return workflow_id
     yield _inner
+
+
+@pytest.fixture
+def complete():
```
oliver-sanders, line 481: Unused.

tests/integration/test_workflow_db_mgr.py
```diff
@@ -127,3 +132,54 @@ async def test_workflow_param_rapid_toggle(
...
+        },
+    })
+    schd = scheduler(id_, paused_start=False, run_mode='simulation')
+
+    async with run(schd):
+        foo = schd.pool.get_task(ISO8601Point(rawpoint), 'foo')
+
+        # Wait until all xtriggers for foo are satisfied:
+        while not all(foo.state.xtriggers.values()):
+            await asyncio.sleep(1)
```
oliver-sanders, line 180: Should wrap this in `async_timeout` to avoid this test hanging indefinitely if something goes wrong.

Try `git grep async_tmeout` for example usage.

oliver-sanders (approved):

tests/integration/conftest.py
MetRonnie: It was supposed to be used as per https://github.com/cylc/cylc-flow/pull/5923#discussion_r1453217855

MetRonnie (changes requested):
`complete` fixture ported from master but still needs to be used in the test

tests/integration/conftest.py
oliver-sanders: The test for completion was changed since then, the new test is reasonable providing it is constrained by a timeout. Otherwise switch to `complete()` which has a default timeout built in.

MetRonnie (approved):
Not merging pending async timeout comment above

wxtim:
@oliver-sanders  - quick check that https://github.com/cylc/cylc-flow/pull/5923/commits/94f3a769360bc78c639f45d9ec25442244c73e5e addresses your comments?

MetRonnie (approved):

PR #5924, opened 2024-01-16
Error if graph has a dependency offset exclusively on the RHS

### LINKED ISSUE #5844: graph parser: absolute dependency offset on the RHS
At Cylc 7, absolute cycle points on the RHS of an expression resulted in a GraphParseError, e.g:

```
crunch_numbers => collate_results[2003]
```

Would result in:

```
GraphParseError: ERROR, illegal cycle point offset on the right: crunch_numbers  =>  collate_results[2003]
```

Whereas at Cylc 8 the error is not raised and the `=> collate_results[2003]` dependency is ignored.

wxtim:
Closes #5844

Raise an error if a graph has a dependency offset only on the RHS of a graph pair.

**Check List**

### DIFF
--- a/changes.d/fix.5924.md
+++ b/changes.d/fix.5924.md
@@ -1,0 +1,1 @@
+Ensure that graphs with dependency offsets on the right hand side are flagged to users.
--- a/changes.d/fix.5924.md
+++ b/changes.d/fix.5924.md
@@ -1,1 +1,1 @@
-Ensure that graphs with dependency offsets on the right hand side are flagged to users.
+Validation: a cycle offset can only appear on the right of a dependency if the task's cycling is defined elsewhere with no offset.
--- a/cylc/flow/graph_parser.py
+++ b/cylc/flow/graph_parser.py
@@ -534,12 +534,6 @@ class GraphParser:
         if right.count("(") != right.count(")"):
             raise GraphParseError(mismatch_msg.format(right))
 
-        # Split right side on AND.
-        rights = right.split(self.__class__.OP_AND)
-        if '' in rights or right and not all(rights):
-            raise GraphParseError(
-                f"Null task name in graph: {left} => {right}")
-
         # Ignore/Error cycle point offsets on the right side:
         # (Note we can only ban this for nodes at the end of chains)
         if '[' in right:
@@ -551,6 +545,12 @@ class GraphParser:
             else:
                 # This right hand side is a left elsewhere:
                 return
+
+        # Split right side on AND.
+        rights = right.split(self.__class__.OP_AND)
+        if '' in rights or right and not all(rights):
+            raise GraphParseError(
+                f"Null task name in graph: {left} => {right}")
 
         lefts: Union[List[str], List[Optional[str]]]
         if not left or (self.__class__.OP_OR in left or '(' in left):
--- a/cylc/flow/graph_parser.py
+++ b/cylc/flow/graph_parser.py
@@ -534,21 +534,23 @@ class GraphParser:
         if right.count("(") != right.count(")"):
             raise GraphParseError(mismatch_msg.format(right))
 
-        # Ignore/Error cycle point offsets on the right side:
-        # (Note we can only ban this for nodes at the end of chains)
-        if '[' in right:
-            if right in terminals:
-                raise GraphParseError(
-                    'ERROR, illegal cycle point offset on the right:'
-                    f' {left} => {right}')
-            else:
-                return
-
         # Split right side on AND.
         rights = right.split(self.__class__.OP_AND)
         if '' in rights or right and not all(rights):
             raise GraphParseError(
                 f"Null task name in graph: {left} => {right}")
+
+        # Ignore/Error cycle point offsets on the right side:
+        # (Note we can only ban this for nodes at the end of chains)
+        if '[' in right:
+            if left and right in terminals:
+                # This right hand side is at the end of a chain:
+                raise GraphParseError(
+                    'ERROR, illegal cycle point offset on the right:'
+                    f' {left} => {right}')
+            else:
+                # This right hand side is a left elsewhere:
+                return
 
         lefts: Union[List[str], List[Optional[str]]]
         if not left or (self.__class__.OP_OR in left or '(' in left):
--- a/cylc/flow/graph_parser.py
+++ b/cylc/flow/graph_parser.py
@@ -463,9 +463,8 @@ class GraphParser:
                 pairs.add((chain[i], chain[i + 1]))
 
         # Get a set of RH nodes which are not at the LH of another pair:
-        terminals = {
-            right for right in {right for _, right in pairs}
-            if right not in {left for left, _ in pairs}}
+        pairs_dict = dict(pairs)
+        terminals = set(pairs_dict.values()).difference(pairs_dict.keys())
 
         for pair in pairs:
             self._proc_dep_pair(pair, terminals)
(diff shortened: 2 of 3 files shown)


MetRonnie:
I think he used the commit message as the PR title and it got truncated, continuing on the body. I updated the title to something a little more readable

tests/unit/test_graph_parser.py
```diff
@@ -133,13 +133,33 @@ def test_graph_syntax_errors_2(seq, graph, expected_err):
...
+        param(
+            "foo => bar[1649] => baz",
+            '!illegal cycle point offset on the right',
+            id='ignore-not-exclusively-RHS-cycle-point'
+        ),
+        param(
+            "foo => bar[1649]\nbar[1649] => baz",
+            '!illegal cycle point offset on the right',
+            id='ignore-not-exclusively-RHS-cycle-point2'
+        ),
```
MetRonnie, line 151: Better suited to including in `test_trigger_equivalence` instead?

cylc/flow/graph_parser.py
```diff
@@ -462,8 +462,13 @@ def parse_graph(self, graph_string: str) -> None:
             for i in range(0, len(chain) - 1):
                 pairs.add((chain[i], chain[i + 1]))
 
+        # Get a set of RH nodes which are not at the LH of another pair:
+        terminals = {
+            right for right in {right for _, right in pairs}
+            if right not in {left for left, _ in pairs}}
```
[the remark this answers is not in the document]
wxtim:
```python
pairs = (('foo', 'bar'), ('baz', 'qux'), ('bear', 'barley'), ('bar', 'beard'))
```

### My Solution


```python
%%timeit
{r for r in {r for _, r in pairs} if r not in {l for l, _ in pairs}}
```

    1.73 µs ± 14 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)


### Ronnie's First Solution


```python
%%timeit
{r for _, r in pairs}.difference(l for l, _ in pairs)
```

    821 ns ± 6.21 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)


### Ronnie's Second Solution


```python
%%timeit
pairs_dict = dict(pairs)
set(pairs_dict.values()).difference(pairs_dict.keys())
```

    691 ns ± 11.3 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

tests/unit/test_graph_parser.py
wxtim: I'm not sure that I see why.

MetRonnie: Because the test is about errors and there are no errors in these two examples

wxtim: I'll move it if you insist, but I think that checking errors aren't raised when they shouldn't be is a valid error test.

MetRonnie: A test that doesn't assert that an exception is raised *is* testing that errors aren't raised

oliver-sanders: It's simpler to have two tests, one for the success and one for the error cases, rather than trying to merge the two:

```python
@pytest.mark.parametrize(...)
def test_do_something_good(foo):
    do_something(foo)

@pytest.mark.parametrize(...)
def test_do_something_bad(foo):
    with pytest.raises(...):
        do_something(foo)
```

Nothing wrong with having multiple tests (note that's what parametrize is doing anyway).

I like to think of test cases in terms of the "good", the "bad" and the "ugly", i.e. expected success cases, expected error cases & nasty edge cases designed to test the extremes. This wild west approach sometimes leaks into test names :grin:.

wxtim:
> > [wxtim](https://github.com/wxtim) requested a review from [MetRonnie](https://github.com/MetRonnie)
> 
> @wxtim, you have an outstanding review comment.

I think that I was waiting for Ronnie and you finish the discussion. Will fix now.

tests/unit/test_graph_parser.py
```diff
@@ -133,13 +133,40 @@ def test_graph_syntax_errors_2(seq, graph, expected_err):
...
     """Test various graph syntax errors."""
-    with pytest.raises(GraphParseError) as cm:
+    if expected_err[0] == '!':
+        # Assert method doesn't fail:
         GraphParser().parse_graph(graph)
-    assert expected_err in str(cm.value)
+    else:
+        with pytest.raises(GraphParseError) as cm:
+            GraphParser().parse_graph(graph)
+        assert expected_err in str(cm.value)
```
MetRonnie, line 152: Revert now that the "!" tests are removed

MetRonnie (approved):

oliver-sanders:
@hjoliver could you check you're happy with this approach.

This is back-filling a check which was removed in: https://github.com/cylc/cylc-flow/pull/4343/files#diff-be87b3fcb20fc3228166ad53af61efb79f0d0c84ee01fcd0c9e3092c788e0b93L374-L377

I think the logic has become more complex in order to permit examples like `a => b[-P1] => c`.

changes.d/fix.5924.md
```diff
@@ -0,0 +1 @@
+Ensure that graphs with dependency offsets on the right hand side are flagged to users.
```
hjoliver, line 1: I think this needs to be rephrased. What is "flagged to users"?  And RHS offsets are only illegal if the the does not appear elsewhere with an offset. 

Expressing this concisely ain't exactly easy though. How about this:

```suggestion
Validation: a cycle offset can only appear on the right of a dependency if the task's cycling is defined elsewhere with n offset.
```
(Ideally we might say the `the task's cycling must `.

cylc/flow/graph_parser.py
```diff
@@ -525,10 +533,16 @@ def _proc_dep_pair(
...
 
-        # Ignore cycle point offsets on the right side.
-        # (Note we can't ban this; all nodes get process as left and right.)
+        # Raise error for cycle point offsets at the end of chains
         if '[' in right:
-            return
+            if left and (right in terminals):
+                # This right hand side is at the end of a chain:
+                raise GraphParseError(
+                    'ERROR, illegal cycle point offset on the right:'
```
hjoliver, line 541: This kinda suggests that all offsets are illegal on the right, which isn't true. (See my comment on the change log line).

Also the `ERROR, ` prefix ends up being redundant (`GraphParseError: ERROR, illegal ...`).

hjoliver (approved):
If we need to merge this quickly for a maintenance release, I'm happy so long as the change log comment is straightened out, and preferably if the error message is too.

As a follow-up it would be nice to clean this up (same problem, very different error messages):

```
foo => bar[-P1] => baz  # TaskDefError: No cycling sequences defined for bar
foo => bar[-P1]  # GraphParseError: illegal cycle point offset on the right: foo => bar[-P1]
```

cylc/flow/graph_parser.py
wxtim:
```suggestion
                    'Cycle point offsets only on the right of dependency pair:'
```

MetRonnie: Or how about
> Invalid cycle point offsets on right hand side of a dependency (must be on left hand side):

wxtim: Not true:

```
foo => bar[42]
bar[42] => baz
```
Is legal.

MetRonnie: Hence the "(must be on left hand side)"

wxtim:
@MetRonnie  - I've used your words. Please merge at your leisure.

MetRonnie:
Tests need updating

MetRonnie (approved):

Issue #5932, opened 2023-12-19
edit runtime strips characters after #


ColemanTom:
### Description

I did an edit-runtime in the UI server, putting in command like `a=${a#*(}`. When I look at it again via edit-runtime, everything after the # is stripped out. Also, job submission fails.

### Reproducible Example

1. In any task, do edit-runtime via the UI. 
2. Add some text to `script` (and probably other areas), like `a=${a#*(}`. You can add extra lines after that too.
3. Everything after and inclusive of that first # is stripped out, including subsequent lines.

### Expected Behaviour

The # should not be treated as a comment and stripped out. I don't think comments should be stripped out in those textareas in general because comments could intentionally be in the textareas.

NOTE: I did try to do `a="${a#*(}"` but that didn't stop the stripping.

PR #5933, opened 2024-01-18
Fix bugs in `cylc broadcast`

MetRonnie:
- Fix traceback when using `cylc broadcast --file` with a relative filepath
- Fix bug where everything after a `#` character in a broadcast string setting (e.g. the script setting) would be  stripped out - #5932

**Check List**

- [x] No dependency changes
- [x] No docs update needed

### DIFF
--- a/cylc/flow/parsec/validate.py
+++ b/cylc/flow/parsec/validate.py
@@ -26,7 +26,7 @@ import re
 import shlex
 from collections import deque
 from textwrap import dedent
-from typing import List, Dict, Any, Tuple
+from typing import List, Dict, Any, Optional, Tuple
 
 from metomi.isodatetime.data import Duration, TimePoint
 from metomi.isodatetime.dumpers import TimePointDumper
@@ -512,6 +512,22 @@ class ParsecValidator:
             return None
 
     @classmethod
+    def _unquote(cls, keys: List[str], value: str) -> Optional[str]:
+        """Unquote value."""
+        for substr, rec in (
+            ("'''", cls._REC_MULTI_LINE_SINGLE),
+            ('"""', cls._REC_MULTI_LINE_DOUBLE),
+            ('"', cls._REC_DQ_VALUE),
+            ("'", cls._REC_SQ_VALUE)
+        ):
+            if value.startswith(substr):
+                match = rec.match(value)
+                if not match:
+                    raise IllegalValueError("string", keys, value)
+                return match[1]
+        return None
+
+    @classmethod
     def strip_and_unquote(cls, keys: List[str], value: str) -> str:
         """Remove leading and trailing spaces and unquote value.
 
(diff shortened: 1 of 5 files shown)


cylc/flow/parsec/validate.py
```diff
@@ -1136,23 +1137,25 @@ def _coerce_type(cls, value):
...
     """Validate and Coerce DB loaded broadcast config to internal objects."""
     def __init__(self):
         CylcConfigValidator.__init__(self)
 
+    @classmethod
+    def coerce_str(cls, value, keys) -> str:
+        """Coerce value to a string.
+
+        Examples:
+            >>> BroadcastConfigValidator.coerce_str('abc#def', None)
```
wxtim, line 1150: Why not give an example of one of the problematic shell scripts?

tests/functional/broadcast/10-file-1/broadcast.cylc
```diff
@@ -1,6 +1,8 @@
 script="""
 printenv CYLC_FOOBAR
 
+# (This hash char should not cause the rest of the script to be stripped out)
```
wxtim, line 4:
```suggestion
# (This hash char should not cause the rest of the script to be stripped out)
# See https://github.com/cylc/cylc-flow/pull/5933
```

wxtim (approved):
- [x] Tried  it before and after - looks like it fixes both bugs.
- [x] Read the code. Made comments, but neither is critical.

cylc/flow/parsec/validate.py
```diff
@@ -1136,23 +1137,25 @@ def _coerce_type(cls, value):
...
+            'abc#def'
+        """
+        # Prevent ParsecValidator from assuming '#' means comments;
+        # '#' has valid uses in shell script such as parameter substitution
+        if isinstance(value, str) and '#' in value:
+            value = f'"{value}"'
```
oliver-sanders, line 1156: This doesn't work if the string is already quoted e.g:

```console
$ cylc broadcast one -n root -s 'script="sleep 30 # ja!"'
ERROR: [{'error': {'message': '(type=string) script = ""sleep 30 # ja!""', 'traceback': ['graphql.error.located_error.GraphQLLocatedError: (type=string) script = ""sleep 30 # ja!""\n']}}]
```

Maybe use `shlex.quote` which has logic to avoid surplus quoting.

cylc/flow/parsec/validate.py
```diff
@@ -1136,23 +1140,25 @@ def _coerce_type(cls, value):
...
+
+        Prevents ParsecValidator from assuming '#' means comments;
+        '#' has valid uses in shell script such as parameter substitution.
+
+        Examples:
+            >>> BroadcastConfigValidator.coerce_str('echo "${FOO#*bar}"', None)
+            'echo "${FOO#*bar}"'
+        """
+        val = ParsecValidator._unquote(keys, value) or value
+        return dedent(val).strip()
```
oliver-sanders, line 1160: Should the dedent/strip happen before the _unquote?

If not, wouldn't `  'foo'  ` end up as `"  'foo'  "`?

cylc/flow/parsec/validate.py
```diff
@@ -1136,23 +1140,25 @@ def _coerce_type(cls, value):
...
-#    8.1.x
-# remove at:
-#    8.x
 class BroadcastConfigValidator(CylcConfigValidator):
     """Validate and Coerce DB loaded broadcast config to internal objects."""
     def __init__(self):
         CylcConfigValidator.__init__(self)
 
+    @classmethod
+    def coerce_str(cls, value, keys) -> str:
```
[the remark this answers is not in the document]
MetRonnie: This is overriding a class method from `ParsecValidator`

cylc/flow/parsec/validate.py
MetRonnie: This stripping is for the stuff inside the quotes, and this is the same as it was before this PR. I've tested
- `cylc broadcast -s 'script =   "foo"  '`
- `cylc broadcast -s 'script =   foo  '`

and both set `script=foo`

oliver-sanders (approved):

PR #5943, opened 2024-01-29
stop after cycle point: support offsets

### LINKED ISSUE #5939: config: support offsets in hold and stop after cycle points
The final cycle point configuration supports offsets e.g:

```ini
final cycle point = 2000
final cycle point = 2000 + P1Y  # i.e. 2001
final cycle point = +P1Y  # i.e. "initial cycle point" + P1Y
```

However, the stop and hold cycle points do not.

It would be nice if these configs could work in the same way.

> Context / why is this an 8.3.0 issue?
>
> A common pattern in some areas is to run a workflow until it stops. Then modify the final cycle point to extend the workflow and restart.
>
> This isn't an easy working pattern to port to Cylc 8 due to the difficulty of re-populating the start tasks. We have a plan to make this easier, see #5416, however, this working pattern is needlessly complex in the first place.
>
> Swap `final cycle point` for `stop after cycle point` and the workflow never "completes" so the pool never needs re-polulating. This is much nicer for Cylc (no discontinuity), the user (no extra commands) and provenance (direct continuation of previous run).

oliver-sanders:
* Closes #5939
* Support offsets (measured from the ICP) for the `stop after cycle point` to mirror the behaviour of the `final cycle point`.
* Add integration test to lock-down `stop after cycle point` interactions.

**Check List**

### DIFF
--- a/cylc/flow/parsec/validate.py
+++ b/cylc/flow/parsec/validate.py
@@ -702,25 +702,21 @@ class CylcConfigValidator(ParsecValidator):
         ),
         V_CYCLE_POINT_WITH_OFFSETS: (
             'cycle point with support for offsets',
-            'An integer of date-time cycle point.',
+            'An integer or date-time cycle point, with optional offset(s).',
             {
                 '1': 'An integer cycle point.',
                 '1 +P5': (
-                    'An integer cycle point with offsets.',
-                    ' (this evaluates as ``6``)'
+                    'An integer cycle point with an offset'
+                    ' (this evaluates as ``6``).'
                 ),
                 '+P5': (
                     'An integer cycle point offset.'
                     ' This offset is added to the initial cycle point'
                 ),
                 '2000-01-01T00:00Z': 'A date-time cycle point.',
-                '2000-01-01T00:00Z +P1D +P1M': (
-                    'A date-time cycle point with offsets.'
-                    ' (this evaluates as ``2000-02-02T00:00Z``'
-                ),
-                '2000-01-01T00:00Z +P1D': (
-                    'A date-time offset.'
-                    ' This offset is added to the initial cycle point'
+                '2000-02-29T00:00Z +P1D +P1M': (
+                    'A date-time cycle point with offsets'
+                    ' (this evaluates as ``2000-04-01T00:00Z``).'
                 ),
             }
         ),
--- a/cylc/flow/parsec/validate.py
+++ b/cylc/flow/parsec/validate.py
@@ -657,6 +657,7 @@ class CylcConfigValidator(ParsecValidator):
     V_CYCLE_POINT = 'V_CYCLE_POINT'
     V_CYCLE_POINT_FORMAT = 'V_CYCLE_POINT_FORMAT'
     V_CYCLE_POINT_TIME_ZONE = 'V_CYCLE_POINT_TIME_ZONE'
+    V_CYCLE_POINT_WITH_OFFSETS = 'V_CYCLE_POINT_WITH_OFFSETS'
     V_INTERVAL = 'V_INTERVAL'
     V_INTERVAL_LIST = 'V_INTERVAL_LIST'
     V_PARAMETER_LIST = 'V_PARAMETER_LIST'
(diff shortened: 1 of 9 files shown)


cylc/flow/parsec/validate.py
```diff
@@ -699,6 +700,30 @@ class CylcConfigValidator(ParsecValidator):
                 '-0830': 'UTC minus 8 hours and 30 minutes.'
             }
         ),
+        V_CYCLE_POINT_WITH_OFFSETS: (
```
oliver-sanders, line 703: This is a new parsec validator type which does the same thing as `V_STRING`.

It exists purely for documentation purposes.

cylc/flow/task_pool.py
```diff
@@ -1026,7 +1026,7 @@ def set_stop_point(self, stop_point: 'PointBase') -> bool:
             LOG.info(f"Stop point unchanged: {stop_point}")
             return False
 
-        LOG.info("Setting stop point: {stop_point}")
+        LOG.info(f"Setting stop point: {stop_point}")
```
oliver-sanders, line 1029: Minor bugfix.

tests/unit/test_config.py
```diff
@@ -114,7 +114,7 @@ def test_xfunction_imports(
         """
         flow_file.write_text(flow_config)
         workflow_config = WorkflowConfig(
-            workflow="name_a_tree", fpath=flow_file, options=Mock(spec=[]),
+            workflow="name_a_tree", fpath=flow_file, options=SimpleNamespace(spec=[]),
```
oliver-sanders, line 117: These tests were written with `Mock`, but none of them were actually relying on mocking in any way so converted them to `SimpleNamespace` which does the same thing, but with less unnecessary extras.

wxtim (changes requested):
You have test failures - should be trivial fixes, as I think both are caused by a change in error message which reflects this PR being OK.

tests/unit/test_config.py
MetRonnie: In that case, might as well be

```suggestion
            workflow="name_a_tree", fpath=flow_file, options=None,
```

or at least need to remove the `spec` arg if using `SimpleNamespace`

```suggestion
            workflow="name_a_tree", fpath=flow_file, options=SimpleNamespace(),
```

(Btw `Mock` is intended be used for stubbing too, it doesn't necessarily have to be used as a full-blown mock - https://docs.python.org/3/library/unittest.mock.html#the-mock-class)

tests/unit/test_config.py
```diff
@@ -692,24 +698,28 @@ def test_process_fcp(
...
         ),
         pytest.param(
             '1993', '1066', '1066', '1066', None,
-            id="From options"
+            id="stop-cp-and-cli-option"
         ),
         pytest.param(
             '1993', 'reload', '1993', None, None,
-            id="From cfg if --stopcp=reload on restart"
+            id="stop-cp-and-cli-reload-option"
```
MetRonnie, line 713: I think the previous ids were more descriptive

tests/unit/test_config.py
```diff
@@ -148,7 +148,7 @@ def test_xfunction_import_error(self, mock_glbl_cfg, tmp_path):
             WorkflowConfig(
                 workflow="caiman_workflow",
                 fpath=flow_file,
-                options=Mock(spec=[])
+                options=SimpleNamespace(spec=[])
```
MetRonnie, line 151:
```suggestion
                options=SimpleNamespace()
```

tests/unit/test_config.py
```diff
@@ -179,7 +179,7 @@ def test_xfunction_attribute_error(self, mock_glbl_cfg, tmp_path):
         flow_file.write_text(flow_config)
         with pytest.raises(XtriggerConfigError) as excinfo:
             WorkflowConfig(workflow="capybara_workflow", fpath=flow_file,
-                           options=Mock(spec=[]))
+                           options=SimpleNamespace(spec=[]))
```
MetRonnie, line 182:
```suggestion
                           options=SimpleNamespace()
```

tests/unit/test_config.py
```diff
@@ -211,7 +211,7 @@ def test_xfunction_not_callable(self, mock_glbl_cfg, tmp_path):
             WorkflowConfig(
                 workflow="workflow_with_not_callable",
                 fpath=flow_file,
-                options=Mock(spec=[])
+                options=SimpleNamespace(spec=[])
```
MetRonnie, line 214:
```suggestion
                options=SimpleNamespace()
```

tests/unit/test_config.py
```diff
@@ -1170,7 +1185,7 @@ def test_check_circular(opt, monkeypatch, caplog, tmp_flow_config):
     # ----- Setup -----
     caplog.set_level(logging.WARNING, CYLC_LOG)
 
-    options = Mock(spec=[], is_validate=True)
+    options = SimpleNamespace(spec=[], is_validate=True)
```
MetRonnie, line 1188:
```suggestion
    options = SimpleNamespace(is_validate=True)
```

tests/unit/test_config.py
```diff
@@ -1739,7 +1754,7 @@ def test_cylc_env_at_parsing(
 
     # Parse the workflow config then check the environment.
     WorkflowConfig(
-        workflow="name", fpath=flow_file, options=Mock(spec=[]),
+        workflow="name", fpath=flow_file, options=SimpleNamespace(spec=[]),
```
MetRonnie, line 1757:
```suggestion
        workflow="name", fpath=flow_file, options=SimpleNamespace(),
```

tests/unit/test_config.py
oliver-sanders: fair enough, I changed them to something more CLI friendly, suggestions welcome.

MetRonnie: Having a short id plus a more descriptive comment would be reasonable IMO

Issue #5946, opened 2024-01-30
pre-initial: inconsistent handling for format 4 reccurences


oliver-sanders:
The pre-initial condition is handled differently for some recurrences to others.

E.G:

```ini
[scheduler]
    cycle point format = CCYY
    allow implicit tasks = True

[scheduling]
    initial cycle point = 2000
    final cycle point = 2010
    [[graph]]
        # cycle forwards
        R3/2000/P1Y = f[-P1Y] => f

        # cycle backwards
        R3/P1Y/2010 = b[-P1Y] => b
```

The pre-initial condition is applied to the first recurrence, but not to the second:

![graph](https://github.com/cylc/cylc-flow/assets/16705946/5e7c3481-5af8-4d14-b3ef-1d8432878d09)

This behaviour appears consistent with Cylc 7 but doesn't make a whole lot of sense.

See also https://github.com/cylc/cylc-flow/issues/5945

hjoliver:
I haven't looked at the code yet, but it might suggest a problem identifying the "first point" in the (so-called) backward recurrence. Or the "first point after the initial cycle point" ...

oliver-sanders:
I tried jamming in a `P1Y` recurrence to check that the issue wasn't an interaction with the runahead limit, but it behaved the same.

hjoliver:
See the discussion on #5945 - if I haven't got my wires badly crossed, this didn't work in Cylc 7 either, because pre-initial ignore of inter-cycle deps was only ever for the workflow ICP, it wasn't applied to individual sequences that start after the ICP.

oliver-sanders:
> This didn't work in Cylc 7 either

Please see the description above where I wrote:

> This behaviour appears consistent with Cylc 7 but doesn't make a whole lot of sense.

I think you've got muddled between the two issues. These may have the same underlying cause but I've written them up separately from the perspective of behaviours.

hjoliver:
So is your main point here essentially a duplicate of #1936 ?

oliver-sanders:
Yes, gotcha.

Issue #5945, opened 2024-01-30
sod: premature shutdown for recurrence format 4


oliver-sanders:
Cylc 8 may silently ignore some format 4 recurrences.

E.G:

```ini
[scheduler]
    cycle point format = CCYY
    allow implicit tasks = True

[scheduling]
    initial cycle point = 2000
    final cycle point = 2010
    [[graph]]
        # cycle forwards
        R3/2000/P1Y = f[-P1Y] => f

        # cycle backwards
        R3/P1Y/2010 = b[-P1Y] => b
```

**Cylc 8:** Spawns the `f` chain of tasks and shuts down on cycle 2003
**Cylc 7:** Spawns both the `f` and `b` chains.

hjoliver:
Just tried with the "backward" recurrence alone:
```
INFO - Cylc version: 8.3.0.dev
INFO - Run mode: live
INFO - Initial point: 2000
INFO - Final point: 2010
INFO - Cold start from 2000
INFO - New flow: 1 (original flow from 2000) 2024-01-31 00:41:23
DEBUG - Runahead: base point 2008
DEBUG - Runahead limit: 2010
```

So this is runahead-related in the sense that at start-up, when there are no tasks in the pool, we compute the runahead limit using sequence points alone - something going wrong there it seems.

MetRonnie:
Looks like it is a problem with the pre-initial dependency (if that's the right term). Adding in `R1//2007 = b` or `R1/2007 = b` causes the sequence to start

hjoliver:
Yeah, I think I understand it ...

hjoliver:
Actually, I'm kinda surprised this worked in Cylc 7. (Did it?)

The pre-initial dependency handling only ever (as I recall) applied to the initial cycle point of the workflow, not individual sequence start points. 

That rings a bell ...

hjoliver:
Looks like I flagged this ages ago: 

https://github.com/cylc/cylc-flow/issues/1936

And see the final comment from 1 year ago:

https://github.com/cylc/cylc-flow/issues/1936#issuecomment-1031825906

MetRonnie:
Ah yes, if you change the start of the format 3 recurrence (the "forwards" example) to a point beyond the ICP e.g. `R3/2002/P1Y = f[-P1Y] => f`, then you get the same immediate shutdown

hjoliver:
[...]

Issue #5416, opened 2023-03-17
easier way to select cycle start tasks


oliver-sanders:
> Meeting 2023-03-16

When a workflow is first run Cylc uses the following logic to determine the tasks which are initially added to the pool:

* Tasks with no parents.
* With cycle points on or after the initial cycle point.
* Subject to the runahead limit.

If we want to start from the beginning in a new flow, say because the workflow has completed and we want to re-run it, Cylc does not currently poses any switches to allow the user to replicate this initial startup logic. Instead the user must work out what tasks they want to (re)start the workflow with and manually specify them as start-tasks.

What we could do with is something with functionality along the lines of:

```
$ cylc play --start-tasks='*' --start-cycle='<cycle>' --flow=new
```

hjoliver:
> When a workflow is first run Cylc uses the following logic to determine the tasks which are initially added to the pool:

We can also start from a specified cycle point, or from specified tasks in the graph.  But only for a first run from scratch, not a restart.

hjoliver:
I have another possible use case, for restarting the same workflow run from a different point (although not the start of the graph).

Consider a DR situation with some workflows that need to fail-over and fail-back between two platforms.  To avoid housekeeping or adapting the system to new run directories it may be convenient to restart the same workflow runs each time you fail over, but from a new point in the graph corresponding to the latest sync point from the other platform (not from wherever the workflow got to last to it ran on this platform).

oliver-sanders:
This could also apply when triggering a new flow e.g:

```bash
# start a new flow at a specified start cycle e.g:
# * I changed the configuration and want to re-run everything from a specified cycle
#   onwards...
# * I fixed the data and want to rerun the workflow from that cycle on...
#
# ...but I don't want to specify all the cycle-start tasks because the graph is complicated
# or I didn't write it, or I don't want to manually specify all of the start tasks.
cylc trigger --flow=new --start-cp=<cycle> <workflow>
```

I've just received a use case for something like:

```bash
# start a new flow at a specified cycle which stops at a specified cycle e.g:
# * I changed the configuration and want to re-run a specified cycle range.
# * I changed the data and want to re-run a specified cycle range.
cylc trigger --flow=new --start-cp=<cycle1> --stop-cp=<cycle2> <workflow>
```

oliver-sanders:
Updated the issue title to reflect the more general problem of selecting the cycle start tasks (i.e. the default cold-start logic)

oliver-sanders:
[...]

PR #5947, opened 2024-01-30
example: extending workflow

oliver-sanders:
Document the preferred approach to "extending" a workflow.

#### Context:

A common pattern in some areas is to run a workflow until it stops. Then modify the final cycle point to extend the workflow and restart.

This isn't an easy working pattern to port to Cylc 8 due to the difficulty of re-populating the start tasks. We have a plan to make this easier, see https://github.com/cylc/cylc-flow/issues/5416, however, this working pattern is needlessly complex in the first place. It would be much simpler if the workflow didn't "finish" in the first place.

To do this, swap final cycle point for stop after cycle point, now the workflow will not "finish" so the pool will not need re-polulating for the workflow to continue. This is much nicer for Cylc (no discontinuity), the user (no extra commands) and provenance (direct continuation of previous run).

#### Blockers.

* https://github.com/cylc/cylc-flow/issues/5939
  This is a blocker to real world usage (but not to this example) where the `final cycle point` is typically specified as an offset from the initial cycle point.
* https://github.com/cylc/cylc-flow/issues/5946
   This causes the "complex" example to stall.
* https://github.com/cylc/cylc-flow/issues/5945
   This causes the final cycle point sequences to be silently skipped.

**Check List**

- [x] `CHANGES.md` - docs only


hjoliver:
Agreed on documenting to use a stop point rather than final cycle point, to allow "extending" a workflow run more easily.

As revealed by the discussion on #5952 automatically extending a workflow run is not trivial if any graph structure is tied to the  stop- or final-point. It will be crucial to document the potential consequences of that. The simple example in that issue can, I think, be handled automatically, but I'm not sure that's universal.

oliver-sanders:
> So I guess we'll still have to document how to do it the hard way as well

We'd rather not, the FCP approach is very awkward and are keen to move users away from it.

We've got a lot of users who rely on this pattern which was very simple at Cylc 7 but very difficult at Cylc 8, we need to provide a simple solution for this use case.

hjoliver:
> We'd rather not, the FCP approach is conceptually awkward and are keen to move users away from it.

That's all very well, but it's going to happen anyway, as in *"Help, my workflow finished, how do it move the FCP and continue it",*

oliver-sanders:
> It's not so difficult now: restart with --pause and trigger the first cycle after the original FCP.

The triggering is often non-trivial, we cannot provide generic advice to do this, or a single command like we used to have.

hjoliver:
> >    We'd rather not, the FCP approach is conceptually awkward and are keen to move users away from it.
> 
> That's all very well, but it's going to happen anyway, as in *"Help, my workflow finished, how do it move the FCP and continue it",*

Case in point https://cylc.discourse.group/t/extending-and-restarting-a-a-workflow/911

oliver-sanders:
True but unhelpful, we would like to avoid using this pattern for workflows which are intended to be extended.

hjoliver:
I don't think we're disagreeing on the fundamental point here. 

I'm just saying in addition to that, we still need to document how to restart after extending the FCP *if you do get yourself into that particular fix.* 

Recommending a better way isn't much help if users don't anticipate the need for extending a workflow, or they don't see the advice, before starting a run.

PR #5956, opened 2024-02-01
cylc lint: deprecate `[cylc-lint]` in favour of `[tool.cylc.lint]` in `pyproject.toml` 

### LINKED ISSUE #5811: Lint: Use [tool.*] table within pyproject.toml
### Problem
Cylc lint currently uses the top level `[cylc-lint]` table in pyproject.toml. This should be under the `[tool]` table to [follow PEP 518](https://peps.python.org/pep-0518/#tool-table), and avoid possible namespace clashes in future.

### Proposed Solution

To avoid namespace clashes the PEP suggests using a table key that is owned on PyPI. For cylc this would be [cylc-flow](https://pypi.org/project/cylc-flow/).

```toml
[tool.cylc-flow]
    # cylc lint settings go here.
```

Alternatively, if it is thought that other bits of cylc would want to use pyproject.toml for configuration, a specific lint table could be added.

```toml
[tool.cylc-flow.lint]
    # cylc lint settings go here.
```

Depending on how widespread usage of cylc lint is currently, the current `[cylc-lint]` table should either be immediately dropped (if very few people are currently using it), or instead it should be checked after the `[tool.cylc-flow]` table, and emit a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) if it is the only source of configuration.

MetRonnie:
Closes #5811 

**Check List**

- [x] No dependency changes 
- [x] Tests are included
- [x] https://github.com/cylc/cylc-doc/pull/687

### DIFF
--- a/cylc/flow/scripts/lint.py
+++ b/cylc/flow/scripts/lint.py
@@ -22,8 +22,6 @@ Checks code style, deprecated syntax and other issues.
 # NOTE: docstring needed for `cylc help all` output
 # (if editing check this still comes out as expected)
 
-LINT_SECTIONS = ['cylc-lint', 'cylclint', 'cylc_lint']
-
 COP_DOC = """cylc lint [OPTIONS] ARGS
 
 Check .cylc and .rc files for code style, deprecated syntax and other issues.
@@ -44,12 +42,12 @@ It is good practice to specify specific errors you wish to ignore using
 """
 
 TOMLDOC = """
-pyproject.toml configuration:{}
-   [cylc-lint]                     # any of {}
-       ignore = ['S001', 'S002']    # List of rules to ignore
-       exclude = ['etc/foo.cylc']  # List of files to ignore
-       rulesets = ['style', '728'] # Sets default rulesets to check
-       max-line-length = 130       # Max line length for linting
+pyproject.toml configuration:
+   [tool.cylc.lint]
+   ignore = ['S001', 'S002']    # List of rules to ignore
+   exclude = ['etc/foo.cylc']   # List of files to ignore
+   rulesets = ['style', '728']  # Sets default rulesets to check
+   max-line-length = 130        # Max line length for linting
 """
 from colorama import Fore
 import functools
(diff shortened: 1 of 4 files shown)


cylc/flow/scripts/lint.py
```diff
@@ -19,11 +19,9 @@
 Checks code style, deprecated syntax and other issues.
 
 """
-# NOTE: docstring needed for `cylc help all` output
+# NOTE: docstring needed for `cylc help all` output and docs
 # (if editing check this still comes out as expected)
 
-LINT_SECTIONS = ['cylc-lint', 'cylclint', 'cylc_lint']
```
MetRonnie, line 25: Using `[cylclint]` or `[cylc_lint]` resulted in traceback so I've gone ahead and removed those

cylc/flow/scripts/lint.py
```diff
@@ -1216,20 +1248,7 @@ def get_cylc_files(
...
-    },
-    'auto gen message': (
-        'U998 and U999 represent automatically generated'
-        ' sets of deprecations and upgrades.'
-    ),
-}
-
-
-def get_reference(linter, output_type):
+def get_reference(linter: str, output_type: 'Literal["text", "rst"]') -> str:
```
oliver-sanders, line 1251: `s/linter/ruleset`?

oliver-sanders (approved):

cylc/flow/scripts/lint.py
wxtim: Massive violation of [PEP20.15](https://github.com/python/cpython/blob/5e6482806482af364ec93823893a49767e5f786f/Lib/this.py#L15) - yes I know I wrote it...

cylc/flow/scripts/lint.py
```diff
@@ -44,12 +42,12 @@
...
 
 TOMLDOC = """
-pyproject.toml configuration:{}
-   [cylc-lint]                     # any of {}
-       ignore = ['S001', 'S002']    # List of rules to ignore
-       exclude = ['etc/foo.cylc']  # List of files to ignore
-       rulesets = ['style', '728'] # Sets default rulesets to check
-       max-line-length = 130       # Max line length for linting
+pyproject.toml configuration:
+   [tool.cylc.lint]
```
wxtim, line 46: NB: there is at least one person that this will break this for.

MetRonnie: They will get a deprecation warning

oliver-sanders: This isn't breaking back compat, yet.

wxtim (approved):

PR #5986, opened 2024-02-21
CLI: glob and hold help tweaks

hjoliver:
- (re)emphasize that globs only match active tasks
- fix a typo and minor grammar issues 
- document that a held running task can still spawn downstream action

One quick review should do.


oliver-sanders (approved):
May as well rebase onto 8.2.x as 8.2.5 will see release before 8.3.0.

hjoliver:
Done. No surprises, will merge.

Issue #5968, opened 2024-02-14
vr: plugin options not respected when workflow is running


oliver-sanders:
The `-S` option does not take effect with the `cylc vr` command in the situation where the workflow is already running (i.e. the reload case).

### Reproducible Example:

```ini
#!Jinja2

[scheduling]
    [[graph]]
        R1 = a

[runtime]
    [[a]]
        script = echo {{ VAR }}
```

```console
$ cylc vip --pause -S VAR=1
$ cylc vr <id> -S VAR=2
$ cd ~/cylc-run/<id>
```

`VAR` *is* updated in the Rose files:

```console
$ cat opt/rose-suite-cylc-install.conf
...
[template variables]
VAR=2
```

But not in the Cylc files:

```console
$ cat log/config/02-reload-01.cylc
...
[runtime]
    [[a]]
        script = echo 1
$ cat log/config/flow-processed.cylc
...
[runtime]
    [[a]]
        script = echo 1
```

Example derived from: https://cylc.discourse.group/t/jinja2-variable-differences-in-config-files-new-ones-are-not-being-respected/885/6

### How it's supposed to work:
* The `-S` option is provided by cylc-rose and applies to (re)install.
* When you provide the `-S` option, it writes the input to the cylc-install optional config (`~/cylc-run/<workflow>/opt/rose-suite-cylc-install.conf`).
* When the workflow is reloaded, the cylc-rose plugin is automatically invoked.
* It reads these inputs from the cylc-install optional config and makes them available for the parsing of the workflow configuration.

### Ongoing Work

Note, there have been recent changes (unreleased) to the reinstall code and there are pending changes to the way the plugins are run: https://github.com/cylc/cylc-flow/pull/5868

Suggest reproducing and fixing on top of that branch.

oliver-sanders:
The root cause of this bug is that the CLI options are held in memory, they are subsequently re-used to re-load the workflow config:

https://github.com/cylc/cylc-flow/blob/8ab57ad0943abeea5b297f7d3b19d3cd8a055175/cylc/flow/scheduler.py#L1272-L1287

Inspecting the options at this point will reveal the previous value:

```
opts.rose_template_vars
['VAR=1', 'ROSE_ORIG_HOST=vld601.cmpd1.metoffice.gov.uk']
```

This makes it look like the plugin as if `-S VAR=1` had been specified on the command line which overrides the `VAR=2` value present in the optional configuration.

wxtim:
[...]

PR #5996, opened 2024-02-23
Clear Template Variables from scheduler options before reload

wxtim:
Closes https://github.com/cylc/cylc-flow/issues/5968

> [!WARNING]
> Branch is based on https://github.com/cylc/cylc-flow/pull/5868 and should not be merged before that.

This passes the manual test described in https://github.com/cylc/cylc-flow/issues/5968, and the equivelent test with `-D`.

I've tried to integration and functional tests, but have hit assorted walls:
- Functional relies on Rose and Cylc-Rose installs, it might be possible to fake a plugin for the test, but I struggled to make it work with Rose.
- Integration makes it easier to mock the plugin, but I've struggled to mock the moving parts required without making the test a meaningless cookie-cutter inversion of the logic.
- Given that this is about CLI interaction Functional tests make more sense IMO.

**Check List**

- [ ] Tests - See explanations above.


cylc/flow/scheduler.py
```diff
@@ -1269,6 +1269,8 @@ def _configure_contact(self) -> None:
     def load_flow_file(self, is_reload=False):
         """Load, and log the workflow definition."""
         # Local workflow environment set therein.
+        self.options.rose_template_vars = []
```
oliver-sanders, line 1272: Please add a comment, else we will forget why we do this.

wxtim:
```suggestion
        # https://github.com/cylc/cylc-flow/issues/5968
        self.options.rose_template_vars = []
```

oliver-sanders (approved):

PR #6008, opened 2024-03-05
Ensure task mail respects `[scheduler][mail]to/from`

### LINKED ISSUE #5976: mail: [scheduler][mail]to is not being used as the default for [runtime][root][mail]to
According to the config change proposal, the scheduler's `[mail]to` is supposed to act at the default for the `[runtime]` equivalent.

https://cylc.github.io/cylc-admin/proposal-config-changes.html#cylceventsmail----schedulermail

So, in this example the failed task mail should be sent to `me@example.com`:

```ini
[scheduler]
    [[mail]]
        to = me@example.com

[scheduling]
    [[graph]]
        R1 = foo

[runtime]
    [[root]]
        [[[events]]]
            mail events = failed

    [[foo]]
        script = false
```

But it doesn't look like that's happening.

MetRonnie:
Closes #5976

Plus tidy some cruft I spotted along the way (sorry, couldn't resist)

**Check List**

- [x] No dependency changes 
- [x] Tests are included
- [x] No docs entry needed


tests/unit/test_task_events_mgr_2.py
MetRonnie, line 1: Moved this into `test_task_events_mgr.py`. Not sure why there was ever a "2" file?

cylc/flow/task_events_mgr.py
```diff
@@ -429,23 +429,23 @@ class TaskEventsManager():
...
+    mail_smtp: Optional[str] = None
+    mail_footer: Optional[str] = None
+
     def __init__(
         self, workflow, proc_pool, workflow_db_mgr, broadcast_mgr,
         xtrigger_mgr, data_store_mgr, timestamp, bad_hosts,
         reset_inactivity_timer_func
     ):
         self.workflow = workflow
-        self.workflow_url = None
```
MetRonnie, line 435: This was unused

hjoliver (approved):
LGTM

wxtim (approved):
I'm satisfied this works. Have double checked that it doesn't break Cylc Review.

Issue #6001, opened 2024-02-28
internals: task "output" vs "message"


hjoliver:
For historical reasons the naming of data structures and variables associated with task outputs and prerequisites is a bit confusing:
- originally task outputs were task messages, full stop
   - consequently tasks messages are the fundamental thing stored as outputs and prereqs in the code
- then we added short names for use as triggers in the graph
   - these got variously called output "names", "labels", or "triggers", and sometimes "outputs"

For Cylc 8, the way we talk about optional and required outputs elevates the importance of the short names.  So in the documentation we should call these "outputs", and we should only call task messages "task messages" not "outputs". 

We should:
1. ensure that naming in the code is  consistent with this
2. consider changing code internals to store the short "outputs" rather than task messages. The translation can be done immediately when task messages are received by the task event manager

hjoliver:
Also, for historical reasons, the way task outputs are store (in the `TaskOutputs` class) is quite ungaily. (3 overlapping dicts: _by_trigger, _by_message, and _required). I'm sure this could be reduced to one simple data structure if we switch using only outputs (as opposed to messages) internally.

PR #6020, opened 2024-03-13
Task expire optionality.

hjoliver:
Salvaged from the aborted #5934, in case it's still useful.

@oliver-sanders dunno if you've already re-implemented this functionality on your new post `cylc set` branch? 

validation:
- clock-expire must be flagged as optional in the graph
- and `:expired` must be optional if used 

runtime expire:
  - can't expire an active task
  - manual expire: stall with incomplete task if expire was not optional
  - otherwise remove the expired task as complete

**Check List**


oliver-sanders:
I'm currently working towards this on my branch, but still useful to see what you've done.

hjoliver:
BTW note my use of  `TaskOutputs._expire_ok` here. That's because `:expired` is either optional or not allowed, and if not allowed then **completing the output makes the task incomplete** 🤯 

To fit the model better there should really be an opposite (*probably just internal*) `:not-expired` output. 
 
- by default, `:not-expired` is required
- if `:expired?` is optional, then so is `:not-expired?`
- if you manually expire a task that was not flagged in the graph as expire-able then it is incomplete because it did not complete its required `:not-expired` output

This would the same internals to be used for all outputs.

hjoliver:
See also https://github.com/cylc/cylc-flow/issues/6001#issuecomment-1996242121

oliver-sanders:
My branch does things a little differently as I've centralised some of the validation and requirement logic into the completion expression so I would rather superseed this branch.

oliver-sanders:
> completing the output makes the task incomplete

There are a couple of issues with the definition of completion as written, I have a cleaner solution that keeps the intended behaviour but fixes the edge cases. I've not put it up yet as I want to get a bit further on with validation and testing to make sure it's watertight.

cylc/flow/task_job_mgr.py
```diff
@@ -235,6 +235,8 @@ def prep_submit_task_jobs(self, workflow, itasks, check_syntax=True):
                 # state transition message reflects the correct submit_num
                 itask.submit_num += 1
                 itask.state_reset(TASK_STATUS_PREPARING)
+                # Avoid clock-expiring an active task:
+                itask.expire_time = None
```
oliver-sanders, line 239: I haven't handled clock-expiry of active tasks (I'm implementing the [opt outputs proposal](https://cylc.github.io/cylc-admin/proposal-optional-output-extension.html#proposal)).

Presumably if a task (submit-)fails and has [submission retries configured, then it should be allowed to expire between retries?

hjoliver:
> Sidenote, the opposites of expired are succeeded, failed and submit-failed 

I would add `submitted` to that, since we don't plan to allow active tasks to expire.

Then my suggestion, to use an internal `not-expired` output is logically equivalent as it (not-expired) would only be "completed" by job submission, But I'll wait to see what you've got.

cylc/flow/task_job_mgr.py
hjoliver: Yes, I think so.

oliver-sanders: Created an issue for the clock-expire of active tasks thing: https://github.com/cylc/cylc-flow/issues/6025

oliver-sanders:
> I would add submitted to that, since we don't plan to allow active tasks to expire.

It makes sense to disable (or postpone) clock-expires for active tasks. However, users can always set the "expired" output on a task.

hjoliver:
Closing as superseded by @oliver-sanders yet-to-posted optional outputs extension PR.

PR #6021, opened 2024-03-13
Fix terminal/log color issues

MetRonnie:
- `--color=never` was not always being respected
- The scheduler log would get a spurious color reset char (`ESC[0m`) at the start of the "DONE" line when not in no-detach mode

**Check List**

- [x] No dependency changes
- [x] Tests are included
- [x] No changelog entry as minor
- [x] No docs entry needed

### DIFF
--- a/cylc/flow/option_parsers.py
+++ b/cylc/flow/option_parsers.py
@@ -26,17 +26,17 @@ from optparse import (
 )
 import os
 import re
+import sys
+from textwrap import dedent
+from typing import Any, Dict, Iterable, Optional, List, Tuple
+
 from ansimarkup import (
     parse as cparse,
     strip as cstrip
 )
 
-import sys
-from textwrap import dedent
-from typing import Any, Dict, Iterable, Optional, List, Tuple
-
 from cylc.flow import LOG
-from cylc.flow.terminal import supports_color, DIM
+from cylc.flow.terminal import should_use_color, DIM
 import cylc.flow.flags
 from cylc.flow.loggingutil import (
     CylcLogFormatter,
@@ -196,7 +196,7 @@ class CylcOption(Option):
 
 
 class CylcHelpFormatter(IndentedHelpFormatter):
-    def _format(self, text):
+    def _format(self, text: str) -> str:
         """Format help (usage) text on the fly to handle coloring.
 
         Help is printed to the terminal before color initialization for general
@@ -208,16 +208,7 @@ class CylcHelpFormatter(IndentedHelpFormatter):
           - Strip any hardwired color tags
 
         """
-        if (
-            hasattr(self.parser.values, 'color')
-            and (
-                self.parser.values.color == "always"
-                or (
-                    self.parser.values.color == "auto"
-                    and supports_color()
-                )
-            )
-        ):
+        if should_use_color(self.parser.values):
             # Add color formatting to examples text.
             text = format_shell_examples(
                 format_help_headings(text)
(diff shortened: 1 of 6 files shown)


cylc/flow/terminal.py
```diff
@@ -257,17 +265,13 @@ def wrapper(*api_args: str) -> None:
...
-                            or (opts.color == 'auto' and supports_color())
-                        )
+                    use_color = should_use_color(opts)
                     wrapped_args = [parser, opts, *args]
                 if 'color' in inspect.signature(wrapped_function).parameters:
                     wrapped_kwargs['color'] = use_color
 
                 # configure Cylc to use colour
-                color_init(autoreset=True, strip=not use_color)
+                color_init(autoreset=False, strip=not use_color)
```
oliver-sanders, line 274: ?

cylc/flow/option_parsers.py
```diff
@@ -335,11 +326,11 @@ class CylcOptionParser(OptionParser):
...
                 ' of strings. E.G.'
-                ' "-z FOO=a,b,c" is shorthand for'
-                ' "-s FOO=[\'a\',\'b\',\'c\']".'
+                ' "-z FOO=a,b,c" is shorthand for'
+                ' "-s FOO=[\'a\',\'b\',\'c\']".'
                 ' Commas can be present in values if quoted, e.g.'
-                ' "-z FOO=a,\'b,c\'" is shorthand for'
-                ' "-s FOO=[\'a\',\'b,c\']".'
+                ' "-z FOO=a,\'b,c\'" is shorthand for'
+                ' "-s FOO=[\'a\',\'b,c\']".'
```
oliver-sanders, line 333: I think you haver replaced functional non-breaking spaces here. Please don't set up your editor to auto-convert.

MetRonnie: I did this manually assuming they were mistaken. But now I see why they were deliberate

cylc/flow/terminal.py
MetRonnie: The autoreset causes `ESC[0m` to be added at the beginning of the last line of log files when the scheduler is run daemonized. Don't know of how else to fix that

oliver-sanders: Right, so this is the change that hides the bug.

oliver-sanders: I'm not keen on this change (it protects users from us missing reset codes), but I also can't see an obvious way to avoid it, daemonization does strange things.

oliver-sanders (approved):

cylc/flow/terminal.py
MetRonnie: Hopefully if we stick to using `ansimarkup.parse` with the xml-style tags it should be easy to avoid accidentally not resetting colour

oliver-sanders: No, that does not protect us:

```python
>>> parse('<green>foo')
'\x1b[32mfoo'
```

MetRonnie: It doesn't guarantee anything, but it's obvious there should be a closing tag at least

hjoliver: I take it (from PR approval) that you've agreed with this approach now @oliver-sanders ?

oliver-sanders: Auto reset is a better approach, but I'm not going to hold this up on that point.

MetRonnie: Opened issue for re-enabling it: #6076

hjoliver (approved):

PR #6035, opened 2024-03-26
Fix seq xtriggers doc-698

dwsutherland:
Fixes spelling failure for https://github.com/cylc/cylc-doc/pull/698

one review will do...

**Check List**

### DIFF
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -811,7 +811,7 @@ with Conf(
             This setting can be overridden by the reserved keyword argument
             ``sequential`` in individual xtrigger declarations.
 
-            One sequential xtrigger on a parentless task with multiple
+            One sequential xtrigger on a parent-less task with multiple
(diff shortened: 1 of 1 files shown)


cylc/flow/cfgspec/workflow.py
```diff
@@ -811,7 +811,7 @@ def get_script_common_text(this: str, example: Optional[str] = None):
             This setting can be overridden by the reserved keyword argument
             ``sequential`` in individual xtrigger declarations.
 
-            One sequential xtrigger on a parentless task with multiple
+            One sequential xtrigger on a parent-less task with multiple
```
hjoliver, line 814: Are we doing this for the spell-checker?  If so, can we just add "parentless" to the dictionary?  I don't think "parent-less" is right.

dwsutherland: Yeah, just for spell checker .. we did this to the other one too .. ok will do

PR #6036, opened 2024-03-26
Fix.sim mode no increment

wxtim:
Fixes a bug not otherwise reported:

### Bug details

Using the following workflow

```
[scheduler]
  allow implicit tasks = true
[scheduling]
  [[graph]]
    R1 = foo
[runtime]
    [[foo]]
        [[[simulation]]]
            fail cycle points = all
            default run length = PT0S
            fail try 1 only = False

```

Open either TUI or GUI and see that the repeated tries to run the job are not displayed correctly:

image

The expected result is 
image

(Thanks to @MetRonnie for the bug report)

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request - N/A
- [x] This is a bugfix against a change introduced on master.

### DIFF
--- a/changes.d/6036.fix.md
+++ b/changes.d/6036.fix.md
@@ -1,0 +1,1 @@
+Fix bug where repeated submissions were not displaying crorrectly in TUI/GUI.
--- a/changes.d/6036.fix.md
+++ b/changes.d/6036.fix.md
@@ -1,1 +1,1 @@
-Fix bug where repeated submissions were not displaying crorrectly in TUI/GUI.
+Fixed bug in simulation mode where repeated submissions were not displaying correctly in TUI/GUI.
--- a/tests/integration/test_task_events_mgr.py
+++ b/tests/integration/test_task_events_mgr.py
@@ -94,7 +94,6 @@ async def test__insert_task_job(flow, one_conf, scheduler, start, validate):
 
         # Check that there are two entries with correct submit
         # numbers waiting for data-store insertion:
-        assert len(schd.data_store_mgr.added['jobs'].keys()) == 2
         assert [
             i.submit_num for i
             in schd.data_store_mgr.added['jobs'].values()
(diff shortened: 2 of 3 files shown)


changes.d/6036.fix.md
MetRonnie, line 1: Isn't this an unreleased bug so not appropriate to have a changelog entry?

tests/integration/test_task_events_mgr.py
```diff
@@ -64,3 +64,38 @@ async def test__reset_job_timers(
...
+        assert not schd.data_store_mgr.added['jobs'].keys()
+
+        # Insert task (twice):
+        schd.task_events_mgr._insert_task_job(itask, 'now', 1)
+        itask.submit_num += 1
+        schd.task_events_mgr._insert_task_job(itask, 'now', 1)
+
+        # Check that there are two entries with correct submit
+        # numbers waiting for data-store insertion:
+        assert len(schd.data_store_mgr.added['jobs'].keys()) == 2
```
MetRonnie, line 97: Not needed as the next assertion tests that there are 2 entries anyway
```suggestion
```

MetRonnie (commented):
Tested, works as expected 👍

changes.d/6036.fix.md
wxtim: Sort-of - you can reproduce something very similar at 8.2.x - it sticks at #01, rather than #00, but otherwise the same. 

However we're not fixing these things on 8.2.x.

Happy to remove the log if you prefer.

changes.d/6036.fix.md
```diff
@@ -0,0 +1 @@
+Fix bug where repeated submissions were not displaying crorrectly in TUI/GUI.
```
MetRonnie, line 1:
```suggestion
Fixed bug in simulation mode where repeated submissions were not displaying correctly in TUI/GUI.
```

MetRonnie (approved):

PR #6046, opened 2024-04-03
Optional outputs extension

### LINKED ISSUE #6025: clock-expire: prevent active tasks from being clock-expired
From discussion on https://github.com/cylc/cylc-flow/pull/6020

Clock expiry can complete a task.

If the task is already active then it should not be completed in this way.

Note, I think we would expect a clock-expired task to be able to expire between retries.

oliver-sanders:
### Summary
    
* Implement optional output extension proposal.
* Closes https://github.com/cylc/cylc-flow/issues/5640
* Closes #6025
* Add a user-configurable completion expression.
  * Automatically define a default when not specified.
* Change completion rules for the expire output.
  * This allows expired tasks to be retained in the pool as "incomplete".
* Expose the completion expression to the data store
  * Add the completion expression to the protobuf and GraphQL schemas.
  * Make the completion expression visible to the data store.
  * Display the completion status in "cylc show".

> [!NOTE]
> 
> As the result of recent terminology wrangling, the term "incomplete task" has now been dropped. I have **not** gone through the cylc-flow codebase to strip these references out (that's a separate task), but I have tried to avoid introducing any new references on this PR.

### Approach

The approach taken on this PR is to replace the current logic for evaluating task output completion with a single expression that returns True/False.

#### Implicit (i.e. derived) Expressions

This expression is automatically determined from the use of optional outputs in the graph.

Take the following workflow for example.

```ini
[scheduler]
    allow implicit tasks = True

[scheduling]
    [[graph]]
        R1 = """
            a:succeeded? => foo
            a:failed? => bar
            foo | bar => baz

            b:expired? => e
            b:succeeded? => s
            b:failed? =? f
            e | s | f => pub
        """
```

The derived completion expressions are exposed in the config, so we can inspect the derived completion expressions with `cylc config`:

```console
$ cylc config .
[scheduler]
    allow implicit tasks = True
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    final cycle point = 1
    [[graph]]
        R1 = """
            a:succeeded? => foo
            a:failed? => bar
            foo | bar => baz
            
            b:expired? => e
            b:succeeded? => s
            b:failed? =? f
            e | s | f => pub
        """
[runtime]
    [[root]]
    [[e]]
        completion = succeeded
    [[b]]
        completion = succeeded or failed or expired
    [[s]]
        completion = succeeded
    [[foo]]
        completion = succeeded
    [[a]]
        completion = succeeded or failed
    [[bar]]
        completion = succeeded
    [[baz]]
        completion = succeeded
    [[pub]]
        completion = succeeded
    [[f]]
        completion = succeeded
```

By default tasks must succeed. If failure, expiry or submission failure are permitted, these are added onto the end of the expression with an `or` operator.

The derived expressions should all match current behaviour with the exception of the handling of the expired state, which may now lead to incomplete outputs and potentially awkward edge cases with no required outputs which may not have been properly handled before.

#### Explicit (i.e. user-provided) expressions

The completion expression can be overridden for a task. If a completion expression is provided, then it must be logically consistent with the use of optional outputs in the graph.

E.G. Here the outputs `a:x`, `a:y` and `a:z` are marked as required in the graph:

```ini
[scheduler]
    allow implicit tasks = True

[scheduling]
    [[graph]]
        R1 = """
            a:x => x
            a:y => y
            a:z => z
        """

[runtime]
    [[a]]
        completion = succeeded and (x or y or z)
        [[[outputs]]]
            x = xxx
            y = yyy
            z = zzz
```

However, they are optional in the completion expression (because the completion expression can be satisfied without `a:x` we can deduce that `a:x` is optional), so this will result in an error:

```console
$ cylc val .
WorkflowConfigError: a:y is required in the graph, but optional in the completion expression
succeeded and (x or y or z)
```

To fix the error, change either the completion expression or the graph to match, e.g:

```diff
diff --git a/flow.cylc b/flow.cylc
index 12a7f70..5b91853 100644
--- a/flow.cylc
+++ b/flow.cylc
@@ -4,9 +4,9 @@
 [scheduling]
     [[graph]]
         R1 = """
-            a:x => x
-            a:y => y
-            a:z => z
+            a:x? => x
+            a:y? => y
+            a:z? => z
         """
 
 [runtime]
```

#### Compatibility Mode

Completion expressions are also used for compatibility mode, however, the logic remains as previously implemented:

```ini
[scheduling]
    [[dependencies]]
        [[[R1]]]
            graph = foo
```

```console
$ cylc config .
2024-04-03T14:00:52+01:00 WARNING - Backward compatibility mode ON
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    final cycle point = 1
    [[graph]]
        R1 = foo
[runtime]
    [[root]]
    [[foo]]
        completion = (submitted and succeeded)
```

### Visibility

Derived completion expressions are added to the config to make them visible via `cylc config`.

Completion expressions have also been added to the Protobuf and GraphQL schemas so they can be queried from the CLI, GUI and Tui.

I've built a simple text based reporting thinggy to help expose to users why a task's outputs are or aren't complete, here's how it looks in the logs:

Screenshot from 2024-04-03 14-05-33

And here's how it looks in cylc-show:

Screenshot from 2024-04-03 14-07-46

(note, switched from `+`/`-` to `✓`/`x` as some folks have been confused by `-` for some reason)

So the completion expression status is visible in the workflow logs (for final tasks with incomplete outputs), on the CLI (via `cylc show`) and in the Tui (via the `show` context option). This will become exposed in the GUI via the "Metadata View".

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc/pull/712) pull request opened if required at https://github.com/cylc/cylc-doc/pull/712


hjoliver:
> The derived completion expressions are exposed in the config, so we can inspect the derived completion expressions with cylc config

That is very cool 🎖️

cylc/flow/task_outputs.py
```diff
@@ -44,226 +67,305 @@
...
+            parts.append(f'({_part})')
+        else:
+            parts.append(_part)
+
+    # (2) handle optional success
+    if (
+        tdef.outputs[TASK_OUTPUT_SUCCEEDED][1] is False
+        or tdef.outputs[TASK_OUTPUT_FAILED][1] is False
+    ):
+        # failure is tolerated -> ensure the task succeess OR fails
```
hjoliver, line 136:
```suggestion
        # failure is tolerated -> ensure the task succeeds OR fails
```

cylc/flow/task_outputs.py
```diff
@@ -44,226 +67,305 @@
...
-
-    def _get_custom_triggers(self, required: bool = False) -> List[str]:
-        """Return list of all, or required, custom trigger messages."""
-        custom = [
-            out[1] for trg, out in self._by_trigger.items()
-            if trg not in SORT_ORDERS
-        ]
+    1. Create a completion expression that ensures all required ouputs are
+       completed.
+    2. If succeess is optional add "or succeeded or failed" onto the end.
```
hjoliver, line 105:
```suggestion
    2. If success is optional add "or succeeded or failed" onto the end.
```

cylc/flow/task_outputs.py
```diff
@@ -44,226 +67,305 @@
...
 
-    def set_completion(self, message, is_completed):
-        """Set output message completion status to is_completed (bool)."""
-        if message in self._by_message:
-            self._by_message[message][_IS_COMPLETED] = is_completed
+    """
+    # determine which triggers are used in the expression
+    used_compvars = get_variable_names(expression)
+
+    # all completion vairables which could appear in the expression
```
hjoliver, line 208:
```suggestion
    # all completion variables which could appear in the expression
```

cylc/flow/task_outputs.py
```diff
@@ -44,226 +67,305 @@
...
+        return self._message_to_trigger[message]
 
-        Args:
-            message (str): Output message.
+    def set_message_complete(self, message):
+        """Set the provided task message as complete."""
+        if (
+            # Note, this interface sometimes gets called erroneously, so we
+            # have to check that the message being set actually applies to
+            # these outputs before setting it
```
hjoliver, line 309:
```suggestion
            # these outputs before setting it.
```

cylc/flow/task_outputs.py
```diff
@@ -44,226 +67,305 @@
...
+            message in self._completed
+            and self._completed[message] is False
+        ):
+            self._completed[message] = True
+            return True
+        return False
+
+    def is_message_complete(self, message: str) -> Optional[bool]:
+        """Return True if this messages is complete.
```
hjoliver, line 318:
```suggestion
        """Return True if this message is complete.
```

hjoliver (commented):
Completed a first pass through the non-test code - it looks really good.

hjoliver:
```bash
$ cylc show ...
output completion: incomplete
  x |  succeeded
```

The output completion formatting here could easily be misinterpreted as saying *the incomplete condition is "output x OR succeeded"*, rather than *"output not completed: succeeded"*.

Maybe we could use a different character for the cross and/or for the separator?

hjoliver:
Had a good play with this today, nice.

tests/flakyfunctional/job-submission/19-chatty/reference.log
oliver-sanders, line 1: Note, because the change from `:submit-failed` to `submit-failed?` allows tasks to be removed from the pool more easily, I've added reference tests to ensure the downstream tasks (stopper in this case) are actually run as necessary.

oliver-sanders:
Tests now mostly there.

oliver-sanders:
Tests are all passing locally, but these two are kicking up a fuss on CI for as yet undetermined reasons:

* tests/f/spawn-on-demand/19-submitted-compat.t
* tests/f/optional-outputs/01-stall-on-incomplete.t

cylc/flow/cfgspec/workflow.py
```diff
@@ -996,6 +996,144 @@ def get_script_common_text(this: str, example: Optional[str] = None):
             can be explicitly configured to provide or override default
             settings for all tasks in the workflow.
         '''):
+            Conf('completion', VDR.V_STRING, desc='''
+                Define the condition for task output completion.
+
+                The completion condition is evaluated when a task attains
+                a final status - i.e. once it finished executing (``succeeded``
```
MetRonnie, line 1003: How about?
```suggestion
                The completion condition is evaluated when a task reaches
                a final state - i.e. once it finished executing (``succeeded``
```

cylc/flow/cfgspec/workflow.py
```diff
@@ -996,6 +996,144 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+                   foo => bar
+                   foo:x => x
+                   foo:y? => y
+
+                In Python syntax that condition looks like this:
+
+                .. code-block:: python
+
+                   # the task must succeed and generate the custom output "x"
+                   succeeded and x
```
MetRonnie, line 1036: I think this seems a bit confusing. How about just:
```suggestion
                The automatically generated completion condition for this is:

                .. code-block:: python

                   succeeded and x
```

cylc/flow/cfgspec/workflow.py
```diff
@@ -996,6 +996,144 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+
+                   For the completion expression, hyphens in task outputs
+                   are converted into underscores to allow
+                   evaluation by Python, e.g.:
+
+                   .. code-block:: cylc
+
+                      [runtime]
+                          [[foo]]
+                              completion = succeeded and my_output # underscore
```
MetRonnie, line 1074: If they are converted, then surely the user can write
```
completion = succeeded and my-output
```
?

Or are you saying that hyphens _must be replaced by_ underscores?

cylc/flow/cfgspec/workflow.py
```diff
@@ -996,6 +996,144 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+                              [[[outputs]]]
+                                  my-output = 'my custom task output' # hyphen
+
+                .. note::
+
+                   In some cases the succeeded output might not explicitly
```
MetRonnie, line 1080:
```suggestion
                   In some cases the ``succeeded`` output might not explicitly
```

cylc/flow/cfgspec/workflow.py
```diff
@@ -996,6 +996,144 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+
+                   In these cases succeess is presumed to be required unless
+                   explicitly stated otherwise, either in the graph:
+
+                   .. code-block:: cylc-graph
+
+                      foo?
+                      foo:x? => x
+
+                   Or in the completion expression:
```
MetRonnie, line 1095:
```suggestion
                   Or by omitting it in the completion expression:
```

cylc/flow/cfgspec/workflow.py
```diff
@@ -996,6 +996,144 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+                          [[graph]]
+                              R1 = """
+                                  # ERROR: this should be "a? => b"
+                                  a => b
+                              """
+                      [runtime]
+                          [[a]]
+                              # this completion condition implies that the
+                              # succeeded output is optional
+                              completion = succeeded or failed
```
MetRonnie, line 1119: Maybe put the runtime section above the graph for clarity

cylc/flow/config.py
```diff
@@ -1008,6 +1015,197 @@ def _check_sequence_bounds(self):
...
+            if expr:
+                # check the user-defined expression
+                self._check_completion_expression(name, expr)
+            else:
+                # derive a completion expression for this taskdef
+                expr = get_completion_expression(taskdef)
+
+            if name not in self.taskdefs:
+                # this is a family -> nothing more to do here
+                continue
```
MetRonnie, line 1038: Shouldn't this be the first thing in the loop?

cylc/flow/cfgspec/workflow.py
```diff
@@ -996,6 +996,144 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+
+                   # the task must succeed and generate at least one of the
+                   # outputs "x" or "y" or "z":
+                   succeeded and (x or y or z)
+
+                .. note::
+
+                   For the completion expression, hyphens in task outputs
+                   are converted into underscores to allow
+                   evaluation by Python, e.g.:
```
MetRonnie, line 1068: This suggests they are automatically converted, but this is not the case?

Did we discuss automatically replacing hyphens with underscores?

Suggestion for the time being at least:
```suggestion
                   For the completion expression, hyphens in task outputs
                   must be replaced with underscores to allow
                   evaluation by Python, e.g.:
```

cylc/flow/config.py
```diff
@@ -1008,6 +1015,197 @@ def _check_sequence_bounds(self):
...
+                graph_opt is True
+                and expr_opt is None
+                and compvar in {'submit_failed', 'expired'}
+            ):
+                raise WorkflowConfigError(
+                    f'{task_name}:{trigger} is permitted in the graph'
+                    ' but is not referenced in the completion'
+                    ' expression (so is not permitted by it).'
+                    f'\nTry: completion = "{expr} or {trigger}"'
+                )
```
MetRonnie, line 1196: I'm struggling to understand this

oliver-sanders: See the truth table above.

cylc/flow/cfgspec/workflow.py
oliver-sanders: We always put `[scheduling]` before `[runtime]`.

oliver-sanders:
@MetRonnie Don't we only use this label when we change a config rather than add one?

I'm not sure what the point of the "config change" label is? Why do we have it?

MetRonnie:
> I'm not sure what the point of the "config change" label is? Why do we have it?

I think originally to keep track of all the config changes from 7 to 8, I see no reason not to continue using it

cylc/flow/util.py
```diff
@@ -15,20 +15,44 @@
...
     Any,
+    Callable,
     List,
     Sequence,
+    Tuple,
 )
 
+# U+2713 (check)
+# U+2A2F (vector cross product)
+BOOL_SYMBOLS: List[str] = ['⨯', '✓']
```
MetRonnie, line 34: Considering how this is used
```suggestion
BOOL_SYMBOLS: Dict[bool, str] = {
    False: '⨯',  # U+2A2F (vector cross product)
    True: '✓'  # U+2713 (check)
}
```

oliver-sanders: In Python bools are integers, so what I;ve done is fine.

cylc/flow/task_outputs.py
```diff
@@ -44,15 +67,207 @@
...
+    """Turn a completion variable into a trigger.
+
+    Examples:
+        >>> completion_variable_to_trigger('succeeded')
+        'succeeded'
+        >>> completion_variable_to_trigger('submit_failed')
+        'submit-failed'
+
+    """
+    return compvar.replace('_', '-')
```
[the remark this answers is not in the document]
oliver-sanders: Then we can't reverse the process :(

This function is only used to format user exceptions for validate-time config issues. So it's non functional, but it would be good to give the error back to the user in the format they gave it in. If we can't easily get the trigger from the completion variable, then I'll strip this.

cylc/flow/config.py
oliver-sanders: > Can we not just do this automatically?

Yes, that is what this code is doing!

Note, not only must they be in the completion expression, but they must also be optional/required in the expression in order to match the graph.

MetRonnie: > Yes, that is what this code is doing!

This code looks like it's raising an error to me 😁

cylc/flow/util.py
MetRonnie: I think this way makes it clearer what the intent is (also gets rid of the mypy complaint when trying to do `BOOL_SYMBOLS[False]`?)

tests/unit/test_task_outputs.py
```diff
@@ -13,30 +13,297 @@
...
-        messages = list(self.TEST_MESSAGES)
-        for _ in range(5):
-            random.shuffle(messages)
-            output = sorted(messages, key=TaskOutputs.msg_sort_key)
-            self.assertEqual(output, self.TEST_MESSAGES, output)
+def tdef(required, optional, completion=None):
+    """Stuf a task definition.
+
+    Args:
+        required: Collection or required outputs.
```
MetRonnie, line 40:
```suggestion
    """Stub a task definition.

    Args:
        required: Collection of required outputs.
```

tests/unit/test_task_outputs.py
```diff
@@ -13,30 +13,297 @@
...
+
+    # this task does not have any required outputs (besides the implicitly
+    # required submitted/started outputs)
+    outputs = TaskOutputs(
+        tdef(
+            {TASK_OUTPUT_SUCCEEDED, 'x', 'y'},
+            {TASK_OUTPUT_FAILED},  # task may fail
+        )
+    )
+    assert set(outputs.iter_required_messages()) == set()
```
MetRonnie, line 274: It's confusing that this `tdef(required={'succeeded', 'x', 'y'}, ...` results in `succeeded`, `x` and `y` being required if `optional={'z'}`, but not being required if `optional={'failed'}`

cylc/flow/config.py
oliver-sanders: As per the spec - https://cylc.github.io/cylc-admin/proposal-optional-output-extension.html

Happy to explain further offline.

tests/unit/test_task_outputs.py
oliver-sanders: This is as per the spec: https://cylc.github.io/cylc-admin/proposal-optional-output-extension.html

If a task succeeds then it is expected to generate its required outputs, if it fails (and failure is permitted), then it is not.

MetRonnie: What does the graph for this look like? It seems to me that this doesn't make sense and it should be
```suggestion
    # this task does not have any required outputs (besides the implicitly
    # required submitted/started outputs)
    outputs = TaskOutputs(
        tdef(
            {},
            {TASK_OUTPUT_SUCCEEDED, 'x', 'y', TASK_OUTPUT_FAILED},  # task may fail
        )
    )
    assert set(outputs.iter_required_messages()) == set()
```

MetRonnie:
Example discussed in person:
```
[scheduling]
    cycling mode = integer
    runahead limit = P0
    [[graph]]
        R1 = """
            a? => b
            a:x => x
        """
[runtime]
    [[a]]
        completion = x and failed
        [[[outputs]]]
            x = x
```
validates without an error. But if you add `a:fail?` to the graph it gives
```
WorkflowConfigError: a:failed is optional in the graph (? symbol), but required in the completion expression:
x and failed
```

tests/unit/test_task_outputs.py
oliver-sanders: It doesn't, this is a unit test probing the `TaskOutputs` stuff directly to ensure consistency at that level. This exact example wouldn't actually get through validation (failure cannot be optional if success is required). If you replace `TASK_OUTPUT_FAILED` with `TASK_OUTPUT_EXPIRED` then validation won't complain and this could be defined in a graph along the lines of:

```
a
a:x
a:y
a:expired?
```

Purpose of the test is to shows that failure can bypass the required outputs if that's what the user wants.

hjoliver:
(I'll do a final(?) review and test of this Friday).

oliver-sanders:
I'll try and get Ronnies example to fail validation and address the review comments, we're running another training session at the moment so might take a couple of days.

cylc/flow/cfgspec/workflow.py
oliver-sanders: This suggestion is not correct as `succeed` may be present in the completion providing that it is optional.

tests/unit/test_task_outputs.py
oliver-sanders: Added a comment.

cylc/flow/cfgspec/workflow.py
MetRonnie: Ah right. The example given
```
completion = x  # no reference to succeeded here
```
still kind of implies that you must omit it. I'm trying to think if this could be worded differently

cylc/flow/cfgspec/workflow.py
MetRonnie: Maybe split it up, as I think seeing
```
# ERROR: this should be "a? => b"
a => b
```
*before* seeing why it is an error may be confusing.

Perhaps:
```suggestion
                      [scheduling]
                          [[graph]]
                              R1 = a => b
                      [runtime]
                          [[a]]
                              # this completion condition means that the
                              # succeeded output is optional
                              completion = succeeded or failed
                              
                   In the example above, the graph could be changed to `a? => b`
                   in order to comply with the completion statement.
```

MetRonnie (commented):
Recent changes looks good.

How easy would it be to tolerate present-tense versions of the task state outputs (`success`, `fail` etc.)?

cylc/flow/cfgspec/workflow.py
oliver-sanders: I don't see an issue with this.

oliver-sanders:
> How easy would it be to tolerate present-tense versions of the task state outputs (success, fail etc.)?

In completion expressions?

Possible, but we would need to run regex substitutions on each completion expression as we load it in the config, this would be sufficient:

`s/\bsuccess\b/succeeded/g`
`s/\bfail\b/failed/g`

Does make more sense to use the past tense in the completion expression though. Could potentially use validation to provide a more helpful error message for these cases?

cylc/flow/cfgspec/workflow.py
MetRonnie: I've updated the suggestion, let me know what you think

MetRonnie:
Seeing as we support the present tense elsewhere and it is always shorter to type, I see value in supporting it here

oliver-sanders:
I don't think that we should support the alternative names here, this muddies the water, makes expressions harder to validate and more expensive to evaluate. I'll make the validation failures a little clearer (suggesting the past tense canonicals).

cylc/flow/config.py
```diff
@@ -1008,6 +1016,218 @@ def _check_sequence_bounds(self):
...
+
+            for alt_qualifier, qualifier in ALT_QUALIFIERS.items():
+                if alt_qualifier in error:
+                    raise WorkflowConfigError(
+                        f'Error in [runtime][{task_name}]completion:'
+                        f'\n  {expr}'
+                        f'\nUse "{qualifier}" not "{alt_qualifier}" '
+                        'in completion expressions.'
+                    )
```
MetRonnie, line 1147: Not quite getting the right result from this
```
WorkflowConfigError: Error in [runtime][a]completion:
  succeeded and (x or submit_fail)
Use "submitted" not "submit" in completion expressions.
```
I think it could be moved to after here:
https://github.com/cylc/cylc-flow/blob/23567bc991111f9e178c3d13ca4112a5deb0ac2f/cylc/flow/task_outputs.py#L230-L231

oliver-sanders: Resolved this via a different approach to keep this in the validate code (also needed to convert qualifier to compvar).

cylc/flow/cfgspec/workflow.py
MetRonnie:
```suggestion
                   Or in the completion expression (either by omitting it or
                   including it with an `or` statement):
```

cylc/flow/cfgspec/workflow.py
oliver-sanders: Still prefer the original TBH but I'll elaborate on the comments a bit if you think it is necessary.

cylc/flow/cfgspec/workflow.py
```diff
@@ -996,6 +996,150 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+                      # "a" may either succeed or fail
+                      a? => b
+
+                .. rubric:: Examples
+
+                ``succeeded``
+                   The task must succeed.
+                ``succeeded or (failed and my_error)``
+                   The task can fail, but only if it also yields the custom
+                   output ``my_error``.
```
MetRonnie, line 1133: The output could be either `my_error` or `my-error` couldn't it?

MetRonnie (approved):
I think it's well explained in the reference docs now. Can't think of any more functional examples to test out

cylc/flow/cfgspec/workflow.py
oliver-sanders: Yes.

cylc/flow/cfgspec/workflow.py
```diff
@@ -996,6 +996,150 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+
+                E.g., in this example, the task ``foo`` must generate the
+                required outputs ``succeeded`` and ``x`` and it may or may not
+                generate the optional output ``y``:
+
+                .. code-block:: cylc-graph
+
+                   foo => bar
+                   foo:x => x
+                   foo:y? => y
```
hjoliver, line 1029: Possible follow-up tweak at worst, but it could be confusing for novices that upstream output and the downstream task look the same in these examples.

hjoliver (approved):
- [x] been through the code again, looks great
- [x] tested the hell out of it, works great

Nice work 👍

cylc/flow/config.py
```diff
@@ -1008,6 +1017,220 @@ def _check_sequence_bounds(self):
...
+            if expr:
+                # check the user-defined expression
+                self._check_completion_expression(name, expr)
+            else:
+                # derive a completion expression for this taskdef
+                expr = get_completion_expression(taskdef)
+
+            if name not in self.taskdefs:
+                # this is a family -> nothing more to do here
+                continue
```
hjoliver, line 1040:
```suggestion
```
The loop is over `self.taskdefs.items()` so this block is useless.

cylc/flow/task_proxy.py
```diff
@@ -578,32 +570,8 @@ def clock_expire(self) -> bool:
...
         """Return True if a final state achieved."""
-        return (
-            self.state(
-                TASK_STATUS_EXPIRED,
-                TASK_STATUS_SUBMIT_FAILED,
-                TASK_STATUS_FAILED,
-                TASK_STATUS_SUCCEEDED
-            )
-        )
+        return self.state(*TASK_STATUSES_FINAL)
```
hjoliver, line 573: Actually this method is no longer used - could remove it.

cylc/flow/config.py
```diff
@@ -1008,6 +1017,220 @@ def _check_sequence_bounds(self):
...
+                        f'\nUse "{_compvar}" not "{_alt_compvar}" '
+                        'in completion expressions.'
+                    )
+
+            raise WorkflowConfigError(
+                # NOTE: str(exc) == "name 'x' is not defined" tested in
+                # tests/integration/test_optional_outputs.py
+                f'Error in [runtime][{task_name}]completion:'
+                f'\nInput {error}'
+            )
```
hjoliver, line 1157: Hmmm this results in declared but undefined outputs being flagged as Inputs, which is likely to confuse users:
```ini
[scheduling]
    [[graph]]
        R1 = "foo:x"
[runtime]
    [[foo]]
        completion = succeeded or x
        #[[[outputs]]]
        #   x = x
```
```
 $ cylc val .
WorkflowConfigError: Error in [runtime][foo]completion:
Input 'x' is not defined
```

hjoliver:
Sorry Oliver, I came back again post-approval and found a couple of very minor issues. This is as good as done though, brilliant!

To get the coverage up a few percent, see also a side PR with some additional integration tests, for a handful of missed error conditions: 

https://github.com/oliver-sanders/cylc-flow/pull/69

cylc/flow/cfgspec/workflow.py
MetRonnie: Yeah actually I was going to comment, ideally I think the downstream tasks could be called e.g. `bar_x`, `bar_y` to avoid any implications

hjoliver: Wider than this PR; punted to https://github.com/cylc/cylc-flow/issues/6085

cylc/flow/task_proxy.py
hjoliver: (Done)

cylc/flow/config.py
hjoliver: (Done)

cylc/flow/config.py
hjoliver: (I'll just delete "Input" there for now).

hjoliver:
To get this show on the road, I deconflicted the branch and pushed several small commits for my trivial and uncontroversial review suggestions. Two tweaks to the CI environment were needed too. See Element chat. Will merge if tests pass.

hjoliver:
Good enough, just waiting on 2/5 macos func tests.

PR #6047, opened 2024-04-03
tutorial: make the tutorial workflow run a little slower

### LINKED ISSUE #6041: tutorial: sleep if not doing anything
### Problem

Tutorial tasks run too fast to see in TUI and GUI, especially for the scheduling tutorials.

### Proposed Solution

Add sleeps to tutorials to slow the running of early examples down.

### Alternative proposed solution

Wait for task-by-task run mode setting provided as part of the skip mode work and use simulation mode instead. In my opinion this is better for learners because `[TASK]run mode = simulation` makes it very explicit that the task isn't running anything real.

oliver-sanders:
* Address feedback that the workflow runs too quickly.
* Add 2-4s sleeps to slow it down.
* Closes #6041

**Check List**

### DIFF
--- a/changes.d/fix.5924.md
+++ b/changes.d/fix.5924.md
@@ -1,0 +1,1 @@
+Validation: a cycle offset can only appear on the right of a dependency if the task's cycling is defined elsewhere with no offset.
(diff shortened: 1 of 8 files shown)


MetRonnie:
Made my suggestions into a PR against this one: https://github.com/oliver-sanders/cylc-flow/pull/67

Also remembered people discovered an issue in the runtime configuration tutorial where part 4 "Test the `get_observations` tasks" wouldn't work because of missing runtime sections, added those in

MetRonnie (approved):

PR #6059, opened 2024-04-08
Clear all Cylc Rose options after Post-Install plugin

### LINKED ISSUE #6058: Changed behaviour in handling template variables passed via cli
### Description

I think that there has been a change in behaviour in handling template variables passed to workflows via the `-S` cli option in `v8.2.5`.
In `v8.2.3`, a template variable passed via `-S` would override that variable set in optional config files. In `v8.2.5`, it looks like this behaviour is reversed, and the setting in optional config will override the setting passed to `-S`.

### Reproducible Example

Example test suite containing;
```
test_workflow
├── flow.cylc
├── rose-suite.conf
├── opt
    └── rose-suite-common.conf
```
[flow.cylc]
```
#!jinja2
[scheduler]
   allow implicit tasks=True

[scheduling]
  initial cycle point = {{ START_CYCLE }}
  final cycle point = {{ END_CYCLE }}
  [[ graph]]
    R1 = cold_start
    T00 = """
      cold_start => a
    a => b
    """
```
[opt/rose-suite-common.conf]
```
[template variables]
START_CYCLE="20240101T00"
END_CYCLE="20240102T00"
```

Variables passed using `-S` are ignored, and the workflow runs with the settings from `opt/rose-suite-common.conf`;
```
$ cylc --version
8.2.5

$ cylc vip -O common -S 'START_CYCLE="20230102"' -S 'END_CYCLE="20230102"' .

$ ls ~/cylc-run/test_workflow/run1/log/job/
20240101T0000Z  20240102T0000Z
```

### Expected Behaviour
The setting passed via `-S` overrides other config
```
$ cylc --version
8.2.3

$ cylc vip -O common -S 'START_CYCLE="20230102"' -S 'END_CYCLE="20230102"' .

$ ls ~/cylc-run/test_workflow/run1/log/job/
20230102T0000Z
```

I seems that I get the behaviour that I'm looking for if I use `-s` instead of `-S`, so I could potentially use that as a workaround, but I don't have a good understanding of the difference between these two options. I note that `-s` will currently not allow variables to be updated on reload (#5965 / cylc/cylc-flow#6045), are there any other significant differences between these two options (i.e., `-s` vs `-S`)?

```
$ cylc --version
8.2.5

$ cylc vip -O common -s 'START_CYCLE="20230102"' -s 'END_CYCLE="20230102"' .

$ ls ~/cylc-run/test_workflow/run1/log/job/
20230102T0000Z
```

wxtim:
Closes #6058 , itself a side effect of #5996.

## Explanation

Cylc play ignores -S but not -O so -O overrides -S, but it shouldn't.

**Check List**

- [x] Tests Should be manually tested using examples in #6058 and #5996

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1263,10 +1263,15 @@ class Scheduler:
     def load_flow_file(self, is_reload=False):
         """Load, and log the workflow definition."""
         # Local workflow environment set therein.
-        # Allow -S and -D to take effect in Cylc VR.
+        # Allow -S and -D and -O to take effect in Cylc VR.
         # https://github.com/cylc/cylc-flow/issues/5968
+        # & https://github.com/cylc/cylc-flow/issues/6058
+        # TODO delete after 8.3.0 when this change will be included in
+        # Cylc Rose
         self.options.rose_template_vars = []
         self.options.defines = []
+        self.options.opt_conf_keys = []
+
         return WorkflowConfig(
             self.workflow,
             self.flow_file,


MetRonnie:
Do you have a minimal test example for this?

MetRonnie (approved):

cylc/flow/scheduler.py
```diff
@@ -1263,10 +1263,15 @@ def _configure_contact(self) -> None:
...
         # Local workflow environment set therein.
-        # Allow -S and -D to take effect in Cylc VR.
+        # Allow -S and -D and -O to take effect in Cylc VR.
         # https://github.com/cylc/cylc-flow/issues/5968
+        # & https://github.com/cylc/cylc-flow/issues/6058
+        # TODO delete after 8.3.0 when this change will be included in
+        # Cylc Rose
         self.options.rose_template_vars = []
         self.options.defines = []
+        self.options.opt_conf_keys = []
```
MetRonnie, line 1273: Why does this work?

PR #6062, opened 2024-04-10
Fix traceback in `cylc config -i`

### LINKED ISSUE #6010: `cylc config -i` traceback
### Description

Trying `cylc config --item <item>`  for an item that is not set in `flow.cylc` causes traceback.

### Reproducible Example

```
[scheduling]
    [[graph]]
        R1 = foo
[runtime]
    [[foo]]
        [[[simulation]]]
            fail cycle points = all
```
```
$ cylc config myworkflow -i '[runtime][foo]execution retry delays'
  File "~/cylc/flow/parsec/config.py", line 145, in get
    cfg = cfg[key]
  File "~/cylc/flow/parsec/OrderedDict.py", line 38, in __getitem__
    return OrderedDict.__getitem__(self, key)
KeyError: 'execution retry delays'

  File "~/.conda/envs/cylc8.1/bin/cylc", line 8, in <module>
    sys.exit(main())
  File "~/cylc/flow/scripts/cylc.py", line 699, in main
  File "~/cylc/flow/scripts/cylc.py", line 330, in execute_cmd
    entry_point.load()(*args)
  File "~/cylc/flow/terminal.py", line 276, in wrapper
  File "~/cylc/flow/scripts/config.py", line 158, in main
    asyncio.run(_main(parser, options, *ids))
  File "~/.conda/envs/cylc8.1/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "~/.conda/envs/cylc8.1/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "~/cylc/flow/scripts/config.py", line 212, in _main
    config.pcfg.idump(
  File "~/cylc/flow/parsec/config.py", line 177, in idump
    self.mdump(mkeys, sparse, prefix, oneline, none_str, handle=handle)
  File "~/cylc/flow/parsec/config.py", line 195, in mdump
    self.dump(keys, sparse, prefix, none_str, handle=handle)
  File "~/cylc/flow/parsec/config.py", line 202, in dump
    cfg = self.get(keys, sparse)
  File "~/cylc/flow/parsec/config.py", line 149, in get
    key in self.spec.get(*parents)
  File "~/cylc/flow/context_node.py", line 149, in get
    node = node[name]
  File "~/cylc/flow/context_node.py", line 134, in __getitem__
    return self._children.__getitem__(name)
KeyError: 'foo'
```

Note the second lot of traceback is particularly confusing as it seems to say `foo` is not in the config even though it is.

### Expected Behaviour

Either a short error message saying e.g. "item not set" or just an empty result like you get when using the `--defaults` option

MetRonnie:
Closes #6010

Fixes bug when trying to get a setting in a `__MANY__` section when the setting is valid but not set in the config.

**Check List**

- [x] No dependency changes 
- [x] Tests are included
- [x] No docs needed


cylc/flow/parsec/config.py
```diff
@@ -145,8 +145,11 @@ def get(self, keys: Optional[Iterable[str]] = None, sparse: bool = False):
                     cfg = cfg[key]
                 except (KeyError, TypeError):
                     if (
+                        # __MANY__ setting not present:
                         parents in self.manyparents or
-                        key in self.spec.get(*parents)
+                        # setting not present in __MANY__ section:
+                        key in self.get(parents, sparse=False)
+                        # (expanding cfg is ok because we raise at this point)
```
MetRonnie, line 152: Partially reverts e7bddf8abf91d4124287514c4b64828ba1f56c42 (from #4887); at the time I was concerned by the fact that we are expanding the config here, but it shouldn't matter because the train terminates here as we're raising an exception

MetRonnie:
I love Parsec... 😕

hjoliver:
> I love Parsec... 😕

Sorry 😬  (I wrote it years ago to fix deficiencies and slow performance of the 3rd party library we originally used, but it leaves much to be desired!)

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1994,26 +1994,34 @@ def platform_dump(
...
         """
-        platforms = '\n'.join(cfg.get(['platforms']).keys())
-        platform_groups = '\n'.join(cfg.get(['platform groups']).keys())
-        print(f'{PLATFORM_REGEX_TEXT}\n\nPlatforms\n---------', file=stderr)
-        print(platforms)
-        print('\n\nPlatform Groups\n--------------', file=stderr)
-        print(platform_groups)
+        platforms, platform_groups = None, None
+        with suppress(ItemNotFoundError):
+            platforms = '\n'.join(cfg.get(['platforms']).keys())
```
wxtim, line 2007: This should never happen because `localhost` is hardcoded into `cfgspec/globalconfig.py`.

Demonstrate by

```
CYLC_CONF_PATH=$HOME/intentionally_empty cylc config --platform-names
```

wxtim (approved):
Looks good. I've pointed to one check which I think may be unecessary, but I don't think it's a blocker. :)

cylc/flow/parsec/config.py
```diff
@@ -145,8 +145,11 @@ def get(self, keys: Optional[Iterable[str]] = None, sparse: bool = False):
                     cfg = cfg[key]
                 except (KeyError, TypeError):
                     if (
+                        # __MANY__ setting not present:
                         parents in self.manyparents or
-                        key in self.spec.get(*parents)
+                        # setting not present in __MANY__ section:
+                        key in copy(self).get(parents, sparse=False)
```
oliver-sanders, line 151: This copy is likely to be a heavy performance hit. Is it possible to check whether the key expects without copying the whole config?

cylc/flow/cfgspec/globalcfg.py
MetRonnie: Ok, I've force-pushed latest commit to simplify this

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1994,26 +1994,29 @@ def platform_dump(
...
     @staticmethod
     def dump_platform_names(cfg) -> None:
         """Print a list of defined platforms and groups.
         """
+        # Platforms is always defined with at least localhost
         platforms = '\n'.join(cfg.get(['platforms']).keys())
-        platform_groups = '\n'.join(cfg.get(['platform groups']).keys())
         print(f'{PLATFORM_REGEX_TEXT}\n\nPlatforms\n---------', file=stderr)
         print(platforms)
-        print('\n\nPlatform Groups\n--------------', file=stderr)
```
oliver-sanders, line 2011: FYI, this extra newline was probably present for compliance with ReST code style.

MetRonnie: Is this being used in the docs somewhere?

cylc/flow/parsec/config.py
MetRonnie: Still, I'm not 100% happy with it as there are cases where we supress/catch the error and continue.

I've come up with a different approach here, modifying `ContextNode.get()` instead: https://github.com/MetRonnie/cylc-flow/commit/4fc740d6abb80080014b5122912dec4721c1615f

MetRonnie:
@wxtim I've changed tack, please re-review. (Diff: https://github.com/cylc/cylc-flow/compare/a87150864721b67496f48c35c680d48f084f948a..f43126bee4a64730458391fb400e6edf8fc00409)

MetRonnie:
Bump @wxtim

cylc/flow/cfgspec/globalcfg.py
wxtim: It'd be pointless using it in the Cylc Docs, but I could imagine feeding it into metomi docs. I don't think that we do that at the moment, though.

wxtim (approved):

Issue #4778, opened 2022-03-28
cli: multi workflow commands


oliver-sanders:
Provide a nicer representation for multi-workflow commands.

Could do something Tui-style (but also colour coded) along the lines of:

```console
$ cylc stop one two tre fou
⊚ one
● two
⊗ tre
  Error: Workflow does not exist
```

Consider also #3329 which should allow commands to yield a stream of messages/updates to the client, providing updates on the requested action. The intention is to do something like this (interactive mode only):

```console
$ cylc stop one
(Ctrl+D to run in background)
● Connecting to workflow
⊚ Waiting for active tasks to complete
```

Combining the two:

```console
$ cylc stop one two tre
(Ctrl+D to run in background)
⊚ one
  ● Connecting to workflow
  ⊚ Waiting for active tasks to complete
● two
  ● Connecting to workflow
  ● Waiting for active tasks to complete
⊗ tre
  ⊗ Connecting to workflow
  Error: Workflow does not exist
```

We can't use Urwid for this as it needs to take control over the whole terminal, however, it should be easy enough to achieve using plain old ANSI (colorama allows us to move the cursor up and down, left and right). Otherwise can use the built-in curses.

**Pull requests welcome!**

PR #6064, opened 2024-04-11
Implement scheduler command method arg validation.

### LINKED ISSUE #6013: Scheduler command arg validation
Follow-up #5658 

The new `cylc set` command has relatively complicated validation requirements. During discussion there, @oliver-sanders suggested validating async scheduler commands server-side before queueing the command, instead of inside the queued command method itself. This enables validation feedback to the CLI and GUI.

I have a branch in progress, but it impacts other commands too, so it's better done as a quick-fire follow-up.

hjoliver:
Close #6013

Validate scheduler command method args before queuing the command, and print mutation response to the terminal:
  - error if method arg validation fails
  - or print command ID if command validated and submitted

Only applied to `cylc set` on this PR - do follow-up work for other commands.

**Check List**


cylc/flow/network/multi.py
```diff
@@ -107,4 +107,4 @@ def _report_single(report, workflow, result):
 
 
 def _report(_):
-    print('Command submitted; the scheduler will log any problems.')
+    pass
```
hjoliver, line 110: Is this client-side "reporting" needed for anything?

ColemanTom:
> Close #6064

I think you mean #6013 ? 6064 is this pull request.

hjoliver:
> > Close #6064
> 
> I think you mean #6013 ? 6064 is this pull request.

Yikes, thanks, good spotting ... corrected.  A pull request that closes itself might cause GitHub to self-destruct 🤯

cylc/flow/command_validation.py
```diff
@@ -0,0 +1,257 @@
...
+ERR_OPT_FLOW_WAIT = (
+    f"--wait is not compatible with --flow={FLOW_NEW} or --flow={FLOW_NONE}"
+)
+
+
+def validate(func: Callable):
+    """Decorate scheduler commands with a callable .validate attribute.
+
+    """
+    # TODO: properly handle "Callable has no attribute validate"?
```
oliver-sanders, line 43: Probs easier just to remove the type hint.

cylc/flow/network/multi.py
oliver-sanders: Not really, the multi-command reporting is more important due to the potential for some, but not all commands to fail.

hjoliver: I'll leave it there for now, in case we think of a use for it.

oliver-sanders: The plan was to do something fancy - https://github.com/cylc/cylc-flow/issues/4778

One day...

cylc/flow/network/multi.py
```diff
@@ -19,6 +19,30 @@
...
+        for _cmd, results in multi_result.items():
+            for result in results.values():
+                for wf_res in result:
+                    wf_id = wf_res["id"]
+                    response = wf_res["response"]
+                    if not response[0]:
+                        # Validation failure
+                        raise InputError(response[1])
+                    else:
+                        print(f"{wf_id}: command {response[1]} queued")
```
oliver-sanders, line 45: Should we be logging the command ID at the info level?

This internally generated string isn't really user facing, it's not useful information to them.

```suggestion
                        print(f"{wf_id}: command queued")
                        LOG.debug(f'command-id: {response[1]}')
```

cylc/flow/command_validation.py
```diff
@@ -0,0 +1,257 @@
...
+    Note the schema defaults flows to ["all"].
+
+    Examples:
+        Good:
+        >>> validate_flow_opts(["new"], False)
+        >>> validate_flow_opts(["1", "2"], False)
+        >>> validate_flow_opts(["1", "2"], True)
+
+        Bad:
+        >>> validate_flow_opts(["none", "1"], False)
```
oliver-sanders, line 62: Validation seems to work for `cylc trigger` but not `cylc set`?

```
$ cylc set foo//1/a --flow=none,1
~osanders/foo: command e1b50751-da49-4588-9fe3-91e3ad2e7bbd queued
$ cylc trigger foo//1/a --flow=none,1
InputError: Flow values must be an integer, or 'all', 'new', or 'none'
```

cylc/flow/command_validation.py
```diff
@@ -0,0 +1,257 @@
...
+        Traceback (most recent call last):
+        cylc.flow.exceptions.InputError: ...
+
+        # Error: invalid format:
+        >>> validate_prereqs(["1/foo::bar"])
+        Traceback (most recent call last):
+        cylc.flow.exceptions.InputError: ...
+
+        # Error: "all" must be used alone:
+        >>> validate_prereqs(["all", "2/foo:baz"])
```
oliver-sanders, line 119: Doesn't seem to be hooked into `cylc set`:

```
$ cylc set foo//1/a --pre=all,2/foo:baz
~osanders/foo: command 5b016bdf-7e26-4073-81b1-dda1c23c9e7c queued
```

oliver-sanders (changes requested):
I don't think these validators have been hooked into the CLI correctly, invalid `cylc set` commands seem to be getting through?

hjoliver:
Possibly now superseded by #6112

oliver-sanders:
If we're happy with the approach this can be superseded by #6112

PR #6065, opened 2024-04-12
clear cylc options at reload

### LINKED ISSUE #5968: vr: plugin options not respected when workflow is running

wxtim:
Closes #5968 _and_ #6058  (Which was an accidental by product of #5996 

I've tried to include as much info as possible here, to prevent reviewers having to re-hash the investigation or handle a web of issues and PRs for what turns out to be a simple problem with complex ramifications.

## Full statement of the problem.

The original problem (#5968 ) was that on reload CLI rose options (`-S`, `-D` and `-O`) are already stored on the scheduler, and because reload doesn't take those arguments they do not get updated. In Cylc VR the user can update the stored copy, but the reload cannot override the variables stored on the scheduler.

The first fix (#5996) cleared `-S` and `-D`, but not `-O` which led to inconsistencies on reload, and an apparent change in priority between variables set in different manner.

## Testing

This issue is a complex interaction between Cylc and the Cylc Rose plugin. Unfortunately the plugin is not as stand-alone as I would like. I think that there are 3 possible ways of testing this change:

1. (https://github.com/cylc/cylc-rose/pull/310) Functional and End-to-End testing in Cylc Rose. This allows me to demonstrate the actual cases where this is a problem, but is ugly because Cylc Rose does not currently have infrastructure equivalent to Cylc for creating schedulers. 
2. Functional testing in Cylc - Cylc has all the infrastructure to carry out these tests neatly, except for a requirement for Cylc Rose to be present. I chose not to because I didn't want to add that requirement, not even for testing.
3. Integration testing in Cylc Rose - Chose not to do this because I'd have to either copy a big pile of infrastructure from Cylc, which would then be something we would have to maintain, or spin the Cylc Test utilities into a new repo or do something strange and complex to make the tests available from Cylc flow if it's an editable install...
4. Unit testing in Cylc: I chose not to do this because it doesn't help document _why_ this change is being made, and ends up mocking so much of what's going on it seems nearly pointless.

**Check List**

- [x] Tests are included in a companion cylc rose PR https://github.com/cylc/cylc-rose/pull/310

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1262,11 +1262,17 @@ class Scheduler:
 
     def load_flow_file(self, is_reload=False):
         """Load, and log the workflow definition."""
-        # Local workflow environment set therein.
-        # Allow -S and -D to take effect in Cylc VR.
-        # https://github.com/cylc/cylc-flow/issues/5968
-        self.options.rose_template_vars = []
-        self.options.defines = []
+        if is_reload:
+            # If the workflow is reloaded clear any existing rose options
+            # The reload command doesn't have the ability to set these but
+            # if the user has used VR to re-install before reload they will
+            # expect the changed values not the ones stored on the scheduler.
+            # Note: Does NOT apply to reload on restart, because the play
+            # command used to restart can have template variables attached.
+            # See https://github.com/cylc/cylc-flow/pull/5996
+            self.options.rose_template_vars = []
+            self.options.defines = []
+            self.options.opt_conf_keys = []
         return WorkflowConfig(
             self.workflow,
             self.flow_file,


oliver-sanders:
@wxtim, tagged as 8.3.0 since this was raised on master.

wxtim:
> @wxtim, tagged as 8.3.0 since this was raised on master.

https://github.com/cylc/cylc-flow/issues/5968 was raised on 8.2.x

cylc/flow/scheduler.py
```diff
@@ -1262,11 +1262,17 @@ def _configure_contact(self) -> None:
 
     def load_flow_file(self, is_reload=False):
         """Load, and log the workflow definition."""
-        # Local workflow environment set therein.
-        # Allow -S and -D to take effect in Cylc VR.
-        # https://github.com/cylc/cylc-flow/issues/5968
-        self.options.rose_template_vars = []
-        self.options.defines = []
+        if is_reload:
```
oliver-sanders, line 1265: Do we also need to consider restarts here (note restart infers reload)?

cylc/flow/scheduler.py
```diff
@@ -1262,11 +1262,17 @@ def _configure_contact(self) -> None:
...
+            # If the workflow is reloaded clear any existing rose options
+            # The reload command doesn't have the ability to set these but
+            # if the user has used VR to re-install before reload they will
+            # expect the changed values not the ones stored on the scheduler.
+            # Note: Does NOT apply to reload on restart, because the play
+            # command used to restart can have template variables attached.
+            # See https://github.com/cylc/cylc-flow/pull/5996
```
oliver-sanders, line 1272: This explanation doesn't make much sense to me as the play command used to start the workflow in the first place can also have template variables attached?

oliver-sanders:
> so I think it should be against 8.2.x

Ok, you'll need to change the base branch to 8.2.x.

cylc/flow/scheduler.py
wxtim: If you do 

```
cylc play x --pause -S 'foo=1'
cylc stop x
cylc play x -S 'foo=2'
```
Do you want foo=1 or 2?

MetRonnie:
While you're here, I think it would be worth improving the description for `-S`. I think the current description is confusing?
```
  -S ROSE_TEMPLATE_VARS, --rose-template-variable=ROSE_TEMPLATE_VARS, --define-suite=ROSE_TEMPLATE_VARS
                        As `--define`, but with an implicit `[SECTION]` for
                        workflow variables.
```

cylc/flow/scheduler.py
```diff
@@ -1262,11 +1262,17 @@ def _configure_contact(self) -> None:
...
-        # https://github.com/cylc/cylc-flow/issues/5968
-        self.options.rose_template_vars = []
-        self.options.defines = []
+        if is_reload:
+            # If the workflow is reloaded clear any existing rose options
+            # The reload command doesn't have the ability to set these but
+            # if the user has used VR to re-install before reload they will
+            # expect the changed values not the ones stored on the scheduler.
+            # Note: Does NOT apply to reload on restart, because the play
+            # command used to restart can have template variables attached.
```
MetRonnie, line 1271: I'm a bit confused
- If the user has used `cylc vr running-workflow -S` (reinstall + reload), how are the changed values getting through to the scheduler then?
- What do you mean by "reload on restart"? If the user has used `cylc vr stopped-workflow -S` (reinstall + restart), then there is no reload, right?

wxtim: > If the user has used cylc vr running-workflow -S (reinstall + reload), how are the changed values getting through to the scheduler then?

The reinstall updates the `rose-suite-cylc-install.conf` which the reload reads. The reload doesn't have the `-S` option so clearing that here only affects options set on the scheduler by an earlier Cylc play.

Your question has however revealed a horrid case where this doesn't work:

```
cylc vip -S 'VAR=1' --pause
cylc vr bugs/5968
```

Do you think that the correct options might be to allow Cylc Reload to set `-S` and it's friends, becuase to me that looks like the simplest way of doing things?


> What do you mean by "reload on restart"? If the user has used cylc vr stopped-workflow -S (reinstall + restart), then there is no reload, right?

Oliver described it as an implicit reload.

MetRonnie: > Your question has however revealed a horrid case where this doesn't work

What does happen with this example, and what is meant to happen?

wxtim:
After much thought, and an offline discussion with @oliver-sanders I'm closing this: We now think that the correct solution to this problem is rather different.

PR #6068, opened 2024-04-16
Play should have no cylc rose options

wxtim:
& Reload should not mess with Cylc Rose CLI options.

Companion to https://github.com/cylc/cylc-rose/pull/312

Cylc Play was ([according to the proposal doc](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-cylc-rose-installing-rose-configs.md)) never intended to take Rose config CLI options (`-O`, `-D` and `-S`).

For example:

```
cylc install -S 'foo=1'
cylc play -S 'foo=2' --pause
```
Would lead to foo = 2 being stored on the scheduler as `self.options.rose_template_vars`: If we then do

```
cylc reinstall -S 'foo=3'
cylc reload  

# or

cylc vr -S 'foo=3'
```

Will lead to the installed files updating, but reload cannot update the `self.options.rose_template_vars` object created by Cylc Play, so we cannot get rid of `foo=2`.

Without the scheduler storing these items, reload can examine all the files and get the updated value.

**Check List**

- [x] Tests not included:
  - [x] Cylc play having rose options was never tested, or intended to work.
  - [x] Tests for the removal of the Cylc-Rose config items were only manual when they were added.

### DIFF
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -309,7 +309,7 @@ def get_option_parser(add_std_opts: bool = False) -> COP:
         argdoc=[WORKFLOW_ID_ARG_DOC]
     )
 
-    options = parser.get_cylc_rose_options() + PLAY_OPTIONS
+    options = PLAY_OPTIONS
     for option in options:
         if isinstance(option, OptionSettings):
             parser.add_option(*option.args, **option.kwargs)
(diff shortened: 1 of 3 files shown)


cylc/flow/scheduler_cli.py
```diff
@@ -309,12 +309,8 @@ def get_option_parser(add_std_opts: bool = False) -> COP:
         argdoc=[WORKFLOW_ID_ARG_DOC]
     )
 
-    options = parser.get_cylc_rose_options() + PLAY_OPTIONS
-    for option in options:
-        if isinstance(option, OptionSettings):
```
wxtim, line 314: There are no longer any options stored as Dicts rather than OptionSettings objects :. this is redundant.

oliver-sanders:
(breaking change entry I'm afraid)

oliver-sanders (approved):

MetRonnie (approved):

PR #6070, opened 2024-04-17
Delete public DB if private DB does not exist on scheduler start

MetRonnie:
On scheduler start, it tries to load workflow template variables from the public database. However, if the user has deleted the private database, it is not a restart, so it shouldn't try to load the public DB either.

The simple solution to this is to delete the public DB file (if it exists) immediately after checking and finding the private DB does not exist.

**Check List**

- [x] No dependency changes 
- [x] No changelog entry needed as minor/niche
- [x] No docs needed


hjoliver (approved):
Makes sense to me. Maybe add a comment to the code to explain it, as per your issue description.

oliver-sanders (approved):

PR #6071, opened 2024-04-17
Include xtrigger function signatures in `cylc config`

### LINKED ISSUE #5877: cylc config shows unhelpful information for xtriggers
### Description

If you have an xtrigger and run `cylc config my_workflow` the xtrigger is listed as `... = [xtrigger-func cmd]` which isn't very informative or helpful.

### Reproducible Example

I just used the test case from [the xtrigger plugin PR](https://github.com/cylc/cylc-flow/pull/5831).

Results was

```
$ cylc config cylctb-20231206T033820Z-uOHK/f/xtriggers/04-respect-cylc-pythonpath
[scheduler]
    cycle point format = %Y
[scheduling]
    initial cycle point = 2010
    final cycle point = 2011
    [[xtriggers]]
        x1 = [xtrigger-func cmd]
    [[graph]]
```

### Expected Behaviour

A resolved string of what the xtrigger was for reference after templating is resolved.

MetRonnie:
Closes #5877

**Check List**

- [x] No dependency changes 
- [x] No docs needed

### DIFF
--- a/cylc/flow/subprocctx.py
+++ b/cylc/flow/subprocctx.py
@@ -158,3 +158,17 @@ class SubFuncContext(SubProcContext):
         args = self.func_args + [
             "%s=%s" % (i, self.func_kwargs[i]) for i in skeys]
         return "%s(%s)" % (self.func_name, ", ".join([str(a) for a in args]))
+
+    def dump(self) -> str:
+        """Output for logging."""
+        return SubProcContext.__str__(self)
+
+    def __str__(self) -> str:
+        """
+        >>> str(SubFuncContext('label', 'my_func', [1, 2], {'a': 3}))
+        'my_func(1, 2, a=3):10.0'
+        """
+        return f"{self.get_signature()}:{self.intvl}"
+
+    def __repr__(self) -> str:
+        return f"<{type(self).__name__} {self}>"
--- a/changes.d/6071.fix.md
+++ b/changes.d/6071.fix.md
@@ -1,0 +1,1 @@
+`cylc config` now shows xtrigger function signatures.
--- a/cylc/flow/xtrigger_mgr.py
+++ b/cylc/flow/xtrigger_mgr.py
@@ -502,7 +502,7 @@ class XtriggerManager:
         Raises:
             ValueError: if the context given is not active
         """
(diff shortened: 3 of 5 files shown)


wxtim (commented):
- [x] Makes sense
- [x] Manually tested as working

No test?

cylc/flow/subprocctx.py
```diff
@@ -158,3 +158,10 @@ def get_signature(self):
         args = self.func_args + [
             "%s=%s" % (i, self.func_kwargs[i]) for i in skeys]
         return "%s(%s)" % (self.func_name, ", ".join([str(a) for a in args]))
+
+    def __str__(self) -> str:
+        """
+        >>> str(SubFuncContext('label', 'func', [1, 2], {'a': 3}))
+        '[xtrigger-func cmd]: func(1, 2, a=3)'
+        """
+        return f"{SubProcContext.__str__(self)}: {self.get_signature()}"
```
hjoliver, line 167: Do we need the `[xtrigger-func cmd]` prefix?
```suggestion
        return self.get_signature()
```

MetRonnie: Dunno 😄

cylc/flow/subprocctx.py
```diff
@@ -158,3 +158,17 @@ def get_signature(self):
...
+    def dump(self) -> str:
+        """Output for logging."""
+        return SubProcContext.__str__(self)
+
+    def __str__(self) -> str:
+        """
+        >>> str(SubFuncContext('label', 'my_func', [1, 2], {'a': 3}))
+        'my_func(1, 2, a=3):10.0'
+        """
+        return f"{self.get_signature()}:{self.intvl}"
```
MetRonnie, line 171: The previous `__str__()` implementation was removed in https://github.com/cylc/cylc-flow/pull/5617/files#r1254330367, but I've ensured the output from xtriggers still gets logged in the workflow log by adding a `dump` method and using that instead

wxtim (approved):

cylc/flow/subprocctx.py
oliver-sanders: That was it! Great.

PR #6072, opened 2024-04-17
cylc.nanorc

wxtim:
Add a limited amount of Cylc syntax highlighting for Nano - I wrote this a while back, because Nano is my in terminal editor of choice, and handy for smaller toy workflows.

image

**Check List**

- [x] Tests are not included: This is provided as is, on the basis of being better than nothing...
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/718

### DIFF
--- a/changes.d/6072.feat.md
+++ b/changes.d/6072.feat.md
@@ -1,0 +1,1 @@
+Nano Syntax Highlighting now available. See
--- a/changes.d/6072.feat.md
+++ b/changes.d/6072.feat.md
@@ -1,1 +1,1 @@
-Nano Syntax Highlighting now available. See
+Nano Syntax Highlighting now available.
--- a/cylc/flow/etc/syntax/cylc.nanorc
+++ b/cylc/flow/etc/syntax/cylc.nanorc
@@ -1,0 +1,30 @@
+# How to use this file:
+# cylc get-resources cylc.nanorc ~/.config/nano/cylc.nanorc
+# Add the following to ~/.nanorc
+# include ~/.config/nano/cylc.nanorc
+
+# Supports `.cylc` files
+syntax "Cylc" "\.cylc$"
+
+## Multiline
+color yellow start="\"\"\"" end="\"\"\""
+color yellow start="'''" end="'''"
+color yellow start="\[" end="\]"
+
+## Values
+color yellow "=(.*)$"
+color green "=[^>]"
+color brightmagenta "=>|&|\|\\"
+
+## Valid headings
+color green "^\[.*\]$"
+color green "^\s{4}\[\[.*\]\]+$"
+color green "^\s{8}\[\[\[.*\]\]\]+$"
+
+## Comments (keep at the end of this file!)
+color cyan "#.*$"
+
+## Jinja2
+icolor brightcyan "^#!Jinja2"
+color brightcyan "\{%.*%\}"
+color brightcyan "\{\{.*\}\}"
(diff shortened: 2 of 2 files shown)


cylc/flow/etc/syntax/cylc.nanorc
```diff
@@ -0,0 +1,30 @@
...
+## Valid headings
+color green "^\[.*\]$"
+color green "^\s{4}\[\[.*\]\]+$"
+color green "^\s{8}\[\[\[.*\]\]\]+$"
+
+## Comments (keep at the end of this file!)
+color cyan "#.*$"
+
+## Jinja2
+icolor red "^#!Jinja2"
```
MetRonnie, line 28: What is `icolor` as opposed to `color`?

Also, not sure red is a good choice as it tends to mean error/problem

MetRonnie:
Btw, enhancement so should be against 8.3.0?

cylc/flow/etc/syntax/cylc.nanorc
wxtim: `icolor` is case insensitive regex so it will highlight `#!Jinja2` or `#!jinja2` or even `#!jInJa2` (which is valid!).

The choice of colors is somewhat limited on the older version of Nano we have at the MetOffice, although newer versions of nano offer more choice - when I try I get:

```
Valid colors are "green", "red", "blue",
"white", "yellow", "cyan", "magenta" and
"black", with the optional prefix "bright"
for foreground colors.

```

(see also https://www.nano-editor.org/dist/latest/nanorc.5.html)

wxtim:
> Btw, enhancement so should be against master / 8.3.0?

It's an enhancement,  but it's very small with no knock on consequences or conceivable side effects and I don't see why it shouldn't go in the earliest possible release.

MetRonnie (commented):
Nice 👍 Approval pending changelog entry. Also will need adding to cylc-doc

changes.d/6072.feat.md
```diff
@@ -0,0 +1 @@
+Nano Syntax Highlighting now available. See
```
MetRonnie, line 1: See?

wxtim:
```suggestion
Nano Syntax Highlighting now available.
```

cylc/flow/etc/syntax/cylc.nanorc
```diff
@@ -0,0 +1,30 @@
...
+
+## Values
+color yellow "=(.*)$"
+color green "=[^>]"
+color brightmagenta "=>|&|\|\\"
+
+## Valid headings
+color green "^\[.*\]$"
+color green "^\s{4}\[\[.*\]\]+$"
+color green "^\s{8}\[\[\[.*\]\]\]+$"
```
MetRonnie, line 22: Amount of whitespace should be irrelevant for a syntax highlighter
```suggestion
color green "^\s*\[.*\]"
color green "^\s*\[\[.*\]\]"
color green "^\s*\[\[\[.*\]\]\]"
```

cylc/flow/etc/syntax/cylc.nanorc
```diff
@@ -0,0 +1,30 @@
...
+syntax "Cylc" "\.cylc$"
+
+## Multiline
+color yellow start="\"\"\"" end="\"\"\""
+color yellow start="'''" end="'''"
+color yellow start="\[" end="\]"
+
+## Values
+color yellow "=(.*)$"
+color green "=[^>]"
```
MetRonnie, line 16: What's this one about?

wxtim: `=` signs that aren't `=>` so we don't include graph deps.

MetRonnie (approved):

PR #6081, opened 2024-04-23
Fix job runner stdout missing newline

jennan:
Kia ora,

While implementing a new job runner, I realised that the if the job runner `submit` method returns its stdout text without a newline, then a workflow with more than one task will fail. After further investigations with @hjoliver , we discovered that this is due to text handling in `JobRunnerManager.jobs_submit`, that doesn't properly add a newline to the text message if it is missing. This PR implements a fix.

You can reproduce the bug by removing the newline of the second output of the background job runner submit method (see https://github.com/cylc/cylc-flow/blob/master/cylc/flow/job_runner_handlers/background.py#L92) and try to run a workflow with at least 2 tasks.

I am not sure how to turn this into a test to reproduce the bug. Any guidance on this would be welcome 🙂.

----

[UPDATE (HO)] Short explanation: without this fix, if a job-runner submit method returns job-ID without a newline, it causes the `cylc jobs-submit` stdout to be misformatted (two lines concatenated) such that the scheduler can't parse it correctly - for one out of the multiple jobs submitted.

----

**Check List**

- [x] No dependency changes 
- [x] Changelog fragment included if this is a change that can affect users
- [x] No cylc-doc PR needed

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -57,3 +57,4 @@ Utheri Wagura              <36386988+uwagura@users.noreply.github.com>
 github-actions[bot]        <github-actions@noreply.github.com>                      <41898282+github-actions[bot]@users.noreply.github.com>
 github-actions[bot]        <github-actions@noreply.github.com>  GitHub Action
 Diquan Jabbour             <165976689+Diquan-BOM@users.noreply.github.com>
+Maxime Rio                 <maxime.rio@niwa.co.nz>
--- a/tests/integration/test_job_runner_mgr.py
+++ b/tests/integration/test_job_runner_mgr.py
@@ -1,0 +1,85 @@
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import errno
+import logging
+from pathlib import Path
+import re
+from textwrap import dedent
+
+from cylc.flow.job_runner_mgr import JobRunnerManager
+from cylc.flow.pathutil import get_workflow_run_job_dir
+from cylc.flow.task_state import TASK_STATUS_RUNNING
+from cylc.flow.subprocctx import SubProcContext
+
+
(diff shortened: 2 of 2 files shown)


jennan:
Hi, I have developed this on `master` and made the PR against branch 8.2.x without realising that it would make a huge diff. Sorry for that. Please could you edit the PR to be against `master` or would you prefer that I reopen a new one but started from branch 8.2.x?

hjoliver:
Thanks @jennan !

@oliver-sanders (UK team) noted overnight that he ran into this some time ago, but wasn't sure it was a bug so added a note on the need for newline in the handler API documentation: https://github.com/cylc/cylc-flow/blob/master/cylc/flow/job_runner_handlers/documentation.py#L426-L428

However, now that you've tracked down the offending code, by inspection (and also by comparison it with similar in `jobs_kill(...)`, it is definitely a bug.

Most of the current job-runners don't provide an explicit `submit` method, they just use the command template as a shortcut. However, the background handler does, and it deliberately adds a newline to its return value. https://github.com/cylc/cylc-flow/blob/master/cylc/flow/job_runner_handlers/background.py#L92

@jennan - can you tweak this PR to:
- not add the newline in the background handler (above), as it's not necessary now
- and remove the newline advice in the docs (above)

hjoliver:
Regarding testing: without this fix, if a job-runner submit method returns job-ID without a newline, it causes the `cylc jobs-submit` stdout to be misformatted (two lines concatenated) such that the scheduler can't parse it correctly - for one out of the multiple jobs submitted.

So I think it would be sufficient to refactor the job runner manger `submit` method to use a helper method that formats and returns the final string to be written to stdout, instead of writing incrementally to stdout as it does now. That helper method could then easily be unit tested with an inline doctest.

hjoliver:
@jennan - no cylc-doc PR is needed for this. It could do with a change log entry though, just in case anyone else is working on job runners at the moment (unlikely but not impossible!). Just add a file `changes.d/6081.fix.md` with a short description (see other md files in there for example).

cylc/flow/job_runner_handlers/background.py
```diff
@@ -89,7 +89,7 @@ def submit(cls, job_file_path, submit_opts):
                 exc.filename = "nohup"
             return (1, None, str(exc))
         else:
-            return (0, "%d\n" % (proc.pid), None)
+            return (0, str(proc.pid), None)
```
hjoliver, line 92: (This was compensating for the bug, pre-fix).

cylc/flow/job_runner_handlers/documentation.py
```diff
@@ -424,8 +424,7 @@ def submit(
             ret_code:
                 Subprocess return code.
             out:
-                Subprocess standard output, note this should be newline
-                terminated.
+                Subprocess standard output.
```
hjoliver, line 427: (We now add the newline automatically)

hjoliver:
@oliver-sanders - I've finished this off, and simplified the affected code a bit.

Not sure how to go about testing it.

oliver-sanders:
> Not sure how to go about testing it.

Have had a go: https://github.com/hjoliver/cylc-flow/pull/59

hjoliver (approved):

hjoliver:
Merged and pushed your new test @oliver-sanders -  this is good to go now.

PR #6086, opened 2024-04-29
Bump pytest-asyncio version

wxtim:
Don't use Python 3.7 on MacOs on CI.

Exclude non-function releases of pytest-asyncio

**Check List**

- [x] This is a small dependency change PR

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -117,7 +117,7 @@ tests =
     flake8>=3.0.0
     mypy>=0.910,<1.9
     # https://github.com/pytest-dev/pytest-asyncio/issues/706
-    pytest-asyncio>=0.17,!=0.23.*
+    pytest-asyncio>=0.17,!=0.23.0,!=0.23.1
     pytest-cov>=2.8.0
     pytest-xdist>=2
     pytest-env>=0.6.2
--- a/setup.cfg
+++ b/setup.cfg
@@ -117,7 +117,7 @@ tests =
     flake8>=3.0.0
     mypy>=0.910,<1.9
     # https://github.com/pytest-dev/pytest-asyncio/issues/706
-    pytest-asyncio>=0.17,!=0.23.0,!=0.23.1
+    pytest-asyncio>=0.21.2,!=0.23.*
     pytest-cov>=2.8.0
     pytest-xdist>=2
     pytest-env>=0.6.2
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -122,7 +122,7 @@ tests =
     pytest-xdist>=2
     pytest-env>=0.6.2
     pytest-mock>=3.7
-    pytest>=6
+    pytest>=6,!=8.2.*
```
MetRonnie, line 125:
```
pytest-asyncio>=0.21.2,!=0.23.*
...
pytest>=6
```

wxtim:
```suggestion
    pytest>=6
```

wxtim:
> No longer needed? - [pytest-dev/pytest-asyncio#706 (comment)](https://github.com/pytest-dev/pytest-asyncio/issues/706#issuecomment-2082788963)

Entirely about bumping the macos version now. :)

setup.cfg
```diff
@@ -117,7 +117,7 @@ tests =
     flake8>=3.0.0
     mypy>=0.910,<1.9
     # https://github.com/pytest-dev/pytest-asyncio/issues/706
-    pytest-asyncio>=0.17,!=0.23.*
+    pytest-asyncio>=0.17,!=0.23.0,!=0.23.1
```
MetRonnie, line 120:
```suggestion
    pytest-asyncio>=0.23.2,!=0.23.0
```

MetRonnie (commented):
pytest-asyncio pin is incorrect I think

setup.cfg
wxtim: >=0.23.3
are ok

MetRonnie: No they aren't: https://github.com/pytest-dev/pytest-asyncio/issues/706

MetRonnie (approved):
Oh, it turns out Python 3.7 is still supported by macos-11 runners, but not macos-13 and above (which only the build workflow uses). Anyway, not too important

hjoliver:
> Oh, it turns out Python 3.7 is still supported by macos-11 runners, but not macos-13 and above (which only the build workflow uses). Anyway, not too important

Yeah in my initial quick hot-fix I only had to tweak this for the build workflow. I mentioned that at at the time, but maybe it got lost in the noise.

PR #6095, opened 2024-05-08
Add "valid for" states to schema def for `cylc set`

MetRonnie:
Follow-up to #5658

This is needed by the UI to disable mutations that are not valid for the workflow state. At present, "set" is not disabled for stopped workflows as it should be:

image

**Check List**

- [x] No dependency changes 
- [x] Tests are included 
- [x] No docs PR needed

### DIFF
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -2103,6 +2103,8 @@ class SetPrereqsAndOutputs(Mutation, TaskMutation):
              - ``started`` implies ``submitted``.
              - ``succeeded`` and ``failed`` imply ``started``.
              - custom outputs and ``expired`` do not imply any other outputs.
+
+            Valid for: paused, running, stopping workflows.
         """)
         resolver = partial(mutator, command='set')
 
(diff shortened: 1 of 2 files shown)


oliver-sanders:
Dammit, good spot! We should probably have a test to assert this.

cylc/flow/network/schema.py
```diff
@@ -2103,6 +2103,8 @@ class Meta:
              - ``started`` implies ``submitted``.
              - ``succeeded`` and ``failed`` imply ``started``.
              - custom outputs and ``expired`` do not imply any other outputs.
+
+            Valid for: paused, running, stopping workflows.
```
oliver-sanders, line 2107: I don't think this command is really valid for stopping workflows:

```suggestion
            Valid for: paused, running.
```

MetRonnie: If waiting for tasks to finish while stopping, is it not valid for users to set task states? I think it can be done from the CLI

oliver-sanders: Ooh, there is a reasonable point there due to this use case:

>  I need to orphan a "stuck" job submission
>
> -- https://github.com/cylc/cylc-admin/blob/master/docs/proposal-interventions.md#4-i-need-to-orphan-a-stuck-job-submission

Which could otherwise block shutdown.

Ok, leave as is.

oliver-sanders (approved):

PR #6101, opened 2024-05-14
tests: reset global state before each test

oliver-sanders:
Should help to resolve possible issues where one test sets a flag (but not via monkeypatch) and another accesses that value.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -27,6 +27,13 @@ from cylc.flow.parsec.config import ParsecConfig
 from cylc.flow.parsec.validate import cylc_config_validate
 
 
+@pytest.fixture(autouse=True)
+def test_reset(monkeypatch):
+    """Reset global state before all tests."""
+    monkeypatch.setattr('cylc.flow.flags.verbosity', 0)
+    monkeypatch.setattr('cylc.flow.flags.cylc7_back_compat', False)
+
+
 @pytest.fixture(scope='module')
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -27,6 +27,13 @@
 from cylc.flow.parsec.validate import cylc_config_validate
 
 
+@pytest.fixture(autouse=True)
+def test_reset(monkeypatch):
+    """Reset global state before all tests."""
+    monkeypatch.setattr('cylc.flow.flags.verbosity', 0)
+    monkeypatch.setattr('cylc.flow.flags.cylc7_back_compat', False)
```
MetRonnie, line 34: What about setting the flags directly?

tests/conftest.py
```diff
@@ -27,6 +27,13 @@
 from cylc.flow.parsec.validate import cylc_config_validate
```
oliver-sanders, line 27:
```suggestion
from cylc.flow.parsec.validate import cylc_config_validate
from cylc.flow import flags
```

tests/conftest.py
oliver-sanders, line 34:
```suggestion
def test_reset():
    """Reset global state before all tests."""
    flags.verbosity = 0
    flags.cylc7_back_compat = False
```

MetRonnie (approved):
This seems to do the trick

PR #6103, opened 2024-05-14
data store: add absolute graph edges

### LINKED ISSUE #5845: tdef: graph parents - absolute dependencies ignored
Absolute dependencies don't appear in the graph.

#### Example:

```ini
[scheduler]
    allow implicit tasks = True

[scheduling]
    initial cycle point = 1
    cycling mode = integer
    [[graph]]
        R1 = build
        P1 = build[^] => run
```

The dependency is drawn in the first cycle, but not in the subsequent ones:

Screenshot from 2023-11-23 14-24-50

But the scheduler does know about the dependency:

```console
$ cylc show icp//2/run
title: (not given)
description: (not given)
URL: (not given)
state: waiting
prerequisites: ('-': not satisfied)
  + 1/build succeeded
outputs: ('-': not completed)
  - 2/run expired
  - 2/run submitted
  - 2/run submit-failed
  - 2/run started
  - 2/run succeeded
  - 2/run failed
```

oliver-sanders:
Closes #5845

Include absolute graph edges into the data store.

At present, absolute graph parents apply only when the parent is in the same cycle as the child.

E.G. Taking this example:

```
R1 = build
P1 = build[^] => run
```

The `build => run` graph edge would only exist in the R1 cycle (the one in which the build task was scheduled to run) rather than all subsequent cycles to which it applied.

The cause of the bug was a bit of code that was purposefully filtering out absolute graph edges which point to tasks in cycles other than the one that the child task is in. The code contained this comment which suggests doing this resolved some other issue:

```
# If 'foo[^] => bar' only spawn off of '^'.
```

Taking this code out fixes #5845, but I'm worried that this code served a purpose and that by deleting it I'm solving one problem by introducing another.

The comment first appeared here:

https://github.com/cylc/cylc-flow/commit/416240f63873b120a26447558767241df6f0b6e5#diff-e67a65e5cc6c5da815c97f19bf1afc269d56a7f032351aa72364a414ae726048R84

I can't track down why it was added by scanning through #3811. Long shot, @dwsutherland you don't have any memory of this (3 years ago!)?

**Check List**

### DIFF
--- a/6103.fix.md
+++ b/6103.fix.md
@@ -1,0 +1,2 @@
+Absolute dependencies (dependenies on tasks in a specified cycle rather than at a specified offset) are now visible in the GUI beyond the specified cycle.
+
--- a/6103.fix.md
+++ b/6103.fix.md
@@ -1,2 +1,2 @@
-Absolute dependencies (dependenies on tasks in a specified cycle rather than at a specified offset) are now visible in the GUI beyond the specified cycle.
+Absolute dependencies (dependencies on tasks in a specified cycle rather than at a specified offset) are now visible in the GUI beyond the specified cycle.
 
--- a/cylc/flow/taskdef.py
+++ b/cylc/flow/taskdef.py
@@ -101,11 +101,7 @@ def generate_graph_parents(tdef, point, taskdefs):
                 # where (point -Px) does not land on a valid point for woo.
                 # TODO ideally validation would flag this as an error.
                 continue
-            is_abs = (trigger.offset_is_absolute or
-                      trigger.offset_is_from_icp)
-            if is_abs and parent_point != point:
-                # If 'foo[^] => bar' only spawn off of '^'.
-                continue
+            is_abs = trigger.offset_is_absolute or trigger.offset_is_from_icp
             graph_parents[seq].append((parent_name, parent_point, is_abs))
 
     if tdef.sequential:
--- a/tests/integration/test_data_store_mgr.py
+++ b/tests/integration/test_data_store_mgr.py
@@ -18,6 +18,7 @@ import pytest
 from typing import TYPE_CHECKING
 
 from cylc.flow.data_store_mgr import (
+    EDGES,
     FAMILY_PROXIES,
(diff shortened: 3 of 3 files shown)


dwsutherland:
I might have shifted some code about also.. I guess this is separate to the issue of spawning off `^` ? (where you'd spawn out to the RH limit if this dependency was in more than just `R1`)

dwsutherland:
Actually yeah, so I've only included it in the same cycle (like you said)... it makes sense WRT limiting the edges to `^`, but it does feel artificial ..

dwsutherland (approved):
Seems fine, tests passing.. Probably desirable for NIWA site also..

I suspect it may have tripped up the old graph walk in some way? hard to tell now.

oliver-sanders:
Added a test and changelog, rebased onto 8.3.x.

@wxtim Assigned for review 2, well worth a good play with the graph view (static `cylc graph` and live in the GUI).

6103.fix.md
```diff
@@ -0,0 +1,2 @@
+Absolute dependencies (dependenies on tasks in a specified cycle rather than at a specified offset) are now visible in the GUI beyond the specified cycle.
```
wxtim, line 1:
```suggestion
Absolute dependencies (dependencies on tasks in a specified cycle rather than at a specified offset) are now visible in the GUI beyond the specified cycle.
```

wxtim (approved):

MetRonnie:
What about this equivalent bit in `generate_graph_children()`?
https://github.com/cylc/cylc-flow/blob/4d6430a7f0905201c1206f18c2f348d26a82b32e/cylc/flow/taskdef.py#L49-L51

oliver-sanders:
> What about this equivalent bit in generate_graph_children()?

(haven't forgotten about this, still catching up with my inbox).

A good point that I either don't have an answer to or simply can't remember.

I don't think I've seen any consequences of this (abs deps seem to work fine, perhaps this is the graph walking algorithm doing its job?).

Issue #5642, opened 2023-07-25
set: implement the new command


oliver-sanders:
Implement accepted proposal [A New Command for Manual Task Forcing](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-cylc-set.md).

* Rename `set-outputs` to `set`.
* Implement the logic for setting prereqs/outputs as per the proposal.
* Update the CLI and GraphQL schema (test this works as expected in the ui).
* Update documentation.
* This is a breaking change, add an entry into the cylc-doc changes page.

hjoliver:
I'm currently working on this.

Subject to discussion, I've changed the new command name to `cylc set-task`.
- there's a CLI ambiguity problem (set-verbosity)
- plain `set` is arguably not specific enough

PR #6112, opened 2024-05-23
Command validation

### LINKED ISSUE #6013: Scheduler command arg validation

oliver-sanders:
Re-implement command argument validation.

Builds on and supersedes #6064 
Closes #6013
Partially addresses: #3329, #4778

#### Summary

Command argument validation is intended to spot issues with arguments provided by the user (e.g. incompatible options). This validation occurs before the command is queued which means we can return the result to the user (rather than leaving them to look in the scheduler log).

We used to have command argument validation at Cylc 7, the vestigial remains of this were removed in #5642. This PR re-implements command argument validation under a new framework which aims to lead on to #3329:

* Scheduler commands are now asynchronous generators.
* All of the code up to the first `yield` is run before the command is queued (outside of the scheduler main loop).
* The remaining code (up to a `return` or subsequent `yield` statement) is run when the command is executed (within the scheduler main loop). 

#### Future 

In the future, this approach should help us to track the command outcome and return it to the client asynchronously e.g:

```python
async def stop(schd, mode):
    stop_mode = get_stop_mode(mode)  # will raise an exception for an invalid stop mode
    if schd.stop_mode == stop_mode:
        raise InputError('Scheduler is already stopping')
    yield 'Command queued'

    schd.set_stop(stop_mode)

    if stop_mode == StopMode.REQUEST_CLEAN:
        while True:
            active_tasks = {task for task in schd.pool.get_tasks() if task.state(*TASK_STATUSES_ACTIVE)}
            yield f'Waiting for {len(active_tasks)} to complete before stopping...'
            await asyncio.sleep(1)

    if stop_mode !== StopMode.REQUEST_NOW_NOW:
        yield 'Waiting for scheduler event handlers to run.'
```

#### CLI Format Change

In order to display the validation error, this changes the CLI output format.

New multi-workflow format (I hacked the resolver to inject an error):

Screenshot from 2024-05-30 13-52-21

With the `-v` option, it also displays the new command UUID:

Screenshot from 2024-05-30 13-52-35

The single workflow format is the same as the multi-workflow but with the workflow-ID redacted:

Screenshot from 2024-05-30 17-05-59

Note, this changes the success-case message which has been the subject of a couple of prior changes:

* 7756f858812eb21586698e7919e52ef2941cc9bd - "Done"
* e5b3302ecd7da013eadb060cd0b348856f25a017 - "Command submitted; the scheduler will log any problems."
* 1b8c7d2e97e470a791efa8227dd2f9282324e111 - ""

#### OP

<details>
<summary>Original post (contains design rationale)</summary>

@hjoliver, I did a bit of thinking on command validation reviewing #6064 and in combination with https://github.com/cylc/cylc-flow/issues/3329 and came up with the following:

* Make each command an async generator.
* Establish a pattern where the first "step" of these generators is performed at queue-time (e.g. for validation).

Take a look at the last commit of this PR to see it in action. Thinking...

We would like commands to be able to "yield" a string of messages back to the client so that the client can track the status of the command it actioned. Python generators are a natural fit for this purpose, initiate once, call multiple times.

Validation is a precursor to running the command, so we could run the validation as part of the command itself (rather than searching for a validator). We would use a `yield` to separate the queue-time logic (e.g. validation [1]) from the main-loop-time logic (e.g. the command itself [2]).

Doing this avoids having duplicate method signatures for each command (one in `command_validation` and one in `scheduler`) because we use a single generator for both. It also means that the validator may parse arguments into a different format without the command having to repeat this parsing. Furthermore, even if a command does not have any explicit validator, its kwargs will still be checked at queue-time which is handy (force compatibility errors to surface early rather than waiting for the command to be run).

Anyway, lemme know what you think, I'm happy to take a stab at this whilst you're working on workflow_state if you're happy with the approach.

[1] There are other potential uses for the queue-time part of this than validation alone, e.g. we could reject commands if the workflow is in a certain state (e.g. "cannot set prereqs/outputs because this workflow is stopping"), or provide a different return message to illistrate what the command is going to do given the context of the operation (e.g. "changing the stop after cycle point from 123 to 234").
[2] There could actually be other uses for the main-loop-time part of this than issuing a command itself, e.g. trackers. For example the "stop" command might send the stop signal, but keep running so as to communicate updates back to the client (e.g. "stopping the workflow", "waiting for 2 tasks to finish", "waiting for 1 task to finish", "workflow stopped").

</details>


hjoliver:
I think that's a very good idea - please go ahead!

oliver-sanders:
I think that should do the trick, some test fixes required but the functionality should be good.

Whilst doing this I've realised that we can provide some form of validation for most commands and potentially provide a more informative response for most cases without actually requiring #3329.

For example, we could perform task filtering at validation time and report and "bad_items" (i.e. non-matching items) straight back to the user *before* we queue the command so any non-matching patterns appear in the command output rather than expecting the user to go fishing around in the logs.

This is actually quite simple to set up, basically just move the `yield` a couple of lines down, but would mean that task matching is performed when the command is queued, rather than when it is executed which could have consequences. I.E. additional tasks could move into the pool before the command is run. On the one hand this is nicer, Cylc did what the user asked for at the time they asked for it, on the other hand this is less nice, e.g. in theory `cylc kill '*'` might not catch all submissions.

I'll probably leave out the task filtering bit for now, but I'll have a flick through and see what validation we can pull forward for better reporting.

oliver-sanders:
Going for minimum change WRT 8.3.0, can look into moving more validation before command queueing in due course.

oliver-sanders:
One functional test has broken, I think as the result of the output format change (I'm sure it passed earlier), will sort out tomorrow. Otherwise should be good to go.

cylc/flow/commands.py
```diff
@@ -0,0 +1,434 @@
...
+    flow_num: Optional[int] = None,
+):
+    yield
+    if flow_num:
+        schd.pool.stop_flow(flow_num)
+        return
+
+    if cycle_point is not None:
+        # schedule shutdown after tasks pass provided cycle point
+        point = TaskID.get_standardised_point(cycle_point)
```
oliver-sanders, line 182: For future work, we move code like this before the first `yield` in order to flag input errors at command queue time.

hjoliver (commented):
(This looks great, but I haven't done a proper review yet)

oliver-sanders:
(added an integration test for multi-workflow broadcast commands)

cylc/flow/command_validation.py
```diff
@@ -0,0 +1,227 @@
...
+            try:
+                int(val)
+            except ValueError:
+                raise InputError(ERR_OPT_FLOW_VAL.format(val))
+
+    if flow_wait and flows[0] in [FLOW_NEW, FLOW_NONE]:
+        raise InputError(ERR_OPT_FLOW_WAIT)
+
+
+def prereqs(prereqs: Optional[List[str]]):
```
wxtim, line 78: Does this type hint make sense?

```suggestion
def prereqs(prereqs: Optional[List[str]] = None):
```

cylc/flow/command_validation.py
```diff
@@ -0,0 +1,227 @@
...
+        # Error: --pre=<word> other than "all"
+        return None
+
+    if prereq != "all" and tokens["task_sel"] is None:
+        prereq += f":{TASK_OUTPUT_SUCCEEDED}"
+
+    return prereq
+
+
+def outputs(outputs: Optional[List[str]]):
```
wxtim, line 172:
```suggestion
def outputs(outputs: Optional[List[str]] = None):
```

cylc/flow/commands.py
```diff
@@ -0,0 +1,434 @@
...
+
+
+def _command(name: str):
+    """Decorator to register a command."""
+    def _command(fcn: 'Command'):
+        nonlocal name
+        COMMANDS[name] = fcn
+        fcn.command_name = name  # type: ignore
+        return fcn
+    return _command
```
wxtim, line 105:
```suggestion
def _command(name: str = None):
    """Decorator to register a command."""
    def _command(fcn: 'Command'):
        nonlocal name
        if not name:
            name = fcn.__name__
        COMMANDS[name] = fcn
        fcn.command_name = name  # type: ignore
        return fcn
    return _command
```

cylc/flow/command_validation.py
oliver-sanders: Yes, Optional[str] means Union[str, None]. This argument must be provided, it may be a string or it may be None.

Optional types are often used in combination with kwargs that have None defaults, but do not have to be.

cylc/flow/commands.py
oliver-sanders: ?

cylc/flow/commands.py
```diff
@@ -0,0 +1,434 @@
...
+
+@_command('pause')
+async def pause(schd: 'Scheduler'):
+    """Pause the workflow."""
+    yield
+    schd.pause_workflow()
+
+
+@_command('set_verbosity')
+async def set_verbosity(schd: 'Scheduler', level: Union[int, str]):
```
wxtim, line 293: It seems odd on a first inspection that this function (formerly an `@staticmethod` requires the scheduler arg, but it does - why?

oliver-sanders: All of the commands must have the same interface so we can handle them in an abstract way:

```python
async def <command>(schd: 'Scheduler', *args, *kwargs):
```

Otherwise we would need to go back to having a seperate implementation of each command in the server code.

wxtim: On the whole this approach is much tidier... 😄

cylc/flow/commands.py
wxtim: Crucial line 

```
name = name if name else fcn.__name__
```

So that if you are happy with the function name (and I think it's only set, where we're not) you can do

```python
@_command
async def sensible_name():
```

cylc/flow/command_validation.py
wxtim: OK - I think you're right about it being a misnomer: With this signature I'd expect

```
prereqs()
```
to work...

oliver-sanders:
Sadly, I've got a bit more fiddling to do in order to get the return codes coming out right.

cylc/flow/commands.py
oliver-sanders: What issue are you trying to resolve?

wxtim: Just thought the decorator would be neater if you didn't have to give it an arg unless you really needed to.

oliver-sanders: Ah, I wanted the mutation:command mapping to be explicit to decouple these concepts so this duplication is purposeful.

oliver-sanders: FYI: This is what I wanted:
 
```python
from cylc.flow.network.schema import Mutations

@command(Mutations.stop)
def stop(schd ...):
```

It's doable, but nasty due to Graphene internals not attaching the field name to the mutation.

oliver-sanders:
With these changes we can unify the server commands (e.g. broadcast) with the scheduler commands (e.g. stop). Off topic for now, but opened a "some day" issue for later: https://github.com/cylc/cylc-flow/issues/6122

oliver-sanders:
If it's the same on master then it ain't to do with the PR.

wxtim:
```
> > cylc set sour.debt/run2//2405/foo/ --pre hello  # Garbage pre-req
hello
```

I think there might be more message here....

wxtim:
Sorry - mistakenly pushed some stuff I hadn't meant to to your branch and undid it again.

oliver-sanders:
Sorry, didn't spot that this comment hadn't been responded to.

> I think there might be more message here which isn't getting back to the commanding terminal?

Prereqs are evaluated at main-loop time, not queue time so this is expected. As with many command errors, e.g. "no matching tasks", we cannot presently return this error to the user in a timely fashion without suspending the scheduler in the process.

wxtim (approved):

hjoliver (approved):
Nice, works well. Code looks good.

I've not found any problems except that prerequisite validation doesn't work well for in-valid prerequisite formatting (which will likely be a common user error):

```
 $ cylc set --pre=bad test//2/foo  # should be "1/bad" (e.g.)
bad
```

Returning the exception rather than raising it strips the information that says its an error this case. Might just need to add some text?

I'm approving this but not merging in case you want to fix the above before merging during Friday UK time. But either way, I think this should be merged today. Along with my PR, and then we can make the release.

oliver-sanders:
I'll have a look at prereq validation.

The error reporting is `<exception-class>: str(<exception>)` so as long as the relevant context is in the string it should get through to the user.

cylc/flow/command_validation.py
```diff
@@ -0,0 +1,227 @@
...
+        p = prereq(pre)
+        if p is not None:
+            prereqs2.append(p)
+        else:
+            bad.append(pre)
+    if bad:
+        raise InputError(
+            "Use prerequisite format <cycle-point>/<task>:output\n"
+            "\n  ".join(bad)
+        )
```
oliver-sanders, line 127: There's a subtle error on this line which was causing the problem.

It should be:

```
+ '\n  '.join(bad)
```

Without the `+` the error contained only `bad`.

oliver-sanders:
@wxtim, could you check the last commit satisfies Hillary's comment and merge if happy.

oliver-sanders:
Merging with two approvals.

PR #6113, opened 2024-05-29
uis subscription data sync info (UIS-597)

dwsutherland:
sibling of https://github.com/cylc/cylc-uiserver/pull/597

Have included testing some of the changes, however the parts not tested are used by cylc-uiserver not cylc-flow.

This end is to:
- hook into resolvers to extract which workflows a subscription/query is targeting
- adds an API endpoint and store stub

**Check List**

- [x] Tests included (limited), as functionality is only used UIS side.

### DIFF
--- a/cylc/flow/network/resolvers.py
+++ b/cylc/flow/network/resolvers.py
@@ -576,9 +576,14 @@ class BaseResolvers(metaclass=ABCMeta):  # noqa: SIM119
         yielded GraphQL subscription objects.
 
         """
-        # NOTE: we don't expect workflows to be returned in definition order
-        # so it is ok to use `set` here
-        workflow_ids = set(args.get('workflows', args.get('ids', ())))
+
+        args['workflows'] = [
+            Tokens(w_id)
+            for w_id in args.get('workflows', args.get('ids', ()))
+        ]
+        # NOTE: we don't expect workflows to be returned in definition
+        # order so it is ok to use `set` here
+        workflow_ids = {w.id for w in args['workflows']}
 
         sub_id = uuid4()
         info.context['sub_id'] = sub_id
@@ -602,19 +607,26 @@ class BaseResolvers(metaclass=ABCMeta):  # noqa: SIM119
         flow_delta_queues: Dict[str, queue.Queue[Tuple[str, dict]]] = {}
         try:
             # Iterate over the queue yielding deltas
-            w_ids = workflow_ids
             sub_resolver = SUB_RESOLVERS.get(to_snake_case(info.field_name))
             interval = args['ignore_interval']
             old_time = 0.0
+            w_ids = workflow_ids
             while True:
+                old_ids = w_ids
                 if not workflow_ids:
-                    old_ids = w_ids
                     # NOTE: we don't expect workflows to be returned in
                     # definition order so it is ok to use `set` here
                     w_ids = set(delta_queues.keys())
-                    for remove_id in old_ids.difference(w_ids):
-                        if remove_id in self.delta_store[sub_id]:
-                            del self.delta_store[sub_id][remove_id]
+                elif not w_ids.issubset(self.data_store_mgr.data):
+                    # make sure new flows matching the args are picked up
+                    w_ids = {
+                        flow[WORKFLOW].id
+                        for flow in self.data_store_mgr.data.values()
+                        if workflow_filter(flow, args)
+                    }
+                for remove_id in old_ids.difference(w_ids):
+                    if remove_id in self.delta_store[sub_id]:
+                        del self.delta_store[sub_id][remove_id]
                 for w_id in w_ids:
                     if w_id in self.data_store_mgr.data:
                         if sub_id not in delta_queues[w_id]:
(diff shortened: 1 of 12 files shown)


cylc/flow/network/resolvers.py
```diff
@@ -339,6 +353,22 @@ async def get_workflows(self, args):
...
+                Tokens(native_id).workflow_id
+            )
+        await self.data_store_mgr.set_query_sync_levels(
+            set(w_ids)
+        )
+        return [
+            data_store[w_id]
+            for w_id in iter_uniq(w_ids)
+            if w_id in data_store
+        ]
```
oliver-sanders, line 355: This is the same as the existing function, just with the addition of the `await self.data_store_mgr.set_query_sync_levels(set(w_ids))` bit.

Consequently, we have to duplicate the `if is_sub: ...; else` logic to toggle between these two implementations for all uses below.

I think we can simplify things by merging the two implementations like so:

```python
async def get_flow_data_from_ids(self, data_store, native_ids, is_sub):
    w_ids = {Tokens(w_id).workflow_id for w_id in native_ids}
    if is_sub:
        await self.data_store_mgr.set_query_sync_levels(w_ids)
    return [data_store[w_id] for w_id in w_ids if w_id in data_store]
```

Then pass trough a bool flag to toggle:

```python
flow_data = get_flow_data_from_ids(
    self.data_store_mgr.data,
    nat_ids,
    bool(args['delta_store'])
)
```

cylc/flow/network/server.py
```diff
@@ -431,14 +431,26 @@ def pb_entire_workflow(self, **_kwargs) -> bytes:
...
+        Args:
+            elements: Keys from DATA_TEMPLATE dictionary.
+
+        Returns serialised Protobuf message
+        """
+        pb_msg = self.schd.data_store_mgr.get_data_elements(elements)
+        return pb_msg.SerializeToString()
+
+    @expose
+    def pb_delta_elements(self, element_type: str, **_kwargs) -> bytes:
```
oliver-sanders: There are now two `pb_data_elements` methods with two different arguments. The latter overrides the former.

Changing this function signature definition creates an inter-version compatibility issue as the UIS needs to be able to communicate with both.

A backwards-compatible option would be:

```python
def pb_data_elements(self, element_type: Union[str, list[str]], **_kwargs):
    if isinstance(element_type, list:)
        ...
    else:
        ...
```

However, in order to maintain compatibility with older workflows, the UIS will need to fall back to requesting all elements (rather than requesting specific elements via the new interface). I think we'll need to switch behaviours based on the Cylc version (found in UIS-side in the `contact_info`). We *could* change the API version number, but that's usually reserved for more series compatibility issues and would cause newer workflows to become inoperable from older interfaces.

oliver-sanders (commented):
As I understand it:

* When a subscription is started, or a query is issued, the timer field is set.
* Once the timeout has elapsed, the subscription will be recalculated.
* (When a subscription is stopped, we should also set the timer, but I'm not sure we currently are?)

Just wondering if it might be easier to add hook methods, e.g, `query_issues`, `subscription_started`, `subscription_stopped` to avoid having to set and track these timeouts, e.g:

```python
def  query_issued(self, w_id):
    self.loop.call_later(TIMEOUT, self.recalculate_subscription, w_id)

def  subscription_stopped(self, w_id):
    self.loop.call_later(TIMEOUT, self.recalculate_subscription, w_id)
```

dwsutherland:
> As I understand it:
> 
> * When a subscription is started, or a query is issued, the timer field is set.
> * Once the timeout has elapsed, the subscription will be recalculated.
> * (When a subscription is stopped, we should also set the timer, but I'm not sure we currently are?)

Off the top of my head, there's no timer on subscriptions (stays up for as long as the subscription is active)~~, however, the query timer trumps the subscription.. So once the query timer is spent, it falls back to the subscription..~~

The query data types are added to the `level` for the duration of the timer, and then the subscription topics are determined from this, both when the query is made and when the timer finishes..

cylc/flow/network/server.py
dwsutherland: hmm.. Yes, the UIS may need handle this with version checks.. So people can upgrade older workflows via the new UIS..

cylc/flow/network/resolvers.py
dwsutherland: Will have to check if it needs to be order preserving (i.e. for tests), but yeah maybe a list would work the same for query sync level setting.

cylc/flow/network/server.py
dwsutherland: Back compat is in.

dwsutherland:
Back compatibility in:
https://github.com/cylc/cylc-uiserver/pull/597#issuecomment-4748428249

cylc/flow/network/resolvers.py
dwsutherland: Although there may be some simplification not thought of, both calls also pass in a different store: `self.delta_store[args['sub_id']]` and `self.data_store_mgr.data` according to `args['delta_store']`..

And although `args['delta_store']` implies `is_sub`, some subs resolve from the main store `self.data_store_mgr.data` .. So `bool(args['delta_store'])` can be `False` while `is_sub` is `True`.

(I did attempt simplification for a while.. but got stuck on this compication)

dwsutherland: I've refactored this `get_flow_data_from_ids` resolver.. Not as complicated as I first thought

PR #6117, opened 2024-05-31
Add `CYLC_TASK_SHARE_CYCLE_DIR` to task script.

### LINKED ISSUE #6098: share/cycle/<cycle> directory
Add a per-cycle share directory to Cylc.

Rose can create `share/cycle/<cycle>` directories which tasks can utilise via `$ROSE_DATAC`. This should really be Cylc functionality not Rose. Building this into Cylc would make it less confusing, more "standard" (i.e. always available) and more useful (i.e. doesn't depend on `eval $(rose task-env)` or Rose in general).

Proposal:
* Create `share/cycle/<cycle>` in the `job.sh` script.
* Export a `$CYLC_SHARE_CYCLE_DIR` environment variable (or similar).
* Do **not** attempt to implement the `rose task-env --offset` functionality for creating cycle offsets (document using `isodatetime` to do this).

wxtim:
Closes #6098 

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at [cylc/cylc-doc/pull/736](https://github.com/cylc/cylc-doc/pull/736).


ColemanTom:
Just as a note. If people don't use `rose task-env`, nor the cycle directory, and never have, then this will be creating a folder that they now have to cleanup else risk leaving lots of empty folders on disk, using up potential inode quotas or having too many files on lustre in a single folder, causing performance degredation.

elliotfontaine:
If I may put my two cents in, this would be really great! I use Rose anyway for the *rose-suite.conf* file, but this should be Cylc functionality as @oliver-sanders said in May. Right now I have `$(eval rose task-env)` in my **[[root]]** task, which is kinda weird.

hjoliver:
IMO this is good. 
- one additional folder per cycle is not much, compared to the per-task folder/file count as Oliver noted
- any cycling workflow is very likely to need to a per-cycle dir like this (and for non-cycling workflows this will only create one folder)
- housekeeping is pretty easy if needed

MetRonnie:
Kicking tests as they last ran 8 months ago

MetRonnie:
Question on the name of the environment variable:
- `CYLC_SHARE_CYCLE_DIR` (✔️ short)
- `CYLC_WORKFLOW_SHARE_CYCLE_DIR` (✔️ consistent with other env vars)

MetRonnie:
This new env var should be added to https://github.com/cylc/cylc-doc/blob/master/src/reference/job-script-vars/var-list.txt

wxtim:
> This new env var should be added to https://github.com/cylc/cylc-doc/blob/master/src/reference/job-script-vars/var-list.txt

Cylc doc PR already exists, and is linked in the heading - though I didn't change the XXX in the link label, so you might have missed it. (Also PR had the wrong variable name, now fixed)

> CYLC_WORKFLOW_SHARE_CYCLE_DIR (✔️ consistent with other env vars)

To my mind that would imply a variable constant across all cycles and tasks in the workflow, which this isn't!

MetRonnie:
Yes you're right, `CYLC_TASK_SHARE_CYCLE_DIR` would be consistent with other vars

wxtim:
> Yes you're right, `CYLC_TASK_SHARE_CYCLE_DIR` would be consistent with other vars

No, because that would imply it's unique on a per task basis.

MetRonnie (approved):
Happy unless there is a consensus to call it `CYLC_TASK_SHARE_CYCLE_DIR`

oliver-sanders:
> Happy unless there is a consensus to call it `CYLC_TASK_SHARE_CYCLE_DIR`

Yes, I think we should follow the naming pattern of other env vars, clunky as it may be.

oliver-sanders (commented):
LGTM

changes.d/6117.feat.md
```diff
@@ -0,0 +1 @@
+Make workflow/share/cycle/<cycle> available as CYLC_TASK_SHARE_CYCLE_DIR
```
MetRonnie, line 1:
```suggestion
Make `workflow/share/cycle/<cycle>` available to jobs as `$CYLC_TASK_SHARE_CYCLE_DIR`.
```

tests/functional/jobscript/02-share-cycle-cycle/flow.cylc
```diff
@@ -0,0 +1,11 @@
...
+    initial cycle point = 19900211
+    final cycle point = 19900212
+    [[graph]]
+        P1D = task
+
+[runtime]
+    [[task]]
+        script = """
+            echo "Hello World" > "${CYLC_TASK_SHARE_CYCLE_DIR}/stuff"
+        """
```
wxtim, line 11:
```suggestion
        """

```

changes.d/6117.feat.md
```diff
@@ -0,0 +1 @@
+Make `workflow/share/cycle/<cycle>` available to jobs as `$CYLC_TASK_SHARE_CYCLE_DIR`.
```
oliver-sanders, line 1: We should communicate that we are also creating this directory as it might not have existed before for non-rose users.

MetRonnie (approved):

oliver-sanders (approved):

PR #6120, opened 2024-06-04
Allow CLI selection of a single rule to check.

### LINKED ISSUE #5544: cylc lint - command line option to only check specific options
### Problem

With the `cylc lint` tool, there is the ability to turn rules off by the command line, but I can't see an option to only run chosen ones. If you want to work through one rule at a time, it would be useful to be able to choose which ones to turn on via the CLI.

### Proposed Solution

Add a new command line argument to only run specific rules.

wxtim:
Closes #5544 

Fix a bug where empty flow.cylc files caused traceback.

**Check List**


oliver-sanders:
PR title and description don't appear to match?

Does this do both? If so maybe a bugfix changelog entry?

wxtim:
> Does this do both? If so maybe a bugfix changelog entry?

Yes. Sorry - original text not super clear. Have added the changelog entry as suggested.

cylc/flow/loggingutil.py
```diff
@@ -434,3 +434,25 @@ def patch_log_level(logger: logging.Logger, level: int = logging.INFO):
...
+        header: To use otherwise
+    """
+    if len(items) == 0:
+        raise ValueError('This function does not take len==0 iterables')
+    elif len(items) == 1:
+        return singular_header.format(items[0])
+    else:
+        msg = header + '\n    * '
+        msg += '\n    * '.join(items)
+        return msg
```
oliver-sanders, line 458: This function feels a bit over-complicated for a single use.

I think it should fail type checking as generators are iterables but are not supported by `len()`?

cylc/flow/scripts/lint.py
```diff
@@ -443,6 +443,7 @@ def list_wrapper(line: str, check: Callable) -> Optional[Dict[str, str]]:
     'S': 'Style'
 }
 LINE_LEN_NO = 'S012'
+EMPTY_FILE = 'S000'   # Reference for a completely empty file.
```
oliver-sanders, line 446: Do we need a code for this?

Can't we just exit `0` (nothing in the file to complain about).

cylc/flow/loggingutil.py
wxtim: It is over complicated for a single use. I think that this is code we've replicated in a number of places, and may want again in future . The idea is that this function is available for refactoring elsewhere and for future use cases.

cylc/flow/scripts/lint.py
wxtim: Yes. Fair point.

cylc/flow/scripts/lint.py
```diff
@@ -1417,6 +1434,14 @@ def main(parser: COP, options: 'Values', target=None) -> None:
         max_line_len=mergedopts[MAX_LINE_LENGTH]
     )
 
+    if options.rule:
+        checks = {k: v for k, v in checks.items() if k in options.rule}
```
oliver-sanders, line 1438: I think this can be simplified to:

```python
checks = {rule: checks[rule] for rule in options.rule}
```

Which has the advantage of raising an error if a rule does not exist.

```suggestion
        try:
            checks = {rule: checks[rule] for rule in options.rule}
        except KeyError as exc:
            LOG.error(f'No such rule {exc.args[0]}')
            sys.exit(1)
```

However, it looks like `--rule` can only accept one rule at present?

I feel we should implement this more like `--select` (ruff/flake8), to allow the selection of rules OR rule sets?

cylc/flow/scripts/lint.py
```diff
@@ -1340,6 +1343,13 @@ def get_option_parser() -> COP:
         action='store_true',
         default=False,
     )
+    parser.add_option(
+        '--rule', '-R',
+        help='Carry out only specific checks.',
+        choices=list(parse_checks(['style', '728']).keys()),
```
oliver-sanders, line 1349: So `--rule` can be one of `style` or `728`.

AND `--ruleset` can be one of `style`, `728` or `all`.

What's the point of `--rule`?

oliver-sanders: The `--select` / `--ignore` interface of `ruff` is a good gauge of how to go about implementing rule selection.

```
--select=R  # select all Rxxx rules
--select=R001  # select only R001
--select=R --ignore=R001  # select all R other than R001
--select=R001,R002  # select both R001 and R002
--select=ALL  # magic
```

wxtim: > AND --ruleset can be one of style, 728 or all.

No - rulest can be any of the items in the parsed checks for both rulesets currently available because 

```python
>>> parse_checks(['style', '728'].keys()
DictKeys('S001', 'S002' ... ... 'R999')

```

wxtim:
Redrafting - not a priority

PR #6123, opened 2024-06-06
Tokens: handle ISO 8601 long format cycle points

### LINKED ISSUE #6110: ID tokeniser bug
This is a valid ISO8601 format: `2050-01-01T00:00Z` but the tokeniser cuts the `:00Z` bit off the end, which ends up with the wrong result if your system clock is not UTC or if you have non-zero minutes.

```python
>>> from cylc.flow.id import Tokens

>>> Tokens('foo//20500101T0000Z')
<id: foo//20500101T0000Z>  # good

>>> Tokens('foo//2050-01-01T00:00Z')
<id: foo//2050-01-01T00>  # bug - later this gets interpreted as local time
```

We disallow this a cycle point format in workflows because `:` is not a valid character in filenames, but cycle points can be specified on the command line too and then converted to the workflow format.

We have a functional test that does this, and fails for non-UTC clocks: `tests/f/workflow-state/06a-noformat.t` (for the `cylc workflow-state` command).

This diff seems to fix the tokeniser (don't disallow ':' in cycle point):

```diff
$ git diff
diff --git a/cylc/flow/id.py b/cylc/flow/id.py
index cba3c4833..b9d7cc2e1 100644
--- a/cylc/flow/id.py
+++ b/cylc/flow/id.py
@@ -403,10 +403,10 @@ class Tokens(dict):
 # //cycle[:sel][/task[:sel][/job[:sel]]]
 RELATIVE_PATTERN = rf'''
     //
-    (?P<{IDTokens.Cycle.value}>[^~\/:\n]+)
+    (?P<{IDTokens.Cycle.value}>[^~\/\n]+)
     (?:
       :
-      (?P<{IDTokens.Cycle.value}_sel>[^\/:\n]+)
+      (?P<{IDTokens.Cycle.value}_sel>[^\/\n]+)
     )?
     (?:
       /
```

Alternatively we might want to detect the `:` here and raise an error instead, but I can't see any harm in allowing this user-friendly format in CLI and UI queries?

MetRonnie:
Closes #6110 

Before:
```python-console
>>> from cylc.flow.id import tokenise

>>> dict(tokenise('//2050-01-01T00:00Z'))
{'cycle': '2050-01-01T00', 'cycle_sel': '00Z'}
```

After:
```pyth-console
>> from cylc.flow.id_cli import cli_tokenise

>>> dict(cli_tokenise('//2050-01-01T00:00Z'))
{'cycle': '2050-01-01T00:00Z'}

>>> dict(cli_tokenise('//2050-01-01T00:fail'))
{'cycle': '2050-01-01T00', 'cycle_sel': 'fail'}
```

I decided to have a go at supporting the long format rather than disallowing it, seeing as it isn't much diff. Performance-wise this should have negligible impact as the "expensive" part of it is only going to be hit once by users running a CLI command.

**Check List**

- [x] No dependency changes 
- [x] Tests are included
- [x] No docs needed

### DIFF
--- a/cylc/flow/id.py
+++ b/cylc/flow/id.py
@@ -403,7 +403,7 @@ class Tokens(dict):
 # //cycle[:sel][/task[:sel][/job[:sel]]]
 RELATIVE_PATTERN = rf'''
     //
-    (?P<{IDTokens.Cycle.value}>[^~\/:\n]+)
+    (?P<{IDTokens.Cycle.value}>[^~\/:\n][^~\/\n]*?)
     (?:
       :
       (?P<{IDTokens.Cycle.value}_sel>[^\/:\n]+)
--- a/changes.d/6123.fix.md
+++ b/changes.d/6123.fix.md
@@ -1,0 +1,1 @@
+Allow long-format datetime cycle points in IDs used on the command line.
--- a/cylc/flow/id_cli.py
+++ b/cylc/flow/id_cli.py
@@ -20,6 +20,9 @@ from pathlib import Path
 import re
 from typing import Optional, Dict, List, Tuple, Any
 
+from metomi.isodatetime.parsers import TimePointParser
+from metomi.isodatetime.exceptions import ISO8601SyntaxError
+
 from cylc.flow import LOG
 from cylc.flow.exceptions import (
     InputError,
@@ -28,6 +31,7 @@ from cylc.flow.hostuserutil import get_user
 from cylc.flow.id import (
     Tokens,
     contains_multiple_workflows,
+    tokenise,
     upgrade_legacy_ids,
 )
 from cylc.flow.pathutil import EXPLICIT_RELATIVE_PATH_REGEX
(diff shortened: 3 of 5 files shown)


cylc/flow/id.py
```diff
@@ -553,6 +553,39 @@ def _dict_strip(dictionary):
     }
 
 
+def _handle_iso_long_fmt(dictionary):
+    """If a cycle selector is present, it could be part of a long-format
+    ISO 8601 datetime that was erroneously split. Re-attach it if it
+    results in a valid datetime.
```
hjoliver, line 559: Good idea.

dwsutherland (approved):
LGTM 👍 
```
>>> dict(Tokens('//2050-01-01T00:00Z:failed'))
{'cycle': '2050-01-01T00:00Z', 'cycle_sel': 'failed', 'task': None, 'task_sel': None, 'job': None, 'job_sel': None}
```

MetRonnie:
On master:
```
In [2]: %timeit Tokens('foo//20500101T0000Z/a')
5.45 µs ± 313 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
On this branch:
```
In [2]: %timeit Tokens('foo//20500101T0000Z/a')
7.64 µs ± 55.7 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [3]: %timeit Tokens('foo//2050-01-01T00:00Z/a')
1.13 ms ± 10.6 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
Ok, so I wasn't expecting it to be that much worse. Turns out init'ing the `TimePointParser` class goes through a process of generating a bunch of regexes? https://github.com/metomi/isodatetime/blob/a4614a2153548ca99e2999ba6f20ae40242b0037/metomi/isodatetime/parsers.py#L140

I don't think this should be a problem for Cylc internals as we don't use the extended datetime format for cycle points, just have to make sure we keep it that way? Or I guess I could move this to `id_cli.py` so it definitely only applies to the CLI

dwsutherland:
> I don't think this should be a problem for Cylc internals as we don't use the extended datetime format for cycle points, just have to make sure we keep it that way?

I think the cycle format is used in the ids of the data-store, so it's whether they are being tokenised from string format frequently.. (which they very well may be)

dwsutherland:
> Note the cycle point format set in `flow.cylc` can't be the extended/long format (as it is used for job dir paths so can't contain `:`)

Ah yes, and only that could be used internally.. So I think we're safe.

MetRonnie:
> Or I guess I could move this to `id_cli.py` so it definitely only applies to the CLI

I've gone ahead and done this as it makes sense to me

hjoliver (approved):

hjoliver:
Post-merge comment - this might warrant an explanation and example in CLI help, otherwise users are unlikely to realize that it's possible.

MetRonnie:
`cylc help id` only shows examples with integer cycles for brevity anyway - I think it's fine the way it is. The important thing is that either the short or long format will work.

PR #6132, opened 2024-06-11
sphinx reference fix

hjoliver:
tiny doc tweak companion of https://github.com/cylc/cylc-doc/pull/727


hjoliver:
(Had to combine with #5809 )

PR #6134, opened 2024-06-13
Fix ID workflow count.

hjoliver:
ID parsing for the CLI is broken with respect to the `max_workflows` constraint - it doesn't count workflows if the ID is "task-like" (which may include workflow).

A pretty trivial fix, so tentatively targeting 8.3.0.

**Check List**

### DIFF
--- a/cylc/flow/id_cli.py
+++ b/cylc/flow/id_cli.py
@@ -465,14 +465,14 @@ def _infer_latest_runs(tokens_list, src_path, alt_run_dir=None):
 def _validate_number(*tokens_list, max_workflows=None, max_tasks=None):
     if not max_workflows and not max_tasks:
         return
-    workflows_count = 0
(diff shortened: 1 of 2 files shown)


hjoliver:
One or two reviews will do. I think Oliver wrote the original code.

MetRonnie:
Is there a CLI example which is broken at present?

oliver-sanders:
> Is there a CLI example which is broken at present?

After a bit of playing, this example doesn't work as expected:

```
$ cylc show 'workflow1//*' 'workflow2//*'
```

You would expect it to list all tasks from both workflows, however, it only lists tasks from the first.

oliver-sanders (approved):

oliver-sanders:
(can easily make 8.3.0 with this one)

MetRonnie (approved):

PR #6136, opened 2024-06-13
protobuf: automate derived file generation

oliver-sanders:
(Training day washup)

Take the guesswork out of protobuf schema changes by automating the updating of the derived Python files:

* Use the bufbuild/buf tool to:
  * Lint the protobuf schema.
  * Check for breaking changes.
  * Configure the generation of derived Python files.
* Address lint failures:
  * Ignore the capitalisation of the URL field (protobuf likes snake case).
  * Define the schema's "package".
  * Define the schema's version (made this match the Cylc API version number).
  * Locate the schema in a directory hierarchy which matches "<package>/<version>" (some protobuf standard).
* Add a GitHub action which:
  * Lints & checks for breaking changes.
  * Automates the checking and generation of derived files.
  * Commits the results back on pull request branches.

Example CI runs (my fork):
* [Adding a new field](https://github.com/oliver-sanders/cylc-flow/pull/74) (files updated and committed back to branch)
* [Removing a field](https://github.com/oliver-sanders/cylc-flow/pull/76) (breaking change)

Note:
* Requires a small cylc-uiserver change to match.

**Check List**


oliver-sanders:
(note the protobuf compatibility test will correctly fail on this branch because the schema file it is comparing against does not exist on the branch it is being merged into)

oliver-sanders:
Moving the protobuf files in with the network code introduced circular import issues due to the definition of code in `cylc.flow.network.__init__.py`. Moved this code into its own modules to resolve the issue.

.github/workflows/protobuf.yml
```diff
@@ -0,0 +1,83 @@
...
+        uses: cylc/release-actions/configure-git@v1
+
+      - name: Install Protobuf
+        uses: mamba-org/setup-micromamba@v1
+        with:
+          # install protobuf into a mamba env (note use shell = "bash -el {0}"
+          # to access this envionment)
+          environment-name: protobuf
+          create-args: protobuf
+          init-shell: bash
```
MetRonnie, line 32: Is there any point using an environment?

.github/workflows/protobuf.yml
```diff
@@ -0,0 +1,83 @@
...
+
+          # NOTE: bufbuild does exist on conda-forge but hasn't been updated for a while
+          brew install bufbuild/buf/buf
+
+      - name: Lint
+        run: |
+          # lint .proto files
+          eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"  # activate homebrew
+          cd cylc/flow/network/protobuf
+          buf lint
```
MetRonnie, line 46: (Looks like `buf` can be installed via npm which might be easier to use in GH Actions, rather than activating homebrew every step)

---

<details><summary>Btw, first result when I googled "buf" 😆</summary>

image

</details>

MetRonnie (commented):
Had a quick look at this so far

.github/workflows/protobuf.yml
oliver-sanders: Conda/Mamba can only install packages into isolated environments, they cannot install into the system environment.

MetRonnie: Why do we need an environment on the temporary GH Actions runner?

.github/workflows/protobuf.yml
```diff
@@ -0,0 +1,83 @@
...
+          # generate .py and .pyi files from the .proto files
+          eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"  # activate homebrew
+          micromamba activate protobuf
+          cd cylc/flow/network/protobuf
+          buf generate
+
+      - name: Commit & Push
+        run: |
+          if [[ -z $(git diff --stat) ]]; then
+            echo '::error:: No functional changes made to the protobuf schema'
```
MetRonnie, line 74:
```suggestion
          if git diff --exit-code --stat; then
            echo '::warning::No functional changes made to the protobuf schema'
```

.github/workflows/protobuf.yml
```diff
@@ -0,0 +1,83 @@
...
+            exit 0
+          else
+            echo '::info:: pushing update commit'
```
MetRonnie, line 77:
```suggestion
            echo '::notice::pushing update commit'
```

.github/workflows/protobuf.yml
```diff
@@ -0,0 +1,83 @@
...
+              git add -u
+              git commit -m 'protobuf: updating generated files'
+              git remote add pr https://github.com/${{ github.event.pull_request.head.repo.owner.login }}/cylc-flow
+              git push pr HEAD:${{ github.head_ref }}
```
MetRonnie, line 81: Ah, I'm not sure this will work for a PR from a fork ☹️

I think `GITHUB_TOKEN` only has read permissions in PRs from forks.

.github/workflows/protobuf.yml
MetRonnie: I am asking why are you using mamba/conda instead of pip (which is simpler - no `bash -el {0}` needed)?

oliver-sanders: Mamba installs low-level dependencies properly, pip doesn't.

.github/workflows/protobuf.yml
oliver-sanders: I wouldn't have used a system package manager if I could have avoided it (I don't like system package managers), but I can't remember.

.github/workflows/protobuf.yml
oliver-sanders: Dammit (although the workflow would work on *their* fork).

Only one way to find out. My master branch is still @ [f3342c2](https://github.com/oliver-sanders/cylc-flow/commit/f3342c25a09f39dcd0cb6dbebfa556dd9d7346bc), maybe try raising a PR there (along the lines of the one in the OP) and see what happens?

I know there are other repos that manage to do this. E.G. checkout jupyter-server#1448 (deliberately not linking their issue), those pre-commit commits are pushed onto the branch automatically.

MetRonnie: https://togithub.com/jupyter-server/jupyter_server/pull/1448

MetRonnie: I think that is done by an app rather than GH Actions: https://github.com/marketplace/pre-commit-ci

MetRonnie: Didn't work ☹️ https://github.com/oliver-sanders/cylc-flow/pull/78#issuecomment-2278325871

oliver-sanders:
Drafting in response to https://github.com/cylc/cylc-flow/pull/6136#discussion_r1711773279

PR #6152, opened 2024-06-18
xtriggers: doc change

oliver-sanders:
The cylc-doc spellcheck doesn't like these return args.


MetRonnie (approved):
I'm getting this one locally
> Warning, treated as error:
../cylc-flow/cylc/flow/exceptions.py:docstring of cylc.flow.exceptions.InvalidCompletionExpression:3: : Spell check: whitelisted: Raised when non-whitelisted syntax is present..

but I think that might be variance in dictionaries between local and GH Actions as I haven't seen it on the latter

PR #6161, opened 2024-06-19
Lint  

oliver-sanders:
Small lint fixes for rules not currently covered by our CI, low priority.

Took `ruff` for a spin to see how handy its code fixers are (it's not just a linter, it can fix some issues for you). Some of the fixers are kinda useful, but most the good rules have to be done manually, so turns out they aren't that helpful for us, worth a try, never mind.

Kept a four of the fixes and turned them into this PR. They're mostly small, but [B904](https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/) created a good bit of diff. Long story short, if you're re-raising an exception with a more helpful class/message, you generally want to do `raise Exception(...) from None`, otherwise the context of the exception you're re-raising is included in the traceback which kinda defates the object.

I.E:

```python

try:
    # the general exception we are catching
    raise Exception('foo')
except Exception as exc:
    # the specific exception we are raising
    raise InputError('bar')
```

Preserves the parent exception context resulting in a confusing traceback:

```
  File "/var/tmp/tmp.wlysihZu3M/mess.py", line 5, in <module>
    raise Exception('foo')
Exception: foo

  File "/var/tmp/tmp.wlysihZu3M/mess.py", line 8, in <module>
    raise InputError('bar')
cylc.flow.exceptions.InputError: bar
```

Whereas:

```python

try:
    # the general exception we are catching
    raise Exception('foo')
except Exception as exc:
    # the specific exception we are raising
    raise InputError('bar') from None
```

Chops off the parent context:

```
  File "/var/tmp/tmp.wlysihZu3M/mess.py", line 8, in <module>
    raise InputError('bar') from None
cylc.flow.exceptions.InputError: bar
```

Using `from exc` is the same as the default behaviour (just more explicit).

Raising the whole exception tree is a useful feature for debugging uncaught errors, but not so good for errors we are trying to niceify for users.

**Check List**

### DIFF
--- a/cylc/flow/broadcast_report.py
+++ b/cylc/flow/broadcast_report.py
@@ -72,7 +72,7 @@ def get_broadcast_change_iter(modified_settings, is_cancel=False):
         value = setting
         keys_str = ""
         while isinstance(value, dict):
-            key, value = list(value.items())[0]
+            key, value = next(iter(value.items()))
             if isinstance(value, dict):
                 keys_str += "[" + key + "]"
             else:
--- a/cylc/flow/clean.py
+++ b/cylc/flow/clean.py
@@ -129,7 +129,7 @@ def _clean_check(opts: 'Values', id_: str, run_dir: Path) -> None:
     except ContactFileExists as exc:
         raise ServiceFileError(
             f"Cannot clean running workflow {id_}.\n\n{exc}"
-        )
+        ) from None
 
 
 def init_clean(id_: str, opts: 'Values') -> None:
@@ -173,7 +173,7 @@ def init_clean(id_: str, opts: 'Values') -> None:
             try:
                 platform_names = get_platforms_from_db(local_run_dir)
             except ServiceFileError as exc:
-                raise ServiceFileError(f"Cannot clean {id_} - {exc}")
+                raise ServiceFileError(f"Cannot clean {id_} - {exc}") from None
             except sqlite3.OperationalError as exc:
                 # something went wrong with the query
                 # e.g. the table/field we need isn't there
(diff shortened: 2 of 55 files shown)


cylc/flow/broadcast_report.py
```diff
@@ -72,7 +72,7 @@ def get_broadcast_change_iter(modified_settings, is_cancel=False):
         value = setting
         keys_str = ""
         while isinstance(value, dict):
-            key, value = list(value.items())[0]
+            key, value = next(iter(value.items()))
```
oliver-sanders, line 75: Get the first item from the iterator rather than working through the whole iterator, turning it into a list, then getting the first item.

cylc/flow/clean.py
```diff
@@ -129,7 +129,7 @@ def _clean_check(opts: 'Values', id_: str, run_dir: Path) -> None:
     except ContactFileExists as exc:
         raise ServiceFileError(
             f"Cannot clean running workflow {id_}.\n\n{exc}"
-        )
+        ) from None
```
oliver-sanders, line 132: Don't include the ContactFileExists exception context in the ServiceFileError.

cylc/flow/network/multi.py
```diff
@@ -235,7 +235,7 @@ def _report(
     """
     try:
         ret: List[Tuple[Optional[str], Optional[str], bool]] = []
-        for _mutation_name, mutation_response in response.items():
+        for mutation_response in response.values():
```
oliver-sanders, line 238: PERF101: don't needless iterate items

cylc/flow/resources.py
```diff
@@ -205,12 +205,9 @@ def get_api_key() -> str:
     load over a larger number of keys to prevent hitting the cap with group
     sessions.
     """
-    keys = []
     with open((TUTORIAL_DIR / 'api-keys'), 'r') as api_keys:
-        for api_key in api_keys:
-            keys.append(api_key)
-    shuffle(keys)
-    return keys[0].strip()
+        return choice(list(api_keys)).strip()  # nosec
```
oliver-sanders, line 209: PERF402: Don't build an entire list when you only need one item

hjoliver (approved):
Nice, worth doing!

PR #6163, opened 2024-06-20
`cylc version --long`: ensure correct path is printed for `cylc` executable

MetRonnie:
While testing on GH Actions MacOS 14, I found that `sys.executable`/`sys.prefix` gave different locations to `$ command -v cylc`. But `sys.argv[0]` is consistent (tested on Ubuntu too).

On the MacOS 14 runner:

```console
$ cylc version --long
8.3.1.dev (/Library/Frameworks/Python.framework/Versions/3.9)
...
```

```console
$ command -v cylc
/Users/runner/hostedtoolcache/Python/3.9.13/arm64/...
```

**Check List**

- [x] No dependency changes 
- [x] Tests are updated
- [x] Minor change, no changelog entry needed
- [x] No docs needed

### DIFF
--- a/cylc/flow/scripts/cylc.py
+++ b/cylc/flow/scripts/cylc.py
@@ -88,7 +88,7 @@ def get_version(long=False):
     from pathlib import Path
     version = f"{__version__}"
     if long:
-        version += f" ({Path(sys.executable).parent.parent})"
+        version += f" ({Path(sys.argv[0])})"
     return version
 
 
--- a/tests/functional/cli/01-help.t
+++ b/tests/functional/cli/01-help.t
@@ -79,12 +79,7 @@ run_ok "${TEST_NAME_BASE}-id" cylc help id
 
 # Check "cylc version --long" output is correct.
 cylc version --long | head -n 1 > long1
-WHICH="$(command -v cylc)"
-PARENT1="$(dirname "${WHICH}")"
-PARENT2="$(dirname "${PARENT1}")"
-echo "$(cylc version) (${PARENT2})" > long2
-# the concise version of the above is a bash quoting nightmare:
-# echo "$(cylc version) ($(dirname $(dirname $(which cylc))))" > long2
+echo "$(cylc version) ($(command -v cylc))" > long2
 cmp_ok long1 long2
 
 # --help with no DISPLAY


cylc/flow/scripts/cylc.py
```diff
@@ -88,7 +88,7 @@ def get_version(long=False):
     from pathlib import Path
     version = f"{__version__}"
     if long:
-        version += f" ({Path(sys.executable).parent.parent})"
+        version += f" ({Path(sys.argv[0])})"
```
oliver-sanders, line 91: I'm a bit worried that this might not produce the desired result in all cases, e.g:

```console
$ python -c 'import sys; print(sys.argv[0])'
-c
```

MetRonnie: Produces the right path in all testing I've done. Not sure how it could not, as it will always be run via the `cylc` script?

oliver-sanders: Yes, but the Cylc script itself is called by a wrapper script that is auto-generated from the entry points.

E.G, try:

```console
$ mamba activate <env>
$ cat $(which cylc)
```

This script may differ between different Mamba/Conda client implementations (incl conda-pack) as well as venv, virtualenv and native Python/pip.

oliver-sanders: How about using `__file__`?

MetRonnie: Granted, it won't give the same result as `command -v cylc` if using a wrapper script; but it will still give the ultimate `cylc` executable path (tested).

I've tested my solution on GH Actions Ubuntu and MacOS 14 runners which aren't using conda, and get the right results.

> How about using `__file__`?

https://github.com/cylc/cylc-flow/blob/5b215901e336f4f8b64738fca3402e4548eef035/cylc/flow/scripts/cylc.py#L83-L86

oliver-sanders: From testing, only the `-c` argument to `python` appears to be able to get `sys.argv[0]` to give the wrong answer.

oliver-sanders (approved):
The idea here is that the location of the Cylc wrapper script is a better indication of the location of the Cylc installation than the location of the Python binary (which is what we were using before).

I think that makes sense.

PR #6169, opened 2024-06-25
Log job failure even when there are retries configured

### LINKED ISSUE #6151: scheduler: task failure not logged when retries are configured
A small logging bug where task failures can be omitted from the logs when execution retries are configured.

We are currently logging all task state changes, however, when retries are configured, the a failed task does not change state until all retries have been exhausted.

### Reproducible Example

```ini
[scheduling]
    [[graph]]
        R1 = foo

[runtime]
    [[foo]]
        execution retry delays = PT1S
        script = """
            if [[ $CYLC_TASK_SUBMIT_NUMBER -gt 1 ]]; then
                true
            else
                false
            fi
        """
```

```console
$ cylc play -N <workflow>

 ██   Copyright (C) 2008-2024 NIWA

...
INFO - [1/foo:waiting(runahead)] => waiting
INFO - [1/foo:waiting] => waiting(queued)
INFO - [1/foo:waiting(queued)] => waiting
INFO - [1/foo/01:preparing] submitted to localhost:background[91088]
INFO - [1/foo/01:preparing] => submitted
INFO - [1/foo/01:submitted] => running
INFO - [1/foo/01:running] => waiting
WARNING - [1/foo:waiting] retrying in PT1S (after 2024-06-18T09:56:09+01:00)
INFO - xtrigger satisfied: _cylc_retry_1/foo = wall_clock(trigger_time=1718700969.3399053)
INFO - [1/foo:waiting] => waiting(queued)
INFO - [1/foo:waiting(queued)] => waiting
INFO - [1/foo/02:preparing] submitted to localhost:background[91380]
INFO - [1/foo/02:preparing] => submitted
INFO - [1/foo/02:submitted] => running
INFO - [1/foo/02:running] => succeeded
INFO - Workflow shutting down - AUTOMATIC
INFO - DONE
```

wxtim:
Closes #6151 

**Check List**


oliver-sanders (commented):
Unfortunately, I don't think it's this simple.

* I think this diff means that polled log messages for task failure will go back to being duplicated.
* This only covers failure, but submission failure also has retries so may require similar treatment.
* I think the failures *before* retries are exhausted will now get logged at CRITICAL level rather than INFO.

wxtim:
> This only covers failure, but submission failure also has retries so may require similar treatment.

I think that submission failure is already handled correctly - it certainly is in the simplistic case where you feed it `platform=garbage` you get a submission failed message at level critical.

oliver-sanders:
> I think this diff means that polled log messages for task failure will go back to being duplicated.
>
>> I think that you have a particular closed issue in mind, but I can't find it... Can you point it out to me?

No, I'm not thinking of the other log message duplication issue.

The change made here bypassed logic that was used for suppressing duplicate log messages (the `f'{FAIL_MESSAGE_PREFIX}ERR'` bit):

https://github.com/cylc/cylc-flow/pull/6169/commits/8f20ab088d2305670667f1beadd31de5611e5ec0#diff-d6de42ef75ecc801c26a6be3a9dc4885b64db89d32bce6f07e319595257b9b2eL930

However, in your more recent "fix" commit, you have put this back the way it was before:

https://github.com/cylc/cylc-flow/pull/6169/commits/3cedf2facefc58c61b14acc9cf44e0fee5f44710#diff-d6de42ef75ecc801c26a6be3a9dc4885b64db89d32bce6f07e319595257b9b2eR930

oliver-sanders (changes requested):
Using the example from the issue with this branch I do not see any job failed notification:

```
INFO - [1/foo:waiting(runahead)] => waiting
INFO - [1/foo:waiting] => waiting(queued)
INFO - [1/foo:waiting(queued)] => waiting
INFO - [1/foo/01:preparing] submitted to localhost:background[75057]
INFO - [1/foo/01:preparing] => submitted
INFO - [1/foo/01:submitted] => running
WARNING - [1/foo/01:running] => failed (retrying in PT1S (after 2024-07-15T13:30:43+01:00))
INFO - [1/foo/01:running] => waiting
INFO - xtrigger satisfied: _cylc_retry_1/foo = wall_clock(trigger_time=1721046643.8823874)
INFO - [1/foo:waiting] => waiting(queued)
INFO - [1/foo:waiting(queued)] => waiting
INFO - [1/foo/02:preparing] submitted to localhost:background[75345]
INFO - [1/foo/02:preparing] => submitted
INFO - [1/foo/02:submitted] => running
INFO - [1/foo/02:running] => succeeded
INFO - Workflow shutting down - AUTOMATIC
INFO - DONE
```

cylc/flow/task_events_mgr.py
```diff
@@ -1325,9 +1325,9 @@ def _process_message_failed(self, itask, event_time, message, forced):
         else:
             # There is an execution retry lined up.
             timer = itask.try_timers[TimerFlags.EXECUTION_RETRY]
-            self._retry_task(itask, timer.timeout)
             delay_msg = f"retrying in {timer.delay_timeout_as_str()}"
-            LOG.warning(f"[{itask}] {delay_msg}")
+            LOG.warning(f'[{itask}] => {TASK_OUTPUT_FAILED} ({delay_msg})')
```
oliver-sanders, line 1329: LGTM.

Could we get the actual failure message into this as it would appear if retry delays were not being used.

(e.g. if you hit wallclock limit you might expect to see XCPU here)

wxtim: > Needs tests.

Yes, wanted you to ok the approach

wxtim:
This does not apply to submit failure, because submit failure will always log a critical warning through the `jobs-submit` command.

MetRonnie (changes requested):
(Test failures)

changes.d/fix.6169.md
MetRonnie, line 1: Invalid filename

cylc/flow/task_events_mgr.py
```diff
@@ -742,7 +742,7 @@ def process_message(
                 # Already failed.
                 return True
             if self._process_message_failed(
-                itask, event_time, self.JOB_FAILED, forced
+                itask, event_time, self.JOB_FAILED, forced, message
```
MetRonnie, line 745: No need for this as this will make the log message
```
[1/foo:waiting] => failed with failed
```

```suggestion
                itask, event_time, self.JOB_FAILED, forced
```

cylc/flow/task_events_mgr.py
```diff
@@ -1327,7 +1334,9 @@ def _process_message_failed(self, itask, event_time, message, forced):
             timer = itask.try_timers[TimerFlags.EXECUTION_RETRY]
             self._retry_task(itask, timer.timeout)
             delay_msg = f"retrying in {timer.delay_timeout_as_str()}"
-            LOG.warning(f"[{itask}] {delay_msg}")
+            LOG.warning(
+                f'[{itask}] => {TASK_OUTPUT_FAILED} with {full_message}'
+                f' \n({delay_msg})')
```
MetRonnie, line 1339: My preference for this log message would be
```suggestion
                f'[{itask}] {full_message or self.EVENT_FAILED} - '
                f'{delay_msg}'
            )
```

cylc/flow/task_events_mgr.py
wxtim: No, it's failed with reason:
```
[1/foo:waiting] => failed with failed/XTERM
```

MetRonnie: Not in this case, it is `self.EVENT_FAILED` which is `'failed'`

changes.d/fix.6169.md
MetRonnie: (#6269 waiting on review)

cylc/flow/task_events_mgr.py
```diff
@@ -790,7 +790,7 @@ def process_message(
             self.workflow_db_mgr.put_update_task_jobs(
                 itask, {"run_signal": signal})
             if self._process_message_failed(
-                itask, event_time, self.JOB_FAILED, forced
+                itask, event_time, self.JOB_FAILED, forced, message
             ):
```
oliver-sanders, line 794: I think you've missed a `_process_message_failed` somewhere which will also need the `message` argument added.

MetRonnie:
Test failures

wxtim:
@oliver-sanders & @MetRonnie : Sorry, but I made a total hash of rebasing this and may have introduced some regressions. :(

changes.d/6169.fix
```diff
@@ -0,0 +1 @@
+Ensure that job failure is logged, even when the presence of retries causes the task not to change state.
```
wxtim, line 1:
```suggestion
Ensure that job failure is logged, even when the presence of retries causes the task not to change state.

```

oliver-sanders:
The first commit doesn't appear to be associated with your GH account?

oliver-sanders (approved):

wxtim:
> The first commit doesn't appear to be associated with your GH account?

I think that I've messed up the setup of NewSpiceVDI.

oliver-sanders:
You'll either need to add a mailmap entry for the new setup, or use `git rebase` to amend the commit to change the author, otherwise the check-contributors action will fail when next run.

wxtim:
> You'll either need to add a mailmap entry for the new setup, or use `git rebase` to amend the commit to change the author, otherwise the check-contributors action will fail when next run.

Fixed - I'm not adding a `metoffic.gov.uk` email address to my history.

cylc/flow/task_events_mgr.py
MetRonnie: (Duplication resolved by changing the logged message)

changes.d/6169.fix
hjoliver, line 1: The task does change state in this case, but to waiting rather than failed.

hjoliver:
@wxtim - we need to choose between this PR and mine #6401 (I made a duplicate without realizing you'd already worked on it, sorry).

Both work, and I quite like your approach, but mine is simpler (a one-liner) and I think the is more consistent between submission and execution failure. 

```cylc
[scheduling]
    [[graph]]
        R1 = """
           a & b
        """
[runtime]
    [[a]]
        script = """
            cylc broadcast -n a -s "script = true" $CYLC_WORKFLOW_ID
            cylc broadcast -n b -s "platform = " $CYLC_WORKFLOW_ID
            false
        """
        execution retry delays = PT5S
    [[b]]
        platform = fake
        submission retry delays = PT5S
```

Log comparison (left me, right you):

shot

So my feeling is, we should use my branch, but cherry-pick your integration test to it.  Do you agree?

MetRonnie:
@hjoliver The CRITICAL level is probably too much though? Surely WARNING is the right level?

hjoliver:
> @hjoliver The CRITICAL level is probably too much though? Surely WARNING is the right level?

Perhaps, but @wxtim's approach still leaves submit-fail (with a retry) at CRITICAL - hence my consistency comment above.

wxtim:
I think that the correct level is error.

@hjoliver  - does your PR fall victim to any of Oliver's comments from https://github.com/cylc/cylc-flow/pull/6169#pullrequestreview-2141436538?

MetRonnie:
If execution/submission retry delays are configured, then execution/submission failures (respectively) are expected to occur. Therefore it is not a CRITICAL message to log. Only if the retries are exhausted should it be a CRITICAL level message?

hjoliver:
> If execution/submission retry delays are configured, then execution/submission failures (respectively) are expected to occur. Therefore it is not a CRITICAL message to log. Only if the retries are exhausted should it be a CRITICAL level message?

I don't disagree with that, but it was kinda off-topic for this Issue - which is about not hiding the job failure from the log - unless we introduce a jarring inconsistency between the way submission and execution failures are logged.

But OK, if we want to two kill two birds with one stone, let's look at unhiding the job failure AND changing the level of both kinds of failure in one go...

I agree with @wxtim 's assertion that the correct level (for both) is ERROR.

hjoliver:
> let's look at unhiding the job failure AND changing the level of both kinds of failure at once, to maintain consistency...

In that case, I would go with your approach @wxtim - but with some tweaks: 

1. don't log failure messages at all in the early `_process_message_check` method (where the demotion to DEBUG is currently done) 

2. instead log for both retries and not-retries in `_process_message_failed` where we already have the right code blocks:
 ```python
 if retries:
    LOG.error(message)
    ...
else:
    LOG.critical(message)
    ...
```

3. and do exactly the same for job submit failure, in `process_message_submit_failed`

**Comment by hjoliver** [comment]
@wxtim - is this ready to go again, in light of the above discussion?

**Comment by wxtim** [comment]
> @wxtim - is this ready to go again, in light of the above discussion?

In my head, yes, but I see that there a load of test failures. Will draft now and undraft when ready.

**Comment by wxtim** [comment]
These test failures were caused by a slight change in the nature of the message caused by moving it: By the time the code reaches `_process_message_failed` some of the info required for the original error messages (the state) is not available, or has changed.

File: cylc/flow/task_events_mgr.py [hunk 6]
```diff
@@ -1327,12 +1338,18 @@ def _process_message_failed(self, itask, event_time, message, forced):
                 self.data_store_mgr.delta_task_output(
                     itask, TASK_OUTPUT_FAILED)
                 self.data_store_mgr.delta_task_state(itask)
+            LOG.error(
+                f'[{itask}] {full_message or self.EVENT_FAILED} - '
```
hjoliver, line 1342:
```suggestion
                f'[{itask}] {full_message or self.EVENT_FAILED}'
```

hjoliver (approved):
One small typo found.

hjoliver:
@wxtim this seems to have diverged a bit from what I thought was agreed [above](https://github.com/cylc/cylc-flow/pull/6169#issuecomment-2434112911), which was:
- log the job (or submit) failure as ERROR whether there are retries or not
- then log the retry message at WARNING level, if there are retries

Now, if there are retries we only get the retry warning.

tests/functional/reload/25-xtriggers.t
```diff
@@ -42,7 +42,7 @@ init_workflow "${TEST_NAME_BASE}" <<'__FLOW_CONFIG__'
     [[reload]]
         script = """
             # wait for "broken" to fail
-            cylc__job__poll_grep_workflow_log -E '1/broken/01.* \(received\)failed/ERR'
+            cylc__job__poll_grep_workflow_log -E '1/broken.*failed/ERR'
```
MetRonnie, line 45: Suggest keeping the job number here
```suggestion
            cylc__job__poll_grep_workflow_log -E '1/broken/01.*failed/ERR'
```

tests/functional/reload/25-xtriggers.t
```diff
@@ -63,7 +63,7 @@ workflow_run_ok "${TEST_NAME_BASE}-run" cylc play "${WORKFLOW_NAME}" --no-detach
 log_scan "${TEST_NAME_BASE}-scan" \
     "$(cylc cat-log -m p "${WORKFLOW_NAME}")" \
     1 1 \
-    '1/broken.* (received)failed/ERR'
+    '1/broken.*failed/ERR'
```
MetRonnie, line 66:
```suggestion
    '1/broken/01.*failed/ERR'
```

tests/integration/test_task_events_mgr.py
```diff
@@ -170,3 +172,44 @@ async def test__always_insert_task_job(
...
+    """
+
+    async with start(one) as LOG:
+        fail_once = one.pool.get_tasks()[0]
+        # Add retry timers:
+        one.task_job_mgr._set_retry_timers(
+            fail_once, {
+                'execution retry delays': [1],
+                'submission retry delays': [1]
+            })
```
MetRonnie, line 191: Less fragile if this was just set in the workflow config for the test?

tests/integration/test_task_events_mgr.py
```diff
@@ -170,3 +172,44 @@ async def test__always_insert_task_job(
...
+
+        # Process submit failed message with and without retries:
+        one.task_events_mgr._process_message_submit_failed(
+            fail_once, None, 1, False)
+        last_record = LOG.records[-1]
+        assert last_record.levelno == logging.WARNING
+        assert '1/one:waiting(queued)' in last_record.message
```
MetRonnie, line 198: Should this not be testing for a "retrying" message?

MetRonnie (commented):
The execution retry log message is now:
```
WARNING - [1/foo:waiting] failed/ERR - retrying in ...
```
Whereas the submission retry log message is still just:
```
WARNING - [1/foo:waiting] retrying in ...
```
I think it would make sense to update this too.

tests/integration/test_task_events_mgr.py
wxtim: Does it really make any difference? - the aim was to avoid fiddling with `one`. Can do if you insist.

tests/integration/test_task_events_mgr.py
wxtim: It _is_ the retrying message, but not the bit of it which says retrying. 🙄

tests/integration/test_task_events_mgr.py
MetRonnie: Surely creating a workflow config with these retry delays set is not any more involved than fiddling the internal retry timers?

hjoliver:
Here's a screenshot comparing two execution failures (one retry) with two submission failures (one retry):

frap-2

Contrary to @MetRonnie 's suggestion (I think) you just need to get rid of the `failed/ERR` bit in the execution retry message. That information is already logged as the error. The warning is just about the resulting retry.

tests/integration/test_task_events_mgr.py
wxtim: Not going to argue. Will change.

MetRonnie:
Or have "execution retrying in..." and "submission retrying in..."?

hjoliver:
> Or have "execution retrying in..." and "submission retrying in..."?

Not really necessary because both entail a full "retry" (i.e., submit the job, then execute it).

hjoliver:
Still waiting on my suggestion [above](https://github.com/cylc/cylc-flow/pull/6169#issuecomment-2576592384), I think. In case it got a bit lost in the log lines:

```console
# submission retry (good)
ERROR - [1/foo/01:preparing] submission failed
INFO - [1/foo/01:preparing] => waiting
WARNING - [1/foo:waiting] retrying in PT5S (after 2025-01-09T11:08:20+13:00)
```

```console
# execution retry (needs tweak)
ERROR - [1/foo/01:running] failed/ERR
INFO - [1/foo/01:running] => waiting
WARNING - [1/foo:waiting] failed/ERR - retrying in PT5S (after 2025-01-09T11:05:49+13:00)
```

(Pretty minor, but there's no need to double-log the `failed/ERR` bit).

wxtim:
> Still waiting on my suggestion https://github.com/cylc/cylc-flow/pull/6169#issuecomment-2576592384, I think. In case it got a bit lost in the log lines:

Done, but not pushed, because I was was also looking at Ronnie's Comments about the tests.

tests/integration/test_task_events_mgr.py
wxtim: Went away and had a look at it - essentially this is more unit-testy than integration test-like: If you set these timers in the config you still need to run this function here to set the retry timers. So I think that I'll leave it.

oliver-sanders:
This PR has stalled, what needs to be done?

wxtim:
> This PR has stalled, what needs to be done?

I thought I'd responded to Hilary's comments. Will double check that I have and poke him. Also rebase.

cylc/flow/task_events_mgr.py
```diff
@@ -1416,21 +1427,22 @@ def _process_message_submit_failed(
         self,
         itask: 'TaskProxy',
         event_time: Optional[str],
+        submit_num: int,
```
hjoliver, line 1430:
```suggestion
```

hjoliver:
(@wxtim - note I committed a small fix to your branch - it was causing all the tests to fail)

hjoliver (approved):
Looks good now 🎉

hjoliver:
This could go to 8.4.1 IMO - a quick review at this point, in spite of the rather long history!

MetRonnie:
The recent merge commit contains `[skip ci]` in the body so the tests have not run. Please can you redo the merge and this time leave the body empty.

oliver-sanders (approved):

PR #6175, opened 2024-06-27
workflow_state: reject invalid configurations

### LINKED ISSUE #6157: workflow-state: handle submitted as a message/trigger not status
The workflow-state command defaults to polling task status rather than task trigger.

The CLI upgrade advise tells you to upgrade an existing message poller into a status poller:

```console
$ cylc workflow-state generic --point 20191209T1200Z --task bar --message=submitted
WARNING - --task, --status, --message, --output, --point, --task-point are deprecated. Please use an ID:
    generic//20191209T1200Z/bar:submitted
```

This is inconsistent with the xtrigger upgrade advice which tells you to use `is_message=True`:

```
        workflow_state(message=submitted, point=1234, task=foo, workflow=ab) -->
    workflow_state(workflow_task_id=ab//1234/foo:submitted, is_message=True)
```

The CLI advice becomes a problem in the case of the `submitted` output. Because `submitted` is a status, it defaults to status polling. However, the submitted status is transitory so this test is not safe.

Suggest:
* Changing the CLI upgrade advice to match the xtrigger advice.
* Raising a warning if workflow status polling is configured for non-final statuses.
* Document trigger polling as the preferred approach.

oliver-sanders:
* Do not allow users to poll for transient statuses.
* Reject invalid task states.
* Reject polling waiting and preparing tasks (not reliably pollable).
* Closes #6157

(minor bug as upgrade advice encouraged users to adopt invalid configurations which were not subsequently rejected)

**Check List**

### DIFF
--- a/cylc/flow/scripts/workflow_state.py
+++ b/cylc/flow/scripts/workflow_state.py
@@ -33,8 +33,8 @@ If the database does not exist at first, polls are consumed waiting for it
 so you can start checking before the target workflow is started.
 
 Legacy (pre-8.3.0) options are supported, but deprecated, for existing scripts:
-    cylc workflow-state --task=NAME --point=CYCLE --status=STATUS
-        --output=MESSAGE --message=MESSAGE --task-point WORKFLOW
+  cylc workflow-state --task=NAME --point=CYCLE --status=STATUS
+      --output=MESSAGE --message=MESSAGE --task-point WORKFLOW
 (Note from 8.0 until 8.3.0 --output and --message both match task messages).
 
 In "cycle/task:selector" the selector will match task statuses, unless:
@@ -55,24 +55,23 @@ Note tasks get recorded in the DB once they enter the active window (n=0).
 
 Flow numbers are only printed for flow numbers > 1.
 
-USE IN TASK SCRIPTING:
+Use in task scripting:
   - To poll a task at the same cycle point in another workflow, just use
     $CYLC_TASK_CYCLE_POINT in the ID.
   - To poll a task at an offset cycle point, use the --offset option to
     have Cylc do the datetime arithmetic for you.
   - However, see also the workflow_state xtrigger for this use case.
 
-WARNINGS:
- - Typos in the workflow or task ID will result in fruitless polling.
- - To avoid missing transient states ("submitted", "running") poll for the
-   corresponding output trigger instead ("submitted", "started").
- - Cycle points are auto-converted to the DB point format (and UTC mode).
- - Task outputs manually completed by "cylc set" have "(force-completed)"
-   recorded as the task message in the DB, so it is best to query trigger
-   names, not messages, unless specifically interested in forced outputs.
+Warnings:
+  - Typos in the workflow or task ID will result in fruitless polling.
+  - To avoid missing transient states ("submitted", "running") poll for the
+    corresponding output trigger instead ("submitted", "started").
+  - Cycle points are auto-converted to the DB point format (and UTC mode).
+  - Task outputs manually completed by "cylc set" have "(force-completed)"
+    recorded as the task message in the DB, so it is best to query trigger
+    names, not messages, unless specifically interested in forced outputs.
 
 Examples:
-
   # Print the status of all tasks in WORKFLOW:
   $ cylc workflow-state WORKFLOW
 
@@ -115,7 +114,11 @@ from cylc.flow.command_polling import Poller
 from cylc.flow.dbstatecheck import CylcWorkflowDBChecker
 from cylc.flow.terminal import cli_function
 from cylc.flow.workflow_files import infer_latest_run_from_id
-from cylc.flow.task_state import TASK_STATUSES_ORDERED
+from cylc.flow.task_state import (
+    TASK_STATUSES_ORDERED,
+    TASK_STATUSES_FINAL,
+    TASK_STATUSES_ALL,
+)
 
 if TYPE_CHECKING:
     from optparse import Values
@@ -363,7 +366,6 @@ def get_option_parser() -> COP:
 
 @cli_function(get_option_parser, remove_opts=["--db"])
 def main(parser: COP, options: 'Values', *ids: str) -> None:
-
     # Note it would be cleaner to use 'id_cli.parse_ids()' here to get the
     # workflow ID and tokens, but that function infers run number and fails
     # if the workflow is not installed yet. We want to be able to start polling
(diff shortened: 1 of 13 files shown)


tests/flakyfunctional/xtriggers/01-workflow_state/flow.cylc
```diff
@@ -8,7 +8,7 @@
     initial cycle point = 2011
     final cycle point = 2016
     [[xtriggers]]
-        upstream = workflow_state("{{UPSTREAM}}//%(point)s/foo:data_ready"):PT1S
+        upstream = workflow_state("{{UPSTREAM}}//%(point)s/foo:data_ready", is_trigger=True):PT1S
```
oliver-sanders, line 11: Behaviour change.

Turned off the implicit, "this isn't a status so handle it as a trigger" behaviour for xtriggers to make them explicit for safety.

cylc/flow/scripts/workflow_state.py
```diff
@@ -427,6 +429,15 @@ def main(parser: COP, options: 'Values', *ids: str) -> None:
...
         else:
             msg += id_.replace(options.depr_point, "$CYLC_TASK_CYCLE_POINT")
+
+        if (
+            options.depr_status
+            and options.depr_status in TASK_STATUSES_ALL
+            and options.depr_status not in TASK_STATUSES_FINAL
+        ):
+            # polling for non-final task stauses is flaky
+            msg += 'and the --triggers option'
```
MetRonnie, line 439:
```suggestion
            msg += ' and the --triggers option'
```

MetRonnie: Although arguably this is unnecessary as it is adequately covered by the `InputError`

MetRonnie:
Ack, just noticed this regression:

Cylc 8.2.7:
```
$ cylc workflow-state Murphid -t foo -p 2001 -S catacomb --max-polls 1
InputError: invalid status 'catacomb'
```

Cylc 8.3.1 and this branch:
```
$ cylc workflow-state Murphid -t foo -p 2001 -S catacomb --max-polls 1
WARNING - --task, --status, --message, --output, --point, --task-point are deprecated. Please use an ID: Murphid//2001/foo:catacomb
ERROR - failed after 1 polls
```

oliver-sanders:
> Ack, just noticed this regression:

Just with the deprecated `-S` presumably, easy fix.

cylc/flow/scripts/workflow_state.py
```diff
@@ -427,6 +429,15 @@ def main(parser: COP, options: 'Values', *ids: str) -> None:
             msg += id_
         else:
             msg += id_.replace(options.depr_point, "$CYLC_TASK_CYCLE_POINT")
+
+        if (
+            options.depr_status
+            and options.depr_status in TASK_STATUSES_ALL
+            and options.depr_status not in TASK_STATUSES_FINAL
+        ):
+            # polling for non-final task stauses is flaky
```
hjoliver, line 438:
```suggestion
            # polling for non-final task statuses is flaky
```

hjoliver (approved):
My bad, but need `--trigger` and `--message` --> `--triggers` and `--messages` here: https://github.com/cylc/cylc-flow/pull/6175/files#diff-16e1252035a0b4a9605f8cf4f45f5d3c4df8d90923b480137d43166beb734113R25-R26|

hjoliver (approved):
https://github.com/cylc/cylc-flow/pull/6175/files#diff-16e1252035a0b4a9605f8cf4f45f5d3c4df8d90923b480137d43166beb734113R25-R26

My bad: need to pluralize `--trigger` and `--message` here

oliver-sanders:
> My bad, but need --trigger and --message --> --triggers and --messages here: [https://github.com/cylc/cylc-flow/pull/6175/files#diff-16e1252035a0b4a9605f8cf4f45f5d3c4df8d90923b480137d43166beb734113R25-R26|](https://github.com/cylc/cylc-flow/pull/6175/files#diff-16e1252035a0b4a9605f8cf4f45f5d3c4df8d90923b480137d43166beb734113R25-R26%7C)

Link appears to be duff, where was is pointing?

changes.d/fix.6175.md
MetRonnie, line 1: Should be `6175.fix.md`

(Using `towncrier create` prevents this mistake)

tests/flakyfunctional/xtriggers/01-workflow_state/flow.cylc
MetRonnie: Why do this for the xtrigger but not the CLI?

oliver-sanders:
> Why do this for the xtrigger but not the CLI?

Back compatibility.

I would prefer the CLI to work this way, and maybe we could make that change, but it would be an embarrassing reversal between 8.3.0 and 8.3.3.

WDYT?

MetRonnie:
Would it be worth having a deprecation warning until 8.4 and then making it an InputError (either for the CLI or both CLI and xtrigger)? Although I don't think it looks great, not sure it's worse than having the xtrigger already an InputError and nothing for the CLI.

oliver-sanders:
If Hillary is ok with this I'm more happy to change the CLI behaviour.

Note that xtriggers are validated as part of workflow startup, whereas the CLI command is not validated until it is run, so it is easier to push xtrigger interface changes than.

hjoliver:
> > My bad, but need --trigger and --message --> --triggers and --messages here: [https://github.com/cylc/cylc-flow/pull/6175/files#diff-16e1252035a0b4a9605f8cf4f45f5d3c4df8d90923b480137d43166beb734113R25-R26|](https://github.com/cylc/cylc-flow/pull/6175/files#diff-16e1252035a0b4a9605f8cf4f45f5d3c4df8d90923b480137d43166beb734113R25-R26%7C)
> 
> Link appears to be duff, where was is pointing?

Branch probably got rebased. ...

hjoliver:
> If Hillary is ok with this I'm more happy to change the CLI behaviour.

I'm happy with that.

Note that I considered this during the main workflow-state overhaul PR, but decided against it because (a) historically we defaulted to polling for status; and (b) the command and xtrigger names also suggest that.  But if y'all are happy with the change, let's make it more robust.

changes.d/fix.6175.md
oliver-sanders: Must have followed the pattern of the [other entries](https://github.com/cylc/cylc-flow/blob/d7540aa62ed8a67f32f21508b7143ef10c03c13c/changes.d/fix.5924.md).

(`towncrier create` is too much faff)

tests/flakyfunctional/xtriggers/01-workflow_state/flow.cylc
oliver-sanders: The same reason this feature was originally implemented, back-compat.

But we now seem to be ok with dropping this (which is much nicer).

changes.d/fix.6175.md
MetRonnie: In towncrier v24.7 there is now an interactive mode if you run `towncrier create` without any args, seems like even less effort than manually creating a fragment file to me

MetRonnie:
(Test failures)

MetRonnie (approved):

PR #6176, opened 2024-06-27
Ensure jobs are created if task is in waiting state

### LINKED ISSUE #6172: submit-failed tasks not showing
It has been reported that, following the instructions to reproduce https://github.com/cylc/cylc-flow/issues/6100, submit-failed tasks are not showing in the GUI/Tui.

We are probably missing a data store call (and if we're missing a data store call, we might be missing a database call too?).

wxtim:
Closes #6172 

Repro:
```
[scheduling]
    [[graph]]
        R1 = foo
[runtime]
    [[foo]]
        platform = borked
        submission retry delays = 5*PT10S
```
With `global.cylc`:
```
[platforms]
    [[borked]]
        hosts = bogus
```

**Check List**

### DIFF
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -1535,11 +1535,21 @@ class TaskEventsManager():
         else:
             job_conf = itask.jobs[-1]
 
+        # Job status should be task status unless task is awaiting a
+        # retry:
+        if itask.state.status == TASK_STATUS_WAITING and itask.try_timers:
+            job_status = (
+                TASK_STATUS_SUBMITTED if submit_status == 0
+                else TASK_STATUS_SUBMIT_FAILED
+            )
+        else:
+            job_status = itask.state.status
+
         # insert job into data store
         self.data_store_mgr.insert_job(
             itask.tdef.name,
             itask.point,
-            itask.state.status,
+            job_status,
             {
                 **job_conf,
                 # NOTE: the platform name may have changed since task
--- a/tests/integration/test_task_events_mgr.py
+++ b/tests/integration/test_task_events_mgr.py
@@ -14,11 +14,18 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from cylc.flow.task_events_mgr import TaskJobLogsRetrieveContext
-from cylc.flow.scheduler import Scheduler
-
+from itertools import product
 import logging
 from typing import Any as Fixture
+
+from cylc.flow.task_events_mgr import TaskJobLogsRetrieveContext
+from cylc.flow.scheduler import Scheduler
+from cylc.flow.data_store_mgr import (
+    JOBS,
+    TASK_STATUSES_ORDERED,
+    TASK_STATUS_WAITING,
+    TASK_STATUS_SUBMIT_FAILED,
+)
 
 
 async def test_process_job_logs_retrieval_warns_no_platform(
@@ -99,3 +106,63 @@ async def test__insert_task_job(flow, one_conf, scheduler, start, validate):
             i.submit_num for i
             in schd.data_store_mgr.added['jobs'].values()
         ] == [1, 2]
+
+
+async def test__always_insert_task_job(
+    flow, scheduler, mock_glbl_cfg, start, run
+):
+    """Insert Task Job _Always_ inserts a task into the data store.
+
+    Bug https://github.com/cylc/cylc-flow/issues/6172 was caused
+    by passing task state to data_store_mgr.insert_job: Where
+    a submission retry was in progress the task state would be
+    "waiting" which caused the data_store_mgr.insert_job
+    to return without adding the task to the data store.
+    """
+    global_config = """
+        [platforms]
+            [[broken1]]
+                hosts = no-such-host-1
+            [[broken2]]
+                hosts = no-such-host-2
+        [platform groups]
+            [[broken]]
+                platforms = broken1
+    """
+    mock_glbl_cfg('cylc.flow.platforms.glbl_cfg', global_config)
+
+    id_ = flow({
+        'scheduling': {'graph': {'R1': 'broken & broken2'}},
+        'runtime': {
+            'root': {'submission retry delays': 'PT10M'},
+            'broken': {'platform': 'broken'},
+            'broken2': {'platform': 'broken2'}
+        }
+    })
+
+    schd = scheduler(id_, run_mode='live')
+    schd.bad_hosts = {'no-such-host-1', 'no-such-host-2'}
+    async with start(schd):
+        schd.task_job_mgr.submit_task_jobs(
+            schd.workflow,
+            schd.pool.get_tasks(),
+            schd.server.curve_auth,
+            schd.server.client_pub_key_dir,
+            is_simulation=False
+        )
+
+        # Both tasks are in a waiting state:
+        assert all(
+            i.state.status == TASK_STATUS_WAITING
+            for i in schd.pool.get_tasks())
+
+        # Both tasks have updated the data store with info
+        # about a failed job:
+        updates = {
+            k.split('//')[-1]: v.state
+            for k, v in schd.data_store_mgr.updated[JOBS].items()
+        }
+        assert updates == {
+            '1/broken/01': 'submit-failed',
+            '1/broken2/01': 'submit-failed'
+        }
--- a/tests/integration/test_task_events_mgr.py
+++ b/tests/integration/test_task_events_mgr.py
@@ -118,6 +118,10 @@ async def test__always_insert_task_job(
     a submission retry was in progress the task state would be
     "waiting" which caused the data_store_mgr.insert_job
     to return without adding the task to the data store.
+    This is testing two different cases:
+
+    * Could not select host from platform
+    * Could not select host from platform group
     """
     global_config = """
         [platforms]
--- a/changes.d/6176.fix.md
+++ b/changes.d/6176.fix.md
@@ -1,0 +1,1 @@
+Fix bug where jobs which fail to submit are not shown in GUI/TUI if submission retries are set.


cylc/flow/data_store_mgr.py
```diff
@@ -1543,13 +1543,18 @@ def _apply_broadcasts_to_runtime(self, tokens, rtconfig):
             poverride(rtconfig, overrides, prepend=True)
         return rtconfig
 
-    def insert_job(self, name, cycle_point, status, job_conf):
+    def insert_job(
```
wxtim, line 1546: This method is only used by `TaskEventsMgr._insert_task_job`. This function looks like it is designed to take a job status, but `_insert_task_job` has been feeding it the task status. 

This may be ok if the task status is also a job status, but the job will not be added to the data store if the status is not a job status, because of the check at line 1578.

In that case, subsequent calls to `delta_job_status` fail, because the data store doesn't have a record of the job.

In any case, `delta_job_status` is always subsequently called, so it doesn't matter if we don't set the job status here.

cylc/flow/task_events_mgr.py
```diff
@@ -1536,10 +1536,11 @@ def _insert_task_job(
             job_conf = itask.jobs[-1]
 
         # insert job into data store
+        job_status = 'submitted' if submit_status == 0 else 'submit-failed'
         self.data_store_mgr.insert_job(
```
wxtim, line 1540: This method takes a job status, not a task status. If the task status is 'waiting' it returns None and doesn't register the job with the data store.

oliver-sanders:
Tui is just showing you what the data store contains (i.e. these Tui test failures are nothing to do with Tui).

tests/integration/tui/screenshots/test_auto_expansion.later-time.html
```diff
@@ -3,7 +3,7 @@
...
 <span style="color:#000000;background:#e5e5e5">      </span><span style="color:#000000;background:#e5e5e5">-</span><span style="color:#000000;background:#e5e5e5"> </span><span style="color:#000000;background:#e5e5e5">● 1                                                                     </span>
-<span style="color:#000000;background:#e5e5e5">         </span><span style="color:#000000;background:#e5e5e5"> </span><span style="color:#000000;background:#e5e5e5"> </span><span style="color:#000000;background:#e5e5e5">● b                                                                  </span>
+<span style="color:#000000;background:#e5e5e5">         </span><span style="color:#000000;background:#e5e5e5">+</span><span style="color:#000000;background:#e5e5e5"> </span><span style="color:#000000;background:#e5e5e5">● </span><span style="color:#00cd00;background:#e5e5e5">■</span><span style="color:#000000;background:#e5e5e5"> b                                                                </span>
```
oliver-sanders, line 6: This test has nothing to do with "submit-failed" jobs so these tests breaking reveals a bug in the PR.

In this test, we run `cylc set` to mark a task as succeeded. Before this change, the test correctly showed the succeeded task with no job. After this change there is a succeeded job in the data which never logically existed.

oliver-sanders (changes requested):

tests/integration/test_task_events_mgr.py
```diff
@@ -99,3 +101,39 @@ async def test__insert_task_job(flow, one_conf, scheduler, start, validate):
...
+
+    n.b. future debuggers may with to convert the for loop to
+    pytest parameterization for ease of debugging, but the test is
+    very much faster with the parameterization inside.
+    """
+    conf = {
+        'scheduler': {'allow implicit tasks': True},
+        'scheduling': {'graph': {'R1': 'skarloey'}},
+    }
+    id_ = flow(conf)
```
oliver-sanders, line 125: Can use the `one_conf` fixture already in the test sig:

```suggestion
    id_ = flow(one_conf)
```

tests/integration/test_task_events_mgr.py
```diff
@@ -99,3 +101,39 @@ async def test__insert_task_job(flow, one_conf, scheduler, start, validate):
...
+    schd = scheduler(id_)
+    async with start(schd):
+        # Set task to running:
+        itask = schd.pool.get_tasks()[0]
+        itask.state.status = 'waiting'
```
oliver-sanders, line 130: Either:

```suggestion
```

Or:

```suggestion
        assert itask.state.status = TASK_STATUS_WAITING
```

tests/integration/test_task_events_mgr.py
```diff
@@ -99,3 +101,39 @@ async def test__insert_task_job(flow, one_conf, scheduler, start, validate):
...
             in schd.data_store_mgr.added['jobs'].values()
         ] == [1, 2]
+
+
+async def test__always_insert_task_job(
+    flow, one_conf, scheduler, start, validate
+):
+    """Insert Task Job _Always_ inserts a task into the data store.
+
+    Bug https://github.com/cylc/cylc-flow/issues/6172 was caused
```
oliver-sanders, line 111: We should be able to go a bit further with this integration test and probe the code pathways which result in these submit-fails to ensure the use case behaves correctly.

E.G, I tested with this:

```ini
# global.cylc

[platforms]    
    [[broken1]]    
        hosts = no-such-host-1
        install target = localhost
    [[broken2]]    
        hosts = no-such-host-2
        install target = localhost
         
[platform groups]    
    [[broken]]    
        platforms = broken1  
```

And this:

```ini
# flow.cylc

[scheduling]
    [[graph]]
        R1 = broken & broken2

[runtime]
    [[root]]
        submission retry delays = PT10M
    [[broken]]
        platform = broken
    [[broken2]]
        platform = broken2  # note broken2 not broken1 to avoid sharing hosts (opens the door to interactions via bad-hosts)
```

I.E, something along the lines of

```python
# untested

async with start(schd):
    for itask in schd.task_job_mgr.submit_task_jobs(
        self.workflow,
        self.pool.get_tasks(),
        self.server.curve_auth,
        self.server.client_pub_key_dir,
        is_simulation=False
    ):
        # check the task pool status
        assert itask.state.status == TASK_STATUS_WAITING
        assert itask.state.outputs.is_completed(TASK_OUTPUT_SUBMIT_FAILED)

        # check the data store status (see tests/integration/test_data_store_mgr.py for inspiration)
        await schd.update_data_store()
        assert {tp.name: tp.jobs[0].state for tp in schd.data_store_mgr.updated[TASK_PROXIES]}
```

oliver-sanders (commented):
Makes sense, tested as working :+1:

oliver-sanders:
(rebase to pick up flake8 fixes on upstream/8.3.x)

tests/integration/test_task_events_mgr.py
wxtim: > assert itask.state.outputs.is_completed(TASK_OUTPUT_SUBMIT_FAILED)

Would you really expect this to be true if the task is waiting on retries?

oliver-sanders: Ach, no, can remove that line.

oliver-sanders: The main thing is to test the job statuses in the store.

tests/integration/test_task_events_mgr.py
```diff
@@ -99,3 +101,52 @@ async def test__insert_task_job(flow, one_conf, scheduler, start, validate):
...
+    schd.bad_hosts = {'no-such-host-1', 'no-such-host-2'}
+    async with start(schd):
+        for itask in schd.task_job_mgr.submit_task_jobs(
+            schd.workflow,
+            schd.pool.get_tasks(),
+            schd.server.curve_auth,
+            schd.server.client_pub_key_dir,
+            is_simulation=False
+        ):
+            assert itask.state.status == 'waiting'
```
oliver-sanders, line 147: Use the TASK_STATUS_WAITING constant.

tests/integration/test_task_events_mgr.py
```diff
@@ -99,3 +101,52 @@ async def test__insert_task_job(flow, one_conf, scheduler, start, validate):
...
+            [[broken]]
+                platforms = broken1, broken2
+    """
+    mock_glbl_cfg('cylc.flow.platforms.glbl_cfg', global_config)
+
+    id_ = flow({
+        'scheduling': {'graph': {'R1': 'broken'}},
+        'runtime': {
+            'root': {'submission retry delays': 'PT10M'},
+            'broken': {'platform': 'broken'},
```
oliver-sanders, line 133: May as well add a second task like in my example so that we are testing both platforms and platform groups (make sure there are no shared hosts to remove interactions, note a platform-group only requires one platform).

tests/integration/test_task_events_mgr.py
```diff
@@ -99,3 +101,52 @@ async def test__insert_task_job(flow, one_conf, scheduler, start, validate):
...
+            schd.workflow,
+            schd.pool.get_tasks(),
+            schd.server.curve_auth,
+            schd.server.client_pub_key_dir,
+            is_simulation=False
+        ):
+            assert itask.state.status == 'waiting'
+            schd.update_data_store()
+            update = [
+                i for i in schd.data_store_mgr.updated['jobs'].values()][0]
```
oliver-sanders, line 150: Use the JOBS constant.

tests/integration/test_task_events_mgr.py
```diff
@@ -99,3 +101,52 @@ async def test__insert_task_job(flow, one_conf, scheduler, start, validate):
...
+            assert update.id.endswith('1/broken/01')
+            assert update.state == 'submit-failed'
```
oliver-sanders, line 152: Use the SUBMIT_FAILED constant.

oliver-sanders (commented):
LGTM

tests/integration/test_task_events_mgr.py
oliver-sanders: Need to remove `broken2` from the `broken` platform group to decouple these two tasks (so they can't interact via bad-hosts)

cylc/flow/task_events_mgr.py
```diff
@@ -1535,11 +1535,18 @@ def _insert_task_job(
         else:
             job_conf = itask.jobs[-1]
 
+        # Job status should be task status unless task is awaiting a
+        # retry:
+        if itask.state.status == 'waiting' and itask.try_timers:
+            job_status = 'submitted' if submit_status == 0 else 'submit-failed'
```
MetRonnie, line 1541:
```suggestion
        if itask.state.status == TASK_STATUS_WAITING and itask.try_timers:
            job_status = (
                TASK_STATUS_SUBMITTED if submit_status == 0
                else TASK_STATUS_SUBMIT_FAILED
            )
```

tests/integration/test_task_events_mgr.py
```diff
@@ -99,3 +106,53 @@ async def test__insert_task_job(flow, one_conf, scheduler, start, validate):
...
+    """
+    mock_glbl_cfg('cylc.flow.platforms.glbl_cfg', global_config)
+
+    id_ = flow({
+        'scheduling': {'graph': {'R1': 'broken & broken2'}},
+        'runtime': {
+            'root': {'submission retry delays': 'PT10M'},
+            'broken': {'platform': 'broken'},
+            'broken2': {'platform': 'broken2'}
+        }
```
MetRonnie, line 140: Couldn't this be simplified to 1 task & platform?

MetRonnie (commented):
Tested manually

tests/integration/test_task_events_mgr.py
```diff
@@ -99,3 +106,53 @@ async def test__insert_task_job(flow, one_conf, scheduler, start, validate):
...
+    })
+
+    schd = scheduler(id_, run_mode='live')
+    schd.bad_hosts = {'no-such-host-1', 'no-such-host-2'}
```
oliver-sanders, line 144: May as well let this happen naturally?

```suggestion
```

tests/integration/test_task_events_mgr.py
```diff
@@ -99,3 +106,53 @@ async def test__insert_task_job(flow, one_conf, scheduler, start, validate):
...
+            schd.pool.get_tasks(),
+            schd.server.curve_auth,
+            schd.server.client_pub_key_dir,
+            is_simulation=False
+        ):
+            assert itask.state.status == TASK_STATUS_WAITING
+            schd.update_data_store()
+            update = [
+                i for i in schd.data_store_mgr.updated[JOBS].values()][0]
+            assert update.id.endswith('1/broken/01')
```
oliver-sanders, line 157: This proves that the test isn't checking the `broken2` task!

Try `[-1]` rather than `[0]` above?

tests/integration/test_task_events_mgr.py
MetRonnie: @wxtim could you add this as a comment?

tests/integration/test_task_events_mgr.py
wxtim: It's not proving anything whatever - that loop isn't being run!

tests/integration/test_task_events_mgr.py
wxtim: Then we have to run the submission script repeatedly....

oliver-sanders: Ach, I think I understand why, it's because we only find out that a host is "bad" when we try to remote-init a remote platform or submit a local job.

That doesn't happen for this test because it's an integration test (the subprocpool isn't running).

Ok, please ignore my comment.

cylc/flow/task_events_mgr.py
```diff
@@ -1544,6 +1544,7 @@ def _insert_task_job(
             )
         else:
             job_status = itask.state.status
+        # breakpoint(header=f"=== {job_status} ===")
```
oliver-sanders, line 1547: debug

tests/integration/test_task_events_mgr.py
MetRonnie: Still need to add code comment

wxtim: It's in the docstring.

tests/integration/test_task_events_mgr.py
```diff
@@ -99,3 +106,63 @@ async def test__insert_task_job(flow, one_conf, scheduler, start, validate):
...
+    flow, scheduler, mock_glbl_cfg, start, run
+):
+    """Insert Task Job _Always_ inserts a task into the data store.
+
+    Bug https://github.com/cylc/cylc-flow/issues/6172 was caused
+    by passing task state to data_store_mgr.insert_job: Where
+    a submission retry was in progress the task state would be
+    "waiting" which caused the data_store_mgr.insert_job
+    to return without adding the task to the data store.
+    """
```
wxtim, line 121:
```suggestion
    This is testing two different cases:

    * Could not select host from platform
    * Could not select host from platform group
    """
```

oliver-sanders (approved):

MetRonnie (approved):

PR #6178, opened 2024-06-28
tui: prevent hang on shutdown

oliver-sanders:
It has been reported that Tui can hang on shutdown.

The easiest way to reproduce this is to press "q" as soon as starting Tui. Sometimes this can cause it to hang.

```
$ cylc vip <workflow>
$ cylc tui <workflow>
q
```

This PR adds a timeout on the updater shutdown to prevent any hanging operation stopping Tui from exiting.

This isn't really the "right" way to solve the problem, getting the updater to exit cleanly in all circumstances would be nicer. However, it's a reasonable fail safe and since the updater is a read-only process (i.e. the only "resources" it's managing are Cylc clients which have timeouts built in) there are no ill effects of terminating it.

**Check List**

- [ ] Not reasonably testable.

### DIFF
--- a/changes.d/fix.6178.md
+++ b/changes.d/fix.6178.md
@@ -1,0 +1,1 @@
+Fix an issue where Tui could hang when closing.
--- a/cylc/flow/tui/app.py
+++ b/cylc/flow/tui/app.py
@@ -216,7 +216,10 @@ def updater_subproc(filters, client_timeout):
         yield updater
     finally:
         updater.terminate()
-        p.join()
+        p.join(4)  # timeout of 4 seconds
+        if p.exitcode is None:
+            # updater did not exit within timeout -> kill it
+            p.terminate()
 
 
 class TuiApp:


wxtim (approved):
I have struggled to repeat the original bug, but this looks ok.

oliver-sanders:
@wxtim, the original bug isn't reproducing as easily as it originally did. To simulate this bug, try changing `break` to `continue` in this code:

https://github.com/cylc/cylc-flow/blob/67a6ba5f429efbf91c1610fba6bf602941a369fe/cylc/flow/tui/updater.py#L148-L159

That will prevent the updater from shutting down. Before this commit, Tui should hang indefinitely, after this PR it should exit after 4 seconds (due to the timeout).

wxtim:
> try changing break to continue in this code

Looks like I can now repeat the problem. Pleasingly this branch fixes it!

PR #6181, opened 2024-06-28
`cylc stop --now`: fix flaky test & clarify docs

MetRonnie:
**Check List**

- [x] No dependency changes 
- [x] Tests are updated
- [x] No docs needed

### DIFF
--- a/tests/functional/shutdown/08-now1/flow.cylc
+++ b/tests/functional/shutdown/08-now1/flow.cylc
@@ -3,7 +3,7 @@
         abort on stall timeout = True
         stall timeout = PT0S
         abort on inactivity timeout = True
-        inactivity timeout = PT3M
+        inactivity timeout = PT1M
 
 [scheduling]
     [[graph]]
@@ -14,6 +14,5 @@
         script = cylc__job__wait_cylc_message_started; cylc stop --now "${CYLC_WORKFLOW_ID}"
         [[[events]]]
             started handlers = sleep 10 && echo 'Hello %(id)s %(event)s'
-            succeeded handlers = echo 'Well done %(id)s %(event)s'
     [[t2]]
         script = true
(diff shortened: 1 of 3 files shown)


tests/functional/shutdown/08-now1/flow.cylc
```diff
@@ -14,6 +14,5 @@
         script = cylc__job__wait_cylc_message_started; cylc stop --now "${CYLC_WORKFLOW_ID}"
         [[[events]]]
             started handlers = sleep 10 && echo 'Hello %(id)s %(event)s'
-            succeeded handlers = echo 'Well done %(id)s %(event)s'
```
MetRonnie, line 17: The `succeeded` handler is not guaranteed to run because the `succeeded` message is not guaranteed to be received by the scheduler before it shuts down. So I have removed it from the test.

oliver-sanders: The test says:

```
Test "cylc stop --now" will wait for event handler.
```

So removing this kinda defeats the object of the test.

There is a way to ensure that the succeeded message is recieved however,

```ini
script = """
cylc message -- succeeded
cylc workflow-state "$CYLC_WORKFLOW_ID//$CYLC_TASK_CYCLE_POINT/$CYLC_TASK_NAME:succeeded"
sleep 1  # give the scheduler time to process the message
cylc stop --now "${CYLC_WORKFLOW_ID}
"""
```

MetRonnie: It's still testing that the long-running `started` handler finishes before shutdown. I'm not sure the `succeeded` handler adds any value to the test.

I tried adding `cylc message -- succeeded` but this results in task `t2` starting, which fails the test

oliver-sanders: > I tried adding cylc message -- succeeded but this results in task t2 starting

We could make that work by using an optional output.

> It's still testing that the long-running started

Gotcha, as long as there is one event handler.

The 10s sleep is a little crude (might be better to poll for the workflow "stopping" state or shutdown message), but that's another matter.

oliver-sanders (approved):

PR #6183, opened 2024-06-28
Tests: fix `contains_ok` UTF-8 issue

MetRonnie:
The `comm` command can't tell the difference between ✓ & ⨯ when `$LANG` is set to a UTF-8 setting e.g. `en_US.UTF-8`

See https://stackoverflow.com/q/78678882/3217306

---

Edit: Ah, interesting!

```console
$ cat c.txt
✓
⨯
✓
⨯

$ LANG=en_US.UTF-8 sort c.txt
✓
⨯
✓
⨯

$ LANG=C sort c.txt
✓
✓
⨯
⨯
```

### DIFF
--- a/tests/functional/lib/bash/test_header
+++ b/tests/functional/lib/bash/test_header
@@ -386,7 +386,7 @@ contains_ok() {
     local FILE_CONTROL="${2:--}"
     local TEST_NAME
     TEST_NAME="$(basename "${FILE_TEST}")-contains-ok"
-    comm -13 <(sort "${FILE_TEST}") <(sort "${FILE_CONTROL}") \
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Can we use a UTF-8 variant of `LANG=C`.

MetRonnie:
I'm not sure there is any such thing as `C.UTF-8`. I checked `locale -a` and couldn't find it in there. But it seems tied to UTF-8, e.g. `de_DE` works but `de_DE.UTF-8` doesn't

oliver-sanders:
I've definitely encountered `C.UTF-8`, however, it looks like not all systems support it (`locale -a` only lists supported locales?)

oliver-sanders (approved):

Issue #6066, opened 2024-04-12
suicide triggers broken


oliver-sanders:
This example works (tasks a and c automatically removed, workflow shuts down) in Cylc 7 but stalls in Cylc 8:

```ini
# suite.rc
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    [[dependencies]]
        [[[R1]]]
            graph = """
                a & b => c
                a:fail => !a & !c
            """

[runtime]
    [[a]]
        script = false
    [[b,c]]
```

This example also stalls in Cylc 8:

```ini
# flow.cylc
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    [[dependencies]]
        [[[R1]]]
            graph = """
                a? & b => c
                a:fail? => !a & !c
            """

[runtime]
    [[a]]
        script = false
    [[b,c]]
```

Slightly related to https://github.com/cylc/cylc-flow/issues/5729 as this covers some or all of the remaining use cases for suicide triggers.

hjoliver:
On 8.2.x, this diff is enough to make it work, but not on master.
```diff
-    [[b,c]]
+    [[b,c]]
+        script = sleep 2
```
Looks like something to do with suicide interfering with the record of previous spawning....

hjoliver:
The two cases look almost identical until:
- "1/c already spawned in flow 1" 👍 
- OR 1/c gets spawned again, which causes the problem 👎 

![image](https://github.com/cylc/cylc-flow/assets/2362137/4ce7449c-8538-4d60-8eea-3103161a0e9f)

hjoliver:
If `b` succeeds before `a` fails:
- `b` spawns `c`, then `a` removes `c` 👍 

If `a` fails before `b` succeeds:
- `a` spawns and removes `c`, then `b` spawns `c` again (causing a stall) 👎

PR #6186, opened 2024-07-01
`cylc set` bug fix for new flows

hjoliver:
Damn it, found a bug in 8.3.0 for `cylc set --flow=new` on future tasks: the task_outputs table won't be updated because of "WHERE flow is n" in the DB statement.

From: https://github.com/hjoliver/ecox-interventions?tab=readme-ov-file#scenario-2-rerun-some-tasks-from-upstream-of-a-failed-task

Example:
```ini
[task parameters]
    m = 1..3
[scheduling]
    [[graph]]
        R1 = "a => b => c<m> => d"
[runtime]
    [[root]]
        pre-script = sleep 5
    [[a, b, d]]
    [[c<m>]]
    [[c<m=2>]]
        script = """
            if (( CYLC_TASK_SUBMIT_NUMBER == 1 )); then
                false
            fi
        """
```
The scenario:
- `1/c_m2` fails, stalling the workflow due to a corrupted output from `1/b`
- want to rerun `1/b` and `1/c_m2`, but not `1/c_m1, m3`, then continue to `1/d` and finish

Method: 
- run a new flow (2) from `1/b` after setting `1/c_m1, m3` to succeeded in the new flow

1. `cylc set test //1/c_m1 //1/c_m2 --flow=2`
2. `cylc trigger test//1/b --flow=2`

On master this reruns `1/c_m1` and `1/c_m3` because the `set` outputs don't end up in the DB.

**Check List**

### DIFF
--- a/cylc/flow/scripts/set.py
+++ b/cylc/flow/scripts/set.py
@@ -177,7 +177,7 @@ def get_option_parser() -> COP:
     return parser
 
 
-def validate_tokens(tokens_list: Tuple['Tokens']) -> None:
+def validate_tokens(tokens_list: Tuple['Tokens', ...]) -> None:
     """Check the cycles/tasks provided.
 
     This checks that cycle/task selectors have not been provided in the IDs.
@@ -214,7 +214,7 @@ def validate_tokens(tokens_list: Tuple['Tokens']) -> None:
 async def run(
     options: 'Values',
     workflow_id: str,
-    *tokens_list
+    *tokens_list: 'Tokens'
 ):
     validate_tokens(tokens_list)
 
--- a/cylc/flow/scripts/set.py
+++ b/cylc/flow/scripts/set.py
@@ -90,7 +90,7 @@ Examples:
 
 from functools import partial
 import sys
-from typing import Tuple, TYPE_CHECKING
+from typing import Iterable, TYPE_CHECKING
 
 from cylc.flow.exceptions import InputError
 from cylc.flow.network.client_factory import get_client
@@ -177,7 +177,7 @@ def get_option_parser() -> COP:
     return parser
 
 
-def validate_tokens(tokens_list: Tuple['Tokens', ...]) -> None:
+def validate_tokens(tokens_list: Iterable['Tokens']) -> None:
     """Check the cycles/tasks provided.
 
     This checks that cycle/task selectors have not been provided in the IDs.
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1578,8 +1578,8 @@ class TaskPool:
 
     def _get_task_history(
         self, name: str, point: 'PointBase', flow_nums: Set[int]
-    ) -> Tuple[bool, int, str, bool]:
-        """Get history of previous submits for this task.
+    ) -> Tuple[int, Optional[str], bool]:
+        """Get submit_num, status, flow_wait for point/name in flow_nums.
 
         Args:
             name: task name
@@ -1587,41 +1587,33 @@ class TaskPool:
             flow_nums: task flow numbers
 
         Returns:
-            never_spawned: if task never spawned before
-            submit_num: submit number of previous submit
-            prev_status: task status of previous sumbit
-            prev_flow_wait: if previous submit was a flow-wait task
+           (submit_num, status, flow_wait)
+           If no matching history, status will be None
 
         """
+        submit_num: int = 0
+        status: Optional[str] = None
+        flow_wait = False
+
         info = self.workflow_db_mgr.pri_dao.select_prev_instances(
             name, str(point)
         )
-        try:
-            submit_num: int = max(s[0] for s in info)
-        except ValueError:
-            # never spawned in any flow
-            submit_num = 0
-            never_spawned = True
-        else:
-            never_spawned = False
-            # (submit_num could still be zero, if removed before submit)
+        with suppress(ValueError):
+            submit_num = max(s[0] for s in info)
 
-        prev_status: str = TASK_STATUS_WAITING
-        prev_flow_wait = False
-
-        for _snum, f_wait, old_fnums, status in info:
+        for _snum, f_wait, old_fnums, old_status in info:
             if set.intersection(flow_nums, old_fnums):
                 # matching flows
-                prev_status = status
-                prev_flow_wait = f_wait
-                if prev_status in TASK_STATUSES_FINAL:
+                status = old_status
+                flow_wait = f_wait
+                if status in TASK_STATUSES_FINAL:
                     # task finished
                     break
                 # Else continue: there may be multiple entries with flow
                 # overlap due to merges (they'll have have same snum and
                 # f_wait); keep going to find the finished one, if any.
 
-        return never_spawned, submit_num, prev_status, prev_flow_wait
+        return submit_num, status, flow_wait
 
     def _load_historical_outputs(self, itask: 'TaskProxy') -> None:
         """Load a task's historical outputs from the DB."""
@@ -1664,40 +1656,32 @@ class TaskPool:
         force: bool = False,
         flow_wait: bool = False,
     ) -> Optional[TaskProxy]:
-        """Return task proxy if not completed in this flow, or if forced.
+        """Return a new task proxy for the given flow if possible.
 
-        If finished previously with flow wait, just try to spawn children.
+        We need to hit the DB for:
+        - submit number
+        - task status
+        - flow-wait
+        - completed outputs (e.g. via "cylc set")
 
-        Note finished tasks may be incomplete, but we don't automatically
-        re-run incomplete tasks in the same flow.
+        If history records a final task status (for this flow):
+        - if not flow wait, don't spawn (return None)
+        - if flow wait, don't spawn (return None) but do spawn children
+        - if outputs are incomplete, don't auto-rerun it (return None)
 
-        For every task spawned, we need a DB lookup for submit number,
-        and flow-wait.
+        Otherwise, spawn the task and load any completed outputs.
 
         """
-        if not self.can_be_spawned(name, point):
-            return None
-
-        never_spawned, submit_num, prev_status, prev_flow_wait = (
+        submit_num, prev_status, prev_flow_wait = (
             self._get_task_history(name, point, flow_nums)
         )
 
-        spawned_but_not_submitted = False
-        if (
-            not never_spawned and
-            not prev_flow_wait and
-            submit_num == 0
-        ):
-            # Task previously spawned but never submitted, which implies:
-            #  - removed before completing any outputs
-            #  - or some outputs were set manually
-            spawned_but_not_submitted = True
-
+        # Create the task proxy with any completed outputs loaded.
         itask = self._get_task_proxy_db_outputs(
             point,
             self.config.get_taskdef(name),
             flow_nums,
-            status=prev_status,
+            status=prev_status or TASK_STATUS_WAITING,
             submit_num=submit_num,
             flow_wait=flow_wait,
         )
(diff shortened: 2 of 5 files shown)


wxtim:
Needs a rebase to pull in flake8 fixes.

wxtim:
I have created a crude integration/fnctional test at https://github.com/hjoliver/cylc-flow/pull/55

I think I can do better/faster, but that test works as a test.

hjoliver:
@wxtim - rebased and added an integration test.

My integration test is presumably similar to what you are working on?  Trouble is, it passes on master too!  I've run out of time to investigate further, for now.

wxtim:
> @wxtim - rebased and added an integration test.
> 
> My integration test is presumably similar to what you are working on? Trouble is, it passes on master too! I've run out of time to investigate further, for now.

Mine is a functional test using the integration test framework. Since you've written most of the test I'll leave the ball in your court I think. I have other 🐟 to  🥘 , not least a sim mode bug which will scupper skip mode too.

hjoliver:
(Update: a similar scenario works fine on master if the set and trigger happens for tasks that have never run before in any flow).

hjoliver:
Test tweaked and validated. It passes on master as an integration test for "the wrong reasons", which are not valid on this branch (it's to do with submit numbers being zero in integration tests - I can explain in detail if necessary). The same scenario does not pass on master as a functional test.

hjoliver:
Damn it, I seem to have broken two integration tests though. So not quite done.

cylc/flow/task_pool.py
```diff
@@ -1696,6 +1704,17 @@ def spawn_task(
...
             return None
 
+        if (
+            spawned_but_not_submitted and
+            not itask.state.outputs.get_completed_outputs()
+        ):
+            # spawned but never submitted, and has no completed outputs.
+            # This implies it was removed, rather than manually set.
+            # TODO: Detecting removal after completion of some outputs probably
+            # TODO: requires recording removal in the DB (set :expired maybe?).
```
hjoliver, line 1714: Beyond the scope of this PR. I'll post a note on Element, then raise an issue.

cylc/flow/scripts/set.py
```diff
@@ -177,7 +177,7 @@ def get_option_parser() -> COP:
     return parser
 
 
-def validate_tokens(tokens_list: Tuple['Tokens']) -> None:
+def validate_tokens(tokens_list: Tuple['Tokens', ...]) -> None:
```
oliver-sanders, line 180: Had to look up the meaning of `...` in this context, curious.

This function doesn't actually *require* a tuple, it just needs something it can iterate over, so this might make more sense:

```suggestion
def validate_tokens(tokens_list: Iterable['Tokens']) -> None:
```

It's generally a good idea to use the loosest description possible in method signatures to give yourself more wiggle room later, e.g. with `Iterable` you could provide a generator instead of a tuple for efficiency reasons.

Note, requires import change on line 93.

cylc/flow/task_pool.py
oliver-sanders: Could you stick up the issue and add a link to it in this comment so we can find our way back to the context when needed.

Note, the `cylc remove` work (will hopefully arrive in 8.4.0) may change this as removed tasks will be bumped into the `None` flow.

tests/integration/test_dbstatecheck.py
```diff
@@ -51,14 +51,18 @@ async def checker(
     })
     schd: Scheduler = mod_scheduler(wid, paused_start=False)
     async with mod_run(schd):
+        print("ONE")
```
oliver-sanders, line 54: Debug

cylc/flow/task_pool.py
hjoliver: Actually I've just removed the TODOs. This needs a bit of discussion first - see my post on Element.

Note this bit of code is designed to avoid respawning a previously-removed task (see #6066) - which is at odds with the proposed future of `cylc remove`.  (If erasing flow history, you *want* to be able to respawn the removed tasks, not avoid it).

hjoliver:
(All discussions resolved except for the TODO one)

tests/integration/test_task_pool.py
```diff
@@ -2019,9 +2019,56 @@ async def test_remove_no_respawn(flow, scheduler, start, log_filter):
...
+    # Scenario: set c2:succeeded in a future flow so when b succeeds
+    # in that flow it will spawn c1 but not c2.
+    id_ = flow({
+        'scheduler': {
+            'allow implicit tasks': True
+        },
+        'scheduling': {
+            'cycling mode': 'integer',
+            'graph': {
+                'R1': 'a => b => c1 & c2 => z',
```
wxtim, line 2048: IMO test becomes clearer if 

```suggestion
                'R1': 'a => good & failonce => z',
```

And is the task `b` necessary?

tests/integration/test_task_pool.py
```diff
@@ -2019,9 +2019,56 @@ async def test_remove_no_respawn(flow, scheduler, start, log_filter):
...
-            log, contains='Not spawning 1/z - task removed'
+            log, contains='Not respawning 1/z - task was removed'
         )
         z1 = schd.pool.get_task(IntegerPoint("1"), "z")
         assert (
             z1 is None
         ), '1/z should have stayed removed (but has been added back into the pool'
+
+
+async def test_set_future_flow(flow, scheduler, start, log_filter):
```
wxtim, line 2030: Checked out onto master this is passing for me.

wxtim (changes requested):
Manually tested, and appears that the fix is working.

I'm not convinced that the test is demonstrating the bug on upstream/8.3.x or master.

Is there a reason why this bug is on 8.3.x milestone but open against master?

tests/integration/test_dbstatecheck.py
```diff
@@ -73,7 +73,7 @@ def test_basic(checker):
         ['output', '10000101T0000Z', 'succeeded'],
         ['output', '10010101T0000Z', 'succeeded'],
         ['good', '10000101T0000Z', 'waiting', '(flows=2)'],
-    ]
+        ['good', '10010101T0000Z', 'waiting', '(flows=2)'], ]
```
oliver-sanders, line 76: Note, this task appears in the DB because the workflow has run on from the triggered task. This is an unintended side-effect of the `await sleep(1)` in the test fixture.

I've opened a PR to replace the sleep with a DB update call https://github.com/cylc/cylc-flow/pull/6212, however, if we want to keep this behaviour to allow the workflow to run to completion again (to test this PR), then replace the `sleep` with `await mod_complete(schd)` (and close #6212).

oliver-sanders:
> I'm not convinced that the test is demonstrating the bug on upstream/8.3.x or master.

I can confirm that `test_set_future_flow` passes on upstream/8.3.x so doesn't appear to be capturing this bug.

MetRonnie:
This should be rebased onto 8.3.x (even if it doesn't make it into the 8.3.2 release)

hjoliver:
> Is there a reason why this bug is on 8.3.2 milestone but open against master?

> This should be rebased onto 8.3.x (even if it doesn't make it into the 8.3.2 release)

I think I probably opened this PR right after 8.3.0 was released, before the 8.3.x branch was made, and forgot to come back and rebase it. But maybe it was just a mistake. Anyhow, rebased now.

hjoliver:
> I'm not convinced that the test is demonstrating the bug on upstream/8.3.x or master.

> I can confirm that test_set_future_flow passes on upstream/8.3.x so doesn't appear to be capturing this bug.

I commented on this [above](https://github.com/cylc/cylc-flow/pull/6186#issuecomment-2213171757)

The integration test also passes on master, but for "the wrong reasons", to do with submit numbers being zero in integration tests.

tests/integration/test_task_pool.py
hjoliver: > IMO test becomes clearer if

Well that makes it look as if success and failure are important for the test, which they're not. What matters is that only one of them gets respawned in the new flow, after the other one get manually set to succeeded (in the new flow).

> And is the task b necessary?

Actually neither b nor z are needed. They were just to make the original functional case easier to manage. I'll remove them.

tests/integration/test_dbstatecheck.py
hjoliver: (I've not had time to think about this today)

hjoliver: (Reviewers - this is test framework issue, so not a reason to delay this PR)

MetRonnie (approved):

MetRonnie:
Merging with missing changelog entry added!

tests/integration/test_task_pool.py
```diff
@@ -2019,9 +2019,50 @@ async def test_remove_no_respawn(flow, scheduler, start, log_filter):
...
+        schd.workflow_db_mgr.process_queued_ops()
+
+        # set task c1:succeeded in flow 2
+        schd.pool.set_prereqs_and_outputs(['1/c1'], prereqs=[], outputs=[], flow=[2])
+        schd.workflow_db_mgr.process_queued_ops()
+
+        # set b:succeeded in flow 2 and check downstream spawning
+        schd.pool.set_prereqs_and_outputs(['1/b'], prereqs=[], outputs=[], flow=[2])
+        assert schd.pool.get_task(IntegerPoint("1"), "c1") is None, '1/c1 (flow 2) should not be spawned after 1/b:succeeded'
+        assert schd.pool.get_task(IntegerPoint("1"), "c2") is not None, '1/c2 (flow 2) should be spawned after 1/b:succeeded'
```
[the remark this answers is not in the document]
hjoliver: It's a sign that writing it like this doesn't work (always True):
```
assert (
   var is None,
   'string'
)
```

However, I suppose I should have done this:
```
assert  var is None, (
   'string'
)
```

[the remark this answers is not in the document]
oliver-sanders: It's a sign that we don't run the linter on the test code!

Limiting line length is one thing that linters can all agree on.

| linter/formatter/guide | length |
| --- | --- |
| pep8 | 79 |
| google | 80 |
| flake8 | 80 |
| black | 88 |
| styleguide (ruby) | 80 |
| rubocop (ruby) | 80 |
| prettier (js/ts) | 80 |
| rustfmt (rust) | 80 |

Line length is arbitrary and each each tool is free to choose its own limit, yet most pick 80. This is not a coincidence:

* Readability (reading vertically is faster than reading horizontally, code structure is clearer when better spaced)
* Reduced diff surface (diff highlights changes more clearly, reduced merge conflicts)
* Accessibility (works better with larger font sizes)
* Works better with tooling (e.g. side-by-side diffs)

> Why Bother with 80 characters in a World of Modern Widescreen Displays?
>
> A lot of people these days feel that a maximum line length of 80 characters is just a remnant of the past and makes little sense today. After all - modern displays can easily fit 200+ characters on a single line. Still, there are some important benefits to be gained from sticking to shorter lines of code.
>
> First, and foremost - numerous studies have shown that humans read much faster vertically and very long lines of text impede the reading process. As noted earlier, one of the guiding principles of this style guide is to optimize the code we write for human consumption.
>
> Additionally, limiting the required editor window width makes it possible to have several files open side-by-side, and works well when using code review tools that present the two versions in adjacent columns.
>
> The default wrapping in most tools disrupts the visual structure of the code, making it more difficult to understand. The limits are chosen to avoid wrapping in editors with the window width set to 80, even if the tool places a marker glyph in the final column when wrapping lines. Some web based tools may not offer dynamic line wrapping at all.
>
> -- https://rubystyle.guide/#max-line-length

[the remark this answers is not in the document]
MetRonnie: I agree lines should be kept short if possible, but sometimes 79 is too short for the content and results in less readable code. I would support moving to black's 88 limit

cylc/flow/task_pool.py
```diff
@@ -1658,44 +1656,52 @@ def spawn_task(
...
+        ):
+            # If itask has any history in this flow but no completed outputs
+            # we can infer it was deliberately removed, so don't respawn it.
+            # TODO (follow-up work):
+            # - this logic fails if task removed after some outputs completed
+            # - this is does not conform to future "cylc remove" flow-erasure
+            #   behaviour which would result in respawning of the removed task
+            # See github.com/cylc/cylc-flow/pull/6186/#discussion_r1669727292
+            LOG.debug(f"Not respawning {point}/{name} - task was removed")
+            return None
```
MetRonnie, line 1703: What is to be done about this RE: `cylc remove`? Should this block just be removed? @hjoliver CC @oliver-sanders

hjoliver: Hmm, I need to remind myself and think again about this.  The linked discussion refers to Element chat. I'll try to find that, and we can continue the conversation there until it's clear what the problem is.

MetRonnie: However there are other ways that `TaskPool.remove()` are called that are not as a result of `cylc remove`, so this might still be needed to prevent a removed task from respawning immediately after being removed in one of those ways?

Thus I am thinking of leaving this in, just updating the comment (it no longer applies to tasks removed by `cylc remove`).

hjoliver: Yes I'm concerned about that too.

oliver-sanders: (without diving into the code in depth)

The proposal does not change the behaviour of suicide-triggered tasks so that will need to be preserved.

This logic should not apply to the remove command any more though.

PR #6187, opened 2024-07-01
Update MacOS version on GH Actions

MetRonnie:
Includes DNS patch

**Check List**

- [x] Infrastructure changes so no need for any funny stuff


MetRonnie:
Lint failure fixed in #6188

tests/functional/cylc-set/00-set-succeeded/flow.cylc
```diff
@@ -15,23 +15,23 @@
...
+                cylc workflow-state \
                     ${CYLC_WORKFLOW_ID}//${CYLC_TASK_CYCLE_POINT}/${TASK}:failed \
-                    --max-polls=10 --interval=1
-             done
-             # set foo succeeded (via --output)
-             cylc set -o succeeded $CYLC_WORKFLOW_ID//$CYLC_TASK_CYCLE_POINT/foo
-             # set bar succeeded (via default)
-             cylc set $CYLC_WORKFLOW_ID//$CYLC_TASK_CYCLE_POINT/bar
-         """
+                    --max-polls=10 --interval=2
```
oliver-sanders, line 31: (increase polls rather than interval to keep test responsive)

```suggestion
                    --max-polls=20 --interval=1
```

cylc/flow/hostuserutil.py
```diff
@@ -113,24 +114,26 @@ def get_host_ip_by_name(target):
...
         if target not in self._host_exs:
-            if target is None:
-                target = socket.getfqdn()
-            if (
-                IS_MAC_OS
-                and target == (
-                    '1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.'
-                    '0.0.0.0.0.0.ip6.arpa'
-                )
-            ):
```
oliver-sanders, line 127: I'm not sure this patch makes sense, `None` should not be in `_host_exs`, if an `arpa` address *is* in `_host_exs`, that's a bug.

What issue were you trying to fix?

MetRonnie: Whenever `target` is `None` then it assumes the host info is not cached when it might be. Because it goes on to call `socket.getfqdn()`.

It makes more sense to call `socket.getfqdn()` so we have a target in mind, then check if it is cached.

oliver-sanders: So no related issue, just arbitrary "tidying"?

oliver-sanders (approved):

wxtim (approved):

oliver-sanders:
Opened separate issue for tracebacks: https://github.com/cylc/cylc-flow/issues/6208

PR #6192, opened 2024-07-04
Allow manual trigger when workflow paused.

### LINKED ISSUE #5963: Allow trigger-now in a paused workflow?
There is a valid use case (especially since the demise of the old `cylc submit` command for running individual tasks sans scheduler) early on in workflow development and debugging, for starting a workflow paused and then manually triggering individual tasks to run.

Currently "paused" means "don't run anything", so if you trigger a task in a paused workflow, the task won't run until the workflow gets un-paused.

I think we probably intended `cylc hold` for this sort of thing, but that doesn't do it without additional seemingly unnecessary effort: holding an already-active task (even via a "hold all" pattern) does not prevent downstream future tasks from running unless every one of those has also been specifically held.

So ... is there any good reason not to allow manual triggering (in the "run right now" sense) of tasks in a paused workflow?

See cylc/cylc-ui#1651

hjoliver:
Close #5963 

I've added the "question" label because there's some disagreement on the issue itself:

@oliver-sanders has argued that manual triggering (i.e., with immediate effect) should only be allowed with `cylc hold */*` (hold all tasks individually) and not `cylc pause` (pause the workflow).

However:
- we haven't implemented `cylc hold */*` yet, and it may not be trivial to do it
- the only solution at the moment is "hold after cycle point", which is not intuitive at all
- two primary Cylc sites have highlighted this as a must-have for operational migration - @dwsutherland, @ColemanTom
- this is very easy to implement (`+13/-7` code lines without tests) so at the very least it provides a working interim solution [UPDATE: unfortunately no longer such a small diff after a change of approach!]
- and finally, I cannot see any problem with allowing manual triggering in a paused workflow - there's nothing wrong with defining workflow pause as *pausing all **automatic** job submission*

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at https://github.com/cylc/cylc-doc/pull/778

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -478,6 +478,7 @@ class Scheduler:
                 get_workflow_test_log_path(self.workflow)))
 
         self.pool = TaskPool(
+            self.start_job_submission,
             self.tokens,
             self.config,
             self.workflow_db_mgr,
@@ -1317,9 +1318,13 @@ class Scheduler:
         # Return, if no tasks to submit.
         else:
             return False
+
         if not pre_prep_tasks:
             return False
 
+        return self.start_job_submission(pre_prep_tasks)
+
+    def start_job_submission(self, pre_prep_tasks) -> bool:
         # Start the job submission process.
         self.is_updated = True
         self.reset_inactivity_timer()
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1291,7 +1291,8 @@ class Scheduler:
 
         """
         if (
-            self.stop_mode is None
+            not self.is_paused
+            and self.stop_mode is None
             and self.auto_restart_time is None
             and self.reload_pending is False
         ):
@@ -1959,7 +1960,7 @@ class Scheduler:
         if msg:
             _msg += f': {msg}'
         LOG.info(_msg)
-        self.is_paused = self.pool.task_queue_mgr.is_paused = True
+        self.is_paused = True
         self.workflow_db_mgr.put_workflow_paused(True)
         self.update_data_store()
 
@@ -1981,7 +1982,7 @@ class Scheduler:
             return
         if not quiet:
             LOG.info("RESUMING the workflow now")
-        self.is_paused = self.pool.task_queue_mgr.is_paused = False
+        self.is_paused = False
         self.workflow_db_mgr.put_workflow_paused(False)
         self.update_data_store()
 
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1291,8 +1291,7 @@ class Scheduler:
 
         """
         if (
-            not self.is_paused
-            and self.stop_mode is None
+            self.stop_mode is None
             and self.auto_restart_time is None
             and self.reload_pending is False
         ):
@@ -1960,7 +1959,7 @@ class Scheduler:
         if msg:
             _msg += f': {msg}'
         LOG.info(_msg)
-        self.is_paused = True
+        self.is_paused = self.pool.task_queue_mgr.is_paused = True
         self.workflow_db_mgr.put_workflow_paused(True)
         self.update_data_store()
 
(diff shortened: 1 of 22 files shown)


oliver-sanders:
I would prefer to have a clean separation of play/pause and hold/release rather than crossing the concepts. I'm also not convinced by the "operational blocker" statement, especially as there is a documented workaround (i.e. this is a preference not a requirement).

However, I am quite happy to concede if this is desired.

This change shouldn't really go into a bugfix release, but it's small enough that I think it's ok to sneak it in if we really want to.

cylc/flow/task_queues/independent.py
```diff
@@ -54,7 +56,10 @@ def release(self, active: Counter[str]) -> List['TaskProxy']:
             except IndexError:
                 # deque empty
                 break
-            if itask.state.is_held:
+            if (
+                itask.state.is_held or
+                (is_paused and not itask.is_manual_submit)
+            ):
```
oliver-sanders, line 62: We would expect this functionality to work the same across all queue implementations, so we could really do with abstracting it into the base class rather than relying on each queue to implement it separately.

Idea (for future work), write some abstract tests that should hold for all queue implementations.

cylc/flow/scheduler.py
```diff
@@ -1225,8 +1225,7 @@ def release_queued_tasks(self) -> bool:
 
         """
         if (
-            not self.is_paused
-            and self.stop_mode is None
+            self.stop_mode is None
```
oliver-sanders, line 1228: > there's nothing wrong with defining workflow pause as pausing all **automatic** job submission

This has gone a little further than that one word change above as we are now continuing to process the queues whilst the workflow is running.

I don't think we actually need to process the entire queue to achieve this, we just need to skip any manually triggered tasks through the queue.

If we pass `self.is_paused` through to `release_queued_tasks` we could reduce the scope of this override such that it only impacts manually triggered tasks.

oliver-sanders (commented):
From a quick functional test, this works as expected:

* `cylc trigger` causes a task to submit
* `cylc set --pre=all` does not.

oliver-sanders:
Needs tests and a cylc-doc PR to update the interventions page.

hjoliver:
(Will undraft again soon, once I've tried your suggestion re queue processing @oliver-sanders )

hjoliver:
(Damn it, thought I had completed this ... it's wanted at NIWA ... coming back to it soon).

hjoliver:
Reminded of this after a couple more support questions at NIWA on how to trigger individual tasks while holding everything else.

> I would prefer to have a clean separation of play/pause and hold/release rather than crossing the concepts. 

I do somewhat agree in principle, but it's a major problem that `cylc hold` doesn't (yet?) work for this common and fairly simple use case. The workaround - hold-after-cycle-point is completely non-intuitive for users. And IMO we don't pollute the concepts much by defining play/pause as pertaining to automatic job submission.

cylc/flow/task_queues/independent.py
hjoliver: Good idea. Opened an Issue for that https://github.com/cylc/cylc-flow/issues/6483

hjoliver:
@oliver-sanders - I've implemented your suggestion. 

If the workflow is paused, do not process the queues. Instead:
- separate initiation of job prep from queue processing
- initiate job prep directly from the force-trigger method, so it works the same whether paused or not
- this means force-triggered tasks no longer need to be remember by the queue manager, for release later in general queue processing

hjoliver:
Ready for review.

TODO: cylc-doc PR

This doesn't need `set --pre=all` tests because the new triggering behaving is all located in the force-trigger method.

oliver-sanders:
@hjoliver: FYI, I'm working on the broadcast management approach we discussed at the meeting the other week.

hjoliver:
@oliver-sanders - just thinking about your original position on this... are you aware of use cases where it might actually be desirable to "pre-trigger" tasks while paused, but not have them run until unpaused?  If so, I could add a small tweak to this PR to support both cases: e.g. a `--force` option to make triggered tasks run now if the workflow is paused, other wait till unpaused.

hjoliver:
(DRAFT again - back in progress)

hjoliver:
I'm superseding this with a new PR. The change of approach has left the initial description outdated and the discussion overly confusing.

PR #6200, opened 2024-07-04
Fix stall status bug

### LINKED ISSUE #6199: Stalled paused workflows incorrectly have running status
### Description

Stall a workflow. Whilst stalled the top level pause button will be unavailable.

image

You can still pause the workflow
  * Via the burger bar drop-down
  * In Tui 🐦‍
  * On the CLI

Which suggests that it's a GUI issue.

MetRonnie:
Closes #6199. A stalled paused workflow would incorrectly have a status of running instead of paused.

**Check List**

- [x] No dependency changes 
- [x] Tests are included 
- [x] No docs entry needed.

### DIFF
--- a/cylc/flow/wallclock.py
+++ b/cylc/flow/wallclock.py
@@ -16,7 +16,7 @@
 """Wall clock related utilities."""
 
 from calendar import timegm
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 from metomi.isodatetime.timezone import (
     get_local_time_zone_format, get_local_time_zone, TimeZoneFormatMode)
@@ -209,7 +209,7 @@ def get_time_string_from_unix_time(unix_time, display_sub_seconds=False,
     to use as the time zone designator.
 
     """
-    date_time = datetime.utcfromtimestamp(unix_time)
+    date_time = datetime.fromtimestamp(unix_time, timezone.utc)
     return get_time_string(date_time,
                            display_sub_seconds=display_sub_seconds,
                            use_basic_format=use_basic_format,
--- a/cylc/flow/workflow_status.py
+++ b/cylc/flow/workflow_status.py
@@ -16,7 +16,7 @@
 """Workflow status constants."""
 
 from enum import Enum
-from typing import Tuple, TYPE_CHECKING
+from typing import TYPE_CHECKING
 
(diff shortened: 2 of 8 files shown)


cylc/flow/workflow_status.py
```diff
@@ -143,62 +143,41 @@ class AutoRestartMode(Enum):
...
-        # fallback - running indefinitely
-        status_msg = 'running'
-
-    return (status.value, status_msg)
+        return f'stopping: {schd.stop_mode.explain()}'
+    if schd.reload_pending:
+        return f'reloading: {schd.reload_pending}'
+    if schd.is_stalled:
+        if schd.is_paused:
+            return 'stalled (paused)'
```
hjoliver, line 163:
```suggestion
            return 'stalled and paused'
```
Just wondering if `stalled (paused)` could be misinterpreted as "stalled, i.e. paused" (which of course is wrong).

cylc/flow/workflow_status.py
```diff
@@ -143,62 +143,41 @@ class AutoRestartMode(Enum):
...
+        return 'stalled'
+    if schd.is_paused:
+        return 'paused'
+    if schd.pool.hold_point:
+        return WORKFLOW_STATUS_RUNNING_TO_HOLD % schd.pool.hold_point
+    if schd.pool.stop_point:
+        return WORKFLOW_STATUS_RUNNING_TO_STOP % schd.pool.stop_point
```
hjoliver, line 170: The order of these should probably depend on which point comes first (if we can have both at once - I don't recall)

hjoliver (approved):
Looks good, a couple of minor suggestions.

cylc/flow/wallclock.py
```diff
@@ -209,7 +209,7 @@ def get_time_string_from_unix_time(unix_time, display_sub_seconds=False,
     to use as the time zone designator.
 
     """
-    date_time = datetime.utcfromtimestamp(unix_time)
+    date_time = datetime.fromtimestamp(unix_time, timezone.utc)
```
oliver-sanders, line 212: For future reference, please raise changes like this against master, not a bugfix release.

oliver-sanders (approved):

hjoliver (approved):

PR #6203, opened 2024-07-05
Use here-document for job script environment variables

### LINKED ISSUE #6202: More robust handing of user provided environment variables in job scripts
### Problem

It is currently difficult to pass arbitrary text into a job from rose edit, due to the many layers of interpretation that variables go through. One of these layers is inside of job scripts, where variables are stored within double quotes (`"`), which can cause contained quotes to be matched and cause trouble, as well as `$MY_VAR` variable references being interpreted.

### Proposed Solution

A [here document](https://www.gnu.org/software/bash/manual/html_node/Redirections.html#Here-Documents) with an obscure terminating sequence should be sufficient to avoid this.

jfrost-mo:
This prevents unwanted quote stripping, etc. As of this commit it does not yet handle expanding ~, which the original code does, and has a comment explaining why.

The here-document is unquoted, so shell parameters are expanded within. This matches the existing behaviour, but if it is desired to make them more robust still it can be easily prevented by quoting the first opaque_terminator with double quotes.

Fixes #6202

**Check List**

### DIFF
--- a/cylc/flow/job_file.py
+++ b/cylc/flow/job_file.py
@@ -18,6 +18,7 @@
 from contextlib import suppress
 import os
 import re
+import secrets
 import stat
 from subprocess import Popen, PIPE, DEVNULL
 from textwrap import dedent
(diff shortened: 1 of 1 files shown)


jfrost-mo:
Questions:

### Do we want shell parameter expansion in environment variables?

The existing code does this.

### Where does the job script creation code run? In the scheduler, or under a user account on the job host?

If tilde handling is this missing bit from this PR currently. If this python code runs in the context of the user, then we can just use [`os.path.expanduser`](https://docs.python.org/3/library/os.path.html#os.path.expanduser).

oliver-sanders:
Here docs are an approach we have previously rejected for performance / robustness reasons (sorry).

I'm going to close this PR, however, discussion over alternative approaches can continue in a new issue.

PR #6216, opened 2024-07-10
check towncrier message format in lint ci

wxtim:
Oliver and I have both now made the mistake of mis-forming the towncrier messages, so here's an auto-check.

Example of it failing (correctly) on master
https://github.com/wxtim/cylc/actions/runs/9871865449/job/27260726283

### DIFF
--- a/.github/workflows/test_fast.yml
+++ b/.github/workflows/test_fast.yml
@@ -100,6 +100,15 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Check Towncrier Files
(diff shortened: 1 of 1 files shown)


MetRonnie:
Hopefully we can get this implemented in towncrier itself: https://github.com/twisted/towncrier/pull/622

wxtim:
Drafted pending outcome of @MetRonnie 's proposed changes to Towncrier.

wxtim:
Good progress on https://github.com/twisted/towncrier/pull/622

Issue #5875, opened 2023-12-13
examples: add more example workflows


oliver-sanders:
https://github.com/cylc/cylc-flow/pull/5773 adds example workflows to the cylc-flow repository.

These are available via the CLI and get auto-documented in cylc-doc. The idea is to document the major implementation patterns for Cylc workflows rather than trying to cover every concrete use case, i.e. keep it generic.

Suggestions for future examples:

* [ ] Site portability using Jinja2 include files.
* [ ] Sub workflows.
* [ ] Perhaps an xtrigger example which listens for an external event, and returns data used to configure the cycle (look at the event-driven-cycling workflow for the configuration side of things).

PR #6223, opened 2024-07-11
example: extending workflow

oliver-sanders:
> [!IMPORTANT]
> This is #5947 with the "complex" example (to which we do not currently have a generic solution) chopped of. The remainder of that PR will need to wait, however there is pressure for us to get the simple bit documented now.

Document the preferred approach to "extending" a workflow.

Partially addresses: https://github.com/cylc/cylc-doc/issues/525
Partially addresses: https://github.com/cylc/cylc-flow/issues/5875

**Context:**

A common pattern in some areas is to run a workflow until it stops. Then modify the final cycle point to extend the workflow and restart.

This isn't an easy working pattern to port to Cylc 8 due to the difficulty of re-populating the start tasks. We have a plan to make this easier, see https://github.com/cylc/cylc-flow/issues/5416, however, this working pattern is needlessly complex in the first place. It would be much simpler if the workflow didn't "finish" in the first place.

To do this, swap final cycle point for stop after cycle point, now the workflow will not "finish" so the pool will not need re-polulating for the workflow to continue. This is much nicer for Cylc (no discontinuity), the user (no extra commands) and provenance (direct continuation of previous run).

**Check List**

### DIFF
--- a/cylc/flow/etc/examples/extending-workflow/index.rst
+++ b/cylc/flow/etc/examples/extending-workflow/index.rst
@@ -1,0 +1,77 @@
+Extending Workflow
+==================
+
+.. cylc-scope:: flow.cylc[scheduling]
+
+Sometimes we may run a workflow to :term:`completion <workflow completion>`,
+but subsequently wish to run it for a few more cycles.
+
+With Cylc 7 this was often done by changing the `final cycle point` and
+restarting the workflow. This approach worked, but was a little awkward.
+It's possible with Cylc 8, but we would recommend moving away from this
+pattern instead.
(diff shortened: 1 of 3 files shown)


cylc/flow/etc/examples/extending-workflow/index.rst
```diff
@@ -0,0 +1,77 @@
+Extending Workflow
+==================
```
wxtim, line 2:
```suggestion
Extending Workflows
===================
```

wxtim (approved):

PR #6232, opened 2024-07-15
pool: warn when triggering queued tasks in different flows

### LINKED ISSUE #6174: triggering a queued task in a different flow has no effect
It would appear that it is not possible to trigger a task in a different flow when it is queued. This could, potentially, be true of waiting states in general, investigation needed.

Example workflow:

```ini
[scheduling]
    [[graph]]
        R1 = foo
[runtime]
    [[foo]]

```

Instructions to reproduce:

```console
$ cylc vip --pause
$ cylc trigger <workflow>//1/foo --flow=2
$ cylc play <workflow>
$ sleep 5
$ sqlite3 ~/cylc-run/one/run1/.service/db 'select * from task_outputs'
1|foo|[1]|["submitted", "started", "succeeded"]
```

The new flow is logged:

```
INFO - New flow: 2 (no description) 
```

However, the task does not run in it:

```
```

As evidenced by the DB showing the resulting task outputs as existing in flow `[1]`.

oliver-sanders:
* The `--flow` argument is ignored if you trigger a queued task.
* It would be possible, though tricky, to merge the flows at this point as queued tasks already have database entries.
* Merging is the correct solution, but for now, this PR will log a warning to let the user know / leave some record that the trigger did not take effect (i.e. continue the status quo but explain what's happening).
* Might be easier to come back to the merging problem post cylc-remove to see what requirement there is for recording the flow number *before* a task has run once that work is in.
* Closes #6174

**Check List**

- [ ] Changelog entry included if this is a change that can affect users (v minor, unreported issue)

### DIFF
--- a/tests/integration/test_task_pool.py
+++ b/tests/integration/test_task_pool.py
@@ -516,6 +516,36 @@ async def test_trigger_states(
         assert task.is_manual_submit == should_trigger
 
 
+async def test_trigger_queue(one, run, db_select, complete, log_filter):
+    """It should handle triggering tasks in the queued state.
+
+    It should not be possible to use "trigger" to add flows to a task that is
+    in the queued state. This is not a firm requirement, we could reconsider
+    this behaviour at a later date, however, it is tricky as at present the
+    task's entry in the task_outputs table already exists by this point.
+
+    This test ensures that a warning is produced in this circumstance.
+
+    See https://github.com/cylc/cylc-flow/issues/6174
+    """
+    async with run(one) as log:
+        # the workflow should start up with one task in the original flow
+        task = one.pool.get_tasks()[0]
+        assert task.state(TASK_STATUS_WAITING, is_queued=True)
+        assert task.flow_nums == {1}
+
+        # we should not be able to trigger this task as it is already queued in
+        # flow 1
+        log.clear()
+        one.pool.force_trigger_tasks([task.identity], '2')
+        assert log_filter(log, contains='ignoring trigger - already queued')
+
+        # only the original flow should continue
+        one.resume_workflow()
+        await complete(one, timeout=2)
+        assert db_select(one, False, 'task_outputs', 'flow_nums') == [('[1]',)]
+
+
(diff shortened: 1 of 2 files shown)


hjoliver:
> It would be possible, though tricky, to merge the flows at this point as queued tasks already have database entries.

I don't think it is tricky, unless I'm missing something. I'll put up another PR to possibly supersede this one...

tests/integration/test_task_pool.py
```diff
@@ -516,6 +516,36 @@ async def test_trigger_states(
         assert task.is_manual_submit == should_trigger
 
 
+async def test_trigger_queue(one, run, db_select, complete, log_filter):
+    """It should handle triggering tasks in the queued state.
+
+    It should not be possible to use "trigger" to add flows to a task that is
+    in the queued state. This is not a firm requirement, we could reconsider
+    this behaviour at a later date, however, it is tricky as at present the
+    task's entry in the task_outputs table already exists by this point.
```
hjoliver, line 525: OK this is why you think its tricky. Actually the same thing happens with any flow merge, since it only happens if a flow runs into an existing `n=0` task.  We just add a new flow row in the DB. 

```console
$ sqlite3 ~/cylc-run/bug/runN/log/db "select * from task_outputs"
1|bar|[1]|{"submitted": "submitted", "started": "started"}
1|foo|[1]|{}                                                       # <--- pre merge, waiting
1|foo|[1, 2]|{"submitted": "submitted", "started": "started"}      # <--- post merge, running
```

oliver-sanders: Isn't it a problem leaving the `[1]` task behind in the DB?

hjoliver: How is it a problem?

On the plus side, you can see exactly what happened (flow merge when foo was waiting).

And, as I said, the same thing happens for normal (non manual trigger) flow merges.

oliver-sanders:
Superseded by https://github.com/cylc/cylc-flow/pull/6241

PR #6234, opened 2024-07-15
clean: change DB corruption advice

oliver-sanders:
Recommend using `--local-only` rather than removing the DB (does the same thing at the end of the day).

**Check List**

- [ ] Tests are included (single log message).
- [ ] Changelog entry (irrelevantly minor functionless change).


MetRonnie (approved):
I can see this is branched off 8.3.x so have changed the base accordingly. Merge if this is correct

PR #6241, opened 2024-07-17
flow merge for force-triggered n=0 (e.g. queued) tasks

### LINKED ISSUE #6174: triggering a queued task in a different flow has no effect

hjoliver:
Supersede #6232 
Close #6174 

**Check List**

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -2367,6 +2367,25 @@ class DataStoreMgr:
         self.state_update_families.add(tproxy.first_parent)
         self.updates_pending = True
 
+    def delta_task_flow_nums(self, itask: TaskProxy) -> None:
+        """Create delta for change in task proxy flow_nums.
+
+        Args:
+            itask (cylc.flow.task_proxy.TaskProxy):
+                Update task-node from corresponding task proxy
+                objects from the workflow task pool.
+
+        """
+        tproxy: Optional[PbTaskProxy]
+        tp_id, tproxy = self.store_node_fetcher(itask.tokens)
+        if not tproxy:
+            return
+        tp_delta = self.updated[TASK_PROXIES].setdefault(
+            tp_id, PbTaskProxy(id=tp_id))
+        tp_delta.stamp = f'{tp_id}@{time()}'
+        tp_delta.flow_nums = serialise_set(itask.flow_nums)
+        self.updates_pending = True
+
     def delta_task_runahead(self, itask: TaskProxy) -> None:
         """Create delta for change in task proxy runahead state.
 
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1387,6 +1387,7 @@ class TaskPool:
             ).relative_id
 
             c_task = self._get_task_by_id(c_taskid)
+            in_pool = c_task is not None
 
             if c_task is not None and c_task != itask:
                 # (Avoid self-suicide: A => !A)
@@ -1411,10 +1412,12 @@ class TaskPool:
                         tasks.append(c_task)
                 else:
                     tasks = [c_task]
+
                 for t in tasks:
                     t.satisfy_me([itask.tokens.duplicate(task_sel=output)])
                     self.data_store_mgr.delta_task_prerequisite(t)
-                    self.add_to_pool(t)
+                    if not in_pool:
+                        self.add_to_pool(t)
 
                     if t.point <= self.runahead_limit_point:
                         self.rh_release_and_queue(t)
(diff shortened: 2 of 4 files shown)


cylc/flow/data_store_mgr.py
```diff
@@ -2367,6 +2367,25 @@ def delta_task_queued(self, itask: TaskProxy) -> None:
         self.state_update_families.add(tproxy.first_parent)
         self.updates_pending = True
 
+    def delta_task_flow_nums(self, itask: TaskProxy) -> None:
```
hjoliver, line 2370: There was no way to update flow_nums in the data_store. 

We got away with this by (erroneously) re-adding flow-merged n=0 tasks to the task pool, which created an entirely new datastore task proxy.

cylc/flow/task_pool.py
```diff
@@ -1411,10 +1412,12 @@ def spawn_on_output(self, itask, output, forced=False):
                         tasks.append(c_task)
                 else:
                     tasks = [c_task]
+
                 for t in tasks:
                     t.satisfy_me([itask.tokens.duplicate(task_sel=output)])
                     self.data_store_mgr.delta_task_prerequisite(t)
-                    self.add_to_pool(t)
+                    if not in_pool:
+                        self.add_to_pool(t)
```
hjoliver, line 1420: This is the add-to-pool bug.

cylc/flow/task_pool.py
```diff
@@ -2169,6 +2172,7 @@ def force_trigger_tasks(
             if itask.state(TASK_STATUS_PREPARING, *TASK_STATUSES_ACTIVE):
                 LOG.warning(f"[{itask}] ignoring trigger - already active")
                 continue
+            self.merge_flows(itask, flow_nums)
```
hjoliver, line 2175: This is all that was needed to trigger an existing `n=0` task (e.g. queued) in a new flow. The rest of the PR is just making sure flow_nums change end up in the data store.

hjoliver:
(One SOD test broken, expecting an easy fix though...)

tests/integration/test_task_pool.py
```diff
@@ -2086,3 +2086,27 @@ async def test_set_future_flow(flow, scheduler, start, log_filter):
...
+        assert task.flow_nums == {1}
+
+        # trigger this task even though is already queued in flow 1
+        log.clear()
+        one.pool.force_trigger_tasks([task.identity], '2')
+
+        # the merged flow should continue
+        one.resume_workflow()
+        await complete(one, timeout=2)
+        assert db_select(one, False, 'task_outputs', 'flow_nums') == [('[1, 2]',), ('[1]',)]
```
oliver-sanders, line 2112: Can remove the logging stuff:

```suggestion
async def test_trigger_queue(one, run, db_select, complete,):
    """It should handle triggering tasks in the queued state.

    Triggering a queued task with a new flow number should result in the
    task running with merged flow numbers.

    See https://github.com/cylc/cylc-flow/pull/6241
    """
    async with run(one):
        # the workflow should start up with one task in the original flow
        task = one.pool.get_tasks()[0]
        assert task.state(TASK_STATUS_WAITING, is_queued=True)
        assert task.flow_nums == {1}

        # trigger this task even though is already queued in flow 1
        one.pool.force_trigger_tasks([task.identity], '2')

        # the merged flow should continue
        one.resume_workflow()
        await complete(one, timeout=2)
        assert db_select(one, False, 'task_outputs', 'flow_nums') == [('[1, 2]',), ('[1]',)]
```

oliver-sanders (approved):
:+1:

wxtim:
Have replicated bug with a runahead limited task - but I still not OK on your branch.

```
[scheduling]
    initial cycle point = 1000
    runahead limit = P2
    [[graph]]
        P1Y = foo
[runtime]
    [[foo]]
        script = sleep 40
```

```
cylc vip
sleep 5
cylc trigger workflow//10030101T0000Z/foo
sleep 40
check db

wxtim (approved):

Issue #6244, opened 2024-07-17
glblcfg: platform not found


oliver-sanders:
We are seeing multiple reports of issues where platforms which are defined in the global configuration are reported as not being found at task submission time.

To activate this bug:
* Write a workflow which gas a task that runs on a remote platform.
* Add a `rose-suite.conf` file.
* Run the workflow in `--debug` mode.

Although some reports suggest that the issue can be replicated without the `--debug` flag, we have yet to reproduce this ourselves.

The origin of the bug is a feature that was added to cylc-rose (released in conjunction with cylc-flow 8.3.0): https://github.com/cylc/cylc-rose/pull/269/files#diff-aa4c6cd8a676c24d08145f471d2e3830f1e41677a6cd1d1b9ce0fd93ced9c6c8R877-R878

This feature reloads the global configuration (`glbl_cfg().load()`). The reload wipes the platforms in the process, but this only appears to happen when `--debug` is set for reasons not yet understood.

oliver-sanders:
Reproducible example?

ScottWales:
A suspicion that I have is that when debug is enabled the global config is read as part of logging messages, which happens before or at the same time as when Rose reloads the configuration. Potentially some sort of async race?

wxtim:
> Reproducible example?

It's in `$HOME/cylc-src/bugs/` with an appropriate name, but it was a very, very simple workflow, with an empty `rose-suite.conf`.

oliver-sanders:
> It's in $HOME/cylc-src/bugs/ with an appropriate name

This example does not reproduce the bug without using `--debug`.

MetRonnie:
This diff fixes the problem by avoiding `.get()`ing the global config on every log call.

```diff
diff --git a/cylc/flow/loggingutil.py b/cylc/flow/loggingutil.py
index 3d77bdcb0..a46c09db3 100644
--- a/cylc/flow/loggingutil.py
+++ b/cylc/flow/loggingutil.py
@@ -161,4 +161,9 @@ class RotatingLogFileHandler(logging.FileHandler):
         self.log_num: Optional[int] = None  # null value until log file created
 
+        self.max_bytes = glbl_cfg().get(
+            ['scheduler', 'logging', 'maximum size in bytes'])
+        if self.max_bytes < self.MIN_BYTES:  # No silly value
+            self.max_bytes = self.MIN_BYTES
+
     def emit(self, record):
         """Emit a record, rollover log if necessary."""
[...]

PR #6245, opened 2024-07-17
glblcfg: replicate platforms rather than overriding them

oliver-sanders:
We are currently replacing the `[platforms]` part of the "sparse" configuration.

This means that the platforms dictionary is being replaced, introducing the potential for some part of the system to retain a reference to the stale dictionary.

This PR uses the parsec replicate method as used elsewhere to decant the new platform entries into the existing dictionary.

Attempted whilst investigating #6244, however, this does not fix that issue, at least, not by itself.

**Check List**


oliver-sanders:
This wasn't part of the solution and we have moved away from in-place modification so this is no longer needed.

PR #6248, opened 2024-07-19
Support EmPy version 4

hjoliver:
A forum post has revealed that EmPy 3 doesn't play well with scheduler daemonization: 

https://cylc.discourse.group/t/empy-import-shell-environment-variables/985/1

If I simply add `#!EmPy` to a `flow.cylc`, the workflow runs fine but:
 - I get a traceback at start-up
- stdout does not end up in the log
- stdout still appears in the terminal

Something to do with EmPy replacing `sys.stdout`
```
--- Logging error ---
  File "/home/oliverh/cylc/cylc-flow/cylc/flow/loggingutil.py", line 167, in emit
    self.do_rollover()
  File "/home/oliverh/cylc/cylc-flow/cylc/flow/loggingutil.py", line 234, in do_rollover
    os.dup2(self.stream.fileno(), sys.stdout.fileno())
AttributeError: 'ProxyFile' object has no attribute 'fileno'
```

The good news: EmPy 4.1 works fine, but it requires a small change to our code.

Anyone aware of any reason not to require EmPy > 4 ?

**Check List**

### DIFF
--- a/cylc/flow/parsec/empysupport.py
+++ b/cylc/flow/parsec/empysupport.py
@@ -22,7 +22,6 @@ from io import StringIO
 import em
 import os
 import typing as t
-import inspect
 
 from cylc.flow.parsec.exceptions import EmPyError
 from cylc.flow.parsec.fileparse import get_cylc_env_vars
@@ -53,20 +52,11 @@ def empyprocess(
     os.chdir(dir_)
     ftempl = StringIO('\n'.join(flines))
     xtempl = StringIO()
-    # Detect EmPy version by interpreter.file() args.
-    # TODO: Use importlib.metadata version() once we drop Python 3.7
-    if 'name' in inspect.signature(em.Interpreter.file).parameters:
-        # Empy 3
-        interpreter = em.Interpreter(
-            output=em.UncloseableFile(xtempl)
-        )
-    else:
-        # Empy 4
-        # dispatcher = False: raise errors to caller
-        interpreter = em.Interpreter(
-            output=em.UncloseableFile(xtempl),
-            dispatcher=False
-        )
+    # Requires EmPy >= 4.1
+    interpreter = em.Interpreter(
+        output=em.UncloseableFile(xtempl),
+        dispatcher=False  # raise errors to caller
+    )
 
     # Add `CYLC_` environment variables to the global namespace.
     interpreter.updateGlobals(
@@ -74,7 +64,6 @@ def empyprocess(
     )
 
     try:
-        # These args work for EmPy versions 3 and 4.
         interpreter.file(ftempl, locals=template_vars)
     except Exception as exc:
         lineno = interpreter.contexts[-1].identify()[1]
--- a/cylc/flow/parsec/empysupport.py
+++ b/cylc/flow/parsec/empysupport.py
@@ -22,6 +22,7 @@ from io import StringIO
 import em
 import os
 import typing as t
+import inspect
 
 from cylc.flow.parsec.exceptions import EmPyError
 from cylc.flow.parsec.fileparse import get_cylc_env_vars
@@ -52,7 +53,20 @@ def empyprocess(
     os.chdir(dir_)
     ftempl = StringIO('\n'.join(flines))
     xtempl = StringIO()
-    interpreter = em.Interpreter(output=em.UncloseableFile(xtempl))
+    # Detect EmPy version by interpreter.file() args.
+    # TODO: Use importlib.metadata version() once we drop Python 3.7
+    if 'name' in inspect.signature(em.Interpreter.file).parameters:
+        # Empy 3
+        interpreter = em.Interpreter(
+            output=em.UncloseableFile(xtempl)
+        )
+    else:
+        # Empy 4
+        # dispatcher = False: raise errors to caller
+        interpreter = em.Interpreter(
+            output=em.UncloseableFile(xtempl),
+            dispatcher=False
+        )
 
     # Add `CYLC_` environment variables to the global namespace.
     interpreter.updateGlobals(
@@ -60,7 +74,8 @@ def empyprocess(
     )
 
     try:
-        interpreter.file(ftempl, '<template>', template_vars)
+        # These args work for EmPy versions 3 and 4.
+        interpreter.file(ftempl, locals=template_vars)
     except Exception as exc:
         lineno = interpreter.contexts[-1].identify()[1]
         raise EmPyError(
--- a/changes.d/6248.fix.md
+++ b/changes.d/6248.fix.md
@@ -1,0 +1,1 @@
+Upgrade to EmPy 4 to fix an issue with stdout in detaching mode.
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -21,7 +21,7 @@ dependencies:
   - tomli >=2 # [py<3.11]
 
 # optional dependencies
-  #- empy >=3.3,<3.4
+  #- empy >=4.1
   #- pandas >=1.0,<2
   #- pympler
   #- matplotlib-base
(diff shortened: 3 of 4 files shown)


cylc/flow/parsec/empysupport.py
```diff
@@ -52,15 +53,29 @@ def empyprocess(
...
+        interpreter = em.Interpreter(
+            output=em.UncloseableFile(xtempl)
+        )
+    else:
+        # Empy 4
+        # dispatcher = False: raise errors to caller
+        interpreter = em.Interpreter(
+            output=em.UncloseableFile(xtempl),
+            dispatcher=False
+        )
```
hjoliver, line 69: If agreed on enforcing EmPy version 4 then we can drop the back-compat bit here.

oliver-sanders:
From the looks of the EmPy changelog, there are quite a few breaking changes at version 4.

However, as far as I'm aware, there are no users of EmPy in Cylc workflows at all. The team who added this support have subsequently moved to Jinja2. Note, their motivation in adding EmPy was stability (i.e. nothing changes in EmPy) at a time when Jinja2 was somewhat volatile. So, TBH we would be quite happy to drop EmPy entirely. Either way, I don't think there's an issue with requiring EmPy 4. Perhaps put a post out on Discourse to see if anyone it using it that we are not aware of (probably not given the bug)?

hjoliver:
> However, as far as I'm aware, there are no users of EmPy in Cylc workflows at all. 

Well there's the user who posted the bug to the forum!

oliver-sanders:
That's a new workflow!

hjoliver:
> The team who added this support have subsequently moved to Jinja2.

Interesting, why did they move to Jinja2?

> Note, their motivation in adding EmPy was stability (i.e. nothing changes in EmPy) at a time when Jinja2 was somewhat volatile.

Didn't they like the more Python-like nature of EmPy too? Stability seems a somewhat crazy reason given that you don't have to go with new versions of a package until some compatibity issue arises, and that no package still in use *never* changes (hence EmPy 4 as we've just discovered!).

> So, TBH we would be quite happy to drop EmPy entirely. Either way, I don't think there's an issue with requiring EmPy 4. Perhaps put a post out on Discourse to see if anyone it using it that we are not aware of (probably not given the bug)?
 
 👍

oliver-sanders:
> Stability seems a somewhat crazy reason given that you don't have to go with new versions of a package until some compatibity issue arises

This was back in the period when Jinja2 was making frequent syntax changes in subsequent releases. Most notably, the highly controversial scoping changes (which bound scope to loops making it impossible to e.g. build a list in a loop) which were introduced one version before `namespaces()` (an object which is permitted to persist beyond the loop scope) were introduced.

Keeping up with these changes was a challenge and a portability nightmare, this ultimately led us to bundle Jinja2 with Cylc so that Jinja2 version changes could be synchronised with Cylc version changes.

cgrudz:
> > However, as far as I'm aware, there are no users of EmPy in Cylc workflows at all.
> 
> Well there's the user who posted the bug to the forum!

Hey there, I'm just chiming in on the discussion, I posted the bug report yesterday.  I'm new to Cylc so I've been evaluating the different ways that I can use its supported features / ecosystem, and I've been testing out its integration with these templating packages.  My use-case is that I'm converting a number of legacy [Bash/Python Rocoto-driven NWP workflows](https://github.com/CW3E/Ensemble-DA-Cycling-Template/tree/develop) and [job-array-driven NWP verification workflows](https://github.com/CW3E/MET-tools) to become a more robust, feature rich and sustainable suite of Cylc-driven workflows for case study analysis and for an eventual integration with our center's near-real-time forecast systems.  I've been testing both of Jinja2 and EmPy, and I think my team could use either one, but without having a lot of existing Jinja2 codes, and as many of my younger colleagues are already proficient with Python, we'd get more value out of templating in Python syntax and idioms.  It's not a huge obstacle for any of us to adopt Jinja2, but I think many new Cylc users will fall into a similar use-case as ours, as NCL is increasingly replaced with Python and Python usage is becoming increasingly widespread for geophysical data science. If it's not a huge burden to support EmPy or a similar package, I can definitely see the value.

oliver-sanders:
Hi,

Sorry for the delayed reply (been a bit busy lately).

The context here is that several years back (2018) someone contributed EmPy support to Cylc, but their project subsequently moved back to Jinja2 (?) and no one returned to the Cylc project to maintain EmPy support. The EmPy v4 release (Nov 2023) appears to have broken Cylc compatibility, but this went unnoticed for nearly a year because there aren't presently any Cylc projects making use of EmPy support.

Supporting other template engines is perfectly reasonable, but maintaining support is a bit of effort if no one uses it. Additionally I'm a little wary of EmPy because:

* The project appears to be the work of a single person.
* The code repository is not public, or at least sufficiently well buried that I can't find it (you can send the maintainer an email, but you can't send them a pull request).
* The project has previously [gone quiet for years at a time](https://pypi.org/project/empy/#history) (there was an 11 year gap between 3.3 and 3.3.1).
* The [conda-forge feedstock](https://github.com/conda-forge/empy-feedstock) appears to be unmaintained (the last Conda release is 3.3.4 which is 5 years old).

Which sadly makes it a high risk dependency. If something breaks in EmPy, you may get a fix (I hear the maintainer is very helpful), but one day they won't be maintaining it any more, in the absence of a succession plan, this will leave EmPy users stranded. The closed nature of the project may also be considered a security risk.

So, if it was two months ago (before you had spotted EmPy support in Cylc), I would have voted to remove support (before anyone actually tried to use it!).

If you're keen and accept the caveats above, perhaps we could consider giving it another go.

cgrudz:
@oliver-sanders, thanks for the detailed reply and your analysis of this dependency, sorry for my own slow response.  I think that you raised a lot of important issues about the long-term support for templating with EmPy and I appreciate that this could be a very dangerous road to go down if there isn't greater transparency about the EmPy project's future.  As far as my team is concerned, we can switch fully to using Jinja2 for templating if that receives first-tier support with Cylc, we don't have too much invested in this yet and both syntaxes are fairly similar to work with.

Cheers!
Colin

oliver-sanders:
We have now decided to drop EmPy support from Cylc rather than updating it ([meeting 2024-11-04](https://github.com/cylc/cylc-admin/issues/188#issuecomment-2225293775)).

Issue: https://github.com/cylc/cylc-flow/issues/6458

TomekTrzeciak:
> Note, their motivation in adding EmPy was stability (i.e. nothing changes in EmPy) at a time when Jinja2 was somewhat volatile. 

I'm a bit late to this discussion, but as the original contributor of this now defunct feature perhaps it's still worth to set the record straight about the history and motivation behind it.

Volatility of Jinja2, however annoying, wasn't the actual driving force behind adding EmPy. At the time we were building increasingly complex workflows with more and more logic ending up in Jinja2 templating. We had macros, recursive for-loops and other crazy stuff. This was getting prohibitively complex and hard to understand with very large, highly interconnected graphs (10k-100k nodes that needed splitting into the top level Cylc graph and Makefile or Dask subgraphs at the task level). At the time EmPy was seen as a better way to tame that complexity, because it is just regular Python with its full power (modules, functions, classes, comprehensions), unlike the Jinja2 presumptuous take.

Eventually it dawned on me that building large, complex graphs by stitching together pieces of stings is never going to work well. No matter how powerful the templating language, it's just not the right tool for the job. Complex logic becomes much more tractable when backed by proper data structures. In the end we gravitated towards NetworkX as the internal representation of our graph, with some custom code to generate and manipulate it. That left us with a relatively small amount of configuration in the template itself - only what was mandated by the needs of our operational team (workflow level variables, family definitions, etc.) - and for that Jinja2 was good enough, so we've stuck with it.

PR #6249, opened 2024-07-19
glbl_cfg: reload without mutating the original

oliver-sanders:
Sibling: https://github.com/cylc/cylc-rose/pull/336

* Use the new `reload` kwarg rather than calling the `.load()` method.
* Fixes https://github.com/cylc/cylc-flow/issues/6244
* The `.load()` method mutates the existing config, due to the use of logging within (and outside of) this routine and the use of `glbl_cfg` in the logging, this created a race condition.
* The new `reload` kwarg creates a new config instance, then sets this as the default.

To assert that the test captures the issue, try applying this diff to test the old reload method:

```diff
diff --git a/tests/integration/test_config.py b/tests/integration/test_config.py
index 231fb12f8..eb56ed5cb 100644
--- a/tests/integration/test_config.py
+++ b/tests/integration/test_config.py
@@ -588,7 +588,9 @@ async def test_glbl_cfg(monkeypatch, tmp_path, caplog):
     # * but during the reload itself, the old config should persist
     #   see https://github.com/cylc/cylc-flow/issues/6244
     expect_platforms_during_reload({'localhost', 'foo'})
-    assert get_platforms(glbl_cfg(reload=True)) == {'localhost', 'foo', 'bar'}
+    cfg = glbl_cfg()
+    cfg.load()
+    assert get_platforms(cfg) == {'localhost', 'foo', 'bar'}
     assert 'ran expect_platforms_during_reload test' in caplog.messages
 
     # the cache should have been updated by the reload
```

**Check List**

### DIFF
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -1978,24 +1978,29 @@ class GlobalConfig(ParsecConfig):
         super().__init__(*args, **kwargs)
 
     @classmethod
-    def get_inst(cls, cached: bool = True) -> 'GlobalConfig':
+    def get_inst(
+        cls, cached: bool = True, reload: bool = False
+    ) -> 'GlobalConfig':
         """Return a GlobalConfig instance.
 
         Args:
-            cached (bool):
-                If cached create if necessary and return the singleton
-                instance, else return a new instance.
+            cached:
+                If True, return a cached instance if present. If False, return
+                a new instance.
+            reload:
+                If true, reload the cached instance (implies cached=True).
+
         """
-        if not cached:
-            # Return an up-to-date global config without affecting the
-            # singleton.
-            new_instance = cls(SPEC, upg, validator=cylc_config_validate)
-            new_instance.load()
-            return new_instance
-        elif not cls._DEFAULT:
-            cls._DEFAULT = cls(SPEC, upg, validator=cylc_config_validate)
-            cls._DEFAULT.load()
-        return cls._DEFAULT
+        if cached and cls._DEFAULT and not reload:
+            return cls._DEFAULT
+
+        new_instance = cls(SPEC, upg, validator=cylc_config_validate)
+        new_instance.load()
+
+        if cached or reload:
+            cls._DEFAULT = new_instance
+
+        return new_instance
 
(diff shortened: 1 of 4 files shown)


cylc/flow/cfgspec/globalcfg.py
```diff
@@ -1978,24 +1978,29 @@ def __init__(self, *args, **kwargs) -> None:
...
         Args:
-            cached (bool):
-                If cached create if necessary and return the singleton
-                instance, else return a new instance.
+            cached:
+                If True, return a cached instance if present. If False, return
+                a new instance.
+            reload:
+                If true, reload the cached instance (implies cached=True).
+
```
wxtim, line 1992: Worth guarding against future misuse?

```suggestion
                If true, reload the cached instance (implies cached=True).
    """
    if reload and cached is False:
        raise CylcError('You cannot reload an un-cached global config')
```

tests/integration/test_config.py
```diff
@@ -503,3 +506,90 @@ def test_special_task_non_word_names(flow: Fixture, validate: Fixture):
...
+
+    def write_global_config(cfg_str):
+        """Write the global.cylc file."""
+        Path(tmp_path, 'global.cylc').write_text(cfg_str)
+
+    def get_platforms(cfg_obj):
+        """Return the platforms defined in the provided config instance."""
+        return {p for p in cfg_obj.get(['platforms']).keys()}
+
+    def expect_platforms_during_reload(platforms):
```
wxtim, line 534: 👍🏼 I enjoy learning about stuff when I review.

wxtim (approved):
- [x] Happy with code
- [x] Tried locally
- [x] Test fails against 8.3.x (and not just because it has an unrecognized keyword arg)

cylc/flow/cfgspec/globalcfg.py
MetRonnie: I think it's ok for `reload=True` to override the `cache` argument

hjoliver (approved):
LGTM!

cylc/flow/cfgspec/globalcfg.py
wxtim: Never mind - I don't think that this that important.

PR #6252, opened 2024-07-22
Logging: do not access global config object during emit

### LINKED ISSUE #6244: glblcfg: platform not found

MetRonnie:
A `glbl_cfg().get()` call can have the side effect of expanding the global config object, so should be avoided in the course of emitting log messages.

This will close #6244 by itself, but this is also supplementary to #6249; IMO both this and #6249 are worth having merged.

The added test is in a separate commit for ease of running against 8.3.x

**Check List**

- [x] No dependency changes
- [x] Tests are included 
- [ ] Changelog entry included if this is a change that can affect users (this doesn't need a changelog entry assuming #6249 is going to be merged)
- [x] No docs needed

### DIFF
--- a/cylc/flow/loggingutil.py
+++ b/cylc/flow/loggingutil.py
@@ -138,7 +138,6 @@ class RotatingLogFileHandler(logging.FileHandler):
 
     FILE_HEADER_FLAG = 'cylc_log_file_header'
     ROLLOVER_NUM = 'cylc_log_num'
-    MIN_BYTES = 1024
 
     header_extra = {FILE_HEADER_FLAG: True}
     """Use to indicate the log msg is a header that should be logged on
@@ -146,7 +145,7 @@ class RotatingLogFileHandler(logging.FileHandler):
 
     def __init__(
         self,
-        log_file_path: str,
+        log_file_path: Union[Path, str],
         no_detach: bool = False,
         restart_num: int = 0,
         timestamp: bool = True,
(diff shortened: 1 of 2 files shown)


cylc/flow/loggingutil.py
```diff
@@ -155,10 +154,19 @@ def __init__(
         self.no_detach = no_detach
         self.formatter = CylcLogFormatter(timestamp=timestamp)
         # Header records get appended to when calling
-        # LOG.info(extra=RotatingLogFileHandler.[rollover_]header_extra)
+        # `LOG.info(extra=RotatingLogFileHandler.[rollover_]header_extra)`:
         self.header_records: List[logging.LogRecord] = []
         self.restart_num = restart_num
         self.log_num: Optional[int] = None  # null value until log file created
+        # Get & cache properties from global config (note: we should not access
+        # the global config object during emit as it can have side effects)
```
hjoliver, line 162: Can you elaborate on this a bit? The side effect is to expand the sparse config, and why is that bad in this context?

hjoliver (approved):

cylc/flow/loggingutil.py
hjoliver: (Approved the PR anyway, as comment is obviously non-functional, but personally I'd like a bit more info here)

wxtim:
```suggestion
        # Get & cache properties from global config (note: we should not access
        # the global config object while running the `emit` method 
        # becuase ...)
```

wxtim (approved):
Approved. Consider note.

cylc/flow/loggingutil.py
MetRonnie: > The side effect is to expand the sparse config, and why is that bad in this context?

IMO that is self-explanatory - logging is a fundamental operation and should not have the possibility of any real side effects at all

wxtim:
```suggestion
        # Get & cache properties from global config (note: we should not access
        # the global config object during the emit method as it can have side effects)
```

MetRonnie:
[Updated](https://github.com/cylc/cylc-flow/compare/e5f7bc8081ffd0a3c801b0b9c9f1c6e72e5e4190..8bb662efc208fc1631beb48f59a90356fb68a359) comment, merging with 2 approvals

PR #6257, opened 2024-07-23
Fix flaky integration tests

MetRonnie:
Fixes state leakage between integration tests that was resulting in `tests/integration/reftests/test_cyclers.py::test_icp_fcp_notation` failing flakily.

<details><summary>Salient traceback</summary>
<p>

```
cylc/flow/scheduler.py:744: in start
    await self.handle_exception(exc)
cylc/flow/scheduler.py:1996: in handle_exception
    raise exc from None
cylc/flow/scheduler.py:741: in start
    await self.configure(params)
cylc/flow/scheduler.py:495: in configure
    self._load_pool_from_point()
cylc/flow/scheduler.py:782: in _load_pool_from_point
    self.pool.load_from_point()
cylc/flow/task_pool.py:197: in load_from_point
    self.spawn_to_rh_limit(tdef, point, {flow_num})
cylc/flow/task_pool.py:787: in spawn_to_rh_limit
    self.get_or_spawn_task(
cylc/flow/task_pool.py:745: in get_or_spawn_task
    ntask = self.spawn_task(tdef.name, point, flow_nums, flow_wait)
cylc/flow/task_pool.py:1686: in spawn_task
    itask = self._get_task_proxy_db_outputs(
cylc/flow/task_pool.py:1793: in _get_task_proxy_db_outputs
    itask = TaskProxy(
cylc/flow/task_proxy.py:294: in __init__
    self.graph_children = generate_graph_children(tdef, self.point)
cylc/flow/taskdef.py:49: in generate_graph_children
    if is_abs and trigger.get_parent_point(point) != point:
cylc/flow/task_trigger.py:74: in get_parent_point
    point = get_point(self.cycle_point_offset).standardise()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = +P0

    def standardise(self):
        """Reformat self.value into a standard representation."""
        try:
            self.value = str(point_parse(self.value))
        except IsodatetimeError as exc:
            if self.value.startswith("+") or self.value.startswith("-"):
                message = WARNING_PARSE_EXPANDED_YEAR_DIGITS % (
                    WorkflowSpecifics.NUM_EXPANDED_YEAR_DIGITS)
            else:
                message = str(exc)
>           raise PointParsingError(type(self), self.value, message)
E           cylc.flow.exceptions.PointParsingError: Incompatible value for <class 'cylc.flow.cycling.iso8601.ISO8601Point'>: +P0: (incompatible with [cylc]cycle point num expanded year digits = 0 ?)

cylc/flow/cycling/iso8601.py:105: PointParsingError
```

</p>
</details> 

Also it was failing in a confusing way, so I altered the integration reftest function to better handle unexpected shutdown


MetRonnie:
MacOS tests failing for unrelated reasons; not tackled in this PR

tests/integration/conftest.py
```diff
@@ -590,56 +590,59 @@ async def _complete(
...
-            )
-
-    # restore regular shutdown logic
-    schd._set_stop = set_stop
+    def done():
+        if wait_tokens:
+            return not tokens_list
+        # otherwise wait for the scheduler to shut down
+        if not schd.contact_data:
+            return True
```
MetRonnie, line 628: This fixture was waiting for 60 secs for the workflow to shut down even if there was no contact data. If there is no contact data it must have already shut down due to an error or something (which bypasses the `set_stop` method).

tests/integration/conftest.py
```diff
@@ -569,7 +569,7 @@ async def _complete(
     Args:
         schd:
             The scheduler to await.
-        tokens_list:
+        wait_tokens:
```
wxtim, line 572: it might be overloaded, but I think `await_tokens` would be better English.

wxtim (approved):
Nice -One entirely subjective comment about naming which you can take or leave.

cylc/flow/scheduler.py
```diff
@@ -408,6 +400,10 @@ async def initialise(self):
 
         self.profiler = Profiler(self, self.options.profile_mode)
 
+        # Reset graph node parser singleton
+        # (mainly to prevent state leakage between integration tests):
+        GraphNodeParser.get_inst().clear()
```
oliver-sanders, line 405: > **mainly** to prevent state leakage between integration tests

Do you think this could cause any issues?

oliver-sanders: If this only matters to integration tests, then it would make more sense to do it in the integration test fixtures (where it is only run by integration tests).

Otherwise, if it makes sense to do this for non-integration purposes, it would probably make more sense to do it in `cylc.flow.config` (where the `GraphNodeParser` is actually used) rather than in the scheduler code?

MetRonnie: Should this be done at the start of `WorkflowConfig.load_graph()` do you think?

oliver-sanders: That sounds reasonable.

I don't think I had realised that there is a singleton here (seems a bit worrying for tests in general, not just integration ones). I wonder if we actually need it.

oliver-sanders: Taking a look at the code, the `GraphNodeParser` appears to cache offsets (e.g. `-P1D`) and nodes (e.g. `foo[-P1D]`).

If my understanding of this is correct (it might not be), then it should be fine to preserve this cache within a workflow and even between workflows (`foo[-P1D]` means the same thing whatever workflow it is in). If so, it doesn't really make sense to clear this cache?

This commit added the `clear()` into the `get_graph_raw` method (https://github.com/cylc/cylc-flow/commit/8972b8eeb44ecce49c3d222ea444fc358be886ae) for performance reasons (which coming to think of it doesn't really make sense as the exact same nodes will have to be re-parsed in the next call, mb, I reviewed it).

MetRonnie: > foo[-P1D] means the same thing whatever workflow it is in

Not if one workflow is in integer cycling mode, hence the test failures

oliver-sanders: > Not if one workflow is in integer cycling mode, hence the test failures

`Foo[-P1D]` has the exact same meaning whether the workflow is in integer, 365 or 360 day cycling mode, i.e. the task `foo` with an offset of `-P1D`. You won't find an offset of `P1D` in an integer cycling workflow, you'll find offsets like `-P1` instead, but having datetime entries in the cache shouldn't be a problem as they will never be retrieved.

MetRonnie: I feel like doing it here in `Scheduler.initialise()` which happens before loading the workflow config is safest for a 8.3.x PR?

oliver-sanders: > I've pasted the traceback in the OP

Cheers.

Here's what the GraphNodeParser is caching about a node:

```
  123             self._nodes[node] = ( 
  124                 name, offset, TaskTrigger.standardise_name(output),
  125                 offset_is_from_icp, offset_is_irregular, offset_is_absolute)
```

I wouldn't expect any of this to clash between integer and datetime cycling modes e.g:

```
foo[-P1D] -> ('foo', ISO8601Duration('-P1D'), 'succeeded', False, False, False)
foo[-P1] -> ('foo', IntegerDuration('-P1'), 'succeeded', False, False, False)
```

It it does clash (which it evidently does), there are two possibilities, zero-duration offsets being stored with their respective class rather than `None`:

```
foo -> ('foo', IntegerDuration('-P0'), 'succeeded', False, False, False)
Or
foo -> ('foo', ISO8601Duration('-P0'), 'succeeded', False, False, False)
```

Or absolute dependencies being stored with their exact cycle point as the offset (in the respective class):

```
foo[^] -> ('foo', ISO8601Duration('<icp>'), 'succeeded', False, False, True)
Or
foo[^] -> ('foo', IntegerDuration('<icp>'), 'succeeded', False, False, True)
```

oliver-sanders: If that reading is correct:

* This could never affect a real workflow, only test workflows.
* This could happen at validate time (because the graph nodes are also parsed then).

So ideally, we would have a fix that only impacts the tests (integration or otherwise) and works both for integration tests which actually start workflows AND any tests that load the config (e.g. validate tests) or poke at the GraphNodeParser.

MetRonnie: Ah I've just noticed we have an autouse fixture for resetting global state between tests

MetRonnie:
MacOS integration test failure is a flaky TUI test

oliver-sanders (approved):

cylc/flow/scheduler.py
oliver-sanders: Cheers.

PR #6264, opened 2024-07-25
Check run name as well as workflow name in Cylc install

### LINKED ISSUE #6246: validate full workflow ID
The workflow "name" is validated, the "run name" is not.

```console
$ cylc install generic -n f? -r r?
WorkflowFilesError: invalid workflow name 'f?' - can only contain: alphanumeric, `/`, `_`, `+`, `-`, `.`, `@`
$ cylc install generic -n f -r r?
INSTALLED f/r? from /net/home/h06/osanders/cylc-src/generic
```

wxtim:
Closes #6246 

**Check List**

### DIFF
--- a/changes.d/6264.fix.md
+++ b/changes.d/6264.fix.md
@@ -1,0 +1,1 @@
+Fix bug where `cylc install` failed to prevent invalid run names.
--- a/cylc/flow/install.py
+++ b/cylc/flow/install.py
@@ -65,7 +65,6 @@ from cylc.flow.workflow_files import (
     check_flow_file,
     get_cylc_run_abs_path,
     is_valid_run_dir,
-    check_reserved_dir_names,
     validate_workflow_name,
 )
 
@@ -285,13 +284,17 @@ def install_workflow(
     source = Path(expand_path(source)).resolve()
     if not workflow_name:
         workflow_name = get_source_workflow_name(source)
-    validate_workflow_name(workflow_name, check_reserved_names=True)
     if run_name is not None:
         if len(Path(run_name).parts) != 1:
             raise WorkflowFilesError(
                 f'Run name cannot be a path. (You used {run_name})'
             )
-        check_reserved_dir_names(run_name)
+        validate_workflow_name(
+            os.path.join(workflow_name, run_name),
+            check_reserved_names=True
+        )
+    else:
+        validate_workflow_name(workflow_name, check_reserved_names=True)
     validate_source_dir(source, workflow_name)
     run_path_base = Path(get_workflow_run_dir(workflow_name))
     relink, run_num, rundir = get_run_dir_info(
--- a/tests/unit/test_install.py
+++ b/tests/unit/test_install.py
@@ -542,3 +542,48 @@ def test_validate_source_dir(tmp_run_dir: Callable, tmp_src_dir: Callable):
     with pytest.raises(WorkflowFilesError) as exc_info:
         validate_source_dir(src_dir, 'ajay')
     assert "exists in source directory" in str(exc_info.value)
+
+
+def test_install_workflow_failif_name_name(tmp_src_dir):
+    """If a run_name is given validate_workflow_name is called on
+    the workflow and the run name in combination.
+    """
+    src_dir: Path = tmp_src_dir('ludlow')
+    # It only has a workflow name:
+    with pytest.raises(WorkflowFilesError, match='can only contain'):
+        install_workflow(src_dir, workflow_name='foo?')
+    # It only has a run name:
+    with pytest.raises(WorkflowFilesError, match='can only contain'):
+        install_workflow(src_dir, run_name='foo?')
+    # It has a legal workflow name, but an invalid run name:
+    with pytest.raises(WorkflowFilesError, match='can only contain'):
+        install_workflow(src_dir, workflow_name='foo', run_name='bar?')
+
+
+def test_install_workflow_failif_reserved_name(tmp_src_dir):
+    """Reserved names cause install validation failure.
+
+    n.b. manually defined to avoid test dependency on workflow_files.
+    """
+    src_dir = tmp_src_dir('ludlow')
+    is_reserved = '(that filename is reserved)'
+    reserved_names = {
+        'share',
+        'log',
+        'runN',
+        'suite.rc',
+        'work',
+        '_cylc-install',
+        'flow.cylc',
+        # .service fails because starting a workflow/run can't start with "."
+        # And that check fails first.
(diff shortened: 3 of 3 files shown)


changes.d/6264.fix.md
```diff
@@ -0,0 +1 @@
+Fix bug where  failed to prevent invalid run names.
```
MetRonnie, line 1:
```suggestion
Fix bug where `cylc install` failed to prevent invalid run names.
```

cylc/flow/install.py
```diff
@@ -285,13 +285,18 @@ def install_workflow(
     source = Path(expand_path(source)).resolve()
     if not workflow_name:
         workflow_name = get_source_workflow_name(source)
-    validate_workflow_name(workflow_name, check_reserved_names=True)
     if run_name is not None:
+        validate_workflow_name(
+            workflow_name + '/' + run_name,
```
MetRonnie, line 290:
```suggestion
            os.path.join(workflow_name, run_name),
```

cylc/flow/workflow_files.py
```diff
@@ -831,31 +832,36 @@ def check_deprecation(path, warn=True, force_compat_mode=False):
 
 
 def validate_workflow_name(
-    name: str, check_reserved_names: bool = False
+    name: str,
+    check_reserved_names: bool = False,
+    being_validated: Literal['workflow', 'workflow and run'] = 'workflow'
```
MetRonnie, line 837: I don't see the need for this argument; I would just keep it simple and drop any changes to this function

MetRonnie (commented):
Lint failing btw

cylc/flow/workflow_files.py
wxtim: I get your point, but if a user does

```
> cylc install --run-name  foo?
```

Then "invalid workflow name" seems unhelpful as a message. 

I'd be happy to modify the error messages without changing the functioning though.

wxtim: Nah, I agree, it's fine

cylc/flow/install.py
```diff
@@ -285,13 +285,15 @@ def install_workflow(
...
     if not workflow_name:
         workflow_name = get_source_workflow_name(source)
-    validate_workflow_name(workflow_name, check_reserved_names=True)
     if run_name is not None:
+        validate_workflow_name(os.path.join(workflow_name, run_name))
         if len(Path(run_name).parts) != 1:
             raise WorkflowFilesError(
                 f'Run name cannot be a path. (You used {run_name})'
             )
         check_reserved_dir_names(run_name)
```
MetRonnie, line 294: Still need to check reserved dir names on the whole path
```suggestion
        validate_workflow_name(
            os.path.join(workflow_name, run_name),
            check_reserved_names=True
        )
        if len(Path(run_name).parts) != 1:
            raise WorkflowFilesError(
                f'Run name cannot be a path. (You used {run_name})'
            )
```

MetRonnie (changes requested):

cylc/flow/install.py
MetRonnie: @wxtim I think there ought to be a test that covers this point

MetRonnie: I think there ought to be a test that covers this point

MetRonnie (changes requested):
See comment above; requested a test

tests/unit/test_install.py
```diff
@@ -542,3 +542,47 @@ def test_validate_source_dir(tmp_run_dir: Callable, tmp_src_dir: Callable):
...
+        # And that check fails first.
+        # '.service'
+    }
+    for name in reserved_names:
+        with pytest.raises(WorkflowFilesError, match=is_reserved):
+            install_workflow(src_dir, workflow_name='ok', run_name=name)
+        with pytest.raises(WorkflowFilesError, match=is_reserved):
+            install_workflow(src_dir, workflow_name=name)
+        with pytest.raises(WorkflowFilesError, match=is_reserved):
+            install_workflow(src_dir, workflow_name=name, run_name=name)
```
MetRonnie, line 588: This doesn't capture the point I raised. Needs:
```suggestion
            install_workflow(src_dir, workflow_name=name, run_name='ok')
```

MetRonnie (changes requested):

MetRonnie (dismissed):

cylc/flow/install.py
```diff
@@ -285,13 +285,18 @@ def install_workflow(
...
     if run_name is not None:
         if len(Path(run_name).parts) != 1:
             raise WorkflowFilesError(
                 f'Run name cannot be a path. (You used {run_name})'
             )
+        validate_workflow_name(
+            os.path.join(workflow_name, run_name),
+            check_reserved_names=True
+        )
         check_reserved_dir_names(run_name)
```
oliver-sanders, line 297: Isn't the subsequent `check_reserved_dir_names(run_name)` superfluous with `check_reserved_names=True`?

tests/unit/test_install.py
```diff
@@ -542,3 +542,47 @@ def test_validate_source_dir(tmp_run_dir: Callable, tmp_src_dir: Callable):
...
+    src_dir: Path = tmp_src_dir('ludlow')
+    # It only has a workflow name:
+    with pytest.raises(WorkflowFilesError, match='can only contain'):
+        install_workflow(src_dir, workflow_name='foo?')
+    # It only has a run name:
+    with pytest.raises(WorkflowFilesError, match='can only contain'):
+        install_workflow(src_dir, run_name='foo?')
+    # It has a legal workflow name, but an invalid run name:
+    with pytest.raises(WorkflowFilesError, match='can only contain'):
+        install_workflow(src_dir, workflow_name='foo', run_name='bar?')
```
oliver-sanders, line 560: With these sorts of tests where we are testing for failure, it is often a good idea to test that the thing you are testing can actually succeed when the conditions are right.

This avoids the situation where the test setup is broken causing the action to fail with the expected error, but for the wrong reasons (though the `match` kwarg provides some protection against this).

cylc/flow/install.py
MetRonnie: Ah yes, good spot (my suggestion https://github.com/cylc/cylc-flow/pull/6264#discussion_r1692729604 removed this but was not applied directly)

MetRonnie (approved):

MetRonnie (changes requested):
Ah, flake8 failing...

MetRonnie:
Lint workflow still failing...

oliver-sanders (approved):

PR #6267, opened 2024-07-30
`cylc play` (restart): fix bug affecting run host reinvocation after interactive upgrade

MetRonnie:
Fixes a bug not recorded in an issue. I think this bug has been present since the introduction of the interactive upgrade mechanism (#5074)

TL;DR - remote reinvocation would not work after interactively upgrading a workflow to a new Cylc version on the CLI.

#### Steps to reproduce

With `run hosts` configured in `global.cylc`

```cylc
[scheduler]
    [[run hosts]]
        available = <server1>, ...
```

Start a workflow and stop it, then fiddle the workflow DB to make it look like it was run with an older version:

```
sqlite3 wflow/.service/db 'UPDATE workflow_params SET value="8.1.7" WHERE key="cylc_version"'
```

Now restart it with `cylc play`, and type 'y' at the interactive upgrade prompt. When it re-invokes on the run host, you should see an error

```
This workflow was previously run with 8.1.7.
This version of Cylc is 8.3.4.dev
Use "--upgrade" to upgrade the workflow.
ERROR: ERROR: command terminated by signal 1: ssh ...
```

#### Check List

- [x] No dependency changes 
- [x] Tests are included
- [x] No docs entry needed

### DIFF
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -390,11 +390,7 @@ async def scheduler_cli(
 
     # check the workflow can be safely restarted with this version of Cylc
     db_file = Path(get_workflow_srv_dir(workflow_id), 'db')
-    if not _version_check(
-        db_file,
-        options.upgrade,
-        options.downgrade,
-    ):
+    if not _version_check(db_file, options):
         sys.exit(1)
 
     # upgrade the workflow DB (after user has confirmed upgrade)
@@ -404,7 +400,7 @@ async def scheduler_cli(
     _print_startup_message(options)
 
     # re-execute on another host if required
-    _distribute(options.host, workflow_id_raw, workflow_id, options.color)
+    _distribute(workflow_id_raw, workflow_id, options)
 
     # setup the scheduler
     # NOTE: asyncio.run opens an event loop, runs your coro,
@@ -474,8 +470,7 @@ async def _resume(workflow_id, options):
 
 def _version_check(
     db_file: Path,
-    can_upgrade: bool,
-    can_downgrade: bool
+    options: 'Values',
 ) -> bool:
     """Check the workflow can be safely restarted with this version of Cylc."""
     if not db_file.is_file():
@@ -491,7 +486,7 @@ def _version_check(
     )):
         if this < that:
             # restart would REDUCE the Cylc version
-            if can_downgrade:
+            if options.downgrade:
                 # permission to downgrade given in CLI flags
                 LOG.warning(
                     'Restarting with an older version of Cylc'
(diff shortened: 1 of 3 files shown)


MetRonnie:
Test failures just macos functional issues and flaky integration test fixed by #6257

wxtim (approved):
👍🏼

cylc/flow/scheduler_cli.py
```diff
@@ -612,8 +609,12 @@ def _distribute(host, workflow_id_raw, workflow_id, color):
         # Prevent recursive host selection
         cmd.append("--host=localhost")
 
+        # Ensure interactive upgrade carries over:
+        if options.upgrade and '--upgrade' not in cmd:
+            cmd.append('--upgrade')
```
oliver-sanders, line 614: Needs duplicating for `--downgrade`?

cylc/flow/scheduler_cli.py
```diff
@@ -390,11 +390,7 @@ async def scheduler_cli(
 
     # check the workflow can be safely restarted with this version of Cylc
     db_file = Path(get_workflow_srv_dir(workflow_id), 'db')
-    if not _version_check(
-        db_file,
-        options.upgrade,
-        options.downgrade,
-    ):
+    if not _version_check(db_file, options):
```
oliver-sanders, line 393: FYI, I purposefully did *not* propagate the `options` object down the call chain. IMO it's a good idea to abstract out the CLI interfaces higher up in the call chain to keep the underlying code more abstract. The more you propagate CLI logic down the call stack, the harder it is to, for example, open a Python API later, or create a wrapper command, you also need to mock more in order to test and typing is less effective too (we know we're getting a `Values` object, but we don't know whether it has an `upgrade` option.

Not a blocker, but not positive "tidying" IMO.

cylc/flow/scheduler_cli.py
MetRonnie: There isn't an interactive downgrade though

oliver-sanders: Gotcha.

oliver-sanders (approved):

cylc/flow/scheduler_cli.py
MetRonnie: I think ideally we want to only alter `options.upgrade` in the case of an interactive upgrade, so the simplest way of doing this is passing in the `options` object to `_version_check`. 

It is a shame about the type-looseness of it, but at least it is relatively easy to test these functions by importing and using `RunOptions` (no need for mocking).

https://github.com/cylc/cylc-flow/blob/6eb8f61c5b9777cb21ac68eff13aa4f0722f691a/cylc/flow/option_parsers.py#L569-L570

oliver-sanders: `options.update = _version_check(db_file, options.upgrade, options.downgrade)` would have done it?

MetRonnie: There are pathways it returns `True` without interactive upgrade

PR #6278, opened 2024-08-01
examples: extending workflow

oliver-sanders:
* Fix incorrect heading level
* Add a note about scheduling tasks to run at the "stop after cycle point".

**Check List**

### DIFF
--- a/cylc/flow/etc/examples/extending-workflow/index.rst
+++ b/cylc/flow/etc/examples/extending-workflow/index.rst
@@ -1,5 +1,5 @@
 Extending Workflow
-==================
+------------------
 
 .. cylc-scope:: flow.cylc[scheduling]
 
@@ -24,7 +24,7 @@ solution avoids this issue.
 
 
 Example
--------
+^^^^^^^
 
 .. admonition:: Get a copy of this example
    :class: hint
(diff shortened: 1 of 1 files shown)


hjoliver (commented):
Looks fine, but I wonder if we should add:
- if you stop the workflow at any point before the final "stop task" (even after some, but not all, stop tasks have run) and restart with a later stop point, the original stop task branch will be truncated, and a new one "rebased" onto the new stop point. 
- add explicit dependence on the main cycling graph so that the example stop tasks really don't start running until just before shutdown. As it stands, they will start running as soon as the runahead limit allows, concurrent with earlier cycles.
 
Both of these points should be kinda obvious, but it doesn't hurt to make examples like this as explicit as possible.

oliver-sanders:
> the original stop task branch will be truncated, and a new one "rebased" onto the new stop point

Sorry, I don't understand this.

I don't think the `stop after cycle point` config has any impact on the task pool does it? Restarting with a later stop CP doesn't appear to have any side effects?

hjoliver:
I mean if the stop-task sun-graph has already started, but not finished, when the user stops or kills the workflow ... and then restarts it with a new stop-after-cycle-point.

oliver-sanders:
Does this not have the same impact as changing the stop cycle (note cycle not task) at runtime (e.g. via `cylc stop` or `cylc reload`)?

wxtim:
> I mean if the stop-task sub-graph has already started, but not finished, when the user stops or kills the workflow ... and then restarts it with a new stop-after-cycle-point.

In the example the stop cycle point is hardcoded and the stop cycle point task will remain 3000 unless the user changes it in the config and reloads.

oliver-sanders:
> What I think Hilary may be getting at, is that in the example the stop cycle point is hardcoded and the stop cycle point task will remain 3000 unless the user changes it in the config and reloads

The same applies to the initial/start/final cycle points?

wxtim (approved):

hjoliver:
OK I'll try to explain myself better. 

I'm just suggesting we should tell users the potential consequences of basing a sub-graph off of a dynamically set cycle point.

Original graph with `stop_cycle = 3`:
```ini
P1 = "foo[-P1] => foo => bar"
R1/3 = "bar => a => b => c & d". # {% set stop_cycle = 3 %}
```
image

But if I run this then shut it down while `3/b` is running, i.e. part-way through the stop-cycle sub-graph, then I restart it with `stop_cycle = 6` - what ends up running is this:

image

It might be obvious to us that the original stop-cycle sub-graph gets truncated and whole new one gets based on cycle 6, but I'm not sure it'll be so obvious to all users.

hjoliver:
I also think the example should have intercycle dependence (like mine above) and anchor the stop-cycle sub-graph to the main graph (like mine above), so that (in lieu of #5090) the stop-cycle sub-graph runs just *before the workflow shuts down at the `stop after cycle point`* and not earlier. That would almost certainly be wanted in a real use case.

oliver-sanders:
> I'm just suggesting we should tell users the potential consequences of basing a sub-graph off of a dynamically set cycle point.

Gotcha.

The same thing applies to the final cycle point (e.g. `R1/$`) so should be expected, but fair enough, I'll add a note.

oliver-sanders:
@hjoliver, done.

hjoliver (approved):

Issue #5643, opened 2023-07-25
remove: extend beyond the task pool


oliver-sanders:
Implement the [cylc remove proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-remove.md):

* Extend `cylc remove` to cover historical (`n<0`) tasks.
* Extend `cylc remove` to cover active `n=0` tasks.
* Extend `cylc remove` to cover future (`n>0`) tasks.
* Update CLI and GraphQL schema to match any changes.
* Update this note in cylc-doc: https://github.com/cylc/cylc-doc/blob/3c2b067e27a4c8e5187fc21a117882466fc0d4f6/src/user-guide/writing-workflows/suicide-triggers.rst?plain=1#L53-L60

MetRonnie:
```
# task_states table
name        cycle       flow_nums   time_created     
----------  ----------  ----------  ---------------
a           1           [1]         <10 mins ago>  
a           1           [1, 2]      <2 mins ago>     <---
```

How should we handle the case where the user removes task `1/a` from flow 2 (a.k.a. remove flow 2 from task `1/a`)?

Bearing in mind `PRIMARY KEY(name, cycle, flow_nums)`, we have 2 choices:
1. Remove row 1
2. Remove row 2

oliver-sanders:
If a task is removed from the pool as completed, then subsequently re-triggered by the user, then the most recent entry will override the older one.

This is the same case by a different mechanism, though a somewhat awkward case. The job state info will be preserved (tied to the submit number).

MetRonnie:
So option 1

oliver-sanders:
The behaviour for re-triggering might be closer to merge than replace (can't remember what else is in the table), otherwise yes.

MetRonnie:
The full schema for `task_states` looks like
```
name   cycle  flow_nums  time_created   time_updated   submit_num   status   flow_wait   is_manual_submit
------ ------ ---------- -------------- -------------- ------------ -------- ----------- ----------------
TEXT   TEXT   TEXT       TEXT           TEXT           INTEGER      TEXT     INTEGER     INTEGER
```

> Note that this will probably happen automatically after the new task has been added to the pool?

Not sure what you mean; I am talking about removing a flow from a historical task

oliver-sanders:
(1) is good

MetRonnie:
```
# task_states table
name        cycle       flow_nums   time_created    ... status  
[...]

PR #1775, opened 2016-04-01
Task state code refactor.

hjoliver:
Refactor the code associated with handling of task state/status.

[updated 13 April - original lines ~~crossed out~~]

Motivation: working on another current pull request, I was bitten by a typo in a task status string comparison, which wouldn't happen if we used (constant) variables to refer to states like "waiting" etc.  (which would be better for internationalization too, if it comes to that at some point...).  Further, each task proxy currently holds its status in a "task_state" object that is (a) next to useless - it contains just a status string, and some minimal logic that is largely duplicated in wrapper methods in the task proxy class anyway; and (b) also holds "spawned/not spawned" status in a silly two-key dict along with "status", which doesn't make much  sense.  This is all dates back a long way, before I learned a few things off @cylc/core members :grimacing: 

So: this PR does the following:
 * separates spawning from task_state
 * deprecates "cylc reset --state=spawn" to a new command "cylc spawn"
 * ~~gets rid of the task state objects (now just a status string held by task proxies)~~
 * the `task_state` module now defines a `TaskState` class that (for each task proxy) holds the simple status string, and prerequisites and outputs (which are intimately connected to task status), plus associated logic.
 * the task state module also defines constants for each status string, e.g. `STATUS_WAITING = 'waiting'`, and these are now used throughout the code. 
 * replaces the `state.is_currently()` and `state.get_status()` methods with static set membership tests, for performance reasons (as pointed out by @benfitzpatrick below).
 
~~@matthewrmshin - this isn't quite ready - I still need to use STATUS_WAITING etc. throughout the code instead of hard-coded strings "waiting" etc.; add a test for "cylc spawn", and address a few TODO comments left in the code (mainly for checking valid status at various points).    However, as this grew a bit bigger than I was expecting, I just wanted to post the branch as-is before the weekend in case of any fundamental objections, or suggestions for improvement.  Test battery passes on this branch at the moment.~~

PR #6287, opened 2024-08-06
`TaskState`: remove cached prereq satisfaction in preparation for `cylc remove`

MetRonnie:
These manually cached satisfaction states have the possibility to drift out of sync with individual `Prerequisite` satisfaction states, especially with `cylc remove` work in progress #5643 (`cylc remove` will have to unset any satisfied prerequisites that the removed task provided).

It looks like they are barely used in the first place (at least not anymore):
- `_is_satisfied` only has any savings at this call to `TaskState.prerequisites_are_not_all_satisfied()`:
    https://github.com/cylc/cylc-flow/blob/80fa008e18a6feb8ecb21123bde16c048fd1b599/cylc/flow/task_pool.py#L1755-L1760

    which is the case for absolute triggers e.g. `foo[2001] => bar` which are not so common?

- `_suicide_is_satisfied` only has any savings at the call above and this call:
    https://github.com/cylc/cylc-flow/blob/80fa008e18a6feb8ecb21123bde16c048fd1b599/cylc/flow/task_pool.py#L1425-L1430

    suicide triggers are also not so common.

I am not sure this is strictly necessary for `cylc remove` (but not happy about the current situation), hence putting it up for review standalone first.

These cached values were first introduced in #1775. There was an issue #1857 mentioning it - looks like it was the cached states drifting out of sync that caused the problem in that case, which is why I am unhappy about it generally. 

**Check List**

- [x] No dependency changes 
- [x] No tests needed
- [x] No changelog needed as not visible to users
- [x] No docs needed
- [x] On master branch as code refactor

### DIFF
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1990,7 +1990,7 @@ class TaskPool:
 
         """
         if prereqs == ["all"]:
-            itask.state.set_all_satisfied()
+            itask.state.set_prerequisites_all_satisfied()
         else:
             # Attempt to set the given presrequisites.
             # Log any that aren't valid for the task.
--- a/cylc/flow/task_state.py
+++ b/cylc/flow/task_state.py
@@ -309,10 +309,9 @@ class TaskState:
         """
         valid: Set[Tokens] = set()
         for prereq in (*self.prerequisites, *self.suicide_prerequisites):
-            yep = prereq.satisfy_me(outputs)
-            if yep:
-                valid = valid.union(yep)
-                continue
+            valid.update(
+                prereq.satisfy_me(outputs)
+            )
         return valid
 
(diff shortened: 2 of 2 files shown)


oliver-sanders:
Makes sense, we have always assumed that prereqs cannot revert to unsatisfied after being satisfied, however, `cylc remove` will need to do this.

I expect this cache was introduced for efficiency reasons, may be worth checking that it doesn't perform poorly for tasks with large numbers of prereqs (though I'm not expecting much change by the looks of the code).

oliver-sanders (approved):
Tested with the "Diamond" workflow. CPU usage actually went down slightly as a result of taking out the caching. It definitely wasn't doing anything useful.

hjoliver (approved):

PR #6297, opened 2024-08-13
Tweak handling of daemon exit to avoid macos traceback.

### LINKED ISSUE #6291: sys.exit traceback running 8.3+ on MacOS
Traceback appears in terminal on start-up in detaching mode, since 8.3.0 (8.2.7 was fine)
The workflow still runs, but a similar traceback appears in the scheduler log on shutdown

```
$ cylc play bug/run20

2024-08-08T11:48:35+12:00 INFO - Extracting job.sh to /Users/oliver/cylc-
    run/bug/run20/.service/etc/job.sh
  File "/Users/oliver/miniforge3/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/Users/oliver/miniforge3/lib/python3.10/asyncio/base_events.py", line 636, in run_until_complete
    self.run_forever()
  File "/Users/oliver/miniforge3/lib/python3.10/asyncio/base_events.py", line 603, in run_forever
    self._run_once()
  File "/Users/oliver/miniforge3/lib/python3.10/asyncio/base_events.py", line 1909, in _run_once
    handle._run()
  File "/Users/oliver/miniforge3/lib/python3.10/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/Users/oliver/cylc/cylc-flow/cylc/flow/scheduler_cli.py", line 417, in scheduler_cli
    daemonize(scheduler)
  File "/Users/oliver/cylc/cylc-flow/cylc/flow/daemonize.py", line 121, in daemonize
    sys.exit(0)
SystemExit: 0
```

hjoliver:
Close #6291 

Avoid inconsequential but frightening tracebacks at startup and shutdown on macOS, in detaching mode.

Uses `os._exit()`, which *I think* should be safe in this context:

```
Help on built-in function _exit in module posix:

_exit(status)
    Exit to the system with specified status, without normal exit processing.
```

**Check List**

- [x] Tests NOT NEEDED (change is non functional)

### DIFF
--- a/changes.d/6297.fix.md
+++ b/changes.d/6297.fix.md
@@ -1,0 +1,1 @@
+Prevent inconsequential traceback on startup and shutdown, since version 8.3.0, on macOS.
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -416,11 +416,21 @@ async def scheduler_cli(
         from cylc.flow.daemonize import daemonize
         try:
             daemonize(scheduler)
-        except SystemExit:
+        except SystemExit as exc:
             # This is supposed to happen on daemonization, but raising
             # SystemExit or calling sys.exit() here results in a traceback
             # to the terminal at start-up on macOS.
-            os._exit(0)
+            excode: int
+            if isinstance(exc.code, int):
+                excode = exc.code
+            elif isinstance(exc.code, str):
+                try:
+                    excode = int(exc.code)
+                except ValueError:
+                    excode = 1
+            else:
+                excode = 1
+            os._exit(excode)
 
     # setup loggers
     _open_logs(
@@ -443,14 +453,12 @@ async def scheduler_cli(
     # NOTE: we must clean up all asyncio / threading stuff before exiting
     # NOTE: any threads which include sleep statements could cause
     #       sys.exit to hang if not shutdown properly
-    msg = "DONE"
-    if ret != 0:
-        msg += f" (exit {ret})"
-    LOG.info(msg)
+    LOG.info("DONE")
     close_log(LOG)
-    # Calling sys.exit() here results in a traceback to the LOG for normal
-    # shutdown on macOS, for a detaching scheduler.
-    os._exit(0)
+    # The sys.exit call here results in a traceback to the LOG on normal
+    # shutdown on macOS for a detaching scheduler, but we can't resort to
+    # os._exit() in this case - it bypasses needed shutdown processing.
+    sys.exit(ret)
 
 
 async def _resume(workflow_id, options):
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -22,6 +22,7 @@ from functools import lru_cache
 from itertools import zip_longest
 from pathlib import Path
 from shlex import quote
+import os
 import sys
 from typing import TYPE_CHECKING
 
@@ -413,7 +414,13 @@ async def scheduler_cli(
     #       ensure you have tidied up all threads etc before daemonizing
     if not options.no_detach:
         from cylc.flow.daemonize import daemonize
-        daemonize(scheduler)
+        try:
+            daemonize(scheduler)
+        except SystemExit:
+            # This is supposed to happen on daemonization, but raising
+            # SystemExit or calling sys.exit() here results in a traceback
+            # to the terminal at start-up on macOS.
+            os._exit(0)
 
     # setup loggers
     _open_logs(
@@ -436,9 +443,14 @@ async def scheduler_cli(
     # NOTE: we must clean up all asyncio / threading stuff before exiting
     # NOTE: any threads which include sleep statements could cause
     #       sys.exit to hang if not shutdown properly
-    LOG.info("DONE")
+    msg = "DONE"
+    if ret != 0:
+        msg += f" (exit {ret})"
+    LOG.info(msg)
     close_log(LOG)
-    sys.exit(ret)
+    # Calling sys.exit() here results in a traceback to the LOG for normal
+    # shutdown on macOS, for a detaching scheduler.
+    os._exit(0)
 
 
 async def _resume(workflow_id, options):


oliver-sanders:
> Uses `os._exit()`, which I think should be safe in this context:

The issue with `os._exit` is that it bypasses shutdown logic (kinda like issuing SIGKILL).

>  os._exit(n)
>
>    Exit the process with status n, without calling cleanup handlers, flushing stdio buffers, etc.
>
> -- https://docs.python.org/3/library/os.html#module-os

This means `with ...` blocks won't call their `__exit__` handlers. E.G. open files, TCP or database connections might not be correctly closed.

hjoliver:
Yes I know, but in this context - exit from fork 2 during daemonizatiom - it seemed to me this should be Ok. Having said that I have not checked with any error conditions yet.

hjoliver:
> Yes I know, but in this context - exit from fork 2 during daemonization - it seemed to me this should be Ok

Might be true. I backed out of the second change (daemon exit, which leaves a traceback in the log) and now all test are passing. The log traceback on shutdown is less serious. The one at start-up in the terminal is pretty nasty and makes it looks as if something has gone horribly wrong.

Question is, can anything go wrong in the daemonizing fork logic that really requires "proper exit processing"?

oliver-sanders:
> Question is, can anything go wrong in the daemonizing fork logic that really requires "proper" exit processing?

It's going to be very hard to say for sure. A lot of things happen in this process before forking occurs, any `atexit` / `__exit__` / `finally` blocks registered in this code will not be run. E.G. we load Cylc configuration files, import libraries, [activate code coverage](https://github.com/cylc/cylc-flow/blob/b8be2e3ef5b37288df5de4d62642b0d84e8eb02d/cylc/flow/scripts/cylc.py#L538-L630), create the Scheduler object, open the workflow database and more.

hjoliver:
Fair enough, hopefully we can find a better solution (than this one) then.

oliver-sanders:
I keep forgetting to try this out at home (sorry), one thing I was wondering, is whether we could capture this `SystemExit` via `asyncio`'s error handling and re-raise it in synchronous code?

oliver-sanders:
Would also like to confirm that it's an interaction between `asynio.run` and `os.fork` that's causing the traceback to be displayed (and not a side effect of `os.fork` alone). If we can boil this down to a simple example it might be work raising it on StackOverflow or the Python forums?

hjoliver:
> one thing I was wondering, is whether we could capture this SystemExit via asyncio's error handling and re-raise it in synchronous code?

Beyond my asyncio knowledge at this stage, I'll try to figure it out if I can squeeze some time.

> ...  is behaving differently on Mac to Linux, it may conceivably be a bug

yeah I was wondering that myself, it's certainly odd that this only happens on Mac.

oliver-sanders:
Ok, I've managed to reproduce this with Python 3.9 from conda-forge. Interestingly I get a slightly different traceback:

```
...
SystemExit: 0

...
OSError: [Errno 9] Bad file descriptor
sys:1: RuntimeWarning: coroutine 'BaseEventLoop.shutdown_asyncgens' was never awaited
one/run2: my-address PID=12345
```

I've also managed to replicate the same traceback with a minmal example for testing:

```python
import os
import sys
from time import sleep

async def run():
    pid = os.fork()

    if pid > 0:
        print('#1')
        sys.exit()

    print('#2')
    sleep(10)

    print('#2')
    sys.exit()

import asyncio
asyncio.run(run())
```

oliver-sanders:
Now replicated without explicitly calling `sys.exit` at all:

```python
import asyncio
import os

loop = asyncio.get_event_loop()

async def main():
    os.fork()

try:
    loop.create_task(main())
    loop.run_forever()
finally:
    loop.close()
```

When I run this `SystemExit` does not appear in the traceback at all (for either process).

It looks like asyncio teardown logic is getting very confused by the forking.

oliver-sanders:
Ok, might be working towards a solution now.

We are trying to run async code, create a fork, then run more async code. Essentially like this:

```
import asyncio
import os

loop = asyncio.get_event_loop()
loop.run_until_complete(asyncio.sleep(0))
loop.close()

os.fork()

loop = asyncio.get_event_loop()
loop.run_until_complete(asyncio.sleep(0))
loop.close()
```

This fails. You can try running whatever asyncio teardown logic you want, I can't seem to fix it.

But apply this diff:

```diff
 
-loop = asyncio.get_event_loop()
+loop = asyncio.new_event_loop()
 loop.run_until_complete(asyncio.sleep(0))
```

And it works fine.

This seems to keep the two loops separate, whereas just closing them doesn't.

oliver-sanders:
And that gives us an answer!

With this patch `cylc play` works with no error:

```diff
diff --git a/cylc/flow/scheduler_cli.py b/cylc/flow/scheduler_cli.py
index 059482068..b52774891 100644
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -351,7 +351,7 @@ def _open_logs(id_: str, no_detach: bool, restart_num: int) -> None:
     )
 
 
-async def scheduler_cli(
+async def scheduler_cli_1(
     options: 'Values',
     workflow_id_raw: str,
     parse_workflow_id: bool = True
@@ -408,6 +408,13 @@ async def scheduler_cli(
     scheduler = Scheduler(workflow_id, options)
     await _setup(scheduler)
 
+    return scheduler, workflow_id
+
+
+def scheduler_cli_2(
+    options: 'Values',
+    scheduler: Scheduler,
+) -> None:
     # daemonize if requested
     # NOTE: asyncio event loops cannot persist across daemonization
     #       ensure you have tidied up all threads etc before daemonizing
@@ -415,6 +422,13 @@ async def scheduler_cli(
         from cylc.flow.daemonize import daemonize
         daemonize(scheduler)
 
+
+async def scheduler_cli_3(
+    options: 'Values',
+    workflow_id: str,
+    scheduler: Scheduler,
+    parse_workflow_id: bool = True,
+) -> None:
     # setup loggers
     _open_logs(
         workflow_id,
@@ -423,14 +437,7 @@ async def scheduler_cli(
     )
 
     # run the workflow
-    if options.no_detach:
-        ret = await _run(scheduler)
-    else:
-        # Note: The daemonization messes with asyncio so we have to start a
-        # new event loop if detaching
-        ret = asyncio.run(
-            _run(scheduler)
-        )
+    ret = await _run(scheduler)
 
     # exit
     # NOTE: we must clean up all asyncio / threading stuff before exiting
@@ -659,4 +666,6 @@ async def _run(scheduler: Scheduler) -> int:
 @cli_function(get_option_parser)
 def play(parser: COP, options: 'Values', id_: str):
     """Implement cylc play."""
-    return asyncio.run(scheduler_cli(options, id_))
+    scheduler, workflow_id = asyncio.run(scheduler_cli_1(options, id_))
+    scheduler_cli_2(options, scheduler)
+    asyncio.run(scheduler_cli_3(options, workflow_id, scheduler))
```

The `scheduler_cli` method has had to be split up into three parts:
* The first async bit.
* The daemonize bit.
* The second async bit.

Not quite as nice to read, but as a result the horrible running `asyncio.run` from within an `async` function bit is gone!

The bad news is that it will be a bit of work to adapt `cylc vip` and `cylc vr` to fit this new, more complex interface. It's tricky because both of these compound commands also need to run their own async code 😨.

hjoliver:
Whoa, that's great - nice work 🥇

hjoliver:
@oliver-sanders - I'll mark this as superseded, it can be closed once you've put up a replacement or updated #6291

oliver-sanders:
I think #6310 should do the trick.

PR #6299, opened 2024-08-13
schema: minor docs changes

oliver-sanders:
Currently poking around the schema looking at possible ways out of the graphql deadlock. Noticed some small things:

* Consistent string quotation.
* Consistent termination punctuation.
* Correct some ReST literals (double backquotes).
* Correct one incorrectly closed markdown literal.

**Check List**

### DIFF
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -2229,7 +2229,10 @@ def delta_subs(root, info: 'ResolveInfo', **args) -> AsyncGenerator[Any, None]:
 
 class Pruned(ObjectType):
     class Meta:
-        description = 'WFS Nodes/Edges that have been removed.'
+        description = sstrip('''
+            Objects (e.g. workflows, tasks, jobs) which have been removed from
+            the store.       
+        ''')   
     workflow = String()
     families = graphene.List(String, default_value=[])
     family_proxies = graphene.List(String, default_value=[])
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -577,9 +577,10 @@ def resolve_mapping_to_list(root, info, **args):
 # Types:
 class NodeMeta(ObjectType):
     class Meta:
-        description = """
-Meta data fields,
-including custom fields in a generic user-defined dump"""
+        description = sstrip("""
+            Meta data fields, including custom fields in a generic user-defined
+            dump.
+        """)
     title = String(default_value=None)
     description = String(default_value=None)
     URL = String(default_value=None)
@@ -588,7 +589,7 @@ including custom fields in a generic user-defined dump"""
 
 class TimeZone(ObjectType):
     class Meta:
-        description = """Time zone info."""
+        description = 'Time zone info.'
     hours = Int()
     minutes = Int()
     string_basic = String()
@@ -597,7 +598,7 @@ class TimeZone(ObjectType):
 
 class Workflow(ObjectType):
     class Meta:
-        description = """Global workflow info."""
+        description = 'Global workflow info.'
     id = ID()  # noqa: A003 (required for definition)
     name = String(
         description='The workflow ID with the ~user/ prefix removed.',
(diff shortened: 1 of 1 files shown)


cylc/flow/network/schema.py
```diff
@@ -985,35 +987,36 @@ def resolve_task_id(root, info, **args):
 
 class Prerequisite(ObjectType):
     class Meta:
-        description = """Task prerequisite."""
+        description = 'Task prerequisite.'
     expression = String()
     conditions = graphene.List(
         Condition,
-        description="""Condition monomers of a task prerequisites.""")
+        description='Condition monomers of a task prerequisites.'
```
hjoliver, line 994:
```suggestion
        description='Condition monomers of a task prerequisite.'
```

cylc/flow/network/schema.py
```diff
@@ -2224,7 +2229,7 @@ def delta_subs(root, info: 'ResolveInfo', **args) -> AsyncGenerator[Any, None]:
 
 class Pruned(ObjectType):
     class Meta:
-        description = """WFS Nodes/Edges that have been removed."""
+        description = 'WFS Nodes/Edges that have been removed.'
```
hjoliver, line 2232: What's "WFS" ?  Workflow Service?   Should be "scheduler"? 

```
> git grep WFS
cylc/flow/network/schema.py:        description = """WFS Nodes/Edges that have been removed."""
cylc/flow/network/schema.py:        description = """Grouped deltas of the WFS publish"""
```

hjoliver (approved):

cylc/flow/network/schema.py
oliver-sanders: Absolutely no idea!

oliver-sanders:
```suggestion
        description = sstrip('''
            Objects (e.g. workflows, tasks, jobs) which have been removed from
            the store.       
        ''')   
```

cylc/flow/network/schema.py
oliver-sanders: "Condition monomers" - that's a good one!

cylc/flow/network/schema.py
hjoliver: Ping @dwsutherland - I think you may have been the original author?

oliver-sanders: This is the description for pruned deltas.

These list Cylc "objects" that have drifted out of the window.

I expect WFS probably meant Workflows?

hjoliver: I have feeling it date backs to early Cylc 8 planning days when we were considering various other names for what is now just the "scheduler" (Workflow Service = WFS?).

dwsutherland: Yeah.. Can't recall 😆 

But "Workflow's" might have been it ...

hjoliver: Ok screw it, merge Oliver's tweak.

cylc/flow/network/schema.py
hjoliver: Yeah who's the wordsmith responsible for that gem?

dwsutherland (approved):
LGTM

dwsutherland:
> ways out of the graphql deadlock.

I think we need to move to more active projects, i.e. only graphql-core, graphene/graphene-tornado  and drop graphql-ws by creating our own solution (that can be used in zmq too).. Or see if there's any other viable graphql ecosystem out there..

The previous generation of graphql-core/graphql-ws is stuck on the `promise` library that was a way of making it more JS like (as it was a port from there).. I believe the new graphql-core uses asyncio more..

(I could be wrong.. but this thing floats around my head a bit too)

oliver-sanders:
@dwsutherland, yep.

I've put some notes here: https://github.com/cylc/cylc-uiserver/issues/333

tldr;

* GraphQL core v3 is a complete re-write of GraphQL core v2, so about as much effort as porting to another server.
* Other servers exist, but it would appear they have GraphQL core as a dependency so are likely wrappers that provide integration support.
* There are no active servers that support Tornado so we will need to develop a bespoke integration for this bit at least.
* It's possible to stick with Graphene, but we could move to another code-driven schema definition framework if it helped.

MetRonnie:
(reminder: do check GH Actions failures before merging)

image

PR #6305, opened 2024-08-15
Remove `c` char from prereq API dump

MetRonnie:
https://github.com/cylc/cylc-ui/pull/1886#discussion_r1718434619

Not sure why "c" was ever included in the prerequisite alias. It is somewhat confusing when you have single-letter task names as we do in test workflows.

**Check List**

- [x] No dependency changes 
- [x] Tests are updated.
- [x] No changelog or docs needed as not user-facing

### DIFF
--- a/cylc/flow/prerequisite.py
+++ b/cylc/flow/prerequisite.py
@@ -16,7 +16,6 @@
 
 """Functionality for expressing and evaluating logical triggers."""
 
-import math
 import re
 from typing import Iterable, Set, TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


oliver-sanders:
So long as this doesn't break anything, we're good to remove it.

cylc/flow/scripts/show.py
```diff
@@ -307,12 +307,12 @@ async def prereqs_and_outputs_query(
                     prereqs.append([
                         True,
                         '',
-                        item['expression'].replace('c', ''),
+                        item['expression'],
                         item['satisfied']
                     ])
                 for cond in item['conditions']:
                     if multi_cond and not options.list_prereqs:
-                        prefix = f'\t{cond["exprAlias"].strip("c")} = '
```
oliver-sanders, line 315: Gotta leave this in for now I'm afraid.

MetRonnie: Ah right, forgot that `cylc show` needs the workflow in question to be running (which might be on an older version)

oliver-sanders (approved):

dwsutherland (approved):
LGTM

dwsutherland:
> Not sure why "c" was ever included in the prerequisite alias. It is somewhat confusing when you have single-letter task names as we do in test workflows.

It was something that was present in the cylc7 api data/dump.. But really was just a shorthand for condition

PR #6309, opened 2024-08-20
GH Actions: include hidden files when uploading `~/cylc-run` artifact

MetRonnie:
https://github.com/actions/upload-artifact/issues/602


.github/workflows/test_fast.yml
```diff
@@ -75,6 +75,7 @@ jobs:
         with:
           name: cylc-run (${{ matrix.os }} py-${{ matrix.python-version }})
           path: ~/cylc-run/
+          include-hidden-files: true
```
MetRonnie, line 78: I tested this and it's ok to add this already even though the option is not yet included in `actions/upload-artifact@v4`

wxtim (approved):

PR #6311, opened 2024-08-20
Flake8: remove wrongly ignored error code

MetRonnie:
Stop ignoring this flake8-bugbear rule:

> B028: No explicit stacklevel argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.

The code that was meant to be ignored according to the comment in `tox.ini` was:

> B907: Consider replacing f"'{foo}'" with f"{foo!r}" which is both easier to read and will escape quotes inside foo if that would appear. The check tries to filter out any format specs that are invalid together with !r. If you're using other conversion flags then e.g. f"'{foo!a}'" can be replaced with f"{ascii(foo)!r}". Not currently implemented for python<3.8 or str.format() calls.

This is an opinionated warning which is not enabled by default

**Check List**

- [x] No dependency changes, docs, tests, changelog entry needed

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,9 +8,6 @@ ignore=
     W504
     ; "experimental" SIM9xx rules (flake8-simplify)
     SIM9
-    ; suggests using f"{!r}" instead of manual quotes (flake8-bugbear)
-    ; Doesn't work at 3.7
-    B028
 
 per-file-ignores=
     ; TYPE_CHECKING block suggestions


oliver-sanders (approved):
This error code was correctly ignored, however, bugbear then renamed it from `B028` to `B907` (making it optional).

> @oliver-sanders felt that !r was excessively obscure

It was, and bugbear conceded this:

> 23.1.17
>
>    Rename B028 to B907, making it optional/opinionated.

Then, at a later date, they then re-used the code for a new rule:

> 23.2.13
>
> Add B028: Check for an explicit stacklevel keyword argument on the warn method from the warnings module.

https://github.com/PyCQA/flake8-bugbear?tab=readme-ov-file#23117

:facepalm:

The new `B907` is probably reasonable, though it doesn't affect us as Cylc uses its own warning system.

PR #6316, opened 2024-08-22
Cylc VR: Validate should not change options

### LINKED ISSUE #6262: `cylc vr` bug with `initial cycle point = now`
```ini
[scheduling]
    initial cycle point = now
    [[graph]]
        PT1M = "foo"
[runtime]
    [[foo]]
```

1. `cylc vip bug`
2. `cylc stop bug`
3. (add a comment to `flow.cylc`)
4. `cylc vr bug`

Result:
```
ERROR - Workflow shutting down - InputError: option --icp is not valid for restart
```

This only happens for `initial cycle point = now` 🤔

wxtim:
Closes #6262  

Validate may alter the content of options objects passed to it: In this case by replacing `icp = now` with `icp = <actual time>` which is an invalid restart option.

### Fix

Pass validate a copy of the options object to prevent changes made to the options by validate affecting subsequent steps.

Whilst in principle we probably shouldn't let the options object be mutated by the `cylc validate` this shouldn't affect us anywhere except in `cylc vr`.

**Check List**

- [x] Check whether this is a problem in `cylc vip`.

### DIFF
--- a/changes.d/6316.fix.md
+++ b/changes.d/6316.fix.md
@@ -1,1 +1,1 @@
-Prevent validate step of VR changing CLI options.
+Fixed bug in `cylc vr` where an initial cycle point of `now`/`next()`/`previous()` would result in an error.
--- a/changes.d/6316.fix.md
+++ b/changes.d/6316.fix.md
@@ -1,0 +1,1 @@
+Prevent validate step of VR changing CLI options.
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -690,7 +690,7 @@ class WorkflowConfig:
         Sets:
             self.initial_point
             self.cfg['scheduling']['initial cycle point']
-            self.options.icp
+            self.evaluated_icp
         Raises:
             WorkflowConfigError - if it fails to validate
         """
@@ -710,10 +710,11 @@ class WorkflowConfig:
                     icp = ingest_time(orig_icp, get_current_time_string())
                 except IsodatetimeError as exc:
                     raise WorkflowConfigError(str(exc))
-        if orig_icp != icp:
+        self.evaluated_icp = None
+        if icp != orig_icp:
             # now/next()/previous() was used, need to store
             # evaluated point in DB
-            self.options.icp = icp
+            self.evaluated_icp = icp
         self.initial_point = get_point(icp).standardise()
         self.cfg['scheduling']['initial cycle point'] = str(self.initial_point)
 
--- a/tests/functional/cylc-combination-scripts/09-vr-icp-now/flow.cylc
+++ b/tests/functional/cylc-combination-scripts/09-vr-icp-now/flow.cylc
@@ -1,0 +1,9 @@
+[scheduling]
+    initial cycle point = 2020
+    stop after cycle point = 2020-01-01T00:01Z
+    [[graph]]
+        PT1M = foo
+[runtime]
+    [[foo]]
+        [[[simulation]]]
+            default run length = PT0S
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -267,7 +267,7 @@ def test_family_inheritance_and_quotes(
 
 
 @pytest.mark.parametrize(
-    ('cycling_type', 'scheduling_cfg', 'expected_icp', 'expected_opt_icp',
+    ('cycling_type', 'scheduling_cfg', 'expected_icp', 'expected_eval_icp',
      'expected_err'),
     [
         pytest.param(
@@ -356,7 +356,7 @@ def test_process_icp(
     cycling_type: str,
     scheduling_cfg: Dict[str, Any],
     expected_icp: Optional[str],
-    expected_opt_icp: Optional[str],
+    expected_eval_icp: Optional[str],
     expected_err: Optional[Tuple[Type[Exception], str]],
     monkeypatch: pytest.MonkeyPatch, set_cycling_type: Fixture
 ) -> None:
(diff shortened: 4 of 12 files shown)


cylc/flow/scripts/validate_install_play.py
```diff
@@ -101,7 +102,11 @@ def main(parser: COP, options: 'Values', workflow_id: Optional[str] = None):
     orig_source = workflow_id
     source = get_source_location(workflow_id)
     log_subcommand('validate', source)
-    asyncio.run(cylc_validate(parser, options, str(source)))
+
+    # Copy options in case validate decides to store information
+    # in the options object:
+    val_opts = copy(options)
+    asyncio.run(cylc_validate(parser, val_opts, str(source)))
```
oliver-sanders, line 109: I was kinda relying on the old behaviour for the rose-stem plugin work (because we need to avoid re-running the plugin for the combined commands).

Did you get to the bottom of why `cylc validate` is modifying the options in the first place?

wxtim: No, but I imagine it's just a place to store data - since you need it for rose stem I'll investigate.

wxtim, line 109:
```suggestion

    asyncio.run(cylc_validate(parser, options, str(source)))
```

cylc/flow/scripts/validate_reinstall.py
```diff
@@ -171,6 +171,7 @@ async def vr_cli(parser: COP, options: 'Values', workflow_id: str):
 
     # Run cylc validate
     log_subcommand('validate --against-source', workflow_id)
+
```
wxtim, line 174:
```suggestion
```

cylc/flow/scripts/validate_install_play.py
```diff
@@ -29,6 +29,7 @@
 """
 
 import asyncio
+from copy import copy
```
wxtim, line 32:
```suggestion
```

cylc/flow/scripts/validate_install_play.py
```diff
@@ -101,6 +102,7 @@ def main(parser: COP, options: 'Values', workflow_id: Optional[str] = None):
     orig_source = workflow_id
     source = get_source_location(workflow_id)
     log_subcommand('validate', source)
+
```
wxtim, line 105:
```suggestion
```

cylc/flow/config.py
```diff
@@ -710,10 +710,10 @@ def process_initial_cycle_point(self) -> None:
...
                 except IsodatetimeError as exc:
                     raise WorkflowConfigError(str(exc))
-        if orig_icp != icp:
-            # now/next()/previous() was used, need to store
-            # evaluated point in DB
-            self.options.icp = icp
+        # Save un-evaluated point for DB
+        self.raw_icp = None
+        if icp != orig_icp:
+            self.raw_icp = icp
```
MetRonnie, line 716: I feel like the previous comment explained it better?
```suggestion
        self.raw_icp = None
        if icp != orig_icp:
            # now/next()/previous() was used, need to store
            # evaluated point in DB
            self.raw_icp = icp
```

cylc/flow/config.py
```diff
@@ -690,7 +690,7 @@ def process_initial_cycle_point(self) -> None:
         Sets:
             self.initial_point
             self.cfg['scheduling']['initial cycle point']
-            self.options.icp
+            self.raw_icp
```
MetRonnie, line 693: I don't find "raw" to be particularly illuminating. Can we avoid using a separate attribute and just use `self.initial_point` for this?

tests/functional/cylc-combination-scripts/09-vr-icp-now/flow.cylc
```diff
@@ -0,0 +1,7 @@
+[scheduling]
+    initial cycle point = now
+    [[graph]]
+        PT1M = foo
+[runtime]
+    [[foo]]
+        script = cylc stop "${CYLC_WORKFLOW_ID}"
```
MetRonnie, line 7: Could just run this in simulation mode with `default run length` of 0?

oliver-sanders:
> Partially closes https://github.com/cylc/cylc-flow/issues/6262 (we should check for other places where the options object might be changed unwisely.)

Please remove the "partially" or create a follow-on issue to cover this.

changes.d/6316.fix.md
```diff
@@ -0,0 +1 @@
+Prevent validate step of VR changing CLI options.
```
MetRonnie, line 1:
```suggestion
Fixed bug in `cylc vr` where an initial cycle point of `now`/`next()`/`previous()` would result in an error.
```

tests/unit/test_config.py
```diff
@@ -396,7 +396,7 @@ def test_process_icp(
         assert mocked_config.cfg[
             'scheduling']['initial cycle point'] == expected_icp
         assert str(mocked_config.initial_point) == expected_icp
-        opt_icp = mocked_config.options.icp
+        opt_icp = mocked_config.evaluated_icp
         if opt_icp is not None:
             opt_icp = str(loader.get_point(opt_icp).standardise())
         assert opt_icp == expected_opt_icp
```
MetRonnie, line 402: Can you replace `opt` with `evaluated` in references to `opt_icp` and `expected_opt_icp`, otherwise this no longer makes sense?

MetRonnie:
MacOS failures addressed by https://github.com/wxtim/cylc/pull/63

MetRonnie:
> MacOS failures caused by different functional tests being introduced into the chunk due to the added functional test. Addressed by [wxtim#63](https://github.com/wxtim/cylc/pull/63)

Still awaiting

changes.d/6316.fix.md
MetRonnie: Unresolved

MetRonnie (commented):
Got a functional test failure
```
tests/f/restart/10-pre-initial-2.t
```

MetRonnie (approved):

PR #6321, opened 2024-08-22
Prevent a couple of unit tests from installing into `~/cylc-run`

MetRonnie:


tests/unit/test_install.py
```diff
@@ -560,7 +558,7 @@ def test_install_workflow_failif_name_name(tmp_src_dir):
         install_workflow(src_dir, workflow_name='foo', run_name='bar?')
 
 
-def test_install_workflow_failif_reserved_name(tmp_src_dir):
+def test_install_workflow_failif_reserved_name(tmp_src_dir, tmp_run_dir):
```
MetRonnie, line 561: This test would fail the second time you ran it 

> WorkflowFilesError: '~/cylc-run/ok/also_ok' already exists

hjoliver (approved):

wxtim (approved):

PR #6322, opened 2024-08-22
param expand: efficiency improvements

oliver-sanders:
Whils**t** investigating a bug, I found the workflow I was investigating loaded extremely slowly (~85-87s).

Over 90% of that time was down to the `_expand_graph` routine. We may be able to implement this more efficiently, but without getting too into it, there were three obvious improvements which bring it down to ~61s (~30% saving) without changing how the method works.

**Check List**


hjoliver:
> `Whils**t**`

🤣  respect ✊

hjoliver (approved):
Good, easy win!

Issue #6314, opened 2024-08-21
polling/set: submitted task cannot be set to succeeded


oliver-sanders:
Spotted in the wild:

```
INFO - + 20210103T0000Z/nemo_cice_obsoper_EN000 submitted
INFO - [20210103T0000Z/nemo_cice_obsoper_EN000/01:submitted(runahead)] => submitted
INFO - + 20210103T0000Z/nemo_cice_obsoper_EN000 poll_timer
INFO - + 20210103T0000Z/nemo_cice_obsoper_EN000 ['try_timers', 'submission-retry']
INFO - + 20210103T0000Z/nemo_cice_obsoper_EN000 ['try_timers', 'execution-retry']
INFO - [20210103T0000Z/nemo_cice_obsoper_EN000/01:submitted] (polled)succeeded
INFO - [20210103T0000Z/nemo_cice_obsoper_EN000/01:submitted] setting implied output: submitted
INFO - [20210103T0000Z/nemo_cice_obsoper_EN000/01:submitted] submitted to xce:None[None]
WARNING - Unhandled jobs-poll output: 2024-08-03T12:18:26Z|20210103T0000Z/nemo_cice_obsoper_EN000/01|{"job_runner_name": "pbs", "job_id": "3094491", "run_status": 0, "time_submit_exit": "2024-08-03T11:17:32Z", "time_run": "2024-08-03T11:39:00Z", "time_run_exit": "2024-08-03T11:45:30Z"}
    set(flow=['all'], flow_wait=False, outputs=[], prerequisites=[], tasks=['20210103T0000Z/nemo_cice_obsoper_EN000'])
INFO - output 20210103T0000Z/nemo_cice_obsoper_EN000:succeeded completed already
    kill_tasks(tasks=['20210103T0000Z/nemo_cice_obsoper_EN000'])
INFO - [20210103T0000Z/nemo_cice_obsoper_EN000/01:submitted] => submitted(held)
ERROR - [jobs-kill cmd] ssh -oBatchMode=yes -oConnectTimeout=8 -oStrictHostKeyChecking=no xcel01 env CYLC_VERSION=8.3.3 CYLC_ENV_NAME=cylc-8.3.3-1 bash --login -c ''"'"'exec "$0" "$@"'"'"'' cylc jobs-kill -- '$HOME/cylc-run/u-di486/run1/log/job' 20210103T0000Z/nemo_cice_obsoper_EN000/01
[...]

PR #6326, opened 2024-08-23
task_events_mgr: handle missing job config

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/6314
* There are niche situations where the job is not stored in "TaskProxy.jobs".
* This handles the situation as gracefully as we are able to.

The solution is far from ideal, the job config info has been lost for starters, however, I think that is non-functional at this stage?

@hjoliver, any better ideas?

**Check List**

### DIFF
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -1538,7 +1538,19 @@ class TaskEventsManager():
         if (itask.tdef.run_mode == RunMode.SIMULATION) or forced:
             job_conf = {"submit_num": itask.submit_num}
         else:
-            job_conf = itask.jobs[-1]
+            try:
+                job_conf = itask.jobs[-1]
+            except IndexError:
+                # we do not have access to the job config (e.g. Scheduler
+                # crashed) - https://github.com/cylc/cylc-flow/pull/6326
+                job_id = itask.tokens.duplicate(
+                    job=itask.submit_num
+                ).relative_id
+                LOG.warning(
+                    f'Could not find the job configuration for "{job_id}".'
+                )
+                itask.jobs.append({"submit_num": itask.submit_num})
+                job_conf = itask.jobs[-1]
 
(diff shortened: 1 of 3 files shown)


cylc/flow/task_events_mgr.py
```diff
@@ -1533,7 +1533,18 @@ def _insert_task_job(
         if (itask.tdef.run_mode == RunMode.SIMULATION) or forced:
             job_conf = {"submit_num": itask.submit_num}
         else:
-            job_conf = itask.jobs[-1]
+            try:
+                job_conf = itask.jobs[-1]
+            except IndexError:
+                # we do not have access to the job config
```
hjoliver, line 1539: Can we add, to this comment, your your theory on why this might be needed? 

```suggestion
                # we do not have access to the job config
                # GitHub #6314: itask.jobs was empty on restart, perhaps because it
                # did not get written to the DB due to an emergency shutdown.
```

hjoliver (approved):
I think your fix is good.

I wonder if, as a follow-up, we should put some checks in place to ensure that all expected critical config exists in `n=0` task loaded for a restart? (And remove this change once that's done.)

hjoliver:
> Ideally, we would probably have Job objects to match Task(Proxy) objects and a strict mapping between them which would allow us to move some job-specific config (e.g. platform, run_mode, etc) out of TaskProxy (where is doesn't belong as job submissions of the same task may have different values) into the Job config. Perfectly possible, however, this would require a reasonably large refactor which isn't something we have time for at present.

Sensible, but I agree it might not be a small job - post an issue?

oliver-sanders:
Definitely a "non-small" job!

> I wonder if, as a follow-up, we should put some checks in place to ensure that all expected critical config exists in n=0 task loaded for a restart?

I think what's happened here is that `itask.jobs` *was* non-critical, once, but has started to be used because the information is hard to get from elsewhere.

wxtim (approved):

PR #6327, opened 2024-08-23
Correctly reset log level when temporarily changing it

MetRonnie:
Fixes tests that were failing on my `cylc remove` branch due to a change in order

#### Context

Running these 2 tests in this order would guarantee `test_Timer` would fail.
```
pytest -n 0 \
  tests/integration/scripts/test_show.py::test_workflow_meta_query \
  tests/unit/test_timer.py::test_Timer
```

This is due to the way we were temporarily changing the Cylc logger level during `Scheduler.start()`

https://github.com/cylc/cylc-flow/blob/cadf1ef057145873f869092ffcab9784ad7c082e/cylc/flow/loggingutil.py#L426-L437

`Logger.getEffectiveLevel()` will return the root logger level if `Logger.level` is `NOTSET`.

In pytest the root logger level is `WARNING`, so in the tests, there were cases where the Cylc logger was being incorrectly reset to `WARNING` instead of `NOTSET`

#### Check List

- [x] No dependency changes 
- [x] Tests are included
- [x] No changelog entry needed as not affecting users
- [x] No docs needed

### DIFF
--- a/tests/unit/test_loggingutil.py
+++ b/tests/unit/test_loggingutil.py
@@ -34,6 +34,7 @@ from cylc.flow.loggingutil import (
     RotatingLogFileHandler,
     get_reload_start_number,
     get_sorted_logs_by_time,
+    patch_log_level,
     set_timestamps,
 )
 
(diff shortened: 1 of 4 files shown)


oliver-sanders (approved):

tests/unit/test_loggingutil.py
```diff
@@ -245,3 +246,32 @@ def test_log_emit_and_glbl_cfg(
...
+    logger.info("nope")
+    assert not caplog.records
+    with patch_log_level(logger, logging.INFO):
+        LOG.info("yep")
+        assert len(caplog.records) == 1
+    logger.info("nope")
+    assert len(caplog.records) == 1
+
+
+def test_patch_log_level__reset(caplog: pytest.LogCaptureFixture):
```
wxtim, line 266: Why do these need to be separate tests?

wxtim (approved):
Looks good - one question about whether both tests are necessary should not stand in the way of merger.

tests/unit/test_loggingutil.py
MetRonnie: I'm not sure they could be (and even if they could, it would probably be very confusing). The above test is a test `patch_log_level()` does what it says. This test is a test that it is not affected by the root logger level. I'll add a bit more info to the docstring of this one

oliver-sanders:
merger-ing

Issue #6320, opened 2024-08-22
platforms: broadcasted platform ignored after ssh failure


oliver-sanders:
We can use broadcasts to change the platform a task submits to.

Under normal circumstances this works fine, however, when hosts go down and the submission is retried, the broadcast seems to be forgotten about and the new submission uses the configured platform.

This could lead to jobs being submitted to the wrong platform.

### Reproducible example:

Run the following workflow.

Once the "remote_init_one" and "remote_init_two" tasks have submitted, break your SSH config to force subsequent calls to fail.

```ini
[scheduling]
    [[graph]]
        R1 = remote_init_one & remote_init_two & local => remote

[runtime]
    # ensure that the workflow has remote-init'ed on platforms "one" and "two"
    [[remote_init_one]]
        platform = one-bg
    [[remote_init_two]]
        platform = two-bg

    # change the platform of "remote" via broadcast
    [[local]]
        script = """
            cylc broadcast "${CYLC_WORKFLOW_ID}" -n remote -p "${CYLC_TASK_CYCLE_POINT}" -s 'platform=one'
            sleep 10
        """

    [[remote]]
        platform = localhost
```

The "remote" task should attempt to submit to each of the hosts in the "one" platform. All SSH connections will fail so the task will run out of hosts and become submit-failed.

However, that's not what happens! Running this command reveals that after running out of hosts, the task then attempted to submit to localhost (the platform defined *before* the broadcast):

```
$ grep 'DEBUG - \[jobs-submit cmd\].*1/remote/01' --color=never ~/cylc-run/<workflow>/log/scheduler/log
... ssh ... one.01 ... cylc jobs-submit ... 1/remote/01
... ssh ... one.02 ... cylc jobs-submit ... 1/remote/01
... cylc jobs-submit ... 1/remote/01
```

Note: This erroneous submission appears to happen *after* all the hosts of the broadcasted platform have been exhausted which may help pin down the offending code pathway.

Interestingly, when I try this, the attempted submission to `localhost` actually fails due to the `qsub` command not being in `$PATH`. In my case platform `one` uses PBS so this suggests that it is attempting to submit to `localhost`, but with the configuration of `one`?!

wxtim:
Looks like it's remote initing on the same host?

```
    [[remote_init_one]]
        platform = one-bg
    [[remote_init_two]]
        platform = one-bg
```

oliver-sanders:
Typo, corrected in OP

wxtim:
[...]

PR #6330, opened 2024-08-27
Ensure that platform from group selection checks broadcast manager

wxtim:
Possible fix for #6320 

**Check List**


cylc/flow/subprocpool.py
```diff
@@ -557,12 +557,12 @@ def _run_callback(callback, args_=None):
                 # Backup, get a platform name from the config:
                 for arg in callback_args:
                     if isinstance(arg, TaskProxy):
-                        platform_name = arg.tdef.rtconfig['platform']
+                        platform_name = arg.platform['name']
```
wxtim, line 560: Separate issue, but a confusing bit of logging output: 

```
WARNING - platform: localhost - Could not connect to xcel00.
    * xcel00 has been added to the list of unreachable hosts
    * jobs-submit will retry if another host is available.
```

Emitted at line 574 as a result of this.

oliver-sanders:
(unsubscribing for now)

cylc/flow/subprocpool.py
wxtim: This turned out to be a 🔴 🐟 : The bad path in the logging sorted itself out as a  result of the other change. 😄

cylc/flow/task_job_mgr.py
```diff
@@ -306,10 +306,12 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
 
                     # Get another platform, if task config platform is a group
                     use_next_platform_in_group = False
-                    if itask.tdef.rtconfig['platform']:
+                    bc_mgr = self.task_events_mgr.broadcast_mgr
+                    rtconf = bc_mgr.get_updated_rtconfig(itask)
+                    if rtconf['platform']:
```
oliver-sanders, line 311: Looking at this line, I've found another similar bug made possible by this fix.

The only case where`if rtconfig['platform']` evaluates as `False` is if the platform is unset. If it does evaluate `False`, the job will submit with `itask.platform` which is what's defined from the config. With the new broadcast awareness, it is possible to get a new bug out of this when using a broadcast to unset the platform.

Here's an example:

```ini
[runtime]
  [[one]]
    platform = my-platform
```

Then start the workflow and unset the platform:

```console
$ cylc vip --pause
$ cylc broadcast -n one -s 'platform='
$ cylc play
```

The job will attempt to submit to localhost, but with the config of `my-platform`? It fails anyway.

Here's a diff that will reproduce the "out of hosts" scenario without having to mess with SSH config:

```diff
diff --git a/cylc/flow/task_job_mgr.py b/cylc/flow/task_job_mgr.py
index 9eec82970..c0da0e6c0 100644
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -292,13 +292,15 @@ class TaskJobManager:
             for itask in itasks:
                 # If there are any hosts left for this platform which we
                 # have not previously failed to contact with a 255 error.
-                if any(
-                    host not in self.task_remote_mgr.bad_hosts
-                    for host in itask.platform['hosts']
-                ):
-                    platform = itask.platform
-                    out_of_hosts = False
-                    break
+                if False:
+                    pass
+                # if any(
+                #     host not in self.task_remote_mgr.bad_hosts
+                #     for host in itask.platform['hosts']
+                # ):
+                #     platform = itask.platform
+                #     out_of_hosts = False
+                #     break
                 else:
                     # If there are no hosts left for this platform.
                     # See if you can get another platform from the group or
@@ -308,6 +310,8 @@ class TaskJobManager:
                     use_next_platform_in_group = False
                     bc_mgr = self.task_events_mgr.broadcast_mgr
                     rtconf = bc_mgr.get_updated_rtconfig(itask)
+                    LOG.critical(f"#1 {rtconf['platform']}")
+                    LOG.critical(f"#2 {itask.platform['name']}")
                     if rtconf['platform']:
                         try:
                             platform = get_platform(
```

```
INFO - [1/one:waiting] => preparing
CRITICAL - #1 
CRITICAL - #2 localhost
ERROR - [jobs-submit cmd] (remote init)
    [jobs-submit ret_code] 1
CRITICAL - [1/one/01:preparing] submission failed
```

oliver-sanders (changes requested):
LGTM, but there's an edge case that needs handling.

cylc/flow/task_job_mgr.py
wxtim: I can't see any benefit, other than possibly a minor speed-up from breaking the loop iteration asap. Testing.

wxtim: What it's achieving is not trying to select another platform from a platform group if there are any hosts left on this one.

oliver-sanders: I don't think `if rtconf['platform']:` will achieve that?

wxtim: Red Herring (I was talking about a different if statement).

However, the if statement is required for the case where platform is set by remote/job settings as per #4570 . I think that I've fixed it by passing the whole rtconfig to the get platform routine, which should handle both scenarios.

oliver-sanders: Oh heck, that's a good point. Presumably this logic would also have failed if you attempted to broadcast Cylc-7 style host/batch-system settings too?!

wxtim: Very likely.

cylc/flow/task_job_mgr.py
```diff
@@ -308,18 +308,17 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
...
-                            pass
-                        else:
-                            # If were able to select a new platform;
-                            if platform and platform != itask.platform:
-                                use_next_platform_in_group = True
+                    try:
+                        platform = get_platform(
+                            rtconf,
+                            bad_hosts=self.bad_hosts
+                        )
```
oliver-sanders, line 315: Would be a good idea to check what other interfaces are using `get_platform(platform_name, ...)` as opposed to `get_platform(platform_config, ...)` to make sure this issue isn't present elsewhere.

wxtim: Searched `get_platform\([^\)]` and found the following cases:

- [x] `subprocpool.SubProcPool._run_command_exit` - Uses a platform name from a realized platform (i.e. `platform['name']` rather than from the config.
- [x] `task_events_mgr.TaskEventsManager._process_job_log_retrieval` - As above.
- [x] `task_job_mgr.TaskJobManager._run_job_cmd` - As above
- [x] `task_pool.TaskPool.load_db_task_pool_for_restart` - Loading a platform name from the database - which is set from `platform['name']`. 
- [x] `cat_log.main`  - as above
- [ ] `cylc check-versions` - is broken for compat mode tasks - shall I fix it here or create a new issue/pr - do people use it much?

cylc/flow/task_job_mgr.py
oliver-sanders: Resolved - happy with the fix

cylc/flow/task_job_mgr.py
oliver-sanders: Great, thanks for checking.

oliver-sanders (approved):
I have tested this with the example in the OP as well as with the Cylc 7 - style variant (https://github.com/cylc/cylc-flow/issues/6320#issuecomment-2326719188) to ensure that it works fine with old-style host selection.

All good :+1:.

PR #6332, opened 2024-08-28
Added f to format string

### LINKED ISSUE #6331: GraphParseError does not print name of non-family namespace that caused error
### Description

In version 8.3.3 of cylc, `GraphParseError` does not include the name of the namespace that produced the error
### Reproducible Example

Run a command like `cylc graph .` on the following workflow:
```
[scheduling]
    [[graph]]
            R1 = "prep => not-fam:succeed-all => final"
[runtime]
    [[prep]]
    [[not-fam]]
    [[final]]

```
### Expected Behaviour

The above workflow will give you an error that says 
```
GraphParseError: family trigger on non-family namespace {expr}
```
When I believe that it should say
```
GraphParseError: family trigger on non-family namespace not-fam
```
I think this error is just caused by a missing `f` in the following format string. Although it is a small issue, I thought I'd let the development team know just in case the issue occurs in other places as well.
https://github.com/cylc/cylc-flow/blob/cadf1ef057145873f869092ffcab9784ad7c082e/cylc/flow/graph_parser.py#L644-L648

uwagura:
Closes #6331 . Test not needed because I literally just added a character
**Check List**

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,4 +1,4 @@
-# FORMAT: <proper-name>  <proper-email>  <commit-name>  <commit-email>   
+# FORMAT: <proper-name>  <proper-email>  <commit-name>  <commit-email>
 # Omit commit-name or commit-email if same as proper
 
 Alex Reinecke              <alex.reinecke@nrlmry.navy.mil>      P. A. Reinecke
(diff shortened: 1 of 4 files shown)


oliver-sanders:
Hi @uwagura and welcome to the Cylc community :tada:

Thanks for your bug report and many thanks for taking the time to fix the issue.

For legal reasons, we need all commit authors to enter their names into the [CONTRIBURINTING.md](https://github.com/cylc/cylc-flow/blob/master/CONTRIBUTING.md) file (no matter how small the fix) in order to agree to the [CLA](https://github.com/cylc/cylc-flow/blob/master/CONTRIBUTING.md#contributor-licence-agreement-and-certificate-of-origin).

Make sure you write your name in the same format it appears in your commit (run `git shortlog -se` and copy your name as it appears there, or add an entry into the `.mailmap` file if you want it to appear differently).

uwagura:
@oliver-sanders , feels a bit strange to call myself a contributor for such a small change, but I've gone ahead and added myself to the contributors list and mailmap. I also added an entry to the changelog using towncrier, though I'm not entirely sure if I did it correctly. Let me know how how everything looks

oliver-sanders:
It's a legal thing we have to do even with the smallest commits, we only need to do it once.

We greatly appreciate bugfixes, thanks for your contribution.

PR #6335, opened 2024-08-28
broadcast: fix issue with multiple namespaces

### LINKED ISSUE #6334: Broadcast: GraphQLLocatedError 'DurationFloat' object has no attribute 'startswith'
### Description

When broadcasting to set the `execution time limit` I'm seeing the error `ERROR: [{'error': {'message': "'DurationFloat' object has no attribute 'startswith'", 'traceback': ["graphql.error.located_error.GraphQLLocatedError: 'DurationFloat' object has no attribute 'startswith'\n"]}}]`. This may be related to #6333 and #6222 
This happens if you have two or more namespaces and are using a ISO duration item.

### Reproducible Example

```
$ cylc broadcast -s 'execution time limit=PT1M' -n root -n root access_g4_rewind
ERROR: [{'error': {'message': "'DurationFloat' object has no attribute 'startswith'", 'traceback': ["graphql.error.located_error.GraphQLLocatedError: 'DurationFloat' object has no attribute 'startswith'\n"]}}]
```

The namespaces can be the same, it doesn't matter. If you have one namespace, there are no issues, 2+ it fails.

### Expected Behaviour

No failure

oliver-sanders:
* Fix an issue that could cause issues when broadcasting "coerced" configurations to multiple namespaces.
* Specifying the same namesapce multiple times doesn't make sense, we should strip duplicates earlier on in the process.
* Closes https://github.com/cylc/cylc-flow/issues/6334

**Check List**

- [x] Changelog entry -  too minor to warrant an entry

### DIFF
--- a/cylc/flow/broadcast_mgr.py
+++ b/cylc/flow/broadcast_mgr.py
@@ -280,8 +280,16 @@ class BroadcastMgr:
         bad_namespaces = []
 
         with self.lock:
-            for setting in settings:
-                for point_string in point_strings:
+            for setting in settings or []:
+                # Coerce setting to cylc runtime object,
+                # i.e. str to  DurationFloat.
+                coerced_setting = deepcopy(setting)
+                BroadcastConfigValidator().validate(
+                    coerced_setting,
+                    SPEC['runtime']['__MANY__'],
+                )
+
+                for point_string in point_strings or []:
                     # Standardise the point and check its validity.
                     bad_point = False
                     try:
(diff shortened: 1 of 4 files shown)


cylc/flow/broadcast_mgr.py
```diff
@@ -280,8 +280,16 @@ def put_broadcast(
         bad_namespaces = []
 
         with self.lock:
-            for setting in settings:
-                for point_string in point_strings:
+            for setting in settings or []:
```
wxtim, line 283: This is 3 levels nested and in theory could be looping a fair bit. Might it be a nicer read and possibly more efficient with:

```
from itertools import product

for setting, point_string, namespace in product(settings or [], point_strings or [], namespaces or []):
    # All below, but with less nesting, and letting C layer do the heavy iterative lifting if necessary.

```

tests/integration/scripts/test_broadcast.py
```diff
@@ -77,3 +78,59 @@ async def test_broadcast_multi(
...
+                'a': {'inherit': 'VOWELS'},
+                'b': {'inherit': 'CONSONANTS'},
+                'c': {'inherit': 'CONSONANTS'},
+            },
+        }
+    )
+    schd = scheduler(id_)
+
+    async with start(schd):
+        # issue a broadcast to multiple namespaces
```
wxtim, line 111: I managed to replicate the error doing this:

```bash
$ cylc broadcast throwaway/wide.dirt -p '*' -n baz -n bar -n root -s 'script=sleep 30'  -s '[simulation]default run length = PT0S'
ERROR: [{'error': {'message': "'DurationFloat' object has no attribute 'startswith'", 'traceback': ["graphql.error.located_error.GraphQLLocatedError: 'DurationFloat' object has no attribute 'startswith'\n"]}}]
```

Perhaps you should test that too?

wxtim (approved):
Reviewed. Possible improvement suggested, but not one which should block merger.

oliver-sanders:
Additional functional review performed here: https://github.com/cylc/cylc-flow/issues/6334#issuecomment-2319756229

cylc/flow/broadcast_mgr.py
oliver-sanders: I don't *think* product is more efficient (it will still use for loops under the hood, it cannot change the number of iterations), but flattening this logic might be nicer.

I would rather avoid refactors in bugfix PRs through for risk management purposes.

oliver-sanders:
Merging with 1 1/2 reviews.

PR #6345, opened 2024-09-02
First hack at a dupl job submission fix.

hjoliver:
**Check List**


oliver-sanders:
> tasks in the preparing state at reload time will submit multiple times.

Yikes!

> It's something to do with the fact that we now wait for preparing tasks to submit before actioning the requested reload (which I think is new-ish)

It is new-ish, but I would have thought that this change would have made reload safer in this regard because there can't be any preparing tasks at the time of the reload?

> I haven't grokked the root cause well enough to see if we can prevent the queue attempt in the first place.

Could definitely do with pinning this down as it's likely an interaction of an internal list or task state with some other part of the system which could potentially spring a leak under other circumstances too?

oliver-sanders:
> but whilst waiting for the preparing tasks to submit we repeatedly add the same job-submit commands to the process pool command queue.

Oh heck, **whilst**!

The ``cylc.flow.commands.reload_workflow` routine contains a mini main-loop containing the small subset of functionality required to flush preparing tasks through contained with the `while schd.release_queued_tasks():` loop. I wonder if there's some other bit of logic required, but not contained within this loop?

hjoliver:
Yes!! I was home in on that flag myself.  I think that's got it.

oliver-sanders:
> I guess the question is, do we need to waiting for preparing tasks to clear

Yes, this shifted other bugs.

hjoliver:
> > I guess the question is, do we need to waiting for preparing tasks to clear
> 
> Yes, this shifted other bugs.

Yes, but I'm asking if those bugs were solely due to `preparing+waiting_on_job_submit`, or `preparing` more generally.

oliver-sanders:
In the case of the auto-restart functionality, it's preparing more generally.

The auto-restart functionality will restart the workflow on another host. Because of this, we must wait for all localhost task submissions to complete first because the localhost platform will be different on the new host.

hjoliver:
So @oliver-sanders - just to make sure we're on the same page here, and so I can hopefully solve this problem tomorrow:

Your suggestion above to use `itask.waiting_on_job_prep` flag, i.e.:

 > We use the `waiting_on_job_prep` flag rather than the `preparing` status in the task_pool for the release-from-queue side of this. 
 
 seems to be at odds with what you've just said (we have to wait for ALL `preparing` tasks to clear, not just those waiting on job prep):
 
 > In the case of the auto-restart functionality, it's `preparing` more generally.
 
Assuming the latter overrides the former, I take it you now think we need to keep the original code in the scheduler module, and come up with a different solution to prevent the duplicate submissions?

oliver-sanders:
Sorry. I wasn't proposing it as a fix (else I would have opened a PR). But pointing out that it seemed to fix the example in the issue.

hjoliver:
OK got it.

I resorted to a functional test as I wasn't sure how to do better in the time available today.

hjoliver:
I can't figure out the one seemingly repeatably functional test failure on this PR: `tests/f/triggering/08-family-finish-any`.

Pretty sure it's unrelated. It only fails in the macos CI run. It passes in the Linux run, and it passes locally on macos for me.

```
    triggering is NOT consistent with the reference log:
    --- reference
    +++ this run
    -1/foo -triggered off ['1/b']
    +1/foo -triggered off ['1/a', '1/b', '1/c']
```
The test workflow graph is:
```cylc
   R1 = """FAM:finish-any => foo"""  # FAM is a, b, c
```
Task `b` has `script = True`, and `a, b` have `script = sleep 10`, but in the test run `b` takes exactly 10 seconds to run just like the other two tasks, so `foo` triggers off of them all at once, hence the result. I downloaded the test artifact, and `b`'s job logs confirm the 10 sec run time and that the scripting is just `true`.  🤯

cylc/flow/task_job_mgr.py
```diff
@@ -536,6 +536,12 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
                 stdin_files = []
                 job_log_dirs = []
                 for itask in itasks_batch:
+                    if not itask.waiting_on_job_prep:
+                        # Avoid duplicate job submissions when flushing
+                        # preparing tasks before a reload. See
+                        # https://github.com/cylc/cylc-flow/pull/6345
+                        continue
```
dwsutherland, line 543: Commenting out this the test will fail (as expected).
```Found 12 (not 1) of 1/foo.*submitted to in /home/sutherlander/cylc-run/cylctb-20240905T051008Z-1U8S/functional/reload/28-preparing/log/scheduler/log```

File: cylc/flow/task_job_mgr.py [hunk 2]
```diff
@@ -554,8 +560,11 @@ def submit_task_jobs(self, workflow, itasks, curve_auth,
                     # write flag so that subsequent manual retrigger will
                     # generate a new job file.
                     itask.local_job_file_path = None
-
                     itask.waiting_on_job_prep = False
+
+                if not job_log_dirs:
+                    continue
```
dwsutherland, line 566: Commenting out this the test passes but:
image

Might be caught with other tests though..

dwsutherland (approved):
Works. 
We would have run into this at NIWA sooner or later 👍

oliver-sanders (approved):

Issue #6284, opened 2024-08-06
expire: expiry is ignored if retries are configured


oliver-sanders:
An expired task will automatically retry if `execution retry delays` are configured.

Example:

```
[scheduling]
    initial cycle point = 2000
    [[special tasks]]
        clock-expire = get_data(+PT10M)
    [[graph]]
        P1Y = """
            get_data => foo
        """

[runtime]
    [[get_data]]
        execution retry delays = PT5S
    [[foo]]
```

The task will expire, then reset back to waiting and retry:

```console
$ grep '20000101T0000Z\/get_data' tmp.bplFgeXhsd/run1/log/scheduler/log 
2024-08-06T15:24:38+01:00 INFO - [20000101T0000Z/get_data:waiting(runahead)] => waiting
2024-08-06T15:24:38+01:00 INFO - [20000101T0000Z/get_data:waiting] => waiting(queued)
2024-08-06T15:24:38+01:00 WARNING - [20000101T0000Z/get_data:waiting(queued)] (internal)expired
2024-08-06T15:24:38+01:00 INFO - [20000101T0000Z/get_data:waiting(queued)] => expired
2024-08-06T15:24:38+01:00 WARNING - [20000101T0000Z/get_data:expired] did not complete the required outputs:
2024-08-06T15:24:52+01:00 INFO - [20000101T0000Z/get_data:expired] => preparing
2024-08-06T15:24:54+01:00 INFO - [20000101T0000Z/get_data/01:preparing] submitted to localhost:background[114986]
2024-08-06T15:24:54+01:00 INFO - [20000101T0000Z/get_data/01:preparing] => submitted
2024-08-06T15:24:59+01:00 INFO - [20000101T0000Z/get_data/01:submitted] => running
2024-08-06T15:25:03+01:00 INFO - [20000101T0000Z/get_data/01:running] => succeeded
```

Setting a completion condition that tolerates expiry does not solve the issue:

```
[runtime]
    [[get_data]]
        completion = succeeded or expired
```

The workaround is to remove the task using an `expired handler`.

```
[runtime]
    [[get_data]]
        [[[events]]]
            expired handlers = cylc remove %(workflow)s//%(id)s
```

Issue #6261, opened 2024-07-25
`cylc vr` - restart with no changes?


hjoliver:
> Validate, reinstall and apply changes to a workflow.
> 
> Validate and reinstall a workflow then either:
>
> * "Reload" the workflow (if it is running),
> * or "play" it (if it is stopped).

Currently, if the workflow is running and no source changes are detected, `cylc vr` aborts with:
```
WARNING - No changes to source: No reinstall or reload required.
```
That's fine, because a reload would achieve absolutely nothing.

However if the workflow is stopped and no source changes are detected, we get:
```
WARNING - No changes to source: No reinstall or play required.
```

IMO the "no play required" bit is wrong, it should restart the workflow. I wouldn't do `cylc vr` if I didn't want the workflow to be running. Maybe I don't recall if if I installed my latest changes already, so I do `cylc vr` instead of `cylc play` just in case.

hjoliver:
Not a big deal since the "workaround" is simply to `cylc play` if this happens. But still, I think the command logic should be this:
 1. validate source
 2. reinstall changes, if any
 3. reload if running, or restart if stopped

oliver-sanders:
> I think the command logic should be this

It is.

oliver-sanders:
The purpose of rejecting a reinstallation with no changes is that this likely reflects user error or at least user expectation error (e.g. I forgot to save the file I was editing). So I think this functionality is useful and should remain (we have the `--yes` option to bypass it).

However, we could consider changing the WARNING into a prompt e.g:

```
Workflow installation is already up to date (nothing to reinstall).

Would you like to restart the workflow with no changes (y/n)?
```

(note, for the reload case, the command may as well exit at this point, any reload would most likely be pointless)

Or we could adapt the WARNING to taste, e.g:

```
Workflow installation is already up to date (nothing to reinstall). Use the "--yes" option if you want to reinstall anyway.
```

hjoliver:
> It is, however, it exits at (2) if there are no changes rather than continuing to (3).

Yes, that's the bit I'm complaining about!

hjoliver:
> Note that you can override this with the --yes option.

OK I just tested that and it works.  I was not aware of it, because it doesn't make sense - there's no interactive prompt to say "yes" to!

```
--yes                 [reinstall] Skip interactive prompts.
```

It would make sense with a new prompt as you suggested above.

hjoliver:
[...]

PR #6354, opened 2024-09-03
vr: restart the workflow with no changes according to user prompt

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/6284
* If there are no changes to reinstall AND the workflow is stopped, rather than just exiting, the `cylc vr` command will now prompt the user to give them the option of restarting the workflow.
* This makes them aware there are no changes to reinstall (which may be an error on their part) but still gives them the option of restart without having to edit or rerun the command.

An odd feature to introduce at a bugfix release, however, it's kinda a UX type issue so it might make sense.

**Check List**


hjoliver:
(Looks good, but I haven't done a proper review yet).

oliver-sanders:
(rebased)

oliver-sanders:
(rebased)

changes.d/6354.feat.md
```diff
@@ -0,0 +1 @@
+Fix an issue that could cause clock-expired tasks to be erroneously retried if "execution retry delays" were configured.
```
wxtim, line 1: This doesn't match the issue description in the PR?

tests/integration/scripts/test_validate_reinstall.py
```diff
@@ -0,0 +1,95 @@
...
+    capsys,
+    log_filter,
+    one_run,
+    capcall,
+):
+    """It should reinstall and restart the workflow with no changes.
+
+    See: https://github.com/cylc/cylc-flow/issues/6261
+
+    We hope to get users into the habbit of "cylc vip" to create a new run,
```
wxtim, line 40:
```suggestion
    We hope to get users into the habit of "cylc vip" to create a new run,
```

tests/integration/scripts/test_validate_reinstall.py
```diff
@@ -0,0 +1,95 @@
...
+    and "cylc vr" to contine an old one (picking up any new changes in the
+    process).
+
+    This works fine, unless there are no changes to reinstall, in which case
+    the "cylc vr" command exits (nothing to do).
+
+    The "nothing to reinstall" situation can be interpretted two ways:
```
wxtim, line 47:
```suggestion
    The "nothing to reinstall" situation can be interpreted two ways:
```

changes.d/6354.feat.md
oliver-sanders: Oh heck!

oliver-sanders: Fixed.

cylc/flow/scripts/validate_reinstall.py
```diff
@@ -189,34 +195,54 @@ async def vr_cli(
...
+            # there are no changes to install but the workflow isn't running
+            # => ask the user if they want to restart it anyway
+            usr = None
+            while usr not in ['y', 'n']:
+                LOG.warning(NO_CHANGES_STR)
+                usr = _input(
+                    cparse('<bold>Restart anyway?</bold> [y/n]: ')
+                ).lower()
+            if usr == 'n':
+                return False
```
wxtim, line 229: Worth testing?

wxtim (approved):
Alles Gut!

hjoliver (changes requested):
I'm not sure the logic is quite right here.

hjoliver:
### no changes, workflow running

```
$  cylc vr twt 
...
WARNING - No changes to reinstall: No reinstall or reload required.
```
⬆️ nothing happens ✔️ 

```
$ cylc vr twt --yes
...
Successfully reinstalled.
```
⬆️ reinstalls and tells the scheduler to reload ❌ 
- this should have the same result regardless of use of `--yes` because there is no prompt to respond to
- and there are no changes, so reinstall reload is pointless

oliver-sanders:
I think that behaviour is as intended (and mostly as before), the `--yes` argument is not *just* for the restart-anyway prompt, it also behaves like `--force`.

I'm guessing you would rather have two separate options for these?

Either way, bumping.

hjoliver:
> I think that behaviour is as intended (and mostly as before), the --yes argument is not just for the restart-anyway prompt, it also behaves like --force.

  1. Command help only says: `  --yes                 [reinstall] Skip interactive prompts.`
  1. I'm not aware of any precedent for `--yes` behaving like `--force`, so that's a bit surprising
  1. What's the point of supporting forced reinstall when we know there are no changes to update? It can't make any difference.
  
  > I'm guessing you would rather have two separate options for these?
  
  Not unless you have a good argument for 3. above - is there any need for forced reinstall?  If there is a need, then yes we should have a separate `--force` option. 
  
  BTW, `--force` could possibly be taken to imply `--yes` but not vice versa IMO.

oliver-sanders:
> What's the point of supporting forced reinstall when we know there are no changes to update? It can't make any difference.

The main use case would be forcing Rose file-installation to be reevaluated. This may cause newer versions of external dependencies to be brought into the workflow.

However, it's also just an easy way to achieve https://github.com/cylc/cylc-flow/issues/6261

hjoliver:
> However, it's also just an easy way to achieve https://github.com/cylc/cylc-flow/issues/6261

(Which is: "cylc vr: restart with no changes")

Not entirely sure I got my point across. I wasn't saying to not reinstall and not restart if there are no changes, just not reinstall.

> The main use case would be forcing Rose file-installation to be reevaluated.

If that's true, fine, but:
- the command should still generate the same no-source-changes-detected message whether or not a forced reinstall is done. 
- the option should be `--force`, not `--yes` which doesn't make sense when there are not prompts to answer to

oliver-sanders:
Ok, can I transfer this PR over to you to implement your desired behavior (might be easier than trying to explain). I don't really see the issue with `--yes` covering `--force` as I doubt users would every want one and not the other, but happy to leave this to your judgement.

Test harness is already in place so it's trivial work to fiddle CLI options.

hjoliver:
Fair enough, I'll tweak the options. Assigning myself...

oliver-sanders:
@hjoliver Poke

hjoliver:
(Blarggh, apologies for the delay)

oliver-sanders:
> Closing as superseded by documenting the current behaviour more clearly 

Erm, you blocked this with an objection to the CLI arguments, you were going to [suggest a new arrangement](https://github.com/cylc/cylc-flow/pull/6354#issuecomment-2502091674).

Don't really see any reason to close? Have you changed your mind about the [need for this](https://github.com/cylc/cylc-flow/issues/6261#issuecomment-2261604868)?

hjoliver:
> but you blocked the PR with a nit-picky objection to the CLI arguments

Not nit-picky, logical and valid!

> Have you changed your mind about the https://github.com/cylc/cylc-flow/issues/6261#issuecomment-2261604868?

Well yes, pretty much.  I commented on this a week ago on the original issue, which you saw and did not object to:

https://github.com/cylc/cylc-flow/issues/6261#issuecomment-3163344864

(Although I can see now the content of your response suggests you didn't grok that this was a change of direction).

Long story short:
- originally the behaviour did not match the documentation, and I thought that it should
- however making the behaviour match the documentation proved to be more difficult than expected (see above plus the fact that the we don't distinguish between "no changes to source" and "user responded No" to the reinstall prompt - as per my recent comment)
- changing the documentation instead of the behaviour is much easier, and I had to touch the command anyway for #6909 

In principle I think `cylc vr` probably should restart even if there are no changes, but the benefits are probably not worth the apparent effort as it turned out.

oliver-sanders:
> However, if you've come to think that it is important to do that

It was never important to us, I did it because you prioritised it, then changed your mind.

The CLI irk is yours to resolve. I don't think this feature is difficult to implement.

hjoliver:
>  I don't think this feature is tricky to implement.

Yeah, but it's logically tricky due to the various different branches of behaviour - hence the discussion above.  IMO the desirability of the feature isn't great enough to justify the effort it was taking to get it right. 

But I'll reopen this and take another look when time allows.

hjoliver:
I've resolved the `--yes` problem by adding a prompt for `reload`, so that `--yes` really does mean "skip interactive prompts" as documented, and the running/reload and stopped/restart logic is the same - which makes it easier to document. Obviously a reload without reinstalled changes is likely useless, but the prompt let's the user bail out of it, and it is possible to do  a `cylc vr` after modifying the run dir in place, so I think it's legit to allow both reload and restart when there were no changes reinstalled from source.

cylc/flow/scripts/validate_reinstall.py
```diff
@@ -18,12 +18,18 @@
...
 
-Validate and reinstall a workflow then either:
+If there are source changes and the user chooses (via prompt) to reinstall, or
+if there are no changes but the user chooses (via prompt) to continue anyway:
+* reload the workflow, if it is running) (see `cylc reload`)
+* or restart the workflow, if it is stopped (see `cylc play`)
 
-* "Reload" the workflow (if it is running),
-* or "play" it (if it is stopped).
+If the command is not running interactively, the command will automatically
```
hjoliver:
```suggestion
If the command is not running interactively, it will automatically
```

hjoliver:
(Not sure if this will require a bit more explaining or discussion, but I wanted to post the tweaked code now because I'm on leave tomorrow).

hjoliver:
(I thought I had made my changes as a separate commit to your branch @oliver-sanders , so you could see what I'd done - but one of us has evidently squashed the branch down ... hopefully the final result is short and simple enough that it won't matter for reviewing).

hjoliver:
Well actually the non-test changes (i.e. the important stuff) are small and in effect all-CLI.

oliver-sanders (commented):
Re-read the code, LGTM, tested interactive functionality, seems fine.

Pushed a commit to simplify the CLI text and remove some unused fixtures / variables in the tests.

(I can't approve my own PR for obvs reasons)

tests/integration/scripts/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+import pytest
+
+from cylc.flow.workflow_files import WorkflowFiles
+
+from ..utils.flow_writer import flow_config_str
+
+
+@pytest.fixture
+def one_src(tmp_path, one_conf):
+    src_dir = tmp_path
```
wxtim, line 31:
```suggestion
    """Create a workflow source directory at tmp_path
    """
    src_dir = tmp_path
```

tests/integration/scripts/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+    (src_dir / 'flow.cylc').write_text(flow_config_str(one_conf))
+    (src_dir / 'rose-suite.conf').touch()
```
wxtim: Is the `rose-suite.conf` necessary here?

tests/integration/scripts/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+    return src_dir
+
+
+@pytest.fixture
+def one_run(one_src, test_dir, run_dir):
```
wxtim, line 37:
```suggestion
def one_run(one_src, test_dir, run_dir):
    """Simplified install of workflow from ``one_src`` into cylc-run.
    """
```

wxtim (approved):
Suggested docstrings.

PR #6356, opened 2024-09-04
Flake8 B904

MetRonnie:


cylc/flow/dbstatecheck.py
```diff
@@ -378,7 +378,7 @@ def check_polling_config(selector, is_trigger, is_message):
         try:
             trigger = TASK_STATE_MAP[selector]
         except KeyError:
-            raise InputError(f'No such task state "{selector}"')
+            raise InputError(f'No such task state "{selector}"') from None
```
wxtim, line 381: Out of curiosity why this and not:

```python
except KeyError as keyerr:
    raise InputError('Message') from keyerr
```

Should not block approval or merging.

wxtim (approved):
1 Comment for my learning: Feel free to merge.

cylc/flow/dbstatecheck.py
MetRonnie: https://docs.python.org/3/tutorial/errors.html#exception-chaining

`KeyError` is the type of error that will never have any useful information to the user, so there is no point including it in the traceback, it would just be noise

`from exc`:

```
KeyError                                  Traceback (most recent call last)
Cell In[3], line 2
      1 try:
----> 2     dict_['x']
      3 except KeyError as exc:

KeyError: 'x'

The above exception was the direct cause of the following exception:

UsefulError                               Traceback (most recent call last)
Cell In[3], line 4
      2     dict_['x']
      3 except KeyError as exc:
----> 4     raise UsefulError('Oopsie') from exc

UsefulError: Oopsie
```

vs `from None`:
```
UsefulError                               Traceback (most recent call last)
Cell In[4], line 4
      2     dict_['y']
      3 except KeyError:
----> 4     raise UsefulError('Oopsie') from None

UsefulError: Oopsie
```

vs no `from` at all:
```
KeyError                                  Traceback (most recent call last)
Cell In[5], line 2
      1 try:
----> 2     dict_['z']
      3 except KeyError:

KeyError: 'z'

UsefulError                               Traceback (most recent call last)
Cell In[5], line 4
      2     dict_['z']
      3 except KeyError:
----> 4     raise UsefulError('Oopsie')

UsefulError: Oopsie
```

PR #6357, opened 2024-09-04
Fix potential type error & a few small mistakes

MetRonnie:
Small change with no associated issue

**Check List**

- [x] No dependency changes 
- [x] tests are not needed
- [x] No changelog entry as not known bug
- [x] No docs needed

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -116,6 +116,7 @@ from cylc.flow.wallclock import (
 if TYPE_CHECKING:
     from cylc.flow.cycling import PointBase
     from cylc.flow.flow_mgr import FlowNums
+    from cylc.flow.scheduler import Scheduler
 
 EDGES = 'edges'
 FAMILIES = 'families'
@@ -468,7 +469,7 @@ class DataStoreMgr:
     ERR_PREFIX_JOB_NOT_ON_SEQUENCE = 'Invalid cycle point for job: '
 
     def __init__(self, schd, n_edge_distance=1):
-        self.schd = schd
+        self.schd: Scheduler = schd
         self.id_ = Tokens(
             user=self.schd.owner,
             workflow=self.schd.workflow,
@@ -1182,10 +1183,7 @@ class DataStoreMgr:
         t_id = self.definition_id(name)
 
         if itask is None:
-            itask = self.schd.pool.get_task(point_string, name)
-
-        if itask is None:
-            itask = TaskProxy(
+            itask = self.schd.pool.get_task(point_string, name) or TaskProxy(
                 self.id_,
                 self.schd.config.get_taskdef(name),
                 point,
@@ -1226,6 +1224,7 @@ class DataStoreMgr:
             depth=task_def.depth,
             graph_depth=n_depth,
             name=name,
+            flow_nums=serialise_set(flow_nums),
         )
         self.all_n_window_nodes.add(tp_id)
         self.n_window_depths.setdefault(n_depth, set()).add(tp_id)
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1332,9 +1332,7 @@ class TaskPool:
         """
         return self.abort_task_failed
 
-    def spawn_on_output(
-        self, itask: TaskProxy, output: str, forced: bool = False
-    ) -> None:
+    def spawn_on_output(self, itask: TaskProxy, output: str) -> None:
         """Spawn child-tasks of given output, into the pool.
 
         Remove the parent task from the pool if complete.
@@ -1355,7 +1353,6 @@ class TaskPool:
 
         Args:
             output: output to spawn on.
-            forced: True if called from manual set task command
 
         """
         if (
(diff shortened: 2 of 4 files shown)


cylc/flow/task_pool.py
```diff
@@ -1329,7 +1329,9 @@ def check_abort_on_task_fails(self):
         """
         return self.abort_task_failed
 
-    def spawn_on_output(self, itask, output, forced=False):
+    def spawn_on_output(
+        self, itask: TaskProxy, output: str, forced: bool = False
```
MetRonnie, line 1333: Actually, I have discovered that `forced` is not used at all 🤔

It only has coverage because there is an integration test that covers it

https://github.com/cylc/cylc-flow/blob/1c5db24c7f98cf9ce749959200732e93431f89d8/tests/integration/test_task_pool.py#L1071

but it is not actually used.

hjoliver: From a quick look, this method used to be called from the `cylc set` logic but no longer is. So you might as well remove the `forced` flag in this PR.

MetRonnie:
Another similar mistake spotted for the `TaskPool.spawn_task()` method (not to be confused with `TaskEventsManager.spawn_task()` mentioned before): https://github.com/cylc/cylc-flow/blob/f30f0c289e1cb803808df4f9131424c23e00b71e/cylc/flow/task_pool.py#L1657-L1664

`flow_wait` is incorrectly passed to the `force` argument in 2 places:

https://github.com/cylc/cylc-flow/blob/f30f0c289e1cb803808df4f9131424c23e00b71e/cylc/flow/task_pool.py#L744-L745

https://github.com/cylc/cylc-flow/blob/f30f0c289e1cb803808df4f9131424c23e00b71e/cylc/flow/task_pool.py#L2015-L2020

hjoliver:
(Ah, yeah I found that in my group-trigger branch, which I haven't exposed to you all yet)

cylc/flow/task_pool.py
```diff
@@ -1419,7 +1417,10 @@ def spawn_on_output(self, itask, output, forced=False):
                     if not in_pool:
                         self.add_to_pool(t)
 
-                    if t.point <= self.runahead_limit_point:
+                    if (
+                        self.runahead_limit_point is not None
+                        and t.point <= self.runahead_limit_point
+                    ):
```
oliver-sanders, line 1423: Shouldn't this line have similar logic:

https://github.com/cylc/cylc-flow/blob/4b1adfc5b777102a8b9d65fdaa5ba1ce896b46a4/cylc/flow/task_pool.py#L781-L785

Given that `compute_runahead` can return `False` before updating the value:

https://github.com/cylc/cylc-flow/blob/4b1adfc5b777102a8b9d65fdaa5ba1ce896b46a4/cylc/flow/task_pool.py#L349-L350

MetRonnie: Are you saying add it to line 785? I.e.
```diff
**diff --git a/cylc/flow/task_pool.py b/cylc/flow/task_pool.py
index 01b0d03e91..c3574ee39a 100644
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -784,4 +784,5 @@ class TaskPool:
 
         is_xtrig_sequential = False
+        if self.runahead_limit_point is not None:
             while point is not None and (point <= self.runahead_limit_point):
                 if tdef.is_parentless(point):
```

oliver-sanders: No idea what the fix should be, but I think it's the same `runahead_limit_point` could be `None` problem you've fixed in the commented lines?

MetRonnie: Addressed

cylc/flow/data_store_mgr.py
```diff
@@ -468,7 +469,7 @@ class DataStoreMgr:
     ERR_PREFIX_JOB_NOT_ON_SEQUENCE = 'Invalid cycle point for job: '
 
     def __init__(self, schd, n_edge_distance=1):
-        self.schd = schd
+        self.schd: Scheduler = schd
```
hjoliver, line 472:
```suggestion
        self.schd: 'Scheduler' = schd
```
(However, you might know something I don't, since lacking of quoting here doesn't seem to cause a problem even though Scheduler is only imported when type checking ??)

hjoliver (approved):
LGTM, just one question.

cylc/flow/data_store_mgr.py
oliver-sanders: Quotes are needed if the type is not imported [1] or if it is declared further down the file.

> [1] This includes dynamic and `if TYPE_CHECKIKNG` imports.

hjoliver: This one is imported in the `if TYPE_CHECKING` block, but weirdly seems to survive non-quoting.

MetRonnie: Quotes for types imported inside the `TYPE_CHECKING` block are only needed when the type is used in function declarations

hjoliver: Ah, that explains it!  (was news to me).

MetRonnie:
Can bump to 8.3.5 if needed

oliver-sanders (approved):

PR #6364, opened 2024-09-10
`cylc clean`: fix bug where `--rm share` would not clean `share/cycle` symlink first

MetRonnie:
Fixes a bug possibly (?) seen before by some users at the Met Office, but not tracked by an issue.

If `share/cycle` is a symlink directory, it must also be cleaned first when only specifying `--rm share`.

#### Reproduceable example

```cylc
# global.cylc
[install]
    [[symlink dirs]]
        [[[localhost]]]
            share = $HOME/sym-other
            share/cycle = $HOME/sym-cycle
```
```
$ cylc play <workflow>
$ cylc stop <workflow>
$ cylc clean <workflow> --rm share
# Restart:
$ cylc play <workflow>
WorkflowFilesError: Symlink dir target already exists: (~/cylc-run/<workflow>/share/cycle ->) ~/sym-cycle/cylc-run/<workflow>/share/cycle
Tip: in future, use 'cylc clean' instead of manually deleting workflow run dirs.
```

#### Check List

- [x] No dependency changes 
- [x] Tests are included 
- [x] No docs needed

### DIFF
--- a/cylc/flow/clean.py
+++ b/cylc/flow/clean.py
@@ -117,12 +117,8 @@ def _clean_check(opts: 'Values', id_: str, run_dir: Path) -> None:
     # Thing to clean must be a dir or broken symlink:
     if not run_dir.is_dir() and not run_dir.is_symlink():
         raise FileNotFoundError(f"No directory to clean at {run_dir}")
-    db_path = (
-        run_dir / WorkflowFiles.Service.DIRNAME / WorkflowFiles.Service.DB
-    )
-    if opts.local_only and not db_path.is_file():
-        # Will reach here if this is cylc clean re-invoked on remote host
-        # (workflow DB only exists on scheduler host); don't need to worry
+    if opts.no_scan:
+        # This is cylc clean re-invoked on remote host; don't need to worry
         # about contact file.
         return
     try:
--- a/cylc/flow/clean.py
+++ b/cylc/flow/clean.py
@@ -259,7 +259,7 @@ def glob_in_run_dir(
     """Execute a (recursive) glob search in the given run directory.
 
     Returns list of any absolute paths that match the pattern. However:
-    * Does not follow symlinks (apart from the spcedified symlink dirs).
+    * Does not follow symlinks (apart from the specified symlink dirs).
     * Also does not return matching subpaths of matching directories (because
         that would be redundant).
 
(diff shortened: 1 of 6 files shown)


oliver-sanders:
> Fixes a bug possibly (?) seen before by some users at the Met Office, but not tracked by an issue.

If it only applies to the `--rm` option, then I don't think this has been reported. There was an unconfirmed report of `cylc clean` (no opts) leaving files behind, however, the evidence was destroyed so it couldn't be verified.

cylc/flow/clean.py
```diff
@@ -117,12 +117,8 @@ def _clean_check(opts: 'Values', id_: str, run_dir: Path) -> None:
...
     if not run_dir.is_dir() and not run_dir.is_symlink():
         raise FileNotFoundError(f"No directory to clean at {run_dir}")
-    db_path = (
-        run_dir / WorkflowFiles.Service.DIRNAME / WorkflowFiles.Service.DB
-    )
-    if opts.local_only and not db_path.is_file():
-        # Will reach here if this is cylc clean re-invoked on remote host
-        # (workflow DB only exists on scheduler host); don't need to worry
+    if opts.no_scan:
+        # This is cylc clean re-invoked on remote host; don't need to worry
```
oliver-sanders, line 121: Might need a bit more to go on here than "Simplify"?

oliver-sanders:
> $ cylc play <workflow>
WorkflowFilesError: Symlink dir target already exists: (~/cylc-run/<workflow>/share/cycle ->) ~/sym-cycle/cylc-run/<workflow>/share/cycle
Tip: in future, use 'cylc clean' instead of manually deleting workflow run dirs.

I actually don't get this traceback, however, I can confirm the share/cycle directory is left behind.

cylc/flow/clean.py
MetRonnie: Both before and after do the same thing: check if this is `cylc clean` reinvoked on the remote host. The original implementation came before the `--no-scan` option was added: https://github.com/cylc/cylc-flow/pull/5631/files#diff-805b35ed176db500a061c8f6127eebd12908264e511589447db63b58eb7a4623R128-R133

cylc/flow/clean.py
```diff
@@ -326,13 +326,16 @@ def _clean_using_glob(
...
         return
     # First clean any matching symlink dirs
-    for path in abs_symlink_dirs:
-        if path in matches:
-            remove_dir_and_target(path)
-            if path == run_dir:
+    for symlink_dir in abs_symlink_dirs:
+        # Note: must clean e.g. share/cycle/ before share/ if the former
+        # is a symlink even if only the latter was specified.
+        if any(is_relative_to(symlink_dir, path) for path in matches):
```
MetRonnie, line 332: @oliver-sanders RE: the traceback you were getting, can you try this in your testing?
```suggestion
        if (
            any(is_relative_to(symlink_dir, path) for path in matches)
            and symlink_dir.is_symlink()
        ):
```

oliver-sanders: Seems to do the trick!

oliver-sanders (approved):

Issue #6143, opened 2024-06-14
pool: check prereqs on task spawn


oliver-sanders:
When we spawn a task, we currently check the DB to retrieve previous submissions. This is required for:
* Flow merge detection.
* Task re-run prevention (within the same flow).
* Preventing the accidental reincarnation of removed tasks.

Whilst we do check for previous submissions, we do not presently check task prerequisites. In some situations, task prerequisites which are satisfied in the DB, are left unsatisfied in the task pool when tasks are spawned. This can happen when tasks are added to the pool via means other than natural task satisfaction (e.g. https://github.com/cylc/cylc-flow/issues/5952).

Easy way to replicate this:
* Empty the task pool.
* Put the tasks back.
* Run the workflow on.

Sadly, I don't this it is sufficient to check the DB _only_ when such interventions are performed, because tasks downstream of the ones being added which are yet to be spawned may also be missing this state and end up with (erroneously) partially satisfied prerequisites.

This is a bad bug as it makes it look to the user like tasks which have run, haven't. It's hard to explain, especially as multiple Cylc interfaces will provide erroneous information. Moreover it's very hard to recover from as the consequences may last as long as the longest inter-cycle dependency in the workflow.

There are efficiency concerns over requesting task prerequisites, however, we haven't ascertained that this would be a problem as yet, however, DB processing is definitely a bottleneck. One way to reduce these overheads might be to only request satisfied prerequisites from the DB. Merging this request in with the existing DB request would also improve performance. As would batching the requests where multiple tasks are spawned in the same main loop iteration.

This issue has relevance to workflow extension use cases (#5952), graph changes either by reload or restart

hjoliver:
> Sadly, I don't this it is sufficient to check the DB only when such interventions are performed, because tasks downstream of the ones being added which are yet to be spawned may also be missing this state and end up with (erroneously) partially satisfied prerequisites

Good point.  It would be sufficient to check the DB only when the workflow stalls, but with significant cons:
- there could be a series of stalls until full recovery
- there could be a confusing period before the stall where tasks appear to be unsatisfied

oliver-sanders:
I think the only way around this (not that I'm suggesting it per-se) would be to load in the task history from the DB and spawn on all outputs. However, this would have the caveat that we would be spawning according to the current graph which may differ from the graph historical tasks were running according to (i.e. it could result in spawning errors).

oliver-sanders:
This has now been discussed in a couple of different places.

[...]

PR #6370, opened 2024-09-11
Make `cylc remove` flow-aware and extend to historical tasks

MetRonnie:
Partially addresses #5643

### Summary

This mostly implements [the "Cylc Remove Extension" proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-remove.md).

#### Flow numbers

`cylc remove` now has a `--flow` option for removing a task from specific flows.

If not used, it will remove the task from all flows that it belongs to.

If the removed task is active/waiting, if it is removed from a subset of flows that it belongs to, it will remain in the task pool; if it is removed from all flows that it belongs to, it will be removed from the task pool (as is the current behaviour).

If a task is removed from all flows that it belongs to, it will become a no-flow task (`flow=None`).

For ease of reviewing, you can use my UI branch that displays flow numbers: https://github.com/MetRonnie/cylc-ui/tree/flow-nums [^c].

[^c]: Waiting tasks that are not yet in the pool have greyed out flow numbers at the moment.

#### Historical tasks

`cylc remove` now can remove tasks that are no longer active, making it look like they never ran. It does this by removing the task from the specified flows in the database (in the `task_states` and `task_outputs` tables)[^a], and un-setting any prerequisites of active tasks that the removed task had naturally satisfied[^b]. If a task is removed from all flows that it belongs to, a no-flow task is left in the DB for provenance.

The above also applies to active/waiting tasks that `cylc remove` is used on.

[^a]: If removing flows would result in two rows in the DB no longer being unique, the SQLite [`UPDATE OR REPLACE`](https://www.sqlite.org/lang_update.html) statement is used, so the first entry will be removed and the most recent entry will remain.
[^b]: Prerequisites manually satisfied by `cylc set --pre` are not affected by `cylc remove`.

### What's left to do

- [ ] When removing an active task from all its flows, kill the task.
- [ ] Should probably add a functional test with the `--flow` option.
- [x] Need to check this one:

  > If removing a task causes all of the prerequisites on a downstream task to be unset (i.e. if the downstream task was spawned as a result of outputs from this task alone) then the downstream task shall be removed from the pool.

### Check List

- [x] No dependency changes
- [x] Tests are included 

### DIFF
--- a/cylc/flow/command_validation.py
+++ b/cylc/flow/command_validation.py
@@ -24,12 +24,22 @@ from typing import (
 )
 
 from cylc.flow.exceptions import InputError
-from cylc.flow.id import IDTokens, Tokens
+from cylc.flow.flow_mgr import (
+    FLOW_ALL,
+    FLOW_NEW,
+    FLOW_NONE,
+)
+from cylc.flow.id import (
+    IDTokens,
+    Tokens,
+)
 from cylc.flow.task_outputs import TASK_OUTPUT_SUCCEEDED
-from cylc.flow.flow_mgr import FLOW_ALL, FLOW_NEW, FLOW_NONE
 
 
-ERR_OPT_FLOW_VAL = "Flow values must be an integer, or 'all', 'new', or 'none'"
+ERR_OPT_FLOW_VAL = (
+    f"Flow values must be an integer, or '{FLOW_ALL}', '{FLOW_NEW}', "
+    f"or '{FLOW_NONE}'"
+)
 ERR_OPT_FLOW_COMBINE = "Cannot combine --flow={0} with other flow values"
 ERR_OPT_FLOW_WAIT = (
     f"--wait is not compatible with --flow={FLOW_NEW} or --flow={FLOW_NONE}"
--- a/cylc/flow/command_validation.py
+++ b/cylc/flow/command_validation.py
@@ -40,13 +40,18 @@ ERR_OPT_FLOW_VAL = (
     f"Flow values must be an integer, or '{FLOW_ALL}', '{FLOW_NEW}', "
     f"or '{FLOW_NONE}'"
 )
+ERR_OPT_FLOW_VAL_2 = f"Flow values must be an integer, or '{FLOW_ALL}'"
 ERR_OPT_FLOW_COMBINE = "Cannot combine --flow={0} with other flow values"
 ERR_OPT_FLOW_WAIT = (
     f"--wait is not compatible with --flow={FLOW_NEW} or --flow={FLOW_NONE}"
 )
 
 
-def flow_opts(flows: List[str], flow_wait: bool) -> None:
+def flow_opts(
+    flows: List[str],
+    flow_wait: bool,
+    allow_new_or_none: bool = True
+) -> None:
     """Check validity of flow-related CLI options.
 
     Note the schema defaults flows to [].
@@ -73,6 +78,10 @@ def flow_opts(flows: List[str], flow_wait: bool) -> None:
         cylc.flow.exceptions.InputError: --wait is not compatible with
         --flow=new or --flow=none
 
+        >>> flow_opts(["new"], False, allow_new_or_none=False)
+        Traceback (most recent call last):
+        cylc.flow.exceptions.InputError: ... must be an integer, or 'all'
+
     """
     if not flows:
         return
@@ -80,9 +89,12 @@ def flow_opts(flows: List[str], flow_wait: bool) -> None:
     flows = [val.strip() for val in flows]
 
     for val in flows:
+        val = val.strip()
         if val in {FLOW_NONE, FLOW_NEW, FLOW_ALL}:
             if len(flows) != 1:
                 raise InputError(ERR_OPT_FLOW_COMBINE.format(val))
+            if not allow_new_or_none and val in {FLOW_NEW, FLOW_NONE}:
+                raise InputError(ERR_OPT_FLOW_VAL_2)
         else:
             try:
                 int(val)
--- a/cylc/flow/commands.py
+++ b/cylc/flow/commands.py
@@ -62,6 +62,7 @@ from typing import (
     List,
     Optional,
     TYPE_CHECKING,
+    TypeVar,
     Union,
 )
 
@@ -86,10 +87,9 @@ if TYPE_CHECKING:
     from cylc.flow.scheduler import Scheduler
 
     # define a type for command implementations
-    Command = Callable[
-        ...,
-        AsyncGenerator,
-    ]
+    Command = Callable[..., AsyncGenerator]
+    # define a generic type needed for the @_command decorator
+    _TCommand = TypeVar('_TCommand', bound=Command)
 
 # a directory of registered commands (populated on module import)
 COMMANDS: 'Dict[str, Command]' = {}
@@ -97,15 +97,15 @@ COMMANDS: 'Dict[str, Command]' = {}
 
 def _command(name: str):
     """Decorator to register a command."""
-    def _command(fcn: 'Command'):
+    def _command(fcn: '_TCommand') -> '_TCommand':
         nonlocal name
         COMMANDS[name] = fcn
-        fcn.command_name = name  # type: ignore
+        fcn.command_name = name  # type: ignore[attr-defined]
         return fcn
     return _command
 
 
-async def run_cmd(fcn, *args, **kwargs):
+async def run_cmd(bound_fcn: AsyncGenerator):
     """Run a command outside of the scheduler's main loop.
 
     Normally commands are run via the Scheduler's command_queue (which is
@@ -120,10 +120,9 @@ async def run_cmd(fcn, *args, **kwargs):
     For these purposes use "run_cmd", otherwise, queue commands via the
     scheduler as normal.
     """
-    cmd = fcn(*args, **kwargs)
-    await cmd.__anext__()  # validate
+    await bound_fcn.__anext__()  # validate
     with suppress(StopAsyncIteration):
-        return await cmd.__anext__()  # run
+        return await bound_fcn.__anext__()  # run
 
 
 @_command('set')
--- a/cylc/flow/commands.py
+++ b/cylc/flow/commands.py
@@ -53,18 +53,23 @@ For more info see: https://github.com/cylc/cylc-flow/issues/3329
 """
 
 from contextlib import suppress
-from time import sleep, time
+from time import (
+    sleep,
+    time,
+)
 from typing import (
+    TYPE_CHECKING,
     AsyncGenerator,
     Callable,
     Dict,
     Iterable,
     List,
     Optional,
-    TYPE_CHECKING,
     TypeVar,
     Union,
 )
+
+from metomi.isodatetime.parsers import TimePointParser
 
 from cylc.flow import LOG
 import cylc.flow.command_validation as validate
@@ -74,14 +79,21 @@ from cylc.flow.exceptions import (
     CylcConfigError,
 )
 import cylc.flow.flags
+from cylc.flow.flow_mgr import get_flow_nums_set
+from cylc.flow.id import tokenise
 from cylc.flow.log_level import log_level_to_verbosity
 from cylc.flow.network.schema import WorkflowStopMode
 from cylc.flow.parsec.exceptions import ParsecError
 from cylc.flow.task_id import TaskID
-from cylc.flow.task_state import TASK_STATUSES_ACTIVE, TASK_STATUS_FAILED
-from cylc.flow.workflow_status import RunMode, StopMode
+from cylc.flow.task_state import (
+    TASK_STATUS_FAILED,
+    TASK_STATUSES_ACTIVE,
+)
+from cylc.flow.workflow_status import (
+    RunMode,
+    StopMode,
+)
 
-from metomi.isodatetime.parsers import TimePointParser
 
 if TYPE_CHECKING:
     from cylc.flow.scheduler import Scheduler
@@ -313,11 +325,22 @@ async def set_verbosity(schd: 'Scheduler', level: Union[int, str]):
 
 
 @_command('remove_tasks')
-async def remove_tasks(schd: 'Scheduler', tasks: Iterable[str]):
+async def remove_tasks(
+    schd: 'Scheduler', tasks: Iterable[str], flow: List[str]
+):
     """Remove tasks."""
     validate.is_tasks(tasks)
+    validate.flow_opts(flow, flow_wait=False, allow_new_or_none=False)
     yield
-    yield schd.pool.remove_tasks(tasks)
+    flow_nums = get_flow_nums_set(flow)
+    schd.pool.remove_tasks(tasks)
+    for task in tasks:
+        task_id = tokenise(task, relative=True)
+        schd.workflow_db_mgr.remove_task_from_flows(
+            point=task_id['cycle'],
+            name=task_id['task'],
+            flow_nums=flow_nums
+        )
 
 
 @_command('reload_workflow')
--- a/cylc/flow/commands.py
+++ b/cylc/flow/commands.py
@@ -80,7 +80,6 @@ from cylc.flow.exceptions import (
 )
 import cylc.flow.flags
 from cylc.flow.flow_mgr import get_flow_nums_set
-from cylc.flow.id import tokenise
 from cylc.flow.log_level import log_level_to_verbosity
 from cylc.flow.network.schema import WorkflowStopMode
 from cylc.flow.parsec.exceptions import ParsecError
@@ -333,14 +332,7 @@ async def remove_tasks(
     validate.flow_opts(flow, flow_wait=False, allow_new_or_none=False)
     yield
     flow_nums = get_flow_nums_set(flow)
-    schd.pool.remove_tasks(tasks)
-    for task in tasks:
-        task_id = tokenise(task, relative=True)
-        schd.workflow_db_mgr.remove_task_from_flows(
-            point=task_id['cycle'],
-            name=task_id['task'],
-            flow_nums=flow_nums
-        )
+    schd.pool.remove_tasks(tasks, flow_nums)
 
 
 @_command('reload_workflow')
--- a/cylc/flow/flow_mgr.py
+++ b/cylc/flow/flow_mgr.py
@@ -16,8 +16,15 @@
 
 """Manage flow counter and flow metadata."""
 
-from typing import Dict, Set, Optional, TYPE_CHECKING
 import datetime
+from typing import (
+    TYPE_CHECKING,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Set,
+)
 
 from cylc.flow import LOG
 
@@ -55,36 +62,62 @@ def add_flow_opts(parser):
     )
 
 
-def stringify_flow_nums(flow_nums: Set[int], full: bool = False) -> str:
-    """Return a string representation of a set of flow numbers
+def get_flow_nums_set(flow: List[str]) -> FlowNums:
+    """Return set of integer flow numbers from list of strings.
 
-    Return:
-    - "none" for no flow
-    - "" for the original flow (flows only matter if there are several)
-    - otherwise e.g. "(flow=1,2,3)"
+    Returns an empty set if the input is empty or contains only "all".
+
+    >>> get_flow_nums_set(["1", "2", "3"])
+    {1, 2, 3}
+    >>> get_flow_nums_set([])
+    set()
+    >>> get_flow_nums_set(["all"])
+    set()
+    """
+    if flow == [FLOW_ALL]:
+        return set()
+    return {int(val.strip()) for val in flow}
+
+
+def stringify_flow_nums(flow_nums: Iterable[int]) -> str:
+    """Return the canonical string for a set of flow numbers.
 
     Examples:
-        >>> stringify_flow_nums({})
-        '(flows=none)'
-
         >>> stringify_flow_nums({1})
+        '1'
+
+        >>> stringify_flow_nums({3, 1, 2})
+        '1,2,3'
+
+        >>> stringify_flow_nums({})
         ''
 
-        >>> stringify_flow_nums({1}, True)
+    """
+    return ','.join(str(i) for i in sorted(flow_nums))
+
+
+def repr_flow_nums(flow_nums: FlowNums, full: bool = False) -> str:
+    """Return a representation of a set of flow numbers
+
+    If `full` is False, return an empty string for flows=1.
+
+    Examples:
+        >>> repr_flow_nums({})
+        '(flows=none)'
+
+        >>> repr_flow_nums({1})
+        ''
+
+        >>> repr_flow_nums({1}, full=True)
         '(flows=1)'
 
-        >>> stringify_flow_nums({1,2,3})
+        >>> repr_flow_nums({1,2,3})
         '(flows=1,2,3)'
 
     """
     if not full and flow_nums == {1}:
         return ""
-    else:
-        return (
-            "(flows="
-            f"{','.join(str(i) for i in flow_nums) or 'none'}"
-            ")"
-        )
+    return f"(flows={stringify_flow_nums(flow_nums) or 'none'})"
 
 
 class FlowMgr:
--- a/cylc/flow/prerequisite.py
+++ b/cylc/flow/prerequisite.py
@@ -39,7 +39,7 @@ from typing_extensions import Literal
 from cylc.flow.cycling.loader import get_point
 from cylc.flow.data_messages_pb2 import PbCondition, PbPrerequisite
 from cylc.flow.exceptions import TriggerExpressionError
-from cylc.flow.id import quick_relative_detokenise
+from cylc.flow.id import quick_relative_id
 
 
 if TYPE_CHECKING:
@@ -58,7 +58,7 @@ class PrereqMessage(NamedTuple):
 
     def get_id(self) -> str:
         """Get the relative ID of the task in this prereq message."""
-        return quick_relative_detokenise(self.point, self.task)
+        return quick_relative_id(self.point, self.task)
 
     @staticmethod
     def coerce(tuple_: AnyPrereqMessage) -> 'PrereqMessage':
@@ -253,12 +253,17 @@ class Prerequisite:
             ) from None
         return res
 
-    def satisfy_me(self, outputs: Iterable['Tokens']) -> 'Set[Tokens]':
-        """Attempt to satisfy me with given outputs.
+    def satisfy_me(
+        self, outputs: Iterable['Tokens'], forced: bool = False
+    ) -> 'Set[Tokens]':
+        """Set the given outputs as satisfied.
 
-        Updates cache with the result.
         Return outputs that match.
 
+        Args:
+            outputs: List of outputs to satisfy.
+            forced: If True, records that this should not be undone by
+                `cylc remove`.
         """
         valid = set()
         for output in outputs:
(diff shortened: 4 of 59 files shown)


oliver-sanders (dismissed):

cylc/flow/scheduler.py
```diff
@@ -16,63 +16,51 @@
 """Cylc scheduler server."""
 
 import asyncio
+import os
+import sys
+import traceback
 from collections import deque
 from contextlib import suppress
```
oliver-sanders, line 23: You seem to have some strange isort configuration that is grouping and sorting `import` and `from * import` statements separately?

MetRonnie: isort config added in #6382; will rebase and sort out

cylc/flow/workflow_db_mgr.py
```diff
@@ -444,7 +439,7 @@ def put_update_task_state(self, itask: 'TaskProxy') -> None:
         }
         # Note tasks_states table rows are for latest submit_num only
         # (not one row per submit).
-        self.db_updates_map.setdefault(self.TABLE_TASK_STATES, []).append(
+        self.db_updates_map[self.TABLE_TASK_STATES].append(
```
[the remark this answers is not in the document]
MetRonnie: I changed it from a `dict` to a `defaultdict` to avoid the need to `setdefault()` all the time

cylc/flow/task_pool.py
```diff
@@ -2032,28 +2044,108 @@ def _get_active_flow_nums(self) -> Set[int]:
...
+                for msg in prereq.naturally_satisfied_dependencies():
+                    id_ = msg.get_id()
+                    if id_ in matched_task_ids:
+                        prereq[msg] = False
+                        if id_ not in removed:
+                            removed[id_] = fnums_to_remove
+
+        # Remove from DB tables:
+        for id_ in matched_task_ids:
+            point, name = id_.split('/', 1)
```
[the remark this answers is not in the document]
MetRonnie: This way was the easiest, considering all the ways `matched_task_ids` is used in this function

tests/conftest.py
```diff
@@ -97,31 +97,26 @@ def _inner(cached=False):
...
-    ) -> List[Tuple[str, int, str]]:
+    ) -> List[Tuple[int, str]]:
         return [
-            (log_name, log_level, log_message)
-            for log_name, log_level, log_message in log.record_tuples
-            if (name is None or name == log_name)
-            and (level is None or level == log_level)
+            (log_level, log_message)
+            for _, log_level, log_message in caplog.record_tuples
+            if (level is None or level == log_level)
```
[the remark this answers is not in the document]
MetRonnie: Should be addressed now

tests/integration/conftest.py
```diff
@@ -534,21 +534,21 @@ def reflog():
 
     """
 
-    def _reflog(schd: 'Scheduler', flow_nums: bool = False) -> Set[tuple]:
+    def _reflog(schd: 'Scheduler', flow_nums: bool = False) -> List[tuple]:
         submit_task_jobs = schd.task_job_mgr.submit_task_jobs
-        triggers = set()
+        triggers = []
```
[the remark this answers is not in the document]
MetRonnie: > It was done this way for a reason! 

Then I've added the reason in the docstring and undone the rest of the changes

oliver-sanders:
Code looks good, trying out some manual testing.

I'm not having much luck removing n=-1 tasks (i.e. tasks which have just succeeded). Here's an example:

```cylc
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    [[graph]]
        P1 = a[-P1] => a
[runtime]
    [[root]]
    [[a]]
        script = sleep 1
        completion = (submitted and succeeded)
```

```bash
# start the workflow
cylc vip --pause

# run task 1/a
cylc play
cylc pause

# remove task 1/a
cylc remove 1/a
```

Results:

* The DB is updated correctly.
* However the prereq `1/a -> 2/a` is marked as satisfied in the `cylc show` output.
* When I resume the workflow `2/a` runs (note it was in the queue when `1/a` was removed).

tests/integration/tui/conftest.py
```diff
@@ -4,7 +4,7 @@
 from pathlib import Path
 import re
 from time import sleep
-from uuid import uuid1
+from secrets import token_hex
```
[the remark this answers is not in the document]
MetRonnie: To shorten the integration test workflow IDs which were really long and making life a bit more difficult than it needed to be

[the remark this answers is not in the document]
oliver-sanders: Gotcha.

cylc/flow/command_validation.py
```diff
@@ -58,21 +73,30 @@ def flow_opts(flows: List[str], flow_wait: bool) -> None:
...
-        cylc.flow.exceptions.InputError: ...
+        cylc.flow.exceptions.InputError: --wait is not compatible ...
+
+        >>> flow_opts(["new"], False, allow_new_or_none=False)
+        Traceback (most recent call last):
+        cylc.flow.exceptions.InputError: ... must be an integer, or 'all'
 
     """
+    if flows is None:
+        return
```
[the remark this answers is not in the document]
MetRonnie: (I don't think `flows` still needs to be `Optional` and so I'll revert this - was probably needed for WIP stuff that was not used)

MetRonnie:
(Test failure is just flaky tui test, not bothering re-running a 3rd time 🤮)

cylc/flow/task_pool.py
```diff
@@ -2032,28 +2044,108 @@ def _get_active_flow_nums(self) -> Set[int]:
...
+                itask.flow_nums.difference_update(fnums_to_remove)
+
+        matched_task_ids = {
+            *removed.keys(),
+            *(quick_relative_id(cycle, task) for task, cycle in inactive),
+        }
+
+        # Unset any prereqs naturally satisfied by these tasks
+        # (do not unset those satisfied by `cylc set --pre`):
+        for itask in self.get_tasks():
```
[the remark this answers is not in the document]
MetRonnie: Done

oliver-sanders:
Have been trying this out for sub-graph re-run use cases. I've been hitting this issue a lot (https://github.com/cylc/cylc-flow/issues/6143) otherwise, works fine :+1:.

oliver-sanders:
What's the status on these TODO items:

> What's left to do
>
>    When removing an active task from all its flows, kill the task.
>
>    Should probably add a functional test with the --flow option.
>
>    Need to check this one:

MetRonnie:
> When removing an active task from all its flows, kill the task.

Will come in follow-up PR

> Should probably add a functional test with the `--flow` option.

Will either come in the follow-up or this one depending on how soon @hjoliver reviews this

> Need to check this one:
>   > If removing a task causes all of the prerequisites on a downstream task to be unset (i.e. if the downstream task was spawned as a result of outputs from this task alone) then the downstream task shall be removed from the pool.

Will ask you about this in-person

oliver-sanders:
> This means prerequisites which have been satisfied by cylc trigger or cylc set will not be unset as the result of cylc remove as these have been manually satisfied by separate interventions.

This statement is in relation to prerequisites satisfied by the task being removed, not the prerequisites of the removed task itself.

cylc/flow/commands.py
```diff
@@ -73,39 +79,43 @@
...
 if TYPE_CHECKING:
     from cylc.flow.scheduler import Scheduler
 
     # define a type for command implementations
-    Command = Callable[
-        ...,
-        AsyncGenerator,
-    ]
+    Command = Callable[..., AsyncGenerator]
+    _TCommand = TypeVar('_TCommand', bound=Command)
```
wxtim, line 102: What does `_TCommand` mean?

cylc/flow/commands.py
```diff
@@ -314,11 +323,15 @@ async def set_verbosity(schd: 'Scheduler', level: Union[int, str]):
...
-async def remove_tasks(schd: 'Scheduler', tasks: Iterable[str]):
+async def remove_tasks(
+    schd: 'Scheduler', tasks: Iterable[str], flow: List[str]
+):
     """Remove tasks."""
     validate.is_tasks(tasks)
+    validate.flow_opts(flow, flow_wait=False, allow_new_or_none=False)
     yield
-    yield schd.pool.remove_tasks(tasks)
+    flow_nums = get_flow_nums_set(flow)
```
wxtim, line 333: Would it make sense to put this before the yield so any failure happens in the validate step?

cylc/flow/flow_mgr.py
```diff
@@ -55,36 +62,65 @@ def add_flow_opts(parser):
...
+
+
+def stringify_flow_nums(flow_nums: Iterable[int]) -> str:
+    """Return the canonical string for a set of flow numbers.
+
+    Examples:
+        >>> stringify_flow_nums({1})
+        '1'
+
+        >>> stringify_flow_nums({3, 1, 2})
```
wxtim, line 89: Consider a different iterable just to demonstrate on a different type.

```
stringify_flow_nums(range(1, 4, 2))
"1,3"
```

cylc/flow/commands.py
MetRonnie: The convention for generic types is to use a single letter like `T` or `S`, in this case I found out how to type this from some GitHub issue on python/typing or python/mypy. I seem to remember adding a link as a comment, but must have lost it

cylc/flow/flow_mgr.py
```diff
@@ -55,36 +62,65 @@ def add_flow_opts(parser):
     )
 
 
-def stringify_flow_nums(flow_nums: Set[int], full: bool = False) -> str:
-    """Return a string representation of a set of flow numbers
+def get_flow_nums_set(flow: List[str]) -> FlowNums:
```
wxtim, line 65: I wonder if there is sufficient complexity to justify making a class `Flows` to gather up some of this functionality?

cylc/flow/flow_mgr.py
```diff
@@ -55,36 +62,65 @@ def add_flow_opts(parser):
...
+
+        >>> stringify_flow_nums({})
+        ''
+
+    """
+    return ','.join(str(i) for i in sorted(flow_nums))
+
+
+def repr_flow_nums(flow_nums: FlowNums, full: bool = False) -> str:
+    """Return a representation of a set of flow numbers
```
wxtim, line 100: Document full in docstring?

MetRonnie: Sorry, what do you mean?

cylc/flow/commands.py
oliver-sanders: I wouldn't try moving that line.

E.G, Imagine there are two queued commands:
* trigger(foo, flow=new)
* trigger(bar, flow=all)

At present we would get:
* foo (flow=2)
* bar (flow=1,2)

If we moved this up a line, we would get:
* foo (flow=2)
* bar (flow=1)

cylc/flow/flow_mgr.py
oliver-sanders: IMO, a set is fine.

We just need two functions to serialise / deserialise the set for DB / presentation purposes.

cylc/flow/flow_mgr.py
wxtim: Can you explain what the `full` boolean is doing in the docstring?

cylc/flow/prerequisite.py
```diff
@@ -268,7 +273,10 @@ def satisfy_me(self, outputs: Iterable['Tokens']) -> 'Set[Tokens]':
             if prereq not in self._satisfied:
                 continue
             valid.add(output)
-            self[prereq] = 'satisfied naturally'
+            if self._satisfied[prereq] != 'satisfied naturally':
```
wxtim, line 276: This should be a straightforward merge with the Skip mode logic here, but we will need to make sure it's well tested to avoid mistakes.

cylc/flow/task_proxy.py
```diff
@@ -516,6 +512,17 @@ def name_match(
             match_func(ns, value) for ns in self.tdef.namespace_hierarchy
         )
 
+    def match_flows(self, flow_nums: 'FlowNums') -> 'FlowNums':
+        """Return which flows the task belongs to that match the given flows.
```
wxtim, line 516:
```suggestion
        """Which of the given flows does the task belong to?
```

cylc/flow/task_pool.py
```diff
@@ -2047,30 +2061,119 @@ def _get_active_flow_nums(self) -> Set[int]:
...
-        if self.compute_runahead():
+            LOG.info(f"Removed task(s): {', '.join(sorted(tasks_str_list))}")
+
+        not_removed.update(matched_task_ids.difference(removed))
+        if not_removed:
+            fnums_str = (
+                repr_flow_nums(flow_nums, full=True) if flow_nums else ''
+            )
+            LOG.warning(
+                "Task(s) not removable: "
```
wxtim, line 2171: I think that an explanation here might be helpful, perhaps:

> Task(s) does not exist in flow(s) so not removed:

wxtim (commented):
I've reviewed the functional code - Will take a bit of a break then have a look at the testing. 

Looking good so far.

MetRonnie:
> If removing a task causes all of the prerequisites on a downstream task to be unset (i.e. if the downstream task was spawned as a result of outputs from this task alone) then the downstream task shall be removed from the pool.

This is proving to be tricky, as there is a difference between how prerequisites are handled for these 2 cases:
```cylc
[scheduling]
    [[graph]]
       R1 = """
           (a1 | a2) & b => c
       """
```
and
```cylc
[scheduling]
    [[graph]]
       R1 = """
           (a1 | a2) => c
           b => c
       """
```
In the first case, there is only 1 prerequisite that includes both `a1 | a2` and `b`, whereas in the second there are 2 prerequisites - 1 for `a1 | a2` and 1 for `b`. In the first case, removing `b` would result in `c` being removed from the task pool even if `a1` or `a2` had succeeded, and I am not sure there is a way to tell this without another refactor of the `Prerequisite` class.

tests/functional/cylc-remove/00-simple/flow.cylc
```diff
@@ -1,24 +1,26 @@
 # Abort on stall timeout unless we remove unhandled failed and waiting task.
 [scheduler]
     [[events]]
-        stall timeout = PT20S
+        stall timeout = PT30S
```
wxtim, line 4: Why was this necessary? Was it flaky without?

tests/unit/test_workflow_db_mgr.py
```diff
@@ -0,0 +1,80 @@
...
+def test_remove_task_from_flows(
+    tmp_path: Path, flow_nums: FlowNums, expected_removed: FlowNums
+):
+    db_flows: List[FlowNums] = [
+        {1, 2},
+        {5},
+        set(),  # FLOW_NONE
+    ]
+    db_mgr = WorkflowDatabaseManager(tmp_path)
+    schd_tokens = Tokens('~asterix/gaul')
```
wxtim, line 52: > O tempora, o mores!

tests/unit/test_task_proxy.py
```diff
@@ -101,3 +103,18 @@ def test_status_match(status_str: Optional[str], expected: bool):
...
 
     assert TaskProxy.status_match(mock_itask, status_str) is expected
+
+
+@pytest.mark.parametrize('itask_flow_nums, flow_nums, expected', [
+    param({1, 2}, {2}, {2}, id="subset"),
+    param({2}, {1, 2}, {2}, id="superset"),
+    param({1, 2}, {3, 4}, set(), id="disjoint"),
+    param({1, 2}, set(), {1, 2}, id="all-matches-num"),
+    param(set(), {1, 2}, set(), id="num-doesn't-match-none"),
```
wxtim, line 113: The apostrophe will make running these tests individually from the shell a nightmare.

tests/unit/test_prerequisite.py
```diff
@@ -138,14 +143,79 @@ def test_get_target_points(prereq):
     }
 
 
-def test_get_resolved_dependencies():
+@pytest.fixture
+def prereq2():
```
wxtim, line 147: suggest `integer_prereq` for this one and `iso8601_prereq` for the other pre-req fixture.

Or shove them in a singe fixutre thus:

```python
@pytest.fixture
def prereq():
    def _inner(iso8601: boolean = False):
        if iso8601:
            # setup prereq 1
            return prereq
        else:
            # setup prereq 2
            return prereq
    return _inner
```

tests/unit/test_prerequisite.py
```diff
@@ -138,14 +143,79 @@ def test_get_target_points(prereq):
...
     prereq = Prerequisite(IntegerPoint('2'))
     prereq[('1', 'a', 'x')] = True
     prereq[('1', 'b', 'x')] = False
     prereq[('1', 'c', 'x')] = 'satisfied from database'
     prereq[('1', 'd', 'x')] = 'force satisfied'
-    assert prereq.get_resolved_dependencies() == [
+    return prereq
```
wxtim, line 153: @wxtim - you'll want to add the strings for skip satisfied dependencies to this list.

tests/integration/test_remove.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+@pytest.fixture
+def example_workflow(flow):
+    return flow({
+        'scheduling': {
+            'graph': {
+                'R1': '''
+                    a1 & a2 => b
+                    a3 => b
```
wxtim, line 36: Why not

```suggestion
                    a1 & a2 & a3 => b
```

tests/integration/test_remove.py
```diff
@@ -0,0 +1,340 @@
...
+            schd, True, 'task_prerequisites', 'satisfied', prereq_name='a1'
+        ) == [
+            ('satisfied naturally',),
+        ]
+        assert get_data_store_flow_nums(schd, a1) == '[1]'
+
+        await run_cmd(remove_tasks(schd, ['1/a1'], [FLOW_ALL]))
+        await schd.update_data_structure()
+
+        assert a1 not in schd.pool.get_tasks()  # removed from pool
```
wxtim, line 76: I think the comments would be easier to read the line above the assert statements.

wxtim (commented):
I've
- [x] Read  the code and tests
- [x] Tried to break it manually.

cylc/flow/flow_mgr.py
MetRonnie: Just because the function can accept any iterable doesn't mean generators are likely to be used

cylc/flow/task_pool.py
MetRonnie: This message will be logged if the user tries to remove a task that either:
- has no intersecting flow numbers with what was specified (as you've said)
- has not run yet
- has already been removed

Should we try and explain this, or stick to the simple "not removable"?

tests/functional/cylc-remove/00-simple/flow.cylc
MetRonnie: Probably on MacOS GH Actions runner

tests/functional/cylc-remove/02-cycling/flow.cylc
```diff
@@ -28,18 +28,6 @@
     [[foo, waz]]
```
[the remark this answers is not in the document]
MetRonnie: I didn't remove `01` in this PR. I'm not sure why the tests are numbered anyway

tests/integration/test_remove.py
MetRonnie: Comment added

tests/unit/test_prerequisite.py
MetRonnie: I think it makes sense to have this fixture here as it is visible while you are looking at the 2 tests that use it

cylc/flow/task_pool.py
wxtim: Honestly I'd prefer the longer explanation, but I agree, the alternative is a bit long. Can you not differentiate between the three causes?

MetRonnie: Nope

oliver-sanders:
Where are we up to with these TODO items?

> What's left to do
>
>    When removing an active task from all its flows, kill the task.
>
>    Should probably add a functional test with the --flow option.
>
>    Need to check this one:
>
>        If removing a task causes all of the prerequisites on a downstream task to be unset (i.e. if the downstream task was spawned as a result of outputs from this task alone) then the downstream task shall be removed from the pool.

MetRonnie:
None of them done yet. The last one is what I have been having a go at and came up against this problem which I also discussed on Teams: https://github.com/cylc/cylc-flow/pull/6370#issuecomment-2444348387

cylc/flow/task_pool.py
wxtim: That's fine. I don't believe this is a blocker.

hjoliver:
Getting in review mode here - sorry for the delay!

cylc/flow/prerequisite.py
```diff
@@ -334,7 +342,7 @@ def get_target_points(self):
             get_point(p) for p in self.iter_target_point_strings()
         ]
 
-    def get_resolved_dependencies(self) -> List[str]:
+    def get_satisfied_dependencies(self) -> List[str]:
```
hjoliver, line 345: [Vaguely interesting historical aside: I think use of "resolved" here goes all the way back to Cylc 2, when we first got the ability to visualize a graph - not from the workflow definition because there was no workflow definition, just a bunch of task definition files with inputs and outputs - from dependencies "resolved" at runtime.]

cylc/flow/scripts/remove.py
```diff
@@ -18,7 +18,26 @@
 
 """cylc remove [OPTIONS] ARGS
 
-Remove one or more task instances from a running workflow.
+Remove task instances from a running workflow and the workflow's history.
```
hjoliver, line 21: Just wondering how to clarify this for users (what are "task instances"?). How about:
```suggestion
Remove tasks from the active window (n=0) and from the workflow's history.
```

cylc/flow/scripts/remove.py
```diff
@@ -18,7 +18,26 @@
...
+
+This removes the task(s) from any specified flows. The task will still exist,
+just not in the specified flows, so will not influence the evolution of
+the workflow in those flows.
```
hjoliver, line 25:
```suggestion
This removes the task(s) from specified flows. The task will still exist,
but it will not influence the evolution of the workflow in those flows.
```

cylc/flow/scripts/remove.py
```diff
@@ -18,7 +18,26 @@
...
+
+If a task is removed from all flows, it and its outputs will be left in the
+`None` flow. This preserves a record that the task ran, but it will not
+influence any flows in any way.
```
hjoliver, line 29:
```suggestion
If a task is removed from all flows, it and its outputs will be left in the
`None` flow to preserve a record that the task ran.
```
The last bit (it will not influence ...) is already stated in the previous paragraph.

cylc/flow/scripts/remove.py
```diff
@@ -18,7 +18,26 @@
...
+
+Examples:
+  # remove a task which has already run
+  # (any tasks downstream of this task which have already run or are currently
+  # running will be left alone The task and its outputs will be left in the
+  # None flow)
```
hjoliver, line 35:
```suggestion
  # Remove a task that already ran.
  # (Any downstream tasks that are already running or finished will
  # be left alone. The task and its outputs will be left in the None flow.)
```

cylc/flow/scripts/remove.py
```diff
@@ -18,7 +18,26 @@
...
+  $ cylc remove <id>
+
+  # remove a task from a specified flow
+  # (the task may remain in other flows)
```
hjoliver, line 39:
```suggestion
  # Remove a task from a specified flow.
  # (The task may remain in other flows.)
```

cylc/flow/scripts/remove.py
```diff
@@ -61,6 +83,19 @@ def get_option_parser() -> COP:
         argdoc=[FULL_ID_MULTI_ARG_DOC],
     )
 
+    parser.add_option(
+        '--flow',
+        action='append',
+        dest='flow',
+        metavar='FLOW',
+        help=(
+            "Remove the task(s) from the specified flow number. "
```
hjoliver, line 92:
```suggestion
            "Remove the task(s) from the specified flow. "
```

cylc/flow/scripts/remove.py
```diff
@@ -61,6 +83,19 @@ def get_option_parser() -> COP:
...
+            "Reuse the option to remove the task(s) from multiple flows. "
+            "If the option is not used at all, the task(s) will be removed "
+            "from all flows."
```
hjoliver, line 95:
```suggestion
            "By default, the task(s) will be removed "
            "from all flows."
```
(We use "by default" for this all over the place).

cylc/flow/task_proxy.py
```diff
@@ -522,6 +518,17 @@ def name_match(
             match_func(ns, value) for ns in self.tdef.namespace_hierarchy
         )
 
+    def match_flows(self, flow_nums: 'FlowNums') -> 'FlowNums':
+        """Return which of the given flow numbers the task belongs to.
+
+        NOTE: If `flow_nums` is empty, it means 'all', whereas
+        if `self.flow_nums` is empty, it means this task is in the 'none' flow
```
hjoliver, line 525: OK nice way to handle that.

cylc/flow/task_state.py
```diff
@@ -527,3 +528,12 @@ def get_unsatisfied_prerequisites(self) -> List['PrereqMessage']:
             for prereq in self.prerequisites if not prereq.is_satisfied()
             for key, satisfied in prereq.items() if not satisfied
         ]
+
+    def any_satisfied_prerequisite_tasks(self) -> bool:
+        """Return True if any of this task's prerequisite tasks are
+        satisfied."""
```
hjoliver, line 534:
```suggestion
        """Return True if any of this task's prerequisites are satisfied."""
```

cylc/flow/task_pool.py
```diff
@@ -1717,12 +1736,10 @@ def spawn_task(
...
             # If itask has any history in this flow but no completed outputs
-            # we can infer it was deliberately removed, so don't respawn it.
+            # we can infer it has just been deliberately removed (N.B. not
+            # by `cylc remove`), so don't immediately respawn it.
             # TODO (follow-up work):
             # - this logic fails if task removed after some outputs completed
-            # - this is does not conform to future "cylc remove" flow-erasure
-            #   behaviour which would result in respawning of the removed task
-            # See github.com/cylc/cylc-flow/pull/6186/#discussion_r1669727292
             LOG.debug(f"Not respawning {point}/{name} - task was removed")
```
hjoliver, line 1743: I think I wrote this comment myself. Presumably the only non-cylc-remove case is suicide triggers. Can we adapt the new machinery on this branch to handle that properly too?

hjoliver (commented):
I've been through almost all of the non-test code, looks good.  I still need to read the task_pool changes more carefully and do some testing.

hjoliver:
@MetRonnie - is this still in draft because of the two "what's left to do" items at the top?

cylc/flow/task_state.py
MetRonnie: Doesn't fit

MetRonnie:
Superseded by #6472

cylc/flow/task_state.py
hjoliver: (my main point was the wording)

cylc/flow/scripts/remove.py
hjoliver: Now outdated.

cylc/flow/scripts/remove.py
hjoliver: Now outdated.

cylc/flow/scripts/remove.py
hjoliver: Now outdated.

cylc/flow/task_pool.py
hjoliver: (I'll transfer this to the superseding PR...)

PR #6376, opened 2024-09-18
subprocpool: populate missing 255 callback arguments

oliver-sanders:
In some situations, these missing arguments could cause tasks to submit fail as the result of a host outage rather than moving onto the next host for the platform.

The bug is as old as the 255 callback - https://github.com/cylc/cylc-flow/pull/4184/files#diff-577155476a8501cbb774dad9dfc50a2ed7b846d577835e58d9bb4dfdb5d56b34R243-R244

**Check List**

### DIFF
--- a/changes.d/6376.fix.md
+++ b/changes.d/6376.fix.md
@@ -1,0 +1,1 @@
+Fixes an issue that could cause Cylc to ignore the remaining hosts in a platform in response to an `ssh` error in some niche circumstances.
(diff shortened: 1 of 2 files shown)


oliver-sanders:
Testing these exact lines is rather awkward, however, they are covered.

wxtim (approved):
Makes sense and fixes a breakage that I introduced all the way back in intelligent host selection (oops). I'd have been happier with it explicitly causing a test failure, but this isn't an easy one to simulate.

Do you want a second review given the lack of tests?

hjoliver (approved):

PR #6390, opened 2024-10-02
tui: fix task state filters

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/5716
* Families cannot be filtered out in the GraphQL query because we can only filter by state (which is an aggreagate state for families), not state totals (i.e. all task states contained within the family).
* This fixes the approach by filtering out empty familes and cycles client-side.

Test this with any odd workflow that contains families. Before this change task state filtering kinda worked, however, sometimes tasks were erroneously filtered out (due to the family state not matching the task state) and some empty families/cycles were left in erroneously.

**Check List**


wxtim (approved):

oliver-sanders:
Mark spotted two issues:

1. Traceback when running the `complex` workflow (solved).
2. Succeeded jobs occasionally appearing when filtering for waiting tasks (replicated but not solved). This seems to be a momentary flicker that quickly vanishes.

oliver-sanders:
I can't get a handle on issue (2). I have only very occasionally seen filtering go wrong and always a quick flash that quickly corrects itself.

Less broken is an improvement.

wxtim:
> Less broken is an improvement.

I'd be happy to accept this PR on that basis? @markgrahamdawson ?

markgrahamdawson (approved):

Issue #6361, opened 2024-09-06
Errant flow merge when triggering


MetRonnie:
### Description

```
R1 = """
    foo
    bar
"""
```

If `foo` fails in flow 1 while `bar` is in flow 2, triggering `foo` without providing a flow in the options causes it to run in flows 1 & 2.

This is because the default `--flow all` is causing it to pick all active flow numbers in the entire task pool instead of all active flow numbers for the task in question.

https://github.com/cylc/cylc-flow/blob/293c3ccba1877adc1061064b410ea54b62beb4ad/cylc/flow/task_pool.py#L2077-L2078

https://github.com/cylc/cylc-flow/blob/293c3ccba1877adc1061064b410ea54b62beb4ad/cylc/flow/task_pool.py#L2025-L2037

Looking at the code, this bug probably affects `cylc set` too

oliver-sanders:
> If foo fails in flow 1 while bar is in flow 2, triggering foo without providing a flow in the options causes it to run in flows 1 & 2.

Hmm, there's nothing untoward here, this is the intended behaviour, all triggers default to `--flow=all`.

`--flow=all` means "all flows in the workflow" *not* "all flows on the task".

MetRonnie:
But maybe the default should not be "all flows in the workflow", but "all flows on the task"?

MetRonnie:
Right so this is a bug then. The default should be "all flows on the task", which is meaningless if it hasn't spawned yet, in which case it should be "all flows in the workflow"?

oliver-sanders:
> What about setting/triggering a task that already ran and is no longer in n=0? Should it not use the flow number(s) from last time it ran?

I don't think there's any special logic for n<0.

hjoliver:
Just to confirm, the plan was:
- if triggering an n=0 (pool) task the default should be that it just keeps its already-assigned flow numbers

@oliver-sanders - it seems like you agree but initially did not realize that @MetRonnie was specifically talking about triggering an n=0 task?

oliver-sanders:
Closed by #6367

PR #6395, opened 2024-10-06
Group trigger

hjoliver:
[UPDATED and refactored post ~informal~ agreement on [the proposal](https://github.com/cylc/cylc-admin/pull/197)]

-------

Major change to manual trigger behaviour: instead of running all target tasks immediately despite unsatisfied prerequisites, **trigger them as a **group** that respects in-group dependencies and satisfies off-group prerequisites that would cause a stall.**

This is an easier way to (re)run a sub-graph when all the tasks can easily be identified - no need to manually identify sub-graph start tasks to trigger and/or set off-flow prerequisites.

(Amazingly, this should not be a breaking change for any sensible `cylc trigger` use cases).

High level implementation (see the [the proposal](https://github.com/cylc/cylc-admin/pull/197) for details):
 1. Examine the group to identify all off-group prerequisites
 2. Remove the (original or specified) flow from group tasks, to allow rerun in the same flow 
 3. Satisfy the off-group prerequisites, to trigger group start tasks and prevent later stalls
 4. (The resulting flow then takes care of itself to propagate through the group)
 
-------

Example from the proposal, tweaked to run automatically via a stall handler:
```cylc
# Task b fails due to an upstream problem not captured by the workflow
# so we need to rerun the graph from task a, to complete the workflow.
[scheduler]
    [[events]]
        expected task failures = 1/b
        stall handlers = \
            "cylc trigger %(workflow)s \
             //1/a //1/f_m1 //1/f_m2 //1/f_m3 //1/g_m1 //1/g_m2 //1/g_m3 //1/b"
[scheduling]
    [[graph]]
        R1 = """
            start => a
            x => f_m1
            a => f_m1 => g_m1 => b
            a => f_m2 => g_m2 => b
            a => f_m3 => g_m3 => b
            b => end
            g_m3 => y
       """
[runtime]
    [[b]]
        script = [[ "${CYLC_TASK_SUBMIT_NUMBER}" != '1' ]]
    [[FAMILY]]  # (anticipates inactive tasks globbing to specify a trigger group)
    [[a, f_m1, f_m2, f_m3, g_m1, g_m2, g_m3, b]]
        inherit = FAMILY
    [[start, x, end, y]]
```
--------


hjoliver:
The test battery passes, so this does not break existing `trigger` and `set` functionality :tada: I just need to get some more coverage of the new code.

hjoliver:
(Updated after major changes on master)

hjoliver:
Note on changes to reference logs: group trigger causes triggering by satisfying prerequisites, rather than by direct triggering despite unsatisfied prerequisites ... so triggering now gets logged as (e.g.) `1/b triggered off ['1/a']` instead of `1/b triggered off [ ]`.

hjoliver:
(All tests were passing until my latest change to incorporate the new `cylc remove` stuff ... TBD)

hjoliver:
(Updated from master and deconflicted)

hjoliver:
(rebased and deconflicted)

hjoliver:
(Now that the proposal has been reviewed, I've converted this back to Draft so we can check on proposal compliance once all the details are agreed).

hjoliver:
(Proposal example works, test battery all good - I should be able to unDraft this early next week).

cylc/flow/scheduler.py
```diff
@@ -1236,6 +1238,141 @@ def remove_tasks(
...
+                        f"[{itask}] ignoring 'flow=none' trigger: task already"
+                        f" has {repr_flow_nums(itask.flow_nums, full=True)}"
+                    )
+                    continue
+
+                if itask.state(
+                    TASK_STATUS_PREPARING,
+                    *TASK_STATUSES_ACTIVE
+                ):
+                    LOG.error(f"[{itask}] ignoring trigger - already active")
```
hjoliver, line 1308: TODO: still need to merge the flow on these guys

hjoliver:
(Updated from master and deconflicted (ouch) to pick up #6691 )

oliver-sanders:
Happy that only the group start-tasks will run. Don't see any real need to work around this.

hjoliver:
Integrated xtrigger prerequisite setting. Tested with this:
```cylc
[scheduling]
    [[xtriggers]]
        xoff = xrandom(0)  # never
    [[graph]]
        R1 = """
            failer? => GROUP => cleanup
            failer:fail? => triggerer
            off => c  # off-group task prerequisite
            @xoff => b  # off-group xtrigger prerequisite
             a => b & c  # GROUP internal triggers
        """
[runtime]
    [[failer, off]]
        script = false
    [[GROUP]]
    [[a, b, c]]
        inherit = GROUP
    [[triggerer]]
        script = "cylc trigger $CYLC_WORKFLOW_ID //1/a //1/b //1/c"
    [[cleanup]]
        script = "cylc remove $CYLC_WORKFLOW_ID//1/*:failed"
```

hjoliver:
TODO:
- [x] code tidied and TODOs addressed
- [ ] test for xtrigger prequisite satisfaction in a group trigger operation
- [ ] replace `tests/f/cylc-trigger/08-group-trigger` with integration tests

hjoliver:
Done.

cylc/flow/scheduler.py
```diff
@@ -1230,11 +1231,150 @@ def remove_tasks(
...
```
oliver-sanders: We are likely to hit this message a lot with group trigger. Maybe demote to warning (?) and batch up the messages rather than issuing one per task.

cylc/flow/scheduler.py
```diff
@@ -1230,11 +1231,150 @@ def remove_tasks(
...
+
+        """
+        active, inactive, _ = self.pool.filter_task_proxies(
+            items, inactive=True, warn_no_active=False,
+        )
+
+        group_ids = (
+            [(tdef.name, str(point)) for (tdef, point) in inactive] +
+            [(itask.tdef.name, str(itask.point)) for itask in active]
+        )
```
oliver-sanders: We do a lot of `entry in group_ids` comparisons below. Convert to a set for O(1) performance:

```suggestion
        group_ids = {
            *((tdef.name, str(point)) for (tdef, point) in inactive),
            *((itask.tdef.name, str(itask.point)) for itask in active),
        }
```

cylc/flow/scheduler.py
```diff
@@ -1230,11 +1231,150 @@ def remove_tasks(
...
+        stuff = inactive
+        if active_tasks_to_remove:
+            stuff.update(
+                {
+                    (itask.tdef, itask.point)
+                    for itask in active_tasks_to_remove
+                }
+            )
+
+        for tdef, point in stuff:
```
oliver-sanders:
```suggestion
        for tdef, point in {*inactive, *active_tasks_to_remove):
```

cylc/flow/scheduler.py
```diff
@@ -1230,11 +1231,150 @@ def remove_tasks(
...
+                    if (key.task, str(key.point)) not in group_ids
+                }
+                in_flow_prereqs = any(
+                    {
+                        key
+                        for pre in tdef.get_prereqs(point)
+                        for key in pre.keys()
+                        if (key.task, str(key.point)) in group_ids
+                    }
+                )
```
oliver-sanders: Write as a generator, otherwise, the entire set will be built before the `any` is allowed to return (even if the first item is `True`).

```suggestion
                in_flow_prereqs = any(
                    key
                    for pre in tdef.get_prereqs(point)
                    for key in pre.keys()
                    if (key.task, str(key.point)) in group_ids
                )
```

cylc/flow/task_pool.py
```diff
@@ -1798,18 +1801,19 @@ def spawn_task(
...
             and not itask.state.outputs.get_completed_outputs()
         ):
             # If itask has any history in this flow but no completed outputs
             # we can infer it has just been deliberately removed (N.B. not
             # by `cylc remove`), so don't immediately respawn it.
             # TODO (follow-up work):
             # - this logic fails if task removed after some outputs completed
-            LOG.debug(f"Not respawning {point}/{name} - task was removed")
+            LOG.info(f"Not respawning {point}/{name} - task was removed")
             return None
```
oliver-sanders: I think this is leftover from before the remove proposal? Remove now strips the specified flow numbers so I don't think this is valid any more.

cylc/flow/task_pool.py
```diff
@@ -2041,64 +2044,67 @@ def set_prereqs_and_outputs(
...
-                # Outputs (may be empty list)
-                trans = self._get_task_proxy_db_outputs(
-                    point, tdef, flow_nums,
-                    flow_wait=flow_wait, transient=True
-                )
-                if trans is not None:
-                    self._set_outputs_itask(trans, outputs)
+        # For inactive tasks, default to all current flows.
+        if inactive_tasks:
+            flow_nums = self.get_flow_nums(flow or [FLOW_ALL], flow_descr)
```
oliver-sanders: Isn't the `FLOW_ALL` default handled upstream of this code?

```suggestion
            flow_nums = self.get_flow_nums(flow, flow_descr)
```

cylc/flow/task_proxy.py
```diff
@@ -334,6 +334,15 @@ def __str__(self) -> str:
             f"{id_}{repr_flow_nums(self.flow_nums)}:{self.state}"
         )
 
+    def __eq__(self, other) -> bool:
+        """Task proxy equality is based point/name only."""
+        # (Needed e.g. for adding to task_pool.tasks_to_trigger sets)
+        return self.identity == other.identity
```
oliver-sanders: `self.transient`, `self.reload_successor`?

cylc/flow/task_proxy.py
```diff
@@ -334,6 +334,15 @@ def __str__(self) -> str:
...
+
+    def __hash__(self):
+        """Task proxy equality is based point/name only."""
+        return hash(self.identity)
```
oliver-sanders: What about reload-successors or transient task proxy objects?

I think these `__eq__` and`__hash__` methods might be a bit dubious, however, it looks like the code works fine without them, so maybe we can just drop them? Note, Python classes get these methods by default, objects only compare equal to themselves which is all we need and wouldn't cause issues in edge cases.

tests/unit/scripts/test_cylc.py
```diff
@@ -133,6 +134,11 @@ def test_pythonpath_manip(monkeypatch):
 
     and adds items from CYLC_PYTHONPATH
     """
+
+    # CYLC_PYTHONPATH needs to be unset for this test.
+    with suppress(KeyError):
+        del os.environ["CYLC_PYTHONPATH"]
```
oliver-sanders: Use `monkeypatch.delenv`

cylc/flow/scheduler.py
```diff
@@ -1230,11 +1231,150 @@ def remove_tasks(
...
             self.pool.release_runahead_tasks()
 
+    def force_trigger_tasks(
+        self,
+        items: Iterable[str],
+        flow: List[str],
+        flow_wait: bool = False,
+        flow_descr: Optional[str] = None,
+        on_resume: bool = False
+    ):
```
oliver-sanders: Co-locate this with the other commands in `cylc.flow.commands` (`s/self/schd/`)

cylc/flow/scripts/trigger.py
```diff
@@ -17,43 +17,51 @@
...
+See `cylc set` for the lower-level flow-based approach to rerunning sub-graphs.
 
 Examples:
-  # trigger task foo in cycle 1234 in test
-  $ cylc trigger test//1234/foo
+  # trigger task foo in cycle 1, in workflow "test"
+  $ cylc trigger test//1/foo
+
+  # trigger all failed tasks in workfow "test"
+  $ cylc trigger 'test//*:failed'  # (quotes required)
```
wxtim, line 57: IMO this is important distinction.

```suggestion
  $ cylc trigger 'test//*:failed'  # (single-quotes required)
```

tests/integration/test_force_trigger.py
```diff
@@ -129,23 +135,136 @@ async def test_trigger_on_resume(
...
+):
+    """Test trigger of a sub-graph of future, then past, tasks.
+
+    It should satisfy off-group task and xtrigger prerequisites automatically.
+
+    """
+    cfg = {
+        'scheduling': {
+            'xtriggers': {
+                'xr': 'xrandom(0)'  # never satisfied
```
wxtim, line 215: Closer to self documenting?
```suggestion
                'xtrigger': 'echo("No", False)'
```

tests/integration/test_force_trigger.py
```diff
@@ -129,23 +135,136 @@ async def test_trigger_on_resume(
...
+            },
+            'graph': {
+                'R1': """
+                    # sub-graph for group trigger: a => b & c => d
+                    x => a => b & c => d => e => y
```
wxtim:
```suggestion
                    x => a
                    # Sub-graph for group trigger
                    a => b & c => d
                    # Later tasks
                    d => e => y
```

tests/integration/test_force_trigger.py
```diff
@@ -129,23 +135,136 @@ async def test_trigger_on_resume(
...
+
+                    # off-group prerequisites:
+                    @xr => x  # stop the flow starting
+                    @xr => c  # off-group xtrigger prerequisite
+                    @xr => off => b  # off-group task prerequisite
```
wxtim, line 234: Why does `off` need a dependence on the xtrigger? How is this different from x?

oliver-sanders: I think the point here is:

* In-group xtriggers should be satisfied by the trigger (just as they would have been if we had triggered the tasks one-by-one).
* However, this should not affect off-group tasks (i.e. satisfying `@xr` for `x` should not also satisfy it for the off-group task `x`).

cylc/flow/scripts/trigger.py
oliver-sanders: Not true:

```console
$ echo "*"  # double quotes
*
$ echo '*'  # single quotess
*
$ echo *  # no quotes
changes.d CHANGES.md conda-environment.yml ...
```

We just use single quotes here since this is the most sensible option when you aren't intending on shell `$substitution`.

hjoliver:
Thanks for starting on this @oliver-sanders  - once again, I'll get back on it ASAP. I'm still stuck in local project admin stuff at the FY-end boundary, plus TAG meeting on Thursday ... but I should be done with all that soon...

hjoliver:
(bug confirmed with a real workflow ... I initially though not, but had borked my environment)

hjoliver:
OK fixed - the bug was a small regression (I had definitely tested this kind of scenario). I had accidentally swapped out int flow_nums for the str version in one place, and a single missed type annotation let it through 😬 

```diff
diff --git a/cylc/flow/scheduler.py b/cylc/flow/scheduler.py
index 37d0aa2e1..4fdeb7b02 100644
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1121,7 +1121,7 @@ class Scheduler:
         if active or inactive:
             self.remove_matched_tasks(active, inactive, flow_nums)

-    def remove_matched_tasks(self, active, inactive, flow_nums):
+    def remove_matched_tasks(self, active, inactive, flow_nums: 'FlowNums'):
         # Mapping of *relative* task IDs to removed flow numbers:
         removed: Dict[Tokens, FlowNums] = {}
         not_removed: Set[Tokens] = set()
@@ -1317,7 +1317,7 @@ class Scheduler:
                 active_tasks_to_remove.append(itask)

         # remove all inactive and selected active group members
-        self.remove_matched_tasks(active_tasks_to_remove, inactive, flow)
+        self.remove_matched_tasks(active_tasks_to_remove, inactive, flow_nums)

         # store removal results before moving on
         self.workflow_db_mgr.process_queued_ops()
```

cylc/flow/scheduler.py
hjoliver: Done.

cylc/flow/scheduler.py
hjoliver: Doesn't work because `active_tasks_to_remove` holds task proxies, not `(tdef, point)`.

(Changed the silly variable name though).

cylc/flow/scheduler.py
hjoliver: Done.

cylc/flow/task_pool.py
hjoliver: Requires some thought, and not really in scope here.

 I think this code avoids automatic respawning of tasks removed by a suicide trigger:

https://github.com/cylc/cylc-flow/pull/6186#discussion_r1669727292
https://github.com/cylc/cylc-flow/pull/6186/files#r1741147438
https://github.com/cylc/cylc-flow/issues/6066

cylc/flow/task_pool.py
hjoliver: At this point, no - I just checked. I might be worth going over flow-num handling and defaulting generally in case it could be streamlined, but maybe not on this PR.

cylc/flow/task_proxy.py
hjoliver: Ah, not sure what your point is?

hjoliver: Ok, you've explained in the next comment below.

cylc/flow/task_proxy.py
hjoliver: I'm reasonably sure I would have added this code for a good reason - but I'll have to try to remind myself, too long ago...

cylc/flow/scheduler.py
hjoliver: It is!  It's just under `def remove_tasks` - although then I split a block of common code out of that method, but left it in-place to make for a minimal code change - is it the positioning of that that you're complaining about?.

tests/integration/test_force_trigger.py
hjoliver: I kinda forgot about the toy `echo` xtrigger, but my comment is fine for self-doc isn't it?  The important thing is, it's an xtrigger that will not be naturally satisfied.

hjoliver:
> Started working on a test to cover --flow interaction but hit the above problem. Here's the test template I was working on:

I added your test @oliver-sanders - it passes.

cylc/flow/scheduler.py
oliver-sanders: > It is! 

???

This code has been yanked out into the Scheduler class for some reason, it should be co-located with the other commands in `cylc.flow.commands`.

tests/integration/test_force_trigger.py
wxtim: Yes - whatever - shouldn't block review.

cylc/flow/task_proxy.py
oliver-sanders: The code seems to work fine with them removed, all unit/integration tests pass.

I can't think of cases where this sort of comparison would be useful. By default an object is only equal to itself which seems like a sensible test for us where `TaskProxy` objects are intended to be unique (besides proxy reload successors and transient objects).

```python
class Foo():

    def __init__(self, x):
        self.x = x


class Bar(Foo):

    def __eq__(self, other):
        return self.x == other.x

    def __hash__(self):
        return hash(self.x)

assert Foo(1) != Foo(1)  # is not the same object
assert Bar(1) == Bar(1)  # obj.x is the same
```

tests/integration/test_force_trigger.py
oliver-sanders: I think `# never satisfied` is clear enough. Call the xtrigger `@fail` for serious clarity!

cylc/flow/scheduler.py
hjoliver: Oh, I see what you mean.  

It was yanked out of the task_pool module, not commands - just as @MetRonnie did to the remove command.

Both trigger and remove have a small wrapper in commands.  I think I was just following Ronnie's example, since the two commands have relevant similarities.

tests/integration/test_force_trigger.py
hjoliver: I considered that, but technically xtriggers don't fail, they're either "succeeded" or "not succeeded yet". I suppose `@never` might be the serious-clarity option!

oliver-sanders:
More tests: https://github.com/hjoliver/cylc-flow/pull/67

cylc/flow/scheduler.py
MetRonnie: I moved the `remove_tasks` method from `task_pool.py` to `scheduler.py` as it needed to call the scheduler's `kill_tasks` method. It looks like you've done a similar thing and need to call the scheduler's `remove_matched_tasks` method

cylc/flow/scheduler.py
```diff
@@ -1330,7 +1330,10 @@ def force_trigger_tasks(
             )
 
         # remove all inactive and selected active group members
-        self.remove_matched_tasks(active_tasks_to_remove, inactive, flow_nums)
+        if flow != ['none'] and flow != ['all']:
+            # TODO: is the enclosing 'if' statement correct here?
```
oliver-sanders: TODO

hjoliver:
I showed the one-liner above: https://github.com/cylc/cylc-flow/pull/6395#issuecomment-2982284844 

(Actually it was a two-liner, but one line was just a missing type annotation).

cylc/flow/scheduler.py
hjoliver: Yes, I just put that there minutes ago. It fixes the problem, but I want to come back tomorrow and make really sure that it's the right thing to do there.

cylc/flow/scheduler.py
hjoliver: Correct!

cylc/flow/scheduler.py
hjoliver: (The one-liner you asked about changed `flow` (`[str]`) to `flow_nums` (`[int]`) in line 1333 above; I understand why it broke the tests now, despite definitely being a bug fix).

hjoliver:
All test passing now, including the ones just merged from you @oliver-sanders 🎉   Nearly there...

hjoliver:
Regarding the new TODO (above) - the DB remove code seems to behave slightly differently to what I expected for the trigger command ... I'll post the question on Element to avoid cluttering the PR more.

hjoliver:
Agreed - out of scope. I just raised it here because I was surprised (my bad) that remove had different default-flow semantics than trigger, when I ran into it on this branch, and thought I'd better confirm that no one expected remove-via-trigger to behave the same as standalone remove in that respect.

cylc/flow/scheduler.py
oliver-sanders: Er, what? That doesn't make sense!

The functions in the commands module are provided with the scheduler instance as the first argument (which is how you called the scheduler method!) so you did not need to move the code out of the commands module in either case. Just `s/self/schd/` and it'll work fine.

We could just move all the commands back into the Scheduler class of course. The intention of the commands module was to encapsulate the interventions code so we can reduce the scope of the bloated scheduler class (IMO, ideally this would reduce to the application control flow and event loop).

cylc/flow/task_pool.py
```diff
@@ -1764,6 +1766,7 @@ def spawn_task(
         point: 'PointBase',
         flow_nums: Set[int],
         flow_wait: bool = False,
+        force: bool = False
```
oliver-sanders: What does `force` do?

A comment or some docstring would be good.

cylc/flow/scheduler.py
MetRonnie: No code was ever moved out of `commands.py`; only out of `task_pool.py`. Yes in principle we could move it into `commands.py`.

hjoliver: What he said. 

Like for remove, trigger was moved out of task_pool to scheduler. It was never in commands. @MetRonnie's change for remove passed code review, so I just followed suit when trigger ran into s similar need for scheduler methods.

In both cases, there was just a small wrapper in commands that called the task pool (now scheduler) methods - and `git blame` fingers you @oliver-sanders for the wrappers - so presumably you made the decision to do it that way?? 

I thought maybe we wanted to keep the commands module brief, just easily readable command interfaces, with the guts out in the task pool and scheduler modules.
If not, happy to move the lot into `commands.py` ...

cylc/flow/scheduler.py
```diff
@@ -1329,8 +1330,23 @@ def force_trigger_tasks(
...
+                self.remove_matched_tasks(
+                    active_tasks_to_remove, empty_i, flow_nums)
+            if inactive:
+                empty_a: List[TaskProxy] = []
+                if not flow_nums:
+                    # remove for all active flows, not all flows.
+                    self.remove_matched_tasks(
+                        empty_a, inactive, self.pool.get_flow_nums([FLOW_ALL]))
+                else:
+                    self.remove_matched_tasks(empty_a, inactive, flow_nums)
```
oliver-sanders: I thought this was right before, I'm not sure I understand this change?

Before, if `--flow=all`, then `flow_nums` is already set to all active flows:

```python
flow_nums = self.pool.get_flow_nums(flow, flow_descr)
```

But now, `flow_nums` is still all active flows, but we remove active tasks (from `flow_nums`), then we recompute the `flow_nums` for the removal:

```python
self.pool.get_flow_nums([FLOW_ALL])
```

Which may now differ from `flow_nums` because we haven't spawned the new task(s) yet?


With this diff, Cylc will drop into a breakpoint any time these two values are not equal:

```diff
diff --git a/cylc/flow/scheduler.py b/cylc/flow/scheduler.py
index fbadbaaef..44e9c17d3 100644
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1347,6 +1347,11 @@ class Scheduler:
                         empty_a, inactive, self.pool.get_flow_nums([FLOW_ALL]))
                 else:
                     self.remove_matched_tasks(empty_a, inactive, flow_nums)
+                try:
+                    assert self.pool.get_flow_nums([FLOW_ALL]) == flow_nums
+                except:
+                    breakpoint()
+        # self.remove_matched_tasks(active_tasks_to_remove, inactive, flow_nums)
 
         # store removal results before moving on
         self.workflow_db_mgr.process_queued_ops()
```

oliver-sanders: To flesh out the edge case, I had a go at extending a test to cover triggering tasks with the `--flow=all` default WHEN the tasks previously ran in a flow that is no longer active:

```diff
--- a/tests/integration/test_force_trigger.py
+++ b/tests/integration/test_force_trigger.py
@@ -17,20 +17,24 @@
 import logging
 from typing import (
     Any as Fixture,
-    Callable
+    Callable,
 )
 
 import pytest
 
+from cylc.flow import LOG
 from cylc.flow.commands import (
     force_trigger_tasks,
-    reload_workflow,
     hold,
+    reload_workflow,
     resume,
     run_cmd,
     set_prereqs_and_outputs,
+    stop,
 )
 from cylc.flow.cycling.integer import IntegerPoint
+from cylc.flow.flow_mgr import FLOW_ALL
+from cylc.flow.workflow_status import StopMode
 
 
 async def test_trigger_workflow_paused(
@@ -485,3 +489,41 @@ async def test_trigger_group_in_flow(
             ('b', '[2]'),
             ('c', '[2]'),
         }
+
+        # breakpoint()
+        # assert schd.pool._get_active_flow_nums() == {'2'}
+        await run_cmd(stop(schd, mode=StopMode.REQUEST_CLEAN, flow_num=1))
+
+        assert schd.pool._get_active_flow_nums() == {2}
+
+        assert [
+            (itask.identity, itask.flow_nums)
+            for itask in schd.pool.get_tasks()
+        ] == [('1/d', {2})]
+
+        LOG.critical('AAAA')  # look for this in the log pytest shows if the test fails
+        await run_cmd(force_trigger_tasks(schd, ['1/a', '1/b', '1/c', '1/d'], [FLOW_ALL]))
+
+        assert set(db_select(
+            schd,
+            True,
+            'task_outputs',
+            'name',
+            'flow_nums',
+        )) == {
+            # original run
+            ('a', '[1]'),
+            ('b', '[1]'),  # flow "2" has been removed
+            ('c', '[1]'),  # flow "2" has been removed
+            ('d', '[1]'),
+
+            # subsequent run
+            ('a', '[]'),
+            ('b', '[]'),
+            ('c', '[]'),
+
+            ('a', '[2]'),
+            ('b', '[2]'),
+            ('c', '[2]'),
+            ('d', '[2]'),
+        }
```

* Use `cylc stop --flow` to stop flow `1`. The only active flow is now `2`.
* Trigger a group of tasks which ran in flows `[1]`, `[1, 2]` and `[2]` to see how the operation handles each.
* Cylc logs that it is triggering/removing the appropriate tasks (in two batches, one for active and one for inactive tasks) - makes sense.
* But strangely, the `flow=2` entries remain in the DB afterwards?

hjoliver: > I thought this was right before, I'm not sure I understand this change? 

I'll double check this soon, but I think what you've missed is flow_nums can be *empty*.  

For trigger that means "all active" flows, but for remove that means "all" flows. 

That's why I'm recomputing them with an explict FLOW_ALL (which means "all active") on calling the remove method.

Relatedly, the `get_flow_nums` method is designed to return flow numbers for *active tasks*, hence the empty set for default case (stick with the existing flow assignment for active tasks).

hjoliver: (I've spent all on day #6813  and #6821 - back to this tomorrow).

oliver-sanders: > but I think what you've missed is flow_nums can be empty.

If that's the case, then the earlier evaluation of `flow_nums = self.pool.get_flow_nums(flow, flow_descr)` at line 1263 would suffer from the same problem:

https://github.com/cylc/cylc-flow/pull/6395/files#diff-90711b005fe27b9f894e6cd09f08816d847dcf853518894c6b7e2139784cbcb8R1263

Either way, I don't think we should be recomputing the flow numbers post-remove?

cylc/flow/scheduler.py
oliver-sanders: > In both cases, there was just a small wrapper in commands that called the task pool (now scheduler) methods - and git blame fingers you @oliver-sanders for the wrappers - so presumably you made the decision to do it that way??

At the time, remove and trigger were implemented in the task pool, so I left them there. This arrangement dates back a long way (Cylc 6 or earlier) and makes reasonable sense since these are operations on the task pool itself.

Now these implementations have been pulled out into the scheduler class which is mostly about application control flow and event loop. The scheduler class is pretty unwieldy, so I've made some small efforts to move logic out of it into more self-contained modules, e.g, the main-loop plugins allowed us to pull out the health check, auto-restart and bad hosts logic; and the commands module allowed us to pull out the command implementations. Little by little...

> If not, happy to move the lot into commands.py ...

Er, yes please.

hjoliver:
Somewhat out-of-scope IMO (given the approved history of the similar `remove` function) but the last rather chunky commit moves trigger and remove out of scheduler and into commands as requested, which had some annoying implications in the tests.

cylc/flow/scheduler.py
hjoliver: (This comment is out of date, the replacement code block is still under review).

cylc/flow/scheduler.py
hjoliver: Done.

cylc/flow/scheduler.py
hjoliver: OK I finally got back to this. 

Sorry, my bad - I missed your main point because I didn't notice that I'd stupidly moved the `get_flow_nums` call into the remove call arg list to avoid a temporary variable ... by jumping over the earlier remove call that could change the task pool 🤦 

I'll try to clean up flow-nums handling a bit. It's confusing as hell at this point. Earlier review comment adds weight to this: https://github.com/cylc/cylc-flow/pull/6395#discussion_r2108973818

hjoliver: I've tried to clean up flow-nums handling, but not pushed the result up yet as I need to adjust a bunch of tests first. I'll do that tomorrow...

cylc/flow/task_pool.py
hjoliver: See #6813 and #6835

hjoliver:
Converted briefly back to draft till I've finished the final tweaks (in progress).

cylc/flow/scheduler.py
hjoliver: I'm resolving this discussion, after significantly refactoring flow-number handling to allow (among other things) simplifying the remove-via-trigger bit. (We can discuss the new code if you want to query anything there).

cylc/flow/task_pool.py
hjoliver: Removed. It was an attempt at handling the long-standing TODO problem that we eventually spun off to https://github.com/cylc/cylc-flow/issues/6813 (and fixed by https://github.com/cylc/cylc-flow/pull/6835)

cylc/flow/task_proxy.py
hjoliver:
``` 
        # (Needed e.g. for adding to task_pool.tasks_to_trigger sets)
```
This comment suggests to me that I must have ran into trouble with multiple instances of the same task proxy being added to the "tasks_to_triggers" sets. 

However, testing just now, I can't make that happen, so maybe it was due to something upstream that I fixed already (e.g. maybe a transient task proxy was being used somewhere instead of an existing active instance).

So ... I've removed the methods. Hopefully that won't come back to haunt us.

cylc/flow/task_pool.py
```diff
@@ -2268,7 +2266,7 @@ def _set_prereqs_tdef(
     ) -> Optional[TaskProxy]:
         """Spawn an inactive task and set prerequisites on it."""
         itask = self.spawn_task(
-            taskdef.name, point, flow_nums, flow_wait=flow_wait, force=True
```
oliver-sanders: This option was being used. Was this a bug?

changes.d/6395.feat.md
```diff
@@ -0,0 +1,3 @@
+When triggering a group of tasks, remove flow history (by default) to allow
+rerun in the same flow, satisfy off-group prerequisites to avoid a stall,
+and leave in-group prerequisites to be satisfied by the flow.
```
oliver-sanders: Great technical description, but it's not obvious what this means to users:

```suggestion
`cylc trigger` now respects the dependencies between tasks allowing it to be used to (re)-run a subgraph of tasks.
```

cylc/flow/flow_mgr.py
```diff
@@ -33,19 +33,23 @@
...
-             f' no flow ("{FLOW_NONE}"); a new flow ("{FLOW_NEW}");'
-             f' or a specific flow (e.g. "2"). The default is "{FLOW_ALL}".'
-             ' Specific flow numbers can be new or existing.'
+        "--flow",
+        action="append",
+        dest="flow",
+        metavar="FLOW",
+        default=[],
+        help='Assign affected tasks to specified flows.'
+             'By default, assign to all active flows; otherwise'
```
oliver-sanders:
```suggestion
             ' By default, assign to all active flows; otherwise'
```

changes.d/6395.feat.md
hjoliver: I'm not super happy about only providing dumbed down explanations aimed at the least savvy users, 

The trouble with not mentioning flow erasure and off-group prerequisites is that more savvy users (and there are at least a few of them) are likely to assume that (or at least wonder if) trigger won't achieve the desired result without manually handling those aspects.

I'll take the simplified description, and add a short "advanced" description at the bottom.

oliver-sanders: Changelog entries are intended to give a quick user-facing overview. Use the user guide to document the fine details.

> and there are at least a few of them

FYI, we don't have *any*.

wxtim: > I'll take the simplified description, and add a short "advanced" description at the bottom.

Was going to suggest this.

hjoliver: Sorry, my bad 🤦too much multi-tasking brain fry - from a quick look here (not via Files Changed) for some reason I assumed you were commenting on text at the top of the CLI help, not the change log (hence my response that I'd add an "advanced" section below it, which is clearly not gonna work for the change log). So, actually, I agree that my text is overkill for the change log.

> FYI, we don't have any.

Jeez, there must be some. Even if not, there *could* be (or *should* be) so we should still explain things well enough for users to understand what's going on if they do want or need to know (but yeah, not in the change log!).

oliver-sanders: > Jeez, there must be some. 

A topic for another day, but, I have good reason/evidence for saying this. Our users are just abandoning their workflow runs rather than using new flows.

And, yeah, obvs no issue with that detail in the `cylc trigger --help`.

oliver-sanders (changes requested):
Still reviewing (sorry, struggling for time ATM). The `--flow=all` change has introduced an inter-version compatibility bug which needs to be fixed - https://github.com/cylc/cylc-flow/pull/6395#issuecomment-3061595557 (if we bump the `--flow=all` change to another PR, this PR will be unstuck, otherwise it will have to be fixed here).

tests/integration/test_force_trigger.py
```diff
@@ -63,39 +74,71 @@ async def test_trigger_workflow_paused(
...
         assert log_filter(
-            level=logging.ERROR,
-            contains="ignoring trigger - already active"
+            level=logging.WARNING,
+            contains="Job already in process - ignoring"
         )
 
 
+async def test_trigger_group_whilst_paused(flow, scheduler, run, complete):
+    """Only group start-tasks should run whilst the scheduler is paused.
```
wxtim: Was this agreed? It feels like slightly odd behaviour - I'd expect that the entire subgraph would run if I'd deliberately triggered it whilst paused.

tests/integration/test_force_trigger.py
```diff
@@ -63,39 +74,71 @@ async def test_trigger_workflow_paused(
...
+
+    When multiple tasks are triggered, only tasks with no dependencies within
+    the group should run whilst the scheduler is paused.
+
+    The remaining tasks will run according to normal prerequsitie satisfaction
```
wxtim: Spilling?
```suggestion
    The remaining tasks will run according to normal prerequisite satisfaction
```

wxtim: Wonder whether it should be

```suggestion
    The remaining tasks will run as normal, when their prerequisites
    are satisfied.
```

wxtim (commented):
.

tests/integration/test_force_trigger.py
oliver-sanders: It wasn't noticed/discussed in the proposal, but it was brought up above (https://github.com/cylc/cylc-flow/pull/6395#issuecomment-2782451295)

* At the moment, if you trigger a task whilst the workflow is paused, it will run.
* This is reasonable and there are use cases for doing so.
* With this change we allow trigger dependent tasks.
* You could use this to re-run an inter-dependent family or even the entire workflow.
* In order for workflow paused to mean something, we have to draw the line somewhere.
* We wouldn't want operators, not to be able to slam on the breaks when they need to.
* You can always push tasks through submission whilst the workflow is paused by triggering them again.

changes.d/6395.feat.md
hjoliver: Well I do have several here who have played around with new flows and commented on it being useful and powerful. But even if that weren't the case we still need to properly document how the system works. It is possible to start new flows, and there are good reasons to do it (group trigger is still using this feature under the hood too), and if someone tries it and needs to understand the behaviour they see then it should be documented (but not in the change log!).

tests/integration/test_force_trigger.py
hjoliver: Damn, "prerequisite" has got to be the most mis-"spilled" word in the codebase. Other than "cylc". 🤣

cylc/flow/task_pool.py
hjoliver: I presume you mean the `force` option, but I'm not sure where this diff came from. It's marked as outdated, and the code now shows no difference from master here (and no `force` option). Maybe the option got removed on master by another PR (and I've since rebased this branch). [... Nope, `git blame` says not].

hjoliver:
OK, last two commits:
- fixed a small off-topic remove bug that I ran noticed (highly unlikely that it affected anyone)
- shimmed `--flow=all` compat for old clients (tested with CLI and web UI, for set, trigger, remove)

tests/integration/test_force_trigger.py
wxtim: Sorry - I missed the comment.

tests/integration/scripts/test_back_compat_flow_all.py
```diff
@@ -0,0 +1,93 @@
...
+    run_cmd,
+    set_prereqs_and_outputs,
+)
+from cylc.flow.exceptions import InputError
+from cylc.flow.scheduler import Scheduler
+from cylc.flow.task_outputs import TASK_OUTPUT_SUCCEEDED
+
+
+async def test_back_compat_flow_all(flow, scheduler, start):
+    """blah"""
```
wxtim: Remove or add helpful comment.

Why does this test back compatibility? As far as I can see it's only testing that providing a garbage flow number causes a failure?

tests/integration/test_force_trigger.py
```diff
@@ -63,39 +74,72 @@ async def test_trigger_workflow_paused(
...
+    schd = scheduler(id_)
+    async with run(schd):
+        # trigger the chain a => c
+        await run_cmd(force_trigger_tasks(schd, ['1/a'], []))
+
+        # 1/a should run whilst the workflow is paused (group start-task)
+        await complete(schd, '1/a', allow_paused=True, timeout=1)
+
+        # 1/b should *not* run whilst the workflow is paused
+        with pytest.raises(AssertionError):
```
wxtim, line 130: This test feels isufficiently fragile - anything which causes a failure to complete task will cause this test to pass - if it's a total failure that's fine, but if it's a delay then the whole test could pass when it shouldn't.

tests/integration/test_force_trigger.py
```diff
@@ -129,23 +173,313 @@ async def test_trigger_on_resume(
...
+
+    It should satisfy off-group task and xtrigger prerequisites automatically.
+
+    """
+    cfg = {
+        'scheduling': {
+            'xtriggers': {
+                'xr': 'xrandom(0)'  # never satisfied
+            },
+            'graph': {
```
wxtim, line 220: Worth covering the old ext-triggers whilst we're at it?

(also needs [this suggestion](https://github.com/cylc/cylc-flow/pull/6395/files#r2210438903))

```suggestion
            'special tasks': {
                'external-trigger': 'ext_trigger("sent")'
            },
            'graph': {
```

tests/integration/test_force_trigger.py
```diff
@@ -129,23 +173,313 @@ async def test_trigger_on_resume(
...
+                    x => a
+
+                    # sub-graph for group trigger:
+                    a => b & c => d
+
+                    # downstream:
+                    d => e => y
+
+                    # off-group prerequisites:
+                    @xr => x  # stop the flow starting
```
wxtim, line 232: See https://github.com/cylc/cylc-flow/pull/6395/files#r2210436921
```suggestion
                    ext_trigger => x
                    @xr => x  # stop the flow starting
```

tests/integration/test_force_trigger.py
```diff
@@ -129,23 +173,313 @@ async def test_trigger_on_resume(
...
+        await complete(schd, '1/e')
+
+        # It should satisfy off-group prerequisites.
+        assert log_filter(
+            regex="1/c:waiting.*prerequisite force-satisfied: xr = xrandom")
+        assert log_filter(
+            regex="1/b:waiting.*prerequisite force-satisfied: 1/off:succeeded")
+        assert log_filter(
+            regex="1/a:waiting.*prerequisite force-satisfied: 1/x:succeeded")
+
```
wxtim, line 262: And a check that ext-trigger has been seen

```suggestion
        assert log_filter(
            regex=(
                "1/a:waiting.* prerequisite force-satisfied: "
                "1/ext_trigger:succeeded"))

```

tests/integration/test_force_trigger.py
```diff
@@ -129,23 +173,313 @@ async def test_trigger_on_resume(
...
+        log.clear()
+
+        # Trigger the group again, now as past tasks, in the same flow.
+        # It should erase flow 1 history to allow the rerun.
+        # It should not flow on to e, which already ran in flow 1.
+
+        # Create an active task that needs removing (to plug a coverage hole).
```
wxtim:
```suggestion
        # Create an active task that needs removing to test that such a task is removed.
```

tests/unit/scripts/test_completion_server.py
```diff
@@ -170,77 +170,75 @@ async def test_complete_cylc(dummy_workflow):
...
         '--flow=new',
     }
 
-    # $ cylc triger --flow=all <tab><tab>
-    assert await _complete_cylc('cylc', 'trigger', '--flow=all', '') == {
-        'foo/run2//'
-    }
-
-    # $ cylc triger --62656566<tab><tab>
+    # $ cylc trigger --62656566<tab><tab>
```
oliver-sanders, line 243: This was testing that the completion moved onto the task argument after `--flow`, not `--flow=all`, we can just change `all` to something else:

```suggestion
    # $ cylc triger --flow=1 <tab><tab>
    assert await _complete_cylc('cylc', 'trigger', '--flow=1', '') == {
        'foo/run2//'
    }

    # $ cylc trigger --62656566<tab><tab>
```

tests/unit/test_taskdef.py
```diff
@@ -195,3 +195,62 @@ def test_get_xtrigs(tmp_flow_config):
...
+            [],
+            id='it.does-not-return-suicide-triggers',
+        ),
+    ],
+)
+def test_get_triggers(tmp_flow_config, task, point, expected):  # noqa: F811
+    """Test that get_triggers() returns the correct triggers for a task.
+
+    """
+    id_ = 'gargle-blaster'
```
oliver-sanders, line 234: :champagne: => :brain: :lemon: :bricks:

tests/integration/scripts/test_back_compat_flow_all.py
oliver-sanders:
```suggestion
    """It should support the `all` argument from older client versions."""
```

hjoliver: Oops (newly added test, temporary "blah" not replaced).

> Why does this test back compatibility?

See my comment just above:
> shimmed `--flow=all` compat for old clients (tested with CLI and web UI, for set, trigger, remove)

The `--flow=all` option is no longer valid ("all" flows is now the default in all cases) but Oliver (rightly) asked me to "shim" it in because it can still come in from old (pre 8.5) clients.

> As far as I can see it's only testing that providing a garbage flow number causes a failure?

It tests that garbage flow numbers fail, but more importantly that a particular garbage flow number ("all") does not fail (as just described).  The new unit test in `commands.py` tests that it trasnslates to the default `flows = [ ]`).

hjoliver:
Actually, disconfirmed, with a caveat.

That's a kinda fiendish example @wxtim , as it turns out, but your expectations are a bit wrong. It should be this:

1.  Start `a` and `b` *and `trigger`*
2.   Run `trigger` which would
    2a. Leave `a` alone.
    2b. Kill `b` *and `trigger`*
    2c. Remove `b` *and `trigger`*
3. the workflow shuts down as completed when `a` succeeds, because `a:started` is a past event - the group start task is not going to complete any outputs, post intervention, to start the new flow.

This is likely not what the user intended, but it is what the proposal says.  The thing is, `b` will not retrigger because `a:started` is a past event.

Currently: user beware if triggering a group that contains active group start tasks (some outputs may already have spawned children).

tests/integration/scripts/test_back_compat_flow_all.py
hjoliver: Fixed the docstring.

tests/integration/test_force_trigger.py
hjoliver: OK, I added an additional check for "waiting" status.

tests/integration/test_force_trigger.py
hjoliver: Yikes, good suggestion - I had entirely forgotten about old-style push external triggers. Luckily that's an easy fix... DONE

tests/integration/test_force_trigger.py
hjoliver: done

tests/integration/test_force_trigger.py
hjoliver: done

cylc/flow/task_events_mgr.py
```diff
@@ -668,6 +668,8 @@ def process_message(
             True: if polling is required to confirm a reversal of status.
 
         """
+        # Useful debug but currently borks tests/f/cylc-message/02-multi.t:
```
wxtim, line 688: IMO a line about why this currently borks tests?

cylc/flow/scripts/trigger.py
```diff
@@ -17,43 +17,46 @@
...
+that would cause a stall.
 
-Triggering a queued task runs it immediately.
+Triggering an unqueued task queues it; triggering a queued task runs it; so you
+may need to trigger unqueued tasks twice to run them now if the queue is full.
 
-Cylc queues restrict the number of jobs that can be active (submitted or
-running) at once. They release tasks to run when their active task count
-drops below the queue limit.
+Preparing, submitted, and running tasks can't be retrigged until finished.
```
wxtim: I think this might confuse becuase tasks inside a sub-graph can be retriggered under these circumstances, and will be killed:
```suggestion
Tasks at the start of a retriggered sub-graph (including single tasks) cannot
be retriggered whilst in preparing, submitted, or running states.
```

P.S. Even if you disagree, check spelling of "retriggered".

cylc/flow/scripts/trigger.py
```diff
@@ -17,43 +17,46 @@
...
 
-Force task(s) to run regardless of prerequisites, even in a paused workflow.
+Manually trigger a group of one or more tasks, automatically satisfying any
+off-group dependencies and respecting dependencies within the group.
 
-Triggering a task that is not yet queued will queue it.
+Group trigger is typically the easiest way to rerun a sub-graph. It erases
+flow history to allow the rerun; it identifies and triggers group start tasks
+to start the flow; and it identifies and satisfies any off-group prerequisites
+that would cause a stall.
```
wxtim: I'm not sure my suggestion actually improves on your text, but I think it's worth a look.

I think that my problem with the orginal is using the term "Group trigger" implies something more special than "triggering >1 task".

I also think that it's worth being super clear about the differences between "start tasks" and tasks inside a sub-graph.

```suggestion
Trigger one or more tasks.

If used to trigger a group of tasks, Cylc will identify tasks
at the start of the sub-graph and start with these tasks.
Cylc trigger will erase the histories of the remaining tasks,
so that dependencies inside the group will be respected.
Dependencies outside the group triggered will be
satisfied to avoid a stall.
```

cylc/flow/scripts/trigger.py
```diff
@@ -99,14 +102,14 @@
 
 def get_option_parser() -> COP:
     parser = COP(
-        __doc__,
+        str(__doc__),
```
wxtim: Why?

cylc/flow/scripts/remove.py
```diff
@@ -87,25 +85,13 @@
 
 def get_option_parser() -> COP:
     parser = COP(
-        __doc__,
+        str(__doc__),
```
wxtim: Again, why?

tests/integration/test_task_pool.py
```diff
@@ -1432,7 +1358,7 @@ async def test_set_bad_prereqs(
     def set_prereqs(prereqs):
         """Shorthand so only varible under test given as arg"""
```
wxtim, line 1359:
```suggestion
def set_prereqs(prereqs):
"""Shorthand so only variable under test given as arg"""
schd.pool.set_prereqs_and_outputs(
```

wxtim: Off topic, but might as well fix
```suggestion
        """Shorthand so only variable under test given as arg"""
```

tests/functional/spawn-on-demand/16-parents-yes-no/flow.cylc
```diff
@@ -7,9 +7,10 @@
         inactivity timeout = PT60S
         abort on inactivity timeout = True
         expected task failures = 1/failer
-        # Unstall, to see if foo gets triggered by baz at point 2.
-        stall handlers = cylc trigger %(workflow)s//2/baz, \
-                           cylc remove %(workflow)s//1/failer
+        # Unstall, to see if foo gets spawned by baz at point 2.
```
wxtim:
```suggestion
        # Un-stall, to see if foo gets spawned by baz at point 2.
```

cylc/flow/scripts/trigger.py
oliver-sanders: Agreed:

* No need to make "group trigger" a user-facing term, we should say "trigger a group of tasks".
* The command should have a simple one-line summary (note, a one line summary is also required by `cylc help all`).

cylc/flow/scripts/trigger.py
```diff
@@ -17,43 +17,46 @@
...
 
 Examples:
-  # trigger task foo in cycle 1234 in test
-  $ cylc trigger test//1234/foo
+  # trigger task foo in cycle 1, in workflow "test"
+  $ cylc trigger test//1/foo
 
-  # trigger all failed tasks in test
-  $ cylc trigger 'test//*:failed'
+  # trigger all failed tasks in workfow "test"
```
wxtim:
```suggestion
  # trigger all failed tasks in workflow "test"
```

tests/integration/test_xtrigger_mgr.py
```diff
@@ -463,29 +463,26 @@ async def test_force_satisfy(flow, start, scheduler, log_filter):
 
         assert foo.state.xtriggers['x']  # satisified
```
wxtim:
```suggestion
        assert foo.state.xtriggers['x']  # satisfied
```

cylc/flow/scripts/trigger.py
oliver-sanders: Also, remember that triggering one task is a more common use case than triggering a group of tasks, so we should bump the group-specific part of this down a bit to make the CLI help clearer to new users who are still learning the basics.

hjoliver:
> Intra-version compatibility:

@wxtim - thanks for the thorough check there.  It looks like every passes - all I see is green ✅s - in which case I don't understand your comment \*-marked boxes and comment *\* These fail both against this branch and master. Should not be fixed here.* ??

cylc/flow/task_events_mgr.py
hjoliver: Done, plus https://github.com/cylc/cylc-flow/issues/6857

cylc/flow/scripts/trigger.py
hjoliver: Yep, you're right.

cylc/flow/scripts/trigger.py
hjoliver: Yep, agreed, dropping it down and rephrasing...

cylc/flow/scripts/trigger.py
hjoliver: I think my neovim setup was complaining about a typing issue there.  Not a good excuse. I'll revert and investigate.

tests/functional/spawn-on-demand/16-parents-yes-no/flow.cylc
hjoliver: Not sure it matters, in a test, but OK!

cylc/flow/scripts/trigger.py
hjoliver: For clarity, I've tried to redo the CLI help by separately addressing single- and multi-task triggering. See what you think.

cylc/flow/scripts/trigger.py
hjoliver: I've completely changed this bit, for an extensive help rewrite - see what you think.

cylc/flow/scripts/trigger.py
hjoliver: Note I've coined a new term "live tasks" for tasks with live jobs (or job scripts, at least) -  preparing, submitting, running -  since we can no longer call these "active tasks" and the full description is verbose as hell. If agreed I'll add it to the glossary...

cylc/flow/scripts/trigger.py
```diff
@@ -17,27 +17,35 @@
...
-
-If the workflow is paused only group start tasks trigger immediately; the flow
-will continue when the workflow resumes.
+Manually trigger tasks, respecting dependencies among them.
+
+Triggering individual tasks:
+  * Triggering an unqueued task queues it; triggering a queued task runs it;
+    so to run an unqueued task immediately you many need to trigger it twice.
+  * Tasks can be triggered to run even if the workflow is paused.
+  * Attempts to live tasks (preparing, submitted, running) will be ignored.
```
wxtim: The rest of this section reads really nicely - thank you!

```suggestion
  * Attempts to trigger live tasks (preparing, submitted, running) will be ignored.
```

wxtim:
>  These fail both against this branch and master. 

@hjoliver  - they fail, but the behaviour of your branch is identical to master.

> Should not be fixed here.

if we decide that a fix is required (it's probably not, becuase the number of users who will encounter this is miniscule, and it can be easily fixed by upgrading your command issuing version of Cylc) it should have it's own bug fix PR - I think getting this done and 8.5 out is higher priority.

wxtim (approved):
1 grammatical issue noted.

oliver-sanders:
I think we're there on the group trigger changes :tada:

But, unfortunately, I've gotta gripe with one of the off-topic changes, the change from triggering n=0 tasks in their assigned flows, to triggering tasks in all flows regardless.

This change is a regression of a recent bugfix https://github.com/cylc/cylc-flow/issues/6361 (we implemented the proposed behaviour wrong and had to fix it).

I get why you did this, without this change group-trigger had the potential to trigger group-start tasks in a different flow to their in-group downstreams which is odd in the extreme. In fact this behaviour is basically the same bug we were fixing with https://github.com/cylc/cylc-flow/issues/6361, so I accept the need for a change.

But the change to triggering n=0 tasks in all active flows is a problem for reflow use cases as it makes it much easier for the user to accidentally force flow-merging without asking for it. E.g, pressing the set/trigger button on any failed task will cause Cylc to merge in ALL active flow numbers as a side-effect (which is just confusing). If you're trying to wash over these tasks with a new flow, that could effectively disable reflow behaviour.

I've come up with a compromise solution which I hope you like. It preserves the old behaviour when triggering non-interdependent tasks, but extends this behaviour to provide a more sensible default for group-trigger use cases (where we are triggering groups of interdependent tasks).

* We take the tasks the user triggered and break them down into interdependent groups (i.e. groups of tasks for us to trigger in order).
* If there are any n=0 tasks in that group, we use the set of these flow numbers when triggering the group.
* We handle each group separately.

With this approach:

* We maintain the old status quo of not messing with an n=0 task's flow numbers unless the user asked us to.
* But all tasks in an interdependent group will trigger with the same flow numbers.
* Bonus, we can safely trigger n=0 tasks from different flows in the same operation without unexpected results.
* It's basically the proposed behaviour but extended from one task to many in order to accommodate group-trigger.

I've put it together here: https://github.com/hjoliver/cylc-flow/pull/68

(I haven't looked through the CLI/docs yet though, can do tomorrow)

hjoliver:
@oliver-sanders -  your side PR looks good except that it doesn't work for task glob inputs - hopefully that's a quick fix.

cylc/flow/scripts/trigger.py
```diff
@@ -17,43 +17,54 @@
...
-
-Attempts to trigger active (preparing, submitted, running)
-tasks will be ignored.
+Manually trigger tasks, respecting dependencies among them.
+
+Triggering individual tasks:
+  * Triggering an unqueued task queues it; triggering a queued task runs it;
+    so to run an unqueued task immediately you many need to trigger it twice.
+  * Tasks can be triggered to run even if the workflow is paused.
+  * Attempts to trigger live tasks (preparing, submitted, running) will be ignored.
```
wxtim: flake8 fail
```suggestion
  * Attempts to trigger live tasks (preparing, submitted, running) will be
    ignored.
```

oliver-sanders:
> Bandit failure seems unreasonable in the context?

Bandit has lost the plot! Added an ignore.

oliver-sanders:
[22e11f1](https://github.com/cylc/cylc-flow/pull/6395/commits/22e11f1ad226f62ef05cf490d02d1d717d51f562) should resolve the mypy issue.

oliver-sanders:
Tests passing!

PR #6400, opened 2024-10-10
examples: using data to define your workflow

oliver-sanders:
A relatively common workflow implementation pattern.

I was writing an example for someone yesterday at a training session and thought I would expand it into a documented example for future reuse.

**Check List**


hjoliver (approved):
Nice. Worth flagging this one as Tip on the forum, once merged.

cylc/flow/etc/examples/external-data-files/index.rst
```diff
@@ -0,0 +1,122 @@
...
+The three files are arranged like so:
+
+.. code-block:: none
+   :caption: File Structure
+
+   |-- flow.cylc
+   |-- lib
+   |   `-- python
+   |       `-- load_data.py
+   `-- stations.json
```
wxtim, line 98: Is it worth encouraging users to put files like this in `etc/`

wxtim (approved):
Nice one. One minor question.

cylc/flow/etc/examples/external-data-files/index.rst
oliver-sanders: * The `load_data.py` file has to go in `lib/python` (in order to be put in `PYTHONPATH`).
* The `stations.json` file could go anywhere.

I'm not sure whether we have given any guidance over the use of `etc/` directories. I think the only thing Cylc does with `etc/` directories is to add them to the [remote install list](https://cylc.github.io/cylc-doc/stable/html/7-to-8/major-changes/cylc-install.html#remote-installation), beyond that they're functionally no different to any other directory.

I guess in this case, we probably don't want the stations list to be installed onto remote platforms because it is only used to define the workflow configuration locally?

> Is it worth encouraging users to put files like this in etc/

TBH, dunno.

wxtim: > I guess in this case, we probably don't want the stations list to be installed onto remote platforms because it is only used to define the workflow configuration locally?

Probably not in that case. Although I'm tidy minded enough to want them in a subfolder it's not germane to this example.

wxtim (approved):

Issue #6307, opened 2024-08-16
GraphQL: updating arrays with delta updates


oliver-sanders:
Data stores maintained by GraphQL updates (i.e. cylc-ui) are not presently able to synchronise some array fields.

This has been discussed before (can't find the issue dammit). We have previously worked around the issue by repeating *all* elements of updated arrays in the returned deltas.

### Example: Outputs

This subscription tracks task outputs:

```graphql
subscription {
  deltas {
    updated {
      workflow {
        taskProxies {
          id
          outputs {
            label
            satisfied
          }
        }
      }
    }
  }
}
```

The corresponding `added` subscription, would initially yield a list of all outputs, i.e:

```js
[
    {'label': 'submitted', 'satisfied': false},
    {'label': 'started', 'satisfied': false},
    {'label': 'succeeded', 'satisfied': false},
    {'label': 'failed', 'satisfied': false},
    {'label': 'submit-failed', 'satisfied': false},
    {'label': 'expired', 'satisfied': false}
]
```

Whenever there is a change, e.g. when the first job submits, we will receive only the element that was updated e.g:

```js
[
    {'label': 'submitted', 'satisfied': true},
]
```

And so on:

```js
[
    {'label': 'started', 'satisfied': true},
]
```

```js
[
    {'label': 'succeeded', 'satisfied': true},
]
```

This makes sense, we are pushing only the minimal information down the wire.

But how is the remote data store supposed to work with these deltas?

The remote data store needs to be able to associate the updated element back to the corresponding element in the store in order to update it.

In this case it's easy to do by inspection, the `label` is the unique key representing the object. The remote data store should look through the outputs until it finds one with the same label, and update the `satisfied` field to match the update. Although this would require the remote data store to know that `label` is the unique key for `outputs` arrays which makes updates complicated.

### Example: Prerequisites

However, it's not always so easy. Take the prerequisites field for example:

```
subscription {
  deltas {
    updated {
      workflow {
        taskProxies {
          id
          prerequisites {
            satisfied
            expression
            conditions {
              taskId
              message
              exprAlias
              satisfied
            }
          }
        }
      }
    }
  }
}
```

There is no unique key here. The `expression` field (which has values like `c0 & c1` or `c0 | c1`) is not unique. The conditions array isn't unique either since the same conditions could be arranged into different prerequisite expressions.

The only way we can get a unique key is to form a composite of the following fields:

```
expression
conditions {
    taskId
    message
    exprAlias
}
```

[...]

PR #6403, opened 2024-10-11
data store: dont update outputs incrementally

oliver-sanders:
An awkward bodge to work around https://github.com/cylc/cylc-flow/issues/6307

I'm not sure we want to go ahead with this, at least not in the long term, however, it should be enough to unblock (at lest the review of) https://github.com/cylc/cylc-ui/pull/1886

**Check List**

- [ ] Internal protocol issue, no changelog required

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -2420,23 +2420,27 @@ class DataStoreMgr:
                 objects from the workflow task pool.
 
         """
-        tproxy: Optional[PbTaskProxy]
-        tp_id, tproxy = self.store_node_fetcher(itask.tokens)
-        if not tproxy:
-            return
-        outputs = itask.state.outputs
-        label = outputs.get_trigger(message)
-        # update task instance
-        update_time = time()
-        tp_delta = self.updated[TASK_PROXIES].setdefault(
-            tp_id, PbTaskProxy(id=tp_id))
-        tp_delta.stamp = f'{tp_id}@{update_time}'
-        output = tp_delta.outputs[label]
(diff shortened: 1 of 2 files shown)


MetRonnie:
Looks like b2accfb885b4a8ea7dce1ad2bfdff4d089ac31b1 (`examples: using data to define your workflow`) included accidentally?

oliver-sanders:
(removed)

cylc/flow/data_store_mgr.py
```diff
@@ -2428,11 +2428,25 @@ def delta_task_output(
...
+                _output = tproxy.outputs[_label]
+                output.label = _output.label
+                output.message = _output.message
+                output.satisfied = _output.satisfied
+
+            output.time = update_time
+
         self.updates_pending = True
 
     def delta_task_outputs(self, itask: TaskProxy) -> None:
```
dwsutherland, line 2452: Might as well only use this call...

dwsutherland: Maybe just call `delta_task_outputs` from within `delta_task_output`

oliver-sanders: This is a bodge workaround, I'm not sure whether we should merge it.

The incremental update is the right thing to do (so I don't want to blast that code aside), we just didn't develop a mechanism for syncing these updates over GraphQL so the UI can't handle it.

oliver-sanders:
Short of a universal solution to #6307, this patch seems a reasonable way to get the info view in the UI off of the ground.

MetRonnie (approved):

PR #6408, opened 2024-10-14
GH actions: Ubuntu 24

MetRonnie:
https://github.com/actions/runner-images/issues/10636 - Ubuntu 24 is rolling out as `ubuntu-latest`, dropping Python 3.7 and bringing a new version of shellcheck.

We can keep using `ubuntu-22.04` for Python 3.7

**Check List**

- [x] No dependency changes, tests, docs etc.

### DIFF
--- a/tests/flakyfunctional/cylc-poll/16-execution-time-limit.t
+++ b/tests/flakyfunctional/cylc-poll/16-execution-time-limit.t
@@ -33,6 +33,7 @@ run_ok "${TEST_NAME_BASE}-validate" cylc validate "${WORKFLOW_NAME}"
 workflow_run_ok "${TEST_NAME_BASE}-run" \
     cylc play --reference-test -v --no-detach "${WORKFLOW_NAME}" --timestamp
 #-------------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


tests/flakyfunctional/cylc-poll/16-execution-time-limit.t
```diff
@@ -33,6 +33,7 @@ run_ok "${TEST_NAME_BASE}-validate" cylc validate "${WORKFLOW_NAME}"
 workflow_run_ok "${TEST_NAME_BASE}-run" \
     cylc play --reference-test -v --no-detach "${WORKFLOW_NAME}" --timestamp
 #-------------------------------------------------------------------------------
+# shellcheck disable=SC2317
```
MetRonnie, line 36: SC2317 is a warning about unreachable code. In this case as we are doing `run_ok $TEST_NAME cmp_times ...` shellcheck is unable to determine whether we are calling the function because we are passing it by name as an argument to a different function.

oliver-sanders (approved):

hjoliver (approved):

Issue #6315, opened 2024-08-21
task reset from succeeded to running


oliver-sanders:
There seems to be some circumstance where succeeded tasks can go back to running (apparently in the same main-loop iteration).

This log is typical of the issue:

```console
$ grep 20190401T0000Z/archive_logs log/scheduler/log
2024-07-29T22:59:59Z INFO - [20190401T0000Z/archive_logs:waiting(runahead)] => waiting
2024-07-30T13:17:30Z INFO - [20190401T0000Z/archive_logs:waiting] => waiting(queued)
2024-07-30T13:17:30Z INFO - [20190401T0000Z/archive_logs:waiting(queued)] => waiting
2024-07-30T13:17:31Z INFO - [20190401T0000Z/archive_logs:waiting] => preparing
2024-07-30T13:17:43Z INFO - [20190401T0000Z/archive_logs/01:preparing] submitted to xce:pbs[2492608]
2024-07-30T13:17:44Z INFO - [20190401T0000Z/archive_logs/01:preparing] => submitted
2024-07-30T13:18:05Z INFO - [20190401T0000Z/archive_logs/01:submitted] => running
2024-07-30T13:18:39Z INFO - [20190401T0000Z/archive_logs/01:running] => succeeded
2024-07-30T13:18:39Z INFO - [20190401T0000Z/archive_logs/01:running(runahead)] => running
```

Every time we see the confusing `running(runahead) => running` transition. There are no obvious exacerbating circumstances.

This seems to be intermittent, but we appear to have an example which yields this bug relatively regularly, though I've not been able to reproduce it yet.

hjoliver:
It might be useful to see the whole log, not just lines associated with the affected task.  Maybe there was some kind of manual intervention that affected runahead tasks?  e.g. triggering a runahead task. I'm not sure that we're properly removing task attribute labels like "queued" and "runahead" every time when we (e.g.) force a runahead-limited task task to run.

oliver-sanders:
> It might be useful to see the whole log, not just lines associated with the affected task

The whole log is very long. I went through carefully and could not find any other relevant log lines.

* There were no client commands at all.
* There were no restarts/reloads.
* There were no unexpected external factors or failures.

oliver-sanders:
I haven't managed to make head or tail of this one yet. One user's workflows have encountered this a few times, they are on leave at the moment, hopefully when they return we can get them to run these workflows in debug mode which might give us a better chance of debugging.

[...]

PR #5619, opened 2023-07-05
pool: ensure the DB is updated when tasks are removed

oliver-sanders:
* Before this change, the `task_pool` table was only being updated when tasks were added, or changed state, but not when they were removed.
* Closes #5598

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #6409, opened 2024-10-14
pool: update DB after removing a task

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/6315
* If the DB is not updated after a task is removed, then it can be respawned in its previous state as the result of upstream output completion.

This fixes the "running(runahead) => running" bug that could cause tasks to get stuck in the running state indefinitely.

Performing a DB write every time a task completes is going to be a performance hit since this is performed per-task not per-main-loop-cycle (i.e. there is no batching efficiency gain). I'm not sure what we can do about this without changes to the task_pool data model. Any ideas?

**Check List**

### DIFF
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -870,6 +870,11 @@ class TaskPool:
                 msg += " - active job orphaned"
 
             LOG.log(level, f"[{itask}] {msg}")
+
+            # ensure this task is written to the DB before moving on
+            # https://github.com/cylc/cylc-flow/issues/6315
+            self.workflow_db_mgr.process_queued_ops()
+
             del itask
 
     def get_tasks(self) -> List[TaskProxy]:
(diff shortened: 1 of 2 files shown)


hjoliver:
> Performing a DB write every time a task completes is going to be a performance hit since this is performed per-task not per-main-loop-cycle

Hopefully not too bad, because the task completion, even of family members, tends to be staggered rather than all-at-once.

>  I'm not sure what we can do about this without changes to the task_pool data model. Any ideas?

Yeah, batching DB ops for efficiency can be problematic if there's any chance of certain events occurring between DB updates.

To avoid any chance of this kind of bug I guess we have to either:
- write "live" info per event and only batch-write the historical-record tables
- batch-write and batch-spawn in the main loop (yuck?)

hjoliver (approved):

hjoliver:
Merging this as the fix is simple and necessary ... the questions are of a wider scope.

oliver-sanders:
> To avoid this kind of bug I guess we have to either:

A fancy solution would be to allow the unwritten data (i.e. the delta) to be queried allowing us to avoid hitting the DB when not necessary, however, this would not be a small job.

MetRonnie:
See also https://github.com/cylc/cylc-flow/pull/5619

Issue #6419, opened 2024-10-15
tui: end key doesn't jump to bottom of log view


oliver-sanders:
I would expect the home/end keys to jump to the top/bottom of the log file, for some reason they don't.

Either fix it or provide some other form of jump to bottom functionality.

Reported in this thread: https://cylc.discourse.group/t/cylc-tui-jump-to-the-end-of-a-file-you-are-viewing/1039

oliver-sanders:
The urwid.ListBox widget does not support the home/end keys, so this event is bubbling and appears to be captured by another item in the list with the effect that it only jumps to the bottom of the header of the log view, not the bottom of the file itself.

PR #6422, opened 2024-10-16
tui: support the home/end keys in the log view

oliver-sanders:
* Allows navigating to the top/bottom of the log file.
* Closes https://github.com/cylc/cylc-flow/issues/6419

The widget we are using (urwid.ListBox) does not support the home/end keys. Other widgets bring other problems as `ListBox` very elegantly handles the issue that we want to keep the focus on the file selector button (so you can change file by pressing enter) whilst being able to scroll the log file (which is a different component all together).

This PR adds a somewhat hacky but rather effective solution. When the user presses "home/end", fire off the "page up/down" key until the window stops scrolling (which happens when we hit the top/bottom of the ListBox).

**Check List**


wxtim (approved):
Nice.
- [x] Makes sense
- [x] Manually tested

PR #6426, opened 2024-10-17
Spelling/formatting

MetRonnie:
https://github.com/cylc/cylc-doc/actions/workflows/test.yml

### DIFF
--- a/cylc/flow/scripts/lint.py
+++ b/cylc/flow/scripts/lint.py
@@ -150,15 +150,16 @@ OBSOLETE_ENV_VARS = {
 
 
 DEPRECATED_STRING_TEMPLATES = {
-    'suite': 'workflow',
-    'suite_uuid': 'uuid',
-    'batch_sys_name': 'job_runner_name',
-    'batch_sys_job_id': 'job_id',
-    'user@host': 'platform_name',
-    'task_url': '``URL`` (if set in :cylc:conf:`[meta]URL`)',
-    'workflow_url': (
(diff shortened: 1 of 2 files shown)


cylc/flow/scripts/lint.py
```diff
@@ -603,21 +603,21 @@ def list_wrapper(line: str, check: Callable) -> Optional[Dict[str, str]]:
...
-            ' with the target job runner\'s platform has been lost.'
-            ' \n\nThe following directives are considered equivelent to'
-            ' execution time limit:\n * '
+            "Using ``[runtime][TASK]execution time limit`` is"
+            " recommended in preference to using job runner"
+            " directives because it allows Cylc to retain awareness"
+            " of whether the job should have finished, even if contact"
+            " with the target job runner's platform has been lost."
+            " \n\nThe following directives are considered equivalent to"
+            " execution time limit:\n * "
```
hjoliver, line 612: Might as well tighten this up - it's a bit verbose IMO.
```suggestion
            "Use ``[runtime][TASK]execution time limit`` to tell
            " Cylc when a job should finish. It is automatically"
            " translated to the right job runner directive:\n * "
```

hjoliver (changes requested):

cylc/flow/scripts/lint.py
MetRonnie: I've made it more concise but left in the explanation for why

hjoliver (approved):

PR #6428, opened 2024-10-17
pool: update the runahead limit after task removal

### LINKED ISSUE #4983: data store: blocked tasks outside of n=1 are missing from the store
This applies to Cylc 7 compat mode, regular Cylc 8 workflows are protected from this by required outputs.

### Explanation

Unsatisfiable tasks will become invisible in the UIs if they drift outside of the N=1 window.
* For regular Cylc 8 workflows the cause of the problem is an incomplete task which will stay in the N=0 window (I think) preventing blocked downstream tasks from leaving the N=1 window. Consequently the blocked tasks remain visible.
* However, in Cylc 7 compat mode we don't have required outputs to save us so if the blocked task(s) drift outside the N=1 window they become invisible to the user making it hard to work out what's going wrong.

### Example

This example is missing the suicide trigger which would remove the `recover` task:

```ini
# suite.rc
[scheduler]
    allow implicit tasks = True

[scheduling]
    initial cycle point = 1
    cycling mode = integer
    [[graph]]
        P1 = """
            foo:fail => recover
            foo | recover => bar
        """
```

When we run this workflow the first 5 cycles run as normal, the sixth cycle is runahead limited.

The blocked "recover" tasks from the first five cycles are not in the N=1 window (note no inter-cycle dependencies here) so disappear from the UIs:

```
foo - running
TUI is experimental and may break with large flows

- foo
   - ̊○ 6
        ○ bar      # ideally we would mark this task as runahead limited too?
        ̊○ foo
        ○ recover  # ideally we would mark this task as runahead limited too?
```

**Pull requests welcome!**

oliver-sanders:
* Update the runahead limit when tasks are removed (i.e. by suicide trigger) to prevent erroneous stall.
* Also closes #4983 (by adding a test for it)

Whilst writing the test, I noticed the n-window values for the "bar" tasks look wrong to me, although that's a different issue.

**Check List**

### DIFF
--- a/changes.d/4983.fix.md
+++ b/changes.d/4983.fix.md
@@ -1,1 +1,1 @@
-Ensure the runahead limit is recomputed when legacy "suicide-triggers" are used to prevent erroneous stall in niche cases.
+Ensure the runahead limit is recomputed when legacy "suicide-triggers" are used, to prevent erroneous stall in niche cases.
--- a/changes.d/4983.fix.md
+++ b/changes.d/4983.fix.md
@@ -1,0 +1,1 @@
+Ensure the runahead limit is recomputed when legacy "suicide-triggers" are used to prevent erroneous stall in niche cases.
--- a/tests/integration/test_compat_mode.py
+++ b/tests/integration/test_compat_mode.py
@@ -1,0 +1,125 @@
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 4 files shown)


changes.d/4983.fix.md
```diff
@@ -0,0 +1 @@
+Ensure the runahead limit is recomputed when legacy "suicide-triggers" are used to prevent erroneous stall in niche cases.
```
hjoliver, line 1:
```suggestion
Ensure the runahead limit is recomputed when legacy "suicide-triggers" are used, to prevent erroneous stall in niche cases.
```
(A bit ambiguous without the comma)

hjoliver (approved):

hjoliver:
> Whilst writing the test, I noticed the n-window values for the "bar" tasks look wrong to me, although that's a different issue.

Needs an issue?

oliver-sanders:
> Needs an issue?

I think so, just checking that it's wrong before doing so.

tests/integration/test_compat_mode.py
```diff
@@ -0,0 +1,125 @@
...
+                            foo:fail => recover
+                            foo | recover => bar
+                        ''',
+                    },
+                },
+            },
+        },
+        filename='suite.rc',
+    )
+    schd: 'Scheduler' = scheduler(id_, paused_start=False, debug=True)
```
wxtim, line 55: I'm a little surprised that `debug=True` isn't the default. Is there a reason why it's not?

wxtim (approved):
✅

PR #6431, opened 2024-10-17
fix CPF wall_clock offset precision constraint

### LINKED ISSUE #6423: xtrigger wall_clock offset trimmed to `cycle point format`
### Description

The xtrigger `wall_clock` offset is trimmed to precision of `cycle point format`

### Reproducible Example
```
[scheduler]
    cycle point format = CCYYMMDDThh
    allow implicit tasks = True
[scheduling]
    initial cycle point = 20241017T00
    [[xtriggers]]
        clock_PT25M = wall_clock(offset=PT25M):PT20S
    [[graph]]
        T-00 = """
@clock_PT25M => a
"""

[runtime]
    [[root]]
        script = sleep $((1 + $RANDOM % 10))
```
image

### Expected Behaviour

The `wall_clock` offset should not be restricted by the cycle point precision

image

dwsutherland:
closes #6423 

```
[scheduler]
    cycle point format = CCYYMMDDThh
    allow implicit tasks = True
[scheduling]
    initial cycle point = 20241017T22
    [[xtriggers]]
        clock_PT25M31S = wall_clock(offset=PT25M31S):PT20S
    [[graph]]
        T-00 = """
@clock_PT25M31S => a
"""
[runtime]
    [[root]]
        script = sleep $((1 + $RANDOM % 10))
```

image

**Check List**

- [x] Tests added.

### DIFF
--- a/changes.d/6431.fix.md
+++ b/changes.d/6431.fix.md
@@ -1,0 +1,1 @@
+The `cycle point format` was imposing an undesirable constraint on `wall_clock` offsets, this has been fixed.
--- a/cylc/flow/task_proxy.py
+++ b/cylc/flow/task_proxy.py
@@ -48,7 +48,6 @@ from cylc.flow.wallclock import get_unix_time_from_time_string as str2time
 from cylc.flow.cycling.iso8601 import (
     point_parse,
     interval_parse,
-    ISO8601Interval
 )
 
 if TYPE_CHECKING:
@@ -424,14 +423,18 @@ class TaskProxy:
         """
         offset_str = offset_str if offset_str else 'P0Y'
         if offset_str not in self.clock_trigger_times:
+            # Convert ISO8601Point into metomi-isodatetime TimePoint at full
+            # second precision (N.B. it still dumps at the same precision
+            # as workflow cycle point format):
+            point_time = point_parse(str(point))
             if offset_str == 'P0Y':
-                trigger_time = point
+                trigger_time = point_time
             else:
-                trigger_time = point + ISO8601Interval(offset_str)
+                trigger_time = point_time + interval_parse(offset_str)
 
-            offset = int(
-                point_parse(str(trigger_time)).seconds_since_unix_epoch)
-            self.clock_trigger_times[offset_str] = offset
+            self.clock_trigger_times[offset_str] = int(
+                trigger_time.seconds_since_unix_epoch
+            )
         return self.clock_trigger_times[offset_str]
 
     def get_try_num(self):
--- a/tests/integration/test_xtrigger_mgr.py
+++ b/tests/integration/test_xtrigger_mgr.py
@@ -29,30 +29,38 @@ def get_task_ids(schd: Scheduler) -> Set[str]:
 
 
 async def test_2_xtriggers(flow, start, scheduler, monkeypatch):
-    """Test that if an itask has 2 wall_clock triggers with different
-    offsets that xtrigger manager gets both of them.
+    """Test that if an itask has 4 wall_clock triggers with different
+    offsets that xtrigger manager gets all of them.
 
     https://github.com/cylc/cylc-flow/issues/5783
 
     n.b. Clock 3 exists to check the memoization path is followed,
     and causing this test to give greater coverage.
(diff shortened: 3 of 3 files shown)


hjoliver:
>  [x] Existing tests cover.

Presumably not, as we've been living with the bug?

dwsutherland:
> > [x] Existing tests cover.
> 
> Presumably not, as we've been living with the bug?

Well coverage is 100%, and trigger offsets are tested ... But I could add a test that has CPF at lower resolution than offset

hjoliver:
Yeah worth doing that I think.

dwsutherland:
> Yeah worth doing that I think.

Good thing I tried.. Found a way to break my solution, and have been working on various solutions all afternoon..
Will have it up by this evening.

changes.d/6431.fix.md
```diff
@@ -0,0 +1 @@
+The `cycle point format` was imposing an undesirable constraint on `wall_clock` offsets, this has been fixed.
```
oliver-sanders, line 1:
```suggestion
Allowed `wall_clock` offsets to use smaller precision than the `cycle point format` (e.g. `wall_clock(offset="PT1M")` in combination with `cycle point format = CCYY`).
```

dwsutherland: Well technically they were capable of it before, but some change broke it.

cylc/flow/task_proxy.py
```diff
@@ -424,14 +423,20 @@ def get_clock_trigger_time(
...
-            offset = int(
-                point_parse(str(trigger_time)).seconds_since_unix_epoch)
-            self.clock_trigger_times[offset_str] = offset
+                trigger_time = (
+                    point_parse(
+                        point_time.__str__(strftime_format='%Y%m%dT%H%M%S'),
+                        '%Y%m%dT%H%M%S'
+                    )
+                    + interval_parse(offset_str)
+                )
```
MetRonnie, line 436: There is no need to mess with the strftime_format:
```suggestion
            # Convert ISO8601Point into metomi-isodatetime TimePoint at full
            # second precision (N.B. it still dumps at the same precision
            # as workflow cycle point format):
            point_time = point_parse(str(point))
            if offset_str == 'P0Y':
                trigger_time = point_time
            else:
                trigger_time = point_time + interval_parse(offset_str)
```

cylc/flow/cycling/iso8601.py
MetRonnie, line 1: There is no need to change this file actually, see above comment

MetRonnie (changes requested):
Tested 👍 Can actually be even simpler

cylc/flow/task_proxy.py
dwsutherland: No way! shocking .. Just tested this and it worked..

Ah I think I know what the issue was ... I was originally converting these objects back to `ISO8601Point` and `ISO8601Interval` adding them together then parsing them again but you get a dump format  and epoc seconds the same as the CPF (trimmed to that precision)...
i.e. if you have `CCYY-MM-DD` the addition will trim off the minutes and seconds.. Hence why I returned a new object with `%Y%m%dT%H%M%S` precision.

And as the TimePoint and Duration objects also have a dump format I assumed they behaved the same. Doh!

Reverting things to this.. Thanks

cylc/flow/cycling/iso8601.py
dwsutherland: Noted above!

cylc/flow/task_proxy.py
dwsutherland: (p.s. sorry, should have merged yours in first then rebased)

hjoliver (approved):

changes.d/6431.fix.md
hjoliver: Approved, but I prefer @oliver-sanders wording here, in terms of clarity for users.

cylc/flow/task_proxy.py
MetRonnie: Yeah I stuck a breakpoint in there and got confused by the dump format precision for a while 😄 Hopefully the comment I included makes it clear for people debugging in the future

MetRonnie (approved):

PR #6440, opened 2024-10-22
Modernise 'cylc dump -t' print format.

hjoliver:
This has been annoying me forever.

Master 🤮: 
```
$ cylc dump -t bug --flows
foo, 100, running, not-held, not-queued, not-runahead, [1]
foo, 101, waiting, not-held, not-queued, runahead, [1]
foo, 96, running, not-held, not-queued, not-runahead, [1]
foo, 97, running, not-held, not-queued, not-runahead, [1]
foo, 98, running, not-held, not-queued, not-runahead, [1]
foo, 99, running, not-held, not-queued, not-runahead, [1]
```

This branch:
```
$ cylc dump -t bug --flows
91/foo:running flows=[1]
92/foo:running flows=[1]
93/foo:running flows=[1]
94/foo:running flows=[1]
95/foo:running flows=[1]
96/foo:waiting (runahead) flows=[1]
```

**QUESTION**: is it OK to make the new format the default? 
(As I have done, with `-l/--legacy` to generate the old format)

**Check List**

### DIFF
--- a/changes.d/6440.feat.md
+++ b/changes.d/6440.feat.md
@@ -1,0 +1,1 @@
+The "cylc dump" command now prints task IDs. Use "--legacy" if you need the old format.
--- a/cylc/flow/scripts/dump.py
+++ b/cylc/flow/scripts/dump.py
@@ -163,6 +163,9 @@ def get_option_parser():
         "-t", "--tasks", help="Task states only.",
         action="store_const", const="tasks", dest="disp_form")
     parser.add_option(
+        "-l", "--legacy", help="Tasks states only; use legacy format.",
+        action="store_true", default=False, dest="legacy_format")
+    parser.add_option(
         "-f", "--flows", help="Print flow numbers with tasks.",
         action="store_true", default=False, dest="show_flows")
     parser.add_option(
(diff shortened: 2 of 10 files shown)


oliver-sanders (approved):
LGTM

dwsutherland (approved):
Looks good.. We adding a legacy format to avoid changing all the tests? 
I guess someone might like it (or use it) 😆

hjoliver:
> Looks good.. We adding a legacy format to avoid changing all the tests? I guess someone might like it (or use it) 😆

Not so much adding it, but keeping it just in case anyone is relying on it. (Not trying to avoid changing the tests, honest!)

dwsutherland:
I don't know how the patch coverage has been calculated? I wouldn't have thought there would be many indirect losses from this change...

hjoliver:
From a quick look on Codecov, I don't get it:

Screenshot (17)

The yellow ("partial") lines are covered by my integration test, and the red (not-covered) lines are not changed on this branch.

oliver-sanders:
To turn those yellow lines green we need a test that makes the `if` fall in either direction, e.g. at least one task which is held and one task which is not held.

hjoliver:
At first glance I thought it was odd because those `if` blocks don't have an `else` clause. But the do-nothing option does affect the following logic of course - so that explains it, thanks!  I'll extend the tests...

hjoliver:
Patch coverage now 100% 🎉

PR #6445, opened 2024-10-23
Ensure `cylc trigger` does not fall back to `flow=none` by default

MetRonnie:
## Summary

Split off from #6370 as I think it's already a bug.

### Problem 1

If you `cylc trigger` a task when there are no active flows (e.g. on restarting a completed workflow), it looks in the DB for the latest flow number(s). However, if the latest was `flow=none`, this would result in the task being triggered with `flow=none` as well.

**Solution:** choose the latest flow number(s) that is not `flow=none`.

### Problem 2

With #6370 making it so that `cylc remove` will demote tasks to `flow=none`, you could end up with the situation where all tasks in the DB are `flow=none`, so again this would still result in the task being triggered with `flow=none`.

**Solution:** fall back to `flow=1`.

## Check List

- [x] No dependency changes 
- [x] Tests are included 
- [x] No docs needed

### DIFF
--- a/changes.d/6445.fix.md
+++ b/changes.d/6445.fix.md
@@ -1,0 +1,1 @@
+Ensure `cylc trigger` does not fall back to `flow=none` when there are no active flows.
--- a/tests/unit/test_task_pool.py
+++ b/tests/unit/test_task_pool.py
@@ -13,3 +13,45 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from typing import List
+from unittest.mock import Mock
+
+import pytest
+
+from cylc.flow.flow_mgr import FlowNums
+from cylc.flow.task_pool import TaskPool
+
+
+@pytest.mark.parametrize('pool, db_fnums, expected', [
+    pytest.param(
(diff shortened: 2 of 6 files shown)


changes.d/6445.fix.md
```diff
@@ -0,0 +1 @@
+Ensure `cylc trigger` does not fall back to `flow=none` in niche circumstances.
```
hjoliver, line 1:
```suggestion
Ensure `cylc trigger` does not fall back to `flow=none` when there are no active flows.

**Review by hjoliver — APPROVED** [review]
Good 👍

File: tests/unit/test_task_pool.py [hunk 2]
```diff
@@ -13,3 +13,45 @@
...
+        {1, 2, 3},
+        id="all-active"
+    ),
+    pytest.param(
+        [set(), set()],
+        {5, 6},
+        {5, 6},
+        id="from-db"
+    ),
+    pytest.param(
```
oliver-sanders, line 40:
```suggestion
    pytest.param(
        # see https://github.com/cylc/cylc-flow/pull/6445
```

oliver-sanders (approved):
Makes sense, looks good, :+1:

PR #6446, opened 2024-10-23
graphdepth update on remove fix

### LINKED ISSUE #6437: n-window calculation error
This example seems to produce an erroneous n-window when run in Cylc 7 compat mode (not tested outside of compat mode yet):

```
    foo:fail => recover
    foo | recover => bar
```

When run, this example stalls after a few cycles, due to unsatisfyable `recover` tasks (because `foo` never fails).

These `recover` tasks are the only tasks in the pool (i.e. the only `n=0` tasks), yet the `bar` tasks are listed as `n=0` despite having run and completed?

dwsutherland:
closes #6437 

**Check List**

- [x] Existing covering tests adjusted.

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -1136,6 +1136,7 @@ class DataStoreMgr:
             self.prune_flagged_nodes.update(
                 self.n_window_node_walks[tp_id]['walk_ids']
             )
+        self.update_window_depths = True
         self.updates_pending = True
 
     def add_pool_node(self, name, point):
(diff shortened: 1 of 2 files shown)


tests/integration/test_compat_mode.py
```diff
@@ -75,13 +75,13 @@ async def test_blocked_tasks_in_n0(flow, scheduler, run, complete):
         } == {
             ('1', 1, 'foo'),
             ('1', 0, 'recover'),
-            ('1', 0, 'bar'),  # NOTE: this should be n=2 not n=0
```
dwsutherland, line 78: This should not have been `n=2` it is one edge away from active task `recover`..

hjoliver (approved):
That's got it, nice.

cylc/flow/data_store_mgr.py
```diff
@@ -1136,6 +1136,7 @@ def remove_pool_node(self, name, point):
             self.prune_flagged_nodes.update(
                 self.n_window_node_walks[tp_id]['walk_ids']
             )
+        self.update_window_depths = True
```
oliver-sanders, line 1139: :tada:

oliver-sanders (approved):

PR #6454, opened 2024-10-29
Remove `UTC mode` from tutorial workflows

wxtim:
`UTC mode` is a legacy setting which isn't required any-more for tutorial workflows.

Companion of https://github.com/cylc/cylc-doc/pull/769

**Check List**

- [x] Tests are **not** included should be tested by the validate script.
- [x] Changelog entry *not* included - very small change
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/0769.
- [x] PR raised against bug fix branch because this is a doc-fix and not a feature.

### DIFF
--- a/cylc/flow/etc/tutorial/consolidation-tutorial/flow.cylc
+++ b/cylc/flow/etc/tutorial/consolidation-tutorial/flow.cylc
@@ -1,7 +1,4 @@
 #!jinja2
-[scheduler]
-    UTC mode = True
-
 [scheduling]
     # Start the workflow 7 hours before now ignoring minutes and seconds
     # * previous(T-00) takes the current time ignoring minutes and seconds.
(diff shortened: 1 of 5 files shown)


wxtim:
> I think the `UTC mode` default is still `False`?
> 
> [https://cylc.github.io/cylc-doc/stable/html/reference/config/global.html#global.cylc[scheduler]UTC%20mode](https://cylc.github.io/cylc-doc/stable/html/reference/config/global.html#global.cylc%5Bscheduler%5DUTC%20mode)

It is, but it no longer has any effect on the `cycle point time zone` [Old docs](https://cylc.github.io/cylc-doc/7.8.8/html/appendices/suiterc-config-ref.html?highlight=utc%20mode#cylc-cycle-point-time-zone) so it doesn't matter any more.

oliver-sanders (approved):
Ok, if it's just logging.

Make sure it doesn't appear in any examples in cylc-doc if removing.

markgrahamdawson (approved):

PR #6039, opened 2024-03-27
Skip Mode

wxtim:
Closes #5641
closes #5975
closes #5820 

[Skip Mode Proposal Doc](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-skip-mode.md#proposal-skip-run-mode-for-tasks)

This branch includes (marked against skip mode proposal):

1. A new task run mode called "skip" to sit alongside the existing "live" and "simulation" and "dummy" modes.
2. Skip mode settings in `[runtime][\<namespace\>][skip].
3. Task-level control of the run mode - allowing run mode of tasks to override that of the workflow using `[runtime][<namespace>]run mode`.
    a. Broadcast can change `run mode` for future task job submissions.
    b. Cylc Validate and lint will warn about the setting not being live.
4. `cylc set --out skip` sets outputs from skip mode.
5. Tests to ensure that `run mode = skip` respects `is_held` flag.
6. Tests to ensure that force triggering a task will not override the run mode.

Extras
7. Run Mode is available as an task attribute in the UI
8. When tasks are run in skip mode, the prerequisites which correspond to the outputs they generate should be marked as satisfied by skip mode rather than satisfied naturally for provenance reasons. For the purpose of cylc remove logic, satisfied by skip mode should be treated the same as satisfied naturally.

There are two extensions, which I haven't dealt with yet, because I want to ensure that the basic functionality works, and move to the substantial documentation PR which need follow this.

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at [cylc/cylc-doc/pull/XXXX](https://github.com/cylc/cylc-doc/pull/755).
- [x] Feature: PR against master

PR #6457, opened 2024-11-04
Refactor task killing & simplify task state deltas

MetRonnie:
In preparation for tackling the last piece of the [`cylc remove` proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-remove.md#proposal):

> When all flows are removed from active tasks, Cylc should first attempt to kill the job as it serves no further purpose and may cause resource contention issues if left running, however, failure of the kill command should be tolerated, though logged.

I extracted and de-deduped the kill mechanism into a scheduler method.

Also, it seems to me the data store manager's `delta_task_state()` method should include `is_held`, `is_queued` and `is_runahead`.

**Check List**

- [x] No dependency changes 
- [x] Changelog entry not needed as refactor.
- [x] Docs not needed


cylc/flow/scheduler.py
```diff
@@ -1010,6 +1014,42 @@ def _set_stop(self, stop_mode: Optional[StopMode] = None) -> None:
...
+    ) -> int:
+        """Kill tasks if they are in a killable state.
+
+        Args:
+            itasks: Tasks to kill.
+            warn: Whether to warn about tasks that are not in a killable state.
+
+        Returns number of tasks that could not be killed.
+        """
+        jobless = self.get_run_mode() == RunMode.SIMULATION
```
MetRonnie, line 1028: (Need to follow-up with skip mode #6039)

tests/integration/tui/screenshots/test_task_states.filter-not-waiting-or-expired.html
```diff
@@ -8,13 +8,13 @@
...
 <span style="color:#000000;background:#e5e5e5">         </span><span style="color:#000000;background:#e5e5e5">-</span><span style="color:#000000;background:#e5e5e5"> </span><span style="color:#000000;background:#e5e5e5">̿⊗ Y                                                                  </span>
 <span style="color:#000000;background:#e5e5e5">            </span><span style="color:#000000;background:#e5e5e5"> </span><span style="color:#000000;background:#e5e5e5"> </span><span style="color:#000000;background:#e5e5e5">̿⊗ b                                                               </span>
-<span style="color:#000000;background:#e5e5e5">      </span><span style="color:#000000;background:#e5e5e5">-</span><span style="color:#000000;background:#e5e5e5"> </span><span style="color:#000000;background:#e5e5e5">̿⊘ 2                                                                     </span>
+<span style="color:#000000;background:#e5e5e5">      </span><span style="color:#000000;background:#e5e5e5">-</span><span style="color:#000000;background:#e5e5e5"> </span><span style="color:#000000;background:#e5e5e5">̎⊘ 2                                                                     </span>
```
MetRonnie, line 11: The reason for the change in icon is that previously, the held state was not updated in the data store by the test fixture: https://github.com/cylc/cylc-flow/blob/478b0d0a77efff25d27fa9ac621281df25921d22/tests/integration/tui/test_app.py#L34-L44

cylc/flow/task_events_mgr.py
```diff
@@ -1322,7 +1322,6 @@ def _process_message_failed(self, itask, event_time, message, forced):
             no_retries = True
             if itask.state_reset(TASK_STATUS_FAILED, forced=forced):
                 self.setup_event_handlers(itask, self.EVENT_FAILED, message)
-                self.data_store_mgr.delta_task_state(itask)
```
MetRonnie, line 1325: Duplicate of L1328 just below

cylc/flow/data_store_mgr.py
```diff
@@ -2291,10 +2291,11 @@ def delta_task_state(self, itask):
         update_time = time()
 
         # update task instance
-        tp_delta = self.updated[TASK_PROXIES].setdefault(
+        tp_delta: PbTaskProxy = self.updated[TASK_PROXIES].setdefault(
             tp_id, PbTaskProxy(id=tp_id))
         tp_delta.stamp = f'{tp_id}@{update_time}'
         tp_delta.state = itask.state.status
+        tp_delta.is_held = itask.state.is_held
```
oliver-sanders, line 2298: Currently:

| attr | updater |
| --- | --- |
| status | delta_task_state |
| is_held | delta_task_held |
| is_queued | delta_task_queued |
| is_runahead | delta_task_runahead |

I'm not sure why we would move `is_held` in with `delta_task_state`?

cylc/flow/scheduler.py
```diff
@@ -1010,6 +1014,42 @@ def _set_stop(self, stop_mode: Optional[StopMode] = None) -> None:
...
+        unkillable: List[TaskProxy] = []
+        for itask in itasks:
+            if itask.state(*TASK_STATUSES_ACTIVE):
+                itask.state_reset(
+                    # directly reset to failed in sim mode, else let
+                    # task_job_mgr handle it
+                    status=(TASK_STATUS_FAILED if jobless else None),
+                    is_held=True,
+                )
+                self.data_store_mgr.delta_task_state(itask)
```
oliver-sanders, line 1039:
```suggestion
                self.data_store_mgr.delta_task_state(itask)
                self.data_store_mgr.delta_task_held(itask)
```

cylc/flow/data_store_mgr.py
MetRonnie: Tome it looks like it makes sense to move all into `delta_task_state`? These are all handled by 1 method on `TaskProxy`, so it would reduce chances of forgetting to update them all in the data store

cylc/flow/scheduler.py
MetRonnie: Not needed if we include `itask.is_held` in `delta_task_state()`

cylc/flow/data_store_mgr.py
oliver-sanders: We could do it all in one, but I think it would be *less* (not more) efficient for the data store to do this all at once as we would be pushing out all four attributes every time any one of them changes.

MetRonnie: If that's a concern, could do
```python
if tproxy.is_held != itask.state.is_held:
    tp_delta.is_held = itask.state.is_held
```
etc.

Or drop the commit as I've separated it out into its own commit now.

oliver-sanders: Sounds good. I think, either we leave as is, or we combine all the related attrs and compare to previous value as you suggested.

dwsutherland (approved):
Approval conditional on your option of doing this:

```
if tproxy.is_held != itask.state.is_held:
    tp_delta.is_held = itask.state.is_held
```

And you'd want to do the same with queue and runahead ..etc 
Also check/do nothing if nothing has changed (unless that's impossible upstream)..

It's only if a field is set that it will be sent as a delta, so if we can avoid sending a field, that's ideal.

MetRonnie:
@dwsutherland I noticed that if you inspect `tp_delta.is_held` it already as a value (`True`/`False`) at time of creation of the `PbTaskProxy` object. However, unless you proactively set a value on it, it does not get included in the delta received at the UI. What sorcery is this? 😅

dwsutherland:
> @dwsutherland I noticed that if you inspect `tp_delta.is_held` it already as a value (`True`/`False`) at time of creation of the `PbTaskProxy` object. However, unless you proactively set a value on it, it does not get included in the delta received at the UI. What sorcery is this? 😅

Yeah, it has a default.. 
But I think it has some sort of flag to know if it's been set, so it's not serialized/sent otherwise...
Then there are some protobuf methods between protobuf object <=> data-store/delta-store, and null-stripping thereafter if flagged in the graphql request.

oliver-sanders (approved):
:+1:, one comment, address or ignore.

cylc/flow/scheduler.py
```diff
@@ -1010,6 +1014,42 @@ def _set_stop(self, stop_mode: Optional[StopMode] = None) -> None:
         self.stop_mode = stop_mode
         self.update_data_store()
 
+    def kill_tasks(
+        self, itasks: 'Iterable[TaskProxy]', warn: bool = True
+    ) -> int:
+        """Kill tasks if they are in a killable state.
```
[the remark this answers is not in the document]
MetRonnie: It needs the workflow run mode, which is not currently available in `task_job_mgr`. However, once #6039 is in, the run mode will be a property of `TaskProxy`s, so could be moved there afterwards?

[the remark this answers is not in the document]
oliver-sanders: Makes sense.

PR #6472, opened 2024-11-11
Implement `cylc remove` proposal

### LINKED ISSUE #5643: remove: extend beyond the task pool

MetRonnie:
Closes #5643. Supersedes #6370.

### Summary

This fully implements [the "Cylc Remove Extension" proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-remove.md).

#### Flow numbers

`cylc remove` now has a `--flow` option for removing a task from specific flows.

If not used, it will remove the task from all flows that it belongs to.

If the removed task is active/waiting, if it is removed from a subset of flows that it belongs to, it will remain in the task pool; if it is removed from all flows that it belongs to, it will be removed from the task pool (as is the current behaviour).

If a task is removed from all flows that it belongs to, it will become a no-flow task (`flow=None`).

For ease of reviewing, you can use my UI branch that displays flow numbers: https://github.com/MetRonnie/cylc-ui/tree/flow-nums [^c].

[^c]: Waiting tasks that are not yet in the pool have greyed out flow numbers at the moment.

#### Historical tasks

`cylc remove` now can remove tasks that are no longer active, making it look like they never ran. It does this by removing the task from the specified flows in the database (in the `task_states` and `task_outputs` tables)[^a], and un-setting any prerequisites of active tasks that the removed task had naturally satisfied[^b]. If a task is removed from all flows that it belongs to, a no-flow task is left in the DB for provenance.

The above also applies to active/waiting tasks that `cylc remove` is used on.

[^a]: If removing flows would result in two rows in the DB no longer being unique, the SQLite [`UPDATE OR REPLACE`](https://www.sqlite.org/lang_update.html) statement is used, so the first entry will be removed and the most recent entry will remain.
[^b]: Prerequisites manually satisfied by `cylc set --pre` are not affected by `cylc remove`.

#### Kill submitted/running tasks

Using `cylc remove` on a submitted/running task will now kill it if you are removing the task from all flows that it belongs to.

Unlike with `cylc kill`, downstream tasks will not spawn off the `:fail` or `:submit-fail` outputs as the task is in `flow=none`, and also the `failed` and `submission failed` handlers will not run.

### Check List

- [x] No dependency changes
- [x] Tests are included 


cylc/flow/scripts/remove.py
```diff
@@ -18,7 +18,27 @@
 
 """cylc remove [OPTIONS] ARGS
 
-Remove one or more task instances from a running workflow.
+Remove tasks from a running workflow and the workflow's history.
```
hjoliver, line 21: IMO as a headline "remove a task from a running workflow" is still pretty ambiguous and kind of harks back to the old task-pool focused `cylc remove` . Can we can focus on the essential new functionality?
```suggestion
Erase the run-history of tasks, allowing them to run again in the same flow.
```

cylc/flow/scripts/remove.py
```diff
@@ -18,7 +18,27 @@
...
+
+By default, the specified task(s) will be removed from all flows, making it
+as if the task(s) never ran. The task(s) and their outputs will be left in the
+`None` flow to preserve a record that they ran, however.
+
+If you specify flow(s) to remove a task from (not all flows that it belongs to)
+then the task will still exist in the remaining flow(s), but it will not
+influence the evolution of the workflow in the specified flow(s).
+
+Removing a submitted or running task will also kill it (see "cylc kill").
```
hjoliver, line 31: This mixes up the default and the specific a bit: you can remove a task from all flows explicitly as well as by default. How about this:

```suggestion
By default, the specified task(s) will be removed from all flows.

Tasks removed from all flows, and any downstream tasks spawned by
their outputs, will be recorded in the `none` flow and will not affect
the evolution of the workflow.

If you remove a task from some but not all of its flows, it will still exist
in the remaining flows, but it will not affect the evolution of the
workflow in the removed flows.

Removing a submitted or running task will also kill it (see "cylc kill").
```

cylc/flow/scripts/remove.py
```diff
@@ -18,7 +18,27 @@
...
+
+Examples:
+  # Remove a task that already ran.
+  # (Any downstream tasks that are already running or finished will be
+  # left alone. The task and its outputs will be left in the None flow)
```
hjoliver, line 36: The bit in parentheses is quite important and should probably be in the main help text where it mentions downstream tasks.

However, this relates to my question on the scheduler module changes above. Why do we not kill downstream tasks that are already running, if they end up with no flow numbers? 

From a quick look back at the proposal it doesn't seem to explicitly address this for downstream tasks:

> If removing a task causes all of the prerequisites on a downstream task to be unset (i.e. if the downstream task was spawned as a result of outputs from this task alone) then the downstream task shall be removed from the pool

.., but the general kill statement would seem to apply equally well to downstream tasks too:

> When all flows are removed from active tasks, Cylc should first attempt to kill the job as it serves no further purpose and may cause resource contention issues if left running

Might be a question for @oliver-sanders?

hjoliver (commented):
Still part way through, looks great so far.

cylc/flow/platforms.py
```diff
@@ -265,6 +265,12 @@ def platform_from_name(
             platform_data['name'] = platform_name
             return platform_data
 
+    # If platform name in run mode and not otherwise defined:
+    if platform_name == 'SIMULATION':
+        platform_data = deepcopy(platforms['localhost'])
+        platform_data['name'] = 'localhost'
+        return platform_data
+
```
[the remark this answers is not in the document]
oliver-sanders: Yes and no, the platform wasn't being written to the DB before, now it is, as a consequence this check is now required.

Ronnie hasn't dotted the t's and crossed the i's as this will change with Tim's skip-mode PR.

cylc/flow/task_state.py
```diff
@@ -527,3 +527,12 @@ def get_unsatisfied_prerequisites(self) -> List['PrereqMessage']:
             for prereq in self.prerequisites if not prereq.is_satisfied()
             for key, satisfied in prereq.items() if not satisfied
         ]
+
+    def any_satisfied_prerequisite_tasks(self) -> bool:
+        """Return True if any of this task's prerequisite tasks are
+        satisfied."""
```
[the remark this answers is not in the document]
MetRonnie: I am using the term "prerequisite tasks" to avoid confusion with `Prerequisite`s which can comprise multiple prerequisite tasks.

cylc/flow/scripts/remove.py
MetRonnie: Ok that reads good to me. What does @oliver-sanders think?

cylc/flow/task_state.py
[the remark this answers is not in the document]
oliver-sanders: Try `Prerequisite outputs".

cylc/flow/prerequisite.py
```diff
@@ -268,7 +273,10 @@ def satisfy_me(self, outputs: Iterable['Tokens']) -> 'Set[Tokens]':
             if prereq not in self._satisfied:
                 continue
             valid.add(output)
-            self[prereq] = 'satisfied naturally'
+            if self._satisfied[prereq] != 'satisfied naturally':
+                self[prereq] = (
+                    'force satisfied' if forced else 'satisfied naturally'
```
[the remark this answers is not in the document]
MetRonnie: I don't think this is worth it, we would lose backwards compatibility

cylc/flow/prerequisite.py
```diff
@@ -345,3 +353,13 @@ def get_resolved_dependencies(self) -> List[str]:
             for msg, satisfied in self._satisfied.items()
             if satisfied
         ]
+
+    def unset_naturally_satisfied_dependency(self, id_: str) -> bool:
+        """Set the matching dependency to unsatisfied and return True only if
+        it was naturally satisfied."""
```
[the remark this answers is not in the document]
MetRonnie: Hmm, I think referring to prerequisites within a class called `Prerequisite` to be somewhat confusing

cylc/flow/prerequisite.py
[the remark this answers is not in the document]
oliver-sanders: Not related to this PR, one for another day.

cylc/flow/scheduler.py
```diff
@@ -1050,6 +1049,133 @@ def kill_tasks(
...
+        matched_task_ids = {
+            *removed.keys(),
+            *(quick_relative_id(cycle, task) for task, cycle in inactive),
+        }
+
+        for id_ in matched_task_ids:
+            point_str, name = id_.split('/', 1)
+            tdef = self.config.taskdefs[name]
+            # Go through downstream tasks to see if any need to stand down
+            # as a result of this task being removed:
```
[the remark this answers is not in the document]
MetRonnie: There are 2 things happening in this loop:
1. Go through downstream tasks to see if any need to stand down
2. Remove the matched task in the DB

Both of these comments come right before what they say they are doing.

As for both removed active tasks and matched inactive ones... the removed active tasks are matched by definition, no?

cylc/flow/scheduler.py
```diff
@@ -1050,6 +1049,133 @@ def kill_tasks(
...
+                self.pool.unqueue_task(child_itask)
+                # Check if downstream task should remain spawned:
+                if (
+                    # Ignoring tasks we are already dealing with:
+                    child_itask.identity in matched_task_ids
+                    or child_itask.state.any_satisfied_prerequisite_tasks()
+                ):
+                    continue
+                # No longer has reason to be in pool:
+                self.pool.remove(child_itask, 'prerequisite task(s) removed')
```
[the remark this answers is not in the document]
MetRonnie: No - https://github.com/cylc/cylc-flow/pull/6472#discussion_r1837879110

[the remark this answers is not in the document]
hjoliver: Agreed, on reading the comment below.

cylc/flow/scripts/remove.py
hjoliver: Maybe ambiguous was the wrong word. 

It's more that I think exactly what it means to "remove tasks from a running workflow" won't be at all obvious to users (e.g. remove tasks from the graph?, remove upcoming tasks - like pre-spawned ones in Cylc 7 - so they won't run? or what?) and it doesn't really suggest the important new history erasure aspect. 

Actually we should add this (or similar) to the main text as well:
```
This command can also be used to remove final status incomplete tasks from the n=0 active window, to prevent or recover from a stall, if they don't need to be completed to continue to the workflow.
```

I think "erase the run-history of tasks" can be viewed as covering all of the above including the task pool operations, if you think of "run-history" as all of task evolution at run time, not just jobs.

cylc/flow/scripts/remove.py
hjoliver: OK makes sense.  Can we amend the proposal to record this, to avoid having to think through it again in the future?

cylc/flow/task_state.py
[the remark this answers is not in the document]
hjoliver: > I am using the term "prerequisite tasks" to avoid confusion with ...

Yes, but tasks don't get "satisfied", prerequisites do. Oliver's suggestion would probably be better.

cylc/flow/prerequisite.py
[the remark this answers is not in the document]
hjoliver: Agreed on that confusing situation, historical reasons, probably my bad. Maybe we should rename the Prerequisite class. Not really a probably for this PR though. Let's just leave it as-is for now, not super-important.

cylc/flow/scheduler.py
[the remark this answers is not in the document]
hjoliver: Fair enough. I guess I was thinking of "stand down", which is not exactly standard terminology for us, as covering both 1. and 2.  

> As for both removed active tasks and matched inactive ones... the removed active tasks are matched by definition, no?

The problem I was trying to address is that I didn't initially see that this handles both active and inactive tasks, because the first loop handles only active tasks. 

I was probably just being dense and making assumptions based on the fact that we quite a few methods that handle active and inactive tasks separately one after the other, but maybe you could just add a slightly more comprehensive comment or two yourself? No biggie though.

cylc/flow/prerequisite.py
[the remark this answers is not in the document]
hjoliver: OK, I forgot we already had both terms in master.  That's mildly unfortunate, but probably not worth fixing.

hjoliver:
(Did some testing today, so far so good, but I'm not done yet.)

cylc/flow/scripts/remove.py
oliver-sanders: Done.

cylc/flow/scripts/remove.py
oliver-sanders: > I think "erase the run-history of tasks" can be viewed as covering the task pool operations too?

I don't think that is obvious personally as it's mixing task spawning (internal concept) with run history (which means job submissions to users).

oliver-sanders:
This PR removes flow nums from the `task_outputs` and `task_states` table which is functionally sufficient.

However, it doesn't remove flow nums from the `task_jobs` table. Removing them here doesn't appear to be functionally required, so no problem. However, I think this table is used to populate "cylc-review", so we might want to remove them here too, possibly as follow-on?

MetRonnie:
> However, it doesn't remove flow nums from the `task_jobs` table.

(It does, but only for submitted/running tasks that get killed by `cylc remove`)

cylc/flow/scheduler.py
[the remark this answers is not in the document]
MetRonnie: Rewritten somewhat, continue this discussion in https://github.com/MetRonnie/cylc-flow/pull/34

cylc/flow/scripts/remove.py
hjoliver: "Task instances" makes perfect sense to me, because it corresponds directly to task proxy objects in the task pool. But I think users will still wonder what in hell an instance is.

We should probably just refer to the user-facing active window (n=0) concept here.

Anyhow, let's move this conversation to Ronnie's side PR.

wxtim:
Functional question: If I do `cylc remove workflow//farfuture/task` I get

```
2024-11-14T08:59:17Z WARNING - Task(s) not removable: 221/foo
```

In the scheduler log, which is great. Any possibility of a warning from command validation? Or is this a sufficiently silly foot-shooting that we're not worrying?

MetRonnie:
> If I do `cylc remove workflow//farfuture/task`... Any possibility of a warning from command validation? 

Not really, the removal processes needs to run to tell if a task is removable or not

> Prerequisites satisfied by set seem to be unset by remove:
> ```bash
> cylc set "${CYLC_WORKFLOW_ID}//1/foo"
> sleep 3
> cylc remove "${CYLC_WORKFLOW_ID}//1/foo"
> ```

You are removing a task that you set _outputs_ on. The proposal point is talking about doing
```bash
cylc set 1/baz --pre 1/foo
cylc remove 1/foo
```

tests/functional/cylc-remove/03-flow.t
```diff
@@ -0,0 +1,43 @@
...
+
+workflow_run_ok "${TEST_NAME_BASE}-run" cylc play "${WORKFLOW_NAME}" --pause
+
+run_ok "${TEST_NAME_BASE}-remove" cylc remove "${WORKFLOW_NAME}//1/foo" --flow 1 --flow 2
+
+cylc stop "${WORKFLOW_NAME}"
+poll_workflow_stopped
+
+grep_workflow_log_ok "${TEST_NAME_BASE}-grep" "Removed task(s): 1/foo (flows=1)"
+
```
wxtim, line 42: Possible simple additional check for the basic rightness of the database?

```suggestion
cylc workflow-state "${WORKFLOW_NAME}" 2>&1 > Workflow-State
cmp_ok Workflow-State <<__HERE__
1/foo:waiting(flows=none)
__HERE__
```

tests/functional/cylc-remove/04-kill/flow.cylc
```diff
@@ -0,0 +1,30 @@
...
+[scheduling]
+    [[graph]]
+        R1 = """
+            a:started => remover
+            a:failed => u
+
+            b:submitted? => remover
+            b:submit-failed? => v
+
+            # Give u & v time to spawn if they're going to (but they shouldn't):
```
wxtim, line 16: This comment makes me surprised that remover doesn't have a sleep in it.

tests/functional/cylc-set/06-parentless/flow.cylc
```diff
@@ -19,4 +15,6 @@
             @wall_clock => x => b
         """
 [runtime]
-    [[a, b, x]]
+    [[a]]
```
wxtim, line 18: Why this refactor? Won't x run without the set?

MetRonnie: 7a9d25305027cdc6decae900d375029a5af4fb28 - De-flakify test (for MacOS CI)

tests/integration/utils/flow_tools.py
```diff
@@ -41,7 +41,7 @@
 
 def _make_src_flow(src_path, conf, filename=WorkflowFiles.FLOW_FILE):
     """Construct a workflow on the filesystem"""
-    flow_src_dir = (src_path / str(uuid1()))
+    flow_src_dir = (src_path / token_hex(4))
```
wxtim, line 44: What's the benefit of this?

tests/unit/test_workflow_db_mgr.py
```diff
@@ -0,0 +1,80 @@
...
+        db_mgr.process_queued_ops()
+
+        removed_fnums = db_mgr.remove_task_from_flows('1', 'a', flow_nums)
+        assert removed_fnums == expected_removed
+
+        db_mgr.process_queued_ops()
+        remaining_fnums: Set[str] = {
+            fnums_str
+            for fnums_str, *_ in dao.connect().execute(
+                'SELECT flow_nums FROM task_states'
```
wxtim, line 75: Workflow_db_mgr does

```python

        for table in (
            self.TABLE_TASK_STATES,
            self.TABLE_TASK_OUTPUTS,
        ):
```

Worth explicitly checking both tables?

tests/unit/test_task_proxy.py
```diff
@@ -101,3 +103,27 @@ def test_status_match(status_str: Optional[str], expected: bool):
...
+    assert TaskProxy.match_flows(mock_itask, flow_nums) == expected
+
+
+def test_match_flows_copy():
+    """Test that this method does not return the same reference as
+    itask.flow_nums."""
+    mock_itask = Mock(flow_nums={1, 2})
+    result = TaskProxy.match_flows(mock_itask, set())
+    assert result == mock_itask.flow_nums
+    assert result is not mock_itask.flow_nums
```
wxtim, line 129:
```suggestion
    assert TaskProxy.match_flows(mock_itask, set()) != mock_itask.flow_nums
```

tests/unit/test_task_proxy.py
```diff
@@ -101,3 +103,27 @@ def test_status_match(status_str: Optional[str], expected: bool):
...
+def test_match_flows(
+    itask_flow_nums: FlowNums, flow_nums: FlowNums, expected: FlowNums
+):
+    mock_itask = Mock(flow_nums=itask_flow_nums)
+    assert TaskProxy.match_flows(mock_itask, flow_nums) == expected
+
+
+def test_match_flows_copy():
+    """Test that this method does not return the same reference as
+    itask.flow_nums."""
```
wxtim, line 125: Worth stating why it's particularly important that it doesn't return the same ref?

MetRonnie: Added to https://github.com/MetRonnie/cylc-flow/pull/34

tests/integration/test_remove.py
```diff
@@ -0,0 +1,481 @@
...
+)
+from cylc.flow.cycling.integer import IntegerPoint
+from cylc.flow.flow_mgr import FLOW_ALL
+from cylc.flow.scheduler import Scheduler
+from cylc.flow.task_outputs import TASK_OUTPUT_SUCCEEDED
+from cylc.flow.task_proxy import TaskProxy
+from cylc.flow.task_state import TASK_STATUS_FAILED
+
+
+def get_pool_tasks(schd: Scheduler) -> Set[str]:
```
wxtim, line 40: IMO this would be a universal good as a fixture in conftest...

tests/unit/test_workflow_db_mgr.py
MetRonnie: Added to https://github.com/MetRonnie/cylc-flow/pull/34

tests/functional/cylc-remove/03-flow.t
MetRonnie: The task never runs in this test. I had a go at adding it to an integration test, but it turned out to be a whole thing because you can't call the `workflow_state` xtrigger from within an async test... So I'll put that in a separate follow-up

tests/integration/test_remove.py
wxtim: Fair enough. Not too bothered./

tests/functional/cylc-remove/03-flow.t
MetRonnie: Ah right, I forgot `workflow-state` could list all tasks in a workflow

oliver-sanders:
I've got one example which I don't think is quite working right:

```
[scheduling]
    cycling mode = integer
    initial cycle point = 1
    [[graph]]
        P1 = a[-P1] => a

[runtime]
    [[a]]
        script = sleep 1
```

```
cylc vip --pause
cylc set //1/a
cylc remove //1/a
```

The downstream task `2/a` was spawned by the manual setting of `1/a:succeeded`. So removing `1/a` should not have caused `2/a` to be removed as the "succeeded" output was not completed naturally (proposal point 3).

oliver-sanders:
I've discovered a small inconsistency in behaviour arising from the proposal. I don't think we should change anything to address this (in general, but at least not in this PR), but raising here for completeness.eness.

We are counteracting the effects of SoD pre-spawning by removing partially satisfied tasks, however, we are not counteracting the effects of SoD flow merging by un-merging flows.

#### Pre Spawning

(using the above example):

```
cylc play --pause
cylc hold //2/a
cylc play  #  let 1/a complete
```

We now have:

```console
$ sqlite3 'SELECT * FROM task_states;'
a|1|[1]|2024-11-20T13:03:53Z|2024-11-20T13:05:45Z|1|succeeded|0|0
a|2|[1]|2024-11-20T13:05:36Z|2024-11-20T13:05:45Z|0|waiting|0|0
```

Now, remove `1/a`:

```
cylc remove //1/a
```

And the workflow stops (flow 1 has been terminated):

```console
$ sqlite3 'SELECT * FROM task_states;'
a|1|[]|2024-11-20T13:03:53Z|2024-11-20T13:05:45Z|1|succeeded|0|0
a|2|[]|2024-11-20T13:05:36Z|2024-11-20T13:05:45Z|0|waiting|0|0
```

:+1:

#### Merging

(using the above example):

```
cylc vip --pause
cylc hold //2/a
cylc play  #  let 1/a complete
```

We now have:

```console
$ sqlite3 'SELECT * FROM task_states;'
a|1|[1]|2024-11-20T13:03:53Z|2024-11-20T13:05:45Z|1|succeeded|0|0
a|2|[1]|2024-11-20T13:05:36Z|2024-11-20T13:05:45Z|0|waiting|0|0
```

Now:

```
cylc trigger 1/a --flow=2
```

The second run of `1/a` caused flow merging at `2/a`, so we now have:

```console
$ sqlite3 'SELECT * FROM task_states;'
a|1|[1]|2024-11-20T13:17:41Z|2024-11-20T13:18:01Z|1|succeeded|0|0
a|2|[1]|2024-11-20T13:17:55Z|2024-11-20T13:17:55Z|0|waiting|0|0
a|1|[2]|2024-11-20T13:18:10Z|2024-11-20T13:18:16Z|2|succeeded|0|0
a|2|[1, 2]|2024-11-20T13:18:16Z|2024-11-20T13:18:16Z|0|waiting|0|0
```

Now, lets remove `1/a`:

```
cylc remove //1/a
```

We end up with:

```console
$ sqlite3 'SELECT * FROM task_states;'
a|1|[]|2024-11-20T13:19:12Z|2024-11-20T13:19:45Z|1|succeeded|0|0
a|2|[1]|2024-11-20T13:19:36Z|2024-11-20T13:19:36Z|0|waiting|0|0
a|1|[2]|2024-11-20T13:19:46Z|2024-11-20T13:19:52Z|2|succeeded|0|0
a|2|[1, 2]|2024-11-20T13:19:52Z|2024-11-20T13:19:52Z|0|waiting|0|0
```

Unlike in the similar pre-spawning example above, the flow 1 continues through the downstream task `2/a` (i.e. it is not terminated).

This is because the flow is merged into the pre-existing task (flow-merging rather than task-spawning).he remove proposal did not consider the effects of downstream flow merging (so not relevant to this PR in any case).

As the effects of flow merging may spread to other tasks (e.g. it could even cause the pre-spawning of parentless tasks) it probably doesn't make sense to un-merge flows (Cylc graph evolution works forwards, not backwards), so I think this is just a quirk of concurrent flow merging. No biggie, if the user wants to terminate the flow, they can `cylc stop --flow=1` or manually remove the flow from affected tasks.

hjoliver:
(Almost done testing here, looking good)

MetRonnie:
> The downstream task `2/a` was spawned by the manual setting of `1/a:succeeded`. So removing `1/a` should not have caused `2/a` to be removed as the "succeeded" output was not completed naturally (proposal point 3).

I thought we discussed this and concluded that it was only prerequisites set by `cylc set downstream_task --prerequisite` that `cylc remove upstream_task` would not undo?

hjoliver:
> > The downstream task `2/a` was spawned by the manual setting of `1/a:succeeded`. So removing `1/a` should not have caused `2/a` to be removed as the "succeeded" output was not completed naturally (proposal point 3).
> 
> I thought we discussed this and concluded that it was only prerequisites set by `cylc set downstream_task --prerequisite` that `cylc remove upstream_task` would not undo?

I read proposal point 3 as saying `2/a` *should* be removed in this case, and it does get removed. And that makes sense to me: `2/a` was spawned solely due to the forced action on `1/a`, so if we erase that upstream action `2/a` should be un-spawned.

hjoliver:
@wxtim  - I'll the conflict resolution to you. Looks like we have used `S014` for two different rules.

oliver-sanders:
> I thought we discussed this and concluded that it was only prerequisites set by cylc set downstream_task --prerequisite that cylc remove upstream_task would not undo?

Aah yes, of course.

hjoliver:
If downstream tasks get removed, I think that should be logged at INFO level rather than DEBUG. It's potentially quite important information.

oliver-sanders:
> (Does that mean you now agree with the behavior on this branch?)

Ish, there are two schools of thought, but I'm happy to go ahead with this.

> If downstream tasks get removed, I think that should be logged at INFO level rather than DEBUG. It's potentially quite important information.

Note: We are only talking about SoD pre-spawning here (as we only remove children with partially satisfied prereqs).

SoD pre-spawning is really an implementation detail, so we could argue that these tasks should not be logged (I completely understand why Ronnie has logged this at DEBUG level). However, on the other hand, the n-window is now part of the model, it defines what the GUI sees and what commands users can issue against the task, so it is pertinent information.

However, given that we log tasks that are pre-spawned by SoD at the INFO level, I agree that we should log their removal at the INFO level too.

hjoliver:
> Note: We are only talking about SoD pre-spawning here (as we only remove children with partially satisfied prereqs).
> 
> SoD pre-spawning is really an implementation detail, so we could argue that these tasks should not be logged (I completely understand why Ronnie has logged this at DEBUG level). However, on the other hand, the n-window is now part of the model, it defines what the GUI sees and what commands users can issue against the task, so it is pertinent information.
> 
> Given that we log tasks that are pre-spawned by SoD at the INFO level, I agree that we should log their removal at the INFO level too

Yes, my thinking exackery. 

I sort of agree that pre-spawning is really an implementation detail, but pragmatically we put partially satisfied tasks into n=0 for good reasons - they, at least some of the time, indicate a problem that should stall the workflow. Given that they are visible in n=0, their removal should be logged pretty visibly.

hjoliver (approved):
I'm approving this sucker, conditional on the small logging change. Nice work 🎖️

MetRonnie:
> If downstream tasks get removed, I think that should be logged at INFO level rather than DEBUG. It's potentially quite important information.

> SoD pre-spawning is really an implementation detail, so we could argue that these tasks should not be logged (I completely understand why Ronnie has logged this at DEBUG level). However, on the other hand...

N.B. The logging at DEBUG level is not something I have touched; this is the default for `TaskPool.remove(itask)`, unless the task is preparing, submitted or running in which case it becomes WARNING.

But I'm happy to change to make it INFO for the case in question

cylc/flow/scheduler.py
```diff
@@ -1050,6 +1049,133 @@ def kill_tasks(
...
+                itask.removed = True
+            itask.flow_nums.difference_update(fnums_to_remove)
+
+        matched_task_ids = {
+            *removed.keys(),
+            *(quick_relative_id(cycle, task) for task, cycle in inactive),
+        }
+
+        for id_ in matched_task_ids:
+            point_str, name = id_.split('/', 1)
```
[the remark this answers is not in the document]
MetRonnie: This was my intention to begin with, however it ended up being cleaner IMO to use the ID as the ID is what is being used for the majority of things

cylc/flow/scheduler.py
```diff
@@ -1050,6 +1049,133 @@ def kill_tasks(
...
+            tdef = self.config.taskdefs[name]
+            # Go through downstream tasks to see if any need to stand down
+            # as a result of this task being removed:
+            for child in set(itertools.chain.from_iterable(
+                generate_graph_children(tdef, get_point(point_str)).values()
+            )):
```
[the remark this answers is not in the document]
MetRonnie: Could do?
```suggestion
            itask_ = self.pool._get_task_by_id(id_)
            tdef = itask_.tdef if itask_ else self.config.taskdefs[name]
            # Go through downstream tasks to see if any need to stand down
            # as a result of this task being removed:
            for child in set(itertools.chain.from_iterable(
                generate_graph_children(tdef, get_point(point_str)).values()
            )):
```

cylc/flow/scheduler.py
```diff
@@ -1050,6 +1049,133 @@ def kill_tasks(
 
         return len(unkillable)
 
+    def remove_tasks(
+        self, items: Iterable[str], flow_nums: Optional['FlowNums'] = None
+    ) -> None:
+        """Remove tasks from the task pool (by command).
```
wxtim, line 1055:
```suggestion
        """Remove tasks from the task pool (when requested to do so by command).
```

cylc/flow/scheduler.py
```diff
@@ -1050,6 +1049,133 @@ def kill_tasks(
...
+
+        Args:
+            items: Relative IDs or globs.
+            flow_nums: Flows to remove the tasks from. If empty or None, it
+                means 'all'.
+        """
+        active, inactive, _unmatched = self.pool.filter_task_proxies(
+            items, warn_no_active=False, inactive=True
+        )
+        if not (active or inactive):
```
wxtim, line 1065: Worth testing?

cylc/flow/flow_mgr.py
```diff
@@ -55,36 +62,62 @@ def add_flow_opts(parser):
     )
 
 
-def stringify_flow_nums(flow_nums: Set[int], full: bool = False) -> str:
-    """Return a string representation of a set of flow numbers
+def get_flow_nums_set(flow: List[str]) -> FlowNums:
```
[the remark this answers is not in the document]
MetRonnie: If there is a need to broaden it, it could be done. But there isn't a need at the moment so I'll leave it.

cylc/flow/scheduler.py
MetRonnie: See https://github.com/MetRonnie/cylc-flow/pull/34/commits/052af843b15b78ced237e1997c2c3ce8d7cd44e5

cylc/flow/scheduler.py
MetRonnie: https://github.com/MetRonnie/cylc-flow/pull/34/commits/3658a86081dfe82f208f3f31f5057d6f4993469c

cylc/flow/scheduler.py
```diff
@@ -1050,6 +1049,133 @@ def kill_tasks(
...
+        self, items: Iterable[str], flow_nums: Optional['FlowNums'] = None
+    ) -> None:
+        """Remove tasks from the task pool (by command).
+
+        Args:
+            items: Relative IDs or globs.
+            flow_nums: Flows to remove the tasks from. If empty or None, it
+                means 'all'.
+        """
+        active, inactive, _unmatched = self.pool.filter_task_proxies(
```
wxtim, line 1062:
```suggestion
        active, inactive, _= self.pool.filter_task_proxies(
```

cylc/flow/scheduler.py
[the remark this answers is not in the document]
oliver-sanders: The code I'm highlighting here converts from components to IDs back to components, i.e. the IDs are an unnecessary intermediary.

cylc/flow/scheduler.py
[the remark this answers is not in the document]
oliver-sanders: Yep, alternatively could capture the `itask.tdef` in the block above to avoid the task pool lookup.

cylc/flow/scheduler.py
MetRonnie: > Or is unmatched meant to be descriptive?

Yes

oliver-sanders: This is fine. It is a Python convention to prefix unused variables with an underscore, or just leave an underscore.

This convention is inherited from functional languages where underscore variables may have a functional meaning (i.e. the variable is truly *ignored*). Python isn't a pure functional language like these, so the underscore is just a convention, it doesn't do anything and the unwanted variable is still assigned.

wxtim (approved):
I've had a decent look at the code, and made a real effort to break it in practice, without success. 🎉

oliver-sanders:
Just had a shock reading back through the proposal:

> At present (8.1.4) the Cylc 8 intervention is:

8.1.4!

cylc/flow/scheduler.py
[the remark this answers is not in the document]
oliver-sanders: Bumped to https://github.com/cylc/cylc-flow/issues/6502

oliver-sanders (approved):
Nice :tada:!

This is a biggie, well done @MetRonnie!

Issue #5802, opened 2023-11-01
cat-log: list job.out and job.err when jobs are running that require a tailer to access


oliver-sanders:
Ensure that the `job.out` and `job.err` files are listed for jobs running on platforms with a configured `tailer` command even if the file is not yet present in the job log directory.

Explanation:

* The `cylc cat-log` command can be used to retrieve job logs.
* The `-m l` option tells `cat-log` to list the available job log files.
* The Cylc UI Server uses the `-m l` option to populate the dropdown of available log files.
* Some platforms have a funny setup where the `job.out` and `job.err` files are moved into the job log directory *after* the job has completed. Until this point they live on a compute node somewhere so do not appear in the `-m l` listing.
* To allow users to access these files whilst the job is still running the batch system provides a command.
* Cylc provides support for these setups using the "tailer" options, these allow system administrators to configure the batch system command to view the files.
* So in these situations, the `job.out` and `job.err` files are viewable via the configured "tailer", however, they don't appear in the file system listings (because they haven't been moved into the job log directory yet).
* To get around this we should add some logic to `cylc cat-log` which:
  * Adds `job.out` and `job.err` to the list of files returned by `cylc cat-log -m l`,
  * *if* the job is running (as determined from the database),
  * *and* there is a "tailer" configured for the platform (as determined from the platform configuration).
* We are (apparently) already performing database access somewhere in `cylc cat-log` so this should be a matter of hooking this extra logic into that check.

PR #6480, opened 2024-11-15
cat-log: list out/err files when available via tailer

oliver-sanders:
Closes https://github.com/cylc/cylc-flow/issues/5802

List log files that are not present on disk but which are available via a tailer.

**Check List**

### DIFF
--- a/cylc/flow/scripts/cat_log.py
+++ b/cylc/flow/scripts/cat_log.py
@@ -596,28 +596,51 @@ def _main(
                 cmd.append('--prepend-path')
             cmd.append(workflow_id)
             # TODO: Add Intelligent Host selection to this
+            proc = None
             with suppress(KeyboardInterrupt):
                 # (Ctrl-C while tailing)
                 # NOTE: This will raise NoHostsError if the platform is not
                 # contactable
-                remote_cylc_cmd(
+                proc = remote_cylc_cmd(
                     cmd,
                     platform,
-                    capture_process=False,
+                    capture_process=(mode == 'list-dir'),
                     manage=(mode == 'tail'),
-                    text=False
+                    text=(mode == 'list-dir'),
                 )
-            if (
-                mode == 'list-dir'
-                and os.path.exists(
-                    os.path.join(
-                        local_log_dir,
-                        'job-activity.log'
-                    )
-                )
-            ):
-                # add the local-only job-activity.log file to the remote-list
-                print('job-activity.log')
+
+            # add and missing items to file listing results
+            if isinstance(proc, Popen):
+                # i.e: if mode=='list-dir' and ctrl+c not pressed
+                out, err = proc.communicate()
+                files = out.splitlines()
+
+                # add files which can be accessed via a tailer
+                if live_job_id is not None:
+                    if (
+                        # NOTE: only list the file if it can be viewed in
+                        # both modes
+                        (platform['out tailer'] and platform['out viewer'])
+                        and 'job.out' not in files
+                    ):
+                        files.append('job.out')
+                    if (
+                        (platform['err tailer'] and platform['err viewer'])
+                        and 'job.err' not in files
+                    ):
+                        files.append('job.err')
+
+                # add the job-activity.log file which is always local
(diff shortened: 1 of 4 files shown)


oliver-sanders:
Note: This approach works just fine, however, it has a couple of small caveats:
* It isn't checking whether the out/err files have appeared in the output already reported - i.e. we could get duplicate `job.out` / `job.err` files in the output.
* The `job.out` and `job.err` files will always appear at the end, not in alphabetical order.

Ideally, we would avoid these caveats. I did start looking at a solution where I added the `job.out` and `job.err` files onto the end of the `mode` variable, e.g. `mode = list-dirs:job.out:job.err`, that way they can be output along with the regular directory listing in a much more elegant way.

But I ran out of time. Writing this up in case anyone else needs to take this one over.

tests/functional/cylc-cat-log/12-remote-out-err-tailer.t
wxtim, line 1: `12-delete-kill.t` already exists - can we make this number 13?

wxtim (approved):
- [x] Read code
- [x] Ran new test
- [x] Tried manually testing, including looking at available log files in TUI before and after.
- [x] Ran new test with coverage. Lines indicated in the diff were covered.

cylc/flow/scripts/cat_log.py
```diff
@@ -596,28 +596,49 @@ def _main(
...
-                    )
-                )
-            ):
-                # add the local-only job-activity.log file to the remote-list
-                print('job-activity.log')
+
+            # add and missing items to file listing results
+            if isinstance(proc, Popen):
+                # i.e: if mode=='list-dir' and ctrl+c not pressed
+                out, err = proc.communicate()
```
MetRonnie, line 615: This is now swallowing stderr and non-zero exit codes (discussed in-person)

MetRonnie (changes requested):

cylc/flow/scripts/cat_log.py
```diff
@@ -596,28 +596,50 @@ def _main(
...
+
+                # add the job-activity.log file which is always local
+                if os.path.exists(
+                    os.path.join(local_log_dir, 'job-activity.log')
+                ):
+                    files.append('job-activity.log')
+
+                files.sort()
+                print('\n'.join(files))
+                print(err, file=sys.stderr)
```
MetRonnie, line 641: Currently on 8.4.0 the `cat-log -m l` command exits non-zero if this subprocess exits non-zero. But it no longer does on this branch

oliver-sanders: Hopefully fixed.

cylc/flow/scripts/cat_log.py
```diff
@@ -596,28 +596,51 @@ def _main(
...
-                    )
-                )
-            ):
-                # add the local-only job-activity.log file to the remote-list
-                print('job-activity.log')
+
+            # add and missing items to file listing results
+            if isinstance(proc, Popen):
+                # i.e: if mode=='list-dir' and ctrl+c not pressed
+                out, err = proc.communicate()
```
MetRonnie, line 615: I think we need to include the `proc.communicate()` in the suppressing of the `KeyboardInterrupt`

MetRonnie: If the whole `if isinstance` block is included in the `suppress` block, it avoids traceback on ctrl+c. But not that important I guess

MetRonnie (approved):
Take or leave my KeyboardInterrupt suggestions

oliver-sanders:
> Take or leave my KeyboardInterrupt suggestions

I think we need to do more than just swallow the traceback. Have opened a low-priority issue: https://github.com/cylc/cylc-flow/issues/6609

cylc/flow/scripts/cat_log.py
oliver-sanders: The traceback is a cosmetic problem, but I think there are pre-existing functional issues here, https://github.com/cylc/cylc-flow/issues/6609

PR #6459, opened 2024-11-05
Lint: Tell users that backslashes are no longer necessary after => & |

wxtim:
Closes #6456 

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [x] Changelog entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] This is a feature, but a lint feature, so suitable to be added to a maintainance release.

PR #6493, opened 2024-11-25
`cylc lint`: S015/U017 line continuation chars

wxtim:
Replaces https://github.com/cylc/cylc-flow/pull/6459 which blocks sync PRs

Otherwise the same diff.

Seealso @MetRonnie's Comment on https://github.com/cylc/cylc-flow/pull/6459#issuecomment-2497686715

Probably OK to be a single review since it's a copy of an already accepted PR.

### DIFF
--- a/changes.d/6459.feat.md
+++ b/changes.d/6459.feat.md
@@ -1,1 +1,1 @@
-Cylc lint now checks for unecessary continuation characters in graph section.
+`cylc lint` now checks for unecessary continuation characters in graph section.
--- a/changes.d/6459.feat.md
+++ b/changes.d/6459.feat.md
@@ -1,1 +1,1 @@
-`cylc lint` now checks for unecessary continuation characters in graph section.
+`cylc lint` now checks for unnecessary continuation characters in the graph section.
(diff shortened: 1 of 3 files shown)


changes.d/6459.feat.md
MetRonnie, line 1: 6493

oliver-sanders: It's ok to reference the issue rather than the PR right?

MetRonnie: Might be better in this case

cylc/flow/scripts/lint.py
```diff
@@ -617,7 +617,13 @@ def list_wrapper(line: str, check: Callable) -> Optional[Dict[str, str]]:
             for job_runner, directive in WALLCLOCK_DIRECTIVES.items()
         )),
         FUNCTION: check_wallclock_directives,
-    }
+    },
+    'S015': {
+        'short': (
+            '`=>` is a line continuation without `\\`.'
```
hjoliver, line 623:
```suggestion
            '`=>` implies line continuation without `\\`.'
```
(Pretty sure you accepted this change on the original!)

hjoliver (changes requested):

hjoliver (approved):
All good 🎉

changes.d/6459.feat.md
MetRonnie: #6497

PR #6503, opened 2024-11-28
convert from ids to tokens

oliver-sanders:
Avoid using `'/'.split(id_)`.

**Check List**

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -75,7 +75,7 @@ from cylc.flow.hostuserutil import (
     get_user,
     is_remote_platform,
 )
-from cylc.flow.id import Tokens, quick_relative_id
+from cylc.flow.id import Tokens
 from cylc.flow.log_level import verbosity_to_env, verbosity_to_opts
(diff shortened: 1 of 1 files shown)


MetRonnie (commented):
Adds slight complexity that there is a risk of dealing with both full and relative tokens here. I have carefully checked and pushed a commit that changes hopefully all of these to relative tokens

oliver-sanders:
Wasn't actually an issue, but ok

cylc/flow/scheduler.py
```diff
@@ -1138,7 +1138,7 @@ def remove_tasks(
                 # Check if downstream task should remain spawned:
                 if (
                     # Ignoring tasks we are already dealing with:
-                    child_itask.identity in matched_task_ids
+                    child_itask.tokens.task in matched_tasks
```
MetRonnie, line 1141: Not sure why Mypy didn't catch this non-overlapping comparison (`if str in set[Tokens]`) 🤔

MetRonnie (approved):

PR #6505, opened 2024-11-29
`cylc set`: log warning when no outputs specified and task has no required outputs

MetRonnie:
```
R1 = a? => b
```
```
$ cylc set workflow//1/a
```
does nothing and currently produces no warning.

This change makes it log a warning.

https://cylc.discourse.group/t/cylc-set-for-a-task-with-multiple-outputs-doesnt-seem-to-do-anything/1074

**Check List**

- [x] No dependency changes 
- [x] Tests are included 
- [x] No changelog entry needed as minor 
- [x] No docs needed

### DIFF
--- a/cylc/flow/id_match.py
+++ b/cylc/flow/id_match.py
@@ -98,7 +98,8 @@ def filter_ids(
             * If IDTokens.Cycle all CyclePoints with any matching tasks will
               be returned.
         warn:
-            Whether to log a warning if no matching tasks are found.
+            Whether to log a warning if no matching tasks are found in the
+            pool.
 
     TODO:
         Consider using wcmatch which would add support for
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1943,7 +1943,7 @@ class TaskPool:
 
         """
         # Get matching pool tasks and inactive task definitions.
-        itasks, inactive_tasks, unmatched = self.filter_task_proxies(
+        itasks, inactive_tasks, _unmatched = self.filter_task_proxies(
             items,
             inactive=True,
             warn_no_active=False,
@@ -1951,6 +1951,7 @@ class TaskPool:
 
         flow_nums = self._get_flow_nums(flow, flow_descr)
 
+        nothing_set: Set[str] = set()
         # Set existing task proxies.
         for itask in itasks:
             if flow == ['none'] and itask.flow_nums != set():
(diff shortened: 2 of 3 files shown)


MetRonnie:
I realised there was a slight mistake and the test was not quite what I wanted.

@oliver-sanders Why, for task `a` in
```
R1 = """
    a? => b
    a:x
"""
```
is the auto-generated completion statement `(x and succeeded) or failed`?

Shouldn't it be `x and (succeeded or failed)`?

hjoliver:
Hmm, that does seem wrong. 

Logically, it should be `(x and succeeded) or (x and failed)` i.e. `x and (succeeded or failed)`.

oliver-sanders:
> is the auto-generated completion statement (x and succeeded) or failed? This is making x optional even though it is not optional in the graph.
>
> Shouldn't it be x and (succeeded or failed)?

No, this is as proposed.

Outputs are associated with the success pathway. The source of truth is in the completion expression.

MetRonnie:
> Doing it as suggested above would prevent clean graph branching (e.g, in your example, you would need a suicide trigger to remove the task if it failed because it didn't produce the output "x").

Can't say I understand this. In my example `x` should be a required output. If it didn't produce the output "x" then that should need an intervention.

hjoliver:
OK I can see that proposal point 3 does indeed lead to `(x and succeeded) or failed`:

> the expression is automatically generated according to the rules:
> 
> i.   All required outputs (referenced in the graph) must be satisfied.
> ii.    Or, if success is optional, then the outputs are complete if it fails.

but I'm no longer entirely convinced that that's correct.

> By default, outputs are associated with the success pathway. The task cannot be expected to complete these outputs in other scenarios

You're saying failure is optional, and failure *could* occur early in execution, before `a:x` can be generated. 

That's true, but I think it's still legit to have a graph that requires the `a:x` path regardless of `a`'s success or failure - and hence manual intervention if `a:x` is not generated.  
```
a? => a-success-path
a:fail? => a-failure-path
a:x => a-x-path  # required either way
```
 
It's conceivable that `a` has a known and handle-able failure mode late in execution, and we need `a:x` to be generated even if it fails.

(Is that what you're saying @MetRonnie ?)

MetRonnie:
Sorry to stoke an argument 🤭 I suggest we re-visit this after 8.4.0 is released. I will probably just stick `completion = x` on the task to make the test work

hjoliver:
It comes down to what should be the default completion condition vs what should require a user-defined one. 

Looking at succeed/fail alone, to my mind we've got it the wrong way around because the current default is not the most literal interpretation of the graph.

Point taken on consistency with submit-failed and expired, but I'm not sure that matters.  Succeed and fail are fundamentally different from those because they involve actually executing the job, so it's seems fine to me to treat them slightly differently on that basis.

But OK, on the back-burner for post-8.4.0 revisit!

MetRonnie:
Btw it seems like instead of logging a warning, we might as well do what skip mode does and just emit the "success pathway" outputs if a user does `cylc set task` without the `--out` option and the task has no required outputs.

Issue #6502, opened 2024-11-28
remove: handle reload


oliver-sanders:
> [!IMPORTANT]
> Don't bump this, it could break!

Follow up to the `cylc remove` implementation (https://github.com/cylc/cylc-flow/pull/6472).

Ensure that it is robust to workflow reloads e.g. adding / removing tasks and dependencies.

For more info see: https://github.com/cylc/cylc-flow/pull/6472#discussion_r1856964148

PR #6516, opened 2024-12-05
Fix KeyError when using `cylc remove` after a reload changed the graph

MetRonnie:
Closes https://github.com/cylc/cylc-flow/issues/6502

Unreleased bug. To reproduce:
1. Start a workflow `a & a:started => b` where `a` is either a long running task or is set to fail
2. Change the graph to just `b` and reload the workflow
3. Do `cylc remove a`
4. This results in a `KeyError` which crashes the scheduler

**Check List**

- [x] No dependency changes 
- [x] Tests are included 
- [x] No changelog entry needed as unreleased bug
- [x] No docs needed
- [x] Master branch is correct

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -70,7 +70,6 @@ from cylc.flow import (
 from cylc.flow.broadcast_mgr import BroadcastMgr
 from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
 from cylc.flow.config import WorkflowConfig
-from cylc.flow.cycling.loader import get_point
 from cylc.flow.data_store_mgr import DataStoreMgr
 from cylc.flow.exceptions import (
     CommandFailedError,
@@ -93,9 +92,7 @@ from cylc.flow.hostuserutil import (
     get_user,
     is_remote_platform,
 )
-from cylc.flow.id import (
-    Tokens,
-)
+from cylc.flow.id import Tokens, quick_relative_id
 from cylc.flow.log_level import (
     verbosity_to_env,
     verbosity_to_opts,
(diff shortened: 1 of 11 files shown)


cylc/flow/scheduler.py
```diff
@@ -1102,17 +1099,20 @@ def remove_tasks(
...
         if flow_nums is None:
             flow_nums = set()
-        # Mapping of *relative* task IDs to removed flow numbers:
-        removed: Dict[Tokens, FlowNums] = {}
-        not_removed: Set[Tokens] = set()
+        # Mapping of relative task IDs to removed flow numbers:
+        removed: Dict[str, FlowNums] = {}
+        not_removed: Set[str] = set()
+        # All the matched tasks (will add applicable active tasks below):
+        matched_tasks = inactive.copy()
```
MetRonnie, line 1106: Ok so I went back to using string IDs... But don't need the yucky `.split('/')` anymore

cylc/flow/task_pool.py
```diff
@@ -2327,7 +2327,7 @@ def filter_task_proxies(
         ids: Iterable[str],
         warn_no_active: bool = True,
         inactive: bool = False,
-    ) -> 'Tuple[List[TaskProxy], Set[Tuple[str, PointBase]], List[str]]':
+    ) -> 'Tuple[List[TaskProxy], Set[Tuple[TaskDef, PointBase]], List[str]]':
```
MetRonnie, line 2330: One thing to note with this change is that the hashability/equality of `TaskDef`s is purely reference based (i.e. `__hash__()` and `__eq__()` fall back to `object` default). This should be fine as the `TaskDef`s are always coming from the `scheduler.config.taskdefs` dictionary, I believe

wxtim (approved):
:)

- [x] Reproduced bug
- [x] Checked that it fixed bug
- [x] Read code
- [x] Read tests
- [x] Got irked that "read" is pronounced differently dependent on tense without changing spelling

hjoliver:
>  Got irked that "read" is pronounced differently dependent on tense without changing spelling

🤣  fair play


@dwsutherland could this be the reload bug your reported in our meeting the other day ?

hjoliver (approved):

PR #6518, opened 2024-12-06
allow workflow config to unset global events config

TomekTrzeciak:
This addresses a bug whereby it is not possible to reset list-type values like `mail events` from the global config to empty list by overriding them in the workflow config.

**Check List**

### DIFF
--- a/changes.d/6518.fix.md
+++ b/changes.d/6518.fix.md
@@ -1,0 +1,1 @@
+Allow setting empty values in `flow.cylc[scheduler][events]` to override the global configuration.
--- a/cylc/flow/parsec/config.py
+++ b/cylc/flow/parsec/config.py
@@ -35,6 +35,10 @@ if TYPE_CHECKING:
     from optparse import Values
 
 
+class DefaultList(list):
+    """List subclass to indicate unassigned list values in expanded config."""
+
+
 class ParsecConfig:
     """Object wrapper for parsec functions."""
 
@@ -112,7 +116,7 @@ class ParsecConfig:
                     else:
                         if node.default == ConfigNode.UNSET:
                             if node.vdr and node.vdr.endswith('_LIST'):
-                                defs[node.name] = []
+                                defs[node.name] = DefaultList()
                             else:
                                 defs[node.name] = None
                         else:
--- a/cylc/flow/parsec/util.py
+++ b/cylc/flow/parsec/util.py
@@ -212,7 +212,7 @@ def replicate(target, source):
                 target[key].defaults_ = pdeepcopy(val.defaults_)
             replicate(target[key], val)
         elif isinstance(val, list):
-            target[key] = type(val)(val)
+            target[key] = val[:]
         else:
             target[key] = val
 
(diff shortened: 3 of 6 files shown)


MetRonnie:
Thanks @TomekTrzeciak :+1: it might be January before we have a chance to look at this however

oliver-sanders:
This would change the meaning of `mail events =` from "reset to global default" to "reset to empty list".

The "reset to empty lisy" behaviour makes sense, but unfortunately, I think there may be valid use cases for both behaviours (especially when you take inheritance into account). I had a skim through workflows at our site to see what list items are being "unset" in this way and found the following:

* `mail events` (a lot of instances)
* `(submission|execution) retry events` (a lot of instances)
* `handler events` (a few instances)
* `collapsed families` (a couple instances)

MetRonnie:
I think this "reset to global default" behaviour only happens for events settings and no other settings?

TomekTrzeciak:
> I think this "reset to global default" behaviour only happens for events settings and no other settings?

My reading is that only `events` and `mail` settings under `scheduler` section would be affected by this change.

> This would change the meaning of `mail events =` from "reset to global default" to "reset to empty list".
> 
> The "reset to empty lisy" behaviour makes sense, but unfortunately, I think there may be valid use cases for both behaviours (especially when you take inheritance into account). I had a skim through workflows at our site to see what list items are being "unset" in this way and found the following:
> 
> * `mail events` (a lot of instances)
> * `(submission|execution) retry events` (a lot of instances)
> * `handler events` (a few instances)
> * `collapsed families` (a couple instances)

Out of this list only `mail events` and `handler events` would be affected by this change. On our site only the former (`mail events`) is set in the global config and thus would behave differently (and as expected IMHO).

hjoliver:
So this change would mean:
- no mention in workflow config: you get the global defaults
- explicit `key = # nothing`: you get nothing

Seems good to me, no blockers as far as NIWA is concerned.

@oliver-sanders - would it be sufficient for you just to ping the owners of the workflows you found and tell them about the change?

oliver-sanders:
> @oliver-sanders - would it be sufficient for you just to ping the owners of the workflows you found and tell them about the change?

Will do. I'll confirm the scope of the change and try to work out the intention of the users impacted. So long as it's just `[scheduler][events` and `[scheduler][mail]`, I'm not expecting any issues.

oliver-sanders (approved):
:+1:

I've taken a look into the change and am happy with the scope.

I had a look on site and found `mail events = ` in 513 run dirs belonging to 47 users internally. I highly doubt that this change has the potential to cause any issues, but I've put up an internal bulletin to ensure that folks are aware. I don't think we need to go any further than that.

A quick changelog would be appreciated.

MetRonnie:
Sorry, just gotten round to this. 

Apparently though, it is already possible to unset events by using a value of `None`, i.e.

```cylc
# flow.cylc
[scheduler]
    [[events]]
        mail events = None
```

oliver-sanders:
^ @TomekTrzeciak thoughts?

hjoliver:
```Python
# cfgspec/workflow.py
                if item in {'handlers', 'handler events', 'mail events'}:
                    # Note: default of None for V_STRING_LIST is used to
                    # differentiate between not set vs set to empty
                    default = None
```

I had forgotten about this too, sorry.  

However, it doesn't seem to be documented, and use of "None" (being Pythonic) is weird in a config file.
So I'd prefer this PR.
And we should document it.

MetRonnie (approved):
I've pushed a changelog entry. Got one optional suggestion, take or leave it 👍

MetRonnie:
Actually, might as well merge

PR #6524, opened 2024-12-13
Graph Parsing Incorrectly Validates if Cycle offset present

wxtim:
The following examples failed to fail validation.

* `foo[-P1]:finish => foo`
* `foo[-P1]:succeeded | foo[-P1]:failed`

Fixed by adding Cycle Offset to Regex for parsing RHS of graph pair

## Check List

### DIFF
--- a/cylc/flow/graph_parser.py
+++ b/cylc/flow/graph_parser.py
@@ -251,6 +251,7 @@ class GraphParser:
         rf"""
         (!)?           # suicide mark
         ({_RE_NODE})   # node name
+        (\[.*?\])?     # cycle offset
         ({_RE_QUAL})?  # trigger qualifier
         ({_RE_OPT})?   # optional output indicator
         """,
@@ -469,10 +470,11 @@ class GraphParser:
                 pairs.add((chain[i], chain[i + 1]))
 
         # Get a set of RH nodes which are not at the LH of another pair:
-        pairs_dict = dict(pairs)
+        pairs_dict = dict([p for p in pairs if p[0] is not None])
         terminals = set(pairs_dict.values()).difference(pairs_dict.keys())
 
-        for pair in pairs:
+        # Sort the pairs so that processing happens in a consistent order:
+        for pair in sorted(pairs, key=lambda p: str(p[0])):
             self._proc_dep_pair(pair, terminals)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


wxtim:
I've not ticked the test box yet, because I want to do some more thinking about testing the Regex.

cylc/flow/graph_parser.py
```diff
@@ -469,10 +470,11 @@ def parse_graph(self, graph_string: str) -> None:
                 pairs.add((chain[i], chain[i + 1]))
 
         # Get a set of RH nodes which are not at the LH of another pair:
-        pairs_dict = dict(pairs)
+        pairs_dict = dict([p for p in pairs if p[0] is not None])
```
wxtim, line 473: I don't think that this was actually causing problems - but I'm concerned that this is bad code because it's very possible that `[i[0] for i in pairs]` may have repeat values, and I'm not sure what the consequences of that are.

cylc/flow/graph_parser.py
```diff
@@ -469,10 +470,11 @@ def parse_graph(self, graph_string: str) -> None:
...
         terminals = set(pairs_dict.values()).difference(pairs_dict.keys())
 
-        for pair in pairs:
+        # Sort the pairs so that processing happens in a consistent order:
```
wxtim, line 476: Process dep pairs in consistent order to prevent non-deterministic behaviour.

cylc/flow/graph_parser.py
```diff
@@ -540,16 +542,12 @@ def _proc_dep_pair(
...
-                # This right hand side is at the end of a chain:
-                raise GraphParseError(
-                    'Invalid cycle point offsets only on right hand '
-                    'side of a dependency (must be on left hand side):'
-                    f' {left} => {right}')
-            else:
-                # This RHS is also a LHS in a chain:
-                return
+        if '[' in right and left and right in terminals:
+            # This right hand side is at the end of a chain:
```
wxtim, line 546:
```suggestion
```

tests/unit/test_graph_parser.py
```diff
@@ -945,11 +977,12 @@ def test_RHS_AND(graph: str, expected_triggers: Dict[str, List[str]]):
...
     (
         # Error if offset in terminal RHS:
-        param((('a', 'b[-P42M]'), {'b[-P42M]'}), 'Invalid cycle point offset'),
+        param(
+            (('a', 'b[-P42M]'), {'b[-P42M]'}), 'Invalid cycle point offset',
+            id='invalid-cycle-point-offset'
+        ),
         # No error if offset in NON-terminal RHS:
-        param((('a', 'b[-P42M]'), {}), None),
-        # Don't check the left hand side if this has a non-terminal RHS:
```
wxtim, line 951: I'm not sure why this was desirable. You will now get an error for this.

wxtim:
Replaced by https://github.com/cylc/cylc-flow/pull/6526

PR #6526, opened 2024-12-16
Fix output consistency check when there are offsets.

### LINKED ISSUE #6523: Graph: Inter-cycle dependencies not correctly checked
@dpmatthews spotted

### Description

graph

```cylc
foo[-P1]:finish => foo
```

... ought to fail validation, because finish implies `foo[-P1]:succeeded? | foo[-P1]:failed?`, so `foo` (implicit `:succeeded`) should also be optional.

I discovered the simpler case `foo[-P1]:succeeded | foo[-P1]:failed => foo` doesn't appear to work either.

### Reproducible Example

This diff will produce failing unit tests (The empty second string is to test what the error message should be, but no error is produced):

```diff
diff --git a/tests/unit/test_graph_parser.py b/tests/unit/test_graph_parser.py
index 75a0bf95a..a38373a14 100644
--- a/tests/unit/test_graph_parser.py
+++ b/tests/unit/test_graph_parser.py
@@ -139,6 +139,16 @@ def test_graph_syntax_errors_2(seq, graph, expected_err):
             'Invalid cycle point offsets only on right',
             id='no-cycle-point-RHS'
         ),
+        param(
+            "foo[-P1]:succeeded | foo[-P1]:failed => foo",
+            '',
+            id='succeed-or-failed-mustbe-optional'
+        ),
+        param(
+            "foo[-P1]:finish => foo?",
+            '',
+            id='lhs-finish-implies-success-optional'
+        ),
     ]
 )
 def test_graph_syntax_errors(graph, expected_err):
```

[diff.txt](https://github.com/user-attachments/files/18092567/diff.txt)

### Expected Behaviour

These graphs should fail validation

hjoliver:
Close #6523 

**Check List**

### DIFF
--- a/cylc/flow/graph_parser.py
+++ b/cylc/flow/graph_parser.py
@@ -251,6 +251,7 @@ class GraphParser:
         rf"""
         (!)?           # suicide mark
         ({_RE_NODE})   # node name
+        ({_RE_OFFSET})?  # cycle point offset
         ({_RE_QUAL})?  # trigger qualifier
         ({_RE_OPT})?   # optional output indicator
         """,
@@ -540,16 +541,12 @@ class GraphParser:
             raise GraphParseError(mismatch_msg.format(right))
 
         # Raise error for cycle point offsets at the end of chains
-        if '[' in right:
-            if left and (right in terminals):
-                # This right hand side is at the end of a chain:
-                raise GraphParseError(
-                    'Invalid cycle point offsets only on right hand '
-                    'side of a dependency (must be on left hand side):'
-                    f' {left} => {right}')
-            else:
-                # This RHS is also a LHS in a chain:
-                return
+        if '[' in right and left and (right in terminals):
+            # This right hand side is at the end of a chain:
+            raise GraphParseError(
+                'Invalid cycle point offsets only on right hand '
+                'side of a dependency (must be on left hand side):'
+                f' {left} => {right}')
 
         # Split right side on AND.
         rights = right.split(self.__class__.OP_AND)
@@ -887,7 +884,7 @@ class GraphParser:
                 raise ValueError(  # pragma: no cover
                     f"Unexpected graph expression: '{right}'"
                 )
-            suicide_char, name, output, opt_char = m.groups()
+            suicide_char, name, offset, output, opt_char = m.groups()
             suicide = (suicide_char == self.__class__.SUICIDE)
             optional = (opt_char == self.__class__.OPTIONAL)
             if output:
@@ -895,7 +892,7 @@ class GraphParser:
 
             if name in self.family_map:
                 fam = True
-                mems = self.family_map[name]
+                rhs_mems = self.family_map[name]
                 if not output:
                     # (Plain family name on RHS).
                     # Make implicit success explicit.
@@ -922,10 +919,13 @@ class GraphParser:
                 else:
                     # Convert to standard output names if necessary.
                     output = TaskTrigger.standardise_name(output)
-                mems = [name]
+                rhs_mems = [name]
                 outputs = [output]
 
-            for mem in mems:
-                self._set_triggers(mem, suicide, trigs, expr, orig_expr)
+            for mem in rhs_mems:
+                if not offset:
+                    # Nodes with offsets on the RHS do not define triggers.
+                    self._set_triggers(mem, suicide, trigs, expr, orig_expr)
                 for output in outputs:
+                    # But they must be consistent with output optionality.
                     self._set_output_opt(mem, output, optional, suicide, fam)
--- a/cylc/flow/graph_parser.py
+++ b/cylc/flow/graph_parser.py
@@ -891,7 +891,7 @@ class GraphParser:
 
             if name in self.family_map:
                 fam = True
-                rhs_mems = self.family_map[name]
+                rhs_members = self.family_map[name]
                 if not output:
                     # (Plain family name on RHS).
                     # Make implicit success explicit.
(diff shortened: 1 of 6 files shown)


tests/unit/test_graph_parser.py
```diff
@@ -948,8 +948,8 @@ def test_RHS_AND(graph: str, expected_triggers: Dict[str, List[str]]):
         param((('a', 'b[-P42M]'), {'b[-P42M]'}), 'Invalid cycle point offset'),
         # No error if offset in NON-terminal RHS:
         param((('a', 'b[-P42M]'), {}), None),
-        # Don't check the left hand side if this has a non-terminal RHS:
-        param((('a &', 'b[-P42M]'), {}), None),
+        # Check the left hand side if this has a non-terminal RHS:
+        param((('a &', 'b[-P42M]'), {}), 'Null task name in graph'),
```
hjoliver, line 952:
```
a & => b[-P25] => b
```
On master this validates OK and the test is confirming that? 🤯

tests/flakyfunctional/cylc-show/04-multi.t
```diff
@@ -44,7 +44,10 @@ outputs: ('⨯': not completed)
   ⨯ 2016/t1 succeeded
   ⨯ 2016/t1 failed
 output completion: incomplete
-  ⨯ ┆  succeeded
+    ┆  (
+  ✓ ┆    started
+  ⨯ ┆    and succeeded
+    ┆  )
```
hjoliver, line 50: This is now consistent with the completion condition if no offset is present.

tests/unit/test_graph_parser.py
```diff
@@ -139,6 +139,36 @@ def test_graph_syntax_errors_2(seq, graph, expected_err):
...
+        param(
+            "foo[-P1]:succeeded? | foo[-P1]:failed? => foo",
+            # Order of processing can result in different error messages
+            # (a) foo:succeeded can't be both optional and required
+            # (b) foo:succeeded and foo:failed most both be optional
+            #     if both are used.
+            # The word "optional" is common to both. Not ideal!
+            'optional',
+            id='succeed-or-failed-implies-success-optional'
+        ),
```
hjoliver, line 171: Took your test cases with attribution (via commit message) @wxtim

tests/integration/reftests/test_pre_initial.py
```diff
@@ -102,8 +102,8 @@ async def test_over_bracketed(flow, scheduler, reftest):
             'final cycle point': '2013-12-25T12:00Z',
             'graph': {
                 'T12': '''
-                    (a[-P1D]:fail | b[-P1D]:fail | c[-P1D]:fail) => d
-                    a & b & c  # Implied by implicit cycling now...
+                    (a[-P1D]:fail? | b[-P1D]:fail? | c[-P1D]:fail?) => d
```
wxtim, line 105: This is absolutely reasobable in the case of recurring cycling, but is it reasonable in general?

i.e. I think that this should fail validation

```cylc
[graph]
    R1/1 = a
    R1/2 = a[-P1]:fail
```

but I'm not sure that _this_ should, because a at -P1 is not the same task as a.

```cylc
R1 = a[-P1]:fail => a
```

I think I may be happy to gloss over this, because the latter recurrance isn't very meaningful.

cylc/flow/graph_parser.py
```diff
@@ -887,15 +884,15 @@ def _compute_triggers(
...
+            suicide_char, name, offset, output, opt_char = m.groups()
             suicide = (suicide_char == self.__class__.SUICIDE)
             optional = (opt_char == self.__class__.OPTIONAL)
             if output:
                 output = output.strip(self.__class__.QUALIFIER)
 
             if name in self.family_map:
                 fam = True
-                mems = self.family_map[name]
+                rhs_mems = self.family_map[name]
```
wxtim, line 895: `rhs_members` for easier reading please?

tests/unit/test_graph_parser.py
```diff
@@ -139,6 +139,36 @@ def test_graph_syntax_errors_2(seq, graph, expected_err):
...
+        param(
+            "foo[-P1]:succeeded | foo[-P1]:failed => bar",
+            # order of outputs varies in the error message
+            'must both be optional if both are used',
+            id='succeed-or-failed-mustbe-optional'
+        ),
+        param(
+            "foo[-P1]:succeeded? | foo[-P1]:failed? => foo",
+            # Order of processing can result in different error messages
```
wxtim, line 164: I fixed this in my version of the pr https://github.com/hjoliver/cylc-flow/pull/62

wxtim (commented):
I think we need to double check the test I've queried, but otherwise this seems to fix the bug in question.

I have a related issue where some code I did early in the year hasn't caused a bug, but I'm unclear whether it's correct. 😢

tests/unit/test_graph_parser.py
hjoliver: took it, thanks

tests/integration/reftests/test_pre_initial.py
hjoliver: So this test was passing on master only because the offsets meant we weren't checking those outputs for consistency with non-offset outputs. 

I agree that "a at -P1 is not the same task as a" but (for non-R1 cases at least) one becomes the other as the cycle point advances, so I think the output optionality does have to be consistent between the two. (If that makes sense!).

wxtim: Happy, unwilling to hold this PR up. As observed elsewhere this code is ripe for refactoring.

markgrahamdawson (approved):

PR #6530, opened 2024-12-19
Clarify trigger command help.

hjoliver:
Currently, in `cylc trigger --help`:

```
...
Attempts to trigger active tasks will be ignored.
...
  Active tasks (in the n=0 window) already belong to a flow.
  * by default, if triggered, they run in the same flow
  *...
...
```

This seems inconsistent at first glance, so I've been a bit more explicit

As far as I'm aware, the terminology we've settled on is:
- "active tasks" are literally active (submitted, running)
  - (users will definitely assume this, so let's not mess with that terminology)
- the "active window of the workflow" (n=0) contains both active and waiting tasks
  - (this is not illogical or inconsistent, but we need to be careful not to call this "the active tasks")

One review will do.

**Check List**

### DIFF
--- a/cylc/flow/scripts/trigger.py
+++ b/cylc/flow/scripts/trigger.py
@@ -27,7 +27,7 @@ Cylc queues restrict the number of jobs that can be active (submitted or
 running) at once. They release tasks to run when their active task count
 drops below the queue limit.
 
-Attempts to trigger active tasks will be ignored.
+Attempts to trigger active (submitted, running) tasks will be ignored.
 
 Examples:
   # trigger task foo in cycle 1234 in test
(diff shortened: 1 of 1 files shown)


cylc/flow/scripts/trigger.py
```diff
@@ -27,7 +27,7 @@
 running) at once. They release tasks to run when their active task count
 drops below the queue limit.
 
-Attempts to trigger active tasks will be ignored.
+Attempts to trigger active (submitted, running) tasks will be ignored.
```
oliver-sanders, line 30: I think preparing tasks will be ignored too:

```suggestion
Attempts to trigger active (preparing, submitted, running) tasks will be ignored.
```

oliver-sanders (approved):

PR #6533, opened 2024-12-20
Fix unlinking old job out/err files.

### LINKED ISSUE #6529: An unlink for JOB_LOG_ERR and JOB_LOG_OUT will always do nothing in job_runner_mgr.py
Reading some code and I saw https://github.com/cylc/cylc-flow/blob/master/cylc/flow/job_runner_mgr.py#L570

```
    def _job_submit_impl(
            self, job_file_path, job_runner_name, submit_opts):
        """Helper for self.jobs_submit() and self.job_submit()."""

        # Create NN symbolic link, if necessary
        self._create_nn(job_file_path)
        for name in JOB_LOG_ERR, JOB_LOG_OUT:
            with suppress(OSError):
                os.unlink(os.path.join(job_file_path, name))
```

I don't think that above `with` block is correct. I've not run things, but, `job_file_path = ".../task/01/job"`, so this would lead to `os.unlink(".../task/01/job/job.out")` which I would doubt exists. I think its meant to be `os.unlink(os.path.join(os.path.dirname(job_file_path), name))` like is done in the `_create_nn` function.

```
    def _create_nn(cls, job_file_path):
        """Create NN symbolic link, if necessary.

        If NN => 01, remove numbered directories with submit numbers greater
        than 01.
        Helper for "self._job_submit_impl".

        """
        job_file_dir = os.path.dirname(job_file_path)
        source = os.path.basename(job_file_dir)
        task_log_dir = os.path.dirname(job_file_dir)
        nn_path = os.path.join(task_log_dir, "NN")
```

hjoliver:
Close #6529 

```diff
diff --git a/cylc/flow/job_runner_mgr.py b/cylc/flow/job_runner_mgr.py
index b8ddaaa05..de7817170 100644
--- a/cylc/flow/job_runner_mgr.py
+++ b/cylc/flow/job_runner_mgr.py
@@ -568,8 +568,8 @@ class JobRunnerManager():
         # Create NN symbolic link, if necessary
         self._create_nn(job_file_path)
         for name in JOB_LOG_ERR, JOB_LOG_OUT:
-            with suppress(OSError):
-                os.unlink(os.path.join(job_file_path, name))
+            with suppress(FileNotFoundError):
+                os.unlink(os.path.join(os.path.dirname(job_file_path), name))

         # Start new status file
```

Suppressing OSError here hid a bug:
```
    NotADirectoryError: [Errno 20] Not a directory: '/home/oliverh/cylc-run/bug/run32/log/job/1/a/01/job/job.err'
```

This fixes the path error and just suppresses FileNotFoundError.

**Check List**

    - (there are no reports of users being affected by this)
   - (simple fix going directly for 8.4 for immediate release)

### DIFF
--- a/cylc/flow/job_runner_mgr.py
+++ b/cylc/flow/job_runner_mgr.py
@@ -373,7 +373,7 @@ class JobRunnerManager():
 
     @classmethod
     def _create_nn(cls, job_file_path):
-        """Create NN symbolic link, or remove old job logs if they exist.
+        """Create NN symbolic link if necessary, and remove any old job logs.
 
         If NN => 01, remove numbered dirs with submit numbers greater than 01.
 
--- a/cylc/flow/job_runner_mgr.py
+++ b/cylc/flow/job_runner_mgr.py
@@ -373,14 +373,15 @@ class JobRunnerManager():
 
     @classmethod
     def _create_nn(cls, job_file_path):
-        """Create NN symbolic link, if necessary.
+        """Create NN symbolic link, or remove old job logs if they exist.
 
-        If NN => 01, remove numbered directories with submit numbers greater
-        than 01.
+        If NN => 01, remove numbered dirs with submit numbers greater than 01.
+
         Helper for "self._job_submit_impl".
 
         """
         job_file_dir = os.path.dirname(job_file_path)
+
         source = os.path.basename(job_file_dir)
         task_log_dir = os.path.dirname(job_file_dir)
         nn_path = os.path.join(task_log_dir, "NN")
@@ -393,6 +394,7 @@ class JobRunnerManager():
             old_source = None
         if old_source is None:
             os.symlink(source, nn_path)
+
         # On submit 1, remove any left over digit directories from prev runs
         if source == "01":
             for name in os.listdir(task_log_dir):
@@ -400,6 +402,11 @@ class JobRunnerManager():
                     # Ignore errors, not disastrous if rmtree fails
                     rmtree(
                         os.path.join(task_log_dir, name), ignore_errors=True)
+
+        # Delete old job logs if necessary
+        for name in JOB_LOG_ERR, JOB_LOG_OUT:
+            with suppress(FileNotFoundError):
+                os.unlink(os.path.join(job_file_dir, name))
 
     @classmethod
     def _filter_submit_output(cls, st_file_path, job_runner, out, err):
@@ -567,9 +574,6 @@ class JobRunnerManager():
 
         # Create NN symbolic link, if necessary
         self._create_nn(job_file_path)
-        for name in JOB_LOG_ERR, JOB_LOG_OUT:
-            with suppress(FileNotFoundError):
-                os.unlink(os.path.join(os.path.dirname(job_file_path), name))
 
         # Start new status file
         with open(f"{job_file_path}.status", "w") as job_status_file:
--- a/cylc/flow/job_runner_mgr.py
+++ b/cylc/flow/job_runner_mgr.py
@@ -568,8 +568,8 @@ class JobRunnerManager():
         # Create NN symbolic link, if necessary
         self._create_nn(job_file_path)
         for name in JOB_LOG_ERR, JOB_LOG_OUT:
-            with suppress(OSError):
-                os.unlink(os.path.join(job_file_path, name))
+            with suppress(FileNotFoundError):
+                os.unlink(os.path.join(os.path.dirname(job_file_path), name))
 
         # Start new status file
         with open(f"{job_file_path}.status", "w") as job_status_file:
--- a/tests/integration/test_job_runner_mgr.py
+++ b/tests/integration/test_job_runner_mgr.py
@@ -88,10 +88,9 @@ async def test_kill_error(one, start, test_dir, capsys, log_filter):
 async def test_create_nn_new(one, start, test_dir, capsys, log_filter):
     """Test _create_nn. 
 
-    It should the NN symlink.
+    It should create the NN symlink.
     """
     async with start(one):
-        # make it look like the task is running
         itask = one.pool.get_tasks()[0]
 
         workflow_job_log_dir = Path(get_workflow_run_job_dir(one.workflow))
--- a/tests/integration/test_job_runner_mgr.py
+++ b/tests/integration/test_job_runner_mgr.py
@@ -85,8 +85,8 @@ async def test_kill_error(one, start, test_dir, capsys, log_filter):
         assert itask.state(TASK_STATUS_RUNNING)
 
 
-async def test_create_nn_new(one, start, test_dir, capsys, log_filter):
-    """Test _create_nn. 
+async def test_create_nn_new(one, start):
+    """Test _create_nn.
 
     It should create the NN symlink.
     """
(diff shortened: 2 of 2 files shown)


cylc/flow/job_runner_mgr.py
```diff
@@ -568,8 +568,8 @@ def _job_submit_impl(
         # Create NN symbolic link, if necessary
         self._create_nn(job_file_path)
         for name in JOB_LOG_ERR, JOB_LOG_OUT:
-            with suppress(OSError):
-                os.unlink(os.path.join(job_file_path, name))
+            with suppress(FileNotFoundError):
+                os.unlink(os.path.join(os.path.dirname(job_file_path), name))
```
ColemanTom, line 572: Perhaps pull the dirname out of the loop for a minor speedup? Or would python be smart enough to do that itself? You could even do a tweak to _create_nn to accept the dirname rather than the job path so it doesn't get determined in multiple spots, `_create_nn` only needs the `01` type path.

hjoliver: I don't think Python would do that automatically.

tests/integration/test_job_runner_mgr.py
```diff
@@ -82,3 +83,59 @@ async def test_kill_error(one, start, test_dir, capsys, log_filter):
             level=logging.WARNING,
         )
         assert itask.state(TASK_STATUS_RUNNING)
+
+
+async def test_create_nn_new(one, start, test_dir, capsys, log_filter):
+    """Test _create_nn. 
+
+    It should the NN symlink.
```
hjoliver, line 91:
```suggestion
    It should create the NN symlink.
```

tests/integration/test_job_runner_mgr.py
```diff
@@ -82,3 +83,59 @@ async def test_kill_error(one, start, test_dir, capsys, log_filter):
...
+    """
+    async with start(one):
+        # make it look like the task is running
```
hjoliver, line 94:
```suggestion
```

hjoliver:
(Note all tests passed, 100% patch coverage, before I committed some non-coding changes with skip-ci)

tests/integration/test_job_runner_mgr.py
```diff
@@ -82,3 +83,58 @@ async def test_kill_error(one, start, test_dir, capsys, log_filter):
...
+    async with start(one):
+        itask = one.pool.get_tasks()[0]
+
+        workflow_job_log_dir = Path(get_workflow_run_job_dir(one.workflow))
+        job_id = itask.tokens.duplicate(job='01').relative_id
+        job_log_dir = Path(workflow_job_log_dir, job_id)
+        job_log_dir.mkdir(parents=True)
+
+        # call _create_nn
+        JobRunnerManager()._create_nn(job_log_dir / 'job.out')
```
MetRonnie, line 102: I think this is usually the path to the job script
```suggestion
        JobRunnerManager()._create_nn(job_log_dir / 'job')
```

tests/integration/test_job_runner_mgr.py
```diff
@@ -82,3 +83,58 @@ async def test_kill_error(one, start, test_dir, capsys, log_filter):
...
+            job_logs.append(job_log_dir / name)
+
+        # create the logs
+        for job_log in job_logs:
+            job_log.touch()
+
+        # check they exist
+        for job_log in job_logs:
+            assert job_log.is_file()
+
```
MetRonnie, line 133: Surely not needed? 😅
```suggestion
```

tests/integration/test_job_runner_mgr.py
```diff
@@ -82,3 +83,58 @@ async def test_kill_error(one, start, test_dir, capsys, log_filter):
             level=logging.WARNING,
         )
         assert itask.state(TASK_STATUS_RUNNING)
+
+
+async def test_create_nn_new(one, start, test_dir, capsys, log_filter):
+    """Test _create_nn.
```
MetRonnie, line 89:
```suggestion
async def test_create_nn_new(one, start):
    """Test _create_nn.
```

tests/integration/test_job_runner_mgr.py
```diff
@@ -82,3 +83,58 @@ async def test_kill_error(one, start, test_dir, capsys, log_filter):
...
+        job_log_dir.mkdir(parents=True)
+
+        # call _create_nn
+        JobRunnerManager()._create_nn(job_log_dir / 'job.out')
+
+        # check the symlink exists
+        assert (job_log_dir.parent / "NN").is_symlink()
+
+
+async def test_create_nn_old(one, start, test_dir, capsys, log_filter):
```
MetRonnie, line 108:
```suggestion
async def test_create_nn_old(one, start):
```

tests/integration/test_job_runner_mgr.py
MetRonnie, line 89:
```suggestion
async def test_create_nn_new(one, start):
    """Test _create_nn.
```

tests/integration/test_job_runner_mgr.py
```diff
@@ -82,3 +83,58 @@ async def test_kill_error(one, start, test_dir, capsys, log_filter):
...
+        for job_log in job_logs:
+            job_log.touch()
+
+        # check they exist
+        for job_log in job_logs:
+            assert job_log.is_file()
+
+        # call _create_nn
+        for job_log in job_logs:
+            JobRunnerManager()._create_nn(job_log)
```
MetRonnie, line 136:
```suggestion
        JobRunnerManager()._create_nn(job_log_dir / 'job')
```

MetRonnie:
Actually, is the unlink really needed if it has been broken this whole time?

wxtim (commented):
Nothing beyond what @MetRonnie said.

tests/integration/test_job_runner_mgr.py
hjoliver: (overzealous copy from another test!)

tests/integration/test_job_runner_mgr.py
hjoliver: Fair point, given the creation method (originally I tried to create them via Cylc methods)

hjoliver:
> **Actually, is the unlink really needed if it has been broken this whole time?**

I thought about this. Presumably the code was put in for a reason, probably (speculation) because it was common to re-run from scratch in an existing run directory in Cylc 7. If an old job log dir had `job.out` and `job.err` files from the previous run, that could leave confusing results after the new run - e.g. if the new run did not generate any stderr but the old one did.

In Cylc 8, it is still *possible* to force a re-run from scratch in an existing run dir (delete the .service dir first). So, it's unlikely to be needed, but perhaps still could be on rare occasions.

That said, I did not test the scenario. I'll try it now...

dpmatthews:
PBS does replace any existing out/err files but only when the job completes which can be confusing. More importantly, we rely on the existence of the job.out file to determine whether the job output is present when doing log retrieval so it's essential we remove any old files.

MetRonnie (approved):

ColemanTom:
One question, is there a reason this is going to master and not 8.3.7? Is 8.3.7 not going to exist as 8.4.0 will soon?

wxtim (approved):

wxtim:
> One question, is there a reason this is going to master and not 8.3.7? Is 8.3.7 not going to exist as 8.4.0 will soon?

releasing 8.4 this week

Issue #5736, opened 2023-09-19
mypy: python 3.10+


oliver-sanders:
Mypy appears to be raising false positives in imported code for Python 3.10+, so currently, we're only running it for Python 3.9.

We will need to address this going forward.

The problem:
* Matplotlib has dropped Python 3.7 support in their latest releases.
* MyPy is complaining for non 3.7 compatible syntax in the non 3.7 compatible matplotlib code.
* But these non compatible versions shouldn't get installed for 3.7 anyway (the install tool *should* handle this).
* Adding an upper pin for Matplotlib which only applies to Python 3.7 doesn't make this go away.

Because `no_silence_site_packages=false` (the default) I wouldn't expect any errors from imported code anyway?

See https://github.com/cylc/cylc-flow/pull/5735

MetRonnie:
https://github.com/python/mypy/issues/9972

I don't think we can do anything about this

oliver-sanders:
@MetRonnie, closed by #6809?

PR #6534, opened 2024-12-27
Fix flake8 E231

MetRonnie:
For some reason, Flake8 was not detecting some problems on python 3.9.

**Check List**

- [x] No dependency changes 
- [x] Tests are not needed
- [x] No docs needed

### DIFF
--- a/.github/workflows/test_fast.yml
+++ b/.github/workflows/test_fast.yml
@@ -108,12 +108,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      # note: exclude python 3.10+ from mypy checks as these produce false
-      # positives in installed libraries for python 3.7
       - name: Configure Python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.9
+          python-version: 3.x
 
       - name: Install
         run: |
--- a/cylc/flow/scripts/dump.py
+++ b/cylc/flow/scripts/dump.py
@@ -314,7 +314,7 @@ async def dump(workflow_id, options, write=print):
                             result += " (" + ",".join(attrs) + ")"
                         if options.show_flows:
                             result += (
-                                f" flows={item['flowNums'].replace(' ','')}"
+                                f" flows={item['flowNums'].replace(' ', '')}"
                             )
                         write(result)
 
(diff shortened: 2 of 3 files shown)


.github/workflows/test_fast.yml
```diff
@@ -130,7 +128,8 @@ jobs:
         run: etc/bin/shellchecker
 
       - name: MyPy
-        run: mypy
+        # specify --python-version to avoid errors regarding site-packages
+        run: mypy --python-version 3.8
```
oliver-sanders, line 132: 3.7?

MetRonnie:
```
$ mypy --python-version=3.7
[...]/python3.12/site-packages/matplotlib/pyplot.py:246: error: Assignment expressions are only supported in Python 3.8 and greater  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```

I suppose to be safe we will have to split off mypy into its own job 😒

oliver-sanders: > Assignment expressions are only supported in Python 3.8 and greater

I think mypy has reported a legit error here that the tests failed to spot! Thankfully this issue only applies to the developer main loop plugins so is of little consequence. Would introducing an upper pin on matplotlib clear the error?

MetRonnie: This version of matplotlib would not install on Python 3.7 so there's no error here. It's just that Mypy can't work in Python 3.7 mode when it comes across 3.7-incompatible code in 3rd-party libraries: https://github.com/python/mypy/issues/9972

MetRonnie: See also https://github.com/cylc/cylc-flow/issues/5736

MetRonnie: Wait I just realised that since we moved to linting in its own job on GH Actions, we're not installing matplotlib during that job so this error is not raised anymore!

oliver-sanders (approved):

.github/workflows/test_fast.yml
oliver-sanders: Ach dammit, we should probably install the `all` group, one for another day...

oliver-sanders:
(re-running the failed test)

PR #6535, opened 2024-12-30
Kill tasks during job prep

### LINKED ISSUE #5746: Kill should abort preparing tasks

MetRonnie:
Supersedes #5749, closes #5746 

This PR allows `cylc kill` [and `cylc remove`] to kill preparing tasks, resulting in the submit-failed state.

> Tested by deliberately extending job preparation like this (and kill tasks whilst still in the preparing state)
> 
> ```
> platform = $(sleep 10; echo localhost)
> ```

**Check List**

- [x] No dependency changes 
- [x] No docs needed.


hjoliver (commented):
Tested as working, with a task somewhere in the preparing state: the killed preparing task goes to submit-failed, and gets a new submit number if retriggered.

Definitely nicer than my hack, if it covers all the bases. Should this work at any stage of the "job preparation pipeline"? (I don't recall why I tried my other approach, but I think I was worried about this aspect for some reason ... hopefully unnecessarily worried).

One thing: presumably the killed task should get held as well, to prevent automatic retries (until released).

oliver-sanders:
This doesn't actually *kill* the job preparation (e.g. if killed during remote-init, we won't actually kill the remote init command itself). This is probably the best we can do as truly "recalling" a preparation operation is tricky and may have unwanted consequences (e.g. a partially remote-installed workflow).

The orphaned operation (e.g. remote-init) is still tracked by the subprocpool and may still be logged, example:

<details>
<summary>setup to reproduce</summary>

```cylc
# global.cylc

[platforms]
  [[myplatform]]
    rsync command = whatevs
    job runner = background
```

```bash
# ~/bin/whatevs

sleep 20
exit 42
```

```console
$ cylc vip myworkflow
$ sleep 2
$ cylc kill 'myworkflow//*'
```

</details>

```
INFO - platform: myplatform - remote init (on myplatform)
INFO - platform: myplatform - remote file install (on myplatform)
INFO - Command "kill_tasks" received. ID=6b6c8363-3bc8-4c99-ba76-9e68a1cb4b5f
    kill_tasks(tasks=['1/foo'])
INFO - [1/foo/01:preparing] => preparing(held)
ERROR - [jobs-submit cmd] (killed in job prep)
    [jobs-submit ret_code] 1
CRITICAL - [1/foo/01:preparing(held)] submission failed
INFO - [1/foo/01:preparing(held)] => submit-failed(held)
WARNING - [1/foo/01:submit-failed(held)] did not complete the required outputs:
    ⨯ ┆  succeeded
INFO - Command "kill_tasks" actioned. ID=6b6c8363-3bc8-4c99-ba76-9e68a1cb4b5f
ERROR - Incomplete tasks:
    * 1/foo did not complete the required outputs:
      ⨯ ┆  succeeded
CRITICAL - Workflow stalled
WARNING - P3D stall timer starts NOW
INFO - testing connectivity for myplatform using ['ssh', '-oBatchMode=yes', '-oConnectTimeout=8', '-oStrictHostKeyChecking=no', 'myplatform', 'true']
ERROR - platform: myplatform - initialisation did not complete
    COMMAND:
        meme --delete \
            --rsh=ssh -oBatchMode=yes -oConnectTimeout=8 -oStrictHostKeyChecking=no \
            --include=/.service/ --include=/.service/server.key -a \
            --checksum --out-format=%o %n%L --no-t --exclude=log \
            --exclude=share --exclude=work --include=/ana/*** \
            --include=/app/*** --include=/bin/*** --include=/etc/*** \
            --include=/lib/*** --exclude=* \
            /home/users/oliver.sanders/cylc-run/tmp.d6SQ6gMhfZ/run11/ \
            myplatform:cylc-run/tmp.d6SQ6gMhfZ/run11/
    RETURN CODE:
        42
```

Strangely, in this situation, if I trigger a downstream task, it goes straight to submit-failed with no attempt to re-run the failed remote-fileinstall operation (bug). But if I re-trigger the downstream task a second time, then it repeats both the remote-init and remote-fileinstall (bug?)!?

> Should this work at any stage of the "job preparation pipeline"

From testing it appears to abort at any stage :+1: (though I'm not entirely sure how this is guaranteed).

We could do with some tests to ensure that this works at each stage and that the following stages are correctly canceled:

* host-select
* bash syntax check
* remote-init
* remote-fileinstall

Hopefully easy to do with integration tests where we can mock the results of each stage of the process.

MetRonnie:
Strictly speaking this "kills the task" but doesn't kill any preparation process it has started.

I think it works by setting `itask.waiting_on_job_prep = False` so that the task will no longer submit a job once the orphaned preparation process finishes. And `_prep_submit_task_job_error()` creates a dummy submit-failed job.

I'll have a look at testing more thoroughly.

hjoliver:
My approach, on the superseded PR, was to set a new flag on the preparing task and then, based on that flag, abort job submission once prep had completed. There must have been a reason why I did not just use `itask.waiting_on_job_prep = False`. Unfortunately i don't recall. Doesn't that flag actually affect the job prep pipeline?

MetRonnie:
The bash syntax check is blocking (not using the subproc pool) so I don't think it matters?

https://github.com/cylc/cylc-flow/blob/26b12e4d88bd7353d3b394d7a7aa779be63217a4/cylc/flow/job_file.py#L80-L93

cylc/flow/task_job_mgr.py
```diff
@@ -346,7 +346,7 @@ def submit_livelike_task_jobs(
                     bc_mgr = self.task_events_mgr.broadcast_mgr
                     rtconf = bc_mgr.get_updated_rtconfig(itask)
                     try:
-                        platform = get_platform(
+                        platform = get_platform(  # type: ignore[assignment]
```
MetRonnie, line 349: Addressed in #6564

cylc/flow/task_job_mgr.py
```diff
@@ -1278,7 +1278,7 @@ def _prep_submit_task_job(
                     workflow, itask, '(platform not defined)', exc)
                 return False
             else:
-                itask.platform = platform
+                itask.platform = platform  # type: ignore[assignment]
```
MetRonnie, line 1281: Addressed in #6564

tests/integration/test_kill.py
```diff
@@ -0,0 +1,167 @@
...
+        await task_state(itask, TASK_STATUS_SUBMIT_FAILED, is_held=True)
+        assert log_filter(logging.ERROR, 'killed in job prep')
+
+
+async def test_kill_preparing_pipeline(
+    flow, scheduler, run, monkeypatch: pytest.MonkeyPatch
+):
+    """Test killing a preparing task through various stages of the preparing
+    pipeline that involve submitting subprocesses and waiting for them to
+    complete."""
```
MetRonnie, line 112: @oliver-sanders Is this what you were after?

cylc/flow/scheduler.py
```diff
@@ -1562,6 +1556,30 @@ def start_job_submission(self, itasks: 'Iterable[TaskProxy]') -> bool:
...
+            )
+        )
+
+        # submit "live" mode tasks (and "dummy" mode tasks)
+        submitted_live_tasks = self.task_job_mgr.submit_livelike_task_jobs(
+            self.workflow,
+            itasks,
+            self.server.curve_auth,
+            self.server.client_pub_key_dir,
+        )
```
oliver-sanders, line 1579: I'm not a big fan of moving this task submission logic into the scheduler.

I'm guessing you did this to avoid having to populate the extra args in integration tests, fair enough.

Maybe just leave a shim in scheduler.py to do this, would rather keep submission logic in one place, Cylc logic has a habit of sprawling.

oliver-sanders (approved):
:+1:

cylc/flow/scheduler.py
MetRonnie: https://github.com/cylc/cylc-flow/compare/378e65b97cd5fefb123a523b204d2057057c50e9..d71cd087a6d1a7e7e93dbbf9c3a0fbdf83c9bbb7

hjoliver (approved):
LGTM. The tests allay the concerns I mentioned earlier, re the preparing pipeline.

PR #6546, opened 2025-01-09
logging: update header timestamp on rollover

oliver-sanders:
When the scheduler log rolls over, the header lines get logged with the original timestamp rather than the rollover time which is a tad confusing.

Test using:
```cylc
# global.cylc

[scheduler]
  [[logging]]
    maximum size in bytes = 200
```

**Check List**

### DIFF
--- a/cylc/flow/loggingutil.py
+++ b/cylc/flow/loggingutil.py
@@ -31,6 +31,7 @@ from pathlib import Path
 import re
 import sys
 import textwrap
+from time import time
 from typing import List, Optional, Union
 
 from ansimarkup import parse as cparse, strip as cstrip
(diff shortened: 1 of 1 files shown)


cylc/flow/loggingutil.py
```diff
@@ -238,15 +239,22 @@ def do_rollover(self) -> None:
             os.dup2(self.stream.fileno(), sys.stderr.fileno())
         # Emit header records (should only do this for subsequent log files)
         for header_record in self.header_records:
+            now = time()
```
wxtim, line 242: IMO it's entirely reasonable not to test assigning the output of a standard library function, and marking this as not needing testing gives us a better idea of what is and isn't tested.

```suggestion
            now = time()   # pragma: no cover
```

wxtim (approved):
- [x] Read code.
- [x] Manually tested.

Doesn't have a unit test, but perhaps the game isn't worth the candle for this change.

cylc/flow/loggingutil.py
oliver-sanders: I'm confused, codecode says this is covered:

> [codecov/patch](https://github.com/cylc/cylc-flow/pull/6546/checks?check_run_id=35367886718) — 100.00% of diff hit (target 97.00%)

wxtim: image

oliver-sanders: Screenshot from 2025-01-15 16-36-48

PR #6551, opened 2025-01-10
Lint S014: Only warn about `-W` when used for LSF

### LINKED ISSUE #6550: `cylc lint` mistakenly picks up PBS `-W` directive as LSF time limit directive
### Description

> The -W option	allows for the specification of additional job attributes.

https://www.jlab.org/hpc/PBS/qsub.html

### Reproducible Example

```cylc
[runtime]
  [[foo]]
    platform = pbs_platform
    [[[directives]]]
      -W umask = 0022
```

### Expected Behaviour

`cylc lint` S014 should look for `-W =` for LSF rather than just `-W`

wxtim:
Closes #6550 

`-W` should only be highlighted by Cylc lint when being used in a way which indicates use of LSF.

**Check List**

- [x] Very small, self documenting.
- [x] Raised against 8.4.x, bugfix

### DIFF
--- a/changes.d/6551.fix.md
+++ b/changes.d/6551.fix.md
@@ -1,0 +1,1 @@
+Lint: Fix bug where lint warned about use of legitimate -W directive for PBS.
--- a/changes.d/6551.fix.md
+++ b/changes.d/6551.fix.md
@@ -1,1 +1,1 @@
-Lint: Fix bug where lint warned about use of legitimate -W directive for PBS.
+Fix bug in `cylc lint` S014 where it warned about use of legitimate `-W` directive for PBS.
--- a/cylc/flow/scripts/lint.py
+++ b/cylc/flow/scripts/lint.py
@@ -189,8 +189,12 @@ def get_wallclock_directives():
             'TIME_LIMIT_DIRECTIVE',
             None
         )
-        if directive:
-            directives[module.name] = directive
+        if directive and directive == '-W':
+            # LSF directive -W needs to have a particular form to
+            # avoid matching PBS directive -W:
+            directives[module.name] = re.compile(r'^-W\s+(\d+:)?\d+$')
+        elif directive:
+            directives[module.name] = re.compile(rf'^{directive}')
     return directives
 
 
@@ -209,9 +213,17 @@ def check_wallclock_directives(line: str) -> Union[Dict[str, str], bool]:
         >>> this = check_wallclock_directives
         >>> this('    -W 42:22')
         {'directive': '-W 42:22'}
+        >>> this('    -W 42:22/hostname')  # Legit LSF use case
+        False
+        >>> this('    -W 422')
+        {'directive': '-W 422'}
+        >>> this('    -W foo=42')  # Legit PBS use case.
+        False
+        >>> this('    -W foo="Hello World"')  # Legit PBS use case.
+        False
     """
     for directive in set(WALLCLOCK_DIRECTIVES.values()):
-        if line.strip().startswith(directive):
+        if directive.findall(line.strip()):
             return {'directive': line.strip()}
     return False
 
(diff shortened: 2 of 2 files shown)


cylc/flow/scripts/lint.py
```diff
@@ -209,9 +213,11 @@ def check_wallclock_directives(line: str) -> Union[Dict[str, str], bool]:
         >>> this = check_wallclock_directives
         >>> this('    -W 42:22')
         {'directive': '-W 42:22'}
+        >>> this('    -W foo=42')
```
wxtim, line 216:
```suggestion
        >>> this('    -W foo=42')  # Legit PBS use case.
```

cylc/flow/scripts/lint.py
```diff
@@ -209,9 +213,17 @@ def check_wallclock_directives(line: str) -> Union[Dict[str, str], bool]:
         >>> this = check_wallclock_directives
         >>> this('    -W 42:22')
         {'directive': '-W 42:22'}
+        >>> this('    -W 42:22/hostname')  # Legit LSF use case
```
wxtim, line 216: Looking at https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=o-w-1, I think that this is an LSF wallclock directive which can't be handled by `execution time limit` and therefore is a legitimate case for using `-W` with LSF.

oliver-sanders: Can you do multiple things with this opt?

e.g `-W 00:02 foo=bar`

wxtim:
> I found some suggestion in the user guide that the time can take a longer format:

This regex is only supposed to match LSF format wallclock strings, avoiding PBS `-W` which look superficially similar. 

The PBS check one will match any case where `-l wallclock` is present.

cylc/flow/scripts/lint.py
wxtim: In LSF or PBS?

In LSF, not really.

In PBS, not really - only key=value pairs.

I don't think that the two are mistakeable.

changes.d/6551.fix.md
```diff
@@ -0,0 +1 @@
+Lint: Fix bug where lint warned about use of legitimate -W directive for PBS.
```
MetRonnie, line 1:
```suggestion
Fix bug in `cylc lint` S014 where it warned about use of legitimate `-W` directive for PBS.
```

oliver-sanders:
Can't find any false positives in the first 2000 workflows scanned, :+1:

cylc/flow/scripts/lint.py
```diff
@@ -189,15 +189,19 @@ def get_wallclock_directives():
             'TIME_LIMIT_DIRECTIVE',
             None
         )
-        if directive:
-            directives[module.name] = directive
+        if directive and directive == '-W':
+            # LSF directive -W needs to have a particular form to
+            # avoid matching PBS directive -W:
+            directives[module.name] = re.compile(r'^-W\s+(\d+:)?\d+$')
```
oliver-sanders, line 195: I think we need to allow for an `=` char here, else we'll get false negatives for LSF directives?

MetRonnie, line 195: What about trailing spaces or comments?
```suggestion
            directives[module.name] = re.compile(r'^-W\s+(\d+:)?\d+[^/]*$')
```

cylc/flow/scripts/lint.py
oliver-sanders: Gotcha

cylc/flow/scripts/lint.py
```diff
@@ -189,15 +189,19 @@ def get_wallclock_directives():
             'TIME_LIMIT_DIRECTIVE',
             None
         )
-        if directive:
-            directives[module.name] = directive
+        if directive and directive == '-W':
+            # LSF directive -W needs to have a particular form to
+            # avoid matching PBS directive -W:
+            directives[module.name] = re.compile(r'^-W\s*=?\s*(\d+:)?\d+$')
```
wxtim, line 195:
```suggestion
            directives[module.name] = re.compile(r'^-W\s*=?\s*(\d+:)?\d+[^/]*$')
```

oliver-sanders (approved):
FYI: There's an (ironic?) lint error on this branch to resolve.

MetRonnie (approved):

PR #6552, opened 2025-01-11
lint: handle multiline strings & jinja2 code

### LINKED ISSUE #6348: `cylc lint`: errors raised for multi-line Jinja2 blocks
```console
$ cat flow.cylc 
#!Jinja2

{% set FOO=[
    1
] %}

$ cylc lint .
[S002] flow.cylc:5: Item not indented.

Checked ... against ['728', 'style'] rules and found 1 issue.

$ echo $?
1
```

### LINKED ISSUE #6507: `cylc lint` is catching things in multi-line strings
### Description

`cylc lint` is flagging items in multi-line strings. I expected these lines to be skipped by the linter.

### Reproducible Example

For this `flow.cylc` file
```ini
[meta]
    title = This is my workflow
    description = """
        This workflow demonstrates how linting flags things in multi-long comments

        Just wanted to say this is how you set UTC mode...

        [scheduler]
            UTC mode = True
    """

[scheduler]
    UTC mode = True

[scheduling]
    initial cycle point = 2024-01-01T00:00Z
    final cycle point = 2024-01-02T00:00Z
```

When I run `cylc lint` I get this...

```
[S003] flow.cylc:8: Top level sections should not be indented.

Checked <myWorkflowDirectory> against ['728', 'style'] rules and found 1 issue.
```

### Expected Behaviour

Line 8 is part of the multiline description and is not "cylc". I expected the linter to ignore these lines.

### LINKED ISSUE #6685: Cylc lint: U9999 is over-eagre
I found `U999` was complaining about something it should not be complaining about. Take this snippet.

```
$ cat flow.cylc
    [[xtriggers]
        foo = abc(a=123, host=hello)
        foo = abc(a=123, \
            host=hello \
        )
$ CYLC_VERSION=8.4 cylc lint $PWD
[U999] flow.cylc:4: [runtime][__MANY__][workflow state polling]host - not available at Cylc 8

Checked /g/ns/sc/user-cylc/tcoleman/code/basic-workflow against ['728', 'style'] rules and found 1 issue.
$ CYLC_VERSION=8.4 cylc version
8.4.1
```

When host is on a new line, it throws an error, thinking it belongs to `workflow state polling`, even though it is part of an xtrigger and host is a completely valid and reasonable kwarg.

wxtim:
Closes #6348 
Closes #6507 
Closes #6685 

**Check List**

### DIFF
--- a/cylc/flow/lint/state.py
+++ b/cylc/flow/lint/state.py
@@ -57,7 +57,9 @@ class LinterState:
             self.is_metadata_section = False
 
         if self.is_metadata_section:
-            if self.TRIPLE_QUOTES.findall(line):
+            # Start quoted section if there is an odd number of triple quotes
+            # i.e. don't match """stuff""".
+            if len(self.TRIPLE_QUOTES.findall(line)) % 2 == 1:
                 self.is_multiline_chunk = not self.is_multiline_chunk
             if self.is_multiline_chunk:
                 return True
--- a/cylc/flow/lint/state.py
+++ b/cylc/flow/lint/state.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Cylc Linter state store object.
+"""
+
+from dataclasses import dataclass
+import re
+
+
+@dataclass
+class LinterState:
+    """A place to keep linter state"""
+    TRIPLE_QUOTES = re.compile(r'\'{3}|\"{3}')
+    JINJA2_START = re.compile(r'{%')
+    JINJA2_END = re.compile(r'%}')
+    NEW_SECTION_START = re.compile(r'^[^\[]*\[[^\[]')
+    is_metadata_section: bool = False
+    is_multiline_chunk: bool = False
+    is_jinja2_block: bool = False
+    jinja2_shebang: bool = False
+    line_no: int = 1
+
+    def skip_line(self, line):
+        """Is this a line we should skip, according to state we are holding
+        and the line content?
+
+        TODO: Testme
+        """
+        return any((
+            self.skip_metatadata_desc(line),
+            self.skip_jinja2_block(line)
+        ))
+
+    def skip_metatadata_desc(self, line):
+        """Should we skip this line because it's part of a metadata multiline
+        description section.
+
+        TODO: Testme
+        """
+        if '[meta]' in line:
+            self.is_metadata_section = True
+        elif self.is_metadata_section and self.is_end_of_meta_section(line):
+            self.is_metadata_section = False
+
+        if self.is_metadata_section:
+            if self.TRIPLE_QUOTES.findall(line):
+                self.is_multiline_chunk = not self.is_multiline_chunk
+            if self.is_multiline_chunk:
+                return True
+
+        return False
+
+    def skip_jinja2_block(self, line):
+        """Is this line part of a jinja2 block?
+
+        TODO: Testme
+        """
+        if self.jinja2_shebang:
+            if (
+                self.JINJA2_START.findall(line)
+                and not self.JINJA2_END.findall(line)
+            ):
+                self.is_jinja2_block = True
+            elif self.is_jinja2_block and self.JINJA2_END.findall(line):
+                self.is_jinja2_block = False
+                return True
+
+        return self.is_jinja2_block
+
+    @staticmethod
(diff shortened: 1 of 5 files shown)


wxtim:
@oliver-sanders  Please could you have a look at this proof of concept before I bother testing it?

oliver-sanders:
I think this is going to be the quickest way to get over the state barrier, but going statefull is a good opportunity to review our approach here before we lay down a pattern which ends up biting us later.

I haven't really given much thought to this so far, I don't have working knowledge of modern linters, a bit of reading would be a good idea...

Mid to long term, it depends how far we want to go with the linter. The initial idea was a quick and dirty line-by-line regex tool, but we thought we would leave it there (we decided it would not be statefull). Now we are clearly at the point where we want to develop this into something more serious.

I think the optimal solution for this sort of problem is something more tokeniser'ey. I.e, a program that reads the file into labeled bits rather than lines, allowing rules to be run on each "bit" in a more structured way than regex would allow by itself.

E.g, here's an example of how a tokeniser'ey type thing might see the file:

```python
>>> for type, content, state in tokeniser(file):
...    print((type, content))
('shebang', '#!Jinja2', <file:flow.cylc, lines:[0], preproc:jinja2, path:>)
('blankline', '', <file:flow.cylc, lines:[1], preproc:jinja2, path:>)
('section', '[scheduling]', <file:flow.cylc, lines:[2], preproc:jinja2, path:[scheduling]>)
('config', '    initial cycle point = 2000', <file:flow.cylc, lines:[3], preproc:jinja2, path:[scheduling]initial cycle point>)
('cycle point', '2000', <file:flow.cylc, lines:[3], preproc:jinja2, path:[scheduling]initial cycle point>)
('section', '    [[graph]]', <file:flow.cylc, lines:[4], preproc:jinja2, path:[scheduling][graph]>)
('config', '        R1 = """\nfoo => bar\nbar => baz\n    """', <file:flow.cylc, lines:[5,6,7,8], path:[scheduling][graph]>)
('graph', 'foo => bar\nbar => baz', <file:flow.cylc, lines:[6,7,8], path:[scheduling][graph]R1>)
('jinja2', '{% if ... \n ... %}', <file:flow.cylc, lines:[9,10], path:[scheduling][graph]>)
('jinja2', '{% include ... %}', <file:flow.cylc, lines:[11], path:?>)
```

It's similar (but less detailed & more annotated) to what our syntax highlighters see. Note, we are carrying the Cylc config "path" where the information is available allowing us to apply specific rules to particular sections or configs and each entry could represent multiple lines. This would allow us, for example, to apply rules, only to graph sections, and to do clever things, like read a few lines ahead to check if different task configs are compatible.

It might be relatively simple to get a paired-down version of this working, e.g something that has only awareness of "plain" Cylc lines and Jinja2 expressions.

There might also be the option of building a linter into one of the Cylc grammars we have already developed? I wouldn't be surprised if this is becoming common in the LSP era.

I mention these things largely to curtail the scope of what we do now. It might be that the Cylc linter never gets much further than it is now, in which case this is a pretty decent pattern. But it is also possible that we end up developing this much further and the community starts contributing rulesets. So we should be cautious of how far we push this pattern, if it's starting to look like this utill might balloon, we should be ready to change tack rather than bang our heads against the pattern we've got. If there's a pattern out there we could easily adopt, this might be a good opportunity to do so before we start trying to get clever with our line reader. And if we do go ahead with this (and I don't think that would be a bad option), we should be cautious over how far we push it.

hjoliver:
Much as I like the linter (I do!) we should think carefully, and maybe look at how widely it is used, before sinking a lot of effort into making it significantly more general than it is now (I think this gels with what Oliver is saying, but in more general terms). (Priorities vs resources, as ever...)

oliver-sanders:
We've had another multiline lint issue reported https://github.com/cylc/cylc-flow/issues/6685 (nothing to do with Jinja2 though this time).

cylc/flow/lint/state.py
```diff
@@ -0,0 +1,112 @@
...
+from dataclasses import dataclass
+import re
+
+
+@dataclass
+class LinterState:
+    """A place to keep linter state"""
+    TRIPLE_QUOTES = re.compile(r'\'{3}|\"{3}')
+    JINJA2_START = re.compile(r'{%')
+    JINJA2_END = re.compile(r'%}')
```
MetRonnie, line 29:
```suggestion
    JINJA2_START = re.compile(r'{%|{{')
    JINJA2_END = re.compile(r'%}|}}')
```

cylc/flow/lint/state.py
```diff
@@ -0,0 +1,112 @@
...
+        return any((
+            self.skip_metatadata_desc(line),
+            self.skip_jinja2_block(line)
+        ))
+
+    def skip_metatadata_desc(self, line):
+        """Should we skip this line because it's part of a metadata multiline
+        description section.
+
+        TODO: Testme
```
MetRonnie, line 52: TODO

cylc/flow/lint/state.py
```diff
@@ -0,0 +1,112 @@
...
+                self.is_multiline_chunk = not self.is_multiline_chunk
+            if self.is_multiline_chunk:
+                return True
+
+        return False
+
+    def skip_jinja2_block(self, line):
+        """Is this line part of a jinja2 block?
+
+        TODO: Testme
```
MetRonnie, line 72: TODO

cylc/flow/lint/state.py
```diff
@@ -0,0 +1,112 @@
...
+    is_multiline_chunk: bool = False
+    is_jinja2_block: bool = False
+    jinja2_shebang: bool = False
+    line_no: int = 1
+
+    def skip_line(self, line):
+        """Is this a line we should skip, according to state we are holding
+        and the line content?
+
+        TODO: Testme
```
MetRonnie, line 41: TODO

wxtim:
Low priority - Branch & PR still available, but not currently worked on

PR #6554, opened 2025-01-13
Add run mode skip to cylc show

wxtim:
**Check List**


cylc/flow/scripts/show.py
```diff
@@ -346,9 +347,15 @@ async def prereqs_and_outputs_query(
                     attrs.append("queued")
                 if t_proxy['isRunahead']:
                     attrs.append("runahead")
+                if (
+                    t_proxy['runtime']['runMode']
+                    and t_proxy['runtime']['runMode'] == 'Skip'
```
oliver-sanders, line 352: Use the RunMode enum?

cylc/flow/scripts/show.py
```diff
@@ -346,9 +347,15 @@ async def prereqs_and_outputs_query(
...
+                ):
+                    attrs.append("skip")
```
oliver-sanders, line 354: * Not especially obvious what `(skip)` means. Perhaps `run-mode=skip`.
* Could also loosen this to cover other run modes, i.e. `and not t_proxy[runtime]['runMode'] == 'live'`...
* Plz, check this value updates with broadcasts.

wxtim: > * Not especially obvious what `(skip)` means. Perhaps `run-mode=skip`.

Yes. Will do.

> * Could also loosen this to cover other run modes, i.e. `and not t_proxy[runtime]['runMode'] == 'live'`...

This would require change to the data stored by sim and dummy modes. Because they are only settable for the workflow the logic to store the info about the tasks isn't there. Will have a look.

> * Plz, check this value updates with broadcasts.

Putting a test up

tests/integration/scripts/test_show.py
```diff
@@ -266,3 +287,48 @@ async def test_task_instance_state_flows(
         ] == [
             expected_flows,
         ]
+
+
+async def test_mode_changes(flow, scheduler, start, capsys):
```
oliver-sanders, line 292:
```suggestion
async def test_task_run_mode_changes(flow, scheduler, start, capsys):
```

tests/integration/scripts/test_show.py
```diff
@@ -266,3 +287,48 @@ async def test_task_instance_state_flows(
...
+    """
+    opts = SimpleNamespace(
+        comms_timeout=5,
+        json=False,
+        task_defs=None,
+        list_prereqs=False,
+    )
+    schd = scheduler(
+        flow({'scheduling': {'graph': {'R1': 'a'}}}),
+        paused_start=True,
```
oliver-sanders, line 303: I don't think this is needed:

```suggestion
```

cylc/flow/scripts/show.py
```diff
@@ -346,9 +347,15 @@ async def prereqs_and_outputs_query(
                     attrs.append("queued")
                 if t_proxy['isRunahead']:
                     attrs.append("runahead")
+                if (
+                    t_proxy['runtime']['runMode']
+                    and t_proxy['runtime']['runMode'] != 'Live'
```
oliver-sanders, line 352: Use enum?

cylc/flow/scripts/show.py
wxtim: Ay - meant to, forgot.

cylc/flow/network/schema.py
```diff
@@ -633,7 +633,8 @@ class Meta:
 # The run mode for the task.
 TaskRunMode = graphene.Enum(
     'TaskRunMode',
-    [(m.capitalize(), m) for m in TASK_CONFIG_RUN_MODES],
+    [(k.capitalize(), k.lower()) for k in RunMode.__members__.keys()],
+    # [(m.capitalize(), m) for m in TASK_CONFIG_RUN_MODES],
     description=lambda x: RunMode(x.value).describe() if x else None,
 )
```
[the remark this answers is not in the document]
MetRonnie: Doesn't it make more sense to keep this the way it was? Simulation and Dummy are not task run modes, they are workflow run modes.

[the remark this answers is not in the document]
wxtim: I think too much can be made of the distinction between task and workflow run modes; they both change the execution of individual tasks. The only difference is how the user can apply them.

@oliver-sanders  - are you suggesting that I carry on with this PR, but try to put up a companion on cylc-ui to filter out the `simulation` and `dummy`?

changes.d/6554.feat.md
```diff
@@ -0,0 +1 @@
+Cylc show now displays that a task has been set to skip mode
```
MetRonnie, line 1:
```suggestion
`cylc show` now displays when a task has been set to skip mode
```

cylc/flow/network/schema.py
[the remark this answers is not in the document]
oliver-sanders: > Doesn't it make more sense to keep this the way it was? Simulation and Dummy are not task run modes, they are workflow run modes.

No, they **are** definitely task run modes, we just don't let people configure them in the workflow config.

I.E, they are readonly values set by the scheduler.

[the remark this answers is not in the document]
oliver-sanders: > @oliver-sanders - are you suggesting that I carry on with this PR, but try to put up a companion on cylc-ui to filter out the simulation and dummy?

Undecided. Edit runtime assumes, not unreasonably, that all possible values of an enum are also valid values for the user to set. Though this isn't actually true of other types (i.e. not all possible strings are valid inputs for a string config).

We could probs filter out these two options by registering a custom type with the form generator which would disable the field if set to certain values?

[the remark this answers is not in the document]
MetRonnie: Is there an issue for this? What are we trying to achieve here? If it's showing when a task is in skip mode in `cylc show`, then surely the simplest option is to leave this Enum alone?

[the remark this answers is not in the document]
wxtim: The idea is that this is the first step of making the task submission mode visible in the GUI.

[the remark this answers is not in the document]
oliver-sanders: @MetRonnie, the task run mode is a field in the data store and something we want to use in `cylc show` and the GUI.

If the workflow is run in simulation mode, then all of the tasks in the workflow will be run in simulation mode (same for dummy). If we do nothing, then this field will incorrectly report that the task is running in live mode. I.E, this is a bug fix.

Tim's change to the Enum is correct, the only issue is the way Edit Runtime is using it. Not necessarily a biggie, but sooner or later, someone is going to try to put a task into simulation mode and will be surprised by the result.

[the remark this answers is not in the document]
MetRonnie: > If the workflow is run in simulation mode, then all of the tasks in the workflow will be run in simulation mode (same for dummy).

Right, so this can simply be handled in `cylc show` by using the task run mode if it's set to Skip else falling back to the workflow run mode? This seems like the simplest solution for now until we implement a way of showing the task run mode in the GUI outside of the Edit Runtime form.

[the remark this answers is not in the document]
oliver-sanders: >  If going down this route, it ought to be an 8.5 change.

Agreed, this should be an 8.5 change whatever we decide to do (feature not a bugfix).


> Note any changes we make to the Enum in the GraphQL schema will affect workflows running in older versions of the UI/UIS 

In theory, the way to handle schema changes is to add a shim in the UIS to provide a default value for the new config for older schedulers which don't provide it. Surprisingly, we haven't had to do this yet.


> Right, so this can be handled in cylc show by using the workflow run mode if it's Simulation or Dummy, else falling back to the task run mode?

Problem with doing that is that the task's `run_mode` will be wrong in the data whenever the workflow run mode is anything other than live. We could cover this up, both in `cylc show` and in the GUI as you suggest, but that would still leave the "Edit Runtime" form displaying the wrong value and allowing the user to edit it (e.g. change live -> skip when the workflow is running in sim mode) even though the change would be invalid. So we would have to change the code in three places (incl Edit Runtime) to smooth this over, so not really that simple.

Options:
1. Ignore the issue, it's not a biggie as long as the error message from broadcast is clear and well presented. This only impacts Edit Runtime.
2. Register a new type with the form generator that only allows this to be set to live/skip and only if the workflow mode is live.
3. Leave the run_mode as live even when the workflow isn't running in live mode, bodge the relevant interfaces to show an adjusted value.
4. Bump this config out of `Runtime` and into the `TaskProxy` in order to exclude it from EditRuntime (would still leave and incorrect value in the store).

Looking ahead to the future, the run mode abstraction opens the door to lots of exciting things, e.g `run_mode = sub-workflow` or `run_mode = xtrigger`. This will make the rules for changing run modes a tad more complicated (i.e. you can only really change the run mode between live and skip).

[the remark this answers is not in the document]
wxtim: 3 was what I initially considered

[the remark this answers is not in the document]
MetRonnie: > We could cover this up... but that would still leave the "Edit Runtime" form displaying the wrong value and allowing the user to edit it (e.g. change live -> skip when the workflow is running in sim mode) even though the change would be invalid

Most things you can change in the Edit Runtime form are not valid for simulation mode!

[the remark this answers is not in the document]
oliver-sanders: I'm not sure which way you're arguing here?

I'm leaning towards option 1: Ignore the issue, it's not a biggie.

* The task's `run_mode` field should be set to the mode that the task will (or did) run in (i.e. live, skip, sim or dummy).
* The only issue with this is with the edit runtime form.
* But invalid broadcasts will result in an error anyway so this issue of little consequence, we just need to make sure the error is clearly presented.
* If we were worried about this issue, we could use a custom Enum so we don't offer users run mode changes that will be rejected by broadcast anyway.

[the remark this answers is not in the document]
MetRonnie: Will discuss at the team meeting

wxtim:
Note to self - failing tests

```
tests/f/cylc-show/05-complex.t
tests/f/xtriggers/04-sequential.t
tests/f/n-window/01-past-present-future.t
tests/f/cylc-show/06-past-present-future.t
tests/f/n-window/02-big-window.t
tests/k/cylc-show/00-simple.t
tests/k/cylc-show/04-multi.t
```

cylc/flow/network/schema.py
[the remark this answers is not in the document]
MetRonnie: > Register a new type with the form generator that only allows this to be set to live/skip and only if the workflow mode is live.

Ok, I understand how to do this now, will open a UI PR for it shortly

cylc/flow/run_modes/__init__.py
```diff
@@ -85,6 +85,14 @@ def get(options: 'Values') -> "RunMode":
             return RunMode(run_mode)
         return RunMode.LIVE
 
+    @classmethod
+    def _missing_(cls, value):
+        value = value.lower()
+        for member in cls:
+            if member.lower() == value:
+                return member
+        return None
```
MetRonnie, line 94: What's this about?

MetRonnie, line 94: Surprised to learn about this. The fact that it's "sunder" (single underscore) is weirding me out. https://docs.python.org/3/library/enum.html#enum.Enum._missing_

cylc/flow/run_modes/__init__.py
```diff
@@ -40,7 +40,7 @@
     ]
 
 
-class RunMode(Enum):
+class RunMode(str, Enum):
```
MetRonnie, line 43: I'm not sure I understand subclassing `str`

wxtim:
```python
RunMode('Live') == 'Live'
True
```

MetRonnie: I'm a bit torn on this. I see its convenience, but also I think leaning too heavily on the convenience could result in problems. For example, your example doesn't work!
```
In [1]: from cylc.flow.run_modes import RunMode

In [2]: RunMode('Live') == 'Live'
Out[2]: False
```
Because the case doesn't match the enum's value.

Without subclassing str, normally Mypy will flag this is a non-overlapping equality check.

wxtim: Don't worry is't not a big deal (i.e. I'm happy to get rid.)

changes.d/6554.feat.md
MetRonnie: Still pending

cylc/flow/scripts/show.py
```diff
@@ -346,9 +348,15 @@ async def prereqs_and_outputs_query(
                     attrs.append("queued")
                 if t_proxy['isRunahead']:
                     attrs.append("runahead")
+                if (
+                    t_proxy['runtime']['runMode']
+                    and RunMode(t_proxy['runtime']['runMode']) != RunMode.LIVE
+                ):
+                    attrs.append(f"run mode={t_proxy['runtime']['runMode']}")
```
MetRonnie, line 355:
```suggestion
                run_mode = t_proxy['runtime']['runMode']
                if run_mode and RunMode(run_mode) != RunMode.LIVE:
                    attrs.append(f"run mode={run_mode}")
```

MetRonnie:
UI accompaniment: https://github.com/cylc/cylc-ui/pull/2057

Btw, there is 1 conflict here now

cylc/flow/config.py
```diff
@@ -2447,6 +2448,13 @@ def _get_taskdef(self, name: str) -> TaskDef:
 
         try:
             rtcfg = self.cfg['runtime'][name]
+
+            # If the workflow mode is simulation or dummy always
+            # override the task config:
+            workflow_run_mode = RunMode.get(self.options)
+            if workflow_run_mode.value in WORKFLOW_ONLY_MODES:
+                rtcfg['run mode'] = workflow_run_mode.value
```
MetRonnie, line 2456: FYI broadcasting a run mode of live or skip seems to override this

MetRonnie (approved):

cylc/flow/config.py
oliver-sanders: Correct right?

MetRonnie: Shouldn't be able to broadcast task run mode = live when the workflow is running in simulation mode? Not that it matters too much

oliver-sanders: Ah right, yes. Unfortunately rather hard to prevent but will result in `run mode` being set (and reported) with an erroneous value :(

wxtim:
@oliver-sanders is this just low priority, or been missed?

oliver-sanders:
@wxtim, there's one unresolved comment (`ShowOptions`) and one unresolved discussion (might just be left as a caveat, but needs your comment).

cylc/flow/config.py
wxtim: I believe that I've now modified broadcast to ignore `[<namespace]run mode` and warn the user that run mode has been ignored if the workflow run mode is set to simulation or dummy.

tests/integration/scripts/test_show.py
```diff
@@ -293,12 +295,7 @@ async def test_task_instance_state_flows(
 async def test_task_run_mode_changes(flow, scheduler, start, capsys):
     """Broadcasting a change of run mode changes run mode shown by cylc show.
     """
-    opts = SimpleNamespace(
-        comms_timeout=5,
-        json=False,
-        task_defs=None,
-        list_prereqs=False,
-    )
+    opts = Options(get_option_parser())()
```
MetRonnie, line 298: (Could have just done)
```py
from cylc.flow.scripts.show import ShowOptions

opts = ShowOptions()
```

Issue #6298, opened 2024-08-13
data-store: incorrect information for n>0 xtriggers


oliver-sanders:
Xtriggers are not evaluated until at least one pre-requisite has been satisfied (where the "runahead limit" is effectively handled as a prerequisite for parentless tasks for this purpose), however, Cylc interfaces e.g. `cylc show` will list them irrespective.

This can cause confusion. E.g. If you inspect a runahead-limited task with a `wall_clock` xtrigger, the xtrigger may be shown as unsatisfied despite being past the trigger time. This is a bug as Cylc is making it look like a `wall_clock` xtrigger is returning `False` rather than explaining that it hasn't started running yet.

Solutions:

* Add a third xtrigger state.
  * E.G. "satisfied", "unsatisfied", or "not yet evaluated".
  * This would need to be a new field (e.g. "status") with the old "satisfied" field being marked as deprecated in order to ensure compatibility with older schedulers.
* Don't list xtriggers until the task has entered n=0.
  * This could cause confusion as it makes it look like the xtrigger does not exist.
  * Cylc interfaces could have some logic to protect against this e.g. put a placeholder message saying "xtriggers pending" or whatnot.
* Add a dummy xtrigger for n>0 tasks and replace it with real xtriggers at spawn time.
  * The dummy xtrigger could have a message along the lines of "xtriggers not yet started".
* Other?

dwsutherland:
2 is a no go for me (like you said, just as confusing)

I think 1 is a good option `pending` or `not yet evaluated` ..
Should be easy.. may want `n/a` or `-` instead or `x` or tick in the UIs/dumps.

Is the prerequisites `n/a for past tasks` a front end message or generated at `cylc-flow`?
Because we may not need a state if it's implied but some other info

hjoliver:
I agree with David, option 1 is best.

oliver-sanders:
Ok, this looks like a consensus, going with option 1.

oliver-sanders:
> Is the prerequisites n/a for past tasks a front end message or generated at cylc-flow?

https://github.com/cylc/cylc-flow/blob/a991faed5d120d04d3a6d620166b93b0ca9985a7/cylc/flow/scripts/show.py#L344

PR #6560, opened 2025-01-16
xtriggers: differentiate "not satisfied" from "not yet evaluated"

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/6298
* The `cylc show` command will now convey if an xtrigger has not yet started polling so that it doesn't look like n>0 xtriggers are not satisfied even if the external condition they will poll for is

**Check List**


hjoliver:
When this goes in, need to update this Note in the User Guide, via https://github.com/cylc/cylc-doc/pull/820 

image

PR #6568, opened 2025-01-20
GH Actions: Drop Ubuntu 20.04

MetRonnie:
Ubuntu 20.04 going away by 1 April. Brownouts starting in March. https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/

**Check List**

- [x] No dependency changes or docs or tests etc

### DIFF
--- a/.github/workflows/test_manylinux.yml
+++ b/.github/workflows/test_manylinux.yml
@@ -33,7 +33,7 @@ jobs:
       fail-fast: false
       matrix:
         manylinux: ['1']
-        os: ['ubuntu-20.04']  # run on the oldest linux we have access to
+        os: ['ubuntu-22.04']  # run on the oldest linux we have access to
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
 
     steps:


oliver-sanders:
I'm not sure this action will actually protect us from much once the image is updated (again).

Plz hold off on this until March so we use the older image for as long as we are able. After that we might just have to drop the check :(

MetRonnie:
How do we know if the test is valid or not on Ubuntu 22?

oliver-sanders:
It's not really valid on 20, but it's better than nothing.

It restricts the manylinux versions that can be used to install wheels. It will fallback to building from source if that fails, this build may fail on older boxes for various reasons. Ideally we would use an ancient SLES image, but we don't have access to that, so we used an ancient Ubuntu image which was close enough. But the more we upgrade the box, the more invalid the comparison becomes and the less protection the test offers.

MetRonnie:
Of the pip packages installed in the Python 3.7 run, it looks like the following are built from source on Ubuntu 22:
- psutil
- pyzmq
- numpy
- PyYAML

The rest of them are either `*-any.whl` or manylinux:
- Pillow-9.5.0-cp37-cp37m-manylinux_2_28_x86_64.whl
- matplotlib-3.5.3-cp37-cp37m-manylinux_2_5_x86_64.manylinux1_x86_64.whl
- coverage-7.2.7-cp37-cp37m-manylinux_2_5_x86_64.manylinux1_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl
- kiwisolver-1.4.5-cp37-cp37m-manylinux_2_5_x86_64.manylinux1_x86_64.whl

oliver-sanders:
Added to Cylc dev meeting agenda: https://github.com/cylc/cylc-admin/issues/200#issuecomment-2607783777

I don't think this workflow is really valid any more, every time we update the base image, the test drifts further from what we need. Suggest deleting this action in March and falling back to manual testing until we can retire our manylinux1 requirement.

PR #6571, opened 2025-01-23
Assume task parameters with underscores to be strings.

wxtim:
PEP-515 allowed integers to be defined with underscores from python 3.6.
This changes the behaviour of Cylc task parameters, as `"123_456"` can now be interpreted as an integer.

n.b. I've tried to get 100% coverage on the method under test.

I'm going to have a quick play around with the code to ensure that we've not missed anything else about this change that needs addressing.

**Check List**

- [ ] Double checked PEP-515
- [ ] Used parameter generation to check the method.


cylc/flow/parsec/validate.py
```diff
@@ -1007,22 +1007,30 @@ def coerce_parameter_list(cls, value, keys):
             >>> CylcConfigValidator.coerce_parameter_list('a, b, c', None)
             ['a', 'b', 'c']
 
+            >>> CylcConfigValidator.coerce_parameter_list('084_132', None)
+            ['084_132']
+
+            >>> CylcConfigValidator.coerce_parameter_list('72, a', None)
+            ['72', 'a']
```
oliver-sanders, line 1014: I think this test probably passed before, maybe worth trying a zero prefix?

```suggestion
            >>> CylcConfigValidator.coerce_parameter_list('072, a', None)
            ['072', 'a']
```

oliver-sanders (commented):
LGTM.

This could do with a changelog. We might want to put it on the 8.4.x branch as it's more of a fix than a feature.

cylc/flow/parsec/validate.py
```diff
@@ -1007,22 +1007,30 @@ def coerce_parameter_list(cls, value, keys):
             >>> CylcConfigValidator.coerce_parameter_list('a, b, c', None)
             ['a', 'b', 'c']
 
+            >>> CylcConfigValidator.coerce_parameter_list('084_132', None)
+            ['084_132']
+
+            >>> CylcConfigValidator.coerce_parameter_list('072, a', None)
```
wxtim, line 1013: This Fixes a coverage hole.

tests/unit/parsec/test_validate.py
```diff
@@ -699,7 +699,7 @@ def test_coerce_parameter_list():
             ('-15, -10, -5, -1..1', [-15, -10, -5, -1, 0, 1])]:
         assert validator.coerce_parameter_list(value, ['whatever']) == result
     # The bad
-    for value in ['foo/bar', 'p1, 1..10', '2..3, 4, p']:
+    for value in ['foo/bar', 'p1, 1..10', '2..3, 4, p', 'x:,']:
```
wxtim, line 702: Fix a coverage hole.

oliver-sanders (approved):

cylc/flow/parsec/validate.py
MetRonnie: Oliver's comment hasn't been resolved

changes.d/6571.fix.md
```diff
@@ -0,0 +1 @@
+Task parameters with single underscores and numbers only were being treated as integers (Python 3.6, PEP-515). This change returns older behaviour.
```
MetRonnie, line 1:
```suggestion
Disabled PEP-515-style integer coercion of task parameters containing underscores (e.g. `084_132` was becoming `84132`). This fix returns older behaviour seen in Cylc 7.

```

cylc/flow/parsec/validate.py
```diff
@@ -1007,22 +1007,30 @@ def coerce_parameter_list(cls, value, keys):
...
+                if can_only_be is str:
                     raise IllegalValueError(
                         'parameter', keys, value, 'mixing int range and str')
                 can_only_be = int
                 items.extend(values)
             elif cls._REC_NAME_SUFFIX.match(item):  # noqa: SIM106
                 try:
+                    if '_' in item:
+                        raise ValueError(
+                            'Artifical value error undoing PEP-515')
```
MetRonnie, line 1030: I don't understand this bit

wxtim: If there is an underscore in the string we don't want it treated as a number but to skip to line 1032, so we artificially raise a value error. 

Since it is in a try-except loop, the only purpose of the string is explanatory.

cylc/flow/parsec/validate.py
wxtim: I've accepted it, because it does not harm, but Oliver's original comment missed the point that this test fixes a coverage hole rather than being related to the matter in hand.

cylc/flow/parsec/validate.py
MetRonnie: Ah got it. Perhaps
```suggestion
                    if '_' in item:
                        # Disable PEP-515 int coercion; go to except block:
                        raise ValueError()
```

MetRonnie (approved):
LGTM, tested, got 1 suggestion

Issue #6429, opened 2024-10-17
command validation: broadcasts


oliver-sanders:
(Easier after https://github.com/cylc/cylc-flow/issues/6605)

If you try to broadcast an invalid configuration on the CLI, you get a helpful error:

```console
$ cylc broadcast generic -s 'elephant = shrew'
IllegalItemError: elephant
```

However, if you try to broadcast an invalid configuration from the GUI, you get no error, and a cryptic message in the log:

```
2024-10-17T15:14:55Z INFO - Command "broadcast" received.
    broadcast(cycle_points=['*'], mode=put_broadcast, namespaces=[None], settings=[{'elephant': 'shrew'}])
Traceback (most recent call last):
...
graphql.error.located_error.GraphQLLocatedError: elephant

2024-10-17T15:14:55Z ERROR - elephant
2024-10-17T15:14:55Z ERROR - graphql.error.located_error.GraphQLLocatedError: elephant
```

The broadcast command is unlike other commands in that it is performed entirely at command queue time (broadcasts aren't queued to be processed later). It's also implemented outside of `cylc.flow.commands` for historical reasons.

~Suggest moving the broadcast code into `cylc.flow.commands` (so it runs like a normal command) as this will give it the same error handling as other commands. But keep it all queue-time. Doing this should also forward the error to the client.~

[edit] This isn't easy to do, https://github.com/cylc/cylc-flow/issues/6605 covers this.

PR #6574, opened 2025-01-24
broadcast: report invalid settings

oliver-sanders:
broadcast: report invalid settings
    
* We already report invalid cycle points and namespaces.
* However, we have been relying on client-side validation for settings  (which doesn't apply to the GraphQL mutation).
* This also raises the potential for inter-version compatibility issues going unreported.
* This commit explicitly handles invalid settings in the same way as invalid cycle points and namespaces so that they are reported back to the user.
* Closes the issue part of #6429.
* Additionally:
  * This change also strips duplicate entries from broadcast reports.
  * And fixes the CLI options in broadcast report to match `cylc broadcast`.

**Check List**

### DIFF
--- a/cylc/flow/broadcast_mgr.py
+++ b/cylc/flow/broadcast_mgr.py
@@ -37,7 +37,7 @@ from cylc.flow.platforms import (
     fail_if_platform_and_host_conflict,
     PlatformLookupError,
 )
-
+from cylc.flow.util import uniq
 
 if TYPE_CHECKING:
     from cylc.flow.id import Tokens
@@ -151,7 +151,7 @@ class BroadcastMgr:
             LOG.error(get_broadcast_bad_options_report(bad_options))
         if modified_settings:
             self.data_store_mgr.delta_broadcast()
-        return modified_settings, bad_options
+        return uniq(modified_settings), bad_options
 
     def expire_broadcast(self, cutoff=None, **kwargs):
         """Clear all broadcasts targeting cycle points earlier than cutoff."""
@@ -284,18 +284,24 @@ class BroadcastMgr:
           bad_options is as described in the docstring for self.clear().
         """
         modified_settings = []
-        bad_point_strings = []
-        bad_namespaces = []
+        bad_settings = []
+        bad_point_strings = set()
+        bad_namespaces = set()
 
         with self.lock:
             for setting in settings or []:
                 # Coerce setting to cylc runtime object,
                 # i.e. str to  DurationFloat.
                 coerced_setting = deepcopy(setting)
-                BroadcastConfigValidator().validate(
-                    coerced_setting,
-                    SPEC['runtime']['__MANY__'],
-                )
+                try:
+                    BroadcastConfigValidator().validate(
+                        coerced_setting,
+                        SPEC['runtime']['__MANY__'],
+                    )
+                except Exception as exc:
+                    LOG.error(exc)
+                    bad_settings.append(setting)
+                    continue
 
                 # Skip and warn if a run mode is broadcast to a workflow
                 # running in simulation or dummy mode.
@@ -308,6 +314,7 @@ class BroadcastMgr:
                         f' running in {self.workflow_run_mode.value} mode'
                         ' will have no effect, and will not be actioned.'
                     )
+                    bad_settings.append(setting)
                     continue
 
                 for point_string in point_strings or []:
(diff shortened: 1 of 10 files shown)


oliver-sanders:
Note to self, if #6554 is merged before this, change the code to return the invalid broadcast to the user rather than just logging it server side:

https://github.com/cylc/cylc-flow/pull/6554/files#diff-713a6d5d5fb4768549c43ae7920483b64760ac4e1c5ef68206b9111cf688d15eR295-R306

wxtim:
> Note to self, if #6554 is merged before this, change the code to return the invalid broadcast to the user rather than just logging it server side:
> 
> https://github.com/cylc/cylc-flow/pull/6554/files#diff-713a6d5d5fb4768549c43ae7920483b64760ac4e1c5ef68206b9111cf688d15eR295-R306

6554 has been merged.

cylc/flow/util.py
```diff
@@ -45,6 +45,45 @@
...
+    Examples:
+        >>> uniq([1, 1, 2, 3, 5, 8, 1])
+        [1, 2, 3, 5, 8]
+
+    """
+    ret = []
+    for item in iterable:
+        if item not in ret:
+            ret.append(item)
+    return ret
```
wxtim, line 63:
```suggestion
return list(iter_uniq(iterable))
```

wxtim (approved):
1 small suggestion.
- [x] Manually tested as working.
- [x] Read - seems reasonable.

cylc/flow/util.py
oliver-sanders: (note, this is a code move from the resolvers module)

I suspect it's probably more efficient to build a list directly, than to build a set, then build a list from the set.

wxtim: No.

```python
import sys
sys.version
```




    '3.8.0 | packaged by conda-forge | (default, Nov 22 2019, 19:11:38) \n[GCC 7.3.0]'




```python
def uniq(iterable):
    """Return a unique collection of the provided items preserving item order.
    Useful for unhashable things like dicts, relies on __eq__ for testing
    equality.
    """
    ret = []
    for item in iterable:
        if item not in ret:
            ret.append(item)
    return ret


def simple_uniq(iterable):
    """wrapper for iter_uniq"""
    return list(iter_uniq(iterable))


def iter_uniq(iterable):
    """Iterate over an iterable omitting any duplicate entries.
    Useful for unhashable things like dicts, relies on __eq__ for testing
    equality.
    """
    cache = set()
    for item in iterable:
        if item not in cache:
            cache.add(item)
            yield item


```


```python
from random import randint
MAX = 100
SAMPLE = sorted([randint(0, MAX) for i in range(MAX)])
SAMPLE[:10]
```




    [0, 1, 2, 2, 3, 6, 9, 9, 9, 10]




```python
%%timeit
uniq(SAMPLE)
```

    45 µs ± 695 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)



```python
%%timeit
simple_uniq(SAMPLE)
```

    8.74 µs ± 199 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)



```python

```

oliver-sanders: Ok, would you like to raise a PR to replace `uniq` calls with `list(iter_uniq)`?

wxtim: Yes, but it will conflict with this PR because you've just moved the functions.

oliver-sanders: Can build atop this branch or attempt after.

tests/integration/run_modes/test_simulation.py
```diff
@@ -424,10 +424,12 @@ async def test_settings_broadcast(
 
         # Assert that setting run mode on a simulation mode task fails with
         # warning:
-        schd.broadcast_mgr.put_broadcast(
+        good, bad = schd.broadcast_mgr.put_broadcast(
             ['1066'], ['one'], [{
                 'run mode': 'live',
             }])
+        assert good == []
+        assert bad == {'settings': [{'run mode': 'live'}]}
```
oliver-sanders, line 432: If you broadcast an incompatible run mode, the error will now be returned to the client (but will still be logged server-side).

Extended this test to cover the new behaviour.

cylc/flow/broadcast_mgr.py
```diff
@@ -303,22 +309,24 @@ def put_broadcast(
                         f' running in {self.workflow_run_mode.value} mode'
                         ' will have no effect, and will not be actioned.'
                     )
+                    bad_settings.append(setting)
```
oliver-sanders, line 312: Return this error to the client.

oliver-sanders:
(mypy failure unrelated)

wxtim (approved):
Rechecked. Still happy.

oliver-sanders:
(rebased)

cylc/flow/broadcast_mgr.py
```diff
@@ -340,6 +348,7 @@ def put_broadcast(
                                 namespace,
                                 coerced_setting,
                             ):
+                                bad_settings.append(setting)
```
oliver-sanders, line 351: @wxtim I have rebased this onto master which now contains your recent change to detect incompatible `[remote]host` / `platform` broadcasts.

This line will report the setting back to the client. I've extended the new tests in `test_broadcast_mgr` to assert this.

hjoliver (approved):
Good 💯

PR #6577, opened 2025-01-28
Get poll to return task failure if job/log has been removed.

### LINKED ISSUE #6425: handle job log directory deleted for active task
Spotted in the wild!

If you delete the job log directory for an active task, Cylc will preserve its last known status indefinitely. I.e, Cylc will consider the job to be submitted/running forever.

In this case it was caused by a housekeep task being triggered whilst other tasks in the cycle were still running. The housekeep task tarred up the `log/job/<cycle>` dir removing the job status files in the process.

This situation should be handled similarly to the job no longer appearing in the queue, i.e, the job is dead, long live the job. Stick it into the failed/submit-failed state as appropriate.

wxtim:
Closes #6425 

> [!NOTE]
> Note to reviewers, you will need to deploy this branch onto remote platforms to confirm it works for remote filesystems.

> Finally have a replicable example (Thank you [@oliver-sanders](https://github.com/oliver-sanders))
> ```
> [scheduling]
>     cycling mode = integer
>     [[graph]]
>         R1 = task
> 
> [runtime]
>     [[task]]
>         script = """
>             rm ${CYLC_WORKFLOW_RUN_DIR}/.service/contact
>             rm -r "${CYLC_WORKFLOW_RUN_DIR}/log/job/${CYLC_TASK_CYCLE_POINT}/${CYLC_TASK_NAME}"
>         """
>         platform = _remote_pbs
> ```

**Check List**


tests/unit/test_job_runner_mgr.py
wxtim, line 1: Only `test__job_poll_status_files_deleted_logdir` is directly related to the PR. Other tests should increase coverage. 😄

cylc/flow/task_job_mgr.py
```diff
@@ -753,6 +753,15 @@ def _manip_task_jobs_callback(
...
             LOG.error(ctx)
+        # A polling task lets us know that a task has failed because it's
+        # log folder has been deleted whilst the task was active:
+        if (
+            getattr(ctx, 'out', None)
+            and 'Job files have been removed' in ctx.out
+        ):
+            LOG.error(
+                f'Task {ctx.cmd[-1]} failed because task log directory'
+                f'\n    {"/".join(ctx.cmd[-2:])}\n    has been removed.')
```
oliver-sanders, line 764: The logger auto-indents hanging lines before they are written to the scheduler log:

```suggestion
                f'Task {ctx.cmd[-1]} failed because task log directory'
                f'\n{"/".join(ctx.cmd[-2:])}\nhas been removed.'
            )
```

cylc/flow/job_runner_mgr.py
```diff
@@ -439,6 +439,12 @@ def _filter_submit_output(cls, st_file_path, job_runner, out, err):
     def _jobs_poll_status_files(self, job_log_root, job_log_dir):
         """Helper 1 for self.jobs_poll(job_log_root, job_log_dirs)."""
         ctx = JobPollContext(job_log_dir)
+        # If the log directory has been deleted prematurely, return a task
+        # failure and an explanation:
+        if not os.path.exists(os.path.join(job_log_root, ctx.job_log_dir)):
+            ctx.run_status = 1
+            ctx.run_signal = 'ERR/Job files have been removed'
```
oliver-sanders, line 446: Should make this a constant to avoid duplicating this string in different places.

oliver-sanders (commented):
LGTM.

oliver-sanders:
Note to reviewers, you will need to deploy this branch onto remote platforms to confirm it works for remote filesystems.

cylc/flow/job_runner_mgr.py
wxtim: It's only used once in `cylc/flow` - are you suggesting importing it into the tests?

oliver-sanders: It's actually used twice in Cylc flow (but only once with the ERR/ prefix). Importing this into the jobs is a good idea.

cylc/flow/job_runner_mgr.py
```diff
@@ -439,6 +442,12 @@ def _filter_submit_output(cls, st_file_path, job_runner, out, err):
     def _jobs_poll_status_files(self, job_log_root, job_log_dir):
         """Helper 1 for self.jobs_poll(job_log_root, job_log_dirs)."""
         ctx = JobPollContext(job_log_dir)
+        # If the log directory has been deleted prematurely, return a task
+        # failure and an explanation:
+        if not os.path.exists(os.path.join(job_log_root, ctx.job_log_dir)):
+            ctx.run_status = 1
+            ctx.run_signal = 'ERR/' + JOB_FILES_REMOVED_MESSAGE
```
MetRonnie, line 449: I think this would be more conventional. `ERR/` is likely to be confused with the `failed/ERR` message
```suggestion
            ctx.run_signal = 'ERR_JOB_FILES_REMOVED'
```

See
https://github.com/cylc/cylc-flow/blob/3a9cd6a3e17dbb9d794e579d25a83daf62f402e3/cylc/flow/job_runner_mgr.py#L119

wxtim: I think it needs that as a prefix?

MetRonnie: `ERR` is not a prefix

MetRonnie:
From the original issue:

> If you delete the job log directory for an active task, Cylc will preserve its last known status indefinitely. I.e, Cylc will consider the job to be submitted/running forever.

This does not seem to be true. It is only true if the job log dir *and* the contact file is removed. The job may well succeed even if its job log dir is removed from under its feet.

oliver-sanders:
> This does not seem to be true. It is only true if the job log dir and the contact file is removed.

Hmmm, I think this is true. Note this is about remote job platforms.

oliver-sanders (commented):
* Code makes sense.
* Tested against the example in the OP, works well, error message clearly logged.

We should talk this one over in tomorrow's meeting.

MetRonnie:
Another example:
```
rm -r "${CYLC_WORKFLOW_RUN_DIR}/log/job/${CYLC_TASK_CYCLE_POINT}/${CYLC_TASK_NAME}"
sleep 2
exit 1
```

This task "fails successfully" but then the workflow hangs on shutdown after `cylc stop`

MetRonnie:
Discussed today:
- Although the job may yet succeed even if the job log dir is deleted, we have decided that it is best to put in the failed state as this is the best we can do if we can't poll anymore.
- We will leave the job log retrieval as it is, as it's user error to delete the job log dir prematurely they will have to suffer the consequences

MetRonnie (commented):
I have some suggestions at https://github.com/wxtim/cylc/pull/72

wxtim:
> I have some suggestions at [wxtim#72](https://github.com/wxtim/cylc/pull/72)

I'm happy with your suggestions. You wield your scapel with a confident hand. I like it.

MetRonnie (approved):

oliver-sanders (approved):

PR #6499, opened 2024-11-28
Trigger now if workflow paused

hjoliver:
Close #5963
Supesede #6192 (it was getting messy)
Close #6398

------

New behaviour for triggering when the workflow is paused

**DEFAULT: run triggered tasks NOW even if the workflow is paused**
  - supports the common use case *"I want to pause the workflow and just trigger individual tasks"*
  - (NIWA priority raised by both ops and research users in early Cylc 8 migration meetings)

**OPTION: `--on-resume` - run triggered tasks only when the paused workflow is resumed**
  - this supports [the broadcast-to-past-tasks workaround](https://github.com/cylc/cylc-flow/pull/6192#issuecomment-2486785465) until broadcast is fixed (prob. 8.5.0)
  - the option can be removed in the future, once broadcast is fixed, if there are no other use cases
  

-----

Implementation:
- separate start of job submission from queue release, in the scheduler
- manual trigger adds tasks to `tasks_to_trigger_(now|on_resume)` lists, after state reset and queue wrangling
- if these lists are populated, the scheduler will release tasks from them as appropriate
- (queues are not processed if the scheduler is paused)

-------

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [x] Changelog entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at https://github.com/cylc/cylc-doc/pull/778
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #6578, opened 2025-01-28
Improve server-client communication error handling

MetRonnie:
In #6499 we added a new field to the `trigger` GraphQL mutation, and this unfortunately broke the ability for `cylc trigger` at 8.4 to work on workflows running in 8.3 (I shall term this "inter-version server-client comms").

This PR does not fix that (we don't think a fix is feasible at this stage).

An additional problem is that the error handling for server-client comms is broken - that is what this PR fixes. And it attempts to ensure we have defined format for server-client comms which should reduce the chance of breaking inter-version server-client comms in future.

#### Before

```console
# Workflow running at Cylc 8.3.6, CLI at Cylc 8.4.0
$ cylc trigger wflow//1/foo
Error processing command:
    AttributeError: 'list' object has no attribute 'values'
```

#### After

```console
# Workflow running at Cylc 8.3.6, CLI at this Cylc <this PR>
$ cylc trigger wflow//1/foo
Error processing command:
    Exception: This command is no longer compatible with the version of Cylc running the workflow. Please stop & restart the workflow with Cylc 8.4.1.dev or higher.

[{'error': {'message': 'Unknown argument "onResume" on field "trigger" of type "Mutations".', 'traceback': ['graphql.error.base.GraphQLError: Unknown argument "onResume" on field "trigger" of type "Mutations".\n']}}]
```

#### Check List

- [x] No dependency changes 
- [x] No docs needed


tests/integration/test_client.py
```diff
@@ -88,3 +91,35 @@ async def test_command_validation_failure(harness):
...
+    one, start, monkeypatch: pytest.MonkeyPatch, sock_response, expected
+):
+    async def mock_recv():
+        return json.dumps(sock_response).encode()
+
+    async with start(one):
+        client = WorkflowRuntimeClient(one.workflow)
+        with monkeypatch.context() as mp:
+            mp.setattr(client, 'socket', Mock(recv=mock_recv))
+            mp.setattr(client, 'poller', Mock())
```
MetRonnie, line 122: For future reference, I originally tried only mocking the `poll` method of `client.poller`, i.e.
```python
mp.setattr(client.poller, 'poll', Mock())
```

but this resulted in pytest hanging near the end of running all integration tests; some strange interaction due to the ZMQ poller using threading or something, dunno.

cylc/flow/network/multi.py
```diff
@@ -220,21 +231,28 @@ def _process_response(
...
 
     Returns:
         (stdout, stderr, outcome)
 
     """
     try:
         ret: List[Tuple[Optional[str], Optional[str], bool]] = []
+        if not isinstance(response, dict):
+            if isinstance(response, list) and response[0].get('error'):
+                # If operating on workflow running in older Cylc version,
```
wxtim, line 252: I know that you are trying to catch as much as possible for the future, but might we be able to identify particular problems  and provide a more obvious error?

```diff
--- a/cylc/flow/network/multi.py
+++ b/cylc/flow/network/multi.py
@@ -38,6 +38,13 @@ from cylc.flow.id_cli import parse_ids_async
 from cylc.flow.terminal import DIM
 
 
+KNOWN_ERRS = {
+    'Unknown argument "onResume" on field "trigger" of type "Mutations".':
+        'Tasks in workflows running using versions of Cylc older than 8.4.0'
+        ' cannot be triggered using Cylc 8.4.0.'
+}
+
+
 def call_multi(*args, **kwargs):
     """Call a function for each workflow in a list of IDs.
 
@@ -251,6 +258,10 @@ def _report(
             if isinstance(response, list) and response[0].get('error'):
                 # If operating on workflow running in older Cylc version,
                 # may get a error response like [{'error': '...'}]
+                if response[0].get('error').get('message') in KNOWN_ERRS:
+                    raise CylcError(
+                        KNOWN_ERRS[response[0].get('error').get('message')]
+                    )
                 raise Exception(response)
             raise Exception(f"Unexpected response: {response}")
         for mutation_response in response.values():

```

tests/integration/test_client.py
wxtim, line 1: Probably over pedantic (and should be a separate commit to any other changes to avoid mixing with code change), but I think it'd be nicer if these tests were in `tests/i/network/`.

wxtim (approved):
I've made two suggestions, but neither should block this PR. :)

cylc/flow/network/multi.py
MetRonnie: I've gone for a slightly simpler version

wxtim: Fair enough.

wxtim (approved):

cylc/flow/network/multi.py
MetRonnie: (Reason: hopefully we won't need to add to this!)

cylc/flow/network/__init__.py
```diff
@@ -30,34 +35,71 @@
...
+        back-compat issues."""
 
 
-def decode_(message):
-    """Convert an encoded message string to JSON with an added 'user' field."""
+def load_server_response(message: str) -> 'ResponseDict':
+    """Convert a JSON message string to dict with an added 'user' field."""
     msg = json.loads(message)
-    msg['user'] = getpass.getuser()  # assume this is the user
+    if 'user' not in msg:
```
oliver-sanders, line 101: Under what circumstances would this be set already, and should we trust it if it is?

cylc/flow/network/multi.py
```diff
@@ -220,21 +239,36 @@ def _process_response(
...
+                # may get a error response like [{'error': '...'}]
+                if response[0]['error'].get('message') in KNOWN_INCOMPAT:
+                    raise Exception(
```
oliver-sanders, line 263: Is raising an exception here the right thing to do?

We may be operating on multiple workflows at multiple versions. Some of the operations may have succeeded, some may still be being processed.

cylc/flow/network/server.py
```diff
@@ -367,41 +402,24 @@ def graphql(
...
-                    'meta': meta or {},
-                },
-                backend=CylcGraphQLBackend(),
-                middleware=list(instantiate_middleware(self.middleware)),
-                executor=AsyncioExecutor(),
-                validate=True,  # validate schema (dev only? default is True)
-                return_promise=False,
-            )
-        except Exception as exc:
-            return 'ERROR: GraphQL execution error \n%s' % exc
```
oliver-sanders, line 385: Where does this error get caught?

cylc/flow/network/server.py
```diff
@@ -367,41 +402,24 @@ def graphql(
...
-                            'message': str(error),
-                            'traceback': formatted_tb
-                        }
-                    })
-                    continue
-                errors.append(getattr(error, 'message', None))
-            return errors
+                LOG.warning(error)
+            if not executed.data:
+                raise Exception(executed.errors[0])
```
oliver-sanders, line 422: Where does this error get caught?

oliver-sanders (commented):
Can you help me understand what bug this fixes.

The changes I can see:
* messages/traceback field removed
* cylc_version field added
* Errors now force exceptions in multi mode (was this the bug?).
* Refactoring (please avoid doing this in bugfix PRs).
* Nicer error for the trigger incompatibility issue.

cylc/flow/network/__init__.py
```diff
@@ -30,34 +35,71 @@
...
+        Going forward, we include this so we can more easily handle any future
+        back-compat issues."""
 
 
-def decode_(message):
-    """Convert an encoded message string to JSON with an added 'user' field."""
+def load_server_response(message: str) -> 'ResponseDict':
+    """Convert a JSON message string to dict with an added 'user' field."""
     msg = json.loads(message)
-    msg['user'] = getpass.getuser()  # assume this is the user
```
oliver-sanders, line 60: The reason we had an encode/decode interface here was to abstract out the transport format in order to allow it to be changed (we've changed it once in Cylc 8 from JSON web tokens to JSON).

cylc/flow/network/__init__.py
```diff
@@ -30,34 +35,71 @@
...
-def encode_(message):
-    """Convert the structure holding a message field from JSON to a string."""
-    try:
-        return json.dumps(message)
-    except TypeError as exc:
-        return json.dumps({'errors': [{'message': str(exc)}]})
+        Confusingly, has similar format to GraphQL execution result.
+        But if we change this now we could break compatibility for
+        issuing commands to/receiving responses from workflows running in
+        different versions of Cylc 8.
```
oliver-sanders, line 68: Sorry about that.

The server interface was developed before we added the GraphQL endpoint. Any resemblance in interchange formats is purely coincidental.

However, it wouldn't make sense to change this to match GraphQL (we need to carry GraphQL over this protocol).

MetRonnie:
> Can you help me understand what bug this fixes

#### Before

```console
# Workflow running at Cylc 8.3.6, CLI at Cylc 8.4.0
$ cylc trigger wflow//1/foo
Error processing command:
    AttributeError: 'list' object has no attribute 'values'
```

#### After

```console
# Workflow running at Cylc 8.3.6, CLI at this Cylc <this PR>
$ cylc trigger wflow//1/foo
Error processing command:
    Exception: This command is no longer compatible with the version of Cylc running the workflow. Please stop & restart the workflow with Cylc 8.4.1.dev or higher.

[{'error': {'message': 'Unknown argument "onResume" on field "trigger" of type "Mutations".', 'traceback': ['graphql.error.base.GraphQLError: Unknown argument "onResume" on field "trigger" of type "Mutations".\n']}}]
```

cylc/flow/network/__init__.py
MetRonnie: This `user` field seems to be used by the [`WorkflowRuntimeServer.receiver()`](https://github.com/cylc/cylc-flow/blob/21d18ba0b22c3e729345aa30e539ee3cee1211c7/cylc/flow/network/server.py#L289) method. It is set by the [`WorkflowReplier.listener()`](https://github.com/cylc/cylc-flow/blob/21d18ba0b22c3e729345aa30e539ee3cee1211c7/cylc/flow/network/replier.py#L103) method. In the server main loop it is listening for zmq messages and then assuming that they are from the current user.

https://github.com/cylc/cylc-flow/blob/21d18ba0b22c3e729345aa30e539ee3cee1211c7/cylc/flow/network/replier.py#L100-L103

It says "authenticating the user in the process" but this is always the current user, so I don't know why we would always set the current user in this dict.

Might be one for @dwsutherland?

cylc/flow/network/multi.py
MetRonnie: This function is called once per workflow, and the raised Exception is caught down below and used to print the error message.
```console
$ cylc trigger w1//1/foo w2//20250221T0000Z/foo
w1: Command queued
w2: Error processing command:
        Exception: This command is no longer compatible ...
```

cylc/flow/network/server.py
MetRonnie: By graphql

cylc/flow/network/__init__.py
oliver-sanders: > "encode" implies converting to bytes

Not really, the meaning of "code" is much broader than that.

cylc/flow/network/__init__.py
oliver-sanders: > Might be one for @dwsutherland?

Nah, that one I put in.

The user field should be filled in for requests inbound from the UIS (if Cylc Hub is being used) so we can log the user who requested the action.

If the request didn't come via the UIS, then we log it as originated from the workflow owner (i.e. the person who actioned this command has control over the owner's user account and can act on its behalf).

cylc/flow/network/server.py
MetRonnie: This function returning a string was broken as it lead to
```console
$ cylc trigger wflow//20250221T0000Z/foo
Error processing command:
    AttributeError: 'str' object has no attribute 'values'
```
with nothing in the scheduler log

cylc/flow/network/server.py
oliver-sanders: Right, GraphQL then swallows this error and injects it into its response?

You've probably already tested this, if not make sure it's nailed down (we don't want the potential for a bad request taking down the server).

cylc/flow/network/__init__.py
MetRonnie: I have gone with `stringify` and `parse` inspired by the JavaScript `JSON` module. Open to other suggestions. Otherwise we have stuff like https://github.com/cylc/cylc-flow/blob/035572ffce71e9e35a74b920761ba0c276d947b9/cylc/flow/network/replier.py#L124

cylc/flow/network/server.py
MetRonnie: I think this one is sufficient
https://github.com/cylc/cylc-flow/blob/f751949dfed0005e069ec4052592409fb80d2fb2/tests/integration/network/test_server.py#L138-L146

cylc/flow/network/__init__.py
MetRonnie: Actually from debugging I'm pretty sure this is vestiges of authorisation functionality that was removed/changed: https://github.com/cylc/cylc-flow/pull/3845/files#diff-97d852e5ae5195e580b347c04b37bd759726680ab8e19dfaa079288d05280b45R46. The `user` field is not actually used anymore.

What is used is `auth_user`: https://github.com/cylc/cylc-flow/blob/21d18ba0b22c3e729345aa30e539ee3cee1211c7/cylc/flow/network/resolvers.py#L717-L730

cylc/flow/network/server.py
```diff
@@ -299,7 +298,6 @@ def receiver(self, message) -> 'ResponseDict':
         try:
             method = getattr(self, message['command'])
             args = message['args']
-            args.update({'user': message['user']})
```
MetRonnie, line 285: I have checked inter-version compatibility by running a workflow in this PR's version of Cylc and the UIS running at Cylc 8.4.0, and testing mutations from the UI still work. And vice versa.

cylc/flow/network/__init__.py
oliver-sanders: > What is used is auth_user

Ach, ok.

I think what's happened here is:
* The user field as added at the ZMQ layer to support multi-user interaction (following on the pattern of Cylc 7).
* Cylc 8 implemented authorisation at the UIS layer, the `auth_user` field was added to GraphQL.
* The user field could still have use at the CLI layer.
* However, we decided that a multi-user CLI would go through the UIS anyway (i.e. via GraphQL) so there should be no present or future use case for having this field at the ZMQ level.

MetRonnie: I've force-pushed to leave the `msg['user']` the way it was, and taken out the removal of the unused auth stuff, to be put on an 8.5.0 PR

cylc/flow/network/client.py
```diff
@@ -270,7 +284,7 @@ async def async_request(
         args: Optional[Dict[str, Any]] = None,
         timeout: Optional[float] = None,
         req_meta: Optional[Dict[str, Any]] = None
-    ) -> object:
+    ) -> Union[bytes, object]:
```
oliver-sanders, line 287: (object was used as a catch-all back before mypy)

```python
>>> isinstance(bytes, object)
True
```

```suggestion
    ) -> object:
```

oliver-sanders:
I think this PR is stripping away exception context?

I tried creating an internal error with this diff:

```diff
diff --git a/cylc/flow/network/resolvers.py b/cylc/flow/network/resolvers.py
index 79b91f97e..702b3f573 100644
--- a/cylc/flow/network/resolvers.py
+++ b/cylc/flow/network/resolvers.py
@@ -787,6 +787,7 @@ class Resolvers(BaseResolvers):
         cutoff: Any = None
     ):
         """Put or clear broadcasts."""
+        raise ValueError('Unsupported broadcast mode')
         if settings is not None:
             # Convert schema field names to workflow config setting names if
             # applicable:
```

Results:

```
# 8.4.0
ERROR: [{'error': {'message': 'Unsupported broadcast mode', 'traceback': ['graphql.error.located_error.GraphQLLocatedError: Unsupported broadcast mode\n']}}]

# this branch
ERROR: {'broadcast': None}
```

MetRonnie:
Should be sorted in 1415531820ea8b3bb13ff7c277973de4b3887059

cylc/flow/network/__init__.py
```diff
@@ -30,32 +35,77 @@
...
+                id: <workflow/task ID>,
+                response: [<success_bool>, <message>]
+            },
+            ...
+            ]
+        }
+        }
+        but this is not 100% consistent unfortunately
+        """
+        error: Union[Exception, str, dict]
```
oliver-sanders, line 86: When is `error` an `Exception` or a `str`?

oliver-sanders:
Using a diff along the lines of this:

```diff
diff --git a/cylc/flow/network/resolvers.py b/cylc/flow/network/resolvers.py
index 79b91f97e..702b3f573 100644
--- a/cylc/flow/network/resolvers.py
+++ b/cylc/flow/network/resolvers.py
@@ -787,6 +787,7 @@ class Resolvers(BaseResolvers):
         cutoff: Any = None
     ):
         """Put or clear broadcasts."""
+        raise ValueError('Unsupported broadcast mode')
         if settings is not None:
             # Convert schema field names to workflow config setting names if
             # applicable:
```

We will now get an error something like this:

```
ClientError: {'message': 'Unsupported broadcast mode'}
```

* We are reporting the string `ClientError` for an error that was received from the server?! Doesn't really make sense.
* We are presenting the error as a dictionary.

This patch changes the string `ClientError` to `RequestError` and uses the `message` field from the dictionary rather than the whole dictionary:

```diff
diff --git a/cylc/flow/exceptions.py b/cylc/flow/exceptions.py
index 802cfaaa9..a3eff7388 100644
--- a/cylc/flow/exceptions.py
+++ b/cylc/flow/exceptions.py
@@ -285,6 +285,10 @@ class ClientError(CylcError):
         return ret

+class RequestError(ClientError):
+    """Represents an error returned by the server."""
+
+
 class WorkflowStopped(ClientError):
     """The Cylc scheduler you attempted to connect to is stopped."""
```
 
```diff
diff --git a/cylc/flow/network/client.py b/cylc/flow/network/client.py
index 656aefd5c..89dac4cbb 100644
--- a/cylc/flow/network/client.py
+++ b/cylc/flow/network/client.py
@@ -40,10 +40,10 @@ from cylc.flow import (
     __version__ as CYLC_VERSION,
 )
 from cylc.flow.exceptions import (
-    ClientError,
     ClientTimeout,
     ContactFileExists,
     CylcError,
+    RequestError,
     WorkflowStopped,
 )
 from cylc.flow.hostuserutil import get_fqdn_by_host
@@ -332,17 +332,23 @@ class WorkflowRuntimeClient(  # type: ignore[misc]
             return response['data']
         except KeyError:
             error = response.get('error')
-            if not error:
-                error = (
+            error_mesage: str
+            if isinstance(error, (str, Exception)):
+                error_mesage = str(error)
+            elif error is None:
+                error_mesage = (
                     f"Received invalid response for Cylc {CYLC_VERSION}: "
                     f"{response}"
                 )
                 wflow_cylc_ver = response.get('cylc_version')
                 if wflow_cylc_ver and wflow_cylc_ver != CYLC_VERSION:
-                    error += (
+                    error_mesage += (
                         f"\n(Workflow is running in Cylc {wflow_cylc_ver})"
                     )
-            raise ClientError(str(error)) from None
+            else:
+                error_mesage = error.get('message', 'ERROR')
+
+            raise RequestError(error_mesage) from None
 
     def get_header(self) -> dict:
         """Return "header" data to attach to each request for traceability.
```

cylc/flow/network/__init__.py
MetRonnie: I think I tried to change it to `Exception` or `str` to begin with, to simplify things, but realised it breaks inter-version compatibility. So presently I don't think it ever is an `Exception` or `str`.

PR #6583, opened 2025-01-30
Fix.no validation of outputs from graph singletons

### LINKED ISSUE #6508: lone custom outputs ignored
This validates OK, and `:x` is ignored in the completion expression (same for `a:x?`):
```
[scheduling]
   [[graph]]
        R1 = "a:x"
[runtime]
   [[a]]  # no outputs defined
```

But triggering another task off the output results in a warning:
```
[scheduling]
   [[graph]]
        R1 = "a:x => b"
[runtime]
   [[a]]  # no outputs defined
   [[b]]
```

```
WorkflowConfigError: Undefined custom output: a:x
```

wxtim:
Closes #6508 

**Check List**

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -2289,10 +2289,12 @@ class WorkflowConfig:
 
         Raises: WorkflowConfigError if a custom output is not defined.
         """
+        # TODO (On drop 3.7): Can be simplified with walrus :=
+        # if (b := a[1].strip("?")) not in TASK_QUALIFIERS
         terminal_outputs = [
-            (a[0].strip("!"), b)
+            (a[0].strip("!"), a[1].strip("?"))
             for a in (t.split(':') for t in terminals if ":" in t)
-            if (b := a[1].strip("?")) not in TASK_QUALIFIERS
+            if (a[1].strip("?")) not in TASK_QUALIFIERS
         ]
 
         for task, output in terminal_outputs:
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -1844,6 +1844,7 @@ class WorkflowConfig:
 
         triggers = {}
         xtrig_labels = set()
+
         for left in left_nodes:
             if left.startswith('@'):
                 xtrig_labels.add(left[1:])
@@ -2265,6 +2266,17 @@ class WorkflowConfig:
             self.workflow_polling_tasks.update(
                 parser.workflow_state_polling_tasks)
             self._proc_triggers(parser, seq, task_triggers)
+            self.check_outputs(
+                [
+                    task_output
+                    for task_output in parser.task_output_opt
+                    if task_output[0]
+                    in [
+                        task_output.split(':')[0]
+                        for task_output in parser.terminals
+                    ]
+                ]
+            )
 
         self.set_required_outputs(task_output_opt)
 
@@ -2277,6 +2289,26 @@ class WorkflowConfig:
 
         for tdef in self.taskdefs.values():
             tdef.tweak_outputs()
+
+    def check_outputs(
+        self, tasks_and_outputs: Iterable[Tuple[str, str]]
+    ) -> None:
+        """Check that task outputs have been registered with tasks.
+
+        Args: tasks_and_outputs: ((task, output), ...)
+
+        Raises: WorkflowConfigError is a user has defined a task with a
+        custom output, but has not registered a custom output.
+        """
+        for task, output in tasks_and_outputs:
+            registered_outputs = self.cfg['runtime'][task]['outputs']
+            if (
+                not TaskOutputs.is_valid_std_name(output)
+                and output not in registered_outputs
+            ):
+                raise WorkflowConfigError(
+                    f"Undefined custom output: {task}:{output}"
+                )
 
     def _proc_triggers(self, parser, seq, task_triggers):
         """Define graph edges, taskdefs, and triggers, from graph sections."""
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -2270,7 +2270,7 @@ class WorkflowConfig:
 
             # Checking for undefined outputs for terminal tasks. Tasks with
             # dependencies are checked in generate_triggers:
-            self.check_outputs(parser.terminals)
+            self.check_terminal_outputs(parser.terminals)
 
         self.set_required_outputs(task_output_opt)
 
@@ -2284,12 +2284,16 @@ class WorkflowConfig:
         for tdef in self.taskdefs.values():
             tdef.tweak_outputs()
 
-    def check_outputs(self, terminals: Iterable[str]) -> None:
+    def check_terminal_outputs(self, terminals: Iterable[str]) -> None:
         """Check that task outputs have been registered with tasks.
+
+
+        Where a "terminal output" is an output for a task at the end of a
+        graph string, such as "end" in `start => middle => end`.
 
         Raises: WorkflowConfigError if a custom output is not defined.
         """
-        # TODO (On drop 3.7): Can be simplified with walrus :=
+        # BACK COMPAT: (On drop 3.7): Can be simplified with walrus :=
         # if (b := a[1].strip("?")) not in TASK_QUALIFIERS
         terminal_outputs = [
             (a[0].strip("!"), a[1].strip("?"))
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -99,6 +99,7 @@ from cylc.flow.task_outputs import (
     get_trigger_completion_variable_maps,
     trigger_to_completion_variable,
 )
+from cylc.flow.task_qualifiers import TASK_QUALIFIERS
 from cylc.flow.run_modes import RunMode
 from cylc.flow.task_trigger import TaskTrigger, Dependency
 from cylc.flow.taskdef import TaskDef
@@ -2266,17 +2267,10 @@ class WorkflowConfig:
             self.workflow_polling_tasks.update(
                 parser.workflow_state_polling_tasks)
             self._proc_triggers(parser, seq, task_triggers)
-            self.check_outputs(
-                [
-                    task_output
-                    for task_output in parser.task_output_opt
-                    if task_output[0]
-                    in [
-                        task_output.split(':')[0]
-                        for task_output in parser.terminals
-                    ]
-                ]
-            )
+
+            # Checking for undefined outputs for terminal tasks. Tasks with
+            # dependencies are checked in generate_triggers:
+            self.check_outputs(parser.terminals)
 
         self.set_required_outputs(task_output_opt)
 
@@ -2290,22 +2284,19 @@ class WorkflowConfig:
         for tdef in self.taskdefs.values():
             tdef.tweak_outputs()
 
-    def check_outputs(
-        self, tasks_and_outputs: Iterable[Tuple[str, str]]
-    ) -> None:
+    def check_outputs(self, terminals: Iterable[str]) -> None:
         """Check that task outputs have been registered with tasks.
 
-        Args: tasks_and_outputs: ((task, output), ...)
-
-        Raises: WorkflowConfigError is a user has defined a task with a
-        custom output, but has not registered a custom output.
+        Raises: WorkflowConfigError if a custom output is not defined.
         """
-        for task, output in tasks_and_outputs:
-            registered_outputs = self.cfg['runtime'][task]['outputs']
-            if (
-                not TaskOutputs.is_valid_std_name(output)
-                and output not in registered_outputs
-            ):
+        terminal_outputs = [
+            (a[0].strip("!"), b)
+            for a in (t.split(':') for t in terminals if ":" in t)
+            if (b := a[1].strip("?")) not in TASK_QUALIFIERS
+        ]
+
+        for task, output in terminal_outputs:
+            if output not in self.cfg['runtime'][task]['outputs']:
                 raise WorkflowConfigError(
                     f"Undefined custom output: {task}:{output}"
                 )
--- a/cylc/flow/graph_parser.py
+++ b/cylc/flow/graph_parser.py
@@ -473,9 +473,9 @@ class GraphParser:
         # Get a set of RH nodes which are not at the LH of another pair:
         # terminals = {p[1] for p in pairs}.difference({p[0] for p in pairs})
 
-        check_terminals = {}
-        lefts = set()
-        rights = set()
+        check_terminals: Dict[str, str] = {}
+        lefts: Set[str] = set()
+        rights: Set[str] = set()
 
         for pair in sorted(pairs, key=lambda p: str(p[0])):
             self._proc_dep_pair(pair, check_terminals, lefts, rights)
--- a/cylc/flow/graph_parser.py
+++ b/cylc/flow/graph_parser.py
@@ -489,14 +489,19 @@ class GraphParser:
             {'bar'}
         """
         lefts = []
-        for left, _ in pairs:
+        rights = []
+        for left, right in pairs:
             if left and ('&' in left or '|' in left):
                 # RE used because don't want to have to worry about
                 # mutiple and/or and cleaning whitespace.
                 lefts += GraphParser._RE_ANDOR.split(left)
             else:
                 lefts.append(left)
-        return {p[1] for p in pairs}.difference(set(lefts))
+            if right and ('&' in right or '|' in right):
+                rights += GraphParser._RE_ANDOR.split(right)
+            else:
+                rights.append(right)
+        return set(rights).difference(set(lefts))
 
     @classmethod
     def _report_invalid_lines(cls, lines: List[str]) -> None:
--- a/cylc/flow/graph_parser.py
+++ b/cylc/flow/graph_parser.py
@@ -168,6 +168,7 @@ class GraphParser:
     _RE_OFFSET = r'\[[\w\-\+\^:]+\]'
     _RE_QUAL = QUALIFIER + r'[\w\-]+'  # task or fam trigger
     _RE_OPT = r'\??'  # optional output indicator
+    _RE_ANDOR = re.compile(r'\s*[&|]\s*')
 
     REC_QUAL = re.compile(_RE_QUAL)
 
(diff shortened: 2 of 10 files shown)


hjoliver:
(See new comment on the issue https://github.com/cylc/cylc-flow/issues/6508#issuecomment-2632169858 )

cylc/flow/config.py
```diff
@@ -2265,6 +2266,17 @@ def load_graph(self):
             self.workflow_polling_tasks.update(
                 parser.workflow_state_polling_tasks)
             self._proc_triggers(parser, seq, task_triggers)
+            self.check_outputs(
```
hjoliver, line 2269: Can you add a comment that this is just checking "terminals" and non-terminals are checked elsewhere.

cylc/flow/config.py
```diff
@@ -2265,6 +2266,17 @@ def load_graph(self):
...
+                [
+                    task_output
+                    for task_output in parser.task_output_opt
+                    if task_output[0]
+                    in [
+                        task_output.split(':')[0]
+                        for task_output in parser.terminals
+                    ]
+                ]
```
hjoliver, line 2278: I think this needs to check every terminal output, but it's actually checking every output (both terminal and non-terminal) of any task that has at least one terminal output. So there's some unnecessary checking going on.

hjoliver (changes requested):

cylc/flow/config.py
hjoliver: (Also, that comprehension would probably be clearer if lines 2275-6 used a different name for `task_output`, as it's a different variable from that in lines 2271-3).

wxtim: I'll see if I can work my way around this, but ther reason I need to use every output is that terminals only carries the raw data. If the user sets `foo:fail` this is not the same as `foo:failed`.

cylc/flow/config.py
```diff
@@ -2278,6 +2283,29 @@ def load_graph(self):
...
+        for terminal in terminals:
+            if ':' in terminal:
+                task = re.findall(TaskID.NAME_RE, terminal)[0]
+                qualifier = re.findall(GraphParser._RE_QUAL, terminal)
+                if qualifier:
+                    qualifier_str = qualifier[0].strip(':')
+                if qualifier_str and qualifier_str not in TASK_QUALIFIERS:
+                    terminal_outputs.append((task, qualifier_str))
+
+        for task, output in terminal_outputs:
```
hjoliver, line 2302: Can't we just split on the `:` here?
```suggestion
        for task, output in [t.split(":") for t in terminals if ":" in t]:
```

hjoliver (dismissed):

cylc/flow/config.py
```diff
@@ -2278,6 +2283,29 @@ def load_graph(self):
...
+            registered_outputs = self.cfg['runtime'][task]['outputs']
+            if output not in registered_outputs:
```
hjoliver, line 2304:
```suggestion
            if output not in  self.cfg['runtime'][task]['outputs']:
```

cylc/flow/config.py
wxtim: Not if there is a suicide trigger, or an optional output or the user has used a present tense qualifier.

hjoliver: Hmm, OK, I'm not sure we should allow outputs to be declared by this sort of construct: `a => ! b:x`, but it seems we do!

Otherwise, that code block strikes me as kinda heavy-weight for what it does, but maybe it doesn't matter.

Still,
- why did you use `re.findall(...)[0]` instead of `re.match()` ?
- `strip(':')` is unnecessary (the `:` is guaranteed so just slice off the final character)?

hjoliver: How about like this?
```python
2303         for task, qual in [t.split(":") for t in terminals if ":" in t]:
2304             if qual in TASK_QUALIFIERS:
2305                 continue
2306             terminal_outputs.append([task.strip("!"), qual.strip("?")])
```

wxtim: > Why use re.findall(...)[0] instead of just re.match() ?

Familiarity. Match works. :)

> I think strip(':') is unnecessary there - just slice off the final character, which is guaranteed to be a :?

First char, surely?

> How about this?

If fails a pile of tests, mostly because terminals have not in fact been well parsed (my bad) - main problem is that fail/failed are not both recognised as standard outputs.

```
tests/integration/validate/test_outputs.py::test_completion_expression_invalid[expire-optional-in-graph-but-not-used-in-completion]
tests/integration/validate/test_outputs.py::test_completion_expression_invalid[alt-compvar2]
tests/integration/validate/test_outputs.py::test_completion_expression_invalid[submit-failed used in completion expression]
tests/integration/test_optional_outputs.py::test_task_completion[2]
tests/integration/reftests/test_pre_initial.py::test_over_bracketed
tests/integration/test_optional_outputs.py::test_implicit_completion_expression[a]
tests/integration/test_optional_outputs.py::test_implicit_completion_expression[b]
tests/integration/test_optional_outputs.py::test_implicit_completion_expression[c]
tests/integration/test_optional_outputs.py::test_implicit_completion_expression[d]
tests/integration/test_optional_outputs.py::test_implicit_completion_expression[e]
tests/integration/test_optional_outputs.py::test_implicit_completion_expression[f]
tests/integration/test_optional_outputs.py::test_implicit_completion_expression[h0]
tests/integration/test_optional_outputs.py::test_implicit_completion_expression[h1]
tests/integration/test_optional_outputs.py::test_implicit_completion_expression[i]
tests/integration/test_optional_outputs.py::test_implicit_completion_expression[j]
tests/integration/test_optional_outputs.py::test_implicit_completion_expression[k]
tests/integration/test_optional_outputs.py::test_implicit_completion_expression[l]
tests/integration/test_dbstatecheck.py::test_basic
tests/integration/test_dbstatecheck.py::test_task
tests/integration/test_dbstatecheck.py::test_point
tests/integration/test_dbstatecheck.py::test_status
tests/integration/test_dbstatecheck.py::test_output
tests/integration/test_dbstatecheck.py::test_flownum
```

hjoliver: >>   I think strip(':') is unnecessary there - just slice off the final character, which is guaranteed to be a :?
>  First char, surely? And isn't strip(':') more meaningful to a future reader than some mysterious slice?

Sorry, yes, first char!

`strip(':')` presumably does some unnecessary computation if the first char is *always* a `:`.  I'd just convey the meaning like this:

`foo[1:] # strip leading ":"`

hjoliver: > If fails a pile of tests, ... main problem is that fail/failed are not both recognised as standard outputs.

Actually `TASK_QUALIFIERS` does include the "present tense" standard outputs.

It's just that my untested attempt should have stripped unwanted chars before comparing with standard outputs.

This passes the tests above, and IMO is computationally simpler and easier to understand:
```python
for left, right in [t.split(":") for t in terminals if ":" in t]:
    taskname = left.strip("!")  # ignore any suicide trigger
    qualifier = right.strip("?")  # ignore any optional output markers
    if qualifier in TASK_QUALIFIERS:  # ignore any standard outputs
        continue
    terminal_outputs.append([taskname, qualifier])
```
(Maybe also replace `?` with `GraphParser.OPTIONAL` etc.)

hjoliver: Even better (tested)! 
```python
2292         terminal_outputs = [
2293             (a[0].strip("!"), b)
2294             for a in (t.split(':') for t in terminals if ":" in t)
2295             if (b := a[1].strip("?")) if b not in TASK_QUALIFIERS
2296         ]
```

cylc/flow/config.py
```diff
@@ -2278,6 +2284,28 @@ def load_graph(self):
...
+        terminal_outputs = []
+        for terminal in terminals:
+            if ':' in terminal:
+                task = re.search(TaskID.NAME_RE, terminal)
+                task = task.group() if task else ''
+                qualifier = re.search(GraphParser._RE_QUAL, terminal)
+                if qualifier:
+                    qualifier_str = qualifier[0].strip(':')
+                if qualifier_str and qualifier_str not in TASK_QUALIFIERS:
+                    terminal_outputs.append((task, qualifier_str))
```
wxtim, line 2301:
```suggestion
        terminal_outputs = [
            (a[0].strip("!"), b)
            for a in (t.split(':') for t in terminals if ":" in t)
            if (b := a[1].strip("?")) not in TASK_QUALIFIERS
        ]
```

cylc/flow/config.py
wxtim: Walrus invalid in Python 3.7.

hjoliver: Aren't we Python 3.8+ for cylc-flow now? (Oh, the UIS?)

oliver-sanders: Yep, cylc-flow, cylc-rose & metomi-rose still 3.7 for a while. Will bump to 3.8 (or likely 3.9) when that requirement relaxes, so no walruses just yet ( :seal: * :unicorn: ).

I am loving the commit message "remove stray walrus" though!

If you're worried about efficiency, swap `[` for `(` to turn the list comp into a generator.

cylc/flow/config.py
```diff
@@ -2278,6 +2284,25 @@ def load_graph(self):
         for tdef in self.taskdefs.values():
             tdef.tweak_outputs()
 
+    def check_outputs(self, terminals: Iterable[str]) -> None:
```
oliver-sanders, line 2287: If this is only checking "terminals", maybe call the method `check_terminal_outputs`.

Would be good to define "terminal output" in the docstring as it's a very niche term.

cylc/flow/task_outputs.py
```diff
@@ -601,7 +601,7 @@ def color_wrap(string, is_complete):
     @staticmethod
     def is_valid_std_name(name: str) -> bool:
         """Check name is a valid standard output name."""
-        return name in SORT_ORDERS
+        return name in TASK_OUTPUTS
```
oliver-sanders, line 604: Slightly surprised by this change.

We were able to use `:finished` outputs before, why is the change only needed now?

tests/unit/test_config.py
```diff
@@ -1713,3 +1719,31 @@ def test_force_workflow_compat_mode(tmp_path):
...
+            [[graph]]
+                R1 = foo
+    """))
+    cfg = WorkflowConfig('', tmp_path / 'flow.cylc', '')
+    cfg.cfg['runtime']['foo']['outputs'] = registered_outputs
+    if fails:
+        with pytest.raises(WorkflowConfigError, match='Undefined custom output'):
+            cfg.check_outputs(tasks_and_outputs)
+    else:
+        assert cfg.check_outputs(tasks_and_outputs) is None
```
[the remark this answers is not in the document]
wxtim:
```suggestion
        assert cfg.check_outputs(tasks_and_outputs) is None

```

cylc/flow/config.py
wxtim:
```suggestion
    def check_terminal_outputs(self, terminals: Iterable[str]) -> None:
```

cylc/flow/config.py
```diff
@@ -2278,6 +2284,25 @@ def load_graph(self):
...
+        """Check that task outputs have been registered with tasks.
+
+        Raises: WorkflowConfigError if a custom output is not defined.
```
wxtim, line 2290:
```suggestion
        Where a "terminal output" is an output for a task at the end of a
        graph string, such as bar in `foo => bar`
        Raises: WorkflowConfigError if a custom output is not defined.
```

cylc/flow/config.py
```diff
@@ -2278,6 +2284,25 @@ def load_graph(self):
...
+        """
+        # TODO (On drop 3.7): Can be simplified with walrus :=
```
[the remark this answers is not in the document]
wxtim:
```suggestion
        # BACK COMPAT: (On drop 3.7): Can be simplified with walrus :=
```

cylc/flow/task_outputs.py
wxtim: Terminals doesn't do this processing anywhere else, other use cases for this do.

wxtim:
```suggestion
        return name in SORT_ORDERS
```

cylc/flow/config.py
hjoliver: > Yep, cylc-flow, cylc-rose & metomi-rose still 3.7 for a while.

cylc-uiserver, ...

But current docs say this 🫨 :

> Via Conda (recommended
> Supported Python versions
    cylc-flow supports Python 3.8+.
    cylc-uiserver currently only supports Python 3.8-3.9.

oliver-sanders: > But current docs say this

That is correct. We can only support Python 3.8+ on conda-forge, but we still support 3.7 via pypi.

cylc/flow/graph_parser.py
```diff
@@ -470,10 +471,37 @@ def parse_graph(self, graph_string: str) -> None:
...
+            if left and ('&' in left or '|' in left):
+                # RE used because don't want to have to worry about
+                # mutiple and/or and cleaning whitespace.
+                lefts += GraphParser._RE_ANDOR.split(left)
+            else:
+                lefts.append(left)
+            if right and ('&' in right or '|' in right):
+                rights += GraphParser._RE_ANDOR.split(right)
+            else:
+                rights.append(right)
```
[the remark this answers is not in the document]
wxtim: Probably. I'll have a look.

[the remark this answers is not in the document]
oliver-sanders: `get_graph_terminals` is iterating over pairs:

```python
def get_graph_terminals(...):
    ....
    for left, right in pairs:
```

And `_proc_dep_pair` is iterat'ed over pairs:

```python
        for pair in sorted(pairs, key=lambda p: str(p[0])):
            self._proc_dep_pair(pair, self.terminals)
```

So I think the two are equivalent?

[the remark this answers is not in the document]
wxtim: Yes, but we _need_ to do the iteration twice, or store lots of extra info in memory (all lefts which could make terminals not actually terminal), or change what is allowable in our syntax (i.e. never allow `[]` in a right).

We need to have iterated over all the pairs to work out which are terminal.
To take my previous example, when looking at pair `food => barley[42]` for the first time `barley[42]` looks illegal. 

My solution is to have iterated through all lefts and rights before to work out that `barley[42]` is on the left of another pair.

If you were to do it in one iteration, you'd have to store `foo`, `bar`, `food` _and_ `barley[42]` as well as possible terminals and do a set comparison afterwards, which could also get expensive.

Alternatively you can just say that `food => barley[42]` is illegal, but that's a change of behaviour.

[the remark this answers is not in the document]
oliver-sanders: Two problems here:

1. Duplication of graph parsing logic (e.g. splitting on and / or symbols).
2. Duplication of iteration over graph pairs.

Here's a really rough and ready diff to show how we could could combine these two loops to avoid the duplications:@

```diff
diff --git a/cylc/flow/graph_parser.py b/cylc/flow/graph_parser.py
index 2b072d659..21a0ad29b 100644
--- a/cylc/flow/graph_parser.py
+++ b/cylc/flow/graph_parser.py
@@ -470,38 +470,22 @@ class GraphParser:
             for i in range(0, len(chain) - 1):
                 pairs.add((chain[i], chain[i + 1]))
 
-        # Get a set of RH nodes which are not at the LH of another pair:
-        self.terminals = self.get_graph_terminals(pairs)
-
+        check_terminals = {}
+        lefts = set()
+        rights = set()
         for pair in sorted(pairs, key=lambda p: str(p[0])):
-            self._proc_dep_pair(pair, self.terminals)
-
-    @staticmethod
-    def get_graph_terminals(pairs):
-        """Get terminating ends of graphs.
-
-        For example in `foo => bar => baz` only `baz` terminates the chain
-        of dependencies.
+            self._proc_dep_pair(pair, check_terminals, lefts, rights)
 
-        Examples:
-            >>> this = GraphParser.get_graph_terminals
-            >>> this({('foo', 'bar')})
-            {'bar'}
-        """
-        lefts = []
-        rights = []
-        for left, right in pairs:
-            if left and ('&' in left or '|' in left):
-                # RE used because don't want to have to worry about
-                # mutiple and/or and cleaning whitespace.
-                lefts += GraphParser._RE_ANDOR.split(left)
-            else:
-                lefts.append(left)
-            if right and ('&' in right or '|' in right):
-                rights += GraphParser._RE_ANDOR.split(right)
-            else:
-                rights.append(right)
-        return set(rights).difference(set(lefts))
+        terminals = rights.difference(lefts)
+        self.terminals = terminals
+        for right in terminals:
+            left = check_terminals.get(right)
+            if left:
+                raise GraphParseError(
+                    'Invalid cycle point offsets only on right hand '
+                    'side of a dependency (must be on left hand side):'
+                    f' {left} => {right}'
+                )
 
     @classmethod
     def _report_invalid_lines(cls, lines: List[str]) -> None:
@@ -532,7 +516,9 @@ class GraphParser:
     def _proc_dep_pair(
         self,
         pair: Tuple[Optional[str], str],
-        terminals: Set[str],
+        check_terminals: Dict[str, str],
+        _lefts: Set[str],
+        _rights: Set[str],
     ) -> None:
         """Process a single dependency pair 'left => right'.
 
@@ -568,12 +554,8 @@ class GraphParser:
             raise GraphParseError(mismatch_msg.format(right))
 
         # Raise error for cycle point offsets at the end of chains
-        if '[' in right and left and (right in terminals):
-            # This right hand side is at the end of a chain:
-            raise GraphParseError(
-                'Invalid cycle point offsets only on right hand '
-                'side of a dependency (must be on left hand side):'
-                f' {left} => {right}')
+        if '[' in right and left:
+            check_terminals[right] = left
 
         # Split right side on AND.
         rights = right.split(self.__class__.OP_AND)
@@ -594,12 +576,15 @@ class GraphParser:
             raise GraphParseError(
                 f"Null task name in graph: {left} => {right}")
 
+        _rights.update(*(rights or []))
+
         for left in lefts:
             # Extract information about all nodes on the left.
 
             if left:
                 info = self.__class__.REC_NODES.findall(left)
                 expr = left
+                _lefts.add(left)
 
             else:
                 # There is no left-hand-side task.

```

We would need to store any "rights" that have inter-cycle offsets.

We could probably get smart with this and housekeep them within the loop.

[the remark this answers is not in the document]
wxtim: But that then makes the list of terminals useless for this PR>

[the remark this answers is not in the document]
wxtim: Looks like it works, and is a good idea, I'm just cleaning up the testrs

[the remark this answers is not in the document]
oliver-sanders: Cool, I'll resolve this conversation now.

I have run this through `cylc validate --profile` on the complex workflow as an efficiency test to see how this approach impacts performance.

The difference in time taken by the graph_parse routine is within the jitter, no noticeable impact :+1:, we good!

cylc/flow/config.py
oliver-sanders: I'm gonna resolve this convo, with or without tusked mammals!

oliver-sanders (approved):
:+1:

Plz squash merge.

hjoliver (approved):
LGTM!

PR #6589, opened 2025-02-04
family add and prune fix

dwsutherland:
closes https://github.com/cylc/cylc-ui/issues/1999

**Check List**

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -1920,9 +1920,10 @@ class DataStoreMgr:
 
         """
         fp_data = self.data[self.workflow_id][FAMILY_PROXIES]
+        fp_added = self.added[FAMILY_PROXIES]
         fp_updated = self.updated[FAMILY_PROXIES]
-        if fp_id in fp_data:
-            fam_node = fp_data[fp_id]
+        fam_node = fp_data.get(fp_id, fp_added.get(fp_id, None))
+        if fam_node is not None:
             # Gather child families, then check/update recursively
             for child_id in fam_node.child_families:
                 if child_id in checked_ids:
@@ -1937,11 +1938,14 @@ class DataStoreMgr:
                     child_tasks.update(fp_updated[fp_id].child_tasks)
                 if fp_updated[fp_id].child_families:
                     child_families.update(fp_updated[fp_id].child_families)
-            # if any child tasks or families are in window, don't prune.
+            # If any child tasks or families are in window,
+            # then don't prune this family.
             if (
                 child_tasks.difference(node_ids)
                 or child_families.difference(prune_ids)
             ):
+                # If any child tasks or families will be pruned,
+                # then update family states.
                 if (
                     child_tasks.intersection(node_ids)
                     or child_families.intersection(prune_ids)


tests/integration/test_data_store_mgr.py
```diff
@@ -131,6 +131,37 @@ async def harness(mod_flow, mod_scheduler, mod_start):
         yield schd, data
 
 
+@pytest.fixture(scope='module')
+async def harness_prune(mod_flow, mod_scheduler, mod_start):
```
wxtim, line 135: Please can we have  at least one line descriptive docstring?

tests/integration/test_data_store_mgr.py
```diff
@@ -313,12 +344,93 @@ async def test_update_data_structure(harness):
...
-    # TODO: fixme
-    # https://github.com/cylc/cylc-flow/issues/4175#issuecomment-1025666413
-    # assert len({t.is_held for t in data[TASK_PROXIES].values()}) == 1
 
 
+async def test_prune_data_store(harness_prune):
+    """Test prune_data_store. This method will expand and reduce the data-store
+    to invoke pruning."""
+    schd, data = harness_prune
+    w_id = schd.data_store_mgr.workflow_id
```
wxtim, line 353: Used in  2/3 tests, suggest that you move it to `harness_prune`:
```
w_id, schd, data = harness_prune
# or
_, schd, data
```

tests/integration/test_data_store_mgr.py
```diff
@@ -313,12 +344,93 @@ async def test_update_data_structure(harness):
...
+        node_ids,
+        parent_ids,
+        checked_ids,
+        schd.data_store_mgr.family_pruned_ids
+    )
+    assert len(checked_ids) == 1
+    assert len(parent_ids) == 0
+
+
+async def test_update_data_structure(harness):
```
wxtim, line 415: Change the name of this to avoid duplicate test names (see line 329)

tests/integration/test_data_store_mgr.py
```diff
@@ -313,12 +344,93 @@ async def test_update_data_structure(harness):
...
 
 
+async def test_prune_data_store(harness_prune):
+    """Test prune_data_store. This method will expand and reduce the data-store
+    to invoke pruning."""
+    schd, data = harness_prune
+    w_id = schd.data_store_mgr.workflow_id
+    schd.data_store_mgr.data[w_id] = data
+    schd.pool.hold_tasks(['*'])
+    await schd.update_data_structure()
```
wxtim, line 356: Could you move this block to harness_prune?

tests/integration/test_data_store_mgr.py
```diff
@@ -313,12 +344,93 @@ async def test_update_data_structure(harness):
...
+    schd.data_store_mgr.set_graph_window_extent(0)
+    schd.data_store_mgr.update_data_structure()
+    assert len({
+        t.is_held
+        for t in data[TASK_PROXIES].values() 
+        if t.graph_depth == 0
+    }) == 1
+
+    # Test rapid addition and removal
+    assert len({
```
wxtim, line 368: These tests might benefit from a quick describe IMO
```suggestion
    # There are no BAR family proxies in .added
    assert len({
```

tests/integration/test_data_store_mgr.py
```diff
@@ -313,12 +344,93 @@ async def test_update_data_structure(harness):
...
+        if f.name == 'BAR'
+    }) == 0
+    assert len({
+        f.name
+        for f in data[FAMILY_PROXIES].values()
+        if f.name == 'BAR'
+    }) == 0
+
+
+async def test_family_ascent_point_prune(harness_prune):
```
wxtim, line 396: Note that this doesn't fail on 8.4.x - were you expecting it to?

tests/integration/test_data_store_mgr.py
```diff
@@ -313,12 +344,93 @@ async def test_update_data_structure(harness):
...
     # state totals changed
     assert TASK_STATUS_FAILED in data[WORKFLOW].state_totals
-    # Shows pruning worked
-    # TODO: fixme
-    # https://github.com/cylc/cylc-flow/issues/4175#issuecomment-1025666413
-    # assert len({t.is_held for t in data[TASK_PROXIES].values()}) == 1
 
 
+async def test_prune_data_store(harness_prune):
+    """Test prune_data_store. This method will expand and reduce the data-store
```
wxtim, line 350: Suggest adding GH issue url

tests/integration/test_data_store_mgr.py
```diff
@@ -313,12 +344,93 @@ async def test_update_data_structure(harness):
...
+    to invoke pruning."""
```
wxtim, line 351:
```suggestion
    to invoke pruning.

    https://github.com/cylc/cylc-ui/issues/1999
    """
```

tests/integration/test_data_store_mgr.py
```diff
@@ -313,12 +344,93 @@ async def test_update_data_structure(harness):
...
+    schd, data = harness_prune
+    w_id = schd.data_store_mgr.workflow_id
+    schd.data_store_mgr.data[w_id] = data
+    schd.pool.hold_tasks(['*'])
```
MetRonnie, line 355: This test is mutating the task pool so it should not be using the module scoped `harness_prune` fixture (or make `harness_prune` function scoped)

dwsutherland:
> I'd really like a reproducible workflow example.
> 

https://github.com/cylc/cylc-ui/issues/1999#issuecomment-2661323292

> ## Must Have
> * Duplicate test `test_update_data_structure`.
> * `test_update_data_structure` failing.
> 

Irrelevant for this PR, I didn't change this test (only removed a comment where pruning wasn't covered)

> ## Nice to have
> * More comments
> 

Will do...

> ## Take on board or ignore
> (as your time and energy permit)
> 
> * Wider refactoring of `test_data_store_mgr.py`

Outside the scope of this small bug fix..

tests/integration/test_data_store_mgr.py
dwsutherland: I've removed this extra module scoped `harness_prune`.. 
Changed the pool modifying prune test to have it's own scheduler to toy with, and shifted the non-pool modifying test to the existing `harness`.

tests/integration/test_data_store_mgr.py
dwsutherland: Have elaborated a little further, noting that the test covers the rapid addition/removal of a task (and it's families) with reference to the issue..

tests/integration/test_data_store_mgr.py
dwsutherland: There is a [safety catch](https://github.com/dwsutherland/cylc-flow/blob/90f69e9615c765690321e110626afa1c335448a4/cylc/flow/data_store_mgr.py#L1929) in the family pruning method to make sure it doesn't trip up if the family proxy is not found in the added or data store..
Theoretically, this should never happen and out tests didn't trigger it.. But in practice I don't want to exclude some exotic situations..

This coverage required I test it, so I added this one to do just that..

tests/integration/test_data_store_mgr.py
dwsutherland: This "harness" has been removed, and included in the prune test scope (as per Ronnie's comment)..

tests/integration/test_data_store_mgr.py
dwsutherland: That was a copy/paste error (was using that test as skeleton for pruning test) .. removed duplicate.

tests/integration/test_data_store_mgr.py
dwsutherland: Again moved `harness_prune` into test function scope.

tests/integration/test_data_store_mgr.py
dwsutherland: Have added more commentary ..

dwsutherland:
Thanks for the review Tim

> I'd really like a reproducible workflow example.
> 

Here's another example in the ticket..
https://github.com/cylc/cylc-ui/issues/1999#issuecomment-2661323292

I'm not sure the criteria for the families being visible in the UI, but graphql will show family proxies that accumulate every cycle.. (due to how suicide triggers add and remove items)


> ## Must Have
> * Duplicate test `test_update_data_structure`.
> * `test_update_data_structure` failing.
> 

Yeah, duplicate removed.

> ## Nice to have
> * More comments
> 

More comments added

> ## Take on board or ignore
> (as your time and energy permit)
> 
> * Wider refactoring of `test_data_store_mgr.py`

I've refactored the code around the tests I added.. Will leave a wider refactor to a bigger change (which there may be in the near future)



**Note:**
If I undo my fix, the tests fail as expected.

wxtim:
@dwsutherland - Unfortunately I couldn't replicate the bug with your example. 😢

dwsutherland:
> @dwsutherland - Unfortunately I couldn't replicate the bug with your example. 😢

Did you use graphiql to try view the accumulation of family proxies?
```
query {
  workflows(ids:["<id>"]) {
    familyProxies {
      id
    }
  }
}
```
with that example workflow, it was 100% reproducible ... (even though the UI doesn't show them)

dwsutherland:
Note `poll_exist` has a hard coded workflow.. but should just be set the itself (`strigger/run1` in this case):
```
[scheduling]
    initial cycle point = 20250204T2240Z
    sequential xtriggers = True
    [[xtriggers]]
        poll_exist = workflow_state(strigger/run1//%(point)s/spawn)
        poll_non_exist = workflow_state(this/workflow//%(point)s/doesnt_exist)
    [[graph]]
        PT1M = """
@wall_clock => spawn
spawn => foo & purge
@poll_non_exist => foo => fiz => fuz
foo => !purge
@poll_exist => purge
purge => !FOOS & !FIZS
"""
[runtime]
    [[root]]
        script = sleep $((1 + $RANDOM % 10))
    [[PURGEFOO]]
    [[spawn]]
        inherit = PURGEFOO
    [[FIZS]]
        inherit = PURGEFOO
    [[fiz]]
        inherit = FIZS
    [[FOOS]]
        inherit = PURGEFOO
    [[foo,fuz]]
        inherit = FOOS
    [[PURGES]]
        inherit = PURGEFOO
    [[purge]]
        inherit = PURGES
```

wxtim:
> > @dwsutherland - Unfortunately I couldn't replicate the bug with your example. 😢
> 
> Did you use graphiql to try view the accumulation of family proxies?
> 
> ```
> query {
>   workflows(ids:["<id>"]) {
>     familyProxies {
>       id
>     }
>   }
> }
> ```
> 
> with that example workflow, it was 100% reproducible ... (even though the UI doesn't show them)

No. I was looking for the bug as reported. Why do the proxies accumulate but not show up in the GUI the same way.

(Yes, I saw the hardcoded inter-workflow xtrigger)

MetRonnie:
I cannot reproduce the GraphiQL family proxies accumulating. I tried both the query and the same thing as a subscription.

Also, I'm not sure if this would be the same bug as https://github.com/cylc/cylc-ui/issues/1999, which seems to be an accumulation of jobs from Tom's GraphiQL output

oliver-sanders:
I tried to replicate using the above example and this diff:

```diff
diff --git a/flow.cylc b/flow.cylc
index 40b932b..2c80133 100644
--- a/flow.cylc
+++ b/flow.cylc
@@ -2,8 +2,7 @@
     initial cycle point = 20250204T2240Z
     sequential xtriggers = True
     [[xtriggers]]
-        # poll_exist = workflow_state(strigger/run1//%(point)s/spawn)
-        poll_exist = workflow_state(tmp.K05ldAjsu1//%(point)s/spawn)
+        poll_exist = workflow_state(%(workflow)s//%(point)s/spawn)
         poll_non_exist = workflow_state(this/workflow//%(point)s/doesnt_exist)
     [[graph]]
         PT1M = """
```

Unfortunately, I couldn't replicate families piling up in GraphiQL OR the empty cycles in the GUI as originally reported.

oliver-sanders:
Have any of us managed to reproduce the "empty cycles in the GUI" issue so far?

dwsutherland:
> Have any of us managed to reproduce the "empty cycles in the GUI" issue so far?

I'm able to easily reproduce the accumulation of families, and the tests show this is fixed (if you undo the fix, but keep the tests, the tests fail).. 

I'm not sure what mechanism produced a stuck UI representation, however, I'm certain that the non-pruning of families addressed here is to blame.. (along with some combination of w/e state/criteria the UI requires to display a node)

dwsutherland:
> 
> Also, I'm not sure if this would be the same bug as [cylc/cylc-ui#1999](https://github.com/cylc/cylc-ui/issues/1999), which seems to be an accumulation of jobs from Tom's GraphiQL output

@MetRonnie  - the main issue (in the title) is the accumulation of old cycles, which is addressed here.. 

The job accumulation is (I assume) a separate issue, where he may be suiciding running (or soon to be) jobs.. It's a Separate issue from [cylc/cylc-ui#1999](https://github.com/cylc/cylc-ui/issues/1999)   ... He should create a new issue for job accumulation..

MetRonnie:
I don't think the job accumulation is a separate issue to old cycles being shown in the UI. It seems the old cycles are showing as a result of the old jobs being retained in the data store. Annoyingly I have not been able to reproduce that issue either, using the `tc_in_domain:noTC` example he gave

MetRonnie (approved):
Having worked on #6656, I think this makes sense, even if none of us were able to reproduce it outside of the added integration test

tests/integration/test_data_store_mgr.py
```diff
@@ -313,10 +313,125 @@ async def test_update_data_structure(harness):
...
+        }) == 1
+        assert len({
+            f.name
+            for f in schd.data_store_mgr.added[FAMILY_PROXIES].values()
+            if f.name == 'BAR'
+        }) == 1
+        # Before updating the data-store, remove bar/BAR.
+        for itask in schd.pool.get_tasks():
+            if itask.tdef.name == 'bar':
+                schd.pool.remove(itask, 'Test removal')
```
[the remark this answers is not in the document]
dwsutherland: Bah, looks like I should have used `1/bar` not `bar` in the `_get_task_by_id` .. will fix up later

dwsutherland:
> Having worked on #6656, I think this makes sense, even if none of us were able to reproduce it outside of the added integration test

Yeah, still a little surprised about this .. Maybe you didn't set `n=0`? I was only able doing that:
https://github.com/cylc/cylc-ui/issues/1999#issuecomment-2632777859

Because, the corresponding task/family wouldn't be pruned in the same batch as it was added otherwise..

But thanks all the same..

MetRonnie (approved):

dwsutherland:
Reproduced even after #6656 is in:
image

dwsutherland:
Found a more simple example to reproduce:
```
[scheduler]
    allow implicit tasks = True

[scheduling]
    cycling mode = integer
    [[graph]]

        P1 = """
foo[-P1] => foo
foo[-P1] => bar => baz
foo => !BAZ
"""

[runtime]
    [[root]]
        [[[simulation]]]
            default run length = PT0S
    [[BAZ]]
    [[baz]]
        inherit = BAZ
```
image

(remember the `n=0`)

hjoliver:
I can confirm the accumulation of family proxies using @dwsutherland's example above.

Initially, with `N=1`, no accumulation occurs. Switch between `N=1,2,3`, still no accumulation.

But as soon as I set `N=0` they start accumulating indefinitely, and after that going back to `N=1,2,3` does not fix the problem.

wxtim (approved):

PR #6590, opened 2025-02-05
Fix evaluation of conditional expressions at negative cycle points

### LINKED ISSUE #6588: `TriggerExpressionError` with inter-cycle dependency and OR operator
### Reproducible Example
```cylc
[scheduler]
    allow implicit tasks = True
[scheduling]
    cycling mode = integer
    [[graph]]
        P1 = """
            a & x
            x[-P2] | a => b
        """
```

Image

```
$ cylc play workflow -N
...
ERROR - Workflow shutting down - TriggerExpressionError: "-1/x succeeded|1/a succeeded":
    invalid syntax (<string>, line 1)
ERROR - No platforms available to remote tidy install targets:
     * localhost
```

hjoliver:
Integer offsets `[-P2]` or bigger aren't getting replaced with Python code in prerequisite conditional expressions because they generate negative cycle points when evaluated at initial cycle point 1. (There's no problem with `[-P1]` - that generates a cycle point of 0).

Close #6588 

**Check List**

### DIFF
--- a/cylc/flow/prerequisite.py
+++ b/cylc/flow/prerequisite.py
@@ -198,16 +198,15 @@ class Prerequisite:
 
         Examples:
             # GH #3644 construct conditional expression when one task name
-            # is a substring of another: foo | xfoo => bar.
-            # Add 'foo' to the 'satisfied' dict before 'xfoo'.
+            # is a substring of another: 11/foo | 1/foo => bar.
             >>> preq = Prerequisite(1)
             >>> preq[(1, 'foo', 'succeeded')] = False
-            >>> preq[(1, 'xfoo', 'succeeded')] = False
-            >>> preq.set_conditional_expr("1/foo succeeded|1/xfoo succeeded")
+            >>> preq[(11, 'foo', 'succeeded')] = False
+            >>> preq.set_conditional_expr("11/foo succeeded|1/foo succeeded")
             >>> expr = preq.conditional_expression
             >>> expr.split('|')  # doctest: +NORMALIZE_WHITESPACE
-            ['bool(self._satisfied[("1", "foo", "succeeded")])',
-            'bool(self._satisfied[("1", "xfoo", "succeeded")])']
+            ['bool(self._satisfied[("11", "foo", "succeeded")])',
+            'bool(self._satisfied[("1", "foo", "succeeded")])']
 
             # GH #6588 integer offset "x[-P2] | a" gives a negative cycle point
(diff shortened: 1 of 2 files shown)


cylc/flow/prerequisite.py
```diff
@@ -211,7 +221,7 @@ def set_conditional_expr(self, expr):
                 # Use '\b' in case one task name is a substring of another
                 # and escape special chars ('.', timezone '+') in task IDs.
                 expr = re.sub(
-                    fr"\b{re.escape(self.MESSAGE_TEMPLATE % t_output)}\b",
+                    fr"{re.escape(self.MESSAGE_TEMPLATE % t_output)}\b",
```
hjoliver, line 224: TODO: consider the preceding comment - might need a more sophisticated fix?

hjoliver: (Tests all pass, but maybe the commented issue only tested the other word boundary for task name sub-strings)

MetRonnie: I've pushed this fix and it reveals that your fix breaks this case

hjoliver: I've fixed it, see below. The problem was for negative cycle points, the word boundary needs to be (e.g.) `-/b1x` not `\b-1/x`.

cylc/flow/prerequisite.py
```diff
@@ -198,26 +198,41 @@ def set_conditional_expr(self, expr):
...
             # Make a Python expression so we can eval() the logic.
             for t_output in self._satisfied:
                 # Use '\b' in case one task name is a substring of another
                 # and escape special chars ('.', timezone '+') in task IDs.
+                msg = self.MESSAGE_TEMPLATE % t_output
+                if msg[0] == '-':
+                    # -ve cycles: \b needs to be to the right of the `-` char.
+                    pattern = fr"-\b{re.escape(msg[1:])}\b"
+                else:
+                    pattern = fr"\b{re.escape(msg)}\b"
```
hjoliver, line 233: Not sure if there's a more elegant way of handling this.

MetRonnie:
Kicking tests

hjoliver:
Codecov patch seems to be confused, it says both 🤯 :

codecov/patch — 50.00% of diff hit (target 97.00%)
codecov/patch — 100.00% of diff hit (target 97.00%)

(Definitely 100% covered though).

MetRonnie (approved):

wxtim (approved):

PR #4970, opened 2022-07-06
Back-compat: fix self-suicide: a:fail => !a

hjoliver:
These changes close #4968

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` and `conda-environment.yml`.
- [x] Appropriate tests are included (unit and/or functional).
- [x] Appropriate change log entry included.
- [x] No documentation update required.

PR #6602, opened 2025-02-10
Suicide trigger prevents later cycles starting

### LINKED ISSUE #6594: Suicide trigger prevents cycles starting
### Description

User reported this case. It's not a blocker (for _that_ user), because the user was over-using suicide triggers and we were able to remove the offending graph section. Nonetheless, this is a bug.

### Reproducible Example

The following workflow will shut down as completed after all 1990's tasks are complete:

```cylc
[scheduler]
    allow implicit tasks = True
    cycle point format = %Y

[scheduling]
    initial cycle point = 1990
    final cycle point = 1992
    [[graph]]
        R1 = install_cold
        P1Y = """
            install_cold[^] => foo? => bar?

            # The line below stops cycle points 1991 and 1992
            # from starting:
            bar:fail? => !foo?
        """
```

### Expected Behaviour

Behaviour should be unchanged with or without the second graph line (assuming bar succeeds): 1991/2 should run the same graph again.

wxtim:
Closes #6594 

**Check List**

### DIFF
--- a/cylc/flow/taskdef.py
+++ b/cylc/flow/taskdef.py
@@ -99,9 +99,10 @@ def generate_graph_parents(
 ) -> Dict['SequenceBase', List[TaskTuple]]:
     """Determine concrete graph parents of task tdef at point.
 
-    Infer parents be reversing upstream triggers that lead to point/task.
+    Infer parents by reversing upstream triggers that lead to point/task.
     """
     graph_parents: Dict['SequenceBase', List[TaskTuple]] = {}
+
     for seq, triggers in tdef.graph_parents.items():
         if not seq.is_valid(point):
             # Don't infer parents if the trigger belongs to a sequence that
@@ -339,8 +340,8 @@ class TaskDef:
                     if (
                         trig.offset_is_absolute or
                         trig.offset_is_from_icp or
-                        # Don't count self-suicide as a normal trigger.
-                        dep.suicide and trig.task_name == self.name
+                        # Don't count suicide as a normal trigger:
+                        dep.suicide
                     ):
                         has_abs = True
                     else:
--- a/changes.d/6602.fix.md
+++ b/changes.d/6602.fix.md
@@ -1,0 +1,1 @@
+Fix a bug where suicide triggers could prevent initial cycle point tasks spawning.
--- a/tests/integration/test_taskdef.py
+++ b/tests/integration/test_taskdef.py
@@ -1,0 +1,44 @@
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+async def test_almost_self_suicide(flow, scheduler, start):
+    """Suicide triggers should not count as upstream tasks when looking
+    to spawn parentless tasks.
+
+    https://github.com/cylc/cylc-flow/issues/6594
+
+    For the example under test, pre-requisites for ``!a`` should not be
+    considered the same as pre-requisites for ``a``. If the are then then
+    is parentless return false for all cases of ``a`` not in the inital cycle
+    and subsequent cycles never run.
+    """
+    wid = flow({
+        'scheduler': {'cycle point format': '%Y'},
+        'scheduling': {
+            'initial cycle point': 1990,
+            'final cycle point': 1992,
+            'graph': {
+                'R1': 'install_cold',
+                'P1Y': 'install_cold[^] => a? => b?\nb:fail? => !a?'
+            }
+        }
+    })
+    schd = scheduler(wid)
+    async with start(schd):
+        tasks = [str(t) for t in schd.pool.get_tasks()]
+        for task in ['1990/a:waiting', '1991/a:waiting', '1992/a:waiting']:
+            assert task in tasks


cylc/flow/taskdef.py
```diff
@@ -339,8 +340,8 @@ def has_only_abs_triggers(self, point):
                     if (
                         trig.offset_is_absolute or
                         trig.offset_is_from_icp or
-                        # Don't count self-suicide as a normal trigger.
-                        dep.suicide and trig.task_name == self.name
+                        # Don't count suicide as a normal trigger:
+                        dep.suicide
```
oliver-sanders, line 344: This undoes a change @hjoliver made in #4970 

https://github.com/cylc/cylc-flow/commit/5e7ffad17524d7a9ef139a040baa1a0b5cc0ba87#diff-b46dc58446bab522ef576c6d2e32a4276f57d60ff62e930e6f47762d47bf9507R296

hjoliver: Good spotting ...

oliver-sanders: (leaving you two to resolve this)

wxtim: @hjoliver  what do you think we should do about this?

hjoliver: So the remaining bit of those two lines protects the original fix for self-suicide (a self suicide trigger is also a suicide trigger).

hjoliver: Actually, by testing, it seems that line is no longer needed at all.

hjoliver: > Isn't that the test for your earlier change though?

Yes, it's one of the two functional tests added in the old PR.

> If bar fails, then foo will only have absolute triggers, but if bar succeeds then foo has a cycling trigger (bar:fail). The outcome of this method depends on somthing which hasn't yet happened!

It's not dynamically determined like that. The trigger exists in the graph, even if runtime events result it in not being needed.

wxtim: > It's not dynamically determined like that. The trigger exists in the graph, even if runtime events result it in not being needed.

Which is why an intervention here is wrong, and the source of the bug lies elsewhere (presumably where we check whether the optional output is required).

wxtim: > It's not dynamically determined like that. The trigger exists in the graph, even if runtime events result it in not being needed.

... and so the correct answer to `has_only_absolute_triggers` should be `False`

wxtim: The problem that leaves us with is that the user expects `a` to act like it is parentless, but it's not: It's not unreasonable to suggest that the non existance of an optional outcome to optional suicide trigger dependency (`b:fail? => !a?`) shouldn't stop `a` from triggering in subsequent cycles.

wxtim: > ~Actually, by testing, it seems that line is no longer needed at all.~ Correction, `17-c7backcompat-self-suicide-cycling.t` fails without the remaining bit of the line.

Does it?

hjoliver: > The problem that leaves us with ...

I'm not sure I follow that. Tasks can be parentless in some cycles and not in others, the system handles that. Optional outputs don't change anything with regard to triggering, just whether or not the task is deemed "final and complete" or "final and incomplete" if/when it runs to a final status. Do you have an example that behaves wrongly?

wxtim: It's not failing for me locally, or in CI, I'm not clear why you're changing the code to make it fail.

> Do you have an example that behaves wrongly?

The one in the ticket.

hjoliver: > It's not failing for me locally, or in CI, I'm not clear why you're changing the code to make it fail.

Wires crossed - my comment above said this test fails *without the remaining bit of that code line* (i.e., it fails if that bit is deleted) - hence the code change to back up my comment. I checking that because you removed half of the line in this PR and I was wondering if the other half was no longer needed either.

hjoliver: > >  Do you have an example that behaves wrongly?
>
> The one in the ticket.

But this PR fixes that one.  

I was responding an exchange that started with you seemingly suggesting that there was a more fundamental problem, or at least that this PR was wrong:

>>>>>> [HO] Why did you make this change?
>>>>>>
>>>>> [TP] Because it appears to work in this case, but on more thought I don't think that its right, ..
>>>>> If bar fails, then foo will only have absolute triggers, but if bar succeeds then foo has a cycling trigger (bar:fail). The outcome of this method depends on somthing which hasn't yet happened!
>>>>>
>>>> [HO] It's not dynamically determined like that. The trigger exists in the graph, even if runtime events result it in not being needed. 
>>>>
>>> [TP] The problem that leaves us with is that the user expects a to act like it is parentless, but it's not: It's not unreasonable to suggest that the non existance of an optional outcome to optional suicide trigger dependency (b:fail? => !a?) shouldn't stop a from triggering in subsequent cycles.
>>>
>> [HO] I'm not sure I follow that. Tasks can be parentless in some cycles and not in others, the system handles that. Optional outputs don't change anything with regard to triggering, just whether the task is deemed "final and complete" or "final and incomplete" if/when it runs to a final status. Do you have an example that behaves wrongly?
>>
> [TP] The one in the ticket.

At this point I'm confused and basically have no idea if you think there's a(nother) problem or not!

wxtim: I think this convo has got into a right mess.

Are you happy with me undoing the change from #4970 ?

oliver-sanders: (unsubscribing for now, ping me when resolved)

hjoliver: > However there is a distinction between graph edges which must happen, and those that may happen.

A bit off topic, but IMO that's not really true. All triggers *must* happen *if the associated output is generated*, but even required outputs don't have to be generated at runtime - they are only required IF the task runs. The tasks with required outputs might not run at all for several reasons, e.g. they could be downstream of an optional branch point, or downstream of a task that expired.

wxtim: > Presumably that's no longer the case? If so, then I'm happy because this change clearly does fix the problem on the ticket. And I failed to trigger any other bug by attempting to target the change.

Yes, I think it's fine.

oliver-sanders: @hjoliver, please resolve this comment when happy.

oliver-sanders (approved):
Tested as working :+1:.

I don't get why we would want to exclude self-edges in this function so I think this change makes sense.

hjoliver:
> I don't get why we would want to exclude self-edges in this function 

Looking back at my original change (from 2022), the explanation at the time was something to do with the way suicide triggers work under SOD.

```
foo:fail => !bar
```
In Cylc 7, `bar` would be pre-spawned as waiting, and if `foo` fails the suicide trigger removes `bar`.

In Cylc 8, the suicide trigger is no longer necessary, but for back-compat it works as follows: if `foo` fails it spawns `bar` (on demand) and then the suicide trigger immediately removes it again.

Self-suicide (`foo:fail => !foo`) works the same way in Cylc 7, but there is a difference (from non-self suicide) in Cylc 8 because the trigger does not need to spawn `foo` (it was already spawned).

Anyhow, that changed fixed a bug at the time, and this one, which partially reverts the change, fixes another without bringing the original bug back (confirmed above), so I'm not going to blow any more time on this now.

hjoliver (approved):
Good, thanks @wxtim

wxtim:
This came up randomly on my music when I was re-checking this...

image

Issue #6506, opened 2024-11-29
running workflows may appear stopped on network filesystems


oliver-sanders:
**Regression of:** https://github.com/cylc/cylc-flow/pull/2943
**Bug report:** https://cylc.discourse.group/t/cylc-set-slow-to-be-able-to-run-after-workflow-started/1073/5

On network filesystems, there is a lag between a file being created on one host and it appearing on another.

The Cylc `contact` file contains the details of the running scheduler and is how we detect if a workflow is running or not.

Through a combination of running `fsync` on the contact file, and listing the directory it is contained in, we can force the network filesystem to synchronize the file reducing this lag.

By default, Cylc installs each run of a workflow into a numbered directory (`run1`, run2`, ...) with a `runN` symlink pointing at the most recent run number. It would appear that we need to perform a filesystem listing on this symlink before the contact file is synchronized to other nodes on the network.

There is another symlink that might trip things up which is configured by [`global.cylc[install][symlink dirs]run`](https://cylc.github.io/cylc-doc/stable/html/reference/config/global.html#global.cylc[install][symlink%20dirs][%3Cinstall%20target%3E]run). We will likely need to perform an additional filesystem listing if this symlink is created.

hjoliver:
(Calling this a bug in Cylc seems a little harsh! but fair enough)

ColemanTom:
We're seeing another potential spot. I could be wrong about it, but

1. Cylc submits a job (so ssh + mkdir + qsub)
2. QSUB submits the job and it starts trying to run on a different node
3. PBS option has `-k` option, so job.out and job.err should go straight into the log/job directory
4. PBS is failing to start running the job because it is failing to open the job.out file
5. The log folder tree is on NFS on the HPC

That to me means a potential that the log folder has no synced across NFS. Does that sound reasonable/possible?

dpmatthews:
Sounds possible but, if so, that would have to be addressed by PBS.

wxtim:
@oliver-sanders  - should #6603  have closed this?

I wonder if 

```md
closes #6603 probably
```

might stop the linking working?

oliver-sanders:
It may well have, however, PRs merged into bugfix branches don't close issues anyway.

oliver-sanders:
Closed by https://github.com/cylc/cylc-flow/pull/6603

PR #6603, opened 2025-02-11
work around NFS caching issues

oliver-sanders:
Closes https://github.com/cylc/cylc-flow/issues/6506 (hopefully!)

NFS filesystems have a caching mechanism that can take a while to clear. This may cause workflows to appear stopped on the CLI, in the Tui & GUI when they are in fact running.

Refresh NFS cache before:

* Scanning workflows:
  * Fixes: Delay between startup and workflows appearing in `cylc scan`, `cylc tui` and `cylc gui` (on other servers).
  * Cost: We've traded a file access for a directory listing (`.service` dir only contains a few files, no biggie).
* Loading the contact file:
  * Fixes: Delay between workflow startup and the CLI being able to operate on the workflow on other servers (the second three cases in the test script).
  * Cost: An additional file access (always) and additional directory listings (only when needed).
* Inferring the latest run:
  * Fixes: Workflow not visible on one server after being installed on another (the last three cases in the test script).
  * Cost: An additional file access (always) and additional directory listings (only when needed).

> Note these cashes are specific to the server, so you may get different results running Cylc commands on different servers.

<details>
<summary>To test, try running this script. It should say "no fix needed" for all cases.</summary>

```bash
#!/bin/bash -l

set -u

WORKFLOW="$1"
SERVER="$2"
RUN_ROOT="$HOME/cylc-run/$WORKFLOW"

reset () {
  cylc stop "*" >/dev/null
  rm -r "$HOME/cylc-run/$WORKFLOW" 2>/dev/null
  rm -r "$DATADIR/cylc-run/$WORKFLOW" 2>/dev/null
  rm -r "$SCRATCH/cylc-run/$WORKFLOW" 2>/dev/null
}

test_startup () {
  local run="$1"

  if cylc ping "$WORKFLOW/$run" >/dev/null 2>/dev/null; then
    echo "  no fix needed"
    return
  fi

  local directory
  for directory in "$HOME/cylc-run" "$HOME/cylc-run/$WORKFLOW" "$HOME/cylc-run/$WORKFLOW/$run"; do
    ls "$directory" >/dev/null
    if cylc ping "$WORKFLOW/$run" >/dev/null 2>/dev/null; then
      echo "  ls $directory needed"
      return
    fi
  done
}


run () {
  echo -e "\n$@"
  "$@" --pause "$WORKFLOW" >/dev/null 2>/dev/null
}

reset

run cylc vip --host=localhost
test_startup run1
reset

run cylc vip --host=localhost --run-name=foo
test_startup foo
reset

run cylc vip --host=localhost --no-run-name
test_startup ""
reset

run cylc vip --host="$SERVER"
test_startup runN
reset

run cylc vip --host="$SERVER" --run-name=foo
test_startup foo
reset

run cylc vip --host="$SERVER" --no-run-name
test_startup ""
reset

run ssh "$SERVER" CYLC_ENV_NAME="$CYLC_ENV_NAME" cylc vip --host=localhost
test_startup runN
reset

run ssh "$SERVER" CYLC_ENV_NAME="$CYLC_ENV_NAME" cylc vip --host=localhost --run-name=foo
test_startup foo
reset

run ssh "$SERVER" CYLC_ENV_NAME="$CYLC_ENV_NAME" cylc vip --host=localhost --no-run-name
test_startup ""
reset

```

</details>

**Check List**

### DIFF
--- a/cylc/flow/network/scan.py
+++ b/cylc/flow/network/scan.py
@@ -316,10 +316,21 @@ async def is_active(flow, is_active):
             False to filter for stopped and unregistered flows.
 
     """
-    contact = flow['path'] / SERVICE / CONTACT
-    _is_active = contact.exists()
+    service = flow['path'] / SERVICE
+    # NOTE: We must list the service directory contents rather than checking
+    # for the existence of the contact file directly, because listing the
+    # directory forces NFS filesystems to recompute their local cache.
+    # See https://github.com/cylc/cylc-flow/issues/6506
+    try:
+        contents = await scandir(service)
+    except FileNotFoundError:
+        _is_active = False
+    else:
+        _is_active = any(
+            path.name == WorkflowFiles.Service.CONTACT for path in contents
+        )
     if _is_active:
-        flow['contact'] = contact
+        flow['contact'] = service / CONTACT
     return _is_active == is_active
 
 
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -25,7 +25,7 @@ See also:
 from contextlib import suppress
 from enum import Enum
 import errno
-from queue import deque
+from collections import deque
 import os
 from pathlib import Path
 import re
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -621,7 +621,6 @@ def get_workflow_srv_dir(id_):
 
 
 def load_contact_file(id_: str, run_dir=None) -> Dict[str, str]:
-    """Load contact file. Return data as key=value dict."""
     if not run_dir:
         path = Path(get_contact_file_path(id_))
     else:
@@ -630,6 +629,19 @@ def load_contact_file(id_: str, run_dir=None) -> Dict[str, str]:
             WorkflowFiles.Service.DIRNAME,
             WorkflowFiles.Service.CONTACT
         )
+
+    # refresh NFS cache before proceeding, the contact file might magically
+    # appear, see https://github.com/cylc/cylc-flow/issues/6506
+    if not path.exists():
+        # list all directories between ~/cylc-run and the the workflow
+        # run dir (inclusive)
+        cylc_run_dir = get_cylc_run_dir()
+        for subdir in reversed(path.relative_to(cylc_run_dir).parents):
+            try:
+                list((cylc_run_dir / subdir).iterdir())
+            except FileNotFoundError as exc:
+                raise ServiceFileError("Couldn't load contact file") from exc
+
     try:
         with open(path) as f:
             file_content = f.read()
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -22,6 +22,7 @@ See also:
 
 """
 
+from contextlib import suppress
 from enum import Enum
 import errno
 import os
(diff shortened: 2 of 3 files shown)


cylc/flow/network/scan.py
```diff
@@ -316,10 +316,21 @@ async def is_active(flow, is_active):
...
+    # directory forces NFS filesystems to recompute their local cache.
+    # See https://github.com/cylc/cylc-flow/issues/6506
+    try:
+        contents = await scandir(service)
+    except FileNotFoundError:
+        _is_active = False
+    else:
+        _is_active = any(
+            path.name == WorkflowFiles.Service.CONTACT for path in contents
+        )
```
MetRonnie, line 331: Is this iteration needed as part of the listing, or could it be instead?
```suggestion
        _is_active = (service / CONTACT).exists()
```

oliver-sanders: Yes, this is the point of the PR.

oliver-sanders: We can either iterate (compute cost) or go back to the filesystem (FS cost).

There are only a couple of items in the directory, and the contact file will probs be the first (alpha order) so this is going to be faster.

MetRonnie:
```suggestion
        _is_active = CONTACT in contents
```

cylc/flow/workflow_files.py
```diff
@@ -620,8 +621,29 @@ def get_workflow_srv_dir(id_):
...
+        path: The directory to refresh.
+
+    Raises:
+        FileNotFoundError: If any of the directories between ~/cylc-run and
+        this directory (inclsive) are not present.
+
+    """
+    cylc_run_dir = get_cylc_run_dir()
+    for subdir in reversed(path.relative_to(cylc_run_dir).parents):
+        list((cylc_run_dir / subdir).iterdir())
```
MetRonnie, line 643: Is is necessary to list all adjacent files along the hierarchy down to the one we care about?

Just thinking about a big job log hierarchy.

Maybe use this to avoid loading into memory
```suggestion
    for subdir in reversed(path.relative_to(cylc_run_dir).parents):
        deque((cylc_run_dir / subdir).iterdir(), maxlen=0)
```

https://docs.python.org/3/library/itertools.html#itertools-recipes

MetRonnie (approved):
LGTM, I have tested using your script, with my code suggestions applied, and it passes

hjoliver (approved):
LGTM 

Not merging as it sounds like @MetRonnie still expects his suggestions to be committed first.

cylc/flow/network/scan.py
```diff
@@ -326,9 +326,7 @@ async def is_active(flow, is_active):
     except FileNotFoundError:
         _is_active = False
     else:
-        _is_active = any(
-            path.name == WorkflowFiles.Service.CONTACT for path in contents
-        )
+        _is_active = CONTACT in contents
```
oliver-sanders, line 329: This suggestion doesn't work, `CONTACT` is a rel path, dir listings are absolute which is why I had used `path.name`.

MetRonnie: D'oh

cylc/flow/workflow_files.py
```diff
@@ -22,8 +22,10 @@
 
 """
 
+from contextlib import suppress
 from enum import Enum
 import errno
+from queue import deque
```
MetRonnie, line 28:
```suggestion
from collections import deque
```

oliver-sanders: Dammit.

MetRonnie (approved):

MetRonnie (changes requested):
Just noticed: needs to be rebased onto 8.4.x?

MetRonnie (approved):

Issue #6567, opened 2025-01-20
data store: blank task state


oliver-sanders:
A user has encountered a traceback in `cylc tui` which suggests that a task had a blank status:

```
  File "/.../cylc-8.3.6-1/lib/python3.9/site-packages/cylc/flow/tui/util.py", line 118, in get_task_icon
    ret.append(TASK_ICONS[status])
KeyError: ''
```

Tasks in the store should always have a status, even far future tasks should be `waiting`.

It's not clear how to reproduce this traceback (I haven't seen it before), possibly just some awkward edge case 🤦. We could protect Tui against this with a fallback to `waiting`, but it would be much better to find and cure the issue than cover it over.

dwsutherland:
I'm not entirely sure how to reproduce it.. 
The `state` of a task is either picked up from:
- the given `itask` for active tasks
- batched for DB extraction
```
        # Active, but not in the data-store yet (new).
        if tp_id in self.n_window_nodes:
            self._process_internal_task_proxy(itask, tproxy)
            # Has run before, so get history.
            # Cannot batch as task is active (all jobs retrieved at once).
            if itask.submit_num > 0:
                flow_db = self.schd.workflow_db_mgr.pri_dao
                for row in flow_db.select_jobs_for_datastore(
                        {itask.identity}
                ):
                    self.insert_db_job(1, row)
        else:
            # Batch non-active node for load of DB history.
            self.db_load_task_proxies[itask.identity] = (
                itask,
                is_parent,
            )
```
The active `itask` should have a status:
```
    def spawn_task(
        self,
        name: str,
        point: 'PointBase',
        flow_nums: Set[int],
        flow_wait: bool = False,
    ) -> Optional[TaskProxy]:
        """Return a new task proxy for the given flow if possible.
        We need to hit the DB for:
        - submit number
        - task status
        - flow-wait
        - completed outputs (e.g. via "cylc set")
        If history records a final task status (for this flow):
        - if not flow wait, don't spawn (return None)
        - if flow wait, don't spawn (return None) but do spawn children
        - if outputs are incomplete, don't auto-rerun it (return None)
        Otherwise, spawn the task and load any completed outputs.
        """
        submit_num, prev_status, prev_flow_wait = (
            self._get_task_history(name, point, flow_nums)
        )
        # Create the task proxy with any completed outputs loaded.
        itask = self._get_task_proxy_db_outputs(
            point,
            self.config.get_taskdef(name),
            flow_nums,
[...]

PR #6607, opened 2025-02-13
data_store: fix traceback

oliver-sanders:
Traceback spotted in the wild.

Found this traceback in a scheduler log. It happened immediately after workflow restart.

**Check List**

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -689,10 +689,13 @@ class DataStoreMgr:
             else:
                 user_defined_meta[key] = val
         workflow.meta.user_defined = json.dumps(user_defined_meta)
-        workflow.tree_depth = max([
+        workflow.tree_depth = max(
             len(val)
-            for val in config.get_first_parent_ancestors(pruned=True).values()
-        ]) - 1
+            for val in (
+                config.get_first_parent_ancestors(pruned=True).values()
+                or (1,)
+            )
+        ) - 1
 
         if get_utc_mode():
             time_zone_info = TIME_ZONE_UTC_INFO


cylc/flow/data_store_mgr.py
```diff
@@ -689,10 +689,13 @@ def generate_definition_elements(self):
...
-        workflow.tree_depth = max([
+        workflow.tree_depth = max(
             len(val)
-            for val in config.get_first_parent_ancestors(pruned=True).values()
-        ]) - 1
+            for val in (
+                config.get_first_parent_ancestors(pruned=True).values()
+                or (1,)
+            )
+        ) - 1
```
oliver-sanders, line 698: Default to a `tree_depth` of `0`.

@dwsutherland does this make sense?

dwsutherland: Not sure how 
`config.get_first_parent_ancestors(pruned=True).values()`
could be zero.. 
Maybe some sort of inheritance error:
```
    def get_first_parent_ancestors(self, pruned=False):
        if pruned:
            # prune non-task namespaces from ancestors dict
            pruned_ancestors = {}
            for key, val in self.runtime['first-parent ancestors'].items():
                if key not in self.taskdefs:
                    continue
                pruned_ancestors[key] = val
            return pruned_ancestors
        else:
            return self.runtime['first-parent ancestors']
```
```
        for name in self.cfg['runtime']:
            try:
                self.runtime['linearized ancestors'][name] = c3.mro(name)
                self.runtime['first-parent ancestors'][name] = (
                    c3_single.mro(name))
            except RecursionError:
                raise WorkflowConfigError(
                    "circular [runtime] inheritance?"
                ) from None
            except Exception as exc:
                # catch inheritance errors
                # TODO - specialise MRO exceptions
                raise WorkflowConfigError(str(exc)) from None
```

I assume it's always 1 minimum otherwise (`root`)...

This fix makes sense as a workaround...

oliver-sanders:
Ok, we don't know why this is happening, but it's easy enough to defend against.

dwsutherland (approved):

oliver-sanders:
Hmmm, I've just been shown a `cylc tui` traceback where `first_parent` was `None`.

I wonder if it's the same root cause as this issue.

@dwsutherland, any ideas about how a task or family could end up with no `first_parent`?

dwsutherland:
>  any ideas about how a task or family could end up with no `first_parent`?

Perhaps with orphans?

The only other way I can think of is..:
If the node is pruned but then reintroduced somehow bypassing the normal creation (setting of static fields), like an update is applied post prune..
However, `updates` should be ignored if there's no node `added` or existing.. and any `added`/existing should have called a populated field..

We really need to be able to reproduce..

oliver-sanders:
> We really need to be able to reproduce..

I can reliably reproduce missing `first_parent` elements in `cylc tui`, however, the example is very complicated.

All I can say so far:
* Some families arriving with no `first_parent` elements.
* This can be seen both in `cylc tui` (direct from scheduler) and `cylc gui` (via cylc-uiserver).
* I haven't found any pattern to the affected families yet. The one's I've spotted so far all inherited (implicitly) from `root`.

dwsutherland:
> Some families arriving with no `first_parent` elements.

Well, `root` would always have no `first_parent`?  I assume you mean others..

> I haven't found any pattern to the affected families yet. The one's I've spotted so far all inherited (implicitly) from `root`.

I wonder if this is similar to the tasks without state problem:
https://github.com/cylc/cylc-flow/issues/6567

oliver-sanders:
> Well, root would always have no first_parent? I assume you mean others.

Yes.

I thought I had pinned this down to `generate_ghost_family` as it is possible for a family to not be assigned a `firstParent` here, however, these families don't seem to match the ones coming through GraphQL with missing `firstParent` fields, so I think I've misunderstood.

PR #6611, opened 2025-02-14
tui: open log files in external application

oliver-sanders:
Requested in: https://cylc.discourse.group/t/feature-request-cylc-tui-searching-job-out-job-err-files/1122/2

* Allow log files to be open in external applications.
* Tui will suspend whilst the external tool is open, and resume once it has closed.
* Options implemented are `$EDITOR`, `$GEDITOR`, `$PAGER` and `vim` as a backup.

tui-external-editor

**Check List**

- [x] Tests are included - very difficult to test
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at [cylc/cylc-doc/pull/797.](https://github.com/cylc/cylc-doc/pull/797)


cylc/flow/tui/overlay.py
```diff
@@ -473,6 +478,67 @@ def open_log(*_, filename=None, close=False):
...
+            host_widget.set_text(f'Host: {host}')
+            file_widget.set_text(f'Path: {path}')
+
+        with tempfile.NamedTemporaryFile('w+') as temp_file:
+            # write the text into a temp file
+            temp_file.write(text)
+            temp_file.seek(0, 0)
+
+            # make the file readonly to avoid confusion
+            os.chmod(temp_file.name, stat.S_IRUSR)
```
oliver-sanders, line 513: This is the approach we used to use `cylc view`. Dump the text into a temp file and mark it as readonly so that people don't think they can edit it.

Note, we can't reliably edit the file locally because it might not be a local file.

changes.d/6611.feat.md
```diff
@@ -0,0 +1 @@
+Tui: Add ability to open log files in external tools. Configure your `$EDITOR`, `$GEDITOR` or `$PAGER` options to configure which tool is used.
```
hjoliver, line 1: Too much "configure".
```suggestion
Tui: Add ability to open log files in external tools. Configure your `$EDITOR`, `$GEDITOR` or `$PAGER` options to choose which tool is used.
```

hjoliver:
For coverage, I think you could test this with a fake editor that just reads the file the in, then exits?

hjoliver (commented):
Very nice, but I think you could go a little bit further to avoid giving the user an editor that that don't know how to use:
- use the value of `$EDITOR` etc.
- and possibly also check that the editor (via environment or default) exists on the system
- and if no editors are available print a help message on what environment vars to configure

hjoliver:
> >  use the value of $EDITOR etc.
> 
> That is exactly what this PR is doing?

I meant print it in the choose-your-editor screen:

image

> If the editor is not installed, then they will see the "command not found" error message in the terminal for a few seconds before Tui restores.

Yeah, I saw that. That's fine, I just thought it would be slightly nicer for check if the command exists in `$PATH` first, and not even provide it as an option if it isn't there.

> If the editor is not installed, then they will see the "command not found" error message in the terminal for a few seconds before Tui restores.

EDITOR can be configured centrally, by default, in which case it is quite possible to be an editor that the user doesn't know how to drive.

> >  and if none are available print a help message on what environment vars to configure
> 
> I have already provided vim as a fallback.

Obviously I love vim and think everyone should be forced to use it, but (a) it is possible for vim not to be installed; and (b) vim is notoriously difficult for the uninitiated. 

(Famously some years back at NIWA a new vim user repeatedly and accidentally encrypted his files with some control sequence that performed the encryption using some part of the sequence as the key and then saved the encrypted file and exited the editor 🤣 )

hjoliver:
> IMO it's clearer this way (serves as functional documentation for how to configure).

Functional documentation is nice, but it's of secondary importance in a UI to functional obviousness. Many users are only vaguely aware, or unaware, of the content of those variables. We could still print a line underneath on how to configure.

> I'm not really sure what you're asking for in the other comments...

OK, TLDR:
- labeling UI buttons with environment variable names rather than the actual application names is unnecessarily obtuse
- `vim` (much as I love it) is a bad default editor - non vim-users can't even figure out how to exit from it

hjoliver:
How about this: I have `EDITOR=nvim`, `GEDITOR` and `PAGER` are not defined, `vim` and `less` are installed, `gvim` is not installed.

image

It is fine to have `vim` as a default now, because it is listed explicitly - the user won't unknowingly choose it.

```diff
diff --git a/cylc/flow/tui/overlay.py b/cylc/flow/tui/overlay.py
index a2fb85f3e..c109bb1e9 100644
--- a/cylc/flow/tui/overlay.py
+++ b/cylc/flow/tui/overlay.py
@@ -41,6 +41,7 @@ from functools import partial
 import os
 import re
 import shlex
+from shutil import which
 import stat
 from subprocess import Popen
 import sys
@@ -551,19 +552,25 @@ def log(app, id_=None, list_files=None, get_log=None):
                     (
                         'pack',
                         urwid.Button(
-                            label,
+                            command,
                             align='left',
                             on_press=partial(open_in_editor, command=command),
                         ),
                     )
-                    for label, command in (
-                        ('$EDITOR', os.environ.get('EDITOR', 'vim')),
-                        ('$GEDITOR', os.environ.get('GEDITOR', 'gvim -f')),
-                        ('$PAGER', os.environ.get('PAGER', 'less')),
-                        ('vim', 'vim'),
+                    for command in set(
+                        opt for opt in (
+                            os.environ.get('EDITOR', 'vim'),
+                            os.environ.get('GEDITOR', 'gvim -f'),
+                            os.environ.get('PAGER', 'less'),
+                            'vim'
+                        ) if which(opt.split()[0])
                     )
                 ),
             ]),
+            urwid.Text(
+                "(Configure apps to open logs via"
+                " $EDITOR, $GEDITOR, $PAGER)"
+            ),
             urwid.Divider(),
             text_widget,
         ]),
```

hjoliver (approved):
All good (noting tests need updating now).

hjoliver:
> Applied suggestion with changes:

Thanks, I'll take that as a decent compromise - approved, so maybe it can be merged while I'm away.

----

> Removed the set() as this scrambles the order making it unclear which env var resulted in which editor.

My intention there was just to prevent `vim` being listed twice, but I see you've dropped that as a final explicit option anyway.

> Removed the which(). I know this was trying to be helpful, but it created more problems than it solved by silently hiding editors the user had configured, preventing them from seeing the "command not found" error they would have seen otherwise.
> ... Note, the user may need to load an environment to get their editor of choice.

It only silently hid editors that are not accessible in the current environment.

It's a matter of opinion which is more of a problem - hiding configured editors that are not available, or hiding the fact that configured editors are not available. But OK, I can see it both sides of that argument, so I'll capitulate on this one!

All good now...

oliver-sanders:
> All good (noting tests need updating now).

Dammit, I did update the tests, but CI has `$EDITOR` configured differently (derp), needed to standardise the environment. Sorted.

markgrahamdawson:
I tested with Gedit (`export GEDITOR=gedit` in bash.rc) and VScode (`export GEDITOR=code`) in .bashrc
gedit worked fine but vs code just opens up a blank file 🤔

oliver-sanders:
Works for me, use `GEDITOR='code --wait'`.

* A new tab will open in vscode.
* Tui will background whilst the tab is open.
* Tui will restore once the tab is closed.

markgrahamdawson (approved):

PR #6623, opened 2025-02-19
host-select: fix compatibility with force-condemned hosts

oliver-sanders:
The "force mode" option in the auto-restart feature hasn't been used for a while.

It still works at Cylc 8, but force-condemning a host causes workflow startup to fail.

This PR:
* Fixes the issue with host-select.
* Adds a note about the "force-condemn" mode in the cfgspec docs.
* A cylc-doc PR will follow in due course. ... Issue https://github.com/cylc/cylc-doc/issues/809

For info on the "force mode", see the Cylc 7 docs: https://cylc.github.io/cylc-doc/7.9.3/html/running-suites.html#auto-stop-restart

**Check List**

- [ ] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened - soon, soon ...

### DIFF
--- a/changes.d/6623.fix.md
+++ b/changes.d/6623.fix.md
@@ -1,4 +1,4 @@
-auto restart: The "force condemn" option (that tells workflows running on a
+Auto restart: The "force condemn" option (that tells workflows running on a
 server to shutdown as opposed to migrate) hasn't worked with the host-selection
 mechanism since Cylc 8.0.0. This has now been fixed and the "force condemn"
 option has been restored in the documentation.
--- a/changes.d/6623.fix.md
+++ b/changes.d/6623.fix.md
@@ -1,0 +1,4 @@
+auto restart: The "force condemn" option (that tells workflows running on a
+server to shutdown as opposed to migrate) hasn't worked with the host-selection
+mechanism since Cylc 8.0.0. This has now been fixed and the "force condemn"
+option has been restored in the documentation.
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -828,23 +828,22 @@ with Conf('global.cylc', desc='''
             Conf('condemned', VDR.V_ABSOLUTE_HOST_LIST, desc=f'''
                 List run hosts that workflows should *not* run on.
 
-                Any hosts listed here will be subtracted from the
-                `available <global.cylc[scheduler][run hosts]>`
-                hosts:
+                These will be subtracted from the
+                `available <global.cylc[scheduler][run hosts]>`:
 
                 * Workflows will not start on condemned hosts.
                 * Workflows that are running on condemned hosts will attempt
-                  to migrate to an uncondemned host (providing the
+                  to migrate to an available host (providing the
                   `auto restart
                   <global.cylc[scheduler][main loop][auto restart]>`
                   plugin is enabled).
 
                 This feature can be used to drain a host for patching, or
-                remove a host that is surplus to requirement.
+                remove a host that is surplus to requirements.
 
-                Hostnames listed here may be followed by a ``!`` character.
-                This activates "force mode", workflows running on a
-                force-condenmed host will shutdown rather than attempting to
+                If a hostname listed here is followed by a ``!`` character
+                ("force mode"), workflows running on it
+                will shutdown rather than attempting to
                 migrate (providing the
                 `auto restart
                 <global.cylc[scheduler][main loop][auto restart]>` plugin
@@ -871,7 +870,7 @@ with Conf('global.cylc', desc='''
 
                 .. versionchanged:: 8.4.2
 
-                   The "force-condemn" option caused issues at workflow
+                   The force-condemn ("!") option caused issues at workflow
                    startup for Cylc versions between 8.0.0 and 8.4.1
                    inclusive.
 
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -826,15 +826,54 @@ with Conf('global.cylc', desc='''
                    range.
             ''')
             Conf('condemned', VDR.V_ABSOLUTE_HOST_LIST, desc=f'''
-                These hosts will not be used to run jobs.
+                List run hosts that workflows should *not* run on.
 
-                If workflows are already running on
-                condemned hosts, Cylc will shut them down and
-                restart them on different hosts.
+                Any hosts listed here will be subtracted from the
+                `available <global.cylc[scheduler][run hosts]>`
+                hosts:
+
+                * Workflows will not start on condemned hosts.
+                * Workflows that are running on condemned hosts will attempt
+                  to migrate to an uncondemned host (providing the
+                  `auto restart
+                  <global.cylc[scheduler][main loop][auto restart]>`
+                  plugin is enabled).
+
+                This feature can be used to drain a host for patching, or
+                remove a host that is surplus to requirement.
+
+                Hostnames listed here may be followed by a ``!`` character.
+                This activates "force mode", workflows running on a
+                force-condenmed host will shutdown rather than attempting to
+                migrate (providing the
+                `auto restart
+                <global.cylc[scheduler][main loop][auto restart]>` plugin
+                is enabled).
+
+                .. rubric:: Example:
+
+                .. code-block:: cylc
+
+                   [scheduler]
+                       [[run hosts]]
+                           # there are three hosts in the "pool"
+                           available = host1, host2, host3
+
+                           # however two have been taken out:
+                           # * workflows running on "host1" will attempt to
+                           #   restart on "host3"
+                           # * workflows running on "host2" will shutdown
+                           condemned = host1, host2!
 
                 .. seealso::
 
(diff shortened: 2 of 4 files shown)


cylc/flow/host_select.py
```diff
@@ -128,6 +128,13 @@ def select_workflow_host(cached=True):
     # be returned with the up-to-date configuration.
     global_config = glbl_cfg(cached=cached)
 
+    # condemned hosts may be suffixed with an "!" to activate "force mode"
+    blacklist = []
+    for host in global_config.get(['scheduler', 'run hosts', 'condemned'], []):
+        if host.endswith('!'):
+            host = host[:-1]
```
oliver-sanders, line 135: Note, this line *is* covered by the test amended in this PR, however, that test is not run in CI due to it's shared filesystem setup.

hjoliver (commented):
Looks fine, just some minor verbosity-reduction suggestions.

A couple of big-picture questions though:
- It's weird that the documentation had been deliberately removed - does that suggest there was an intention to remove this option??
- Maybe it's bedded in for historical reasons as this point but I'm not sure force-condemn" is a very good name for this feature. Is it because there are some workflows that auto-migrate won't touch? (long-running background tasks?). But even then, shouldn't force-condemn try to auto-migrate first, then kill any remaining workflows that failed to migrate?

oliver-sanders:
> It's weird that the documentation had been deliberately deleted

I think this more slipped the net than was deliberately deleted.

> Maybe it's bedded in for historical reasons as this point but I'm not sure "force-condemn" is a very good name

Happy to change, suggest an alternative.

> shouldn't force-condemn try to auto-migrate first, then kill any remaining workflows that failed to migrate

Some workflows cannot auto-migrate (e.g. workflows started in `--no-detach` mode) or might fail to migrate (e.g. user has scrambled their SSH config since starting the workflow).

The idea was to condemn the host first (i.e. ask workflows to migrate nicely), then mop up any stragglers with the "force" mode.

hjoliver:
> The idea was to condemn the host first (i.e. ask workflows to migrate nicely), then mop up any stragglers with the "force" mode.

OK that's fine. Leave the decision on when to force it up to the humans.

We should document that though, as some will assume the options are migrate-all or kill-all, and not think of migrate-if-possible before changing the global config again to kill the rest.

We should also document the circumstances under which workflows will not successfully auto-migrate.

I think the best place for that is the general "pool of hosts" doc in the user guide, would you agree?

oliver-sanders:
Yes, a cylc-doc PR will follow (see OP), it shouldn't go here in the config docs.

hjoliver:
I put up a cylc-doc issue, just not that we need to document *why* force-mode is needed in the first place (some workflows will not auto-migrate): https://github.com/cylc/cylc-doc/issues/809

hjoliver:
Regards the name "force condemn" ...

A node has been "condemned" as soon as it is listed as such, whether or not it still has workflows running on it.

(C.f. a condemned man hasn't been hung by the neck just yet; and a condemned build hasn't been demolished yet.)

I think we can just say that workflows running on condemned hosts will be told to auto-migrate, unless the hostname is prefixed by "!" in which case they will be killed.  There's probably no need to call that forced condemnation, right?

changes.d/6623.fix.md
```diff
@@ -0,0 +1,4 @@
+Auto restart: The "force condemn" option (that tells workflows running on a
+server to shutdown as opposed to migrate) hasn't worked with the host-selection
+mechanism since Cylc 8.0.0. This has now been fixed and the "force condemn"
+option has been restored in the documentation.
```
hjoliver, line 4:
```suggestion
Restored the option to shut down rather than auto-migrate workflows on a condemned
host. This option hasn't worked with the host-selection mechanism since Cylc 8.0.0. 
```

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -826,16 +826,54 @@ def default_for(
...
 
-                If workflows are already running on
-                condemned hosts, Cylc will shut them down and
-                restart them on different hosts.
+                These will be subtracted from the
+                `available <global.cylc[scheduler][run hosts]>`:
+
+                * Workflows will not start on condemned hosts.
+                * Workflows that are running on condemned hosts will attempt
+                  to migrate to an available host (providing the
```
hjoliver, line 836:
```suggestion
                * By default, workflows running on condemned hosts will attempt
                  to migrate to an available host (providing the
```

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -826,16 +826,54 @@ def default_for(
...
+                  <global.cylc[scheduler][main loop][auto restart]>`
+                  plugin is enabled).
+
+                This feature can be used to drain a host for patching, or
+                remove a host that is surplus to requirements.
+
+                If a hostname listed here is followed by a ``!`` character
+                ("force mode"), workflows running on it
+                will shutdown rather than attempting to
+                migrate (providing the
```
hjoliver, line 847:
```suggestion
                If a hostname listed here is followed by a ``!`` character
                workflows running on it will be told to shut down
                rather than migrate to another host (providing the
```

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -826,16 +826,54 @@ def default_for(
...
+
+                   [scheduler]
+                       [[run hosts]]
+                           # there are three hosts in the "pool"
+                           available = host1, host2, host3
+
+                           # however two have been taken out:
+                           # * workflows running on "host1" will attempt to
+                           #   restart on "host3"
+                           # * workflows running on "host2" will shutdown
```
hjoliver, line 864:
```suggestion
                           # * workflows running on "host2" will shut down
```

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -826,16 +826,54 @@ def default_for(
...
 
                 .. seealso::
 
                    :ref:`auto-stop-restart`
 
+                .. versionchanged:: 8.4.2
+
+                   The force-condemn ("!") option caused issues at workflow
+                   startup for Cylc versions between 8.0.0 and 8.4.1
+                   inclusive.
```
hjoliver, line 875:
```suggestion
                   The option to shut down ("hostname!") rather than migrate
                   ("hostname") workflows on condemned hosts caused problems
                   at workflow startup for Cylc versions 8.0.0 through 8.4.1.
```

tests/functional/restart/43-auto-restart-force-override-normal.t
```diff
@@ -50,7 +50,10 @@ create_test_global_config '' "
 ${BASE_GLOBAL_CONFIG}
 [scheduler]
     [[run hosts]]
-        available = ${CYLC_TEST_HOST_1}
+        available = ${CYLC_TEST_HOST_1}, ${CYLC_TEST_HOST_2}
+        # ensure the workflow can start if a host is force-condemned
+        # see #6623
```
hjoliver, line 55:
```suggestion
        # ensure the workflow can start if previously shut down on a
        # condemned host. See #6623
```

hjoliver (changes requested):

oliver-sanders:
> I think we can just say that workflows running on condemned hosts will be told to auto-migrate, unless the hostname is followed by "!" in which case they will be shut down. There's probably no need to call that forced condemnation, right?

I think we need a name for this. It would be clunky to cross-reference this from the docs otherwise.

Please suggest an alternative to "force". I will keep using "force" as a placeholder, happy to change, but need something to change it to.

hjoliver:
See my response on the cylc-doc issue: https://github.com/cylc/cylc-doc/issues/809#issuecomment-2686933685

The point is, hosts are by definition condemned from the moment they are declared as such. They don't suddenly become condemned only later, once/if all their workflows have disappeared. So "force condemned" simply doesn't make sense.

We need to document this:

> Condemned hosts have been sentenced to death, so:
>  - new workflows will not be started on a condemned host
>  - and existing workflows need to be stopped (with `!`) or migrated (without `!`)
>     - either way, the host is still condemned

changes.d/6623.fix.md
oliver-sanders: This isn't actually correct.

cylc/flow/cfgspec/globalcfg.py
oliver-sanders: Superfluous in combo with next paragraph.

oliver-sanders:
> It's weird that the documentation had been deliberately deleted - does that suggest there was an intention to remove this option??

@hjoliver 

Aah, the docs weren't deleted, they were simply moved.

https://cylc.github.io/cylc-doc/stable/html/plugins/main-loop/built-in/cylc.flow.main_loop.auto_restart.html#module-cylc.flow.main_loop.auto_restart

I'll update the pre-existing documentation in line with your terminology irks.

oliver-sanders:
Turns out this was already well documented, I just didn't look very hard.

https://cylc.github.io/cylc-doc/stable/html/plugins/main-loop/built-in/cylc.flow.main_loop.auto_restart.html#module-cylc.flow.main_loop.auto_restart

I'm going to close this PR and re-raise it with simpler docs pointing at the canonical documentation.

PR #6639, opened 2025-02-27
Kill foreground event handlers on timeout

### LINKED ISSUE #6615: shutdown event handler not killed on timeout
Seen in the wild, event handler not respecting its timeout?

* Workflow stalled.
* "workflow stalled" event handler was run, it timed out after 10 mins.
* Workflow hit stall timeout.
* "abort on stall timeout" event handler was run, this did not time out and had to be killed manually.

```
CRITICAL - Workflow stalled    
WARNING - PT3H stall timer starts NOW    
ERROR - [('workflow-event-handler-00', 'stall') cmd] suite_report.py 'stall' 'r7091_physics_test/run11' 'workflow stalled'
    [('workflow-event-handler-00', 'stall') ret_code] -9    
    [('workflow-event-handler-00', 'stall') err] killed on timeout (PT10M)
ERROR - stall EVENT HANDLER FAILED    
WARNING - stall timer timed out after PT3H    
ERROR - Workflow shutting down - "abort on stall timeout" is set    
INFO - platform: xce - remote tidy (on xcel01)    
ERROR - [('workflow-event-handler-00', 'shutdown') cmd] suite_report.py 'shutdown' 'r7091_physics_test/run11' '"abort on stall timeout" is set'
    [('workflow-event-handler-00', 'shutdown') ret_code] -15
ERROR - shutdown EVENT HANDLER FAILED    
INFO - DONE    
```

Why did the "workflow stalled" event handler abide by the timeout and the "abort on stall timeout" event handler ignore it?

This causes issues with auto-restart functionality because if an event handler ends up hanging for whatever reason, there is no timeout to stop it and the workflow will remain in the shutting down state indefinitely.

hjoliver:
Close #6615 

Shutdown handlers aren't run via the subprocess pool, which gets closed on shutdown, so they weren't subject to the process pool timeout. This PR imposes
the timeout on them anyway, outside of the pool.

**Check List**

### DIFF
--- a/cylc/flow/subprocpool.py
+++ b/cylc/flow/subprocpool.py
@@ -365,16 +365,22 @@ class SubProcPool:
             )
 
     @classmethod
-    def run_command(cls, ctx):
+    def run_command(cls, ctx, timeout: Optional[float] = None):
         """Execute command in ctx and capture its output and exit status.
+
+        Raises subprocess.TimeoutExpired (via subprocess.communicate) if
+        the command gets killed for exceeding a given timeout.
 
         Arguments:
             ctx (cylc.flow.subprocctx.SubProcContext):
                 A context object containing the command to run and its status.
+            timeout:
+                Timeout in seconds, after which to kill the command.
         """
         proc = cls._run_command_init(ctx)
         if proc:
-            ctx.out, ctx.err = (f.decode() for f in proc.communicate())
+            ctx.out, ctx.err = (
+                f.decode() for f in proc.communicate(timeout=timeout))
             ctx.ret_code = proc.wait()
             cls._run_command_exit(ctx)
 
--- a/cylc/flow/subprocpool.py
+++ b/cylc/flow/subprocpool.py
@@ -25,7 +25,7 @@ import shlex
 from tempfile import SpooledTemporaryFile
 from threading import RLock
 from time import time
-from subprocess import DEVNULL, run  # nosec
+from subprocess import DEVNULL, TimeoutExpired, run  # nosec
 from typing import TYPE_CHECKING, Any, Callable, List, Optional, Set
 
 from cylc.flow import LOG, iter_entry_points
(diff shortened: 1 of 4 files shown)


cylc/flow/workflow_events.py
```diff
@@ -302,14 +305,26 @@ def _send_mail(
...
+            self.proc_pool.put_command(ctx, callback=callback)
         else:
-            # Run command using process pool otherwise
-            self.proc_pool.put_command(
-                proc_ctx, callback=self._run_event_mail_callback)
+            # Run it in the foreground, but use the subprocess pool timeout.
+            try:
+                self.proc_pool.run_command(ctx, float(self.proc_timeout))
+            except TimeoutExpired:
+                ctx.ret_code = -9
```
oliver-sanders, line 325: The `timeout` command exits `124` if the timeout is exceeded. If setting an exit code, we might as well match this for consistency:

```suggestion
                ctx.ret_code = 124
```

cylc/flow/subprocpool.py
```diff
@@ -365,16 +365,22 @@ def put_command(
...
             ctx (cylc.flow.subprocctx.SubProcContext):
                 A context object containing the command to run and its status.
+            timeout:
+                Timeout in seconds, after which to kill the command.
         """
         proc = cls._run_command_init(ctx)
         if proc:
-            ctx.out, ctx.err = (f.decode() for f in proc.communicate())
+            ctx.out, ctx.err = (
+                f.decode() for f in proc.communicate(timeout=timeout))
```
oliver-sanders, line 383: The subprocess `timeout` option just raises an exception once the timeout has elapsed. You can catch this exception and retry the `proc.communicate` again because it doesn't actually kill the process. 

> The child process is not killed if the timeout expires, so in order to cleanup properly a well-behaved application should kill the child process and finish communication:
>
> -- https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate

Under normal operation (background rather than foreground) subprocpool implements its own timeout kill:

https://github.com/cylc/cylc-flow/blob/335216488793401ce3f7f90ebf07c11d3187b63c/cylc/flow/subprocpool.py#L280-L291

oliver-sanders (changes requested):

cylc/flow/subprocpool.py
hjoliver: Huh, good spotting. It doesn't stop the scheduler from shutting down anymore, but the subprocess is still running. I just assumed that the `communicate` timeout would kill it. (Isn't that what a "timeout" should do? otherwise it's more of a "timer").

hjoliver: Huh, good spotting. It doesn't stop the scheduler from shutting down anymore, but the subprocess is still running. I just assumed that the `communicate` timeout would kill it. (Isn't that what a "timeout" should do? otherwise it's more of a "timer").

cylc/flow/workflow_events.py
hjoliver: I initially took this suggestion, but I've backed out of it on implementing the process kill in response to your main comment.

The return code `-9` from `subprocess.wait()` (which is how I'm getting it now, not hardwiring it) reflects the value of the signal used to kill the process (SIGKILL is 9). Is that OK with you?

cylc/flow/subprocpool.py
hjoliver: I've added kill logic to the "foreground" `run_command` function.

oliver-sanders (approved):

PR #6640, opened 2025-02-27
Address some warnings

MetRonnie:
Spotted in pytest output

> [!NOTE]
> There is a risk with this because using comparison operators (e.g. `<`) between timezone-aware and naive (timezone-unaware) `datetime`s results in a type error. However I cannot find any instances of doing any such comparisons, and all the tests are passing.

> [!IMPORTANT]
> Python 3.15 is coming in Oct 2026, hence I think this ought to go into 8.7.0 at the latest

### Check List

- [x] No dependency changes 
- [x] No tests/docs/changelog needed

### DIFF
--- a/cylc/flow/wallclock.py
+++ b/cylc/flow/wallclock.py
@@ -45,7 +45,6 @@ TIME_FORMAT_EXTENDED = "%H:%M:%S"
 TIME_FORMAT_EXTENDED_SUB_SECOND = "%H:%M:%S.%f"
 
 TIME_ZONE_STRING_UTC = "Z"
-TIME_ZONE_UTC_UTC_OFFSET = (0, 0)
 
 TIME_ZONE_LOCAL_INFO = {
     "hours": get_local_time_zone()[0],
@@ -57,8 +56,8 @@ TIME_ZONE_LOCAL_INFO = {
 }
 
 TIME_ZONE_UTC_INFO = {
-    "hours": TIME_ZONE_UTC_UTC_OFFSET[0],
-    "minutes": TIME_ZONE_UTC_UTC_OFFSET[1],
+    "hours": 0,
+    "minutes": 0,
     "string_basic": TIME_ZONE_STRING_UTC,
     "string_extended": TIME_ZONE_STRING_UTC
 }
@@ -122,7 +121,6 @@ def get_time_string(
     override_use_utc: bool | None = None,
     use_basic_format: bool = False,
     date_time_is_local: bool = False,
-    custom_time_zone_info: dict | None = None,
 ):
     """Return a string representing the current system time.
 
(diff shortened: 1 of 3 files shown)


cylc/flow/wallclock.py
```diff
@@ -85,9 +87,9 @@ def now(override_use_utc=None):
 
     """
     if override_use_utc or (override_use_utc is None and _FLAGS['utc_mode']):
-        return datetime.utcnow(), False
+        return datetime.now(timezone.utc), False
```
oliver-sanders, line 90: There is a difference between the old and new methods:

```python
>>> str(datetime.utcnow())
'2025-02-28 13:18:42.595223'
>>> str(datetime.now(timezone.utc))
'2025-02-28 13:18:54.379479+00:00'
```

We may be stringifying datetimes like this, e.g. for the database. Need to make sure it doesn't cause issues.

cylc/flow/wallclock.py
```diff
@@ -85,9 +87,9 @@ def now(override_use_utc=None):
...
     else:
-        return datetime.now(), True
+        return datetime.now().astimezone(), True
```
oliver-sanders, line 92: Also differs in the same way:

```python
>>> str(datetime.now())
'2025-02-28 13:21:39.364016'
>>> str(datetime.now().astimezone())
'2025-02-28 13:21:39.715789+00:00'
```

tests/unit/cycling/test_iso8601.py
```diff
@@ -892,10 +895,10 @@ def test_next_simple_no_now(set_cycling_type):
     point = "next(T00Z)+P1D"
     output = ingest_time(point, my_now)
 
-    current_time = datetime.utcnow()
+    current_time = datetime.now(timezone.utc)
     # my_now is None, but ingest_time will have used a similar time, and
     # the returned value must be after current_time
-    output_time = datetime.strptime(output, "%Y%m%dT%H%MZ")
+    output_time = datetime.strptime(output, "%Y%m%dT%H%M%z")
```
oliver-sanders, line 901: Time zone format has changed?

cylc/flow/wallclock.py
MetRonnie: I've checked all uses of `cylc.flow.wallclock.now()` and we're not directly stringifying the native `datetime` object it returns anywhere. I've also added a test for `get_time_string()` to ensure it doesn't treat naive and tz-aware `datetime`s any differently. (Actually a lot of the logic in the wallclock module could be simplified to make better use of `datetime`'s handling of time zones but that will be for another day)

tests/unit/cycling/test_iso8601.py
MetRonnie: Using `Z` just means it matches and discards the `Z` at the end of `output`, leading to a naive `datetime`. Whereas using `%z` means it recognises `Z` as UTC, leading to a tz-aware `timezone`. The latter is needed for the comparison operation

cylc/flow/wallclock.py
```diff
@@ -133,46 +136,21 @@ def get_time_string(date_time, display_sub_seconds=False,
...
-        if date_time_is_local:
-            date_time_hours = TIME_ZONE_LOCAL_UTC_OFFSET_HOURS
-            date_time_minutes = TIME_ZONE_LOCAL_UTC_OFFSET_MINUTES
-        else:
-            date_time_hours, date_time_minutes = (0, 0)
-        diff_hours = custom_hours - date_time_hours
-        diff_minutes = custom_minutes - date_time_minutes
-        date_time = date_time + timedelta(
-            hours=diff_hours, minutes=diff_minutes)
-        time_zone_string = custom_string
```
MetRonnie, line 160: After noticing this block had no coverage, I realised it is not used

cylc/flow/wallclock.py
oliver-sanders: > we're not directly stringifying 

:+1:

> could be simplified

This might be true, but be aware we support alternative calendars that datetime does not which might explain this?

tests/unit/cycling/test_iso8601.py
oliver-sanders: What I'm asking is why is this diff present? The test passed as was before these changes.

Is this arbitrary "tidying", or was it needed due to a change in timestamp format?

oliver-sanders (approved):

MetRonnie:
@hjoliver ping

cylc/flow/wallclock.py
```diff
@@ -133,46 +136,20 @@ def get_time_string(date_time, display_sub_seconds=False,
...
```
MetRonnie, line 160: > @hjoliver, this PR removes a code block that isn't used any more. Can you have a quick look at it and merge if ok.

This code block was previously only used by the Cylc 7 GUI (7.8.15):

https://github.com/cylc/cylc-flow/blob/e60e8f7a6d77256af0e312237c6c8f7bf4fd38f8/lib/cylc/gui/updater_tree.py#L269-L277

However, it looks like it was only used for the same purpose that is already provided by `override_use_utc`:

https://github.com/cylc/cylc-flow/blob/e60e8f7a6d77256af0e312237c6c8f7bf4fd38f8/lib/cylc/gui/updater_tree.py#L190

https://github.com/cylc/cylc-flow/blob/e60e8f7a6d77256af0e312237c6c8f7bf4fd38f8/lib/cylc/state_summary_mgr.py#L125-L128

oliver-sanders: @hjoliver poke

oliver-sanders: @hjoliver poke

hjoliver:
Apologies for the massive delay!

hjoliver (approved):

hjoliver:
Several conflicts now, sorry.

PR #6643, opened 2025-02-28
host-select: fix compatibility with force-condemned hosts

oliver-sanders:
Supersedes https://github.com/cylc/cylc-flow/pull/6623

Fix an issue with host-select which prevented us from condemning hosts in "force mode".

Also, better link into the "auto restart" docs which can be found here: https://cylc.github.io/cylc-doc/stable/html/plugins/main-loop/built-in/cylc.flow.main_loop.auto_restart.html#module-cylc.flow.main_loop.auto_restart

I am **not** going to go applying terminology changes to this long-standing feature as part of this bugfix (out of scope).

**Check List**

### DIFF
--- a/changes.d/6623.fix.md
+++ b/changes.d/6623.fix.md
@@ -1,0 +1,3 @@
+Auto restart: The option to condemn a host in "force" mode (that tells
+workflows running on a server to shutdown as opposed to migrate) hasn't worked
+with the host-selection mechanism since Cylc 8.0.0. This has now been fixed.
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -833,7 +833,7 @@ with Conf('global.cylc', desc='''
                 preventing new workflows from starting on the "condemned" host.
 
                 Any workflows running on these hosts will either migrate
-                to another host, or shutdown according to
+                to another host, or shut down according to
                 :py:mod:`the configuration <cylc.flow.main_loop.auto_restart>`.
 
                 This feature requires ``auto restart`` to be listed
--- a/cylc/flow/cfgspec/globalcfg.py
+++ b/cylc/flow/cfgspec/globalcfg.py
@@ -826,15 +826,51 @@ with Conf('global.cylc', desc='''
                    range.
             ''')
             Conf('condemned', VDR.V_ABSOLUTE_HOST_LIST, desc=f'''
-                These hosts will not be used to run jobs.
+                List run hosts that workflows should *not* run on.
 
-                If workflows are already running on
-                condemned hosts, Cylc will shut them down and
-                restart them on different hosts.
+                These hosts will be subtracted from the
+                `available <global.cylc[scheduler][run hosts]>` hosts
+                preventing new workflows from starting on the "condemned" host.
+
+                Any workflows running on these hosts will either migrate
+                to another host, or shutdown according to
+                :py:mod:`the configuration <cylc.flow.main_loop.auto_restart>`.
+
+                This feature requires ``auto restart`` to be listed
+                in `global.cylc[scheduler][main loop]plugins`.
+
+                For more information, see the
+                :py:mod:`auto restart <cylc.flow.main_loop.auto_restart>`
+                plugin.
+
+                .. rubric:: Example:
+
+                .. code-block:: cylc
+
+                   [scheduler]
+                       [[main loop]]
+                           # activate the "auto restart" plugin
+                           plugins = auto restart
+                       [[run hosts]]
+                           # there are three hosts in the "pool"
+                           available = host1, host2, host3
+
+                           # however two have been taken out:
+                           # * workflows running on "host1" will attempt to
+                           #   restart on "host3"
+                           # * workflows running on "host2" will shutdown
+                           condemned = host1, host2!
 
                 .. seealso::
 
+                   :py:mod:`cylc.flow.main_loop.auto_restart`
                    :ref:`auto-stop-restart`
+
+                .. versionchanged:: 8.4.2
+
+                   The "force mode" (activated by a "!" suffix) caused issues
+                   at workflow startup for Cylc versions between 8.0.0 and
+                   8.4.1 inclusive.
 
                 .. versionchanged:: 8.0.0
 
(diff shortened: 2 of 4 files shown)


cylc/flow/host_select.py
```diff
@@ -128,6 +128,13 @@ def select_workflow_host(cached=True):
     # be returned with the up-to-date configuration.
     global_config = glbl_cfg(cached=cached)
 
+    # condemned hosts may be suffixed with an "!" to activate "force mode"
+    blacklist = []
+    for host in global_config.get(['scheduler', 'run hosts', 'condemned'], []):
+        if host.endswith('!'):
+            host = host[:-1]
```
oliver-sanders, line 135: Note, this line is covered by the test amended in this PR (confirm by running it on master), however, that test is not run in CI due to it's shared filesystem setup.

hjoliver:
> I am not going to go applying terminology changes to this long-standing feature as part of this bugfix (out of scope).

This needs a name change, which you agreed to on the PR that this supersedes - why have you changed your mind?

For readers who didn't follow the original discussion, "force condemned" is just bad terminology, given the universally accepted meaning of the word condemned (the subject has been sentenced to some *future* dire fate). Once a host is condemned we need to drain it of existing running workflows - either by telling them to migrate or by telling them to stop. The host is 100% condemned either way, even before the workflows migrate or stop, so "force condemned" just doesn't make sense.

There's really no need to use a specific term for this anyway since there is no associated config item. There are two ways of writing hostnames in the `condemned hosts` list: one with no `!` (migrate workflows) and one with a `!` (stop workflows). We can just document that.

changes.d/6623.fix.md
```diff
@@ -0,0 +1,3 @@
+Auto restart: The option to condemn a host in "force" mode (that tells
+workflows running on a server to shutdown as opposed to migrate) hasn't worked
```
hjoliver, line 2:
```suggestion
Auto restart: The option to to tell 
workflows running on a server to shutdown as opposed to migrate hasn't worked
```

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -826,16 +826,52 @@ def default_for(
...
 
-                If workflows are already running on
-                condemned hosts, Cylc will shut them down and
-                restart them on different hosts.
+                These hosts will be subtracted from the
+                `available <global.cylc[scheduler][run hosts]>` hosts
+                preventing new workflows from starting on the "condemned" host.
+
+                Any workflows running on these hosts will either migrate
+                to another host, or shutdown according to
```
hjoliver, line 836:
```suggestion
                to another host, or shut down according to
```

cylc/flow/cfgspec/globalcfg.py
```diff
@@ -826,16 +826,52 @@ def default_for(
...
+                           condemned = host1, host2!
 
                 .. seealso::
 
+                   :py:mod:`cylc.flow.main_loop.auto_restart`
                    :ref:`auto-stop-restart`
 
+                .. versionchanged:: 8.4.2
+
+                   The "force mode" (activated by a "!" suffix) caused issues
```
hjoliver, line 871:
```suggestion
                   The stop (as opposed to migrate) option (activated by a "!" suffix)
                   caused issues
```

hjoliver (changes requested):

hjoliver:
> There's really no need to use a specific term for this anyway since there is no associated config item. 

See my review tweaks above - I only had to nuke "force" in two rather unimportant places - the change log and one version-changed note, which confirms it's not something that even needs special terminology beyond the concepts of "migrate" and "stop".

oliver-sanders:
> I presume you mean the "force condemned" term? That needs to be dropped, which you more or less agreed to on the original PR - why have you changed your mind?

Please see the OP.

You were "irked" by me saying "force-condemned", however, this term has never actually been used in the documentation.

These auto-restart modes have long been refereed to as "force" and "normal". There's no reason to change this so I have left as is.

cylc/flow/cfgspec/globalcfg.py
oliver-sanders: This PR will continue to use "force mode" as documented:

https://cylc.github.io/cylc-doc/stable/html/plugins/main-loop/built-in/cylc.flow.main_loop.auto_restart.html#module-cylc.flow.main_loop.auto_restart

changes.d/6623.fix.md
oliver-sanders: This PR will continue to use "force mode" as documented:

https://cylc.github.io/cylc-doc/stable/html/plugins/main-loop/built-in/cylc.flow.main_loop.auto_restart.html#module-cylc.flow.main_loop.auto_restart

hjoliver:
I'm somewhat happier with that, but note that you muddied the waters here somewhat by literally using "force-condemned hosts" in the title of *this* PR!

> host-select: fix compatibility with force-condemned hosts

cylc/flow/cfgspec/globalcfg.py
hjoliver: > This PR will continue to use the longstanding "force mode" terms documented in the plugins section:

Well, it will so long as your reviewers agree to that 😜

oliver-sanders:
@hjoliver, if you see a problem with this, please raise a PR.

hjoliver:
My opinion - this is not exactly a time-critical fix (it's been broken for years and evidently no one noticed) and we often fix closely related, terminology or otherwise, when other changes directly touch them - as here.

However, it seems you're not going to budge, and this isn't important enough to keep arguing forever, so fine, I'll punt it to a follow-up PR!

hjoliver (approved):

cylc/flow/host_select.py
```diff
@@ -128,6 +128,13 @@ def select_workflow_host(cached=True):
     # be returned with the up-to-date configuration.
     global_config = glbl_cfg(cached=cached)
 
+    # condemned hosts may be suffixed with an "!" to activate "force mode"
+    blacklist = []
```
wxtim, line 132: prefer `condemn_list`?

oliver-sanders: The name `blacklist` is used elsewhere in the file, so using the same name here for consistency. To explain:

The `select_host` method below is abstract (as we may have other uses for this beyond Cylc server selection) so uses the generic name `blacklist`.

This is the concrete method for Cylc server selection. It populates the `blacklist` argument of `select_host` from condemned hosts found in the configuration.

wxtim (commented):
- [x] Run test locally
- [x] Read code

I've left one comment, which you should consider.

wxtim (approved):

PR #6647, opened 2025-03-04
Ensure `cylc message` exceptions are printed to stderr

MetRonnie:
Fixes problem reported internally.

> I had a problem... where task messaging wasn't working but there was no error reported by `cylc message`.

It was only printing errors in debug mode.

This change ensures it prints errors if they occur. But `cylc message` should not fail even if the workflow runtime client fails to initialize.

**Check List**

- [x] No dependency changes 
- [x] Tests are included 
- [x] No docs needed

### DIFF
--- a/cylc/flow/task_message.py
+++ b/cylc/flow/task_message.py
@@ -102,7 +102,9 @@ def write_messages(workflow, job_id, messages, event_time):
     _append_job_status_file(workflow, job_id, event_time, messages)
 
 
-def send_messages(workflow, job_id, messages, event_time):
+def send_messages(
+    workflow: str, job_id: str, messages: List[list], event_time: str
+) -> None:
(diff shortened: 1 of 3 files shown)


cylc/flow/task_message.py
```diff
@@ -138,7 +141,8 @@ def _append_job_status_file(workflow, job_id, event_time, messages):
     try:
         job_status_file = open(job_log_name + '.status', 'a')  # noqa: SIM115
         # TODO: niceify read/write/appending messages to this file
-    except IOError:
+    except IOError as exc:
+        print(f"{type(exc).__name__}: {exc}", file=sys.stderr)
```
oliver-sanders, line 145: Would be nice if we could cover this one too.

oliver-sanders (approved):

PR #6654, opened 2025-03-05
GH Actions: use explicit `bash` shell & other defaults

MetRonnie:
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#defaultsrunshell

>| Supported platform | `shell` parameter | Description | Command run internally |
>| --- | --- | --- | --- |
>| Linux / macOS | unspecified | The default shell on non-Windows platforms. **Note that this runs a different command to when `bash` is specified explicitly.** If `bash` is not found in the path, this is treated as `sh`. | **`bash -e {0}`** |
>| All | `bash` | The default shell on non-Windows platforms with a fallback to `sh`. When specifying a bash shell on Windows, the bash shell included with Git for Windows is used.| **`bash --noprofile --norc -eo pipefail {0}`** |

This was added to the docs in https://github.com/actions/runner/issues/1955; I was unaware that there was a difference between not specifying the default and specifying `bash`.

We seem to be missing `-o pipefail` by default. (This change will also introduce `--norc`, `--noprofile`; not sure why these are included 🤷)

> [!TIP]
> If/when we change to using setup-micromamba/miniconda, we will need to provide the full set of options we want, which will have to include `-l`. 

**Check List**

- [x] No dependency changes 
- [x] tests, docs, etc not needed

### DIFF
--- a/.github/workflows/1_create_release_pr.yml
+++ b/.github/workflows/1_create_release_pr.yml
@@ -11,6 +11,19 @@ on:
         required: false
         default: 'master'
 
+concurrency:
+  # Only let this run 1 at a time
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  FORCE_COLOR: 2
+  PIP_PROGRESS_BAR: off
+
 jobs:
   create-release-pr:
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


oliver-sanders:
Will be revoked by https://github.com/cylc/cylc-flow/pull/6655

oliver-sanders:
Should be superseded by #6655

Similar PRs are up for metomi-rose and cylc-rose.

oliver-sanders (approved):
LGTM

MetRonnie:
Build tests failing, needs https://github.com/cylc/release-actions/pull/102

MetRonnie:
MacOS unit test failing 3 times in a row with

> shell-init: error retrieving current directory: getcwd: cannot access parent directories: Undefined error: 0

but is unrelated

oliver-sanders (approved):

Issue #6630, opened 2025-02-24
sqlite3: migrate [tests] to 3.46.0+


oliver-sanders:
Double quotes are being deprecated in sqlite3.

https://sqlite.org/quirks.html#dblquote

This means we need to run through our SQL statements and switch double to single quotes and test our code against sqlite 3.46.0+.

All sqlite3 3.46.0 issues:
* https://github.com/cylc/cylc-flow/issues/6630
* https://github.com/cylc/cylc-uiserver/issues/669
* https://github.com/metomi/rose/issues/2854

hjoliver:
Double quotes are required for identifiers, single quotes for string literals.

PR #6655, opened 2025-03-05
actions: provision python from conda-forge

oliver-sanders:
* Provision Python , Bash and other deps from conda-forge.
* Divorce from the OS environment and give us more flexibility in the software versions we utilise.
* Reduce usage of `apt get` (to `at`) and remove usage `brew install` completely.
* Fix SQL statements in the functional tests to comply with changes in sqlite3 3.46.0+ (closes https://github.com/cylc/cylc-flow/issues/6630)
* Fix a functional test (wasn't failing before for unknown reasons, but the fix makes sense)
* Install subversion to ensure `svn` tests aren't skipped.
* Fix installation of the mock `mail` command to ensure mail tests aren't skipped.

**Check List**

### DIFF
--- a/.github/workflows/test_fast.yml
+++ b/.github/workflows/test_fast.yml
@@ -12,6 +12,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+defaults:
+  run:
+    shell: bash -c "exec $CONDA_PREFIX/bin/bash -elo pipefail {0}"
+
 jobs:
   test:
     runs-on: ${{ matrix.os }}
@@ -20,12 +24,10 @@ jobs:
       fail-fast: false # don't stop on first failure
       matrix:
         os: ['ubuntu-latest']
-        python-version: ['3.8', '3.10', '3.11', '3']
+        python-version: ['3.7', '3.8', '3.10', '3.11', '3.12', '3']
         include:
-          - os: 'ubuntu-22.04'
-            python-version: '3.7'
           - os: 'macos-latest'
-            python-version: '3.9'  # oldest supported version
+            python-version: '3.8'  # oldest version with arm64 builds
           # non-utc timezone test
           - os: 'ubuntu-latest'
             python-version: '3.9'  # not the oldest, not the most recent version
@@ -39,23 +41,24 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Configure Python
-        uses: actions/setup-python@v5
+      - name: Install System Dependencies
+        uses: mamba-org/setup-micromamba@v2
         with:
-          python-version: ${{ matrix.python-version }}
+          cache-environment: true
+          post-cleanup: 'all'
+          environment-name: cylc-fast-test
+          create-args: >-
+            python=${{ matrix.python-version }}
+            pip
+            bash
+            sqlite
 
-      - name: Apt-Get Install
-        if: startsWith(matrix.os, 'ubuntu')
+      - name: Install Python Dependencies
         run: |
-          sudo apt-get update
-          sudo apt-get install -y sqlite3
+          pip install -e ."[all]"
 
       - name: Patch DNS
         uses: cylc/release-actions/patch-dns@v1
-
-      - name: Install
-        run: |
-          pip install -e ."[all]"
 
       - name: Configure git  # Needed by the odd test
         uses: cylc/release-actions/configure-git@v1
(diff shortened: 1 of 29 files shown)


.github/workflows/test_functional.yml
```diff
@@ -56,34 +60,34 @@ jobs:
...
             os: 'macos-latest'
-            python-version: '3.9'
+            python-version: '3.8'  # oldest available
             test-base: 'tests/f'
             chunk: '1/5'
             platform: '_local_background*'
           - name: 'macos 2/5'
             os: 'macos-latest'
-            python-version: '3.9'
+            python-version: '3'  # newest available
```
MetRonnie, line 90: The original intention was to run the `1/4` chunk in MacOS on the oldest Python version. (However at some point this was split into `1/5` and `2/5` for faster completion.)

What is the intention behind running different chunks in different Python versions?

oliver-sanders: I thought we should probably have a test for latest Python on Mac OS as a canary. Happy to revert (unrelated really).

.github/workflows/test_functional.yml
```diff
@@ -167,6 +157,8 @@ jobs:
       - name: Configure Atrun
         if: contains(matrix.platform, '_local_at')
         run: |
+          sudo apt-get update
+          sudo apt-get install -y at
```
oliver-sanders, line 161: `at` is the one thing we need to get from the OS package manager.

.github/workflows/test_fast.yml
```diff
@@ -12,6 +12,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+defaults:
+  run:
+    shell: bash -c "exec $CONDA_PREFIX/bin/bash -elo pipefail {0}"
```
oliver-sanders, line 17: Setting `-eo pipefail` for error checking.

.github/workflows/test_fast.yml
```diff
@@ -39,24 +41,26 @@ jobs:
...
-        if: startsWith(matrix.os, 'ubuntu')
+          generate-run-shell: true
+          cache-environment: true
+          post-cleanup: 'all'
+          environment-name: cylc-fast-test
+          create-args: >-
+            python=${{ matrix.python-version }}
+            pip
+            bash
+            sqlite
```
MetRonnie, line 55: I've noticed the SVN unit tests are being skipped. Which is strange because we didn't install SVN before. Is it possible to include SVN here?

oliver-sanders: Good spotting.

This suggests that if subversion is installed, the `svn` executable is no longer in the `$PATH`.

I'm not sure if this is a consequence of the switch to micromamba (doesn't make sense) or a consequence of dropping `apt-get` installation (doesn't make sense).

oliver-sanders: We could just do this:

```suggestion
            sqlite
            subversion
```

MetRonnie:
I can't remember where you got to with the mail command, but looks like it's not working
```
[18:37:31] tests/f/events/13-task-event-mail-globalcfg.t .......... skipped: "mail" command not available
```

.github/workflows/test_fast.yml
oliver-sanders: > Ah wait, it's because it was installed on Ubuntu 22 but not latest. Is it possible to include SVN here?

I have just discovered this myself.

Let's try installing it from conda-forge on the newer images.

oliver-sanders:
> Looks like another sqlite quote issue (although test passing for some reason)

Fixed and turned the `sqlite3` command calls into `run_ok` tests to make sure they can't fail silently.

oliver-sanders:
> I can't remember where you got to with the mail command, but looks like it's not working:

The mock mail implementation is added to the `$PATH` in the `.bashrc` file. I've got the action to `cat` the `.bashrc` file so we can see what's going on here.

.github/workflows/test_fast.yml
oliver-sanders: Done, will see if it works...

oliver-sanders:
Ok, comments addressed:

* Fixed the mocked `mail` command (I think this was broken before). The mail tests are now running [e.g.](https://github.com/cylc/cylc-flow/actions/runs/13787511565/job/38561716670?pr=6655#step:14:226).
* Installed the `svn` command (this was broken by GH image changes).
* Fixed an SQL statement and updated the test to ensure the failure of this operation fails the test.

.github/workflows/test_functional.yml
```diff
@@ -128,36 +144,12 @@ jobs:
...
-      - name: Apt-Get Install
-        if: startsWith(matrix.os, 'ubuntu')
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y sqlite3 tree at
-
       - name: Add .github/bin/ to PATH
         # Sets up mocked mail command & any other custom executables
-        run: echo "${{ github.workspace }}/.github/bin" >> $GITHUB_PATH
+        # run: echo "${{ github.workspace }}/.github/bin" >> $GITHUB_PATH
```
MetRonnie, line 149:
```suggestion
        # Adding to $GITHUB_PATH does not work when using setup-micromamba and/or login shell
```

MetRonnie (approved):
Nice. Got one comment suggestion that I would rather have merged but not a blocker

oliver-sanders:
Done.

PR #6658, opened 2025-03-06
Fix reinstall not detecting changed permissions by itself

MetRonnie:
This fixes a bug where `cylc reinstall` in interactive mode would not pick up on permissions changes for files.

The reason for this is that `rsync` does not report files as included to be sent if only the permissions have been changed, unless `%i` is added to the `--out-format` option. `%i` is an itemized changes summary.

https://linux.die.net/man/1/rsync, https://man.freebsd.org/cgi/man.cgi?query=rsync

> The "%i" escape has a cryptic output that is 11 letters long. The general format is like the string YXcstpoguax, where Y is replaced by the type of update being done, X is replaced by the file-type, and the other letters represent attributes that may be output if they are being modified.

I have hidden this itemized changes summary unless using `cylc reinstall -v` or even higher verbosity. 

**Check List**

- [x] No dependency changes 
- [x] Tests are included
- [x] No docs needed


cylc/flow/install.py
```diff
@@ -213,19 +213,12 @@ def reinstall_workflow(
...
-
     # Run rsync command:
     reinstall_log.info(cli_format(rsync_cmd))
     LOG.debug(cli_format(rsync_cmd))
     proc = Popen(rsync_cmd, stdout=PIPE, stderr=PIPE, text=True)  # nosec
     # * command is constructed via internal interface
-    stdout, stderr = proc.communicate()
-
-    # Strip unwanted output.
-    stdout = ('\n'.join(re.findall(r'\+\+\+(.*)\+\+\+', stdout))).strip()
```
MetRonnie, line 227: This seemed to be duplicating the work of `format_rsync_output()`, but removing all `cannot delete non-empty directory:` lines, not just the `opt` dir

cylc/flow/scripts/reinstall.py
```diff
@@ -331,20 +339,23 @@ def format_rsync_out(out: str) -> List[str]:
...
+        if match:
+            summary, operation, file = match.groups()
+            color = 'green' if operation == 'send' else 'red'
+            formatted_line = f"<{color}>{operation} {file}</{color}>"
+            if cylc.flow.flags.verbosity > 0:
+                formatted_line = f"<{DIM}>{summary}</{DIM}> {formatted_line}"
+            lines.append(cparse(formatted_line))
         else:
-            # other uncategorised log line
+            # shouldn't happen; tolerate unknown rsync implementation?
```
wxtim, line 358: If you don't think it should happen, then perhaps a message warning that this is an unexpected rsync implementation, please contact the dev team might be appropriate.

cylc/flow/scripts/reinstall.py
```diff
@@ -331,20 +339,23 @@ def format_rsync_out(out: str) -> List[str]:
...
+        if not line or line.startswith('cannot delete non-empty directory:'):
+            # This can arise as a result of deleting a subdir when we are
+            # excluding files within the subdir.
+            # This is likely to occur for the opt dir when a rose-suite.conf
             # file is present.
+            # Skip this line as nothing will happen to this dir.
             continue
+        match = re.match(r'^(.{11}) (send|del\.) (.*)$', line)
+        if match:
+            summary, operation, file = match.groups()
```
wxtim, line 351: [should not block/nice to have]
If nothing has changed except permissions can get this info from the rsync output and pass it on to the user?

wxtim (approved):
- [x] Tested manually
- [x] Read
- [x] Made 2 non-blocking suggestions

cylc/flow/install.py
oliver-sanders: The point of the `+++` prefix/suffix will have been to separate intended `rsync` output from any unexpected "junk" that might be output by the command / shell, ensuring that we only display what we need, even if the output differs from expectation in other ways.

cylc/flow/remote.py
```diff
@@ -180,9 +180,14 @@ def get_includes_to_rsync(rsync_includes=None):
 DEFAULT_RSYNC_OPTS = [
     '-a',
     '--checksum',
-    '--out-format=%o %n%L',
+    '--out-format=%i %o %n%L',  # see comment below
```
oliver-sanders, line 183: Do we want to include `%i` in the logs?

* If this is only really relevant to the dry-run functionality, then continue to override the CLI argument as we did before.
* If we think we should have this in the logs, then to keep them readable it might be better to keep the operation at the start of the line?

cylc/flow/install.py
MetRonnie: From the context I could gather in https://github.com/cylc/cylc-flow/pull/5125, the only unwanted output that was being targeted was `cannot delete non-empty directory:`.

cylc/flow/remote.py
MetRonnie: The rsync output isn't included in the reinstall logs. The only thing we do with the rsync output is print it to stdout in the dry-run case.

cylc/flow/install.py
oliver-sanders: This `rsync` command doesn't just apply to the dry-run check (which is passed through`format_rsync_output`). And this prefix stripped out other potential output.

cylc/flow/remote.py
oliver-sanders: I think it is, install a workflow, make changes, then check `~/cylc-run/<workflow>/log/install/*`.

MetRonnie: Ah ok - it's only included in the `01-install.log`. So I should make this option with `%i` only apply to reinstall

cylc/flow/install.py
MetRonnie: https://github.com/cylc/cylc-flow/blob/48d3d3898fd06c0f25c3214b220389a336c01fc3/cylc/flow/install.py#L226-L227
is at odds with
https://github.com/cylc/cylc-flow/blob/48d3d3898fd06c0f25c3214b220389a336c01fc3/cylc/flow/scripts/reinstall.py#L346-L348

The former seems to have been implemented to cover `cannot delete non-empty directory:`, there was no mention of any other potential rsync output in #5125. And the latter seems to have been implemented to deliberately include other potential rsync output in case the rsync implementation results in non-standard output format.

Hence I am not sure we should be stripping other potential rsync output.

oliver-sanders: It's not at odds.

* The `+++` prefix ensures that only the intended output lines get through.
* The `else` handles output lines that don't match `send` or `del.`.

cylc/flow/scripts/reinstall.py
MetRonnie: Although the itemized changes summary should be consistent, I am wary of doing this as the docstring of this function says to keep it simple.

wxtim (approved):
- [x] Checked changes since last review
- [x] Deconflicted
- [x] Lint check failures unrelated

cylc/flow/scripts/reinstall.py
oliver-sanders: There are multiple potential causes of this:

* Alternative `rsync` implementation (e.g. BSD rsync).
* Newer/older version of GNU `rsync`
* Other misc stdout created by `ssh` or `bash`.

We just need to keep our scripts as robust to output changes as possible. No need to bother the user about it.

oliver-sanders (approved):

cylc/flow/remote.py
oliver-sanders: Will tackle in follow-up.

cylc/flow/install.py
oliver-sanders: https://github.com/cylc/cylc-flow/pull/6954

PR #6661, opened 2025-03-11
Refine documentation of Validators

### LINKED ISSUE #6288: Documentation for task names is slightly incorrect
### Description

The [cylc.flow.unicode_rules.TaskNameValidator](https://cylc.github.io/cylc-doc/stable/html/_modules/cylc/flow/unicode_rules.html#TaskNameValidator) documentation states that workflows must start with an alphanumeric character, and contain only alphanumeric, -, +, %, and @ characters. However it can also contain `_` characters, which is not documented.

Specifically this comes from using the `\w` regex class, which:
> Matches Unicode word characters; this includes all Unicode alphanumeric characters (as defined by [str.isalnum()](https://docs.python.org/3/library/stdtypes.html#str.isalnum)), as well as the underscore (_).

This class also contains many characters outside the possibly expected `[a-zA-Z0-9_]`, such as `⅓` (U+2153 Vulgar Fraction One Third), though these are likely less frequently used in workflows than underscores.

```py
>>> "⅓".isalnum()
True
>>> "日本".isalnum()
True

```

### Expected Behaviour

Underscores should be contained in the documentation, probably also that other characters are supported, maybe by linking to the python isalnum str method. At least it should check that everyone is properly supporting Unicode.

wxtim:
Closes #6288 

**Check List**

- [x] Is purely a documentation change and doesn't require tests.
- [x] Has been tested by building the documentation.

### DIFF
--- a/cylc/flow/unicode_rules.py
+++ b/cylc/flow/unicode_rules.py
@@ -45,7 +45,7 @@ def regex_chars_to_text(chars):
         >>> regex_chars_to_text([r'\-', r'\.', r'\/'])
         ['``-``', '``.``', '``/``']
         >>> regex_chars_to_text([r'\w'])
-        ['alphanumeric (regex word characters - ``\w``)']
+        ['alphanumeric (regex word characters - ``\\w``)']
         >>> regex_chars_to_text(['not_in_map'])
         ['``not_in_map``']
 
--- a/cylc/flow/unicode_rules.py
+++ b/cylc/flow/unicode_rules.py
@@ -27,7 +27,7 @@ from cylc.flow.task_qualifiers import TASK_QUALIFIERS
 from cylc.flow.task_state import TASK_STATUSES_ORDERED
 
 ENGLISH_REGEX_MAP = {
-    r'\w': 'alphanumeric (regex word characters - ``\w``),
+    r'\w': 'alphanumeric (regex word characters - ``\w``'),
     r'a-zA-Z0-9': 'latin letters and numbers',
     r'\d': 'numbers',
     r'\-': '``-``',
--- a/cylc/flow/unicode_rules.py
+++ b/cylc/flow/unicode_rules.py
@@ -27,7 +27,7 @@ from cylc.flow.task_qualifiers import TASK_QUALIFIERS
 from cylc.flow.task_state import TASK_STATUSES_ORDERED
 
 ENGLISH_REGEX_MAP = {
-    r'\w': 'alphanumeric',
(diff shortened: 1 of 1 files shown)


cylc/flow/unicode_rules.py
```diff
@@ -27,7 +27,7 @@
 from cylc.flow.task_state import TASK_STATUSES_ORDERED
 
 ENGLISH_REGEX_MAP = {
-    r'\w': 'alphanumeric',
+    r'\w': 'alphanumeric (defined by ``str.isalnum()``) or ``_``',
```
oliver-sanders, line 30: Technically correct, but a tad cryptic!

Not sure how to communicate this clearly, perhaps:

```
alphanumeric (regex word characters - `\w+`)
```

cylc/flow/unicode_rules.py
```diff
@@ -45,7 +45,7 @@ def regex_chars_to_text(chars):
         >>> regex_chars_to_text([r'\-', r'\.', r'\/'])
         ['``-``', '``.``', '``/``']
         >>> regex_chars_to_text([r'\w'])
-        ['alphanumeric']
+        ['alphanumeric (defined by ``str.isalnum()``) or ``_``']
```
oliver-sanders, line 48:
```suggestion
        ['alphanumeric (regex word characters - ``\w``)']
```

cylc/flow/unicode_rules.py
```diff
@@ -27,7 +27,7 @@
 from cylc.flow.task_state import TASK_STATUSES_ORDERED
 
 ENGLISH_REGEX_MAP = {
-    r'\w': 'alphanumeric',
+    r'\w': 'alphanumeric (regex word characters - ``\w``),
```
oliver-sanders, line 30:
```suggestion
    r'\w': 'alphanumeric (regex word characters - ``\w``'),
```

cylc/flow/unicode_rules.py
```diff
@@ -45,7 +45,7 @@ def regex_chars_to_text(chars):
         >>> regex_chars_to_text([r'\-', r'\.', r'\/'])
         ['``-``', '``.``', '``/``']
         >>> regex_chars_to_text([r'\w'])
-        ['alphanumeric']
+        ['alphanumeric (regex word characters - ``\w``)']
```
oliver-sanders, line 48:
```suggestion
        ['alphanumeric (regex word characters - ``\\w``)']
```

hjoliver (approved):

PR #6670, opened 2025-03-13
Fix flaky test

oliver-sanders:
I think the flaky test is caused by some stderr that it output from the `.bash_profile` file. This adds `--noprofile` to suppress the problem.

I initially though I would need monkeypatch so converted to pytest, kept the change anyway. I've got a crude script to do this, should get around to finishing off the last unittests at some point...

**Check List**

### DIFF
--- a/tests/unit/test_subprocpool.py
+++ b/tests/unit/test_subprocpool.py
@@ -14,16 +14,16 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from pathlib import Path
+from types import SimpleNamespace
 from tempfile import (
     NamedTemporaryFile,
     SpooledTemporaryFile,
     TemporaryFile,
     TemporaryDirectory,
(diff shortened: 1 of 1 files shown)


tests/unit/test_subprocpool.py
```diff
@@ -79,7 +79,10 @@ def test_run_command_writes_to_out():
 
 def test_run_command_writes_to_err():
     """Test basic usage, command writes to STDERR"""
-    ctx = SubProcContext('parrot2', ['bash', '-c', 'echo pirate errrr >&2'])
+    ctx = SubProcContext(
+        'parrot2',
+        ['bash', '-noprofile', '-c', 'echo pirate errrr >&2']
```
oliver-sanders, line 84:
```suggestion
        ['bash', '--noprofile', '-c', 'echo pirate errrr >&2']
```

wxtim (approved):
Looks good. Just re-running CI to ensure lack of flake isn't a fluke.

Issue #6686, opened 2025-03-27
Delayed poll result contaminate the datastore.


hjoliver:
(Symptoms reported by @ColemanTom on the forum: https://cylc.discourse.group/t/submitted-jobs-cant-be-polled/1146)

Delayed poll results that would incorrectly revert the task state are correctly ignored by the scheduler (i.e. they don't affect the task pool) but they seem to get into the datastore. This can result in a task in the UI in the wrong state that is not fixable by polling - there's no longer an active task in the `n=0` window so the scheduler refuses to poll it.

hjoliver:
Diagnosis: I think the job poll callback holds a reference to the task proxy, and when the delayed poll result is received we don't actually know that the task has completed and left the task pool.

hjoliver:
I think I have a fix.

Issue #6412, opened 2024-08-01
`cylc set` on task no longer in the pool can cause task state regressions


ColemanTom:
### Description

Doing `cylc set` in order (at least of) `succeeded, failed, succeeded` gives incorrect symbol colouring. The status of outputs appears correct in cylc show, but incorrect (maybe?) in UI.

![ui_bug_set_not_update](https://github.com/user-attachments/assets/728a5962-a8ff-493f-882b-5d95a77d78a3)

### Reproducible Example

NOTE: Look at UI symbols in between each `cylc set` command, and on CLI to see the inconsistency.

1. Run any workflow
2. `cylc set workflow//cycle/task` (circle is fully coloured in)
3. `cylc set -o failed workflow//cycle/task` (circle now has a cross in it)
4. `cylc set workflow//cycle/task` (circle has no symbol or colouring)
5. `cylc set workflow//cycle/task` (circle still has no symbol or colouring)

### Expected Behaviour

I think it should be showing as fully coloured in. However, I recognise that it is perhaps in a confused state.

oliver-sanders:
I think the root cause of this issue is the "transient task proxy" objects that are used by task pool in some cases.

When we create a "transient" task proxy, we extract any previously completed outputs from the DB, however, we do not retrieve the previous task status, instead defaulting to "waiting".

oliver-sanders:
Taking a look at the loading of the outputs:

https://github.com/cylc/cylc-flow/blob/62fb15b961b06639000c9a10f8453f8358961298/cylc/flow/task_pool.py#L1634-L1635

@hjoliver, shouldn't that function be filtering entries by flow number(s)?

hjoliver:
> I think the root cause of this issue is the "transient task proxy" objects that are used by task pool in some cases.

Yeah I think you're right.

wxtim:
> I think the root cause of this issue is the "transient task proxy" objects that are used by task pool in some cases.
> 
> When we create a "transient" task proxy, we extract any previously completed outputs from the DB, however, we do not retrieve the previous task status, instead defaulting to "waiting".

Having investigated #6613 and come to the same conclusion - the transient task proxy seems a bit weird. 

I think that there are a bunch of ways of tackling this, so in order of most radical -> most computationally expensive

* a) Junk the concept of transient task proxies - simply don't allow people to use set on inactive tasks (probably reasonable, but a big change). I'm unclear whether this might cause problems with expiry.
[...]

Issue #4513, opened 2021-11-15
Polling can incorrectly return a failed task to the running state


dpmatthews:
If a task fails due to exceeding the wallclock limit it can return to the running state if it is polled before it has exited the queue.
The following workflow illustrates the problem:
```
[scheduling]
    [[graph]]
        R1 = "timeout:started => poll"
[runtime]
    [[timeout]]
        script = "sleep 600"
        err-script = "sleep 180"
        platform = remote-hpc
        execution time limit = PT30S
    [[poll]]
        # Poll the task after it has failed but before it exits the queue
        script = "sleep 50; cylc poll $CYLC_SUITE_NAME timeout.1"
```
(note: `err-script` keeps the job in the queue until the workflow manager kills it)
I performed this test with our HPC using PBS but the same result should occur with, e.g Slurm.

Relevant log messages:
```
2021-11-15T19:15:37Z CRITICAL - [timeout.1 running job:01 flows:1] (received)failed/TERM at 2021-11-15T19:15:36Z
2021-11-15T19:15:37Z INFO - [timeout.1 running job:01 flows:1] => failed
2021-11-15T19:15:46Z DEBUG - [jobs-poll cmd] ssh -oBatchMode=yes -oConnectTimeout=8 -oStrictHostKeyChecking=no hpc env CYLC_VERSION=8.0b3.dev CYLC_ENV_NAME=cylc-8.0b3.dev bash --login -c ''"'"'exec "$0" "$@"'"'"'' cylc jobs-poll --debug -- '$HOME/cylc-run/bug.polling3.c8/run8/log/job' 1/timeout/01
    [jobs-poll ret_code] 0
    [jobs-poll out] [TASK JOB SUMMARY]2021-11-15T19:15:46Z|1/timeout/01|{"job_runner_name": "pbs", "job_id": "5698928", "job_runner_exit_polled": 0, "run_status": 1, "run_signal": "TERM", "time_submit_exit": "2021-11-15T19:14:34Z", "time_run": "2021-11-15T19:14:42Z", "time_run_exit": "2021-11-15T19:15:36Z"}
2021-11-15T19:15:46Z INFO - [timeout.1 failed job:01 flows:1] (polled)started at 2021-11-15T19:14:42Z
2021-11-15T19:15:46Z INFO - [timeout.1 failed job:01 flows:1] => running
```
This is clearly a bug.

I think this is the relevant code which causes the problem:
[...]

PR #6687, opened 2025-03-27
Mark tasks as transient on removal from task pool.

hjoliver:
Bug #6686 is caused by job polling callback, with a delayed poll result, processing tasks that have already left the task pool as completed.

We already have a task proxy flag whose exact purpose is to indicate "not task pool": `TaskProxy.transient`.

So the easiest way to fix this bug seems to be:
 1. set the transient flag on a task proxy when it is removed from the pool
 2. don't process task messages if the target task is marked as transient

**Check List**


hjoliver:
(If test failures are meaningful, might have to consider a different approach such as checking task completion before processing messages...).

hjoliver:
(OK the initial two-line "fix" is a bit too simplistic - turns out we are calling `process_message` artificially, with a transient task proxy, to spawn on the outputs of inactive tasks targeted by `cylc set`).

oliver-sanders:
Is there any relationship to https://github.com/cylc/cylc-flow/issues/6412?

ColemanTom:
Does this relate to https://github.com/cylc/cylc-flow/issues/4513 ?

oliver-sanders:
> Does this relate to https://github.com/cylc/cylc-flow/issues/4513 ?

That issue predates "transient" task proxies.

hjoliver:
For reference, "transient task proxies" explained here: https://github.com/cylc/cylc-flow/issues/6412#issuecomment-2746488195

dwsutherland (approved):

hjoliver:
(Yes, superseded).

PR #6692, opened 2025-03-31
flake8++

oliver-sanders:
Fix violations of recently added F824.

The `global` and `nonlocal` statements allow variables to be rebound from within a nested scope (more info: [1](https://stackoverflow.com/a/1261961), [2](https://docs.python.org/3/reference/simple_stmts.html#the-nonlocal-statement)).

The overwhelming majority of these statements were added by myself. It's a once bitten, always cautious thing:

```python
x = 0
y = []

def foo():
   x = 1  # this rebinds x (so does nothing without `global` / `nonlocal`)

def bar():
   y.append(1)  # this adds to x (but does not rebind it)

>>> foo()
>>> x
0   # <= with `global` / `nonlocal`, this would be 1

>>> bar()
y
[1]
```

Because these two examples behave differently, I always used `nonlocal` to prevent the potential for confusion. I also thought that it was good to always indicate where a variable is being "borrowed" from a nonlocal scope due to the implications of this (be careful with this variable, it is shared with other scopes).

However, I'm happy to be overruled on it. If `flake8` has swung this way (don't use `nonlocal` / `global` unless you are actually rebinding the variable(s)), I'm happy to follow.

**Check List**

### DIFF
--- a/tests/integration/tui/test_logs.py
+++ b/tests/integration/tui/test_logs.py
@@ -101,7 +101,7 @@ def wait_log_loaded(monkeypatch):
             delay: The delay between retries.
 
         """
-        nonlocal before, count
+        nonlocal before
         for _try in range(tries):
             if count > before:
                 await asyncio.sleep(0)
--- a/tests/unit/main_loop/test_main_loop.py
+++ b/tests/unit/main_loop/test_main_loop.py
@@ -72,9 +72,9 @@ def test_wrapper_calls_function():
     flag = False
 
     async def test_coro(arg1, arg2):
+        nonlocal flag
         assert arg1 == 'arg1'
         assert arg2 == 'arg2'
-        nonlocal flag
         flag = True
 
     coro = _wrapper(
@@ -152,7 +152,6 @@ def basic_plugins():
     calls = []
 
     def capture(*args):
-        nonlocal calls
         calls.append(args)
 
     plugins = {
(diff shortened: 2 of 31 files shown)


wxtim (approved):
- [x] Read, understood reasoning
- [x] Checked functioning of `cylc tui` overlays, since that is the most changed file.
- [x] Run flake 8 locally.
- [ ] Wait for all tests to pass before hitting merge.

oliver-sanders:
Force pushed change to also cover the tests (not included in flake8 on this branch)

tests/integration/tui/test_logs.py
```diff
@@ -101,7 +101,7 @@ async def _wait_log_loaded(tries: int = 25, delay: float = 0.1):
             delay: The delay between retries.
 
         """
-        nonlocal before, count
+        nonlocal before
```
oliver-sanders, line 104: This `nonlocal before` is needed.

Warning, don't remove `nonlocal` or `global` statements unless flake8 tells you to, it can break things.

tests/unit/main_loop/test_main_loop.py
```diff
@@ -72,9 +72,9 @@ def test_wrapper_calls_function():
     flag = False
 
     async def test_coro(arg1, arg2):
+        nonlocal flag
         assert arg1 == 'arg1'
         assert arg2 == 'arg2'
-        nonlocal flag
         flag = True
```
oliver-sanders, line 78: Another example where `nonlocal` is required. Because the variable is a primitive, not an object, we have to rebind it rather than modify it in place.

PR #6671, opened 2025-03-14
xtrigger: ui support

oliver-sanders:
This exposes xtriggers in Tui and lays the groundwork for exposing them in the GUI.

* Closes #6742 
* Unblocks https://github.com/cylc/cylc-ui/issues/331
* Unblocks https://github.com/cylc/cylc-ui/issues/2103
* Partially addresses https://github.com/cylc/cylc-flow/issues/6307

Visibility of xtriggers in the GUI (retry triggers in particular) has been flagged as a high priority development for production use cases and a desirable feature in general.

#### xtriggers: fix update and add is_retry, is_wallclock and is_xtriggered attrs

* Don't update xtriggers incrementally:
  * Xtriggers were being updated in the data store incrementally.
  * This doesn't work with GraphQL subscriptions (see https://github.com/cylc/cylc-flow/issues/6307).
  * This turns off incremental updates ensuring that all xtriggers are included in all updates.
  * Unblocks https://github.com/cylc/cylc-ui/issues/2103
* Add is_retry, is_wallclock and is_xtriggered task attributes.
  * We would like to be able to incorporate xtriggers into the task icons used in the GUI and Tui.
  * These attributes allow clients to access the required information without having to subscribe to and monitor all xtriggers.
  * Unblocks https://github.com/cylc/cylc-ui/issues/331

#### tui: add retry, wallclock and xtrigger task modifiers

* Display information about a task's xtriggers as part of the task icon using a modifier.
* Incorporate a text-based summary of these modifiers into the task menu.
* Change the runahead task modifier icon to match the placement of theother modifiers.

Screenshot from 2025-03-14 16-33-10

Screenshot from 2025-03-14 16-33-26

#### TODO:
* [x] Confirm viability of selected Tui icons in various terminals.
  * [x] linux/xterm (good)
  * [x] linux/terminator (good)
  * [x] chromium/vscode (task alignment good, modifier alignment differs, ok IMO)
  * [x] browser/jupyterlab (task alignment good, modifier alignment differs, ok IMO)
  * [x] macos/terminal.app (good)
* [ ] Settle the names and descriptions of the is_retry, is_wallclock and is_xtriggered attributes (if folk would like to debate them).
* [x] Update Tui integration tests.
* [x] Back-compat support for Tui (detect workflow version and modify the query).
* [x] ~Back-compat support for GUI (requires a cylc-uiserver shim).~ would appear not to be necessary https://github.com/cylc/cylc-flow/pull/6671#issuecomment-2758849647

**Check List**

[...]

PR #6695, opened 2025-04-02
set command: manually satisfy dependence on xtriggers

### LINKED ISSUE #3221: Allow xtriggers to be reset
We need to:
- [ ] allow xtriggers to be reset (forced to be satisfied or unsatisfied)
- [ ] document the current workaround in the meantime

REF: https://cylc.discourse.group/t/can-i-reset-an-xtrigger/87/2?u=hilary.j.oliver

hjoliver:
- partially address #3221
- needed for group trigger #6395
- see also #6725 and #6741 

Satisfy ~~or unsatisfy~~ xtrigger (including clock trigger) prerequisites in target tasks.
- this spawns the target task into `n=0` with the xtrigger satisfied
- results (set and unset) persist via storing xtrigger prerequisites in the DB task prerequisites table,
- if no other tasks depend on the xtrigger, the scheduler stops calling it

[UPDATE 28 April] punted unsatisfy to #6741
 
To avoid changing the `cylc set` API I've leveraged the existing prerequisite setting syntax: 

-----

Note I've promoted basic xtrigger logging to INFO.  [UPDATE]: now just logs initiation of xtrigger checking.

-----

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/820


oliver-sanders:
I'm happy with the `xtrigger` cycle point, makes sense.

Ensuring that the consequence of these xtrigger set operations are visible in Cylc interfaces is going to be important here. E.g, if you set an xtrigger in an n>0 task, we would want the data store to reflect that so the user can see the consequences of their action (e.g. in `cylc show`). Note xtriggers will be added to the GUI "Info view" shortly.

You might want to consider rolling this in with something along the lines of: https://github.com/cylc/cylc-flow/pull/6560

hjoliver:
Drafting this for the moment - need to distinguish between satisfying an xtrigger (which multiple tasks might depend on) and satisfying a task's dependence on an xtrigger.  (C.f. setting outputs vs setting prerequisites, for task triggers).

hjoliver:
Undrafting now - tentatively, I think it's all working and I only need to up the test coverage.

hjoliver:
cylc-doc TODO:
- mention this in the xtriggers section
- (and document as part of group trigger, via #6395 )

hjoliver:
@wxtim - I've tentatively pinged you as reviewer, given your related xtrigger work atm.

wxtim:
> Note I've promoted basic xtrigger logging to INFO. On master it's completely hidden, which IMO is not good (unless perhaps you're calling a lot of xtriggers very frequently). It's important to be able to see what xtriggers are currently being checked, and how often. It also helps elucidate the cycle point specificity (or not) of xtriggers, because the function signature is logged.

Possibly a future enhancement, rather than a "thing" for this PR - but might we consider grouping xtrigger reports in the way we do emails?

oliver-sanders:
> It's important to be able to see what xtriggers are currently being checked, and how often.

IMO, we really don't want to log every xtrigger function call at the INFO level. Not really helpful information.

cylc/flow/xtrigger_mgr.py
```diff
@@ -770,11 +769,64 @@ def callback(self, ctx: 'SubFuncContext'):
 
         LOG.debug('%s: returned %s', sig, results)
         if not satisfied:
+            LOG.info('xtrigger not satisfied: %s = %s', ctx.label, sig)
```
wxtim, line 772: This logging would be more helpful to users if they could see which tasks are depending on this xtrigger.

cylc/flow/xtrigger_mgr.py
```diff
@@ -577,6 +574,9 @@ def load_xtrigger_for_restart(self, row_idx: int, row: Tuple[str, str]):
         # Tell the datastore this xtrigger is satisfied.
         self.data_store_mgr.delta_task_xtrigger(sig, True)
 
+        # Tell the datastore this xtrigger is satisfied.
+        self.data_store_mgr.delta_xtrigger(sig, True)
+
```
wxtim, line 579: This looks a bit odd: The two methods look like duplicates, but on examination

- They do different things - the comments probably need to explain this.
- `delta_task_xtrigger` has args `itask, label, sig, satisfied`.

cylc/flow/xtrigger_mgr.py
oliver-sanders: That could be a very long list.

I don't think we should be logging this at the INFO level at all. The number of unsuccessful xtrigger polls is not of interest to the user or developer.

cylc/flow/task_pool.py
```diff
@@ -1871,12 +1921,24 @@ def _get_task_proxy_db_outputs(
         self._load_historical_outputs(itask)
         return itask
 
-    def _standardise_prereqs(self, prereqs: 'List[str]') -> 'Set[PrereqTuple]':
-        """Convert trigger prerequisites to task messages."""
+    def _standardise_prereqs(self, prereqs: 'Iterable[str]') -> 'Set[Tokens]':
```
wxtim, line 1924: Can be punted, but might be easier to include now?

I'm almost certain that at some point someone will want `--set=xtrigger/all` to be available.

cylc/flow/task_pool.py
```diff
@@ -1997,13 +2061,14 @@ def set_prereqs_and_outputs(
                 continue
 
             if prereqs:
-                set_all, valid_prereqs = (
+                set_all, valid_prereqs, valid_xtrigs = (
```
wxtim, line 2064: You will almost certainly need to review the docstring for `set_prereqs_and_outputs`.

cylc/flow/xtrigger_mgr.py
oliver-sanders: We could, but then what would be the point of doing it in the first place.

And again, I really don't think we should log this (at INFO level) at all.

cylc/flow/task_pool.py
```diff
@@ -119,6 +121,51 @@
...
+
+    Args:
+        prereqs: prerequisites and xtriggers in string form
+            xtriggers format "xtrigger/label:state"
+
+    Returns: {label: (succeeded, set-output)}
+        (set-output means set the xtrigger output for all dependent tasks)
+
+    Examples:
+        >>> _standardise_xtrigs(["1/foo:started"])
```
wxtim, line 144: MIght be nice to see some examples using other types of iterable, just to be on the safe side?

dwsutherland (approved):
LGTM
image

cylc/flow/data_store_mgr.py
```diff
@@ -2525,29 +2521,53 @@ def delta_task_ext_trigger(
         ext_trigger.time = update_time
         self.updates_pending = True
 
-    def delta_task_xtrigger(self, sig, satisfied):
-        """Create delta for change in task proxy xtrigger.
+    def _delta_xtrigger(
```
wxtim, line 2524: This will conflict with #6671 ?

cylc/flow/scripts/set.py
```diff
@@ -18,41 +18,59 @@
...
-For custom outputs, use the output names not the associated task messages:
-[runtime]
-  [[my-task]]
-    # ...
-    [[[outputs]]]
-      # <output-name> = <task-message>
-      x = "file x completed and archived"
+Command to manually set task prerequisites and outputs, and xtriggers.
+
+By default, this sets all required outputs plus the "submitted", "started",
```
wxtim, line 23: Probably worth being super clear that you can't set outputs for an xtrigger?

```suggestion
By default, this sets all required outputs of a task plus the "submitted", "started",
```

cylc/flow/data_store_mgr.py
oliver-sanders: We've got a lot of xtrigger stuff going on ATM, I would like to get 6671 reviewed and in soon (especially as it unblocks working on xtriggers in the GUI which is being raised as a high requirement)

cylc/flow/xtrigger_mgr.py
hjoliver: Ah yeah, sorry about that - it's due to a typo in my latest change, had it working before that.

cylc/flow/data_store_mgr.py
hjoliver: The trouble is, the existing method (which you've highlighted here) is for recording satisfied xtriggers in the DB, regardless of which tasks depend on them. Whereas the new method I've added here is to record a specific task's dependence on an xtrigger - so the old name is more appropriate for the new method. I can revert this and come up with a new name, supported by explanatory docstrings, if this change causes more trouble than it's worth?

cylc/flow/task_pool.py
hjoliver: Yes, good point.

cylc/flow/task_pool.py
hjoliver: OK I converted one to a set.

cylc/flow/scripts/set.py
hjoliver: Done.

hjoliver:
@wxtim and @oliver-sanders - short on time today (local HPC and user issues) BUT:
- I've addressed a bunch of @wxtim's comments
- I've added support for `--pre=xtrigger/all` (suggestion from @wxtim )
- I've changed the xtrigger "states" (for the CLI args) from ":succeeded" and ":waiting" to ":satisfied" and ":unsatisfied" (attempting to reuse task statuses for this was a bit lazy, and potentially confusing)
- I agree with comments on logging to INFO being too much. It annoyed me that nothing at all was logged to INFO, but just logging commencement of xtrigger checking would be best.

So: TODO on this branch tomorrow:
- extend tests a little to cover `--pre=xtrigger/all`
- reduce logging as just described

hjoliver:
Xtrigger logging rationalized:

Untitled

hjoliver:
(Drafting - made a bunch of changes that undoubtedly require some tidying and may break tests).

tests/unit/test_prerequisite.py
```diff
@@ -199,10 +199,9 @@ def test_satisfy_me():
     assert not prereq.is_satisfied()
     assert prereq._cached_satisfied is False
 
-    valid = prereq.satisfy_me(
+    prereq.satisfy_me(
         [Tokens('//1/a:x'), Tokens('//1/d:x'), Tokens('//1/c:y')],
     )
-    assert {detok(tokens) for tokens in valid} == {'1/a:x'}
```
hjoliver, line 205: (We were not using the return value of this method anywhere, so I ditched it).

cylc/flow/data_store_mgr.py
```diff
@@ -2525,29 +2521,53 @@ def delta_task_ext_trigger(
...
+        """Helper for the two xtrigger delta methods."""
+        tp_delta = self.updated[TASK_PROXIES].setdefault(
+            tp_id, PbTaskProxy(id=tp_id))
+        tp_delta.stamp = f'{tp_id}@{t_update}'
+        xtrigger = tp_delta.xtriggers[sig]
+        xtrigger.id = sig
+        xtrigger.label = label
+        xtrigger.satisfied = satisfied
+        xtrigger.time = t_update
+        self.updates_pending = True
```
oliver-sanders, line 2537: FYI, this does not work with subscriptions ATM as a result of https://github.com/cylc/cylc-flow/issues/6307

This issue is resolved on #6671, but requires a change of approach because whenever one xtrigger changes state, we need to list all xtriggers.

wxtim:
I've updated my test branch https://github.com/hjoliver/cylc-flow/pull/65

oliver-sanders:
> I've changed the xtrigger "states" (for the CLI args) from ":succeeded" and ":waiting" to ":satisfied" and ":unsatisfied" (attempting to reuse task statuses for this was a bit lazy, and potentially confusing because only a subset of states is relevant and the meaning is kinda different - we can't unsatisfy task prerequisites)

Dammit!

I get why you have done this, but I actually think the task state was a better option.

I had been planning on looking into dragging xtriggers into the task state model in the future. This is mostly off topic (shift any convo onto Element?), but very quickly, here's why I'm leaning towards task states:

* "Satisfied" / "unsatisfied" is too binary, we also need a "not yet evaluated" state, else the GUI gives incorrect information.
  * See the bug report - https://github.com/cylc/cylc-flow/issues/6298
  * And POC PR (also linked above as the third state is relevant with this PR) - https://github.com/cylc/cylc-flow/pull/6560
  * Using the task state makes this problem go away:
    * waiting - xtrigger has not yet been evaluated.
    * running - the xtrigger manager is actively polling this function at the configured interval.
    * succeeded - the xtrigger has returned a success status.
* We have the plan to allow async xtrigger functions soon.
  * The current xtrigger implementation has severe efficiency issues. After a few xtrigger server load issues, we have had to advise against their use 😢.
  * So this issue is higher up our priority ladder than it would be for functionality sake alone.
  * Async xtrigger functions will map cleanly onto task states:
    * waiting - xtrigger has not yet started running.
    * running - xtrigger async function has started.
    * succeeded - xtrigger async function has returned successfully.
    * failed - xtrigger async function has errored (i.e. crashed).
    * expired?
* I have an undocumented plan to utilize the run_mode to implement async xtrigger functions (and more).
  * This will make them proper tasks, just with a different execution mode (i.e. async function in a dedicated process).
  * There are lots of benefits to this, highlights:
    * Code reuse - It's easier to implement it this way because it follows the simulation / skip / dummy mode code abstraction, there would be no need for an async xtrigger manager.
    * Xtrigger dependencies & use in conditional expressions - Will work with no effort.
    * Retries - Xtrigger retries can be configured like task retries with no effort.
    * Broadcasts - Xtriggers can be configured on the fly just like tasks with no effort.
    * Etc.
  * However, this would *require* xtriggers to fit in with the task state model.

This PR has now been extended to unsetting xtriggers? This isn't actually necessary right?

cylc/flow/data_store_mgr.py
oliver-sanders: Ok, I have now re-implemented my xtrigger branch (https://github.com/cylc/cylc-flow/pull/6671) in a way that is compatible with your changes here.

It just needs your new `delta_task_xtrigger` bit jammed in around here:

https://github.com/cylc/cylc-flow/pull/6671/files#diff-aaaccca49464232ef4889b5092201eb44575e83602d68b38f9608755ae90982aR2561

cylc/flow/task_pool.py
```diff
@@ -2043,37 +2106,69 @@ def set_prereqs_and_outputs(
...
+        for prereq in req_pre - valid_pre:
             # But log bad ones with triggers, not messages.
             trg = self.config.get_taskdef(
-                prereq.task
+                str(prereq.task)
             ).get_output(prereq.output)
             LOG.warning(
                 f'{point}/{tdef.name} does not depend on '
-                f'"{prereq.get_id()}:{trg}"'
+                f'"{prereq.point}/{prereq.task}:{trg}"'
```
oliver-sanders, line 2141: Plz run this through the detokeniser.

cylc/flow/scripts/set.py
```diff
@@ -72,15 +90,17 @@
 
   # satisfy the 3/foo:succeeded prerequisite of 3/bar:
   $ cylc set --pre=3/foo my_workflow//3/bar
-  #   or:
+    # or:
```
oliver-sanders, line 93: Funky indentation, does not match line 82.

cylc/flow/scripts/set.py
```diff
@@ -72,15 +90,17 @@
...
   # satisfy all prerequisites (if any) of 3/bar and promote it to
   # the active window (and start checking its xtriggers, if any):
   $ cylc set --pre=all my_workflow//3/bar
 
-  # complete the "file1" custom output of 3/bar:
-  $ cylc set --out=file1 my_workflow//3/bar
+  # satisfy the dependence of 3000/bar on clock-trigger @clock1:
+  $ cylc set --pre=xtrigger/clock1 my_worklfow//3000/bar
+    # or reset it back to unsatisfied:
+  $ cylc set --pre=xtrigger/data:unsatisfied my_worklfow//3000/bar
```
oliver-sanders, line 103: More funky indentation.

oliver-sanders (commented):
Testing going well so far.

Was able to merge this with #6671 fairly easily, just needs the new bit of `delta_task_xtrigger` (from #6671) jammed into the new location. Will continue reviewing tomorrow with Tui showing xtrigger modifiers.

<details>
<summary>Merge diff</summary>

```diff
diff --cc cylc/flow/data_store_mgr.py
index 4e73bd382,ff688d043..000000000
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@@ -2521,54 -2539,78 +2535,109 @@@ class DataStoreMgr
          ext_trigger.time = update_time
          self.updates_pending = True
  
++
+     @staticmethod
+     def _set_task_xtrigger_modifiers(node_or_delta):
+         # update the xtrigger task modifiers
+         node_or_delta.is_retry = False
+         node_or_delta.is_wallclock = False
+         node_or_delta.is_xtriggered = False
+         for xtrigger in node_or_delta.xtriggers.values():
+             if xtrigger.satisfied:
+                 continue
+             if (
+                 xtrigger.label.startswith('_cylc_retry')
+                 or xtrigger.label.startswith('_cylc_submit_retry')
+             ):
+                 node_or_delta.is_retry = True
+             elif xtrigger.id.startswith('wall_clock'):
+                 node_or_delta.is_wallclock = True
+             else:
+                 node_or_delta.is_xtriggered = True
+ 
 -    def delta_task_xtrigger(self, sig, satisfied):
 -        """Create delta for change in task proxy xtrigger.
 +    def _delta_xtrigger(
-         self, tp_id: str, label: str, sig: str, satisfied: bool,
-         t_update: float
++        self,
++        tp_id: str,
++        label: str,
++        sig: str,
++        satisfied: bool,
++        update_time: float,
 +    ) -> None:
 +        """Helper for the two xtrigger delta methods."""
++        # fetch the task from the store
++        tp_id, tproxy = self.store_node_fetcher(Tokens(tp_id))
++        if not tproxy:
++            return
++
++        # create or fetch the updated delta
 +        tp_delta = self.updated[TASK_PROXIES].setdefault(
-             tp_id, PbTaskProxy(id=tp_id))
-         tp_delta.stamp = f'{tp_id}@{t_update}'
++            tp_id, PbTaskProxy(id=tp_id)
++        )
++        tp_delta.stamp = f'{tp_id}@{update_time}'
++
++        # populate all xtriggers on the delta if not already present
++        if not tp_delta.xtriggers:
++            # NOTE: if one xtrigger changes, we must include all in the
++            # delta, see https://github.com/cylc/cylc-flow/issues/6307
++            for _sig, xtrigger in tproxy.xtriggers.items():
++                if _sig == sig:
++                    # don't copy the xtrigger we are changing
++                    continue
++                _xtrigger = tp_delta.xtriggers[_sig]
++                _xtrigger.id = xtrigger.id
++                _xtrigger.label = xtrigger.label
++                _xtrigger.satisfied = xtrigger.satisfied
++                _xtrigger.time = xtrigger.time
++
++        # modify the xtrigger that has changed
 +        xtrigger = tp_delta.xtriggers[sig]
 +        xtrigger.id = sig
 +        xtrigger.label = label
 +        xtrigger.satisfied = satisfied
-         xtrigger.time = t_update
-         self.updates_pending = True
++        xtrigger.time = update_time
 +
-     def delta_xtrigger(self, sig: str, satisfied: bool) -> None:
-         """Create delta for an xtrigger, which multiple tasks might depend on.
++        # update the xtrigger task modifiers
++        self._set_task_xtrigger_modifiers(tp_delta)
 +
-         Args:
-             sig: xtrigger function call signature.
-             satisfied: xtrigger satisfied or not.
++        # ensure family modifier counts are updated
++        self.state_update_families.add(tproxy.first_parent)
 +
-         """
-         update_time = time()
-         # update all dependent task instances
-         for tp_id, label in self.xtrigger_tasks.get(sig, set()):
-             self._delta_xtrigger(tp_id, label, sig, satisfied, update_time)
++        self.updates_pending = True
 +
 +    def delta_task_xtrigger(
 +        self, itask: 'TaskProxy', label: str, sig: str, satisfied: bool
 +    ) -> None:
 +        """Create delta for a change in a task's xtrigger satisfaction status.
 +
 +        (Xtrigger satisfaction is normally handled by the xtrigger_mgr for all
 +        dependent tasks, but "cylc set" can set individual task dependence).
  
          Args:
 -            itask (cylc.flow.task_proxy.TaskProxy):
 -                Update task-node from corresponding task proxy
 -                objects from the workflow task pool.
 -            sig (str): Context of function call (name, args).
 -            satisfied (bool): Trigger message.
 +            itask: scheduler trask proxy that has changed status.
 +            sig: xtrigger function call signature.
 +            satisfied: satisfied or not.
  
          """
 -        update_time = time()
 -        for tp_id, label in self.xtrigger_tasks.get(sig, set()):
 -            # fetch the task from the store
 -            tp_id, tproxy = self.store_node_fetcher(Tokens(tp_id))
 -            if not tproxy:
 -                continue
 -
 -            # create or fetch the updated delta
 -            tp_delta = self.updated[TASK_PROXIES].setdefault(
 -                tp_id, PbTaskProxy(id=tp_id)
 -            )
 -            tp_delta.stamp = f'{tp_id}@{update_time}'
 -
 -            # populate all xtriggers on the delta if not already present
 -            if not tp_delta.xtriggers:
 -                # NOTE: if one xtrigger changes, we must include all in the
 -                # delta, see https://github.com/cylc/cylc-flow/issues/6307
 -                for _sig, xtrigger in tproxy.xtriggers.items():
 -                    if _sig == sig:
 -                        # don't copy the xtrigger we are changing
 -                        continue
 -                    _xtrigger = tp_delta.xtriggers[_sig]
 -                    _xtrigger.id = xtrigger.id
 -                    _xtrigger.label = xtrigger.label
 -                    _xtrigger.satisfied = xtrigger.satisfied
 -                    _xtrigger.time = xtrigger.time
 +        tp_id = self.id_.duplicate(
 +            cycle=str(itask.point),
 +            task=itask.tdef.name,
 +        ).id
 +        self._delta_xtrigger(tp_id, label, sig, satisfied, time())
  
 -            # modify the xtrigger that has changed
 -            xtrigger = tp_delta.xtriggers[sig]
 -            xtrigger.id = sig
 -            xtrigger.label = label
 -            xtrigger.satisfied = satisfied
 -            xtrigger.time = update_time
++    def delta_xtrigger(self, sig, satisfied):
++        """Create delta for change in task proxy xtrigger.
+ 
 -            # update the xtrigger task modifiers
 -            self._set_task_xtrigger_modifiers(tp_delta)
++        Args:
++            sig: xtrigger function call signature.
++            satisfied: xtrigger satisfied or not.
+ 
 -            # ensure family modifier counts are updated
 -            self.state_update_families.add(tproxy.first_parent)
++        """
++        update_time = time()
++        # update all dependent task instances
++        for tp_id, label in self.xtrigger_tasks.get(sig, set()):
++            self._delta_xtrigger(tp_id, label, sig, satisfied, update_time)
+ 
 -            self.updates_pending = True
+ 
      def delta_from_task_proxy(self, itask: TaskProxy) -> None:
          """Create delta from existing pool task proxy.
  
diff --cc tests/integration/test_xtrigger_mgr.py
index 33f2d4e95,b3c75af44..000000000
--- a/tests/integration/test_xtrigger_mgr.py
+++ b/tests/integration/test_xtrigger_mgr.py
@@@ -365,54 -286,132 +366,185 @@@ async def test_1_seq_clock_trigger_2_ta
          )
  
  
 +async def test_force_satisfy(flow, start, scheduler, log_filter):
 +    """It should satisfy valid xtriggers and ignore invalid ones."""
 +    id_ = flow({
 +        'scheduling': {
 +            'xtriggers': {
 +                'x': 'xrandom(0)'
 +            },
 +            'graph': {
 +                'R1': '@x => foo'
 +            },
 +        }
 +    })
 +    schd = scheduler(id_)
 +    async with start(schd):
 +        foo = schd.pool.get_tasks()[0]
 +
 +        # check x not satisfied yet
 +        assert not foo.state.xtriggers['x']  # not satisified
 +
 +        # force satisfy it
 +        xtrigs = {
 +            "x": True,  # it should satisfy this one
 +            "y": True  # it should just ignore this one
 +        }
 +        schd.xtrigger_mgr.force_satisfy(foo, xtrigs)
 +
 +        assert foo.state.xtriggers['x']  # satisified
 +        assert log_filter(
 +            contains=(
 +                'xtrigger prerequisite satisfied (forced): x = xrandom(0)'))
 +
 +        # force satisfy it again
 +        schd.xtrigger_mgr.force_satisfy(foo, xtrigs)
 +        assert foo.state.xtriggers['x']  # satisified
 +        assert log_filter(
 +            contains=(
 +                'xtrigger prerequisite already satisfied: x = xrandom(0)'))
 +
 +        # force unsatisfy it
 +        schd.xtrigger_mgr.force_satisfy(foo, {"x": False})
 +        assert not foo.state.xtriggers['x']  # not satisified
 +        assert log_filter(
 +            contains=(
 +                'xtrigger prerequisite unsatisfied (forced): x = xrandom(0)'))
 +
 +        # force unsatisfy it again
 +        schd.xtrigger_mgr.force_satisfy(foo, {"x": False})
 +        assert not foo.state.xtriggers['x']  # not satisified
 +        assert log_filter(contains=(
 +            'xtrigger prerequisite already unsatisfied: x = xrandom(0)'
 +        ))
++
++
+ async def test_data_store(flow, start, scheduler):
+     """It should update the data store with xtrigger state."""
+     id_ = flow({
+         'scheduling': {
+             'initial cycle point': 'previous(T00)',
+             'xtriggers': {
+                 'clock1': 'wall_clock()',
+                 'clock2': 'wall_clock(offset="P1Y")',
+             },
+             'graph': {
+                 'R1': '''
+                     @clock1 => foo
+                     @clock2 => foo
+                 '''
+             }
+         }
+     })
+     schd: Scheduler = scheduler(id_)
+     async with start(schd):
+         await schd.update_data_structure()
+         itask = schd.pool.get_tasks()[0]
+ 
+         # extract xtrigger entry from the data store
+         xtriggers = schd.data_store_mgr.data[
+             schd.tokens.id
+         ][TASK_PROXIES][itask.tokens.id].xtriggers
+         clock1, clock2 = sorted(x for x in xtriggers if 'wall_clock' in x)
+ 
+         # it should not be satisfied (yet)
+         assert xtriggers[clock1].label == 'clock1'
+         assert xtriggers[clock1].satisfied is False
+ 
+         # execute the xtrigger
+         schd.xtrigger_mgr.call_xtriggers_async(itask)
+ 
+         # an update delta should be produced
+         # NOTE: both xtriggers should be present in the update
+         # (see https://github.com/cylc/cylc-flow/issues/6307)
+         task_delta = schd.data_store_mgr.updated[TASK_PROXIES][itask.tokens.id]
+         assert task_delta.xtriggers[clock1].satisfied is True
+         assert task_delta.xtriggers[clock2].satisfied is False
+ 
+         # the xtrigger should be satisfied in the data store
+         await schd.update_data_structure()
+         assert xtriggers[clock1].label == 'clock1'
+         assert xtriggers[clock1].satisfied is True
+ 
+ 
+ def satisfy_xtrigger_functions(schd, stdout='[true, {}]', ret_code=0):
+     """Satisfy and dequeue any xtrigger subprocesses."""
+     for item in list(schd.proc_pool.queuings):
+         ctx, _, callback, *_ = item
+         if isinstance(schd.proc_pool.queuings[0][0], SubFuncContext):
+             # dequeue from the proc pool
+             schd.proc_pool.queuings.remove(item)
+ 
+             # mock the xtrigger output
+             ctx.ret_code = ret_code
+             ctx.out = stdout
+ 
+             # run the callback
+             callback(ctx)
+ 
+ 
+ async def test_xtrigger_modifiers(flow, scheduler, start):
+     """It should update xtrigger derived task modifiers."""
+     id_ = flow({
+         'scheduling': {
+             'initial cycle point': 'previous(T00)',
+             'xtriggers': {
+                 'echo': 'echo("whatever", succeed=True)',
+             },
+             'graph': {
+                 'R1': '''
+                     @wall_clock => foo
+                     @echo => foo
+                 '''
+             },
+         },
+         'runtime': {
+             'foo': {
+                 'execution retry delays': 'PT0S',
+             },
+         },
+     })
+     schd = scheduler(id_)
+     async with start(schd):
+         # configure the task to retry
+         itask = schd.pool.get_tasks()[0]
+         schd.task_events_mgr._retry_task(itask, 0)
+         await schd.update_data_structure()
+ 
+         # the task proxy object in the data store
+         ds_tproxy = schd.data_store_mgr.data[schd.tokens.id][TASK_PROXIES][
+             itask.tokens.id
+         ]
+ 
+         # the "root" family proxy object in the data store
+         ds_fproxy = schd.data_store_mgr.data[schd.tokens.id][FAMILY_PROXIES][
+             itask.tokens.duplicate(task='root').id
+         ]
+ 
+         # the task modifiers should be initialised on task creation
+         assert len(itask.state.xtriggers) == 3
+         assert ds_tproxy.is_retry is True
+         assert ds_tproxy.is_wallclock is True
+         assert ds_tproxy.is_xtriggered is True
+ 
+         # the modifiers should bubble up the family tree
+         assert ds_fproxy.is_retry is True
+         assert ds_fproxy.is_wallclock is True
+         assert ds_fproxy.is_xtriggered is True
+ 
+         # run the xtriggers
+         schd.xtrigger_mgr.call_xtriggers_async(itask)  # run xtrigs
+         satisfy_xtrigger_functions(schd)  # mock results
+         schd.xtrigger_mgr.call_xtriggers_async(itask)  # process callbacks
+ 
+         await schd.update_data_structure()
+ 
+         # the task modifiers should have been updated
+         assert ds_tproxy.is_retry is False
+         assert ds_tproxy.is_wallclock is False
+         assert ds_tproxy.is_xtriggered is False
+ 
+         # the modifiers should bubble up the family tree
+         assert ds_fproxy.is_retry is False
+         assert ds_fproxy.is_wallclock is False
+         assert ds_fproxy.is_xtriggered is False
diff --git a/cylc/flow/task_events_mgr.py b/cylc/flow/task_events_mgr.py
index bb9a88ba3..3f4eee9fd 100644
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -1307,7 +1307,7 @@ class TaskEventsManager():
             .setdefault(sig, set())
             .add((itask.tokens.id, label))
         )
-        self.data_store_mgr.delta_task_xtrigger(sig, False)
+        self.data_store_mgr.delta_xtrigger(sig, False)
 
         if itask.state_reset(TASK_STATUS_WAITING):
             self.data_store_mgr.delta_task_state(itask)
diff --git a/tests/integration/tui/test_app.py b/tests/integration/tui/test_app.py
index 9ae32f8b4..d384761cc 100644
--- a/tests/integration/tui/test_app.py
+++ b/tests/integration/tui/test_app.py
@@ -46,7 +46,7 @@ def _add_xtrig(schd, itask, sig, label, satisfied=False):
         .setdefault(sig, set())
         .add((itask.tokens.id, label))
     )
-    schd.data_store_mgr.delta_task_xtrigger(sig, satisfied)
+    schd.data_store_mgr.delta_xtrigger(sig, satisfied)
 
 
 def set_task_state(schd, task_states):
```

</summary>

cylc/flow/data_store_mgr.py
hjoliver: OK, but no implications for this PR, right?

cylc/flow/data_store_mgr.py
hjoliver: >We have both attempted to solve a similar problem for different reasons, months apart. But, unfortunately, it is gonna conflict bad!

Actually only about 1 week apart according to the dates on our PRs. In that week, I didn't look closely enough at your PR, sorry!, because the headline suggests it's UI focused, whereas this one is (I thought) just about task dependence on xtriggers, in the scheduler. But yes, I can see the overlap now.

> This has actually conflicted with yet another xtrigger PR - https://github.com/cylc/cylc-flow/pull/6721 - which makes acquiring the itask difficult.

That one was only a one-line bug fix though.

hjoliver:
> I've updated my test branch [hjoliver#65](https://github.com/hjoliver/cylc-flow/pull/65)

I must have missed the notification for that @wxtim (don't think you mentioned it anywhere above?). I have good coverage already here, but unit tests are a better first line of defense than integration, so I'll take that - thanks.

hjoliver:
> This PR has now been extended to unsetting xtriggers? This isn't actually necessary right?

Not extended, it had that right from the start. 

Group trigger needs setting of xtriggers, not unsetting. That was the motivation for doing this PR now of course, but it technically it addresses #3221 (2019) which is for both set and unset, and that issue links to a forum post where the use case is actually for unsetting.

hjoliver:
> >    I've changed the xtrigger "states" (for the CLI args) from ":succeeded" and ":waiting" to ":satisfied" and ":unsatisfied" (attempting to reuse task statuses for this was a bit lazy, and potentially confusing because only a subset of states is relevant and the meaning is kinda different - we can't unsatisfy task prerequisites)
> 
> Dammit!
> I get why you have done this, but I actually think the task state was a better option.

It's easy to revert if necessary - just change the value of two constants (plus some tests that don't use the constants but probably should).

I'll leave it as-is for now, and will respond on Element.

cylc/flow/data_store_mgr.py
oliver-sanders: Direct, no, but it would be good to build this atop the bugfix rather than in parallel. The bugfix is blocking https://github.com/cylc/cylc-ui/pull/2102 which would be handy for reviewing this if nothing else.

hjoliver:
I'm not sure you've grokked my main argument here. It is really that the `:succeeded` and `:waiting` syntax strongly (and wrongly) suggests "setting the xtrigger state". This PR does not do that, it only sets (satisfies or unsatisfies) the target task's dependence on the xtrigger.

We can use that kind of syntax for setting xtrigger state or outputs - which I've punted to #6725 - but prerequisites are different. A prerequisite is a dependency on *a particular state/output*, and that dependency can be satisfied or not - either way it is the same state/output. 

Dependence on custom outputs is perhaps the clearest example of this, because they don't have an "opposite state" that could be "set" as a proxy for unsatisfying the prerequisite. (As I noted above there is actually a case for unsetting task prerequisites too - to undo a mistakenly set one).

So imagine an xtrigger `x` has a custom output `a` as well as `succeeded` (it's conceivable that we will support this in future, but that's somewhat beside the point).  It would be perfectly sensible to satisfy `foo`'s dependence on `x:a` like this:
```console
$ cylc set --pre=xtrigger/x:a 1/foo  # OK  (satisfy foo's dependence on x:a)
```
but there's no way to unsatisfy it using the `:output` field.

Similarly, we can satisfy `foo`'s dependence on `x;succeeded` like this:
```console
$ cylc set --pre=xtrigger/x[:succeeded] 1/foo  # OK (satisfy foo's dependence on x:succeeded)
```
but this doesn't really make sense (for a *prerequisite*):
```console
$ cylc set --pre=xtrigger/x:waiting 1/foo  # WRONG (unsatisfy foo's dependence on x:succeeded!)
```

That syntax only makes sense for setting outputs:
```console
$ cylc set xtrigger/:waiting  # OK-but-not-supported yet (return the xtrigger to wait-and-check status)
```

hjoliver:
OK, latest commit:
- remove the ability to unsatisfy xtrigger prerequisites (punted to #6741)
- go back to `--pre=xtrigger/x[:succeeded]`
  - (this makes sense for satisfying the prerequisite, it's `xtrigger/x[:waiting]` that does not make sense for unsatisfying it!)
- made a start on cleaning up some ambiguities:
  - distinguish between "a satisfied xtrigger prerequisite" and "a succeeded xtrigger"
  - say "satisfy a prerequisite" and "complete an output" - some places we just said "set" for both (which is OK but could be clearer) or even "satisfied an output" (which is 🤮 !)

wxtim (approved):
- [x] Read latest changes
- [x] Final manual test

oliver-sanders:
> To speed this, for group trigger, I will remove the command interface for unset, and we can up hook the internals to the agreed API later via a new issue: https://github.com/cylc/cylc-flow/issues/6741

Good idea.

oliver-sanders (commented):
Nearly done with review, some small comments.

hjoliver:
@oliver-sanders - 
>>     * --pre=xtrigger/<xtrigger>[:succeeded]
> (we don't really need to document the :succeeded bit as there's no other output you can set, so could trim it for simplicity)

I've put `[:succeeded]` in deliberately because:
- it shows the syntax is consistent with task prerequisite setting 
- it's possible we'll allow xtrigger custom outputs in the future, for xtrigger-driven optional branching, consistent with bringing xtriggers more into the task model.

But I'll add a comment to indicate that `:succeeded` is the only xtrigger output at the moment - ok?

cylc/flow/task_proxy.py
```diff
@@ -564,20 +564,45 @@ def state_reset(
 
     def satisfy_me(
         self,
-        task_messages: 'Iterable[Tokens]',
+        outputs: 'Iterable[Tokens]',
```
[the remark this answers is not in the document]
hjoliver: Agreed, I'm not sure why I made this change. Will revert it.

cylc/flow/task_pool.py
hjoliver: I've used quick_relative_id

cylc/flow/task_pool.py
```diff
@@ -2043,37 +2102,69 @@ def set_prereqs_and_outputs(
...
+        """
+        valid_x_labels = tdef.get_xtrigs(point)
+
+        # weed out task prerequisites and convert state to bool
+        req_x = _get_xtrig_prereqs(prereqs)
+
+        for xtrig in set(req_x.keys()) - valid_x_labels:
+            if xtrig != "all":
+                LOG.warning(
+                    f'{point}/{tdef.name} does not depend on'
```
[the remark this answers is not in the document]
hjoliver: Used quick_relative_id

tests/integration/test_xtrigger_mgr.py
```diff
@@ -203,15 +207,91 @@ def mytrig(*args, **kwargs):
...
 
+async def test_set_xtriggers_restart(flow, start, scheduler, db_select):
+    """It should write satisfied xtrig prerequisites to the DB for restart.
+
+    This checks that individual task dependence on xtriggers, which can be
+    artificially satisfied by "cylc set" persists across restart.
+
+    See also test_xtriggers_restart, for persistence of xtrigger results.
+
+    """
```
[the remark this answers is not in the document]
hjoliver: Good call, that actually wasn't working. Done, and integration test added.

hjoliver:
All comments addressed.

wxtim:
Sorry - I think I've found a fairly large problem:

```cylc
# flow.cylc
[scheduler]
    allow implicit tasks = True
    cycle point format = %Y

[scheduling]
    initial cycle point = 2026
    [[graph]]
        P1Y = """
            @english => foo
            @cornish => foo
        """
    [[xtriggers]]
        english = echo('Hello, World!', succeed=False)
        cornish = echo('Helo, Byd!', succeed=False)
```

Start the workflow then:

```bash
> cylc set xtriggers/manual//2026/foo --pre=xtrigger/all
> cylc log xtriggers/manual
# ...
2025-04-29T09:31:04+01:00 INFO - Command "set" received. ID=871d7dca-09de-4c75-b4dc-93f0617ee84f
    set(flow=[], flow_wait=False, outputs=[], prerequisites=['xtrigger/all'], tasks=['2026/foo'])
2025-04-29T09:31:05+01:00 WARNING - Invalid prerequisite cycle point:
    Incompatible value for <class 'cylc.flow.cycling.iso8601.ISO8601Point'>: xtrigger: Invalid ISO 8601 date representation: xtrigger
2025-04-29T09:31:05+01:00 INFO - Command "set" actioned. ID=871d7dca-09de-4c75-b4dc-93f0617ee84f

**Comment by wxtim** [comment]
### Manual testing

<details>
<summary>
I created a workflow with two xtriggers which could only be manually set to succeeded.
</summary>

```cylc
[scheduler]
    allow implicit tasks = True
    cycle point format = %Y

[scheduling]
    initial cycle point = 2026
    [[graph]]
        P1Y = """
            @english => foo
            @cornish => foo
        """
    [[xtriggers]]
        english = echo('Hello, World!', succeed=False)
        cornish = echo('Helo, Byd!', succeed=False)
```
</details>

Steps:
* Workflow run.
* Set fist xt satisfied.
* Stopped workflow.
* Restarted workflow.
  - [x] xtrigger remained satisfied.
* Commented other xtrigger in source
* Ran `cylc vr`.
  - [x] Task ran.

oliver-sanders:
Minor defect (non-blocking): A bad cycle point (a typo of xtrigger in this case) can get logged once for each task. Ideally this validation would only be performed once. Not a biggie, but could make a mess of the logs where larger numbers of tasks are involved.

```
2025-04-29T09:24:30Z INFO - Command "set" received. ID=215017ab-1c2a-4d34-a2eb-4a5f1d3b25e0
    set(flow=[], flow_wait=False, outputs=[], prerequisites=['xtirgger/all:succeeded'], tasks=['*/*'])
2025-04-29T09:24:30Z WARNING - Invalid prerequisite cycle point:
    Incompatible value for <class 'cylc.flow.cycling.iso8601.ISO8601Point'>: xtirgger: Invalid ISO 8601 date representation: xtirgger
2025-04-29T09:24:30Z WARNING - Invalid prerequisite cycle point:
    Incompatible value for <class 'cylc.flow.cycling.iso8601.ISO8601Point'>: xtirgger: Invalid ISO 8601 date representation: xtirgger
2025-04-29T09:24:30Z WARNING - Invalid prerequisite cycle point:
    Incompatible value for <class 'cylc.flow.cycling.iso8601.ISO8601Point'>: xtirgger: Invalid ISO 8601 date representation: xtirgger
2025-04-29T09:24:30Z WARNING - Invalid prerequisite cycle point:
    Incompatible value for <class 'cylc.flow.cycling.iso8601.ISO8601Point'>: xtirgger: Invalid ISO 8601 date representation: xtirgger
```

hjoliver:
> 2) Retry xtriggers cannot be selectively satisfied

Hmm, do we want to allow that? Retries are implemented via xtriggers under the hood, but we don't really expose that to users. The somewhat non-obvious xtrigger name is not logged when the retry gets teed up, and we don't reveal it in `cylc show`. It might be best to prevent `xtrigger/all` from satisfying retry xtriggers.

> I expect the /foo is likely the cause of this. 

That might be a problem too, but the initial problem is that we reject invalid xtriggers from user input by checking them against xtriggers recorded in the taskdef.

oliver-sanders:
>  and we don't reveal it in cylc show.

No, we do reveal the retry xtrigger label in `cylc show`, that's where I got it from.

hjoliver:
Huh, I just tried the same thing in a single-task workflow with an execution retry, and show didn't reveal it. 🤔

oliver-sanders:
Make sure there is enough retry delay for the xtrigger to be displayed. Shows up every time for me.

wxtim:
I'm seeing it in show, but only _after_ I've triggered it.

hjoliver:
```cylc
[scheduling]
    [[graph]]
        R1 = foo
[runtime]
    [[foo]]
        execution retry delays = PT5M
        script = false
```
For this ⬆️  `cylc show` refuses to show the retry xtrigger.

oliver-sanders:
Curious, seems to need something to cause them to show. Happy to shift this reporting part of the issue onto https://github.com/cylc/cylc-flow/issues/6742.

hjoliver:
> I agree these labels aren't especially user friendly, they contain an auto-incremented number and the task name,

Actually that's not an autoincremented number, it's just the task ID `1/foo`.  Maybe it should have an additional autoincremented number?

oliver-sanders:
> Actually that's not an autoincremented number, it's just the task ID 1/foo. Maybe it should have an additional autoincremented number?

That makes sense actually. If a subsequent retry occurs is recycles the old xtrigger (i.e. resets it) so doesn't need an auto-incrementer (since we cannot have an execution and submission retry active at the same time).

oliver-sanders:
The easiest way to get retry xtriggers to show up on this branch (without https://github.com/cylc/cylc-flow/pull/6671 merged in) is to reload the workflow.

hjoliver:
> 5) Duplicate xtriggers must all be set but are not all reported as prerequisites.

Same problem on master.

I think actually validation should detect false1 and false2 as duplicates and flag that as an error. There's not point in have two different xtrigger names (labels) for the exact some function call. It gives the impression there are two prerequisites when really (for natural xtrigger completion) there's only one. That does mess with this branch a bit though, since the point of it is to satisfy individual xtrigger prerequisites independently of xtrigger completion.

oliver-sanders:
Yep, I'm aware it's an issue one master, only mentioned here because this PR turns it from a reporting issue into an interface issue.

I thought you might suggest we consider this a validation problem, and can bump this to an issue if we decide to go that way.

I think duplicate xtrigger labels could be considered valid (and are certainly implemented as such). Note, there's no performance impact because the xtrigger_mgr deduplicates them internally. The use case would be for programmatically generated workflows, e.g:

```
[scheduling]
    [[xtriggers]]
{% for lead in lead_times + fcst_times %}
        {{ clock }}_lead = wall_clock(offset=PT{{ lead }}M)
{% endfor %}
```

Or workflows consisting of modules e,g:

```
# inc/graph_a.cylc

[scheduling]
    [[xtriggers]]
         @model_a_clock => model_a
```

```
# inc/graph_b.cylc

[scheduling]
    [[xtriggers]]
         @model_b_clock => model_b
```

hjoliver:
> In such circumstances, it is easy to accidentally generate duplicate xtriggers, but should this be considered an error? Or just two graph-variables referencing the same xtrigger?

Fair point. OK, I think we should continue to support multiple labels for the same xtrigger, fix the reporting issue, and maybe print a warning in validation to alert users that multiple labels are actually pointing to the same xtrigger.

The set command is doing the right thing on this branch though, I think. It's just a reporting issue.

hjoliver:
> 1) [Minor] Xtrigger validation errors repeated for each task.

This is also a problem on master.

hjoliver:
Duplicate warning done:

```cylc
[scheduling]
  [[xtriggers]]
    f1 = xrandom(0)
    f2 = xrandom(0)  # dup
    f3 = xrandom(1)
    f4 = xrandom(1)  # dup
    g = xrandom(3)
  [[graph]]
    R1 = """
      @f1 & @f2 & @f3 & @f4 & @g => foo
    """
[runtime]
  [[foo]]
```
Validation:
```console
$ cylc val .
INFO - Duplicate xtriggers: f1,f2 = xrandom(0)
INFO - Duplicate xtriggers: f3,f4 = xrandom(1)
WARNING - Duplicate xtrigger prerequisites get satisfied naturally at once, but they can be
    satisfied separately with `cylc set`.
Valid for cylc-8.5.0.dev
```

hjoliver:
> 3) [Critical] Premature shutdown for parentless tasks w/ sequential xtriggers

Fixed.

hjoliver:
Summary:

> 1) [Minor] Xtrigger validation errors repeated for each task. ✅ 
> 2) Retry xtriggers cannot be selectively satisfied ✅ 
> 3) [Critical] Premature shutdown for parentless tasks w/ sequential xtriggers ✅ 
> 4) [Could Be Better] Prereq validation could detect incorrect xtrigger specification. ❌ (follow-up?)
> 5) Duplicate xtriggers must all be set but are not all reported as prerequisites. ✅ 

Plus (follow-up?) - consider the full reload problem - after adding or removing task or xtrigger prerequisites (same problem on master for task prerequisites - see `pool._get_valid_prereqs()` - it use the taskdef to checkout validity of requested prereqs).

oliver-sanders:
> 2. Retry xtriggers cannot be selectively satisfied ✅

I think this one is still outstanding. Quick example:

```cylc
[scheduling]
    [[graph]]
        R1 = foo

[runtime]
    [[foo]]
        script = false
        execution retry delays = 10 * PT1M
```

To reproduce:

```console
$ cylc vip
...

$ cylc reload  # makes the retry xtrigger show up (without #6671 merged)
...

$ cylc show //1/foo
...
other: ('⨯': not satisfied)
  ⨯ xtrigger "_cylc_retry_1/foo = wall_clock(trigger_time=2025-05-07T10:27:46,950535Z)"

$ cylc set //1/foo --pre=xtrigger/_cylc_retry_1/foo
$ cylc set //1/foo --pre=xtrigger/_cylc_retry_1_foo

$ cylc cat-log
...
WARNING - 1/foo does not depend on xtrigger "_cylc_retry_1_foo"
WARNING - 1/foo does not depend on xtrigger "_cylc_retry_1"
```

oliver-sanders:
Have opened issues for the other matters:

* https://github.com/cylc/cylc-flow/issues/6747
* https://github.com/cylc/cylc-flow/issues/6748
* https://github.com/cylc/cylc-flow/issues/6749

@hjoliver, I assigned you to the ["same problem on master: see pool._get_valid_prereqs()"](https://github.com/cylc/cylc-flow/issues/6749) issue, hopefully quick.

hjoliver:
> >   Retry xtriggers cannot be selectively satisfied ✅
> I think this one is still outstanding. Quick example:

Pretty sure I fixed that, investigating...

oliver-sanders:
MB, not sure how I got that result.

oliver-sanders (approved):
:+1:

PR #6697, opened 2025-04-02
more nonlocals

wxtim:
See also discussion in https://github.com/cylc/cylc-flow/pull/6692

These were added to branch master separately to those fixed in 8.4.x/

### DIFF
--- a/cylc/flow/tui/overlay.py
+++ b/cylc/flow/tui/overlay.py
@@ -482,7 +482,6 @@ def log(app, id_=None, list_files=None, get_log=None):
                 restored when the command exits.
 
         """
-        text_widget
 
         with tempfile.NamedTemporaryFile('w+') as temp_file:
             # write the text into a temp file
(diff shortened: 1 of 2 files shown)


cylc/flow/tui/overlay.py
```diff
@@ -482,7 +482,7 @@ def open_in_editor(*_, command):
                 restored when the command exits.
 
         """
-        nonlocal text_widget
+        text_widget
```
wxtim, line 485:
```suggestion
```

tests/integration/tui/test_logs.py
```diff
@@ -409,7 +409,7 @@ async def test_external_editor(
 
     class FakePopen:
         def __init__(self, cmd, *args, raises=None, **kwargs):
-            nonlocal fake_popen_instances
+            fake_popen_instances
```
wxtim, line 412:
```suggestion
```

wxtim:
MyPy Failure applies to 8.4.x and master, will fix separately and let sync PR sort it out.

oliver-sanders (approved):

Issue #5979, opened 2024-02-19
platforms: 7/8 syntax mismatch should result in submission failure


oliver-sanders:
A user had upgraded their workflow to Cylc 8 syntax, but broadcasted a `[remote]host` (Cylc 7 syntax) causing a syntax conflict.

Cylc detected this issue correctly, but rather than marking the task as submit failed (e.g. like we would do for a Bash syntax error), the scheduler aborted (i.e. shut down on unexpected error). This isn't terribly helpful as it doesn't give the user a chance to fix the error. When the workflow is restarted, the task (still in the waiting state) attempted to submit again, causing the workflow to abort again :facepalm:.

Additionally, the error that is raised is missing some context:

https://github.com/cylc/cylc-flow/blob/088450fae08abc3d10e093edc1e807ffa6399395/cylc/flow/task_job_mgr.py#L1126-L1130

It says `the following are not compatible:\n`, then the error ends without telling you what configs are incompatible.

PR #6711, opened 2025-04-07
Prevent illegal platform host broadcast

### LINKED ISSUE #6693: Invalid Broadcast of Illegal Config breaks scheduler
### Description

Broadcast reject try to merge Cylc 7 platform-like configs (`[remote]` and `[host]`) with the running config. If the running config uses platforms the configs cannot be merged and pull the scheduler down.

### Reproducible Example

```cylc
[scheduler]
    allow implicit tasks = True

[scheduling]
    cycling mode = integer
    [[graph]]
        R1 = foo => bar

[runtime]
    [[bar]]
        platform = localhost
    [[foo]]
        script = cylc broadcast ${CYLC_WORKFLOW_ID} -p 1 -n bar -s '[remote]host=localhost'
```

### Expected Behaviour

Broadcast shouldn't allow this.

wxtim:
Closes #6693 

**Check List**

### DIFF
--- a/cylc/flow/broadcast_mgr.py
+++ b/cylc/flow/broadcast_mgr.py
@@ -320,11 +320,12 @@ class BroadcastMgr:
                                 self.broadcasts.get(point_string, {})
                                 .get(namespace, {})
                             )
-                            self.check_for_old_and_new_platform_settings(
+                            if self.bc_mixes_old_and_new_platform_settings(
                                 newconfig,
                                 namespace,
                                 coerced_setting,
-                            )
+                            ):
+                                continue
 
                             if namespace not in self.broadcasts[point_string]:
                                 self.broadcasts[point_string][namespace] = {}
@@ -355,7 +356,7 @@ class BroadcastMgr:
         return modified_settings, bad_options
 
     @staticmethod
-    def check_for_old_and_new_platform_settings(
+    def bc_mixes_old_and_new_platform_settings(
         task_config, namespace, coerced_setting
     ):
         """Check for combination of old ([remote]host) and new (platform)
@@ -366,8 +367,10 @@ class BroadcastMgr:
             fail_if_platform_and_host_conflict(
                 task_config, namespace
             )
+            return False
         except PlatformLookupError as exc:
             LOG.error('Cannot apply broadcast:\n' + '\n    '.join(exc.args))
+            return True
 
     @staticmethod
     def _cancel_keys_in_prunes(prunes, cancel_keys):
--- a/cylc/flow/broadcast_mgr.py
+++ b/cylc/flow/broadcast_mgr.py
@@ -310,19 +310,18 @@ class BroadcastMgr:
                         if namespace not in self.linearized_ancestors:
                             bad_namespaces.append(namespace)
                         elif not bad_point:
-                            # Check broadcast against config:
-                            self.check_for_old_and_new_platform_settings(
-                                self.schd.config.get_config(
-                                    ['runtime', namespace]
-                                ).copy(),
-                                namespace,
-                                coerced_setting,
-                            )
-                            # Check broadcast against existing broadcasts:
-                            self.check_for_old_and_new_platform_settings(
+                            # Check broadcast against config and against
+                            # existing broadcasts:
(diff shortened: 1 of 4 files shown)


cylc/flow/broadcast_mgr.py
```diff
@@ -304,6 +310,14 @@ def put_broadcast(
...
                             bad_namespaces.append(namespace)
                         elif not bad_point:
+                            self.check_for_old_and_new_platform_settings(
+                                self.schd.config.get_config(
+                                    ['runtime', namespace]
+                                ).copy(),
+                                namespace,
+                                coerced_setting,
+                            )
+
```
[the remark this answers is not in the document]
wxtim: Turned out to be fairly simple to cater for.

cylc/flow/broadcast_mgr.py
```diff
@@ -304,6 +310,23 @@ def put_broadcast(
...
+                            )
+                            # Check broadcast against existing broadcasts:
+                            self.check_for_old_and_new_platform_settings(
+                                self.broadcasts.get(point_string, {})
+                                .get(namespace, {})
+                                .copy(),
+                                namespace,
+                                coerced_setting,
+                            )
+
```
oliver-sanders, line 329: I think you can combine these two into something like:

```python
# apply any broadcasts to the task config
reconfig = pdeepcopy(self.schd.config.get_config(...))
poveride(rtconfig, self.broadcast_mgr.get(...))

# check for conflicting configs
self.check_for_old_and_new_platform_settings(
    rtconfig,
    ...
)
```

oliver-sanders (approved):

tests/integration/test_broadcast_mgr.py
```diff
@@ -0,0 +1,69 @@
...
+    """`put_broadcast` gracefully rejects invalid broadcast:
+
+    Existing config = [task][remote]host = foo
+    Broadcast       = [task]platform = bar
+
+    https://github.com/cylc/cylc-flow/issues/6693
+    """
+    conf = one_conf.copy()
+    conf.update({'runtime': {'root': {'platform': 'foo'}}})
+    wid = flow(conf)
```
MetRonnie, line 32:
```suggestion
    one_conf.update({'runtime': {'root': {'platform': 'foo'}}})
    wid = flow(one_conf)
```

cylc/flow/broadcast_mgr.py
```diff
@@ -304,6 +310,22 @@ def put_broadcast(
...
+                            poverride(
+                                newconfig,
+                                self.broadcasts.get(point_string, {})
+                                .get(namespace, {})
+                            )
+                            self.check_for_old_and_new_platform_settings(
+                                newconfig,
+                                namespace,
+                                coerced_setting,
+                            )
```
MetRonnie, line 327: All this does is log an error. The broadcast is still applied

tests/integration/test_broadcast_mgr.py
```diff
@@ -0,0 +1,69 @@
...
+    schd = scheduler(flow(one_conf))
+    async with start(schd):
+        bc_mgr = schd.broadcast_mgr
+        bc_mgr.put_broadcast(
+            point_strings=['1'],
+            namespaces=['one'],
+            settings=[{'remote': {'host': 'bar'}}]
+        )
+        # this should not be allowed, if it is the scheduler will crash
+        # when unpaused:
```
MetRonnie, line 63: Please test this

MetRonnie (changes requested):

tests/integration/test_broadcast_mgr.py
wxtim: That would rather have prevented me putting a non-functioning PR...

MetRonnie (approved):
Only problem is that there is no error reported in the command line, however this is probably good enough

wxtim:
> Only problem is that there is no error reported in the command line, however this is probably good enough

Oliver tells me it can't be, yet!

oliver-sanders:
@MetRonnie, see https://github.com/cylc/cylc-flow/pull/6574

wxtim:
Also closes #5979

PR #6719, opened 2025-04-08
Fix.6701.scheduler time zone change bug

### LINKED ISSUE #6701: UTC Edge effect
### Description

Symptoms: Workflows which remained up over the weekend the clocks changed in the UK displayed odd times for submit time such as:

Image

This seems to be caused by the database containing the local time, but retaining the Z timezone for `time_submit` in the `task_jobs` table.

wxtim:
Closes #6701 

**Check List**


cylc/flow/wallclock.py
```diff
@@ -37,21 +37,16 @@
...
-    TimeZoneFormatMode.reduced)
-TIME_ZONE_STRING_LOCAL_EXTENDED = get_local_time_zone_format(
-    TimeZoneFormatMode.extended)
 TIME_ZONE_STRING_UTC = "Z"
 TIME_ZONE_UTC_UTC_OFFSET = (0, 0)
-TIME_ZONE_LOCAL_UTC_OFFSET = get_local_time_zone()
-TIME_ZONE_LOCAL_UTC_OFFSET_HOURS = TIME_ZONE_LOCAL_UTC_OFFSET[0]
-TIME_ZONE_LOCAL_UTC_OFFSET_MINUTES = TIME_ZONE_LOCAL_UTC_OFFSET[1]
 
 TIME_ZONE_LOCAL_INFO = {
```
wxtim, line 43: This is only used in `cylc.flow.data_store_mgr.DataStoreMgr.generate_definition_elements` which seems to be for fields which are not updated. @dwsutherland might wish to comment on whether this safe.

tests/integration/test_workflow_db_mgr.py
```diff
@@ -178,3 +183,36 @@ async def test_record_only_non_clock_triggers(
...
+            r'SELECT time_submit, time_submit_exit FROM task_jobs'
+        ).fetchone()
+        time_submit = datetime.strptime(time_submit, '%Y-%m-%dT%H:%M:%S%z')
+        time_submit_exit = datetime.strptime(
+            time_submit_exit, '%Y-%m-%dT%H:%M:%S%z'
+        )
+
+        # If submission time is > 20 seconds we may have other problems:
+        assert time_submit_exit >= time_submit
+        assert time_submit_exit < time_submit + timedelta(seconds=10)
```
MetRonnie, line 218: Confusingly, comment refers to 20s, whereas the assertion tests it is within 10s

cylc/flow/wallclock.py
```diff
@@ -162,17 +157,19 @@ def get_time_string(date_time, display_sub_seconds=False,
         time_zone_string = TIME_ZONE_STRING_UTC
         if date_time_is_local:
             date_time = date_time - timedelta(
-                hours=TIME_ZONE_LOCAL_UTC_OFFSET_HOURS,
-                minutes=TIME_ZONE_LOCAL_UTC_OFFSET_MINUTES
+                hours=get_local_time_zone()[0],
+                minutes=get_local_time_zone()[1]
```
MetRonnie, line 161: `get_local_time_zone()` ought to be called only once at the start of this function.

cylc/flow/wallclock.py
```diff
@@ -149,8 +145,8 @@ def get_time_string(date_time, display_sub_seconds=False,
         else:
             custom_string = custom_time_zone_info["string_extended"]
         if date_time_is_local:
-            date_time_hours = TIME_ZONE_LOCAL_UTC_OFFSET_HOURS
-            date_time_minutes = TIME_ZONE_LOCAL_UTC_OFFSET_MINUTES
+            date_time_hours = hours
+            date_time_minutes = minutes
```
MetRonnie, line 149:
```suggestion
            date_time_hours, date_time_minutes = local_tz
```

cylc/flow/wallclock.py
```diff
@@ -162,17 +158,19 @@ def get_time_string(date_time, display_sub_seconds=False,
         time_zone_string = TIME_ZONE_STRING_UTC
         if date_time_is_local:
             date_time = date_time - timedelta(
-                hours=TIME_ZONE_LOCAL_UTC_OFFSET_HOURS,
-                minutes=TIME_ZONE_LOCAL_UTC_OFFSET_MINUTES
+                hours=hours,
+                minutes=minutes
```
MetRonnie, line 162:
```suggestion
                hours=local_tz[0],
                minutes=local_tz[1]
```

cylc/flow/wallclock.py
```diff
@@ -162,17 +158,19 @@ def get_time_string(date_time, display_sub_seconds=False,
...
+                TimeZoneFormatMode.reduced)
         else:
-            time_zone_string = TIME_ZONE_STRING_LOCAL_EXTENDED
+            time_zone_string = get_local_time_zone_format(
+                TimeZoneFormatMode.extended)
         if not date_time_is_local:
-            diff_hours = TIME_ZONE_LOCAL_UTC_OFFSET_HOURS
-            diff_minutes = TIME_ZONE_LOCAL_UTC_OFFSET_MINUTES
+            diff_hours = hours
+            diff_minutes = minutes
```
MetRonnie, line 173:
```suggestion
            diff_hours, diff_minutes = local_tz
```

changes.d/6719.fix.md
```diff
@@ -0,0 +1 @@
+Fix bug causing time string to be wrong after a clock change.
```
MetRonnie, line 1:
```suggestion
Fix incorrect job submission time reported after a time zone change.
```

tests/integration/test_workflow_db_mgr.py
```diff
@@ -178,3 +179,36 @@ async def test_record_only_non_clock_triggers(
...
+    (time_submit, time_submit_exit), = db_select(
+        schd, False, 'task_jobs', 'time_submit', 'time_submit_exit'
+    )
+    time_submit = datetime.strptime(time_submit, '%Y-%m-%dT%H:%M:%S%z')
+    time_submit_exit = datetime.strptime(
+        time_submit_exit, '%Y-%m-%dT%H:%M:%S%z'
+    )
+
+    # If submission time is > 10 seconds we may have other problems:
+    assert time_submit_exit >= time_submit
```
MetRonnie, line 213: Still a bit confused by this comment. Do you mean:
```suggestion
    assert time_submit_exit >= time_submit
    # The two times should be approx the same:
```

tests/unit/test_wallclock.py
```diff
@@ -60,3 +63,13 @@ def test_get_unix_time_from_time_string_360(time_str, time_sec):
 def test_get_unix_time_from_time_string_error(value, error):
     with pytest.raises(error):
         get_unix_time_from_time_string(value)
+
+
+def test_get_current_time_string(set_nonexistent_timezone):
+    """Now doesn't change timezones if system time is
+    different to scheduler's timezone.
```
MetRonnie, line 70: Huh?
```suggestion
    """It reacts to local time zone changes.
```

tests/integration/test_workflow_db_mgr.py
```diff
@@ -178,3 +179,36 @@ async def test_record_only_non_clock_triggers(
...
+    db_select,
+):
+    """Don't store scheduler startup timezone forever.
+
+    https://github.com/cylc/cylc-flow/issues/6701
+    """
+    wid = flow(one_conf)
+    schd = scheduler(wid, paused_start=False, run_mode='live')
+    async with run(schd):
+        await complete(schd, timeout=20)
```
MetRonnie, line 200: Unfortunately the integration test running in live mode is falling foul of MacOS runner performance, looks like the job is taking more than 20s to run. But actually we don't need the job or workflow to run, can instead just use `async with start(schd)` and prod the task through the submission process manually by `schd.submit_task_jobs()` or something.

oliver-sanders:
I suspect it's an issue with the test? Maybe the setting of the time zone isn't quite right?

tests/conftest.py
```diff
@@ -202,3 +203,17 @@ def _call(*args, **kwargs):
...
 
     return _capcall
+
+
+@pytest.fixture
+def set_nonexistent_timezone(monkeypatch):
+    """Fixture to temporarily set a non-existent time zone."""
+    def patch():
+        with monkeypatch.context() as mp:
+            mp.setenv('TZ', 'XXX-19:17')  # Set to a non-existent time zone
```
MetRonnie, line 213: Might as well make this 
```suggestion
def set_timezone(monkeypatch):
    """Fixture to temporarily set a system time zone."""
    def patch(time_zone: str):
        with monkeypatch.context() as mp:
            mp.setenv('TZ', time_zone)
```

MetRonnie:
> I suspect it's an issue with the test? Maybe the setting of the time zone isn't quite right?

The timezone change in the test looks right to me. The problem seems to be the bug still happens on MacOS

tests/conftest.py
wxtim: I agree - Easy reviewing twhen we are thinking the same thing, though I was going to make `time_zone` a kwarg with your `19:17` value as default.

wxtim:
> > I suspect it's an issue with the test? Maybe the setting of the time zone isn't quite right?
> 
> The timezone change in the test looks right to me. The problem does indeed seem to be the bug still happens on MacOS despite the attempt at fixing

It is, I've somehow pushed the bug to `time_submit_exit`, but only on Mac. 🤮

tests/conftest.py
```diff
@@ -202,3 +203,19 @@ def _call(*args, **kwargs):
...
+    Will use a very implausible timezone if none is provided.
+    """
+    def patch(time_zone: str = 'XXX-19:17'):
+        monkeypatch.setenv('TZ', time_zone)
+        time.tzset()
+
+    try:
+        yield patch
+    finally:
+        time.tzset()  # Reset to the original time zone after the test
```
MetRonnie, line 221: Argh, unfortunately I have debugged this and found that the TZ env var is remaining set when `tzset()` is called in the `finally` block, so this does not reset the TZ on Linux.

However, the solution is simple:
```suggestion
        monkeypatch.undo()
        time.tzset()  # Reset to the original time zone after the test
```

tests/integration/test_workflow_db_mgr.py
```diff
@@ -178,3 +179,39 @@ async def test_record_only_non_clock_triggers(
...
+        schd, False, 'task_jobs', 'time_submit', 'time_submit_exit'
+    )
+    time_submit = datetime.strptime(time_submit, '%Y-%m-%dT%H:%M:%S%z')
+    time_submit_exit = datetime.strptime(
+        time_submit_exit, '%Y-%m-%dT%H:%M:%S%z'
+    )
+
+    assert time_submit_exit >= time_submit
+    # The two times should be approx the same:
+    assert time_submit_exit < time_submit + timedelta(seconds=10)
```
MetRonnie, line 217: The use of `schd.task_events_mgr.process_message(itask, 'INFO', 'submitted')` is a bit artificial as it is just getting the current time, whereas normally this is called with a timestamp that comes from the jobs-submit context.

However, come to think of it, the `time_submit_exit` is not important, the fact of the matter is that the `time_submit` was bugged and that's all we need to test. So I would boil this test down to:

```suggestion
    set_timezone('XXX-19:00')
    schd = scheduler(flow(one_conf), paused_start=False, run_mode='live')
    async with start(schd):
        itask = schd.pool.get_tasks()[0]
        now = datetime.now().astimezone()
        set_timezone('XXX-19:17')
        schd.submit_task_jobs([itask])

    # Check the db time_submit:
    (time_submit,) = db_select(schd, False, 'task_jobs', 'time_submit')[0]
    time_submit = datetime.strptime(time_submit, '%Y-%m-%dT%H:%M:%S%z')
    # The submit time should be approx correct:
    assert (
        abs(time_submit - now) < timedelta(seconds=10)
    ), f"{time_submit} ~= {now}"
```

MetRonnie (approved):

tests/unit/test_wallclock.py
```diff
@@ -60,3 +66,75 @@ def test_get_unix_time_from_time_string_360(time_str, time_sec):
...
+            id='dt_is_local',
+        ),
+        param(
+            datetime(2000, 12, 13, 15, 30, 12, 123456),
+            {
+                'custom_time_zone_info': {
+                    'hours': 0,
+                    'minutes': -20,
+                    'string_basic': 'XXX+00:20',
+                },
```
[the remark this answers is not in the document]
wxtim: Which I guess raises the question of whether we ought to remove unused stuff.

wxtim:
@MetRonnie  Requesting a re-review based on https://github.com/cylc/cylc-flow/pull/6719/commits/5ad2dfa6bb21d512e63037d70b3a343f67a00ea8

tests/unit/test_wallclock.py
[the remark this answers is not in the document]
wxtim: Also, it is used  - see the tests - it's just that it's referenced positionally.

wxtim:
@hjoliver - Review Poke

tests/unit/test_wallclock.py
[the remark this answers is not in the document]
MetRonnie: Erm, it just looks like you removed the kwarg from some functions but not all of them. Nothing to do with the tests.

cylc/flow/wallclock.py
dwsutherland: Well the time zone is recorded on the workflow object:
https://github.com/cylc/cylc-flow/blob/d5ac2e2b7bf349458a8723f1235013b1b301873e/cylc/flow/data_store_mgr.py#L697

So perhaps that'll need a delta or made a field that gets checked whenever other workflow deltas are created.

The job nodes just get passed the time by the events or job manager.

wxtim: That leads to a real possibility that the time zones in the GUI will be wrong on DST change?

MetRonnie: What is the use case for having in the GraphQL API the local system time zone on scheduler start (except not when the workflow is running in UTC mode)? Note this is not the [cycle point time zone](https://cylc.github.io/cylc-doc/latest/html/reference/config/workflow.html#flow.cylc[scheduler]cycle%20point%20time%20zone)

wxtim:
@hjoliver Review poke

wxtim:
@hjoliver Review poke

hjoliver (approved):
LGTM, sorry for the long delay @wxtim

PR #6721, opened 2025-04-10
Update datastore xtriggers on restart.

### LINKED ISSUE #6720: Xtriggers not updated in the datastore on restart
At restart, the task pool will correctly load satisfied xtriggers (i.e., so they don't need to be re-satisfied) but the datastore doesn't, so `cylc show` shows them as not satisfied.

Install an easy-to-manipulate xtrigger:
```python
from pathlib import Path

def xfile(fname):
    if Path(fname).is_file():
        return (
            True,
            {
                "thing": "stuff",
            }
        )
    else:
        return (
            False,
            {}
        )
```

```cylc
[scheduling]
    [[xtriggers]]
        xa = xfile("/tmp/a")
        xb = xfile("/tmp/b")
    [[graph]]
        R1 = """
           @xa => foo
           @xb => foo
        """
[runtime]
    [[foo]]
```

Run this, satisfy one xtrigger with `touch /tmp/a`, then shut the workflow down and restart it. Task `foo` will have `xa` satisfied (if you satisfy `xb` as well, `foo` will run) but `cylc show` says it is not satisfied.

hjoliver:
Close #6720 

Just needed one call to the datastore after loading xtriggers for restart.

**Check List**

### DIFF
--- a/changes.d/6721.fix.md
+++ b/changes.d/6721.fix.md
@@ -1,0 +1,1 @@
+Fixed a bug that could mark satisfied xtriggers as unsatisfied after a restart, in task queries.
--- a/cylc/flow/xtrigger_mgr.py
+++ b/cylc/flow/xtrigger_mgr.py
@@ -574,6 +574,8 @@ class XtriggerManager:
             LOG.info("LOADING satisfied xtriggers")
         sig, results = row
         self.sat_xtrig[sig] = json.loads(results)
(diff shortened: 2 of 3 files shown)


dwsutherland (approved):
LGTM 👍

oliver-sanders:
(@wxtim, maybe one to test along with your GUI side xtrigger work)

wxtim (approved):

PR #6722, opened 2025-04-10
Tui memory leak

oliver-sanders:
Fix a slow memory leak in Tui:
* Urwid has a cache (CanvasCache) that caches the render of each widget to speed up future render cycles.
* Tui is a dynamic tool, it creates a new TuiNode for each update.
* The old TuiNode and its child nodes were not being cleared out of the urwid cache preventing Python's garbage clearance from doing its job.
* This change explicitly clears the urwid cache for each Tui update.
* Preserving the cache post-update isn't really a performance boost for Tui as pretty much every node on the screen has changed anyway (besides the application header and footer). So this is likely of little consequence performance wise. The cache is still active between updates so things like expand/collapse operations will still benefit.
* There is an open issue in urwid for a similar issue with the cache so we may be able to drop this workaround one day.

How to test:

* The `memray` profiler didn't seem to pick up on this issue (I guess caches are unlikely to be noticed by an internal sampling profiler).
* I found easier to spot using an external profiler tool like the Python `memory_profiler`, but you can also spot it by counting objects using `pympler.muppy`.
* Start a workflow in paused mode.
* Launch Tui and open it on the given workflow.
* Find the lead process ID and attach your profiler tool (note the subprocess will have the PPID of the lead process), e.g. `mprof attach PID`.
* Leave it for ~60 seconds and stop your profiler, e.g. quit Tui and run `mprof graph $PWD/<data-file>`.
* It should reveal a steady ratcheting of memory before this change and a flat line after (nothing is changing in the workflow, so the line should be quite flat).

Example output from memory_profiler before:

before

And after:

after

**Check List**

- [ ] Tests are included - hard to test for memory leaks.

### DIFF
--- a/cylc/flow/tui/app.py
+++ b/cylc/flow/tui/app.py
@@ -22,13 +22,8 @@ from multiprocessing import Process
 import re
 
 import urwid
-try:
-    from urwid.widget import SelectableIcon
-except ImportError:
-    # BACK COMPAT: urwid.wimp
-    # From: urwid 2.0
-    # To: urwid 2.2
-    from urwid.wimp import SelectableIcon
+from urwid.canvas import CanvasCache
+from urwid.widget import SelectableIcon
 
 from cylc.flow.id import Tokens
 from cylc.flow.task_state import (
@@ -540,6 +535,11 @@ class TuiApp:
         # schedule the next run of this update method
         if self.loop:
             self.loop.set_alarm_in(self.UPDATE_INTERVAL, self.update)
+
+        # NOTE: prevent a memory leak by clearing out any caches that urwid
+        # may have accumulated for the previous TuiNode instance
+        # (and its child widgets)
+        CanvasCache.clear()
 
         return True
 
--- a/cylc/flow/tui/app.py
+++ b/cylc/flow/tui/app.py
@@ -515,8 +515,11 @@ class TuiApp:
         _, old_node = self.listbox.body.get_focus()
 
         # nuke the tree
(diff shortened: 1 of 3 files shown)


cylc/flow/tui/app.py
```diff
@@ -515,8 +510,11 @@ def update(self, *_):
         _, old_node = self.listbox.body.get_focus()
 
         # nuke the tree
-        self.tree_walker = urwid.TreeWalker(topnode)
-        self.listbox.body = self.tree_walker
+        if not (self.tree_walker):
+            self.tree_walker = urwid.TreeWalker(topnode)
+            self.listbox.body = self.tree_walker
+        else:
+            self.tree_walker.set_focus(topnode)
```
oliver-sanders, line 517: Unrelated, there is a nicer way to do this now.

setup.cfg
```diff
@@ -76,7 +76,7 @@ install_requires =
     pyzmq>=22
     importlib_metadata>=5.0; python_version < "3.12"
     # NOTE: exclude two urwid versions that were not compatible with Tui
-    urwid==2.*,!=2.6.2,!=2.6.3
+    urwid>=2.2,!=2.6.2,!=2.6.3,<3
```
oliver-sanders, line 79: Unrelated, drop support for urwid <2.2

hjoliver (approved):
Confirmed with a quick profiler run as per your instructions. Nice.

oliver-sanders:
@wxtim, could you give this a quick sanity check.

wxtim (approved):

Issue #5995, opened 2024-02-22
pytest-asyncio: current version is incompatible with module scoped event loop


oliver-sanders:
Issue to track a problem with pytest-asyncio which is preventing us from moving to the latest version.

https://github.com/cylc/cylc-flow/blob/4ac6f3d10500a8df6ddfca24c23b1ac7b040d7df/setup.cfg#L121-L122

Upstream issue: https://github.com/pytest-dev/pytest-asyncio/issues/706

See also: https://github.com/cylc/cylc-rose/issues/299

oliver-sanders:
Upstream issue now closed, so we should be able to upgrade?

PR #6726, opened 2025-04-14
tests: switch from pytest.fixture to pytest_asyncio.fixture for async code

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/5995
* Asynchronous fixtures should use `pytest_asyncio.fixture` rather than `pytest.fixture` when attempting to override the default event loop scope.
* In practice, this has to be done whenever the fixture scope is not "function" (which is the pytest_asyncio default).

Getting this change out the way now before they drop support for the status quo.

To test, find a differently scoped fixture and jam a breakpoint into the code to ensure that it is called the expected number of times.

E.g, add a breakpoint into `tests/integration/network/test_client.py::harness`, then run `pytest -n0 --dist=no tests/integration/network/test_client.py::test_graphql tests/integration/network/test_client.py::test_protobuf`. The breakpoint will only be hit once.

**Check List**


setup.cfg
```diff
@@ -118,6 +118,10 @@ tests =
     flake8>=3.0.0
     mypy>=0.910,<1.9
     # https://github.com/pytest-dev/pytest-asyncio/issues/706
+    # BACK COMAPAT: pytest-asyncio 0.21.2
+    # FROM: python 3
+    # TO: python 3.7
+    # URL: https://github.com/cylc/cylc-flow/pull/6726
     pytest-asyncio>=0.21.2,!=0.23.*
```
oliver-sanders, line 125: We can't drop 0.21 support just yet, because it's the last version to support Python 3.7 :(

oliver-sanders:
Note: This does not silence the Python warnings because setting the default loop scope (as recommended) appears to break our ability to override it. Other projects seem to be suppressing these warnings by the look of their GH issues. Soon they will finish this migration and the warnings will vanish.

MetRonnie:
According to the warnings:

> The event loop scope for asynchronous fixtures will default to the fixture caching scope. Future versions of pytest-asyncio will default the loop scope for asynchronous fixtures to function scope. Set the default fixture loop scope explicitly in order to avoid unexpected behavior in the future.

So for now, `@pytest_asyncio.fixture(scope="module")` is the same as `scope="module", loop_scope="module"`, but in the future this will change to `scope="module", loop_scope="function"`?

But this sounds like it will break, according to the [docs](https://pytest-asyncio.readthedocs.io/en/stable/reference/decorators/index.html#decorators):

> The loop_scope of a fixture can be chosen independently from its caching scope. However, **the event loop scope must be larger or the same as the fixture’s caching scope**. In other words, it’s possible to reevaluate an async fixture multiple times within the same event loop, but it’s not possible to switch out the running event loop in an async fixture.

Also, according to the [migration guide from v0.23](https://pytest-asyncio.readthedocs.io/en/stable/how-to-guides/migrate_from_0_23.html):

> Explicitly set the loop_scope of async fixtures by replacing occurrences of `@pytest.fixture(scope="…")` and `@pytest_asyncio.fixture(scope="…")` with `@pytest_asyncio.fixture(loop_scope="…", scope="…")` such that loop_scope and scope are the same.

So I'm rather confused

tests/unit/test_id_cli.py
```diff
@@ -42,7 +44,7 @@ def mock_exists(monkeypatch: pytest.MonkeyPatch):
     monkeypatch.setattr('pathlib.Path.exists', lambda *a, **k: True)
 
 
-@pytest.fixture(scope='module')
+@pytest_asyncio.fixture(scope='module')
```
MetRonnie, line 47: Are you supposed to change these when the fixture doesn't do anything async?

oliver-sanders: Should be harmless, but not necessary.

MetRonnie: It's just occurred to me that any `@pytest.fixture`s added on 8.4.x will likely make it into master unnoticed

oliver-sanders: Yup.

oliver-sanders:
Drafting this whilst we investigate whether there is another way.

oliver-sanders:
I'm going to close this now as there's nothing we can reasonably do to evade the deprecation warning (this change really isn't "reasonable"!).

**Summary:**

Our tests require the `loop_scope` to match the fixture's scope.

This is the default behaviour, but it's kinda deprecated without replacement.

Our hope is that a replacement will arrive with https://github.com/pytest-dev/pytest-asyncio/issues/1095

PR #6727, opened 2025-04-14
protobuf upb memory accumulation fix

dwsutherland:
closes https://github.com/cylc/cylc-uiserver/issues/682

no-fix:
image

fix:
image

**Check List**


cylc/flow/data_store_mgr.py
```diff
@@ -330,6 +341,9 @@ def apply_delta(key, delta, data):
                         'on update application: %s' % str(exc)
                     )
                     continue
+    # Clear memory accumulation
+    if key == WORKFLOW:
+        data[key] = reset_protobuf_object(MESSAGE_MAP[key], data[key])
```
dwsutherland, line 346: `PbWorkflow` is the only long-lived protobuf object in the store, all the other nodes/edges/objects are transitory (added then eventually pruned)..

oliver-sanders, line 346: Why do we only do this for WORKFLOW fields, not all fields?

cylc/flow/data_store_mgr.py
```diff
@@ -295,6 +295,17 @@ def runtime_from_config(rtconfig):
...
 
 
+def reset_protobuf_object(msg_class, msg_orig):
+    """Reset object to clear memory build-up."""
+    # See: https://github.com/protocolbuffers/protobuf/issues/19674
+    # The new message instantiation needs happen on a separate line.
+    new_msg = msg_class()
+    new_msg.ParseFromString(
+        msg_orig.SerializeToString(deterministic=True)
+    )
```
oliver-sanders, line 305: Ouch, we have to serialise and deserialise the whole workflow object every time something change in it?

That must be a performance hit?

cylc/flow/data_store_mgr.py
oliver-sanders: What is it about PbWorkflow that makes it long-lived by contrast to the other fields?

dwsutherland: Because we never delete it, we only add-to/update/prune-from fields.

dwsutherland: Because `PbWorkflow` the only long-lived field, whereas nodes/edges/etc will cause a little accumulation but usually have finite updates and existence (are pruned).. See:
https://github.com/cylc/cylc-uiserver/issues/682#issuecomment-2800872753

cylc/flow/data_store_mgr.py
dwsutherland: Well, I had to weigh up doing this instead of using the python implementation (i.e. `export PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python`):
- the underlying `upb` libraries are faster than `python` implementation at doing this..
- Also it's not a large object, and only one.
- when I timed it, it was ~0.0001sec before and after this change

dwsutherland: Essentially what it's doing is copying the content of the old message into a new message.. I'll try some alternatives, cause it may be enough just to create a new object and copy from the old...  One moment

(I just did it this way to ensure I wasn't transferring memory references from the old object)

dwsutherland: Well, what do you know.. It worked with existing tools (`CopyFrom`), changed to that.

cylc/flow/data_store_mgr.py
oliver-sanders: So updating **any** repeated field in protobuf causes a memory leak, but we are only really concerned by the workflow field due to its longevity?

changes.d/6727.fix.md
```diff
@@ -0,0 +1 @@
+A memory leak effecting both scheduler and UI server has been fixed.
```
MetRonnie, line 1:
```suggestion
Fixed a memory leak affecting both scheduler and UI server.
```

cylc/flow/data_store_mgr.py
oliver-sanders: Trying to work out whether it would make more sense to run this as a periodic activity (with a set interval), rather than doing it for every update (which could potentially be every main loop cycle):

E.g, like the health check routine: https://github.com/cylc/cylc-flow/blob/master/cylc/flow/main_loop/health_check.py

oliver-sanders:
I profiled using the "complex" workflow, this branch reduced memory usage by 1.36%. An improvement, but a fairly small one.

It sounds like you've been seeing much more dramatic memory usage? Have you got an example workflow that can replicate this?

dwsutherland:
> I profiled using the "complex" workflow, this branch reduced memory usage by 1.36% (134'672kb -> 132'844kb). An improvement, but a fairly small one.

How long was it run for? This isn't a improvement in the memory footprint, this is a accumulation of updates on the workflow object.. So

> It sounds like you've been seeing much more dramatic memory usage? Have you got an example workflow that can replicate this?

If any workflow was updating quickly and running for a long time, you should see a big difference..
It has a more major impact on the long running UIS, that may be running many of these workflows..

dwsutherland:
While this stabilizes the UIS without the UI open:
image

`apply_delta` does appear to accumulate memory when the graphql subscriptions kick off:
image
So there may be accumulation of `delta-stores` perhaps..


Changed this PR to "partially addresses" association to the UIS issue.

cylc/flow/data_store_mgr.py
dwsutherland: I think it's not that expensive to run on `PbWorkflow` alone .. Especially since we've swapped to `CopyFrom`, also, this solution will be adopted by the UIS without the need to implement the same here

cylc/flow/data_store_mgr.py
dwsutherland: Correct, "memory leak" for the most part on nodes would be changes in state or other fields.. But these typically only have a finite number of updates done to them, and are ultimately deleted..

The only other object I can think of is the "mean_elapse_time" on the `PbTask` objects... 
(now reset)

oliver-sanders:
> How long was it run for?

That's an old operational workflow run for 1 days worth of cycles with a 0 second task runtime. That represents lots and lots of updates. I would expect a sizeable reduction in the max RSS if there were a memory leak active.

> Multiply that by ~50 workflows, and that's 14Gb a day in accumulated memory.

This is where I'm confused.

We have ~70 Cylc 8 workflows running on a single server which only gets rebooted once a month, yet memory usage is fairly constant. I'm not seeing gigabytes worth of memory accumulation per month let alone per day.

oliver-sanders:
I took your example above, removed the `sleep`s and ran it in simulation mode (all that's needed to replicate this issue):

```cylc
[scheduler]
    UTC mode = True
    allow implicit tasks = True

[scheduling]
    initial cycle point = 20200101T00
    final cycle point = 20210101  # run for one year
    [[graph]]
        R1 = prep => foo
        PT12H = foo[-PT12H] => foo => bar

[runtime]
    [[root]]
        [[[simulation]]]
            default run length = PT0S
```

```console
$ cylc vip -N --mode=simulation
$ mprof attach <pid>
$ mprof plot <file>
```

Peak memory usage reduced by about 5%, but a clear memory leak remains visible.

### Before (upstream/master)

Image

### After (dwsutherland/protobuf-upb-mem-accumulation-fix)

Image

dwsutherland:
@oliver-sanders : sent a message on Element..

On last thing before I head away, could you try rerunning that test with:
`export PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python`
?

dwsutherland:
I was able to reproduce your accumulation:
image

At first I thought it was the simulation mode, because I ran without and `script = sleep 5` and it was flat:
image

however I reduced the sleep to `script = sleep 0` and was able to reproduce the problem again:
image

It doesn't appear to be the data-store:
image
We should really remove the memory check plugin from the profiler, but the only other thing I can see increasing is the proc pool:
image

(Is there something else in the plugins?)


Anyway, it's clear this a new and separate memory leak .. Will check with memray

dwsutherland:
Memray can't pick it up:
image

dwsutherland:
Scalene can't reproduce the leak either:
https://github.com/plasma-umass/scalene

image

Kinda hard to pin point if I can look more granular

dwsutherland:
I was able to find something growing with `muppy`:
```
(8.4.x) sutherlander@cortex-vbox:cylc-flow$ git diff
index 1b31e4bd9..833a28940 100644
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -177,6 +177,8 @@ from cylc.flow.workflow_status import (
 )
 from cylc.flow.xtrigger_mgr import XtriggerManager
 
+from pympler import muppy, summary
+
 
 if TYPE_CHECKING:
     from optparse import Values
@@ -190,6 +192,7 @@ if TYPE_CHECKING:
     from cylc.flow.network.resolvers import TaskMsg
 
+MUPPY_INTERVAL = 30
 
 class SchedulerStop(CylcError):
     """Scheduler normal stop."""
@@ -352,6 +355,8 @@ class Scheduler:
         # {install_target: platform}
         self.incomplete_ri_map: Dict[str, Dict] = {}
 
+        self.muppy_check_time = time()
+
     async def install(self):
         """Get the filesystem in the right state to run the flow.
         * Validate flowfiles
@@ -1892,6 +1897,14 @@ class Scheduler:
         else:
             duration = self.INTERVAL_MAIN_LOOP - elapsed
         await asyncio.sleep(duration)
+
+        muppy_elapsed = time() - self.muppy_check_time
+        if (muppy_elapsed >= MUPPY_INTERVAL):
+            all_objects = muppy.get_objects()
+            sum1 = summary.summarize(all_objects)
+            summary.print_(sum1)
+            self.muppy_check_time = time()
+
         # Record latest main loop interval
         self.main_loop_intervals.append(time() - tinit)
         # END MAIN LOOP
```
A start:
```
                       types |   # objects |   total size
============================ | =========== | ============
                         str |       38549 |      5.22 MB
                        dict |       15846 |      4.88 MB
                        code |       12243 |      2.09 MB
                        type |        1886 |      1.70 MB
                       tuple |       14261 |    861.78 KB
                        list |        2315 |    270.41 KB
                         set |         524 |    267.03 KB
     collections.OrderedDict |         435 |    222.30 KB
       weakref.ReferenceType |        2723 |    191.46 KB
                 abc.ABCMeta |         183 |    183.66 KB
          wrapper_descriptor |        2444 |    171.84 KB
                         int |        4229 |    127.37 KB
                  re.Pattern |         360 |    123.09 KB
  builtin_function_or_method |        1684 |    118.41 KB
         function (__init__) |         824 |    115.88 KB
```
The `tuple` and `dict` continues to grow with each snapshot:
```
                       types |   # objects |   total size
============================ | =========== | ============
                        dict |       15846 |      5.31 MB
                         str |       40070 |      5.31 MB
                        code |       12242 |      2.37 MB
                        type |        1886 |      1.70 MB
                       tuple |       28215 |      1.65 MB
                        list |        2342 |    312.70 KB
                         set |         524 |    267.53 KB
     collections.OrderedDict |         435 |    222.30 KB
       weakref.ReferenceType |        2723 |    191.46 KB
                 abc.ABCMeta |         183 |    183.66 KB
          wrapper_descriptor |        2444 |    171.84 KB
                         int |        4956 |    147.25 KB
                  re.Pattern |         360 |    123.09 KB
  builtin_function_or_method |        1684 |    118.41 KB
         function (__init__) |         824 |    115.88 KB
```

I'll try drill down further

oliver-sanders:
> I was able to reproduce your accumulation:
>
> At first I thought it was the simulation mode, because I ran without and script = sleep 5 and it was flat:
>
> however I reduced the sleep to script = sleep 0 and was able to reproduce the problem again:

Interesting. Perhaps this suggests that the memory leak occurs when multiple state changes happen within the same main loop cycle? Perhaps a result of updating a pre-existing delta?

<br />

> It doesn't appear to be the data-store:

If this is a protobuf memory leak, it's likely too low-level for internal profilers like pympler or memray to pick up on. The memory_profiler works by polling the OS for memory usage so will pick up on everything (irrespective of what level of the stack the memory is associated with).

So, unfortunately, this may or may not be true.

<br />

> I was able to find something growing with muppy:
>
> The tuple and dict continues to grow with each snapshot:

Unfortunately, that is probably the profiler results accumulating?

dwsutherland:
>  Perhaps this suggests that the memory leak occurs when multiple state changes happen within the same main loop cycle? Perhaps a result of updating a pre-existing delta?

Not sure about that, the deltas get discarded.. It would seem to point to some race condition in Cylc, but we just fixed the known ones (the accumulating jobs and families problem)

> Unfortunately, that is probably the profiler results accumulating?

Can't be, you can see from the diff (above), it's a new snapshot every time (no long running profiler)..

I'm having a pretty hard time isolating the problem:
```
                       types |   # objects |   total size
============================ | =========== | ============
                         str |       38555 |      5.22 MB
                        dict |       15841 |      4.88 MB
                        code |       12243 |      2.10 MB
                        type |        1886 |      1.70 MB
                       tuple |       14272 |    862.45 KB
                        list |        2311 |    270.04 KB
                         set |         524 |    267.28 KB
     collections.OrderedDict |         435 |    222.30 KB
       weakref.ReferenceType |        2723 |    191.46 KB
                 abc.ABCMeta |         183 |    183.66 KB
          wrapper_descriptor |        2444 |    171.84 KB
                         int |        4229 |    127.37 KB
                  re.Pattern |         360 |    123.09 KB
  builtin_function_or_method |        1684 |    118.41 KB
         function (__init__) |         824 |    115.88 KB
------------------
Schd: 493.56 KB
config: 261.368 KB
Task Pool: 36.952 KB
Proc Pool: 1.68 KB
Workflow DB Mgr: 49.016 KB
Broadcast Mgr: 0.864 KB
Task Events Mgr: 1.576 KB
Task Job Mgr: 70.28 KB
Xtrigger Mgr: 2.824 KB
Flow Mgr: 0.992 KB
Data Store Mgr: 35.264 KB
Server: 105.72 KB
data - size: 6808  length: 1
deltas - size: 1800  length: 7
added - size: 1848  length: 7
updated - size: 1848  length: 7
delta_queues - size: 368  length: 1
xtrigger_tasks - size: 64  length: 0
n_window_nodes - size: 1968  length: 3
n_window_node_walks - size: 9136  length: 3
n_window_depths - size: 1200  length: 2
db_load_task_proxies - size: 64  length: 0
prune_trigger_nodes - size: 1176  length: 2
------------------
```
after:
```
                       types |   # objects |   total size
============================ | =========== | ============
                        dict |       16103 |      5.35 MB
                         str |       40071 |      5.31 MB
                        code |       12242 |      2.38 MB
                        type |        1886 |      1.70 MB
                       tuple |       27837 |      1.63 MB
                        list |        2336 |    309.06 KB
                         set |         524 |    267.53 KB
     collections.OrderedDict |         435 |    222.30 KB
       weakref.ReferenceType |        2723 |    191.46 KB
                 abc.ABCMeta |         183 |    183.66 KB
          wrapper_descriptor |        2444 |    171.84 KB
                         int |        5141 |    152.66 KB
                  re.Pattern |         360 |    123.09 KB
  builtin_function_or_method |        1684 |    118.41 KB
         function (__init__) |         824 |    115.88 KB
------------------
Schd: 552.64 KB
config: 321.256 KB
Task Pool: 37.072 KB
Proc Pool: 1.68 KB
Workflow DB Mgr: 49.016 KB
Broadcast Mgr: 0.864 KB
Task Events Mgr: 1.576 KB
Task Job Mgr: 70.28 KB
Xtrigger Mgr: 2.824 KB
Flow Mgr: 0.992 KB
Data Store Mgr: 35.872 KB
Server: 106.384 KB
data - size: 7216  length: 1
deltas - size: 1800  length: 7
added - size: 1848  length: 7
updated - size: 1848  length: 7
delta_queues - size: 368  length: 1
xtrigger_tasks - size: 64  length: 0
n_window_nodes - size: 1968  length: 3
n_window_node_walks - size: 9136  length: 3
n_window_depths - size: 1200  length: 2
db_load_task_proxies - size: 64  length: 0
prune_trigger_nodes - size: 1176  length: 2
------------------
```
no comparable object found yet

> If this is a protobuf memory leak, it's likely too low-level for internal profilers like pympler or memray to pick up on.

You did reproduce the `mprof` accumulation with `export PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python`, so not sure about that.

I might try the above with that, just in case..
Failing that, I will try recreate the data-store every minute, which should clearly show if the problem is there.

dwsutherland:
By the way, this fix is still valid.. it fixed the long lived protobuf issue.. So is good to go WRT that, and also fixes the UIS end in this regard too.

dwsutherland:
What we've found is a niche problem with fast jobs (it appears)

dwsutherland:
So it doesn't appear to be data-store related, as re-creating the data-store with this reload modification:
```
diff --git a/cylc/flow/commands.py b/cylc/flow/commands.py
index b05bb9660..c6b3aad44 100644
--- a/cylc/flow/commands.py
+++ b/cylc/flow/commands.py
@@ -79,6 +79,7 @@ from cylc.flow.exceptions import (
     CylcConfigError,
 )
 import cylc.flow.flags
+from cylc.flow.data_store_mgr import DataStoreMgr
 from cylc.flow.flow_mgr import get_flow_nums_set
 from cylc.flow.log_level import log_level_to_verbosity
 from cylc.flow.network.schema import WorkflowStopMode
@@ -412,7 +413,12 @@ async def reload_workflow(schd: 'Scheduler'):
         schd._update_workflow_state()
 
         # Re-initialise data model on reload
+        schd.data_store_mgr = DataStoreMgr(schd)
         schd.data_store_mgr.initiate_data_model(schd.is_reloaded)
+        schd.xtrigger_mgr.data_store_mgr = schd.data_store_mgr
+        schd.task_job_mgr.data_store_mgr = schd.data_store_mgr
+        schd.task_events_mgr.data_store_mgr = schd.data_store_mgr
+        schd.pool.data_store_mgr = schd.data_store_mgr
 
         # Reset the remote init map to trigger fresh file installation
         schd.task_job_mgr.task_remote_mgr.remote_init_map.clear()
```
doesn't reset the accumulation:
image

dwsutherland:
This PR is about the `upb` fix, which it does fix, and given the data-store doesn't appear to be a cause of the `sleep 0`/simulation accumulation (both `upb` and `python` implementations),
image

See https://github.com/cylc/cylc-flow/pull/6727#issuecomment-2841164577

and because I've wasted so much time banging my head against a wall trying to find it, I'm going to make this PR live and open a separate issue for the other problem..

dwsutherland:
fast job memory growth spun off to a separate issue:
https://github.com/cylc/cylc-flow/issues/6751

oliver-sanders:
I have had a go a profiling this against the "complex workflkow". The results are a little surprising:

| protobuf implementation | branch | max RSS (mb) | elapsed time | CPU [user + sys] (s) |
| --- | --- | --- | --- | --- |
| upb | upstream/master | 158.604 | 1:04:44 | 706.30 |
| upb | dwsutherland/protobuf-upb-mem-accumulation-fix | 154.136<br />(-2.8%) | 1:13:11<br />(+13%) | 717.10<br />(+1.5%) |
| python | upstream/master | 158.604 | 1:03:38 | 838.72 |
| python | dwsutherland/protobuf-upb-mem-accumulation-fix | 121.640<br />(-23%) | 1:04:10<br />(+5.5%) | 835.07<br />(-0.4%) |

Conclusions:
* For a real workflow, the memory saving with the UPB backend are small (2.8%).
* Yet the savings for the Python backend are much higher. This doesn't make sense as the Python backend is not supposed to be affected by this.
* The CPU impact is within the jitter of the results, but may indicate a noticeable increase, would require further tests.

hjoliver:
@dwsutherland can you give a quick summary of where we are at with this PR, without the "sleep 0" case that you've spun off to another issue (I think that problem was contaminating the results above until you realized it was something different?).

dwsutherland:
> For a real workflow, the memory saving with the UPB backend are small (2.8%).

We have to remember, this about memory accumulation over time..

> The CPU impact is within the jitter of the results, but may indicate an increase, would require repeat tests.

Note the CPU difference between `upb` and `python` implementations, serialization/de-serialization is mean to be where `upb` is suppose to win..


> @dwsutherland can you give a quick summary of where we are at with this PR, without the "sleep 0" case that you've spun off to another issue (I think that problem was contaminating the results above until you realized it was something different?).

@hjoliver - This PR fixes the contribution from `upb` Protobuf in the memory accumulation over time, and is complete in that regard. (and ready to go)

The spin off issue was shown to be nothing to do with Protobuf/data-store.

hjoliver:
(Note my speculation on memory fragmentation in the spinoff PR #6751 , and Element chat).

oliver-sanders:
I don't think memory fragmentation is causing us issues. Whether it is or isn't, the UBP memory leak David has identified is very much real, though I haven't managed to quantify the amount just yet.

dwsutherland:
> > can you give a quick summary of where we are at with this PR
> 
> FYI: Summary of where I'm at with review (sorry, it's going slow ATM):
> 
> * [ ]  Need to confirm this PR fixes all of the UPB memory leak.
>   

I've seen the memory for protobuf objects stabilize with this fix, I've targeted all the long-lived objects that get updates.. Should be enough
Also the memory now acts the same as the `export PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python`

>   * Should try to confirm the "sleep 0" thing is actually a different issue.
>   * I haven't managed to convince myself that this memory accumulation is _not_ being caused by the data store.
> * [ ]  Need to confirm this PR doesn't cost too much CPU.

I think you showed that it's better than the python implementation:
https://github.com/cylc/cylc-flow/pull/6727#issuecomment-2879403425
I'm pretty confident that creating a new object and copying from existing isn't that expensive, the fields aren't that big in themself


>   * There's an easy solution if it is.
> * [ ]  Could do with getting a handle on how much memory this is actually leaking.
>   

I believe this is the main leak that's plaguing us at NIWA, I can't confirm that easy until I implement it there.. but it makes sense that this is only multiplied by 57 at our UIS and that we go from  3Gb to 30Gb in ~3-4 days

dwsutherland:
@oliver-sanders - you can play around with this to get an idea of mem/cpu impact:
```
#!/usr/bin/python

from time import sleep

from cylc.flow.data_messages_pb2 import PbWorkflow


wbuf = PbWorkflow(id='this')
wbuf2 = PbWorkflow(id='that', status='sprinting')
while True:
    wbuf.status_msg = 'workflow is running'
    wbuf.MergeFrom(wbuf2)
    sleep(0.00001)
```
(careful it can eat up your memory very quickly without the sleep)
And then the fixed version:
```
#!/usr/bin/python

from time import sleep

from cylc.flow.data_messages_pb2 import PbWorkflow
from cylc.flow.data_store_mgr import reset_protobuf_object


wbuf = PbWorkflow(id='this')
wbuf2 = PbWorkflow(id='that', status='sprinting')
while True:
    wbuf.status_msg = 'workflow is running'
    wbuf.MergeFrom(wbuf2)
    wbuf = reset_protobuf_object(PbWorkflow, wbuf)
    sleep(0.00001)
```
**Note:** realistically we only do this every 1.5sec (main loop) if there's an update

dwsutherland:
Another note, out of all the data types:
```
DATA_TEMPLATE = {
    EDGES: {},
    FAMILIES: {},
    FAMILY_PROXIES: {},
    JOBS: {},
    TASKS: {},
    TASK_PROXIES: {},
    WORKFLOW: PbWorkflow(),
}
```
Only `TASKS`, `FAMILIES`, and `WORKFLOW` are long lived, the others are transitory (added then pruned)..
All three of these have fields on them that change, `WORKFLOW` most of all.
`TASKS` - the mean elapse time, and proxy references are altered.
`FAMILIES` - proxy references.

The other protobuf objects like deltas are cleared after application:
```
    def clear_delta_batch(self):
        """Clear current deltas."""
        # Potential shared reference, avoid clearing
        self.deltas = {
            EDGES: EDeltas(),
            FAMILIES: FDeltas(),
            FAMILY_PROXIES: FPDeltas(),
            JOBS: JDeltas(),
            TASKS: TDeltas(),
            TASK_PROXIES: TPDeltas(),
            WORKFLOW: WDeltas(),
        }

    def clear_delta_store(self):
        """Clear current delta store."""
        # Potential shared reference, avoid clearing
        self.added = deepcopy(DATA_TEMPLATE)
        self.updated = deepcopy(DATA_TEMPLATE)
```

So only `TASKS`, `FAMILIES`, and `WORKFLOW` are renewed (`new.CopyFrom(old)`)

MetRonnie (changes requested):

changes.d/6727.fix.md
dwsutherland: reapplied

oliver-sanders (approved):
I've run some tests overnight.

The fastest I could get memory to accumulate was 10mb per hour, which works out at 120mb per day, 1.2gb per week, etc. 

<table>
<tbody>
<tr>
<td>

before

</td>
<td>

after

</td>
</tbody>
</table>

However, this depends on workflow events occurring at the fastest rate possible. Real workflows accumulate memory much more slowly. Our longest running workflows are not the highest memory users at our site and we have several workflows that have been running for ~2 weeks, but are only using ~50mb (which is about the Cylc baseline memory usage) showing no signs of a memory leak.

memory

So, I'm not convinced that this memory leak *alone* can explain the high memory usage that you've reported which is annoying. Further investigations are likely required :(

I've done some CPU profiling, the `CopyFrom` call chain isn't showing up too clearly in profiling. It's unnecessary to do this for every update, but as long as it's harmless there's no reason not to do this. If it ever shows up in profiling, we can always shift it into a main loop plugin.

This PR definitely identifies and fixes the peculiar UPB leak :+1:, let's get it in.

MetRonnie (approved):
LGTM

PR #6733, opened 2025-04-16
psutil: allow accessing psutil object methods

oliver-sanders:
We have recently seen a case where the `_is_process_running` check would timeout every time for a certain user.

Why only this one user is affected is something I do not understand, however, it's clear that the `cylc psutil` call made by this function is returning way too much data, we only need one field!

Reviewers, please ensure this change is back-compatible, note you can invoke the `cylc psutil` command like so:

```
# return the process listing for PID=1
$ cylc psutil <<< '[["Process", 1]]'
```

To active this logic in a functional test, try:

```
cylc vip --pause <x>
cylc vr --yes <x>
```

You might want to use this patch to confirm that the logic is being called and the cmdline is being returned as expected:

```diff
diff --git a/cylc/flow/workflow_files.py b/cylc/flow/workflow_files.py
index add79adbb..64e748e47 100644
--- a/cylc/flow/workflow_files.py
+++ b/cylc/flow/workflow_files.py
@@ -459,6 +459,7 @@ def _is_process_running(
             f'\n{command}'
         )
 
+    print('#', cli_format(process['cmdline']), command)
     return cli_format(process['cmdline']) == command
 
 
```

---

* The `_is_process_running` check was requesting the `psutil.Process` object via the `cylc psutil` command.
* It would appear that these objects can get large enough to fill up the buffer. This means that `subprocess.Process.communicate` or `.wait` will hang indefinitely unless something reads from stdout to clear out the buffers.
* This commit changes the `is_process_running` check to request only the one field it actually requires, `psutil.Process.cmdline` to reduce the data volume to a level where pipe polling is not required.

---

**Check List**

### DIFF
--- a/cylc/flow/scripts/psutil.py
+++ b/cylc/flow/scripts/psutil.py
@@ -46,25 +46,48 @@ def get_option_parser():
     return OptionParser(__doc__)
 
 
+def _call_psutil_interface(interface, *args):
+    """Call a psutil interface with the provided arguments.
+
+    Args:
+        interface:
+            The psutil method we want to call, e.g. "cpu_percent".
+
+            In the case of objects, this may include attribures, e.g.
+            "Process.cmdline".
+        args:
+            The arguments to provide to the psutil method.
+
+    Returns:
+        The result of the psutil method call.
+
+    """
+    result = psutil
+    is_first = True
+    for fcn in interface.split('.'):
+        try:
+            method = getattr(result, fcn)
+        except AttributeError as exc:
+            # error obtaining interfaces from psutil e.g:
+            # * requesting a method which does not exist
+            print(exc, file=sys.stderr)
+            sys.exit(2)
+
+        if is_first:
+            args = args
+            is_first = False
+        else:
+            args = ()
+
+        result = method(*args)
+    return result
+
+
 def _psutil(metrics_json):
     metrics = parse_dirty_json(metrics_json)
 
     try:
-        methods = [
-            getattr(psutil, key[0])
-            for key in metrics
-        ]
-    except AttributeError as exc:
-        # error obtaining interfaces from psutil e.g:
-        # * requesting a method which does not exist
-        print(exc, file=sys.stderr)
-        sys.exit(2)
-
-    try:
-        ret = [
-            method(*key[1:])
-            for key, method in zip(metrics, methods)
-        ]
+        ret = [_call_psutil_interface(*key) for key in metrics]
     except Exception as exc:
         # error extracting metrics from psutil e.g:
         # * requesting information on a resource which does not exist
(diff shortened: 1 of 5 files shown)


wxtim:
> This commit changes the is_process_running check to request only the one field it actually requires, psutil.Process.cmdline to reduce the data volume to a level where pipe polling is not required.

I can't see evidence of this when I try:

```
@@ -427,6 +427,7 @@ def _is_process_running(
     try:
         # Terminate command after 10 seconds to prevent hanging, etc.
         out, err = proc.communicate(timeout=10, input=metric)
+        print(out)
     except TimeoutExpired:
```

oliver-sanders:
Do me mean that code is not being called? If so, you need to tweak your example so that it is. Make sure the workflow is still running when you call the `cylc vr` command, I think that's enough, but it might take something else.

wxtim:
> Do me mean that code is not being called? If so, you need to tweak your example so that it is. Make sure the workflow is still running when you call the `cylc vr` command, I think that's enough, but it might take something else.

No, it's that I can't see any changes to `is_process_running`, and when I look at the output of it's subprocess call to `cylc psutil` I don't see any difference.

oliver-sanders:
That's because I'm a clot who forgets to commit his changes!

wxtim (approved):

wxtim:
You have a broken test.

wxtim:
> That's because I'm a clot who forgets to commit [his changes](https://github.com/cylc/cylc-flow/compare/3212535dd4a97303bf8bd03fe5da00c735d186c4..7ca9daba4266c452687f73e7822a35d2d83bc9e5)!

I'd love to say I've _never_ done anything like this. Although I usually forget to push!

cylc/flow/scripts/psutil.py
```diff
@@ -46,25 +46,48 @@ def get_option_parser():
...
 
 
+def _call_psutil_interface(interface, *args):
+    """Call a psutil interface with the provided arguments.
+
+    Args:
+        interface:
+            The psutil method we want to call, e.g. "cpu_percent".
+
+            In the case of objects, this may include attribures, e.g.
```
MetRonnie, line 56:
```suggestion
            In the case of objects, this may include attributes, e.g.
```

cylc/flow/scripts/psutil.py
```diff
@@ -46,25 +46,48 @@ def get_option_parser():
...
+        try:
+            method = getattr(result, fcn)
+        except AttributeError as exc:
+            # error obtaining interfaces from psutil e.g:
+            # * requesting a method which does not exist
+            print(exc, file=sys.stderr)
+            sys.exit(2)
+
+        if is_first:
+            args = args
```
MetRonnie, line 77:
```suggestion
```

MetRonnie (approved):
Two niggles

MetRonnie (approved):

PR #6745, opened 2025-05-02
Run hosts random-selection bug

MetRonnie:
This fixes a bug where remote reinvocation would exit 0 even if it failed when `[scheduler][run hosts]ranking` is empty (which means random host selection).
- Now it exits non-zero if it failed.
- It also keeps trying hosts if any are uncontactable, until it either finds one that is contactable or exhausts all listed run hosts. (Contact-ability is checked by an empty `cylc psutil <<< '[]'` call on the host)
- Also fixes traceback if an invalid run host is specified in the global config

#### Steps to reproduce

```cylc
# global.cylc
[scheduler]
    [[run hosts]]
        available = somehost1, somehost2, somehost3
        ranking =  # (empty!)
```
```ini
# ~/.ssh/config
Host somehost*
  User borked
```
```console
$ cylc play AnyWorkflow

 ▪ ■  Cylc Workflow Engine 8.4.2
 ██   Copyright (C) 2008-2025 NIWA

borked@somehost3: Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
$ echo $?
0
```

#### After fix

```console
$ cylc play AnyWorkflow

 ▪ ■  Cylc Workflow Engine 8.4.3.dev
 ██   Copyright (C) 2008-2025 NIWA

WARNING - Could not contact somehost2.example.com:
    borked@somehost2.example.com: Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
WARNING - Could not contact somehost1.example.com:
    borked@somehost1.example.com: Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
WARNING - Could not contact somehost3.example.com:
    borked@somehost3.example.com: Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
HostSelectException: Could not select host from:
    somehost1.example.com
        returncode: 255
    somehost2.example.com
        returncode: 255
    somehost3.example.com
        returncode: 255

Cylc could not establish SSH connection to the run hosts.
Ensure you can SSH to these hosts without having to answer any prompts.
$ echo $?
1
```

#### Check List

- [x] No dependency changes 
- [x] No docs needed

### DIFF
--- a/cylc/flow/exceptions.py
+++ b/cylc/flow/exceptions.py
@@ -397,30 +397,34 @@ class PlatformLookupError(CylcConfigError):
 
 
 class HostSelectException(CylcError):
-    """No hosts could be selected from the provided configuration."""
+    """No hosts could be selected from the provided configuration.
 
-    def __init__(self, data: Dict[str, dict]):
+    Args:
+        data: Mapping of hostnames to error info.
+        ranking: The ranking expression used to select the host.
+    """
+
+    def __init__(self, data: Dict[str, dict], ranking: Optional[str] = None):
         self.data = data
+        self.ranking = ranking
 
     def __str__(self) -> str:
         ret = 'Could not select host from:'
         for host, data in sorted(self.data.items()):
-            if host != 'ranking':
-                ret += f'\n    {host}:'
-                for key, value in data.items():
-                    ret += f'\n        {key}: {value}'
+            ret += f'\n    {host}:'
+            for key, value in data.items():
+                ret += f'\n        {key}: {value}'
         hint = self.get_hint()
         if hint:
             ret += f'\n\n{hint}'
         return ret
 
-    def get_hint(self):
+    def get_hint(self) -> Optional[str]:
         """Return a hint to explain this error for certain cases."""
         if all(
             # all procs came back with special SSH error code 255
             datum.get('returncode') == 255
-            for key, datum in self.data.items()
-            if key != 'ranking'
+            for datum in self.data.values()
         ):
             # likely SSH issues
             return (
@@ -431,18 +435,17 @@ class HostSelectException(CylcError):
 
         if (
             # a ranking expression was used
-            self.data.get('ranking')
+            self.ranking
             # and all procs came back with special 'cylc psutil' error code 2
             # (which is used for errors relating to the extraction of metrics)
             and all(
                 datum.get('returncode') == 2
-                for key, datum in self.data.items()
-                if key != 'ranking'
+                for datum in self.data.values()
             )
         ):
             # likely an issue with the ranking expression
             lines = wrap(
-                self.data.get("ranking"),
+                self.ranking,
                 initial_indent='    ',
                 subsequent_indent='    ',
             )
--- a/cylc/flow/host_select.py
+++ b/cylc/flow/host_select.py
@@ -260,8 +260,10 @@ def select_host(
         # no metrics or ranking required, pick host at random
         random.shuffle(hosts)
         for host in hosts:
-            # check host is contactable
-            if _get_metrics([host], [], data):
+            if (not is_remote_host(host)) or (
+                # check host is contactable
+                _get_metrics([host], [], data)
+            ):
                 return hostname_map[host], host
         raise HostSelectException(data)
 
--- a/cylc/flow/host_select.py
+++ b/cylc/flow/host_select.py
@@ -255,10 +255,12 @@ def select_host(
 
     if not rankings:
         # no metrics or ranking required, pick host at random
-        hosts = [random.choice(list(hosts))]  # nosec
-
-    if not rankings and len(hosts) == 1:
-        return hostname_map[hosts[0]], hosts[0]
+        random.shuffle(hosts)
+        for host in hosts:
+            # check host is contactable
+            if _get_metrics([host], [], data):
+                return hostname_map[host], host
+        raise HostSelectException(data)
 
     # filter and sort by rankings
     metrics = list({x for x, _ in rankings})  # required metrics
@@ -592,10 +594,11 @@ def _get_metrics(hosts, metrics, data):
             out, err = (stream.strip() for stream in proc.communicate())
             if proc.wait():
                 # Command failed
-                LOG.warning(
-                    'Error evaluating ranking expression on'
-                    f' {host}: \n{err}'
+                msg = (
+                    'Could not contact' if proc.returncode == 255
+                    else 'Error evaluating ranking expression on'
                 )
+                LOG.warning(f'{msg} {host}:\n{err}')
             else:
                 host_stats[host] = dict(zip(
                     metrics,
--- a/cylc/flow/host_select.py
+++ b/cylc/flow/host_select.py
@@ -67,15 +67,17 @@ from functools import lru_cache
 from io import BytesIO
 import json
 import random
+from socket import gaierror
 from time import sleep
 import token
 from tokenize import tokenize
 from typing import (
-    Any,
     Dict,
     Iterable,
     List,
     Optional,
+    Set,
+    Tuple,
 )
 
 from cylc.flow import LOG
@@ -133,8 +135,6 @@ def select_workflow_host(cached=True):
     Raises:
         HostSelectException:
             See `select_host` for details.
-        socket.gaierror:
-            See `select_host` for details.
 
     """
     # get the global config, if cached = False a new config instance will
@@ -164,11 +164,11 @@ def select_workflow_host(cached=True):
 
 
 def select_host(
-    hosts,
-    ranking_string=None,
-    blacklist=None,
-    blacklist_name=None
-):
+    hosts: List[str],
+    ranking_string: Optional[str] = None,
+    blacklist: Optional[Iterable[str]] = None,
+    blacklist_name: Optional[str] = None,
+) -> Tuple[str, str]:
     """Select a host from the provided list.
 
     If no ranking is provided (in `ranking_string`) then random selection
@@ -196,10 +196,10 @@ def select_host(
                getloadavg()
 
             Comments are allowed using `#` but not inline comments.
-        blacklist (list):
+        blacklist:
             List of host names to filter out.
             Can be short host names (do not have to be fqdn values)
-        blacklist_name (str):
+        blacklist_name:
             The reason for blacklisting these hosts
             (used for exceptions).
 
@@ -207,38 +207,42 @@ def select_host(
         HostSelectException:
             In the event that no hosts are available / meet the specified
             criterion.
-        socket.gaierror:
-            This may be raised in the event of unknown host names
-            for some installations or not for others.
+            This may also be raised in the event of unknown host names.
 
     Returns:
         tuple - (hostname, fqdn) the chosen host
 
-        hostname (str):
+        hostname:
             The hostname as provided to this function.
-        fqdn (str):
+        fqdn:
             The fully qualified domain name of this host.
 
     """
-    # standardise host names - remove duplicate items
-    hostname_map = {  # note dictionary keys filter out duplicates
-        get_fqdn_by_host(host): host
-        for host in hosts
-    }
-    hosts = list(hostname_map)
-
     # dict of conditions and whether they have been met (for error reporting)
-    data = {
-        host: {}
-        for host in hosts
-    }
+    data: Dict[str, dict] = {}
+
+    # standardise host names - remove duplicate items
+    hostname_map = {}  # note dictionary keys filter out duplicates
+    for host in hosts:
+        try:
+            hostname_map[get_fqdn_by_host(host)] = host
+        except gaierror as exc:
+            data.setdefault(host, {})[type(exc).__name__] = str(exc)
+    hosts = list(hostname_map)
 
     # filter out `filter_hosts` if provided
     if blacklist:
-        blacklist = list(set(map(get_fqdn_by_host, blacklist)))
-        _filter_by_hostname(
+        blacklist_fqdns: Set[str] = set()
+        for host in blacklist:
+            try:
+                blacklist_fqdns.add(get_fqdn_by_host(host))
+            except gaierror as exc:
+                LOG.warning(
+                    f'Could not resolve blacklisted host {host}: {exc}'
+                )
+        hosts = _filter_by_hostname(
             hosts,
-            blacklist,
+            blacklist_fqdns,
             blacklist_name,
             data=data
         )
@@ -251,7 +255,6 @@ def select_host(
     if ranking_string:
         # parse rankings
         rankings = list(_get_rankings(ranking_string))
-        data['ranking'] = ranking_string
 
     if not rankings:
         # no metrics or ranking required, pick host at random
@@ -271,7 +274,7 @@ def select_host(
     # stop here if we don't need to proceed
     if not hosts:
         # no hosts provided / left after filtering
-        raise HostSelectException(data)
+        raise HostSelectException(data, ranking_string)
     if not rankings and len(hosts) == 1:
         return hostname_map[hosts[0]], hosts[0]
 
@@ -285,18 +288,18 @@ def select_host(
 
     if not hosts:
         # no hosts provided / left after filtering
-        raise HostSelectException(data)
+        raise HostSelectException(data, ranking_string)
 
     return hostname_map[hosts[0]], hosts[0]
 
 
 def _filter_by_hostname(
-    hosts: List[str],
+    hosts: Iterable[str],
     blacklist: Iterable[str],
     blacklist_name: Optional[str],
-    data: Dict[str, Any],
-) -> None:
-    """Filter out (in-place) any hosts present in `blacklist`.
+    data: Dict[str, dict],
+) -> List[str]:
+    """Return hosts, having filtered out any present in `blacklist`.
 
     Args:
         hosts:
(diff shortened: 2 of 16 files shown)


cylc/flow/host_select.py
```diff
@@ -243,15 +255,17 @@ def select_host(
...
-        hosts = [random.choice(list(hosts))]  # nosec
-
-    if not rankings and len(hosts) == 1:
-        return hostname_map[hosts[0]], hosts[0]
+        random.shuffle(hosts)
+        for host in hosts:
+            # check host is contactable
+            if _get_metrics([host], [], data):
+                return hostname_map[host], host
+        raise HostSelectException(data)
```
MetRonnie, line 263: Thoughts on testing this? A unit test would likely end up with lots of mocking

MetRonnie:
Test failure `tests/functional/workflow-host-self-id/00-address.t` seems to be an uncovered bug:

```python
>>> from cylc.flow.hostuserutil import is_remote_host, get_host_ip_by_name
>>> is_remote_host(get_host_ip_by_name('localhost'))  # -> 127.0.0.1
True
>>> is_remote_host(get_host_ip_by_name(get_host()))  # -> network-accessible IP address
True
```

MetRonnie:
Potential fix: https://github.com/MetRonnie/cylc-flow/pull/36

wxtim:
I can't replicate this:

```bash
> cylc play throwaway/rare.beach/run7

 ▪ ■  Cylc Workflow Engine 8.4.3.dev
 ██   Copyright (C) 2008-2025 NIWA

  File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/scheduler_cli.py", line 695, in cylc_play
    asyncio.get_running_loop()
    ~~~~~~~~~~~~~~~~~~~~~~~~^^

  File "/home/users/tim.pillinger/conda-envs/cylc313/bin/cylc", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/scripts/cylc.py", line 702, in main
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/scripts/cylc.py", line 333, in execute_cmd
    entry_point.load()(*args)
    ~~~~~~~~~~~~~~~~~~^^^^^^^
  File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/terminal.py", line 298, in wrapper
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/scheduler_cli.py", line 675, in play
    cylc_play(options, id_)
    ~~~~~~~~~^^^^^^^^^^^^^^
  File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/scheduler_cli.py", line 698, in cylc_play
    scheduler, workflow_id = asyncio.run(
                             ~~~~~~~~~~~^
        _scheduler_cli_1(options, id_, parse_workflow_id=parse_workflow_id)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/users/tim.pillinger/conda-envs/cylc313/lib/python3.13/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/home/users/tim.pillinger/conda-envs/cylc313/lib/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/users/tim.pillinger/conda-envs/cylc313/lib/python3.13/asyncio/base_events.py", line 721, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/scheduler_cli.py", line 411, in _scheduler_cli_1
    _distribute(workflow_id_raw, workflow_id, options)
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/scheduler_cli.py", line 611, in _distribute
    host = options.host or select_workflow_host()[0]
                           ~~~~~~~~~~~~~~~~~~~~^^
  File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/host_select.py", line 138, in select_workflow_host
    return select_host(
        # list of workflow hosts
    ...<9 lines>...
        blacklist_name='condemned host'
    )
  File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/host_select.py", line 212, in select_host
    get_fqdn_by_host(host): host
    ~~~~~~~~~~~~~~~~^^^^^^
  File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/hostuserutil.py", line 266, in get_fqdn_by_host
    return HostUtil.get_inst().get_fqdn_by_host(target)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/hostuserutil.py", line 172, in get_fqdn_by_host
    return self._get_host_info(target)[0]
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/hostuserutil.py", line 136, in _get_host_info
    self._host_exs[target] = socket.gethostbyname_ex(target)
                             ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
socket.gaierror: [Errno -2] Name or service not known: 'somehost1'

```

cylc/flow/host_select.py
wxtim: I don't like the bash test battery, but this is probably a case for needing it. I wonder if you could create your own script in the test dir, call it `ssh` and add it to `$PATH` as a very low level form of mocking?

oliver-sanders: Would need a functional test with `REQUIRE_PLATFORM='loc:remote fs:shared'.

Could do a trivial single task workflow and `run_ok "${TEST_NAME_BASE}" cylc play --no-detach --mode=simulation`.

oliver-sanders:
Whilst this is a good way to fix the problem (because it makes random selection tolerant of dead hosts). I feel that the real issue here, is that the `cylc play` command fails to exit non-zero if the detached process exits non-zero.

Note: For context, when this code was written, it made sense to differentiate between "host does not exist" and "could not contact host" because the former is a site-configuration issue (that cannot be resolved by users), and the latter a user connection issue (that can). As we move into the world of elastic services, this distinction is less useful (as hosts may be taken down when idle, or listed before being provisioned), so don't worry about breaking this paradigm.

cylc/flow/scheduler_cli.py
```diff
@@ -641,8 +646,9 @@ def _distribute(
         # Re-invoke the command
         # NOTE: has the potential to raise NoHostsError, however, this will
         # most likely have been raised during host-selection
-        cylc_server_cmd(cmd, host=host)
-        sys.exit(0)
+        sys.exit(
+            cylc_server_cmd(cmd, host=host)
+        )
```
MetRonnie, line 651: > I feel that the real issue here, is that the `cylc play` command fails to exit non-zero if the detached process exits non-zero immediately (at the distribution stage).

@oliver-sanders Does this changed line not address this?

cylc/flow/host_select.py
MetRonnie: How do I ensure the host is not contactable?

MetRonnie: How do I make sure the hosts are uncontactable? Note they have to be valid hosts

oliver-sanders: One option would be to alter the configured `ssh command`. E.g, we could change it to `ssh command = ssh -F /var/tmp/mock.ssh`.

MetRonnie: This is what I have tried:
```sh
# create an SSH config that will fail to connect to a valid host
mock_ssh_config_file="/var/tmp/cylc-test-ssh-config"
cat > "${mock_ssh_config_file}" << __EOF__
Host badger*
    HostName localhost
    User BorkedMushroom25
__EOF__

[scheduler]
    [[run hosts]]
        available = badger1, badger2  # Two hosts minimum to ensure all are tried
        ranking =
[platforms]
    [[localhost]]
        ssh command = ssh -F ${mock_ssh_config_file}
"
```

Unfortunately this is not sufficient as Python's `socket` library does not recognise the `badger` hostnames, so the test does not get the right error.

oliver-sanders: Try breaking a connection that would work rather than one that wouldn't (badger).

E.G, in the functional tests, you can ask for a remote platform, this hostname will be known (!= :badger:) so should suffice for testing.

cylc/flow/scheduler_cli.py
oliver-sanders: Sorry, I'm not quite sure what I meant there, plz ignore.

cylc/flow/host_select.py
MetRonnie: How do I get 2 different hostnames?

oliver-sanders: localhost + remote test host

https://github.com/cylc/cylc-flow/blob/4082f43032e4e5a5f6034761d1925a44a031d78d/tests/functional/restart/34-auto-restart-basic.t#L36-L38

cylc/flow/remote.py
```diff
@@ -102,13 +126,13 @@ def run_cmd(
...
           successfully.
-        * Else True if the remote command is executed successfully, or
-          if unsuccessful and capture_status=True the remote command exit code.
-        * Otherwise exit with an error message.
+        * Else if capture_status=True: the remote command exit code.
+        * Else if the remote command is executed successfully: 0.
+        * Else exits with an error message instead of returning.
 
-    Exits with code 1 in the event of certain command errors.
+    Exits with an error message in the event of certain command errors.
```
oliver-sanders: There is no such thing as "exit with an error message", Python just dumps the msg to stderr and exits 1.

```suggestion
    Exits with code 1 in the event of certain command errors.
```

I'm not sure what the "certain command errors" were, but with your changes, isn't this:


```suggestion
    Exits with code 1 in the event of command error.
```

cylc/flow/main_loop/auto_restart.py
```diff
@@ -157,20 +157,19 @@ def _should_auto_restart(scheduler, current_glbl_cfg):
...
+    except Exception as exc:
         # Any unexpected error in host selection shouldn't be able to take
         # down the workflow.
-        LOG.critical(
-            'Workflow cannot automatically restart because:\n' +
-            'Error in host selection:\n' +
-            traceback.format_exc())
+        LOG.critical(f"{err_msg}: Error in host selection.")
+        LOG.exception(exc)
         return False
```
oliver-sanders: Unrelated to bugfix?

The `log.critical` message was better before (one message at CRITICAL level, now two messages one at CRITICAL, one at ERROR).

cylc/flow/host_select.py
MetRonnie: Can't use localhost as the hosts have to look like remote hosts

cylc/flow/scheduler_cli.py
```diff
@@ -608,7 +608,7 @@ def _distribute(
 
     """
     # Check whether a run host is explicitly specified, else select one.
-    host = options.host or select_workflow_host()[0]
+    host = options.host or select_workflow_host()[1]
```
oliver-sanders: What's the reason for this change?

cylc/flow/exceptions.py
```diff
@@ -397,30 +397,34 @@ class PlatformLookupError(CylcConfigError):
...
+    """
+
+    def __init__(self, data: Dict[str, dict], ranking: Optional[str] = None):
         self.data = data
+        self.ranking = ranking
 
     def __str__(self) -> str:
         ret = 'Could not select host from:'
         for host, data in sorted(self.data.items()):
-            if host != 'ranking':
```
oliver-sanders, line 408: You've removed a bunch of guards against "ranking".

Do you understand what these were intended to do?

cylc/flow/host_select.py
oliver-sanders: Taking a second look at this code, surely we only need one host to test this (shuffle aside)?

cylc/flow/main_loop/auto_restart.py
MetRonnie: > Unrelated to bugfix?

This was in response to https://github.com/cylc/cylc-flow/pull/6745#discussion_r2075371964

> It might be better to catch `gaierror` in `cylc.flow.select_host.select_host` and re-raise as `HostSelectException` (then no need to catch and log for each individual usage).

Because now `gaierror` is replaced by `HostSelectException` the details would be lost without including the error message here (L165).

cylc/flow/scheduler_cli.py
MetRonnie: This is changing to use the FQDN - we have already obtained it but we were making `is_remote_host` on the next line look it up again.

cylc/flow/main_loop/auto_restart.py
oliver-sanders: I don't think this block is in relation to that comment and the exception was already logged, this diff has just rejigged some log messages to taste and split the latter in two?

cylc/flow/exceptions.py
MetRonnie: I found it confusing that ranking information was included in a mapping of hostnames to exception messages. It's been pulled out into a separate field on `HostSelectException`.

cylc/flow/scheduler_cli.py
oliver-sanders: Sounds like this is unrelated to the bugfix?

Changing to FQDN here could have consequences because it changes the re-invocation argument. I think this should be reverted.

cylc/flow/main_loop/auto_restart.py
MetRonnie: I've undone the split if that's what you're talking about.

cylc/flow/scheduler_cli.py
oliver-sanders: This is a regression of a bugfix - 506da4732b9dff1af894d0f1fdb51d656e69c0f8

cylc/flow/main_loop/auto_restart.py
oliver-sanders: Yes. The only other change here was the text of the log message (unrelated).

cylc/flow/host_select.py
MetRonnie: Using one host would at least test the exit code part of the bug, but the other part was that it would only try one host instead of keeping on trying until it found one that was good

oliver-sanders: Ach right. Yes, that's not something you'll be able to test unless you can make localhost fail the is remote host check, OR make an alias of the remote host not resolve to the same FQDN. Both would require mocking.

cylc/flow/scheduler_cli.py
MetRonnie: Reverted, but note this is inconsistent with `_get_metrics()` where we call `cylc_server_cmd()` with the FQDN

cylc/flow/host_select.py
MetRonnie: Ok, I've pushed the test with one host configured

tests/flakyfunctional/restart/39-auto-restart-no-suitable-host.t
```diff
@@ -64,9 +64,9 @@ ${BASE_GLOBAL_CONFIG}
 FILE=$(cylc cat-log "${WORKFLOW_NAME}" -m p |xargs readlink -f)
 log_scan "${TEST_NAME_BASE}-no-auto-restart" "${FILE}" 20 1 \
     'The Cylc workflow host will soon become un-available' \
-    'Workflow cannot automatically restart because:' \
+    'Workflow cannot automatically restart: No alternative host' \
     'No alternative host to restart workflow on.' \
-    'Workflow cannot automatically restart because:' \
+    'Workflow cannot automatically restart: No alternative host' \
     'No alternative host to restart workflow on.'
```
oliver-sanders: This is now grepping for the `No alternative host` part twice for each event.

Since this change removed a `\n` character from the message, these two parts are now on the same line. The fact this test still passes means that this test is no longer valid as a single `Workflow cannot automatically restart: No alternative host to restart workflow on` line could satisfy all four tests here (whereas it looks like it is designed to ensure the event occurs twice).

I think you can workaround this by including a search for the following line:

```
Workflow cannot automatically restart: No alternative host to restart workflow on
Traceback
Workflow cannot automatically restart: No alternative host to restart workflow on
Traceback
```

MetRonnie: Ah, oops. But the test is still valid, it's just waiting for 4 auto restart messages now before stopping the workflow, instead of 2. I've changed it back to 2 in latest force push

oliver-sanders:
Have manually run the auto-restart tests:

* tests/functional/restart/41-auto-restart-local-jobs.t passes on this PR (with these fixes https://github.com/cylc/cylc-flow/pull/6776)
* tests/functional/restart/42-auto-restart-ping-pong.t appears to be broken on this PR.
* All other tests passed.

oliver-sanders:
I have figured out why the test was failing, before no SSH was being performed because no ranking was defined. After an SSH (and `cylc psutil`) was performed because this PR performs the test, even for random ranking. The FQDN wasn't in my keys so the test failed.

The new SSH is a design feature, so long as it is called the same way (FQDN) that it is when rankings are involved (which it is), there isn't a problem here.

oliver-sanders (approved):

wxtim (approved):

PR #6753, opened 2025-05-09
duplicate xtrigger data-store fix

### LINKED ISSUE #6747: xtriggers: duplicate xtrigger labels should all be reported
Tasks can have multiple xtrigger triggers which reference the same underlying function call, e.g:

```cylc
[scheduling]
  [[xtriggers]]
    true = xrandom(100)
    false = xrandom(0)
    false2 = xrandom(0)
  [[graph]]
    R1 = """
      @false => foo
      @true => foo
      @false2 => foo
    """

[runtime]
  [[foo]]
```

This works fine, the xtrigger manager de-duplicates them under the hood, so the `xrandom(0)` call is only made once.

However, only one of the duplicate labels is present in the data store:

```console
$ cylc show 'tmp.8wWcobsLXP//*'
title: (not given)
description: (not given)
URL: (not given)
state: waiting
prerequisites: (None)
outputs: ('⨯': not completed)
  ⨯ 1/foo expired
  ⨯ 1/foo submitted
  ⨯ 1/foo submit-failed
  ⨯ 1/foo started
  ⨯ 1/foo succeeded
  ⨯ 1/foo failed
other: ('⨯': not satisfied)
  ⨯ xtrigger "false2 = xrandom(0)"
  ✓ xtrigger "true = xrandom(100)"
output completion: incomplete
  ⨯ ┆  succeeded
```

After https://github.com/cylc/cylc-flow/pull/6695, this could lead to confusion as, by manually setting xtrigger prereqs, it is possible to end up in the situation where `cylc show` reports all of the task's prerequsites as satisfied, yet the task is still waiting on an xtrigger.

To resolve this, we should include duplicate xtrigger labels in the data store.

For context, see point (5) in [this comment](https://github.com/cylc/cylc-flow/pull/6695#issuecomment-2838082634), and the discussion thereafter.

(note, I've named this as an "xtrigger" issue as opposed to a "data store" issue, as the deduplication of xtriggers will be happening in the xtrigger manager class so the fix will likely need to change that code, not just the data store).

dwsutherland:
closes #6747

**Check List**


cylc/flow/data_store_mgr.py
```diff
@@ -1520,7 +1521,7 @@ def _process_internal_task_proxy(
         for label, satisfied in itask.state.xtriggers.items():
             sig = self.schd.xtrigger_mgr.get_xtrig_ctx(
                 itask, label).get_signature()
-            xtrig = tproxy.xtriggers[sig]
+            xtrig = tproxy.xtriggers[f'{sig}{XTRIGGER_DELIM}{label}']
```
hjoliver, line 1524: This is a better solution than my initial suggestion (mapping signature to a list of labels, which would require changing the schema). But just wondering:
- for maximum clarity, might we might as well use `=` as the delimiter, and key the map on `{label}={signature}`
- is it slightly inefficient that the information in the `tproxy.xtriggers` map values are now almost entirely redundant with the keys (could we just store boolean `satisfied` instead of the `xtrig` object?)

dwsutherland: > * for maximum clarity, might we might as well use `=` as the delimiter, and key the map on `{label}={signature}`

The only trouble with using `=` or something common, is you can't deconstruct the ID easily  (because the signature can contain a plethora of characters).. Although if we have the label on the left, it's probably ok

> * is it slightly inefficient that the information in the `tproxy.xtriggers` map values are now almost entirely redundant with the keys (could we just store boolean `satisfied` instead of the `xtrig` object?)

We could, but then we'd need to change a lot more downstream.. Also, perhaps there's more information we could add? (i.e. message from the latest attempt of the xtrigger?)

dwsutherland: > * for maximum clarity, might we might as well use `=` as the delimiter, and key the map on `{label}={signature}`

Done

hjoliver:
Unfortunately that broke some tests!

dwsutherland:
> Unfortunately that broke some tests!

Yeah, for some reason we now have duplicates:
image

Will have to figure out what the changes did with my rebase

dwsutherland:
@oliver-sanders - I found a bug with https://github.com/cylc/cylc-flow/pull/6671 .. There's no fields in the `data_messages.proto` , yet you've generated the code for them

dwsutherland:
Found the duplication problem, and fixed

hjoliver:
> @oliver-sanders - I found a bug with #6671 .. There's none of the new fields in the `data_messages.proto` , yet you've generated the code for them

Huh, yeah - we missed that in review, sorry. Oliver must have missed a `git add` there.

hjoliver (approved):

hjoliver:
(Putting this to 8.5 - it's small, easy to review, and pretty much necessary to have it in 8.5 since #6695 got merged)

Issue #4148, opened 2021-03-26
stream log events


oliver-sanders:
Add log events into the data store / schema.

This is intended to allow the UI to display warnings when workflows require user intervention in order to continue - https://github.com/cylc/cylc-ui/issues/632

We can do this by creating a new log handler that creates deltas in the data store rather than writing to a file/stream and attaching that to the scheduler log. I think we should configure the handler for the logging level "warning".

This could be implemented in the same way as the latest task statuses as a dequeue listing the last N warnings. We don't need to maintain a comprehensive list of all warnings.

> **Historical Context:**
>
> In the Cylc7 GUI there is a warning icon in the bottom right-hand corner which lights up whenever
> a "warning" or higher is written to the log (I think it's warning, it might be "error" would have to look).
> The user can click on the icon to open the log / dismiss it.
>
> In the Cylc7 GScan there is a warning icon next to each workflow which appears next to the suite whenever
> a task enters the failed or submit-failed statues. The user can then click on the icon to dismiss it.

**Pull requests welcome!**

PR #6755, opened 2025-05-12
GraphQL: stream log events

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/4148
* Unblocks https://github.com/cylc/cylc-ui/issues/535
* Stream log events at the warning level or above over Protobuf/GraphQL for use in the UIs.

**Check List**


dwsutherland (approved):
LGTM 🚀 
image

cylc/flow/data_store_mgr.py
```diff
@@ -2280,6 +2297,20 @@ def _generate_broadcast_node_deltas(self, node_data, node_type):
...
 
+    def delta_log_record(self, record: 'LogRecord') -> None:
+        w_delta = self.updated[WORKFLOW]
+        w_delta.id = self.workflow_id
+        w_delta.last_updated = time()
+        w_delta.stamp = f'{w_delta.id}@{w_delta.last_updated}'
+
+        if not hasattr(w_delta, 'log_records'):
+            w_delta.log_records = []
+        w_delta.log_records.append(
```
dwsutherland, line 2308: You may want to add this as a clear field entry:
https://github.com/cylc/cylc-flow/blob/0b1ec93d370066ee9fb15cd3b2d3c02259a1e088/cylc/flow/data_store_mgr.py#L196-L209

Otherwise it will pile up on the `PbWorkflow` objects when applied to the data-store

dwsutherland (changes requested):

cylc/flow/data_store_mgr.py
dwsutherland: The UI can just hold these messages until dismissed (or something)

dwsutherland: After making this change:
```
index 40b879655..e1514156f 100644
@@ -205,7 +205,7 @@ CLEAR_FIELD_MAP = {
     JOBS: {'messages'},
     TASKS: set(),
     TASK_PROXIES: {'prerequisites'},
-    WORKFLOW: {'latest_state_tasks', 'state_totals', 'states'},
+    WORKFLOW: {'latest_state_tasks', 'state_totals', 'states', 'log_records'},
 }
```
The log entries disappear when refreshing the page (as expected)

image

Otherwise they persist

wxtim (approved):
Please note @dwsutherland's change - I've had a really decent practical play around with this change, and it seems to work. The code also makes sense.

cylc/flow/data_store_mgr.py
oliver-sanders: > The log entries disappear when refreshing the page (as expected)

Ideally they would not disappear, but also would not pile up.

At the cylc-ui end of this PR, we only store the 10 most recent log messages (a dequeue). Would it be possible to override the way this field is updated in the data store so that it only stores the last N messages?

hjoliver: Ping @dwsutherland ⬆️

hjoliver (commented):
(Code looks good, happy to approve this as soon as the incremental update question is settled).

cylc/flow/data_store_mgr.py
oliver-sanders: Ok, I've had a quick play, but haven't had the time to do anything proper yet.

I think I can shift the deque mechanism form the cylc-ui end to the cylc-flow end. I.E, cylc-flow would only retain the last N log entries in memory. When the UIS is restarted, the log messages are persisted (because they're held in scheduler memory), when the workflow is restarted the log messages are cleated (correctly).

This diff almost does the trick but will cause all log messages to be cleared unless there is at least one log message in the delta:

```diff
diff --git a/cylc/flow/data_store_mgr.py b/cylc/flow/data_store_mgr.py
index 40b879655..c25ca7448 100644
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -205,7 +205,7 @@ CLEAR_FIELD_MAP = {
     JOBS: {'messages'},
     TASKS: set(),
     TASK_PROXIES: {'prerequisites'},
-    WORKFLOW: {'latest_state_tasks', 'state_totals', 'states'},
+    WORKFLOW: {'latest_state_tasks', 'state_totals', 'states', 'log_records'},
 }
 
 
@@ -2303,12 +2303,18 @@ class DataStoreMgr:
         w_delta.last_updated = time()
         w_delta.stamp = f'{w_delta.id}@{w_delta.last_updated}'
 
-        if not hasattr(w_delta, 'log_records'):
-            w_delta.log_records = []
+        if len(w_delta.log_records) == 0:
+            w_delta.log_records.extend(
+                self.data[self.workflow_id][WORKFLOW].log_records
+            )
+
         w_delta.log_records.append(
             PbLogRecord(level=record.levelname, message=record.message)
         )
 
+        if len(w_delta.log_records) > 5:
+            w_delta.log_records.pop(0)
+
         self.updates_pending = True
 
     # -----------
```

dwsutherland: Looks like you have a solution, yes this would work to keep the last 5

cylc/flow/data_store_mgr.py
```diff
@@ -311,17 +336,23 @@ def apply_delta(key, delta, data):
...
+
+            # Apply the updated delta
             data[key].MergeFrom(delta.updated)
+
+            # Deque (i.e. length restrict) selected fields
+            for field_name, max_len in DEQUE_FIELD_MAP.get(key, {}).items():
+                if field_name in field_set:
+                    lst = getattr(data[key], field_name)
+                    while len(lst) > max_len:
+                        lst.pop(0)
```
oliver-sanders, line 348: This block will ensure the data store only holds the last 10 log messages.

oliver-sanders:
(rebased)

cylc/flow/data_store_mgr.py
```diff
@@ -320,16 +345,24 @@ def apply_delta(key, delta, data):
...
         else:
             for element in delta.updated:
                 try:
                     data_element = data[key][element.id]
                     # Clear fields that require overwrite with delta
-                    if CLEAR_FIELD_MAP[key]:
```
dwsutherland, line 329: The reason I had this is to avoid iterating over the fields unnecessarily if there are no fields flagged for clearing for this type/key.

dwsutherland: However, that's only `Families`, `Tasks`, `Edges` ... So probably not a big deal

dwsutherland (approved):
LGTM, tested with front end, only keeps last 10 now 👍

cylc/flow/data_store_mgr.py
wxtim: Consider adding to code comment?

wxtim:
@oliver-sanders : Have rechecked - happy for you to merge this. Only refrained from merging because unhelpful GH message makes it unclear if you want Hilary to check.

```
if (you want Hilary to check is False):
  merge
```

cylc/flow/data_store_mgr.py
oliver-sanders: Already done:

```
# Deque (i.e. length restrict) selected fields
```

oliver-sanders:
> Only refrained from merging because unhelpful GH message makes it unclear if you want Hilary to check.

Click on the list of reviewers to see who's actually assigned and who has just commented.

PR #6765, opened 2025-05-23
demote unnecessary warnings

### LINKED ISSUE #6760: Demote uneccessary warnings
These warnings can be demoted to avoid popping up orange triangles after https://github.com/cylc/cylc-flow/pull/6755.

* [19/foo/01:running] active with pre-reload settings & batch?
* More than 100 tasks, circular dep checking skipped

wxtim:
Closes #6760 

https://github.com/cylc/cylc-ui/pull/2169 adds streaming warnings to the UI, but this highlights some warnings which may be either very large, or of no interest to the user.

I am aware that this change may engender controversy - feel free to persuade me that my choices are wrong.

**Check List**

- [x] Refactor and log level change - should be covered by existing tests, and indeed a number broke.
- [x] ~Changelog entry included if this is a change that can affect users~ No: These changes are being made because these warning should NOT affect users.
- [x] ~[Cylc-Doc](https://github.com/cylc/cylc-doc)~ N/A
- [x] Not a bug fix => master

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -914,10 +914,10 @@ class WorkflowConfig:
         """Check for circular dependence in graph."""
         if (len(self.taskdefs) > self.CHECK_CIRCULAR_LIMIT and
                 not getattr(self.options, 'check_circular', False)):
-            LOG.warning(
+            LOG.info(
                 f"Number of tasks is > {self.CHECK_CIRCULAR_LIMIT}; will not "
-                "check graph for circular dependencies. To enforce this "
-                "check, use the option --check-circular.")
+                "check graph for circular dependencies. To run this check"
+                " anyway use the option --check-circular.")
             return
         start_point_str = self.cfg['scheduling']['initial cycle point']
         raw_graph = self.get_graph_raw(start_point_str,
--- a/cylc/flow/rundb.py
+++ b/cylc/flow/rundb.py
@@ -535,7 +535,7 @@ class CylcWorkflowDAO:
                 )
                 raise
             self.n_tries += 1
-            LOG.warning(
+            LOG.info(
                 "%(file)s: write attempt (%(attempt)d)"
                 " did not complete: %(error)s\n"
                 " SQLite error code: %(error_code)s\n"
@@ -560,7 +560,7 @@ class CylcWorkflowDAO:
                 table.update_queues.clear()
             # Report public database retry recovery if necessary
             if self.n_tries:
-                LOG.warning(
+                LOG.info(
                     "%(file)s: recovered after (%(attempt)d) attempt(s)\n" % {
                         "file": self.db_file_name, "attempt": self.n_tries})
             self.n_tries = 0
@@ -601,7 +601,10 @@ class CylcWorkflowDAO:
             for i, stmt_args in enumerate(stmt_args_list):
                 err_log += ("\nstmt_args[%(i)d]=%(stmt_args)s" % {
                     "i": i, "stmt_args": stmt_args})
-            LOG.warning(err_log)
+            if self.is_public:
+                LOG.info(err_log)
+            else:
+                LOG.warning(err_log)
             raise
 
     def pre_select_broadcast_states(self, order=None):
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -1375,7 +1375,7 @@ class TaskEventsManager():
         """Helper for process_message, handle a started message."""
         if itask.job_vacated:
             itask.job_vacated = False
-            LOG.warning(f"[{itask}] Vacated job restarted")
+            LOG.info(f"[{itask}] Vacated job restarted")
         job_tokens = itask.tokens.duplicate(job=str(itask.submit_num))
         self.data_store_mgr.delta_job_time(job_tokens, 'started', event_time)
         self.data_store_mgr.delta_job_state(job_tokens, TASK_STATUS_RUNNING)
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1169,7 +1169,9 @@ class TaskPool:
                     # Job file might have been written at this point?
                     _warn_tasks.append(str(itask))
 
-        for may, tasks in (('', warn_tasks), ('may be', warn_tasks)):
+        for may, tasks in [('', n) for n in warn_tasks] + [
+            ('may be', n) for n in _warn_tasks
+        ]:
             if tasks:
                 _tasks = "\n * ".join(tasks)
                 LOG.info(
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1169,9 +1169,7 @@ class TaskPool:
                     # Job file might have been written at this point?
                     _warn_tasks.append(str(itask))
 
-        for may, tasks in [('', n) for n in warn_tasks] + [
-            ('may be', n) for n in _warn_tasks
-        ]:
+        for may, tasks in (('', warn_tasks), ('may be', _warn_tasks)):
             if tasks:
                 _tasks = "\n * ".join(tasks)
                 LOG.info(
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1123,7 +1123,11 @@ class TaskPool:
         # Log tasks orphaned by a reload but not currently in the task pool.
         for name in orphans:
             if name not in (itask.tdef.name for itask in tasks):
-                LOG.warning("Removed task: '%s'", name)
+                LOG.info("Removed task: '%s'", name)
+        # Store lists of tasks which were active before reload.
+        warn_tasks: List[str] = []
+        _warn_tasks: List[str] = []
+
         for itask in tasks:
             if itask.tdef.name in orphans:
                 if (
@@ -1136,7 +1140,7 @@ class TaskPool:
                 else:
                     # Keep active orphaned task, but stop it from spawning.
                     itask.graph_children = {}
-                    LOG.warning(
+                    LOG.info(
                         f"[{itask}] will not spawn children "
                         "- task definition removed"
                     )
@@ -1158,15 +1162,18 @@ class TaskPool:
                 self._swap_out(new_task)
                 self.data_store_mgr.delta_task_prerequisite(new_task)
                 LOG.info(f"[{itask}] reloaded task definition")
+
                 if itask.state(*TASK_STATUSES_ACTIVE):
-                    LOG.warning(
-                        f"[{itask}] active with pre-reload settings"
-                    )
+                    warn_tasks.append(str(itask))
                 elif itask.state(TASK_STATUS_PREPARING):
                     # Job file might have been written at this point?
-                    LOG.warning(
-                        f"[{itask}] may be active with pre-reload settings"
-                    )
+                    _warn_tasks.append(str(itask))
+
+        for may, tasks in (('', warn_tasks), ('may be', warn_tasks)):
+            if tasks:
+                _tasks = "\n * ".join(tasks)
+                LOG.info(
+                    f"Tasks {may} active with pre-reload settings:\n{_tasks}")
 
         # Reassign live tasks to the internal queue
         del self.task_queue_mgr
@@ -1250,22 +1257,18 @@ class TaskPool:
                     orphans_kill_failed.append(itask)
                 else:
                     orphans.append(itask)
-        if orphans_kill_failed:
-            LOG.warning(
-                "Orphaned tasks (kill failed):\n"
-                + "\n".join(
-                    f"* {itask.identity} ({itask.state.status})"
-                    for itask in orphans_kill_failed
+
+        for orphanlist, extra_text in (
+            (orphans_kill_failed, ' (kill failed)'),
+            (orphans, '')
+        ):
+            if orphanlist:
+                LOG.warning(
+                    f"Orphaned tasks{extra_text}:\n"
+                    + "\n".join(
+                        f"* {itask.identity} ({itask.state.status})"
+                        for itask in orphanlist)
                 )
-            )
-        if orphans:
-            LOG.warning(
-                "Orphaned tasks:\n"
-                + "\n".join(
-                    f"* {itask.identity} ({itask.state.status})"
-                    for itask in orphans
-                )
-            )
 
         for id_key in self.task_events_mgr._event_timers:
             LOG.warning(
--- a/tests/unit/test_workflow_files.py
+++ b/tests/unit/test_workflow_files.py
@@ -207,7 +207,7 @@ def test_infer_latest_run_warns_for_runN(
     monkeypatch.setattr('cylc.flow.LOG.level', logging.INFO)
     infer_latest_run(runN_path, warn_runN=warn_arg)
     filtered_log = log_filter(
-        logging.INFO, contains="You need only include"
+        logging.WARNING, contains="You do not need to include"
     )
     assert filtered_log if warn_arg else not filtered_log
 
--- a/tests/unit/test_workflow_files.py
+++ b/tests/unit/test_workflow_files.py
@@ -198,14 +198,16 @@ def test_infer_latest_run_warns_for_runN(
     warn_arg: bool,
     log_filter: Callable,
     tmp_run_dir: Callable,
+    monkeypatch: pytest.MonkeyPatch
 ):
     """Tests warning is produced to discourage use of /runN in workflow_id"""
     (tmp_run_dir() / 'run1').mkdir()
     runN_path = tmp_run_dir() / 'runN'
     runN_path.symlink_to('run1')
+    monkeypatch.setattr('cylc.flow.LOG.level', logging.INFO)
     infer_latest_run(runN_path, warn_runN=warn_arg)
     filtered_log = log_filter(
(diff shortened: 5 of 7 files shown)


cylc/flow/rundb.py
```diff
@@ -560,7 +560,7 @@ def execute_queued_items(self):
                 table.update_queues.clear()
             # Report public database retry recovery if necessary
             if self.n_tries:
-                LOG.warning(
```
oliver-sanders, line 563: Not sure.

This isn't an issue for the public DB, but is for the private.

cylc/flow/rundb.py
```diff
@@ -601,7 +601,7 @@ def _execute_stmt(self, stmt, stmt_args_list):
             for i, stmt_args in enumerate(stmt_args_list):
                 err_log += ("\nstmt_args[%(i)d]=%(stmt_args)s" % {
                     "i": i, "stmt_args": stmt_args})
-            LOG.warning(err_log)
+            LOG.info(err_log)
```
oliver-sanders, line 604: Ditto.

cylc/flow/task_events_mgr.py
```diff
@@ -887,7 +887,7 @@ def _process_message_check(
             timestamp = ""
         if flag == self.FLAG_RECEIVED and submit_num != itask.submit_num:
             # Ignore received messages from old jobs
-            LOG.warning(
```
oliver-sanders, line 890: Not sure about this one.

cylc/flow/task_events_mgr.py
```diff
@@ -917,13 +917,13 @@ def _process_message_check(
             # Ignore messages if task has a retry lined up
             # (caused by polling overlapping with task failure)
             if flag == self.FLAG_RECEIVED:
-                LOG.warning(
```
oliver-sanders, line 920: Not sure about this.

cylc/flow/task_events_mgr.py
```diff
@@ -917,13 +917,13 @@ def _process_message_check(
...
+                LOG.info(
                     f"[{itask}] "
                     f"{self.FLAG_RECEIVED_IGNORED}{message}{timestamp}"
                 )
 
             else:
-                LOG.warning(
```
oliver-sanders, line 926: Not sure about this.

cylc/flow/task_events_mgr.py
```diff
@@ -1375,7 +1375,7 @@ def _process_message_started(self, itask, event_time, forced):
         """Helper for process_message, handle a started message."""
         if itask.job_vacated:
             itask.job_vacated = False
-            LOG.warning(f"[{itask}] Vacated job restarted")
```
oliver-sanders, line 1378: Don't actually know what a "vacated job" is!

cylc/flow/task_pool.py
```diff
@@ -1123,7 +1123,11 @@ def _reload_taskdefs(self) -> None:
         # Log tasks orphaned by a reload but not currently in the task pool.
         for name in orphans:
             if name not in (itask.tdef.name for itask in tasks):
-                LOG.warning("Removed task: '%s'", name)
```
oliver-sanders, line 1126: Agreed.

The user specifically got Cylc to remove tasks from the workflow, why would we warn them about this.

cylc/flow/task_pool.py
```diff
@@ -1136,7 +1140,7 @@ def _reload_taskdefs(self) -> None:
                 else:
                     # Keep active orphaned task, but stop it from spawning.
                     itask.graph_children = {}
-                    LOG.warning(
```
oliver-sanders, line 1139: Agreed.

cylc/flow/task_pool.py
```diff
@@ -1158,15 +1162,18 @@ def _reload_taskdefs(self) -> None:
...
                 LOG.info(f"[{itask}] reloaded task definition")
+
                 if itask.state(*TASK_STATUSES_ACTIVE):
-                    LOG.warning(
-                        f"[{itask}] active with pre-reload settings"
-                    )
+                    warn_tasks.append(str(itask))
                 elif itask.state(TASK_STATUS_PREPARING):
                     # Job file might have been written at this point?
-                    LOG.warning(
```
oliver-sanders, line 1167: Agreed.

* This is what the user told Cylc to do.
* The user does not need to review this list of tasks.
* There is no call to action here.

cylc/flow/task_pool.py
```diff
@@ -1250,22 +1257,18 @@ def warn_stop_orphans(self) -> None:
...
+            (orphans, '')
+        ):
+            if orphanlist:
+                LOG.warning(
+                    f"Orphaned tasks{extra_text}:\n"
+                    + "\n".join(
+                        f"* {itask.identity} ({itask.state.status})"
+                        for itask in orphanlist)
                 )
-            )
```
oliver-sanders, line 1268: Agreed, it is a good idea to combine these messages.

cylc/flow/rundb.py
```diff
@@ -535,7 +535,7 @@ def execute_queued_items(self):
                 )
                 raise
             self.n_tries += 1
-            LOG.warning(
```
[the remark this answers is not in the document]
wxtim: That seems reasonable - I'll see if I can sort out the difference.

cylc/flow/task_pool.py
wxtim: > Do we actually need this message?

Need? Possibly - it can be useful to present evidence after the fact, probably. Perhaps it should go to debug?

cylc/flow/rundb.py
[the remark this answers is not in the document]
wxtim: The difference is already handled at line 519  - 

```python
if not self.is_public:
  LOG.error(...)
  raise

File: cylc/flow/rundb.py (hunk 1 above)
Comment (wxtim) [reply cylc/flow/rundb.py:563]: This will not get called if it's the private DB.

File: cylc/flow/workflow_files.py [hunk 12]
```diff
@@ -963,7 +963,7 @@ def infer_latest_run(
     if path.name == WorkflowFiles.RUN_N:
         runN_path = path
         if warn_runN:
-            LOG.warning(
+            LOG.info(
```
oliver-sanders: No need to demote this?

Would only interact with the warning triangles on startup if `runN` is explicitly specified to `cylc play`?

tests/unit/test_workflow_files.py
```diff
@@ -205,7 +205,7 @@ def test_infer_latest_run_warns_for_runN(
     runN_path.symlink_to('run1')
     infer_latest_run(runN_path, warn_runN=warn_arg)
     filtered_log = log_filter(
-        logging.WARNING, "You do not need to include runN in the workflow ID"
+        logging.INFO, "You do not need to include runN in the workflow ID"
```
oliver-sanders: No need to demote this?

cylc/flow/workflow_files.py
wxtim: It's not really warning worthy - it's helping the users a little, but there's no real consequences to not noting this?

tests/unit/test_workflow_files.py
wxtim: I think warning was too high - what are the consequences of not seeing the message?

cylc/flow/workflow_files.py
oliver-sanders: This is a warning because we expect the user to spot this and to take action (i.e. don't specify the runN bit).

If this is at the info level, it won't be spotted and the user won't take action.

I don't think why we would want to demote this?

cylc/flow/config.py
```diff
@@ -914,7 +914,7 @@ def _check_circular(self):
         """Check for circular dependence in graph."""
         if (len(self.taskdefs) > self.CHECK_CIRCULAR_LIMIT and
                 not getattr(self.options, 'check_circular', False)):
-            LOG.warning(
+            LOG.info(
                 f"Number of tasks is > {self.CHECK_CIRCULAR_LIMIT}; will not "
                 "check graph for circular dependencies. To enforce this "
                 "check, use the option --check-circular.")
```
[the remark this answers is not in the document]
hjoliver: Maybe there's no such algorithm - it might be inherently difficult.

We almost need a different category for things like this. Arguably it should be a warning because it alerts the user that an important validation check is being skipped. But on the other hand the user can't do anything to address it because it's based on the number of tasks.

All things considered, INFO is best, I guess.

(An extreme solution: have the CLI store some stuff on the filesystem to allow (e.g.) Warning about cyclic dependency checking once per day, per workflow).

@wxtim - could you change `enforce` to `force` in the message?  IMO "enforce" has slightly different and slightly incorrect connotations.

> To force somebody or something means using force to make them do something. To enforce something - usually a law or other rule - means making sure that that law or rule is obeyed. Enforcement may or may not require the use of force - it depends on the rule, and on how obedient the people are.

cylc/flow/task_events_mgr.py
hjoliver: Does loadleveler still exist?...

It is **old** - found in some paper online:

> LoadLeveler [3] is a distributed job scheduling product of IBM. It is developed
based on the licensed codes from the Condor system [13] in mid 1990. The archi-
tectural framework of LoadLeveler, as shown in Figure 1, still retains the core
structure of Condor

It seems to have been absorbed into Tivoli Workload Scheduler, but recent editions of the guide for that don't seem to mention it by name. (However, I'm not going to waste time looking further).

cylc/flow/task_pool.py
hjoliver: IMO we should keep the message. It shows that Cylc correctly identified and handled a live task whose definition has been removed by the reload. That's a fairly unusual circumstance, with potentially unpleasant results if we get it wrong.

cylc/flow/workflow_files.py
hjoliver: This was just added to the Issue yesterday: https://github.com/cylc/cylc-flow/issues/6760#issuecomment-2929729888

I think it should not be a warning - it's just informing the user that its OK to omit the run number. To specify it, however, is not any kind of wrong.

Plus, as per my comment [here](https://github.com/cylc/cylc-flow/issues/6760#issuecomment-2932955032) the text should be changed to say that omitting the run number is only an option if targeting the latest run.

tests/unit/test_workflow_files.py
hjoliver: (Duplicate of previous comment - this is only a test)

cylc/flow/config.py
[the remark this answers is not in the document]
wxtim: Enforce sounds right in this context. But I'm going to propose a rewording which sidesteps the issue.

cylc/flow/task_pool.py
wxtim: @hjoliver  But at what level?

cylc/flow/workflow_files.py
wxtim: > the text should be changed to say that omitting the run number is only an option if targeting the latest run.

Does my text not say that?

cylc/flow/config.py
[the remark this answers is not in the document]
oliver-sanders: > Maybe there's no such algorithm - it might be inherently difficult.

We're actually using a standard algorithm for this:

https://github.com/cylc/cylc-flow/blob/4082f43032e4e5a5f6034761d1925a44a031d78d/cylc/flow/config.py#L958-L960

It's fairly simple, I'm sure we can speed this up. I also think we can increase the 100 limit (this was an arbitrary choice).

Anyway, topic for another issue. Suggest keeping this one to WARNING stuff.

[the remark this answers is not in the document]
hjoliver: (I meant "maybe there's no such *fast* algorithm" -  I know we're using a standard one now, but it's slow).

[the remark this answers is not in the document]
hjoliver: > Enforce sounds right in this context. But I'm going to propose a rewording which sidesteps the issue.

Not according to my understanding of the definitions and what the option does - but OK, let's see the new wording.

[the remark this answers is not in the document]
oliver-sanders: > (I meant "maybe there's no such fast algorithm" - I know we're using a standard one now, but it's slow).

I think the algorithm itself is quick. The algorithm is all statically typed so a low level imp would give a decent speedup. It's also possible that Python's JIT compiler (alpha feature ATM) could make a substantial passive improvement.

Def things we can do...

cylc/flow/task_pool.py
oliver-sanders: I'm not so convinced of the importance of this, the the task will continue with its previous config. I think this is a common circumstance right (?), I can't think what unpleasant / unexpected results this could lead to?

But if you disagree (no need to state your argument, I don't want to haggle this), I'm happy to keep this on at the info level.

cylc/flow/workflow_files.py
oliver-sanders: >  think it should not be a warning - it's just informing the user that its OK to omit the run number

Not quite.

This isn't telling you that it's unnecessary to specify the run name (we don't log that at all).

This is telling you that it's unnecessary to specify `runN` as the run name.

This is a usage pattern we've tried to prevent, hence the warning.

cylc/flow/workflow_files.py
```diff
@@ -963,10 +963,9 @@ def infer_latest_run(
     if path.name == WorkflowFiles.RUN_N:
         runN_path = path
         if warn_runN:
-            LOG.warning(
-                f"You do not need to include {WorkflowFiles.RUN_N} in the "
-                "workflow ID; Cylc will select the latest run if just the "
-                "workflow name is used"
+            LOG.info(
+                f"You need only include {WorkflowFiles.RUN_N} in the"
```
wxtim:
```suggestion
                f"You need only include a run number in the"
```

cylc/flow/workflow_files.py
wxtim: Reverted. Not interested in courting controversy.

cylc/flow/task_pool.py
```diff
@@ -1158,15 +1162,18 @@ def _reload_taskdefs(self) -> None:
...
-                    )
+                    warn_tasks.append(str(itask))
                 elif itask.state(TASK_STATUS_PREPARING):
                     # Job file might have been written at this point?
-                    LOG.warning(
-                        f"[{itask}] may be active with pre-reload settings"
-                    )
+                    _warn_tasks.append(str(itask))
+
+        for may, tasks in (('', warn_tasks), ('may be', warn_tasks)):
```
oliver-sanders: Logging the same `warn_tasks` list twice, once with `may be` and once without?

oliver-sanders: @wxtim, poke incase you missed this

wxtim: Well poked - I had missed it.

oliver-sanders: I think it's still iterating over the same list twice with the latest change?

wxtim: `warn_tasks != _warn_tasks`?

```python
>>> warn_tasks = ['foo', 'bar']
>>> _warn_tasks = ['baz', 'qux']
>>> for may, tasks in [('', n) for n in warn_tasks] + [
...     ('may be', n) for n in _warn_tasks
... ]:
...     print(may, tasks)
... 
 foo
 bar
may be baz
may be qux

Comment (oliver-sanders) [reply cylc/flow/task_pool.py]: Aah, that was the change.

File: cylc/flow/task_pool.py [hunk 17]
```diff
@@ -1169,7 +1169,9 @@ def _reload_taskdefs(self) -> None:
                     # Job file might have been written at this point?
                     _warn_tasks.append(str(itask))
 
-        for may, tasks in (('', warn_tasks), ('may be', warn_tasks)):
+        for may, tasks in [('', n) for n in warn_tasks] + [
+            ('may be', n) for n in _warn_tasks
+        ]:
```
oliver-sanders: The list comprehensions aren't necessary here:

```suggestion
        for may, tasks in (('', warn_tasks), ('may be', _warn_tasks)):
```

cylc/flow/workflow_files.py
hjoliver: Unresolving this as I just saw @oliver-sanders response to my comment. 

OK so I guess I didn't read carefully enough and misunderstood the text as saying you don't need to specify the run number at all.

If the warning is actually just triggered by use of the literal "runN" then I agree it should be a warning.

However, can we tweak the text to make it clear that it means the literal "runN" string, not say "run2" etc. ?

oliver-sanders: >  If the warning is actually just triggered by use of the literal "runN" then I agree it should be a warning.

I think Tim has now reverted this change so it's back to warning again.

> However, can we tweak the text to make it clear that it means the literal "runN" string, not say "run2" etc. ?

This PR is about demoting warnings, it's likely to drag on if we start debating re-wordings, suggest punting these to issues.

Quick response, I think the meaning of `runN` is well explained to the user, it's just not so obvious when you're reading the message in the Python code:

```console
$ cylc hold one/runN
WARNING - You do not need to include runN in the workflow ID; Cylc will select the latest run if just the workflow name is used
```

hjoliver: OK, actually I'm happy with that wording, in context.

oliver-sanders:
FYI: Have checked back through the comments, I think we're all agreed on these changes.

PR #6768, opened 2025-05-29
Always submit preparing tasks.

### LINKED ISSUE #6767: trigger: remote-init will not run if the workflow is paused
We can trigger tasks whilst a workflow is paused, however, this will only work if remote-init+remote-fileinstall has completed (or is not required).

#### Reproducible example

```cylc
[scheduling]
  [[graph]]
    R1 = foo

[runtime]
  [[foo]]
    platform = excd-bg
    script = true
```

```console
$ cylc vip --pause
$ cylc trigger //1/foo
# task gets stuck in preparing
$ cylc play
# task submits
```

hjoliver:
Close #6767 

Remote platform init was thwarting the trigger-if-paused logic.

**Check List**

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1354,19 +1354,20 @@ class Scheduler:
                 pre_prep_tasks.update(self.pool.tasks_to_trigger_now)
                 self.pool.tasks_to_trigger_now = set()
 
-            if not self.is_paused:
+            if self.is_paused:
+                # finish processing preparing tasks
+                pre_prep_tasks.update({
+                    itask for itask in self.pool.get_tasks()
+                    if itask.state(TASK_STATUS_PREPARING)
+                })
+            else:
                 # release queued tasks
                 pre_prep_tasks.update(self.pool.release_queued_tasks())
                 if self.pool.tasks_to_trigger_on_resume:
                     # and manually triggered tasks to run once workflow resumed
                     pre_prep_tasks.update(self.pool.tasks_to_trigger_on_resume)
                     self.pool.tasks_to_trigger_on_resume = set()
-            else:
-                # finish processing preparing tasks
-                pre_prep_tasks.update({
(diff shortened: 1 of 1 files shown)


hjoliver:
(Hmm, maybe not quite as simple as this!)

hjoliver:
My original commit worked (I think) and simplified the logic quite a bit, but it resulted in ref tests failing due to double logging of `x triggered off y` lines. I think that happens already whenever `pre_prep_tasks` is taken from preparing pool tasks, but that's not hit by any of our test. I'll look closer tomorrow.

hjoliver:
@oliver-sanders - I've confirmed that this small change fixes the reported bug, so we might as well get it in. Pretty sure I left it in Draft just because I wasn't sure how to test it - any suggestions?

oliver-sanders:
>  I wasn't sure how to test it - any suggestions?

Extend or copy this test:

https://github.com/cylc/cylc-flow/blob/3a484542c3dc20064a6af4a0fbc272e9ca823095/tests/integration/test_force_trigger.py#L114-L147

cylc/flow/scheduler.py
```diff
@@ -1354,7 +1354,13 @@ def release_tasks_to_run(self) -> bool:
...
                 self.pool.tasks_to_trigger_now = set()
 
-            if not self.is_paused:
+            if self.is_paused:
+                # finish processing preparing tasks
+                pre_prep_tasks.update({
+                    itask for itask in self.pool.get_tasks()
+                    if itask.state(TASK_STATUS_PREPARING)
+                })
+            else:
```
MetRonnie, line 1363: I've just rearranged the if condition from `if not is_paused ... else` to `if is_paused ... else`

MetRonnie (approved):

cylc/flow/scheduler.py
oliver-sanders: Err, right

Issue #3497, opened 2020-02-05
XTriggers: re-implement as async functions


oliver-sanders:
**The Problem:**

At the moment XTriggers are run by the subprocess pool, consequences:

* Each XTrigger call involves a subprocess call to a script.
* Python is invoked for each xtrigger call.
* XTriggers are re-imported for each xtrigger call.
* Missing XTriggers *[can]* cause runtime errors

Botching the XTrigger loading mechanism for #3465 seems nasty.

**The Proposal:**

Convert XTriggers to asynchronous functions and call them directly from the main loop piggy-backing on the main loop plugin functionality introduced in https://github.com/cylc/cylc-flow/pull/3492.

This will turn XTriggers into a blocking stage of the main loop (where they would effectively be running in an unlimited thread pool) something like this:

```python
async def main_loop(self):
    # ...
    await asyncio.gather(*[
        trigger(args)
        for trigger in xtriggers
    ])
    # ...
    await asyncio.sleep()
```

Consequences:

* This should be an efficiency boost as we are calling the XTrigger directly rather than going through `SubProcPool`, Subprocess, Python and `importlib`.
* XTriggers would not be re-imported whilst the suite is running, if you want to change XTrigger code you need to restart your suite.
* XTriggers would get loaded on suite start, missing xtriggers could not cause runtime errors.
* XTriggers gain the logging, timing and standardisation offered by main loop plugins.
* The responsibility for performing non-blocking IO lies with the XTrigger function writer.

**Questions:**

* [ ] Is there any reason we would want XTriggers to be re-imported whilst the suite is running?
* [ ] Are we happy with an async implementation?
* [ ] Should XTriggers by run with the other main loop plugins to increase parallelism?

**Pull requests welcome!**

hjoliver:
Closes #2917 !

oliver-sanders:
Didn't spot that!

hjoliver:
(I'll move that "close" comment to the description...)

dwsutherland:
I think this is blocking (hence why `WFS => UIS` is run in the thread pool executor):
```
async def push(*args):
    socket = ...
    await msg = socket.recv(). # could take hours, doesn’t matter
    return json.loads(msg)
```
However, there's an option to make it `socket.recv()` non-blocking (then you can loop over it like a queue).. And I hadn't tried `asyncio.gathers(iterable_of_asyncfuncs)` because the the `iterable_of_asyncfuncs` changed size for the UIS (amongst other things).

[...]

PR #6772, opened 2025-06-02
ensure that xtrigger or fails

### LINKED ISSUE #6771: Xtrigger or dosn't fail validation
[Reference to Discourse](https://cylc.discourse.group/t/should-a-b-c-fail-validation/1178/4)

### Description

Using xtriggers in a logical "or" (i.e. `@xtriggerA | @xtriggerB`) will not work. It will be treated as an "and". However this structure passes validation.

### Reproducible Example

```python
# Test added to tests/integration/test_config.py
def test_xtrig_or_fails_validation(
    flow: "Fixture",
    validate: "Fixture",
):
    id_ = flow(
        {
            "scheduling": {
                "initial cycle point": "2024",
                "xtriggers": {
                    "xrandom": "xrandom(100)",
                    "echo": "echo(succeed=True)"
                },
                "graph": {"R1": "@xrandom | @echo => foo"},
            }
        }
    )
    expected_msg = "'|', 'or' operator, is not supported in xtriggers."
    with pytest.raises(XtriggerConfigError, match=expected_msg):
        validate(id_)
```

wxtim:
Closes #6771 

**Check List**


cylc/flow/graph_parser.py
```diff
@@ -186,6 +186,7 @@ class GraphParser:
 
     # Match @xtriggers.
     REC_XTRIG = re.compile(r'@[\w\-+%]+')
+    REC_XTRIG_OR = re.compile(r'@.*\s*\|\s*@.*')
```
ColemanTom: Will this fail for something like `@a | foo => bar`? Or does validation fail for this combination already? This regex only captures two xtriggers obviously, but maybe other spots capture the other cases with or being used,which should also fail?

wxtim: Absolutely right - I'm not totally convinced this is a reasonable solution, but I'll have another look at the regex.

changes.d/6772.fix.md
```diff
@@ -0,0 +1 @@
+Prevent validation accepting two xtriggers with an | between them
```
MetRonnie:
```suggestion
Xtriggers with an OR operator (`|`) between them now correctly fails validation as Cylc does not support this.
```

cylc/flow/graph_parser.py
oliver-sanders: I don't think we can catch this in a regex, only by looking over the graph edges.

Xtriggers cannot be used in conditional expressions of any form (i.e. if you declare an xtrigger condition, it *must* be satisfied).

hjoliver: @oliver-sanders are you suggesting that's a more than just a possibly-temporary restriction on xtrigger use? It's true that xtriggers are either not-yet-satisfied or satisfied, they don't have an equivalent of the task `failed` state, but we could still in principle have `@x | foo => bar` and just stop checking the xtrigger condition if `bar` triggers off of `foo:succeeded`.

oliver-sanders: @hjoliver, this issue is not about working around the OR conditional limitation of xtriggers, it is about ensuring that `cylc validate` fails in this situation in order to communicate the limitation.

I'm wasn't saying anything about the long-term future of conditional graph strings involving xtriggers, however, since you asked... I'm keen to look into an alternative xtrigger implementation that would support `async` xtriggers, this would also resolve the conditional logic issue amongst other limitations. In other words, I'm happy for this to remain a permanent limitation of xtriggers in their current form and pursue a more general solution to this and other problems.

For info, I dumped a quick outline here: https://github.com/cylc/cylc-flow/issues/3497#issuecomment-2167875786 (the thing I brought this up in a recent xtrigger discussion).

I think it will actually be quite easy to implement this approach as it reuses existing code and wouldn't require any form of "manager" class. This might be something we look at once we've got our next opps migration out of the way.

hjoliver: > @hjoliver, this issue is not about working around the OR conditional limitation of xtriggers, it is about ensuring that cylc validate fails in this situation in order to communicate the limitation.

I know that, but your statement (the second bit of it, in parentheses) looked suspiciously like a statement about xtriggers in principle that I don't agree with - but OK, just a misunderstanding.

oliver-sanders: At present that statement is absolutely correct! Whatevs.

wxtim: > I don't think we can catch this in a regex, only by looking over the graph edges.

The regex is being applied to one end of a graph edge. I'm not sure where you'd go about checking the graph edges. Are they stored somwhere
 
> Xtriggers cannot be used in conditional expressions of any form (i.e. if you declare an xtrigger condition, it _must_ be satisfied).

Does that mean that we should further limit the scope of expressions in which xtriggers are used?

hjoliver:
Out of curiosity, this fails here and on master, but I wonder if it needs to?
(It's been a long time since I looked at - and probably wrote - the associated code, so I'm not sure of the reason for the restriction any more, and whether it should apply to this sort of expression...)
```
@x & (a | b) => c
```
```
$ cylc val .
WorkflowConfigError: Xtriggers cannot be used in conditional graph expressions:
@x&(a:succeeded|b:succeeded) => c
```

oliver-sanders:
> This fails here and on master, but I wonder if it needs to?

If this PR isn't changing behaviour, suggest moving this to another issue:

https://github.com/cylc/cylc-flow/issues/6949

oliver-sanders:
As Hillary pointed out, there is already logic intended to catch xtriggers used in conditional expressions:

https://github.com/cylc/cylc-flow/blob/700eb0217f25fbe178b0eed2b432ca7f7c4a00db/cylc/flow/scripts/validate.py#L196-L210

However, this check isn't catching situations where there are no tasks on the LHS, only xtriggers.

Sidenote: Strangely, this code is in `cylc.flow.validate` (which means you can bypass this error using `cylc install; cylc play`). I suspect this is purely historical.

If the new test covers all of the logic of the old one, suggest ripping the old check out. There are probably tests for the old check which could do with being combined / erased.

hjoliver:
> However, beyond @wall_clock xtriggers are not used at our site, might be worth getting Hillary to make sure it doesn't cause any false positives with NIWA workflows?

Unfortunately, due to enforced home dir privacy here, I can't see our workflows.

cylc/flow/prerequisite.py
```diff
@@ -253,12 +253,8 @@ def _eval_satisfied(self) -> bool:
             # * the expression is constructed internally
             # * https://github.com/cylc/cylc-flow/issues/4403
         except (SyntaxError, ValueError) as exc:
-            err_msg = str(exc)
```
wxtim, line 270: From 3.10 Python no longer emits the message being searched for here. [ref](https://github.com/python/cpython/blob/6826166280d6518441a729b444173db205c4ab20/Doc/whatsnew/3.10.rst#syntaxerrors)

changes.d/6772.fix.md
```diff
@@ -0,0 +1 @@
+Xtriggers with an OR operator (`|`) between them now correctly fails validation as Cylc does not support this.
```
oliver-sanders:
```suggestion
Xtriggers with an OR operator (`|`) between them now correctly fail validation as Cylc does not support this.
```

dwsutherland (approved):
LGTM
image

PR #6775, opened 2025-06-03
Getlogin

### LINKED ISSUE #6773: Add $SUDO_USER to scheduler log "Command received" lines if present
### Problem

The scheduler log provides an excellent source of who did what to which workflow, when those actions are performed via the WUI. For example:

```
2025-06-02T02:34:56Z INFO - Command "set_verbosity" received from jarich. ID=e98fa3fd-ce49-4fce-953e-d86c36aed48b
    set_verbosity(level=10)
2025-06-02T02:34:56Z DEBUG - Processing 1 queued command(s)
```

When actions are performed by the workflow owner (for example via sudo on the command line) no user is attributed:

```
2025-06-02T11:58:21Z INFO - Command "force_trigger_tasks" received. ID=974d42cb-b151-4ea6-877d-28e92f53340f
    force_trigger_tasks(flow=[], flow_wait=False, tasks=['20250602T0000Z/get_winds*'])
2025-06-02T11:58:22Z DEBUG - Processing 1 queued command(s)
```

### Proposed Solution

If `$SUDO_USER` is set in the environment, add that to the message. For example:

```
2025-06-02T11:58:21Z INFO - Command "force_trigger_tasks" received (from jarich via sudo). ID=974d42cb-b151-4ea6-877d-28e92f53340f
    force_trigger_tasks(flow=[], flow_wait=False, tasks=['20250602T0000Z/get_winds*'])
2025-06-02T11:58:22Z DEBUG - Processing 1 queued command(s)
```

### Rationale

When a user is working directly with their workflow, they can probably be counted on to remember what they have been doing. However, I imagine it is not unusual for production accounts to not be the same as user accounts, especially if workflows need 24/7 support. Therefore it is very helpful to know who did what, ideally without having to examine system log sources (like `/var/log/secure`).

psa:
Provide more consistent logging messages so they're easier to parse and log the username from the system if a command isn't being run by the owner or a scheduler user.

Fixes #6773

**Check List**

- [N/A] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [X] Tests are included (or explain why tests are not needed).
- [N/A] Changelog entry included if this is a change that can affect users
- [N/A] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [X] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.


cylc/flow/network/resolvers.py
```diff
@@ -724,10 +725,17 @@ async def _mutation_mapper(
 
         """
         user = meta.get('auth_user', self.schd.owner)
-        if user == self.schd.owner:
-            log_user = ""  # don't log user name if owner
```
oliver-sanders, line 704: I think this PR is effectively equivalent to:

```suggestion
        if user == self.schd.owner or not user:
            user = os.getlogin()
```

I.E:
* Log the user name if provided with the request.
* Otherwise log the scheduler owner.

oliver-sanders (changes requested):
The resolver code is run by the scheduler (i.e. the server) not the client, so this isn't going to extract the user who actioned the operation, you would need to shift this into the Cylc client code (`cylc.flow.network.client`).

However, the client cannot be trusted in this situation (because it is being run by a different user), so it is not possible to achieve this (https://github.com/cylc/cylc-flow/issues/6773#issuecomment-2934468089).

psa:
I'm working on re-writing this based on the feedback in https://github.com/cylc/cylc-flow/issues/6773#issuecomment-2934680239

cylc/flow/network/client.py
```diff
@@ -374,6 +385,7 @@ def get_header(self) -> dict:
                     os.getenv(
                         "CLIENT_COMMS_METH",
                         default=CommsMeth.ZMQ.value
-                    )
+                    ),
+                'behalf_of': behalf_of
```
oliver-sanders: Send the full message through using the existing `auth_user` field rather than creating a new `behalf_of` one.

```python
auth_user = '<actor> (possibly on behalf of <user>)'
```

That way no changes will be required in `cylc/flow/network/resolvers.py` to log this and the feature will be fully backwards compatible.

psa: @oliver-sanders 
I've been poking at this but need help from someone with  more py-zmq knowledge.

I was expecting `self.socket.get(zmq.PLAIN_USERNAME)` in `async_request()` to achieve this. I've not found a way to get the username on the client side.

Ideas on how I might approach this?

oliver-sanders: @psa, ZMQ doesn't come into it, just replace `behalf_of` with `auth_user` and supply it with the full text:

```suggestion
                'auth_user': f'{actor} (possibly on behalf of {user})'
```

Where `actor` is the actual user account (i.e. `$USER`) and `user` is the name of the person who is acting via `sudo` (i.e. `$LOGNAME`).

oliver-sanders:
Ping @psa, let us know if you need any assistance to unstick this PR.

PR #6776, opened 2025-06-04
tests/f: fix restart/41

oliver-sanders:
The `cylc workflow-state` call in this test wasn't properly updated after recent(ish) changes.


tests/functional/restart/41-auto-restart-local-jobs.t
```diff
@@ -83,7 +85,7 @@ log_scan "${TEST_NAME}-stop-log-scan" "${LOG_FILE}" 40 1 \
 # have waited for them to complete
 grep_fail 'orphaned task' "$LOG_FILE"
 
-poll_workflow_restart
+run_ok "${TEST_NAME_BASE}-restart" poll_workflow_restart 10 "$UUID"
```
oliver-sanders, line 88: Before, if this `poll_workflow_restart` call failed, the test would continue (after the timeout) with no indication that it had failed :(

Manually specifying the UUID prevents the test from failing if this test doesn't start running until after the restart for improved robustness.

MetRonnie (approved):
Ran locally

PR #6780, opened 2025-06-06
check-circular: assorted optimisations

oliver-sanders:
I took a look at the --check-circular code as a training day activity to see how it might fare with JIT compilation.

On closer inspection, it turned out that the `get_graph_raw` method called by the check-circular logic was actually the main bottleneck, with time largely spent elsewhere in the codebase. So not that suitable for JIT after all :(

Without changing the algorithms I found some low-hanging fruit which brings a reasonable speedup:

* avoid unnecessary sort in get_graph_raw
* simplify get_graph_raw conditionals
* switch lists to sets
* avoid unnecessary dupliace get_inst calls

| Workflow | Viz points | Before (s) | After (s) | Saving % |
| --- | --- | --- | --- | --- |
| complex | 3 (default) | 0.111 | 0.0809 | 27% |
| stress test | 100 | 18.4 | 11.0 | 40% |

<details>
<summary>stress test config</summary>

```cylc
[scheduler]
    allow implicit tasks = True

[task parameters]
    b = 1..100
    c = 1..100

[scheduling]
    initial cycle point = 2000
    [[graph]]
        P1D = """
            a => <b> => <c> => d
            <c=1>[-P1D] => <b=1>
            <b=2>[-P2D] => <c=2>
            a[-P3D] => d
        """
```

</details>

This is just a quick low-hanging fruit optimisation. To get better performance out of this code in the future, we should look at re-writing the `get_graph_raw` end of this code into a form that is more compatible with the `check_circular` code.

**Check List**

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -920,8 +920,11 @@ class WorkflowConfig:
                 "check, use the option --check-circular.")
             return
         start_point_str = self.cfg['scheduling']['initial cycle point']
-        raw_graph = self.get_graph_raw(start_point_str,
-                                       stop_point_str=None)
+        raw_graph = self.get_graph_raw(
+            start_point_str,
+            stop_point_str=None,
+            sort=False,
+        )
         lhs2rhss = {}  # left hand side to right hand sides
         rhs2lhss = {}  # right hand side to left hand sides
         for lhs, rhs in raw_graph:
@@ -1815,8 +1818,9 @@ class WorkflowConfig:
 
     def add_sequence(self, nodes, seq, suicide):
         """Add valid sequences to taskdefs."""
+        graph_node_parser = GraphNodeParser.get_inst()
         for node in nodes:
-            name, offset = GraphNodeParser.get_inst().parse(node)[:2]
+            name, offset = graph_node_parser.parse(node)[:2]
             taskdef = self.get_taskdef(name)
             # Only add sequence to taskdef if explicit (not an offset).
             if offset:
@@ -1847,6 +1851,7 @@ class WorkflowConfig:
         triggers = {}
         xtrig_labels = set()
 
+        graph_node_parser = GraphNodeParser.get_inst()
         for left in left_nodes:
             if left.startswith('@'):
                 xtrig_labels.add(left[1:])
@@ -1854,7 +1859,8 @@ class WorkflowConfig:
             # (GraphParseError checked above)
             (name, offset, output, offset_is_from_icp,
              offset_is_irregular, offset_is_absolute) = (
-                GraphNodeParser.get_inst().parse(left))
+                graph_node_parser.parse(left)
+            )
 
             # Qualifier.
             outputs = self.cfg['runtime'][name]['outputs']
@@ -1982,7 +1988,12 @@ class WorkflowConfig:
         return stop_point
 
     def get_graph_raw(
-            self, start_point_str=None, stop_point_str=None, grouping=None):
+        self,
+        start_point_str=None,
+        stop_point_str=None,
+        grouping=None,
+        sort=True,
+    ):
         """Return concrete graph edges between specified cycle points.
 
         Return a family-collapsed graph if the grouping arg is not None:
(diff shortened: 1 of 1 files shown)


cylc/flow/config.py
```diff
@@ -1815,8 +1818,9 @@ def generate_taskdef(self, orig_expr, node):
 
     def add_sequence(self, nodes, seq, suicide):
         """Add valid sequences to taskdefs."""
+        graph_node_parser = GraphNodeParser.get_inst()
         for node in nodes:
-            name, offset = GraphNodeParser.get_inst().parse(node)[:2]
```
oliver-sanders, line 1819: Don't call `get_inst` within a for loop. It's cached, but methods aren't free.

cylc/flow/config.py
```diff
@@ -2042,23 +2053,25 @@ def get_graph_raw(
...
         start_point_offset_cache = {}
         point_offset_cache = None
+        graph_node_parser = GraphNodeParser.get_inst()
         for sequence, edges in self.edges.items():
             # Get initial cycle point for this sequence
             point = sequence.get_first_point(start_point)
-            new_points = []
+            new_points = set()
             while point is not None:
-                if point not in new_points:
```
oliver-sanders, line 2050: In comparisons on lists are O(N) vs O(1) for sets.

cylc/flow/config.py
```diff
@@ -2042,23 +2053,25 @@ def get_graph_raw(
...
+                        # Take VIS_N_POINTS cycles from each sequence.
+                        break
+                elif point > stop_point:
                     # Beyond requested final cycle point.
                     break
                 if (workflow_final_point is not None
                         and point > workflow_final_point):
                     # Beyond workflow final cycle point.
                     break
-                if stop_point is None and len(new_points) > self.VIS_N_POINTS:
```
oliver-sanders, line 2059: Refactored to avoid testing `if stop_point is None` twice, functionally equivalent.

cylc/flow/config.py
```diff
@@ -2091,19 +2105,19 @@ def get_graph_raw(
                         l_point = point
                     l_id = (name, l_point)
 
-                    if l_id is None and r_id is None:
-                        continue
-                    if l_id is not None and actual_first_point > l_id[1]:
+                    if actual_first_point > l_point:
                         # Check that l_id is not earlier than start time.
-                        if (r_id is None or r_id[1] < actual_first_point or
```
oliver-sanders, line 2098: Re-arranged the conditions in this `if` statement so that the expensive `r_id[1] < actual_first_point` check comes last (so will only be performed if absolutely necessary).

cylc/flow/config.py
```diff
@@ -2126,7 +2140,10 @@ def get_graph_raw(
             # Flatten nested list.
             graph_raw_edges = (
                 [i for sublist in gr_edges.values() for i in sublist])
-        graph_raw_edges.sort(key=lambda x: [y if y else '' for y in x[:2]])
```
oliver-sanders, line 2129: The biggest speedup: this sort was not needed for `check_circular` purposes.

hjoliver:
That's a pretty good speed up. I think `get_graph_raw` is very old, not surprised it could do with a revamp.

cylc/flow/config.py
```diff
@@ -2067,14 +2080,15 @@ def get_graph_raw(
                         r_id = (right, point)
                     else:
                         r_id = None
-                    if left.startswith('@'):
+                    if left[0] == '@':
```
ChrisPaulBennett, line 2083: Getting really picky here, but is this faster? I'd argue it's less readable

ChrisPaulBennett: Had a bit of a look.
Indexing into the array is O(1).
And according to co pilot, startswith() is O(k), where k is the length of the prefix (in this case 1).
Also, indexing into the list will raise an error if it's empty (But looking at the logic above, I'm not convinced that possible. Anyway, defensive programming would go with the method.
But as I said, I'm being picky so I don't just write "LGTM"

oliver-sanders: Fair question.

Going to be honest, I didn't look into the `startswith` implementation, but yes, indexing seems to be about 3x faster:

```python
$ python -m timeit '"abcdefghijklmnop"[0] == "a"'
10000000 loops, best of 5: 21.2 nsec per loop
$ python -m timeit '"abcdefghijklmnop".startswith("a")'
5000000 loops, best of 5: 74.1 nsec per loop
```

Yes it's not very defensive code, but it can't have length zero. In this code "left" is the left-hand-side of a dependency, e.g. `foo` is the LHS of `foo => bar`, so, due to Cylc graph validation (tasks can't have zero-length names), this will always have length 1 or higher.

Normally I wouldn't go changing something like this, it was fine the way it was before, but this code gets hammered for large graphs, so I've gone with some minimally evil optimisations.

oliver-sanders:
@ChrisPaulBennett, poke

PR #6478, opened 2024-11-14
graphene/graphql-core v3 upgrade

dwsutherland:
partially addresses https://github.com/cylc/cylc-uiserver/issues/333

This is the first part described here:
https://github.com/cylc/cylc-uiserver/issues/333#issuecomment-2463424670

The second, and arguably trickier, part will be the sibling at the UIS.

There were some breaking changes to contend with (mainly around more support for Enums):
https://github.com/graphql-python/graphene/wiki/v3-release-notes

But the most involved part was re-implementing the null stripping, but some time reading the [graphql-core](https://github.com/graphql-python/graphql-core) I found some tools to help, simplifying the code.

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests altered to cover.
- [ ] Changelog entry included if this is a change that can affect users

PR #6781, opened 2025-06-09
Fix CLI flow number help text

MetRonnie:
Noticed that this help text wasn't correct during review of #6478 

**Check List**

- [x] No dependency changes 
- [x] Tests, Changelog not needed for minor docs change

### DIFF
--- a/cylc/flow/flow_mgr.py
+++ b/cylc/flow/flow_mgr.py
@@ -42,11 +42,16 @@ FLOW_NONE = "none"
 def add_flow_opts(parser):
     parser.add_option(
         "--flow", action="append", dest="flow", metavar="FLOW",
-        help=f'Assign new tasks to all active flows ("{FLOW_ALL}");'
-             f' no flow ("{FLOW_NONE}"); a new flow ("{FLOW_NEW}");'
-             f' or a specific flow (e.g. "2"). The default is "{FLOW_ALL}".'
-             ' Specific flow numbers can be new or existing.'
-             ' Reuse the option to assign multiple flow numbers.'
+        help=(
+            f"Assign the task(s) to all active flows using '{FLOW_ALL}'; "
+            f"no flow using '{FLOW_NONE}'; a new flow using '{FLOW_NEW}'; "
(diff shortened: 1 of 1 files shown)


cylc/flow/flow_mgr.py
```diff
@@ -42,11 +42,16 @@
...
+        help=(
+            f"Assign the task(s) to all active flows using '{FLOW_ALL}'; "
+            f"no flow using '{FLOW_NONE}'; a new flow using '{FLOW_NEW}'; "
+            f"or a specific flow number, e.g. '2' (can be new or "
+            "pre-existing). "
+            "By default, tasks already in the active window will keep their "
+            "existing flow(s), whereas tasks outside the active window will "
+            "be assigned to all active flows. "
+            "Reuse the option to assign multiple flow numbers."
+        )
```
hjoliver, line 54: By "new tasks" I think I meant, in current terminology, tasks newly spawned into n=0 by the trigger or set operation. 

At the time of writing, I think we were only changing the flow numbers of those tasks. Existing n=0 tasks were stuck with their existing flow numbers.  I think now we can update the flow numbers of existing tasks, so this text does need updating, but I don't think your new version is 100% correct either - e.g. we can't set no-flow on existing n=0 tasks.

(I'll need to experiment for a few minutes, to check!)

MetRonnie: > tasks newly spawned into n=0 by the trigger or set operation

Isn't that a bit of a tautology? All tasks spawned, by definition, end up in the n=0 window, but the behaviour of what flow they end up in depends on whether they were already n=0 before triggering/setting?

hjoliver: The phrase "tasks newly spawned into n=0" is redundant, but arguably helpfully so. I wouldn't call that tautological, but maybe a linguistic tautology is looser than a logical one. 

> whether they were already n=0 before triggering/setting?

Task already in `n=0` were already spawned, so the phrase does not refer to them.

MetRonnie:
Superseded by https://github.com/cylc/cylc-flow/pull/6395

PR #6789, opened 2025-06-13
tests/f: fix tests which hang on restart timeout

oliver-sanders:
* These tests restart workflows with empty task pools.
* In early Cylc 8, the workflow would have just shut down (empty pool).
* In later Cylc 8, a restart timeout was introduced.
* These tests subsequently hung on the restart timeout and failed (due to hitting the inactivity timeout) when the restart timeout was extended in #6787.


oliver-sanders:
Functional tests passed on master, now rebasing against 8.4.x

dwsutherland (approved):
LGTM - Tests passing

wxtim (approved):
I'm 🦛 With this PR

Issue #6797, opened 2025-06-17
infinite unconfigured submission retries


oliver-sanders:
A niche bug where tasks can enter an infinite submission retry loop even though Cylc was not configured to do this.

The behaviour is intermittent, probably triggered by some internal race condition. It happens as the result of 255 errors to a remote platform.

```cylc
[meta]
  description = """
    Add this into your SSH config file, the test will toggle these lines on and
    off to fix / break the connection.

    # ~/.ssh/config
    Host my-login-node-*  # MARKER
        User nosuchuser  # MARKER
  """


[scheduling]
  initial cycle point =1
  final cycle point =12
  cycling mode = integer
  runahead limit = P10
  [[graph]]
    R1 = """
      fix_ssh => remote_init => break_ssh
    """
    P1 = """
      break_ssh[^] => test
    """

[runtime]
  [[test, remote_init]]
    platform = my-remote-platform   # TODO: changeme
    [[[directives]]]
      -q = shared

  [[fix_ssh]]
    script = """
      sed -i 's/^\(.*\)# MARKER/#\1# MARKER/' "$HOME/.ssh/config"
    """

  [[break_ssh]]
    script = """
      sed -i 's/^#\(.*\)# MARKER/\1# MARKER/' "$HOME/.ssh/config"
    """
```

When I run this, I usually get one or more tasks entering a rapid submission-retry loop. Their submission numbers are being incremented, but the retries never stop.

PR #6798, opened 2025-06-17
fix infinite submission retry bug

oliver-sanders:
* Closes https://github.com/cylc/cylc-flow/issues/6797
* Fixes an issue where multiple tasks being submitted in the same batch could interact via the "bad hosts" set.
* This would result in one task entering an infinite submission retry loop despite this not having been configured.

I fear this one is going to be a right pain to test, or that the test will become a false negative check. The cause was simple variable scope error, so maybe not in need of a direct test?

**Check List**

### DIFF
--- a/changes.d/6798.fix.md
+++ b/changes.d/6798.fix.md
@@ -1,0 +1,1 @@
+Prevent unintended submission reties that could result from platform connection issues in certain circumstances.
--- a/changes.d/6798.fix.md
+++ b/changes.d/6798.fix.md
@@ -1,1 +1,1 @@
-Prevent unintended submission reties that could result from platform connection issues in certain circumstances.
+Prevent unintended submission retries that could result from platform connection issues in certain circumstances.
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -340,10 +340,11 @@ class TaskJobManager:
             # Find the first platform where >1 host has not been tried and
             # found to be unreachable.
             # If there are no good hosts for a task then the task submit-fails.
+            out_of_hosts = False
             for itask in itasks:
                 # If there are any hosts left for this platform which we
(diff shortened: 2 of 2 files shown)


cylc/flow/task_job_mgr.py
```diff
@@ -340,10 +340,11 @@ def submit_livelike_task_jobs(
             # Find the first platform where >1 host has not been tried and
             # found to be unreachable.
             # If there are no good hosts for a task then the task submit-fails.
+            out_of_hosts = False
             for itask in itasks:
                 # If there are any hosts left for this platform which we
                 # have not previously failed to contact with a 255 error.
-                if any(
+                if not out_of_hosts and any(
```
oliver-sanders, line 347: At this point in the code we are iterating over tasks to be submitted to the same platform.

The `out_of_hosts` flag gets set to `True` if we detect that there are no more hosts to submit to. However, this branch sets `out_of_hosts = False` irrespective of whether it had previously been set to `True` short-circuiting the logic.

MetRonnie:
(This will conflict with https://github.com/cylc/cylc-flow/pull/6564, I'm having a look at that)

MetRonnie (approved):

MetRonnie:
Btw, changelog entry?

changes.d/6798.fix.md
```diff
@@ -0,0 +1 @@
+Prevent unintended submission reties that could result from platform connection issues in certain circumstances.
```
MetRonnie:
```suggestion
Prevent unintended submission retries that could result from platform connection issues in certain circumstances.
```

MetRonnie (approved):

wxtim:
Is there a reason you've not tested? It was fairly tricky to replicate reliably and you'd probably have an inherently flaky test?

wxtim (approved):

PR #6800, opened 2025-06-17
Fix flaw in logic for selecting new platform from group

MetRonnie:
### Description

This fixes a small bug where tasks could be recorded in the DB and logged as running on a bad platform when they are actually running on a good platform.

### Repro

Reproduction is a bit flaky; need a good amount of tasks to trigger the problem.

```cylc
# flow.cylc
[task parameters]
    n = 1..20
[scheduling]
    [[graph]]
        R1 = foo<n> & bar<n>

[runtime]
    [[foo<n>]]
        platform = group1
    [[bar<n>]]
        platform = group2
```
```cylc
# global.cylc
[platforms]
    [[bad]]
        hosts = bad_host
[platform groups]
    [[group1]]
        platforms = bad, <valid-remote-bg-platform-1>
    [[group2]]
        platforms = bad, <valid-remote-bg-platform-2>
```
You should see in the logs lines like this:
```
INFO - [1/bar_n06/01:preparing] submitted to bad:background[102190]
```
and at the end
```
WARNING - platform: bad - clean up did not complete
    Unable to find valid host for bad
ERROR - Unable to find a platform from install target bad during remote tidy.
INFO - DONE
```
And also if you run
```console
$ sqlite3 .service/db 'SELECT name, platform_name FROM task_jobs ORDER BY name'
```
you should see `bad` recorded for some tasks in the DB.

### Additional context

I have checked https://github.com/cylc/cylc-flow/issues/6797 is still fixed on this branch

This branch contains a `Tidy` commit, may want to review commits independently

### Check List

- [x] No dependency changes 
- [ ] Tests
- [x] Changelog entry
- [x] Docs n/a

### DIFF
--- a/changes.d/6800.fix.md
+++ b/changes.d/6800.fix.md
@@ -1,0 +1,1 @@
+Fixed a bug where tasks could be erroneously reported as running on an uncontactable platform.


MetRonnie:
@wxtim ready for review

cylc/flow/task_job_mgr.py
```diff
@@ -324,20 +321,35 @@ def submit_livelike_task_jobs(
...
+            )
 
         # Submit task jobs for each platform
         # Non-prepared tasks can be considered done for now:
         done_tasks = bad_tasks
 
-        for _, itasks in sorted(auth_itasks.items()):
-            platform = self._get_platform_with_good_host(itasks, done_tasks)
-            if not platform:
+        for _platform_name, itasks in sorted(platform_itasks.items()):
```
oliver-sanders, line 337: [minor]

```suggestion
        for itasks in sorted(platform_itasks.values()):
```

cylc/flow/task_job_mgr.py
```diff
@@ -334,9 +334,19 @@ def submit_livelike_task_jobs(
...
+            # All tasks in this iteration have the same platform
+            platform = itasks[0].platform
+
+            if self.bad_hosts.issuperset(platform['hosts']):
+                # Out of hosts for this platform.
+                for itask in itasks:
+                    # Get another platform, if task config platform is a group
+                    # (Note there may be tasks with different but intersecting
+                    # platform groups)
+                    if not self._select_new_platform(itask):
```
oliver-sanders, line 347: Because we are batching `itasks` by rtconfig, I think the outcome of `_select_new_platform(itask)` will be the same for all `itasks`.

If so, then the previous batched approach makes more sense.

If not, what variable(s) can cause the outcome to differ?

cylc/flow/task_job_mgr.py
```diff
@@ -555,74 +566,59 @@ def submit_livelike_task_jobs(
...
+            PlatformError(
+                f"{PlatformError.MSG_INIT} (no hosts were reachable)",
+                itask.platform['name'],
             )
-            out_of_hosts = True
-            done_tasks.append(itask)
-
-        return None
+        )
+        done_tasks.append(itask)
```
oliver-sanders: Now this has been extracted into its own method, it makes sense to bump the `done_tasks` addition out into the calling code.

cylc/flow/task_job_mgr.py
MetRonnie: This would change it sort by length of `itasks` whereas we currently are sorting by platform name (Although perhaps we don't need to sort at all? Might upset the tests if we changed it though)

cylc/flow/task_job_mgr.py
MetRonnie: I hoped to make this clear in my comment, but perhaps it needs rewording. We are batched by platform name here, but the tasks in question could have different platform groups. `self._select_new_platform(itask)` could then return different results for the tasks, as it would be looking in different parts of the Venn diagram

image

oliver-sanders: (Ah, because it's grouped by `itask.platform` not `itask.rtconfig['platform']`)

PR #6801, opened 2025-06-19
tui: increase comms timeout

oliver-sanders:
* Some people have encountered timeouts.
* Because Tui uses a global update, the app updates at the rate of the slowest workflow it is connected to.
* So it is advantageous to set a lower timeout to keep the app responsive.
* However, multi-workflow monitoring is less common and it is unhelpful not to be able to view the workflow at all.
* Increasing the timeout from 3 to 5 seconds as a compromise.

**Check List**


wxtim (approved):
- Sane
- Tests passing

Issue #6813, opened 2025-06-24
TODO: this logic fails if task removed after some outputs completed


oliver-sanders:
TODO comment with no issue found in `task_pool.py`:

https://github.com/cylc/cylc-flow/blob/01c2308754c33776f292e521bfde7712f7408997/cylc/flow/task_pool.py#L1804-L1810

It would appear the comment originally read:

```python
            # If itask has any history in this flow but no completed outputs
            # we can infer it was deliberately removed, so don't respawn it.
            # TODO (follow-up work):
            # - this logic fails if task removed after some outputs completed
            # - this is does not conform to future "cylc remove" flow-erasure
            #   behaviour which would result in respawning of the removed task
            # See github.com/cylc/cylc-flow/pull/6186/#discussion_r1669727292
```

@hjoliver, any ideas what needs to be done to resolve this?

oliver-sanders:
I expect this is something we should have re-visited post cylc-remove reimplementation but didn't spot. Hopefully quick?

hjoliver:
I have managed to confirm that the logic encoded by this code block:
  * is still needed, to prevent suicide-triggered tasks from respawning via another dependency
  * is flawed and fails as commented, *"if task removed after some outputs completed"*

However, for the kind of use case that motivated this code block, the second part (logic failure) was hidden by a bug that wipes out prior outputs of newly spawned tasks - described, with a fix, here: #6821

hjoliver:
Why we added this code block:
- it was a quick hack at preventing post-suicide re-spawning without recording suicide in the DB, by inferring indirectly that the task must have been removed
   - (I've shown above that it still does that, for the no-existing-outputs case)
- however it was known to be flawed logic that fails if there are completed outputs, hence the TODO comments
   - (I've confirmed this flaw above, too)

To do this properly, I think we need to record in the DB that the task was removed by suicide trigger and thus should not be respawned.

**I have a good solution for this problem - will post it tomorrow**

hjoliver:
### The Problem

We need to know if a task was removed by suicide trigger, so as not to respawn it later by another dependency. 

There's no way to reliably infer the removal from current task pool information, which is what the offending code block tries to do, imperfectly.

### Proposed Solution

Note that suicide triggers effectively *expire* tasks:
- **suicide**: don't run the task even if it becomes ready
    - (and remove it from n=0 if it is there)
- **expire**: don't run the task even if it becomes ready
[...]

PR #6821, opened 2025-06-30
Don't add new db flow rows if prior history exists.

hjoliver:
Bug found while investigating #6813 

Newly spawned tasks add new rows to the DB, which wipes out prior outputs (e.g. via use of `cylc set`).

This is not as bad as it seems because newly spawned tasks load prior outputs before the new rows are added, so the original outputs get restored to the DB quickly at the next outputs table update.

However, it is still a bug, and there is at least one niche situation that exposes it: if a suicide-triggered task has prior completed outputs (e.g. via `cylc set`):

```cylc
[scheduling]
    [[graph]]
        R1 = """
            setter
            slow:fail? => foo
            slow? => !foo
            foo:x => x
        """
[runtime]
    [[setter]]
        script = "cylc set -o x ${CYLC_WORKFLOW_ID}//1/foo"
    [[slow]]
        script = "sleep 5"
    [[foo]]
        [[[outputs]]]
            x = x
    [[x]]
```
Example: 
- `setter` sets future output `foo:x`, which spawns task `x`
- a little later, `slow` succeeds and suicides `foo` (so `foo` never runs)

In Cylc 8 `foo => ! bar` is usually unnecessary, because `bar` does not get pre-spawned and so does not actually need to be removed. But to continue supporting suicide triggers in Cylc 8 we made it work like this: the suicide trigger spawns `bar` then immediately removes it. This quick spawn-and-remove exposes the bug as described above.

On 8.4.x and master, the workflow runs and leaves this in the DB:
```console
$ sqlite3 -header ~/cylc-run/bug/runN/.service/db 'select * from task_outputs;'
cycle|name|flow_nums|outputs
1|setter|[1]|{"submitted": "submitted", "started": "started", "succeeded": "succeeded"}
1|slow|[1]|{"submitted": "submitted", "started": "started", "succeeded": "succeeded"}
1|x|[1]|{"submitted": "submitted", "started": "started", "succeeded": "succeeded"}
1|foo|[1]|{}  # <------  output x missing
```

On this branch, the manually completed output correctly remains in the DB:
```console
$ sqlite3 -header ~/cylc-run/bug/runN/.service/db 'select * from task_outputs;'
cycle|name|flow_nums|outputs
1|setter|[1]|{"submitted": "submitted", "started": "started", "succeeded": "succeeded"}
1|slow|[1]|{"submitted": "submitted", "started": "started", "succeeded": "succeeded"}
1|foo|[1]|{"x": "(manually completed)"}
1|x|[1]|{"submitted": "submitted", "started": "started", "succeeded": "succeeded"}
```

**Check List**

### DIFF
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1754,6 +1754,9 @@ class TaskPool:
         if itask is None:
             return None
 
+        if prev_status is None:
+            self.db_add_new_flow_rows(itask)
+
         if (
             prev_status is not None
             and not itask.state.outputs.get_completed_outputs()
@@ -1820,7 +1823,6 @@ class TaskPool:
                     for cycle, task, output in self.abs_outputs_done
                 ])
 
-        self.db_add_new_flow_rows(itask)
         return itask
 
     def _spawn_after_flow_wait(self, itask: TaskProxy) -> None:


cylc/flow/task_pool.py
```diff
@@ -1754,6 +1754,9 @@ def spawn_task(
         if itask is None:
             return None
 
+        if prev_status is None:
+            self.db_add_new_flow_rows(itask)
+
```
oliver-sanders, line 1759: With this change, the `db_add_new_flow_rows` operation has jumped up the function a bit here.

However, there are multiple `return` statements between its new and old locations. That means that we will now add new flow rows to the DB in situations where we didn't before, namely:

* `Not respawning {point}/{name} - task was removed`
*  `if prev_status in TASK_STATUSES_FINAL and itask.transient`
* `[task] not spawned: a prerequisite is beyond the workflow stop point`

I don't think we want to do this?

I tested this diff with the example and it gave the same result:

```diff
**diff --git a/cylc/flow/task_pool.py b/cylc/flow/task_pool.py
index 2be68e13a..628b0fb66 100644
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1871,7 +1871,11 @@ class TaskPool:
                     for cycle, task, output in self.abs_outputs_done
                 ])
 
-        self.db_add_new_flow_rows(itask)
+        if prev_status is None:
+            # only add new flow rows if this task has not run before
+            # see https://github.com/cylc/cylc-flow/pull/6821
+            self.db_add_new_flow_rows(itask)
+
         return itask
 
     def _spawn_after_flow_wait(self, itask: TaskProxy) -> None:
```

oliver-sanders: @hjoliver , I've raised this suggestion as a PR, https://github.com/cylc/cylc-flow/pull/6856, hopefully it makes sense.

If not, pick the test from my PR, consider bumping to 8.5.1, and continue development here.

hjoliver: Yes you're right. I think that was a hurried fix that I hadn't got back to yet; will superseded with your better one...

PR #6832, opened 2025-06-30
pyproject: add ruff isort config

oliver-sanders:
Add configuration for the ruff isort plugin.

```console
$ ruff check --select I --fix
```

This is a copy of the plain isort config with one option commented out because it hasn't been implemented, yet.


MetRonnie (approved):
Not merging myself in case you want to change base to 8.4.x

oliver-sanders:
Good spot.

PR #6835, opened 2025-07-01
Reimpliment suicide trigger as expire trigger

### LINKED ISSUE #6813: TODO: this logic fails if task removed after some outputs completed

hjoliver:
Close #6813

This simplifies the system a bit by unifying suicide triggers with task expiry.

Also fixes a long-standing bug: tasks removed by suicide triggers can respawn by other dependencies.

Rationale explained at https://github.com/cylc/cylc-flow/issues/6813#issuecomment-3021516979. Summary:

> On master, suicide triggers just delete tasks from the active pool (n=0)
> - (the bug is: no record of the removal, so suicided tasks can respawn by other dependencies)
> 
> On this branch, suicide triggers just expire tasks
> - expired tasks leave the active pool immediately (which is exactly what suicide is supposed to do)
>   but they're recorded as expired, which stops them respawning
> - every existing suicide test passes as-is
> 
> **The only functional change is**: if a suicide-triggered task also has an :expired trigger, that will now run on suicide - that is (a) very niche; and (b) actually sensible

**Check List**

- [x] [Cylc-Doc issue](https://github.com/cylc/cylc-doc/issues/841) 

### DIFF
--- a/changes.d/6835.feat.md
+++ b/changes.d/6835.feat.md
@@ -1,0 +1,3 @@
+Reformulate suicide triggers to expire tasks.
+Fixes a rare bug that could allow tasks to run after suicide triggering.
+The "expired" output is now completed on suicide triggered tasks.
--- a/changes.d/6835.feat.md
+++ b/changes.d/6835.feat.md
@@ -1,3 +1,4 @@
-Reformulate suicide triggers to expire tasks.
-Fixes a rare bug that could allow tasks to run after suicide triggering.
-The "expired" output is now completed on suicide triggered tasks.
+Reformulate suicide triggers to expire tasks. The task's "expired" output will
+now be completed by the trigger, and will automatically be marked optional,
+but custom completion conditions must be adapted accordingly.
+This also fixes a rare bug that could allow tasks to run after suicide triggering.
--- a/changes.d/6835.feat.md
+++ b/changes.d/6835.feat.md
@@ -1,4 +1,5 @@
-Reformulate suicide triggers to expire tasks. The task's "expired" output will
-now be completed by the trigger, and will automatically be marked optional,
-but custom completion conditions must be adapted accordingly.
-This also fixes a rare bug that could allow tasks to run after suicide triggering.
+A new experimental feature that can be switched on in workflow config:
+Suicide triggers expire tasks rather than just remove them. This fixes
+a bug that could allow tasks to run after suicide triggering. The
+"expired" output will automatically be marked as optional for the
+task, but custom completion conditions must be adapted accordingly.
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -435,7 +435,7 @@ with Conf(
                 This reimplements "suicide triggers" as "expire triggers".
 
                 * When the condition is met, the task will generate the
-                  ``expired`` output rather than just being removed from the pool.
+                  ``expired`` output rather than just being removed.
                 * The triggered task's
                   `flow.cylc[runtime][<namespace>]completion condition`
                   will be automatically modified so that expiry completes the
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -435,7 +435,7 @@ with Conf(
                 This reimplements "suicide triggers" as "expire triggers".
 
                 * When the condition is met, the task will generate the
-                  ``expired`` output rather than being removed from the pool.
+                  ``expired`` output rather than just being removed from the pool.
                 * The triggered task's
                   `flow.cylc[runtime][<namespace>]completion condition`
                   will be automatically modified so that expiry completes the
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -416,6 +416,40 @@ with Conf(
                The default time zone is now ``Z`` instead of the local time of
                the first workflow start.
         ''')
+        with Conf('experimental', desc='''
+            Activate experimental features.
+
+            These are preview features which will become the default in future
+            releases.
+
+            .. versionadded:: 8.6.0
+        '''):
+            Conf('all', VDR.V_BOOLEAN, False, desc='''
+                Activate all experimental features.
+
+                Encouraged for canary testing.
+
+                .. versionadded:: 8.6.0
+            ''')
+            Conf('expire triggers', VDR.V_BOOLEAN, False, desc='''
+                This reimplements "suicide triggers" as "expire triggers".
+
+                * When the condition is met, the task will generate the
+                  ``expired`` output rather than being removed from the pool.
+                * The triggered task's
+                  `flow.cylc[runtime][<namespace>]completion condition`
+                  will be automatically modified so that expiry completes the
+                  task's outputs.
+                * This should be functionally equivalent to "suicide triggers"
+                  in that the triggered task will not run.
+                * However, the triggered task will now be left in the
+                  ``expired`` state making it clearer in the GUI/logs that
+                  the task has been triggered in this way.
+                * It is possible to trigger other tasks off of this ``expired``
+                  output for more advanced failure recovery.
+
+                .. versionadded:: 8.6.0
+            ''')
 
         with Conf(   # noqa: SIM117 (keep same format)
             'main loop',
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -436,10 +436,10 @@ with Conf(
 
                 * When the condition is met, the task will generate the
                   ``expired`` output rather than just being removed.
-                * The triggered task's
-                  `flow.cylc[runtime][<namespace>]completion condition`
-                  will be automatically modified so that expiry completes the
-                  task's outputs.
+                * The ``expired`` output will be marked as :term:`optional`
+                  for the triggered task, but a custom
+                  `flow.cylc[runtime][<namespace>]completion condition
+                  will need to be modified accordingly.
                 * This should be functionally equivalent to "suicide triggers"
                   in that the triggered task will not run.
                 * However, the triggered task will now be left in the
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -1827,7 +1827,7 @@ class TaskPool:
                 # revive as incomplete.
                 msg = "incomplete"
 
-            if LOG.level == logging.DEBUG:
+            if LOG.level <= logging.DEBUG:
                 # avoid unnecessary compute when we are not in debug mode
                 id_ = itask.tokens.duplicate(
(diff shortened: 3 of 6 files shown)


oliver-sanders:
This is a functional change which could potentially have consequences in the nice situation where the same task is clock-expired *and* suicide-triggered. I don't think this is a problem, but we shouldn’t sneak it into a bugfix release.

hjoliver:
I sort of commented on functional vs bug-fix above. I don't mind either way. I'll rebase on to master (actually where I originally did it). It certainly isn't just a bug fix if we document the conceptual difference straight off.

Functionally I don't think it makes any difference except where a task is both suicide triggered and clock-expired (as you note) with something triggered of the "expired" output. It's pretty unlikely that anyone has ever done that IMO - but I could be wrong.

hjoliver:
Rebased to master, for 8.5

cylc/flow/task_pool.py
```diff
@@ -1818,7 +1820,7 @@ def spawn_task(
                 # revive as incomplete.
                 msg = "incomplete"
 
-            if cylc.flow.flags.verbosity >= 1:
+            if LOG.level == logging.DEBUG:
```
hjoliver: Log level does not currently affect verbosity in tests. (This change is correct anyway, because the code block applies specifically to debug log level).

oliver-sanders: The log level is an integer, `logging.DEBUG = 10`, `logging.INFO = 20`, etc.

The handler level can be set to any arbitrary integer:

```suggestion
            if LOG.level <= logging.DEBUG:
```

hjoliver: Fair enough.

Redoing - GitHub UI can't apply suggested changes involving deleted lines...

oliver-sanders:
As discussed on Element, time pressures are a great concern due to the tight timeframes we're working with at the moment. But additionally, we're especially concerned by potential risk of behaviour changes at this particular moment in time which was the big reason for pushing for 8.7.0.

When behaviour changes of this kind are introduced to other tools, they are often released as "opt-in" experimental features to allow the community to evaluate them before rolling out on-mass. If no issues are found, the "experimental" behaviour then becomes the default in a subsequent release. Helps reduce teething pains often experienced in such situations as well as raising awareness of upcoming changes.

I've put together a side PR which puts this change behind an "experimental" flag. This way you can push ahead with it on the 8.6.0 timescale (desired for reasons outlined above) and we can derisk our operational implementation (desired for reasons outlined on the PR): https://github.com/hjoliver/cylc-flow/pull/70

I think this is a good approach in general. Where possible (and of course it isn't possible for all changes), I think that this would be a good way to preview proposed changes to the community going forward. I have added an "all" option to allow users to easily opt-in to all experimental features and will encourage the owners of test workflows to use this (as they are excellent canary testers).

This also avoids the need to strip out all of the remaining "suicide" code in the project "right now". This is good as we haven't figured out how we want to handle the downstream implications of this feature yet (e.g. [cylc graph](https://github.com/cylc/cylc-flow/blob/3706f18bc49fe7a2d0cc40cd128cbd86278b1ac1/cylc/flow/scripts/graph.py#L511-L514), https://github.com/cylc/cylc-ui/issues/1239). We might want to look at renaming / reusing the existing "suicide" field in the graph code going forward?

hjoliver:
I'm confident this change won't cause any real problems, but I think your "experimental" flag is a good idea in general, @oliver-sanders, so I'm keen to go forward with it.

hjoliver:
(Merged the side branch to make this an opt-in "experimental" feature.)

cylc/flow/cfgspec/workflow.py
```diff
@@ -416,6 +416,40 @@ def get_script_common_text(this: str, example: Optional[str] = None):
...
+            ''')
+            Conf('expire triggers', VDR.V_BOOLEAN, False, desc='''
+                This reimplements "suicide triggers" as "expire triggers".
+
+                * When the condition is met, the task will generate the
+                  ``expired`` output rather than just being removed.
+                * The triggered task's
+                  `flow.cylc[runtime][<namespace>]completion condition`
+                  will be automatically modified so that expiry completes the
+                  task's outputs.
```
oliver-sanders: > Note: I think this PR currently has a buggy interaction with the completion condition configuration as the logic added to make expiry optional wont apply if a completion condition is specified.

I've just tried this out with the following example:

```cylc
[scheduler]
  [[experimental]]
    all = True

[scheduling]
  [[graph]]
    R1 = """
      a? => b
      a:fail? => !a
    """

[runtime]
  [[a]]
    completion = succeeded or failed
  [[b]]
```

The result is:

```
WorkflowConfigError: a:expired is permitted in the graph but is not referenced in the completion expression (so is not permitted by it).
Try: completion = "succeeded or failed or expired"
```

Which is reasonable, the discrepancy comes out as a validation error rather than a runtime stall which is good.

I think this is ok, although, we should probably clarify the `WorkflowConfigError` message for this case to make it clear that the "expire trigger" is the cause of this (as the expired output is unlikely to appear in the graph).

Also this bullet point will need to be updated (sorry made an invalid assumption here):

```suggestion
                * The ``expired`` output will be marked as
                  :term:`optional` for the triggered task.
```

changes.d/6835.feat.md
```diff
@@ -0,0 +1,3 @@
+Reformulate suicide triggers to expire tasks.
+Fixes a rare bug that could allow tasks to run after suicide triggering.
+The "expired" output is now completed on suicide triggered tasks.
```
oliver-sanders: (this needs reformulating)

hjoliver: That's still true as written. I'll add a note on custom completion conditions, if that's what you mean.

oliver-sanders: I'm thinking about the experimental flag.

hjoliver: Done.

cylc/flow/cfgspec/workflow.py
```diff
@@ -416,6 +416,40 @@ def get_script_common_text(this: str, example: Optional[str] = None):
                The default time zone is now ``Z`` instead of the local time of
                the first workflow start.
         ''')
+        with Conf('experimental', desc='''
```
dwsutherland, line 419: Think we can have this as a global option too? That way we don't have to alter all our workflows twice (once to test, and once to undo when it goes live)

dwsutherland (approved):
Makes sense to me 👍   .. Just the one comment about making `experimental` a global option (if it isn't already inherited)

cylc/flow/cfgspec/workflow.py
oliver-sanders: Experimental features are for canary testing. I'm not sure it's a good idea to turn on experimental features by default, that way users don't know what features they are opting into so aren't going to be attentive to any issues caused by them.

I did think we should add a CLI flag though to avoid the need to modify the workflows at all.

oliver-sanders:
I put together a workflow that implements the [documented use patterns](https://cylc.github.io/cylc-doc/stable/html/user-guide/writing-workflows/suicide-triggers.html), running them in and out of experimental mode. Comparing the reflogs, all examples run the same both ways :+1:


### ./flow.cylc
```cylc
#!Jinja2

{% from "cylc.flow" import LOG %}

[scheduler]
  allow implicit tasks = True
{% if experimental %}
  {% do LOG.warning('[scheduler][experimental]all = True') %}
  [[experimental]]
    all = True
{% endif %}

[scheduling]
  cycling mode = integer
  initial cycle point = 1
  final cycle point = 2
  runahead limit = P0
  [[xtriggers]]
    fail1 = myxtrig(%(point)s, 1):PT1S
    fail2 = myxtrig(%(point)s, 2):PT1S
  [[graph]]
    P1 = """
{% if pattern == 'basic' %}
      a? => b
      a:fail? => r
      a? => !r
      b | r => c
{% elif pattern == 'flaky' %}
      a? => b? => c?
{% elif pattern == 'multiple' %}
      a? & b? => c
      a? | b? => !c
{% elif pattern == 'xtrigger' %}
      @fail1 => x => !y
      @fail2 => y => !x
{% endif %}
    """

[runtime]
  [[a]]
    script = [[ $CYLC_TASK_CYCLE_POINT == 1 ]]
```

### ./lib/python/myxtrig.py
```python
def myxtrig(current_cycle, fail_at_cycle):
    return (int(current_cycle) != fail_at_cycle, {})
```

### ./run
```bash
#!/usr/bin/env bash

set -euo pipefail

exp_id () {
  pattern="$1"
  experimental="$2"
  local name="trig/$pattern"
  if [[ $experimental == True ]]; then
    name="$name/expire"
  else
    name="$name/suicide"
  fi
  echo "$name"
}

for pattern in basic flaky multiple xtrigger; do
  ids=()
  for experimental in True False; do
    id="$(exp_id "$pattern" "$experimental")"
    ids+=("$id")
    cylc vip . \
      -N \
      --no-run-name \
      --reference-log \
      -n "$id" \
      -s "pattern='$pattern'" \
      -s "experimental=$experimental"
  done
done

for pattern in basic flaky multiple xtrigger; do
  ids=()
  for experimental in True False; do
    id="$(exp_id "$pattern" "$experimental")"
    ids+=("$HOME/cylc-run/$id/reference.log")
  done
  echo diff "${ids[@]}"
  diff "${ids[@]}"
  echo -e '\n\n\n'
done
```

cylc/flow/cfgspec/workflow.py
dwsutherland: global configs can be user specific `~/.cylc/flow/8/global.cylc`  (as we do with out oper/test role users)

hjoliver: Fair point - a user with a lot of workflows might well want to opt in for all of them at once.

We could just recommend against doing it centrally in site config.

dwsutherland: I also wouldn't recommend site wide.

Even if I implement it local to my user, I would have to stop and start (or reload?) my workflows for them to pick it up (I believe).
It would save me CLI or workflow mods to do it this way..

Follow on PRs I suppose

hjoliver: `cylc reload   -g, --global          # also reload global configuration`

hjoliver: (Since 8.5.0 I think)

PR #6836, opened 2025-07-03
6808.fix Platform selection from subshell

### LINKED ISSUE #6808: Subshell platform selection appears to only work for the first task of each name
### Description

Where a config set a platform using a subshell it doesn't appear to be able to change the platform dynamically: The platform selected for the first instance of the task seems to stick.

### Reproducible Example

The following workflow should demonstrate flip-flop platform selection, but doesn't.

```cylc
[scheduling]
    cycling mode = integer
    [[graph]]
        P1 = remote_task[-P1] => toggler => remote_task

[runtime]
    [[remote_task]]
        platform = $(cat ${HOME}/pretend-hall-info)

    [[toggler]]
        script = """
            HPC1=address
            HPC2=address

            # Create pretend-hall-info file if it does not exist
            if [[ ! -f ${HOME}/pretend-hall-info ]]; then
                echo ${HPC1} > ${HOME}/pretend-hall-info
            fi

            # Read the current platform from the file and toggle it
            this=$(cat ${HOME}/pretend-hall-info)
            echo $this
            if [[ $this == "${HPC1}" ]]; then
                echo ${HPC2} > ${HOME}/pretend-hall-info
                cylc message -- "WARNING:changing platform to ${HPC2}"
            else
                echo "${HPC1}" > ${HOME}/pretend-hall-info
                cylc message -- "WARNING:changing platform to '${HPC1}'"
            fi
        """
```

wxtim:
Closes #6808  TL;DR `rtconfig['platform']` was being changed to the result of `platform = $(subshell commands)`, causing the subshell command being run once for each `rtconfig` and that result being fixed.

**Check List**


tests/functional/platforms/12-ping-pong.t
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
```
wxtim, line 1: I tried to write this as an integration test... But couldn't make it play ball.

oliver-sanders:
@wxtim, did you mean to raise this on 8.4.x?

MetRonnie:
I assume so as this is branched off 8.4.x. I've updated the PR

MetRonnie (approved):

cylc/flow/task_job_mgr.py
```diff
@@ -1306,6 +1308,13 @@ def _prep_submit_task_job(
                 # Retry delays, needed for the try_num
                 self._set_retry_timers(itask, rtconfig)
 
+        # Put the original platform and host names back into the
+        # rtconfig. Prevents storing first evaluation of shell expression.
+        if orig_platform_name:
+            rtconfig['platform'] = orig_platform_name
+        if orig_host_name:
+            rtconfig['remote']['host'] = orig_host_name
```
oliver-sanders: With this change, we're still storing the platform in the `rtconfig`, we're just putting it back the way it was after.

Unfortunately, this means the potential for interaction is still there. I think all instances of the task which submit in the same batch will use the same platform?

Because of the `return`s above, I think eternal caching is still possible, e.g, if the shbshell returns a broken platoform:

E.g, take this example and change `functional` to the name of a platform:

```cylc
[scheduling]
  cycling mode = integer
  initial cycle point = 1
  runahead limit = P2
  [[graph]]
    P1 = foo

[runtime]
  [[foo]]
    platform = $(python -c 'import random; import sys; print(random.choice(sys.argv[1:]))' broken1 broken2 broken3 functional)
```

Once you've got some submit-fails, trigger all submit-failed tasks over and over. It will keep picking the same broken platform over and over.

I think we've going to have to pass the platform as an argument and cut out all `rtconfig` manipulation.

wxtim: Reasonable. I did wonder.

oliver-sanders: BTW, I'm not sure what the intended behaviour of platform subshells is when interacting with batched submissions.

E.G, if you have 5 tasks with the same `platform = $(subshell)`:
* Should all 5 go to the same platform (single batched submission -> more efficient).
* Should the subshell be evaluated 5 times (multiple submissions -> more even load).

I would be tempted to say, whatever it's doing at the moment, assume it's right and preserve that behaviour.

MetRonnie: From our operational PoV, I think it's fine for all tasks with the same subshell to only evaluate this once if they are submitted in the same batch, as the subshell simply selects the "live HPC" platform

oliver-sanders: I *suspect* that opps doesn't actually need subshells since platforms are updated by broadcast when changed anyway and switching hall at an arbitrary point in a workflow won't work (unless that arbitrary point just happens to be a data sync task).

I'm guessing the only reason they do this is to provide tasks with a default platform when the workflow is first started? From there in broadcasts take over? If so, we can probs just flatten this out in the config:

```cylc
{% import "os" as os %}
{% set live_hall = os.fdopen(os.open('live-hall-file', os.O_RDONLY)).read() %}

[runtime]
  [[HPC]]
    platform = {{ live_hall }}
```

This would save a lot of subshell calls and make submissions a tad snappier.

MetRonnie: There are tasks that run several hours in advance of the broadcast, to create PBS reservations on the live HPC. So unfortunately the Jinja2 example wouldn't cut it.

I've tested `platform = $( echo evaluated >> ~/platform_subshell.txt )` with one of the workflows and I don't think it gets called very often, as most of the time the broadcast applies.

wxtim:
@MetRonnie - I've dismissed your earlier review since this now works rather differently

cylc/flow/task_job_mgr.py
```diff
@@ -1258,31 +1258,34 @@ def _prep_submit_task_job(
...
                 and rtconfig['platform']
                 and rtconfig['platform'] != platform_name
             ):
                 LOG.debug(
                     f"for task {itask.identity}: platform = "
                     f"{rtconfig['platform']} evaluated as {platform_name}"
                 )
-                rtconfig['platform'] = platform_name
+                # orig_platform_name = rtconfig['platform']
+                # rtconfig['platform'] = platform_name
```
MetRonnie:
```suggestion
```

cylc/flow/task_job_mgr.py
```diff
@@ -1236,10 +1236,10 @@ def _prep_submit_task_job(
                 f"\"{itask.identity}\" the following are not compatible:\n"
             )
 
-        host_n, platform_name = None, None
+        host_name, platform_name = [None] * 2
```
MetRonnie: IMO
```suggestion
        host_name, platform_name = None, None
```
Also:
```
In [1]: %timeit a, b = None, None
14.9 ns ± 0.489 ns per loop ...

In [2]: %timeit a, b = [None] * 2
64 ns ± 0.98 ns per loop ...
```

tests/functional/platforms/13-ping-pong-host.t
```diff
@@ -0,0 +1,41 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-----------------------------------------------------------------------------
+
+# If a task has a platform set using a subshell this should be evaluated
```
MetRonnie:
```suggestion
# If a task has [remote]host=$(subshell) this should be evaluated
```

MetRonnie (approved):
Only linckcheck failing (twice in a row for 2 different links)

wxtim:
@oliver-sanders  Poke

cylc/flow/task_job_mgr.py
```diff
@@ -1242,27 +1242,26 @@ def _prep_submit_task_job(
...
-                host_n is None
+                host_name is None
                 and rtconfig['platform']
                 and rtconfig['platform'] != platform_name
             ):
                 LOG.debug(
                     f"for task {itask.identity}: platform = "
                     f"{rtconfig['platform']} evaluated as {platform_name}"
                 )
-                rtconfig['platform'] = platform_name
```
wxtim, line 1256: For reference in #6990

PR #6838, opened 2025-07-04
Validate workflow and task handler/mail event names

### LINKED ISSUE #2756: handler events list not validated
For generic event handlers, the `handler events` list (in task config) is not checked for validity.  We should have a list of all valid event names held in one place, in `task_events_mgr.py`, and then simply use that as an option list in the suite config spec.

MetRonnie:
Closes #2756

The names set in these settings are now validated:
- `[scheduler][events]handler events`
- `[scheduler][events]mail events`
- `[runtime][X][events]handler events`
- `[runtime][X][events]mail events`

Any invalid event names will raise an `IllegalValueError` or `WorkflowConfigError`.

If there are any outdated Cylc 7 `[scheduler][events]handler/mail events` they are automatically upgraded.

**Check List**

- [x] No dependency changes 
- [x] https://github.com/cylc/cylc-doc/pull/840

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -542,7 +542,7 @@ class WorkflowConfig:
             self.cfg['scheduling']['special tasks'][s_type] = result
 
         self.process_config_env()
-        self._check_and_upg_wflow_handler_events()
+        self._upg_wflow_event_names()
 
         self.mem_log("config.py: before load_graph()")
         self.load_graph()
@@ -560,7 +560,7 @@ class WorkflowConfig:
         self.configure_workflow_state_polling_tasks()
 
         for taskdef in self.taskdefs.values():
-            self._check_task_handler_events(taskdef)
+            self._check_task_event_names(taskdef)
             self._check_task_event_handlers(taskdef)
         self._check_special_tasks()  # adds to self.implicit_tasks
         self._check_explicit_cycling()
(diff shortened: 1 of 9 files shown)


cylc/flow/config.py
```diff
@@ -2672,3 +2725,33 @@ def upgrade_clock_triggers(self):
             taskdef = self.get_taskdef(task_name)
             for seq in taskdef.sequences:
                 taskdef.add_xtrig_label(label, seq)
+
+    def _check_and_upg_wflow_handler_events(self) -> None:
```
wxtim: Can you not just use the `parsec.upgrader.deprecate` logic? You would lose the back-compat mode discrimination, but I'm not sure that's a problem.

oliver-sanders:
> Closes https://github.com/cylc/cylc-flow/issues/2756

Wow, 2018 issue!

Maybe a matter for another PR, but I think `mail events` aren't handled either:

```console
$ cat global.cylc 
[scheduler]
  [[events]]
    mail events = elephant
$ CYLC_CONF_PATH=$PWD cylc config
[scheduler]
    [[events]]
        mail events = elephant
```

<br />

> Both [scheduler][events]handler events and [runtime][X][events]handler events names are now validated.

Parsec has an `options` parameter which allows us to define valid options, e.g:

https://github.com/cylc/cylc-flow/blob/0711cc9cfa43028d29eb08e41df720c4c9188534/cylc/flow/cfgspec/globalcfg.py#L1005-L1007

This provides validation, but also gets pulled through into the config when documented:

https://cylc.github.io/cylc-doc/stable/html/reference/config/global.html#global.cylc[scheduler][host%20self-identification]method

I don't know if this supports `VDR.V_STRING_LIST` configs like this one where multiple options can be picked, but it would be great if it did, especially on the docs side.

<br />

> Can you not just use the parsec.upgrader.deprecate logic

I think parsec only offers `deprecate` and `obsolete` functions. We don't want to deprecate / obsolete the `handler events` config, only particular values of it.

The `deprecate` thinggy accepts a function that can be used to upgrade values from one format to another which is what we want, but I think we would need a new in-situ upgrade function for parsec for this use case.

We could develop this functionality in parsec, it might come in handy again and the closer the functionality is to parsec, the closer it is to the docs. But at the end of the day, meh, whatever works!

cylc/flow/task_events_mgr.py
```diff
@@ -421,7 +421,25 @@ class TaskEventsManager():
     EVENT_EXPIRED = TASK_OUTPUT_EXPIRED
     EVENT_SUBMIT_FAILED = "submission failed"
     EVENT_SUBMIT_RETRY = "submission retry"
+    EVENT_SUBMIT_TIMEOUT = "submission timeout"
+    EVENT_EXEC_TIMEOUT = "execution timeout"
     EVENT_SUCCEEDED = TASK_OUTPUT_SUCCEEDED
+    NON_UNIQUE_EVENTS = ('warning', 'critical', 'custom')
+    STD_EVENTS = [
```
oliver-sanders, line 428: Even if we can't validate these at the parsec level, it would be great to document them there.

I think something like this will make list the options in the same format the parsec autodoc would:

```python
With conf('task events', desc = f'''
    :Options: ``{"``, ``".join(STD_EVENTS)}}``, any custom event

    Configure the task event handling system.
''')
```

Punt to issue if short on time.

cylc/flow/config.py
```diff
@@ -2672,3 +2726,24 @@ def upgrade_clock_triggers(self):
...
+                if event in WorkflowEventHandler.EVENTS_DEPRECATED:
+                    event_names[i] = upgraded[event] = (
+                        WorkflowEventHandler.EVENTS_DEPRECATED[event]
+                    )
+            if upgraded and not cylc.flow.flags.cylc7_back_compat:
+                LOG.warning(
+                    f"{upgrader.DEPR_MSG}\n"
+                    f" * (8.0.0) [scheduler][events][{setting}] "
+                    + ', '.join(f'{k} -> {v}' for k, v in upgraded.items())
+                )
```
oliver-sanders, line 2749: [minor] We could potentially make `depr_options` a dict of `{deprecated_opt: replacement}` and handle this in the parsec layer if we thought this functionality would be reused elsewhere.

cylc/flow/task_events_mgr.py
oliver-sanders: https://github.com/cylc/cylc-flow/issues/6870

Issue #5774, opened 2023-10-18
Simulation Mode: Outputs


wxtim:
Simulation Mode currently causes a task to register all optional outputs and succeeded or failed, depending on settings. 

Behaviour of sim mode tasks with respect to `sumbit`, `started` and optional outputs is poorly defined.

This task is to investigate the exact behaviour of outputs and decide what these should be.

In my opinion succeeded and failed states both imply that the task has submitted and started.

For examples testing this logic see [this discussion](https://github.com/cylc/cylc-doc/pull/641#discussion_r1360651583_)

hjoliver:
I haven't read the preceding discussion very closely, but shouldn't sim mode just do this: all required outputs.

hjoliver:
Randomly choose one?

Skip mode should only be used if the outcome of one-or-another graph branching doesn't matter due to branches rejoining before the end of the skip.  This would typically be the case when skipping to next cycle, say.

And it should be highlighted in docs that if random graph branching (which is the best you can expect Cylc to do automatically) might result in a "wrong" result, you should start a new flow downstream instead of using skip mode.

Do we do the DB lookup in sim mode at the moment? We probably need to in case flow flow-merging needs to occur inside the skip block. Then, manually setting outputs to direct the simulated flow would be another option for users.

dpmatthews:
I'd suggest we apply exactly the same logic that will be used in skip mode. That implies that, by default, no custom outputs will be generated (which is just as valid as all or one - we can't know the purpose of these outputs).
We definitely need to allow the outputs to be configured.

hjoliver:
@dpmatthews - I think we're agreeing, except for the default?

Non-default: ✔️ 
- allow users to specify which outputs to generate in sim mode 

Default: ❓ 
- DM: no custom outputs will be generated
- HO: all custom outputs generated

My thinking is:
-  ALL custom outputs will run all (custom) optional branches, but that's OK in sim mode, and it will pass any completion condition
- NO custom outputs will fail some completion conditions, stalling the workflow

wxtim:
> NO custom outputs will fail some completion conditions, stalling the workflow

Well that's exactly the same as running a non-simulated task without having that task produce the custom outputs.

What about comparing tdef's defined outputs and comparing them to simulation defined outputs and emitting a warning to users?

wxtim:
[...]

PR #6842, opened 2025-07-09
Fix `cylc set` causing erroneous state changes

### LINKED ISSUE #6412: `cylc set` on task no longer in the pool can cause task state regressions

### LINKED ISSUE #6704: set: setting task outputs can change job outputs
The `cylc set --output` command allows us to satisfy **task** outputs, the task state may update to reflect the change. This should have no effect on **job** states, however, it can in some cases.

### Reproducible example

Get a task to fail (e.g. `script=false`) then run `cylc set <task>`.

### LINKED ISSUE #6955: Tasks set to suceeded should not have their time added to task means
Tasks set to succeeded have their run time added to the average run time. They should not be added to the average run time (Skip mode tasks are already exempted).

Example

```cylc
[scheduling]
  cycling mode = integer
  [[graph]]
    P1 = foo
[runtime]
  [[foo]]
    script = sleep 3600
```
```shell
cylc vip .
N=1
while [[ N < 10 ]]; do
  cylc set 'workflow//*/foo'
  sleep 1
  N=$((N+1))
done
```

MetRonnie:
Closes #6704
- `cylc set` no longer affects job states

Closes #6412 
- `cylc set`, on a task that has already succeeded and is no longer in the pool, no longer causes the task state to regress to waiting

Closes #6955
- `cylc set` no longer affects task mean timings info

"High risk" bugfix hence on next minor milestone rather than patch milestone.

**Check List**

- [x] No dependency changes 
- [x] Docs not needed


cylc/flow/run_modes/simulation.py
```diff
@@ -95,22 +99,20 @@ def submit_task_job(
     itask.jobs.append(
         task_job_mgr.get_simulation_job_conf(itask)
     )
-    task_job_mgr.task_events_mgr.process_message(
-        itask, INFO, TASK_OUTPUT_SUBMITTED,
-    )
+    for output in (TASK_OUTPUT_SUBMITTED, TASK_OUTPUT_STARTED):
+        task_job_mgr.task_events_mgr.process_message(itask, INFO, output)
```
oliver-sanders, line 99: This is a functional change in a commit called "Simplify simulation mode stuff"?

The started output is semi-intentionally omitted at present pending a review of how sim mode outputs are generated - https://github.com/cylc/cylc-flow/issues/5774

cylc/flow/task_events_mgr.py
```diff
@@ -965,6 +976,44 @@ def _process_message_check(
...
         return True
 
+    def process_job_message(
+        self,
+        job_tokens: 'Tokens',
+        tdef: 'TaskDef',
+        message: str,
+        event_time: str,
+    ) -> bool:
+        """Process a job message only.
```
wxtim: I'm not clear about the "only"?
Can you say what it's not doing?

```suggestion
        """Process a job message without altering job state.
```

cylc/flow/scheduler.py
```diff
@@ -994,11 +995,20 @@ def process_queued_task_messages(self) -> None:
         # Remaining unprocessed messages have no corresponding task proxy.
         # For example, if I manually set a running task to succeeded, the
         # proxy can be removed, but the orphaned job still sends messages.
-        for tms in messages.values():
-            warn = "Undeliverable task messages received and ignored:"
-            for _, msg in tms:
-                warn += f'\n  {msg.job_id}: {msg.severity} - "{msg.message}"'
-            LOG.warning(warn)
+        warn = ""
+        for tm in unprocessed_messages:
```
wxtim, line 1004: Please either use `task_msg` or put a comment in saying `# tm = task message`

cylc/flow/scheduler.py
```diff
@@ -994,11 +995,20 @@ def process_queued_task_messages(self) -> None:
...
+            job_tokens = self.tokens.duplicate(
+                Tokens(tm.job_id, relative=True)
+            )
```
oliver-sanders: [minor] Job tokens already parsed in the above logic.

Suggest changing the dictionary key from string ID to tokenised ID.

cylc/flow/run_modes/simulation.py
MetRonnie: It is not omitted at present (and the fact it is not omitted looks very much intentional to me):

https://github.com/cylc/cylc-flow/blob/9f7039abd50f61699dc00483c0bb92f04ba5b542/cylc/flow/task_events_mgr.py#L803-L805

https://github.com/cylc/cylc-flow/blob/9f7039abd50f61699dc00483c0bb92f04ba5b542/cylc/flow/task_events_mgr.py#L1524-L1531

https://github.com/cylc/cylc-flow/blob/9f7039abd50f61699dc00483c0bb92f04ba5b542/cylc/flow/task_events_mgr.py#L1555-L1557

This commit does not contain any functional changes to the best of my knowledge

cylc/flow/scheduler.py
MetRonnie: Before I change anything, does considering the fact the editors can provide hover info reduce the need for the change? (increasing length of variable means having to re-wrap lines to placate flake8)

image

cylc/flow/scheduler.py
MetRonnie: Sorry, which dictionary?

cylc/flow/scheduler.py
oliver-sanders: I'm not overly fussed, but just to say that editor features are not an excuse for bad variable names.

cylc/flow/scheduler.py
oliver-sanders: The ids are passed when the task ID is determined above:

```python
tokens = Tokens(task_msg.job_id, relative=True)
# task ID (job stripped)
task_id = tokens.duplicate(job=None).relative_id
messages.setdefault(task_id, [])
```

Using it as the dict key wouldn't work though, sry.

cylc/flow/scheduler.py
MetRonnie: TBH I am in favour of giving `NamedTuple` instances the shortest possible variable name, as their main purpose is a container. What you really care about is accessing the contents (just without the need to unpack the tuple in a specific order).

cylc/flow/run_modes/simulation.py
oliver-sanders: Gotcha, it's explicitly done, but from a different bit of code.

cylc/flow/scheduler.py
```diff
@@ -1005,11 +1006,20 @@ def process_queued_task_messages(self) -> None:
...
+            tdef = self.config.get_taskdef(job_tokens['task'])
+            if not self.task_events_mgr.process_job_message(
+                job_tokens, tdef, tm.message, tm.event_time
+            ):
+                warn += f'\n  {tm.job_id}: {tm.severity} - "{tm.message}"'
+        if warn:
+            LOG.warning(
+                f"Undeliverable task messages received and ignored:{warn}"
+            )
```
hjoliver, line 1014: Coverage says this bit isn't tested - add a quick integration test?

cylc/flow/scheduler.py
```diff
@@ -1005,11 +1006,20 @@ def process_queued_task_messages(self) -> None:
...
-            LOG.warning(warn)
+        warn = ""
+        for tm in unprocessed_messages:
+            job_tokens = self.tokens.duplicate(
+                Tokens(tm.job_id, relative=True)
+            )
+            tdef = self.config.get_taskdef(job_tokens['task'])
+            if not self.task_events_mgr.process_job_message(
+                job_tokens, tdef, tm.message, tm.event_time
+            ):
```
hjoliver, line 1009: An undeliverable task message (no task proxy in the pool) with no job submit number brings the scheduler down here, on this branch, but not on master.

 `cylc message <workflow> 1/a "hello"`  (as opposed to `1/a/01`) with no `1/a` in the pool.

Weirdly, with `1/a` in the pool, this is fine - the submit number seems to default to 01.

That should not happen naturally, but maybe we should protect against it?

```
ERROR - int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/scheduler_cli.py", line 719, in cylc_play
        asyncio.get_running_loop()
        ~~~~~~~~~~~~~~~~~~~~~~~~^^
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/scheduler.py", line 713, in run_scheduler
        await self._main_loop()
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/scheduler.py", line 1690, in _main_loop
        self.process_queued_task_messages()
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/scheduler.py", line 1015, in
    process_queued_task_messages
        if not self.task_events_mgr.process_job_message(
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
            job_tokens, tdef, tm.message, tm.event_time
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        ):
        ^
      File "/home/oliverh/cylc/cylc-flow/cylc/flow/task_events_mgr.py", line 997, in process_job_message
        submit_num=int(job_tokens['job']),
                   ~~~^^^^^^^^^^^^^^^^^^^
    TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

hjoliver (changes requested):

cylc/flow/scheduler.py
hjoliver: ... side PR https://github.com/MetRonnie/cylc-flow/pull/37

cylc/flow/scheduler.py
MetRonnie: Fixed by https://github.com/cylc/cylc-flow/pull/6842/commits/3d5bdc9cac021c2704c7463c519c410567b04bd1

wxtim (approved):

cylc/flow/id.py
```diff
@@ -365,9 +366,24 @@ def is_null(self) -> bool:
...
+
+        Examples:
+            >>> Tokens('//c/t/01').submit_num
+            1
+            >>> Tokens('//c/t').submit_num is None
+            True
+        """
+        if self['job'] is None:
+            return None
+        return int(self['job'])
```
hjoliver:
```suggestion
        return int(self['job']) if self['job'] else None
```
(Minor and arguable, but IMO this is easier to read.)

cylc/flow/rundb.py
```diff
@@ -737,7 +739,7 @@ def select_task_job(self, cycle, name, submit_num=None):
             '''  # nosec B608
             # * table name is code constant
             # * keys are code constants
-            stmt_args = [cycle, name]
+            stmt_args: List[Any] = [cycle, name]
```
hjoliver:
```suggestion
            stmt_args: list[Any] = [cycle, name]
```

cylc/flow/rundb.py
```diff
@@ -714,7 +714,9 @@ def select_task_action_timers(self, callback):
         for row_idx, row in enumerate(self.connect().execute(stmt)):
             callback(row_idx, list(row))
 
-    def select_task_job(self, cycle, name, submit_num=None):
+    def select_task_job(
+        self, cycle: str, name: str, submit_num: Union[str, int, None] = None
+    ) -> Optional[Dict[str, Any]]:
```
hjoliver:
```suggestion
    def select_task_job(
        self, cycle: str, name: str, submit_num: str | int | None = None
    ) -> dict[str, Any] | None:
```

Start using the new syntax ... with possible impact on typing imports, of course. 
[I see you have done this in places].

cylc/flow/task_events_mgr.py
```diff
@@ -187,7 +218,13 @@ def get_event_id(event: str, itask: 'TaskProxy') -> str:
     return event
 
 
-def log_task_job_activity(ctx, workflow, point, name, submit_num=None):
+def log_task_job_activity(
+    ctx: 'SubProcContext',
+    workflow: str,
+    point: Union[str, 'PointBase'],
```
hjoliver:
```suggestion
    point: str | 'PointBase',
```

cylc/flow/task_events_mgr.py
```diff
@@ -187,7 +218,13 @@ def get_event_id(event: str, itask: 'TaskProxy') -> str:
...
+    name: str,
+    submit_num: Union[str, int, None] = None,
```
hjoliver:
```suggestion
    submit_num: str | int | None = None,
```

cylc/flow/task_events_mgr.py
```diff
@@ -705,36 +745,37 @@ def process_message(
...
-            completed_output = (
-                itask.state.outputs.set_message_complete(msg0, forced)
+        task_output, run_signal = split_run_signal(message)
+        job_aborted = (
+            run_signal is not None and task_output == ABORT_MESSAGE_PREFIX
+        )
+        if job_aborted:
+            task_output = TASK_OUTPUT_FAILED
+
+        output_completed: Optional[bool] = False
```
hjoliver:
```suggestion
        output_completed: bool | None = False
```

cylc/flow/task_events_mgr.py
```diff
@@ -1162,29 +1219,12 @@ def _process_job_logs_retrieval(
         if ctx.max_size:
             cmd.append("--max-size=%s" % (ctx.max_size,))
         # Includes and excludes
-        includes = set()
+        includes: Set[str] = set()
```
hjoliver, line 1222:
```suggestion
        includes: set[str] = set()
```

cylc/flow/task_events_mgr.py
```diff
@@ -1328,10 +1368,11 @@ def _retry_task(self, itask, wallclock_time, submit_retry=False):
     def _process_message_failed(
         self,
         itask: 'TaskProxy',
-        event_time: Optional[str],
+        event_time: str,
         message: str,
         forced: bool,
         full_message: str,
+        run_signal: Optional[str] = None,
```
hjoliver:
```suggestion
        run_signal: str | None = None,
```

cylc/flow/task_events_mgr.py
```diff
@@ -1383,16 +1417,32 @@ def _process_message_failed(
         self._reset_job_timers(itask)
         return no_retries
 
-    def _process_message_started(self, itask, event_time, forced):
+    def _process_job_failed(
+        self,
+        itask: 'TaskProxy',
+        event_time: str,
+        run_signal: Optional[str] = None,
```
hjoliver:
```suggestion
        run_signal: str | None = None,
```

cylc/flow/task_job_logs.py
```diff
@@ -40,7 +50,13 @@
 NN = "NN"
 
 
-def get_task_job_log(workflow, point, name, submit_num=None, suffix=None):
+def get_task_job_log(
+    workflow: str,
+    point: Union[str, 'PointBase'],
```
hjoliver:
```suggestion
    point: str | 'PointBase',
```

cylc/flow/task_job_logs.py
```diff
@@ -40,7 +50,13 @@
...
+    name: str,
+    submit_num: Union[str, int, None] = None,
```
hjoliver:
```suggestion
    submit_num: str | int | None = None,
```

cylc/flow/task_job_logs.py
```diff
@@ -40,7 +50,13 @@
...
+    suffix: Optional[str] = None,
```
hjoliver:
```suggestion
    suffix: str | None = None,
```

cylc/flow/task_job_logs.py
```diff
@@ -55,12 +71,23 @@ def get_task_job_log(workflow, point, name, submit_num=None, suffix=None):
     return os.path.join(*args)
 
 
-def get_task_job_activity_log(workflow, point, name, submit_num=None):
+def get_task_job_activity_log(
+    workflow: str,
+    point: Union[str, 'PointBase'],
```
hjoliver:
```suggestion
    point: str | 'PointBase',
```

cylc/flow/task_job_logs.py
```diff
@@ -55,12 +71,23 @@ def get_task_job_log(workflow, point, name, submit_num=None, suffix=None):
...
+    name: str,
+    submit_num: Union[str, int, None] = None,
```
hjoliver:
```suggestion
    submit_num: str | int | None = None,
```

cylc/flow/task_job_logs.py
```diff
@@ -55,12 +71,23 @@ def get_task_job_log(workflow, point, name, submit_num=None, suffix=None):
...
     return get_task_job_log(
-        workflow, point, name, submit_num, JOB_LOG_ACTIVITY)
+        workflow, point, name, submit_num, JOB_LOG_ACTIVITY
+    )
 
 
-def get_task_job_job_log(workflow, point, name, submit_num=None):
+def get_task_job_job_log(
+    workflow: str,
+    point: Union[str, 'PointBase'],
```
hjoliver:
```suggestion
    point: str | 'PointBase',
```

cylc/flow/task_job_logs.py
```diff
@@ -55,12 +71,23 @@ def get_task_job_log(workflow, point, name, submit_num=None, suffix=None):
...
+    name: str,
+    submit_num: Union[str, int, None] = None,
```
hjoliver:
```suggestion
    submit_num: str | int | None = None,
```

cylc/flow/task_message.py
```diff
@@ -69,6 +82,20 @@
 '''
 
 
+def split_run_signal(message: str) -> Tuple[str, Optional[str]]:
```
hjoliver:
```suggestion
def split_run_signal(message: str) -> tuple[str, str | None]:
```

hjoliver (approved):
No problems spotted, just thought we should start taking advantage of the new typing syntax.

cylc/flow/task_events_mgr.py
```diff
@@ -745,9 +786,8 @@ def process_message(
             )
 
         if message == self.EVENT_STARTED:
-            if (
-                flag == self.FLAG_RECEIVED
-                and itask.state.is_gt(TASK_STATUS_RUNNING)
+            if flag == self.FLAG_RECEIVED and itask.state.is_gt(
+                TASK_STATUS_RUNNING
```
hjoliver, line 790: IMO we should avoid making purely style changes that aren't mandated by our agreed CI checks. Otherwise blocks like this might slowly oscillate between different developers personal preferences over time. In this case I think the original is more readable as the "state > running"  bit is not broken up. However, not a biggie, probably not worth changing back. (And I'm sure I'm guilty of the same thing sometimes 🤣 )

MetRonnie: Well we have a black/ruff config in our pyproject.toml. Although it isn't part of CI don't be suprised if autoformatting occurs that blackens/ruffens the code

oliver-sanders: Use a selection range when autoformatting and this won't happen.

MetRonnie: I always use selection ranges otherwise there would be a lot more diff than this! It's just that the selection ranges happened to include these bits

PR #6848, opened 2025-07-11
Allow platform setting to have prefix/suffix around subshell

MetRonnie:
This so that we can do e.g.
```cylc
platform = $(rose config platform hpc_live)-bg
```
Currently we have to resort to
```cylc
platform = $(echo "$(rose config platform hpc_live)-bg")
```

**Check List**

- [x] No dependency changes 
- [x] Tests are included 
- [x] No docs PR applicable

### DIFF
--- a/cylc/flow/platforms.py
+++ b/cylc/flow/platforms.py
@@ -57,7 +57,7 @@ SINGLE_HOST_JOB_RUNNERS = ['background', 'at']
 
 # Regex to check whether a string is a command
 HOST_REC_COMMAND = re.compile(r'(`|\$\()\s*(.*)\s*([`)])$')
-PLATFORM_REC_COMMAND = re.compile(r'(\$\()\s*(.*)\s*([)])$')
+PLATFORM_REC_COMMAND = re.compile(r'(\$\()\s*(.*)\s*(\))')
 
 HOST_SELECTION_METHODS = {
     'definition order': lambda goodhosts: goodhosts[0],
--- a/changes.d/6848.fix.md
+++ b/changes.d/6848.fix.md
@@ -1,0 +1,1 @@
+Allow `flow.cylc[runtime][<task>]platform` setting to have a prefix/suffix around a subshell expression.
--- a/cylc/flow/cfgspec/workflow.py
+++ b/cylc/flow/cfgspec/workflow.py
@@ -1211,6 +1211,7 @@ with Conf(
 
                    # run a command to select the platform (or platform group):
                    platform = $(select-platform)
+                   platform = prefix-$(select-platform)-suffix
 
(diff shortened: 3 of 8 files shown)


oliver-sanders:
The change makes sense, however, platform subshell is really Cylc 7 hangover, `[platforms]` were intended to mitigate the need for this sort of thing.

It looks like this approach is using Rose to configure Cylc with the Rose config edited when `hpc_live` updates?

Could this be done the other way around, editing the Cylc config instead?

```cylc
# global.cylc

[platform groups]
  [[hpc-live]]
    platforms = aaa
  [[hpc-live-bg]]
    platforms = aaa-bg
```

Or even pull in the value from an external source of truth:

```cylc
#!Jinja2

{% import "os" as os %}

[platform groups]
  [[hpc-live]]
    platforms = {{ os.fdopen("hpc_live_file").read() }}
  [[hpc-live-bg]]
    platforms = {{ os.fdopen("hpc_live_file").read() }}-bg

**Comment by MetRonnie** [comment]
> It looks like this approach is using Rose to configure Cylc with the Rose config edited when `hpc_live` updates?

Only using Rose for the simple parsing of the config file (`~/.metomi/rose.conf`) to pick out the `hpc_live` setting. Could use something like `jq` with a JSON config file, but `jq` is not generally installed. The config file itself has a few other settings used for switching HPCs, e.g.
```
[platform]
hpc_live = zone1
hpc_mirror = zone2
hpc_switch = zone3
```

> Could this be done the other way around, editing the Cylc config instead?

This would not pick up changes to `hpc_live` without a reload, which would rather be avoided for an operational suite of dozens of workflows.

Brainstorming here, would it be worth implementing dynamically updatable platforms in some other way, as an 8.6.0 feature?

File: cylc/flow/platforms.py [hunk 1]
```diff
@@ -57,7 +57,7 @@
 
 # Regex to check whether a string is a command
 HOST_REC_COMMAND = re.compile(r'(`|\$\()\s*(.*)\s*([`)])$')
-PLATFORM_REC_COMMAND = re.compile(r'(\$\()\s*(.*)\s*([)])$')
+PLATFORM_REC_COMMAND = re.compile(r'(\$\()\s*(.*)\s*(\))')
```
oliver-sanders, line 60: [minor] can prevent this first group from being captured using `?:`:

```suggestion
PLATFORM_REC_COMMAND = re.compile(r'(?:\$\()\s*(.*)\s*(\))')
```

oliver-sanders (approved):
LGTM.

There is one, very minor flaw in that the regex will match something like `$(foo)-$(bar)` and attempt to run it leaving a cryptic error. But doesn't really matter.

wxtim (approved):

PR #6852, opened 2025-07-15
no data-store n>0 flow num predictions

### LINKED ISSUE #6810: data store: do not project n>0 flow numbers
The data store currently projects flow numbers ahead for future tasks by expanding around the n=0 window and projecting these flow numbers forward.

[We have decided](https://github.com/cylc/cylc-admin/issues/200#issuecomment-2904775098) to drop this feature, all n>0 tasks should have `flow=None`.

See also: https://github.com/cylc/cylc-ui/issues/2179#issuecomment-2903963242

dwsutherland:
closes #6810 

**Before:**
image

**After:**
image

**Check List**


dwsutherland:
TUI SS failing.. Does it show flows now?

oliver-sanders:
I checked the Tui test failures, they were caused by tasks changing colour (as expected), actually a good test that this feature is working as intended, example diff:

Screenshot from 2025-07-15 15-36-11

Pushed the rebuilt screenshots to this branch.

tests/integration/test_data_store_mgr.py
```diff
@@ -652,7 +652,7 @@ async def test_flow_numbers(flow, scheduler, start):
         # the task should exist in the original flow
```
MetRonnie:
```suggestion
        # the task should not have a flow number as it is n>0
```

cylc/flow/data_store_mgr.py
```diff
@@ -1254,7 +1254,7 @@ def generate_ghost_task(
                 self.id_,
                 self.schd.config.get_taskdef(name),
                 point,
-                flow_nums,
+                None,
```
MetRonnie: For clarity
```suggestion
                flow_nums=None,
```

oliver-sanders:
> There is still a protobuf task proxy created with the value of the flow_nums arg: 

Great spot!

dwsutherland:
> no-flow tasks go transparent and the flow numbers are visible in the context menus

This is actually the result I wanted (about n>0 being translucent to make it different from active jobs, (although past tasks are still bold))

dwsutherland:
> here is still a protobuf task proxy created with the value of the flow_nums arg

Yes that's fine .. the created `TaskProxy` object takes `flow_nums=None` and I assume translates it to an empty set.

cylc/flow/data_store_mgr.py
```diff
@@ -1218,7 +1213,6 @@ def generate_ghost_task(
         self,
         tokens: Tokens,
         point: 'PointBase',
-        flow_nums: 'FlowNums',
         is_parent: bool = False,
         itask: Optional['TaskProxy'] = None,
         n_depth: int = 0,
```
MetRonnie, line 1218: Is the `replace_existing` arg below (added in https://github.com/cylc/cylc-flow/pull/6115) still needed?

MetRonnie (approved):
LGTM, but have 1 question

dwsutherland:
> > should go ahead and assign a default waiting (or other) state
> 
> I think:
> 
> * If there **is** an `itask` in the pool, use its status.
> * Else, if there is a record in the DB, use it's status.
> * Else, default to `waiting`.
> 
> I would avoid setting a generic default of `waiting` as `itask` objects and DB entries should always have a status so we shouldn't need a more global default here, if there is a bug in either of these, we wouldn’t necessarily want the data store to mask it.
> 
> Lets do this in another PR though.

This is already how it works, except the assignment to waiting happens after the DB search.. and the DB search is batched, so there is time between the creation of the protobuf object and the DB search..

If the `itask` is not in the pool, we create a dummy `itask` (which is by default `waiting` state).. We then batch these dummy `itask` and check/update these `itask` with DB data (if any)... So there is a slither or time between immediate creation of active tasks, and the batch processing of `n>0` ..

cylc/flow/data_store_mgr.py
dwsutherland: Yes, it's used to change/update an n>0 task to an active task (n=0).

I understand that you'd think that these changes should have come in the form of deltas, however, if a tasks information was populated by the DB then when it becomes live again it may have changed since then (maybe there was a reload or broadcast/runtime change)..

oliver-sanders:
> So it's not a bug, it's more a consequence of us employing batching for DB lookups (to reduce the amount of queries hammering the DB for every task in the n>0 window)

Shouldn't the data store manager just wait for complete information before adding the task to the store?

cylc/flow/data_store_mgr.py
oliver-sanders: Can confirm there are calls with both `replace_existing = False` and `True`.

PR #6854, opened 2025-07-15
demote warnings

oliver-sanders:
Some more warnings that I think are likely to frustrate users.

I don't think these should be controversial.

**Please bump to 8.6.0 rather than delay 8.5.0 if it comes to it.**

**Check List**

### DIFF
--- a/cylc/flow/commands.py
+++ b/cylc/flow/commands.py
@@ -419,7 +419,7 @@ async def reload_workflow(schd: 'Scheduler', reload_global: bool = False):
         # logged by the TaskPool.
         add = set(schd.config.get_task_name_list()) - old_tasks
         for task in add:
-            LOG.warning(f"Added task: '{task}'")
+            LOG.info(f"Added task: '{task}'")
         schd.workflow_db_mgr.put_workflow_template_vars(schd.template_vars)
         schd.workflow_db_mgr.put_runtime_inheritance(schd.config)
         schd.workflow_db_mgr.put_workflow_params(schd)
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -2351,7 +2351,7 @@ class WorkflowConfig:
                     suicides += 1
 
         if suicides and not cylc.flow.flags.cylc7_back_compat:
-            LOG.warning(
+            LOG.info(
                 f"{suicides} suicide trigger(s) detected. These are rarely "
                 "needed in Cylc 8 - see https://cylc.github.io/cylc-doc/"
                 "stable/html/7-to-8/major-changes/suicide-triggers.html"
--- a/cylc/flow/run_modes/skip.py
+++ b/cylc/flow/run_modes/skip.py
@@ -188,4 +188,4 @@ def skip_mode_validate(taskdefs: 'Dict[str, TaskDef]') -> None:
         message = 'The following tasks are set to run in skip mode:'
         for taskname in skip_mode_tasks:
             message += f'\n    * {taskname}'
-        LOG.warning(message)
+        LOG.info(message)
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -2179,7 +2179,7 @@ class Scheduler:
             return
         if not self.is_paused:
             if not quiet:
-                LOG.warning("No need to resume - workflow is not paused")
+                LOG.info("No need to resume - workflow is not paused")
             return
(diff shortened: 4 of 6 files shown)


cylc/flow/commands.py
```diff
@@ -419,7 +419,7 @@ async def reload_workflow(schd: 'Scheduler', reload_global: bool = False):
         # logged by the TaskPool.
         add = set(schd.config.get_task_name_list()) - old_tasks
         for task in add:
-            LOG.warning(f"Added task: '{task}'")
+            LOG.info(f"Added task: '{task}'")
```
oliver-sanders, line 422: Tasks which have been added by reload are currently logged at the warning level.

This shouldn't be a warning:
* This is design behaviour, Cylc has done what the user asked.
* We don't expect the user to review this list.
* There are no required actions.

We agreed to remove the opposing `Removed task` message in Tim's PR: https://github.com/cylc/cylc-flow/pull/6765/files#r2104444586

cylc/flow/config.py
```diff
@@ -2351,7 +2351,7 @@ def _proc_triggers(self, parser, seq, task_triggers):
                     suicides += 1
 
         if suicides and not cylc.flow.flags.cylc7_back_compat:
-            LOG.warning(
+            LOG.info(
```
oliver-sanders, line 2354: Suicide triggers are sometimes needed and there is no way for users to opt out of this warning if they are. So it's harsh to force it on them every time they run their workflow.

cylc/flow/run_modes/skip.py
```diff
@@ -188,4 +188,4 @@ def skip_mode_validate(taskdefs: 'Dict[str, TaskDef]') -> None:
         message = 'The following tasks are set to run in skip mode:'
         for taskname in skip_mode_tasks:
             message += f'\n    * {taskname}'
-        LOG.warning(message)
+        LOG.info(message)
```
oliver-sanders, line 191: We (Met Office) added this warning message in the skip mode proposal. The intention was to reduce the potential for users configuring tasks in skip mode (e.g. for testing), and forgetting to take them out of skip mode (e.g. for production).

I recently saw a workflow where dummy tasks were parameterised making this list rather long.

This warning was a bit dubious when we added it (there was some discussion at the time). On reflection, and with the new context of the warning triangles, I'm happy to demote it.

I think only Dave and I would be inclined to see this as controversial?

cylc/flow/scheduler.py
```diff
@@ -2179,7 +2179,7 @@ def resume_workflow(self, quiet: bool = False) -> None:
             return
         if not self.is_paused:
             if not quiet:
-                LOG.warning("No need to resume - workflow is not paused")
+                LOG.info("No need to resume - workflow is not paused")
```
oliver-sanders, line 2182: User tried to resume a workflow that was already running, e.g:

* They hit the play button twice.
* The GUI lagged.
* They ran the `cylc play` command against a workflow without checking whether it was running first.

In any case, it's not really a warning, Cylc has done what the user asked (the workflow is not paused).

cylc/flow/run_modes/skip.py
MetRonnie: Should re-word this then: https://github.com/cylc/cylc-flow/blob/7bf22339b297a7a7a93b17daf6332fc6cd2c1597/cylc/flow/cfgspec/workflow.py#L1333-L1340

hjoliver (approved):
These changes are good IMO.

cylc/flow/run_modes/skip.py
oliver-sanders: I think the note is still fine? The warning is no longer a WARNING but the content is still valid.

MetRonnie: Well it's not warning you anymore, and in your own words the fact that it was a warning was dubious to begin with.

MetRonnie (approved):

cylc/flow/run_modes/skip.py
hjoliver: Sorry, I didn't notice this unresolved question before merging with two approvals.

I guess the point is *"Cylc will warn you about any tasks set to run in skip mode"* should now be changed to *"Cylc will flag any tasks..."* (or similar)?

MetRonnie: Well, not a biggie

PR #6859, opened 2025-07-18
Pre batch processing task state (Option 2)

### LINKED ISSUE #6567: data store: blank task state

dwsutherland:
closes #6567, closes #6860, closes #6861

**Note:** The following view, will almost never been seen (otherwise TUI would've be crashing all the time previously), and has been generated by commenting out this batch block:
image

With that said, momentarily the pre batch processed DB would show:
image

(The WUI should change the ICON to `(?)` or something appropriate)

description
```
    TASK_STATUS_UNKNOWN:
        'Job status yet to be determined.'
```

**Check List**

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -114,6 +114,7 @@ from cylc.flow.task_state import (
     TASK_STATUS_RUNNING,
     TASK_STATUS_SUCCEEDED,
     TASK_STATUS_FAILED,
+    TASK_STATUS_UNKNOWN,
     TASK_STATUSES_ORDERED
 )
 from cylc.flow.task_state_prop import extract_group_state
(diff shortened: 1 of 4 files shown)


oliver-sanders:
Really don't want to go down the route of introducing an "unknown" task state to cover up a bug.

But strangely, I've managed to reproduce the bug on this branch too. Something stranger is at work...

I'll close this to help tidy up open PRs.

PR #6861, opened 2025-07-18
(no) Pre batch processing task state (Option 3)

### LINKED ISSUE #6567: data store: blank task state

dwsutherland:
closes #6567 

Tasks batched (for DB read) during a graph walk, and processed at the end of that walk.

Window resizes typically include many walks (for each active task), so the DB will have that many calls.

**Check List**

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -1143,6 +1143,9 @@ class DataStoreMgr:
                 self.prune_trigger_nodes[active_id])
             del self.prune_trigger_nodes[active_id]
 
+        # load database history for flagged nodes
+        self.apply_task_proxy_db_history()
+
     def generate_edge(
         self,
         parent_tokens: Tokens,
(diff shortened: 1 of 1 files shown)


oliver-sanders:
>> If I'm understanding correctly, the root cause is that the data store is trying to push out incomplete deltas before it has actually gathered all of the required information. This (only?) happens on reload.
>
> it applies stateless deltas before the historical batch is applied, yes.

I'm not convinced this is the case.

The data store shouldn't yield any deltas until the process has completed (both logically and from the code):

https://github.com/cylc/cylc-flow/blob/a12a45855fb5d86c742c797a34ee62e594e50856/cylc/flow/data_store_mgr.py#L707

I have also managed to reproduce this example without any historical data to load - https://github.com/cylc/cylc-flow/issues/6567#issuecomment-3778291392

Unfortunately, the issue still reproduces on this branch :(

oliver-sanders:
(unassigning this to re-focus my review queue, feel free to re-assign when undrafted)

PR #6874, opened 2025-07-23
Validation of order of initial/start/stop/final cycle points

### LINKED ISSUE #6866: Lack of validation for start cycle point and stop after cycle point
### Description

There is not hint of a problem if you:
- Have a start cycle point that is after the stop after cycle point, or after the final cycle point
- Have a stop after cycle point that is before the initial cycle point

### Expected Behaviour

This should be raise an input error?

wxtim:
Closes #6866 

Follow-up to #4302

Apparently there was no systematic validation of whether users were supplying sensible values for these, except in a small number of cases.

In this PR, I write a thorough set of cases and ensure that they all pass.

Currently this targets 8.4.x, but I understand that given the refactoring it might be sensible enough to target 8.6.0.

**Check List**

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -491,6 +491,10 @@ class WorkflowConfig:
         self.process_start_cycle_point()
         self.process_final_cycle_point()
         self.process_stop_cycle_point()
+        if self.start_point:
+            self.cycle_point_warning('start', 'before', 'initial')
+            self.cycle_point_warning('start', 'after', 'final')
+            self.cycle_point_warning('stop', 'before', 'start')
 
         # Parse special task cycle point offsets, and replace family names.
         LOG.debug("Parsing [special tasks]")
@@ -813,6 +817,7 @@ class WorkflowConfig:
             if self.cycling_type == ISO8601_CYCLING_TYPE:
                 self.options.startcp = _parse_iso_cycle_point(startcp)
             self.start_point = get_point(self.options.startcp).standardise()
+
         elif starttask:
             # Start from designated task(s).
             # Select the earliest start point for use in pre-initial ignore.
@@ -830,6 +835,24 @@ class WorkflowConfig:
         else:
             # Start from the initial point.
             self.start_point = self.initial_point
+
+    def cycle_point_warning(self, label1, order, label2):
+        """Centralized logic for warning that start, stop, initial and
+        final cycle points are not sensibly ordered.
+        """
+        point1 = getattr(self, label1 + '_point')
+        point2 = getattr(self, label2 + '_point')
+        if (
+            order == 'before' and point1 < point2
+            or order == 'after' and point1 > point2
+        ):
+            msg = (
+                f"{label1} cycle point '{point1}' will have no effect as"
+                f" it is {order} the {label2} cycle point '{point2}'."
+            )
+            LOG.warning(msg)
+            return True
+        return False
 
     def process_final_cycle_point(self) -> None:
         """Validate and set the final cycle point from flow.cylc or options.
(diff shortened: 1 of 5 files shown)


wxtim:
Broken the one test which did test any case

cylc/flow/config.py
```diff
@@ -815,6 +820,25 @@ def process_start_cycle_point(self) -> None:
...
+        point2 = getattr(self, label2 + '_point')
+        if (
+            order == 'before' and point1 < point2
+            or order == 'after' and point1 > point2
+        ):
+            msg = (
+                f"{label1} cycle point '{point1}' will have no effect as"
+                f" it is {order} the {label2} cycle point '{point2}'."
+            )
+            # Capitalize() makes the ISO strings look very odd:
```
MetRonnie: Sorry, what does this mean?

cylc/flow/config.py
```diff
@@ -815,6 +820,25 @@ def process_start_cycle_point(self) -> None:
...
+        """
+        point1 = getattr(self, label1 + '_point')
+        point2 = getattr(self, label2 + '_point')
+        if (
+            order == 'before' and point1 < point2
+            or order == 'after' and point1 > point2
+        ):
+            msg = (
+                f"{label1} cycle point '{point1}' will have no effect as"
+                f" it is {order} the {label2} cycle point '{point2}'."
```
MetRonnie, line 851: I think there is conflation of 2 different cases here.

1. Valid but useless X cycle point. E.g.:
    > start cycle point '1' before the initial cycle point '2'

    This should give a warning.

2. Invalid X cycle point. E.g.:
    > stop cycle point '1' before the initial cycle point '2'

    I think this ought to raise an input error.

cylc/flow/config.py
wxtim: I'm explaining why I've done the slightly strange `x[0].upper + x[1:]` thing, rather than using the built in string method `capitalize`.

```python
>>> 'foo bar baz 20250101T0000Z'.capitalize()
'Foo bar baz 20250101t0000z'
```

cylc/flow/config.py
wxtim: There are already errors raised elsewhere for seriously bad cases (mostly final < initial).

I can't see, however why you think these two cases should be treated differently - Neither make a lick of sense, but I don't think one is more clearly wrong than the other?

cylc/flow/config.py
wxtim:
```suggestion
            # str.capitalize() will turn T & Z in isodate strings lowercase
```

cylc/flow/config.py
MetRonnie: Case 2 is a user error that will result in the workflow shutting down immediately. It would be more helpful to flag up this error instead of just warning?

cylc/flow/config.py
MetRonnie: I don't see any need to capitalize the message in the first place

cylc/flow/config.py
wxtim: > will result in the workflow shutting down immediately

I think that's less dangerous than case 1 where the workflow runs, but if the user intended startcp=10 it won't work as expected, but will do so quietly.

cylc/flow/config.py
wxtim: English grammar.

MetRonnie: At the cost of code simplicity/readability?

cylc/flow/config.py
MetRonnie: I think I am happy for case 1 to raise an error too. However we may want to discuss this

cylc/flow/config.py
hjoliver: >```
> > >>> 'foo bar baz 20250101T0000Z'.capitalize()
> > 'Foo bar baz 20250101t0000z'
>```
Not the point of your example, by since you've done it: task name case should not be messed with like this because task names are case sensitive: `Foo` could be an entirely different task from `foo`.

I guess the log message under discussion here starts with e.g. *initial cycle point*. Capitalizing that is less consequential (than task names), but I still don't think it should be done because `initial cycle point` is a case sensitive config item (I know the log message could be taken as referring to the concept, not the literal config item, but there's still some potential to confuse the user there).

cylc/flow/config.py
```diff
@@ -815,6 +820,25 @@ def process_start_cycle_point(self) -> None:
...
+            )
+            # Capitalize() makes the ISO strings look very odd:
+            LOG.warning(msg[0].upper() + msg[1:])
```
wxtim:
```suggestion
            LOG.warning(msg)
```

oliver-sanders:
(assigning commenters as reviewers)

wxtim:
Bump

MetRonnie:
Closing this one, someone will need to revisit this

PR #6875, opened 2025-07-24
Cylc Review: support alternate home dirs.

hjoliver:
This is a minimal change to support viewing symlinked run dirs directly instead of via $HOME.

At ESNZ our HPC home directories now have to be private, but we have shared project dirs at the end of our `cylc install` symlinks. This change allows a Cylc Review service running as a project member to see the run dirs of all project members, via the symlinked dirs instead of `$HOME`.

From `cylc review --help` on this branch:

```
If $CYLC_REVIEW_HOME is defined, look there for run directories instead of
of user home directories. This can be used to view symlinked run directories
directly when home directories are private. For example with symlinking as:
    /home/USER/cylc-run/WORKFLOW -> /project/PROJECT/USER/cylc-run/WORKFLOW
set the environment variable as:
    CYLC_REVIEW_HOME=/project/PROJECT
```

There were several places in the code where home dirs were assumed. To avoid any chance of breaking existing use cases, I just banged handling of `$CYLC_REVIEW_HOME` in first, in those locations: if the variable exists, do it the new way; else do it the old way.

**Check List**

### DIFF
--- a/bin/cylc-review
+++ b/bin/cylc-review
@@ -23,8 +23,15 @@ logs via an HTTP interface.
 
 With no arguments, the status of the ad-hoc web service server is printed.
 
-For 'cylc review start', if 'PORT' is not specified, port 8080 is used."""
+For 'cylc review start', if 'PORT' is not specified, port 8080 is used.
 
+If $CYLC_REVIEW_HOME is defined, look there for run directories instead of
+in user home directories. This can be used to view symlinked run directories
+directly when home directories are private. For example with symlinking as:
+  /home/USER/cylc-run/WORKFLOW -> /project/PROJECT/USER/cylc-run/WORKFLOW
+set the environment variable as:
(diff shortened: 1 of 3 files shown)


wxtim (approved):
LGTM. Consider whether this should be rebased onto 8.7.0 Cylc review ticket.

oliver-sanders:
(note, since there'll be no further Cylc 7 releases there might not be much benefit to merging&releasing this one *now* since the ESNZ release is already deployed and patched)

oliver-sanders:
The Cylc Review code has now moved to cylc-uiserver, so this patch will need to be re-applied on the new codebase.

I've opened an issue in cylc-uiserver to make sure this isn't forgotten: https://github.com/cylc/cylc-uiserver/issues/837

Closing this PR.

hjoliver:
Superseded by https://github.com/cylc/cylc-uiserver/pull/840

Issue #2972, opened 2019-02-28
CLI refactor


oliver-sanders:
With Cylc8 and the upgrade to Python3 (which brings with it `argparse`) we have the opportunity (read excuse) to make some meaningful changes to the CLI.

This issue is for collating (and hashing out the details of) changes to make.

My suggestions:

1. **Use argparse**
    * Shouldn't be contriversial, may do this on the Python3 branch.
2. **Migrate Cylc subcommands to Python** so we can use argparse for them all.
    * Perhaps controversial, we may want to exclude internal cylc commands for performance reasons.
    * Move the scripts into `cylc.cli` or something of that ilk.
    * This ties in with Bruno's packaging efforts.
3. **Use argparse to implement cylc subcommands** as [proper subcommands](https://docs.python.org/3/library/argparse.html#sub-commands).
4. **Use [argument groups](https://docs.python.org/3/library/argparse.html#argument-groups)** to rationalise help pages.
   * Lots of commands group options, `pytest` is a good example.
   * E.g. we could have a format group in which to group options like `--format=json`, `--terse` and `--color=never`.
5. **Collect the various "color" options**
    * `--bold`, `--no-bold`, `--color`, `--no-color`
    * Make them orthogonal and standardise them across cylc subcommands using the standard(ish) convention `--color=WHEN`.
    * Use colour automatically when the terminal supports it.
    * Make more use of colour (tastefully not just for the sake of it) in user commands.
    * Use colour in Cylc loggers when output to a terminal? (e.g. grey for debug, orange for warnings, red for errors, purple for critical)
    * See the changes to `cylc scan` on my Python3 branch.
6. **Collect the various "format" options**
    * `--raw`, `--plain`, `--json`, `--terse`
    * Make them orthogonal and standardise them across cylc subcommands as `--format=FMT, -t FMT`.
    * See the changes to `cylc scan` on my Python3 branch.
7. Consider **moving some options onto the `cylc` command** itself rather than the subcommands.
    * `--host` might be a good candidate i.e. `cylc --host=myhost subcommand --subcmdoption`.
    * I think this may allow us to perform re-infocation *before* parsing the sub-command options.
8. **Implement a simple batch mode** on the `cylc` command
    ```console
        $ cylc -s foo bar baz/* stop --kill
        cylc stop foo --kill
        cylc stop bar --kill
        cylc stop baz/bool --kill
        cylc stop baz/been --kill
    ```
[...]

PR #6892, opened 2025-07-30
Add `--format=json` option to `cylc broadcast` command

### LINKED ISSUE #6873: `cylc broadcast --display` JSON output
It should be trivially simple to add a `--json` option to `cylc broadcast` so that when using the `--display` and `--display-task` options, you can output the result in JSON.

There is already a `--raw` option that outputs in a Python dict format, the code for this should be able to be adapted with a `json.dumps()` call.

MetRonnie:
Closes #6873

**Check List**

- [x] No dependency changes 
- [x] Tests are included 
- [x] No docs needed
- [x] Very low risk feature hence on 8.5.x branch

### DIFF
--- a/cylc/flow/scripts/broadcast.py
+++ b/cylc/flow/scripts/broadcast.py
@@ -111,6 +111,11 @@ if TYPE_CHECKING:
     from optparse import Values
 
 
+RAW_DEPR_MSG = (
+    "DEPRECATED: the --raw option will be removed at Cylc 8.7; "
+    "use --format=raw instead."
+)
+
 REC_ITEM = re.compile(r'^\[([^\]]*)\](.*)$')
 
 MUTATION = '''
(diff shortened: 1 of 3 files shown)


cylc/flow/scripts/broadcast.py
```diff
@@ -318,6 +318,17 @@ def get_option_parser() -> COP:
...
+    parser.add_option(
+        '--json',
+        help=(
+            "With -d/--display or -k/--display-task, write out "
+            "the broadcast config structure in JSON format."
+        ),
+        action='store_true',
+        default=False,
+        dest='json',
+    )
```
oliver-sanders: In an attempt to standardise the Cylc CLI options, can we have `--format=json`, `-t=json`.

Context: https://github.com/cylc/cylc-flow/issues/2972

cylc/flow/scripts/broadcast.py
```diff
@@ -312,12 +317,33 @@ def get_option_parser() -> COP:
...
-             "the broadcast config structure in raw Python form.",
+        help=(
+            "With -d/--display or -k/--display-task, write out "
+            "the broadcast config structure in raw Python form. "
+            f"{RAW_DEPR_MSG}"
+        ),
         action="store_true", default=False, dest="raw")
 
+    parser.add_option(
+        '--format',
```
oliver-sanders, line 334: Consider adding the short option `-t`.

oliver-sanders (approved):
LGTM

cylc/flow/scripts/broadcast.py
MetRonnie: I had a look at scripts that we might implement `--format` for:
- `broadcast`: No conflict for `-t` in this case. However...
- `dump`: `-t` is `--tasks`
- `list`: `-t` is `--tree`
- `play`: (which already has `--format`) `-t` is `--start-task`
- `graph`: `-t` is `--transpose`
- `workflow-state`: `-t` is `--task`

So I'm not sure there is going to be consistency with `scan`

wxtim (approved):

PR #6903, opened 2025-08-06
restart timeout: fire event for stop after cycle point scenarios

oliver-sanders:
* If a workflow hits the `final cycle point` and is restarted, it is considered complete.
* If a `restart timeout` is configured, the scheduler will stay up for a configured period.
* However, if a workflow hits the `stop after after cycle point` and is restarted, the `restart timeout` event is not yielded and the scheduler will instantly shut down again.

This has been reported as a problem for people following this design pattern: https://cylc.github.io/cylc-doc/stable/html/user-guide/examples/extending-workflow/index.html

**Check List**

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -555,8 +555,22 @@ class Scheduler:
                     timer.reset()
                 self.timers[event] = timer
 
-        if self.is_restart and not self.pool.get_tasks():
-            # This workflow completed before restart; wait for intervention.
+        if self.is_restart and (
+            # workflow has completed
+            not self.pool.get_tasks()
+            # workflow has hit the "stop after cycle point"
+            or (
+                self.config.stop_point
+                and all(
+                    cycle > self.config.stop_point
+                    for cycle in {
+                        itask.point for itask in self.pool.get_tasks()
+                    }
+                )
+            )
+        ):
+            # This workflow will shut down immediately once restarted
+            # => Give the user a grace period to intervene first
             with suppress(KeyError):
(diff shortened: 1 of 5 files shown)


tests/integration/test_workflow_events.py
```diff
@@ -188,6 +196,46 @@ async def test_restart_timeout(test_scheduler, scheduler, run, complete):
     assert schd2.get_events() == {'startup', 'restart timeout', 'shutdown'}
 
 
+async def test_restart_timeout_workflow_stop_after_cycle_point(
```
oliver-sanders, line 199: Copy of above test, shimmed to test `stop after cycle point`.

hjoliver (approved):
👍

cylc/flow/scheduler.py
```diff
@@ -555,8 +555,22 @@ async def configure(self, params):
                     timer.reset()
                 self.timers[event] = timer
 
-        if self.is_restart and not self.pool.get_tasks():
-            # This workflow completed before restart; wait for intervention.
+        if self.is_restart and (
+            # workflow has completed
+            not self.pool.get_tasks()
+            # workflow has hit the "stop after cycle point"
```
wxtim, line 561: Suggest that this should be documented in `globalcfg.py` since this setting now applies to stopped as well as completed workflows.

oliver-sanders: Under what config?

oliver-sanders: Done in [8e3afc7](https://github.com/cylc/cylc-flow/pull/6903/commits/8e3afc7a016ec6a9c1de4e82fb262ab7a52102c8)

PR #6909, opened 2025-08-11
Address play and vr timeouts: ping first

### LINKED ISSUE #6846: resume: replace `detect_old_contact_file` check with a "ping" type check
We currently perform the `detect_old_contact_file` check before we resume a workflow (i.e. `cylc play`) and additionally for `cylc vr`.

With all other commands, we only perform this check if a network timeout has occurred. But in these situations we just do it anyway.

I don't think there is any good reason to perform these checks, we should be able to replace it with a simple `cylc ping` type check (we are going to create a Cylc client to send the resume request anyway). If this ping experiences a timeout, the `detect_old_contact_file` check will be called, so there would be no compromise to safety.

For context, here's a thread discussing the addition of this check:

https://github.com/cylc/cylc-flow/pull/2693/files#r213234919

hjoliver:
Close #6846 
~~Partially address #6261 (by documenting exactly what `cylc vr` does now, rather than changing its behaviour)~~ [off-topic, punted to #6354]

> We currently perform the `detect_old_contact_file` check before we resume a workflow (i.e. `cylc play`) and additionally for`cylc vr`.
> With all other commands, we only perform this check if a network timeout has occurred. But in these situations we just do it anyway.
> I don't think there is any good reason to perform these checks, we should be able to replace it with a simple cylc ping type check 

The `detect_old_contact_file` check runs `cylc psutil` over ssh on a hard-wired 10 second timeout. With NFS latency issues on HPC, that may not be nearly long enough.

This change:
- do a (typically much faster) ping type check first, and only resort to `detect_old_contact_file` if that times out (#6846)
- make the timeout configurable - it's still needed for orphaned contact files, and 10 sec ain't enough
- fix badly named exception `ContactFileExists` -> `SchedulerAlive`
  - (it is not raised if the contact file exists but the scheduler is not alive)
- (somewhat relatedly) remove the ability of `cylc ping` to detect running tasks (as opposed to running workflows)
  - this very old feature was untested and broken, so I can conclude no one was using it

**Check List**

### DIFF
--- a/cylc/flow/exceptions.py
+++ b/cylc/flow/exceptions.py
@@ -136,8 +136,8 @@ class WorkflowFilesError(CylcError):
     bullet = "\n    -"
 
 
-class ContactFileExists(CylcError):
-    """Workflow contact file exists."""
+class SchedulerAlive(CylcError):
+    """Workflow contact file exists and scheduler is alive."""
 
 
 class FileRemovalError(CylcError):
--- a/cylc/flow/network/client.py
+++ b/cylc/flow/network/client.py
@@ -41,7 +41,7 @@ from cylc.flow import (
 )
 from cylc.flow.exceptions import (
     ClientTimeout,
-    ContactFileExists,
+    SchedulerAlive,
     CylcError,
     RequestError,
     WorkflowStopped,
@@ -181,7 +181,7 @@ class WorkflowRuntimeClientBase(metaclass=ABCMeta):
         # behind a contact file?
         try:
             detect_old_contact_file(self.workflow)
-        except ContactFileExists:
+        except SchedulerAlive:
             # old contact file exists and the workflow process still alive
             return
         else:
--- a/cylc/flow/network/client.py
+++ b/cylc/flow/network/client.py
@@ -158,6 +158,10 @@ class WorkflowRuntimeClientBase(metaclass=ABCMeta):
             WorkflowStopped: if the workflow has already stopped.
             CyclError: if the workflow has moved to different host/port.
         """
+        LOG.warning(
+            f"{self.workflow} {self.host}:{self.port}:"
+            f" Connection timed out ({self.timeout} ms)"
+        )
         contact_host, contact_port, *_ = get_location(self.workflow)
         if (
             contact_host != get_fqdn_by_host(self._orig_host)
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -486,7 +486,8 @@ async def _resume(workflow_id, options):
     try:
         await cylc_ping(options, workflow_id, pclient)
     except WorkflowStopped:
-        # Not running (orphaned contact file).
+        # Not running, restart instead of resume.
+        # (Orphaned contact file will be removed by cylc_ping client logic).
         return
     except CylcError as exc:
         # PID check failed - abort.
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -34,9 +34,7 @@ from cylc.flow.exceptions import (
     ServiceFileError,
     WorkflowStopped,
 )
-from cylc.flow.scripts.ping import (
-    run as cylc_ping,
-)
+from cylc.flow.scripts.ping import run as cylc_ping
 import cylc.flow.flags
 from cylc.flow.id import upgrade_legacy_ids
 from cylc.flow.host_select import select_workflow_host
--- a/cylc/flow/scheduler_cli.py
+++ b/cylc/flow/scheduler_cli.py
@@ -30,9 +30,12 @@ from packaging.version import Version
 
 from cylc.flow import LOG, __version__
 from cylc.flow.exceptions import (
-    ContactFileExists,
     CylcError,
     ServiceFileError,
+    WorkflowStopped,
+)
+from cylc.flow.scripts.ping import (
+    run as cylc_ping,
 )
 import cylc.flow.flags
 from cylc.flow.id import upgrade_legacy_ids
@@ -60,7 +63,6 @@ from cylc.flow.run_modes import WORKFLOW_RUN_MODES
 from cylc.flow.workflow_db_mgr import WorkflowDatabaseManager
 from cylc.flow.workflow_files import (
     SUITERC_DEPR_MSG,
-    detect_old_contact_file,
     get_workflow_srv_dir,
 )
 from cylc.flow.terminal import (
@@ -474,13 +476,28 @@ async def _scheduler_cli_3(
 async def _resume(workflow_id, options):
     """Resume the workflow if it is already running."""
     try:
-        detect_old_contact_file(workflow_id)
-    except ContactFileExists as exc:
-        print(f"Resuming already-running workflow\n\n{exc}")
         pclient = WorkflowRuntimeClient(
             workflow_id,
             timeout=options.comms_timeout,
         )
+    except WorkflowStopped:
+        # Not running - don't resume.
+        return
+
+    # Is it running? If yes, send resume command.
+    try:
+        await cylc_ping(options, workflow_id, pclient)
+    except WorkflowStopped:
+        # Not running (orphaned contact file).
+        return
+    except CylcError as exc:
+        # PID check failed - abort.
+        LOG.error(exc)
+        LOG.critical('Cannot tell if the workflow is running')
+        sys.exit(1)
+    else:
+        # It's running: resume it and exit.
+        print("Resuming already-running workflow")
         mutation_kwargs = {
             'request_string': RESUME_MUTATION,
             'variables': {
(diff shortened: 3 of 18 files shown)


hjoliver:
This is not really a bug fix, but it's needed pretty urgently. Can retarget to 8.6 if necessary.

cylc/flow/exceptions.py
```diff
@@ -136,8 +136,8 @@ class WorkflowFilesError(CylcError):
     bullet = "\n    -"
 
 
-class ContactFileExists(CylcError):
-    """Workflow contact file exists."""
+class SchedulerAlive(CylcError):
+    """Workflow contact file exists and scheduler is alive."""
```
oliver-sanders, line 140: I keep meaning to refactor this! Raising an exception if the scheduler is running is a tad confusing.

cylc/flow/network/client.py
```diff
@@ -158,6 +158,10 @@ def timeout_handler(self) -> None:
             WorkflowStopped: if the workflow has already stopped.
             CyclError: if the workflow has moved to different host/port.
         """
+        LOG.warning(
+            f"{self.workflow} {self.host}:{self.port}:"
+            f" Connection timed out ({self.timeout} ms)"
+        )
```
oliver-sanders, line 164: Perhaps move this warning into an `else` branch on the following `if`.

Otherwise, the same information will get written to stderr twice in certain situations.

cylc/flow/scheduler_cli.py
```diff
@@ -474,13 +474,28 @@ async def _scheduler_cli_3(
...
+    except WorkflowStopped:
+        # Not running - don't resume.
+        return
+
+    # Is it running? If yes, send resume command.
+    try:
+        await cylc_ping(options, workflow_id, pclient)
+    except WorkflowStopped:
+        # Not running (orphaned contact file).
+        return
```
oliver-sanders, line 491: With the previous logic, we would clean up the orphaned contact file at this stage. This is important for allowing crashed workflows to be restarted.

I think we need to bring back the `detect_old_contact_file` check, but only for this code branch.

```python
        detect_old_contact_file(workflow_id)
        # TODO: handle the different outcomes
```

cylc/flow/scheduler_cli.py
```diff
@@ -474,13 +474,28 @@ async def _scheduler_cli_3(
...
+    except CylcError as exc:
```
oliver-sanders, line 492: Note that `ClientTimeout` is a subclas of `CylcError`.

Not sure whether you want special handling for that?

cylc/flow/scripts/clean.py
```diff
@@ -220,7 +220,7 @@ async def run(*ids: str, opts: 'Values') -> None:
         workflows, multi_mode = await scan(workflows, multi_mode)
 
         if not workflows:
-            LOG.warning(f"No workflows matching {', '.join(ids)}")
+            LOG.warning(f"No stopped workflows matching {', '.join(ids)}")
```
oliver-sanders, line 223: :+1:

cylc/flow/scripts/ping.py
```diff
@@ -48,16 +46,6 @@
...
-  }
-}
-'''
-
-TASK_QUERY = '''
-query ($tProxy: ID!) {
-  taskProxy (id: $tProxy) {
-    state
-    id
   }
```
oliver-sanders, line 49: Fair enough, I'm not sure what the purpose of this was?

cylc/flow/exceptions.py
hjoliver: Agreed, it was a bit of a head-scratcher.

cylc/flow/scripts/ping.py
hjoliver: See Issue description, final bullet point.

cylc/flow/scheduler_cli.py
hjoliver: No, `cylc_ping` does that already, via the usual client logic, and then the scheduler restarts instead of resumes.

I've tweaked the comment to clarify.

```
$ cylc vip test
...
test/run1: NIWA-1022450.niwa.local PID=20799

$ kill -9 20799  # => orphaned contact file

$ cylc play test
WARNING - test/run1 NIWA-1022450.niwa.local:43041: Connection timed out (5000 ms)
INFO - Removed contact file for test/run1 (workflow no longer running).
INFO - HELLO FROM scheduler_cli line 489
...
test/run1: NIWA-1022450.niwa.local PID=21056
```

cylc/flow/scheduler_cli.py
hjoliver: No need, client timeout is handled in the `cylc_ping` client code.

cylc/flow/network/client.py
hjoliver: It's not the same information: (1) the connection timed out, then (2) reason for the timeout. IMO that's good because it explains why the command hung for a while. 

In principle we could combine the two messages (`timed out: reason`) but that would require a bit of refactoring as other "reasons" are printed down in `detect_old_contact_files` (i.e., not inside this method).

cylc/flow/scripts/ping.py
oliver-sanders: I did, I'm still not sure what the purpose of this feature was.

cylc/flow/scripts/validate_reinstall.py
```diff
@@ -18,12 +18,14 @@
...
 
-Validate and reinstall a workflow then either:
+If the source has changed and you respond Yes to the reinstall prompt:
+* reinstall and reload the workflow, if it is running
+* or reinstall and restart the workflow, if it is not running
 
-* "Reload" the workflow (if it is running),
-* or "play" it (if it is stopped).
+If the source has not changed or you respond No to the reinstall prompt, the
+workflow will not be reloaded or restarted.
```
oliver-sanders: This isn't quite right:
* It infers that the workflow will still be reinstalled, even if there were no changes.
* It fails to mention that the `--yes` option overrides this behaviour.
* The command will still function without prompts in non-interactive mode.

The original description was a clearer overview of the command's behaviour. Suggest keeping the simple description and extending it to cover edge cases.

Suggestion:

```suggestion
Validate and reinstall a workflow then either:

* "Reload" the workflow (if it is running),
* or "play" it (if it is stopped).

This command will only reinstall, then reload/restart the workflow if the
source has changed since it was installed. Use the `--yes` option to reinstall
it anyway.
```

cylc/flow/scripts/ping.py
hjoliver: Ah OK, it was a crude and ancient feature for interrogating workflow state (running tasks only) that predated commands for dumping the task pool or interrogating the DB (or even the text file that preceded the DB).

cylc/flow/scripts/validate_reinstall.py
hjoliver: It is difficult to document these multi-faceted commands accurately without being overly verbose 😓 

> It infers that the workflow will still be reinstalled, even if there were no changes.

I presume you mean "implies".  If you read my description as a whole (as opposed to the second sentence in isolation) it is pretty clear, but OK I can make it even clearer by explicitly saying no reinstallation is done in the second case.

> It fails to mention that the --yes option overrides this behaviour.

Damn, it I forgot about the behaviour of `--yes` that contradicts the option documentation: https://github.com/cylc/cylc-flow/issues/6261#issuecomment-2261815358
Maybe I can amend this PR to deal with that properly as well, as discussed on the issue.

-----

Your suggested change isn't that clear BTW - first it implies that reinstallation is always done:
> Validate and reinstall a workflow then either:

... and then later contradicts itself for certain cases:
> This command will only reinstall ... if ...

hjoliver: > The original description was a clearer overview of the command's behaviour.

Here's the original text:

> Validate, reinstall and apply changes to a workflow.
>
> Validate and reinstall a workflow then either:
> 
> * "Reload" the workflow (if it is running),
> * or "play" it (if it is stopped).

Here's why it's not clear (point 3 is in fact why I posted #6261 in the first place):

- there are at least 3 possible interpretations of "apply changes to a workflow" 
- it implies that reinstallation is always done, which is not true (even if there are source changes)
- it implies the workflow will always be restarted if stopped, which is not true
- it doesn't mention the interactive prompt
- it doesn't mention that `--yes` does more than just "skip interactive prompts" as documented
- (minor, style: the second line is nearly redundant with the first)

hjoliver: (I'll have another go at getting it right, but have run out of time today).

hjoliver: @oliver-sanders - I've reverted the CLI help to the original, to resolve this discussion - will address with #6354

hjoliver:
@oliver-sanders - any suggestions on how to test this?

dwsutherland (approved):
LGTM 👍  (also used in ESNZ operations)

PR #6910, opened 2025-08-11
Group trigger: re-use already-completed outputs of active group start tasks.

### LINKED ISSUE #6858: Group trigger follow-up for "live" start tasks with completed outputs
Placeholder: see
- https://github.com/cylc/cylc-flow/pull/6395#issuecomment-3082054766
- https://github.com/cylc/cylc-flow/pull/6395#issuecomment-3083488141
- https://github.com/cylc/cylc-flow/pull/6395#issuecomment-3086208463

hjoliver:
Close #6858 

Re-spawn **in-group** children of already-completed group start task outputs - because triggering is event-driven and group trigger removes in-group tasks and their already-completed prerequisites.

**Check List**

### DIFF
--- a/cylc/flow/scripts/trigger.py
+++ b/cylc/flow/scripts/trigger.py
@@ -17,39 +17,23 @@
 
 """cylc trigger [OPTIONS] ARGS
 
-For tasks to trigger, while respecting dependencies among them.
+Trigger a group of one or more tasks, respecting dependencies among them.
 
-Triggered tasks can run even if the workflow is paused.
+Prerequisites on tasks outside of the group will be satisfied automatically.
 
-Live tasks (preparing, submitted, running) can't be triggered until finished.
+Tasks will be removed if necessary to allow re-run without intervention, so
+triggered tasks that are preparing, submitted, or running may be killed.
 
-Triggering an unqueued task queues it, and triggering a queued task runs it
-regardless of the queue limit; so it may take two triggers to run a task now.
+Tasks that lead into a group will run immediately even if the workflow is
+paused; activity will flow on from them once the workflow is resumed.
 
-Triggering a group of tasks: prerequisites on tasks outside of the group will
-be satisfied automatically; tasks in the group that already ran will be removed
-to allow re-run without intervention, and live tasks will be removed if
-necessary to allow the group to re-run in order.
+Triggering an unqueued task queues it; triggering a queued task runs it.
 
-If the workflow is paused, group start tasks will trigger immediately and the
-flow will continue downstream of them once the workflow is resumed.
-
-  Cylc will automatically:
-  * Erase the run history of members, so they can re-run in the same flow.
-  * Identify group start tasks, and trigger them to start the flow.
-  * Identify off-group dependencies, and satisfy them to avoid a stall.
-  * Leave in-group dependencies to be satisfied by the triggered flow.
-
-  How live (preparing, submitted, or running) tasks are handled:
-  * Live in-group tasks are killed and removed, to rerun in the triggered flow.
-  * Live group-start tasks are left alone; they don't need to be retriggered.
-
-Flow number assignment in triggered tasks:
-  Active tasks (n=0) already have flows assigned; inactive tasks (n>0) do not.
+How flow numbers are assigned to triggered tasks:
+  Active tasks (n=0) already have assigned flows; inactive tasks (n>0) do not.
   * If an interdependent group of triggered tasks includes active tasks, the
     flow will be assigned the existing flow numbers of those active tasks.
   * Otherwise the flow will be assigned all current active flow numbers.
-
 
 Examples:
   # trigger task foo in cycle 1, in workflow "test"
@@ -64,16 +48,6 @@ Examples:
 
   # rerun sub-graph "a => b & c" in the same flow, ignoring "off => b"
   $ cylc trigger test //1/a //1/b //1/c
-
- Flow numbers of triggered tasks are determined as follows:
-  Active tasks (n=0) already have existing flow numbers.
-   * default: merge active and existing flow numbers
-   * --flow=INT or "new": merge given and existing flow numbers
-   * --flow="none": ERROR (not valid for already-active tasks)
-  Inactive tasks (n>0) do not have flow numbers assigned:
-   * default: run with all active flow numbers
-   * --flow=INT or "new": run with the given flow numbers
-   * --flow="none": run as no-flow (activity will not flow on downstream)
 
 """
 
--- a/cylc/flow/scripts/trigger.py
+++ b/cylc/flow/scripts/trigger.py
@@ -35,13 +35,9 @@ Triggering a group of tasks at once (e.g. members of a sub-graph):
   If the workflow is paused, group start tasks will trigger immediately; the
   flow will continue downstream of them when you resume the workflow.
 
-  Beware of triggering live (preparing, submitted, or running) tasks:
-  * Live in-group tasks will be killed and their run history erased, to allow
-    them to re-run in the triggered flow.
-  * Live group-start tasks are left to run; they don't need to be retriggered.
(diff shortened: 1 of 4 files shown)


hjoliver:
Simple example:
```cylc
[scheduling]
    [[graph]]
    R1 = """
         a:x => b => c
         a:x => offgroup
      """
[runtime]
  [[root]]
      pre-script = sleep 5
  [[b, c, offgroup]]
  [[a]]
    script = """
        # Complete output x and wait for downstream tasks to finish
        cylc message "of cabbages and kings"
        cylc__job__poll_grep_workflow_log -E "1/offgroup.*=> succeeded"
        cylc__job__poll_grep_workflow_log -E "1/c.*=> succeeded"

        # Now trigger a, b, c as group.
        # 1. I (a) should be left alone as running
        # 2. b and c should be removed, to run again in the triggered flow
        # 3. my already-completed output x should satisfy b for the triggered flow
        #    but NOT the offgroup task
        # ... so a:x => b => c should complete, and workflow shuts down.

        # First remove offgroup to prove we got 3. right.
        cylc remove $CYLC_WORKFLOW_ID//1/offgroup
        cylc__job__poll_grep_workflow_log -E "Removed.*1/off"

        # Now trigger the group.
        cylc trigger $CYLC_WORKFLOW_ID //1/a //1/b //1/c
    """
    [[[outputs]]]
        x = "of cabbages and kings"
```

wxtim:
@hjoliver - my test case is basically the same as your example, except that I test both

```cylc
a:started => b => end
k:kustom => l => end
```

hjoliver:
@wxtim - my manual test case also checks that the "re-emitted" outputs don't also cause off-group activity.  I'll merge your test (thanks) and extend it a bit to cover that.

hjoliver:
(All tests passed)

Rebased to 8.5.x and labelled as a bug fix.

@oliver-sanders - this was easier than anticipated because rather than respawning on the already-completed outputs we can just set the corresponding prerequisites on in-group downstream tasks (thus naturally avoiding off-group effects).

cylc/flow/scripts/trigger.py
```diff
@@ -35,13 +35,9 @@
...
-  Beware of triggering live (preparing, submitted, or running) tasks:
-  * Live in-group tasks will be killed and their run history erased, to allow
-    them to re-run in the triggered flow.
-  * Live group-start tasks are left to run; they don't need to be retriggered.
-    WARNING: if they already completed outputs that other group members depend
-    on, you must manually satisfy those prerequisites again for the triggered
-    flow (run history erasure wipes out the original satisfied prerequisites).
+  How live (preparing, submitted, or running) tasks are handled:
+  * Live in-group tasks are killed and removed, to rerun in the triggered flow.
+  * Live group-start tasks are left alone; they don't need to be retriggered.
```
oliver-sanders: Technical and correct, but I don't think users are going to make much sense of this.

I would recommend explaining group trigger in terms of its behaviours rather than the details of its implementation.

Here's how I explained it in the changes section:

> * Prerequisites on any tasks that are outside of the group of tasks being triggered are automatically satisfied.
> * Any tasks which have already run within the group will be automatically removed (i.e. cylc remove) to allow them to be re-run without intervention.
> * Any preparing, submitted or running tasks within the group will also be removed if necessary to allow the tasks to re-run in order.

This avoids using the terms "gruop-start", "in-group", "flow" and "run history" but is much more likely to be read and understood.

oliver-sanders (approved):
Code makes sense, tested as working, :+1:.

cylc/flow/scripts/trigger.py
hjoliver: I'll put a general note on approaches to documentation on Element, but for now:

I don't really agree that terms like "in-group" are technical in this context: i.e., in a section entitled "triggering a group of tasks at once", and where "group" here is perfectly compatible with normal colloquial usage of the word.

And we *should* be using "flow" to describe (intuitively, not technically) workflow activity flowing through the graph. Without a good name for concepts like this we end up having to repeat verbose descriptions all over the place.

That said, in this case maybe I have overcooked it, and I do quite like your description, so I'm considering a re-do ...

oliver-sanders: One to take off issue.

<br />

> I don't really agree that terms like "in-group" are technical [...] in a section entitled "triggering a group of tasks at once"

I agree! These aren't especially technical terms so long as they are contextualised in a section as you say.

But if we *can* explain the behaviours more simply (without using these terms), we should.

<br />

> And we should be using "flow" to describe (intuitively, not technically) workflow activity flowing through the graph.

I understand why you want to use the term "flow" and convey the flow model in documentation.

However, the "flow" term is only really necessary to explain concurrent flows. I don't think it is helpful outside of this context, we can explain single-flow behaviour more clearly without this term/model (just like we did at Cylc 7).

Concurrent flows are an advanced feature, the terminology and models associated with it should be constrained to the documentation for concurrent flows (we asked for this at proposal time). The default single-flow cases can and should be more simply explained without invoking this model.

hjoliver: > I understand why you want to use the term "flow" and convey the flow model in documentation. ...

That's really not what I'm doing. I certainly agree that our recent intervention improvements have relegated concurrent flows to an advanced feature that won't need to be used much. By "flow" I simply mean when I trigger a task (or set its prerequisites or whatever) activity "flows on from" that intervention through the graph. That's very much an intuitive concept, and IMO it's the most natural way to explain what Cylc does.  Even with the new group trigger feature, which erases run history so that a "concurrent flow" is not needed, the activity that results from the trigger still "flows" through the graph in this way and as such is most simply described as triggering a flow.

wxtim (approved):

cylc/flow/scripts/trigger.py
oliver-sanders: > By "flow" I simply mean when I trigger a task (or set its prerequisites or whatever) activity "flows on from" that intervention through the graph

I understand this, and I do get where you are coming from. Of course "flow on" is reasonable, but there shouldn't be any need to talk about "flows" outside of concurrent flows IMO.

Outside of concurrent flows, I don't think the "flow" model / terminology is particularly helpful and can make docs confusing for users. We can explain Cylc things clearly without introducing this term (e.g. we don't use it once in the tutorial).

hjoliver:
> But if we can explain the behaviours more simply (without using these terms), we should.

Most concepts can be described without domain-specific terminology, but there's a trade-off with verbosity. A good (natural or intuitive) name for a concept can avoid very verbose descriptions, which are particularly painful when the concepts come up frequently.

I much prefer concise writing, and I think users often won't bother reading long tracts of text.  "Group" and "flow" (as described above) strike me as extremely natural and obvious terms for concepts that take many more words to explain without them.
- "in-group" (or "off-group") vs "a task that belongs (or does not belong) to the set of tasks that you triggered at once"
- "flow" vs "the activity that flows on through the graph downstream of your manual intervention"

Unfortunately where we land on the trade-off is a subjective judgment. 

Perhaps a good compromise, in each major piece of documentation (a self-contained CLI help page, or a section of the user guide, say), is to use the verbose description once at the top, to define the term in-place (as opposed to assuming that the user knows it already) and then use the concise term below that - in cases where the concept needs to be referred to multiple times.

cylc/flow/scripts/trigger.py
```diff
@@ -17,34 +17,20 @@
...
+Tasks will be removed if necessary to allow re-run without intervention, so
+triggered tasks that are preparing, submitted, or running may be killed.
+
+Tasks that lead into a group will run immediately even if the workflow is
+paused; activity will flow on from them once the workflow is resumed.
+
+Triggering an unqueued task queues it; triggering a queued task runs it.
+
+How flow numbers are assigned to triggered tasks:
+  Active tasks (n=0) already have assigned flows; inactive tasks (n>0) do not.
```
oliver-sanders, line 33: Much simpler!

Issue #6598, opened 2025-02-07
cat-log: new auto cat/tail mode


oliver-sanders:
* `cylc cat-log <job-id>` will cat/tail the requested log file for the specified job.
  * `--mode=cat` (CLI default) will cat the file.
  * `--mode=tail` (UIServer default) will tail the file.

Issues:
1. If the job has finished, the job log files are no longer growing and the tail operation is not required.
2. However, the UI Server will always request `--mode=tail` because it doesn't know any better.
3. Excess tail commands may cause load.

Solution:
* Add a new `--mode` called "auto" (or something of that ilk) that will tail if the job is running, and cat otherwise.

oliver-sanders:
(optimistically tagged 8.6.0, but this should not delay release)

PR #2966, opened 2019-02-22
Python3

oliver-sanders:
#### Python 2.6 -> 3.7

* For the most part a minimal effort to get Cylc running under Py3
* Dependency base is now quite pleasing :champagne:.
* All but two remote tests should pass :champagne: 
  * *batch schedulers not tested*
  * **bumping this to future work**
  * `tests/authentication/10-remote-suite-passphrase-cache.t` - awaiting conclusion of #2975 
  * `tests/cylc-message/00-ssh.t` - awaiting conclusion of #2975 
* All local tests should pass :champagne:.
* All unittests should pass :champagne:.

#### Related Issues

closes #2858 
closes #2647 
closes #2858 
closes #2864 
??? #2785 

GUI issues which become obsolete:
closes #164 
closes #1620 
closes #1671
closes #1996
closes #2466 
closes #2498 
closes #2632 
closes #2898 
closes #2633 
closes #2956 

GUI issues which we may want to keep around for the moment:
addresses #2570 
addresses #2580 
addresses #2284 

#### TODO:

* [x] Scan suites in parallel
* [x] Fix remaining test failures
* [x] Docs
* [x] pep8 (will do as last commit to make reviewing easier)
* [x] Work out what we should do with uuid (removed from client-server interface at the moment).

#### Highlights:

* Remove non-py3 compatible components
  * GTK related code
  * `cylc review` (awaiting tornado re-write)
  * cherrypy
* Upgrade isodatetime
* Upgrade to Python 3.7 (unittests are now run under 3.6 as well, functional tests should pass under 3.6, untested)
* Replace cherrypy with minimal zmq network layer
  * The aim here was to put in the absolute minimum, we will want to improve on this later
  * Anon auth has been removed awaiting a superior mechanism, tests have been disabled
* Replace `cylc graph --reference`
* Refactor `cylc scan`
  * Aggressive refactor with interface changes, I can "tone it down" a little if we think this is likely to cause problems
  * See commit notes for details

#### Further Work

* ZMQ interface improvements: #2978
* CLI improvements: #2972 
* Convert main loop to asyncio #2866 
* Cylc unicode support #2979, #2281 
* Fully replace `__cmp__` logic

**Take a look through the commits, there have been a few changes beyond Python3.**

PR #6913, opened 2025-08-11
#6598 cat-log: new auto cat/tail mode

ChrisPaulBennett:
This PR addresses issue #6598.
This adds a new 'auto' mode to the cylc cat-log command. When asking for job logs (Not workflow logs), the logic will determine if the workflow is live, tail will be used otherwise cat will be used.

**Check List**

### DIFF
--- a/cylc/flow/scripts/cat_log.py
+++ b/cylc/flow/scripts/cat_log.py
@@ -158,6 +158,7 @@ MODES = {
     'd': 'print-dir',
     'c': 'cat',
     't': 'tail',
+    'a': 'auto',
 }
 
 
@@ -367,7 +368,7 @@ def get_task_job_attrs(workflow_id, point, task, submit_num):
     """Retrieve job info from the database.
 
     * live_job_id is the job ID if job is running, else None.
-    * submit_failed is True if the the submission failed.
+    * submit_failed is True if the submission failed.
 
     Returns:
         tuple - (platform, job_runner_name, live_job_id, submit_failed)
@@ -466,6 +467,10 @@ def _main(
         file_name: str = options.filename or 's'
         log_file_path: Path
 
+        # auto mode only applies to task logs. Default to cat mode.
+        if mode == 'auto':
+            mode = 'cat'
+
         if mode == 'list-dir':
             # list workflow logs
             print('\n'.join(sorted(
@@ -540,6 +545,10 @@ def _main(
                 # KeyError: Is already long form (standard log, or custom).
         platform_name, _, live_job_id, submit_failed = get_task_job_attrs(
             workflow_id, point, task, submit_num)
+        if mode == 'auto' and live_job_id is None:
+            mode = 'cat'
+        elif mode == 'auto' and live_job_id:
+            mode = 'tail'
         platform = get_platform(platform_name)
         batchview_cmd = None
         if live_job_id is not None:
--- a/tests/functional/cylc-cat-log/14-auto-mode.t
+++ b/tests/functional/cylc-cat-log/14-auto-mode.t
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#-------------------------------------------------------------------------------
+# Test "cylc cat-log" using auto mode with a custom tail /command
+. "$(dirname "$0")/test_header"
+#-------------------------------------------------------------------------------
+set_test_number 4
+install_workflow "${TEST_NAME_BASE}" "${TEST_NAME_BASE}"
+create_test_global_config "" "
+[platforms]
+   [[localhost]]
+        tail command template = $PWD/bin/my-tailer.sh %(filename)s
+"
+#-------------------------------------------------------------------------------
+TEST_NAME="${TEST_NAME_BASE}-validate"
+run_ok "${TEST_NAME}" cylc validate "${WORKFLOW_NAME}"
+workflow_run_ok "${TEST_NAME_BASE}-run" cylc play "${WORKFLOW_NAME}"
+cylc workflow-state "${WORKFLOW_NAME}//1/foo:started" --interval=1
+sleep 3
+# Test that the custom tail command is used when the task is live
+TEST_NAME=${TEST_NAME_BASE}-cat-log
+cylc cat-log "${WORKFLOW_NAME}//1/foo" -f o -m a > "${TEST_NAME}.out"
+grep_ok "HELLO from foo 1" "${TEST_NAME}.out"
+#-------------------------------------------------------------------------------
+cylc stop --kill --max-polls=20 --interval=1 "${WORKFLOW_NAME}"
+#-------------------------------------------------------------------------------
+# Test that the custom tail command is not used when the task is not live
+# (i.e., auto mode should not use the custom tail command)
+cylc cat-log "${WORKFLOW_NAME}//1/foo" -f o -m a > "${TEST_NAME}.out"
+grep_fail "HELLO from foo 1" "${TEST_NAME}.out"
+purge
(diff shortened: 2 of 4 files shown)


tests/functional/cylc-cat-log/14-auto-mode.t
```diff
@@ -0,0 +1,45 @@
...
+[platforms]
+   [[localhost]]
+        tail command template = $PWD/bin/my-tailer.sh %(filename)s
+"
+#-------------------------------------------------------------------------------
+TEST_NAME="${TEST_NAME_BASE}-validate"
+run_ok "${TEST_NAME}" cylc validate "${WORKFLOW_NAME}"
+workflow_run_ok "${TEST_NAME_BASE}-run" cylc play "${WORKFLOW_NAME}"
+cylc workflow-state "${WORKFLOW_NAME}//1/foo:started" --interval=1
+sleep 3
```
oliver-sanders: The `cylc workflow-state` command should render this unncessary:

```suggestion
```

ChrisPaulBennett: With sleep 1 (Which is what the 04-local-tail test does) the test does not work, neither does sleep 2, sleep 3 does work though.
With sleep 1 or sleep 2, it sees the job as not live and uses cat instead of tail.

cylc/flow/scripts/cat_log.py
```diff
@@ -466,6 +467,10 @@ def _main(
         file_name: str = options.filename or 's'
         log_file_path: Path
 
+        # auto mode only applies to task logs. Default to cat mode.
+        if mode == 'auto':
```
wxtim: It's a bit more refactoring, but in my opinion these strings should be assigned a variable at the top, just before modes:

```python
PRINT = 'print'
LISTDIR = 'list-dir'
PRINTDIR = 'print-dir'
CAT = 'cat'
TAIL = 'tail'
AUTO = 'auto'

MODES = {
    'p': PRINT,
    'l': LISTDIR,
    'd': PRINTDIR,
    'c': CAT,
    't': TAIL,
    'a': AUTO,
}
```
Then using these variables. Means that we don't have to search through the code if we ever need to change these values.

(In theory there's probably some clever use of `enums`, `dataclasses`, `namedtuple` or something which would be really elegant here, but this is fine for the use case)

cylc/flow/scripts/cat_log.py
```diff
@@ -158,6 +158,7 @@
     'd': 'print-dir',
     'c': 'cat',
     't': 'tail',
+    'a': 'auto',
```
wxtim: Please add this list to the help text.

cylc/flow/scripts/cat_log.py
ChrisPaulBennett: Oliver said to me that Ronnie would find the use of hard coded strings infuriating and ask me to change it. Looks like you beat him to it!
Done

tests/functional/cylc-cat-log/14-auto-mode.t
wxtim: Perhaps you need to test for `:finished`?

cylc/flow/scripts/cat_log.py
ChrisPaulBennett: `parser.add_option(`
 `    "-m", "--mode",`
 `    help="Mode: %s. Default c(cat)." % (`
 `    ', '.join(['%s(%s)' % (i, j) for i, j in MODES.items()])),`
 `    action="store", choices=list(MODES.keys()) + list(MODES.values()),`
 `    default='c', dest="mode")`

This code auto makes the help text for the arg parser.
The help text reads;
`-m MODE, --mode=MODE  Mode: p(print), l(list-dir), d(print-dir), c(cat),`
`    t(tail), a(auto). Default c(cat).`

Am I missing some help text?

tests/functional/cylc-cat-log/14-auto-mode.t
wxtim: I get message `InputError: No such task state "started"` - probably why this isn't working?

I see that this sleep is copied from similar tests (added in https://github.com/cylc/cylc-flow/pull/2966/files) - and that workflow-state doesn't actually appear to be working correctly. You could try and solve, or decide it's OOS and create a new ticket. 😞

oliver-sanders: If you need to use a sleep, then it means something in the test is timing dependent. We should replace this `sleep` with a command which waits for whatever event you are awaiting to happen. 

Sleeps are fragile and result in flaky tests which may pass locally, but fail on a loaded box running tests in parallel.

ChrisPaulBennett: Agreed, but suggest that someone with a bit more experience with these functional tests attempts to solve the issue.
So, suggest this is OOS for little old me.

cylc/flow/scripts/cat_log.py
wxtim: No, I'm a muppet.

cylc/flow/scripts/cat_log.py
oliver-sanders: Knew it!

cylc/flow/scripts/cat_log.py
oliver-sanders: > No, I'm a muppet.

So tempted to AI this true!

tests/functional/cylc-cat-log/14-auto-mode.t
```diff
@@ -0,0 +1,55 @@
...
```
wxtim: This should fix this:

```suggestion
cylc workflow-state "${WORKFLOW_NAME}//1/foo:started" --interval=1 --messages
```

tests/functional/cylc-cat-log/14-auto-mode.t
```diff
@@ -0,0 +1,55 @@
...
+grep_ok "HELLO from foo 1" "${TEST_NAME}.out"
+#-------------------------------------------------------------------------------
+cylc stop --kill --max-polls=20 --interval=1 "${WORKFLOW_NAME}"
+#-------------------------------------------------------------------------------
+# Test that the custom tail command is not used when the task is not live
+# (i.e., auto mode should not use the custom tail command)
+TEST_NAME=${TEST_NAME_BASE}-cat-log-auto-cat
+cylc cat-log "${WORKFLOW_NAME}//1/foo" -f o -m a > "${TEST_NAME}.out"
+grep_fail "HELLO from foo 1" "${TEST_NAME}.out"
+grep_ok "from foo 1" "${TEST_NAME}.out"
```
oliver-sanders: The `run_ok` assertion thinggy redirects its output into stdout/err files for you:

```suggestion
run_ok "$TEST_NAME" cylc cat-log "${WORKFLOW_NAME}//1/foo" -f o -m a
grep_fail "HELLO from foo 1" "${TEST_NAME}.stdout"
grep_ok "from foo 1" "${TEST_NAME}.stdout"
```

You will need to bump the test number at the top of the file when doing this.

tests/functional/cylc-cat-log/14-auto-mode.t
```diff
@@ -0,0 +1,55 @@
...
+"
+#-------------------------------------------------------------------------------
+TEST_NAME="${TEST_NAME_BASE}-validate"
+run_ok "${TEST_NAME}" cylc validate "${WORKFLOW_NAME}"
+workflow_run_ok "${TEST_NAME_BASE}-run" cylc play "${WORKFLOW_NAME}"
+cylc workflow-state "${WORKFLOW_NAME}//1/foo:started" --interval=1
+sleep 3
+# Test that the custom tail command is used when the task is live
+TEST_NAME=${TEST_NAME_BASE}-cat-log-auto-tail
+cylc cat-log "${WORKFLOW_NAME}//1/foo" -f o -m a > "${TEST_NAME}.out"
```
oliver-sanders: Ditto (use `run_ok`).

tests/functional/cylc-cat-log/14-auto-mode.t
```diff
@@ -0,0 +1,55 @@
...
+TEST_NAME=${TEST_NAME_BASE}-cat-log-auto-cat
+cylc cat-log "${WORKFLOW_NAME}//1/foo" -f o -m a > "${TEST_NAME}.out"
+grep_fail "HELLO from foo 1" "${TEST_NAME}.out"
+grep_ok "from foo 1" "${TEST_NAME}.out"
+#-------------------------------------------------------------------------------
+# Test that the custom tail command is not used when workflow logs are requested
+# (i.e., auto mode should not use the custom tail command)
+TEST_NAME=${TEST_NAME_BASE}-cat-log-auto-workflow
+# Use the workflow name without the task ID to get the workflow logs
+cylc cat-log "${WORKFLOW_NAME}" -m a > "${TEST_NAME}.out"
```
oliver-sanders: Ditto (use run_ok).

oliver-sanders (commented):
LGTM

tests/functional/cylc-cat-log/14-auto-mode.t
ChrisPaulBennett: Done

tests/functional/cylc-cat-log/14-auto-mode.t
ChrisPaulBennett: Done

tests/functional/cylc-cat-log/14-auto-mode.t
ChrisPaulBennett: Done

tests/functional/cylc-cat-log/14-auto-mode.t
ChrisPaulBennett: Thanks for that

PR #6914, opened 2025-08-12
Fix data store bug where late arrival of submitted message would cause backwards state change

### LINKED ISSUE #6686: Delayed poll result contaminate the datastore.

MetRonnie:
Fixes #6686

### Bug

Sometimes the `submitted` message can come through after `started`, and when the `submitted` message came through this was causing the job in the data store to regress to `submitted`.

```console
$ grep '1/foo' ~/cylc-run/wflow/log/scheduler/log 
INFO - [1/foo:waiting(runahead)] => waiting
INFO - [1/foo:waiting] => waiting(queued)
INFO - [1/foo:waiting(queued)] => waiting
INFO - [1/foo/01:preparing] setting implied output: submitted
INFO - [1/foo/01:preparing] submitted to hpc:pbs[None]
INFO - [1/foo/01:preparing] => submitted
INFO - [1/foo/01:submitted] => running
INFO - [1/foo/01:running] => succeeded
INFO - [1/foo/01:succeeded] submitted to hpc:pbs[5920620]
```

In the GUI this would result in the task correctly showing as succeeded but the job incorrectly showing as submitted.

### Fix

In the data store, skip any job state deltas that would take it backwards. There shouldn't be any legitimate reason for a job to go backwards in state (the sole exception being the loadleveler "vacated" signal, which I have hopefully handled although this is not likely to be used much).

> [!NOTE]
> See also #6842 which ensures `cylc set` does not affect job states

### Check List

- [x] No dependency changes 
- [x] Tests are included
- [x] Docs PR not needed

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -2771,12 +2771,12 @@ class DataStoreMgr:
 
     def delta_job_attr(
         self,
-        tokens: Tokens,
+        itask: 'TaskProxy',
         attr_key: str,
         attr_val: Any,
     ) -> None:
         """Set job attribute."""
-        j_id, job = self.store_node_fetcher(tokens)
+        j_id, job = self.store_node_fetcher(itask.job_tokens)
         if not job:
             return
         j_delta = PbJob(stamp=f'{j_id}@{time()}')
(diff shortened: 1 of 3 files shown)


hjoliver:
> (the sole exception being the loadleveler "vacated" signal, which I have hopefully handled although this is not likely to be used much).

I'd be very surprised if it's used at all.

PR #6919, opened 2025-08-13
Fix workflow trigger commands broken in functional tests

wxtim:
Tests relying on :started and :stopping needed --messages adding to make them work.
Tests relying on :running triggers will not have been working as workflow-trigger will return an error in this case.
Don't swallow errors!

**Check List**

### DIFF
--- a/tests/functional/restart/38-auto-restart-stopping.t
+++ b/tests/functional/restart/38-auto-restart-stopping.t
@@ -17,7 +17,7 @@
 #-------------------------------------------------------------------------------
 export REQUIRE_PLATFORM='loc:remote fs:shared runner:background'
 . "$(dirname "$0")/test_header"
-set_test_number 2
+set_test_number 3
 if ${CYLC_TEST_DEBUG:-false}; then ERR=2; else ERR=1; fi
 #-------------------------------------------------------------------------------
 # ensure that workflows don't get auto stop-restarted if they are already stopping
@@ -52,7 +52,8 @@ ${BASE_GLOBAL_CONFIG}
 "
 
 run_ok "${TEST_NAME}-workflow-start" cylc play "${WORKFLOW_NAME}" --host=localhost
-cylc workflow-state "${WORKFLOW_NAME}//1/foo:running" --interval=1 --max-polls=20 >& $ERR
+run_ok "wait-for-task-foo-to-start" \
+    cylc workflow-state "${WORKFLOW_NAME}//1/foo:started" --triggers --interval=1 --max-polls=20 >& $ERR
 
 # condemn localhost
 create_test_global_config '' "
--- a/tests/functional/cylc-cat-log/04-local-tail.t
+++ b/tests/functional/cylc-cat-log/04-local-tail.t
@@ -18,7 +18,7 @@
 # Test "cylc cat-log" with a custom tail /command
 . "$(dirname "$0")/test_header"
 #-------------------------------------------------------------------------------
-set_test_number 3
+set_test_number 4
 install_workflow "${TEST_NAME_BASE}" "${TEST_NAME_BASE}"
 create_test_global_config "" "
 [platforms]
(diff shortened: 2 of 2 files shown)


hjoliver:
I don't think we were deliberately swallowing errors here. There was probably an incorrect expectation that the test would abort if `cylc workflow-state` failed (whether because the desired state was not achieved, or a proper error condition like this).

All functional test scripts should probably have `set -eu` at the top, to abort if any command fails.

hjoliver:
> (These probably became broken with #5809, @hjoliver is in the best place to advise)

And also by #6175 which disallowed polling for transient states like `running`.

tests/functional/restart/38-auto-restart-stopping.t
```diff
@@ -52,7 +52,7 @@ ${BASE_GLOBAL_CONFIG}
 "
 
 run_ok "${TEST_NAME}-workflow-start" cylc play "${WORKFLOW_NAME}" --host=localhost
-cylc workflow-state "${WORKFLOW_NAME}//1/foo:running" --interval=1 --max-polls=20 >& $ERR
+cylc workflow-state "${WORKFLOW_NAME}//1/foo:started" --messages --interval=1 --max-polls=20
```
hjoliver: The redirection to `$ERR` was deliberately added by @oliver-sanders by the look of it - probably there to capture or hide command output depending on debug mode. Does it need to be removed?

tests/functional/restart/41-auto-restart-local-jobs.t
```diff
@@ -63,7 +63,7 @@ cylc play "${WORKFLOW_NAME}"
 TEST_NAME="${TEST_NAME_BASE}-normal-mode"
 
 cylc workflow-state "${WORKFLOW_NAME}//1/foo:started" \
-  --triggers --interval=1 --max-polls=20 >& $ERR
+  --messages --triggers --interval=1 --max-polls=20
```
hjoliver: This one doesn't need changing. The `--triggers` option is fine for `started` (as is `--messages`, but don't need both).

tests/functional/restart/41-auto-restart-local-jobs.t
```diff
@@ -95,7 +95,7 @@ TEST_NAME="${TEST_NAME_BASE}-force-mode"
 
 cylc trigger "${WORKFLOW_NAME}//1/bar"
 cylc workflow-state "${WORKFLOW_NAME}//1/bar:started" \
-  --triggers --interval=1 --max-polls=20 >& $ERR
+  --messages --triggers --interval=1 --max-polls=20
```
hjoliver: ditto

tests/functional/shutdown/08-now1/flow.cylc
```diff
@@ -24,7 +24,7 @@
         [[[events]]]
             # wait for the stopping message, sleep a bit, then echo some stuff
             started handlers = """
-                cylc workflow-state %(workflow)s//%(point)s/%(name)s:stopping --triggers >/dev/null && sleep 1 && echo 'Hello %(id)s %(event)s'
+                cylc workflow-state %(workflow)s//%(point)s/%(name)s:stopping --messages --triggers >/dev/null && sleep 1 && echo 'Hello %(id)s %(event)s'
```
hjoliver: ditto

tests/functional/restart/38-auto-restart-stopping.t
wxtim: It was hiding the failure of this command to do anything useful - without `--messages`, the `:started` modifier doesn't do anything (except return a warning that started is not a task state)

wxtim:
> I don't think we were deliberately swallowing errors here. There was probably an incorrect expectation that the test would abort if `cylc workflow-state` failed (whether because the desired state was not achieved, or a proper error condition like this).
> 
> All functional test scripts should probably have `set -eu` at the top - shall we do that?

This didn't seem to do the trick. Using `run_ok` did though.

oliver-sanders:
> I tested set -e and it did work. Maybe you tested with one of "ditto" cases above that weren't actually broken?
>
> run_ok is probably better though, in any case.

Would recommend going with `run_ok` for now as `set -euo pipefail` is likely to break tests and cause headaches, maybe later.

Issue #5677, opened 2023-08-10
cylc show for future tasks


hjoliver:
Currently we just report *task not found in n=0*.

Instead, it would be easy to instantiate a transient task proxy, and report its prerequisites.

oliver-sanders:
We could potentially allow `cylc show` to query task definitions (note the prereqs may differ based on cycle point), however, this value may not match the one in the workflow due to restarts/reloads.

Low priority, I'm not quite sure what the use case for this is?

hjoliver:
> Low priority, I'm not quite sure what the use case for this is?

The use case is, "I want to know what prerequisites this upcoming task has".  And they might click on (say) an `n=1` task to request that information, which currently fails and logs the above rather cryptic error message. 

> f the data isn't there, we could resort to instantiating a transient task proxy, just to return its prerequisites.

My point here was, it should be pretty easy to instantiate a task proxy at a particular cycle point, get a list of its prerequisites to satisfy the query, and then discard it.

oliver-sanders:
> The use case is, "I want to know what prerequisites this upcoming task has"

But what's the requirement for that?! It's a reasonable feature, but this feels like a weak use-case to me, can we bump this back to some-day (we have over 180 issues on 8.x).

> My point here was, it should be pretty easy to instantiate a task proxy at a particular cycle point

Yep, this would avoid any issues with reloads/restarts too.

oliver-sanders:
Should wait for the discussions on #5695

hjoliver:
> Merge with #5695 

I'm not really seeing a common thread. This is about showing information now, about a future task. That is about teeing up future behaviour (hold) of a future task.

oliver-sanders:
IMO these are all closely related, they all come down to the selection of tasks in the task pool, they all have the same CLI interface, they all share the same limitations, at the implementation level they share a large part of the code.

Addressing these limitations for one command may address the limitations for all commands or at least provide a pattern for doing so. Moreover going different ways with different commands is needlessly confusing for the end user, consistency across the CLI is important here.

hjoliver:
OK then I think we're arguing about nothing :-) We should certainly use general code for task matching (some of which already exists, of course). My point is just that there is only one command (hold) that matters for at the moment (talking about the future task matching for teeing up future behaviour), and the handful of issues you're concerned about arose from specific user reports (mostly from @ColemanTom) in relation to that command (and "show", which is different and easier, although it should use the same code and interface for choosing the tasks of course).

[...]

Issue #4357, opened 2021-08-13
task globbing: add cli interface


oliver-sanders:
Some Cylc subcommands allow us to glob over the active task window e.g. `cylc trigger`.

It would be useful if there was a way to get a list of the tasks that this would trigger before triggering.

Ideas:

* Separate command (prints list of tasks).
* `--dry` option (prints list of tasks).
* `-i, --interactive` option (prints list of tasks with a y/n prompt).

We can also print out the list of the tasks that were triggered, note this is currently blocked by #3329 due to the command queueing system.

**Additional context**

**Pull requests welcome!**

MetRonnie:
Having a separate command might be tricky, due to the two types of task matching (dumping explanation here as it may be useful to whoever tackles this):

https://github.com/cylc/cylc-flow/blob/4686ba63e26777e2e828752b1c0dfbb48cdf0ed9/cylc/flow/task_pool.py#L1523-L1526

which matches spawned tasks (hidden + main pool). It looks like this is used by `cylc kill` for example.

And

https://github.com/cylc/cylc-flow/blob/4686ba63e26777e2e828752b1c0dfbb48cdf0ed9/cylc/flow/task_pool.py#L1348-L1351

which matches any task that is in the graph, whether it has spawned or not. I think this is use by `cylc trigger`.

Additionally, `cylc hold` uses both

PR #6809, opened 2025-06-24
Bump min python from 3.7 to 3.12

MetRonnie:
Closes #6804

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests not needed
- [x] Changelog entry included if this is a change that can affect users
- [ ] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

Issue #6442, opened 2024-10-23
TaskProxy.jobs: refactor job config


oliver-sanders:
Separate out the task and job specific data into separate objects and ensure it is persisted correctly.

The `TaskProxy` object contains some data that is specific to the task, e.g:

* `tdef` - The task definition.
* `tokens` - The parsed task ID.
* `clock_trigger_time` - External wallclock dependency.
* `state` - Aggregate task state.

However, it also contains some data that is specific to particilar job submissions:

* `platform` - The platform that the current/next job submission is using / will use.
* `submit_num` - The submit num of the current / next job.
* `is_manual_submit` - Were any(?) of the submissions the result of a manual trigger?
* `local_job_file_path` - Path of the current submission.

Additionally, the task proxy has two separate objects containing specific attributes of job submissions:

* `jobs` - (potentially incomplete) list of job selected configuration settings.
* `mode_settings` - Transient job state pertaining to the current/next submission.

This is a bit of a jumble, and code interfaces can easily end up relying on interfaces which are not stable for the intended purposes (e.g. #6326).

Proposal:
* Create a `Job` object to take over from `TaskProxy.job[item]` and `TaskProxy.mode_settings`. Use an efficient store, e.g. a class with `__slots__` or a protobuf serialisation.
* Migrate other job-specific `TaskProxy` fields into the job.
* Ensure that these fields are either re*populated* from the DB on task-re-spawn/workflow-reload/workflow-restart OR re*computed* as required.
* Manage job data lifecycle for efficiency (i.e. don't hang onto data we no longer need).

PR #6920, opened 2025-08-13
id_match: extend task matching to inactive tasks for globs and families

### LINKED ISSUE #5750: `cylc release`: only need to match in the tasks-to-hold list

### LINKED ISSUE #5827: Command task matching requirements
Supersedes: #5763, #5695

See also: #5752, #5416, #5677

### TODO:

* [ ] hold
* [ ] set
* [ ] trigger
* [ ] release
* [ ] kill
* [ ] poll
* [ ] remove
* show
  * [ ] show command
  * [ ] Update the [troubleshooting](https://github.com/cylc/cylc-doc/pull/638#discussion_r1583928712 ) page to remove the note about `cylc show` not working outside the n=0 window.

### target task population

Different commands need to select tasks from different populations:

| command | population |
| ------------------- | --------------- |
| `hold`, `set`, `trigger` |  pool, future, past |
| `release`  |  held-tasks list only #5752 |
| `kill`, `poll`  |  pool only |
| `remove` |  pool, past |
| `show` | pool only (currently, but users want future & past too #5677) |

### task name wildcards

We need to support matching task names by glob (or regex) and family name in all cases (not just in the pool.

For future and past tasks this means:
- find all matching task def names (or member tasks of matching family names)
- check which are valid for the given cycle point(s)

### cycle point wildcards

We need to support matching cycle points by glob in the pool, e.g. to target all incomplete tasks.

We should not support cycle point wildcards outside of the pool (well,  `*` is certainly bad; `[5-8]` (e.g.) may be OK in principle but still dangerous and not that much of a plus to users anyway).

### task qualifiers

Use of some **task qualifiers** (i.e. `:output`) necessarily restricts matching to the pool. E.g. `cylc set --out=succeeded "*:failed"` should set all `n=0` **incomplete** failed tasks to succeeded. 

We can probably decide that use of any qualifier means pool only (is there any need to distinguish between e.g. completed-succeeded and completed-failed past tasks?)

---- 

Terminology:
- **pool:** task proxies in the `n=0` active window 
- **future**, **past**: abstract tasks anywhere else in the graph, i.e. `n>0` and `n<0`

oliver-sanders:
* Implement a single task ID matching interface which will satisfy all currently documented use cases / requirements.
* Closes #5827. Allow IDs including families or globs to match against inactive tasks.
* Closes #5750. Only match tasks to release against the "pool" of tasks which are actually held.
* Addresses #4357 & #5677 Abstract task matching from task pool objects. This simplifies its logic, but also unlocks the ability to add a command / GraphQL interface for listing IDs that match provided patterns. The remaining work required for this is outlined in https://github.com/cylc/cylc-uiserver/issues/720.
* Additionally:
  * Standardise task ID parsing, validation and standardisation and perform it upfront in `cylc.flow.commands` to ensure validation is not required in internal interfaces and sanitisation errors cannot occur there.
  * Add a subclass `TaskTokens(Tokens)` type which enforces the presence of cycle and task fields. This makes it easier to use `Tokens` objects in typed code.
  * Fix a `__hash__` computation error for `Tokens` objects which caused erroneous inequality.

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at https://github.com/cylc/cylc-doc/pull/867

### DIFF
--- a/cylc/flow/command_validation.py
+++ b/cylc/flow/command_validation.py
@@ -16,14 +16,18 @@
 
 """Cylc command argument validation logic."""
 
-
 from typing import (
+    TYPE_CHECKING,
+    Dict,
     Iterable,
     List,
     Optional,
+    Set,
+    cast,
 )
 
-from cylc.flow.exceptions import InputError
+from cylc.flow.cycling.loader import standardise_point_string
+from cylc.flow.exceptions import InputError, PointParsingError
 from cylc.flow.flow_mgr import (
     FLOW_NEW,
     FLOW_NONE,
@@ -32,8 +36,13 @@ from cylc.flow.id import (
     IDTokens,
     Tokens,
 )
-from cylc.flow.task_outputs import TASK_OUTPUT_SUCCEEDED
+from cylc.flow.id_cli import contains_fnmatch
 from cylc.flow.scripts.set import XTRIGGER_PREREQ_PREFIX
+from cylc.flow.task_outputs import TASK_OUTPUT_SUCCEEDED
+
+
+if TYPE_CHECKING:
+    from cylc.flow.id import TaskTokens
 
 
 ERR_OPT_FLOW_VAL_INT_NEW_NONE = (  # for set and trigger commands
@@ -49,7 +58,7 @@ ERR_OPT_FLOW_WAIT = (
 def flow_opts(
     flows: List[str],
     flow_wait: bool,
-    allow_new_or_none: bool = True
+    allow_new_or_none: bool = True,
 ) -> None:
     """Check validity of flow-related CLI options.
 
@@ -235,9 +244,7 @@ def prereq(prereq: str) -> Optional[str]:
         return None
 
     if tokens["cycle"] == XTRIGGER_PREREQ_PREFIX:
-        if (
-            tokens["task_sel"] not in {None, TASK_OUTPUT_SUCCEEDED}
-        ):
+        if tokens["task_sel"] not in {None, TASK_OUTPUT_SUCCEEDED}:
             # Error: xtrigger status must be default or succeeded.
             return None
         if tokens["task_sel"] is None:
@@ -291,7 +298,7 @@ def outputs(outputs: Optional[List[str]]):
 
 def consistency(
     outputs: Optional[List[str]],
-    prereqs: Optional[List[str]]
+    prereqs: Optional[List[str]],
 ) -> None:
     """Check global option consistency
 
@@ -309,40 +316,73 @@ def consistency(
         raise InputError("Use --prerequisite or --output, not both.")
 
 
-def is_tasks(tasks: Iterable[str]):
-    """All tasks in a list of tasks are task ID's without trailing job ID.
+def is_tasks(ids: Iterable[str]) -> 'Set[TaskTokens]':
+    """Ensure all IDs are task IDs and standardise them.
 
-    Examples:
-        # All legal
-        >>> is_tasks(['1/foo', '1/bar', '*/baz', '*/*'])
+    * Parses IDs.
+    * Filters out job ids and ensures at least the cycle point is provided.
+    * Standardises the cycle point format.
+    * Defaults the namespace to "root" unless provided.
 
-        # Some legal
-        >>> is_tasks(['1/foo/NN', '1/bar', '*/baz', '*/*/42'])
-        Traceback (most recent call last):
-        ...
-        cylc.flow.exceptions.InputError: This command does not take job ids:
-        * 1/foo/NN
-        * */*/42
+    Args:
+        ids: The strings to parse.
 
-        # None legal
-        >>> is_tasks(['*/baz/12'])
-        Traceback (most recent call last):
-        ...
-        cylc.flow.exceptions.InputError: This command does not take job ids:
-        * */baz/12
+    Returns:
+        The parsed IDs as TaskTokens objects.
 
-        >>> is_tasks([])
-        Traceback (most recent call last):
-        ...
-        cylc.flow.exceptions.InputError: No tasks specified
+    Raises:
+        InputError: If any of the IDs cannot be parsed or formatted.
+
     """
-    if not tasks:
+    if not ids:
         raise InputError("No tasks specified")
-    bad_tasks: List[str] = []
-    for task in tasks:
-        tokens = Tokens('//' + task)
+    ret: 'Set[TaskTokens]' = set()
+    errors: Dict[str, List[str]] = {}
+
+    for id_ in ids:
+        # parse id
+        try:
+            tokens = Tokens(id_, relative=True)
+        except ValueError:
+            errors.setdefault('Invalid ID', []).append(id_)
+            continue
+
+        # filter out job IDs
         if tokens.lowest_token == IDTokens.Job.value:
-            bad_tasks.append(task)
-    if bad_tasks:
-        msg = 'This command does not take job ids:\n * '
-        raise InputError(msg + '\n * '.join(bad_tasks))
+            errors.setdefault('This command does not take job IDs', []).append(
+                id_
+            )
+            continue
+
+        # if the task is not specified, default to "root"
+        if tokens['task'] is None:
+            tokens = tokens.duplicate(task='root')
+
+        # if the cycle is not a glob or reference, standardise it
+        if (
+            # cycle point is a glob
+            not contains_fnmatch(cast('str', tokens['cycle']))
+            # cycle point is a reference to the ICP/FCP
+            and tokens['cycle'] not in {'^', '$'}
+        ):
+            try:
+                cycle = standardise_point_string(tokens['cycle'])
+            except PointParsingError:
+                errors.setdefault('Invalid cycle point', []).append(id_)
+                continue
+            else:
+                if cycle != tokens['cycle']:
+                    tokens = tokens.duplicate(cycle=cycle)
+
+        # we have confirmed that both cycle and task have been provided
+        ret.add(cast('TaskTokens', tokens))
+
+    if errors:
+        raise InputError(
+            '\n'.join(
+                f'{message}: {", ".join(sorted(_ids))}'
+                for message, _ids in sorted(errors.items())
+            )
+        )
+
+    return ret
--- a/cylc/flow/commands.py
+++ b/cylc/flow/commands.py
@@ -730,6 +730,7 @@ def _force_trigger_tasks(
 
     # Record off-group prerequisites, and active tasks to be removed.
     active_to_remove: List[TaskTokens] = []
+    active_to_resume: Set[TaskTokens] = set()
 
     warnings_flow_none = []
     warnings_has_job = []
@@ -794,6 +795,7 @@ def _force_trigger_tasks(
 
         else:
             active_to_remove.append(itask.tokens.task)
+            active_to_resume.add(itask.tokens.task)
 
     if warnings_flow_none:
         msg = '\n  * '.join(warnings_flow_none)
@@ -807,6 +809,12 @@ def _force_trigger_tasks(
     if flow != [FLOW_NONE]:
         # (No need to remove tasks if triggering with no-flow).
         _remove_matched_tasks(schd, {*active_to_remove, *inactive}, flow_nums)
+
+        # trigger should override the held state, however, in-group tasks may
+        # have previously been held and active in-group tasks will become
+        # held as the result of removal
+        schd.pool.release_held_tasks({*active_to_resume, *inactive})
+
         # Store removal results before moving on.
         schd.workflow_db_mgr.process_queued_ops()
 
--- a/cylc/flow/commands.py
+++ b/cylc/flow/commands.py
@@ -67,7 +67,6 @@ from typing import (
     List,
     Optional,
     Set,
-    Tuple,
     TypeVar,
 )
 
@@ -84,7 +83,7 @@ from cylc.flow.exceptions import (
 )
 import cylc.flow.flags
 from cylc.flow.flow_mgr import FLOW_NONE, repr_flow_nums
-from cylc.flow.id import Tokens
+from cylc.flow.id import TaskTokens
 from cylc.flow.log_level import log_level_to_verbosity
 from cylc.flow.parsec.exceptions import ParsecError
 from cylc.flow.prerequisite import PrereqTuple
@@ -102,11 +101,9 @@ from cylc.flow.workflow_status import StopMode
 if TYPE_CHECKING:
     from enum import Enum
 
-    from cylc.flow.cycling import PointBase
     from cylc.flow.flow_mgr import FlowNums
     from cylc.flow.scheduler import Scheduler
     from cylc.flow.task_proxy import TaskProxy
-    from cylc.flow.taskdef import TaskDef
 
     # define a type for command implementations
     Command = Callable[..., AsyncGenerator]
@@ -144,45 +141,61 @@ def _command(name: str):
     return _command
 
 
+def _report_unmatched(unmatched: Set[TaskTokens]):
+    """Log unmatched IDs."""
+    if len(unmatched) == 1:
+        LOG.warning(
+            'No tasks match'
+            f' "{next(iter(unmatched)).relative_id_with_selectors}"'
+        )
+    elif len(unmatched) > 1:
+        LOG.warning(
+            'No tasks match the IDs:\n* '
+            + '\n* '.join(
+                sorted([id_.relative_id_with_selectors for id_ in unmatched])
+            )
+        )
+
+
 def _remove_matched_tasks(
     schd: 'Scheduler',
-    active: 'List[TaskProxy]',
-    inactive: 'Set[Tuple[TaskDef, PointBase]]',
-    flow_nums: 'FlowNums'
+    ids: Set[TaskTokens],
+    flow_nums: 'FlowNums',
 ):
     """Remove matched tasks."""
-
     # Mapping of *relative* task IDs to removed flow numbers:
-    removed: Dict[Tokens, FlowNums] = {}
-    not_removed: Set[Tokens] = set()
-    # All the matched tasks (will add applicable active tasks below):
-    matched_tasks = inactive.copy()
+    removed: Dict[TaskTokens, FlowNums] = {}
+    not_removed: Set[TaskTokens] = set()
     to_kill: List[TaskProxy] = []
 
-    for itask in active:
-        fnums_to_remove = itask.match_flows(flow_nums)
-        if not fnums_to_remove:
-            not_removed.add(itask.tokens.task)
-            continue
-        removed[itask.tokens.task] = fnums_to_remove
-        matched_tasks.add((itask.tdef, itask.point))
-        if fnums_to_remove == itask.flow_nums:
-            # Need to remove the task from the pool.
-            # Spawn next occurrence of xtrigger sequential task (otherwise
-            # this would not happen after removing this occurrence):
-            schd.pool.check_spawn_psx_task(itask)
-            schd.pool.remove(itask, 'request')
-            to_kill.append(itask)
-            itask.removed = True
-        itask.flow_nums.difference_update(fnums_to_remove)
+    for id_ in ids:
+        itask = schd.pool._get_task_by_id(id_.relative_id)
 
-    for tdef, point in matched_tasks:
-        tokens = Tokens(cycle=str(point), task=tdef.name)
+        if itask:
+            # remove active task from the pool
+            fnums_to_remove = itask.match_flows(flow_nums)
+            if not fnums_to_remove:
+                not_removed.add(itask.tokens.task)
+                continue
+            removed[itask.tokens.task] = fnums_to_remove
+            if fnums_to_remove == itask.flow_nums:
+                # Need to remove the task from the pool.
+                # Spawn next occurrence of xtrigger sequential task (otherwise
+                # this would not happen after removing this occurrence):
+                schd.pool.check_spawn_psx_task(itask)
+                schd.pool.remove(itask, 'request')
+                to_kill.append(itask)
+                itask.removed = True
+            itask.flow_nums.difference_update(fnums_to_remove)
+
+        # remove task from the DB
+        tdef = schd.config.taskdefs[id_['task']]
+        icycle = get_point(id_['cycle'])
 
         # Go through any tasks downstream of this matched task to see if
         # any need to stand down as a result of this task being removed:
         for child in set(itertools.chain.from_iterable(
-            generate_graph_children(tdef, point).values()
+            generate_graph_children(tdef, icycle).values()
         )):
             child_itask = schd.pool.get_task(child.point, child.name)
             if not child_itask:
@@ -197,9 +210,9 @@ def _remove_matched_tasks(
             ):
                 # Unset any prereqs naturally satisfied by these tasks
                 # (do not unset those satisfied by `cylc set --pre`):
-                if prereq.unset_naturally_satisfied(tokens.relative_id):
+                if prereq.unset_naturally_satisfied(id_.relative_id):
                     prereqs_changed = True
-                    removed.setdefault(tokens, set()).update(
+                    removed.setdefault(id_, set()).update(
                         fnums_to_remove
                     )
             if not prereqs_changed:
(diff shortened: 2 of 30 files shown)


cylc/flow/id_match.py
oliver-sanders, line 1: The bulk of the work is in this file (closes #5827), most of the rest of this PR is just interface changes.

cylc/flow/command_validation.py
```diff
@@ -309,40 +316,73 @@ def consistency(
...
+def is_tasks(ids: Iterable[str]) -> 'Set[TaskTokens]':
+    """Ensure all IDs are task IDs and standardise them.
 
-    Examples:
-        # All legal
-        >>> is_tasks(['1/foo', '1/bar', '*/baz', '*/*'])
+    * Parses IDs.
+    * Filters out job ids and ensures at least the cycle point is provided.
+    * Standardises the cycle point format.
+    * Defaults the namespace to "root" unless provided.
```
oliver-sanders, line 325: On this PR, all parsing, validating and standardising of IDs is done upfront at validate time.

There is no longer any re-parsing of IDs so errors can only crop up at validate time, not in the lower-level code.

oliver-sanders:
I couldn't implement the task matching for Cylc show (#5677) in this PR because it requires UIS changes, however, I have adapted the interface into something that can work independently of the task pool which unblocks this work, pending https://github.com/cylc/cylc-uiserver/issues/720.

cylc/flow/run_modes/simulation.py
```diff
@@ -162,7 +162,7 @@ def __init__(
             db_info = db_mgr.pri_dao.select_task_job(
                 itask.tokens['cycle'],
                 itask.tokens['task'],
-                itask.tokens['job'],
+                itask.submit_num,
```
oliver-sanders, line 161: Unrelated fix spotted by MyPy as the result of type changes on this branch:

* `TaskProxy.tokens['job']` is always `None` (because it's a task proxy not a [job proxy](https://github.com/cylc/cylc-flow/issues/6442)).
* This caused the DB check to retrieve the latest job, which is probably what we want.
* Made this explicit by passing through the actual job number.

hjoliver:
(Resolved some minor conflicts via GitHub UI, so the test could run).

cylc/flow/commands.py
```diff
@@ -144,45 +141,56 @@ def _command(fcn: '_TCommand') -> '_TCommand':
     return _command
 
 
+def _report_unmatched(unmatched: Set[TaskTokens]):
+    """Log unmatched IDs."""
+    if len(unmatched) == 1:
+        LOG.warning(f'No tasks match "{next(iter(unmatched)).relative_id}"')
```
hjoliver:
```suggestion
        LOG.warning(f'No tasks match "{next(iter(unmatched)).relative_id_with_selectors}"')
```
(Otherwise the "No tasks match" message looks wrong, if it was the selector that did not match).

hjoliver:
This seems to work really well.

Need to update the ID matching help in:
 -  `scripts/cylc.py` 
 -  `option_parsers.py`

cylc/flow/commands.py
oliver-sanders: Also needed to be done on the following branch, addressed in latest commit.

cylc/flow/option_parsers.py
```diff
@@ -241,21 +241,28 @@ class CylcOptionParser(OptionParser):
...
 
-            Globs (note: quote globs; they only match in the active-window):
-                # Match any active-window task "foo" in all cycles
+                # Operate on all members of a family
+                //cycle-1/FAMILY-NAME
+
+                # Operate on all tasks in a cycle
+                //cycle  # shorthand for '//cycle/root'
+
+            Globs (note: quote globs or they will expand in your shell):
```
hjoliver:
```suggestion
            Globs (note: quote globs or they might expand in your shell):
```
No biggie, but "might" is more accurate (they won't expand if they don't match anything).

cylc/flow/option_parsers.py
```diff
@@ -241,21 +241,28 @@ class CylcOptionParser(OptionParser):
...
+                # Match all tasks in cycle "1" with bames beginning with "foo"
+                ''//1/foo*'
+
+                # Match the task "foo" in all active (i.e. n=0) cycles
                 '//*/foo'
 
-                # Match the tasks "foo-1" and "foo-2"
+                # Match the tasks "foo-1" and "foo-2"c in all active cycles.
```
hjoliver:
```suggestion
                # Match the tasks "foo-1" and "foo-2" in all active cycles.
```

cylc/flow/id.py
```diff
@@ -37,6 +39,11 @@
 if TYPE_CHECKING:
     from cylc.flow.cycling import PointBase
 
+# BACK COMPAT: typing_extensions.Literal
+# FROM: Python 3.7
+# TO: Python 3.8
+    from typing_extensions import Literal
+
```
hjoliver: Will need tweaking very soon: #6809

cylc/flow/scripts/cylc.py
```diff
@@ -199,8 +199,6 @@ def get_version(long=False):
 
     Warning:
       Quote IDs on the command line to protect them from shell expansion.
-      Patterns only match tasks in the n=0 active window (except for the
-      `cylc show` command where they match in the wider n-window).
```
hjoliver, line 197: Can you tweak the main text above this as necessary too.

Maybe this one thing is enough (I haven't got time right now to think too deeply about it):

> `workflow//*            # All cycles in workflow`

Probably need to say "all active cycles" there.

oliver-sanders: Done.

cylc/flow/id.py
oliver-sanders: Rebased.

cylc/flow/option_parsers.py
oliver-sanders: Done.

cylc/flow/option_parsers.py
oliver-sanders: Done.

oliver-sanders:
(rebased)

cylc/flow/scripts/cylc.py
```diff
@@ -203,7 +201,7 @@ def get_version(long=False):
 
     Examples:
       *:running                       # All running workflows
-      workflow//*:running             # All running cycles in workflow
+      workflow//*:running             # All running tasks in workflow
```
hjoliver, line 204: Still need to say (e..g above under the "Filters" heading) that filters only match in the `n=0` window.

Otherwise there could be an expectation that e.g. `cycle/*:failed` will match final-status failed tasks in a past cycle.

cylc/flow/option_parsers.py
```diff
@@ -241,21 +241,28 @@ class CylcOptionParser(OptionParser):
...
+
+                # Match the task "foo" in all active (i.e. n=0) cycles
                 '//*/foo'
 
-                # Match the tasks "foo-1" and "foo-2"
+                # Match the tasks "foo-1" and "foo-2" in all active cycles.
                 '//*/foo-[12]'
 
-            Selectors (note: selectors only match in the active window too):
+            Selectors (note: selectors only match active tasks):
```
hjoliver, line 265: (See my comment above on "filters" in `cylc.py`)

We seem to be calling these both "selectors" and "filters".  I think I prefer "selectors", which can be explained as "selecting specific task states". Whereas "filter" might suggest more general filtering behaviour.

cylc/flow/scripts/cylc.py
oliver-sanders: Selectors can also be applied to workflows, hence the vagueness.

Will add a note for task selection.

cylc/flow/option_parsers.py
oliver-sanders: Will change to selector.

oliver-sanders: > (See my comment above on "filters" in cylc.py)

Isn't this good enough:

> Selectors (note: selectors only match active tasks):

?

hjoliver: (Yes, it's good enough here, I just meant I made my other comment on the same topic before seeing use of "selector" here rather than "filter").

oliver-sanders:
(rebased to pick up test fix)

cylc/flow/commands.py
```diff
@@ -144,45 +141,61 @@ def _command(fcn: '_TCommand') -> '_TCommand':
     return _command
 
 
+def _report_unmatched(unmatched: Set[TaskTokens]):
```
wxtim, line 144: Why not build this in to `id_match`?

cylc/flow/task_pool.py
```diff
@@ -2460,129 +2483,28 @@ def log_task_pool(self, log_lvl=logging.DEBUG):
...
-                * Specifies a cycle point.
-                * Is not a pattern. (e.g. `*/foo`).
-                * Does not contain a state selector (e.g. `:failed`).
-
-        Returns:
-            (matched, inactive_matched, unmatched)
+        ids: Set[TaskTokens],
+        only_match_pool: bool = False,
+    ) -> Tuple[Set[TaskTokens], Set[TaskTokens]]:
+        """Match IDs against active tasks in the pool."""
```
wxtim, line 2491: Why have we lost so much of the explanation (especially the `returns:` bit)

cylc/flow/commands.py
oliver-sanders: We don't always want to report unmatched tasks (e.g, we have some internal uses for this interface). And we might want to change the message / reporting on a command by command basis.

We could use a `log_unmatched` flag, but this can get messy. I've centralised the part of the code which handles user input  in one place, and not leaked its logic elsewhere.

A few commands aren't logging unmatched IDs at all, I haven't attempted to resolve that in this PR though.

cylc/flow/task_pool.py
```diff
@@ -2460,129 +2483,28 @@ def log_task_pool(self, log_lvl=logging.DEBUG):
             )
         )
 
-    def filter_task_proxies(
+    def id_match(
```
wxtim, line 2486: Would it not be tidier to remove this entirely and either
* Pass the task pool object as an argument to `id_match.id_match`
* Pass `pool.active_tasks.values()` and `pool.config` to `id_match.id_match`

oliver-sanders: This adaptor exists in order to abstract the ID matching code from the task pool.

This is necessary to close https://github.com/cylc/cylc-flow/issues/5750 (where we need to match against the list of held tasks, not the task pool), as well as for future plans e.g, https://github.com/cylc/cylc-flow/issues/4357 & https://github.com/cylc/cylc-flow/issues/5677 (which need ID matching in cylc-uiserver where the task pool is not available).

It also simplifies the `id_match` code which no longer has to juggle itask objects and can focus on pure string manipulation.

cylc/flow/id_match.py
```diff
@@ -14,250 +14,234 @@
...
+                        ):
+                            LOG.warning(
+                                'Invalid cycle point for task:'
+                                f' {id__["task"]}, {id__["cycle"]}'
                             )
-                            or match(itask.state.status, task_sel)
-                        )
-                    ):
-                        tasks.append(itask)
+                except (ValueError, KeyError):
```
wxtim, line 196: This isn't tested by `tests/u/test_id_match` - would a unit test be in order?

cylc/flow/id_match.py
```diff
@@ -14,250 +14,234 @@
...
+                for pool_task_id in pool
+            })
+        else:
+            # filter for on-sequence task instances
+            for id__ in list(_matched):
+                try:
+                    taskdef = config.taskdefs[id__['task']]
+                    if not taskdef.is_valid_point(get_point(id__['cycle'])):
+                        _matched.remove(id__)
+                        if (
```
wxtim, line 184: None of the unit tests in `tests/u/test_id_match` hit this and retunr `False`

cylc/flow/id_match.py
oliver-sanders: This is covered in `tests/i/test_id_match.py`.

There are two exceptions here:

* `ValueError` - Caused by an invalid cycle point.
  * This is covered by the `not-a-cycle` value in the tests.
* `KeyError` - caused by a missing taskdef.
  * I was unable to find a scenario in which the taskdef was not present, however, this still fealt like an appropriate defence (we have done the same elsewhere).
  * The `test_match_removed_task` test orphans a task (the most likely thing I could think of to cause a missing taskdef).
  * At present the taskdef is present, so the `KeyError` does not get raised.
  * But if that changes in the future, this test will cover it.

cylc/flow/id_match.py
oliver-sanders: Yes, and that fact revealed that the check was buggy!

Have added an explicit test (in `tests/i`) and some comments to explain why this check is here: https://github.com/cylc/cylc-flow/compare/ebccd5127508882150f27ed7eed24626831b6ca0..86c4b933048574aca62bda8da6d1ed3844e39655

oliver-sanders: resulting test failures are invalid, will resolve...

...done

tests/unit/test_id_match.py
```diff
@@ -14,376 +14,262 @@
...
+
+            [scheduling]
+                cycling mode = integer
+                initial cycle point = 1
+                [[graph]]
+                    P1 = a => b => c => d
+                    P3 = z
+        '''))
+
+    return WorkflowConfig('test', str(path), SimpleNamespace())
```
wxtim, line 59: Off topic, but you can avoid touching IO here.

Also takes about .7 seconds not 2.7.

It might also be worth separating the config and the method so that you can re-use this for the last test. I even wonder if we might want to make the "create workflow config without touching file system" a generically available fixture?
```suggestion
    msg = (dedent('''
            [scheduler]
                allow implicit tasks = True

            [scheduling]
                cycling mode = integer
                initial cycle point = 1
                [[graph]]
                    P1 = a => b => c => d
                    P3 = z
        '''))
    def mock_read_and_proc(fpath, *_, **__):
        if fpath.endswith('yellow'):
            return [f.rstrip() for f in msg.split('\n')]
        else:
            return read_and_proc(fpath, *_, **__)
    monkeypatch.setattr(
        'cylc.flow.parsec.fileparse.read_and_proc',
        mock_read_and_proc
    )
```

oliver-sanders: This is a tad off-topic as this comment could also be made about all other integration tests (the `flow` fixture writes to the FS too).

But a valid comment anyway. Mocking in this way is a bit ugly and may cause errors as the `WorkflowConfig` does make use of the filepath, e.g:

https://github.com/cylc/cylc-flow/blob/700eb0217f25fbe178b0eed2b432ca7f7c4a00db/cylc/flow/config.py#L293

https://github.com/cylc/cylc-flow/blob/700eb0217f25fbe178b0eed2b432ca7f7c4a00db/cylc/flow/config.py#L950

https://github.com/cylc/cylc-flow/blob/700eb0217f25fbe178b0eed2b432ca7f7c4a00db/cylc/flow/config.py#L1706

https://github.com/cylc/cylc-flow/blob/700eb0217f25fbe178b0eed2b432ca7f7c4a00db/cylc/flow/config.py#L2001

So it remains correct for the integration tests to use the filesystem at present.

However, the desire to get the `WorkflowConfig` to the point where it can function without the filesystem path (e.g, so you can pipe in a string config) is there. Unlikely useful for Cylc, but defo handy for tests.

A refactor opportunity for another day.

tests/unit/test_command_validation.py
```diff
@@ -37,3 +42,56 @@ async def test_trigger_invalid(flow_strs, expected_msg):
...
+
+    # the namespace should default to "root" unless provided
+    assert is_tasks({'*', '2000'}) == {
+        TaskTokens('*', 'root'),
+        TaskTokens('20000101T0000Z', 'root'),
+    }
+
+    # invalid IDs result in errors
+    with pytest.raises(InputError, match='Invalid ID: //, ///, ////'):
+        is_tasks({'//', '///', '////'})
```
wxtim: Convince me that it _only_ identifies the invalid ones?
```suggestion
        is_tasks({'//', '///', '////', '2000/a'})
```

tests/unit/test_command_validation.py
```diff
@@ -37,3 +42,56 @@ async def test_trigger_invalid(flow_strs, expected_msg):
...
+
+    # invalid cycle points result in errors
+    with pytest.raises(
+        InputError,
+        match=re.escape('Invalid cycle point: (42)/answer, 2000Z, abc'),
+    ):
+        is_tasks({'(42)/answer', '2000Z', 'abc'})
```
wxtim: And here
```suggestion
        is_tasks({'(42)/answer', '2000Z', 'abc', '2000/a'})
```

hjoliver:
(I'll likely approve this tomorrow, just need to spend a little more time on it).

tests/integration/test_id_match.py
```diff
@@ -0,0 +1,214 @@
...
+
+        # task pool state:
+        # * cycle 1
+        #   * n=0 a1 waiting
+        #   * n=1 b1 waiting
+        #   * n=2 c1 waiting
+        #   * n=1 a2 succeeded
+        #   * n=0 b2 failed
+        #   * n=1 c2 waiting
+        # * cycle 2
```
wxtim, line 81: Point 3 would help the reader with the expected results below - Cycle 2 is ambiguous.

tests/integration/test_id_match.py
```diff
@@ -0,0 +1,214 @@
...
+        #   * n=1 b2 waiting
+        #   * n=2 c2 waiting
+
+        # check the n=0 window matches expecations before proceeding
+        assert {
+            itask.tokens.relative_id for itask in schd.pool.get_tasks()
+        } == {'1/a1', '1/b2', '3/a1', '3/a2'}
+
+        # test active task selection
+        assert (
```
wxtim, line 95: If at some point in the future someone breaks a selection (but not all) of these equalities, this will be painful to debug. My heart's not set on it, but what about:

```python
n0_waiting = ({'1/a1', '3/a1', '3/a2'}, set())
assert match('*:waiting') == n0_waiting
assert match('*/root:waiting') == n0_waiting
...

or even:
```python
for search_term in [...]:
    assert match(search_term) == n0_waiting

```

tests/integration/test_id_match.py
```diff
@@ -0,0 +1,214 @@
...
+            == match('1/b2:failed')
+            == match('^/b2:failed')
+            == ({'1/b2'}, set())
+        )
+
+        assert (
+            match('1/b1:failed', '1/b2:failed')
+            == match('1/B:failed', '1/b1:failed')
+            == match('*:failed', '1/B:failed', '1/b1:failed')
+            == ({'1/b2'}, {'1/b1:failed'})
```
wxtim, line 121: The unmatched result here surprised me. I'll be poking this to see if I'm unhappy with the consequences.

tests/integration/test_id_match.py
```diff
@@ -0,0 +1,214 @@
...
+        )
+
+        assert match('*:submit-failed') == (set(), {'*:submit-failed'})
+
+        # test regular task selection
+        assert (
```
wxtim, line 130: Can we add an example to the workflow to check that match('*') can't produce infinite results?

tests/integration/test_id_match.py
```diff
@@ -0,0 +1,214 @@
...
+
+        assert (
+            match('1/A')
+            == match('1/a*')
+            == match('^/a*')
+            == match('1/a1', '1/a2')
+            == ({'1/a1', '1/a2'}, set())
+        )
+
+        assert match('1/X') == (set(), {'1/X'})
```
wxtim, line 163: I think that these need comments about what they are testing.

tests/integration/test_id_match.py
```diff
@@ -0,0 +1,214 @@
...
+        assert match('5:waiting') == (set(), {'5:waiting'})
+
+        # cycle 2 is not on sequence for task a1
+        assert match('1/a1', '2/a1', '3/a1') == ({'1/a1', '3/a1'}, {'2/a1'})
+
+        # errors
```
wxtim: Can you explain why there is not a check for an error or a warning in the comments?

tests/integration/test_force_trigger.py
```diff
@@ -752,3 +756,52 @@ async def test_trigger_with_sequential_task(flow, scheduler, run, log_filter):
...
+    """Test task matching with the trigger command.
+
+    This test is intended to extend the other integration tests for ID matching
+    with a real use case to ensure the code in cylc.flow.commands
+    (which parses and standardises IDs) is working correctly.
+    """
+    id_ = flow({
+        'scheduling': {
+            'graph': {
+                'R1': 'a & b & c & d & e & f & g'
```
wxtim, line 771: Easier to read IMO
```suggestion
                'R1': '&'.join('abcdefg')
```

tests/integration/test_id_match.py
oliver-sanders: There are other tests for `*` which already do this.

tests/integration/test_force_trigger.py
oliver-sanders: Harder to read IMO.

cylc/flow/id_match.py
```diff
@@ -14,250 +14,238 @@
...
-
-def point_match(
-    point: 'PointBase', value: str, pattern_match: bool = True
-) -> bool:
-    """Return whether a cycle point matches a string/pattern.
+    return matched, unmatched
+
+
+def _fnmatchcase_glob(pattern: str, values: Iterable[str]) -> Set[str]:
+    """Convenience function for globing over a list of values.
```
hjoliver:
```suggestion
    """Convenience function for globbing over a list of values.
```

cylc/flow/option_parsers.py
```diff
@@ -241,21 +241,28 @@ class CylcOptionParser(OptionParser):
...
-            Globs (note: quote globs; they only match in the active-window):
-                # Match any active-window task "foo" in all cycles
+                # Operate on all members of a family
+                //cycle-1/FAMILY-NAME
+
+                # Operate on all tasks in a cycle
+                //cycle  # shorthand for '//cycle/root'
+
+            Globs (note: quote globs or they might expand in your shell):
+                # Match all tasks in cycle "1" with bames beginning with "foo"
```
hjoliver:
```suggestion
                # Match all tasks in cycle "1" with names beginning with "foo"
```

tests/integration/test_flow_assignment.py
```diff
@@ -129,77 +134,52 @@ async def test_flow_assignment(
...
-            do_command(['1/a'], flow=[])
-        else:
-            await run_cmd(do_command(['1/a'], flow=[]))
+        await run_cmd(do_command(['1/a'], flow=[]))
         assert schd.pool._get_task_by_id('1/a').flow_nums == {1, 2, 3}
 
         # Else assign requested flows.
         # Run as no-flow:
-        from cylc.flow import LOG
-        LOG.critical("ZER")
```
hjoliver, line 181: Oops 🤯

hjoliver (approved):
💯

wxtim (approved):
- [x] Read code changes.
- [x] Read tests (Made some suggestions, but none which ought to block merge).
- [x] Manually tested. Not found any problems caused by this PR, although independently re-discovered other bugs.
- [x] Manually tested whilst reading https://github.com/cylc/cylc-doc/pull/867/files
- [x] Saved my Gradmother from the ravenous bugblatter beast of Traal.

tests/integration/test_id_match.py
oliver-sanders: The test contains this check at the start:

```python
        assert {
            itask.tokens.relative_id for itask in schd.pool.get_tasks()
        } == {'1/a1', '1/b2', '3/a1', '3/a2'}
```

Which will catch any potential future changes to the n=0 window.

I don't think it'll help to capture that in a variable for reuse, partly as the checks often select a subset of these values, partly because the task matching rules will probably need to be reinterpreted for these cases if this were to occur.

oliver-sanders:
(added more comments and resolved some above comments)

hjoliver:
Two approvals, no important unresolved comments, unrelated test failures (swarm build mamba solve failed, due to slow download?) ... merging.

PR #6926, opened 2025-08-15
Fix data store missing info for submit-failed jobs

MetRonnie:
- Submitted time[^1]
- Job runner
- Job ID in the job runner (on restart, as it was wiped from DB on submit-failure)

[^1]: Submitted time in this case is really the time of failed submission, but useful to know. It is recorded in the DB but wasn't showing in the UI.

**Check List**

- [x] No dependency changes
- [x] Docs PR not needed

### DIFF
--- a/cylc/flow/data_store_mgr.py
+++ b/cylc/flow/data_store_mgr.py
@@ -67,7 +67,6 @@ from typing import (
     Set,
     TYPE_CHECKING,
     Tuple,
-    Union,
 )
 import zlib
 
@@ -1612,25 +1611,18 @@ class DataStoreMgr:
 
     def insert_job(
         self,
-        name: str,
-        cycle_point: Union['PointBase', str],
+        itask: 'TaskProxy',
         status: str,
         job_conf: dict,
-    ):
+    ) -> None:
         """Insert job into data-store.
 
         Args:
-            name: Corresponding task name.
-            cycle_point: Cycle point string
             status: The task's state.
             job_conf:
                 Dictionary of job configuration used to generate
                 the job script.
                 (see TaskJobManager._prep_submit_task_job_impl)
-
-        Returns:
-
-            None
 
         """
         if status not in JOB_STATUS_SET:
@@ -1639,10 +1631,7 @@ class DataStoreMgr:
             return
 
         sub_num = job_conf['submit_num']
-        tp_tokens = self.id_.duplicate(
-            cycle=str(cycle_point),
-            task=name,
-        )
+        tp_tokens = self.id_.duplicate(itask.tokens)
         tproxy: Optional[PbTaskProxy]
         tp_id, tproxy = self.store_node_fetcher(tp_tokens)
         if not tproxy:
@@ -1665,6 +1654,7 @@ class DataStoreMgr:
             execution_time_limit=job_conf.get('execution_time_limit'),
             platform=job_conf['platform']['name'],
             job_runner_name=job_conf.get('job_runner_name'),
+            job_id=itask.summary.get('submit_method_id'),
         )
         # Not all fields are populated with some submit-failures,
         # so use task cfg as base.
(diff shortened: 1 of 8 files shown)


cylc/flow/task_job_mgr.py
```diff
@@ -1286,6 +1286,12 @@ def _prep_submit_task_job(
                     # bad_hosts:
                     self.bad_hosts -= exc.hosts_consumed
                     self._set_retry_timers(itask, rtconfig)
+                # Provide dummy platform otherwise it will show as localhost
+                # in the data store:
```
oliver-sanders: Context easily lost, worth linking this in:

```suggestion
                # in the data store:
                # see https://github.com/cylc/cylc-flow/pull/6926
```

oliver-sanders (approved):
LGTM

cylc/flow/task_events_mgr.py
```diff
@@ -794,13 +794,7 @@ def process_message(
             self._process_message_submitted(itask, event_time, forced)
             self.spawn_children(itask, TASK_OUTPUT_SUBMITTED, forced)
 
-            # ... but either way update the job ID in the job proxy (it only
-            # comes in via the submission message).
```
MetRonnie, line 798: Now it is done by `data_store_mgr.insert_job()`

cylc/flow/task_events_mgr.py
```diff
@@ -1465,7 +1459,6 @@ def _process_message_submit_failed(
             "time_submit_exit": event_time,
             "submit_status": 1,
         })
-        itask.summary['submit_method_id'] = None
```
MetRonnie, line 1468: Don't know why the job ID was being wiped on submit-failure

cylc/flow/data_store_mgr.py
```diff
@@ -1612,37 +1611,27 @@ def _apply_broadcasts_to_runtime(self, tokens, rtconfig):
 
     def insert_job(
         self,
-        name: str,
-        cycle_point: Union['PointBase', str],
+        itask: 'TaskProxy',
```
oliver-sanders, line 1614: Please try to avoid passing itask objects to the data store where possible.

We have had to do this in a couple of places, but we don't need to update the remaining interfaces to match.

In theory, we are supposed to be able to populate the data store out of the data base (without the Scheduler or its runtime objects, e.g. TaskProxy) so we can provide offline data.

In truth that isn't possible right now, but we should try to reduce the pain of refactor when the time comes.

cylc/flow/task_events_mgr.py
oliver-sanders: I don't know either, but there may be a reason, if not part of the bugfix, plz bump.

oliver-sanders: Actually, is this the job ID or the job submit method?

cylc/flow/data_store_mgr.py
```diff
@@ -1665,6 +1654,7 @@ def insert_job(
             execution_time_limit=job_conf.get('execution_time_limit'),
             platform=job_conf['platform']['name'],
             job_runner_name=job_conf.get('job_runner_name'),
+            job_id=itask.summary.get('submit_method_id'),
```
oliver-sanders, line 1657: Note, there is no Job ID for a submission failure caused by a platform lookup error because no job submission was made.

cylc/flow/task_events_mgr.py
MetRonnie: Job ID in the job runner, it's part 3 of the bugfix

cylc/flow/data_store_mgr.py
MetRonnie: `.get()` will return `None` in this case, which is fine

wxtim (approved):

PR #6944, opened 2025-08-27
Make a new standard template variable "CYLC_SOURCE_DIRECTORY"

### LINKED ISSUE #6830: config: provide access to source dir location
Provide the workflow source directory as a template variable.

This is a barrier to converting `rose-stem` suites into plain Cylc workflows as the rose-stem sources logic handles this at present.

I thought we could sneakily so something like this via cylc-rose:

```
# rose-suite.conf

[template variables]
SOURCE_PATH="$PWD/"
```

But it turns out that this isn't presently an option. I think the issue was `$PWD` differences between `cylc install` and `cylc vip`?

We could do with a solution to this before November 2025 to assist with Lfric git migration.

wxtim:
available to users.

Closes #6830 

**Check List**


tests/functional/cylc-combination-scripts/10-CYLC_SOURCE_DIRECTORY.t
```diff
@@ -0,0 +1,89 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#------------------------------------------------------------------------------
+# Test `cylc vip` (Validate Install Play)
+
+. "$(dirname "$0")/test_header"
+
+CYLC_ROSE=$(python -c """
```
wxtim: I'm in two minds about this?

Reasons this belongs in Cylc:
* It's testing code from Cylc.
* The Rose-stem case is just one edge case in a group and can sit with its friends.
* I only have to have one PR.

Reasons this belongs in Cylc Rose
* The test might get missed?
* Rose stem is Cylc-Rose functionality

oliver-sanders: Cylc Rose isn't installed in CI so this will fail/skip anyway.

tests/integration/test_config.py
```diff
@@ -671,3 +672,44 @@ async def test_invalid_starttask(one_conf, flow, scheduler, start):
...
+            target,
+            viewcfg={
+                'mark': False,
+                'single': False,
+                'label': False,
+                'jinja2': True,
+                'contin': True,
+                'inline': True,
+            },
+        )
```
oliver-sanders, line 699: We run a few CLI things from integration tests:
* Import the option parser function.
* Turn it into an "Options" object.
* Pull out the content of `def main` into `[async] def _main` as needed (to drop the CLI / asyncio stuff).

wxtim: I'm not sure what your point is here? Is it that my function comment has made you think that this is some sort of work-around for testing a CLI function? 

It's not - and I've renamed the function to make that clearer.

tests/functional/cylc-combination-scripts/10-CYLC_SOURCE_DIRECTORY.t
wxtim: So I should remove this? Or leave it there to show I've tested this?

oliver-sanders: I would be inclined to drop the Rose part of these tests from cylc-flow as we don't have other Rose tests in cylc-flow and this is unlikely to get run.

If there are any interactions with cylc-rose that we would like to ensure are upheld (e.g. if we wanted to make `CYLC_SOURCE_DIRECTORY` available for use in the `rose-suite.conf` file [which I don't think we do?]) then we could add tests to cylc-rose for that.

cylc/flow/parsec/fileparse.py
```diff
@@ -471,6 +471,11 @@ def read_and_proc(
     template_vars = merge_template_vars(template_vars, extra_vars)
     template_vars['CYLC_TEMPLATE_VARS'] = template_vars
 
+    # Add CYLC_SOURCE_DIRECTORY to template variables.
+    sourcepath = get_workflow_source_dir(Path(fpath).parent)[1]
+    template_vars['CYLC_SOURCE_DIRECTORY'] = (
```
oliver-sanders: There's a convention to prefix workflow variables with `CYLC_WORKFLOW` which would be worth upholding for consistency.

We have a variable called `CYLC_WORKFLOW_RUN_DIR`, perhaps this should be `CYLC_WORKFLOW_SRC_DIR` to match?

wxtim:
```suggestion
    template_vars['CYLC_WORKFLOW_SRC_DIR'] = (
```

tests/functional/cylc-combination-scripts/10-CYLC_SOURCE_DIRECTORY.t
wxtim: The original problem was related to rose stem.

changes.d/6944.feat.md
```diff
@@ -0,0 +1 @@
+Make WORKFLOW_SOURCE_DIRECTORY available as a template variable.
```
hjoliver:
```suggestion
Make CYLC_WORKFLOW_SRC_DIR available as a template variable.
```

cylc/flow/parsec/fileparse.py
```diff
@@ -471,6 +471,11 @@ def read_and_proc(
     template_vars = merge_template_vars(template_vars, extra_vars)
     template_vars['CYLC_TEMPLATE_VARS'] = template_vars
 
+    # Add CYLC_SOURCE_DIRECTORY to template variables.
```
hjoliver:
```suggestion
    # Add CYLC_WORKFLOW_SRC_DIR to template variables.
```

tests/integration/test_config.py
```diff
@@ -671,3 +672,44 @@ async def test_invalid_starttask(one_conf, flow, scheduler, start):
     with pytest.raises(InputError, match='a///b'):
         async with start(schd):
             pass
+
+
+async def test_CYLC_SOURCE_DIRECTORY_correctly_set(tmp_path, install, run_dir):
```
hjoliver:
```suggestion
async def test_CYLC_WORKFLOW_SRC_DIR_correctly_set(tmp_path, install, run_dir):
```

tests/integration/test_config.py
```diff
@@ -671,3 +672,44 @@ async def test_invalid_starttask(one_conf, flow, scheduler, start):
...
+    """CYLC_SOURCE_DIRECTORY is set correctly:
```
hjoliver:
```suggestion
    """CYLC_WORKFLOW_SRC_DIR is set correctly:
```

tests/integration/test_config.py
```diff
@@ -671,3 +672,44 @@ async def test_invalid_starttask(one_conf, flow, scheduler, start):
...
+                'label': False,
+                'jinja2': True,
+                'contin': True,
+                'inline': True,
+            },
+        )
+
+    # Setup a source directory:
+    (tmp_path / 'flow.cylc').write_text(
+        '#!jinja2\n{{ CYLC_SOURCE_DIRECTORY }}'
```
hjoliver:
```suggestion
        '#!jinja2\n{{ CYLC_WORKFLOW_SRC_DIR }}'
```

hjoliver (changes requested):
(I'm sure you were about to do the var name changes, sorry!)

For completeness, should this be added as a job environment variable too?

Obviously we don't want jobs to mess with source dirs, as a matter of principle, but other template vars are also job vars (?), and maybe a valid use case could be conjured up...

oliver-sanders:
> For completeness, should this be added as a job environment variable too?

Nooooo!

We definitely shouldn't be encouraging workflow writers to make use of resources in the workflow src dir.

hjoliver:
> Nooooo! 

🤣  - well that's why I said "probably NO", just checking.

hjoliver:
@wxtim  - you have resolved the first two of my suggestions without applying them. If that was deliberate, I'll argue the point: They're not functional code but should use the actual variable name `CYLC_WORKFLOW_SRC_DIR` or normal text "workflow source directory". Currently they use two different capitalized names: `WORKFLOW_SOURCE_DIRECTORY` (change log); and `CYLC_SOURCE_DIRECTORY` (in-code comment).

cylc/flow/parsec/fileparse.py
wxtim:
```suggestion
    # Add CYLC_WORKFLOW_SRC_DIR to template variables.
```

changes.d/6944.feat.md
wxtim:
```suggestion
Make CYLC_WORKFLOW_SRC_DIR available as a template variable.
```

wxtim:
> @wxtim - you have resolved the first two of my suggestions without applying them. If that was deliberate, I'll argue the point: They're not functional code but should use the actual variable name `CYLC_WORKFLOW_SRC_DIR` or normal text, e.g. "workflow source directory". Currently they use two different capitalized names: `WORKFLOW_SOURCE_DIRECTORY` (change log); and `CYLC_SOURCE_DIRECTORY` (in-code comment).

Not intentional - Either some sort of regression of failure of mine making the changes locally, or GH having some sort of brain fart and not applying changes I asked it to.

tests/functional/cylc-combination-scripts/10-CYLC_WORKFLOW_SRC_DIR.t
```diff
@@ -0,0 +1,60 @@
...
+    cylc vr "${WORKFLOW_NAME}"
+poll_grep "meta" "${RUN_DIR}/${WORKFLOW_NAME}/log/config/flow-processed.cylc"
+
+# It can get CYLC_WORKFLOW_SRC_DIR
+named_grep_ok "src-path-available" \
+    "$PWD" "${RUN_DIR}/${WORKFLOW_NAME}/log/config/flow-processed.cylc"
+
+purge "${WORKFLOW_NAME}"
+
+exit 0
```
hjoliver:
```suggestion
```

hjoliver (approved):

tests/functional/cylc-combination-scripts/10-CYLC_WORKFLOW_SRC_DIR.t
```diff
@@ -0,0 +1,58 @@
...
+    [[graph]]
+        R1 = foo
+__HERE__
+
+# It starts playing:
+run_ok "${TEST_NAME_BASE}-vip" \
+    cylc vip \
+        --pause \
+        --no-run-name \
+        --workflow-name "${WORKFLOW_NAME}"
```
oliver-sanders, line 47: I'm not sure we need to run the workflow for this test?

```suggestion
    cylc install \
        --no-run-name \
        --workflow-name "${WORKFLOW_NAME}"
```

tests/functional/cylc-combination-scripts/10-CYLC_WORKFLOW_SRC_DIR.t
```diff
@@ -0,0 +1,58 @@
...
+
+# It can get CYLC_WORKFLOW_SRC_DIR
+named_grep_ok "src-path-available" \
+    "$PWD" "${RUN_DIR}/${WORKFLOW_NAME}/log/config/flow-processed.cylc"
+
+# It can be updated with Cylc VR
+echo "[meta]" >> flow.cylc
+run_ok "${TEST_NAME_BASE}-vr" \
+    cylc vr "${WORKFLOW_NAME}"
```
oliver-sanders, line 56: This will restart/reload the workflow, but there is no `cylc stop` command to shut it down.

tests/functional/cylc-combination-scripts/10-CYLC_WORKFLOW_SRC_DIR.t
```diff
@@ -0,0 +1,58 @@
...
+# It starts playing:
+run_ok "${TEST_NAME_BASE}-vip" \
+    cylc vip \
+        --pause \
+        --no-run-name \
+        --workflow-name "${WORKFLOW_NAME}"
+
+# It can get CYLC_WORKFLOW_SRC_DIR
+named_grep_ok "src-path-available" \
+    "$PWD" "${RUN_DIR}/${WORKFLOW_NAME}/log/config/flow-processed.cylc"
```
oliver-sanders: Can we put the "TEST" prefix here?

```suggestion
    "TEST: $PWD" "${RUN_DIR}/${WORKFLOW_NAME}/log/config/flow-processed.cylc"
```

tests/functional/cylc-combination-scripts/10-CYLC_WORKFLOW_SRC_DIR.t
```diff
@@ -0,0 +1,58 @@
...
+
+# It can be updated with Cylc VR
+echo "[meta]" >> flow.cylc
+run_ok "${TEST_NAME_BASE}-vr" \
+    cylc vr "${WORKFLOW_NAME}"
+poll_grep "meta" "${RUN_DIR}/${WORKFLOW_NAME}/log/config/flow-processed.cylc"
+
+# It can get CYLC_WORKFLOW_SRC_DIR
+named_grep_ok "src-path-available" \
+    "$PWD" "${RUN_DIR}/${WORKFLOW_NAME}/log/config/flow-processed.cylc"
```
oliver-sanders: Can we put the "TEST" prefix here?

```suggestion
    "TEST: $PWD" "${RUN_DIR}/${WORKFLOW_NAME}/log/config/flow-processed.cylc"
```

tests/functional/cylc-combination-scripts/10-CYLC_WORKFLOW_SRC_DIR.t
wxtim: Subsequent test requires the `flow-processed.cylc` to check that the new variable is available and correct. As far as I can tell, template variables provided by Cylc (as opposed to those stored by rose-suite-cylc-install.conf) are not stored on installation.

tests/functional/cylc-combination-scripts/10-CYLC_WORKFLOW_SRC_DIR.t
wxtim: Have changed this.

tests/functional/cylc-combination-scripts/10-CYLC_WORKFLOW_SRC_DIR.t
wxtim: Added a `cylc stop` command. It speeds things up quite a lot, because `purge` has a "Wait for lsof" routine.

tests/functional/cylc-combination-scripts/10-CYLC_WORKFLOW_SRC_DIR.t
```diff
@@ -0,0 +1,60 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#------------------------------------------------------------------------------
+# Test `cylc vip` (Validate Install Play)
```
oliver-sanders:
```suggestion
# Test `CYLC_WORKFLOW_SRC_DIR.t` is set
```

tests/functional/cylc-combination-scripts/10-CYLC_WORKFLOW_SRC_DIR.t
```diff
@@ -0,0 +1,60 @@
...
+[scheduler]
+    allow implicit tasks = true
+[scheduling]
+    [[graph]]
+        R1 = foo
+__HERE__
+
+# It starts playing:
+run_ok "${TEST_NAME_BASE}-vip" \
+    cylc vip \
```
oliver-sanders, line 44: Would be good to add a `cylc view -p` test here (works when I try it manually) to make sure this variable is provided in that context too (note config parsing doesn't happen here so it's a likely breakage).

Would be something like:

```bash
run_ok ... cylc view -p
grep_ok ... "${TEST_NAME}.stdout"
```

PR #6948, opened 2025-08-28
Raise an error if template variables are set with no shebang

wxtim:
line in the `flow.cylc`.

Fixes an otherwise unreported bug.

### Real world example of bug:

```bash
#!/bin/bash
TMP=$(mktemp -d)
echo Files extracted to ${TMP}
mkdir -p "${TMP}/."

cat > "${TMP}/./flow.cylc" <<__ICI__
[scheduler]
    allow implicit tasks = true

[scheduling]
    initial cycle point = {{ STARTDATE }}T0000Z

    [[graph]]
        P1M = foo

__ICI__

cat > "${TMP}/./rose-suite.conf" <<__ICI__
[template variables]
STARTDATE="20000101"

__ICI__

cylc val ${TMP}
# .... lots of traceback
```

**Check List**


oliver-sanders (changes requested):
This change is overzealous for this example:

```
# ./flow.cylc
[scheduling]
  [[graph]]
    R1 = foo

[runtime]
  [[foo]]

# ./rose-suite.conf
[env]
FOO=bar
```

```console
$ cylc val . -s FOO=1
TemplateVarLanguageClash: No shebang line (#!jinja2) in config file.
```

But underzealous for this one 

```
# ./flow.cylc
[scheduling]
  [[graph]]
    R1 = foo

[runtime]
  [[foo]]
```

```console
$ cylc val . -s FOO=1
Valid for cylc-8.5.2.dev
```

cylc/flow/parsec/fileparse.py
```diff
@@ -114,6 +114,7 @@
     TEMPLATE_VARIABLES: {},
     TEMPLATING_DETECTED: None
 }
+JINJA2_SHEBANG = '#!jinja2'
```
hjoliver, line 117: Note we allow `#!Jinja2` (capitalized) as well... but this is just for insertion, not detection?

wxtim: Where it's used for detection it'd be used with ~`JINJA2_SHEBANG = var.lower()`~

Correction: It's a regex

wxtim:
Converted to draft.

wxtim:
I'm slightly worried that this has the potential to throw out a lot of valid workflows, just because they have a `rose-suite.conf`. Perhaps it should be a warning with a plan to make an error in future?

@oliver-sanders  what think you?

tests/functional/events/11-cycle-task-event-job-logs-retrieve/flow.cylc
```diff
@@ -1,3 +1,4 @@
+#!jinja2
```
oliver-sanders: Why?

This config doesn't use Jinja2!

cylc/flow/parsec/fileparse.py
```diff
@@ -553,6 +557,10 @@ def hashbang_and_plugin_templating_clash(
             f"A plugin set the templating engine to {templating}"
             f" which does not match {flines[0]} set in flow.cylc."
         )
+
+    # if templating and not hashbang:
+    #     raise TemplateVarLanguageClash(
+    #         f'No shebang line ({JINJA2_SHEBANG}) in config file.')
```
oliver-sanders: Comments

tests/functional/platforms/05-host-to-platform-upgrade-fail/flow.cylc
```diff
@@ -1,3 +1,4 @@
+#!jinja2
```
oliver-sanders: Workflow also doesn't require shebang.

tests/functional/remote/01-file-install.t
```diff
@@ -74,6 +75,7 @@ purge
 # Test the [scheduler]install configuration
 TEST_NAME="${TEST_NAME_BASE}-configured-paths"
 init_workflow "${TEST_NAME}" <<__FLOW_CONFIG__
+#!jinja2
```
oliver-sanders: Ditto.

tests/functional/remote/01-file-install.t
```diff
@@ -38,6 +38,7 @@ create_files () {
 # (ana, app, bin, etc, lib) are correctly installed on the remote platform.
 TEST_NAME="${TEST_NAME_BASE}-default-paths"
 init_workflow "${TEST_NAME}" <<__FLOW_CONFIG__
+#!jinja2
```
oliver-sanders: Ditto.

cylc/flow/parsec/fileparse.py
```diff
@@ -497,6 +498,9 @@ def read_and_proc(
             flines = jinja2process(
                 fpath, flines, fdir, template_vars
             )
+        elif template_vars.keys() - {"CYLC_VERSION", "CYLC_TEMPLATE_VARS"}:
```
oliver-sanders, line 501: Does cylc-rose provide default template vars even if templating was not configured?

wxtim: If you turn Cylc Rose on it provides template variables. You turn it on by adding a blank `rose-suite.conf`.

tests/functional/events/11-cycle-task-event-job-logs-retrieve/flow.cylc
wxtim: But template variables are supplied - looks like there's vestigial garbage in the test.

cylc/flow/parsec/fileparse.py
oliver-sanders: So adding a `rose-suite.conf` file will force the user to add a Jinja2 shebang line, even if they are not using Jinja2 in the config?

wxtim: I don't think there's a simple/sane* way around that. If you think that's a deal breaker IMO we just call this PR off. It's too small an issue for me to spend much more time on it.

\* Non simple ways
* with supress: from cylc.rose.utilites import STANDARD_VARS (less general change, but I don't like adding a single plugin exception)
* Modify the plugin interface (more general, but I don't want to change the interface!)

oliver-sanders: > I don't think there's a simple/sane* way around that.

I think forcing people to add Jinja2 shebangs where they aren't needed is going to be more of a problem than failing to spot when shebangs aren't needed but should have been added.

wxtim:
Solutions worse than problem. Bug recorded at https://github.com/cylc/cylc-flow/issues/6971

PR #6954, opened 2025-09-01
Fix reinstall test failing on MacOS due to rsync change

MetRonnie:
For some reason on GH Actions MacOS we have started seeing "recv" instead of "send" in rsync here:
```
REINSTALL cit-20250901T091622Z/test_reinstall/test_permissions_change/52f05762 from /private/var/folders/x7/ch5v91h56_zbvbd1y2f600dm0000gn/T/pytest-of-runner/pytest-1/popen-gw4/test_permissions_change0/src
.f...p... recv myscript

\x1b[38;5;248mTIP: You can "exclude" files/dirs to prevent Cylc from installing or overwriting
     them by adding them to the .cylcignore file. See cylc reinstall --help.\x1b[0m

\x1b[1mReinstall would make the above changes.\x1b[0m
\x1b[35mNo changes made: reinstall cancelled.\x1b[0m
```

Possibly macos-latest has changed to 15

**Check List**

- [x] No dependency changes or tests or changelog or docs needed

### DIFF
--- a/cylc/flow/scripts/reinstall.py
+++ b/cylc/flow/scripts/reinstall.py
@@ -348,6 +348,7 @@ def format_reinstall_output(out: str) -> List[str]:
             # Skip this line as nothing will happen to this dir.
             continue
         match = re.match(r'^(.{11}) (send|del\.|recv) (.*)$', line)
+        # (On some systems, may get "recv" instead of "send")
         if match:
             summary, operation, file = match.groups()
             color = 'green' if operation == 'send' else 'red'


cylc/flow/scripts/reinstall.py
```diff
@@ -347,7 +347,7 @@ def format_reinstall_output(out: str) -> List[str]:
             # file is present.
             # Skip this line as nothing will happen to this dir.
             continue
-        match = re.match(r'^(.{11}) (send|del\.) (.*)$', line)
+        match = re.match(r'^(.{11}) (send|del\.|recv) (.*)$', line)
```
hjoliver, line 350: Worth a short comment here to explain?

MetRonnie:
```suggestion
        match = re.match(r'^(.{11}) (send|del\.|recv) (.*)$', line)
        # (On some systems, may get "recv" instead of "send")
```
?

PR #6959, opened 2025-09-03
Fix docker FROM AS/as error.

hjoliver:
After swarm build failed in CI (for other reasons - slow download?):

```
 135.6 critical libmamba Could not solve for environment specs
------

 1 warning found (use docker --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 24)
Dockerfile:8
---------------
```

(Does that really matter? 🤣 )


oliver-sanders (approved):
I think this is just lint, but whatever.

hjoliver:
Yes it was just a warning (I said the build failed for other reasons).

Issue #6402, opened 2024-10-11
broadcast: operate on broadcasts by ID


oliver-sanders:
(Something we have previously talked about but do not presently have an issue for).

At present, we issue a broadcast and set its scope using the namespace (i.e. task/family) and point (i.e. cycle) options.

We can revoke broadcasts by "clearing" them, however, this interface is a little peculiar:
* To clear a broadcast, you must issue a "clear" operations with the exact same scope as the "set" operation you want to clear.
* This will clear all broadcasts with the same scope (i.e. there may not be a way to target a single broadcast to clear).

> Active broadcasts can be revoked using the "clear" mode. Any broadcasts
> matching the specified cycle points and namespaces will be revoked.
>
> Note: a "clear" broadcast for a specific cycle or namespace does *not* clear
> all-cycle or all-namespace broadcasts.
>
> cylc broadcast --help

This makes broadcast clearing rather awkward. When we [expose broadcasts in the GUI](https://github.com/cylc/cylc-ui/issues/463), it would be good to open up the ability to just click on a broadcast and press clear.

Additionally, it might be good to carry some form of metadata with a broadcast, e.g. who issued this broadcast, or a one-line description of the intention (might be useful for broadcasts made by automated scripts). Note I think we might already store broadcast-time as a metadata field.

Suggestions:

1. Assign each broadcast a unique ID (suggest a simple incremented integer, return this on the CLI when a broadcast is made).
2. Allow broadcasts to be cleared by ID.
3. Allow additional metadata to be stored along with the broadcast.
4. Be very careful to avoid breaking changes, there are a lot of scripts out there that are built into the behaviour of broadcast which we left unchanged between Cylc 7 & 8 (on purpose).

ColemanTom:
Hah, I was just about to ask if there was a way to clear all broadcasts for a workflow and/or a cycle. Glad to see it being thought about.

hjoliver:
>   To clear a broadcast, you must issue a "clear" operations with the exact same scope as the "set" operation you want to clear.
 >  This will clear all broadcasts with the same scope (i.e. there may not be a way to target a single broadcast to clear).

It's been a while, but I think you can always clear broadcasts with the same specificity as used when setting them (e.g. an individual environment variable). But if you only specify the scope to clear, not the specifics, that may of course clear multiple broadcasts at once. (Or have I misunderstood you?)

The ID is a good idea, so that we don't have to repeat the original scope information when clearing a broadcast.

[...]

Issue #6308, opened 2024-08-20
Modify runtime of a succeeded task so you can trigger it


ColemanTom:
Discussion took place in https://cylc.discourse.group/t/cylc-broadcast-is-being-cleared-automatically/991/3

I'm posting this here to hope it doesn't get lost/forgotten.

Two of us encountered this same issue today. When you do a cylc broadcast on a task which has already run and succeeded, the broadcast is automatically cleared.

I was using CYLC_VERSION=8.3.1 and my colleague using CYLC_VERSION=8.3.3. We both first noticed it via the WUI but we have shown it also happens via CLI. I don’t know the behaviour in 8.2.

From the scheduler log:

```
2024-07-31T02:00:40Z INFO - Command "broadcast" received.
    broadcast(cycle_points=['20240327T0000Z'], mode=put_broadcast, namespaces=['ct_to_simpler'], settings=[{'environment': {'FORCE_RESEND': 'no'}}])
2024-07-31T02:00:40Z INFO - Broadcast set:
    + [20240327T0000Z/ct_to_simpler] [environment]FORCE_RESEND=no
2024-07-31T02:00:40Z INFO - Broadcast cancelled:
    - [20240327T0000Z/ct_to_simpler] [environment]FORCE_RESEND=no
```

Pausing the workflow, triggering the task, and then broadcasting allowed the broadcast to be kept.

```
2024-07-31T02:04:48Z INFO - Pausing the workflow
2024-07-31T02:04:48Z INFO - Command "pause" actioned. ID=0ea0b0b8-5b94-4d06-b90b-81b43962e8a9
2024-07-31T02:05:06Z INFO - Command "force_trigger_tasks" received. ID=8055b39c-10e3-4ccf-97ac-ed3a9bdd5d17
    force_trigger_tasks(flow=['all'], flow_wait=False, tasks=['20240327T0000Z/ct_to_simpler'])
2024-07-31T02:05:07Z INFO - [20240327T0000Z/ct_to_simpler:waiting(runahead)] => waiting
2024-07-31T02:05:07Z INFO - [20240327T0000Z/ct_to_simpler:waiting] => waiting(queued)
2024-07-31T02:05:07Z INFO - Command "force_trigger_tasks" actioned. ID=8055b39c-10e3-4ccf-97ac-ed3a9bdd5d17
2024-07-31T02:06:07Z INFO - Command "broadcast" received.
[...]

PR #6964, opened 2025-09-03
broadcast: reduce the broadcast expire limit

oliver-sanders:
* Partially addresses #6308
* Broadcasts for a cycle are expired when there are no more active tasks in the cycle.
* At Cylc 7, there was (almost) always a succeeded task in the pool one cycle behind each active task.
* So the broadcast expire limit was the oldest cycle with active tasks *minus* the duration of the longest cycling sequence.
* This commit restores the Cylc 7 broadcast expire limit which ensures that the broadcasts for the previous instance of any given task are preserved.
* This better satisfies the use case of winding back a workflow to the previous cycle (i.e. to retrigger a task, reflow from a task or warm restart).

### The originally proposed solution

The solution suggested in #6308 was:
* Don't expire broadcasts automatically.
* Maintain a window of broadcasts in memory.
* Go to the DB to retrieve broadcasts when this window changed.

Unfortunately, this turned out to be a right pig to implement. Unfortunately, because broadcasts are managed from the server thread, it is not possible to go to the DB. Working around this problem is a major refactor on top of the work itself, which was also a major refactor.

### The newly proposed solution

This is essentially a Cylc 7 -> 8 migration issue. Although it doesn't make sense to auto-expire broadcasts (except from a memory management perspective) this expiry didn't cause any issues before.

This PR reduces the broadcast expire limit back to more-or-less what it was at Cylc 7 resolving the the "issue" part of the problem.

Removing auto-expiry may still be a mid/long term goal, but this would need to be done with a large scale refactor of the broadcast code. The broadcast code has a few issues, so a refactor may well be in order:

* The CLI does validation (and upgrading) client side (fingers crossed the client is at the same version!), the GUI gets no validation or upgrading whatsoever (#6429).
* Broadcasts are performed in the server thread, but I don't think there's a good reason for this.
* Broadcasts can only be cleared/expired by reproducing the exact settings they were issued with (#6402).

So we can always come back to this.

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at https://github.com/cylc/cylc-doc/pull/871.

### DIFF
--- a/cylc/flow/config.py
+++ b/cylc/flow/config.py
@@ -2346,6 +2346,23 @@ class WorkflowConfig:
             # dependencies are checked in generate_triggers:
             self.check_terminal_outputs(parser.terminals)
 
+        # set of all cycling intervals containined within the workflow
+        cycling_intervals = {
+            sequence.get_interval()
+            for sequence in self.sequences
+        } | {
+            # add a null interval to handle async workflows
+            get_interval_cls().get_null()
+        }
+
+        # determine the longest cycling interval in the workflow
+        self.interval_of_longest_sequence: 'IntervalBase' = max({
+            interval for interval in cycling_intervals
+            # NOTE: None type sorts above a null interval for strange
+            # historical reasons so must be filtered out
+            if interval is not None
+        })
+
         self.set_required_outputs(task_output_opt)
 
         # Detect use of xtrigger names with '@' prefix (creates a task).
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -58,6 +58,7 @@ from typing import (
 )
 from uuid import uuid4
 
+from metomi.isodatetime.exceptions import TimePointDumperBoundsError
 import psutil
 
 from cylc.flow import (
@@ -109,8 +110,8 @@ from cylc.flow.loggingutil import (
 from cylc.flow.network import API
 from cylc.flow.network.authentication import key_housekeeping
 from cylc.flow.network.server import WorkflowRuntimeServer
-from cylc.flow.parsec.OrderedDict import DictTree
 from cylc.flow.parsec.exceptions import ParsecError
+from cylc.flow.parsec.OrderedDict import DictTree
 from cylc.flow.parsec.validate import DurationFloat
 from cylc.flow.pathutil import (
     get_workflow_name_from_id,
@@ -1680,7 +1681,18 @@ class Scheduler:
             # A simulated task state change occurred.
             self.reset_inactivity_timer()
 
-        self.broadcast_mgr.expire_broadcast(self.pool.get_min_point())
+        # auto expire broadcasts
+        with suppress(TimePointDumperBoundsError):
+            # NOTE: TimePointDumperBoundsError will be raised for negative
+            # cycle points, we skip broadcast expiry in this circumstance
+            # (pre-initial condition)
+            if min_point := self.pool.get_min_point():
+                # NOTE: the broadcast expire limit is the oldest active cycle
+                # MINUS the longest cycling interval
+                self.broadcast_mgr.expire_broadcast(
+                    min_point - self.config.interval_of_longest_sequence
+                )
+
         self.late_tasks_check()
 
         self.process_queued_task_messages()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,7 +18,7 @@ import logging
 import re
 from pathlib import Path
 from shutil import rmtree
-from typing import List, Optional, Tuple
+from typing import Callable, List, Optional, Tuple
 
 import pytest
 
@@ -175,9 +175,11 @@ def capcall(monkeypatch):
         function_string:
             The function to replace as it would be specified to
             monkeypatch.setattr.
-        substitute_function:
-            An optional function to replace it with, otherwise the captured
-            function will return None.
+        mock:
+            * If True, the function will be replaced by a "return None".
+            * If False, the oringal function will be run.
+            * If a Callable is provided, this will be run in place of the
+              original function.
 
     Returns:
         [(args: Tuple, kwargs: Dict), ...]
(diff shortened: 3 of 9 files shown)


cylc/flow/config.py
```diff
@@ -2346,6 +2346,10 @@ def load_graph(self):
             # dependencies are checked in generate_triggers:
             self.check_terminal_outputs(parser.terminals)
 
+        self.interval_of_longest_sequence = max(
+            {sequence.get_interval() for sequence in self.sequences}
+        )
```
hjoliver: This is killing the tests - it is `None` for non-cycling workflows.

hjoliver:
Seems reasonable. 

- can you put the rationale in code comments?
- and copy your broadcast refactor notes to a new issue?
- we probably need to document the broadcast expire limit for users

oliver-sanders:
> and copy your broadcast refactor notes to a new issue?

https://github.com/cylc/cylc-flow/issues/6402#issuecomment-3253477095

tests/functional/broadcast/00-simple.t
oliver-sanders, line 1: This test has changed because a cycle which was expired under the old logic, is not expired under the new logic.

Explanation:

Before this change, `expire_broadcast` was called for cycles:
* 201008**08**T00 (i.e. expire broadcasts for cycle **07** and older)
* 201008**09**T00 (i.e. expire broadcasts for  **08** and older)

After this change, broadcasts expired for cycles before:
* 201008**07**T00 (i.e. expire broadcasts for  **06** and older)
* 201008**08**T00 (i.e. expire broadcasts for  **07** and older)

Due the the broadcast expire limit offset of `P1D`.

The `BCAST|FOO` broadcast which is made FROM the **08** cycle, targetting the **08** cycle. Previously, this broadcast was expired at the end of the second cycle. After this change it would be expired after the end of the third cycle (which never runs because of the final cycle point).

> [!NOTE]
> The `expire_broadcast` interface, expires broadcasts before but not including the specified cutoff cycle.

> [!NOTE]
> These KGO lines have not changed since Cylc 7! In this case, the change actually makes the broadcast window larger than Cylc 7 because there are no inter-cycle dependencies in this workflow, so older cycles were not kept open by succeeded tasks at Cylc 7. 
>
> This change ensures you can always rerun the previous cycle, with Cylc 7 this was often but not always the case.

tests/flakyfunctional/xtriggers/01-workflow_state.t
oliver-sanders, line 1: Before this change the 2015 cycle expired just before the scheduler shut down.

With this change that has been pushed back to the 2014 cycle.

The exact cycle being tested doesn't matter here (the final cycle - 2016 - matters to a test above, but not this one), so I have bumped these tests back to the 2014 cycle so that they continue to cover both broadcasting and broadcast expiring.

cylc/flow/scheduler.py
```diff
@@ -1674,7 +1675,18 @@ async def _main_loop(self) -> None:
             # A simulated task state change occurred.
             self.reset_inactivity_timer()
 
-        self.broadcast_mgr.expire_broadcast(self.pool.get_min_point())
+        # auto expire broadcasts
+        with suppress(TimePointDumperBoundsError):
+            # NOTE: TimePointDumperBoundsError will be raised for negitive
```
hjoliver:
```suggestion
            # NOTE: TimePointDumperBoundsError will be raised for negative
```

cylc/flow/scheduler.py
```diff
@@ -1674,7 +1675,18 @@ async def _main_loop(self) -> None:
...
+            # cycle points, we skip broadcast expiry in this circumstance
+            # (pre-initial condition)
+            if min_point := self.pool.get_min_point():
+                # NOTE: the broadcast expire limit is the oldest ascive cycle
```
hjoliver:
```suggestion
                # NOTE: the broadcast expire limit is the oldest active cycle
```

hjoliver (approved):
Looks good.

cylc/flow/config.py
```diff
@@ -2346,6 +2346,23 @@ def load_graph(self):
...
+            get_interval_cls().get_null()
+        }
+
+        # determine the longest cycling interval in the workflow
+        self.interval_of_longest_sequence: 'IntervalBase' = max({
+            interval for interval in cycling_intervals
+            # NOTE: None type sorts above a null interval for strange
+            # historical reasons so must be filtered out
+            if interval is not None
+        })
```
wxtim, line 2364: Is the implicit set creation needed?

cylc/flow/scheduler.py
```diff
@@ -1680,7 +1681,18 @@ async def _main_loop(self) -> None:
             # A simulated task state change occurred.
             self.reset_inactivity_timer()
 
-        self.broadcast_mgr.expire_broadcast(self.pool.get_min_point())
+        # auto expire broadcasts
+        with suppress(TimePointDumperBoundsError):
+            # NOTE: TimePointDumperBoundsError will be raised for negative
+            # cycle points, we skip broadcast expiry in this circumstance
+            # (pre-initial condition)
+            if min_point := self.pool.get_min_point():
```
wxtim, line 1689: WalrusWalrusWavingGIF

tests/conftest.py
```diff
@@ -175,9 +175,11 @@ def capcall(monkeypatch):
         function_string:
             The function to replace as it would be specified to
             monkeypatch.setattr.
-        substitute_function:
-            An optional function to replace it with, otherwise the captured
-            function will return None.
+        mock:
+            * If True, the function will be replaced by a "return None".
+            * If False, the oringal function will be run.
```
wxtim:
```suggestion
            * If False, the original function will be run.
```

cylc/flow/config.py
oliver-sanders: Good point, technically no, though it might be faster?

I suspect that `__hash__` comparisons are probably faster than `IntervalBase._cmp` comparisons, so filtering out duplicates (of which there will likely be many, e.g. `T00` and `T01` have the same interval of `P1D`) would make the `max` faster.

Haven't tested and speed isn't important here. Don't much mind either way.

cylc/flow/scheduler.py
oliver-sanders: Oh hai walrus

tests/integration/test_broadcast_mgr.py
```diff
@@ -90,3 +95,120 @@ async def test_reject_valid_broadcast_is_remote_clash_with_broadcast(
...
+
+    The broadcast expiry limit is the oldest active cycle MINUS the longest
+    cycling sequence.
+
+    See https://github.com/cylc/cylc-flow/pull/6964
+    """
+    # capture broadcast expiry calls
+    _expires = capcall('cylc.flow.broadcast_mgr.BroadcastMgr.expire_broadcast')
+
+    def expires():
```
wxtim, line 122: expired? (IMO tense of function name and docstring matching make it easire to understand what's going on.

oliver-sanders: It returns a list (plural) of expiry events.

cylc/flow/config.py
wxtim: > Don't much mind either way.

Stet (let it stand)

wxtim (approved):

hjoliver:
Note also #6976

Issue #6904, opened 2025-08-07
protobuf: evaluate sensible version range


oliver-sanders:
We pin the version of protobuf to ensure that the "protos" we compile are compatible with the runtime they are used in.

(note that as a distributed system, it's possible that different components of a Cylc system may be using different protobuf versions)

There is now an authoritative statement on cross version compatibility:

https://protobuf.dev/support/cross-version-runtime-guarantee/#major

https://protobuf.dev/support/cross-version-runtime-guarantee/#:~:text=minor%2C%20or%20micro).-,Python%2Dspecific%20Guarantees,-Since%20the%203.20.0

Given this, we may be able to loosen the version range.

Someone needs to carefully read through the content on this page and work out what a sensible range would be for us going forward.

PR #6972, opened 2025-09-05
Update protobuf version pin

### LINKED ISSUE #6885: dependencies: bump protobuf
We are behind on Protobuf releases.

* Upgrade to permit use of latest releases.
* Determine whether the version pin can be relaxed (e.g. could we do `4.*` rather than `4.24.*`).

wxtim:
Partially closes #6885 - Upgrades Protobuf to 

**Check List**


cylc/flow/data_messages_pb2.py
```diff
@@ -1,12 +1,22 @@
 # -*- coding: utf-8 -*-
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
+# NO CHECKED-IN PROTOBUF GENCODE
 # source: data_messages.proto
-# Protobuf Python Version: 4.25.3
+# Protobuf Python Version: 5.28.0
```
MetRonnie: Not regenerated since upping to 6.32?

setup.cfg
```diff
@@ -67,7 +67,7 @@ install_requires =
     metomi-isodatetime>=1!3.0.0,<1!3.2.0
     # Constrain protobuf version for compatible Scheduler-UIS comms across hosts
     packaging
-    protobuf>=4.24.4,<4.25.0
+    protobuf==6.32.0
```
MetRonnie: Based on https://protobuf.dev/support/cross-version-runtime-guarantee/#python I think we can relax this to
```suggestion
    protobuf>=4.25,<8
```

At the very least,

```suggestion
    protobuf==6.32.*
```

cylc/flow/data_messages_pb2.py
wxtim: Yes. It is - this is what I was curious about!

```console
> pip list | grep protobuf
protobuf                       6.32.0
types-protobuf                 5.29.1.20250208
> cd cylc/flow/
> protoc -I=./ --python_out=./ --pyi_out=./ data_messages.proto
> git st
> git diff


```

wxtim: I think that this is correct, however.

oliver-sanders: If the support bracket is as stated, then this file should not need to be recompiled for this change.

And if we were to recompile this file, it should be done using the oldest supported version:

> gencode should never be paired with a runtime that predates the protoc and plugin that was used to generate those bindings.
>
> -- https://protobuf.dev/support/cross-version-runtime-guarantee/#backwards

setup.cfg
oliver-sanders: I *think* that's correct, but was reluctant to make this change at the time without more info. I find the v8 choice rather arbitrary (will they really stick to that?).

oliver-sanders: (Unlocking to v8 would close #6904, good to do, but we don't *need* to close it right now)

cylc/flow/data_messages_pb2.py
```diff
@@ -1,12 +1,22 @@
...
 from google.protobuf import symbol_database as _symbol_database
 from google.protobuf.internal import builder as _builder
+_runtime_version.ValidateProtobufRuntimeVersion(
+    _runtime_version.Domain.PUBLIC,
+    5,
+    28,
+    0,
+    '',
+    'data_messages.proto'
+)
```
MetRonnie: Could this be the "poison pill" mentioned in https://protobuf.dev/support/cross-version-runtime-guarantee/#backwards?

wxtim: It already leaves that info in the header, so it'd be available, but I guess this would make it easier to run checks.

oliver-sanders: (resolving this as the changes have been reverted)

cylc/flow/data_messages_pb2.py
oliver-sanders: (resolving as the changes have been reverted)

dwsutherland:
> Code generated for a major version V (full version: V.x.y) will be supported by protobuf runtimes of version V and V+1.

This is the safe general rule.. 

> Python may break generated code compatibility in specific future major version releases, but it will be coupled with poison pill warnings and errors in advance. As of 6.32.0, all generated code since 3.20.0 will be supported until at least 8.x.y.

This is the python exception to the rule.. as shown [here](https://protobuf.dev/support/version-support/):
image


The python runtime only supports Python 3.9+ from v6.x, i.e.:
https://pypi.org/project/protobuf/6.32.1/#files

And from grok:
image


Our generated code is from `4.25.3` but that shouldn't be an issue with python version..

However, if we want to constrain the runtime to the same as `cylc-flow` then we need a runtime compatible with 3.12? which would be 5+ (i.e. 6.32.1).. 
Although I assume pip or w/e package manager will find the appropriate runtime version according to the python as long as our runtime range covers the python so we definitely need i.e.`protobuf>=4.25,<7`...

setup.cfg
```diff
@@ -67,7 +67,7 @@ install_requires =
     metomi-isodatetime>=1!3.0.0,<1!3.2.0
     # Constrain protobuf version for compatible Scheduler-UIS comms across hosts
     packaging
-    protobuf>=4.24.4,<4.25.0
+    protobuf>=4.25,<8
```
MetRonnie: As v4 is out of support, I suggest we up the lower end of this so that we are better placed in future for using a newer gencode version
```suggestion
    protobuf>=5,<8
```

dwsutherland: As long as the package manager is smart enough, it will be ok? (as it should install a corresponding version of protobuf?)

No harm in going to 5 though

MetRonnie: > as it don't work with Python 3.12

Wait, it's Python 3.13 actually (and only on conda-forge). I've rewritten the comment to avoid confusion

MetRonnie (approved):

PR #6974, opened 2025-09-07
Remove obsolete trigger --on-resume option.

### LINKED ISSUE #6675: trigger: remove --on-resume option
The `--on-resume` option was added as a temporary bodge to ensure that a certain use case would continue to function after the "triggering a task whilst the workflow is paused causes it to run instantly" change.

This option is scheduled for demolition at 8.5.0.

hjoliver:
Close #6675

**Check List**


oliver-sanders:
> Not ideal, but all software has its limitations :-)

Err, didn't you add the `--on-resume` option specifically to work around this "limitation"?!

> Don't expire while paused seems a good idea though.

:+1:, will document.

hjoliver:
>  >  Not ideal, but all software has its limitations :-)
> 
> Err, didn't you add the --on-resume option specifically to work around this "limitation"?!

Yes, but your recent broadcast expiry change means broadcasts now last at least as long as they did in Cylc 7, which makes the option much less important - it only matters for tasks beyond the expiry limit (and you seemed very keen to get rid of the option, which I had decided not to argue against).

tests/integration/test_remove.py
```diff
@@ -538,13 +538,12 @@ async def test_remove_triggered(flow, scheduler, start):
 
         # trigger bar (on resume)
```
oliver-sanders:
```suggestion
        # trigger bar
```

cylc/flow/network/schema.py
```diff
@@ -2221,13 +2221,7 @@ class Meta:
         resolver = partial(mutator, command='force_trigger_tasks')
 
     class Arguments(TaskMutation.Arguments, FlowMutationArguments):
-        on_resume = Boolean(
-            default_value=False,
-            description=sstrip('''
-                If the workflow is paused, wait until it is resumed before
-                running the triggered task(s).
-            ''')
-        )
```
oliver-sanders: This may create an inter-version compatibility issue.

Will test. @wxtim, could you check forward/backward compatibility for this too. Old/new GUI, old/new scheduler, old/new CLI.

I think we might be ok as we made a change so as not to populate arguments which aren't being changed from the default.

oliver-sanders (commented):
LGTM.

* [x] Use case covered by https://github.com/cylc/cylc-flow/pull/6976 (approved)
* [x] Legacy code removed.
* [x] Tested inter-version compat.
* [x] Tests pass (branch probs just needs a rebase).

cylc/flow/network/schema.py
oliver-sanders: @hjoliver, dammit, I think I've caught an issue between a scheduler run from this branch and the CLI run from 8.5.x.

```console
$ CYLC_VERSION=8.6.0.dev cylc vip -n new
...

$ CYLC_VERSION=8.5.1 cylc trigger new//1/a
new: RequestError: Unknown argument 'onResume' on field 'Mutations.trigger'.
    (Workflow is running in Cylc 8.6.0.dev)
```

The best option to avoid issues is likely to leave this GraphQL option as it was, but update the description to say that it is deprecated and no longer in use. Add a back-compat comment to remind us to remove this in a future version.

hjoliver:
> >   which I had decided not to argue against
>
> Quick refresh on the context:
>
> > OPTION: --on-resume - run triggered tasks only when the paused workflow is resumed

Not sure what your point is here!

You convinced us of the need for a compromise on the broadcast fix that would allow this option to be removed, and then reminded me to remove it - so I figured you wanted it removed despite the fact that the compromise obviously meant that broadcasts beyond the expiry limit would (by definition) not stick. 

I had merely decided that was OK because the compromise was (a) pragmatically necessary; and (b) good enough for most use cases, in the context of a strong push against keeping the option (or even adding in the first place) from you.

cylc/flow/network/schema.py
hjoliver: (Yeah, I should've done this anyway).

hjoliver: Done - bit of a rush job at the end of the day, but hopefully all good.

hjoliver:
> @hjoliver I'm afraid lots of tests are failing.

Gah, see "bit of a rush job at the end of the day" 😮‍💨    - fixed

oliver-sanders:
@wxtim, the inter-version compat issue should be resolved, worth testing.

oliver-sanders:
As outlined above, I'm leaving it to you, [b] would be "good", but that doesn't overrule [a], that I think this is good enough.

PR #6976, opened 2025-09-08
broadcast: suspend auto-expiry whilst scheduler paused

oliver-sanders:
Relates to: https://github.com/cylc/cylc-ui/pull/2299

We have a functionality gap for the case where we want to broadcast to a historical cycle *before* re-triggering it because the broadcast will be expired before the cycle can be triggered.

The `--on-resume` option handled this, so we need a workaround before removing it.

Minor change: don't auto-expire broadcasts whilst the scheduler is paused.

**Check List**


cylc/flow/scheduler.py
```diff
@@ -1682,16 +1682,18 @@ async def _main_loop(self) -> None:
...
-            # cycle points, we skip broadcast expiry in this circumstance
-            # (pre-initial condition)
-            if min_point := self.pool.get_min_point():
-                # NOTE: the broadcast expire limit is the oldest active cycle
-                # MINUS the longest cycling interval
-                self.broadcast_mgr.expire_broadcast(
-                    min_point - self.config.interval_of_longest_sequence
-                )
+        if not self.is_paused:
+            # NOTE: Don't auto-expire broadcasts whilst the scheduler is paused
```
hjoliver:
```suggestion
            # NOTE: Don't auto-expire broadcasts whilst the scheduler is paused.
            # This allows broadcast-and-trigger beyond the expiry limit, by
            # pausing before doing it (after which the expiry limit moves back).
```

hjoliver: (We could document this as a reasonable rationale: past broadcasts have to expire for memory management reasons, but you can work around this limitation by pausing the workflow before broadcast and trigger, because expiry is suspended while the workflow is paused).

hjoliver: (It's good to put a reason for statements like this, in case we forget why and think it's not important during some future refactor).

MetRonnie (approved):
One code comment suggestion from Hilary still unresolved

PR #6986, opened 2025-09-17
doc: fix missing term

oliver-sanders:
The "optional" term has been removed, changed to "optional output".


wxtim:
I believe that there is a second issue

```
/home/users/tim.pillinger/repos/cylc-doc/src/reference/config/workflow.rst:369: WARNING: Inline interpreted text or phrase reference start-string without end-string.

```

wxtim (approved):
- [x] I have built the docs locally.

wxtim:
I'm happy that this can be merged on 1 review if you want

PR #6990, opened 2025-09-18
Fix erroneous submissions to localhost when using deprecated `batch system` setting

wxtim:
Fixes a bug not otherwise described. 

Bug introduced in #6836 . 

## Description of bug

Before #6836 hostname was (unwisely) cached on the rtconfig object and hence passed (evaluated) to the platform selection routine, but inside the following logic:

```
            elif (
                host_name is None
                and rtconfig['platform']
                and rtconfig['platform'] != platform_name
            ):
                rtconfig['platform'] = platform
```

[see diff](https://github.com/cylc/cylc-flow/pull/6836/files#r2360012173)

so that platform remained unset if `rtconfig['platform']` had not been set in the config then it wouldn't be updated and `get_platform` would still see this config as a host/batch system format one.

### Example of bug

We have a platform where jobs are submitted to Slurm from localhost. The user set up the following valid config.

```cylc
# flow.cylc
[scheduling]
    cycling mode = integer
    [[graph]]
        R1 = foo

[runtime]
    [[foo]]
        [[[job]]]
            batch system = slurm
```

The task will run on background.

**Check List**

### DIFF
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -1226,9 +1226,9 @@ class TaskJobManager:
             # We need to assume that we want a host if any of the items
             # for the old host/batch system plaform selection system are set.
             if any(
-                rtconfig[section][key[0]] is not None
+                rtconfig[section][key] is not None
                 for section, values in FORBIDDEN_WITH_PLATFORM.items()
-                for key in values.items()
+                for key in values
             ):
                 host_name = self.task_remote_mgr.eval_host(
                     rtconfig['remote']['host']


wxtim:
@MetRonnie @oliver-sanders @hjoliver Please add yourselves to review...

cylc/flow/task_job_mgr.py
```diff
@@ -1222,14 +1223,21 @@ def _prep_submit_task_job(
 
         host_name, platform_name = None, None
         try:
-            if rtconfig['remote']['host'] is not None:
+            # We need to assume that we want a host if any of the items
+            # for the old host/batch system plaform selection system are set.
+            if any(
+                rtconfig[section][key[0]] is not None
+                for section, values in FORBIDDEN_WITH_PLATFORM.items()
```
oliver-sanders, line 1230: I understand checking `[job]batch system`, but I'm not sure whether `[host]retrieve job logs` should influence platform selection?

For reference:

```python
FORBIDDEN_WITH_PLATFORM: Dict[str, Dict[str, Set[Optional[str]]]] = {
    'remote': {
        # setting: exclusions
        'host': {'localhost', None},
        'retrieve job logs': {None},
        'retrieve job logs max size': {None},
        'retrieve job logs retry delays': {None},
    },
    'job': {
        'batch system': {None},
        'batch submit command template': {None}
    }
}
```

oliver-sanders (commented):
Confirmed that the test fails before and passes after.

Tried out the reported example and confirmed it works now.

To reproduce, try something like this (note you'll need to have a compatible platform defined):

```cylc
[scheduling]
    [[graph]]
        R1 = foo

[runtime]
    [[foo]]
        [[[job]]]
            batch system = slurm
```

cylc/flow/task_job_mgr.py
MetRonnie: If `[host]retrieve job logs` is set we can safely assume the workflow must not be using the Cylc 8 platforms syntax, as it would fail validation. So this seems fine to me

cylc/flow/task_job_mgr.py
```diff
@@ -1222,14 +1223,21 @@ def _prep_submit_task_job(
 
         host_name, platform_name = None, None
         try:
-            if rtconfig['remote']['host'] is not None:
+            # We need to assume that we want a host if any of the items
+            # for the old host/batch system plaform selection system are set.
+            if any(
+                rtconfig[section][key[0]] is not None
```
MetRonnie:
```suggestion
                rtconfig[section][key] is not None
```

cylc/flow/task_job_mgr.py
```diff
@@ -1222,14 +1223,21 @@ def _prep_submit_task_job(
...
+                for section, values in FORBIDDEN_WITH_PLATFORM.items()
+                for key in values.items()
```
MetRonnie:
```suggestion
                for key in values
```

MetRonnie:
(I noticed the type annotations for `eval_host` are too strict)
```diff
diff --git a/cylc/flow/task_remote_mgr.py b/cylc/flow/task_remote_mgr.py
index d2a8a9160..ffad00573 100644
--- a/cylc/flow/task_remote_mgr.py
+++ b/cylc/flow/task_remote_mgr.py
@@ -122,3 +122,3 @@ class TaskRemoteMgr:
     def _subshell_eval(
-        self, eval_str: str, command_pattern: re.Pattern
+        self, eval_str: Optional[str], command_pattern: re.Pattern
     ) -> Optional[str]:
@@ -177,3 +177,3 @@ class TaskRemoteMgr:
         #     Cylc8.x
-    def eval_host(self, host_str: str) -> Optional[str]:
+    def eval_host(self, host_str: Optional[str]) -> Optional[str]:
         """Evaluate a host from a possible subshell string.
@@ -193,3 +193,3 @@ class TaskRemoteMgr:
 
-    def eval_platform(self, platform_str: str) -> Optional[str]:
+    def eval_platform(self, platform_str: Optional[str]) -> Optional[str]:
         """Evaluate a platform from a possible subshell string.
```

oliver-sanders (approved):
LGTM, tested as working.

cylc/flow/task_job_mgr.py
wxtim: What Ronnie says - this is imported from `platforms.py` and is the canonical list of things which we assume are a mistake if used in combination with platforms: This is the same list as used by validation.

PR #6999, opened 2025-09-20
Don't infer optionality from RHS implicit outputs.

### LINKED ISSUE #6744: Tweak output optionality inference to avoid confusing syntax.
(Pasted from Element chat; this proposes a small non-breaking tweak to the status quo; not worth a full proposal on cylc-admin).

------

## Summary

I've had negative feedback from users on the need for flagging optional outputs on the right side of dependency pairs.
- the syntax is confusing 
- a single optional success/fail branch can require adding right-side `?`s in many places, which is annoying

## Background

Intuitively, a dependency pair describes how a (right) task triggers off of a (left) task output:
```
# OK, makes sense:
a => b
a:succeeded => b
a:x => b
```

Flagging optional outputs on the left is also intuitive, because an optional left-side output effectively makes the dependency itself optional (so the syntax works from both perspectives):

```
# OK, makes sense:
a? = > b
a:succeeded? = > b
a:x? = > b
```

But inferring right-side output optionality from dependencies is not intuitive, because right-side outputs have nothing to do with triggering! So the following all look kinda nonsensical: 

```
# Makes users righteously cranky:
a = > b?  
a = > b:succeeded?
a = > b:x?
```

*"If a succeeds, trigger b:succeeded?"* - what???

## Proposed tweak

**Relax the requirement to flag optional outputs on the right of dependencies**:
- by default, right sides define only triggering (the syntax is intuitive)
  - (but still support right side `?` as well - opt-in confusion, for back-compat)
- left sides determine output optionality as now (the syntax is intuitive)
  - optional branching - the most important thing - is still explicit everywhere
- for optional outputs that are terminal (i.e., no tasks trigger off of them):
    - use a standalone graph item `a?` (the syntax is intuitive - there's no dependency arrow!)
    - (or opt-in to current right-side syntax)

### Examples:

```
x => y  # (implies nothing about y's outputs, but y:succeeded is required *by default*)
```

```
x =>y
y => z  # y:succeeded is required
```

```
x => y
y? => z  # y:succeeded is optional
```

```
x => y
y?  # y:succeeded is optional (note: no confusing arrow pointing at it!)
```

```
x => y?  # y:succeeded is optional (opt-in confusion, for back-compat)
```

hjoliver:
Close #6744 

**Check List**


cylc/flow/graph_parser.py
```diff
@@ -933,8 +941,13 @@ def _compute_triggers(
                     ) from None
             else:
                 fam = False
-                if not output:
-                    # Make implicit success explicit.
+                # RHS: don't infer ":succceeded", except:
```
wxtim: Technically this is a slightly different expression, but I've put some thought into accessing the `else:` statement by trying to create a situation where `not output & not optional & expr`, and have been unable to do it.

cylc/flow/graph_parser.py
```diff
@@ -705,6 +708,11 @@ def _families_all_to_all(
             else:
                 n_info += [(name, offset, trig)]
 
+            # Check output optionality consistency for left side nodes.
+            # (We don't infer output optionality from right side nodes).
+            # Note suicide is False by definition for left side nodes.
+            self._set_output_opt(name, trig, opt, False, fam_member)
```
wxtim: Is this not called as part of `_compute_triggers`?

Besides, this is now adding families to `self.task_output_opt` that were not previously being added.

consider

```cylc
    [[graph]]
        # P1 = foo & foo[-P1] & @xt & qux:finished => bar
        P1 = FOO:succeed-all => bar

[runtime]
    [[FOO]]
    [[foo, food]]
        inherit = FOO

```

Contents of `self.task_output_opt`

```diff
{
+    ('FOO': 'succeed-all'): (False, False, False)
    ('bar', 'succeeded'): (False, False, True),
    ('foo', 'succeeded'): (False, False, False),
    ('food', 'succeeded'): (False, False, False),
```

tests/integration/test_optional_outputs.py
```diff
@@ -508,3 +512,289 @@ async def test_removed_taskdef(
...
+            """,
+            # omit the specific outputs here - order not deterministic
+            "must both be optional if both are used"
+            " (via family trigger defaults)",
+            id='7',
+        ),
+    ],
+)
+async def test_optional_outputs_consistency(flow, validate, graph, err):
+    """stuff."""
```
wxtim: .. that for a docstring?

cylc/flow/graph_parser.py
hjoliver: Hmm, good spot. I don't think that will have any deleterious effects, but best not to do it ... checking ...

hjoliver: > Is this not called as part of `_compute_triggers`?

Yes, but it is only called for right-side nodes there. 

On master, that was it, because optionality was only set and checked for right-side nodes. (Because all nodes appear on the right at some point, even lone and first-in-chain nodes appear as `None => foo` pairs). 

On this branch optionality is set and checked definitively for left-side nodes and only for right-side nodes with explicit outputs (or explicit `?` short for `:succeeded?`)

```
foo:out => bar  # foo:out required; no implications for bar:succeeded (but defaults to required)
foo:out => bar?  # foo:out required; bar:succeeded optional
```

hjoliver: > Besides, this is now adding families to self.task_output_opt that were not previously being added.

Good spotting. It doesn't break anything because families aren't used in there, but I've fixed it anyway as I think it might be possible to get a bug due to it not doing what I was intending there in the family case (which indicates a coverage hole - I'll try to hunt that down and add an associated test tomorrow).

cylc/flow/graph_parser.py
```diff
@@ -685,15 +685,17 @@ def _families_all_to_all(
         n_info = []
         n_expr = expr
         for name, offset, trig, opt in info:
-            fam_member = False
             if (name, trig) in family_trig_map:
-                fam_member = True
+                fam = True
```
wxtim: Too short to be self documenting: I think fam_member was more helpful - that or `in_family` to distinguish from `is_family`.

Also a great walrus opportunity?

```python
if family := (name, trig) in family_trig_map:
```

Which would also save you setting it in the else loop:

```python
>>> if x := (1 == 2):
>>>    print(f'{x=}')
>>> else:
>>>    print(f'{x=}')
x=False
```

cylc/flow/graph_parser.py
```diff
@@ -922,46 +924,46 @@ def _compute_triggers(
                 fam = True
                 rhs_members = self.family_map[name]
                 if not output:
-                    # (Plain family name on RHS).
-                    # Make implicit success explicit.
-                    output = QUAL_FAM_SUCCEED_ALL
+                    if not expr:
```
wxtim: This section seems over-nested can we flatten it?

```python
if not output and not expr:
    ...
elif output and ouput.startswith('finish') and optional:
    ...

```


I think that the comments would make more sense if they have examples: Is this one `R1 = FAMILY`?

cylc/flow/graph_parser.py
```diff
@@ -922,46 +924,46 @@ def _compute_triggers(
...
                 if not offset:
                     # Nodes with offsets on the RHS do not define triggers.
                     self._set_triggers(mem, suicide, trigs, expr, orig_expr)
                 for output in outputs:
-                    if output is None:
-                        # no explicit output on RHS
-                        continue
-                    # But they must be consistent with output optionality.
-                    self._set_output_opt(mem, output, optional, suicide, fam)
+                    if output:
```
wxtim, line 1008: More elegant and less nested as a comprehension IMO

```python
[
    self._set_output_opt(mem, output, optional, suicide, fam)
    for output in outputs if output
]

```

cylc/flow/graph_parser.py
hjoliver: Good suggestion if the variable was being used anywhere else as well, but it wasn't (anymore) so better to remove it and use a named arg in the method call (the one place it was being used) instead.

cylc/flow/graph_parser.py
hjoliver: Done for the first block, but not the `elif` block which really has a sneaky `else` inside it that I've now made explicit.

cylc/flow/graph_parser.py
hjoliver: Well it's elegant, but it unnecessarily constructs a list.

cylc/flow/graph_parser.py
hjoliver: (Outdated)

cylc/flow/graph_parser.py
```diff
@@ -621,6 +664,10 @@ def _proc_dep_pair(
                         expr = re.sub(this, that, expr)
                 else:
                     # Make success triggers explicit.
+                    if name in self.family_map:
+                        raise GraphParseError(
+                            f"Family trigger required: {left} => {right}"
+                        )
```
hjoliver, line 670: Bug in master: the code here was adding `:succeeded` to bare family names on the left side of trigger pairs: `FAM --> FAM:succeeded`.

This results in failed validation later, but with a cryptic error message.

E.g. for `a => FAM => c` on master:
```
$ cylc val .
GraphParseError: Illegal family trigger in FAM:succeeded
```

On this branch:
```
$ cylc val .
GraphParseError: Family trigger required: FAM => c
```

cylc/flow/graph_parser.py
```diff
@@ -796,6 +843,7 @@ def _set_output_opt(
             for outp in [TASK_OUTPUT_SUCCEEDED, TASK_OUTPUT_FAILED]:
                 self._set_output_opt(
                     name, outp, optional, suicide, fam_member)
+            return
```
hjoliver, line 846: Bug in master, possibly inconsequential though.

cylc/flow/graph_parser.py
wxtim:
```python
(
    self._set_output_opt(mem, output, optional, suicide, fam)
    for output in outputs if output
)

File: cylc/flow/config.py [hunk 9]
```diff
@@ -2394,6 +2394,35 @@ def load_graph(self):
...
+            # Print inferred output optionality, for debugging graph parser.
+            # (Note: this excludes tasks that default to success-required
+            # via the RHS of a trigger with no explicit output attached).
+            outputs = [
+                f" \u2022 {name}:{output} "
+                f"{'optional' if optional else 'required'}"
+                for (name, output), (optional, _, _) in task_output_opt.items()
+            ]
+            if outputs:
+                print(
```
wxtim: Why not do this with logging?
```python
LOG.info(optional outputs)
LOG.debug(required outputs)
```

cylc/flow/graph_parser.py
```diff
@@ -104,8 +104,44 @@ class GraphParser:
...
+
+    RIGHT sides are just TASKS to trigger when the left side outputs complete:
+        [left] => d  # does NOT imply d:succeeded is required
+
+    But output optionality can be *explicitly* set on the right, if you like:
+        [left] => d?
+        [left] => d:x?
+    means
+        [left] => d  # AND d:succeeded is optional
+        [left] => d  # AND d:x is required
```
wxtim: Sorry? `d:x?` makes `d:x` required?

cylc/flow/graph_parser.py
```diff
@@ -104,8 +104,44 @@ class GraphParser:
             NODE(<REMOTE-WORKFLOW-QUALIFIER>)(:QUALIFIER)
         * Outputs (boo:x) are ignored as triggers on the RHS to allow chaining:
             "foo => bar:x => baz & qux"
-    """
 
+    Chained triggers get decomposed into pairs for processing, e.g.:
```
wxtim, line 108: This is very helpful in my opinion.

cylc/flow/graph_parser.py
hjoliver: That constructs a generator object.  All we  want to do here is call a function for each non-null output, for the function side effects, so I think a loop is fine for that

hjoliver: Could construct a list of non-null outputs first, in a comprehension, to avoid the `if` inside the loop, but is that worth it?

cylc/flow/config.py
hjoliver: I don't think we'd want to log optional outputs as INFO by default, without the `-v` option. At the moment, non-verbose validation just says its valid or prints the error if not valid.  [I think...]

I suppose we could log as INFO behind a `-v` option as well, but that's probably not what you mean?

cylc/flow/graph_parser.py
hjoliver: Oops, fixing.

hjoliver: Pushed with `[skip ci]'.

cylc/flow/config.py
hjoliver: Also, we are printing rather logging non-debug validation output. [I think...]

wxtim: Perhaps in that case we should have `#` at the start of each line so that what it printed is a valid `flow.cylc` file?

oliver-sanders: TBH, I don't think this warrants logging (at any level).

Only developers would ever be interested in this, but there's kinda a better way to get at this info by using `cylc config` and looking at the `completion` setting which contains more info (e.g. consider `completion = expired or failed or (succeeded and x)`).

cylc/flow/graph_parser.py
wxtim: I don't care _that_ much.

cylc/flow/config.py
oliver-sanders: > I don't think we'd want to log optional outputs as INFO by default, without the -v option

Yes, and we can still do that, but we should still use `LOG.info` or `LOG.debug` rather than `print` in any case. Note, at present this is polluting stdout, hence Tim's comment about adding `#` (as this is presumably breaking `cylc config`, etc).

oliver-sanders: > Also, we are printing rather logging non-debug validation output. [I think...]

I don't think so?

```
$ git grep 'print(' -- cylc/flow/config.py
```

hjoliver: Fair enough, wires evidently crossed (mine) on print vs logging in validation. 

I hadn't noticed the `cylc config` contamination as it only occurs with `-v` which is unlikely to be used with that command.

I'm not particularly attached to this -  hence flagged as a "IMO" only in Element chat.  I just thought it might be useful to highlight optional outputs at validation, because users should be aware of their potential to silently terminate graph branches. But you're probably right that viewing completion via `config` is a better way to do that.

OK: I'll do a quick move to `LOG.debug` tomorrow, as it is useful for developers to be able to see what the parser has inferred from the graph.

tests/integration/test_optional_outputs.py
```diff
@@ -508,3 +513,509 @@ async def test_removed_taskdef(
...
+            id='10',
+        ),
+        pytest.param(
+            """
+            a => FAM => c
+            """,
+            "Family trigger required: FAM => c",
+            id='11',
+        ),
+    ],
```
wxtim, line 597: You can replace all the ids with:

```suggestion
    ],
    ids=range(1, 12)
```

<details>
<summary>...or, if you wanted to be fancy</summary>
```python

def massage_id(val):
    # Remove chars likely to make a bash shell sad
    out = dedent(val).strip(
    ).replace('\n', '-'
    ).replace(' ', ''
    ).replace('\'', '')
    if len(out) > 20:
        return ''
    return out


# ....

ids=massage_id
```
</details>

oliver-sanders: > `ids=range(1, 12)`

:+1:

tests/integration/test_optional_outputs.py
```diff
@@ -508,3 +513,509 @@ async def test_removed_taskdef(
...
+            id='14',
+        ),
+        pytest.param(
+            """
+            a => c:x?
+            """,
+            {
+                ("a", "succeeded"): True,  # inferred
+                ("c", "succeeded"): True,  # default
+                ("c", "x"): False,  # inferred
```
wxtim, line 826:
```suggestion
                ("c", "x"): False,  # manually set
```

tests/integration/test_optional_outputs.py
hjoliver: Nice, didn't know that.

cylc/flow/config.py
hjoliver: Done.

tests/integration/test_optional_outputs.py
hjoliver: Even better:

```
ids=itertools.count()
```

tests/integration/test_optional_outputs.py
hjoliver: Fixed a few of those (inferred -> explicit).

tests/integration/test_optional_outputs.py
wxtim: The more I learn about itertools the better I like it.

wxtim (approved):

cylc/flow/config.py
```diff
@@ -2394,34 +2394,28 @@ def load_graph(self):
...
+        if optionals:
+            LOG.debug(
+                "Optional outputs inferred from the graph:\n"
+                f"{'\n'.join(optionals)}"
+            )
+        if requireds:
+            LOG.debug(
+                "Required outputs inferred from the graph:\n"
+                f"{'\n'.join(requireds)}"
+            )
```
oliver-sanders, line 2418: I'll move this suggestion to a new PR so as not to hold up 8.6.0

This forces the computation of these two lists and the log message strings, even if the messages are not emitted.

Suggest keeping the `if cylc.flow.flags.verbosity > 1` statement to shortcut this.

```suggestion
        if cylc.flow.flags.verbosity > 1:
            # Print inferred output optionality, for debugging graph parser.
            # Note this excludes tasks that just default to success-required.
            optionals = [
                f" \u2022 {name}:{output}"
                for (name, output), (optional, _, _) in task_output_opt.items()
                if optional
            ]
            requireds = [
                f" \u2022 {name}:{output}"
                for (name, output), (optional, _, _) in task_output_opt.items()
                if not optional
            ]
            if optionals:
                LOG.debug(
                    "Optional outputs inferred from the graph:\n"
                    f"{'\n'.join(optionals)}"
                )
            if requireds:
                LOG.debug(
                    "Required outputs inferred from the graph:\n"
                    f"{'\n'.join(requireds)}"
                )
````

hjoliver: Yeah, rush job before exiting the office sorry - will follow-up!

hjoliver: #7008

PR #7013, opened 2025-10-01
packaging: check pip/conda compatibility

oliver-sanders:
An attempt to ensure that the `conda-environment.yml` file is kept in sync with the pip config file(s).

For context, see https://github.com/conda-forge/cylc-flow-feedstock/pull/116#issuecomment-3356769809

**Check List**

### DIFF
--- a/.github/workflows/test_conda-build.yml
+++ b/.github/workflows/test_conda-build.yml
@@ -5,6 +5,8 @@ on:
     paths:
       - 'conda-environment.yml'
       - '.github/workflows/test_conda-build.yml'
+      - 'pyproject.toml'
+      - 'setup.cfg'
   schedule:
     - cron: '17 22 * * 6'  # Every Saturday at 22:17
   workflow_dispatch:
@@ -31,6 +33,7 @@ jobs:
         python-version: ['3.12', '3']
     env:
       ENV_FILE: conda-environment.yml
+      PIP_NO_DEPS: True
     steps:
       - name: checkout cylc-flow
         uses: actions/checkout@v5
@@ -52,7 +55,12 @@ jobs:
       - name: check cylc installation
         shell: bash -el {0}
         run: |
+          # make sure Cylc is functioning
           conda run -n test cylc version --long
+
+          # make sure pip is happy with the packages installed
+          # (checks pip deps are compatible with conda deps)
+          conda run -n test pip check
 
       - name: check for activate scripts
         shell : bash -el {0}


oliver-sanders:
That did the trick, the conda builds failed with this error:

```
ERROR conda.cli.main_run:execute(127): `conda run pip check` failed. (See above for error)
cylc-flow 8.6.0 has requirement protobuf<8,>=5, but you have protobuf 4.24.4.
```

Reflecting the incompatibility between the Conda and pip metadata.

This will *help* in detecting cases where pip deps have been updated, but conda deps haven't, but won't catch **all** cases.

oliver-sanders:
(note, the test will fail until the dependency is fixed!)

.github/workflows/test_conda-build.yml
```diff
@@ -31,6 +33,7 @@ jobs:
         python-version: ['3.12', '3']
     env:
       ENV_FILE: conda-environment.yml
+      PIP_NO_DEPS: True
```
MetRonnie, line 37: Why is this needed?

oliver-sanders: If the deps are incompatible, pip will just reinstall some stuff at compatible versions, invalidating the test.

```suggestion
      # prevent "pip install" from meddling with the conda installed dependencies
      PIP_NO_DEPS: True
```

MetRonnie: But `pip install` is not run, only `pip check`?

oliver-sanders: It does, but via conda.

The conda-environment.yml file only defines the dependencies of Cylc, creating this environment would not install Cylc itself. So we wouldn't be able to test anything (including pip dependenices).

So this workflow modifies the conda-environment,yml file to add this onto the end:

```
  - pip
  - pip:
    - .
```

Which is how cylc-flow gets pip installed.

MetRonnie: Ah, got it

MetRonnie:
```suggestion
      # prevent pip from meddling with the conda installed dependencies (during the conda build step)
      PIP_NO_DEPS: True
```

.github/workflows/test_conda-build.yml
```diff
@@ -52,8 +56,13 @@ jobs:
       - name: check cylc installation
         shell: bash -el {0}
         run: |
+          # make sure Cylc is functioning
```
MetRonnie, line 59: One thing I noticed when testing locally is that conda seems pretty keen on installing Python 3.12 even when the current is 3.14. Can we check that 3.14 is not being overridden?
```suggestion
          conda run -n test python --version
          
          # make sure Cylc is functioning
```

oliver-sanders: That would just be testing the `conda-incubator/setup-miniconda@v3` action?

MetRonnie: I'm concerned the `python-version` we specify is not working (e.g. for `3` it looks like we might be getting `3.12`: https://github.com/cylc/cylc-flow/actions/runs/18500486273/job/52715670153?pr=7013#step:4:1438

Possibly this issue: https://github.com/conda-incubator/setup-miniconda/issues/114

oliver-sanders: Right.

We could hard pin the python version at the Conda level (i.e add `python = <version>` into the conda environment file) which should be as safe as the Conda environment itself (if that dependency brokering is broken, all bets are off).

oliver-sanders: Implemented the above suggestion in c91c4029c

Unfold the last section in the "build Conda env" step to reveal the version installed.

MetRonnie: Not sure it has worked: https://github.com/cylc/cylc-flow/actions/runs/18534258646/job/52824976536?pr=7013#step:4:1447

(My `python --version` suggestion doesn't have to be permanent btw, at least would make it easier to see)

oliver-sanders: It's worked fine! We never told it to install 3.14, we just hoped it would.

The Python version is set to "3" not "3.14".

oliver-sanders: I we want to **force** our actions (including this one) to run with the latest Python, we'll have to explicitly state it.

But that's a right maintenance pain as we'll have to keep updating it :(

* Using an action to extract the version is possible, but ugly, but at least automated.
* Configuring it in a file is still a maintenance pain, but at least a centralised one.
* Setting it as a repository secret/variable is a maintenance pain and non-obvious, but at least centralised.

Orthogonal to this new check either way, but probs something we should look into soon to help spot leading edge issues sooner as current tests ain't gonna do that.

MetRonnie: Ah right, I was thinking of `actions/setup-python` which does this effortlessly. I guess updating the latest one once a year isn't too bad... But at present there is no point having a separate `3.12` and `3` in the matrix

oliver-sanders: With setup-python, `3.x` always picks the latest Python version for the operating system, not necessarily the latest Python.

With Conda, `3` always picks the latest Python compatible with the preferable solve of the environment, not necessarily the latest Python.

Neither is perfect, but setup-python is more likely to give the preferred answer more of the time.

Specifying `3` isn't "pointless", once the dependency issue that this PR is identifying has been resolved, it will pick up 3.14 as intended.

MetRonnie: Ohhh it was choosing 3.12 because of the old protobuf pin, got it.

Still, it is useful to have a `conda run -n test python --version` or better yet `conda list -n test` step for debugging purposes

oliver-sanders: Yep. It isn't ideal that this can happen, am mulling possible workarounds for future...

PR #7024, opened 2025-10-09
Ensure next/prev syntax works for `--startcp` option

MetRonnie:
E.g. `cylc play --startcp 'previous(T00)'`

**Check List**

- [x] No dependency changes 
- [x] Docs updated

### DIFF
--- a/changes.d/7024.feat.md
+++ b/changes.d/7024.feat.md
@@ -1,1 +1,1 @@
-Ensure `next()/previous()` syntax works for the `cylc play --startcp` option.
+Add support for the `next()/previous()` syntax with the `cylc play --startcp` option.


cylc/flow/scheduler_cli.py
```diff
@@ -162,10 +173,14 @@
     OptionSettings(
         ["--start-cycle-point", "--startcp"],
         help=(
-            "Set the start cycle point, which may be after"
-            " the initial cycle point. If the specified start point is"
-            " not in the sequence, the next on-sequence point will"
-            " be used. (Not to be confused with the initial cycle point)"),
```
MetRonnie, line 168: (I notice leading spaces have crept into the codebase - not sure what the motivation is, as this is not how we write paragraphs ordinarily. Perhaps we should specify the style to follow to avoid inconsistency?)

changes.d/7024.feat.md
```diff
@@ -0,0 +1 @@
+Ensure `next()/previous()` syntax works for the `cylc play --startcp` option.
```
oliver-sanders:
```suggestion
Add support for the `next()/previous()` syntax with the `cylc play --startcp` option.
```

cylc/flow/scheduler_cli.py
oliver-sanders: ???

Whether the space goes at the start or end of the line is arbitrary, neither is "how we write paragraphs ordinarily" since we don't write this whitespace in plain text. So neither way is right or wrong, natural or unnatural, it's just a style choice like any other.

Leading spaces are preferable because it is much easier to spot when a space has been omitted as they are all aligned in the first column (we have had several cases where tailing spaces were omitted) and with this arrangement commenting out or removing a line will never leave an erroneous space.

It's exactly the same reason why [placing operators at the start rather than end of the line is considered preferable](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) (PEP8, flake8, black, ruff, prettier).

FYI: There are no linting rules for strings, so style cannot be enforced.

cylc/flow/option_parsers.py
```diff
@@ -123,10 +123,11 @@ def _update_sources(self, other):
...
     help=(
-        "Set the initial cycle point."
-        " Required if not defined in flow.cylc."
-        "\nMay be either an absolute point or an offset: See"
-        f" {SHORTLINK_TO_ICP_DOCS} (Cylc documentation link)."
+        "Set the initial cycle point. "
+        "Required if not defined in flow.cylc.\n"
+        "Can be an absolute point or an offset relative to the "
+        "current time - see "
+        f"{SHORTLINK_TO_ICP_DOCS} (Cylc documentation link)."
```
oliver-sanders, line 130: Why change this?

cylc/flow/scheduler_cli.py
```diff
@@ -162,10 +173,14 @@
...
+            "Set the start cycle point, which may be after "
+            "the initial cycle point. "
+            "If the specified start point is not in the sequence, the next "
+            "on-sequence point will be used.\n"
+            "Can be an absolute point or an offset relative to the "
+            "current time, like the --initial-cycle-point option - see "
+            f"{SHORTLINK_TO_ICP_DOCS} (Cylc documentation link)."
```
oliver-sanders, line 182: Why change this?

cylc/flow/scheduler_cli.py
wxtim: I agree with @oliver-sanders

wxtim:
Looks LGTM, and seemed to work when I tried it out, apart from a bug where Cylc VIP or Play seems to freeze if the startcp >> icp because checking whether it is in the sequence(s) can take a long time if there are lots of items in the sequence.

cylc/flow/option_parsers.py
MetRonnie: For clarity

MetRonnie:
@wxtim are you going to approve?

changes.d/7024.feat.md
hjoliver: I prefer @oliver-sanders phrasing here, although it doesn't matter that much.

"Ensure it works" suggests that it was supposed to work but sometimes doesn't.

cylc/flow/option_parsers.py
hjoliver: @MetRonnie 's version is a bit clearer (offset *relative to current time*)

cylc/flow/scheduler_cli.py
hjoliver: I think it can be argued both ways, but I'm somewhat more persuaded by @oliver-sanders leading space argument.

cylc/flow/scheduler_cli.py
hjoliver: The new text added at the end adds some helpful clarity.  (Separately from the leading/trailing space issue).

hjoliver (approved):
LGTM.

PR #7025, opened 2025-10-09
replace deprecated ast constants

### LINKED ISSUE #7021: Deprecated `ast` classes removed in Python 3.14
https://docs.python.org/3/whatsnew/3.14.html#id9

>Remove the following classes, which have been deprecated aliases of [`Constant`](https://docs.python.org/3/library/ast.html#ast.Constant) since Python 3.8 and have emitted deprecation warnings since Python 3.12:
>- `Bytes`
>- `Ellipsis`
>- `NameConstant`
>- `Num`
>- `Str`
>
>As a consequence of these removals, user-defined `visit_Num`, `visit_Str`, `visit_Bytes`, `visit_NameConstant` and `visit_Ellipsis` methods on custom [`NodeVisitor`](https://docs.python.org/3/library/ast.html#ast.NodeVisitor) subclasses will no longer be called when the `NodeVisitor` subclass is visiting an AST. Define a `visit_Constant` method instead.

We are using `Num` and `Str`. Also `Index` which is also deprecated.

https://github.com/cylc/cylc-flow/blob/3302cbfaa29f79ca0414a9668bc3b92158f23dd7/cylc/flow/host_select.py#L104-L114

```
$ cylc play workflow
...
AttributeError: module 'ast' has no attribute 'Num'
```

oliver-sanders:
* Closes #7021

Unblocks Py 3.14 compat.

**Check List**

### DIFF
--- a/cylc/flow/host_select.py
+++ b/cylc/flow/host_select.py
@@ -31,6 +31,8 @@ Examples:
     True
     >>> RankingExpressionEvaluator('1 in (1, 2, 3)')
     True
+    >>> RankingExpressionEvaluator('[1,2,3][-1] ** 2')
+    9
     >>> import psutil
     >>> RankingExpressionEvaluator(
     ...     'a.available > 0',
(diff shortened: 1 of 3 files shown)


cylc/flow/host_select.py
```diff
@@ -31,6 +31,8 @@
     True
     >>> RankingExpressionEvaluator('1 in (1, 2, 3)')
     True
+    >>> RankingExpressionEvaluator('[1,2,3][-1] ** 2')
+    9
```
oliver-sanders, line 35: Tests for the evaluator can be found here.

PR #7026, opened 2025-10-10
Fix small xtrigger housekeeping bug.

hjoliver:
Found while investigating [this forum post](https://cylc.discourse.group/t/cylc-xtrigger-question/1255/6)

The "next call" dict in `xtrigger_mgr` is not being housekept, so if the same xtrigger comes back into the pool later the "commencing xtrigger call" message will not be logged again when the calls do re-commence.

**Check List**

### DIFF
--- a/tests/integration/test_xtrigger_mgr.py
+++ b/tests/integration/test_xtrigger_mgr.py
@@ -27,6 +27,7 @@ from cylc.flow.id import TaskTokens
 from cylc.flow.pathutil import get_workflow_run_dir
 from cylc.flow.scheduler import Scheduler
 from cylc.flow.subprocctx import SubFuncContext
+from cylc.flow.task_outputs import TASK_OUTPUT_SUCCEEDED
 
 
 async def test_2_xtriggers(flow, start, scheduler, monkeypatch):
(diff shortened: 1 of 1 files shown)


cylc/flow/xtrigger_mgr.py
```diff
@@ -711,6 +711,7 @@ def call_xtriggers_async(self, itask: 'TaskProxy'):
             # General case: potentially slow asynchronous function call.
             if sig in self.sat_xtrig:
                 # Already satisfied, just update the task
+                LOG.info(f"[{itask}] satisfying xtrigger prerequisite: {sig}")
```
hjoliver, line 714: IMO we should log this as INFO, as confirmation that an xtrigger prerequisite is being satisfied by an existing xtrigger result.

cylc/flow/xtrigger_mgr.py
```diff
@@ -760,7 +761,10 @@ def housekeep(self, itasks):
                 itask, sigs_only=True, unsat_only=True)
         for sig in list(self.sat_xtrig):
             if sig not in all_xtrig:
+                LOG.debug(f"Housekeeping xtrigger result: {sig}")
                 del self.sat_xtrig[sig]
+                with suppress(KeyError):
+                    del self.t_next_call[sig]
```
hjoliver, line 767: This is the bug fix.

Issue #6994, opened 2025-09-19
Platform subshell can cause remote init to happen on a different platform if triggering a task while paused


MetRonnie:
### Reproducible Example

Start this workflow paused (it helps if you have as many platforms as possible to choose from in the subshell):

```cylc
[scheduling]
    [[graph]]
        R1 = foo
[runtime]
    [[foo]]
        script = cylc message -- "$(hostname)"; false
        platform = $(python -m random -c platform1 platform2 platform3 platform4)
```

Trigger the task - you will see the scheduler report it is on one platform but the actual platform it is submitted to will be revealed to be different by the message that comes in.

### Expected Behaviour

The platform subshell is being re-evaluated when it should be cached. The expected behaviour is that the the task remote inits on the platform that has already been evaluated. Subsequent submissions should re-evaluate it once on each submission.

PR #5062, opened 2022-08-12
Wait for preparing tasks to submit before auto restart

MetRonnie:
Fix bug where any preparing local jobs would get orphaned during the shutdown before an auto restart.

You could reproduce this bug by changing
```diff
diff --git a/tests/functional/restart/42-auto-restart-ping-pong.t b/tests/functional/restart/42-auto-restart-ping-pong.t
index 5f897b33e..ecf58e99f 100644
--- a/tests/functional/restart/42-auto-restart-ping-pong.t
+++ b/tests/functional/restart/42-auto-restart-ping-pong.t
@@ -74,5 +74,5 @@ log_scan2() {
 }
 
-EARS=5  # number of times to bounce the workflow between hosts
+EARS=40  # number of times to bounce the workflow between hosts
 NO_TESTS="$(( EARS * 5 + 1 ))"
 set_test_number "${NO_TESTS}"
```

and running the test, eventually it would orphan a preparing job and cause the test to fail.

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] No dependency changes 
- [x] Tests are included
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [x] No docs needed
- [ ] If this is a bug fix, PRs raised to both master and the relevant maintenance branch.

PR #7035, opened 2025-10-13
Fix platform subshell evaluating more than once for tasks triggered in paused workflow

MetRonnie:
Closes https://github.com/cylc/cylc-flow/issues/6994

Stop trying to submit the triggered tasks once `itask.waiting_on_job_prep == False`.

This is kind of a follow-up to https://github.com/cylc/cylc-flow/pull/6836,  https://github.com/cylc/cylc-flow/pull/6768, https://github.com/cylc/cylc-flow/pull/6345, https://github.com/cylc/cylc-flow/pull/5062. It also predates and fixes the same bug as #7054.

Also fixes a scheduler TypeError crash if platform subshell evaluates to an empty string.

Also fixes an incomplete warning message

> A mixture of Cylc 7 (host) and Cylc 8 (platform) logic should not be used. In this case for the task "1/foo" the following are not compatible:

**Check List**

- [x] No dependency changes 
- [x] No docs needed

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -1355,8 +1355,8 @@ class Scheduler:
                 pre_prep_tasks.update(self.pool.release_queued_tasks())
 
         if (
-            # Manually triggered tasks will be in preparing state and should
-            # be submitted even if paused (unless stopping).
+            # Manually triggered tasks will be preparing and should
+            # be submitted even if paused (unless workflow is stopping).
             self.is_paused and not self.stop_mode
         ) or (
             # Need to get preparing tasks to submit before auto restart
@@ -1369,8 +1369,7 @@ class Scheduler:
             pre_prep_tasks.update({
                 itask
                 for itask in self.pool.get_tasks()
-                if itask.state(TASK_STATUS_PREPARING)
-                and itask.waiting_on_job_prep
+                if itask.waiting_on_job_prep
             })
 
         # Return, if no tasks to submit.
--- a/tests/integration/test_force_trigger.py
+++ b/tests/integration/test_force_trigger.py
@@ -42,7 +42,17 @@ from cylc.flow.task_state import (
 )
 
 
-async def test_trigger_workflow_paused(
+async def test_workflow_paused_simple(
+    one_conf, flow, scheduler, run, complete
+):
+    """It should run triggered tasks even if the workflow is paused."""
+    schd: Scheduler = scheduler(flow(one_conf), paused_start=True)
+    async with run(schd):
+        await run_cmd(force_trigger_tasks(schd, ['1/one'], ['1']))
+        await complete(schd, '1/one', allow_paused=True, timeout=1)
+
+
+async def test_workflow_paused_queues(
     flow: 'Fixture',
     scheduler: 'Fixture',
     start: 'Fixture',
@@ -111,7 +121,8 @@ async def test_trigger_workflow_paused(
 
 
 async def test_trigger_group_whilst_paused(flow, scheduler, run, complete):
-    """Only group start tasks should run whilst the scheduler is paused.
+    """Only group start tasks should run if the group is triggered whilst
+    the scheduler is paused.
 
     Group start tasks have only off-group dependencies.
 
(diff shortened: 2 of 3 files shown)


cylc/flow/platforms.py
```diff
@@ -329,7 +329,7 @@ def get_platform_from_group(
         return HOST_SELECTION_METHODS[method](platform_names)
 
 
-def platform_name_from_job_info(
+def _platform_name_from_job_info(
```
MetRonnie, line 332: (Sundering this function as it shouldn't be called elsewhere)

tests/integration/test_force_trigger.py
```diff
@@ -689,40 +683,13 @@ async def test_trigger_with_sequential_task(flow, scheduler, run, log_filter):
...
                 if log_filter(contains='[2/foo/02:running] (received)failed'):
                     break
                 await asyncio.sleep(0)
 
 
-async def test_trigger_whilst_paused_preparing(flow, scheduler, run, complete):
-    """It should run "preparing" tasks even if the workflow is paused.
-
-    Remote init leaves tasks as preparing for a while. These must still be
-    pushed through to running, even if triggered while the workflow is paused.
```
MetRonnie: ~This test was not actually testing a paused workflow.~ My bad, integration tests start paused by default (which is unintuitive but one for another time).

But it kind of needs to test live mode in order to be accurate. Really I think this is testing implementation details a bit too much

tests/functional/platforms/04-host-to-platform-upgrade-fail-inherit.t
```diff
@@ -29,16 +29,11 @@ create_test_global_config '' "
 
 install_workflow "${TEST_NAME_BASE}" "${TEST_NAME_BASE}"
 
-# Both of these cases should validate ok.
-run_ok "${TEST_NAME_BASE}-validate" \
-    cylc validate "${WORKFLOW_NAME}"
-
```
MetRonnie, line 35: I don't think it should validate ok! This is detectable at config parse time, however we are unlikely to fix this now as the deprecated syntax is to be removed in the next minor release

cylc/flow/scheduler.py
```diff
@@ -1350,36 +1350,30 @@ def release_tasks_to_run(self) -> bool:
...
                 self.pool.tasks_to_trigger_now = set()
 
-            if self.is_paused:
-                # finish processing preparing tasks
-                pre_prep_tasks.update({
-                    itask for itask in self.pool.get_tasks()
-                    if itask.state(TASK_STATUS_PREPARING)
-                })
-            else:
+            if not self.is_paused:
```
hjoliver, line 1353: This change nukes the bug fix that just went in with #7054 

Image

cylc/flow/scheduler.py
```diff
@@ -1350,36 +1350,30 @@ def release_tasks_to_run(self) -> bool:
...
-            # finish processing preparing tasks first
-            pre_prep_tasks = {
-                itask for itask in self.pool.get_tasks()
+            pre_prep_tasks.update({
+                itask
+                for itask in self.pool.get_tasks()
                 if itask.state(TASK_STATUS_PREPARING)
-            }
+                and itask.waiting_on_job_prep
+            })
```
MetRonnie, line 1373: @oliver-sanders This PR pre-dated #7054 and had basically the same fix.

However  #7054 missed L1376 - I have consolidated the two in this PR.

One difference here is that I did not delete the check for `itask.state(TASK_STATUS_PREPARING)`, I added the check for `itask.waiting_on_job_prep` in addition.

tests/integration/test_force_trigger.py
MetRonnie: Test added back in https://github.com/cylc/cylc-flow/pull/7035/commits/d3f5f68d57357fa77c597d5dd90bd2cfb857c235 since it was changed on 8.6.x

cylc/flow/scheduler.py
hjoliver: OK now waiting on @oliver-sanders to check that.

oliver-sanders: I don't think there's any need to have `itask.state(TASK_STATUS_PREPARING)` in combination with `itask.waiting_on_job_prep`, but it can't do any harm.

PR #7041, opened 2025-10-16
Remove unnecessary "tasks not killable" warnings for `cylc stop --kill`

MetRonnie:
Small change with no associated issue.

**Check List**

- [x] No dependency changes 
- [ ] Tests are not included as testing for absence of stuff is fragile.
- [x] Docs not needed

### DIFF
--- a/changes.d/7041.fix.md
+++ b/changes.d/7041.fix.md
@@ -1,0 +1,1 @@
+Removed spurious "tasks not killable" warnings for `cylc stop --kill`.


cylc/flow/scheduler.py
```diff
@@ -1511,8 +1511,8 @@ async def workflow_shutdown(self):
             self.time_next_kill is not None
             and time() > self.time_next_kill
         ):
-            await commands.run_cmd(commands.poll_tasks(self, ['*/*']))
-            await commands.run_cmd(commands.kill_tasks(self, ['*/*']))
+            # self.task_job_mgr.poll_task_jobs(self.pool.get_tasks())
+            self.kill_tasks(self.pool.get_tasks(), warn=False)
```
MetRonnie, line 1515: I have realised the poll will not necessary have finished before the kill is executed (this is the case before or after this change)

oliver-sanders: I think it runs this code in some sort of loop so I'm not sure that matters.

MetRonnie: Looking at `subprocpool.py` I believe it will run the kill processes while the poll processes are still potentially running if there is space in the process pool

oliver-sanders: Of course, but I think both are called in a loop, so the results of a poll will influence a future kill.

I.E, the kill will (eventually) be responsive to state changes resulting from the poll.

MetRonnie: Reinstated the poll as unrelated to this fix

oliver-sanders: Will resolve this comment.

The existing approach is fine. The main loop calls `workflow_shutdown` over and over, each time it issues a fresh batch of poll and kill commands until there are no active tasks left in the pool. The lack of wait between the two is not an issue. Worst case scenario, we issue a kill command for a task which is still active, no ill consequences.

oliver-sanders (approved):

PR #7042, opened 2025-10-16
Avoid traceback on Ctrl+C during interactive prompt

MetRonnie:
This change makes Ctrl+C during an interactive prompt more graceful - instead of giving traceback it will have the same effect as typing <kbd>n</kbd>+<kbd>Enter</kbd>

**Check List**

- [x] No dependency changes 
- [x] Docs not needed

### DIFF
--- a/changes.d/7042.feat.md
+++ b/changes.d/7042.feat.md
@@ -1,0 +1,1 @@
+Ctrl+C during interactive prompts is now handled more gracefully.


cylc/flow/scheduler_cli.py
```diff
@@ -565,15 +566,16 @@ def _version_check(
...
-                options.upgrade = prompt(
-                    cparse(
-                        'Are you sure you want to upgrade from'
-                        f' <yellow>{last_run_version}</yellow>'
-                        f' to <green>{__version__}</green>?'
-                    ),
-                    {'y': True, 'n': False},
-                    process=str.lower,
-                )
+                with handle_sigint():
```
oliver-sanders, line 569: Why not use:

```python
with suppress(KeyboardInterrupt):
```

Or:


```python
try:
    ...
except KeyboardInterrupt:
    sys.exit(0)
```

(As appropriate)

MetRonnie: I think that requires Ctrl+C twice while you are inside of the `input()` loop (or maybe doesn't work at all, can't remember), whereas this handles Ctrl+C immediately

oliver-sanders: So, `KeyboardInterrupt` seems to work fine, even in a loop:

```python
try:
    usr = None
    while usr not in ['y', 'n']:
        usr = input('type something: ')
except KeyboardInterrupt:
    print('interrupted')
```

If we have to do something different, so be it. But if we can stick with `KayboardInterrupt`, it is preferable because:

* It doesn't clear pre-existing signal handlers.
* It's more flexible because you can handle both cases (see example).
* It's simpler / no implicit `sys.exit`.

https://github.com/cylc/cylc-flow/blob/2d33667a78ee256b3d263ad3dbeb1d65a3019212/cylc/flow/scripts/reinstall.py#L237-L240

MetRonnie: I re-tested `try... except KeyboardInterrupt`. For `scheduler_cli` and `clean` it requires 2 presses of Ctrl+C vs 1 on this branch. `reinstall` was already using `except KeyboardInterrupt` but it was still giving traceback if you Ctrl+C'd twice, whereas it exits cleanly with 1 press on this branch.

oliver-sanders: Ok, I haven't poked at this but will take your word for it.

MetRonnie:
(Not re-running failing MacOS `tests/f/cylc-reinstall/01-file-transfer.t` but it has been fixed on master)

wxtim (approved):

PR #7044, opened 2025-10-17
tutorial: replace deprecated DataPoint API

### LINKED ISSUE #6053: tutorial workflow: move away from datapoint
Datapoint is past its sell by and no longer supported. It has become increasingly unstable for our usage so we will need to look at moving away.

Its replacement is the "Weather Observations Website" or WOW - https://www.wow.metoffice.gov.uk/

Unfortunately I don't think we will be able to bundle API keys for WOW (in order to skip the registration step) so would require users to sign up.

The current tutorial extracts cloud cover from radar images which may or may not be available in WOW, if not we could use the "current weather" field. This would help us drop the Pillow dependency.

> **Note:** Ideally the tutorial workflow would not have any additional dependencies beyond the regular cylc-flow dependencies.

wxtim:
Closes #6053 

Sibling to
- https://github.com/cylc/cylc-doc/pull/887
- https://github.com/metomi/rose/pull/2959

Datapoint is to be switched off from the first of December.

The data is available from the [Met Office via the Amazon Sustainability Data Initiative](https://www.metoffice.gov.uk/services/data/external-data-channels#ASDI). [Amazon link](https://registry.opendata.aws/met-office-uk-radar-observations/).

Annoyingly the metadata there seems quite limited, so I applied to a bunch of contacts in Obs R&D who insisted that the data has the following properties:

image

Which raises 2 problems:
1. The image is in a different projection to previous input data (Transverse rather than obverse Mercator projection)
2. The domain doesn't seem quite right with the values given (though these could just be a side effect of (1)

Rather than re-jigging the mathematics I've made something vaguely plausible by fiddling with domain values. Hopefully this creates a product good enough for the training purpose for which it is built.

Finally, I have fixed a bug I introduced when I added the SYNOP collecting routine - these wind observations are in meteorological convention (wind is blowing from), but we need where the wind is going to, so all wind directions were 180° off!

### DIFF
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -23,3 +23,5 @@ dependencies:
   #- pympler
   #- matplotlib-base
   #- sqlparse
+  #- h5py
+  #- requests
--- a/cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/consolidate-observations
+++ b/cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/consolidate-observations
@@ -61,7 +61,7 @@ def plot_wind_data(wind_x, wind_y, x_range, y_range, x_coords, y_coords,
                [x[0] for x in z_coords],
                [y[1] for y in z_coords],
                color='red')
-    plt.savefig('wind.png')
+    plt.savefig(f'{os.environ["CYLC_TASK_LOG_DIR"]}/wind.png')
 
 
 def get_wind_fields():
--- a/cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/forecast
+++ b/cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/forecast
@@ -150,11 +150,12 @@ def push_rainfall(rainfall, wind_data, step, resolution, spline_level):
                                 dim_x, dim_y, resolution, resolution,
                                 spline_level)
 
+    domain = util.parse_domain(os.environ['DOMAIN'])
+
     while True:
         out_of_bounds = []
         for itt in range(len(x_values)):
             try:
-                domain = util.parse_domain(os.environ['DOMAIN'])
                 lng = domain['lng1'] + x_values[itt]
                 lat = domain['lat1'] + y_values[itt]
 
--- a/cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-observations
+++ b/cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-observations
@@ -24,11 +24,8 @@ Usage:
     get-observations
 
 Environment Variables:
-    SITE_ID: The four digit DataPoint site code identifying the weather station
-        we are to fetch results for.
-    API_KEY: The DataPoint API key, required for getting live weather data.
-        If un-specified then get-observations will fall back to archive data
-        from the workflow directory.
+    SITE_ID: The four digit WMO (World Meteorological Organization)
+    site code identifying the weather station we are to fetch results for.
 
 """
 
@@ -42,10 +39,6 @@ import requests
 
 import util
 
-
-BASE_URL = ('http://datapoint.metoffice.gov.uk/public/data/'
-            'val/wxobs/all/json/{site_id}'
-            '?res=hourly&time={time}&key={api_key}')
 
 # Compass bearings for ordinate directions.
 # NOTE: We measure direction by where the winds have come from!
@@ -68,22 +61,52 @@ ORDINATES = {
     'NNW': '157.5'
 }
 
+class Meteorology:
+    """
+    Surface winds tend to be 20-30 degrees backed (vector anticlockwise)
+    from the winds which steer weather systems:
+    Friction with the ground surface tends to mean that land surface winds
+    are as slow as half the speed of the wind at 2000ft. This fudge factor is
+    a more conservative 1.5:
+
+    .. seealso::
+
+       [Source Book to the Forecaster's Reference Book](https://digital.nmla
+       .metoffice.gov.uk/IO_011f7cd4-50fc-4903-b556-d24480ea883d/), section
+       1.2
+    """
+    SURFACE_BACKING = 2
+    SURFACE_FRICTION = .66
+    KT_TO_MPH = 1.15078
+
+    @staticmethod
+    def process_direction(direction: str) -> str:
+        """Process raw wind direction:
+
+        * Convert direction from 10s of degrees to degrees.
+        * Convert from Oceanographic (wind to) to Meteorological (wind from)
+          convention.
+        * Surface Friction Correction
+        """
+        return str(
+            ((int(direction) + 18 - Meteorology.SURFACE_BACKING) % 36) * 10)
+
+    @staticmethod
+    def process_speed(speed: str) -> str:
+        """Process Raw wind speed
+
+        * Convert to KT to MPH
+        * Surface Friction Correction
+        """
+        return str(
+            (
+                int(speed) * Meteorology.KT_TO_MPH
+            ) / Meteorology.SURFACE_FRICTION
+        )
+
 
 class NoDataException(Exception):
     ...
-
-
-def get_datapoint_data(site_id, time, api_key):
-    """Get weather data from the DataPoint service."""
-    # The URL required to get the data.
-    time = datetime.strptime(time, '%Y%m%dT%H%MZ').strftime('%Y-%m-%dT%H:%MZ')
-    url = BASE_URL.format(time=time, site_id=site_id, api_key=api_key)
-    req = requests.get(url)
-    if req.status_code != 200:
-        raise Exception(f'{url} returned exit code {req.status_code}')
-    # Get the data and parse it as JSON.
-    print('Opening URL: %s' % url)
-    return req.json()['SiteRep']['DV']['Location']
 
 
 def get_archived_data(site_id, time):
@@ -152,11 +175,12 @@ def synop_grab(site_id, time):
         raise NoDataException(
             f'Request for data failed, raw request was {req.text}')
 
-    # Parse the direction from 10's of degrees to degrees:
-    data['direction'] = str(int(data['direction']) * 10)
+    # * Parse the direction from 10's of degrees to degrees
+    # * Convert direction from to direction it's blowing to
+    data['direction'] = Meteorology.process_direction(data['direction'])
 
     # Convert data in KT to MPH:
-    data['speed'] = str(int(data['speed']) * 1.15078)
+    data['speed'] = Meteorology.process_speed(data['speed'])
 
     # Get lat and long from MO Data:
     lat, lon = reference_lat_long(site_id)
@@ -185,7 +209,7 @@ def get_nearby_site(site_id, badsites):
     return int(result[0]), dist
 
 
-def main(site_id, api_key=None):
+def main(site_id):
     cycle_point = os.environ['CYLC_TASK_CYCLE_POINT']
 
     # Try to get the information from SYNOPS.
@@ -202,13 +226,8 @@ def main(site_id, api_key=None):
             site_id, dist = get_nearby_site(site_id, badsites)
 
     if obs is None:
-        if api_key:
-            print('Attempting to get weather data from DataPoint...')
-            data = get_datapoint_data(site_id, cycle_point, api_key)
-        else:
-            print('No API key provided, falling back to archived data')
-            data = get_archived_data(site_id, cycle_point)
-
+        print('Obs unavailable, falling back to archived data')
+        data = get_archived_data(site_id, cycle_point)
         obs = extract_observations(data)
 
     # Write observations.
@@ -218,5 +237,4 @@ def main(site_id, api_key=None):
 
 if __name__ == '__main__':
     util.sleep()
-    main(os.environ['SITE_ID'],
-         os.environ.get('API_KEY'))
+    main(os.environ['SITE_ID'])
--- a/cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-rainfall
+++ b/cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-rainfall
@@ -24,9 +24,6 @@ Usage:
     get-rainfall
 
 Environment Variables:
-    API_KEY: The DataPoint API key, required for getting live weather data.
-        If un-specified then get-observations will fall back to archive data
-        from the workflow directory.
     DOMAIN: The area in which to generate forecasts for in the format
         (lng1, lat1, lng2, lat2).
     RESOLUTION: The length/width of each grid cell in degrees.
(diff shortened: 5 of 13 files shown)


cylc/flow/etc/tutorial/cylc-forecasting-workflow/flow.cylc
```diff
@@ -40,14 +40,15 @@
             # The dimensions of each grid cell in degrees.
             RESOLUTION = 0.2
             # The area to generate forecasts for (lng1, lat1, lng2, lat2)
-            DOMAIN = -12,48,5,61  # Do not change!
+
```
wxtim, line 42: What Obs R&D Claim the Domain Should be

image

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-rainfall
```diff
@@ -160,37 +157,64 @@ def process_rainfall_data(filename, resolution, domain):
...
 
-    image = Image.open(filename)
-    scale = get_scale(domain, image.width)
+    # image = Image.open(filename)
+    height, width = data.shape
+
+    # TODO - work out why this isn't working
+    scale = get_scale(domain, width)
     offset = get_offset(domain, scale)
+    scale = (1672.2334443981335, 3344.466888796267)
```
wxtim: @oliver-sanders  - These numbers seem to work (roughly) might I poke you to have a look at why the functions do or don't?

If you run the workflow in debug mode you get piuctures

Compare my images 
* [With these lines](https://cylchub/services/cylc-review/view/tim.pillinger?&suite=cylc-forecasting-workflow%2Fmanually-set&no_fuzzy_time=0&path=log/job/20251017T0900Z/get_rainfall/01/processed.data.png)
* [With these lines commented](https://cylchub/services/cylc-review/view/tim.pillinger?&suite=cylc-forecasting-workflow%2Fdomain-set&no_fuzzy_time=0&path=log/job/20251017T0900Z/get_rainfall/01/processed.data.png)

Actually, only get_offset isn't giving a sensible answer.

NVM - I'll keep poking - I think that this is a problem with the projection.

cylc/flow/etc/tutorial/cylc-forecasting-workflow/flow.cylc
wxtim: But I'm not convinced -

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/consolidate-observations
```diff
@@ -61,7 +61,7 @@ def plot_wind_data(wind_x, wind_y, x_range, y_range, x_coords, y_coords,
                [x[0] for x in z_coords],
                [y[1] for y in z_coords],
                color='red')
-    plt.savefig('wind.png')
+    plt.savefig(f'{os.environ["CYLC_TASK_LOG_DIR"]}/wind.png')
```
wxtim, line 64: We're only saving this for people to look at, so it ought to go here?

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-observations
```diff
@@ -73,19 +73,6 @@ class NoDataException(Exception):
...
-    url = BASE_URL.format(time=time, site_id=site_id, api_key=api_key)
-    req = requests.get(url)
-    if req.status_code != 200:
-        raise Exception(f'{url} returned exit code {req.status_code}')
-    # Get the data and parse it as JSON.
-    print('Opening URL: %s' % url)
-    return req.json()['SiteRep']['DV']['Location']
-
-
 def get_archived_data(site_id, time):
```
wxtim, line 112: I can't see any evidence of actual archived data - Perhaps we should save some at some point?

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-observations
```diff
@@ -202,13 +193,8 @@ def main(site_id, api_key=None):
...
 
     if obs is None:
-        if api_key:
-            print('Attempting to get weather data from DataPoint...')
-            data = get_datapoint_data(site_id, cycle_point, api_key)
-        else:
-            print('No API key provided, falling back to archived data')
-            data = get_archived_data(site_id, cycle_point)
-
+        print('No API key provided, falling back to archived data')
```
wxtim:
```suggestion
        print('Obs unavailable, falling back to archived data')
```

oliver-sanders:
@wxtim, FYI, this will need to go into a bugfix release.

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-observations
```diff
@@ -152,8 +132,12 @@ def synop_grab(site_id, time):
         raise NoDataException(
             f'Request for data failed, raw request was {req.text}')
 
+    # Convert direction from (meteorlogical convention) to
+    # direction to:
+    data['direction'] = (int(data['direction']) + 18) % 36
+
     # Parse the direction from 10's of degrees to degrees:
-    data['direction'] = str(int(data['direction']) * 10)
+    data['direction'] = str(data['direction'] * 10)
```
oliver-sanders: Please could we re-combine these two lines (a bit confusing otherwise).

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-rainfall
```diff
@@ -38,20 +35,20 @@ import math
 import os
 import shutil
 
-import requests
+import h5py
```
oliver-sanders, line 38: Dependency change not reflected in the pip/conda dependency files?

```diff
+ h5py
- Pil?
```

cylc/flow/etc/tutorial/cylc-forecasting-workflow/flow.cylc
```diff
@@ -68,9 +75,12 @@
...
         [[[environment]]]
-            # The key required to get weather data from the DataPoint service.
-            # To use archived data comment this line out.
-            API_KEY = DATAPOINT_API_KEY
+            # A template Map file
+            MAP_FILE = "${CYLC_TASK_LOG_ROOT}-map.html"
+            # Create the html map file in the task's log directory.
+            MAP_FILE = "${CYLC_TASK_LOG_ROOT}-map.html"
+            # The path to the template file used to generate the html map.
+            MAP_TEMPLATE = "$CYLC_WORKFLOW_RUN_DIR/lib/template/map.html"
```
oliver-sanders: We didn't need to specify these here before.

Might be easier to stick with using the work or share directory as appropriate?

cylc/flow/etc/tutorial/cylc-forecasting-workflow/lib/python/util.py
```diff
@@ -28,6 +28,7 @@
 import jinja2
 
 R_0 = 6371.  # Radius of the Earth (km).
+INFO = '[INFO] '
```
oliver-sanders: ?

cylc/flow/etc/tutorial/cylc-forecasting-workflow/flow.cylc
wxtim: I did this because I thought that it was quite nice to view the intermediate stages from Cylc Review. Happy to remove if you think I should though.,

conda-environment.yml
```diff
@@ -23,3 +23,4 @@ dependencies:
   #- pympler
   #- matplotlib-base
   #- sqlparse
+  #- h5py
```
oliver-sanders, line 26: I think we install the tutorial dependencies by default in the conda-forge feedstock, so this might need to move to the main section, not sure.

But oddly, I don't see `requests` here?

oliver-sanders:
```suggestion
  #- h5py
  #- requests
```

oliver-sanders:
FYI: There's a wrap-around problem with rainfall leaking out of the west of the domain into the east:

(possible that this was a pre-existing issue?)

Screenshot from 2025-10-30 12-48-25

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-observations
oliver-sanders: I think we are using archive data as part of the Rose tutorial. It gets bundled with the app for use in the test config.

I think this test data will need to be re-generated from the new data source (check if this is true) otherwise the Rose app might fail due to a resolution mismatch?

oliver-sanders:
The `get-rainfall` step is really slow ~1:40. After a bit of fiddling, I managed to get that down to ~0:30 which is reasonable.

Note: It uses ~25% CPU (but negligible RAM) on my box.

https://github.com/wxtim/cylc/pull/75

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-observations
wxtim: Wierdly no, -we're using archived output of this task for the `forecast` task. This looks like legacy. Still need a rose PR to follow up this though. PR at https://github.com/metomi/rose/pull/2959

I think we should have a follow up issue to check and fix.

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-observations
```diff
@@ -68,22 +61,52 @@ ORDINATES = {
     'NNW': '157.5'
 }
 
+class Meteorology:
```
wxtim, line 64: This is probably a bit bonkers. I won't mind (very much) if you tell me to take it out again.

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/forecast
```diff
@@ -150,11 +150,12 @@ def push_rainfall(rainfall, wind_data, step, resolution, spline_level):
                                 dim_x, dim_y, resolution, resolution,
                                 spline_level)
 
+    domain = util.parse_domain(os.environ['DOMAIN'])
```
wxtim, line 153: Seems odd to run this every single loop.

cylc/flow/etc/tutorial/cylc-forecasting-workflow/flow.cylc
oliver-sanders: I would probably suggest removing these to keep the amount of configuration that users have to do in the tutorial minimal.

cylc/flow/etc/tutorial/cylc-forecasting-workflow/flow.cylc
```diff
@@ -39,15 +39,22 @@
...
+
+        [[[meta]]]
+            metadata source = Met Office Observations R&D
+            ellipsiod = "Airy 1830"
+            projection = "Transverse Mercator"
+            origin = "49,2"
+            SW = 43.8341,-12.0118
+            NW = 62.9207,-17.9739
+            NE = 62.6515,16.1289
+            SE = 437009,941578
```
oliver-sanders: I get that this is nice from the perspective of the workflow.

But we probs don't want to inject this into the tutorial?

(every change we make in the `.cylc` files has to be reflected throughout the tutorial**s**)

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/consolidate-observations
oliver-sanders: Was really added for the developers not the trainees.

But fair.

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-rainfall
```diff
@@ -160,37 +156,66 @@ def process_rainfall_data(filename, resolution, domain):
...
-    image = Image.open(filename)
-    scale = get_scale(domain, image.width)
-    offset = get_offset(domain, scale)
+    _height, width = data.shape
+
+    scale = get_scale(domain, width)
+    # TODO Fix Mercator - the new data is in a totally 
+    # different projection, transverse Mercator, rather
+    # than Mercator.
+    # offset = get_offset(domain, scale)
```
oliver-sanders: Floating TODO.

Either fix it or open an issue and link it here.

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-rainfall
```diff
@@ -160,37 +156,66 @@ def process_rainfall_data(filename, resolution, domain):
...
     time = os.environ['CYLC_TASK_CYCLE_POINT']
     resolution = float(os.environ['RESOLUTION'])
     domain = util.parse_domain(os.environ['DOMAIN'])
-    api_key = os.environ.get('API_KEY')
 
-    if api_key:
-        print('Attempting to get weather data from the DataPoint service.')
-        get_datapoint_radar_image('rainfall-radar.png', time, api_key)
+    # Acts as a switch - if a file-name is given, use that file-name
+    # TODO - keep the nice S3 formatting with implied metadata
```
oliver-sanders: Ditto.

If we *should* do this, open an issue and link it in.

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-rainfall
wxtim:
```suggestion
```

cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/get-rainfall
wxtim:
```suggestion
```

cylc/flow/etc/tutorial/cylc-forecasting-workflow/flow.cylc
wxtim:
```suggestion
```

tests/functional/cylc-cat-log/14-auto-mode.t
oliver-sanders, line 1: Might need to delete some commits post base change.

oliver-sanders:
Looking good, assigned @ChrisPaulBennett for a quick skim.

Note, we will need followup PRs in:
* cylc-doc
* metomi-rose

tests/functional/cylc-cat-log/14-auto-mode.t
oliver-sanders: @wxtim, soz, this branch still has 8.7.x commits post rebase.

setup.cfg
```diff
@@ -121,7 +121,7 @@ tests =
     types-Jinja2>=0.1.3
     types-protobuf>=0.1.10,!=5.29.1.20250402
 tutorials =
-    pillow
+    h5py
```
ChrisPaulBennett, line 124: I had to install h5py to get the workflow to work. But I assume that when this gets deployed h5py will be apart of the cylc deployment?

PR #7048, opened 2025-10-17
Fix icp = now changing on reload

### LINKED ISSUE #7047: `initial cycle point = now` - reload causes sequences to go out of whack
### Reproducible Example

```cylc
[scheduler]
    allow implicit tasks = True
    cycle point format = CCYYMMDDThhmmssZ  # second precision for easy repro

[scheduling]
    initial cycle point = now
    [[graph]]
        R1 = cold => foo
        PT15M = foo[-PT15M] => foo
```
```
$ cylc play example --pause
```
```
INFO - Initial point: 20251017T150648Z
```
Reload:
```
$ cylc reload example
```
```
INFO - Reloading the workflow definition.
INFO - LOADING workflow parameters
...
INFO - + initial point = 2025-10-17T16:06:48+01:00
```
Small thing - the initial point has changed representation.

But more importantly you will see `foo` disappear from the GUI.

And if you try triggering the `cold` task
```
$ cylc trigger example//20251017T150648Z/cold
```
```
WARNING - Invalid cycle point for task: cold, 20251017T150648Z
WARNING - No tasks match "20251017T150648Z/cold"
```

MetRonnie:
Closes #7047

Follow-up to https://github.com/cylc/cylc-flow/pull/6316

Built upon:
- https://github.com/cylc/cylc-flow/pull/7024

**Check List**

- [x] No dependency changes
- [x] Tests are included 
- [x] Changelog entry included

### DIFF
--- a/tests/functional/cylc-combination-scripts/09-vr-icp-now.t
+++ b/tests/functional/cylc-combination-scripts/09-vr-icp-now.t
@@ -16,7 +16,8 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 #------------------------------------------------------------------------------
-# Ensure that validate step of Cylc VR cannot change the options object.
+# Ensure that validate step of `cylc vr` does not set the --icp option for the
+# restart step, as this would cause an InputError.
 # See https://github.com/cylc/cylc-flow/issues/6262
 
 . "$(dirname "$0")/test_header"
@@ -26,14 +27,13 @@ WORKFLOW_ID=$(workflow_id)
 
 cp -r "${TEST_SOURCE_DIR}/${TEST_NAME_BASE}/flow.cylc" .
 
-run_ok "${TEST_NAME_BASE}-vip" \
-    cylc vip . \
-        --workflow-name "${WORKFLOW_ID}" \
-        --no-detach \
-        --no-run-name
+run_ok "${TEST_NAME_BASE}-vip" cylc vip . \
+    --workflow-name "${WORKFLOW_ID}" \
+    --no-detach \
+    --no-run-name \
+    --mode simulation
 
 echo "# Some Comment" >> flow.cylc
 
 run_ok "${TEST_NAME_BASE}-vr" \
-    cylc vr "${WORKFLOW_ID}" \
-        --stop-cycle-point 2020-01-01T00:02Z
+    cylc vr "${WORKFLOW_ID}" --no-detach --mode simulation
--- a/tests/functional/cylc-combination-scripts/09-vr-icp-now/flow.cylc
+++ b/tests/functional/cylc-combination-scripts/09-vr-icp-now/flow.cylc
@@ -1,8 +1,11 @@
+[scheduler]
+    [[events]]
+        restart timeout = PT0S
 [scheduling]
     initial cycle point = 2020
-    stop after cycle point = 2020-01-01T00:01Z
+    final cycle point = 2020
     [[graph]]
-        PT1M = foo
+        P1Y = foo
 [runtime]
     [[foo]]
         [[[simulation]]]


cylc/flow/workflow_db_mgr.py
```diff
@@ -353,11 +353,8 @@ def put_workflow_params(self, schd: 'Scheduler') -> None:
         ])
 
         # Store raw initial cycle point in the DB.
-        value = schd.config.evaluated_icp
-        value = None if value == 'reload' else value
```
MetRonnie, line 357: Initial cycle point cannot be "reload" (https://github.com/cylc/cylc-flow/pull/4313)

changes.d/7048.fix.md
```diff
@@ -0,0 +1 @@
+Fixed `initial cycle point = now` changing on reload/restart.
```
wxtim: Fixed is ambiguous IMO

```suggestion
Stop `initial cycle point = now` changing on reload/restart.
```

cylc/flow/config.py
```diff
@@ -670,10 +685,10 @@ def prelim_process_graph(self) -> None:
             all(item in ['graph', '1', 'R1'] for item in graphdict)
         ):
             # Pure acyclic graph, assume integer cycling mode with '1' cycle
-            self.cfg['scheduling']['cycling mode'] = INTEGER_CYCLING_TYPE
             for key in ('initial cycle point', 'final cycle point'):
                 if key not in self.cfg['scheduling']:
                     self.cfg['scheduling'][key] = '1'
+            self.cfg['scheduling']['cycling mode'] = INTEGER_CYCLING_TYPE
```
wxtim, line 691: Why this change?

tests/functional/cylc-combination-scripts/09-vr-icp-now.t
```diff
@@ -26,14 +26,12 @@ WORKFLOW_ID=$(workflow_id)
 
 cp -r "${TEST_SOURCE_DIR}/${TEST_NAME_BASE}/flow.cylc" .
 
-run_ok "${TEST_NAME_BASE}-vip" \
-    cylc vip . \
-        --workflow-name "${WORKFLOW_ID}" \
-        --no-detach \
-        --no-run-name
+run_ok "${TEST_NAME_BASE}-vip" cylc vip . \
```
wxtim, line 30: I'm not sure that there's a crushing advantage in style to this being re-written one way or the other. That being the case this inflates the diff surface with non-functional changes. Must we?

tests/functional/param_expand/01-basic.t
```diff
@@ -381,9 +381,9 @@ cmp_ok '19.cylc' <<'__FLOW_CONFIG__'
     [[templates]]
         lang = %(lang)s
 [scheduling]
-    cycling mode = integer
     initial cycle point = 1
     final cycle point = 1
+    cycling mode = integer
```
wxtim, line 386: What's the benefit?

cylc/flow/config.py
MetRonnie: This is part of [Minor fix to ensure order of ICP vs cycling mode in processed config does not change on reload vs start](https://github.com/cylc/cylc-flow/pull/7048/commits/a398b6008941b75c42823e80b179d7661f50c5a9)

Due to `schd._set_workflow_params()` moving earlier on, `[scheduling]initial cycle point` will now be set in the config before the cycling mode on reload, so will appear first in flow-processed.cylc

tests/functional/cylc-combination-scripts/09-vr-icp-now.t
MetRonnie: I was confused by the inclusion of the `--stopcp` option in the `cylc vr` command below. Because that _would_ "change the options object"

tests/functional/param_expand/01-basic.t
MetRonnie: No benefit, necessary change to keep the tests passing: https://github.com/cylc/cylc-flow/pull/7048#discussion_r2523360395

tests/functional/cylc-combination-scripts/09-vr-icp-now.t
```diff
@@ -26,14 +26,12 @@ WORKFLOW_ID=$(workflow_id)
...
+    --no-detach \
+    --no-run-name \
+    --stop-cycle-point 2020
 
 echo "# Some Comment" >> flow.cylc
 
-run_ok "${TEST_NAME_BASE}-vr" \
-    cylc vr "${WORKFLOW_ID}" \
-        --stop-cycle-point 2020-01-01T00:02Z
+run_ok "${TEST_NAME_BASE}-vr" cylc vr "${WORKFLOW_ID}" --no-detach
```
oliver-sanders: There doesn't appear to be any check that the FCP is preserved here? Can this test fail?

tests/functional/param_expand/01-basic.t
oliver-sanders: Are you saying that the order these configs are defined in makes a functional difference?

MetRonnie: It makes a difference to the order they end up in `cylc config`'s output

tests/functional/cylc-combination-scripts/09-vr-icp-now.t
MetRonnie: I'm not sure what relevance the final cycle point has?

oliver-sanders: > I'm not sure what relevance the final cycle point has?

I think I had meant the ICP, however, that does raise a point as the final cycle point can be relative to the initial cycle point, e.g:

```
initial cycle point = now
final cycle point = +P1D
```


>> Can this test fail?

There doesn't appear to be a reference test here or a check of the the ICP (?), so can this test fail?

tests/functional/cylc-combination-scripts/09-vr-icp-now.t
```diff
@@ -26,14 +27,13 @@ WORKFLOW_ID=$(workflow_id)
...
-run_ok "${TEST_NAME_BASE}-vip" \
-    cylc vip . \
-        --workflow-name "${WORKFLOW_ID}" \
-        --no-detach \
-        --no-run-name
+run_ok "${TEST_NAME_BASE}-vip" cylc vip . \
+    --workflow-name "${WORKFLOW_ID}" \
+    --no-detach \
+    --no-run-name \
+    --mode simulation
```
oliver-sanders, line 34: Good idea (we should do this more in functional tests), but be aware that simulation mode sleeps for some random interval unless configured otherwise (like we do in the integration tests), so may need to be backed up with a `[simulation]default run length = PT0S` config.

wxtim: Definitely not random. If you've set both the `execution time limit` and `[simulation]speedup factor` you get the execution time limit / speedup factor, otherwise you will fall back to default run length. You should then fall back to 10 seconds - though I can't see the mechanism at a quick glance. (All timings accurate to 1 main loop cycle)

tests/functional/cylc-combination-scripts/09-vr-icp-now.t
MetRonnie: The test is checking that `cylc vr` can be called without failing due to the error in #6262

> ERROR - Workflow shutting down - InputError: option --icp is not valid for restart

tests/functional/cylc-combination-scripts/09-vr-icp-now.t
MetRonnie: Tim originally added `[simulation]default run length = PT0S` in the test, but I think he forgot to add the mode option to the command

oliver-sanders: Ach, have found the `speedup factor` in the flow.cylc file, one line outside the diff.

PR #7051, opened 2025-10-21
Avoid spurious logging on restart with waiting parentless tasks.

### LINKED ISSUE #7037: Spurious "Not respawning... task was removed" log messages
```cylc
[scheduler]
    allow implicit tasks = True
[scheduling]
    initial cycle point = 2024-01-01
    [[graph]]
        PT3H = get_observations => consolidate_observations

        +PT6H/PT6H = """
            consolidate_observations => forecast
            consolidate_observations[-PT3H] => forecast
            consolidate_observations[-PT6H] => forecast
            get_rainfall => forecast => post_process
        """

        +PT12H/PT6H = """
            forecast[-PT6H] => forecast
        """
```

```console
$ cylc play simple-datetime --pause
...
$ cylc stop simple-datetime
...
$ cylc play simple-datetime --pause --no-detach
...
INFO - LOADING task proxies
INFO - + 20240101T0000+0100/get_observations waiting
INFO - [20240101T0000+0100/get_observations:waiting(runahead)] => waiting
INFO - [20240101T0000+0100/get_observations:waiting] => waiting(queued)
INFO - Not respawning 20240101T0300+0100/get_observations - task was removed
INFO - Not respawning 20240101T0600+0100/get_observations - task was removed
INFO - Not respawning 20240101T0900+0100/get_observations - task was removed
INFO - Not respawning 20240101T1200+0100/get_observations - task was removed
INFO - Not respawning 20240101T1500+0100/get_observations - task was removed
INFO - + 20240101T0300+0100/get_observations waiting
INFO - [20240101T0300+0100/get_observations:waiting(runahead)] => waiting
INFO - [20240101T0300+0100/get_observations:waiting] => waiting(queued)
...
```

hjoliver:
Close #7037 

If auto-spawned parentless tasks are present at shutdown, on restart they will be resurrected from the DB; **and** auto-spawned to the runahead limit *again* - because we don't record whether or not parentless tasks were spawned already.

This double-spawning is prevented (by luck - [see footnote]) by the imperfect bodge that stops suicided tasks from being resurrected by other dependencies - which has no functional effect, but it results in misleading log messages.

Said imperfect bodge was fixed by re-implementing suicide triggers as expire triggers (#6835) - but for now that's experimental.

Solution implemented here:
- if the experimental expire trigger feature is turned on on:
   - don't run the bodge (this should have been done in #6835)
   - (and handle the parentless-waiting-task-at-restart problem by simply not adding tasks to the pool if already added)
- if the expire trigger feature is not turned on:
   - do run the bodge (it's still needed for suicide triggers, despite being imperfect)
   - but demote the log message to DEBUG and change it slightly to work for both cases (suicide and parentless)

-----

[footnote] - if the double-spawning had not been prevented by the bodge I think the new instance would just replace the old in the task pool, which would actually be fine because the task had not done anything yet (however, better not to do it).

**Check List**

### DIFF
--- a/tests/integration/test_task_pool.py
+++ b/tests/integration/test_task_pool.py
@@ -2583,15 +2583,7 @@ async def test_add_to_pool(
     flow, scheduler, start, caplog
 ):
     """It should log attempts to add the same task again."""
-    id_ = flow(
-        {
-            'scheduling': {
-                'graph': {
-                    'R1': 'a',
-                },
-            },
-        }
-    )
+    id_ = flow('a')
     schd = scheduler(id_)
 
     async with start(schd):


cylc/flow/task_pool.py
```diff
@@ -266,12 +266,22 @@ def db_add_new_flow_rows(self, itask: TaskProxy) -> None:
...
+            # will be resurrected from the DB again *and* try to auto-spawn
+            # again because we don't record that parentless tasks spawned.
+            # (But debug-log it anyway for any unexpected cases).
+            LOG.debug(f"{itask.identity} not added to n=0: already exists")
+            return None
+
         self.active_tasks[itask.point][itask.identity] = itask
         self.active_tasks_changed = True
-        LOG.debug(f"[{itask}] added to the n=0 window")
+        LOG.info(f"[{itask}] added to the n=0 window")
```
hjoliver, line 284: Deliberate change - it's important for users to know when tasks enter the n=0 window (otherwise it only gets logged as INFO at the first state change).

tests/integration/test_task_pool.py
```diff
@@ -2576,3 +2577,31 @@ def list_spawn_task_calls():
...
+    """It should log attempts to add the same task again."""
+    id_ = flow(
+        {
+            'scheduling': {
+                'graph': {
+                    'R1': 'a',
+                },
+            },
+        }
+    )
```
MetRonnie: Can make this shorter
```suggestion
    id_ = flow('a')
```

cylc/flow/task_pool.py
MetRonnie: I don't see why this is useful info to have in the logs

oliver-sanders: @hjoliver poke (though I suspect we'll bump this to 8.6.2).

I'm with Ronnie, this isn't `INFO` material, the logs are crazy enough as it is!

hjoliver: IMO it's good to know when tasks enter and leave active window, because it helps users to understand how the system works, and understanding how the system works helps when unexpected things happen.

In this case, tasks entering the active window when their first prerequisite gets satisfied is a pretty important concept. Of course I know that *most* users don't care to understand the system, but (a) they'd be better off if they did; and (b) at least a few do; and (c) we should make the evolution of workflows as clear as possible anyway, in case anyone does try to follow it.

This is just one small aspect of that discussion though - if I'm outvoted so be it.

hjoliver: (Omitting this change from the superseding PR).

hjoliver:
Closing as superseded by #7135

Issue #7001, opened 2025-09-22
Poll: polling is not attempted if the platform has run out of hosts


wxtim:
### Description

* If we experience issues contacting a host, the host is added to bad hosts.
* Polling will purposefully exclude bad hosts.
* So if there are no good hosts left, each poll attempt will in effect be skipped.
* The bad hosts are reset every `PT3M` (default) so it will take a long time before the polling resumes.

This also applies to other IHS (intelligent host select) operations such as job log retrieval.

### Reproducible Example

Using this workflow

```cylc
[scheduling]
    cycling mode = integer
    [[graph]]
        R1 = pbs:started => breaker
[runtime]
    [[pbs]]
        platform = remote
        script = sleep 1h
        execution polling intervals = PT15S
    [[breaker]]
        # Breaks your ssh config
        script = mv ~/.ssh/config ~/.ssh/config.bk
```

* Wait for polling to report your platform unavailable.
* Fix your ssh config.
* Monitor as polling continues, without success.
* Try killing the task (it doesn't have any effect)

Similarly, when one starts the workflow, waits for the task submission then `cylc stop --now`, then breaks one's ssh and restarts the workflow the initial polling doesn't reset the list of bad host after the attempt to poll finishes...

oliver-sanders:
If a host is detected as uncontactable, it is added to bad hosts. It will then be omitted for use by subsequent commands.

The bad hosts list is reset:

* Every `PT30M` (configurable).
* On manual submission if there are no good hosts to submit to.

(as per prior proposals)

The idea behind clearing the bad hosts every `PT30M` is to allow us to detect resurrected hosts and avoid the situation this bug reports. However `PT30M` is a long time to wait and this approach is somewhat naive.

It has been suggested that other commands (e.g poll) might also reset bad hosts if there are no good hosts left. However, this is as it s can cause resonance between operations since this state is global.

---

### Proposal (a rethink of the "bad host" logic)

The "bad hosts" set contains valuable information, just blindly wiping it every `PT30M` seems silly.

Rather than **reset**ting bad hosts at a configured time interval, why not **test** them instead.

This proposal is to refactor `reset bad hosts` into `check bad hosts`:

* Every so many minutes, the plugin would "check" the first "N" bad hosts.
  * Only check the first N hosts to avoid flooding the system with SSH calls.
* It would do this by running an SSH command, if the command returns 0, then the host would be removed from bad hosts.
  * `ssh true` should be enough/
[...]

PR #7055, opened 2025-10-22
polling: reset bad hosts on manual poll

### LINKED ISSUE #7029: polling: reset bad hosts on *manual* poll
If there are no good hosts left for a platform and a task is manually triggered, the bad hosts are reset.

I.E, if the user manually triggered a task, we should attempt a submission, even if we know it is likely to fail.

This is designed to allow users to quickly "unstick" a workflow in the event of resolved platform issues.

We should extend this to manual poll.

Note, we cannot extend this to automated actions as the resetting of bad hosts by one command can interact with another remote operation running in parallel, leading to an infinite loop. Manual operations, however, are safe (we already do this for manual trigger).

Takes the sting out of #7001, the remainder becomes a low priority.

oliver-sanders:
* Closes #7029
* This reduces the pressure on #7001 (polling is not attempted if a platform runs out of hosts until the next run of reset-bad hosts) by making it easier for operators to reset bad hosts and recover their workflow in the event of platform outages.

**Check List**

### DIFF
--- a/cylc/flow/task_job_mgr.py
+++ b/cylc/flow/task_job_mgr.py
@@ -223,7 +223,10 @@ class TaskJobManager:
         self._prep_submit_task_job_error(itask, '(killed in job prep)', '')
 
     def poll_task_jobs(
-        self, itasks: 'Iterable[TaskProxy]', msg: str | None = None
+        self,
+        itasks: 'Iterable[TaskProxy]',
+        msg: str | None = None,
+        manual_request: bool = False,
     ):
         """Poll jobs of specified tasks.
 
@@ -245,7 +248,8 @@ class TaskJobManager:
                     if itask.state.status != TASK_STATUS_WAITING
                 ],
                 self._poll_task_jobs_callback,
-                self._poll_task_jobs_callback_255
+                self._poll_task_jobs_callback_255,
+                continue_if_no_good_hosts=manual_request,
             )
 
     def prep_submit_task_jobs(
@@ -918,12 +922,33 @@ class TaskJobManager:
         log_task_job_activity(ctx, self.workflow, itask.point, itask.tdef.name)
 
     def _run_job_cmd(
-        self, cmd_key, itasks, callback, callback_255
+        self,
+        cmd_key,
+        itasks,
+        callback,
+        callback_255,
+        continue_if_no_good_hosts=False,
     ):
         """Run job commands, e.g. poll, kill, etc.
 
         Group itasks with their platform_name and host.
         Put a job command for each group to the multiprocess pool.
+
+        Args:
+            cmd_key:
+                Identifier for the command to run.
+            itasks:
+                List of task proxies to run the command against.
+            callback:
+                Callback to run on command completion.
+            callback_255:
+                Callback to run on SSH error.
+            continue_if_no_good_hosts:
+                If True, the bad hosts set will be reset in the event that
+                there are no "good hosts" to run the command on. This should
+                only be turned on for manual operations, e.g. manual poll. Use
+                of this option for automatic commands may result in feedback
+                loops between parallel opererations.
 
         """
         if not itasks:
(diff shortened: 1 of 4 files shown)


cylc/flow/task_job_mgr.py
```diff
@@ -968,15 +993,22 @@ def _run_job_cmd(
...
+                        # -> reset the bad hosts and try again
+                        self.task_events_mgr.reset_bad_hosts()
+                        host = get_host_from_platform(
+                            platform, bad_hosts=self.bad_hosts
+                        )
+                    else:
+                        ctx.err = f'No available hosts for {platform["name"]}'
+                        LOG.debug(ctx)
+                        callback_255(ctx, itasks)
+                        continue
```
MetRonnie: Both branches here continue to the next iteration of the `for` loop. `host` is not used in the first branch. Did you mean to remove the `else` condition on the line below?

cylc/flow/task_job_mgr.py
```diff
@@ -968,15 +993,22 @@ def _run_job_cmd(
...
-                    )
                 except NoHostsError:
-                    ctx.err = f'No available hosts for {platform["name"]}'
-                    LOG.debug(ctx)
-                    callback_255(ctx, itasks)
-                    continue
+                    if continue_if_no_good_hosts:
+                        # no hosts available for this platform
+                        # -> reset the bad hosts and try again
+                        self.task_events_mgr.reset_bad_hosts()
```
wxtim: Could you ignore rather than clear?

cylc/flow/task_job_mgr.py
wxtim: Mildly concerned that if Ronnie is right (I think he is) that the test isn't failing....

cylc/flow/task_job_mgr.py
oliver-sanders: Yes, but that would only apply to this one operation. Clearing is the mechanism which ensures all hosts are retried (same thing we do for trigger).

MetRonnie:
Converting to draft pending resolution of https://github.com/cylc/cylc-flow/pull/7055#discussion_r2490318207

cylc/flow/task_job_mgr.py
oliver-sanders: Ronnie is very much correct.

The test was passing because of the `host = 'localhost'` above which was providing an unwelcome default pathway through the code.

MB, that was some AI level BS.

cylc/flow/task_job_mgr.py
```diff
@@ -960,36 +985,48 @@ def _run_job_cmd(
             cmd.append("--")
             cmd.append(get_remote_workflow_run_job_dir(self.workflow))
             job_log_dirs = []
-            host = 'localhost'
```
oliver-sanders, line 963: Have removed the `localhost` default.

cylc/flow/task_job_mgr.py
```diff
@@ -960,36 +985,48 @@ def _run_job_cmd(
...
-            if remote_mode:
-                try:
+            # select a host to run the operation on
+            host = None
+            with suppress(NoHostsError):
+                host = get_host_from_platform(
+                    platform, bad_hosts=self.bad_hosts
+                )
+
+            if not host and continue_if_no_good_hosts:
```
oliver-sanders, line 996: Have flattened the `if/try/except/if/else` pathway.

cylc/flow/task_job_mgr.py
```diff
@@ -960,36 +985,48 @@ def _run_job_cmd(
...
 
-            ctx = SubProcContext(cmd_key, cmd, host=host)
-            if remote_mode:
-                try:
+            # select a host to run the operation on
+            host = None
+            with suppress(NoHostsError):
+                host = get_host_from_platform(
+                    platform, bad_hosts=self.bad_hosts
+                )
```
oliver-sanders, line 994: Have pushed localhost submissions through this interface to flatten the code.

This is a change, though `localhost` can't get into the bad hosts, so functionless. Happy to change if desired.

tests/integration/test_task_job_mgr.py
```diff
@@ -260,6 +264,67 @@ async def test_poll_job_deleted_log_folder(
...
+        await run_cmd(poll_tasks(schd, [itask.tokens.relative_id]))
+
+        # the bad hosts set should be empty
+        assert not schd.task_events_mgr.bad_hosts
+
+        # and a poll command should have been issued
+        assert len(schd.proc_pool.queuings) == 1
+        poll_ctx = schd.proc_pool.queuings.pop()[0]
+        assert poll_ctx.cmd_key == TaskJobManager.JOBS_POLL
+        assert poll_ctx.host == 'abc'
```
oliver-sanders, line 325: Test hardened.

PR #7056, opened 2025-10-22
expire: fix interaction with automatic retries

### LINKED ISSUE #6717: expire: bad interaction with automatic retries
If a task expires before automatic retry, the task gets left in the waiting state and Cylc enters a polling loop.

### Reproducible Example

```cylc
[scheduling]
  initial cycle point = now
  [[special tasks]]
    clock-expire = foo(PT1M)
  [[graph]]
    R1 = foo

[runtime]
  [[foo]]
    script = false
    execution retry delays = PT2M
```

Example output:

```
INFO - [20250408T0814Z/foo/01:preparing] => submitted
INFO - [20250408T0814Z/foo/01:submitted] => running
ERROR - [20250408T0814Z/foo/01:running] failed/ERR
INFO - [20250408T0814Z/foo/01:running] => waiting
WARNING - [20250408T0814Z/foo:waiting] - retrying in PT2M (after 2025-04-08T09:16:13+01:00)
WARNING - [20250408T0814Z/foo:waiting] (polled-ignored)expired
WARNING - [20250408T0814Z/foo:waiting] (polled-ignored)expired
WARNING - [20250408T0814Z/foo:waiting] (polled-ignored)expired
WARNING - [20250408T0814Z/foo:waiting] (polled-ignored)expired
WARNING - [20250408T0814Z/foo:waiting] (polled-ignored)expired
```

### Expected Behaviour

The task should enter the expired state and retries should be cancelled.

Note, setting a task to a final state (i.e. succeeded, failed, submit-failed, expired) should cancel any scheduled retries ([see proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-interventions.md#5-i-need-to-terminate-a-chain-of-automatic-retries)).

oliver-sanders:
* Closes #6717
* Blocks #7175

(Note, this issue needs to be resolved before expire triggers can be used in anger)

**Check List**

### DIFF
--- a/tests/integration/test_task_pool.py
+++ b/tests/integration/test_task_pool.py
@@ -2323,7 +2323,7 @@ async def test_expire_dequeue_with_retries(
     }
 
     if expire_type == 'clock-expire':
-        # configre foo to clock-expire
+        # configure foo to clock-expire
         conf['scheduling']['special tasks'] = {'clock-expire': 'foo(PT0S)'}
 
         # run the clock-expire logic


cylc/flow/task_pool.py
```diff
@@ -2440,6 +2440,7 @@ def clock_expire_tasks(self):
                     itask,
                     logging.WARNING,
                     TASK_OUTPUT_EXPIRED,
+                    forced=True,
```
oliver-sanders: @hjoliver, is this a valid use of the `force` argument?

Looks like the tutorial workflow tests failed as this change removed the `(internal)` prefix from the expired log message.

hjoliver: I think it's valid, but maybe update the docstring in the message to say that expiry counts as forcing too.

Do we also need this for suicide-induced task expiry? (See line 1565)

oliver-sanders: The message currently reads:

https://github.com/cylc/cylc-flow/blob/89efa74603224f674fefa1ae3a6e4128c7787bcf/cylc/flow/task_events_mgr.py#L720-L721

So, should this be:

```
If this message was set for internal (i.e, artificial) as opposed to natural satisfaction.
```

If so, are there any other situations we should be using this in?

oliver-sanders:
(rebased)

cylc/flow/task_pool.py
oliver-sanders: > Do we also need this for suicide-induced task expiry? (See line 1565)

I think we do. I've had a go with this example:

```cylc
[scheduler]
    [[experimental]]
        expire triggers = True

[scheduling]
    [[graph]]
        R1 = """
            foo:x => bar => baz
            foo:x & bar => !foo
        """

[runtime]
    [[foo]]
        script = """
            cylc message -- xxx
            false
        """
        # execution retry delays = PT1M
        [[[outputs]]]
            x = xxx
    [[bar]]
        script = """
            cylc__job__poll_grep_workflow_log 'foo.*failed'
        """
    [[baz]]
        script = sleep 60
```

Which seems to behave differently when the `execution retry delays` are turned on.

cylc/flow/etc/examples/expiry/.validate
```diff
@@ -30,7 +30,7 @@ test_one () {
         ./one
 
     # the start task should have expired
-    grep 'start.*(internal)expired' "$HOME/cylc-run/$ID/log/scheduler/log"
```
oliver-sanders, line 33: The fix (`force=True`) causes this change in message.

cylc/flow/task_events_mgr.py
```diff
@@ -718,7 +718,8 @@ def process_message(
                 The submit number of the task relevant for the message.
                 If not specified, use latest submit number.
             forced:
-                If this message is due to manual completion or not (cylc set)
+                True, if this message is due to manual (as opposed to natural)
```
wxtim: I'd suggest that you give some examples of cases where it'd be true or false. Cylc set was the primary example of "forced" before.

oliver-sanders: See above discussion.

tests/integration/test_task_pool.py
```diff
@@ -2298,48 +2301,91 @@ async def test_expire_dequeue_with_retries(
...
             },
         },
         'runtime': {
             'foo': {
-                'execution retry delays': 'PT0S'
-            }
-        }
+                'execution retry delays': 'PT0S',
+                'outputs': {
+                    'x': 'xxx',
```
wxtim, line 2319: Suggest that tests involving outputs are safer if `key not in value`
```suggestion
                    'x': 'yyy',
```

tests/integration/test_task_pool.py
```diff
@@ -2298,48 +2301,91 @@ async def test_expire_dequeue_with_retries(
...
+                },
+            },
+        },
     }
 
     if expire_type == 'clock-expire':
+        # configre foo to clock-expire
```
wxtim:
```suggestion
        # configure foo to clock-expire
```

tests/integration/test_task_pool.py
oliver-sanders: Outputs must be an exact match, so that's not any safer.

cylc/flow/task_events_mgr.py
wxtim: I'm not sure that "manual (as opposed to natural)" is particularly helpful

cylc/flow/task_pool.py
wxtim: Are you going to fix this in this PR or spin it out into a new issue?

cylc/flow/task_events_mgr.py
hjoliver: *"manual (as opposed to natural) output completion"* seems OK to me. Natural output completion means "in the course of normal automatic workflow evolution".  I suppose we could say "automatic" instead.

hjoliver: So "manual" means - I hope - any manual intervention.  This is only a docstring, so we could just be more verbose to avoid any confusion.

E..g *True if this message results from any manual intervention, as opposed to automatic workflow evolution.*

hjoliver: (Of course clock expiry is not "manual" so this PR muddies those waters somewhat).

cylc/flow/task_pool.py
oliver-sanders: Have already fixed in this issue, have added clock-expiry to the integration test.

Had left this question open as haven't quite got to the bottom of the purpose of the `force` argument.

Since you've opened another thread for that below, I'll resolve this as duplicated by https://github.com/cylc/cylc-flow/pull/7056#discussion_r2846891761

hjoliver:
The only thing holding me back here is I want to follow up on tracing the "forced" argument ... top of the todo list.

cylc/flow/task_events_mgr.py
oliver-sanders: "manual" => "automated"?

hjoliver:
OK, on master, the `forced` argument in `process_task_message()` only comes from the `set` command:

On manually setting an output, we call `process_task_message()` just like a naturally completed output would, so that all the normal consequences of completing an output follow; but we add the `forced` arg to indicate that it was not natural completion, because a few of those consequences need to be different in the manual case.

hjoliver (changes requested):
Given my analysis of the problem above, I think ideally `forced=True` should be reserved for manual output completion (i.e. `cylc set`) and a better fix for the bug is to tweak `task_message_check()` to stop it ignoring natural expired messages for tasks waiting on retries. 

I've checked this diff on master works:
```diff
diff --git a/cylc/flow/task_events_mgr.py b/cylc/flow/task_events_mgr.py
index 91d0ec0c6..44e82d8b4 100644
--- a/cylc/flow/task_events_mgr.py
+++ b/cylc/flow/task_events_mgr.py
@@ -932,6 +932,7 @@ class TaskEventsManager():

         if (
             itask.state(TASK_STATUS_WAITING)
+            and message != TASK_OUTPUT_EXPIRED
             # Polling in live mode only:
             and itask.run_mode == RunMode.LIVE
             and (
```
The comment in this code block is:
```
            # Ignore messages if task has a retry lined up
            # (caused by polling overlapping with task failure)
```
I guess that means e.g. the task polls as running, then fails and goes to waiting on retry before the poll result comes in, in which case the poll result should be ignored.

That scenario doesn't apply to automatic expiry, which is not like a delayed poll result.

MetRonnie (approved):

PR #7058, opened 2025-10-23
Turn off flake8-simplify for Python≥3.14

wxtim:
Backport of #7050

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,6 +16,9 @@ ignore=
     SIM9
     ; explicitly concatenated strings (flake8-implicit-str-concat)
     ISC003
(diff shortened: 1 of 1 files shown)


oliver-sanders:
The change looks ok, but `flake8` does not appear to be happy?

oliver-sanders:
Some our exceptions don't appear to call `super().__init__` which is bad, however, adding this doesn't make B042 go away.

I think (once the `__init__` stuff is fixed), these are false positives.

oliver-sanders:
The B042 stuff is tangential to this PR, but the tests ain't gonna pass without it.

@wxtim, suggest disabling this check here. I'll raise a PR against master to add in the missing `__init__` calls.

PR #7075, opened 2025-11-12
focus family and workflow state totals about n=0 tasks

### LINKED ISSUE #7070: data store: task state totals should only represent n=0
This is related to the redesign/rethink of the workflows sidebar in cylc-ui. See https://github.com/cylc/cylc-ui/pull/2331, https://github.com/cylc/cylc-ui/issues/2350

Workflow task state counts are currently computed for the whole n-window. We think it would make more sense for the workflow state totals to only represent n=0 tasks. This way:

* Changing the n-window doesn't change the state count.
* Failed/submit-failed, but completed tasks are not included in the failed/submit-failed counts.
* Makes it easier to support is_held and is_retry counts (note, historical tasks may be held, but displaying this might not be considered helpful).

As this is data-store only, we feel this is something we can put into a bugfix release.

dwsutherland:
closes #7070

From element
> David Sutherland
12:30
> The only trouble states are failed/submit-failed and succeeded (which would be a pointless total with this change, as it would always be zero)..

> David Sutherland
13:14
The other question would be; what do we do with family states? (which selects the most important state of it's tasks)
Should they represent n=0 too?

> Maybe represent n=0 state if an active task(s) exist, otherwise n>0

This is what's done here, n=0 tasks have precedence with family state totals, workflow state totals.

n0-state-totals

**Check List**

- [ ] Covered by existing tests.

### DIFF
--- a/tests/integration/test_data_store_mgr.py
+++ b/tests/integration/test_data_store_mgr.py
@@ -407,8 +407,6 @@ async def test_update_data_structure(mod_harness):
     assert TASK_STATUS_FAILED in set(collect_states(data, TASK_PROXIES))
     # family state changed and applied
     assert TASK_STATUS_FAILED in set(collect_states(data, FAMILY_PROXIES))
-    # state totals changed
-    assert TASK_STATUS_FAILED in data[WORKFLOW].state_totals
 
 
 async def test_prune_data_store(flow, scheduler, start):


dwsutherland:
> > I appreciate that the aggregate family states are awkwardly related to the top-level workflow state counts due to the implementation which makes this difference a bit fun :(
> 
> Yeah, this.. I think I'm gonna need another field, or internal dict, if I can't use the same.

Actually, I think it's fine the way it is implemented here..
Because, let's think about this:
- if the family contains only n>0 states (child families & tasks) then it will be the aggregate of all states
- if the family contains only n=0 states, then it will be an aggregate of those states
- if the family contains a mix of n=0 and n>0, then the n=0 states are represented and already have higher priority in `extract_group_state`, 
  ```
    ordered_states = [
        TASK_STATUS_SUBMIT_FAILED,
        TASK_STATUS_FAILED,
        TASK_STATUS_EXPIRED,
        TASK_STATUS_RUNNING,
        TASK_STATUS_SUBMITTED,
        TASK_STATUS_PREPARING,
        TASK_STATUS_WAITING,
        TASK_STATUS_SUCCEEDED
    ]

  ```
  with the exception/inconsistency being handled `failed`/`submit-failed` that will represent n>0 families if present. i.e.

image

but I think this is acceptable.. Because all it says is n=0 is higher priority than n>0 (which includes handled failures), and then within those `ordered_states` is followed.

hjoliver:
(I'm planning to have a close look at this one tomorrow...)

cylc/flow/data_store_mgr.py
```diff
@@ -2195,36 +2195,58 @@ def _family_ascent_point_update(self, fp_id):
...
+                    graph_depth = tp_depth.graph_depth
+
+                if tp_id in self.all_task_pool:
+                    is_active = True
+
                 tp_state = self.from_delta_or_node(tp_delta, tp_node, 'state')
                 if tp_state:
                     task_states.append(tp_state)
+                    if is_active:
+                        active_states.append(tp_state)
```
oliver-sanders: I think we can do `active_counter[tp_state] += 1` to avoid creating a list of N states.

cylc/flow/data_store_mgr.py
```diff
@@ -2195,36 +2195,58 @@ def _family_ascent_point_update(self, fp_id):
             tp_added = self.added[TASK_PROXIES]
             # Count child family states, set is_held, is_queued, is_runahead
             state_counter = Counter({})
+            active_counter = Counter({})
             is_held_total = 0
             is_queued_total = 0
             is_runahead_total = 0
```
oliver-sanders, line 2205: With this change the `is_held_total` has changed to reflect n=0 as needed.

However, the `is_queued_total` and `is_runahead_total` are presumably still n=x which might become confusing in the future.

I don't think we have any use case for these at present, so I think we can just take them out. Otherwise, we could make them follow the `is_held_total` and stick to n=0.

changes.d/7075.feat.md
```diff
@@ -0,0 +1 @@
+State totals shown in gscan/tui now represent the active (n=0) window of tasks.
```
oliver-sanders:
```suggestion
State totals shown in GUI/Tui now represent the active (n=0) window of tasks.
```

(The GScan sidebar is now called the "Workflows View" or the "Workflows Sidebar", GScan is no longer a user-facing term)

cylc/flow/data_store_mgr.py
dwsutherland: Have refactored this section.. also taken out the `if tp_state` as the state should always be available (given that this family update occurs post DB history application)

cylc/flow/data_store_mgr.py
dwsutherland: I've implemented the latter, all represent `n=0`

changes.d/7075.feat.md
dwsutherland: Yes, I meant gscan view in the WUI.. So, also sorry for the confusion.

hjoliver (approved):
LGTM - but could you put comments in the code to explain the state total meanings:
- task totals n=0 only (these are the important states for monitoring at full-workflow summary)
- family totals n-window (so that family totals in workflow views - e.g. treeview - reflect the members shown in the view)

cylc/flow/data_store_mgr.py
```diff
@@ -2168,6 +2168,10 @@ def _family_ascent_point_update(self, fp_id):
         first called with this function, which then adds it's first parent
         ancestor to the set of families flagged for update.
 
+        State totals of families reflect zero n-window (n=0), if no n=0
+        children (tasks/families) exist then the totals include the states of
+        all children. Family group state, however, is determined from all child
+        states via the n>=0 state totals.
         """
```
wxtim: Possibly easier to read?

```suggestion
        If any tasks in a family are in the zero n-window (n=0) then
        state totals reflect the n=0 window.
        If a family has no n=0 children (tasks/families) then the state 
        totals include the states of all children.
        
        Family group state, however, is determined from all child
        states via the n>=0 state totals.
        """
```

cylc/flow/data_store_mgr.py
```diff
@@ -2168,6 +2168,10 @@ def _family_ascent_point_update(self, fp_id):
         first called with this function, which then adds it's first parent
         ancestor to the set of families flagged for update.
 
+        State totals of families reflect zero n-window (n=0), if no n=0
+        children (tasks/families) exist then the totals include the states of
+        all children. Family group state, however, is determined from all child
+        states via the n>=0 state totals.
```
oliver-sanders: This works, but it's a bit confusing having the state totals mean different things depending on whether the family is n=0 or not.

We don't actually need to compute the parent family state from the state totals as we can derive it from the child states contained within it instead.

I put together a small change which:
* Makes family state totals reflect n=0 in all cases.
* Computes the parent family state from the child states.

https://github.com/dwsutherland/cylc-flow/pull/25

dwsutherland: > Computes the parent family state from the child states.

We couldn't do this before, because child states weren't homogenous (WRT what they represented).. but can now.. merged

wxtim (approved):

PR #7077, opened 2025-11-12
developer: add a configuration for turning lru_caching off

oliver-sanders:
In order to perform memory profiling for Cylc it is generally necessary to disable the `lru_cache` mechanism that we use to speedup datetime / recurrence operations.

Let's make this configurable via an environment variable rather than performing this hack every time.

**Check List**

### DIFF
--- a/cylc/flow/cycling/iso8601.py
+++ b/cylc/flow/cycling/iso8601.py
@@ -18,33 +18,46 @@
 
 import contextlib
 from functools import lru_cache
+import os
 import re
-from typing import List, Optional, TYPE_CHECKING, Tuple
+from typing import TYPE_CHECKING, List, Optional, Tuple
 
-from metomi.isodatetime.data import Calendar, CALENDAR, Duration
+from metomi.isodatetime.data import CALENDAR, Calendar, Duration
 from metomi.isodatetime.dumpers import TimePointDumper
-from metomi.isodatetime.timezone import (
-    get_local_time_zone, get_local_time_zone_format, TimeZoneFormatMode)
 from metomi.isodatetime.exceptions import IsodatetimeError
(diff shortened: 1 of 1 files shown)


cylc/flow/cycling/iso8601.py
```diff
@@ -57,6 +70,17 @@
...
 
+# NOTE: We cache some datetime cycling operations to improve compute
+# perforance. For profiling, this can be disabled by setting the environment
+# variable CYLC_CYCLER_LRU_CACHE_SIZE=0.
+
+# The number of cycling operations to cache:
+_LRU_CACHE_SIZE = int(os.environ.get('CYLC_CYCLER_LRU_CACHE_SIZE', '10000'))
+
+# A smaller cache for use with larger objecs (to reduce memory impact):
+_LARGE_LRU_CACHE_SIZE = int(_LRU_CACHE_SIZE / 100) if _LRU_CACHE_SIZE else 0
```
ChrisPaulBennett, line 81: Why is it called `_LARGE_LRU_CACHE_SIZE` when it is the "smaller" cache?

oliver-sanders: For larger objects.

Happy to accept a name suggestion.

cylc/flow/cycling/iso8601.py
```diff
@@ -57,6 +70,17 @@
     "(incompatible with [cylc]cycle point num expanded year digits = %s ?)")
 
 
+# NOTE: We cache some datetime cycling operations to improve compute
+# perforance. For profiling, this can be disabled by setting the environment
```
wxtim, line 74:
```suggestion
# performance. For profiling, this can be disabled by setting the environment
```

PR #7081, opened 2025-11-12
zmq client: cleanup sockets and contexts

oliver-sanders:
* Resolves Python warnings about unclosed sockets and contexts.
* Particularly important for commands which repeatably open clients such as `cylc scan --ping` and `cylc gui`.

To see the warning for yourself, try:

```
$ PYTHONWARNINGS=all cylc scan --ping # with at least one workflow running
```

It is possible that this may be a memory leak for Cylc GUI / Tui processes, although I can't confirm that, tracemalloc did identify some zmq entries during investigation of https://github.com/cylc/cylc-uiserver/issues/732, so it's possible.

I'm uncertain over whether this should go into a bugfix or not. It is a bugfix, however, it's not clear if this is causing issues ATM and the diff has the potential to cause issues if not tested thoroughly.

**Check List**

### DIFF
--- a/cylc/flow/network/client.py
+++ b/cylc/flow/network/client.py
@@ -401,3 +401,6 @@ class WorkflowRuntimeClient(  # type: ignore[misc]
                     )
             }
         }
+
+    def __del__(self):
+        self.stop(stop_loop=False)
--- a/cylc/flow/network/__init__.py
+++ b/cylc/flow/network/__init__.py
@@ -165,8 +165,12 @@ class ZMQSocketBase:
         self.bind = bind
         if context is None:
             self.context: 'Context' = zmq.asyncio.Context()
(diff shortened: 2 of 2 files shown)


cylc/flow/network/client.py
```diff
@@ -188,6 +188,9 @@ def timeout_handler(self) -> None:
             # the workflow has stopped
             raise WorkflowStopped(self.workflow)
 
+    def __del__(self):
+        self.stop(stop_loop=False)
```
MetRonnie: A `stop()` method is not defined here in `WorkflowRuntimeClientBase`. Why not move this `__del__()` method to `ZMQBase`, or `WorkflowRuntimeClient`?

oliver-sanders: Done.

PR #7082, opened 2025-11-12
daemonize: fix unclosed file handle

oliver-sanders:
Fix an unclosed file handle by addressing an ancient `TODO`.

The unclosed file handle is reported as a Python warning, try running:

```console
$ PYTHONWARNINGS=all cylc vip
```

**Check List**

- [x] Changelog entry included if this is a change that can affect users - no symptoms observed thus far

### DIFF
--- a/cylc/flow/daemonize.py
+++ b/cylc/flow/daemonize.py
@@ -16,6 +16,7 @@
 
 """Turn a cylc scheduler into a Unix daemon."""
 
+import atexit
 import json
 import os
 import sys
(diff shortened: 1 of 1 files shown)


cylc/flow/daemonize.py
```diff
@@ -129,6 +130,6 @@ def daemonize(schd):
     # Note that simply reassigning the sys streams is not sufficient
     # if we import modules that write to stdin and stdout from C
     # code - evidently the subprocess module is in this category!
-    # TODO: close resource? atexit?
```
oliver-sanders, line 132: Yes! Good idea!

MetRonnie (approved):
TIL about https://docs.python.org/3/library/atexit.html

hjoliver (approved):

PR #7084, opened 2025-11-13
Coverage: set core to ctrace to avoid performance issues with sysmon

MetRonnie:
Further investigation in https://github.com/coveragepy/coveragepy/issues/2082 has revealed that we can remove the upper pin on coverage as long as we configure it to use ctrace 

**Check List**

- [x] Dependency changes do not apply to conda file
- [x] No tests etc. needed

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -19,7 +19,8 @@ omit =
     cylc/flow/scripts/report_timings.py
 parallel = True
 source = ./cylc
-timid = False
+# https://github.com/coveragepy/coveragepy/issues/2082:
+core = ctrace
 
 [report]
 exclude_lines =
--- a/setup.cfg
+++ b/setup.cfg
@@ -96,8 +96,7 @@ tests =
     aiosmtpd
     async_generator
     bandit>=1.7.0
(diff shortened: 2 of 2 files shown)


.coveragerc
```diff
@@ -19,7 +19,8 @@ omit =
     cylc/flow/scripts/report_timings.py
 parallel = True
 source = ./cylc
-timid = False
```
MetRonnie, line 22: `timid` is false by default. And I think if set to true it seems to be the same as setting `core = pytrace`

https://coverage.readthedocs.io/en/7.11.3/config.html#run-timid

setup.cfg
```diff
@@ -96,8 +96,7 @@ tests =
     aiosmtpd
     async_generator
     bandit>=1.7.0
-    # https://github.com/coveragepy/coveragepy/issues/2082
-    coverage>=5.0.0,<7.11.1
+    coverage>=7.9.0
```
MetRonnie, line 99: New lower pin is because the `[run]core` config setting is new in this version

oliver-sanders (approved):
Assuming this does the job regards timings, plz merge.

PR #7088, opened 2025-11-17
Replace `select` with `DefaultSelector` in pipe poller

MetRonnie:
Fixes an error seen intermittently (more often than not on GH Actions lately) when running the integration tests.
> Error in plugin cylc.post_install.log_vc_info
> ValueError: filedescriptor out of range in select()

<details><summary>Full traceback</summary>

```
    async def test_install_scan_ping(
        src_run_dirs: Tuple[Path, Path],
        capsys: pytest.CaptureFixture,
        caplog: pytest.LogCaptureFixture,
        patch_graphql_query: Callable
    ) -> None:
        """At install, running intances should be reported.
    
        Ping = True case: but mock scan's scheduler query method.
        """
        opts = InstallOptions()
        opts.no_ping = False
    
>       await install_cli(opts, id_='w1')

tests/integration/test_install.py:139: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
cylc/flow/scripts/install.py:280: in install_cli
    wf_name, wf_id = await install(opts, id_)
                     ^^^^^^^^^^^^^^^^^^^^^^^^
cylc/flow/scripts/install.py:320: in install
    async for _entry_point, _plugin_result in run_plugins_async(
cylc/flow/plugins.py:85: in run_plugins_async
    _raise_plugin_exception(exc, plugin_namespace, entry_point)
cylc/flow/plugins.py:71: in run_plugins_async
    plugin_result = meth(*args, **kwargs)
                    ^^^^^^^^^^^^^^^^^^^^^
cylc/flow/install_plugins/log_vc_info.py:378: in main
    vc_info = get_vc_info(srcdir)
              ^^^^^^^^^^^^^^^^^^^
cylc/flow/install_plugins/log_vc_info.py:170: in get_vc_info
    out = _run_cmd(vcs, args, cwd=path)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cylc/flow/install_plugins/log_vc_info.py:258: in _run_cmd
    out, err = pipe_poller(proc, proc.stdout, proc.stderr)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cylc/flow/pipe_poller.py:68: in pipe_poller
    _read()
[...]

PR #7093, opened 2025-11-19
Replace `select`/`poll` with `DefaultSelector`

MetRonnie:
Follow-up to #7088 

This PR replaces the use of specific select/poll methods with [selectors.DefaultSelector()](https://docs.python.org/3/library/selectors.html#selectors.DefaultSelector)

As Oliver put it below:
 
> * `select.epoll` isn't portable.
> * `select.select` is portable, but has limitations.
> * This PR uses "the best" poller implementation available on any given OS.

**Check List**

- [x] No dependency changes 
- [x] Tests, changelog, docs not needed
- [x] Not a confirmed bug, so raising on master

### DIFF
--- a/cylc/flow/subprocpool.py
+++ b/cylc/flow/subprocpool.py
@@ -435,7 +435,9 @@ class SubProcPool:
         self.process()
         self.pipepoller.close()
 
-    def _poll_proc_pipes(self, proc, ctx):
+    def _poll_proc_pipes(
+        self, proc: 'Popen[bytes]', ctx: 'SubProcContext'
+    ) -> None:
         """Poll STDOUT/ERR of proc and read some data if possible.
 
         This helps to unblock the command by unblocking its pipes.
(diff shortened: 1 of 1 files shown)


cylc/flow/scheduler.py
```diff
@@ -1876,10 +1876,7 @@ async def _shutdown(self, reason: BaseException) -> None:
 
         if hasattr(self, 'proc_pool'):
             try:
-                self.proc_pool.close()
-                if self.proc_pool.is_not_done():
-                    self.proc_pool.terminate()
-                self.proc_pool.process()
```
MetRonnie, line 1882: These 4 lines did not do anything different to what `self.proc_pool.terminate()` already does!

oliver-sanders (approved):
LGTM, should manually test to double check the poller can unblock a process (if we don't already have an automated test).

oliver-sanders:
@wxtim, context:

* `select.epoll` isn't portable.
* `select.select` is portable, but has limitations.
* This PR uses "the best" poller implementation available on any given OS.

wxtim (commented):
The code makes sense - @MetRonnie do you have any thoughts on how I might go about checking this manually?

wxtim (approved):

PR #7100, opened 2025-11-25
set: clear task modifiers on manual output set

### LINKED ISSUE #7065: set: clear task modifiers on manual state change
If a task is manually set to a "final" status (ala `cylc set`), any task modifiers (held, retry, etc) are obsolete and should be cleared (both internally and in the data store) for clarity.

Leaving these modifiers in place is harmless, but confusing, and will make a mess of modifier filtering when that arrives.

oliver-sanders:
set: clear task modifiers when no longer relevant

* Closes #7065
* When a task is removed, e.g, via `cylc remove`, `cylc set` or natural task completion, the held state and any unsatisfied xtriggers are no longer relevant and should be cleared.
* Additionally, when a task naturally enters the preparing state, any unsatisfied xtriggers are no longer relevant and should also be cleared.
* This commit clears these states at the task-pool and data-store levels to make the task states clearer and more meaningful to users, especially in relation to task state filtering, which can get rather confusing otherwise.

**Check List**

### DIFF
--- a/changes.d/7100.fix.md
+++ b/changes.d/7100.fix.md
@@ -1,2 +1,2 @@
-Task held, retry, wallclock and xtriggered states are now cleared when they
+Task hold, retry, wallclock and xtriggers are now cleared when they
 are no longer applicable.


changes.d/7100.fix.md
```diff
@@ -0,0 +1,3 @@
+Task `is_held`, `is_retry`, `is_wallclock` and `is_xtriggered` states are now
```
hjoliver: I suppose it's obvious what that means, but they're not supposed to be user-facing terms. Maybe this?
```suggestion
The task hold, wallclock trigger, and xtrigger attributes are now
```

cylc/flow/task_pool.py
```diff
@@ -2258,6 +2273,14 @@ def _set_outputs_itask(
         if not itask.state(TASK_STATUS_WAITING):
             # Can't be runahead limited or queued.
             itask.state_reset(is_runahead=False, is_queued=False)
+            if status_geq(itask.state.status, TASK_STATUS_PREPARING):
+                # clear all xtriggers
+                self.xtrigger_mgr.force_satisfy_all(itask)
+
+                if is_manual:
+                    # clear the held state (if output is manually set)
+                    self.release_held_active_task(itask)
```
hjoliver: Wondering about future held tasks, which should also be released if we `cylc set` them to a later state ... 

... from testing, that works properly on this branch, because this code gets called even for non-active tasks (i.e., for a transient task proxy spawned just for setting outputs of the future task) ... which means `release_held_active_task` is confusingly named.  Not the fault of this PR but probably worth changing the method name to `release_held_task`?

hjoliver (approved):
A couple of minor comments, otherwise looks good and tests as working.

cylc/flow/task_pool.py
```diff
@@ -2258,6 +2273,14 @@ def _set_outputs_itask(
         if not itask.state(TASK_STATUS_WAITING):
             # Can't be runahead limited or queued.
             itask.state_reset(is_runahead=False, is_queued=False)
+            if status_geq(itask.state.status, TASK_STATUS_PREPARING):
+                # clear all xtriggers
+                self.xtrigger_mgr.force_satisfy_all(itask)
```
wxtim: I was trying to check that this would be safe if `@wc => foo & bar` - It's not, but I'm not sure that I understand the error fully, but I suspect that the other task requires the xtrigger to be loaded into the task pool.

```
[scheduler]
    allow implicit tasks = True
    cycle point format = %Y

[scheduling]
    initial cycle point = 2100

    [[xtriggers]]
        wc = wall_clock()
    [[graph]]
        R1 = """
            @wc => foo & bar
        """
```

If you `cylc set <workflow>//2100/foo # (or bar)` the workflow crashed thus:

```
CRITICAL - An uncaught error caused Cylc to shut down.
    If you think this was an issue in Cylc, please report the following traceback to the developers.
    https://github.com/cylc/cylc-flow/issues/new?assignees=&labels=bug&template=bug.md&title=;
ERROR - 'NoneType' object has no attribute 'is_xtrigger_sequential'
      File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/scheduler_cli.py", line 739, in cylc_play
        asyncio.get_running_loop()
        ~~~~~~~~~~~~~~~~~~~~~~~~^^
      File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/scheduler.py", line 711, in run_scheduler
        await self._main_loop()
      File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/scheduler.py", line 1654, in _main_loop
        self.pool.spawn_parentless_sequential_xtriggers()
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
      File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/task_pool.py", line 2421, in spawn_parentless_sequential_xtriggers
        self.check_spawn_psx_task(itask)
        ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
      File "/home/users/tim.pillinger/repos/cylc-flow/cylc/flow/task_pool.py", line 2428, in check_spawn_psx_task
        itask.is_xtrigger_sequential
    AttributeError: 'NoneType' object has no attribute 'is_xtrigger_sequential'
CRITICAL - Workflow shutting down - 'NoneType' object has no attribute 'is_xtrigger_sequential'

```

This does not, however, happen on 8.6.x, not even doing
```
> cylc set used.marble//2100/bar 
> cylc set used.marble//2100/bar --pre  xtrigger/wc
```

 😢

oliver-sanders: Hmmm, the fix for this issue is:

```diff
diff --git a/cylc/flow/task_pool.py b/cylc/flow/task_pool.py
index fbce6e200..bc0b3da22 100644
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -2417,8 +2417,8 @@ class TaskPool:
         """Spawn successor(s) of parentless wall clock satisfied tasks."""
         while self.xtrigger_mgr.sequential_spawn_next:
             taskid = self.xtrigger_mgr.sequential_spawn_next.pop()
-            itask = self._get_task_by_id(taskid)
-            self.check_spawn_psx_task(itask)
+            if itask := self._get_task_by_id(taskid):
+                self.check_spawn_psx_task(itask)
 
     def check_spawn_psx_task(self, itask: 'TaskProxy') -> None:
         """Check and spawn parentless sequential xtriggered task (psx)."""
```

However, this is concerning as the issue appears to be a pre-existing problem with sequential xtrigger spawning which raises questions over whether this spawning is guaranteed to happen. Have written this up in https://github.com/cylc/cylc-flow/issues/7120

oliver-sanders: Added the fix in https://github.com/cylc/cylc-flow/compare/5a3663d09188062daf6d6d37a562f7ac836aa444..10231df249ba1d1994bf158232bd06f0dbcdfb44

cylc/flow/task_pool.py
oliver-sanders: This method is only ever provided with `TaskProxy` objects from the pool, so they are always n=0 by definition, even if they have just been spawned and/or are in the process of being completed as the result of the set.

changes.d/7100.fix.md
oliver-sanders: Agreed & simplified further.

tests/flakyfunctional/hold-release/14-hold-kill.t
oliver-sanders, line 1: Have replaced this icky functional test with an integration test which also covers the state changes and interaction with `cylc set` and `cylc remove`.

wxtim (approved):
- [x] Read code
- [x] Checked new test against old functional test
- [x] Thrown some IRL stuff against PR
- [x] Checked that I cannot repeat earlier bug

changes.d/7100.fix.md
```diff
@@ -0,0 +1,2 @@
+Task held, retry, wallclock and xtriggered states are now cleared when they
```
hjoliver: Not a biggie to use "states" in the colloquial sense in this context (change log), but probably still better to avoid muddying our definition of "task state" in any way.

```suggestion
Task held, retry, wallclock and xtriggered attributes are now cleared when they
```
Or "modifiers" or avoid the problem:
```suggestion
Task hold, retry, wallclock and xtriggers are now cleared when they
```

cylc/flow/task_pool.py
hjoliver: Yes it's only ever provided with `TaskProxy` objects, but no they're not always from the pool.

 They can be transient task proxies instantiated to use for spawning the outputs of `n>0` tasks (after a `cylc set` operation) and then discarded without adding to the task pool.  

```
a => b => c => d
```
with:
- `a` fails to stall the workflow
- then `cylc hold //1/c` (held future task)
- then `cylc set //1/c` (set the outputs the held future task)

Result:  `release_held_active_tasks(...)` gets called with non-active task proxy `1/c`, and this is actually what releases the hold on the future task too - it gets removed from the `tasks_to_hold` list inside that method.

changes.d/7100.fix.md
oliver-sanders: Not particularly fussed, but I don't think either "attributes" or "modifers" are better.

"Attributes" is not a user facing term (it derives from Python attributes).

"Modifiers" is not a user facing term (just something we use as developers).

"State" is an accurate use of language that users will understand it is also historically just (and, although irrelevant, it's also technically correct as modifiers are a part of the `TaskState` object).

cylc/flow/task_pool.py
oliver-sanders: Ach, `TaskProxy` objects created in/by the pool, versus `TaskProxy` objects stored in a particular field in the pool.

Arguably, `1/c` becomes active (and it literally IS active) in order for the operation to be performed onto it.

Completely arbitrary IMO, not interested in arguing.

hjoliver: Fair enough, I guess, but going by the definitions we've always used to date, future tasks do not become active if their outputs are set ... which does make the name of that function misleading. (We've always said active tasks form "the task pool" and the TaskPool object manages other associated stuff in addition to the active tasks). 

I do agree that the distinction is somewhat arbitrary - we could decide to say such tasks are briefly active then disappear immediately - but we have never done that.

(The use of "transient" task proxies is kind of a bodge; ideally we might (e.g.) spawn on the outputs of future tasks  without instantiating temporary task proxies, which would be less confusing, but a little more painful to code).

I'll pull up a small tidy PR since you can't be bothered with it (which is fair enough).

hjoliver: See #7142

PR #7101, opened 2025-11-26
Fix spawn order when group triggering tasks before the start cycle point

MetRonnie:
Following a warm start, group triggering tasks that exist in the initial cycle point only, such as so-called "install cold" tasks, has a bug where the prerequisites within the group are not obeyed - they end up force-satisfied and the whole group submits at the same time.

#### Repro

```cylc
[scheduler]
    allow implicit tasks = True

[scheduling]
    cycling mode = integer
    initial cycle point = 1
    runahead limit = P2
    [[graph]]
        R1 = herring => cold1 => cold2 => foo
        P1 = foo[-P1] => foo

[runtime]
    [[COLD]]
    [[cold1, cold2]]
        inherit = COLD
```

```
$ cylc play wflow --startcp 5
$ cylc trigger wflow//^/COLD
```

#### Check List

- [x] No dependency changes 
- [x] Tests are included 
- [x] No docs needed

### DIFF
--- a/tests/unit/parsec/__init__.py
+++ b/tests/unit/parsec/__init__.py
@@ -1,0 +1,15 @@
+# THIS FILE IS PART OF THE CYLC WORKFLOW ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.


cylc/flow/commands.py
```diff
@@ -267,15 +268,14 @@ def _remove_matched_tasks(
...
+    if warn_unremovable and not_removed:
         fnums_str = (
             repr_flow_nums(flow_nums, full=True) if flow_nums else ''
         )
         tasks_str = ', '.join(
             sorted(tokens.relative_id for tokens in not_removed)
         )
-        # This often does not indicate an error - e.g. for group trigger.
-        LOG.debug(f"Task(s) not removable: {tasks_str} {fnums_str}")
+        LOG.warning(f"Task(s) not removable: {tasks_str} {fnums_str}")
```
oliver-sanders, line 278: The comment suggests it's not just group trigger where this warning can be erroneous.

I found one example:

```
cylc trigger task --flow=2
cylc remove FAMILY --flow=2
```

(where `task` is a member of `FAMILY`)

oliver-sanders:
(coverage failing on uncovered __repr__ methods)

tests/integration/test_force_trigger.py
```diff
@@ -742,3 +742,38 @@ def _force_trigger_tasks(_schd, ids, *_, **__):
         await run_cmd(force_trigger_tasks(schd, ['*:failed'], []))
         assert trigger_calls == [{'1/d'}]
         trigger_calls.clear()
+
+
+async def test_pre_warm_start_group_trigger(flow, scheduler, run, complete):
+    """Group-triggered tasks that are before the start point (in a warm-started
+    workflow) should run in order."""
```
wxtim:
```suggestion
    workflow) should run in order.
    
    https://github.com/cylc/cylc-flow/pull/7101
    """
```

tests/integration/test_force_trigger.py
```diff
@@ -742,3 +742,38 @@ def _force_trigger_tasks(_schd, ids, *_, **__):
...
+    schd: Scheduler = scheduler(
+        flow({
+            'scheduling': {
+                'initial cycle point': '1',
```
wxtim: Not required?

cylc/flow/task_trigger.py
```diff
@@ -247,30 +252,25 @@ def get_prerequisite(
...
-                cpre[(
-                    task_trigger.get_point(point),
-                    task_trigger.task_name,
-                    task_trigger.output
-                )] = (
-                    (prereq_offset_point < tdef.start_point) &
-                    (point >= tdef.start_point)
+                    if tdef.max_future_prereq_offset is None or (
+                        prereq_offset > tdef.max_future_prereq_offset
+                    ):
```
wxtim: This feels more readable to me.
```suggestion
                    if (
                        tdef.max_future_prereq_offset is None
                        or prereq_offset > tdef.max_future_prereq_offset
                    ):
```

cylc/flow/commands.py
MetRonnie: That seems more debatable to me than obviously erroneous. Not sure how we can address this easily either. One for a new issue?

oliver-sanders: Bump to another PR? Or ask Hillary about it.

MetRonnie:
The workflow flowing on from the triggered tasks is expected (but problematic) at the moment; I have not found a fix for that yet. However the runahead stall in your example is probably preferential to it not stalling, as it gives a chance to remove the unwanted tasks!

oliver-sanders:
Understood, but unfortunately, I think that we need to come up with a solution to the flow-on problem before the intervention for this use case can be used in anger as it's too caveat-prone without this.

No idea what that solution would be however!

cylc/flow/commands.py
MetRonnie: https://github.com/cylc/cylc-flow/issues/7124

hjoliver: See my comments on #7124 - it seems to me we don't need to warn here under any circumstances, and therefore don't need the `warn_unremovable` flag either.

cylc/flow/task_pool.py
```diff
@@ -836,9 +836,11 @@ def spawn_to_rh_limit(
         longer parentless, and/or hit the runahead limit.
 
         """
-        if not flow_nums or point is None:
-            # Force-triggered no-flow task.
-            # Or called with an invalid next_point.
+        if (
+            not flow_nums  # Force-triggered no-flow task
+            or point is None  # Reached end of sequence?
+            or point < self.config.start_point  # Warm start
```
wxtim, line 842: [not a blocker]
Do we want to continue to propogate the warm start terminology?

wxtim (approved):

cylc/flow/task_pool.py
oliver-sanders: This isn't user facing.

Issue #7104, opened 2025-11-27
cylc set: allow selecting tasks by state


oliver-sanders:
Cylc commands can select tasks by state, e.g, `cylc trigger '//*:failed`.

However, we blocked this for `cylc set` to avoid a likely misunderstanding.

I recently encountered a use case where this would be useful:

```
cylc set '//1/*:failed' --out=expired
InputError: Use "--output=failed" to specify outputs, not "*:failed"
```

PR #7108, opened 2025-11-28
set: remove state selection limitation

oliver-sanders:
Closes https://github.com/cylc/cylc-flow/issues/7104

@hjoliver, it would be useful to be able to select tasks by state when using the `cylc set` command.

However, this was purposefully disabled when the `cylc set` command was added. I think this was to prevent users accidentally running `cylc set <task>:out` when they meant `cylc set <task> --out=<out>`? However, this check was only ever added for the CLI, not the GUI.

Are you ok with removing this check?

---

This has randomly come up twice in the last two days.

One use case is mopping up failed/submit-failed tasks from otherwise expired cycles:

```
$ cylc set '//1/*:failed' --out=expired
```

The other, from opps, was setting failed tasks to succeeded where manually clicking them one-by-one was tedious.

**Check List**

- [ ] Tests are included (or explain why tests are not needed). - don't need a test to ensure we don't do something unusual

### DIFF
--- a/changes.d/7108.fix.md
+++ b/changes.d/7108.fix.md
@@ -1,0 +1,1 @@
+Allow use of task state selectors with the `cylc set` command.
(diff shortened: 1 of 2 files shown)


hjoliver:
Actually, pre merge I'll just to a quick side-PR to document that point of possible confusion that we were presumably trying to avoid...

PR #7116, opened 2025-12-03
Move paused start logic to before initiating data store and task pool

MetRonnie:
This may be needed to avoid sending a delta with "running" status before "paused" status is applied.

I have seen on several occasions a workflow started with `--pause` appear as running for several seconds in the GUI before becoming paused, which is slightly alarming as it makes you think you might have forgotten the `--pause` option.

**Check List**

- [x] No dependency changes 
- [x] Changelog entry not needed as minor
- [x] Docs not needed

### DIFF
--- a/tests/integration/network/test_graphql.py
+++ b/tests/integration/network/test_graphql.py
@@ -463,9 +463,9 @@ async def test_subscription_deltas(one, start):
         assert aitem.data['deltas']['added']['workflow'] == {
             'id': one.id,
             'host': one.host,
-            'status': 'running',
+            'status': 'paused',
         }
-        # Workflow one is paused on start, but this hasn't been processed yet.
+        # Workflow one is paused on start
         await one.update_data_structure()
         assert (
             one.data_store_mgr.data[one.id]['workflow'].status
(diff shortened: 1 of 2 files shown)


tests/integration/network/test_graphql.py
```diff
@@ -486,7 +486,6 @@ async def test_subscription_deltas(one, start):
         aitem = await subscription.__anext__()
         assert aitem.data['deltas']['updated']['workflow'] == {
             'id': one.id,
-            'status': get_workflow_status(one).value,
         }
```
MetRonnie, line 489: This test could probably do with improvement now that I've removed the one change it was testing in a delta?

dwsutherland:
> > Partially addresses [cylc/cylc-uiserver#485](https://github.com/cylc/cylc-uiserver/issues/485)
> 
> The change in this PR seems sensible, but that Issue doesn't mention the paused state so far as I can see?

Yes, seems it would turn it into a paused and empty delta problem (instead of running and empty) for this particular case. (?)

However, it does seem sensible..

dwsutherland (approved):

MetRonnie:
> > Partially addresses [cylc/cylc-uiserver#485](https://github.com/cylc/cylc-uiserver/issues/485)
> 
> The change in this PR seems sensible, but that Issue doesn't mention the paused state so far as I can see?

You're right; think I got my wires crossed looking into several related issues here. I've removed that line from the OP

PR #7121, opened 2025-12-04
Fix bug when broadcasting deprecated config

MetRonnie:
#### Repro

Any simple workflow

```
$ cylc broadcast wflow -p 1 -n foo -s '[job]execution time limit'

  File ".../bin/cylc", line 9, in <module>
    sys.exit(main())
             ^^^^^^
  File ".../lib/python3.12/site-packages/cylc/flow/scripts/cylc.py", line 646, in main
    ret = _main(opts, cmd_args)
          ^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.12/site-packages/cylc/flow/scripts/cylc.py", line 706, in _main
    return execute_cmd(command, *cmd_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.12/site-packages/cylc/flow/scripts/cylc.py", line 329, in execute_cmd
    entry_point.load()(*args)
  File ".../lib/python3.12/site-packages/cylc/flow/terminal.py", line 298, in wrapper
  File ".../lib/python3.12/site-packages/cylc/flow/scripts/broadcast.py", line 506, in main
    rets = asyncio.run(_main(options, *ids))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.12/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File ".../lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.12/asyncio/base_events.py", line 691, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File ".../lib/python3.12/site-packages/cylc/flow/scripts/broadcast.py", line 511, in _main
    return await call_multi_async(
           ^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.12/site-packages/cylc/flow/network/multi.py", line 139, in call_multi_async
    out, err, outcome = _process_response(
                        ^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.12/site-packages/cylc/flow/network/multi.py", line 227, in _process_response
    raise response
  File ".../lib/python3.12/site-packages/cylc/flow/async_util.py", line 436, in _inner
    return await coroutine(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.12/site-packages/cylc/flow/scripts/broadcast.py", line 476, in run
    modified_settings = result['response'][0]
                        ~~~~~~~~~~~~~~~~~~^^^
KeyError: 0
```

#### Check List

- [x] No dependency changes 
- [x] Tests are included 
- [x] Docs not needed

### DIFF
--- a/changes.d/7121.fix.md
+++ b/changes.d/7121.fix.md
@@ -1,0 +1,1 @@
+Fixed `cylc broadcast` traceback when broadcasting an auto-upgraded deprecated setting.


cylc/flow/scripts/broadcast.py
```diff
@@ -196,13 +196,36 @@ def get_rdict(left, right=None, for_cancel_broadcast=False):
...
+    rdict = strip_empty_sections(rdict)
     # Perform validation, but don't coerce the original (deepcopy).
     cylc_config_validate(deepcopy(rdict), SPEC['runtime']['__MANY__'])
     return rdict
 
 
+def strip_empty_sections(data):
+    """Recursively strip data structure of empty dicts.
+
+    This is needed post-upgrade as the upgrader can leave empty sections
```
wxtim, line 210: Can we not make the upgrade not do that? Or is it a feature that the upgrade actually uses?

MetRonnie: Not sure. Seems safer not to change that

cylc/flow/parsec/upgrade.py
```diff
@@ -43,17 +43,28 @@ class upgrader:
...
         self.cfg = cfg
         self.descr = descr
         # upgrades must be ordered in case several act on the same item
-        self.upgrades = OrderedDict()
+        self.upgrades: OrderedDict[str, list[dict]] = OrderedDict()
+        if broadcast:
+            self.DEPR_MSG = self.DEPR_TEMPLATE.format('broadcast')
+            self.OBSLT_MSG = (
+                "Obsolete config items were rejected by the broadcast."
+            )
```
oliver-sanders: Overloading class constants isn't the nicest pattern (actually gets flagged by flake8 C002 as redefinition of constant).

Both only get used once, so might be more appropriate to do the formatting there than storing the modified template on the instance?

cylc/flow/scripts/broadcast.py
```diff
@@ -196,13 +196,36 @@ def get_rdict(left, right=None, for_cancel_broadcast=False):
...
+    >>> strip_empty_sections({'section': {'subsection': {}}})
+    {}
+    """
+    if not isinstance(data, dict):
+        return data
+    return {
+        key: stripped
+        for key, val in data.items()
+        if (stripped := strip_empty_sections(val)) != {}
+    }
```
oliver-sanders, line 226: Wow a recursive comprehension which implicitly modifies itself via the walrus operator!

It's cool this works, though I'm kinda surprised flak8 doesn't complain.

For context, it's often considered bad practice for a comprehension to have side effects ([SO](https://stackoverflow.com/questions/5753597/is-it-pythonic-to-use-list-comprehensions-for-just-side-effects)).

oliver-sanders (commented):
LGTM, tested as working.

Gotta couple of QA things which aren't covered by the linter, open to debate.

cylc/flow/parsec/upgrade.py
MetRonnie: Is that just because the variables have uppercase names? Otherwise there is nothing constant about class attributes and there can be no harm in reassigning a instance attribute that shares the same name as a class attribute

cylc/flow/scripts/broadcast.py
MetRonnie: Unless I'm missing something, this comprehension does not have any side effects? The function does not modify anything, it just returns a subset of the given dictionary

wxtim: A quick check suggests that @MetRonnie  is right?

```python
data = {'foo': 1, 'bar': 2, 'baz': 3}
```
```python
{
    key: mod_val
    for key, val in data.items()
    if (mod_val := val % 2) == 1
}
```
```
    {'foo': 1, 'baz': 1}
```
```python
data
```
```
    {'foo': 1, 'bar': 2, 'baz': 3}
```

cylc/flow/parsec/upgrade.py
oliver-sanders: Yep. Python has no top-level concept of a constant, so there is a convention instead, we use upper case for constants and lower case for variables. Once defined a constant should not be redefined (C002).

This pattern is also using class/instance variable shadowing because we have two variables with the same name, `cls.DEPR_MSG` and `self.DEPR_MSG` which can be confusing as `self.DEPR_MSG` will return `cls.DEPR_MSG` if the instance variable is unset.

cylc/flow/scripts/broadcast.py
oliver-sanders: Fair, if it's not modifying.

PR #7132, opened 2025-12-10
event driven seq xtrig task spawn

### LINKED ISSUE #7103: clock-expire: premature shutdown due to sequential xtriggers triggers
There is a bad interaction between sequential xtriggers and clock-expire which can result in premature shutdown.

#### Reproducible Example

```cylc
[scheduling]
    initial cycle point = 2000
    [[special tasks]]
        clock-expire = a
    [[graph]]
        P1D = """
            @wall_clock => a => b => c
        """

[runtime]
    [[a]]
        completion = succeeded or expired
    [[b,c]]
```

```console
$ cylc vip -N .
$ cylc validate /var/tmp/tmp.e74BWQakBa
Valid for cylc-8.6.1.dev
$ cylc install /var/tmp/tmp.e74BWQakBa
INSTALLED tmp.e74BWQakBa/run6 from /var/tmp/tmp.e74BWQakBa
$ cylc play -N tmp.e74BWQakBa/run6

 ▪ ■  Cylc Workflow Engine 8.6.1.dev
 ██   Copyright (C) 2008-2025 NIWA

INFO - Extracting job.sh to ~/cylc-run/tmp.e74BWQakBa/run6/.service/etc/job.sh
INFO - Workflow: tmp.e74BWQakBa/run6
INFO - Scheduler: url=tcp://... pid=...
INFO - Workflow publisher: url=tcp://...
INFO - Run: (re)start number=1, log rollover=1
INFO - Cylc version: 8.6.1.dev
INFO - Run mode: live
INFO - Initial point: 20000101T0000Z
INFO - Final point: None
INFO - Cold start from 20000101T0000Z
INFO - New flow: 1 (original flow from 20000101T0000Z) 2025-11-27T14:18:58
INFO - [20000101T0000Z/a:waiting(runahead)] => waiting
INFO - xtrigger succeeded: wall_clock = wall_clock(trigger_time=946684800)
INFO - [20000101T0000Z/a:waiting] => waiting(queued)
INFO - [20000102T0000Z/a:waiting(runahead)] => waiting
WARNING - [20000101T0000Z/a:waiting(queued)] (internal)expired
INFO - [20000101T0000Z/a:waiting(queued)] => expired
WARNING - [20000102T0000Z/a:waiting] (internal)expired
INFO - [20000102T0000Z/a:waiting] => expired
INFO - Workflow shutting down - AUTOMATIC
INFO - DONE
```

### Workaround

Define a wall_clock xtrigger and set `sequential=False`.

### LINKED ISSUE #7120: sequential xtriggers: edge cases
I'm not quite sure the sequential xtrigger mechanism is watertight.

A [recent bug](https://github.com/cylc/cylc-flow/pull/7100/files#r2588262122) raised whilst working on https://github.com/cylc/cylc-flow/pull/7100 revealed a flaw where sequential spawning fails if the task has been removed from the pool.

The spawning happens in the `Scheduler`'s event loop:

https://github.com/cylc/cylc-flow/blob/e757f44cf5ad5bc0c8e4360c0dc397dfbf6d272a/cylc/flow/scheduler.py#L1653-L1654

However, the housekeeping of the list of triggers to spawn is tied to the `TaskProxy` lifecycle (in the remove method):

https://github.com/cylc/cylc-flow/blob/e757f44cf5ad5bc0c8e4360c0dc397dfbf6d272a/cylc/flow/task_pool.py#L892-L896

So if the `TaskProxy` lifecycle steps out of sync with the `Scheduler` event loop, bugs can occur.

In [this case](https://github.com/cylc/cylc-flow/pull/7100/files#r2588262122) I had made a call to `XtriggerMgr.force_satisfy` as part of task completion which caused the task to be removed too soon, breaking the logic. The fix was to suppress spawning if the task is no longer in the pool, relying on the pre-existing assumption that if a task is no longer in the pool then the sequential spawning must already have happened. But is this guaranteed?

Consider also shutdown/restart cases. We don't back this list up in the DB and we don't perform sequential spawning as part of shutdown, and can't in the event of a crash. See also https://github.com/cylc/cylc-flow/issues/7103

I suspect that we need to shift to performing the sequential spawning at event time, rather than after the fact.

dwsutherland:
closes #7120
closes #7103 

**Check List**

- [x] Changelog not required

### DIFF
--- a/cylc/flow/scheduler.py
+++ b/cylc/flow/scheduler.py
@@ -404,12 +404,7 @@ class Scheduler:
         self.workflow_event_handler = WorkflowEventHandler(self.proc_pool)
 
         self.xtrigger_mgr = XtriggerManager(
-            self.workflow,
-            user=self.owner,
-            broadcast_mgr=self.broadcast_mgr,
-            workflow_db_mgr=self.workflow_db_mgr,
-            data_store_mgr=self.data_store_mgr,
-            proc_pool=self.proc_pool,
+            self,
             workflow_run_dir=self.workflow_run_dir,
             workflow_share_dir=self.workflow_share_dir,
         )
@@ -1649,9 +1644,6 @@ class Scheduler:
 
             if itask.is_ready_to_run() and not itask.is_manual_submit:
                 self.pool.queue_task(itask)
-
-        if self.xtrigger_mgr.sequential_spawn_next:
-            self.pool.spawn_parentless_sequential_xtriggers()
 
         if self.xtrigger_mgr.do_housekeeping:
             self.xtrigger_mgr.housekeep(self.pool.get_tasks())
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -896,7 +896,6 @@ class TaskPool:
         itask.transient = True
 
         if itask.is_xtrigger_sequential:
-            self.xtrigger_mgr.sequential_spawn_next.discard(itask.identity)
             self.xtrigger_mgr.sequential_has_spawned_next.discard(
                 itask.identity
             )
@@ -2400,13 +2399,6 @@ class TaskPool:
         # Task may be set running before xtrigger is satisfied,
         # if so check/spawn if xtrigger sequential.
         self.check_spawn_psx_task(itask)
-
-    def spawn_parentless_sequential_xtriggers(self):
-        """Spawn successor(s) of parentless wall clock satisfied tasks."""
-        while self.xtrigger_mgr.sequential_spawn_next:
-            taskid = self.xtrigger_mgr.sequential_spawn_next.pop()
-            if itask := self._get_task_by_id(taskid):
-                self.check_spawn_psx_task(itask)
 
     def check_spawn_psx_task(self, itask: 'TaskProxy') -> None:
         """Check and spawn parentless sequential xtriggered task (psx)."""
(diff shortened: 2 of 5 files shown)


cylc/flow/scheduler.py
```diff
@@ -404,12 +404,7 @@ async def initialise(self):
         self.workflow_event_handler = WorkflowEventHandler(self.proc_pool)
 
         self.xtrigger_mgr = XtriggerManager(
-            self.workflow,
-            user=self.owner,
-            broadcast_mgr=self.broadcast_mgr,
-            workflow_db_mgr=self.workflow_db_mgr,
-            data_store_mgr=self.data_store_mgr,
-            proc_pool=self.proc_pool,
+            self,
```
dwsutherland, line 407: Task pool isn't instantiated yet, but will be when `xtrigger_mgr` calls it (via `self.schd.pool...`).. (also `xtrigger_mgr` is passed into pool already)

oliver-sanders:
> Covered by existing tests.

We don't have a test covering the interaction of these two features (and expiry testing is a bit light in general), so worth sticking a simple integration test in, raised:

https://github.com/dwsutherland/cylc-flow/pull/26

oliver-sanders (approved):
This makes sense to me, fixes the particular case in https://github.com/cylc/cylc-flow/issues/7103. This approach should be shudown/restart safe, so will avoid any possible issues there (https://github.com/cylc/cylc-flow/issues/7120).

(added the example from #7103 as an integration test on a side pr)

wxtim (approved):

Issue #7020, opened 2025-10-08
Multiple isolated graphs.


hjoliver:
A clean issue for a problem with a long and convoluted history (e.g. see #4903, #4912, #5036, #5090).

# the problem

Cylc famously has no barrier between cycles. This is usually a very good thing, but there are certain situations where a hard barrier would be convenient because it stands in for ALL dependence on tasks prior to the barrier.

The following diagrams illustrate for a simple workflow with tasks in the first cycle that *everything else* depends on (e.g. to deploy code and data); and tasks in the final cycle that should not run until *everything else* has finished. 

## what we want

<img width="1592" height="857" alt="Image" src="https://github.com/user-attachments/assets/aab3c7bc-417a-43de-a0f2-e767219b3eac" />

## how we currently achieve it (1/2)

The blue bits represent perpetual dependence on the initial tasks, to ensure that nothing runs before they finish. This is reasonably intuitive and reflects real dependence - not a workaround - but it causes problems:
- it makes an unbelievable mess of graph visualizations for real workflows
- it makes retriggering the startup graph difficult or confusing (will it "flow on" again?)
- it can cause performance issues in the cycling computations, far from the ICP (is this still true?)

The red bits represent dependence on special dummy tasks inserted purely to ensure that final-cycle tasks wait on everything else - i.e. it's a workaround.  It is not possible to do blue thing for the final cycle - that would require `foo => bar[$]` 
 - pragmatically the workaround (red) is probably better than the proper solution (blue)
 - but it can get nasty for some workflows (with lots of parentless tasks at the top of a cycle)
 - and evidence at ESNZ shows users don't naturally think of it (all of our workflows have the blue bits)

<img width="1248" height="1434" alt="Image" src="https://github.com/user-attachments/assets/f76fa7e8-6957-45ca-a8d6-08150a5dab0d" />

## how we currently achieve it (2/2)

Probably the cleanest solution at the moment is to use the dummy task workaround for both startup and shutdown tasks - but see comments above on the downsides of the workaround.

<img width="1248" height="1553" alt="Image" src="https://github.com/user-attachments/assets/3846838e-df15-467e-adb2-91ec33e33f9a" />

oliver-sanders:
> @oliver-sanders, @hjoliver - Thoughts? If we can agree on this basic implementation, then let's get the ball rolling

This is going to require more thought before work can begin.

The bad news is that we're really blocked out at our end and will be until ~ the end of January. Sorry about that.

[...]

PR #7134, opened 2025-12-10
reinstate no-spawn option for remove

dwsutherland:
At present there's no way to remove parentless tasks from the scheduler and not have them spawn their next instance.

This is not ideal, and at ESNZ I've run into this problem many times while adding new tasks to a workflow and them being spawned in at the distant past (i.e. the ICP), and then the only way to fix the workflow is to spawn them forward one cycle at a time to the desired cycle point.

Also, in general, perhaps you wish not to run an isolate/branch for a gap then introduce again at some future cycle point.

This PR fixes the problem by reinstating `--no-spawn` option to `cylc remove`..

Example:
```
[scheduling]
    initial cycle point = 20260101T0000Z
    [[graph]]
        P1D = """
            @wall_clock => a => b => c
            d
        """

[runtime]
    [[root]]
        script = sleep 5
    [[a,b,c,d]]
```
running shows:
image
Removing `d` normally (`$ cylc remove couple/run1//20260106T0000Z/d`) spawns it into the next cycle:
image
With `--no-spawn`:
image

And of course removing the sequentially spawned `a` (`cylc remove --no-spawn couple/run1//20260101T0000Z/a`), with this option, gives:
image
(as expected)

Code changes were made in such a way that they should be conflict-free with #7132 

Seems our forms are smart enough at the UI end:
image

**Check List**

- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX. (not needed I think, [docs](https://cylc.github.io/cylc-doc/stable/html/user-guide/interventions/index.html#remove-tasks) aren't that detailed about removal and options are explained in UI/CLI info)

### DIFF
--- a/changes.d/7134.feat.md
+++ b/changes.d/7134.feat.md
@@ -1,0 +1,1 @@
+Remove without spawning option, needed for parentless tasks, has been reinstated.
--- a/cylc/flow/commands.py
+++ b/cylc/flow/commands.py
@@ -162,6 +162,7 @@ def _remove_matched_tasks(
     ids: Set[TaskTokens],
     flow_nums: 'FlowNums',
     warn_unremovable: bool = True,
+    no_spawn: bool = False,
 ):
     """Remove matched tasks."""
     # Mapping of *relative* task IDs to removed flow numbers:
@@ -183,6 +184,7 @@ def _remove_matched_tasks(
                 # Need to remove the task from the pool.
                 # Spawn next occurrence of xtrigger sequential task (otherwise
                 # this would not happen after removing this occurrence):
+                itask.no_spawn = no_spawn
                 schd.pool.check_spawn_psx_task(itask)
                 schd.pool.remove(itask, 'request')
                 to_kill.append(itask)
@@ -493,13 +495,17 @@ async def set_verbosity(schd: 'Scheduler', level: 'Enum'):
 
 @_command('remove_tasks')
 async def remove_tasks(
-    schd: 'Scheduler', tasks: Iterable[str], flow: List[str]
+    schd: 'Scheduler',
+    tasks: Iterable[str],
+    flow: List[str],
+    no_spawn: bool = False
 ):
     """Match and remove tasks (`cylc remove` command).
 
     Args:
         tasks: Relative IDs or globs to match.
         flow: flows to remove the tasks from.
+        no_spawn: Do not spawn successors before removal.
     """
     flow = back_compat_flow_all(flow)  # BACK COMPAT (see func def)
     ids = validate.is_tasks(tasks)
@@ -512,7 +518,8 @@ async def remove_tasks(
         _remove_matched_tasks(
             schd,
             matched,
-            schd.pool.flow_mgr.cli_to_flow_nums(flow)
+            schd.pool.flow_mgr.cli_to_flow_nums(flow),
+            no_spawn=no_spawn,
         )
 
 
--- a/cylc/flow/network/schema.py
+++ b/cylc/flow/network/schema.py
@@ -2185,6 +2185,21 @@ class Remove(Mutation, TaskMutation):
                 By default, tasks will be removed from all flows.
             ''')
         )
+        no_spawn = Boolean(
+            default_value=False,
+            description=sstrip('''
+                "Do not spawn successors before removal."
+
+                This usually only applies to parentless tasks whose next
+                instance(s) are either spawned automatically to the runahead
+                limit or sequentially on xtrigger satisfaction.
+
+                If `false` the scheduler will spawn the next task instance
+                as normal (default).
+
+                If `true` the scheduler will not spawn the next task instance.
+            ''')
+        )
 
 
 class SetPrereqsAndOutputs(Mutation, TaskMutation):
--- a/cylc/flow/scripts/remove.py
+++ b/cylc/flow/scripts/remove.py
@@ -71,11 +71,13 @@ mutation (
   $wFlows: [WorkflowID]!,
   $tasks: [NamespaceIDGlob]!,
   $flow: [Flow!],
+  $noSpawn: Boolean,
 ) {
   remove (
     workflows: $wFlows,
     tasks: $tasks,
-    flow: $flow
+    flow: $flow,
+    noSpawn: $noSpawn
   ) {
     result
   }
@@ -91,7 +93,12 @@ def get_option_parser() -> COP:
         multiworkflow=True,
         argdoc=[FULL_ID_MULTI_ARG_DOC],
     )
+
     add_flow_opts_for_remove(parser)
+    parser.add_option(
+        "--no-spawn",
+        help="Do not spawn successors before removal.",
+        action="store_true", default=False, dest="no_spawn")
     return parser
 
 
@@ -107,6 +114,7 @@ async def run(options: 'Values', workflow_id: str, *tokens_list):
                 for tokens in tokens_list
             ],
             'flow': options.flow,
+            'noSpawn': options.no_spawn,
         }
     }
 
--- a/cylc/flow/task_pool.py
+++ b/cylc/flow/task_pool.py
@@ -888,7 +888,7 @@ class TaskPool:
         # xtriggers are no longer relevant -> remove them
         self.xtrigger_mgr.force_satisfy_all(itask, log=False)
 
-        if itask.state.is_runahead and itask.flow_nums:
+        if itask.state.is_runahead and itask.flow_nums and not itask.no_spawn:
             # If removing a parentless runahead-limited task
             # auto-spawn its next instance first.
             self.spawn_if_parentless(
@@ -2436,6 +2436,7 @@ class TaskPool:
                 itask.identity not in
                 self.xtrigger_mgr.sequential_has_spawned_next
             )
+            and not itask.no_spawn
         ):
             self.xtrigger_mgr.sequential_has_spawned_next.add(
                 itask.identity
--- a/cylc/flow/task_proxy.py
+++ b/cylc/flow/task_proxy.py
@@ -166,6 +166,9 @@ class TaskProxy:
         .removed:
             A flag to indicate this task has been removed by command (used
             e.g. to disable failed/submit-failed event handlers).
+        .no_spawn:
+            A flag to indicate whether this task should spawn a successor
+            (usually only applies to parentless tasks).
 
     Args:
         tdef: The definition object of this task.
@@ -212,6 +215,7 @@ class TaskProxy:
         'transient',
         'is_xtrigger_sequential',
         'removed',
+        'no_spawn',
     )
 
     def __init__(
@@ -288,6 +292,7 @@ class TaskProxy:
         self.is_late = is_late
         self.waiting_on_job_prep = False
         self.removed: bool = False
+        self.no_spawn: bool = False
 
         self.state = TaskState(tdef, self.point, status, is_held)
 
--- a/tests/integration/test_remove.py
+++ b/tests/integration/test_remove.py
@@ -545,3 +545,47 @@ async def test_remove_triggered(flow, scheduler, start):
         )
         assert not schd.pool.get_tasks()
         assert not schd.pool.tasks_to_trigger_now
+
+
+async def test_remove_no_spawn(flow, scheduler, start):
+    """Test the no-spawn removal of parentless tasks."""
+    schd: Scheduler = scheduler(
+        flow({
+            'scheduler': {
+                'cycle point format': 'CCYY',
+            },
+            'scheduling': {
+                'initial cycle point': '2000',
+                'runahead limit': 'P0',
+                'graph': {
+                    'R3//P1Y': '''
+                        @wall_clock => a
+                        b
+                        c
+                    ''',
+                },
+            },
+        })
+    )
+    async with start(schd):
+        assert schd.pool.get_task_ids() == {
+            '2000/a',
+            '2000/b',
+            '2000/c',
+            '2001/b',
+            '2001/c',
+        }
+
+        # normal removal of parentless runahead and sequential xtrigger
+        # spawned tasks
+        await run_cmd(remove_tasks(schd, ['2000/a', '2000/b', '2001/b'], []))
+        assert schd.pool.get_task_ids() == {
+            '2000/c',
+            '2001/a',
+            '2001/c',
+            '2002/b',
+        }
+
+        # no spawn removal
+        await run_cmd(remove_tasks(schd, ['2001/a', '2001/c'], [], True))
+        assert schd.pool.get_task_ids() == {'2000/c', '2002/b'}


hjoliver:
>  I've run into this problem many times while adding new tasks to a workflow and them being spawned in at the distant past (i.e. the ICP),

How did they get spawned back at the ICP - hopefully that does not happen automatically?

(Even if that was human error, it's a still a problem that we need to be able to fix easily, but just wondering...)

oliver-sanders:
As it stands, the `cylc remove` command is a bit odd, it can do one of three things:

1. Remove a single instance of a task.
2. Temporally remove a single instance of a task (if it gets re-spawned as the result of another output).
3. Remove all instances of all tasks (i.e, stop the workflow) or lop off an entire branch of the workflow.

All depending on the exact details of the graph structure and the precise state of the task pool at the time the command was issued. The user can't really tell what will happen without a lot of graph understanding and SoD knowledge which is a shame.

The new option might be a little bit confusing (i.e, it only applies to parentless / sequential-xtriggererd tasks), but the status quo is a bit confusing, so whatever.

@hjoliver, thoughts?

dwsutherland:
> Yes, however, the flip side of the flip side is that the remove solution requires you to know enough about the graph structure

This:
> we may still need some lower-level commands or options for expert use in some situations

Also, it's not an option slapping users in the face.. It's there if needed.

> Another option might be to set the tasks

Problem here is you run into the same skip-broadcast issue; you may inadvertently trigger a cascade of downstream tasks to run/spawn..
This is one main reason why the `--no-spawn` option is useful..

oliver-sanders:
To be clear, not arguing against the use case here, and we do need some kind of a solution to enable the removal of orphan tasks (#7209).

This argument is absolutely an option, but:

* It might not work reliably:
   * E.g, It's not preserved across restarts / reloads, we can handle this of course.
  * But it's going to introduce a bunch more interactions for us to test, e.g, how does this interact with set, trigger, plain-remove, plain-restart, auto-restart, graph-changes, etc.
* It will most likely only ever be used by one person!
  * It's [really hard to explain what `cylc remove` does](https://github.com/cylc/cylc-flow/pull/7134#issuecomment-3950432581) in the first place
    * It's completely contextual, so its behaviours can't be easily predicted ahead of time.
    * In the extreme, removing just one task could cause an entire infinitely-cycling workflow to shut down!
  * It's really really hard to explain the situations in which this new option would be needed in the help text.
    * This involves explaining parentless and sequentially-xtriggered tasks.
    * As well as explaining that sequential xtriggers may be implicit, e.g, as in the case of `@wall_clock`.
  * It's really really really hard for an operator to determine whether an incoming workflow configuration change is such a situation:
    * They need to diff the before/after graphs, then identify all tasks which are parentless OR triggered by a sequential xtrigger.
  * It's really really really really hard for an operator to compose an intervention to mitigate against this.
    * They need to identify the link tasks which join the old tasks onto the new ones, note these dependencies may differ from cycle to cycle.
  * Odds are that there is only one operator in the world with the insight required to use this feature.
    * Impressive skills, but I feel there should be a less involved way to achieve this!

Hence trying to explore whether there are other options here. If we can't come up with a better idea...

In the SoS days, the remove option had spawn/no-spawn modes which worked consistently in all scenarios, whereas as it stands here, the absence of `--no-spawn` does not infer that future instances of the task will be spawned (status quo remove behaviour).

<br />

>> Another option might be to set the tasks
>
> Problem here is you run into the same skip-broadcast issue; you may inadvertently trigger a cascade of downstream tasks to run/spawn..

This plays both ways, with the remove approach, you may inadvertently trigger a cascade of downstream tasks if:
* You fail to identify [all of] the correct task(s) to remove.
* You fail to use the `--no-spawn` option.

With the set approach, you may inadvertently trigger a cascade of downstream tasks if:
* You fail to identify [all of] the correct task(s) to set.

The set approach would be a bit harder to get wrong because we can automatically generate the list of newly added tasks (and already log this for reloads), whereas we cannot automatically determine the correct task(s) to remove (which may be different from one cycle to another).

Moreover, with the remove approach, you may create the conditions for a workflow stall due to missing task outputs, whereas the set and skip approaches ensure that this cannot happen.

dwsutherland:
> Would it help to construct a dummy workflow...

A basic example would be to create a gap, i.e.  remove without spawning at one cycle and reintroduce at some future cycle..
The tests here just about do that, would just need a reintroduction (`cylc set --pre=all  . . .`)..

hjoliver:
Thanks @oliver-sanders - I'll attempt to find the time to digest your write-up in detail later today.

As a first comment, I was preferring the remove option as much simpler in terms of both implementation and command syntax (if perhaps not conceptually simpler, being a low level intervention) - you only need to target a single task.

> cylc set //<2027/{a,b,c} - (theoretical cycle selection syntax)

Your other suggestions rely on selecting a potentially large number of tasks over many cycles and modifying all of those in the DB. Maybe that's OK though, with the right syntax (I like your suggested syntax there).

dwsutherland:
Bear with me, I do come to an agreement in the end. . .

> **Caveats to the remove approach**
> 
> To conclude this lengthy debate, several caveats to the remove approach have been identified.

If you don't like it from an idealistic point of view, that's fine, however these points are off target.. 
I'm not looking for something to replace `set/skip`, or some general solution or conceptual reframing (although maybe you are), I'm looking for a solution to particular scenarios where `set/skip` are less practical (scenarios common at ESNZ).. 


> The following major points have been conceded (by acknowledgement or lack of counter):
> 
> 1. The remove operation needs to be performed on the whole subgraph **in the general case** in order to reliably remove these tasks
>    
>    * But not in your special case where all added tasks in a single chain.

This point is void.. I've explained already:
- I would **not** use it on the "whole subgraph" just the parentless beginning of a sub-graph.
- The special case is the usual case.

> 2. The remove operation may need to be performed multiple times
>    
>    * In the likely event that not all downstream tasks spawn simultaneously.

I object here too.. To create/avoid a gap I would usually just target the parentless task(s) of a cycle/glob-of-cycle point(s) if necessary. 


> 3. Remove can only mop up tasks after they have spawned (and potentially after they have submitted / started).
>    
>    * Note that your approach of pausing the workflow to avoid this will only work if there is no inter-dependence between the R1 tasks, otherwise the workflow will have to be resumed to allow them to run.

Yes, this is a post spawn intervention.. Yes, it would be nice to avoid the situation in the first place, but people make mistakes.. And it's nice that they can visualize the thing they need to act on (as opposed to the "whole subgraph" which they would have to know about.

> 4. The same tasks will have to be re-removed with any subsequent re-running of the R1 tasks.
>    
>    * Really surprised you're not concerned about this! From your description, this applies to your use case!
> 
> Additionally, this point was not conceded but was demonstrated in a worked example:

I already addressed this:

> I would `set` the first downstream R1 task(s), in this case, and `remove` the any downstream and next instance spawned...
This would have to out weigh the trade offs of using `set` here and future cycles...

That's why I'm not concerned..
 
> 5. Unintended active tasks may occur.
>    
>    * The logical result of (3) above, proven in example.

And again, I've already addressed this.. There are "unintended active tasks" with `skip/set` approaches, and I would say it's more pernicious.. I would rather things don't kick off because of fake succeeded states in the workflow..


---



> **General case solution**
> 
> > If you want to convince me to not have remove --no-spawn as an option,
> 
> I'm not necessarily trying to convince you not to have `remove --no-spawn` as an option (read back, I noted that this actually necessary, at least internally, for #7198), I'm pointing out that it's not a great solution and doesn't work for all cases, there is more likely than not a better resolution to this scenario.
> 
> 
> > you'll need to show me that there is a more practical way to handle these examples.
> 
> Great, let's talk alternatives!
> 
> Outline of the scenario:
> 
> * We have run a bunch of cycles.
> * We have added some tasks, and reloaded/restarted the workflow.
> * We have triggered an explicit selection of tasks somewhere in the workflow's history.
> * We expected only the explicitly triggered tasks to spawn/run, but some other tasks got spawned too.
> * (note that this problem is not coupled to R1 tasks, though this is a pressing use case at present)
> 
> Outline of the problem:
> 
> * When we trigger a historical task, it _may_ cause newly added tasks to spawn and either run (if dependencies are satisfied) or force a workflow-stall depending on the specifics of the graph change.
> 
> 
> While the status-quo behaviour is "logical" by SoD, it's rather illogical from a user perspective. It can certainly be argued that this is a bug (I lean towards the position that it is, more on that below).
> 
> Irrespective of whether we accept this as a bug or not, there are two possible ways for Cylc to handle this situation:
> 
> 1. Spawn these tasks.
> 2. Don't spawn these tasks.
> 
> We have to choose one of these as the default behaviour, currently, we choose (1), however, I would argue that:
> 
> * (2) is a more logical default.
> * Not spawning these tasks is the most likely user intention.
> * The inverse intervention (i.e, to spawn these tasks) is easier and less caveat-prone this way around.
> 

I'm with you..

> 
> But it would be much easier to prevent this from happening in the first place. We can always provide a mechanism for running these new-old tasks if anyone were ever to want this.

Agree, again..


> This problem is actually very similar to the issue of pre-start-cycle-point tasks in the warm start scenario (#7178), an issue which is impacting our operational workflows at the moment:
> 
> * With warm starts, all tasks are newly added, we presume that all historical task instances (tasks before the warm-start point) have completed. This is the so-called "pre-initial" logic.
> * With this scenario, it is arguable that we should consider all historical task instances (tasks before the reload-point) as completed.

There's another issue I've run into, which this will solve, and that's off ICP start tasks (when you spin up a workflow on specified tasks, cycles ahead of the ICP)..
You can accidentally end up spawning in these tasks earlier when running R1  (as this gap is meaningless after a reload/restart)..  Setting earlier instances as complete would help with this. 


> So there's two simple alternatives:
> 
> * Consider historical instances of newly-added tasks to be completed.
> * Set the start cycle point of the recurrences of newly-added tasks to the oldest active cycle point.
> 
> Both of these approaches would completely resolve this situation, removing the need for the operator to run the `cylc remove --no-spawn` command.

**Agreed!**
Also, to remove gaps and alleviate the extent to which these alternatives need to play a part:
I want isolated-graphs/inter-graph-dependencies to be a thing:
https://github.com/cylc/cylc-flow/issues/7020#issuecomment-3644811990

Along with ideas around parentless tasks https://github.com/cylc/cylc-flow/issues/7228   (which I agree that xtriggers should be like automatic python tasks separate from what they trigger)


**For now:**

This PR is a small change that can be undone once we have some of these solutions Ironed out and ready to go (which I assume would be something like `8.8` release).

dwsutherland:
I might give this a crack/go.

> We would probably need to find an alternative way to implement group-trigger to make this approach handle the use case here.

WRT this -- I feel that, post reload, if triggering a group (say `FAM1`) contains new tasks, then they should be targeted (as the group encompasses the new tasks)...  

However, you're correct, the way it's implemented now will mean off-group new tasks could also spawn in... So perhaps we could either:
- Remember the previous record before wiping it (in a temporary dict or something), and check it before spawning off shoots of this group..
- Or internally create a temporary flow or pass/ticket for the group, that gives them an exemption to run (as the previous record).

The 2nd option seems cleaner; the temporary pass/one-off-ticket implementation could by-pass the record-checking system (that determines whether a task should run or not), and the group ticket could be handed down to dependents when the corresponding outputs are produced (if they're in the group specified on the ticket).
For reload/restart this ticket would have to be recorded in the DB against the current ticket holders.

hjoliver:
I just ran into a nasty problem myself for which `remove --no-spawn` is the best solution: I added (via a colleague who vouched for it) a large number (several hundred!) of mostly independent tasks to our operational "downstream sync" workflow, to sync operational ensemble output filesets to our other site, then manually kicked off the new tasks with bulk trigger commands. A lot failed due to bad file matching globs, others succeeded even though they should have failed (because evidently rsync failure due to ssh connection limiting (for DDOS protection!) doesn't return error status), and another lot were in the wrong offset recurrence (our ensemble has different tasks in 03, 15 vs 09, 21).

The easiest way out of this total nightmare would have been to `remove` (via a small number of globbed commands) all the new instances that ran (and succeeded or failed) and all those that had spawned ahead and were waiting in future cycles - i.e. to back out of the runtime change to make it look as if it had never happened (then fix the config and trigger the corrected stuff). But I could not do that because removing the waiting instances spawns the next ones, ad infinitum.

Stopping the workflow to warm start it wasn't practical because of the spread over cycles and the sheer number of reruns that would cause. I actually restarted it with my own hacked scheduler (with david's change) to allow remove to do what I wanted - which worked like a charm 💯

dwsutherland:
Well.. I think this one is still ready to go, Oliver may have mentioned something about restart/reload  (from memory), but maybe that was for something else because this would work for your case and they wouldn't spawn back on reload/restart (only when triggered manually or automatically via outputs)

hjoliver:
Yes, `remove` makes it look as if the target task had never spawned. There's no reload/restart issue, at least for this sort of use case. After removing a task, it *should* spawn again if upstream circumstances demand that - that's the whole point of remove: (a) get rid of  the current (or past) instance; and (b) allow it run again (if needed) as if it had never spawned in the first place.

(I've posted a comment on Element)

hjoliver:
Note after doing the above, I realized I can set runahead limited waiting instances to `failed` and then remove them, and it doesn't spawn the next instance - but that's actually a bug, it *should* spawn the next instance under normal circumstances!  Is it a known bug...?  (And in any case, it is more difficult and less intuitive than `remove --no-spawn`).

[Update: #7413]

oliver-sanders:
> I just ran into a nasty problem myself for which remove --no-spawn is the best solution
>
> The easiest way out of this total nightmare would have been to remove (via a small number of globbed commands) all the new instances that ran (and succeeded or failed) 

Yes, `cylc remove` seems like the natural solution to this problem, ideally we wouldn't *need* a `--no-spawn` option for this operation to work correctly.

For the ICP/trigger scenario (which started this off), we have agreed that it would be *better* if these tasks were never spawned in the first place (avoiding the need for any subsequent remove operation) - this is an option.

<br />

> Stopping the workflow to warm start it wasn't practical

IMO: Warm start is an emergency operation, a great capability for Cylc to have as an ultimate backstop, but not something which should be considered for regular interventions.

dwsutherland:
Partially addressing PR up: https://github.com/cylc/cylc-flow/pull/7419

oliver-sanders:
Can we close this PR as superseded by https://github.com/cylc/cylc-flow/pull/7419?

hjoliver:
Ping @dwsutherland ☝️

dwsutherland:
> Can we close this PR as superseded by #7419?

Yes, will do a separate thing for the no spawn on re-outputs thing..

oliver-sanders:
I've opened https://github.com/cylc/cylc-flow/issues/7456 to handle the issue of historical instances of newly-added tasks in the general case.

I.e, for historical events applying to any cycle (not just the ICP) in a workflow of any design (i.e, one in which a wider range of the caveats of the `cylc remove --no-spawn` intervention apply).

The ideas we ran up above can be re-injected into that issue.

The `cylc remove --no-spawn` intervention provides a workaround, suitable for the ESNZ use case - for now, however, has many caveats which may make it unsuitable for other workflow designs.

Isolated graphs may partially-address this issue (assuming that the isolated graph triggers the subsequent cycles once and only once [per flow?]). However, the general case will still remain at large for:
* Cases where this happens in a non-isolated cycle (this problem isn't ICP-specific, it can originate from any cycle).
* Cases where the isolated approach isn't suitable (i.e, if you want fast startup where tasks only wait for the relevant prereqs - not the entire startup graph).
* Workflows which have not been adapted to use isolated-cycles.

PR #7142, opened 2025-12-12
Change a method name, for clarity.

hjoliver:
A tiny non-functional code-tidy change for clarity.
 - change `release_held_active_task(itask)` to `release_held_task(itask)`
 - and add to the method's docstring

This method is not just called for active (n=0) tasks.

The task proxy argument can also be a transient (a temporary task proxy not intended to be added to the task pool, but used for other purposes) representing a future (n>0, non-active) task. 

**Check List**


hjoliver:
(I should probably have based this on master, but it doesn't really matter)

MetRonnie:
It looks to me like this method is only called for active tasks, where are you seeing otherwise?

PR #7144, opened 2025-12-13
family state fix on n-window resize

### LINKED ISSUE #7115: set-n-window: family state icons not updated
The family state icons are derived from the n-window, however, they are not updated when the n-window is changed.

### Workaround:

A reload fixes the issue.

### Reproducible Example

Image

dwsutherland:
closes #7115

**Check List**

### DIFF
--- a/tests/integration/test_data_store_mgr.py
+++ b/tests/integration/test_data_store_mgr.py
@@ -664,7 +664,7 @@ async def test_absolute_graph_edges(flow, scheduler, start):
         }
 
 
-async def test_group_state_on_window_rezise(flow, scheduler, start):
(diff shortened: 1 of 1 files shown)


tests/integration/test_data_store_mgr.py
```diff
@@ -663,6 +664,118 @@ async def test_absolute_graph_edges(flow, scheduler, start):
         }
 
 
+async def test_group_state_on_window_rezise(flow, scheduler, start):
```
wxtim:
```suggestion
async def test_group_state_on_window_resize(flow, scheduler, start):
```

dwsutherland: Oops, thanks.

